PR #196, opened 2013-05-25
Provide lower keep-alive default on Android.

### LINKED ISSUE #190: Keep alive duration default and configurability
The default keep alive duration is 5 minutes and might too much in the context of an Android application.

I think we should:
- make this configurable,
- set it to a better default when running on Android.

JakeWharton:
Closes #190.

### DIFF
--- a/okhttp/src/main/java/com/squareup/okhttp/internal/Platform.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/internal/Platform.java
@@ -50,6 +50,8 @@ import javax.net.ssl.SSLSocket;
  */
 public class Platform {
   private static final Platform PLATFORM = findPlatform();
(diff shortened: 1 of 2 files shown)


okhttp/src/main/java/com/squareup/okhttp/internal/Platform.java
```diff
@@ -50,6 +50,8 @@
  */
 public class Platform {
   private static final Platform PLATFORM = findPlatform();
+  private static final int KEEP_ALIVE_MS_ANDROID = 60 * 1000; // 60 seconds.
+  private static final int KEEP_ALIVE_MS_JVM = 5 * 50 * 1000; // 5 minutes.
```
rogers-steve: typo: 5 * 60 * 1000

PR #908, opened 2014-06-10
Blow up on an unexpected request body.

swankjesse:


phileo:
Just curious to know why GET with request body is not allowed ?

Issue #943, opened 2014-06-18
SetInstanceFollowRedirects(false) isn't respected when using OkUrlFactory


MiguelLavigne:
The code below will still follow the redirect even though we're specifying otherwise.  Expected behaviour would be complete with appropriate 30\* status.

OkUrlFactory factory = new OkUrlFactory(new OkHttpClient());
HttpURLConnection connection = factory.open(request.url());
connection.setInstanceFollowRedirects(false);
... exercise the connection

PR #944, opened 2014-06-18
Fixes the do not follow redirects bug

MiguelLavigne:
The HttpEngine would not obey the set state because there was no
correlation between the HttpURLConnection.setInstanceFollowRedirects
method and the OkHttpClient.

Fixed the missing link by adding a new setFollowRedirects method to the
OkHttpClient class.  Bridged the gap for the HttpURLConnection.setInstanceFollowRedirects
by forwarding that state into the OkHttpClient.

Now the HttpEngine will always obey the OkHttpClient redirect state when
attempting the followUp phase.

Added necessary test to both OkUrlFactoryTest and CallTest.

https://github.com/square/okhttp/issues/943

### DIFF
--- a/okhttp-tests/src/test/java/com/squareup/okhttp/CallTest.java
+++ b/okhttp-tests/src/test/java/com/squareup/okhttp/CallTest.java
@@ -1345,6 +1345,25 @@ public final class CallTest {
     assertNull(recordedRequest.getHeader("User-Agent"));
   }
 
+  @Test
+  public void setFollowRedirectsFalse() throws Exception {
+    server.enqueue(new MockResponse()
+        .setResponseCode(302)
+        .addHeader("Location: /b")
+        .setBody("A"));
+    server.enqueue(new MockResponse()
+        .setBody("B"));
+    server.play();
+
+    client.setFollowRedirects(false);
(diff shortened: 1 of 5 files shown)


danielgomezrico:
Why this?

```
"If the 307 or 308 status code is received in response to a request other than GET
or HEAD, the user agent MUST NOT automatically redirect the request"
```

swankjesse:
It’s now-obsolete advice actually.

```
   The set of request methods that are safe to automatically redirect is
   no longer closed; user agents are able to make that determination
   based upon the request method semantics.  The redirect status codes
   301, 302, and 307 no longer have normative requirements on response
   payloads and user interaction.  (Section 6.4)
```

https://tools.ietf.org/html/rfc7231#appendix-B

danielgomezrico:
@swankjesse then no redirect is done by _OkHttp_ with any _http code_?

I tried with OkHttp `3.4.1` but:

- `301` / `302` change `POST` to `GET`
- `307` / `308` there's no redirect, just the body with the redirect Uri

swankjesse:
Nope, we follow redirects. Just the spec was recently revised and we could track the changes.

danielgomezrico:
@swankjesse will be cool some sample or entry on FAQ for this :)

PR #1208, opened 2014-12-16
Modify the cipher suite / TLS version selection behavior

nfuller:
Updated commit + added test based on discussion below. Please take a look.

### DIFF
--- a/okhttp-tests/src/test/java/com/squareup/okhttp/ConnectionSpecTest.java
+++ b/okhttp-tests/src/test/java/com/squareup/okhttp/ConnectionSpecTest.java
@@ -1,0 +1,172 @@
+/*
+ * Copyright (C) 2015 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.okhttp;
+
+import com.squareup.okhttp.internal.http.AuthenticatorAdapter;
+
+import org.junit.Test;
+
+import java.net.InetSocketAddress;
+import java.net.Proxy;
+import java.net.ProxySelector;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
+import javax.net.ssl.SSLSocket;
+import javax.net.ssl.SSLSocketFactory;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+
+public final class ConnectionSpecTest {
+
+  private static final Proxy PROXY = Proxy.NO_PROXY;
+  private static final InetSocketAddress INET_SOCKET_ADDRESS =
+      InetSocketAddress.createUnresolved("host", 443);
+  private static final Address HTTPS_ADDRESS = new Address(
+      INET_SOCKET_ADDRESS.getHostString(), INET_SOCKET_ADDRESS.getPort(), null, null, null, null,
+      AuthenticatorAdapter.INSTANCE, PROXY, Arrays.asList(Protocol.HTTP_1_1),
+      Arrays.asList(ConnectionSpec.MODERN_TLS), ProxySelector.getDefault());
+
+  @Test
+  public void cleartextBuilder() throws Exception {
+    ConnectionSpec cleartextSpec = new ConnectionSpec.Builder(false).build();
+    assertFalse(cleartextSpec.isTls());
+  }
+
+  @Test
+  public void tlsBuilder_explicitCiphers() throws Exception {
+    ConnectionSpec tlsSpec = new ConnectionSpec.Builder(true)
+        .cipherSuites(CipherSuite.TLS_RSA_WITH_RC4_128_MD5)
+        .tlsVersions(TlsVersion.TLS_1_2)
+        .supportsTlsExtensions(true)
+        .build();
+    assertEquals(Arrays.asList(CipherSuite.TLS_RSA_WITH_RC4_128_MD5), tlsSpec.cipherSuites());
+    assertEquals(Arrays.asList(TlsVersion.TLS_1_2), tlsSpec.tlsVersions());
+    assertTrue(tlsSpec.supportsTlsExtensions());
+  }
+
+  @Test
+  public void tlsBuilder_defaultCiphers() throws Exception {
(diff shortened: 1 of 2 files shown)


okhttp/src/main/java/com/squareup/okhttp/ConnectionSpec.java
```diff
@@ -129,6 +129,9 @@ void apply(SSLSocket sslSocket, Route route) {
       if (socketSupportsFallbackScsv) {
         // Add the SCSV cipher to the set of enabled ciphers iff it is supported.
         String[] oldEnabledCipherSuites = cipherSuitesToEnable;
```
swankjesse: might read better as single assignment ?
```
String[] oldEnabledCipherSuites = cipherSuitesToEnable != null
    ? cipherSuitesToEnable
    : sslSocket.getEnabledCipherSuites();
```

okhttp/src/main/java/com/squareup/okhttp/ConnectionSpec.java
```diff
@@ -146,16 +152,23 @@ void apply(SSLSocket sslSocket, Route route) {
...
+    String[] cipherSuitesToEnable = null;
+    if (cipherSuites != null) {
+      String[] cipherSuitesToSelectFrom = sslSocket.getEnabledCipherSuites();
+      List<String> cipherSuitesToEnableList =
+          Util.intersect(cipherSuites, cipherSuitesToSelectFrom);
+      cipherSuitesToEnable = cipherSuitesToEnableList.toArray(
+          new String[cipherSuitesToEnableList.size()]);
+    }
+
+    String[] protocolsToSelectFrom = sslSocket.getEnabledProtocols();
```
swankjesse: Hang on, what if `tlsVersions == null` ?

okhttp/src/main/java/com/squareup/okhttp/ConnectionSpec.java
```diff
@@ -146,16 +152,23 @@ void apply(SSLSocket sslSocket, Route route) {
...
+   * enabled by {@code sslSocket}.
    */
   private ConnectionSpec supportedSpec(SSLSocket sslSocket) {
-    List<String> supportedCipherSuites =
-        Util.intersect(cipherSuites, sslSocket.getSupportedCipherSuites());
-    List<String> supportedTlsVersions =
-        Util.intersect(tlsVersions, sslSocket.getSupportedProtocols());
+    String[] cipherSuitesToEnable = null;
+    if (cipherSuites != null) {
+      String[] cipherSuitesToSelectFrom = sslSocket.getEnabledCipherSuites();
```
swankjesse: Would it be simpler to offer a variant of `Util.intersect` that accepts and returns arrays?

okhttp/src/main/java/com/squareup/okhttp/ConnectionSpec.java
```diff
@@ -186,7 +199,9 @@ private ConnectionSpec supportedSpec(SSLSocket sslSocket) {
 
   @Override public String toString() {
     if (tls) {
-      return "ConnectionSpec(cipherSuites=" + cipherSuites()
+      List<CipherSuite> cipherSuites = cipherSuites();
+      String cipherSuitesString = cipherSuites == null ? "[use default]" : cipherSuites.toString();
+      return "ConnectionSpec(cipherSuites=" + cipherSuitesString
           + ", tlsVersions=" + tlsVersions()
```
swankjesse: Isn't this nullable also?

swankjesse: ... ahhh , it isn't. These are asymmetric. That's fine.

okhttp/src/main/java/com/squareup/okhttp/ConnectionSpec.java
swankjesse: Nevermind. These are asymmetric.

okhttp/src/main/java/com/squareup/okhttp/ConnectionSpec.java
```diff
@@ -110,11 +114,7 @@ public boolean supportsTlsExtensions() {
 
   /** Applies this spec to {@code sslSocket} for {@code route}. */
   void apply(SSLSocket sslSocket, Route route) {
-    ConnectionSpec specToApply = supportedSpec;
```
swankjesse: Caching dropped deliberately, because the enabled set might change? That's fair.

In a follow up I'll be tempted to move the SCSV stuff so we only end up rebuilding this array 1x.

okhttp/src/main/java/com/squareup/okhttp/ConnectionSpec.java
```diff
@@ -89,6 +90,9 @@ public boolean isTls() {
   }
 
   public List<CipherSuite> cipherSuites() {
```
swankjesse: Document that this returns null if the socket's enabled set will be used?

okhttp-tests/src/test/java/com/squareup/okhttp/ConnectionSpecTest.java
```diff
@@ -0,0 +1,172 @@
...
+
+    Set<String> expectedCipherSet = createSet(CipherSuite.TLS_RSA_WITH_RC4_128_MD5.javaName);
+    if (Arrays.asList(socket.getSupportedCipherSuites()).contains("TLS_FALLBACK_SCSV")) {
+      expectedCipherSet.add("TLS_FALLBACK_SCSV");
+    }
+    assertEquals(expectedCipherSet, expectedCipherSet);
+  }
+
+  private static Set<String> createSet(String... values) {
+    return new HashSet<String>(Arrays.asList(values));
```
swankjesse: nit: please use LinkedHashSet. Never use naked HashSet. The negligible performance gain is not worth the determinism lost.

swankjesse:
LGTM. Nice.

mfietz:
I'll just ask this here because I'm not sure if this is an actual issue or this is working as intended and I am missing something.

Why do you choose the cipher suites and TLS versions by intersecting with the enabled ones, not the supported ones?

The problem: I have a user that wants to connect to a server only supporting TLSv1.1 and TLSv1.2 on an Android 4.4 device. Android 4.1 to 4.4 support TLS 1.1 and 1.2, but don't have in enabled by default.
At the moment, I have to work around OkHttp by setting a custom SSLSocketFactoring wrapping the actual one and enabling the higher TLS versions on newly created sockets manually before handing them to OkHttp.

nfuller:
@mfietz Android "supports" various legacy ciphers but disables them. The list of "supported but disabled" may grow in future. We don't want OkHttp's TLS fallback enabling them (e.g. turning on SSLv3 today) in situations where the socket factory "knows better".

Do you have the option of using Google Play Services "dynamic security provider"? That _should_ give you more up-to-date socket / socketfactory behavior on older releases of Android. https://developer.android.com/training/articles/security-gms-provider.html

mfietz:
@nfuller I see. Sounds reasonable, the issue is probably more on Google's side... Thanks for the dynamic security provider suggestion, I'll definitely have a look into it. But we have a lot of users with custom roms, so we'll probably keep the (pretty silly) workaround.
Perhaps this could be added to the HTTPs wiki page. I'm probably not the last dev that wonders why TLS 1.1 and 1.2 won't work with 4.x devices

PR #1224, opened 2014-12-23
Use native Okio copy for retrying.

JakeWharton:
For your consideration and review, because this has a slight behavioral problem...

This method is called from two places in two different ways:
 * `SpdyTransport.writeRequestBody` calls it with an instance of `SpdyDataSink`. This class is a `Sink` implementation which adds framing around a SPDY `FrameWriter` who ultimately delegates to a `BufferedSink` around the `Socket`.
 * `HttpConnection.writeRequestBody` on the other hand passes a `BufferedSink` which is directly wrapped around the `Socket`.

The distinction of the former being a `Sink` and the latter being a `BufferedSink` means that `Okio.buffer` is going to behave differently. When we ultimately call `.emit()` to ensure we're not leaking segments, one writes directly to a `Socket` whereas the other hands off to another `Buffer` and lets emission to the socket happen at the discretion of that sink.

So, are we fine with that behavior? Do we revert Okio's behavior?

### DIFF
--- a/okhttp/src/main/java/com/squareup/okhttp/internal/http/RetryableSink.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/internal/http/RetryableSink.java
@@ -72,8 +72,9 @@ public final class RetryableSink implements Sink {
   }
 
   public void writeToSocket(Sink socketOut) throws IOException {
-    // Clone the content; otherwise we won't have data to retry.
-    Buffer buffer = content.clone();
+    // Copy the content; otherwise we won't have data to retry.
+    Buffer buffer = new Buffer();
+    content.copyTo(buffer, 0, content.size());
     socketOut.write(buffer, buffer.size());
   }
 }


JakeWharton:
LGTY now that we're Okio 1.2.0-ing all up in this piece?

okhttp/src/main/java/com/squareup/okhttp/internal/http/RetryableSink.java
```diff
@@ -72,8 +74,8 @@ public long contentLength() throws IOException {
   }
 
   public void writeToSocket(Sink socketOut) throws IOException {
-    // Clone the content; otherwise we won't have data to retry.
-    Buffer buffer = content.clone();
-    socketOut.write(buffer, buffer.size());
+    BufferedSink bufferedOut = Okio.buffer(socketOut);
```
swankjesse: I don't think the BufferedSource here helps?

```
Buffer buffer = new Buffer();
content.copyTo(buffer, 0, content.size());
socketOut.write(buffer, buffer.size());
```

swankjesse:
I don't really get what the motivation is here. Is `clone()` slower than `copyTo` or something?

JakeWharton:
I asked you that very same question a few weeks ago but I think our conversation got derailed.

swankjesse:
Cool. As it stands now, I don't think this changes behavior. LGTM.

Issue #1043, opened 2014-09-04
Don't recover if a non-GET request fails


ValleZ:
Currently it can silently retry requests, but now it's forbidden:

http://tools.ietf.org/html/rfc7230

A user agent MUST NOT automatically retry a request with a non-
   idempotent method unless it has some means to know that the request
   semantics are actually idempotent, regardless of the method, or some
   means to detect that the original request was never applied.

related to https://github.com/square/okhttp/issues/258

PR #1259, opened 2014-12-30
New API to disable retry for an OkHttpClient.

swankjesse:
Most applications won't want to disable retry globally. Instead, use
clone() to get an OkHttpClient for a specific, non-idempotent request,
then configure that client with the setting.

Closes https://github.com/square/okhttp/issues/1043

### DIFF
--- a/okhttp-tests/src/test/java/com/squareup/okhttp/CallTest.java
+++ b/okhttp-tests/src/test/java/com/squareup/okhttp/CallTest.java
@@ -704,6 +704,34 @@ public final class CallTest {
     callback.await(request.url()).assertHandshake();
   }
 
+  @Test public void recoverWhenRetryOnConnectionFailureIsTrue() throws Exception {
+    server.enqueue(new MockResponse().setSocketPolicy(SocketPolicy.DISCONNECT_AT_START));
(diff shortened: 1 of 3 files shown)


1zaman:
Why not provide an option for disabling retries based on the potential non-idempotency of the HTTP method itself? I am sure that the overwhelming majority of use-cases for disabling retries are for this. Absent this option, I ended up writing my own [delegator class][] to implement this.

[delegator class]: https://gist.github.com/1zaman/94892c192ca318e2dd6a

afinas-em:
OkHttpClient okHttpClient= null;
        okHttpClient = new OkHttpClient.Builder()
                .sslSocketFactory(new TLSSocketFactory(),trustManager)
                .connectTimeout(2, TimeUnit.MINUTES)
                .readTimeout(2, TimeUnit.MINUTES)
                .writeTimeout(2, TimeUnit.MINUTES)
                //.sslSocketFactory(sslSocketFactory, trustManager)
                .followRedirects(false)
                .followSslRedirects(false)
                .retryOnConnectionFailure(false)
                .cache(null)//new Cache(sContext.getCacheDir(),10*1024*1024)
                .build();

PR #1268, opened 2015-01-02
Always include OkHttp in thread names.

swankjesse:


okhttp/src/main/java/com/squareup/okhttp/internal/ws/WebSocketCall.java
```diff
@@ -178,10 +178,10 @@ private void createWebSocket(Response response, WebSocketListener listener)
     BufferedSink sink = Okio.buffer(Okio.sink(socket));
 
     final RealWebSocket webSocket =
-        new ConnectionWebSocket(connection, source, sink, random, listener);
+        new ConnectionWebSocket(response, connection, source, sink, random, listener);
 
     // Start a dedicated thread for reading the web socket.
-    new Thread(new NamedRunnable("WebSocketReader " + request.urlString()) {
```
swankjesse: This was _potentially_ problematic for URLs containing `%s`

swankjesse: or more plausibly, %20s

PR #1270, opened 2015-01-02
Update MockWebServer to use Headers.

JakeWharton:


JakeWharton:
Public API changes to `MockResponse` and `RecordedRequest` to expose `Headers` rather than hide as implementation detail will come next.

swankjesse:
LGTM

Rats, the tests are hanging!

JakeWharton:
Oh, snap. These changes don't look like they could have broken anything either...

JakeWharton:
I can get `com.squareup.okhttp.mockwebserver.MockWebServerTest.expect100ContinueWithNoBody` to hang via CLI but it does not hang when run in IDEA.

JakeWharton:
I can reproduce this locally but the thread dump isn't clear to me what the problem is.

JakeWharton:
I'm at a loss for why this still hangs. The thread dumps clearly indicate a SPDY client is waiting to read but that MWS isn't doing anything.

mockwebserver/src/main/java/com/squareup/okhttp/mockwebserver/MockResponse.java
```diff
@@ -78,7 +78,17 @@ public MockResponse setStatus(String status) {
...
+    Headers headers = this.headers.build();
+    int size = headers.size();
+    List<String> headerList = new ArrayList<>(size);
+    for (int i = 0; i < size; i++) {
+      headerList.add(headers.name(i) + ": " + headers.value(i));
+    }
+    return headerList;
+  }
+
+  Headers getNewHeaders() {
```
JakeWharton: This will eventually replace `getHeaders()` but that requires a good bit of refactoring which will be in a follow-up.

mockwebserver/src/main/java/com/squareup/okhttp/mockwebserver/RecordedRequest.java
```diff
@@ -73,6 +74,15 @@ public String getPath() {
...
   public List<String> getHeaders() {
+    int size = headers.size();
+    List<String> headerList = new ArrayList<>(size);
+    for (int i = 0; i < size; i++) {
+      headerList.add(headers.name(i) + ": " + headers.value(i));
+    }
+    return headerList;
+  }
+
+  Headers getNewHeaders() {
```
JakeWharton: Ditto

okhttp-tests/src/test/java/com/squareup/okhttp/internal/http/HttpOverHttp20Draft16Test.java
```diff
@@ -32,9 +33,12 @@ public HttpOverHttp20Draft16Test() {
   }
 
   @Test public void serverSendsPushPromise_GET() throws Exception {
-    MockResponse response = new MockResponse().setBody("ABCDE").setStatus("HTTP/1.1 200 Sweet")
-        .withPush(new PushPromise("GET", "/foo/bar", Arrays.asList("foo: bar"),
-            new MockResponse().setBody("bar").setStatus("HTTP/1.1 200 Sweet")));
+    PushPromise pushPromise = new PushPromise("GET", "/foo/bar", Headers.of("foo", "bar"),
```
adriancole: looks nicer

adriancole:
confirmed.. after updating to latest JRE (needed as alpn is pinned to recent ssl classes), master builds fine, but this branch hangs on `CacheAdapterTest`. Looking into it.

adriancole:
Maybe this doesn't fail in idea because the bootclasspath doesn't have the alpn library set?

JakeWharton:
When I add the ALPN jar to the test config in the IDE it still passes.

JakeWharton:
I found it!!!

JakeWharton:
Filed at https://github.com/square/okhttp/issues/1294

mockwebserver/src/main/java/com/squareup/okhttp/mockwebserver/MockWebServer.java
```diff
@@ -817,7 +806,7 @@ private RecordedRequest readRequest(SpdyStream stream) throws IOException {
         } else if (name.equals(Header.VERSION)) {
           version = value;
         } else {
-          httpHeaders.add(name.utf8() + ": " + value);
+          httpHeaders.add(name.utf8(), value);
```
JakeWharton: This was the bug. `name` can start with a colon causing the line-based parsing to blow up.

JakeWharton:
Had to make two tiny changes:

 * Expose a means of replacing all headers. This allows tests to use `Headers.addLenient` to test invalid cached values. `MockResponse` no longer allows creating invalid headers.
 * Change `truncateViolently` to not assume the headers `List<String>` was mutable.

JakeWharton:
(What an ordeal this was)

mockwebserver/src/main/java/com/squareup/okhttp/mockwebserver/MockWebServer.java
adriancole: ahah!

PR #1271, opened 2015-01-02
Use TlsVersion enum for recorded requests.

JakeWharton:


mockwebserver/src/main/java/com/squareup/okhttp/mockwebserver/RecordedRequest.java
```diff
@@ -141,12 +143,14 @@ public int getSequenceNumber() {
     return sequenceNumber;
   }
 
-  /**
-   * Returns the connection's SSL protocol like {@code TLSv1}, {@code SSLv3},
-   * {@code NONE} or null if the connection doesn't use SSL.
-   */
+  /** @deprecated Use {@link #getTlsVersion()}. */
```
JakeWharton: Not sure how much we care about this. I'm going to be making other breaking changes in this API...

okhttp/src/main/java/com/squareup/okhttp/OkHttpClient.java
```diff
@@ -120,6 +120,10 @@
         connection.connectAndSetOwner(client, owner, request);
       }
 
+      @Override public TlsVersion tlsVersionForJavaName(String name) {
+        return TlsVersion.forJavaName(name);
```
swankjesse: I think we want to make `forJavaName` public both for TlsVersion and also for `CipherSuite`. Turns out there's lots of code that needs to convert between string and enum.

PR #1276, opened 2015-01-03
Prefer JUnit rules for temporary folders, MockWebServer.

JakeWharton:


JakeWharton:
Two of these tests were polluting the temporary folder of the host machine by not deleting their files.

Issue #1096, opened 2014-10-22
Explore Mock OkHttpClient Testing Implementation


JakeWharton:
An `OkHttpClient` instance which can be scripted similar to `MockWebServer`. The advantages here are explicit control over the objects passed back to the caller and much lower overhead.

JakeWharton:
https://github.com/square/retrofit/blob/763fe1639e3348480214873323caacd00cbd7fa6/retrofit/src/test/java/retrofit/MockClient.java

gmazzo:
Hi guys, based on @swankjesse comment I've created this repo: https://github.com/gmazzo/okhttp-client-mock. I can make a PR to include it in your mock API if you like it.

lucasoares:
Hello guys.. Any updates here? I'd love to see this feature as native on okhttpclient.

yschimke:
Closing out as https://github.com/gmazzo/okhttp-client-mock is probably the right way to go here.

gmazzo:
> Closing out as https://github.com/gmazzo/okhttp-client-mock is probably the right way to go here.

So, @yschimke are you interested in making it official?

yschimke:
I think official is a link here https://square.github.io/okhttp/works_with_okhttp/ and you are already there.

gmazzo:
Ohh great then, I didn't note it before

PR #1277, opened 2015-01-03
Make MockWebServer directly usable as a rule.

JakeWharton:
This eliminates the awkward sometimes-calls to get() in order to access APIs which were not exposed directly on the rule.


JakeWharton:
In a case of the very odd... Travis CI passed this on JDK 7 but not JDK 8. This is because the 'Sample Server' example uses MockWebServer but does not have a dependency on JUnit.

I guess the advantage to a separate class was that the JUnit dependency was optional.

What do we feel here? Is MockWebServer testing only and have a JUnit dependency or should it be able to stand on its own without it?

swankjesse:
Making MockWebServer depend on JUnit is fine by me.

Here's one option for MockWebServer In 2015:

 * New `okhttp-testing` module that uses interceptor to make setting up scenarios and confirming interactions possible. ([issue 1096](https://github.com/square/okhttp/issues/1096)). Probably with a new API.
 * MockWebServer degraded to be test support for OkHttp's own implementation.
 * Possibly permit MockWebServer to also implement the `okhttp-testing` API as a real server. For example, for unit testing vs. fuctional testing.

mockwebserver/src/main/java/com/squareup/okhttp/mockwebserver/MockWebServer.java
```diff
@@ -122,17 +132,17 @@ public void setServerSocketFactory(ServerSocketFactory serverSocketFactory) {
   }
 
   public int getPort() {
-    if (port == -1) throw new IllegalStateException("Call play() before getPort()");
+    startOrThrow();
```
swankjesse: awww, does it really have to be this automatic? I liked the explicit lifecycle

swankjesse:
LGTM

mockwebserver/src/main/java/com/squareup/okhttp/mockwebserver/MockWebServer.java
JakeWharton: The explicit life cycle is maintained for non-rule usage. I think we can
kill this though.

swankjesse: Yeah, let's go 100% in on the rule. The `play`/`shutdown` names aren't necessary. That comes from an earlier time, when you needed to provide all the responses before you called `play`, and it was intended to offer a record/replay experience.

JakeWharton:
Having second thoughts. Low value prop here. This can wait until we have the scriptable mock interceptor.

PR #1280, opened 2015-01-03
Teach MockWebServer to speak WebSockets.

JakeWharton:


mockwebserver/src/main/java/com/squareup/okhttp/mockwebserver/MockWebServer.java
```diff
@@ -565,7 +585,83 @@ private RecordedRequest readRequest(Socket socket, InputStream in, OutputStream
...
+        .message(response.getStatus().split(" ", 3)[2])
+        .request(fancyRequest)
+        .protocol(Protocol.HTTP_1_1);
+    List<String> responseHeaders = response.getHeaders();
+    Headers.Builder fancyResponseHeaders = new Headers.Builder();
+    for (int i = 0, size = responseHeaders.size(); i < size; i++) {
+      fancyRequestHeaders.add(responseHeaders.get(i));
+    }
+    fancyRequestBuilder.headers(fancyResponseHeaders.build());
+    final Response fancyResponse = fancyResponseBuilder.build();
```
JakeWharton: All the above junk gets a lot nicer once #1270 lands since the loops disappear.

mockwebserver/src/main/java/com/squareup/okhttp/mockwebserver/MockWebServer.java
```diff
@@ -565,7 +585,83 @@ private RecordedRequest readRequest(Socket socket, InputStream in, OutputStream
         new Buffer().write(requestBody.toByteArray()), sequenceNumber, socket);
   }
 
-  private void writeResponse(Socket socket, OutputStream out, MockResponse response)
+  private void handleWebSocketUpgrade(Socket socket, InputStream in, OutputStream out,
+      RecordedRequest request, MockResponse response) throws IOException {
+    response.setHeader("Connection", "Upgrade");
```
swankjesse: I think we should do this work in `MockResponse.withWebSocketUpgrade`. That way we can test scenarios that later remove those headers.

mockwebserver/src/main/java/com/squareup/okhttp/mockwebserver/MockWebServer.java
```diff
@@ -565,7 +585,83 @@ private RecordedRequest readRequest(Socket socket, InputStream in, OutputStream
...
+      @Override public void run() {
+        try {
+          listener.onOpen(webSocket, fancyRequest, fancyResponse);
+        } catch (IOException e) {
+          // TODO try to write close frame?
+          listener.onFailure(e);
+          connectionClose.countDown();
+        }
+      }
+    }, "Vivisection WebSocket Writer").start();
```
swankjesse: typo here.

mockwebserver/src/main/java/com/squareup/okhttp/mockwebserver/MockWebServer.java
```diff
@@ -565,7 +585,83 @@ private RecordedRequest readRequest(Socket socket, InputStream in, OutputStream
...
+
+    // Use this thread to continuously read messages.
+    while (webSocket.readMessage()) {
+    }
+
+    // Even if messages are no longer being read we need to wait for the connection close signal.
```
swankjesse: Nic.

mockwebserver/src/main/java/com/squareup/okhttp/mockwebserver/MockWebServer.java
```diff
@@ -565,7 +585,83 @@ private RecordedRequest readRequest(Socket socket, InputStream in, OutputStream
...
+    final Response fancyResponse = fancyResponseBuilder.build();
+
+    // The callback might act synchronously. Give it its own thread.
+    new Thread(new Runnable() {
+      @Override public void run() {
+        try {
+          listener.onOpen(webSocket, fancyRequest, fancyResponse);
+        } catch (IOException e) {
+          // TODO try to write close frame?
+          listener.onFailure(e);
```
swankjesse: Should we notify the listener of its own failures? I think in this case we might want to do what Callback does, which is just give up and log it.

okhttp-tests/src/test/java/com/squareup/okhttp/internal/ws/WebSocketRecorder.java
```diff
@@ -78,29 +78,54 @@ public void setNextMessageDelegate(MessageDelegate delegate) {
   public void assertTextMessage(String payload) {
     Message message = new Message(TEXT);
     message.buffer.writeUtf8(payload);
-    assertEquals(message, events.pollFirst());
+    try {
+      assertEquals(message, events.poll(10, TimeUnit.SECONDS));
```
swankjesse: Maybe a method, nextEvent() that does the try/poll/catch nonsense?

swankjesse:
LGTM

I really like how the tests work.

mockwebserver/src/main/java/com/squareup/okhttp/mockwebserver/MockWebServer.java
JakeWharton: copy/pasta!

PR #1283, opened 2015-01-04
Make cache-control remember its header value.

### LINKED ISSUE #1255: CacheControl should remember its toString form
Yo dog.

JakeWharton:
Closes #1255.

### DIFF
--- a/okhttp-tests/src/test/java/com/squareup/okhttp/CacheControlTest.java
+++ b/okhttp-tests/src/test/java/com/squareup/okhttp/CacheControlTest.java
@@ -20,6 +20,8 @@ import org.junit.Test;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertSame;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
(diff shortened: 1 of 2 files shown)


okhttp/src/main/java/com/squareup/okhttp/CacheControl.java
```diff
@@ -41,9 +41,11 @@
   private final boolean onlyIfCached;
   private final boolean noTransform;
 
+  volatile String headerValue; // Lazily computed, if absent.
```
swankjesse: Is volatile necessary here? I fear this might be a pessimization with `volatile`, since we're invalidating CPU caches. (And we can implement this correctly without it.)

okhttp-tests/src/test/java/com/squareup/okhttp/CacheControlTest.java
```diff
@@ -99,6 +101,44 @@
...
+  @SuppressWarnings("RedundantStringConstructorCall") // Testing instance equality.
+  @Test public void parseCacheControlHeaderValueIsRetained() {
+    String value = new String("max-age=12");
+    Headers headers = Headers.of("Cache-Control", value);
+    CacheControl cacheControl = CacheControl.parse(headers);
+    assertSame(value, cacheControl.toString());
+  }
+
+  @Test public void parseCacheControlHeaderValueInvalidatedByPragma() {
+    Headers headers = Headers.of("Cache-Control", "max-age=12", "Pragma", "must-revalidate");
```
swankjesse: Test two cache control headers?

okhttp/src/main/java/com/squareup/okhttp/CacheControl.java
```diff
@@ -149,73 +152,93 @@ public static CacheControl parse(Headers headers) {
...
-            pos = HeaderParser.skipUntil(string, pos, ",;");
-            parameter = string.substring(parameterStart, pos).trim();
+            pos = HeaderParser.skipUntil(value, pos, ",;");
+            parameter = value.substring(parameterStart, pos).trim();
           }
         }
 
         if ("no-cache".equalsIgnoreCase(directive)) {
           noCache = true;
+          if (!isCacheControl) headerValue = null;
```
swankjesse: Might be simpler to do the assignment at the end, if there's exacts one Cache-Control header and nothing else?

okhttp/src/main/java/com/squareup/okhttp/CacheControl.java
```diff
@@ -149,73 +152,93 @@ public static CacheControl parse(Headers headers) {
...
       }
     }
 
     return new CacheControl(noCache, noStore, maxAgeSeconds, sMaxAgeSeconds, isPublic,
-        mustRevalidate, maxStaleSeconds, minFreshSeconds, onlyIfCached, noTransform);
+        mustRevalidate, maxStaleSeconds, minFreshSeconds, onlyIfCached, noTransform, headerValue);
   }
 
   @Override public String toString() {
+    String result = headerValue;
```
swankjesse: Nice. Because you read the field exactly once, you don't need volatile for correctness.

okhttp-tests/src/test/java/com/squareup/okhttp/CacheControlTest.java
JakeWharton: I didn't even realize that was a thing!

JakeWharton:
I updated this to be a bit less clever. The presence of any "Pragma" header or second "Cache-Control" header invalidates the cached value.

swankjesse:
LGTM

PR #1284, opened 2015-01-04
Name threads based on service, as it helps narrow context.

adriancole:
This will likely need revision when we do alternate services. For now, it helps correlate okhttp client threads with MWS.

### DIFF
--- a/mockwebserver/src/main/java/com/squareup/okhttp/mockwebserver/MockWebServer.java
+++ b/mockwebserver/src/main/java/com/squareup/okhttp/mockwebserver/MockWebServer.java
@@ -123,6 +123,7 @@ public final class MockWebServer {
   private boolean protocolNegotiationEnabled = true;
   private List<Protocol> protocols
       = Util.immutableList(Protocol.HTTP_2, Protocol.SPDY_3, Protocol.HTTP_1_1);
+  private String service;
 
   public void setServerSocketFactory(ServerSocketFactory serverSocketFactory) {
     if (serverSocketFactory == null) throw new IllegalArgumentException("null serverSocketFactory");
@@ -183,6 +184,7 @@ public final class MockWebServer {
    */
   public void setProtocolNegotiationEnabled(boolean protocolNegotiationEnabled) {
     this.protocolNegotiationEnabled = protocolNegotiationEnabled;
+    this.protocols = Util.immutableList(Protocol.HTTP_1_1);
   }
 
   /**
@@ -275,14 +277,18 @@ public final class MockWebServer {
    */
   public void play(int port) throws IOException {
     if (executor != null) throw new IllegalStateException("play() already called");
-    executor = Executors.newCachedThreadPool(Util.threadFactory("MockWebServer", false));
     inetAddress = InetAddress.getByName(null);
     serverSocket = serverSocketFactory.createServerSocket();
     serverSocket.setReuseAddress(port != 0); // Reuse the port if the port number was specified.
     serverSocket.bind(new InetSocketAddress(inetAddress, port), 50);
 
     this.port = serverSocket.getLocalPort();
-    executor.execute(new NamedRunnable("MockWebServer %s", this.port) {
+    this.service = sslSocketFactory != null ? protocols + ":" : "" + this.port;
+    executor = Executors.newCachedThreadPool(
+        Util.threadFactory("MockWebServer " + service + " idle", false)
+    );
+
+    executor.execute(new NamedRunnable("MockWebServer %s accept", service) {
       @Override protected void execute() {
         try {
           logger.info(MockWebServer.this + " starting to accept connections");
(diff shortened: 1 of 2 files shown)


mockwebserver/src/main/java/com/squareup/okhttp/mockwebserver/MockWebServer.java
```diff
@@ -275,14 +277,18 @@ public void play() throws IOException {
...
     serverSocket = serverSocketFactory.createServerSocket();
     serverSocket.setReuseAddress(port != 0); // Reuse the port if the port number was specified.
     serverSocket.bind(new InetSocketAddress(inetAddress, port), 50);
 
     this.port = serverSocket.getLocalPort();
-    executor.execute(new NamedRunnable("MockWebServer %s", this.port) {
+    this.service = sslSocketFactory != null ? protocols + ":" : "" + this.port;
+    executor = Executors.newCachedThreadPool(
+        Util.threadFactory("MockWebServer " + service + " idle", false)
+    );
```
swankjesse: nit: we don't typically wrap this line

mockwebserver/src/main/java/com/squareup/okhttp/mockwebserver/MockWebServer.java
```diff
@@ -275,14 +277,18 @@ public void play() throws IOException {
...
     if (executor != null) throw new IllegalStateException("play() already called");
-    executor = Executors.newCachedThreadPool(Util.threadFactory("MockWebServer", false));
     inetAddress = InetAddress.getByName(null);
     serverSocket = serverSocketFactory.createServerSocket();
     serverSocket.setReuseAddress(port != 0); // Reuse the port if the port number was specified.
     serverSocket.bind(new InetSocketAddress(inetAddress, port), 50);
 
     this.port = serverSocket.getLocalPort();
-    executor.execute(new NamedRunnable("MockWebServer %s", this.port) {
+    this.service = sslSocketFactory != null ? protocols + ":" : "" + this.port;
```
swankjesse: add parens? the `":"` string makes this difficult to parse visually!

swankjesse: I was parsing reading this as

```
  sslSocketFactory != null ? (protocols + ":") : ("" + this.port);
```

instead of
```
 (sslSocketFactory != null ? protocols + ":" : "") + this.port;
```

mockwebserver/src/main/java/com/squareup/okhttp/mockwebserver/MockWebServer.java
```diff
@@ -341,7 +347,7 @@ public void shutdown() throws IOException {
   }
 
   private void serveConnection(final Socket raw) {
-    executor.execute(new NamedRunnable("MockWebServer %s", raw.getRemoteSocketAddress()) {
```
swankjesse: Why drop the remote address from the thread name?

okhttp/src/main/java/com/squareup/okhttp/internal/spdy/SpdyConnection.java
```diff
@@ -147,14 +147,14 @@ private SpdyConnection(Builder builder) throws IOException {
       okHttpSettings.set(Settings.INITIAL_WINDOW_SIZE, 0, OKHTTP_CLIENT_WINDOW_SIZE);
     }
 
-    hostName = builder.hostName;
+    service = protocol + ":" + builder.hostName + ":" + builder.socket.getPort();
```
swankjesse: standard to use a colon between protocol and host? `http/1.1:twitter.com:443`  ? I'd expect different formatting; perhaps `twitter.com:443 (http/1.1)` ?

okhttp/src/main/java/com/squareup/okhttp/internal/spdy/SpdyConnection.java
```diff
@@ -79,7 +79,7 @@
    */
   private final IncomingStreamHandler handler;
   private final Map<Integer, SpdyStream> streams = new HashMap<>();
-  private final String hostName;
+  private final String service;
```
swankjesse: service isn't the name I'd expect for this. Maybe call it `name`?

okhttp/src/main/java/com/squareup/okhttp/internal/spdy/SpdyConnection.java
adriancole: Oh, in alternate services, it is ( alpn, host, port )

Maybe I shouldn't tie to alt svc, but the bit I was troubleshooting when
looking at this had to do w/ protocol nego.

Other ideas?

swankjesse: I think `host:port (protocol)` is obvious. Is the protocol strictly necessary?

adriancole: Probably not. The stack trace can answer that question. I will lighten this
change.

adriancole:
in the current form, this has questionable value. I'll defer this to when we add Alt-Svc support.

PR #1286, opened 2015-01-05
Change MockResponse body to Source.

JakeWharton:


JakeWharton:
Sigh. More hanging tests.

mockwebserver/src/main/java/com/squareup/okhttp/mockwebserver/MockResponse.java
```diff
@@ -130,15 +128,20 @@ public MockResponse removeHeader(String name) {
...
+    return body;
   }
 
-  Source getBodySource() {
-    return bodyStream != null ? Okio.source(bodyStream) : getBody();
+  /**
+   * The number of bytes in {@linkplain #getBody() the body}. Note: this is not synonymous with
+   * {@code Content-Length} (e.g., in the case of chunked body).
+   */
+  public long getBodySize() {
```
swankjesse: Looks like we only ever compare this to 0. Perhaps instead we could just have a null response from body source? The fact that this doesn't line up with `Content-Length` is ugly.

mockwebserver/src/main/java/com/squareup/okhttp/mockwebserver/MockWebServer.java
```diff
@@ -855,15 +855,16 @@ private void writeResponse(SpdyStream stream, MockResponse response) throws IOEx
...
-      if (body == null) body = new Buffer();
-      boolean closeStreamAfterHeaders = body.size() > 0 || !response.getPushPromises().isEmpty();
+
+      Source body = response.getBody();
+      long bodySize = response.getBodySize();
+      boolean closeStreamAfterHeaders = bodySize > 0 || !response.getPushPromises().isEmpty();
       stream.reply(spdyHeaders, closeStreamAfterHeaders);
       pushPromises(stream, response.getPushPromises());
-      if (body.size() > 0) {
+      if (bodySize > 0) {
```
swankjesse: `if (body != null) ` ?

swankjesse:
LGTM

JakeWharton:
I went full-buffer which means you can stream anymore. Do we care?

The problem is that some tests re-enqueue the same `MockResponse` instance for multiple requests. This worked if and only if you were using a `Buffer`/`byte[]` but not if you were specifying an `InputStream` (which nobody was internally). Do we need to support one-shot streaming responses?

swankjesse:
Let's kill one-shot streaming responses. They were there for testing large payloads.

PR #1289, opened 2015-01-05
Clean up time-based APIs for throttle and delay.

JakeWharton:


mockwebserver/src/main/java/com/squareup/okhttp/mockwebserver/MockResponse.java
```diff
@@ -31,13 +31,14 @@
 
   private Buffer body;
 
-  private int throttleBytesPerPeriod = Integer.MAX_VALUE;
-  private long throttlePeriod = 1;
-  private TimeUnit throttleUnit = TimeUnit.SECONDS;
+  private long throttleBytesPerPeriod = Long.MAX_VALUE;
```
JakeWharton: lol `int`? What is this, `java.io` streams?

swankjesse:
LGTM

PR #1297, opened 2015-01-06
Use 'Headers' in recorded response.

JakeWharton:


JakeWharton:
Please review closely. The tests should be verifying behavior but we want to make sure semantics don't change as this no longer does any regex-based validation. I'm confident it doesn't.

A guide:
`assertTrue`/`assertContains` --> `assertEquals`
`assertFalse`/`assertNotContains` --> `assertNull`

PR #1301, opened 2015-01-07
WIP: Connection route selection

nfuller:
Work in progress. This is to get feedback and to show you where I'm heading with this. The ultimate aim is to implement multiple-level fallback, but without creating unnecessary connections when an SSLSocketFactory returns a socket that does not support all the protocols.

In Android I only moved the protocol selection out of RouteSelector / Route. Here, I have kept the RouteSelector, but made Connection responsible for creating the socket connection; it handles the retries.

See https://android-review.googlesource.com/#/c/113411 for the Android change against an older OkHttp

This pull request adds a commit on top of the first refactoring commit I sent out earlier. Please see the commit comment there for more information and potential benefits from the change.

Several files have TODO(nfuller) in the code and I would like feedback on. All OkHttp tests pass.

I expect the next commit would further modify the RouteSelector / Connection interaction. An SSLSocket can be interrogated by the RouteSelectorImpl to work out whether further attempts with other TLS protocols are worth trying. As things stand and plain socket is created, an SSLSocket wrapped around it, and *then* the Connection code would be able to work out there's no point in going further. This way we can avoid making unnecessary connections.

### DIFF
--- a/okhttp/src/main/java/com/squareup/okhttp/Call.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/Call.java
@@ -270,6 +270,10 @@ public class Call {
       try {
         engine.sendRequest();
         engine.readResponse();
+      } catch (ConnectionFailureException e) {
+        // TODO(nfuller): It would be nice to log or expose all exceptions somehow. E.g. >= JDK1.7
+        // with "suppressed" exceptions.
+        throw e.getLastConnectException();
       } catch (IOException e) {
         HttpEngine retryEngine = engine.recover(e, null);
         if (retryEngine != null) {
--- a/okhttp/src/main/java/com/squareup/okhttp/ConnectionFailureException.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/ConnectionFailureException.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright (C) 2015 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.okhttp;
+
+import java.io.IOException;
+import java.util.LinkedList;
+
+/**
+ * An exception thrown to indicate a problem connecting to an address. Multiple attempts may have
+ * been made, none of which were successful.
+ */
+public class ConnectionFailureException extends Exception {
+
+  private final LinkedList<IOException> connectExceptions = new LinkedList<IOException>();
+
+  public ConnectionFailureException(IOException cause) {
+    super(cause);
+    connectExceptions.add(cause);
+  }
+
+  public IOException getLastConnectException() {
+    return connectExceptions.getLast();
+  }
+
+  public void addConnectException(IOException e) {
+    connectExceptions.add(e);
+  }
+}
--- a/okhttp/src/main/java/com/squareup/okhttp/internal/http/HttpEngine.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/internal/http/HttpEngine.java
@@ -20,6 +20,7 @@ package com.squareup.okhttp.internal.http;
 import com.squareup.okhttp.Address;
 import com.squareup.okhttp.CertificatePinner;
 import com.squareup.okhttp.Connection;
+import com.squareup.okhttp.ConnectionFailureException;
 import com.squareup.okhttp.ConnectionPool;
 import com.squareup.okhttp.Headers;
 import com.squareup.okhttp.Interceptor;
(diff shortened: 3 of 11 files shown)


swankjesse:
I like the direction!

swankjesse:
(But I haven't reviewed this too carefully yet.)

nfuller:
Rebased.

okhttp/src/main/java/com/squareup/okhttp/Call.java
```diff
@@ -270,6 +270,10 @@ Response getResponse(Request request, boolean forWebSocket) throws IOException {
       try {
         engine.sendRequest();
         engine.readResponse();
+      } catch (ConnectionFailureException e) {
+        // TODO(nfuller): It would be nice to log or expose all exceptions somehow. E.g. >= JDK1.7
+        // with "suppressed" exceptions.
```
swankjesse: nit: scare quotes aren't necessary; you can just say "with suppressed exceptions."

okhttp/src/main/java/com/squareup/okhttp/ConnectionFailureException.java
```diff
@@ -0,0 +1,41 @@
...
+import java.io.IOException;
+import java.util.LinkedList;
+
+/**
+ * An exception thrown to indicate a problem connecting to an address. Multiple attempts may have
+ * been made, none of which were successful.
+ */
+public class ConnectionFailureException extends Exception {
+
+  private final LinkedList<IOException> connectExceptions = new LinkedList<IOException>();
```
swankjesse: LinkedList? I think you mistyped ArrayList

okhttp/src/main/java/com/squareup/okhttp/internal/http/HttpEngine.java
```diff
@@ -413,18 +398,14 @@ public HttpEngine recover(IOException e, Sink requestBodyOut) {
 
     Connection connection = close();
 
+    // TODO(nfuller): I'm not convinced about the retention of RouteSelector in HttpEngine. Its use
```
swankjesse: This doesn't really seem like a TODO; perhaps it's more of a bug report. What's the bug?

okhttp/src/main/java/com/squareup/okhttp/internal/http/HttpEngine.java
```diff
@@ -399,9 +385,8 @@ public Connection getConnection() {
    * body is buffered.
    */
   public HttpEngine recover(IOException e, Sink requestBodyOut) {
-    if (routeSelector != null && connection != null) {
-      connectFailed(routeSelector, e);
-    }
+    // TODO(nfuller):  Note: With this change a failure to issue a request/read a
```
swankjesse: Again, not a TODO.

okhttp/src/main/java/com/squareup/okhttp/internal/http/RouteSelectorImpl.java
```diff
@@ -45,7 +47,7 @@
  * choice of proxy server, IP address, and TLS mode. Connections may also be
  * recycled.
  */
-public final class RouteSelector {
+public final class RouteSelectorImpl implements RouteSelector {
```
swankjesse: nit: rename to `RealRouteSelector` ? I really loathe the `Impl` suffix. (And want to nuke it from everything in OkHttp.)

okhttp/src/main/java/com/squareup/okhttp/internal/http/HttpEngine.java
swankjesse: Sorry; I didn't realize this was intended as an annotation for my benefit!

swankjesse:
I'm under the impression that connection failures are asynchronous. Ie. if a host is unreachable, I won't necessarily discover it until I attempt to read; my writes will proceed "successfully" but won't go anywhere. Is this assumption valid? And if it is, does it invalidate the premise of your change?

swankjesse:
. . . after further thinking on this, and some experimentation, I'm thinking that a lot of the extra recovery is exclusively for the benefit of pooled connections and not failed connections. For whatever reason I was thinking that connect could return without a completed 3-way handshake.

nfuller:
I wrote a long response but got side side-tracked with a simper change so I didn't send. The one I'm working on instead is (I think) is less risky that the one I was going to propose next (without precluding it or this one). It also allows me to focus on the core issue of TLS fallback as I did in the Android codebase.

Summary I agree, though I did think of these things, which also mostly agree:

I can't rule out alternative implementations of socket that "pretend" to have connected when they haven't, but I'm not particularly worrying about those: I consider them edge cases (unless we have an example otherwise) and non-compliant.

I also can't rule out proxies that could delay the actual connection the backend until the request has been attempted. I could see this being a problem with non-TLS proxy use, but I don't think that would possible with tunnelling, where the TLS handshake with the ultimate server has to have taken place to know whether you can establish a connection at all.

I do worry about the layer above. For example, pooled connections. Just because a socket worked once, doesn't mean that it will when it is tried again later. This is one reason a resisted removing retry logic at the HTTP layer. There you really might find out that the pooled Connection has gone bad and you would want to retry (although I dispute that you would want to retry with a different route, and I think I would only want to do so if the request failed, not the response). Here I figured a later change might look at whether a connection originated from the pool. An HTTP request/response attempt with a pooled connection should be given some leniency if it failed on request, and a new socket connection would be set up. If a brand new socket connection (i.e. not a pooled one) failed it should generally be considered a failure against the route and given no leniency.

I hope to have my improved / simpler PR out today.

swankjesse:
Excellent. I agree, if the connection came from a pool then its route "doesn't count", and anything that fails shouldn't be held against the route.

nfuller:
Closing. I'd rather we focused on https://github.com/square/okhttp/pull/1339

PR #1302, opened 2015-01-08
Update documentation of mockwebserver

josketres:


JakeWharton:
This is not true for master. If you are using version 2.2 (or otherwise) browse its documentation by switching to the 2.2 tag.

josketres:
You're right!

PR #1304, opened 2015-01-08
OkHttp now needs to explicitly close the response InputStream before it ...

nfuller:
...will be cached

I think this is a regression introduced by commit
e74e3f3bf744ef7f4d8ee724a7cf2347e486cfab

Can it go back the way it was?

[Do not submit as is - right now the test hangs due to the change in behavior]


swankjesse:
Yes, this is a deliberate behavior change. Is this a problem?

swankjesse:
Yup. We'll spend up to 100ms trying to cache the entire thing. If we can't cache it in that deadline, we don't cache it.

PR #1307, opened 2015-01-09
Source-incompatible: ResponseBody methods throw IOException

nfuller:
This is a binary-compatible change, but code that currently calls
ResponseBody [contentLength()|byteStream()|source()|charStream()]
that doesn't necessarily catch or declare IOException will need
to with this change.

These methods could be performing I/O (opening streams, etc.).
Throwing IOException seems reasonable in this case.

contentType() and charset() have not been changed.


nfuller:
Happy to change contentType() as well, but I don't currently see any need. It might be worth to do it all at once.

PR #1308, opened 2015-01-09
Fix ResponseCache handling for Android usecases

nfuller:
Please see the commit comments for details.

I'd also like to propose a subsequent change to make the purpose of this code more obvious:

We move the following classes into an okhttp/okhttp-android-support sub-build + include it in the standard build (but in the same packages as they are today. The okhttp-android-support code would only be for Android to use, but by keeping it in the okhttp git repository we avoid future pain when Android pull from upstream, plus we reduce the amount of Android-specific junk in okhttp-httpurlconnection.

Move these:
1) AndroidShimResponseCache
2) OkCacheContainer
3) android/HttpResponseCache + Test
4) internal/huc/JavaApiConverter.java + Test
5) internal/huc/CacheAdapter.java

AFAIK, the only thing stopping this is the following method on OkUrlFactory. We could move this code into okhttp-android-support in the appropriate package (e.g. a static method in com.squareup.okhttp.AndroidInternal):

  /** Sets the response cache to be used to read and write cached responses. */
  public OkUrlFactory setResponseCache(ResponseCache responseCache) {
    if (responseCache instanceof OkCacheContainer) {
      // Avoid adding layers of wrappers. Rather than wrap the ResponseCache in yet another layer to
      // make the ResponseCache look like an InternalCache, we can unwrap the Cache instead.
      // This means that Cache stats will be correctly updated.
      OkCacheContainer okCacheContainer = (OkCacheContainer) responseCache;
      client.setCache(okCacheContainer.getCache());
    } else {
      client.setInternalCache(responseCache != null ? new CacheAdapter(responseCache) : null);
    }
    return this;
  }

What do you think?

### DIFF
--- a/okhttp-android-support/src/main/java/com/squareup/okhttp/AndroidInternal.java
+++ b/okhttp-android-support/src/main/java/com/squareup/okhttp/AndroidInternal.java
@@ -15,13 +15,14 @@
  */
 package com.squareup.okhttp;
 
-import com.squareup.okhttp.internal.InternalCache;
 import com.squareup.okhttp.internal.huc.CacheAdapter;
 
 import java.net.ResponseCache;
 
 /**
- * A class for back doors for Android's use of OkHttp within the Android platform.
+ * Back doors to enable the use of OkHttp within the Android platform libraries. OkHttp is used to
+ * provide the default {@link java.net.HttpURLConnection} / {@link javax.net.ssl.HttpsURLConnection}
+ * implementation including support for a custom {@link ResponseCache}.
  */
 public class AndroidInternal {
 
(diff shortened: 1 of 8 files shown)


okhttp-urlconnection/src/main/java/com/squareup/okhttp/OkUrlFactory.java
```diff
@@ -39,8 +39,16 @@ public OkHttpClient client() {
   }
 
   /** Sets the response cache to be used to read and write cached responses. */
-  OkUrlFactory setResponseCache(ResponseCache responseCache) {
-    client.setInternalCache(responseCache != null ? new CacheAdapter(responseCache) : null);
+  public OkUrlFactory setResponseCache(ResponseCache responseCache) {
+    if (responseCache instanceof OkCacheContainer) {
+      // Avoid adding layers of wrappers. Rather than wrap the ResponseCache in yet another layer to
```
swankjesse: Great idea.

swankjesse:
This is a great idea, and will make future maintenance much easier! Want me to review as-is? Or do you want to create the new `okhttp-android-support` build first?

nfuller:
I'm happy either way. It looks like I have the morning here before you will be awake so I'll make a start on the rearrangement and see how I get on. If I make good progress I should be able to update this one easily enough afterwards.

nfuller:
Done, with a slight detour after I discovered a change over the weekend that broke the Android code. Feeling good we'll spot these things more easily in future with the third / main commit.

swankjesse:
Dig. Do you mind rebasing this?

nfuller:
Rebased. Down to a single commit now.

okhttp-android-support/src/main/java/com/squareup/okhttp/AndroidInternal.java
```diff
@@ -30,12 +31,15 @@ private AndroidInternal() {
...
-    okUrlFactory.client().setInternalCache(
-        responseCache != null ? new CacheAdapter(responseCache) : null);
-  }
-
-  public static ResponseCache getResponseCache(OkUrlFactory okUrlFactory) {
-    InternalCache cache = okUrlFactory.client().internalCache();
-    return cache instanceof CacheAdapter ? ((CacheAdapter) cache).getDelegate() : null;
+    OkHttpClient client = okUrlFactory.client();
+    if (responseCache instanceof OkCacheContainer) {
+      // Avoid adding layers of wrappers. Rather than wrap the ResponseCache in yet another layer to
```
swankjesse: This amazing

swankjesse:
Thanks!

PR #1310, opened 2015-01-11
Lazily initialize the cache.

JakeWharton:
For Android this defers a non-zero amount of I/O when applications are initializing. Since this is both happening on app startup and on the main thread it has a large chance of non-trivially impacting startup even though the work is relatively small.

### DIFF
--- a/okhttp/src/main/java/com/squareup/okhttp/internal/DiskLruCache.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/internal/DiskLruCache.java
@@ -150,6 +150,9 @@ public final class DiskLruCache implements Closeable {
   private final LinkedHashMap<String, Entry> lruEntries = new LinkedHashMap<>(0, 0.75f, true);
   private int redundantOpCount;
 
+  // Must be read and written when synchronized on 'this'.
+  private boolean initialized;
+
   /**
    * To differentiate between old and current snapshots, each entry is given
    * a sequence number each time an edit is committed. A snapshot is stale if
@@ -162,7 +165,7 @@ public final class DiskLruCache implements Closeable {
   private final Runnable cleanupRunnable = new Runnable() {
     public void run() {
       synchronized (DiskLruCache.this) {
-        if (journalWriter == null) {
+        if (isClosed()) {
           return; // Closed.
         }
         try {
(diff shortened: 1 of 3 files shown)


JakeWharton:
Both `Cache` and `DiskLruCache` get binary compatible but source incompatible changes. The net win is that constructing either one no longer requires handling (read: ignoring) `IOException`.

okhttp/src/main/java/com/squareup/okhttp/internal/DiskLruCache.java
```diff
@@ -206,29 +214,31 @@ void initialize() throws IOException {
       try {
         readJournal();
         processJournal();
+        initialized = true;
```
JakeWharton: I hate initializing this in two places and having to reset `closed` below. This method's flow control is wonky.

okhttp/src/main/java/com/squareup/okhttp/internal/DiskLruCache.java
```diff
@@ -190,7 +193,11 @@ public void run() {
   }
 
   // Visible for testing.
-  void initialize() throws IOException {
+  synchronized void initialize() throws IOException {
```
swankjesse: Should this assume the caller already did synchronized? Maybe with a guard on like `Thread.holdsLock(this)` ?

okhttp/src/main/java/com/squareup/okhttp/internal/DiskLruCache.java
```diff
@@ -432,6 +440,7 @@ public synchronized Snapshot get(String key) throws IOException {
    * edit is in progress.
    */
   public Editor edit(String key) throws IOException {
+    initialize();
```
swankjesse: move this to be inside the edit overload that takes a sequence number.

swankjesse: (otherwise you obtain the lock, release it, and then obtain it again)

swankjesse:
Laughing Gas Tickles Me. :shipit: 

Really love this. This is going to be a big deal for Android, 'cause now initialization will happen on the background thread. Hooray!

PR #1312, opened 2015-01-12
Create a new maven artifact to hold things needed for Android embedding

nfuller:
okhttp-android-support contains classes needed for Android embedding and
not for normal OkHttp development. The classes here can be excluded from
okhttp-urlconnection, shrinking that artifact. More classes will be
added to this component to make Android maintenance easier.


swankjesse:
Nice.

PR #1313, opened 2015-01-12
Fix isClosed() when a cache has not been initialized

nfuller:
isClosed() now means "has close() been called", after
commit ea565b2e30e15cd52ddfc2ddc6db4ea8f1c3de88
it meant "has not been initialized or close() has been
called".

Introduced explicit closed state. Minor tweak to use
initialized state to determine whether cleanup will
do anything. Added a test.


JakeWharton:
Thanks! I had something like this initially but thought we might be able to get away without it.

nfuller:
No problem. Not sure my change is particularly good (especially the need to set the closed back to false after the call to delete()) but was the smallest I could get away with and get everything passing.

I spotted the issues via an Android test I'm proposing to add to OkHttp here: https://github.com/square/okhttp/pull/1308 

I'm hoping that commit will make my life substantially easier without making your lives too much harder.

PR #1315, opened 2015-01-12
Move stringy test methods to test class.

JakeWharton:


JakeWharton:
This found a bug!

okhttp-tests/src/test/java/com/squareup/okhttp/internal/DiskLruCacheTest.java
```diff
@@ -1225,15 +1207,23 @@ private void assertAbsent(String key) throws Exception {
...
-    assertEquals(value1, snapshot.getString(1));
-    assertEquals(value1.length(), snapshot.getLength(1));
+    assertSnapshotValue(snapshot, value0, 0);
+    assertSnapshotValue(snapshot, value1, 1);
     assertTrue(getCleanFile(key, 0).exists());
     assertTrue(getCleanFile(key, 1).exists());
     snapshot.close();
   }
 
+  private void assertSnapshotValue(DiskLruCache.Snapshot snapshot, String value, int index)
```
swankjesse: nit: typically it's (index, value), not (value, index). As in List.set() among others.

JakeWharton: Yeah I do that in the other method. I like it better, too.

swankjesse:
LGTM

Issue #1069, opened 2014-09-29
MockWebserver 2.0 play() does not seem to work.


ksarmalkar:
I have code that uses `MockWebServer`, however it seems like the server does not accept incoming connections. This seems to have started happening since we moved to 2.0.0, on 1.6.0 things were working fine. 

The dependencies are 
`compile 'com.squareup.okhttp:okhttp:2.0.0'`
`compile 'com.squareup.okhttp:okhttp-urlconnection:2.0.0'`
`androidTestCompile ('com.squareup.okhttp:mockwebserver:2.0.0'){
        exclude group: 'com.squareup.okhttp'
    }`

Tried a sample test as below, once play is called, I try to hit the server URL:port using curl/browser, but it does not seem to accept any connection, Just says connection refused. Isn't it supposed to return the enqueued responses ?

``` java
public void testMockWebServer() throws Exception {
        // Create a MockWebServer. These are lean enough that you can create a new
        // instance for every unit test.
        MockWebServer server = new MockWebServer();

        // Schedule some responses.
        server.enqueue(new MockResponse().setBody("hello, world!"));
        server.enqueue(new MockResponse().setBody("sup, bra?"));
        server.enqueue(new MockResponse().setBody("yo dog"));

        // Start the server.
        server.play();
        server.getPort();
        Thread.sleep(5000000000L);
        // Shut down the server. Instances cannot be reused.
        server.shutdown();
    }
```

josketres:
I'm having the same issue using version 2.2.0 running in a real device and in an emulator.
When I change to **version 2.1.0 the test runs without problems** in the device and in the emulator.

    compile 'com.squareup.okhttp:okhttp:2.2.0'
    androidTestCompile('com.squareup.okhttp:mockwebserver:2.2.0') {
        exclude group: 'com.squareup.okhttp'
    }

my test looks like this

```java
import android.test.AndroidTestCase;

import com.squareup.okhttp.OkHttpClient;
import com.squareup.okhttp.Request;
import com.squareup.okhttp.mockwebserver.MockResponse;
import com.squareup.okhttp.mockwebserver.MockWebServer;

public class UseOfMockWebServerTest extends AndroidTestCase {

    public void test() throws Exception {

        MockWebServer server = new MockWebServer();
        server.enqueue(new MockResponse().setBody("hello, world!"));
        server.play();
        String url = server.getUrl("/").toString();
[...]

PR #1320, opened 2015-01-13
Modify the way MockWebServer selects a host IP to bind to

nfuller:
On Android getByName(null) returns the first of
[IPv6 loopback, IPv4 loopback], binding the server to ::1.
The IPv6 loopback address returns "localhost". "localhost"
does not resolve to the IPv6 address.

Instead: explicitly select "localhost" ensuring
MockWebServer binds to an IP that can be looked up by name
and should also be a loopback address.

Suggested fix for https://github.com/square/okhttp/issues/1069


swankjesse:
This was originally changed here: b78030a630dc0b757d77e8565241f984e11feaf8

PR #1321, opened 2015-01-13
Fix the user agent test so it works with a different user agent string

nfuller:
The important thing is that the Version.userAgent() string is used and
not the exact string.


JakeWharton:
Test failure (only one one JDK) was a flake.

PR #1326, opened 2015-01-15
Fix for MockWebServer on Android

nfuller:
On Android L (and probably prior releases) it is necessary to
close a socket explicitly rather than relying on
InputStream/OutputStream to do it for you.

See bug here: https://code.google.com/p/android/issues/detail?id=97564


swankjesse:
Thanks!

PR #1327, opened 2015-01-15
(Hopefully) fix a flake in Spdy3ConnectionTest on Android

nfuller:
In some cases serverSocket is found to be null in
readAndWriteFrames().

The suspicion is that it's because of a lack of synchronization
between the thread setting the field, and the thread (started in
play()) reading it. serverSocket / port have been made
volatile to ensure the field state is flushed / read and not
cached. No flakes seen in several runs since the change.


JakeWharton:
Not sure where Travis CI is. Closing and re-opening to kick it.

swankjesse:
The synchronization occurs with the call to `executor.execute()`. From the ExecutorService docs, _"Memory consistency effects: Actions in a thread prior to the submission of a Runnable or Callable task to an ExecutorService happen-before any actions taken by that task, which in turn happen-before the result is retrieved via Future.get()._"

swankjesse:
That is, if you're seeing those fields not initialized, it's more likely that `close()` is being called before `readAndWriteFrames()` runs. That could be a problem, but I don't think this fixes it.

(We also had that problem in MockWebServer; the fix was `12f858caac6b9215b7d3cd84a781c812fae6c2bb`

nfuller:
Think you're correct.

Issue #782, opened 2014-04-27
Media range matching


swankjesse:
We could implement support for [matching media types](http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html):

```
       Accept: audio/*; q=0.2, audio/basic
```

As suggested by @rubUx4 here https://github.com/square/okhttp/pull/762

swankjesse:
Iceboxing this. Accepting small, well-motivated patches.

yschimke:
Closing as low priority.  Re-open if there is a compelling sue case we are missing.

PR #1329, opened 2015-01-15
Add helper class with commonly used MediaType constants

joschi:
This change set adds the helper class `MediaTypes` defining commonly used media types.

The class has been inspired by [`MediaType`](http://docs.oracle.com/javaee/7/api/javax/ws/rs/core/MediaType.html) from JAX-RS and is intended to save users from typing or having to define commonly used media types by themselves again and again.

The selection of media types in `MediaTypes` is rather arbitrary but should be sensible.

### DIFF
--- a/okhttp/src/main/java/com/squareup/okhttp/MediaType.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/MediaType.java
@@ -31,6 +31,38 @@ public final class MediaType {
   private static final Pattern PARAMETER = Pattern.compile(
       ";\\s*(?:" + TOKEN + "=(?:" + TOKEN + "|" + QUOTED + "))?");
 
+  /**
+   * A {@link MediaType} constant representing the "application/xml; charset=utf-8" media type.
+   */
+  public static final MediaType APPLICATION_XML =
+                  new MediaType("application/xml; charset=utf-8", "application", "xml", "utf-8");
+  /**
+   * A {@link MediaType} constant representing the "application/json; charset=utf-8" media type.
+   */
+  public static final MediaType APPLICATION_JSON =
+                  new MediaType("application/json; charset=utf-8", "application", "json", "utf-8");
+  /**
+   * A {@link MediaType} constant representing the "application/x-www-form-urlencoded" media type.
+   */
+  public static final MediaType APPLICATION_FORM_URLENCODED =
+                  new MediaType("application/x-www-form-urlencoded",
+                                "application", "x-www-form-urlencoded", null);
+  /**
+   * A {@link MediaType} constant representing the "multipart/form-data" media type.
+   */
+  public static final MediaType MULTIPART_FORM_DATA =
+                  new MediaType("multipart/form-data", "multipart", "form-data", null);
+  /**
+   * A {@link MediaType} constant representing the "application/octet-stream" media type.
+   */
+  public static final MediaType APPLICATION_OCTET_STREAM =
+                  new MediaType("application/octet-stream", "application", "octet-stream", null);
+  /**
+   * A {@link MediaType} constant representing the "text/plain; charset=utf-8" media type.
+   */
+  public static final MediaType TEXT_PLAIN =
+                  new MediaType("text/plain; charset=utf-8", "text", "plain", "utf-8");
+
   private final String mediaType;
   private final String type;
   private final String subtype;
--- a/okhttp-tests/src/test/java/com/squareup/okhttp/MediaTypeTest.java
+++ b/okhttp-tests/src/test/java/com/squareup/okhttp/MediaTypeTest.java
@@ -156,6 +156,15 @@ public class MediaTypeTest {
     assertEquals("text/plain;", mediaType.toString());
   }
 
+  @Test public void testValidTypes() throws Exception {
+    assertEquals(MediaType.parse("application/x-www-form-urlencoded"), MediaType.APPLICATION_FORM_URLENCODED);
+    assertEquals(MediaType.parse("application/octet-stream"), MediaType.APPLICATION_OCTET_STREAM);
+    assertEquals(MediaType.parse("multipart/form-data"), MediaType.MULTIPART_FORM_DATA);
+    assertEquals(MediaType.parse("application/json; charset=utf-8"), MediaType.APPLICATION_JSON);
+    assertEquals(MediaType.parse("application/xml; charset=utf-8"), MediaType.APPLICATION_XML);
+    assertEquals(MediaType.parse("text/plain; charset=utf-8"), MediaType.TEXT_PLAIN);
+  }
+
   private void assertMediaType(String string) {
     MediaType mediaType = MediaType.parse(string);
     assertEquals(string, mediaType.toString());


okhttp/src/main/java/com/squareup/okhttp/MediaTypes.java
```diff
@@ -0,0 +1,108 @@
+package com.squareup.okhttp;
+
+/**
+ * Constants for commonly used media types.
+ *
+ * @see com.squareup.okhttp.MediaType
+ */
+public abstract class MediaTypes {
```
artem-zinnatullin: Special class for constants is not a good OOP design.. But okay, if you do so, please make sure that the class is not instantiable (private constructor), here is good thread about storing constants in Java http://stackoverflow.com/questions/320588/interfaces-with-static-fields-in-java-for-sharing-constants

artem-zinnatullin:
Also, if I'd want to use only one constant from `MediaTypes`, for example `MediaTypes.APPLICATION_JSON`, device (Android in bad case) would have to compute all static values from `MediaTypes` just to load the class and this will cause multiple calls to `MediaType.parse()` which is quite complex method + memory allocations

okhttp/src/main/java/com/squareup/okhttp/MediaTypes.java
joschi: I thought about putting the constants into the existing `MediaType` class but thought it might just blow up the code there.

Good point about the instantiation.

joschi:
I thought about that case (memory restricted devices) as well but I couldn't come up with a better idea. Having only String constants in `MediaTypes` kind of defeats its purpose.

Do you have a suggestion on how to prevent bad effects on restricted machines? The only solution I could come up with is by moving the constants into `MediaType` and use its private constructor (which would work since we know how to build proper media types there).

joschi:
Upon thinking a little bit more about the issues I've moved the constants into the `MediaType` class and use the private constructor to build them.

okhttp/src/main/java/com/squareup/okhttp/MediaType.java
```diff
@@ -31,6 +31,64 @@
   private static final Pattern PARAMETER = Pattern.compile(
       ";\\s*(?:" + TOKEN + "=(?:" + TOKEN + "|" + QUOTED + "))?");
 
+  /**
+   * A {@link MediaType} constant representing the wildcard media type.
+   */
+  public static final MediaType WILDCARD = new MediaType("*/*", "*", "*", null);
```
swankjesse: Are you anticipating using `MediaType` in an `Accept` header? At the moment we don't implement media range matching.

https://github.com/square/okhttp/issues/782

I'd prefer to omit this until we support that.

okhttp/src/main/java/com/squareup/okhttp/MediaType.java
```diff
@@ -31,6 +31,64 @@
...
+  /**
+   * A {@link MediaType} constant representing the "application/xml" media type.
+   */
+  public static final MediaType APPLICATION_XML =
+                  new MediaType("application/xml", "application", "xml", null);
+  /**
+   * A {@link MediaType} constant representing the "application/atom+xml" media type.
+   */
+  public static final MediaType APPLICATION_ATOM_XML =
```
swankjesse: Drop this? I'd like to limit this to a very small set of very popular media types:

XML,
JSON,
FORM_URLENCODED,
FORM_DATA,
OCTET_STREAM,
and TEXT_PLAIN

Do we need two different media types for XML? That's unfortunate!

okhttp/src/main/java/com/squareup/okhttp/MediaType.java
```diff
@@ -31,6 +31,64 @@
...
+  /**
+   * A {@link MediaType} constant representing the "application/svg+xml" media type.
+   */
+  public static final MediaType APPLICATION_SVG_XML =
+                  new MediaType("application/svg+xml", "application", "svg+xml", null);
+  /**
+   * A {@link MediaType} constant representing the "application/json" media type.
+   */
+  public static final MediaType APPLICATION_JSON =
+                  new MediaType("application/json", "application", "json", null);
```
swankjesse: Please include `"utf-8"` as the charset for all of the applicable content types, and also mention that in the member documentation? If we're going to make it easy to get a content type, we should make sure you get one that's fully-specified.

okhttp/src/main/java/com/squareup/okhttp/MediaType.java
```diff
@@ -31,6 +31,64 @@
...
+   */
+  public static final MediaType TEXT_PLAIN = new MediaType("text/plain", "text", "plain", null);
+  /**
+   * A {@link MediaType} constant representing the "text/xml" media type.
+   */
+  public static final MediaType TEXT_XML = new MediaType("text/xml", "text", "xml", null);
+  /**
+   * A {@link MediaType} constant representing the "text/html" media type.
+   */
+  public static final MediaType TEXT_HTML = new MediaType("text/html", "text", "html", null);
```
swankjesse: Is this appropriate for OkHttp? I assume that if you're reading & writing HTML, you're probably doing it with a webview and not OkHttp.

okhttp/src/main/java/com/squareup/okhttp/MediaType.java
swankjesse: (I guess you'd also have to change the first argument, to be like `application/json; charset=utf-8`)

okhttp/src/main/java/com/squareup/okhttp/MediaType.java
joschi: Fair enough, I can change that.

The media types are currently exactly the same that are also provided by the JAX-RS class [`MediaType`](http://docs.oracle.com/javaee/7/api/javax/ws/rs/core/MediaType.html).

joschi:
@swankjesse I've updated the PR with your proposed changes.

swankjesse:
Perfect. LGTM.

Please rebase & squash & I'll merge.
rebaseandsqua.sh

joschi:
@swankjesse Done. Thanks!

JakeWharton:
I'm :-1: for this change.

The included media types will never cover everyone's use cases which still requires me to define my own (for example, we use protos almost exclusively which is not included here). This will result in subsequent pull requests for people wanting to add their own favorite media type of the week. Additionally, some endpoints (like GitHub) don't use standard media types despite serving JSON.

Conversely, if I only use, say, JSON, I still pay the cost of having all these objects be initialized and in the public API.

I'd personally just rather have a policy of "bring your own". A lot of the time I could use the ones in this PR, but I think operationally having zero is the better decision from the library's perspective.

joschi:
I'm totally fine if this PR doesn't get merged and I can understand it from a library author's perspective. While OkHttp is a generic HTTP client, the choice of media types in this PR is still rather arbitrary (and based on my use cases).

swankjesse:
Despite it not being merged, this was a good discussion & decision. Thanks for the pull request @joschi!

PR #1330, opened 2015-01-15
Allow to create Headers from Map

joschi:
This change set adds a method to the `Headers` class which allows users to create a new `Headers` instance from a `Map<String, String>` (similar to `Headers#of(String[])`).

### DIFF
--- a/okhttp-tests/src/test/java/com/squareup/okhttp/internal/http/HeadersTest.java
+++ b/okhttp-tests/src/test/java/com/squareup/okhttp/internal/http/HeadersTest.java
@@ -22,7 +22,11 @@ import com.squareup.okhttp.Response;
 import com.squareup.okhttp.internal.spdy.Header;
 import java.io.IOException;
 import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
+
 import org.junit.Test;
 
 import static com.squareup.okhttp.TestUtil.headerEntries;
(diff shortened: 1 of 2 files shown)


okhttp-tests/src/test/java/com/squareup/okhttp/internal/http/HeadersTest.java
```diff
@@ -234,4 +238,68 @@
     } catch (IllegalArgumentException expected) {
     }
   }
+
+  @Test public void ofMapThrowsOnNull() {
```
swankjesse: Nice tests!

okhttp/src/main/java/com/squareup/okhttp/Headers.java
```diff
@@ -167,6 +168,33 @@ public static Headers of(String... namesAndValues) {
...
+  public static Headers of(Map<String, String> headers) {
+    if (headers == null) {
+      throw new IllegalArgumentException("Expected map with header names and values");
+    }
+
+    // Make a defensive copy and clean it up.
+    final String[] namesAndValues = new String[headers.size() * 2];
+    int i = 0;
+    for (Map.Entry<String, String> header : headers.entrySet()) {
+      if (header.getKey() == null || header.getValue() == null)
```
swankjesse: nitpick: our code style always uses `{` for wrapped lines, and never uses `final` for locals

swankjesse:
Nice.

Please accept our contributor's agreement, address my nitpicks & I'll merge your change.
https://docs.google.com/forms/d/13WR8m5uZ2nAkJH41k7GdVBXAAbzDk00vxtEYjd6Imzg/viewform?ndplr=1&formkey=dDViT2xzUHAwRkI3X3k5Z0lQM091OGc6MQ

joschi:
@swankjesse I've signed the CLA and adapted to the project's code style.

swankjesse:
Thanks!

Issue #1253, opened 2014-12-29
OkHttp cache-by-default status codes out of date


swankjesse:
The spec wants these:
200, 203, 204, 206, 300, 301, 404, 405, 410, 414, and 501

http://tools.ietf.org/html/rfc7231#section-6.1

swankjesse:
Thanks @jbornemann !

PR #1332, opened 2015-01-16
Update http default cache codes

jbornemann:


xdevient:
1 _bash

Seeing these tests fail on master.  Any ideas?

okhttp-android-support/src/test/java/com/squareup/okhttp/internal/huc/ResponseCacheTest.java
```diff
@@ -1113,10 +1113,34 @@ public void assertCookies(CookieManager cookieManager, URL url, String... expect
   }
 
   @Test public void cachePlusRange() throws Exception {
```
swankjesse: I'd prefer to avoid suggesting we support any range caching in our tests because the built-in Cache class can't handle ranges. They're a lot of work. For now can we just avoid offering support for HTTP_PARTIAL ?

okhttp-tests/src/test/java/com/squareup/okhttp/CacheTest.java
```diff
@@ -113,41 +118,50 @@
...
-    assertCached(true, 200);
+    assertCached(true,  200);
     assertCached(false, 201);
     assertCached(false, 202);
-    assertCached(true, 203);
-    assertCached(false, 204);
+    assertCached(true,  203);
+    assertCached(true,  204);
     assertCached(false, 205);
-    assertCached(false, 206); // we don't cache partial responses
```
swankjesse: Yeah, I'd prefer we keep this as-is.

okhttp-tests/src/test/java/com/squareup/okhttp/CacheTest.java
```diff
@@ -113,41 +118,50 @@
     // assertCached(false, 100);
```
swankjesse: Quick summary of changes:

Newly cached with this update:
204 No Content: cool
206 Partial Content: **I'd prefer to leave this as-is.** Implementing this is difficult (and incomplete).
404 Not Found: cool
405 Method Not Allowed: cool
410 Gone: cool
414 URI too long: cool
501 Not implemented: cool

Newly not cached:
308 Permanent Redirect. **I'd prefer to leave this as-is**. [RFC 7238](https://tools.ietf.org/html/rfc7238) says, _"A 308 response is cacheable by default"_

okhttp/src/main/java/com/squareup/okhttp/Cache.java
```diff
@@ -66,9 +66,9 @@
  * valid. Such responses increment both the network count and hit count.
  *
  * <p>The best way to improve the cache hit rate is by configuring the web
- * server to return cacheable responses. Although this client honors all <a
+ * server to return cacheable responses. This client honors all <a
  * href="http://www.ietf.org/rfc/rfc2616.txt">HTTP/1.1 (RFC 2068)</a> cache
- * headers, it doesn't cache partial responses.
```
swankjesse: Yeah, we still don't actually do the hard work of caching partial responses properly.

okhttp/src/main/java/com/squareup/okhttp/Cache.java
```diff
@@ -142,12 +142,17 @@ public Cache(File directory, long maxSize) {
     cache = DiskLruCache.create(directory, VERSION, ENTRY_COUNT, maxSize);
   }
 
-  private static String urlToKey(Request request) {
+  private static String generateEntryKey(Request request) {
+    //If the request is for partial content, we want a unique key for the partial response
```
swankjesse: This is not how I'd like to implement partial responses. If we request ranges 0–100 and 100–200 of `http://square.com`, that should yield one 200-byte cache entry and not two 100-byte cache entries. I don't think this change will increase cache hit rates.

swankjesse:
Yeah, something is unhappy with SSL and the test suite:

```
Tests in error: 
  CallTest.matchingPinnedCertificate:864 » NoSuchMethod sun.security.ssl.SSLSess...
  URLConnectionTest.sslFallbackNotUsedWhenRecycledConnectionFails:635->assertContent:3076->assertContent:3072 » NoSuchMethod
```

Running on Java8 should fix until I investigate & fix.

swankjesse:
Nice. This should close https://github.com/square/okhttp/issues/1253

A few requests and then I'll merge!

1. Please revert behavior for 206 and 308. I'm not comfortable keeping range-specific responses in the cache.

2. Please complete out our contributor's agreement:
https://docs.google.com/forms/d/13WR8m5uZ2nAkJH41k7GdVBXAAbzDk00vxtEYjd6Imzg/viewform?ndplr=1&formkey=dDViT2xzUHAwRkI3X3k5Z0lQM091OGc6MQ

3. Also please squash your commits to a single commit so our revision history shows only what we reviewed! Instructions here: http://rebaseandsqua.sh/

okhttp/src/main/java/com/squareup/okhttp/Cache.java
jbornemann: Yeah.  Makes complete sense.  Thinking about it further, it's much more practical of a case that a request for 101-200 bytes proceed a request for 0 - 100 bytes.  I could see cache hit rates increase for cases of requesting a full body, then subsequently requesting partial bodies; but even that is probably something of an edge case.

jbornemann:
Cool, all CR comments addressed.  Btw, if you enjoy using http://rebaseandsqua.sh/ , you might check out https://github.com/jdigger/git-process

swankjesse:
Thanks!

PR #1334, opened 2015-01-17
Fix for SSL test

jbornemann:
@swankjesse thought I'd look into this.  Introduced in 03bb6bef 

The core of the issue is around the use of sun.security.ssl.SSLSessionImpl.setAsResumption(boolean) which is used in ALPN 7.1, but not introduced until Java 8. 

I wasn't seeing this fail in IDEA due to IDEA loading the OOTB 1.7 sun.security.ssl.ClientHandShaker over the jetty jar's ClientHandShaker.  

Ha, but sure enough, the reason the version bump was introduced (CallTest.matchingPinnedCertification) is brought back

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -36,7 +36,7 @@
     <java.version>1.7</java.version>
     <okio.version>1.2.0</okio.version>
     <!-- ALPN library targeted to Java 7 -->
-    <alpn.jdk7.version>7.1.2.v20141202</alpn.jdk7.version>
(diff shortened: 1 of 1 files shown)


jbornemann:
Feel free to commit to this branch, or whatever is easiest.  I'll look back at this next week

swankjesse:
Is there a tracking bug for the regression in Jetty ALPN ?

swankjesse:
No action to take here. I think we're already fixed upstream.

swankjesse:
Though we didn't merge the code, I appreciate the fix!

PR #1339, opened 2015-01-19
Move TLS fallback into Connection + make TLS fallback try 3 times (but only if appropriate)

nfuller:
Ready for a proper review.

### DIFF
--- a/okhttp-urlconnection/src/main/java/com/squareup/okhttp/internal/huc/HttpURLConnectionImpl.java
+++ b/okhttp-urlconnection/src/main/java/com/squareup/okhttp/internal/huc/HttpURLConnectionImpl.java
@@ -27,11 +27,13 @@ import com.squareup.okhttp.Response;
 import com.squareup.okhttp.Route;
 import com.squareup.okhttp.internal.Internal;
 import com.squareup.okhttp.internal.Platform;
+import com.squareup.okhttp.internal.http.RouteException;
 import com.squareup.okhttp.internal.Util;
 import com.squareup.okhttp.internal.http.HttpDate;
 import com.squareup.okhttp.internal.http.HttpEngine;
 import com.squareup.okhttp.internal.http.HttpMethod;
 import com.squareup.okhttp.internal.http.OkHeaders;
+import com.squareup.okhttp.internal.http.RequestException;
 import com.squareup.okhttp.internal.http.RetryableSink;
 import com.squareup.okhttp.internal.http.StatusLine;
 import java.io.FileNotFoundException;
@@ -432,7 +434,20 @@ public class HttpURLConnectionImpl extends HttpURLConnection {
       }
 
       return true;
+    } catch (RequestException e) {
+      // An attempt to interpret a request failed.
+      throw e.getCause();
+    } catch (RouteException e) {
+      // The attempt to connect via a route failed. The request will not have been sent.
+      HttpEngine retryEngine = httpEngine.recover(e);
+      if (retryEngine != null) {
+        httpEngine = retryEngine;
+        return false;
+      }
+      // Give up; recovery is not possible.
+      throw e.getLastConnectException();
     } catch (IOException e) {
+      // An attempt to communicate with a server failed. The request may have been sent.
       HttpEngine retryEngine = httpEngine.recover(e);
       if (retryEngine != null) {
         httpEngine = retryEngine;
--- a/okhttp/src/main/java/com/squareup/okhttp/Address.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/Address.java
@@ -45,13 +45,12 @@ public final class Address {
   final CertificatePinner certificatePinner;
   final Authenticator authenticator;
   final List<Protocol> protocols;
-  final List<ConnectionSpec> connectionSpecs;
   final ProxySelector proxySelector;
 
   public Address(String uriHost, int uriPort, SocketFactory socketFactory,
       SSLSocketFactory sslSocketFactory, HostnameVerifier hostnameVerifier,
       CertificatePinner certificatePinner, Authenticator authenticator, Proxy proxy,
-      List<Protocol> protocols, List<ConnectionSpec> connectionSpecs, ProxySelector proxySelector) {
+      List<Protocol> protocols, ProxySelector proxySelector) {
     if (uriHost == null) throw new NullPointerException("uriHost == null");
     if (uriPort <= 0) throw new IllegalArgumentException("uriPort <= 0: " + uriPort);
     if (authenticator == null) throw new IllegalArgumentException("authenticator == null");
@@ -66,7 +65,6 @@ public final class Address {
     this.certificatePinner = certificatePinner;
     this.authenticator = authenticator;
     this.protocols = Util.immutableList(protocols);
-    this.connectionSpecs = Util.immutableList(connectionSpecs);
     this.proxySelector = proxySelector;
   }
 
@@ -119,10 +117,6 @@ public final class Address {
     return protocols;
   }
 
-  public List<ConnectionSpec> getConnectionSpecs() {
-    return connectionSpecs;
-  }
-
   /**
    * Returns this address's explicitly-specified HTTP proxy, or null to
    * delegate to the {@linkplain #getProxySelector proxy selector}.
@@ -139,6 +133,13 @@ public final class Address {
     return proxySelector;
   }
 
+  /**
+   * Returns this address's certificate pinner. Only used for secure connections.
+   */
+  public CertificatePinner getCertificatePinner() {
+    return certificatePinner;
+  }
+
   @Override public boolean equals(Object other) {
     if (other instanceof Address) {
       Address that = (Address) other;
@@ -150,7 +151,6 @@ public final class Address {
           && equal(this.certificatePinner, that.certificatePinner)
           && equal(this.authenticator, that.authenticator)
           && equal(this.protocols, that.protocols)
-          && equal(this.connectionSpecs, that.connectionSpecs)
           && equal(this.proxySelector, that.proxySelector);
     }
     return false;
@@ -166,7 +166,6 @@ public final class Address {
     result = 31 * result + (certificatePinner != null ? certificatePinner.hashCode() : 0);
     result = 31 * result + authenticator.hashCode();
     result = 31 * result + protocols.hashCode();
-    result = 31 * result + connectionSpecs.hashCode();
     result = 31 * result + proxySelector.hashCode();
     return result;
   }
--- a/okhttp/src/main/java/com/squareup/okhttp/Call.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/Call.java
@@ -16,7 +16,9 @@
 package com.squareup.okhttp;
 
 import com.squareup.okhttp.internal.NamedRunnable;
+import com.squareup.okhttp.internal.http.RouteException;
 import com.squareup.okhttp.internal.http.HttpEngine;
+import com.squareup.okhttp.internal.http.RequestException;
 import java.io.IOException;
 import java.net.MalformedURLException;
 import java.net.ProtocolException;
@@ -270,7 +272,20 @@ public class Call {
       try {
         engine.sendRequest();
         engine.readResponse();
+      } catch (RequestException e) {
+        // The attempt to interpret the request failed. Give up.
+        throw e.getCause();
+      } catch (RouteException e) {
+        // The attempt to connect via a route failed. The request will not have been sent.
+        HttpEngine retryEngine = engine.recover(e);
+        if (retryEngine != null) {
+          engine = retryEngine;
+          continue;
+        }
+        // Give up; recovery is not possible.
+        throw e.getLastConnectException();
       } catch (IOException e) {
+        // An attempt to communicate with a server failed. The request may have been sent.
         HttpEngine retryEngine = engine.recover(e, null);
         if (retryEngine != null) {
           engine = retryEngine;
--- a/okhttp/src/main/java/com/squareup/okhttp/Connection.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/Connection.java
@@ -16,29 +16,18 @@
  */
 package com.squareup.okhttp;
 
-import com.squareup.okhttp.internal.Platform;
-import com.squareup.okhttp.internal.Util;
 import com.squareup.okhttp.internal.http.HttpConnection;
 import com.squareup.okhttp.internal.http.HttpEngine;
 import com.squareup.okhttp.internal.http.HttpTransport;
-import com.squareup.okhttp.internal.http.OkHeaders;
+import com.squareup.okhttp.internal.http.RouteException;
+import com.squareup.okhttp.internal.http.SocketConnector;
 import com.squareup.okhttp.internal.http.SpdyTransport;
 import com.squareup.okhttp.internal.http.Transport;
 import com.squareup.okhttp.internal.spdy.SpdyConnection;
-import com.squareup.okhttp.internal.tls.OkHostnameVerifier;
 import java.io.IOException;
-import java.net.Proxy;
 import java.net.Socket;
-import java.net.URL;
-import java.security.cert.X509Certificate;
-import java.util.concurrent.TimeUnit;
-import javax.net.ssl.SSLSocket;
-import okio.Source;
-
-import static com.squareup.okhttp.internal.Util.getDefaultPort;
-import static com.squareup.okhttp.internal.Util.getEffectivePort;
-import static java.net.HttpURLConnection.HTTP_OK;
-import static java.net.HttpURLConnection.HTTP_PROXY_AUTH;
+import java.net.SocketException;
+import java.util.List;
 
 /**
  * The sockets and streams of an HTTP, HTTPS, or HTTPS+SPDY connection. May be
@@ -141,23 +130,41 @@ public final class Connection {
     socket.close();
   }
 
-  void connect(int connectTimeout, int readTimeout, int writeTimeout, Request tunnelRequest)
-      throws IOException {
+  void connect(int connectTimeout, int readTimeout, int writeTimeout, Request request,
+      List<ConnectionSpec> connectionSpecs, boolean connectionRetryEnabled) throws RouteException {
     if (connected) throw new IllegalStateException("already connected");
 
-    if (route.proxy.type() == Proxy.Type.DIRECT || route.proxy.type() == Proxy.Type.HTTP) {
-      socket = route.address.socketFactory.createSocket();
+    SocketConnector socketConnector = new SocketConnector(this, pool);
+    SocketConnector.ConnectedSocket connectedSocket;
+    if (route.address.getSslSocketFactory() != null) {
+      // https:// communication
+      connectedSocket = socketConnector.connectTls(connectTimeout, readTimeout, writeTimeout,
+          request, route, connectionSpecs, connectionRetryEnabled);
     } else {
-      socket = new Socket(route.proxy);
+      // http:// communication.
+      if (!connectionSpecs.contains(ConnectionSpec.CLEARTEXT)) {
+        throw new RouteException(
+            new SocketException("CLEARTEXT communication not supported: " + connectionSpecs));
+      }
+      connectedSocket = socketConnector.connectClearText(connectTimeout, readTimeout, route);
     }
 
-    socket.setSoTimeout(readTimeout);
-    Platform.get().connectSocket(socket, route.inetSocketAddress, connectTimeout);
+    socket = connectedSocket.socket;
+    handshake = connectedSocket.handshake;
+    protocol = connectedSocket.alpnProtocol == null
+        ? Protocol.HTTP_1_1 : connectedSocket.alpnProtocol;
 
-    if (route.address.sslSocketFactory != null) {
-      upgradeToTls(tunnelRequest, readTimeout, writeTimeout);
-    } else {
-      httpConnection = new HttpConnection(pool, this, socket);
+    try {
+      if (protocol == Protocol.SPDY_3 || protocol == Protocol.HTTP_2) {
+        socket.setSoTimeout(0); // SPDY timeouts are set per-stream.
+        spdyConnection = new SpdyConnection.Builder(route.address.uriHost, true, socket)
+            .protocol(protocol).build();
+        spdyConnection.sendConnectionPreface();
+      } else {
+        httpConnection = new HttpConnection(pool, this, socket);
+      }
+    } catch (IOException e) {
+      throw new RouteException(e);
     }
     connected = true;
   }
@@ -166,13 +173,14 @@ public final class Connection {
    * Connects this connection if it isn't already. This creates tunnels, shares
    * the connection with the connection pool, and configures timeouts.
    */
-  void connectAndSetOwner(OkHttpClient client, Object owner, Request request) throws IOException {
+  void connectAndSetOwner(OkHttpClient client, Object owner, Request request)
+      throws RouteException {
     setOwner(owner);
 
     if (!isConnected()) {
-      Request tunnelRequest = tunnelRequest(request);
-      connect(client.getConnectTimeout(), client.getReadTimeout(),
-          client.getWriteTimeout(), tunnelRequest);
+      List<ConnectionSpec> connectionSpecs = client.getConnectionSpecs();
+      connect(client.getConnectTimeout(), client.getReadTimeout(), client.getWriteTimeout(),
+          request, connectionSpecs, client.getRetryOnConnectionFailure());
       if (isSpdy()) {
         client.getConnectionPool().share(this);
       }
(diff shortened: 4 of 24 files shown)


nfuller:
BTW - this is a replacement for my other PR, which is groping towards the behavior I want.

This goes straight for what I'm aiming for:
1) 3-stage fallback - TLSv1.2, TLSv1.1, TLSv1.0 (and, if you like, SSLv3 too).

2) Only keep retrying if the protocols make sense, e.g. if the socket only supports TLSv1.0, don't try to connect 3 times using TLSv1 (e.g. TLSv1.2 attempt would negotiate TLSv1, TLSv1.1 would attempt TLSv1, etc.) . If the socket supports TLSv1.2 and TLSv1.1 but doesn't support TLSv1.0, don't try the third time - we could have foreseen it wasn't going to work when we performed the TLSv1.1 connection.

swankjesse:
(Will take a look at this this weekend.)

okhttp-urlconnection/src/main/java/com/squareup/okhttp/internal/huc/HttpURLConnectionImpl.java
```diff
@@ -432,7 +434,20 @@ private boolean execute(boolean readResponse) throws IOException {
       }
 
       return true;
+    } catch (RequestException e) {
```
swankjesse: Neat.

okhttp/src/main/java/com/squareup/okhttp/Address.java
```diff
@@ -45,13 +45,12 @@
   final CertificatePinner certificatePinner;
   final Authenticator authenticator;
   final List<Protocol> protocols;
-  final List<ConnectionSpec> connectionSpecs;
```
swankjesse: Hmmm

okhttp/src/main/java/com/squareup/okhttp/Call.java
```diff
@@ -270,7 +272,20 @@ Response getResponse(Request request, boolean forWebSocket) throws IOException {
       try {
         engine.sendRequest();
         engine.readResponse();
+      } catch (RequestException e) {
+        // The attempt to interpret the request failed. Give up.
```
swankjesse: Can we make request validation more eager? What happens if a redirect target is malformed?

okhttp/src/main/java/com/squareup/okhttp/Connection.java
```diff
@@ -166,13 +173,14 @@ void connect(int connectTimeout, int readTimeout, int writeTimeout, Request tunn
...
-  void connectAndSetOwner(OkHttpClient client, Object owner, Request request) throws IOException {
+  void connectAndSetOwner(OkHttpClient client, Object owner, Request request)
+      throws RouteException {
     setOwner(owner);
 
     if (!isConnected()) {
-      Request tunnelRequest = tunnelRequest(request);
-      connect(client.getConnectTimeout(), client.getReadTimeout(),
-          client.getWriteTimeout(), tunnelRequest);
+      List<ConnectionSpec> connectionSpecs = client.getConnectionSpecs();
```
swankjesse: Why isn't this a property of the address? Shouldn't it be a part of the cache key for connection pooling?

okhttp/src/main/java/com/squareup/okhttp/Connection.java
```diff
@@ -360,12 +277,17 @@ void setProtocol(Protocol protocol) {
...
+      throws RouteException {
     if (!connected) throw new IllegalStateException("setTimeouts - not connected");
 
     // Don't set timeouts on shared SPDY connections.
     if (httpConnection != null) {
-      socket.setSoTimeout(readTimeoutMillis);
+      try {
+        socket.setSoTimeout(readTimeoutMillis);
+      } catch (IOException e) {
+        throw new RouteException(e);
```
swankjesse: Just curious: does this ever happen?

okhttp/src/main/java/com/squareup/okhttp/Connection.java
```diff
@@ -382,56 +304,10 @@ int recycleCount() {
...
-        default:
-          throw new IOException(
-              "Unexpected response code for CONNECT: " + response.code());
-      }
-    }
-  }
-
   @Override public String toString() {
+    if (!connected) {
+      return "Connection{unconnected}";
```
swankjesse: Shouldn't we still include the route & address? They're known. Maybe just add connected/unconnected as another property of toString ?

okhttp/src/main/java/com/squareup/okhttp/ConnectionSpec.java
```diff
@@ -23,44 +22,47 @@
...
+      CipherSuite.TLS_DHE_RSA_WITH_AES_256_CBC_SHA,
+      CipherSuite.TLS_RSA_WITH_AES_128_GCM_SHA256,
+      CipherSuite.TLS_RSA_WITH_AES_128_CBC_SHA,
+      CipherSuite.TLS_RSA_WITH_AES_256_CBC_SHA,
+      CipherSuite.TLS_RSA_WITH_3DES_EDE_CBC_SHA,
+      CipherSuite.TLS_RSA_WITH_RC4_128_SHA,
+      CipherSuite.TLS_RSA_WITH_RC4_128_MD5,
+  };
+
+  public static final ConnectionSpec TLS_1_2_AND_BELOW = new Builder(true)
```
swankjesse: Can you put `MODERN_TLS` back? End-users of the OkHttp API don't want to know what the difference is between TLS 1.1 and TLS 1.2 is; they're trusting us to make TLS decisions for them. We offer `MODERN_TLS` as a convenience for "State of the art for the current release of OkHttp".

okhttp/src/main/java/com/squareup/okhttp/ConnectionSpec.java
```diff
@@ -23,44 +22,47 @@
...
+      .cipherSuites(APPROVED_CIPHER_SUITES)
       .tlsVersions(TlsVersion.TLS_1_2, TlsVersion.TLS_1_1, TlsVersion.TLS_1_0)
       .supportsTlsExtensions(true)
       .build();
 
-  /** A backwards-compatible fallback connection for interop with obsolete servers. */
-  public static final ConnectionSpec COMPATIBLE_TLS = new Builder(MODERN_TLS)
+  public static final ConnectionSpec TLS_1_1_AND_BELOW = new Builder(TLS_1_2_AND_BELOW)
```
swankjesse: There's two changes going on with this PR. You're changing where the TLS fallback happens (from HttpEngine to Connection). That's a great change, though it's tricky.

The other change is a policy change. Today we attempt a connection supporting [TLS 1.0, TLS 1.1 and TLS 1.2], and if that fails we attempt a connection supporting [TLS 1.0] only. If we want to also make a policy change, we should do that, but as a separate PR.

Have you any evidence that real world HTTPS servers negotiate TLS 1.1 connections when offered [TLS 1.0, TLS 1.1] but not when offered [TLS 1.0, TLS 1.1, TLS 1.2] ? It's my understanding that only the lousiest TLS servers crash when offered a TLS version they don't understand. I would expect a diminishingly-small number of servers like that to exist in the wild.

Regardless, I think that discussion goes in a separate PR.

okhttp/src/main/java/com/squareup/okhttp/ConnectionSpec.java
```diff
@@ -117,14 +123,32 @@ public boolean supportsTlsExtensions() {
...
-    ConnectionSpec specToApply = supportedSpec(sslSocket);
+  /** Applies this spec to {@code sslSocket}. */
+  void apply(SSLSocket sslSocket, boolean isFallback) {
+    ConnectionSpec specToApply = supportedSpec(sslSocket, isFallback);
 
     sslSocket.setEnabledProtocols(specToApply.tlsVersions);
 
     String[] cipherSuitesToEnable = specToApply.cipherSuites;
-    if (route.shouldSendTlsFallbackIndicator) {
+    // null means "use default set".
```
swankjesse: Nice comment.

okhttp/src/main/java/com/squareup/okhttp/ConnectionSpec.java
```diff
@@ -143,35 +167,65 @@ void apply(SSLSocket sslSocket, Route route) {
...
+        .cipherSuites(cipherSuitesToEnable)
+        .tlsVersions(protocolsToEnable)
+        .build();
   }
 
   /**
-   * Returns a copy of this that omits cipher suites and TLS versions not
-   * enabled by {@code sslSocket}.
+   * Returns {@code true} if the socket, as currently configured, supports this ConnectionSpec.
+   * In order for a socket to be compatible the enabled cipher suites and protocols must match.
```
swankjesse: I read _match_ as being equal. I think you might want _intersect_ or _overlap_ ?

okhttp/src/main/java/com/squareup/okhttp/ConnectionSpec.java
```diff
@@ -143,35 +167,65 @@ void apply(SSLSocket sslSocket, Route route) {
...
+   *
+   * <p>For cipher suites, at least one of the {@link #cipherSuites() required cipher suites} must
+   * match the socket's enabled cipher suites. If there are no required cipher suites the socket
+   * must have at least one cipher suite enable.
```
swankjesse: enable –> enabled ?

okhttp/src/main/java/com/squareup/okhttp/ConnectionSpec.java
```diff
@@ -143,35 +167,65 @@ void apply(SSLSocket sslSocket, Route route) {
...
+   * connection spec containing [TLSv1.2, TLSv1.1, etc.] is only compatible if the socket has
+   * TLSv1.2 enabled.
    */
-  private ConnectionSpec supportedSpec(SSLSocket sslSocket) {
-    String[] cipherSuitesToEnable = null;
-    if (cipherSuites != null) {
-      String[] cipherSuitesToSelectFrom = sslSocket.getEnabledCipherSuites();
-      cipherSuitesToEnable =
-          Util.intersect(String.class, cipherSuites, cipherSuitesToSelectFrom);
+  public boolean isCompatible(SSLSocket socket) {
```
swankjesse: This is a fair bit of per-socket work. Is it worthwhile?

swankjesse: (in terms of lines of code, and work performed by the CPU)

swankjesse: ... the alternative would be to just make the call, and see if the handshake succeeds!

okhttp/src/main/java/com/squareup/okhttp/ConnectionSpec.java
```diff
@@ -143,35 +167,65 @@ void apply(SSLSocket sslSocket, Route route) {
...
+        if (contains(enabledCipherSuites, requiredCipher)) {
+          requiredCiphersEnabled = true;
+          break;
+        }
+      }
+    }
+    return requiredProtocolEnabled && requiredCiphersEnabled;
+  }
+
+  private static <T> boolean contains(T[] array, T value) {
```
swankjesse: Honest question: is null ever passed here?

okhttp/src/main/java/com/squareup/okhttp/TlsVersion.java
```diff
@@ -43,4 +43,12 @@ public static TlsVersion forJavaName(String javaName) {
     }
     throw new IllegalArgumentException("Unexpected TLS version: " + javaName);
   }
+
+  public static String[] javaNames(TlsVersion... tlsVersions) {
```
swankjesse: Need this be public? Maybe adding a single accessor javaName() would be sufficient for what's public?

okhttp/src/main/java/com/squareup/okhttp/internal/ConnectionSpecSelector.java
```diff
@@ -0,0 +1,132 @@
...
+import javax.net.ssl.SSLHandshakeException;
+import javax.net.ssl.SSLProtocolException;
+import javax.net.ssl.SSLSocket;
+
+/**
+ * Handles the connection spec fallback strategy: When a secure socket connection fails
+ * due to a handshake / protocol problem the connection may be retried with different protocols.
+ * Instances are stateful and should be created and used for a single connection attempt.
+ */
+public class ConnectionSpecSelector {
```
swankjesse: Neat. Can this be final?

okhttp/src/main/java/com/squareup/okhttp/internal/ConnectionSpecSelector.java
```diff
@@ -0,0 +1,132 @@
...
+  public ConnectionSpecSelector(List<ConnectionSpec> connectionSpecs) {
+    this.nextModeIndex = 0;
+    this.connectionSpecs = connectionSpecs;
+  }
+
+  /**
+   * Configure the supplied {@link SSLSocket} to connect to the specified host using an appropriate
+   * {@link ConnectionSpec}.
+   *
+   * @return the chosen {@link ConnectionSpec}
```
swankjesse: delete this line; I don't think it's useful

okhttp/src/main/java/com/squareup/okhttp/internal/ConnectionSpecSelector.java
```diff
@@ -0,0 +1,132 @@
...
+        tlsConfiguration = connectionSpec;
+        nextModeIndex = i + 1;
+        break;
+      }
+    }
+
+    if (tlsConfiguration == null) {
+      // This may be the first time a connection has been attempted and the socket does not support
+      // any the required protocols, or it may be a retry (but this socket supports fewer
+      // protocols than was suggested by a prioHttpr socket).
```
swankjesse: typo, prioHttpr

okhttp/src/main/java/com/squareup/okhttp/internal/ConnectionSpecSelector.java
```diff
@@ -0,0 +1,132 @@
...
+
+    if (e instanceof SSLHandshakeException) {
+      // If the problem was a CertificateException from the X509TrustManager,
+      // do not retry.
+      if (e.getCause() instanceof CertificateException) {
+        return false;
+      }
+    }
+
+    // On Android, SSLProtocolExceptions can be caused by TLS_FALLBACK_SCSV failures, which means we
```
swankjesse: I don't understand this comment. Is it pointing out a bug?

okhttp/src/main/java/com/squareup/okhttp/internal/ConnectionSpecSelector.java
```diff
@@ -0,0 +1,132 @@
...
+    return ((e instanceof SSLHandshakeException || e instanceof SSLProtocolException))
+        && isFallbackPossible;
+  }
+
+  /**
+   * Returns {@code true} if any later {@link ConnectionSpec} in the fallback strategy looks
+   * possible based on the supplied {@link SSLSocket}. It assumes that a future socket will have the
+   * same capabilities as the supplied socket.
+   */
+  private boolean isFallbackPossible(SSLSocket socket) {
```
swankjesse: Do we really need to look ahead here? Seems like `isCompatible()` is an expensive op, and the code would be simpler to just attempt and fail instead of checking first

okhttp/src/main/java/com/squareup/okhttp/internal/Internal.java
```diff
@@ -35,6 +38,12 @@
  */
 public abstract class Internal {
   public static final Logger logger = Logger.getLogger(OkHttpClient.class.getName());
+
+  public static void initializeInstanceForTests() {
```
swankjesse: This is quite unfortunate. Seems like tests shouldn't need to know about `Internal` !

okhttp/src/main/java/com/squareup/okhttp/internal/RouteDatabase.java
```diff
@@ -22,9 +22,8 @@
 /**
  * A blacklist of failed routes to avoid when creating a new connection to a
  * target address. This is used so that OkHttp can learn from its mistakes: if
- * there was a failure attempting to connect to a specific IP address, proxy
- * server or TLS mode, that failure is remembered and alternate routes are
- * preferred.
+ * there was a failure attempting to connect to a specific IP address or proxy
```
swankjesse: Good catch.

okhttp/src/main/java/com/squareup/okhttp/Call.java
swankjesse: ... ahh we can't, because sometimes it depends on DNS. Lame.

okhttp/src/main/java/com/squareup/okhttp/internal/http/RouteSelector.java
```diff
@@ -66,27 +61,22 @@
...
     this.client = client;
     this.routeDatabase = Internal.instance.routeDatabase(client);
     this.network = Internal.instance.network(client);
-    this.request = request;
 
     resetNextProxy(uri, address.getProxy());
   }
 
   public static RouteSelector get(Address address, Request request, OkHttpClient client)
       throws IOException {
```
swankjesse: It turns out that we should just get rid of this IOException completely, which gets rid of a lot of the motivation for RequestException.

This only comes up when the URL is malformed when converted to a URI. We could either detect that eagerly, or possibly improve `toUriLenient` to not throw.

I'd love to not need RequestException if we can avoid it. Our code is better spent on more important problems.

okhttp/src/main/java/com/squareup/okhttp/internal/http/SocketConnector.java
```diff
@@ -0,0 +1,278 @@
...
+          throw new IOException("Hostname " + address.getUriHost() + " not verified:"
+              + "\n    certificate: " + CertificatePinner.pin(cert)
+              + "\n    DN: " + cert.getSubjectDN().getName()
+              + "\n    subjectAltNames: " + OkHostnameVerifier.allSubjectAltNames(cert));
+        }
+
+        // Check that the certificate pinner is satisfied by the certificates presented.
+        address.getCertificatePinner().check(address.getUriHost(), handshake.peerCertificates());
+
+        return new ConnectedSocket(route, sslSocket, alpnProtocol, handshake);
```
swankjesse: Nice.

okhttp/src/main/java/com/squareup/okhttp/internal/http/SocketConnector.java
```diff
@@ -0,0 +1,278 @@
...
+    String proxyAuthorization = request.header("Proxy-Authorization");
+    if (proxyAuthorization != null) {
+      result.header("Proxy-Authorization", proxyAuthorization);
+    }
+
+    return result.build();
+  }
+
+  /**
+   * A connected socket with metadata.
```
swankjesse: Sadly, this is almost the exact same definition as `Connection`. Do we really need both?

swankjesse: Could this thing return a `Connection` ?

swankjesse:
There's a lot of code here, and I'm having some trouble differentiating some of the motivations & reasons behind some of the changes.

There are some things I think are awesome. I think having a `SocketConnector` class that builds `Connection` instances is awesome. It's dumb that `Connection` can be created in an unconnected state. Instead of having a stateful object that can be partially-formed, your `SocketConnector` class is better.

I don't quite grok the ceremony around socket compatibility testing. I like the current approach: attempt each option in a loop until one succeeds. Testing compatibility before and after seems like unnecessary work that we can avoid.

Can we eliminate RequestException?

okhttp/src/main/java/com/squareup/okhttp/internal/Internal.java
nfuller: Yes. This is required for any test that interacts with anything that touches Internal without first touching OkHttpClient. Having an explicit method seemed a way of explaining that. Without it you can get to a point where it tries to use Internal.instance and finds it null.

Not typically an issue unless you are running tests individually (as I was).

okhttp/src/main/java/com/squareup/okhttp/Connection.java
nfuller: I will return it to the Address, which means the "connection specs that could be used" becomes part of the connection pooling key and anything else that keys off of Address or things that hold it. e.g. Route, Connection.

I'm unsure whether that's actually a good thing or not for all cases. Route seems like it should be independent of the ConnectionSpecs to me with this change, but I can see why Address might want to have it which means something about the design is incorrect.

okhttp/src/main/java/com/squareup/okhttp/Connection.java
nfuller: Seems unlikely to me.

okhttp/src/main/java/com/squareup/okhttp/Connection.java
nfuller: Will remove

okhttp/src/main/java/com/squareup/okhttp/ConnectionSpec.java
nfuller: Will do. I think it's worth discussing that separately as a follow up.

okhttp/src/main/java/com/squareup/okhttp/Address.java
nfuller: Added back in for https://github.com/square/okhttp/pull/1388

nfuller:
I have produced a cut-down version that doesn't change the fallback behavior and has been rebased against ToT. This one shows where I was going.

https://github.com/square/okhttp/pull/1388

Some of the changes have been addressed, others I'd like to discuss and I'll copy your comments over there.

okhttp/src/main/java/com/squareup/okhttp/Call.java
nfuller: See comment https://github.com/nfuller/okhttp/commit/0b564359a1bd552c931af9098dd2e90c85824173#diff-b1d38a032a8d105e2ff59c621343b05cR275

okhttp/src/main/java/com/squareup/okhttp/Connection.java
nfuller: Added back in for #1388

okhttp/src/main/java/com/squareup/okhttp/Connection.java
nfuller: Removed in #1388

okhttp/src/main/java/com/squareup/okhttp/ConnectionSpec.java
nfuller: Mostly reverted in #1388

okhttp/src/main/java/com/squareup/okhttp/ConnectionSpec.java
nfuller: Reverted behavior for #1388

okhttp/src/main/java/com/squareup/okhttp/ConnectionSpec.java
nfuller: ConnectionSpec behavior for isCompatible(SSLSocket) should now do what it used to in #1388

okhttp/src/main/java/com/squareup/okhttp/ConnectionSpec.java
nfuller: Fixed in #1388

okhttp/src/main/java/com/squareup/okhttp/ConnectionSpec.java
nfuller: This method is used in two places:
1) To judge whether a socket is compatible with a given ConnectionSpec
2) To judge whether a socket is _going to be_ compatible with a future ConnectionSpec (in order to determine whether we need to create another plain socket).

In the first case we burn the plain socket in the attempt (if it fails), in the second we open sockets we might not be able to use.

It seemed better to trade CPU for I/O.

In this change (with the fallback modification) we also have more connection specs, and the isCompatible() is important to skip over those that don't apply to satisfy the "socket must support the primary protocol" logic.

There was a obvious optimization I could make here (in #1388) but it was mostly "terminate early on success". To optimize further I could see us creating a bitmask fingerprint of ciphers / protocols and memoizing it rather than keep doing all the string comparisons (may not work for ciphers if we need to deal with ciphers that we don't know about).

okhttp/src/main/java/com/squareup/okhttp/ConnectionSpec.java
nfuller: AFAIK, no. The first arg comes from getEnabledProtocols() and getEnabledCipherSuites() which should always return an array, even if empty. I can make it more defensive though because this could be interpreted differently by different implementers of SSLSocket.

okhttp/src/main/java/com/squareup/okhttp/internal/ConnectionSpecSelector.java
nfuller: Done in #1388

okhttp/src/main/java/com/squareup/okhttp/internal/ConnectionSpecSelector.java
nfuller: Done in #1388, but I beefed up the method comment to make it clearer.

okhttp/src/main/java/com/squareup/okhttp/internal/ConnectionSpecSelector.java
nfuller: Fixed in #1388

okhttp/src/main/java/com/squareup/okhttp/internal/ConnectionSpecSelector.java
nfuller: Seem my response here: https://github.com/nfuller/okhttp/commit/0b564359a1bd552c931af9098dd2e90c85824173#diff-e4ffef89d58c7312e934db99192a4556R113

okhttp/src/main/java/com/squareup/okhttp/internal/ConnectionSpecSelector.java
nfuller: Response here: https://github.com/nfuller/okhttp/commit/0b564359a1bd552c931af9098dd2e90c85824173#diff-3cb5f046f3603e0f409df86106d842edR185

okhttp/src/main/java/com/squareup/okhttp/internal/http/RouteSelector.java
nfuller: Tried to cover the justification here: https://github.com/nfuller/okhttp/commit/0b564359a1bd552c931af9098dd2e90c85824173#diff-b1d38a032a8d105e2ff59c621343b05cR275

swankjesse:
Closing because this is obsoleted by https://github.com/square/okhttp/issues/1388

(Though I suspect there may be further changes along this theme.)

PR #1341, opened 2015-01-21
Trivial optimization for comparing ConnectionSpecs

nfuller:
ConnectionSpec instances will usually originate from
final statics, particularly ConnectionSpec.CLEARTEXT. This seems
like a no-brainer to avoid array comparisons.


okhttp/src/main/java/com/squareup/okhttp/ConnectionSpec.java
```diff
@@ -176,6 +176,7 @@ private ConnectionSpec supportedSpec(SSLSocket sslSocket) {
 
   @Override public boolean equals(Object other) {
     if (!(other instanceof ConnectionSpec)) return false;
+    if (other == this) return true;
```
JakeWharton: You can actually make this the first statement in the method for an even more optimized optimization!

swankjesse:
LGTM

PR #1347, opened 2015-01-23
Private cache improvements

nfuller:
See commit comment. Minor improvements to the cache strategy now that OkHttp is a private, not a shared cache.

### DIFF
--- a/okhttp/src/main/java/com/squareup/okhttp/CacheControl.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/CacheControl.java
@@ -34,6 +34,7 @@ public final class CacheControl {
   private final boolean noStore;
   private final int maxAgeSeconds;
   private final int sMaxAgeSeconds;
+  private final boolean isPrivate;
   private final boolean isPublic;
   private final boolean mustRevalidate;
   private final int maxStaleSeconds;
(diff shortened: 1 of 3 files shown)


okhttp/src/main/java/com/squareup/okhttp/CacheControl.java
```diff
@@ -257,6 +264,10 @@ private String headerValue() {
     return result.toString();
   }
 
+  public boolean isPrivate() {
```
swankjesse: nit: define this with the other accessors

swankjesse:
LGTM

okhttp/src/main/java/com/squareup/okhttp/internal/http/CacheStrategy.java
```diff
@@ -67,10 +67,13 @@ public static boolean isCacheable(Response response, Request request) {
       case HTTP_MOVED_TEMP:
       case HTTP_TEMP_REDIRECT:
         // These codes can only be cached with the right response headers.
+        // http://www.w3.org/Protocols/rfc2616/rfc2616-sec13.html#sec13.4
+        // s-maxage is not checked because OkHttp is a private cache that should ignore s-maxage as
+        // per http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.9.3
```
swankjesse: There's new RFCs! The new hotness for caching is here:

http://tools.ietf.org/html/rfc7234
(and friends)

nfuller:
Updated reference. The new RFC doesn't spell it out as neatly. I haven't read the newer RFC in any detail so please do confirm this conforms.

PR #1348, opened 2015-01-23
Reduce flakiness and document reasons for flakiness

nfuller:
Android has been receiving reports of some tests being flaky
on what are probably lower-spec devices.

This introduces delays into tests where sockets are being
poisoned after the entire response body has been written to
them *and* where there are follow-up requests.

This change also improves the documentation for the problematic
SocketPolicy values.

### DIFF
--- a/mockwebserver/src/main/java/com/squareup/okhttp/mockwebserver/SocketPolicy.java
+++ b/mockwebserver/src/main/java/com/squareup/okhttp/mockwebserver/SocketPolicy.java
@@ -16,7 +16,19 @@
 
 package com.squareup.okhttp.mockwebserver;
 
-/** What should be done with the incoming socket. */
+/**
+ * What should be done with the incoming socket.
+ *
+ * <p>Be careful when using values like {@link #DISCONNECT_AT_END}, {@link #SHUTDOWN_INPUT_AT_END}
+ * and {@link #SHUTDOWN_OUTPUT_AT_END} that close a socket after a response, and where there are
+ * follow-up requests. The client is unblocked and free to continue as soon as it has received the
+ * entire response body. If and when the client makes a subsequent request using a pooled socket the
+ * server may not have had time to close the socket. The socket will be closed at an indeterminate
+ * point before or during the second request. It may be closed after client has started sending the
+ * request body. If a request body is not retryable then the client may fail the request, making
+ * client behavior non-deterministic. Add delays in the client to improve the chances that the
+ * server has closed the socket before follow up requests are made.
+ */
 public enum SocketPolicy {
 
   /**
@@ -28,6 +40,8 @@ public enum SocketPolicy {
   /**
    * Close the socket after the response. This is the default HTTP/1.0
    * behavior.
+   *
+   * <p>See {@link SocketPolicy} for reasons why this can cause test flakiness and how to avoid it.
    */
   DISCONNECT_AT_END,
 
@@ -56,17 +70,21 @@ public enum SocketPolicy {
   /**
    * Shutdown the socket input after sending the response. For testing bad
    * behavior.
+   *
+   * <p>See {@link SocketPolicy} for reasons why this can cause test flakiness and how to avoid it.
    */
   SHUTDOWN_INPUT_AT_END,
 
   /**
    * Shutdown the socket output after sending the response. For testing bad
    * behavior.
+   *
+   * <p>See {@link SocketPolicy} for reasons why this can cause test flakiness and how to avoid it.
    */
   SHUTDOWN_OUTPUT_AT_END,
 
   /**
-   * Don't response to the request but keep the socket open. For testing
+   * Don't respond to the request but keep the socket open. For testing
    * read response header timeout issue.
    */
   NO_RESPONSE
--- a/okhttp-tests/src/test/java/com/squareup/okhttp/internal/http/URLConnectionTest.java
+++ b/okhttp-tests/src/test/java/com/squareup/okhttp/internal/http/URLConnectionTest.java
@@ -425,6 +425,11 @@ public final class URLConnectionTest {
     connection1.setReadTimeout(100);
     assertContent("This connection won't pool properly", connection1);
     assertEquals(0, server.takeRequest().getSequenceNumber());
+
+    // Give the server time to enact the socket policy if it's one that could happen after the
+    // client has received the response.
+    Thread.sleep(500);
+
     HttpURLConnection connection2 = client.open(server.getUrl("/b"));
     connection2.setReadTimeout(100);
     assertContent("This comes after a busted connection", connection2);
@@ -632,6 +637,10 @@ public final class URLConnectionTest {
     client.client().setHostnameVerifier(new RecordingHostnameVerifier());
 
     assertContent("abc", client.open(server.getUrl("/")));
+
+    // Give the server time to disconnect.
+    Thread.sleep(500);
+
     assertContent("def", client.open(server.getUrl("/")));
 
     Set<TlsVersion> tlsVersions =
@@ -1196,6 +1205,9 @@ public final class URLConnectionTest {
 
     // Seed the pool with a bad connection.
     assertContent("a", client.open(server.getUrl("/")));
+
+    // Give the server time to disconnect.
+    Thread.sleep(500);
 
     // This connection will need to be recovered. When it is, transparent gzip should still work!
     assertContent("b", client.open(server.getUrl("/")));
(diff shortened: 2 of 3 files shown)


mockwebserver/src/main/java/com/squareup/okhttp/mockwebserver/SocketPolicy.java
```diff
@@ -28,6 +28,11 @@
   /**
    * Close the socket after the response. This is the default HTTP/1.0
    * behavior.
+   *
+   * <p>Note: Be careful if this used in a test and it is not the final queued request. The client
```
swankjesse: For what it's worth, triggering this flakiness is the entire motivation here.

mockwebserver/src/main/java/com/squareup/okhttp/mockwebserver/SocketPolicy.java
```diff
@@ -56,17 +61,27 @@
...
+   * the server may not have had time to closed the socket. Add delays in the client to improve
+   * the chances that the server has closed the socket.
    */
   SHUTDOWN_INPUT_AT_END,
 
   /**
    * Shutdown the socket output after sending the response. For testing bad
    * behavior.
+   *
+   * <p>Note: Be careful if this used in a test and it is not the final queued request. The client
```
swankjesse: Not a fan of copy-pasting comments. This trains the reader to not read the documentation, which is bad. Promote this comment to top-level documentation?

okhttp-tests/src/test/java/com/squareup/okhttp/internal/http/URLConnectionTest.java
```diff
@@ -425,6 +425,11 @@ private void testServerClosesOutput(SocketPolicy socketPolicy) throws Exception
     connection1.setReadTimeout(100);
     assertContent("This connection won't pool properly", connection1);
     assertEquals(0, server.takeRequest().getSequenceNumber());
+
+    // Give the server time to enact the socket policy if it's one that could happen after the
```
swankjesse: Ew. If this is flaky, why don't we recover from it automatically in OkHttp?

swankjesse:
You fixed the tests, but I think the right solution is fixing OkHttp itself.

mockwebserver/src/main/java/com/squareup/okhttp/mockwebserver/SocketPolicy.java
nfuller: The tests appear to be testing "what happens if a pooled connection has gone bad?", not "what happens if the server closes the connection at an arbitrary point during the request / response". 

You agree it's non-deterministic? Just to be clear in case my comments didn't explain:

I've seen this policy used in two ways:

1) Advertise a body as being N bytes, but disconnect before delivering the full N bytes.
2) Advertise a body as being N bytes, deliver the full N bytes, then (after time T) disconnect.

(1) is fine, since the client is blocked waiting for N bytes.
(2) is problematic if the test is trying to test "what happens if a pooled connection is no longer valid" : the client receives the N bytes, then the result depends on how long T is. Usually the test is expecting the connection to be closed by the time a second request is issued. However, because time T is varied, the test can move on to make another request and have the connection closed in the middle or after the client has issued the request.

If the server closes the connection before request 2 has been made the mock server needs to be set up to expect 2 requests. If it closes it after request 2 has been made and OkHttp retries, then it needs to be set up to expect 3 requests.

okhttp-tests/src/test/java/com/squareup/okhttp/internal/http/URLConnectionTest.java
nfuller: See comment above. It probably does try to recover, but the tests are no typically written to actually deal with the different points when the connection may be closed, only the common case when it happens before the 2nd request has been issued.

nfuller:
Updated the documentation.

It's possible there are also problems with OkHttp not retrying these when it should but I haven't investigated too much beyond working out a reason for why they were flaky.

The problems manifest as problems when reading the response:

01-09 11:23:44 I/VGLV8S7999999999: com.squareup.okhttp.internal.http.URLConnectionTest#postFailsWithChunkedRequestForLargeRequest FAIL 
java.net.SocketException: recvfrom failed: ECONNRESET (Connection reset by peer)
at libcore.io.IoBridge.maybeThrowAfterRecvfrom(IoBridge.java:592)
at libcore.io.IoBridge.recvfrom(IoBridge.java:556)
at java.net.PlainSocketImpl.read(PlainSocketImpl.java:485)
at java.net.PlainSocketImpl.access$000(PlainSocketImpl.java:37)
at java.net.PlainSocketImpl$PlainSocketInputStream.read(PlainSocketImpl.java:237)
at okio.Okio$2.read(Okio.java:113)
at okio.RealBufferedSource.indexOf(RealBufferedSource.java:147)
at okio.RealBufferedSource.readUtf8LineStrict(RealBufferedSource.java:94)
at com.squareup.okhttp.internal.http.HttpConnection.readResponse(HttpConnection.java:179)
at com.squareup.okhttp.internal.http.HttpTransport.readResponseHeaders(HttpTransport.java:101)
at com.squareup.okhttp.internal.http.HttpEngine.readResponse(HttpEngine.java:628)
at com.squareup.okhttp.internal.http.HttpURLConnectionImpl.execute(HttpURLConnectionImpl.java:388)
at com.squareup.okhttp.internal.http.HttpURLConnectionImpl.getResponse(HttpURLConnectionImpl.java:332)
at com.squareup.okhttp.internal.http.HttpURLConnectionImpl.getInputStream(HttpURLConnectionImpl.java:199)
at com.squareup.okhttp.internal.http.URLConnectionTest.assertContent(URLConnectionTest.java:3016)
at com.squareup.okhttp.internal.http.URLConnectionTest.assertContent(URLConnectionTest.java:3020)
at com.squareup.okhttp.internal.http.URLConnectionTest.reusedConnectionFailsWithPost(URLConnectionTest.java:2644)
at com.squareup.okhttp.internal.http.URLConnectionTest.postFailsWithChunkedRequestForLargeRequest(URLConnectionTest.java:2613)
at java.lang.reflect.Method.invoke(Native Method)
at java.lang.reflect.Method.invoke(Method.java:372)
at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:45)
at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:15)
at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:42)
at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:20)
at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:28)
at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:30)
at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:263)
at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:68)
at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:47)
at org.junit.runners.ParentRunner$3.run(ParentRunner.java:231)
at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:60)
at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:229)
at org.junit.runners.ParentRunner.access$000(ParentRunner.java:50)
at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:222)
at org.junit.runners.ParentRunner.run(ParentRunner.java:300)
at org.junit.runners.Suite.runChild(Suite.java:128)
at org.junit.runners.Suite.runChild(Suite.java:24)
at org.junit.runners.ParentRunner$3.run(ParentRunner.java:231)
at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:60)
at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:229)
at org.junit.runners.ParentRunner.access$000(ParentRunner.java:50)
at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:222)
at org.junit.runners.ParentRunner.run(ParentRunner.java:300)
at org.junit.runner.JUnitCore.run(JUnitCore.java:157)
at org.junit.runner.JUnitCore.run(JUnitCore.java:136)
at android.support.test.runner.AndroidJUnitRunner.onStart(AndroidJUnitRunner.java:245)
at android.app.Instrumentation$InstrumentationThread.run(Instrumentation.java:1853)
Caused by: android.system.ErrnoException: recvfrom failed: ECONNRESET (Connection reset by peer)
at libcore.io.Posix.recvfromBytes(Native Method)
at libcore.io.Posix.recvfrom(Posix.java:185)
at libcore.io.BlockGuardOs.recvfrom(BlockGuardOs.java:250)
at libcore.io.IoBridge.recvfrom(IoBridge.java:553)

swankjesse:
I still fear that we're fixing the tests when we should be fixing the implementation. The motivation for these wacky socket policies is to put the client in an awkward state, and to confirm that the client isn't flaky even when the server is flaky.

swankjesse:
Rebased & merged.

swankjesse:
Epic delay I know.

PR #1353, opened 2015-01-30
Fix a major bug where threads weren't being created.

swankjesse:
The LinkedBlockingQueue and core pool size of 0 do not interact well.
This prevents more threads from being created, harming overall
concurrency for async requests.


okhttp/src/main/java/com/squareup/okhttp/Dispatcher.java
```diff
@@ -23,6 +23,7 @@
 import java.util.Iterator;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.LinkedBlockingQueue;
```
JakeWharton: Delete import

JakeWharton:
LGTM

PR #1357, opened 2015-01-31
Do not allow timeouts less than 1ms.

### LINKED ISSUE #1356: Verify timeouts don't round to zero
The code for setting timeouts looks like
```java
  /**
   * Sets the default connect timeout for new connections. A value of 0 means no timeout.
   *
   * @see URLConnection#setConnectTimeout(int)
   */
  public final void setConnectTimeout(long timeout, TimeUnit unit) {
    if (timeout < 0) throw new IllegalArgumentException("timeout < 0");
    if (unit == null) throw new IllegalArgumentException("unit == null");
    long millis = unit.toMillis(timeout);
    if (millis > Integer.MAX_VALUE) throw new IllegalArgumentException("Timeout too large.");
    connectTimeout = (int) millis;
  }
```
The docs don't make clear that times less than 1ms will be rounded down to 0, which means no timeout.  This is obvious if you have an understanding of the Java socket apis, but is not to anyone else.  I suggest making any value between 1 and 999,999 nano seconds illegal.

JakeWharton:
Closes #1356.

### DIFF
--- a/okhttp-tests/src/test/java/com/squareup/okhttp/OkHttpClientTest.java
+++ b/okhttp-tests/src/test/java/com/squareup/okhttp/OkHttpClientTest.java
@@ -32,6 +32,7 @@ import java.net.URLConnection;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.TimeUnit;
 import javax.net.SocketFactory;
 import org.junit.After;
 import org.junit.Test;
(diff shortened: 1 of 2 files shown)


JakeWharton:
Upon second reading of the Javadoc it's a bit contradictory. Tweaked it slightly in b92c1b9db1c27b8035157a9016344db9e5f198ec.

PR #1364, opened 2015-02-05
ConnectionSpec: Allow custom specifications

raggi:
I would like to be able to create ConnectionSpec's, like so:

``` java
ConnectionSpec mySpec = new ConnectionSpec.Builder(ConnectionSpec.MODERN_TLS)
    .cipherSuites( ... strings or enums ... ).build();
```

 * Users can use the String... API to set CiperSuites and TLS versions, enabling forward-compatibility. You may wish to encourage users to continue to use the enums, but we already have a use case where the enum is insufficient.

Thank you in advance for your review, and thank you for the library!

### DIFF
--- a/okhttp-tests/src/test/java/com/squareup/okhttp/ConnectionSpecTest.java
+++ b/okhttp-tests/src/test/java/com/squareup/okhttp/ConnectionSpecTest.java
@@ -166,6 +166,16 @@ public final class ConnectionSpecTest {
     assertEquals(expectedCipherSet, expectedCipherSet);
   }
 
+  @Test
+  public void tls_stringCiphersAndVersions() throws Exception {
+    // Supporting arbitrary input strings allows users to enable suites and versions that are not
+    // yet known to the library, but are supported by the platform.
+    ConnectionSpec tlsSpec = new ConnectionSpec.Builder(ConnectionSpec.MODERN_TLS)
+        .cipherSuites("MAGIC-CIPHER")
+        .tlsVersions("TLS9k")
+        .build();
+  }
+
   private static Set<String> createSet(String... values) {
     return new LinkedHashSet<String>(Arrays.asList(values));
   }
--- a/okhttp/src/main/java/com/squareup/okhttp/ConnectionSpec.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/ConnectionSpec.java
@@ -238,12 +238,20 @@ public final class ConnectionSpec {
       for (int i = 0; i < cipherSuites.length; i++) {
         strings[i] = cipherSuites[i].javaName;
       }
-
-      return cipherSuites(strings);
+      this.cipherSuites = strings;
+      return this;
     }
 
-    Builder cipherSuites(String[] cipherSuites) {
-      this.cipherSuites = cipherSuites; // No defensive copy.
+    public Builder cipherSuites(String... cipherSuites) {
+      if (!tls) throw new IllegalStateException("no cipher suites for cleartext connections");
+
+      if (cipherSuites == null) {
+        this.cipherSuites = null;
+      } else {
+        // This makes a defensive copy!
+        this.cipherSuites = cipherSuites.clone();
+      }
+
       return this;
     }
 
@@ -255,12 +263,20 @@ public final class ConnectionSpec {
       for (int i = 0; i < tlsVersions.length; i++) {
         strings[i] = tlsVersions[i].javaName;
       }
-
-      return tlsVersions(strings);
+      this.tlsVersions = strings;
+      return this;
     }
 
-    Builder tlsVersions(String... tlsVersions) {
-      this.tlsVersions = tlsVersions; // No defensive copy.
+    public Builder tlsVersions(String... tlsVersions) {
+      if (!tls) throw new IllegalStateException("no TLS versions for cleartext connections");
+
+      if (tlsVersions == null) {
+        this.tlsVersions = null;
+      } else {
+        // This makes a defensive copy!
+        this.tlsVersions = tlsVersions.clone();
+      }
+
       return this;
     }
 


okhttp/src/main/java/com/squareup/okhttp/ConnectionSpec.java
```diff
@@ -242,7 +242,7 @@ public Builder cipherSuites(CipherSuite... cipherSuites) {
       return cipherSuites(strings);
     }
 
-    Builder cipherSuites(String[] cipherSuites) {
+    public Builder cipherSuites(String[] cipherSuites) {
```
JakeWharton: Did you mean to change this to varargs?

okhttp/src/main/java/com/squareup/okhttp/ConnectionSpec.java
```diff
@@ -219,7 +219,7 @@ private ConnectionSpec supportedSpec(SSLSocket sslSocket) {
     private String[] tlsVersions;
     private boolean supportsTlsExtensions;
 
-    Builder(boolean tls) {
+    public Builder(boolean tls) {
```
JakeWharton: This is an unfortunate constructor to expose publicly. At least with methods that take a boolean you're afforded a bit of context as to what the value corresponds to. I mean, would anyone even want to call this with `false` publicly?

raggi: Good question. I'd be happy with a zero-arg constructor that defaults to true. It seems that there's no use for the builder in the cleartext case today. Sound good?

okhttp/src/main/java/com/squareup/okhttp/ConnectionSpec.java
raggi: I'm happy either way, the most important thing is being able to do this.

Would you prefer varargs for a public API?

JakeWharton: Well I only ask because your test case is written with a single string which made me assume it's what you meant.

okhttp/src/main/java/com/squareup/okhttp/ConnectionSpec.java
JakeWharton: I would wait for @swankjesse to weigh in on the change as whole. This code is his territory!

okhttp/src/main/java/com/squareup/okhttp/ConnectionSpec.java
raggi: Hmm, and it passed I thought!

okhttp/src/main/java/com/squareup/okhttp/ConnectionSpec.java
raggi: I will do so, and fix the args issue you mentioned above. Thanks!

swankjesse: I'm quite reluctant to offer a naked ConnectionSpec. In the future we may add more stuff to ConnectionSpec: for example, four years ago we didn't have NPN, SNI, or SCSV. The defaults for these things will be different depending on whether you're going for maximum security (`MODERN_TLS`) or maximum compatibility (`COMPATIBLE_TLS`).

If we add an open ended default constructor, we're in trouble because we prevent the caller from choosing whether they prefer security or compatibility.

The current design is surprising because it makes you choose. That's deliberate!

okhttp/src/main/java/com/squareup/okhttp/ConnectionSpec.java
```diff
@@ -242,7 +242,7 @@ public Builder cipherSuites(CipherSuite... cipherSuites) {
       return cipherSuites(strings);
     }
 
-    Builder cipherSuites(String[] cipherSuites) {
+    public Builder cipherSuites(String... cipherSuites) {
       this.cipherSuites = cipherSuites; // No defensive copy.
```
swankjesse: Now that this is public, we're going to need to do a defensive `clone()`.

okhttp/src/main/java/com/squareup/okhttp/ConnectionSpec.java
```diff
@@ -259,7 +259,7 @@ public Builder tlsVersions(TlsVersion... tlsVersions) {
       return tlsVersions(strings);
     }
 
-    Builder tlsVersions(String... tlsVersions) {
+    public Builder tlsVersions(String... tlsVersions) {
       this.tlsVersions = tlsVersions; // No defensive copy.
```
swankjesse: similarly.

okhttp/src/main/java/com/squareup/okhttp/ConnectionSpec.java
```diff
@@ -259,7 +259,7 @@ public Builder tlsVersions(TlsVersion... tlsVersions) {
       return tlsVersions(strings);
```
swankjesse: ... and if you do add the defensive copy, you should change this method to avoid self-use because that will save an unnecessary allocation. You can just assign `this.tlsVersions` right here and `return this`.

okhttp/src/main/java/com/squareup/okhttp/ConnectionSpec.java
```diff
@@ -242,7 +242,7 @@ public Builder cipherSuites(CipherSuite... cipherSuites) {
       return cipherSuites(strings);
```
swankjesse: Similarly. Lose this self-use when you add the defensive copy.

swankjesse:
I like the string methods. But we have to be aggressive about immutability (these things are value objects that can be shared).

swankjesse:
CLA is good!

okhttp/src/main/java/com/squareup/okhttp/ConnectionSpec.java
raggi: Right, I'm just wondering if one wanted to use, say, CHACHA20, that's currently unavailable. The string API enables this. My proposal is that we'd encourage users to use the standard APIs rather than making decisions themselves, but still allow other users to make the selections they need if  they (think they) know that they're doing.

okhttp/src/main/java/com/squareup/okhttp/ConnectionSpec.java
raggi: varargs are mutable?

raggi:
Ok, a few changes for your review. I have done the following:

 * The introduced public Builder constructor now no longer accepts arguments. It defaults to TLS, as there's no reason presently to use the builder for insecure connections. (expecting some contention still)
 * Switched to varargs for the Strings API for consistency, and guard against explicit nulls (one test raised this).
 * Cloned the varargs and avoiding extra copies/allocations.

I feel like there's a lot of duplication now, although that may be fine if the preference is allocations and performance over repetition. They're at least consistent.

okhttp/src/main/java/com/squareup/okhttp/ConnectionSpec.java
```diff
@@ -233,34 +237,68 @@ public Builder(ConnectionSpec connectionSpec) {
...
+      this.cipherSuites = strings;
+      return this;
     }
 
-    Builder cipherSuites(String[] cipherSuites) {
-      this.cipherSuites = cipherSuites; // No defensive copy.
+    public Builder cipherSuites(String... cipherSuites) {
+      if (!tls) throw new IllegalStateException("no cipher suites for cleartext connections");
+
+      if (cipherSuites == null) {
```
JakeWharton: This case is strange. I would just throw an NPE. For a varargs to be null the user would have had to explicitly call it with `(String[]) null`.

okhttp/src/main/java/com/squareup/okhttp/ConnectionSpec.java
```diff
@@ -233,34 +237,68 @@ public Builder(ConnectionSpec connectionSpec) {
...
+        this.cipherSuites = new String[0];
+        return this;
+      }
+
+      // This makes a defensive copy!
+      String[] strings = new String[cipherSuites.length];
+      for (int i = 0; i < cipherSuites.length; i++) {
+        strings[i] = cipherSuites[i];
+      }
+      this.cipherSuites = strings;
```
JakeWharton: This could just be `this.cipherSuites = cipherSuites.clone();`

okhttp/src/main/java/com/squareup/okhttp/ConnectionSpec.java
raggi: I think that actually gets autoboxed. I'll track down which test does this.

raggi: So supportedSpec can call cipherSuites with null, which is where this needs to be covered. It wasn't a problem for the String[] API, but varargs box differently on nulls.

raggi: Given the varargs, I'm not sure the clone() is necessary, but my Java is still rusty.

JakeWharton: Yeah it's a tough call. The only time the call site has access to the array is when they explicitly pass an array, rather than use varargs. :cry:

raggi: Ah well. Clone done.

okhttp-tests/src/test/java/com/squareup/okhttp/ConnectionSpecTest.java
```diff
@@ -166,6 +166,16 @@ public void tls_explicitCiphers() throws Exception {
     assertEquals(expectedCipherSet, expectedCipherSet);
   }
 
+  @Test
+  public void tls_customCipersAndVersions() throws Exception {
+    // Supporting arbitrary input strings allows users to enable suites and versions that are not
+    // yet known to the library, but are supported by the platform.
+    ConnectionSpec tlsSpec = new ConnectionSpec.Builder()
```
swankjesse: Instead, please do `new ConnectionSpec.Builder(ConnectionSpec.MODERN_TLS)`. That way we can set TBA-future-defaults to either modern (maximum security) or compatible (maximum compatibility).

okhttp/src/main/java/com/squareup/okhttp/ConnectionSpec.java
```diff
@@ -219,6 +219,10 @@ private ConnectionSpec supportedSpec(SSLSocket sslSocket) {
     private String[] tlsVersions;
     private boolean supportsTlsExtensions;
 
+    public Builder() {
```
swankjesse: Drop this please! I want new instances to have to choose whether they're derived from `MODERN_TLS` or `COMPATIBLE_TLS`.

raggi:
Addressed comments, PTAL - thank you!

okhttp-tests/src/test/java/com/squareup/okhttp/ConnectionSpecTest.java
```diff
@@ -166,6 +166,16 @@ public void tls_explicitCiphers() throws Exception {
     assertEquals(expectedCipherSet, expectedCipherSet);
   }
 
+  @Test
+  public void tls_stringCiphersAndVersions() throws Exception {
+    // Supporting arbitrary input strings allows users to enable suites and versions that are not
+    // yet known to the library, but are supported by the platform.
+    ConnectionSpec tlsSpec = new ConnectionSpec.Builder(ConnectionSpec.MODERN_TLS)
```
swankjesse: Thank you

swankjesse:
Sweet.

PR #1388, opened 2015-02-11
Move TLS retries / fallback into Connection.

nfuller:
API changes:
1) Added: Address.getCertificatePinner()
2) Incompatible API change: Address constructor no longer
includes ConnectionSpecs. Removed getConnectionSpecs().
3) Added: ConnectionSpec.isCompatible(SSLSocket)
4) Added: TlsVersion.javaNames()

Implicit / semantic / internal changes:

1) Connection now handles all attempts to connect via a route
(effectively a {proxy, socket address} pair), rather than just
one attempt. i.e. Connection now handles all the TLS negotiation
fallbacks internally.

2) Route no longer deals with TLS versions. Individual TLS
failures are not counted against a Route. If no connection
attempts to a route were successful the failure is counted
against the route.

3) The code makes a distinction between when various
IOExceptions occur, with the intention making retries a bit
smarter. It is now more obvious which exceptions happen during
setup, connection, HTTP communication and thus which can be
retried and whether the request might have been sent.

### DIFF
--- a/okhttp-tests/src/test/java/com/squareup/okhttp/ConnectionSpecTest.java
+++ b/okhttp-tests/src/test/java/com/squareup/okhttp/ConnectionSpecTest.java
@@ -15,13 +15,8 @@
  */
 package com.squareup.okhttp;
 
-import com.squareup.okhttp.internal.http.AuthenticatorAdapter;
-
 import org.junit.Test;
 
-import java.net.InetSocketAddress;
-import java.net.Proxy;
-import java.net.ProxySelector;
 import java.util.Arrays;
 import java.util.LinkedHashSet;
 import java.util.Set;
@@ -34,14 +29,6 @@ import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 
 public final class ConnectionSpecTest {
-
-  private static final Proxy PROXY = Proxy.NO_PROXY;
-  private static final InetSocketAddress INET_SOCKET_ADDRESS =
-      InetSocketAddress.createUnresolved("host", 443);
-  private static final Address HTTPS_ADDRESS = new Address(
-      INET_SOCKET_ADDRESS.getHostString(), INET_SOCKET_ADDRESS.getPort(), null, null, null, null,
-      AuthenticatorAdapter.INSTANCE, PROXY, Arrays.asList(Protocol.HTTP_1_1),
-      Arrays.asList(ConnectionSpec.MODERN_TLS), ProxySelector.getDefault());
 
   @Test
   public void cleartextBuilder() throws Exception {
@@ -89,9 +76,8 @@ public final class ConnectionSpecTest {
         TlsVersion.TLS_1_1.javaName,
     });
 
-    Route route = new Route(HTTPS_ADDRESS, PROXY, INET_SOCKET_ADDRESS, tlsSpec,
-        false /* shouldSendTlsFallbackIndicator */);
-    tlsSpec.apply(socket, route);
+    assertTrue(tlsSpec.isCompatible(socket));
+    tlsSpec.apply(socket, false /* isFallback */);
 
     assertEquals(createSet(TlsVersion.TLS_1_2.javaName), createSet(socket.getEnabledProtocols()));
 
@@ -119,9 +105,8 @@ public final class ConnectionSpecTest {
         TlsVersion.TLS_1_1.javaName,
     });
 
-    Route route = new Route(HTTPS_ADDRESS, PROXY, INET_SOCKET_ADDRESS, tlsSpec,
-        true /* shouldSendTlsFallbackIndicator */);
-    tlsSpec.apply(socket, route);
+    assertTrue(tlsSpec.isCompatible(socket));
+    tlsSpec.apply(socket, true /* isFallback */);
 
     assertEquals(createSet(TlsVersion.TLS_1_2.javaName), createSet(socket.getEnabledProtocols()));
 
@@ -153,9 +138,8 @@ public final class ConnectionSpecTest {
         TlsVersion.TLS_1_1.javaName,
     });
 
-    Route route = new Route(HTTPS_ADDRESS, PROXY, INET_SOCKET_ADDRESS, tlsSpec,
-        true /* shouldSendTlsFallbackIndicator */);
-    tlsSpec.apply(socket, route);
+    assertTrue(tlsSpec.isCompatible(socket));
+    tlsSpec.apply(socket, true /* isFallback */);
 
     assertEquals(createSet(TlsVersion.TLS_1_2.javaName), createSet(socket.getEnabledProtocols()));
 
@@ -174,6 +158,52 @@ public final class ConnectionSpecTest {
         .cipherSuites("MAGIC-CIPHER")
         .tlsVersions("TLS9k")
         .build();
+  }
+
+  public void tls_missingRequiredCipher() throws Exception {
+    ConnectionSpec tlsSpec = new ConnectionSpec.Builder(true)
+        .cipherSuites(CipherSuite.TLS_RSA_WITH_RC4_128_MD5)
+        .tlsVersions(TlsVersion.TLS_1_2)
+        .supportsTlsExtensions(false)
+        .build();
+
+    SSLSocket socket = (SSLSocket) SSLSocketFactory.getDefault().createSocket();
+    socket.setEnabledProtocols(new String[] {
+        TlsVersion.TLS_1_2.javaName,
+        TlsVersion.TLS_1_1.javaName,
+    });
+
+    socket.setEnabledCipherSuites(new String[] {
+        CipherSuite.TLS_RSA_WITH_RC4_128_SHA.javaName,
+        CipherSuite.TLS_RSA_WITH_RC4_128_MD5.javaName,
+    });
+    assertTrue(tlsSpec.isCompatible(socket));
+
+    socket.setEnabledCipherSuites(new String[] {
+        CipherSuite.TLS_RSA_WITH_RC4_128_SHA.javaName,
+    });
+    assertFalse(tlsSpec.isCompatible(socket));
+  }
+
+  @Test
+  public void tls_missingTlsVersion() throws Exception {
+    ConnectionSpec tlsSpec = new ConnectionSpec.Builder(true)
+        .cipherSuites(CipherSuite.TLS_RSA_WITH_RC4_128_MD5)
+        .tlsVersions(TlsVersion.TLS_1_2)
+        .supportsTlsExtensions(false)
+        .build();
+
+    SSLSocket socket = (SSLSocket) SSLSocketFactory.getDefault().createSocket();
+    socket.setEnabledCipherSuites(new String[] {
+        CipherSuite.TLS_RSA_WITH_RC4_128_MD5.javaName,
+    });
+
+    socket.setEnabledProtocols(
+        new String[] { TlsVersion.TLS_1_2.javaName, TlsVersion.TLS_1_1.javaName });
+    assertTrue(tlsSpec.isCompatible(socket));
+
+    socket.setEnabledProtocols(new String[] { TlsVersion.TLS_1_1.javaName });
+    assertFalse(tlsSpec.isCompatible(socket));
   }
 
   private static Set<String> createSet(String... values) {
--- a/okhttp-tests/src/test/java/com/squareup/okhttp/internal/ConnectionSpecSelectorTest.java
+++ b/okhttp-tests/src/test/java/com/squareup/okhttp/internal/ConnectionSpecSelectorTest.java
@@ -1,0 +1,144 @@
+/*
+ * Copyright (C) 2015 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.okhttp.internal;
+
+import com.squareup.okhttp.ConnectionSpec;
+import com.squareup.okhttp.TlsVersion;
+
+import org.junit.Test;
+
+import java.io.IOException;
+import java.security.cert.CertificateException;
+import java.util.Arrays;
+import java.util.LinkedHashSet;
+import java.util.Set;
+import javax.net.ssl.SSLContext;
+import javax.net.ssl.SSLHandshakeException;
+import javax.net.ssl.SSLSocket;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+public class ConnectionSpecSelectorTest {
+
+  static {
+    Internal.initializeInstanceForTests();
+  }
+
+  private static final SSLContext sslContext = SslContextBuilder.localhost();
+
+  public static final SSLHandshakeException RETRYABLE_EXCEPTION = new SSLHandshakeException(
+      "Simulated handshake exception");
+
+  @Test
+  public void nonRetryableIOException() throws Exception {
+    ConnectionSpecSelector connectionSpecSelector =
+        createConnectionSpecSelector(ConnectionSpec.MODERN_TLS, ConnectionSpec.COMPATIBLE_TLS);
+    SSLSocket socket = createSocketWithEnabledProtocols(TlsVersion.TLS_1_1, TlsVersion.TLS_1_0);
+    connectionSpecSelector.configureSecureSocket(socket);
+
+    boolean retry = connectionSpecSelector.connectionFailed(
+        new IOException("Non-handshake exception"));
+    assertFalse(retry);
+    socket.close();
+  }
+
+  @Test
+  public void nonRetryableSSLHandshakeException() throws Exception {
+    ConnectionSpecSelector connectionSpecSelector =
+        createConnectionSpecSelector(ConnectionSpec.MODERN_TLS, ConnectionSpec.COMPATIBLE_TLS);
+    SSLSocket socket = createSocketWithEnabledProtocols(TlsVersion.TLS_1_1, TlsVersion.TLS_1_0);
+    connectionSpecSelector.configureSecureSocket(socket);
+
+    SSLHandshakeException trustIssueException =
+        new SSLHandshakeException("Certificate handshake exception");
+    trustIssueException.initCause(new CertificateException());
+    boolean retry = connectionSpecSelector.connectionFailed(trustIssueException);
+    assertFalse(retry);
+    socket.close();
+  }
+
+  @Test
+  public void retryableSSLHandshakeException() throws Exception {
+    ConnectionSpecSelector connectionSpecSelector =
+        createConnectionSpecSelector(ConnectionSpec.MODERN_TLS, ConnectionSpec.COMPATIBLE_TLS);
+    SSLSocket socket = createSocketWithEnabledProtocols(TlsVersion.TLS_1_1, TlsVersion.TLS_1_0);
+    connectionSpecSelector.configureSecureSocket(socket);
+
+    boolean retry = connectionSpecSelector.connectionFailed(RETRYABLE_EXCEPTION);
+    assertTrue(retry);
+    socket.close();
+  }
+
+  @Test
+  public void someFallbacksSupported() throws Exception {
+    ConnectionSpec sslV3 =
+        new ConnectionSpec.Builder(ConnectionSpec.MODERN_TLS)
+            .tlsVersions(TlsVersion.SSL_3_0)
+            .build();
+
+    ConnectionSpecSelector connectionSpecSelector = createConnectionSpecSelector(
+        ConnectionSpec.MODERN_TLS, ConnectionSpec.COMPATIBLE_TLS, sslV3);
+
+    TlsVersion[] enabledSocketTlsVersions = { TlsVersion.TLS_1_1, TlsVersion.TLS_1_0 };
+    SSLSocket socket = createSocketWithEnabledProtocols(enabledSocketTlsVersions);
+
+    // MODERN_TLS is used here.
+    connectionSpecSelector.configureSecureSocket(socket);
+    assertEnabledProtocols(socket, TlsVersion.TLS_1_1, TlsVersion.TLS_1_0);
+
+    boolean retry = connectionSpecSelector.connectionFailed(RETRYABLE_EXCEPTION);
+    assertTrue(retry);
+    socket.close();
+
+    // COMPATIBLE_TLS is used here.
+    socket = createSocketWithEnabledProtocols(enabledSocketTlsVersions);
+    connectionSpecSelector.configureSecureSocket(socket);
+    assertEnabledProtocols(socket, TlsVersion.TLS_1_0);
+
+    retry = connectionSpecSelector.connectionFailed(RETRYABLE_EXCEPTION);
+    assertFalse(retry);
+    socket.close();
+
+    // sslV3 is not used because SSLv3 is not enabled on the socket.
+  }
+
+  private static ConnectionSpecSelector createConnectionSpecSelector(
+      ConnectionSpec... connectionSpecs) {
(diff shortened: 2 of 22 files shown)


nfuller:
On #1339 Jesse wrote:
> There's a lot of code here, and I'm having some trouble differentiating some of the motivations & 
> reasons behind some of the changes.

This is a cutdown version of #1339 without the intentional behavior changes around TLS fallback. That one has been kept as a reference for where I intend to get to. One possibility if you don't like the ConnectionSpec.MODERN_TLS api changes there might be for Android to extend / modify ConnectionSpec to modify isCompatible() to the implement the behavior I'd like, without affecting OkHttp API. This change makes that a small delta. I'd like to avoid local Android changes where possible, though.

> There are some things I think are awesome. I think having a SocketConnector class that builds 
> Connection instances is awesome. It's dumb that Connection can be created in an unconnected 
> state. Instead of having a stateful object that can be partially-formed, your SocketConnector class is 
> better.

My original motivation was to keep Connection around to satisfy the existing contract for Connection, but move the actual "get me a socket" logic to another class. IIRC, I didn't quite manage to sever the relationship because of the need to have Connection as the owner for pooling. Maybe a follow up change could allow that?

> I don't quite grok the ceremony around socket compatibility testing. I like the current approach: 
> attempt each option in a loop until one succeeds. Testing compatibility before and after seems like 
> unnecessary work that we can avoid.

This is covered in the comments around the code in question. I'm trying to avoid unnecessary network comms (at the cost of some CPU). I/O with odious timeouts seems worse that CPU to me.

> Can we eliminate RequestException?

Also covered in comments.

okhttp-tests/src/test/java/com/squareup/okhttp/ConnectionSpecTest.java
```diff
@@ -176,6 +160,52 @@ public void tls_stringCiphersAndVersions() throws Exception {
         .build();
   }
 
+  public void tls_missingRequiredCipher() throws Exception {
```
swankjesse: Nice.

okhttp-tests/src/test/java/com/squareup/okhttp/internal/ConnectionSpecSelectorTest.java
```diff
@@ -0,0 +1,153 @@
...
+        createConnectionSpecSelector(ConnectionSpec.MODERN_TLS, ConnectionSpec.COMPATIBLE_TLS);
+    SSLSocket socket = createSocketWithEnabledProtocols(TlsVersion.TLS_1_1, TlsVersion.TLS_1_0);
+    try {
+      connectionSpecSelector.configureSecureSocket(socket);
+
+      boolean retry = connectionSpecSelector.connectionFailed(
+          new IOException("Non-handshake exception"));
+      assertFalse(retry);
+    } finally {
+      socket.close();
```
swankjesse: nit: as a general rule, I don't think it's useful to bother with `try/finally` within unit tests. We assume the tests are going to pass, and if failing tests leak resources, that's okay. (and in general, it's not worthwhile to worry about efficiency of incorrect code.)

okhttp-tests/src/test/java/com/squareup/okhttp/internal/ConnectionSpecSelectorTest.java
```diff
@@ -0,0 +1,153 @@
...
+  public void someFallbacksSupported() throws Exception {
+    ConnectionSpec sslV3 =
+        new ConnectionSpec.Builder(ConnectionSpec.MODERN_TLS)
+            .tlsVersions(TlsVersion.SSL_3_0)
+            .build();
+
+    ConnectionSpec[] connectionSpecs = {
+        ConnectionSpec.MODERN_TLS, ConnectionSpec.COMPATIBLE_TLS, sslV3
+    };
+    ConnectionSpecSelector connectionSpecSelector = createConnectionSpecSelector(connectionSpecs);
```
swankjesse: nit: should this be varargs? save some lines of code

okhttp-tests/src/test/java/com/squareup/okhttp/internal/ConnectionSpecSelectorTest.java
```diff
@@ -0,0 +1,153 @@
...
+
+    TlsVersion[] enabledSocketTlsVersions = { TlsVersion.TLS_1_1, TlsVersion.TLS_1_0 };
```
swankjesse: similarly, varargs?

okhttp-tests/src/test/java/com/squareup/okhttp/internal/ConnectionSpecSelectorTest.java
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  private SSLSocket createSocketWithEnabledProtocols(TlsVersion... tlsVersions) throws IOException {
+    SSLSocket socket = (SSLSocket) sslContext.getSocketFactory().createSocket();
+    socket.setEnabledProtocols(TlsVersion.javaNames(tlsVersions));
+    return socket;
+  }
+
+  private static void assertEnabledProtocols(SSLSocket socket, TlsVersion... required) {
+    Set<String> actual = new HashSet<String>(Arrays.asList(socket.getEnabledProtocols()));
```
swankjesse: Never use regular `HashedSet`. Always `LinkedHashSet`. The cost is negligible, and the predictability is invaluable.

okhttp-tests/src/test/java/com/squareup/okhttp/internal/http/RouteSelectorTest.java
```diff
@@ -379,7 +316,8 @@
     client.setProxy(Proxy.NO_PROXY);
     RouteSelector routeSelector = RouteSelector.get(address, httpsRequest, client);
 
-    dns.inetAddresses = makeFakeAddresses(255, 1);
+    final int numberOfAddresses = 2;
```
swankjesse: why final?

okhttp/src/main/java/com/squareup/okhttp/Connection.java
```diff
@@ -142,23 +130,41 @@ void closeIfOwnedBy(Object owner) throws IOException {
...
+      connectedSocket = socketConnector.connectTls(connectTimeout, readTimeout, writeTimeout,
+          request, route, connectionSpecs, connectionRetryEnabled);
     } else {
-      socket = new Socket(route.proxy);
+      // http:// communication.
+      if (!connectionSpecs.contains(ConnectionSpec.CLEARTEXT)) {
+        throw new RouteException(
+            new SocketException("CLEARTEXT communication not supported: " + connectionSpecs));
+      }
+      connectedSocket = socketConnector.connectClearText(connectTimeout, readTimeout, route);
```
swankjesse: nit: for consistency please rename `ClearText` to `Cleartext`. (Otherwise it'd have to be `ConnectionSpec.CLEAR_TEXT`, and it's not!)

okhttp/src/main/java/com/squareup/okhttp/Connection.java
```diff
@@ -361,12 +277,17 @@ void setProtocol(Protocol protocol) {
     this.protocol = protocol;
   }
 
-  void setTimeouts(int readTimeoutMillis, int writeTimeoutMillis) throws IOException {
+  void setTimeouts(int readTimeoutMillis, int writeTimeoutMillis)
+      throws RouteException {
```
swankjesse: I'm on board with RouteException. I was skeptical at first, but you've completely won me over. Good idea.

okhttp/src/main/java/com/squareup/okhttp/ConnectionSpec.java
```diff
@@ -143,35 +162,70 @@ void apply(SSLSocket sslSocket, Route route) {
...
-        .build();
+    String[] enabledProtocols = socket.getEnabledProtocols();
+    boolean requiredProtocolsEnabled = false;
+    for (String requiredProtocol : tlsVersions) {
+      if (contains(enabledProtocols, requiredProtocol)) {
+        requiredProtocolsEnabled = true;
+        break;
+      }
+    }
+
```
swankjesse: nit: newline

okhttp/src/main/java/com/squareup/okhttp/ConnectionSpec.java
```diff
@@ -143,35 +162,70 @@ void apply(SSLSocket sslSocket, Route route) {
...
+    boolean requiredCiphersEnabled;
+    if (cipherSuites == null) {
+      requiredCiphersEnabled = socket.getEnabledCipherSuites().length > 0;
+    } else {
+      String[] enabledCipherSuites = socket.getEnabledCipherSuites();
+      requiredCiphersEnabled = false;
+      for (String requiredCipher : cipherSuites) {
+        if (contains(enabledCipherSuites, requiredCipher)) {
```
swankjesse: nit: asymmetry with the two loops is lame. would be more regular to assign `requiredCiphersEnabled = true` and break, as above.

okhttp/src/main/java/com/squareup/okhttp/ConnectionSpec.java
```diff
@@ -143,35 +162,70 @@ void apply(SSLSocket sslSocket, Route route) {
...
+    }
+    return requiredCiphersEnabled;
+  }
+
+  private static <T> boolean contains(T[] array, T value) {
+    if (array == null) {
+      return false;
+    }
+    for (T arrayValue : array) {
+      if (value == arrayValue || (value != null && value.equals(arrayValue))) {
```
swankjesse: `Util.equal()`, if you want it.

okhttp/src/main/java/com/squareup/okhttp/ConnectionSpec.java
```diff
@@ -143,35 +162,70 @@ void apply(SSLSocket sslSocket, Route route) {
...
+      for (String requiredCipher : cipherSuites) {
+        if (contains(enabledCipherSuites, requiredCipher)) {
+          return true;
+        }
+      }
+    }
+    return requiredCiphersEnabled;
+  }
+
+  private static <T> boolean contains(T[] array, T value) {
```
swankjesse: nit: maybe add another method, `nonEmptyIntersection(T[] a, T[] b)` ? You'd be able to use that on 210.

okhttp/src/main/java/com/squareup/okhttp/ConnectionSpec.java
```diff
@@ -257,6 +311,9 @@ public Builder cipherSuites(String... cipherSuites) {
 
     public Builder tlsVersions(TlsVersion... tlsVersions) {
       if (!tls) throw new IllegalStateException("no TLS versions for cleartext connections");
+      if (tlsVersions.length == 0) {
+        throw new IllegalArgumentException("At least one TlsVersion is required");
```
swankjesse: Nice.

okhttp/src/main/java/com/squareup/okhttp/TlsVersion.java
```diff
@@ -43,4 +43,12 @@ public static TlsVersion forJavaName(String javaName) {
     }
     throw new IllegalArgumentException("Unexpected TLS version: " + javaName);
   }
+
+  public static String[] javaNames(TlsVersion... tlsVersions) {
```
swankjesse: nit: API wise, I think we need at least a method `javaName()` for a single instance call. Then this would be non-public, and possibly not in this file.

okhttp/src/main/java/com/squareup/okhttp/internal/ConnectionSpecSelector.java
```diff
@@ -0,0 +1,131 @@
...
+import javax.net.ssl.SSLHandshakeException;
+import javax.net.ssl.SSLProtocolException;
+import javax.net.ssl.SSLSocket;
+
+/**
+ * Handles the connection spec fallback strategy: When a secure socket connection fails
+ * due to a handshake / protocol problem the connection may be retried with different protocols.
+ * Instances are stateful and should be created and used for a single connection attempt.
+ */
+public final class ConnectionSpecSelector {
```
swankjesse: Dig.

okhttp/src/main/java/com/squareup/okhttp/internal/ConnectionSpecSelector.java
```diff
@@ -0,0 +1,131 @@
...
+
+  /**
+   * Configures the supplied {@link SSLSocket} to connect to the specified host using an appropriate
+   * {@link ConnectionSpec}. Returns the chosen {@link ConnectionSpec}, never {@code null}.
+   *
+   * @throws IOException if the socket does not support any of the TLS modes available
+   */
+  public ConnectionSpec configureSecureSocket(SSLSocket sslSocket) throws IOException {
+    ConnectionSpec tlsConfiguration = null;
+    for (int i = nextModeIndex; i < connectionSpecs.size(); i++) {
```
swankjesse: nit: we have a slightly-different form for our for loops:
```
for (int i = nextModeIndex, size = connectionSpecs.size(); i < size; i++) {
```

okhttp/src/main/java/com/squareup/okhttp/internal/ConnectionSpecSelector.java
```diff
@@ -0,0 +1,131 @@
...
+      ConnectionSpec connectionSpec = connectionSpecs.get(i);
+      if (connectionSpec.isCompatible(sslSocket)) {
+        tlsConfiguration = connectionSpec;
+        nextModeIndex = i + 1;
```
swankjesse: do we really need both `i` and `nextModeIndex` ?  Seems like just tracking `nextModeIndex` would be sufficient.

okhttp/src/main/java/com/squareup/okhttp/internal/ConnectionSpecSelector.java
```diff
@@ -0,0 +1,131 @@
...
+        break;
+      }
+    }
+
+    if (tlsConfiguration == null) {
```
swankjesse: Dig.

okhttp/src/main/java/com/squareup/okhttp/internal/ConnectionSpecSelector.java
```diff
@@ -0,0 +1,131 @@
...
+      // This may be the first time a connection has been attempted and the socket does not support
+      // any the required protocols, or it may be a retry (but this socket supports fewer
+      // protocols than was suggested by a prior socket).
+      throw new SocketException(
```
swankjesse: Curious: isn't this a RouteException? (Or do we wrap it with RouteException in the call site?)

okhttp/src/main/java/com/squareup/okhttp/internal/ConnectionSpecSelector.java
```diff
@@ -0,0 +1,131 @@
...
+      // do not retry.
+      if (e.getCause() instanceof CertificateException) {
+        return false;
+      }
+    }
+
+    // On Android, SSLProtocolExceptions can be caused by TLS_FALLBACK_SCSV failures, which means we
+    // retry those when we probably should not.
+    return ((e instanceof SSLHandshakeException || e instanceof SSLProtocolException))
+        && isFallbackPossible;
```
swankjesse: Could you compute `isFallbackPossible()` on demand?

okhttp/src/main/java/com/squareup/okhttp/internal/http/RequestException.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+package com.squareup.okhttp.internal.http;
+
+import java.io.IOException;
+
+/**
+ * Indicates a problem with interpreting a request. It may indicate there was a problem with the
+ * request itself, or the environment being used to interpret the request (network failure, etc.).
+ */
+public class RequestException extends Exception {
```
swankjesse: nit: final

okhttp/src/main/java/com/squareup/okhttp/internal/http/RouteException.java
```diff
@@ -0,0 +1,42 @@
...
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * An exception thrown to indicate a problem connecting via a single Route. Multiple attempts may
+ * have been made with alternative protocols, none of which were successful.
+ */
+public class RouteException extends Exception {
```
swankjesse: nit: final

okhttp/src/main/java/com/squareup/okhttp/internal/http/SocketConnector.java
```diff
@@ -0,0 +1,280 @@
...
+
+  public ConnectedSocket connectTls(int connectTimeout, int readTimeout,
+      int writeTimeout, Request request, Route route, List<ConnectionSpec> connectionSpecs,
+      boolean connectionRetryEnabled) throws RouteException {
+
+    Address address = route.getAddress();
+    ConnectionSpecSelector connectionSpecSelector = new ConnectionSpecSelector(connectionSpecs);
+    RouteException routeException = null;
+    do {
+      Socket socket = connectRawSocket(readTimeout, connectTimeout, route);
```
swankjesse: Neat.

swankjesse:
LGTM, with some nitpicks.

Sincere apologies for my delay on this. I've had a combination of personal & work things that have had me busier and less responsive than I want to be.

okhttp-tests/src/test/java/com/squareup/okhttp/internal/ConnectionSpecSelectorTest.java
nfuller: Fixed in next version.

okhttp-tests/src/test/java/com/squareup/okhttp/internal/ConnectionSpecSelectorTest.java
nfuller: Fixed in next version.

okhttp-tests/src/test/java/com/squareup/okhttp/internal/ConnectionSpecSelectorTest.java
nfuller: Fixed in next version. Duplicated the TLS versions on the line below.

okhttp-tests/src/test/java/com/squareup/okhttp/internal/ConnectionSpecSelectorTest.java
nfuller: Sorry. This seems like a hard habit for me to break. Will try.

okhttp-tests/src/test/java/com/squareup/okhttp/internal/http/RouteSelectorTest.java
nfuller: Typically I do this for method-scoped constants. i.e. it is a constant that I've pulled out to give it a clear name and/or because it's used in multiple places that must agree, but which don't make sense outside of the test. Where coding conventions permit I make them upper-case too. By marking it final the reader can mentally dismiss the chance of it changing.

In this case, the 1 in the code that was replaced was unclear as to its meaning.

This seemed preferable to having:

2 /* numberOfAddresses */ in a couple of places.

okhttp/src/main/java/com/squareup/okhttp/Connection.java
nfuller: Time will tell!

okhttp/src/main/java/com/squareup/okhttp/ConnectionSpec.java
nfuller: Fixed in next version.

okhttp/src/main/java/com/squareup/okhttp/ConnectionSpec.java
nfuller: Agreed! Fixed in next version.

okhttp/src/main/java/com/squareup/okhttp/ConnectionSpec.java
nfuller: Fixed in next version.

okhttp/src/main/java/com/squareup/okhttp/ConnectionSpec.java
nfuller: Done in next version. It's still as inefficient so take a look.

okhttp/src/main/java/com/squareup/okhttp/TlsVersion.java
nfuller: Fixed in next version. Added javaName() and moved javaNames() to ConnectionSpecSelectorTest.

okhttp/src/main/java/com/squareup/okhttp/internal/ConnectionSpecSelector.java
nfuller: Fixed in next version.

okhttp/src/main/java/com/squareup/okhttp/internal/ConnectionSpecSelector.java
nfuller: Yes, I see what you're saying, but it ends up looking like this:

    for (int size = connectionSpecs.size(); nextModeIndex < size; nextModeIndex++) {
      ConnectionSpec connectionSpec = connectionSpecs.get(nextModeIndex);
      if (connectionSpec.isCompatible(sslSocket)) {
        tlsConfiguration = connectionSpec;
        nextModeIndex++;
        break;
      }
    }

That's a pretty unusual-looking loop. I'm sure it can be improved, though.

okhttp/src/main/java/com/squareup/okhttp/internal/ConnectionSpecSelector.java
nfuller: Potentially. We lose information about what the initial enabled set of protocols / ciphers is (having changed them), so we'd need to persist that instead of isFallbackPossible.

Not changed this.

okhttp/src/main/java/com/squareup/okhttp/internal/http/RequestException.java
nfuller: Fixed in next version.

okhttp/src/main/java/com/squareup/okhttp/internal/http/RouteException.java
nfuller: Fixed in next version.

okhttp/src/main/java/com/squareup/okhttp/Connection.java
nfuller: Fixed in next version.

okhttp/src/main/java/com/squareup/okhttp/internal/ConnectionSpecSelector.java
nfuller: Yes: Any IOException at this level is turned into a RouteException in the caller, SocketConnector.

okhttp/src/main/java/com/squareup/okhttp/internal/ConnectionSpecSelector.java
swankjesse: Sounds good.

swankjesse:
LGTM

Any idea on the failing tests? If they're flakes, we should merge this and fix those independently.

nfuller:
I was ready to dismiss the failures as flakes because they pass for me, but the failures are consistent across both, suggesting I broke them. I'll see if I can work it out.

swankjesse:
> Question 1: Should RouteException cause the whole thing to give up in all cases?

Agreed.

> Question 2: Should a "SSLHandshakeException wrapping a CertificateException" and SSLPeerUnverifiedException cause the whole thing to give up?

Agreed. The only SSL failure we want to recover from is the fallback case.


> Any suggestions for refactoring to reduce the duplication?

The original, lousy strategy for this was to put all recovery in one place, at the top-level.

swankjesse:
Sorry for the awful delay.

swankjesse:
Thanks for the code!

okhttp/src/main/java/com/squareup/okhttp/internal/http/HttpEngine.java
```diff
@@ -393,8 +409,75 @@ public Connection getConnection() {
...
+    IOException ioe = e.getLastConnectException();
+
+    // TODO(nfuller): This is the same logic as in ConnectionSpecSelector
+    // If there was a protocol problem, don't recover.
+    if (ioe instanceof ProtocolException) {
+      return false;
+    }
+
+    // If there was an interruption or timeout, don't recover.
+    if (ioe instanceof InterruptedIOException) {
```
narayank: What was the purpose of this change (lines 452 - 455) ? This appears to have caused a regression on some devices where there are (network, not device) issues with IPV6 addresses, but where IPV4 addresses work ok.

PR #1393, opened 2015-02-12
A way to catch that cleartext HTTP request was attempted.

### LINKED ISSUE #1392: Provide a way to catch that cleartext HTTP traffic was attempted
This is a follow-up to feature request 1058 where okhttp offers a way to disallow cleartext traffic. Just blocking cleartext traffic is insufficient for some apps. Some app developers want to not only block cleartext traffic but also report to their servers that their app attempted cleartext traffic. This is so that regressions from HTTPS to cleartext HTTP can be investigated and fixed quicker.

At the moment, when okhttp blocks a cleartext HTTP request, it throws a new SocketException("No route to ... ; exhausted connection specs: ..."). Please provide a way to differentiate the exception thrown in this case from other IOExceptions. For example, when blocking cleartext HTTP okhttp could throw its own special subclass of IOException or perhaps a java.lang.SecurityException or an IOException caused by a SecurityException.

I'm happy to implement the changes once a decision reached about what exception to throw.

klyubin:
This makes the HTTP stack throw java.net.UnknownServiceException if
a request is attempted using a protocol which matches none of the
configured connection specs. For example, this will happen if a cleartext
HTTP request is attempted when ConnectionSpec.CLEARTEXT is not
configured.

Fixes #1392

### DIFF
--- a/okhttp-tests/src/test/java/com/squareup/okhttp/CallTest.java
+++ b/okhttp-tests/src/test/java/com/squareup/okhttp/CallTest.java
@@ -31,7 +31,7 @@ import java.io.InputStream;
 import java.net.CookieManager;
 import java.net.HttpCookie;
 import java.net.HttpURLConnection;
-import java.net.SocketException;
+import java.net.UnknownServiceException;
 import java.net.URL;
 import java.security.cert.Certificate;
 import java.util.Arrays;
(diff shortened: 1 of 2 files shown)


swankjesse:
Thanks!

Issue #1395, opened 2015-02-12
Handilng authentication exception


busylee999:
I faced the problem, now realization of Client use this scenario:
1. make request, if error (401 - Unauthenticate, interesting for issue context)
2. if 401 code of previous response, invoke Authenticator for specifying credentials and follow up request.

Consider credentials is wrong (for example user specified wrong credentials).
than next request will return 401 again, and this repeats MAX_FOLLOW_UPS times. After all it throws ProtocolException("Too many follow-up requests: ").

And question is: "How I can detect wrong credentials?"

There is urge way for client uses this HTTPClient is to detect count of equal requests or something the same now. It seems as spike.

Is there any different way to handle bad credentials?

swankjesse:
You can test your credentials against the credentials of the unauthenticated request:
```
      @Override public Request authenticate(Proxy proxy, Response response) {
        System.out.println("Authenticating for response: " + response);
        System.out.println("Challenges: " + response.challenges());
        String credential = Credentials.basic("jesse", "password1");

        if (credential.equals(response.request().header("Authorization"))) {
          return null; // If we already failed with these credentials, don't retry.
        }

        return response.request().newBuilder()
            .header("Authorization", credential)
            .build();
      }
```

swankjesse:
I've added this example and another to our recipes doc.

PR #1396, opened 2015-02-12
fix authentication issue, 401 response code have never returns

busylee999:
https://github.com/square/okhttp/issues/1395

### DIFF
--- a/okhttp/src/main/java/com/squareup/okhttp/Call.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/Call.java
@@ -282,7 +282,7 @@ public class Call {
       }
 
       Response response = engine.getResponse();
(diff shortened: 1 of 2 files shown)


swankjesse:
I don't think this works for applications that prompt for a password interactively. This is not backwards compatible.

busylee999:
Yep, It seems not good. But I think it is bad never return 401 error. I faced this problem when I can not detect then user has bad credentials.

swankjesse:
I think we need to surface the attempt number in the Response object.

busylee999:
If we just pass number into response, it leads to excess attempts to connect using wrong credentials. I think it is better pass response with correct 401 error.

Consider we pass number of attempts. Where do you suppose user to handle it and do some thing?

PR #1445, opened 2015-02-19
Streaming Response

ipsi:
Having been attempting to do some integration testing against a service that was using Twitters streaming API, I was hoping to use MockHTTP to do that, but it didn't work out so well, since as soon as the connection was closed, everything fell over in a horrible fashion. As such, I have made some modifications so that it is possible to create a connection to the mock service and have that connection stay open until we're done with it, and to send a "Keep Alive" chunk (basically a blank string) every so often.

This is my first draft of the code, so it might be a little wonky.

### DIFF
--- a/mockwebserver/src/main/java/com/squareup/okhttp/mockwebserver/MockResponse.java
+++ b/mockwebserver/src/main/java/com/squareup/okhttp/mockwebserver/MockResponse.java
@@ -17,6 +17,8 @@ package com.squareup.okhttp.mockwebserver;
 
 import com.squareup.okhttp.Headers;
 import com.squareup.okhttp.internal.ws.WebSocketListener;
+
+import java.nio.charset.Charset;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
@@ -24,7 +26,7 @@ import okio.Buffer;
 
 /** A scripted response to be replayed by the mock web server. */
 public final class MockResponse implements Cloneable {
-  private static final String CHUNKED_BODY_HEADER = "Transfer-encoding: chunked";
+  private static final String CHUNKED_BODY_HEADER = "transfer-encoding: chunked";
 
   private String status = "HTTP/1.1 200 OK";
   private Headers.Builder headers = new Headers.Builder();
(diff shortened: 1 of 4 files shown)


mockwebserver/src/main/java/com/squareup/okhttp/mockwebserver/MockResponse.java
```diff
@@ -164,6 +166,26 @@ public MockResponse setChunkedBody(Buffer body, int maxChunkSize) {
     return this;
   }
 
+    /**
+     * Provide a list of chunks to be streamed back to the client.
+     *
```
swankjesse: nit: configure your editor to use 2 space indents & avoid tabs.

mockwebserver/src/main/java/com/squareup/okhttp/mockwebserver/MockResponse.java
```diff
@@ -164,6 +166,26 @@ public MockResponse setChunkedBody(Buffer body, int maxChunkSize) {
...
+     * @param chunks
```
swankjesse: delete this.

mockwebserver/src/main/java/com/squareup/okhttp/mockwebserver/MockResponse.java
```diff
@@ -164,6 +166,26 @@ public MockResponse setChunkedBody(Buffer body, int maxChunkSize) {
...
+     * @return
```
swankjesse: delete this.

mockwebserver/src/main/java/com/squareup/okhttp/mockwebserver/MockResponse.java
```diff
@@ -164,6 +166,26 @@ public MockResponse setChunkedBody(Buffer body, int maxChunkSize) {
...
+     */
+  public MockResponse setChunkedBody(List<String> chunks) {
+    removeHeader("Content-Length");
+    headers.add(CHUNKED_BODY_HEADER);
+      Buffer bytesOut = new Buffer();
+    for (String s : chunks) {
+        bytesOut.writeUtf8(Integer.toHexString(s.getBytes(Charset.forName("UTF-8")).length));
```
swankjesse: Please add two methods, one that takes a `List<String>` and another that takes a `List<ByteString>`, and rewrite one to call the other. Then we can also rewrite the overload on line 151 to call through to the method that takes a `List<ByteString>`, and only have _one_ place where we call `Integer.toHexString()`.

mockwebserver/src/main/java/com/squareup/okhttp/mockwebserver/MockWebServer.java
```diff
@@ -701,8 +701,48 @@ private void writeHttpResponse(Socket socket, BufferedSink sink, MockResponse re
 
     Buffer body = response.getBody();
     if (body == null) return;
-    sleepIfDelayed(response);
-    throttledTransfer(response, socket, body, sink, Long.MAX_VALUE);
+    if (response.getSocketPolicy() == SocketPolicy.STREAM_REPEAT_FINAL_CHUNK) {
```
swankjesse: I don't think this makes much sense for general use.

swankjesse: .. . . can't you accomplish the same with the existing API, using throttled transfer and a sufficiently large response?

swankjesse:
I don't think we want this feature. You can already accomplish it with the throttled transfer, right?

ipsi:
@swankjesse Yes, you can, and this code only works by luck, so will close the pull request. I may raise another, smaller, one for adding some utility changes around this, though, and maybe some examples of how to do it, as it took me a while to click that it would always pause after writing out a set if bytes if the body was throttled, even if you'd written out the entire response - I'm not sure if that was intentional or not, which is what I needed: some way to keep the connection open for a second or two after writing out the body to give everything a chance to process.

swankjesse:
Glad it's working for you. Thanks @ipsi. Looking forward to the follow up.

Issue #1349, opened 2015-01-27
alternatives to content-length -1 MultipartBuilder


felipeska:
I am doing the following test:
````
 @Test
  public void testMultipartBuilder() throws Exception {
    MockWebServer mockWebServer = new MockWebServer();
    mockWebServer.play();
    OkHttpClient client = new OkHttpClient();
    mockWebServer.setDispatcher(new Dispatcher() {
      @Override
      public MockResponse dispatch(RecordedRequest request) throws InterruptedException {
        if (request.getHeader("Content-Length") == null) {
          return new MockResponse()
                  .setStatus("HTTP/1.1 411")
                  .setResponseCode(411)
                  .addHeader("content-type: text/plain; charset=utf-8")
                  .setBody("Length Required");
        } else
          return new MockResponse()
                  .setStatus("HTTP/1.1 200")
                  .setResponseCode(Constants.HTTP_OK)
                  .addHeader("content-type: text/plain; charset=utf-8")
                  .setBody("server: nginx");
      }
    });

    RequestBody requestBody = new MultipartBuilder()
            .type(MultipartBuilder.FORM)
            .addFormDataPart("title", "Square Logo")
            .addFormDataPart("image", null,
                    RequestBody.create(MediaType.parse("image/png"),
                            new File("website/static/logo-square.png")))
            .build();
    com.squareup.okhttp.Request request = new com.squareup.okhttp.Request.Builder()
            .url(mockWebServer.getUrl("/images/"))
            .post(requestBody)
            .build();
    com.squareup.okhttp.Response response = client.newCall(request).execute();
    assertTrue(response.isSuccessful());
  }
```
 the  ```contentLength()``` from ```MultipartBuilder``` returns the number of bytes that will be written to out in a call to ```writeTo```, or -1 if that count is unknown.

As this happens it does not send the header  ``Content-Length``, and my need this header to process the request.

what options exist to obtain the content-length?

JakeWharton:
Write the multipart body to a `Buffer`. Then use that `Buffer` as the body for the actual request.

felipeska:
Can be?
``` 
Buffer buffer = new Buffer();
requestBody.writeTo(buffer);
com.squareup.okhttp.Request request = new com.squareup.okhttp.Request.Builder()
            .url(mockWebServer.getUrl("/images/"))
            .post(RequestBody.create(MultipartBuilder.FORM,buffer.readByteArray()))
            .build();
```

JakeWharton:
[...]

PR #1449, opened 2015-02-23
Compute the content length of multipart bodies.

swankjesse:
Closes https://github.com/square/okhttp/issues/1349


okhttp-tests/src/test/java/com/squareup/okhttp/MultipartBuilderTest.java
```diff
@@ -46,7 +46,8 @@ public void onePartRequired() throws Exception {
 
     Buffer buffer = new Buffer();
     requestBody.writeTo(buffer);
-    assertEquals(-1L, requestBody.contentLength());
+    assertEquals(53, requestBody.contentLength());
```
swankjesse: reverted from previous!

PR #1456, opened 2015-02-24
Test that DELETE can have a body.

swankjesse:


edenman:
LG

JakeWharton:
A wild @edenman appears!

PR #1457, opened 2015-02-25
request create from byte[] range

longbai:


longbai:
have signed CLA.

okhttp/src/main/java/com/squareup/okhttp/RequestBody.java
```diff
@@ -74,6 +74,26 @@ public static RequestBody create(final MediaType contentType, final byte[] conte
     };
   }
 
+  /** Returns a new request body that transmits {@code content}. */
```
swankjesse: Can you fix the overload on line 60 to call through to this? No sense in having more code than we need.

swankjesse:
Works for me. Just one nitpick.

longbai:
done.

swankjesse:
Awesome. (Sorry for sitting on this!)

CLA looks good. My last request: do you mind squashing this into a single commit? See http://rebaseandsqua.sh for instructions.

longbai:
done

swankjesse:
Thanks!

PR #1464, opened 2015-03-03
unsigned short frame length bug fixed

dmide:


okhttp/src/main/java/com/squareup/okhttp/internal/ws/WebSocketReader.java
```diff
@@ -135,7 +135,8 @@ private void readHeader() throws IOException {
     // Get frame length, optionally reading from follow-up bytes if indicated by special values.
     frameLength = b1 & B1_MASK_LENGTH;
     if (frameLength == PAYLOAD_SHORT) {
-      frameLength = source.readShort();
+      //we may receive an unsigned short
+      frameLength = 0xffff & source.readShort();
     } else if (frameLength == PAYLOAD_LONG) {
       frameLength = source.readLong();
```
JakeWharton: Long is also unsigned in the spec. This fix will need to be more robust (and have tests). I'll keep this open to track the bug.

JakeWharton:
Thanks for finding this!

dmide:
glad to help

PR #1465, opened 2015-03-03
Payload length continuations are unsigned values.

JakeWharton:
The short and long continuations of payload length should be treated as unsigned values. Due to the sheer size of an unsigned long, any value over the signed Long.MAX_VALUE is forbidden.

Closes #1464.

### DIFF
--- a/okhttp-tests/src/test/java/com/squareup/okhttp/internal/ws/WebSocketReaderTest.java
+++ b/okhttp-tests/src/test/java/com/squareup/okhttp/internal/ws/WebSocketReaderTest.java
@@ -123,6 +123,28 @@ public class WebSocketReaderTest {
     callback.assertTextMessage("Hello");
   }
 
+  @Test public void clientFramePayloadShort() throws IOException {
+    data.write(ByteString.decodeHex("817E000548656c6c6f")); // Hello
+    clientReader.processNextFrame();
+    callback.assertTextMessage("Hello");
+  }
+
+  @Test public void clientFramePayloadLong() throws IOException {
+    data.write(ByteString.decodeHex("817f000000000000000548656c6c6f")); // Hello
+    clientReader.processNextFrame();
+    callback.assertTextMessage("Hello");
+  }
+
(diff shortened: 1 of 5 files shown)


okhttp-tests/src/test/java/com/squareup/okhttp/internal/ws/WebSocketReaderTest.java
```diff
@@ -123,6 +123,27 @@
...
+  @Test public void clientFramePayloadLong() throws IOException {
+    data.write(ByteString.decodeHex("817f000000000000000548656c6c6f")); // Hello
+    clientReader.processNextFrame();
+    callback.assertTextMessage("Hello");
+  }
+
+  @Test public void clientFramePayloadTooLongThrows() throws IOException {
+    data.write(ByteString.decodeHex("817f8000000000000000"));
+    try {
+      clientReader.processNextFrame();
```
swankjesse: fail

okhttp-tests/src/test/java/com/squareup/okhttp/internal/ws/WebSocketReaderTest.java
```diff
@@ -123,6 +123,27 @@
...
+    } catch (IllegalStateException e) {
```
swankjesse: Is this the right exception? I would expect a ProtocolException

swankjesse:
+1

okhttp-tests/src/test/java/com/squareup/okhttp/internal/ws/WebSocketReaderTest.java
JakeWharton: Dunno. It's not technically a protocol problem.

okhttp-tests/src/test/java/com/squareup/okhttp/internal/ws/WebSocketReaderTest.java
JakeWharton: fail() fail

PR #1469, opened 2015-03-06
null out Request.Builder.url when setting urlString

nashira:
If you have a request that has called `Request.url` a URL will be initialized and cached on the object. If you then call `Request.newBuilder` on the same request object it will copy over `url` and `urlString` to the builder.  If you then call `Request.Builder.url(String)` on the builder the `urlString` field is set, but not the `url` field.  Then when you call `Request.Builder.build`, you get the original `Request`'s url, *not* the url string set on the builder.

The suggested fix is to not copy `Request.url` to `Request.Builder.url` on initialization.

### DIFF
--- a/okhttp-tests/src/test/java/com/squareup/okhttp/RequestTest.java
+++ b/okhttp-tests/src/test/java/com/squareup/okhttp/RequestTest.java
@@ -114,6 +114,18 @@ public final class RequestTest {
     assertEquals(new URL("http://localhost/api"), request.url());
   }
 
+  @Test public void newBuilderUrlResetsUrl() throws Exception {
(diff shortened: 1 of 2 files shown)


JakeWharton:
I think a better change would be to null the Url if you set a string url.

nashira:
That makes sense, thanks.  Pull request updated.

swankjesse:
Test is good; CLA is good. One final request: can you rebase & squash? Once that's done I'll merge this.
http://rebaseandsqua.sh

nashira:
I think it's up to date with master and I squashed the two commits down to one.  Let me know if that works.

PR #1260, opened 2014-12-31
Add a web socket call concept for connecting.

JakeWharton:
Similar to HTTP and Call, the WebSocketCall is a representation of a pending HTTP request and subsequent upgrade to speak web sockets. Upon synchronous execution you are handed a WebSocket instance for synchronous writing and also pass in a WebSocketListener for async callbacks due to reading.

The API changes in this commits also generalize WebSocket such that it's agnostic to being a client or server peer.

Closes #1173.

Missing tests:

 * [x] Non-101 response
 * [x] Response with failed web socket upgrade (multiple)

PR #1481, opened 2015-03-09
Use the now-generalized reply executor for peer close.

JakeWharton:
This was feedback on https://github.com/square/okhttp/pull/1260 from @swankjesse

### DIFF
--- a/okhttp-tests/src/test/java/com/squareup/okhttp/internal/ws/RealWebSocketTest.java
+++ b/okhttp-tests/src/test/java/com/squareup/okhttp/internal/ws/RealWebSocketTest.java
@@ -18,6 +18,7 @@ package com.squareup.okhttp.internal.ws;
 import java.io.IOException;
 import java.net.ProtocolException;
 import java.util.Random;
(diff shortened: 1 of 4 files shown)


JakeWharton:
One side-effect of this is that close might be delayed pending any queued pongs. Before this close would happen immediately and any queued pongs would just fail.

swankjesse:
LGTM.

JakeWharton:
This change might justify injecting the executor as it's what is causing the tests to fail.

okhttp-tests/src/test/java/com/squareup/okhttp/internal/ws/RealWebSocketTest.java
```diff
@@ -50,7 +51,15 @@
   @Before public void setUp() {
     Random random = new Random(0);
     String url = "http://example.com/websocket";
-    client = new RealWebSocket(true, server2client, client2Server, random, clientListener, url) {
+
+    Executor synchronousExecutor = new Executor() {
```
swankjesse: Nice.

PR #1485, opened 2015-03-11
Use ExecutorService.execute, not ExecutorService.submit.

swankjesse:
The latter returns a Future that we never call get() on, and
never get the exception from.


adriancole:
oops

swankjesse:
Bad APIs are bad. submit/execute are a dangerous pair.

PR #1489, opened 2015-03-11
Move WebSocket to the public API in a sibling module.

JakeWharton:


okhttp-ws/src/main/java/com/squareup/okhttp/ws/WebSocket.java
```diff
@@ -13,13 +13,12 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package com.squareup.okhttp.internal.ws;
+package com.squareup.okhttp.ws;
```
JakeWharton: Note: In the core these probably would have been in `com.squareup.okhttp` not `com.squareup.okhttp.ws`. There is usage of some package-private APIs in `com.squareup.okhttp` which currently has to go through `Internal` in order to work. If these were in `com.squareup.okhttp` those indirections could be removed.

Issue #1461, opened 2015-02-28
Cache javadoc wildly out of date.


JakeWharton:
References "connection" (meaning an `HttpURLConnection` instance) and no mention of `CacheControl` at the very least.

PR #1491, opened 2015-03-13
Update cache docs to use Call and CacheControl.

swankjesse:
Fixes https://github.com/square/okhttp/issues/1461


JakeWharton:
Excellent :+1:

Issue #1211, opened 2014-12-18
Cache explodes on abortQuietly


swankjesse:
https://code.google.com/p/android/issues/detail?id=82889

The stacktrace suggests that `journalWriter.flush()` throwing an `IOException` can cause a later `IllegalStateException` crash.

```
java.lang.IllegalStateException
       at com.android.okhttp.internal.DiskLruCache.completeEdit(DiskLruCache.java:506)
       at com.android.okhttp.internal.DiskLruCache.access$2200(DiskLruCache.java:86)
       at com.android.okhttp.internal.DiskLruCache$Editor.abort(DiskLruCache.java:816)
       at com.android.okhttp.HttpResponseCache.abortQuietly(HttpResponseCache.java:259)
       at com.android.okhttp.HttpResponseCache.update(HttpResponseCache.java:251)
       at com.android.okhttp.internal.http.HttpEngine.readResponse(HttpEngine.java:642)
       at com.android.okhttp.internal.http.HttpURLConnectionImpl.execute(HttpURLConnectionImpl.java:379)
       at com.android.okhttp.internal.http.HttpURLConnectionImpl.getResponse(HttpURLConnectionImpl.java:323)
       at com.android.okhttp.internal.http.HttpURLConnectionImpl.getResponseCode(HttpURLConnectionImpl.java:491)
```

li2:
I also met the same crash, code as following:

```java
try {
    if (mDiskLruCache != null && mDiskLruCache.get(key) == null) {
        editor = mDiskLruCache.edit(key);
        if (editor != null) {
            if (addBitmapToDiskCache(editor, bitmap)) {
                mDiskLruCache.flush();
                editor.commit();
                Log.d(TAG, "image put on disk cache " + key);
            } else {
                editor.abort();
                Log.e(TAG, "ERROR on: image put on disk cache " + key);
            }
        }
    }
} catch (IOException ioe) {
    Log.e(TAG, "ERROR on: image put on disk cache " + key + ", " + ioe);
} finally {
    try {
        // TODO IllegalStateException
        // at com.jakewharton.disklrucache.DiskLruCache.completeEdit
        if (editor != null) {
            editor.abort();
        }
    } catch (IOException ignore) {
    }
}
```

then I move the `try-catch` code snippet from `finally` into `catch(IOException ioe)`, and then it works well and didn't crash any more.

```java
try {
    ...
} catch (IOException ioe) {
[...]

PR #1495, opened 2015-03-14
New file system abstraction.

swankjesse:
Working towards https://github.com/square/okhttp/issues/1211


okhttp-tests/src/test/java/com/squareup/okhttp/internal/io/InMemoryFileSystem.java
```diff
@@ -0,0 +1,75 @@
...
+import java.io.IOException;
+import java.util.Iterator;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import okio.Buffer;
+import okio.Sink;
+import okio.Source;
+
+/** A simple file system where all files are held in memory. Not safe for concurrent use. */
+public final class InMemoryFileSystem implements FileSystem {
```
adriancole: I like this

Issue #746, opened 2014-04-21
Cache should recover gracefully when disk space is exhausted


cevcev:
I wonder if any cache clear process automatically or should i do it my self ? In some conditions phones disk has no space and it throws java.io.IOException while using externalcachedir.

Using picasso 2.2.0 and okhttp 1.5.2 

``` java
java.lang.RuntimeException: java.io.IOException: write failed: ENOSPC (No space left on device)
at com.squareup.okhttp.internal.DiskLruCache$1.run(DiskLruCache.java:175)
at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1080)
at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:573)
at java.lang.Thread.run(Thread.java:838)
Caused by: java.io.IOException: write failed: ENOSPC (No space left on device)
at libcore.io.IoBridge.write(IoBridge.java:506)
at java.io.FileOutputStream.write(FileOutputStream.java:187)
at com.squareup.okhttp.internal.okio.Okio$1.write(Okio.java:70)
at com.squareup.okhttp.internal.okio.RealBufferedSink.close(RealBufferedSink.java:150)
at com.squareup.okhttp.internal.DiskLruCache.rebuildJournal(DiskLruCache.java:366)
at com.squareup.okhttp.internal.DiskLruCache.access$300(DiskLruCache.java:86)
at com.squareup.okhttp.internal.DiskLruCache$1.run(DiskLruCache.java:171)
... 3 more
Caused by: libcore.io.ErrnoException: write failed: ENOSPC (No space left on device)
at libcore.io.Posix.writeBytes(Native Method)
at libcore.io.Posix.write(Posix.java:189)
at libcore.io.BlockGuardOs.write(BlockGuardOs.java:197)
at libcore.io.IoBridge.write(IoBridge.java:501)
... 9 more
```

swankjesse:
We're getting closer on this. One full-disk bug fixed in 2.3 is here:
https://github.com/square/okhttp/issues/1211

Still remaining is improving `rebuildJournal()` to work with an unwritable disk.

yschimke:
Issue is general, but within a JVM app it's ops fault.  Within a mobile, it's awkward to handle except in OkHttp.

yschimke:
Closing as low scope. Re-open if this is impacting you.

PR #1497, opened 2015-03-14
Don't get corrupted when journal writing fails.

swankjesse:
Closes https://github.com/square/okhttp/issues/1211

Working towards https://github.com/square/okhttp/issues/746

### DIFF
--- a/okhttp/src/main/java/com/squareup/okhttp/internal/DiskLruCache.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/internal/DiskLruCache.java
@@ -36,7 +36,6 @@ import java.util.regex.Pattern;
 import okio.Buffer;
 import okio.BufferedSink;
 import okio.BufferedSource;
-import okio.ForwardingSink;
 import okio.Okio;
 import okio.Sink;
 import okio.Source;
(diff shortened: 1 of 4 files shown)


okhttp/src/main/java/com/squareup/okhttp/internal/DiskLruCache.java
```diff
@@ -291,13 +291,23 @@ private void readJournal() throws IOException {
...
     } finally {
       Util.closeQuietly(source);
     }
   }
 
+  private BufferedSink newJournalWriter() throws FileNotFoundException {
+    Sink fileSink = fileSystem.appendingSink(journalFile);
+    Sink faultHidingSink = new FaultHidingSink(fileSink) {
+      @Override protected void onException(IOException e) {
+        hasJournalErrors = true;
```
JakeWharton: Synchronize on the instance. Reads are guarded by it.

JakeWharton:
LGTM

okhttp/src/main/java/com/squareup/okhttp/internal/DiskLruCache.java
swankjesse: (We only write to the journal inside a synchronized block, but I'll make that explicit)

PR #1498, opened 2015-03-15
Introduce a buffer in SpdyStream's outgoing frames.

swankjesse:
Previously we'd consistently send 2048 byte frames because that was
Okio's default segment size. This increases OkHttp's typical segment
size to the 16 KiB frame size that all HTTP/2 endpoints support.

### DIFF
--- a/okhttp/src/main/java/com/squareup/okhttp/internal/spdy/SpdyStream.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/internal/spdy/SpdyStream.java
@@ -53,7 +53,6 @@ public final class SpdyStream {
 
   private final int id;
   private final SpdyConnection connection;
-  private long readTimeoutMillis = 0;
 
   /** Headers sent by the stream initiator. Immutable and non null. */
   private final List<Header> requestHeaders;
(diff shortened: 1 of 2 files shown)


okhttp/src/main/java/com/squareup/okhttp/internal/spdy/SpdyStream.java
```diff
@@ -478,26 +485,34 @@ private void cancelStreamIfNecessary() throws IOException {
...
         }
 
-        byteCount -= toWrite;
-        connection.writeData(id, false, source, toWrite);
+        checkOutNotClosed(); // Kick out if the stream was reset or closed while waiting.
+        toWrite = Math.min(bytesLeftInWriteWindow, sendBuffer.size());
+        bytesLeftInWriteWindow -= toWrite;
       }
+
+      connection.writeData(id, outFinished && toWrite == sendBuffer.size(), sendBuffer, toWrite);
```
adriancole: ok I see the relationship. In flush(), outFinished will be true, but it will end up false here as we've not yet exhausted the sendBuffer.

okhttp/src/main/java/com/squareup/okhttp/internal/spdy/SpdyStream.java
```diff
@@ -518,7 +536,15 @@ private void cancelStreamIfNecessary() throws IOException {
         if (closed) return;
       }
       if (!sink.finished) {
-        connection.writeData(id, true, null, 0);
+        // Emit the remaining data, setting the END_STREAM flag on the last frame.
+        if (sendBuffer.size() > 0) {
+          while (sendBuffer.size() > 0) {
+            emitDataFrame(true);
```
adriancole: basically, send frames until you can close the stream (outFinished = true).

adriancole:
LGTM: kicking build as travis was a bit stormy

adriancole:
I don't have a rebuild button, so kicking build old school

adriancole:
winky travis. one passed one failed. good enough for me

PR #1499, opened 2015-03-15
Stream the response in okcurl.

swankjesse:
Beneficial for this website in particular:
https://http2.golang.org/clockstream


adriancole:
neat!

JakeWharton:
What was the behavior before? It would only emit after 2048?

adriancole:
@JakeWharton I guess the trick is pulling until exhausted is true. I'm guessing the former code would only read the first chunk.

adriancole:
suppose we could start adding mws tests for things like this in the future..

swankjesse:
Previous behavior emitted every 2048 bytes. This emits more frequently.

Issue #1482, opened 2015-03-10
[OkHttp] Difference in error reporting bw interceptors+sync and interceptors+async


circlespainter:
This test https://github.com/circlespainter/okhttp/commit/7d0a1ef6ff9c16064ddcbc9b159ab149fe3176cf fails, showing a difference in behaviour between the sync and async APIs when reporting errors related to interceptors.
If this is known and meant to be like this, how is it best to perform interceptors error handling in the async case? TIA

swankjesse:
I haven't really considered how interceptors interact with unchecked exceptions.

For async, we shouldn't call onFailure with the unchecked exception; that method is for I/O failures only. Instead we're better off crashing to the worker thread's uncaught exception handler.

For sync, we have to throw to the caller; there's no other alternative.

I need to write a test for this.

swankjesse:
And if you want to throw in an interceptor, throw a subclass of IOException. Those are handled more consistently.

circlespainter:
In this case it's not an user-generated unchecked exception but it's thrown by the API while validating the user attempt to change the server address within an interceptor (and there are other similar validation cases like attempting to invoke `call` multiple times or trying to short-circuit the response).

The reason I need the async and sync behaviour to be made similar is that I'm using the async API to implement a fiber-blocking version of the sync one.
Is there any way for me to make the async API's unchecked exception handling similar to the sync one?

swankjesse:
Can you throw an IOException?

swankjesse:
Or perhaps return an HTTP response indicating a bad response?

```
new Response.Builder()
            .request(request)
            .protocol(Protocol.HTTP_1_1)
            .code(400)
            .message("Malformed request!")
            .body(ResponseBody.create(null, 0L, new Buffer());)
            .build()
```

This is probably your best bet, especially if you report unexpected 4xx errors to a service like Bugsnag.

circlespainter:
Thanks, I'll have a look and see if I can add some additional behaviour when interceptors are added, I think it's better since I'm not working on a client application but on an integration.

PR #1501, opened 2015-03-15
Add a test for interceptors throwing unchecked exceptions.

swankjesse:
Our behavior here is not good. Right now it looks like we're leaking
resources; nothing is releasing the socket in a 'finally' clause when
interceptors crash. We're also not notifying the callback that the
call has failed; that's left hanging.

That said, we don't expect applications to recover from these
exceptions.

Closes https://github.com/square/okhttp/issues/1482


JakeWharton:
lgtm

Issue #1328, opened 2015-01-15
com.squareup.okhttp.internal.spdy.Spdy3ConnectionTest#getResponseHeadersTimesOut is flaky on Android


nfuller:
I haven't looked closely at why. This shows up when running the OkHttp tests under Android's CTS runner (an InstrumentationTestCase arrangement).

Sometimes it passes, sometimes it fails with:

01-15 11:43:09 I/0201ad2c860e49b1: com.squareup.okhttp.internal.spdy.Spdy3ConnectionTest#getResponseHeadersTimesOut FAIL 
java.lang.AssertionError: expected:<0> but was:<1>
at org.junit.Assert.fail(Assert.java:93)
at org.junit.Assert.failNotEquals(Assert.java:647)
at org.junit.Assert.assertEquals(Assert.java:128)
at org.junit.Assert.assertEquals(Assert.java:472)
at org.junit.Assert.assertEquals(Assert.java:456)
at com.squareup.okhttp.internal.spdy.Spdy3ConnectionTest.getResponseHeadersTimesOut(Spdy3ConnectionTest.java:888)
at java.lang.reflect.Method.invoke(Native Method)
at java.lang.reflect.Method.invoke(Method.java:372)
at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:45)
at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:15)
at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:42)
at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:20)
at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:30)
at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:263)
at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:68)
at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:47)
at org.junit.runners.ParentRunner$3.run(ParentRunner.java:231)
at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:60)
at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:229)
at org.junit.runners.ParentRunner.access$000(ParentRunner.java:50)
at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:222)
[...]

PR #1503, opened 2015-03-15
Fix a flaky test by waiting until the Okio watchdog is caught up.

swankjesse:
We had tests that flake because they expected the watchdog to have
completed closing a connection, but we weren't blocking until that
work was done.

Closes https://github.com/square/okhttp/issues/1328


JakeWharton:
LGTM

PR #1505, opened 2015-03-15
Relax expectations when streamed requests fail.

swankjesse:
We don't have machinery to recover from this when streamed requests
fail, and we also don't have machinery to reliably detect when they
will fail.

This is sad. Fortunately the new API doesn't suffer this problem.


JakeWharton:
Green build... Is this real life?!

swankjesse:
Green builds mean you can code-review on the go & GitHub won't punish you!

Issue #1336, opened 2015-01-18
OkHttpClient API to trust a PEM file


Lobooo:
Hi, how can I load a self signed certificate using okhttp?
Using apache standard library I just register my SocketFactory and it works like this:

public class HttpClientUtil {
    public static DefaultHttpClient getNewHttpClient() {
        try {
            KeyStore trustStore = KeyStore.getInstance(KeyStore.getDefaultType());
            trustStore.load(null, null);

            SSLSocketFactory sf = new MySSLSocketFactory(trustStore);
            sf.setHostnameVerifier(SSLSocketFactory.ALLOW_ALL_HOSTNAME_VERIFIER);

            HttpParams params = new BasicHttpParams();
            HttpProtocolParams.setVersion(params, HttpVersion.HTTP_1_1);
            HttpProtocolParams.setContentCharset(params, HTTP.UTF_8);

            SchemeRegistry registry = new SchemeRegistry();
            registry.register(new Scheme("http", PlainSocketFactory.getSocketFactory(), 80));
            registry.register(new Scheme("https", sf, 443));

            ClientConnectionManager ccm = new ThreadSafeClientConnManager(params, registry);

            return new DefaultHttpClient(ccm, params);
        } catch (Exception e) {
            return new DefaultHttpClient();
        }
    }

}

public class MySSLSocketFactory extends SSLSocketFactory {
    SSLContext sslContext = SSLContext.getInstance("TLS");

    public MySSLSocketFactory(KeyStore truststore) throws NoSuchAlgorithmException, KeyManagementException, KeyStoreException, UnrecoverableKeyException {
        super(truststore);

        TrustManager tm = new X509TrustManager() {
            public void checkClientTrusted(X509Certificate[] chain, String authType) throws CertificateException {
            }

            public void checkServerTrusted(X509Certificate[] chain, String authType) throws CertificateException {
            }

            public X509Certificate[] getAcceptedIssuers() {
                return null;
            }
        };

        sslContext.init(null, new TrustManager[] { tm }, null);
    }

    @Override
    public Socket createSocket(Socket socket, String host, int port, boolean autoClose) throws IOException, UnknownHostException {
        return sslContext.getSocketFactory().createSocket(socket, host, port, autoClose);
    }

    @Override
    public Socket createSocket() throws IOException {
        return sslContext.getSocketFactory().createSocket();
    }
}

JakeWharton:
Register your socket factory with your OkHttpClient.

[...]

PR #1506, opened 2015-03-15
Custom trust recipe.

swankjesse:
To my great delight, Java and Android can read .pem files without
the BKS or JKS store files we typically use to distribute trusted
certificates.

This recipe includes a big copy-and-pasteable method for converting
an InputStream of PEM files into an SSLContext. The rest of it is
warnings about how dangerous that is.

Closes https://github.com/square/okhttp/issues/1336

### DIFF
--- a/samples/guide/src/main/java/com/squareup/okhttp/recipes/CustomTrust.java
+++ b/samples/guide/src/main/java/com/squareup/okhttp/recipes/CustomTrust.java
@@ -1,0 +1,206 @@
+/*
+ * Copyright (C) 2015 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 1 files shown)


samples/guide/src/main/java/com/squareup/okhttp/recipes/CustomTrust.java
```diff
@@ -0,0 +1,207 @@
...
+   * available. Or in production, to avoid reliance on third-party certificate authorities.
+   *
+   * <p>See also {@link CertificatePinner}, which can limit trusted certificates while still using
+   * the host platform's built-in trust store.
+   *
+   * <h3>Warning: Customizing Trusted Certificates is Dangerous!</h3>
+   * Relying on your own trusted certificates limits your server team's ability to update their TLS
+   * certificates. By installing a specific set of trusted certificates, you take on additional
+   * operational complexity and limit your ability to migrate between certificate authorities. Do
+   * not use custom trusted certificates in production without hte blessing of your server's TLS
```
JakeWharton: "the"

JakeWharton:
Lgtm

PR #1508, opened 2015-03-16
Fix some crawler bugs.

swankjesse:
We were interpretting links relative to the URLs requested, rather
than the redirect targets of those URLs.

### DIFF
--- a/samples/crawler/src/main/java/com/squareup/okhttp/sample/Crawler.java
+++ b/samples/crawler/src/main/java/com/squareup/okhttp/sample/Crawler.java
@@ -19,6 +19,7 @@ import com.squareup.okhttp.Cache;
 import com.squareup.okhttp.OkHttpClient;
 import com.squareup.okhttp.Request;
 import com.squareup.okhttp.Response;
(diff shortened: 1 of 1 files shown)


samples/crawler/src/main/java/com/squareup/okhttp/sample/Crawler.java
```diff
@@ -76,12 +80,18 @@ private void drainQueue() throws Exception {
   }
 
   public void fetch(URL url) throws IOException {
+    // Skip hosts that we've visited many times.
+    AtomicInteger hostnameCount = new AtomicInteger();
+    AtomicInteger previous = hostnames.putIfAbsent(url.getHost(), hostnameCount);
```
JakeWharton: Isn't this a Java 8 method?

JakeWharton:
Lgtm

samples/crawler/src/main/java/com/squareup/okhttp/sample/Crawler.java
swankjesse: Java 5. It's ConcurrentHashMap, not Map.

PR #1510, opened 2015-03-16
Remove -SNAPSHOT from okio version

nfuller:
This seems to be required for me to build successfully.


JakeWharton:
Prior to last night you needed to `mvn install` Okio beside OkHttp since both were being developed in tandem. But now that 1.3 is out, LGTM.

PR #1516, opened 2015-03-19
Set readerSentClose to break reading loop and close frame

cy6erGn0m:
We have to set readerSentClose = true to break frames reading loop otherwise we get IOException that will call callback.onFailure.. it is worst if onFailure will be called before onClose (most likely) so user can't make decition.

Would be great to integrate to 2.3 if it is planned for release 2.3.1


JakeWharton:
I need to take a closer look at this. I'll try to do it this evening. Thanks for bringing it to our attention.

cy6erGn0m:
Any luck with it?

Issue #1471, opened 2015-03-07
Recipe: upload & download progress


swankjesse:

yschimke:
Can we close this, as addressed? https://github.com/square/okhttp/blob/master/samples/guide/src/main/java/okhttp3/recipes/Progress.java

swankjesse:
I wanna do upload progress ’cause the examples on StackOverflow are wrong and people will copy+paste it until we provide a better one.

aried3r:
Whoops, this is partly on me, I began work on this more than 3 years ago but never got around to fixing it. And I don't have time right now to rebase and and adapt to the new API. 🙈 

https://github.com/square/okhttp/pull/1528

thalisvilela:
@aried3r Any update on this matter? Nothing works! The progress just goes to 100 while the file still being uploaded.

Shusshu:
> I wanna do upload progress ’cause the examples on StackOverflow are wrong and people will copy+paste ’em until we provide a better one.

Is there a nice approach to this today?

vincent-paing:
> I wanna do upload progress ’cause the examples on StackOverflow are wrong and people will copy+paste ’em until we provide a better one.

Any updates on this?

yschimke:
Better as a standalone library or a canonical stackoverflow post.

PR #1526, opened 2015-03-21
Progress recipe

aried3r:
This is the first draft of a (download) progress recipe, mentioned in #1471. 

Looking forward to suggestions for improvements. For example, the output is a bit verbose right now, for testing purposes, also the file is too small to actually show anything other than 100%, because it's <2KB which is the segment size of okio (from what I've gathered).

Also, am I missing something, there is no progress if I comment out line 56 (the `System.out` call), but that shouldn't influence the request afaik!?

Another thing, I'd like to provide another example for upload and split it into two files, `DownloadProgress.java` and `UploadProgress.java` or should it stay in one?

### DIFF
--- a/samples/guide/src/main/java/com/squareup/okhttp/recipes/Progress.java
+++ b/samples/guide/src/main/java/com/squareup/okhttp/recipes/Progress.java
@@ -1,0 +1,112 @@
+/*
+ * Copyright (C) 2015 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.okhttp.recipes;
+
+import com.squareup.okhttp.Interceptor;
+import com.squareup.okhttp.MediaType;
+import com.squareup.okhttp.OkHttpClient;
+import com.squareup.okhttp.Response;
+import com.squareup.okhttp.ResponseBody;
+import com.squareup.okhttp.Request;
+import java.io.IOException;
+import okio.Buffer;
+import okio.BufferedSource;
+import okio.ForwardingSource;
+import okio.Okio;
+import okio.Source;
+
+public final class Progress {
+
+  private final OkHttpClient client = new OkHttpClient();
+
+  public void run() throws Exception {
+    Request request = new Request.Builder()
+        .url("https://publicobject.com/helloworld.txt")
+        .build();
+
+    final ProgressListener progressListener = new ProgressListener() {
+      @Override
+      public void update(long bytesRead, long contentLength, boolean done) {
+        System.out.println(bytesRead);
+        System.out.println(contentLength);
+        System.out.println(done);
+        System.out.format("%d%% done\n", (100 * bytesRead) / contentLength);
+      }
+    };
+
+    client.networkInterceptors().add(new Interceptor() {
+      @Override public Response intercept(Chain chain) throws IOException {
+        Response originalResponse = chain.proceed(chain.request());
+        return originalResponse.newBuilder()
+            .body(new ProgressResponseBody(originalResponse.body(), progressListener))
+            .build();
+        }
+    });
+
+    Response response = client.newCall(request).execute();
+    if (!response.isSuccessful()) throw new IOException("Unexpected code " + response);
+
+    System.out.println(response.body().string());
+  }
+
+  public static void main(String... args) throws Exception {
+    new Progress().run();
+  }
+
+  private static class ProgressResponseBody extends ResponseBody {
+
+    private final ResponseBody responseBody;
+    private final ProgressListener progressListener;
+    private BufferedSource bufferedSource;
+
+    public ProgressResponseBody(ResponseBody responseBody, ProgressListener progressListener) {
+      this.responseBody = responseBody;
+      this.progressListener = progressListener;
+    }
+
+    @Override public MediaType contentType() {
+      return responseBody.contentType();
+    }
+
+    @Override public long contentLength() throws IOException {
+      return responseBody.contentLength();
+    }
+
+    @Override public BufferedSource source() throws IOException {
+      if (bufferedSource == null) {
+        bufferedSource = Okio.buffer(source(responseBody.source()));
+      }
+      return bufferedSource;
+    }
+
+    private Source source(Source source) {
+      return new ForwardingSource(source) {
+        long totalBytesRead = 0L;
+        @Override public long read(Buffer sink, long byteCount) throws IOException {
+          long bytesRead = super.read(sink, byteCount);
+          // read() returns the number of bytes read, or -1 if this source is exhausted.
+          totalBytesRead += bytesRead != -1 ? bytesRead : 0;
+          progressListener.update(totalBytesRead, responseBody.contentLength(), bytesRead != -1);
+          return bytesRead;
+        }
+      };
+    }
+  }
+
+  interface ProgressListener {
+    void update(long bytesRead, long contentLength, boolean done);
+  }
+}


samples/guide/src/main/java/com/squareup/okhttp/recipes/Progress.java
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  private static class ProgressResponseBody extends ResponseBody {
+
+    private long limit;
+    private MediaType contentType;
+    private ProgressListener progressListener;
+    private Source source;
+
+    public ProgressResponseBody(final long limit, final MediaType contentType,
```
swankjesse: instead of taking a content type, content length and source, this could just take a ResponseBody

samples/guide/src/main/java/com/squareup/okhttp/recipes/Progress.java
```diff
@@ -0,0 +1,119 @@
...
+        @Override public void close() throws IOException {
+          super.close();
+          progressListener.update(totalBytesRead, limit, true);
+        }
+      };
+    }
+  }
+
+  private static class ProgressListener {
+    public ProgressListener() { }
```
swankjesse: drop this?

swankjesse: (it's not necessary)

samples/guide/src/main/java/com/squareup/okhttp/recipes/Progress.java
```diff
@@ -0,0 +1,119 @@
...
+    public ProgressResponseBody(final long limit, final MediaType contentType,
+        final ProgressListener progressListener, Source source) {
+      this.limit = limit;
+      this.contentType = contentType;
+      this.progressListener = progressListener;
+      this.source = source;
+    }
+
+    @Override public MediaType contentType() {
+      return contentType;
```
swankjesse: ... and then these can just call `.contentType()` on the passed-in response

samples/guide/src/main/java/com/squareup/okhttp/recipes/Progress.java
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    private Source source(Source source) {
+      return new ForwardingSource(source) {
+        long totalBytesRead = 0L;
+        @Override public long read(Buffer sink, long byteCount) throws IOException {
+          long bytesRead = super.read(sink, byteCount);
+          // read() returns the number of bytes read, or -1 if this source is exhausted.
+          totalBytesRead += bytesRead != -1 ? bytesRead : 0;
+          progressListener.update(totalBytesRead, limit, false);
```
swankjesse: Instead of passing `false` here unconditionally, you could pass `bytesRead != -1`.

swankjesse: (That way it'll get the update call before close gets called)

samples/guide/src/main/java/com/squareup/okhttp/recipes/Progress.java
```diff
@@ -0,0 +1,119 @@
...
+          return bytesRead;
+        }
+
+        @Override public void close() throws IOException {
+          super.close();
+          progressListener.update(totalBytesRead, limit, true);
```
swankjesse: And if you drop the bytesRead special case above, you don't need to override `close()` at all? Or do you want to do that in case the request gets canceled? I suspect it's simpler to just not override close and not make promises.

samples/guide/src/main/java/com/squareup/okhttp/recipes/Progress.java
```diff
@@ -0,0 +1,119 @@
...
+    @Override public MediaType contentType() {
+      return contentType;
+    }
+
+    @Override public long contentLength() {
+      return limit;
+    }
+
+    @Override public BufferedSource source() {
+      return Okio.buffer(source(source));
```
swankjesse: You should cache this object and return it if the caller calls `source()` multiple times. Otherwise you'll end up creating multiple instances that are out of sync with each other

samples/guide/src/main/java/com/squareup/okhttp/recipes/Progress.java
```diff
@@ -0,0 +1,119 @@
...
+    System.out.println(response.body().string());
+  }
+
+  public static void main(String... args) throws Exception {
+    new Progress().run();
+  }
+
+  private static class ProgressResponseBody extends ResponseBody {
+
+    private long limit;
```
swankjesse: can these be final?

samples/guide/src/main/java/com/squareup/okhttp/recipes/Progress.java
```diff
@@ -0,0 +1,119 @@
...
+        .build();
+
+    client.networkInterceptors().add(new Interceptor() {
+      @Override public Response intercept(Chain chain) throws IOException {
+        Response originalResponse = chain.proceed(chain.request());
+        return originalResponse.newBuilder()
+            .body(new ProgressResponseBody(
+                originalResponse.body().contentLength(),
+                originalResponse.body().contentType(),
+                new ProgressListener(),
```
swankjesse: One idea: you could declare ProgressListener as an interface below, and then here in the `run()` method you could implement that interface. That way you cleanly separate one from the other.

In this method, it'd be something like this:

ProgressListener progressListener = new ProgressListener() {
  @Override public void update(long bytesRead, long contentLength, boolean exhausted) {
    System.out.println("%d of %d read, done=%s", bytesRead, contentLength, exhausted);
  }
};

samples/guide/src/main/java/com/squareup/okhttp/recipes/Progress.java
```diff
@@ -0,0 +1,119 @@
...
+          progressListener.update(totalBytesRead, limit, true);
+        }
+      };
+    }
+  }
+
+  private static class ProgressListener {
+    public ProgressListener() { }
+
+    public void update(final long bytesRead, final long limit, final boolean done) {
```
swankjesse: nit: I think it should be `contentLength` and not `limit` (the difference is that content length can be -1.

swankjesse:
Dig. I've got lots of comments but the code is solid.

samples/guide/src/main/java/com/squareup/okhttp/recipes/Progress.java
aried3r: I really just went with your code in https://github.com/square/okhttp/issues/1042#issuecomment-68012313, guess I shouldn't have used it verbatim, but fair point.

aried3r:
All good points, I'll see that I get to it this weekend.

samples/guide/src/main/java/com/squareup/okhttp/recipes/Progress.java
swankjesse: ha! Yeah.

samples/guide/src/main/java/com/squareup/okhttp/recipes/Progress.java
aried3r: As for the `contentLength()`, how do you want me to handle the IOException contentLength throws since OkHttp 2.2.0?

aried3r: nvm, I'll just make it `throws IOException` as per the documentation.

samples/guide/src/main/java/com/squareup/okhttp/recipes/Progress.java
```diff
@@ -0,0 +1,112 @@
...
+    new Progress().run();
+  }
+
+  private static class ProgressResponseBody extends ResponseBody {
+
+    private final ResponseBody responseBody;
+    private final ProgressListener progressListener;
+    private BufferedSource bufferedSource;
+
+    public ProgressResponseBody(final ResponseBody body, final ProgressListener listener) {
```
aried3r: I called the arguments `body` and `listener` for short so it would fit in one line, I hope that's ok?

aried3r:
I hope I didn't miss anything, once you give me the ok, I'll squash the commits, I made a separate one so it's easier to see the changes I applied.

As for the upload progress, shall I write another recipe or do you think people can apply this one to uploading just fine?

swankjesse:
If you'd like to do an upload progress recipe, I'd like that too! After we get this one merged.

samples/guide/src/main/java/com/squareup/okhttp/recipes/Progress.java
swankjesse: yup. The `final` on locals isn't necessary; only crazy people re-assign values to locals!

samples/guide/src/main/java/com/squareup/okhttp/recipes/Progress.java
```diff
@@ -0,0 +1,112 @@
...
+    @Override public MediaType contentType() {
+      return responseBody.contentType();
+    }
+
+    @Override public long contentLength() throws IOException {
+      return responseBody.contentLength();
+    }
+
+    @Override public BufferedSource source() throws IOException {
+      if (bufferedSource == null) {
```
swankjesse: Nice.

samples/guide/src/main/java/com/squareup/okhttp/recipes/Progress.java
```diff
@@ -0,0 +1,112 @@
...
+          totalBytesRead += bytesRead != -1 ? bytesRead : 0;
+          progressListener.update(totalBytesRead, responseBody.contentLength(), bytesRead != -1);
+          return bytesRead;
+        }
+      };
+    }
+  }
+
+  interface ProgressListener {
+    void update(final long bytesRead, final long contentLength, final boolean done);
```
swankjesse: `final` here doesn't do anything useful. You can strip it!

swankjesse:
Strip the finals, squash the commits and I'll merge this!

swankjesse:
Thanks!

swankjesse:
(http://rebaseandsqua.sh makes the squashing part easy)

aried3r:
Should be fine now.

swankjesse:
Win.

Tolriq:
@swankjesse sorry to hijack this thread but since I found the question by using this recipe I prefer to ask here first. Is it possible to have interceptors only for a cloned OkHttpClient ? or is by design that interceptors added to cloned clients are applied to all clients ?

Use case would be that I have a client with already interceptors to add authentication and some headers, and would need to keep this configuration but add a new interceptor only for the cloned client used in my download thread.

swankjesse:
@Tolriq our current behavior is by mistake. If you don't mind using a snapshot, the fix is already merged.

Tolriq:
@swankjesse Thanks, This is good news, for now I add manually the wrapper but I'll switch when 2.4 is out.

bakua:
Hello. How does it look like with Upload recipe please? I was able to find one only for download.

Thank you

lbt05:
I'm confused about this sample code... what should I do when I need to share the okhttpClient instance for two download request which both need progressListener?

swankjesse:
One option is to clone the OkHttpClient before attaching the progress listener.

WalkingRoy:
@swankjesse hi, any other options else ?

yschimke:
I would say the main options are

1) Clone OkHttpClient (cheap as it can share the original dispatcher and connection pool)
2) Change the Progress example and pass in a ListenerFactory, this is similar to how EventListener currently works and instead builds a new ProgressListener per call.

ngaya-ll:
Another solution is to "decorate" each request with its own listener and use that to update the relevant listener in the interceptor.  See https://github.com/swagger-api/swagger-codegen/pull/8053

PR #1528, opened 2015-03-23
Upload progress recipe

aried3r:
Let's have it, @swankjesse! ;)

There's a few things I'm not quite sure about how to handle (basically the whole `writeTo()` method), so I'm awaiting your comments. But I hope I'll stay below 10 this time :p

### DIFF
--- a/samples/guide/src/main/java/com/squareup/okhttp/recipes/UploadProgress.java
+++ b/samples/guide/src/main/java/com/squareup/okhttp/recipes/UploadProgress.java
@@ -1,0 +1,120 @@
+/*
+ * Copyright (C) 2015 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.okhttp.recipes;
+
+import com.squareup.okhttp.MediaType;
+import com.squareup.okhttp.OkHttpClient;
+import com.squareup.okhttp.Request;
+import com.squareup.okhttp.RequestBody;
+import com.squareup.okhttp.Response;
+import okio.Buffer;
+import okio.BufferedSink;
+import okio.ForwardingSink;
+import okio.Okio;
+import okio.Sink;
+import okio.Source;
+
+import java.io.File;
+import java.io.IOException;
+
+public final class UploadProgress {
+  /**
+   * The imgur client ID for OkHttp recipes. If you're using imgur for anything
+   * other than running these examples, please request your own client ID!
+   * https://api.imgur.com/oauth2
+   */
+  private static final String IMGUR_CLIENT_ID = "9199fdef135c122";
+  private static final MediaType MEDIA_TYPE_PNG = MediaType.parse("image/png");
+
+  private final OkHttpClient client = new OkHttpClient();
+
+  public void run() throws Exception {
+    final ProgressListener progressListener = new ProgressListener() {
+      @Override public void update(long bytesWritten, long contentLength, boolean done) {
+        System.out.println(bytesWritten);
+        System.out.println(contentLength);
+        System.out.println(done);
+        System.out.format("%d%% done\n", (100 * bytesWritten) / contentLength);
+      }
+    };
+
+    // Use the imgur image upload API as documented at https://api.imgur.com/endpoints/image
+    Request request = new Request.Builder()
+        .header("Authorization", "Client-ID " + IMGUR_CLIENT_ID)
+        .url("https://api.imgur.com/3/image")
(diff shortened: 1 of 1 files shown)


samples/guide/src/main/java/com/squareup/okhttp/recipes/UploadProgress.java
```diff
@@ -0,0 +1,120 @@
...
+    new UploadProgress().run();
+  }
+
+  private static class ProgressRequestBody extends RequestBody {
+
+    private final ProgressListener progressListener;
+    private final MediaType contentType;
+    private final File file;
+
+    public ProgressRequestBody(MediaType contentType, File file,
```
swankjesse: Could you change this so that it wraps another RequestBody? That way the caller can do whatever they want: upload data being generated, a file, or whatever.

samples/guide/src/main/java/com/squareup/okhttp/recipes/UploadProgress.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    public Sink sink(Sink sink) {
+      return new ForwardingSink(sink) {
+        long totalBytesWritten = 0L;
+
+        @Override public void write(Buffer source, long byteCount) throws IOException {
+          super.write(source, byteCount);
+          totalBytesWritten += byteCount;
+          progressListener.update(totalBytesWritten, contentLength(),
```
swankjesse: I think this case is asymmetric with the download case, in that you don't get a call to `write()` for the end of the stream. I think the fix is to add a close() method.

swankjesse: ... (and the close method calls `update()`.)

swankjesse:
I think this is pretty awesome. Consider building on the `PostFile` endpoint instead. It's simpler & so there's less to distract from the main content of the example. With this one it's unclear whether you're showing me about multipart vs. progress monitor.

swankjesse:
This is great!

samples/guide/src/main/java/com/squareup/okhttp/recipes/UploadProgress.java
aried3r: I wanted to, but from what I've gathered I can't get the upload data out of the request body, so it would have to look like this, `file` being passed to the `RequestBody` and again to the `ProgressRequestBody` so it can be accessed for `contentLength()` etc.
```Java
File file = new File("website/static/logo-square.png");
RequestBody requestBody = RequestBody.create(MEDIA_TYPE_PNG, file);
  .header("Authorization", "Client-ID " + IMGUR_CLIENT_ID)
  .url("https://api.imgur.com/3/image")
  .post(new ProgressRequestBody(requestBody, file, progressListener)) // Here file is passed a second time
  .build();
```

samples/guide/src/main/java/com/squareup/okhttp/recipes/UploadProgress.java
aried3r: I know I don't get a `-1`, which is why I test for `totalBytesWritten == contentLength()`.
Sample output, without overriding `close()`:
```
2048
3740
false
54% done
3740
3740
true
100% done
```
It seems to work without it, is there a case where overriding `close()` is absolutely necessary?

aried3r:
> With this one it's unclear whether you're showing me about multipart vs. progress monitor.

The only difference is `.header("Authorization", "Client-ID " + IMGUR_CLIENT_ID)`, I'm not even explicitly using a multipart builder, I'm not passing a name or a description or anything, but I can change it, if you want.

The reason why I used the imgur example was that the README is below 2KB which means there's no real progress to watch, but I guess it should just serve as an example?

samples/guide/src/main/java/com/squareup/okhttp/recipes/UploadProgress.java
swankjesse: You can, but you need to be a bit sneaky. You'd just need to call `delegate.writeTo(buffer(progressSink(sink)))` in your own `writeTo` method.

samples/guide/src/main/java/com/squareup/okhttp/recipes/UploadProgress.java
swankjesse: Yes. Sometimes you don't know content length in advance.

samples/guide/src/main/java/com/squareup/okhttp/recipes/UploadProgress.java
```diff
@@ -0,0 +1,120 @@
...
+              totalBytesWritten == contentLength());
```
swankjesse: (You can look at the PostStreaming example to see an unknown content length in practice.)

swankjesse:
@aried3r follow me on the PostStreaming thing?

aried3r:
Sorry, been busy. I'll try to get this done during the weekend. Stay tuned! ;)

swankjesse:
I'm going to close this for now. If you'd like to revisit, please do!

durunvo:
Hi, i used this code and get this error in onFailure 

unexpected end of stream

Another thing, i don't understand that the progress is not from the uploading, but just writing the buffer?
Because the progress took like 20ms to finish and the is too fast because 10 seconds later after progress done, it callback into onFailure
screen shot 2558-07-10 at 1 47 20 am

LouisCAD:
Does the code found [here](https://gist.github.com/kyawkyaw/54d9e2feb1c969c90673) sticks to your recommendations @swankjesse ?

PR #1532, opened 2015-03-25
Make OkHttpClient and Call final.

swankjesse:
This will break mocks. We have a much simpler solution now, with
interceptors. Instead of mocking the HTTP client, use interceptors
to change its behavior.

Fix a bug where interceptor lists aren't independent after cloning
the OkHttpClient.


adriancole:
checkstyle is quite unhappy. otherwise LGTM especially substituting mocks with an alternative.

iNoles:
@adriancole checkstyle is not only one, javac complains about unchecked or unsafe operations in Util.java too.

JakeWharton:
LGTM. Grab `WebSocketCall` too?

PR #1541, opened 2015-04-01
Adding more content and example to Dispatcher

lalbuquerque:
I realized that Dispatcher session in the README.md had few explanation and no quick example.
So I added one single line of explanation and a little example to quickly understanding.


rotilho:
+1

swankjesse:
Awesome.

PR #1547, opened 2015-04-06
Add Protocol.SHOUTCAST to distinguish Shoutcast protocol from Request ob...

kenmasumitsu:
Shoutcast protocol is handled by HTTP_1_0 protocol. So there are no ways to distinguish HTTP/1.0 and Shoutcast protocol from the application.  Some application needs to distinguish them.

The Pull Requests add Shoutcast protocol to distinguish them.

### DIFF
--- a/okhttp/src/main/java/com/squareup/okhttp/Protocol.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/Protocol.java
@@ -35,6 +35,8 @@ public enum Protocol {
    */
   HTTP_1_0("http/1.0"),
 
+  SHOUTCAST("icy"),
+
   /**
    * A plaintext framing that includes persistent connections.
    *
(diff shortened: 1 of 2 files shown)


swankjesse:
Just curious... why do you need to differentiate? Is it possible to fix your server to do a proper `HTTP/1.1` status line? (The `ICY` thing is obsolete so far as I know.)

kenmasumitsu:
My app is HTTP proxy.  I uses okhttp to requst HTTP server. 
User-Agent  <---> Proxy <---> HTTP server

If HTTP server is Shoutcast server, the proxy want to receives Shoutcast protocol and then response the Shoutcast protocol. Therefore I want to distinguish the protocol whether it is Shoutcast or HTTP.

swankjesse:
What happens if you upgrade `ICY` to `HTTP`? My guess is that your clients don't break!

kenmasumitsu:
If I sent HTTP instead of ICY, the client can accept the data.  The client (typically it is music player) plays the stream with **noise**, because I guess the client ignores some of important header in Shoutcast.

swankjesse:
Noise? Like audible noise? That's amazing, I have no idea how that would be possible.

Right now I think I have to reject this because it's just such a corner case. I don't want everyone using OkHttp to be exposed to weird and obsolete `ICY` stuff. Especially since that won't be coming with us to the future with http/2.

Sorry & I hope you understand where I'm coming from.

kenmasumitsu:
Yes, it's audible noise.

I see your decision. Thanks for giving me your feedback.

PR #1548, opened 2015-04-07
Interceptor request not returned in response

quinnjn:
I found an issue when using interceptors to change the `"User-Agent"` header, that `response.request().headers()` was showing the me the old default `"User-Agent"`.

I took a peek and noticed that `networkRequest` does not update with the *possibly* modified interceptors request. 

Let me know if I'm missing something here! Thanks!

### DIFF
--- a/okhttp/src/main/java/com/squareup/okhttp/internal/http/HttpEngine.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/internal/http/HttpEngine.java
@@ -813,6 +813,9 @@ public final class HttpEngine {
 
       transport.writeRequestHeaders(request);
 
+      //Update the networkRequest with the possibly updated interceptor request.
+      networkRequest = request;
+
       if (permitsRequestBody() && request.body() != null) {
         Sink requestBodyOut = transport.createRequestBody(request, request.body().contentLength());
         BufferedSink bufferedRequestBody = Okio.buffer(requestBodyOut);
--- a/okhttp-tests/src/test/java/com/squareup/okhttp/InterceptorTest.java
+++ b/okhttp-tests/src/test/java/com/squareup/okhttp/InterceptorTest.java
@@ -477,6 +477,30 @@ public final class InterceptorTest {
     interceptorThrowsRuntimeExceptionAsynchronous(client.networkInterceptors());
   }
 
+  @Test public void networkInterceptorModifiedRequestIsReturned() throws IOException {
+    server.enqueue(new MockResponse());
+
+    Interceptor modifyHeaderInterceptor = new Interceptor() {
+      @Override public Response intercept(Chain chain) throws IOException {
+        return chain.proceed(chain.request().newBuilder()
+          .header("User-Agent", "intercepted request")
+          .build());
+      }
+    };
+
+    client.networkInterceptors().add(modifyHeaderInterceptor);
+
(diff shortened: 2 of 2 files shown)


okhttp/src/main/java/com/squareup/okhttp/internal/http/HttpEngine.java
```diff
@@ -674,6 +674,7 @@ public void readResponse() throws IOException {
       return; // No network response to read.
     }
 
+    Request sentRequest = userRequest;
```
swankjesse: Curious : do tests fail if you drop this field, and just use the initial `networkRequest` ?

quinnjn: One fails:
`URLConnectionTest.authenticateWithGetAndTransparentGzip:1720 expected:<[Successful auth!]> but was:<[
                                                                                                       .MNN-.N+ÍQH,-ÉP¤%N{]>`

swankjesse: iiiiiinteresting. I find it strange that we have a sentRequest and a networkRequest that aren't necessarily the same thing, particularly because networkRequest is supposed to be the request that is sent. But it isn't really; it's the request we sent into the interceptor chain.

I'm going to play with this change.

swankjesse:
Right now OkHttp models two different requests:
 * The one the application layer passed in. This is called the "userRequest" and it's what we surface by default.
 * The one we transmitted on the wire. This is called the "networkRequest" and it's buried in the API. It's intended for people who are mucking with low-level HTTP.

I'd like to surface the interceptor's user agent, but only under the `networkRequest`. If you want to get it as the user request, you need to register a regular interceptor, not a network interceptor.

Ie. the test would change to this:

```
  @Test public void networkInterceptorModifiedRequestIsReturned() throws IOException {

    Interceptor modifyHeaderInterceptor = new Interceptor() {
      @Override public Response intercept(Chain chain) throws IOException {
        return chain.proceed(chain.request().newBuilder()
          .header("User-Agent", "intercepted request")
          .build());
      }
    };

    client.networkInterceptors().add(modifyHeaderInterceptor);

        .url(server.getUrl("/"))
        .header("User-Agent", "user request")
        .build();

    assertNotNull(response.request().header("User-Agent"));
    assertEquals("user request", response.request().header("User-Agent"));
    assertEquals("intercepted request", response.networkResponse().request().header("User-Agent"));
  }
```

Thoughts?

swankjesse:
... And in this case, you're using a network interceptor to do an application interceptor's job. The distinction is here: https://github.com/square/okhttp/wiki/Interceptors

quinnjn:
Ahhh nice! Totally understand this better now. I'll update the PR to remove `sentRequest` since `networkRequest` is sent through `networkResponse()`.

swankjesse:
Looks good. Can you please rebaseandsqua.sh so our project's history has one commit instead of three? Also please sign our CLA if you haven't already. Then I'll merge!

swankjesse:
http://rebaseandsqua.sh

quinnjn:
Squashed + Signed CLA. Thanks!

swankjesse:
Thanks @quinnjn.

PR #1553, opened 2015-04-13
Expose running and queued call counts for monitoring

kryali:
goal: add monitoring around how many requests are queued, in-flight, and maximum capacity.

### DIFF
--- a/okhttp/src/main/java/com/squareup/okhttp/Dispatcher.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/Dispatcher.java
@@ -177,4 +177,12 @@ public final class Dispatcher {
   synchronized void finished(Call call) {
     if (!executedCalls.remove(call)) throw new AssertionError("Call wasn't in-flight!");
   }
+
+  public synchronized int getRunningCallCount() {
+    return runningCalls.size();
+  }
+
+  public synchronized int getQueuedCallCount() {
+    return readyCalls.size();
+  }
 }


okhttp/src/main/java/com/squareup/okhttp/Dispatcher.java
```diff
@@ -177,4 +177,12 @@ synchronized void executed(Call call) {
   synchronized void finished(Call call) {
     if (!executedCalls.remove(call)) throw new AssertionError("Call wasn't in-flight!");
   }
+
+  public int getRunningCallCount() {
```
swankjesse: Needs `synchronized`.

okhttp/src/main/java/com/squareup/okhttp/Dispatcher.java
```diff
@@ -177,4 +177,12 @@ synchronized void executed(Call call) {
...
+    return runningCalls.size();
+  }
+
+  public int getQueuedCallCount() {
```
swankjesse: Needs `synchronized`.

swankjesse:
Nice.

okhttp/src/main/java/com/squareup/okhttp/Dispatcher.java
kryali: Added.

Also, I have an irrational fear that metrics systems may call this too many times and then interfere with actual request processing. That's irrational right?

swankjesse: Not irrationally, but unlikely. If it comes up, we'll figure something out.

Issue #1552, opened 2015-04-13
New breakages in OkHttp tests on Android AOSP/master


nfuller:
Just putting together a pull of recent changes to OkHttp for Android. These changes failed 2 out of 2 runs on Android. The test may actually be flaky, not sure yet. All in SpdyConnection. Death of the worker thread takes down the test process, as they would (by default) an Android app.

In the past these kind of problems have not always been from the test that fails, but something happening in the tear down from a prior tests.

This is a heads up. I'll look closer and there may be fixes coming. Any insights from OkHttp about recent changes from maintainers would be appreciated. Because the problems are related to SPDY they won't necessarily stop me from committing.

WIP Android commit is here: https://android-review.googlesource.com/146428

Not looked if there are related changes in the commits I have pulled, which are:

    OkHttp:
    From b609edd07864d7191dcda8ba1f6c833c9fe170ad
    to b40f99a950cb407eff52537a97420bd253a64f63
    
    Okio:
    From 654ddf5e8f6311fda77e429c22d5e0e15f713b8d
    to b5811711b141b230e4e58f577c79cfbf4c2d4028

Stack traces below:

04-13 08:07:25.218  8560  8576 I TestRunner: started: cancelInFlightBeforeResponseReadThrowsIOE_SPDY_3(com.squareup.okhttp.CallTest)
04-13 08:07:25.218  8560  9199 E AndroidRuntime: FATAL EXCEPTION: OkHttp SpdyConnection
04-13 08:07:25.218  8560  9199 E AndroidRuntime: Process: android.core.tests.runner, PID: 8560
04-13 08:07:25.218  8560  9199 E AndroidRuntime: java.lang.RuntimeException: java.io.IOException: stream was reset: CANCEL
04-13 08:07:25.218  8560  9199 E AndroidRuntime:        at com.squareup.okhttp.internal.spdy.SpdyConnection$Reader$1.execute(SpdyConnection.java:644)
04-13 08:07:25.218  8560  9199 E AndroidRuntime:        at com.squareup.okhttp.internal.NamedRunnable.run(NamedRunnable.java:33)
04-13 08:07:25.218  8560  9199 E AndroidRuntime:        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1112)
04-13 08:07:25.218  8560  9199 E AndroidRuntime:        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:587)
[...]

Issue #1294, opened 2015-01-05
SpdyConnection IncomingStreamHandler exceptions are swallowed


JakeWharton:
```java
executor.submit(new NamedRunnable("OkHttp %s stream %d", hostName, streamId) {
  @Override public void execute() {
    try {
      handler.receive(newStream);
    } catch (IOException e) {
      throw new RuntimeException(e);
    }
  }
});
```
A `RuntimeException` was being thrown in MockWebServer's handler which caused the test to hang indefinitely with no explanation.

PR #1554, opened 2015-04-14
Fix MockWebServer with SPDY on Android.

nfuller:
Fix for issue #1552
Related to commit f78f74f from issue #1294

SpdyConnection creates a non-daemon thread to handle the readerRunnable. If
that thread dies from a RuntimeException it triggers Android's default
UncaughtExceptionHandler, which kills the test process. Previously exceptions
were being swallowed, but commit f78f74f caused the exceptions to be
propagated.

Rather than fix the problem just in SpdyConnection the fix has been
made in Util.threadFactory() for consistency. A factory method Util.newThread()
has been created and all occurances (outside of tests) of new Thread() have
been moved over.

This does alter behavior on Android across OkHttp: Previously any unexpected
thread death due to an exception or error would have taken down the Android
process. Now it will be logged to System.err, but the process will continue.

### DIFF
--- a/mockwebserver/src/main/java/com/squareup/okhttp/mockwebserver/MockWebServer.java
+++ b/mockwebserver/src/main/java/com/squareup/okhttp/mockwebserver/MockWebServer.java
@@ -667,7 +667,7 @@ public final class MockWebServer {
         .build();
 
     // The callback might act synchronously. Give it its own thread.
-    new Thread(new Runnable() {
+    Runnable webSocketWriterRunnable = new Runnable() {
       @Override public void run() {
         try {
           listener.onOpen(webSocket, fancyRequest, fancyResponse);
@@ -676,7 +676,9 @@ public final class MockWebServer {
           connectionClose.countDown();
         }
       }
-    }, "MockWebServer WebSocket Writer " + request.getPath()).start();
+    };
+    Util.newThread("MockWebServer WebSocket Writer " + request.getPath(), false /* daemon */,
+        webSocketWriterRunnable).start();
 
     // Use this thread to continuously read messages.
     while (webSocket.readMessage()) {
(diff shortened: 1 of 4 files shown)


swankjesse:
I think this is worse behavior. If OkHttp crashes due to a bug, we shouldn't stamp to sweep it under the rug. The problem is that applications will hang on never-completing progress bars, which is an aggravating user experience.

I'd rather fix the root cause!

nfuller:
Thought you might say that :-)
Not sure taking down the whole app is less aggravating, but I agree it's more likely to be noticed by developers and probably the right thing to do.

Previous behavior was, of course, to swallow the exceptions from SpdyConnection.

OkHttp has different behavior on desktop and Android due to the more aggressive behavior on Android, which is (I suspect) why you haven't seen it on the continuous build. Perhaps we should explicitly install a System.exit() default uncaught exception handler during all tests so these are more obvious to less sporadic contributors than me?

If you can suggest suitable handling in the various places affected I'll try to implement. I started with a change in the SpdyConnection$Reader Runnable (| RuntimeException), but it just died in another place instead suggesting more point changes needed. WebSocketCall may also need some handling.

Let me know if you agree and I can have a go.

swankjesse:
Yeah, I love that idea. Not sure the easiest mechanism but I'm sure you'll find something!

nfuller:
#1555 should do the trick. I'll see what I can do about this pull to make it all pass with that change in place.

swankjesse:
Cool. Let's close this and focus on #1555  ?

nfuller:
Sorry - didn't understand what you were saying and should have pushed for clarity.

#1555 is semi-independent and I think we should do separately, unless you have philosophical objections. It has the benefit for me in that your CI build will experience the similar test failures as I do when I run the tests on Android. Right now, when I do a pull there's a risk I can't submit because tests are only broken on Android. Some of those failures may be real Android issues.

Before #1555 can be submitted, a fix like #1554 needs to be applied.

PR #1555, opened 2015-04-14
Install an UncaughtExceptionHandler during tests

nfuller:
This is to make the tests more brittle around uncaught exceptions
as they are on Android. No uncaught exceptions should escape.

For issue #1552 .

Probably not necessary to say this but right now it will cause the CI to fail. Needs #1554 to pass.

### DIFF
--- a/okhttp/src/main/java/com/squareup/okhttp/internal/spdy/SpdyConnection.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/internal/spdy/SpdyConnection.java
@@ -575,7 +575,7 @@ public final class SpdyConnection implements Closeable {
         }
         connectionErrorCode = ErrorCode.NO_ERROR;
         streamErrorCode = ErrorCode.CANCEL;
-      } catch (IOException e) {
+      } catch (RuntimeException | IOException e) {
         connectionErrorCode = ErrorCode.PROTOCOL_ERROR;
         streamErrorCode = ErrorCode.PROTOCOL_ERROR;
       } finally {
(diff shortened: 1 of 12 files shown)


okhttp/src/main/java/com/squareup/okhttp/internal/spdy/SpdyConnection.java
```diff
@@ -640,8 +640,11 @@ private Reader() {
             @Override public void execute() {
               try {
                 handler.receive(newStream);
-              } catch (IOException e) {
-                throw new RuntimeException(e);
+              } catch (RuntimeException | IOException e) {
```
swankjesse: So now we're swallowing the exceptions thrown by the higher-level handler. I don't think this is what we want to do here; we're hiding bugs.

swankjesse: ...if a handler that throws a NullPointerException, then it disappears!

nfuller: Sorry - you'll have to specify what it should do. This was just a guess to get the tests to pass.

Previously, if an exception was thrown it was being swallowed everywhere.
Now on desktop it's being logged and kills the thread.
On Android it's killing the process.

swankjesse:
Sorry for sitting on this. My hope is that instead of adding this somewhat invasive test harness permanently, we can use it, fix the underlying bugs, and keep this on the shelf if/when we run into problems in the future.

nfuller:
I'd like to make a case for the UncaughtExceptionHandler change being submitted: An uncaught exception (and the runtime stopping on Android) _may_ be a real OkHttp issue, but you won't notice it on the CI using desktop Java. It'll not be until a user reports it, or until I perform a pull and the OkHttp tests are run as a CTS suite. Then there's a bit of a scramble to work out what introduced the problem and fix it. It would be nice if we caught the problems as/before they went in. An alternative would be to run the tests on Android as part of the CI, if possible.

PR #1558, opened 2015-04-15
Bugfix - Avoid null pointer when building an empty RequestBody

ajwhite:
Theoretically, if a request does not merit a request body, then one shouldn't bother creating a request body when constructing the request object. However, if one passes `null` into `RequestBody.create(mediaType, null)`, then an error becomes thrown when getting the bytes for the request body's content https://github.com/square/okhttp/blob/8d0b2b9d3ca870a5b3a2b91c59d19c1e3bfc966a/okhttp/src/main/java/com/squareup/okhttp/RequestBody.java#L54

This addresses that breaking point by providing an empty string when generating the byte array when no content is provided, producing an empty byte array that gets used when producing the network request content and content length. In a perfect world, the developer wouldn't bother building a RequestBody when there's no actual.. request body. However, we encountered an issue where our abstraction didn't check if the request body was empty, and would include this step in the builder pattern for each `PUT`, `POST`, etc regardless.

This safeguards against trying to essentially perform `(null).getBytes()`, and instead, defaults to an empty string and thus an empty `byte` array.

### DIFF
--- a/okhttp/src/main/java/com/squareup/okhttp/RequestBody.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/RequestBody.java
@@ -51,7 +51,7 @@ public abstract class RequestBody {
         contentType = MediaType.parse(contentType + "; charset=utf-8");
       }
     }
-    byte[] bytes = content.getBytes(charset);
+    byte[] bytes = (null == content ? "" : content).getBytes(charset);
     return create(contentType, bytes);
   }
 


Rican7:
Nice, simple fix. :+1:

swankjesse:
I think what we really want to do is throw at the callsite, as soon as an unwanted null enters the system. We shouldn't be lenient to bogus input, we should explode!

(You could imagine a bug in the calling code passing null when it wanted a `byte[]`. We want to crash rather than coalescing that to an empty array. Otherwise the application would reply "upload success" after losing your data.)

ajwhite:
I see your point. But at the same time HTTP spec allows for `PUT` to perform without a request body, which the library doesn't quite reflect, as there's only [`Request.Builder.put`](https://github.com/square/okhttp/blob/master/okhttp/src/main/java/com/squareup/okhttp/Request.java#L221-L223) that enforces a single `RequestBody` argument.

At this time, I'm pigeon holded into passing an empty string into `RequestBody.create(mediaType, "")` in order to make a successful `PUT` request with empty content. I could adversely call `put(null)` and avoid using `RequestBody` in a way it's not intended to be used (that.. would actually have been better on my part in where I'm using OkHttp).

At that point, a better solution to this (making a `PUT` request without a request body) might be to create an alias for `put` that doesn't require any arguments to coincide with [`put(RequestBody body)`](https://github.com/square/okhttp/blob/master/okhttp/src/main/java/com/squareup/okhttp/Request.java#L221-L223)

```java
public Builder put() {
  return put(null);
} 
```

I had done something similar for [`delete()`](https://github.com/square/okhttp/blob/master/okhttp/src/main/java/com/squareup/okhttp/Request.java#L213-L219) back over in https://github.com/square/okhttp/pull/1317. So a revised approach, such as this, might adhere to an existing precedent for empty request bodies

swankjesse:
@ajwhite ugh, you're right. We're wildly inconsistent here. Here's a fix for that:
https://github.com/square/okhttp/pull/1559/files

swankjesse:
Obsoleted by 1559!

Issue #229, opened 2013-07-03
Support custom HTTP methods with bodies.


blackdigger:
PROPFIND was added. Re-purposing this issue for arbitrary methods allowing bodies. These would obviously be forbidden from any cache interaction as no semantics can be known.

— @JakeWharton

~~Is it possible to add support for custom HTTP methods such as PROPFIND?
I realize there is a limitation of sun's implementation of HttpUrlConnection, but maybe you have some idea how to work around that?
we use apache http client in our project at the moment mainly because it allows to define custom http method by extending HttpEntityEnclosingRequestBase.
we tried to switch to OkHttp, both core version with HttpUrlConnection and OkApacheClient version - they both throw ProtocolException (probably because both use HttpUrlConnection inside)~~

maevius:
I think the implementors of HttpURLConnection got it wrong here. Clearly the spec states "extension-method". While this is an optional feature for a Server, it doesn't leave any space for a client side provider to speculate on the server side...

http://www.w3.org/Protocols/rfc2616/rfc2616-sec5.html#sec5.1
The Method token indicates the method to be performed on the resource identified by the Request-URI. The method is case-sensitive.

       Method         = "OPTIONS"                ; Section 9.2
                      | "GET"                    ; Section 9.3
                      | "HEAD"                   ; Section 9.4
                      | "POST"                   ; Section 9.5
                      | "PUT"                    ; Section 9.6
                      | "DELETE"                 ; Section 9.7
                      | "TRACE"                  ; Section 9.8
                      | "CONNECT"                ; Section 9.9
                      | extension-method
       extension-method = token

rfc2822:
Is there currently any way to use PROPFIND with okhttp?

rfc2822:
See also PR #1901. Please tell me if I can do further changes, tests, etc for this modification.

alphatan:
1. How to have:

    1. make user-defined method (supported now, by **Request.method(String, RequestBody)**) 

    2. define the restriction of permitting or not the corresponding body (**NOT** work, **Request.method(String, RequestBody)**, **HttpMethod.permitsRequestBody(method)**).

2. Problem

    As all related classes are marked as '**FINAL**' (terrible pattern), I need to hack inside the source.

    **Hope it can be more open for extension**.

3. Tested version: 

    okhttp-2.5.0


Thanks.

dahankzter:
This would be very neat! Now I have to include Apache Http Components for this isolated little thing. :(

[...]

PR #1559, opened 2015-04-15
Never convert null into an empty request body.

swankjesse:
This is a behavior change.

### DIFF
--- a/okhttp/src/main/java/com/squareup/okhttp/Request.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/Request.java
@@ -16,7 +16,6 @@
 package com.squareup.okhttp;
 
 import com.squareup.okhttp.internal.Platform;
-import com.squareup.okhttp.internal.Util;
 import com.squareup.okhttp.internal.http.HttpMethod;
 import java.io.IOException;
 import java.net.MalformedURLException;
@@ -215,7 +214,7 @@ public final class Request {
     }
 
     public Builder delete() {
-      return method("DELETE", null);
+      return delete(RequestBody.create(null, new byte[0]));
     }
 
     public Builder put(RequestBody body) {
@@ -233,8 +232,8 @@ public final class Request {
       if (body != null && !HttpMethod.permitsRequestBody(method)) {
         throw new IllegalArgumentException("method " + method + " must not have a request body.");
       }
-      if (body == null && HttpMethod.permitsRequestBody(method)) {
-        body = RequestBody.create(null, Util.EMPTY_BYTE_ARRAY);
+      if (body == null && HttpMethod.requiresRequestBody(method)) {
+        throw new IllegalArgumentException("method " + method + " must have a request body.");
       }
       this.method = method;
       this.body = body;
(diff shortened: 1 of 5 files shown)


okhttp/src/main/java/com/squareup/okhttp/Request.java
```diff
@@ -233,8 +233,8 @@ public Builder method(String method, RequestBody body) {
       if (body != null && !HttpMethod.permitsRequestBody(method)) {
         throw new IllegalArgumentException("method " + method + " must not have a request body.");
       }
-      if (body == null && HttpMethod.permitsRequestBody(method)) {
-        body = RequestBody.create(null, Util.EMPTY_BYTE_ARRAY);
+      if (body == null && HttpMethod.requiresRequestBody(method)) {
+        throw new IllegalArgumentException("method " + method + " must have a request body.");
```
ajwhite: This would create a breaking change in the API, as now it completely disallows an empty `PUT`, `POST` or `PATCH` by the definition of [`com.squareup.okhttp.internal.http.HttpMethod.requiresRequestBody`](https://github.com/square/okhttp/blob/master/okhttp/src/main/java/com/squareup/okhttp/internal/http/HttpMethod.java#L26-L30).

The big issue I was facing over in https://github.com/square/okhttp/pull/1558 was that I needed to send an empty `PUT`. I personally, as a consumer of your library, don't necessarily think it should disallow a request unless the HTTP spec explicitly states certain request methods must adhere to the existence or lack of existence of a request body, which it does not state AFAIK. 

If I can't make an empty `PUT` request, which I can right now, it's just with some unconventional and unintended approaches, then I'm simply SOL and have to go back to apache lib or something crappy. :(

---

It'd be great if we took a couple steps back and looked into allowing an empty `PUT`, and making this more accessible by having an alias `put()` in the `Request.Builder` (described in https://github.com/square/okhttp/pull/1558#issuecomment-93545013)

ajwhite: OH...

You're making it so that we can indeed pass empty body's, we must just explicitly create a RequestBody with an empty byte array as the content....

So I can still make `PUT` requests, we'd just want to have
```java
public Builder put() {
  return put(RequestBody.create(null, new byte[0]));
}
```

swankjesse: Awesome. Yeah sorry I'm terse!

swankjesse: Maybe we want a constant like RequestBody.EMPTY that is zero-bytes and has a null `Content-Type`?

JakeWharton: SGTM.

JakeWharton:
LGTM

okhttp/src/main/java/com/squareup/okhttp/Request.java
```diff
@@ -233,8 +232,8 @@ public Builder method(String method, RequestBody body) {
       if (body != null && !HttpMethod.permitsRequestBody(method)) {
         throw new IllegalArgumentException("method " + method + " must not have a request body.");
       }
-      if (body == null && HttpMethod.permitsRequestBody(method)) {
-        body = RequestBody.create(null, Util.EMPTY_BYTE_ARRAY);
+      if (body == null && HttpMethod.requiresRequestBody(method)) {
```
JakeWharton: Does this solve https://github.com/square/okhttp/issues/229?

swankjesse: Possibly?

PR #1563, opened 2015-04-16
Switch Form URL encoding builder and body to use Okio.

JakeWharton:


okhttp/src/main/java/com/squareup/okhttp/FormEncodingBuilder.java
```diff
@@ -45,12 +45,32 @@ public FormEncodingBuilder add(String name, String value) {
   }
 
   public RequestBody build() {
-    if (content.length() == 0) {
+    if (content.size() == 0) {
       throw new IllegalStateException("Form encoded body must have at least one part.");
     }
+    return new FormEncodingRequestBody(content.snapshot());
```
JakeWharton: Now that we have `snapshot()`, maybe it's worth having an overload to `RequestBody.create` which takes a `Buffer` and does a "defensive" non-copy with it?

swankjesse: I think that might be weird? Passing a mutable object in?

swankjesse: Perhaps just call this:

```
return RequestBody.create(
    MediaType.parse("application/x-www-form-urlencoded"), content.readByteArray());
```

swankjesse:
LGTM

okhttp/src/main/java/com/squareup/okhttp/FormEncodingBuilder.java
JakeWharton: I figured `ByteString` would be optimal because if it's `Segment`-backed (as it is here) we could potentially pull those immutable segments back out when writing.

swankjesse: Yup, good call. We need a `RequestBody` method that takes a ByteString.

Issue #1560, opened 2015-04-15
SPDY Request stuck


raggi:
I have an Android bugreport in which a user has all communications with a server halted due to what appears to be a hanging SSL handshake. Full report isn't provided due to confidentiality, but I can provide some more information as needed.

Relevant stacks appear to be:

```
"OkHttp <REMOVED>" prio=5 tid=22 Native
  | group="main" sCount=1 dsCount=0 obj=0x12c07940 self=0xaec4b000
  | sysTid=23547 nice=0 cgrp=bg_non_interactive sched=0/0 handle=0xaef1c800
  | state=S schedstat=( 5023230 434323 20 ) utm=0 stm=0 core=1 HZ=100
  | stack=0xa232a000-0xa232c000 stackSize=1036KB
  | held mutexes=
  kernel: (couldn't read /proc/self/task/23547/stack)
  native: #00 pc 0003b070  /system/lib/libc.so (__pselect6+20)
  native: #01 pc 00016347  /system/lib/libc.so (select+60)
  native: #02 pc 000152fb  /data/app/com.google.android.gms-1/lib/arm/libgmscore.so (???)
  native: #03 pc 000196ff  /data/app/com.google.android.gms-1/lib/arm/libgmscore.so (???)
  native: #04 pc 0207e411  /data/dalvik-cache/arm/data@app@com.google.android.gms-1@base.apk@classes.dex (Java_com_google_android_gms_org_conscrypt_NativeCrypto_SSL_1read__JLjava_io_FileDescriptor_2Lcom_google_android_gms_org_conscrypt_NativeCrypto_00024SSLHandshakeCallbacks_2_3BIII+196)
  at com.google.android.gms.org.conscrypt.NativeCrypto.SSL_read(Native method)
  at com.google.android.gms.org.conscrypt.r.read(SourceFile:698)
  - locked <@addr=0x12c001d0> (a java.lang.Object)
  at okio.Okio$2.read(Okio.java:137)
  at okio.AsyncTimeout$2.read(AsyncTimeout.java:211)
  at okio.RealBufferedSource.request(RealBufferedSource.java:71)
  at okio.RealBufferedSource.require(RealBufferedSource.java:64)
  at okio.RealBufferedSource.readInt(RealBufferedSource.java:217)
  at com.squareup.okhttp.internal.spdy.Spdy3$Reader.nextFrame(Spdy3.java:129)
[...]

PR #1570, opened 2015-04-19
Don't get stuck waiting on the connection window.

swankjesse:
We weren't applying the stream's timeout to writes to the underlying
connection. This was just a dumb bug.

When the stream does time out waiting on a write that's blocked inside
the connection awaiting write window, we need to break out.

Closes https://github.com/square/okhttp/issues/1560

### DIFF
--- a/okhttp/src/main/java/com/squareup/okhttp/internal/spdy/SpdyStream.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/internal/spdy/SpdyStream.java
@@ -501,7 +501,7 @@ public final class SpdyStream {
         writeTimeout.enter();
         try {
           while (bytesLeftInWriteWindow <= 0 && !finished && !closed && errorCode == null) {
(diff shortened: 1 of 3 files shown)


okhttp/src/main/java/com/squareup/okhttp/internal/spdy/SpdyStream.java
```diff
@@ -522,8 +527,8 @@ private void emitDataFrame(boolean outFinished) throws IOException {
       }
       while (sendBuffer.size() > 0) {
         emitDataFrame(false);
+        connection.flush();
```
swankjesse: This change is not strictly necessary, but unlikely to cause problems. If we don't flush between frames there's a risk of deadlock.

JakeWharton:
LGTM. Travis failing in something.

swankjesse:
My test was flaky. Added a ping to force the timing fixed it.

PR #1571, opened 2015-04-19
Update reader state on the reader thread.

JakeWharton:
This causes the read loop to immedaitely break–the desired behvior after receiving a close frame–without having to wait for the runnable to execute.

Closes #1516.

### DIFF
--- a/okhttp-ws-tests/src/test/java/com/squareup/okhttp/internal/ws/RealWebSocketTest.java
+++ b/okhttp-ws-tests/src/test/java/com/squareup/okhttp/internal/ws/RealWebSocketTest.java
@@ -19,7 +19,10 @@ import com.squareup.okhttp.ws.WebSocketRecorder;
 import java.io.IOException;
 import java.net.ProtocolException;
 import java.util.Random;
+import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.Executor;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
 import okio.Buffer;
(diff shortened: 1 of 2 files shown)


okhttp-ws-tests/src/test/java/com/squareup/okhttp/internal/ws/RealWebSocketTest.java
```diff
@@ -128,9 +128,9 @@
 
   @Test public void close() throws IOException {
     client.close(1000, "Hello!");
-    server.readMessage(); // This will trigger a close response.
+    assertFalse(server.readMessage()); // This will trigger a close response.
     serverListener.assertClose(1000, "Hello!");
-    client.readMessage();
+    assertFalse(client.readMessage());
```
JakeWharton: Ugh these asserts don't really assert the behavior because the executor is synchronous.

JakeWharton: Fixed!

okhttp-ws-tests/src/test/java/com/squareup/okhttp/internal/ws/RealWebSocketTest.java
```diff
@@ -300,4 +303,20 @@
     server.readMessage();
     serverListener.assertClose(1000, "Bye!");
   }
+
+  private static void waitForExecutor(Executor executor) {
```
swankjesse: Nice.

PR #1578, opened 2015-04-22
HttpUrl initial API.

swankjesse:
This is very early and subject to change. I'd like to evaluate the API first
because it's difficult to do properly and because if it's done right, the
implementation should be mostly mechanical once this is set.

### DIFF
--- a/okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
@@ -1,0 +1,302 @@
+/*
+ * Copyright (C) 2015 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.okhttp;
+
+import java.io.IOException;
+import java.net.InetAddress;
+import java.net.URI;
+import java.net.URL;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * A <a href="https://url.spec.whatwg.org/">URL</a> with an {@code http} or {@code https} scheme.
+ *
+ * TODO: discussion on canonicalization
+ *
+ * TODO: discussion on encoding-by-parts
+ *
+ * TODO: discussion on this vs. java.net.URL vs. java.net.URI
+ */
+public final class HttpUrl {
+  private HttpUrl(Builder builder) {
+    throw new UnsupportedOperationException();
+  }
+
+  public URL url() {
+    throw new UnsupportedOperationException();
+  }
+
+  public URI uri() throws IOException {
+    throw new UnsupportedOperationException();
+  }
+
+  /** Returns either "http" or "https". */
+  public String scheme() {
+    throw new UnsupportedOperationException();
+  }
+
+  public boolean isHttps() {
+    return scheme().equals("https");
+  }
+
+  public String user() {
+    throw new UnsupportedOperationException();
+  }
+
+  public String encodedUser() {
+    throw new UnsupportedOperationException();
+  }
+
+  public String password() {
+    throw new UnsupportedOperationException();
+  }
+
+  public String encodedPassword() {
+    throw new UnsupportedOperationException();
+  }
+
+  /**
+   * Returns the decoded (potentially non-ASCII) hostname. The returned string may contain non-ASCII
+   * characters and is <strong>not suitable</strong> for DNS lookups; for that use {@link
+   * #encodedHost}. For example, this may return {@code ☃.net} which is a user-displayable IDN that
+   * cannot be used for DNS lookups without encoding.
+   */
+  public String host() {
+    throw new UnsupportedOperationException();
+  }
+
+  /**
+   * Returns the host address suitable for use with {@link InetAddress#getAllByName(String)}. May
+   * be a regular host name ({@code android.com}), an IPv4 address ({@code 127.0.0.1}), an IPv6
+   * address ({@code ::1}; note that there are no square braces), or an encoded IDN ({@code
+   * xn--n3h.net}).
+   */
+  public String encodedHost() {
+    throw new UnsupportedOperationException();
+  }
+
+  /**
+   * Returns the explicitly-specified port if one was provided, or the default port for this URL's
+   * scheme. For example, this returns 8443 for {@code https://square.com:8443/} and 443 for {@code
+   * https://square.com/}.
+   */
+  public int port() {
+    throw new UnsupportedOperationException();
+  }
+
+  /**
+   * Returns 80 if {@code scheme.equals("http")}, 443 if {@code scheme.equals("https")} and -1
+   * otherwise.
+   */
+  public static int defaultPort(String scheme) {
+    throw new UnsupportedOperationException();
+  }
+
+  /**
+   * Returns the entire path of this URL, encoded for use in HTTP resource resolution. The
+   * returned path is always nonempty and is prefixed with {@code /}.
+   */
+  public String encodedPath() {
+    throw new UnsupportedOperationException();
+  }
+
+  public List<String> pathSegments() {
+    throw new UnsupportedOperationException();
+  }
+
+  /**
+   * Returns the query of this URL, encoded for use in HTTP resource resolution. The returned string
+   * may be null (for URLs with no query), empty (for URLs with an empty query) or non-empty (all
+   * other URLs).
+   */
+  public String encodedQuery() {
+    throw new UnsupportedOperationException();
+  }
+
+  /**
+   * Returns the first query parameter named {@code name} decoded using UTF-8, or null if there is
+   * no such query parameter.
+   */
+  public String queryParameter(String name) {
+    throw new UnsupportedOperationException();
+  }
+
+  public Set<String> queryParameterNames() {
+    throw new UnsupportedOperationException();
+  }
+
+  public List<String> queryParameterValues(String name) {
+    throw new UnsupportedOperationException();
+  }
+
+  public String queryParameterName(int index) {
+    throw new UnsupportedOperationException();
+  }
+
+  public String queryParameterValue(int index) {
+    throw new UnsupportedOperationException();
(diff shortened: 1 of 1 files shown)


okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
```diff
@@ -0,0 +1,294 @@
...
+  }
+
+  public String encodedUser() {
+    throw new UnsupportedOperationException();
+  }
+
+  /**
+   * Returns the decoded (potentially non-ASCII) hostname. The returned string may contain non-ASCII
+   * characters and is <strong>not suitable</strong> for DNS lookups; for that use {@link
+   * #encodedHost}. For example, this may return {@code ☃.net} which is a user-displayable IDN that
```
swankjesse: real URL. http://☃.net

swankjesse: Also, unlike Chrome, Safari shows the Unicode character in the address bar.

screen shot 2015-04-21 at 11 45 21 pm

okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
```diff
@@ -0,0 +1,294 @@
...
+   */
+  public static int defaultPort(String scheme) {
+    throw new UnsupportedOperationException();
+  }
+
+  /**
+   * Returns the entire path of this URL, encoded for use in HTTP resource resolution. The
+   * returned path is always nonempty and is prefixed with {@code /}.
+   */
+  public String encodedPath() {
```
swankjesse: Note there's no way to get the complete path, not encoded. I feel like `/foo%2Fbar/baz` is ambiguous for that case.

okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
JakeWharton: Awesome. I actually own uoʇɹɐɥʍǝʞɐɾ.com :)

okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
```diff
@@ -0,0 +1,294 @@
...
+
+  public Builder newBuilder() {
+    return new Builder(this);
+  }
+
+  /**
+   * Returns a new {@code OkUrl} representing {@code url} if it is a well-formed HTTP or HTTPS URL,
+   * or null if it isn't.
+   */
+  public static HttpUrl parse(String url) {
```
swankjesse: parse or get?

okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
JakeWharton: Ew! That's a great way to get paypal.com except with a no-space character or letter from a different alphabet in-between and phish people! Thanks, Safari.

swankjesse: that's awesome. Maybe that should be my example?

okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
```diff
@@ -0,0 +1,294 @@
...
+  /** Returns either "http" or "https". */
+  public String scheme() {
+    throw new UnsupportedOperationException();
+  }
+
+  public boolean isHttps() {
+    return scheme().equals("https");
+  }
+
+  public String user() {
```
JakeWharton: password?

okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
JakeWharton: :+1:

okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
```diff
@@ -0,0 +1,294 @@
...
+
+  @Override public int hashCode() {
+    throw new UnsupportedOperationException();
+  }
+
+  @Override public String toString() {
+    throw new UnsupportedOperationException();
+  }
+
+  public static class Builder {
```
JakeWharton: +`final`

okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
```diff
@@ -0,0 +1,294 @@
...
+  }
+
+  public String fragment() {
+    throw new UnsupportedOperationException();
+  }
+
+  /**
+   * Returns the URL that would be retrieved by following {@code link} from this URL.
+   *
+   * TODO: explain better.
```
JakeWharton: Might be able to copy/paste from https://docs.oracle.com/javase/7/docs/api/java/nio/file/Path.html#resolve(java.nio.file.Path)

swankjesse:
Right now I'm missing a public API to do general encoding & decoding of `application/x-www-form-urlencoded`.

You can accomplish the same thing via hacks:
```
HttpUrl url = HttpUrl.parse("http://foo/").newBuilder()
    .addQueryParameter("toppings", "lettuce, tomato")
    .addQueryParameter("cheese", "shredded mozzarella & cheddar")
    .build();
String requestBody = url.encodedQuery();
```

Plus similar hacks to do decoding. Or we could have a `Query` class that looks & works a lot like our Headers class.

okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
swankjesse: fixed.

okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
swankjesse: fixed

okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
```diff
@@ -0,0 +1,294 @@
...
+    throw new UnsupportedOperationException();
+  }
+
+  public List<String> pathSegments() {
+    throw new UnsupportedOperationException();
+  }
+
+  /**
+   * Returns the query of this URL, encoded for use in HTTP resource resolution. The returned string
+   * may be null (for URLs with no query), empty (for URLs with an empty query) or non-empty (all
```
JakeWharton: Interesting distinction here. You talked about how an explicit port 80 would be dropped for an implicit one. Is there reason for an empty query string to be treating differently than the lack of a query string?

okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
swankjesse: Definitely not from there. But possibly I can take guidance from here:
https://url.spec.whatwg.org/#concept-relative-url

okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
swankjesse: It's not full-unicode. There's a full spec on case folding to avoid character sequences that are visually similar. Will be hard but fun to implement. http://www.unicode.org/reports/tr46/#Introduction

okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
swankjesse: Ultimately it's about what gets transmitted over the wire. The receiving webserver can differentiate between these URLs:
```
http://publicobject.com/
http://publicobject.com/?
```

but not these ones:
```
http://publicobject.com/
http://publicobject.com:80/
```

swankjesse: (easy way to check this is to type each of the four addresses into a major browser's address bar and watch it canonicalize)

JakeWharton: Ah, yep. That makes total sense.

okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
```diff
@@ -0,0 +1,294 @@
...
+
+  /**
+   * Returns a new {@code OkUrl} representing {@code url} if it is a well-formed HTTP or HTTPS URL,
+   * or null if it isn't.
+   */
+  public static HttpUrl parse(String url) {
+    throw new UnsupportedOperationException();
+  }
+
+  public static HttpUrl get(URL url) {
```
JakeWharton: Might not be wise to use `get` here as that is what's used in `Request.Builder` for a GET request. Perhaps `from`?

JakeWharton: Granted, I'm reaching on as to whether this would actually confuse someone.

okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
JakeWharton: I like parse. There will be much more work in handling this case than conversion from URL or URI and will be much more prone to error.

okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
swankjesse: The alternative I was considering was a constructor overload. But I felt like that diminished the Builder. (And is asymmetric with `parse()` which can return null)

JakeWharton:
> Right now I'm missing a public API to do general encoding & decoding of application/x-www-form-urlencoded.

Also potentially something that could be omitted and just duplicated. I don't think there's any value in exposing something like a `Query` type in the public API.

okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
```diff
@@ -0,0 +1,294 @@
...
+
+import java.io.IOException;
+import java.net.InetAddress;
+import java.net.URI;
+import java.net.URL;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * A <a href="https://url.spec.whatwg.org/">URL</a> with an {@code http} or {@code https} scheme.
```
JakeWharton: This means that web sockets urls (`ws://localhost:9001`) can't be passed in or around with the `ws` scheme nor can this class be used to build web socket urls. Now in every case we allow 'http' in place of 'ws' and 'https' in place of 'wss' so this isn't really a concern, but I bet it comes up.

JakeWharton:
lovin' it. :shipit:

okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
swankjesse: Agreed.

The thing I like about this approach is that it's 1:1 with the URLs supported by our request/response model. It'd be weird if a page could redirect to `wss://localhost:80` and we attempted to traverse that link.

I reserve the right to change my mind on this later.

PR #1581, opened 2015-04-24
Allow an empty PUT and POST to keep compatibility

nfukuzaki:
This PR is related to https://github.com/square/okhttp/pull/1559

### DIFF
--- a/okhttp-apache/src/main/java/com/squareup/okhttp/apache/OkApacheClient.java
+++ b/okhttp-apache/src/main/java/com/squareup/okhttp/apache/OkApacheClient.java
@@ -7,11 +7,9 @@ import com.squareup.okhttp.Request;
 import com.squareup.okhttp.RequestBody;
 import com.squareup.okhttp.Response;
 import com.squareup.okhttp.ResponseBody;
-
 import java.io.IOException;
 import java.net.InetSocketAddress;
 import java.net.Proxy;
-
 import org.apache.http.Header;
 import org.apache.http.HttpEntity;
 import org.apache.http.HttpEntityEnclosingRequest;
--- a/okhttp-apache/src/main/java/com/squareup/okhttp/apache/OkApacheClient.java
+++ b/okhttp-apache/src/main/java/com/squareup/okhttp/apache/OkApacheClient.java
@@ -7,9 +7,11 @@ import com.squareup.okhttp.Request;
 import com.squareup.okhttp.RequestBody;
 import com.squareup.okhttp.Response;
 import com.squareup.okhttp.ResponseBody;
+
 import java.io.IOException;
 import java.net.InetSocketAddress;
 import java.net.Proxy;
+
 import org.apache.http.Header;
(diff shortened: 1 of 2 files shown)


okhttp-apache/src/test/java/com/squareup/okhttp/apache/OkApacheClientTest.java
```diff
@@ -113,6 +117,16 @@
     assertEquals("Hello, world!", request.getBody().readUtf8());
     assertEquals(request.getHeader("Content-Length"), "13");
   }
+  @Test public void postEmptyEntity() throws Exception {
+    server.enqueue(new MockResponse());
+    final HttpPost post = new HttpPost(server.getUrl("/").toURI());
+    client.execute(post);
```
JakeWharton: These tests don't assert anything. Copy the assertions from another test and ensure the body was indeed missing.

okhttp-apache/src/test/java/com/squareup/okhttp/apache/OkApacheClientTest.java
```diff
@@ -3,19 +3,23 @@
 import com.squareup.okhttp.mockwebserver.MockResponse;
 import com.squareup.okhttp.mockwebserver.MockWebServer;
 import com.squareup.okhttp.mockwebserver.RecordedRequest;
+
```
swankjesse: PROTIP: if you want to win the hearts & minds of open source developers, don't make unrelated formatting changes when you submit a pull request!

okhttp-apache/src/main/java/com/squareup/okhttp/apache/OkApacheClient.java
```diff
@@ -67,6 +69,8 @@ private static Request transformRequest(HttpRequest request) {
         if (encoding != null) {
           builder.header(encoding.getName(), encoding.getValue());
         }
+      } else {
```
swankjesse: When would a user of Apache HTTP client return a null entity? Isn't that a bug?

nfukuzaki: Apache HttpClient is able to send POST/PUT request without entity body.
For example following code works properly.

```java
HttpPut put = new HttpPut("URL");
HttpClient client = new DefaultHttpClient();
client.execute(put);
```

But OkApacheClient(master blanch) will throw IllegalArgumentException.
java.lang.IllegalArgumentException: method PUT must have a request body.

```java
HttpPut put = new HttpPut("URL");
OkApacheClient client = new OkApacheClient();
client.execute(put);
```

okhttp-apache/src/test/java/com/squareup/okhttp/apache/OkApacheClientTest.java
nfukuzaki: Thank you for your advice.
I will fix it.

swankjesse:
Thank you! I rebased & squashed & merged this.

PR #1586, opened 2015-04-28
Allow port number of the MockWebServerRule to be predefined.

wjam:
An example of when this feature would be required would be when testing a Spring Boot application which talks to some other web application.
When running tests in Spring, all of the configuration is read when the Spring context is first started which means that everything, including the port number of the other web application, must be the same for every Spring test.
It is possible to get hold of a free port before the tests start through the use of the `build-helper-maven-plugin`.


swankjesse:
Your test fails.

wjam:
Damn it. Worked fine locally :-/.

swankjesse:
Please reopen when you'd like me to take another look.

PR #1588, opened 2015-04-28
Add support for being able to fluently mock responses

wjam:
This adds support for being able to easily mock out some other web application. Note that this API has been written with Java 8 in mind.

```java
expectedExchanges.post("/url/to/get").with(RequestMatchers.body("expected body"::equals)
  .willReturn(new MockResponse().setBody(expected));
```


swankjesse:
This is fantastic but also outside of the scope of the project. Can you build this separately from MockWebServer?

wjam:
I presume you mean as in a separate repository rather than a separate module?

swankjesse:
Yeah. There's a lot of runway with this and I don't want to tightly couple it to the core project. Small, focused projects are our thing, and this one is already way too big.

Issue #1589, opened 2015-04-28
Vary headers not working on Android


nfuller:
This exception showed up during Android testing with ToT:

04-25 09:19:23.659  5471  5495 E AndroidRuntime: java.lang.NullPointerException: Attempt to invoke virtual method 'com.android.okhttp.Request com.android.okhttp.Response.request()' on a null object reference
04-25 09:19:23.659  5471  5495 E AndroidRuntime:         at com.android.okhttp.internal.http.OkHeaders.varyHeaders(OkHeaders.java:180)
04-25 09:19:23.659  5471  5495 E AndroidRuntime:         at com.android.okhttp.Cache$Entry.<init>(Cache.java:547)
04-25 09:19:23.659  5471  5495 E AndroidRuntime:         at com.android.okhttp.Cache.put(Cache.java:228)
04-25 09:19:23.659  5471  5495 E AndroidRuntime:         at com.android.okhttp.Cache.-wrap0(Cache.java)
04-25 09:19:23.659  5471  5495 E AndroidRuntime:         at com.android.okhttp.Cache$1.put(Cache.java:142)
04-25 09:19:23.659  5471  5495 E AndroidRuntime:         at com.android.okhttp.AndroidShimResponseCache.put(AndroidShimResponseCache.java:71)
04-25 09:19:23.659  5471  5495 E AndroidRuntime:         at android.net.http.HttpResponseCache.put(HttpResponseCache.java:209)
04-25 09:19:23.659  5471  5495 E AndroidRuntime:         at com.squareup.okhttp.internal.http.ResponseCacheAdapter.put(ResponseCacheAdapter.java:109)
04-25 09:19:23.659  5471  5495 E AndroidRuntime:         at com.squareup.okhttp.internal.http.HttpEngine.readResponse(HttpEngine.java:633)
04-25 09:19:23.659  5471  5495 E AndroidRuntime:         at com.squareup.okhttp.internal.http.HttpURLConnectionImpl.execute(HttpURLConnectionImpl.java:381)
04-25 09:19:23.659  5471  5495 E AndroidRuntime:         at com.squareup.okhttp.internal.http.HttpURLConnectionImpl.getResponse(HttpURLConnectionImpl.java:328)
04-25 09:19:23.659  5471  5495 E AndroidRuntime:         at com.squareup.okhttp.internal.http.HttpURLConnectionImpl.getHeaderFields(HttpURLConnectionImpl.java:175)
[...]

PR #1590, opened 2015-04-28
Fix for issue #1589: Vary caching on Android

nfuller:
Keep the embedded OkHttp on Android working with Vary headers:
It breaks if an app bundles OkHttp and uses the platform's
android.net.http.HttpResponseCache: the OkHttp caching
requires a networkResponse to check the vary headers and it was
not being provided, leading to a NullPointerException. Further
changes were needed to actually retain the request headers
needed for the vary check itself.

To avoid regressions the existing ResponseCacheTest has been
changed to use the real cache. Previously, the use of
InMemoryResponseCache was hiding the Vary problem.

The ResponseCacheTest is now run with an AndroidShimResponseCache
to prove the tests work with real cases. The case without the shim
is covered with the CacheTest.

Switching the test to a real cache revealed a problem in the
JavaApiConverter and bad headers, which has been fixed by
making Headers.Builder.addLenient(String, String) accessible.

### DIFF
--- a/okhttp-android-support/src/main/java/com/squareup/okhttp/internal/huc/JavaApiConverter.java
+++ b/okhttp-android-support/src/main/java/com/squareup/okhttp/internal/huc/JavaApiConverter.java
@@ -22,6 +22,7 @@ import com.squareup.okhttp.Request;
 import com.squareup.okhttp.RequestBody;
 import com.squareup.okhttp.Response;
 import com.squareup.okhttp.ResponseBody;
+import com.squareup.okhttp.internal.Internal;
 import com.squareup.okhttp.internal.Util;
 import com.squareup.okhttp.internal.http.CacheRequest;
 import com.squareup.okhttp.internal.http.HttpMethod;
@@ -41,6 +42,7 @@ import java.security.cert.Certificate;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import javax.net.ssl.HostnameVerifier;
 import javax.net.ssl.HttpsURLConnection;
 import javax.net.ssl.SSLPeerUnverifiedException;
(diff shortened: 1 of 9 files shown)


okhttp-android-support/src/main/java/com/squareup/okhttp/internal/huc/JavaApiConverter.java
```diff
@@ -59,15 +61,19 @@ private JavaApiConverter() {
...
     HttpURLConnection httpUrlConnection = (HttpURLConnection) urlConnection;
 
     Response.Builder okResponseBuilder = new Response.Builder();
 
     // Request: Create one from the URL connection.
-    // A connected HttpURLConnection does not permit access to request headers.
-    Map<String, List<String>> requestHeaders = null;
-    Request okRequest = createOkRequest(uri, httpUrlConnection.getRequestMethod(), requestHeaders);
+
+    // Android bug 20566983: Some request headers are needed for Vary caching.
```
swankjesse: Unfortunate that there's no public record for 20566983

okhttp-android-support/src/main/java/com/squareup/okhttp/internal/huc/JavaApiConverter.java
```diff
@@ -107,15 +117,72 @@ public static Response createOkResponse(URI uri, URLConnection urlConnection) th
...
+    }
+
+    Headers.Builder result = new Headers.Builder();
+    for (String fieldName : varyFields) {
+      String fieldValue = urlConnection.getRequestProperty(fieldName);
+      if (fieldValue != null) {
+        result.add(fieldName, fieldValue);
+      }
+    }
+    return OkHeaders.toMultimap(result.build(), null);
```
swankjesse: This call converts from a Headers object to a Map . . . but then in the call to createOkRequest we immediately convert back?

okhttp-android-support/src/main/java/com/squareup/okhttp/internal/huc/JavaApiConverter.java
```diff
@@ -107,15 +117,72 @@ public static Response createOkResponse(URI uri, URLConnection urlConnection) th
...
+      }
+    }
+    return builder.build();
+  }
+
+  private static Map<String, List<String>> varyHeaders(URLConnection urlConnection,
+      Headers responseHeaders) {
+
+    if (OkHeaders.hasVaryAll(responseHeaders)) {
+      // "*" means that this will be treated as uncacheable anyway.
```
swankjesse: Hmmm. . . . do we have test coverage for this case?

swankjesse:
LGTM

swankjesse:
Thanks!

swankjesse:
You said, "It breaks if an app bundles OkHttp and uses the platform's android.net.http.HttpResponseCache" . . . there's no way to do this. The bundled OkHttpClient doesn't support changing out the cache implementation.

okhttp-android-support/src/main/java/com/squareup/okhttp/internal/huc/JavaApiConverter.java
nfuller: I'll just kill the bug reference. I could reference issue #1589, which has the pertinent stack trace, but nobody will actually care in future that there was a bug here once.

okhttp-android-support/src/main/java/com/squareup/okhttp/internal/huc/JavaApiConverter.java
nfuller: True. I was trying to avoid a ton of overloads, and make use of some of the validation / eliding / trimming, etc. that Headers provides. I have inlined some code instead.

okhttp-android-support/src/main/java/com/squareup/okhttp/internal/huc/JavaApiConverter.java
nfuller: Not sure. Will add.

nfuller:
Historically wasn't the case. This stack trace suggests this is what is happening.

android.net.http.HttpResponseCache.put(HttpResponseCache.java:209)
04-25 09:19:23.659 5471 5495 E AndroidRuntime: at com.squareup.okhttp.internal.http.ResponseCacheAdapter.put(ResponseCacheAdapter.java:109)
04-25 09:19:23.659 5471 5495 E AndroidRuntime: at com.squareup.okhttp.internal.http.HttpEngine.readResponse(HttpEngine.java:633)
04-25 09:19:23.659 5471 5495 E AndroidRuntime: at com.squareup.okhttp.internal.http.HttpURLConnectionImpl.execute(HttpURLConnectionImpl.java:381)

This fix means that older OkHttp releases _should_ work. I've not tested that scenario myself. The fact that the android-support ResponseCacheTest broke when I switched to the real cache suggests there is an issue generally anyway.

nfuller:
Having gone looking for tests for "Vary: *" I discovered we were missing a lot of coverage for ResponseCache on Android around Vary headers. Probably because I'd stripped  them out when I created ResponseCacheTest from CacheTest.

A few new problems identified by the tests and now fixed. I've rearranged ResponseCacheTest to be very close to CacheTest in terms of structure so they're easier to compare.

I decided to explicitly lose support for caching Vary responses when an unusual URLConnection is passed to the ResponseCache because it would involve calling urlConnection.getRequestProperties() and we can't be sure it will work be cause the java.net.HttpURLConnection API is poor. I don't know if it has ever worked properly.

Generally wishing I could drop support for all this, but I can't.

Issue #1352, opened 2015-01-30
Certificate pinning for self signed certificates


lwasyl:
I've tried to pin certificate using example from `CertificatePinner` code, but it doesn't work. Specifically it instructs user to to:

```
String hostname = "publicobject.com";
CertificatePinner certificatePinner = new CertificatePinner.Builder()
    .add(hostname, "sha1/BOGUSPIN")
    .build();
OkHttpClient client = new OkHttpClient();
client.setCertificatePinner(certificatePinner);

Request request = new Request.Builder()
    .url("https://" + hostname)
    .build();
client.newCall(request).execute();
```
and then catch `SSLPeerUnverifiedException`. Unfortunately I only get following error: 
```
javax.net.ssl.SSLHandshakeException: java.security.cert.CertPathValidatorException: Trust anchor for certification path not found.
```

Adding `client.setSslSocketFactory()` with socket factory that doesn't check certificates at all makes code fail with `HostnameNotVerified` error, and finally adding `setHostnameVerifier` makes certificate pinner work as expected.

Is it correct way to use certificate pinning for self signed certificates?

swankjesse:
Certificate pinning adds additional constraints on which certificates are trusted. It doesn't replace the certificate-authority requirements, it adds to them. If you want to do self-signed certificates, you need to do that with a trust manager.

PR #1591, opened 2015-04-28
Add Javadoc about self-signed certificates

artem-zinnatullin:
This addition to Javadoc will help developers who want to use `CertificatePinner` to pin self-signed certificates. 

See https://github.com/square/okhttp/issues/1352

P.S. When I saw `CertificatePinner` I decided to pin self-signed certificate via without configuring `TrustManager` & `SSLSocketFactory`, of course I failed :)

But I didn't find any info about self-signed certificates in [CertificatePinner's Javadoc](http://square.github.io/okhttp/javadoc/com/squareup/okhttp/CertificatePinner.html), so I decided to help others

### DIFF
--- a/okhttp/src/main/java/com/squareup/okhttp/CertificatePinner.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/CertificatePinner.java
@@ -100,6 +100,13 @@ import static java.util.Collections.unmodifiableList;
  * complexity and limit your ability to migrate between certificate authorities.
  * Do not use certificate pinning without the blessing of your server's TLS
  * administrator!
+ *
+ * <h4>Note about self-signed certificates</h4>
+ * {@link CertificatePinner} can not be used to pin self-signed certificate
+ * if such certificate is not accepted by {@link javax.net.ssl.TrustManager}.
+ *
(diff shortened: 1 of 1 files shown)


okhttp/src/main/java/com/squareup/okhttp/CertificatePinner.java
```diff
@@ -100,6 +100,18 @@
  * complexity and limit your ability to migrate between certificate authorities.
  * Do not use certificate pinning without the blessing of your server's TLS
  * administrator!
+ * 
+ * <h3>Note about self-signed certificates</h3>
+ * {@link CertificatePinner} is second layer of certificate check,
```
swankjesse: Can you say this with fewer words and with fewer `{@link }` tags? We try to keep our docs very short and to the point so that people read them. The first sentence is all that's really essential here.

swankjesse:
Thank you. Though I'd like to hold you to a high standard for copy-editing. (We strive to that standard but don't always reach it.)

artem-zinnatullin:
Left one main sentence with just two `{@link}` + very useful link to OWASP, Checkstyle violation fixed, `<h4>` instead of `<h3>`, commits squashed and rebased into one

swankjesse:
Thank you!

Issue #1486, opened 2015-03-11
OkHttp Url API


swankjesse:
An immutable value class with a builder. Features to support:

 * [x] toUrl()
 * [x] toUri()
 * [x] toString()
 * [x] resolve(String link)
 * [x] compose from scheme, authority, path, query, fragment
 * [x] compose query parameters, which may come already-encoded or not
 * [x] Pragmatic validation. For example, permit `|` chars. See [issue 1044](https://github.com/square/okhttp/issues/1044).
 * [x] effectivePort() method that knows HTTP is 80 and HTTPS is 443
 * [x] static defaultPort(String scheme) method to lookup 80 and 443
 * [x] static parse(String) method that returns null if the input is not well formed
 * [x] Pragmatic canonicalization. `HTTP://SQUARE.COM:80` should be canonicalized to `http://square.com/`.
 * [x] equals and hashCode that leverage the pragmatic canonicalization

We need to be thoughtful around how the APIs work when data needs encoding. If I call `path()` does that return `/foo%7cbar` or `/foo|bar` ? My guess is that as a rule, we _never_ decode `%7c` into `|` because encoding and decoding URL components necessarily lives at the application level. (Otherwise consider the path `/foo%2fbar/baz`, which decodes to `/foo/bar/baz`). This strategy makes more sense when you recognize that the charset is itself unspecified and lives at the application level.

I'd like to consider deliberately not supporting URL schemes other than `http` and `https`. There's marginal benefit to handling `mailto:` and `file:` in the same class, and it means that more code would need to validate that the URL they're holding is of the type they expect. (What about `ws:` and `wss:` ?)

I'd like to support internationalized domain names, though not necessarily immediately.

What do we name it? `Url`? `HttpUrl` ? `OkUrl` ?

Resources:
http://developer.android.com/reference/java/net/URL.html
http://developer.android.com/reference/java/net/URI.html
http://developer.android.com/reference/android/net/Uri.html

JakeWharton:
Great coverage of features. You hit almost everything I'd want (read: Retrofit would want). The missing things could probably be supported on top of it (e.g., templating path parts).

JakeWharton:
Reposting https://github.com/square/okhttp/issues/1044#issuecomment-68342233, because I think it's good to 

[...]

PR #1595, opened 2015-05-01
Some initial URL test cases.

swankjesse:
It turns out that even the spec that tries to describe how browsers actually
work doesn't completely accomplish that. So I'm trying to reconcile differences
between Firefox, Chrome and the WHATWG spec. In some cases I get three different
behaviors.

This adds tests for some interesting cases.


swankjesse:
These webpages contain links that use strange links: the first contains a null character and the second contains a backspace character. I used 'em to confirm behavior in Firefox and Chrome.

https://publicobject.com/null.html
https://publicobject.com/backspace.html

JakeWharton:
LGTM

JakeWharton:
:shipit:

swankjesse:

Issue #1454, opened 2015-02-24
CertificatePinner does not support wildcards


bryanstern:
I want to pin everything at `*.example.com` to a specific certificate, but `CertificatePinner#check` does not resolve `beers.example.com` to `*.example.com` and the certificates permitted by it.

Is there a reason why OkHttp does not support this? If not, I'm more than happy to submit a PR to add wildcard support.

artem-zinnatullin:
JFYI recently I found that domain name wildcards are weird (in compare to regular wildcards in Unix).

I bought wildcard certificate for domain `*.example.com` and everything was okay for first level subdomains like `sub.example.com`, but for next sub-levels it won't work `sub.sub.example.com` won't be trusted :disappointed: 

Here is some useful information: http://security.stackexchange.com/questions/10538/what-certificates-are-needed-for-multi-level-subdomains
RFC 2818 http://tools.ietf.org/html/rfc2818

I think @swankjesse will want to have wildcards support similar to RFC 2818

P.S. very funny, that security.stackexchange.com does not use https by default

swankjesse:
Sure. The format should be consistent with hostname verifier. In particular, `*` shouldn't match through`.`.

bryanstern:
Yep, each wildcard should only match a single level. I'll take a stab this.

swankjesse:
Nice. The crazy method you'll want to use is here: https://github.com/square/okhttp/blob/master/okhttp/src/main/java/com/squareup/okhttp/internal/tls/OkHostnameVerifier.java#L165

artem-zinnatullin:
@swankjesse can I prepare PR for this?

swankjesse:
@artem-zinnatullin sure. Send the tests first!

swankjesse:
Fixed!

PR #1598, opened 2015-05-01
Wildcard hostnames support for CertificatePinner

artem-zinnatullin:
Part of #1454 

@swankjesse PTAL

### DIFF
--- a/okhttp-tests/src/test/java/com/squareup/okhttp/CertificatePinnerTest.java
+++ b/okhttp-tests/src/test/java/com/squareup/okhttp/CertificatePinnerTest.java
@@ -19,10 +19,15 @@ import com.squareup.okhttp.internal.SslContextBuilder;
 import java.security.GeneralSecurityException;
 import java.security.KeyPair;
 import java.security.cert.X509Certificate;
+import java.util.Set;
 import javax.net.ssl.SSLPeerUnverifiedException;
+import okio.ByteString;
 import org.junit.Test;
 
+import static com.squareup.okhttp.TestUtil.setOf;
+import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
@@ -32,10 +37,16 @@ public final class CertificatePinnerTest {
   static KeyPair keyPairA;
   static X509Certificate keypairACertificate1;
   static String keypairACertificate1Pin;
+  static ByteString keypairACertificate1PinBase64;
 
   static KeyPair keyPairB;
   static X509Certificate keypairBCertificate1;
   static String keypairBCertificate1Pin;
+  static ByteString keypairBCertificate1PinBase64;
+
+  static KeyPair keyPairC;
+  static X509Certificate keypairCCertificate1;
+  static String keypairCCertificate1Pin;
 
   static {
     try {
@@ -44,13 +55,23 @@ public final class CertificatePinnerTest {
       keyPairA = sslContextBuilder.generateKeyPair();
       keypairACertificate1 = sslContextBuilder.selfSignedCertificate(keyPairA, "1");
       keypairACertificate1Pin = CertificatePinner.pin(keypairACertificate1);
+      keypairACertificate1PinBase64 = pinToBase64(keypairACertificate1Pin);
 
       keyPairB = sslContextBuilder.generateKeyPair();
       keypairBCertificate1 = sslContextBuilder.selfSignedCertificate(keyPairB, "1");
       keypairBCertificate1Pin = CertificatePinner.pin(keypairBCertificate1);
+      keypairBCertificate1PinBase64 = pinToBase64(keypairBCertificate1Pin);
+
+      keyPairC = sslContextBuilder.generateKeyPair();
+      keypairCCertificate1 = sslContextBuilder.selfSignedCertificate(keyPairC, "1");
+      keypairCCertificate1Pin = CertificatePinner.pin(keypairCCertificate1);
     } catch (GeneralSecurityException e) {
       throw new AssertionError(e);
     }
+  }
+
+  static ByteString pinToBase64(String pin) {
+    return ByteString.decodeBase64(pin.substring("sha1/".length()));
   }
 
   @Test public void malformedPin() throws Exception {
@@ -134,5 +155,99 @@ public final class CertificatePinnerTest {
   @Test public void absentHostnameMatches() throws Exception {
     CertificatePinner certificatePinner = new CertificatePinner.Builder().build();
     certificatePinner.check("example.com", keypairACertificate1);
+  }
+
+  @Test public void successfulCheckForWildcardHostname() throws Exception {
+    CertificatePinner certificatePinner = new CertificatePinner.Builder()
+        .add("*.example.com", keypairACertificate1Pin)
+        .build();
+
+    certificatePinner.check("a.example.com", keypairACertificate1);
+  }
+
+  @Test public void successfulMatchAcceptsAnyMatchingCertificateForWildcardHostname() throws Exception {
+    CertificatePinner certificatePinner = new CertificatePinner.Builder()
+        .add("*.example.com", keypairBCertificate1Pin)
+        .build();
+
+    certificatePinner.check("a.example.com", keypairACertificate1, keypairBCertificate1);
+  }
+
+  @Test public void unsuccessfulCheckForWildcardHostname() throws Exception {
+    CertificatePinner certificatePinner = new CertificatePinner.Builder()
+        .add("*.example.com", keypairACertificate1Pin)
+        .build();
+
+    try {
+      certificatePinner.check("a.example.com", keypairBCertificate1);
+      fail();
+    } catch (SSLPeerUnverifiedException expected) {
+    }
+  }
+
+  @Test public void multipleCertificatesForOneWildcardHostname() throws Exception {
+    CertificatePinner certificatePinner = new CertificatePinner.Builder()
+        .add("*.example.com", keypairACertificate1Pin, keypairBCertificate1Pin)
+        .build();
+
+    certificatePinner.check("a.example.com", keypairACertificate1);
+    certificatePinner.check("a.example.com", keypairBCertificate1);
+  }
+
+  @Test public void successfulCheckForOneHostnameWithWildcardAndDirectCertificate() throws Exception {
+    CertificatePinner certificatePinner = new CertificatePinner.Builder()
+        .add("*.example.com", keypairACertificate1Pin)
+        .add("a.example.com", keypairBCertificate1Pin)
+        .build();
+
+    certificatePinner.check("a.example.com", keypairACertificate1);
+    certificatePinner.check("a.example.com", keypairBCertificate1);
+  }
+
+  @Test public void unsuccessfulCheckForOneHostnameWithWildcardAndDirectCertificate() throws Exception {
+    CertificatePinner certificatePinner = new CertificatePinner.Builder()
+        .add("*.example.com", keypairACertificate1Pin)
+        .add("a.example.com", keypairBCertificate1Pin)
+        .build();
+
+    try {
+      certificatePinner.check("a.example.com", keypairCCertificate1);
+      fail();
+    } catch (SSLPeerUnverifiedException expected) {
+    }
+  }
+
+  @Test public void successfulFindMatchingPins() {
+    CertificatePinner certificatePinner = new CertificatePinner.Builder()
+        .add("first.com", keypairACertificate1Pin, keypairBCertificate1Pin)
+        .add("second.com", keypairCCertificate1Pin)
+        .build();
+
+    Set<ByteString> expectedPins = setOf(keypairACertificate1PinBase64, keypairBCertificate1PinBase64);
+    Set<ByteString> matchedPins  = certificatePinner.findMatchingPins("first.com");
+
+    assertEquals(expectedPins, matchedPins);
+  }
+
+  @Test public void successfulFindMatchingPinsForWildcardAndDirectCertificates() {
+    CertificatePinner certificatePinner = new CertificatePinner.Builder()
+        .add("*.example.com", keypairACertificate1Pin)
+        .add("a.example.com", keypairBCertificate1Pin)
+        .add("b.example.com", keypairCCertificate1Pin)
+        .build();
+
+    Set<ByteString> expectedPins = setOf(keypairACertificate1PinBase64, keypairBCertificate1PinBase64);
+    Set<ByteString> matchedPins  = certificatePinner.findMatchingPins("a.example.com");
+
+    assertEquals(expectedPins, matchedPins);
+  }
+
+  @Test public void wildcardHostnameShouldNotMatchThroughDot() throws Exception {
+    CertificatePinner certificatePinner = new CertificatePinner.Builder()
+        .add("*.example.com", keypairACertificate1Pin)
+        .build();
+
+    assertNull(certificatePinner.findMatchingPins("example.com"));
+    assertNull(certificatePinner.findMatchingPins("a.b.example.com"));
   }
 }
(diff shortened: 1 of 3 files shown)


okhttp-tests/src/test/java/com/squareup/okhttp/CertificatePinnerTest.java
```diff
@@ -135,4 +143,73 @@
...
+        .build();
+
+    try {
+      certificatePinner.check("a.example.com", keypairCCertificate1);
+      fail();
+    } catch (SSLPeerUnverifiedException expected) {
+    }
+  }
+
+  @Test public void wildcardHostnameShouldNotWorkThroughDot() throws Exception {
```
swankjesse: Nice.

okhttp-tests/src/test/java/com/squareup/okhttp/CertificatePinnerTest.java
```diff
@@ -135,4 +143,73 @@
...
+    CertificatePinner certificatePinner = new CertificatePinner.Builder()
+        .add("*.example.com", keypairACertificate1Pin)
+        .build();
+
+    certificatePinner.check("example.com", keypairBCertificate1);
+    certificatePinner.check("a.b.example.com", keypairBCertificate1);
```
swankjesse: shouldn't this fail?

okhttp-tests/src/test/java/com/squareup/okhttp/CertificatePinnerTest.java
```diff
@@ -135,4 +143,73 @@
...
+    } catch (SSLPeerUnverifiedException expected) {
+    }
+  }
+
+  @Test public void wildcardHostnameShouldNotWorkThroughDot() throws Exception {
+    CertificatePinner certificatePinner = new CertificatePinner.Builder()
+        .add("*.example.com", keypairACertificate1Pin)
+        .build();
+
+    certificatePinner.check("example.com", keypairBCertificate1);
```
swankjesse: shouldn't this fail?

okhttp-tests/src/test/java/com/squareup/okhttp/CertificatePinnerTest.java
```diff
@@ -135,4 +143,73 @@
...
+  @Test public void multipleCertificatesForOneWildcardHostname() throws Exception {
+    CertificatePinner certificatePinner = new CertificatePinner.Builder()
+        .add("*.example.com", keypairACertificate1Pin, keypairBCertificate1Pin)
+        .build();
+
+    certificatePinner.check("a.example.com", keypairACertificate1);
+    certificatePinner.check("a.example.com", keypairBCertificate1);
+  }
+
+  @Test public void successfulCheckForOneHostnameWithWildcardAndDirectCertificate() throws Exception {
```
swankjesse: neat

swankjesse:
Excellent. Are the expectations correct for matching through dots? Other than that, LGTM

okhttp-tests/src/test/java/com/squareup/okhttp/CertificatePinnerTest.java
artem-zinnatullin: Current implementation of `CertificatePinner` just skips hostnames without pinned certificates so we can not understand "hostname check passed, or hostname just doesn't have pinned certificates". 

May be `CertificatePinner#check()` should also return `boolean` in addition to throwing exception: `true` if hostname has pinned certificates and check passed and `false` if no pinned certificates were found? 

It will allow us to check it's behavior and fail such tests

okhttp-tests/src/test/java/com/squareup/okhttp/CertificatePinnerTest.java
artem-zinnatullin: Same as below

okhttp-tests/src/test/java/com/squareup/okhttp/CertificatePinnerTest.java
swankjesse: Ahh, yes of course. Works for me.

swankjesse:
Sweet, this is good to merge. Unfortunately our tests are now red, so I present you two fun options!

1. Add another commit to implement this feature, same pull request.
2. Amend the current commit to add `@Ignore` tests for the new tests. You can implement at your leisure.

okhttp-tests/src/test/java/com/squareup/okhttp/CertificatePinnerTest.java
artem-zinnatullin: Can I make this change (boolean result) as part of PR for #1454?

artem-zinnatullin:
I don't think that merging tests for not implemented feature is good idea :) 

I prefer 1st variant, you can close this PR and we'll reopen it when I'll prepare implementation

swankjesse:
Works for me. Looking forward to the implementation!

artem-zinnatullin:
@swankjesse can you please reopen this PR? I don't have access rights for that, or I can create new one  (GitHub suggests me to do so)

artem-zinnatullin:
Comment to implementation:

1. Now all tests for `CertificatePinner` checks "hostname has pinned certificates or not".
2. According to [this](http://security.stackexchange.com/questions/10538/what-certificates-are-needed-for-multi-level-subdomains), [this](http://serverfault.com/questions/104160/wildcard-ssl-certificate-for-second-level-subdomain) and [RFC 6125](https://tools.ietf.org/html/rfc6125) hostname can contain multiple `*` wildcard symbols, but most of Http Clients support only the complete left-most label (e.g., *.example.com),  `OkHostnameVerifier`: "Asterisk (*) is only permitted in the left-most domain name label"

So with the rule "Asterisk (*) is only permitted in the left-most domain name label", I was able to implement check algorithm with complexity `O(1)`, it's also possible to use  `OkHostnameVerifier#verifyHostname(hostname, pattern)` as @swankjesse suggested in #1454 but `CertificatePinner#check(hostname)` will have complexity `O(n)` where `n` is number of pinned hostnames, also`OkHostnameVerifier#verifyHostname(hostname, pattern)` does a lot of unneeded (for `CertificatePinner`) checks.

Build on Travis failed with message "The forked VM terminated without properly saying goodbye. VM crash or System.exit called" on both of environments, I can build my branch locally without problems..

swankjesse:
(our builds are currently flaky; if your new tests pass locally then the red build is probably unrelated.)

artem-zinnatullin:
Yep, I also merged this PR with current master -> tests pass

okhttp/src/main/java/com/squareup/okhttp/CertificatePinner.java
```diff
@@ -123,17 +123,45 @@ private CertificatePinner(Builder builder) {
    * pinned for {@code hostname}. OkHttp calls this after a successful TLS
    * handshake, but before the connection is used.
    *
+   * @return {@code true} if hostname has pinned certificates and check passed,
```
swankjesse: After all discussion, I think we should probably skip the boolean here. We don't need it, and it's a bit confusing because there's three states, but most users would expect only two. (Okay, not okay).

okhttp/src/main/java/com/squareup/okhttp/CertificatePinner.java
```diff
@@ -123,17 +123,45 @@ private CertificatePinner(Builder builder) {
...
+
+    List<ByteString> directPins = hostnameToPins.get(hostname);
+    List<ByteString> wildcardPins = null;
+
+    int indexOfFirstDot = hostname.indexOf('.');
+
+    // Skip hostNames with one dot symbol for wildcard pattern search
+    //   example.com   will  be skipped
+    //   a.example.com won't be skipped
+    if (indexOfFirstDot != hostname.lastIndexOf('.')) {
```
swankjesse: extract a field like `indexOfLastDot` for symmetry? Right now this is not as intuitive.

okhttp/src/main/java/com/squareup/okhttp/CertificatePinner.java
```diff
@@ -123,17 +123,45 @@ private CertificatePinner(Builder builder) {
...
+      // a.example.com -> search for wildcard pattern *.example.com
+      wildcardPins = hostnameToPins.get("*." + hostname.substring(indexOfFirstDot + 1));
```
swankjesse: Dig.

okhttp/src/main/java/com/squareup/okhttp/CertificatePinner.java
```diff
@@ -123,17 +123,45 @@ private CertificatePinner(Builder builder) {
...
+    }
+
+    if (directPins == null && wildcardPins == null) return false;
+
+    // Optimization: allocate only required amount of memory
```
swankjesse: Don't bother with this optimization. We're not holding this memory for very long!

okhttp/src/main/java/com/squareup/okhttp/CertificatePinner.java
```diff
@@ -180,9 +208,12 @@ private static ByteString sha1(X509Certificate x509Certificate) {
     private final Map<String, List<ByteString>> hostnameToPins = new LinkedHashMap<>();
 
     /**
-     * Pins certificates for {@code hostname}. Each pin is a SHA-1 hash of a
-     * certificate's Subject Public Key Info, base64-encoded and prefixed with
-     * "sha1/".
+     * Pins certificates for {@code hostname}.
+     *
+     * @param hostname lower-case host name or wildcard pattern such as {@code *.example.com}.
```
swankjesse: Please add a big note to warn that `*.example.com` doesn't match `example.com`; that's likely to trip up some users.

okhttp/src/main/java/com/squareup/okhttp/CertificatePinner.java
swankjesse: (if you really wanted this to run fast, you'd just run two loops)

swankjesse:
I like where this is going. Comments!

okhttp/src/main/java/com/squareup/okhttp/CertificatePinner.java
artem-zinnatullin: Just to be sure: you mean remove it from javadoc or remove it from the code too? (without this boolean tests won't be able to check that behavior is correct, but it's okay for `CertificatePinner#check()` general contract), I agree that it's a bit confusing

swankjesse: Yeah. I think if you want to make this more testable, extract a package-protected method that returns a `List<ByteString>` of matching pins? (The method does two things: it finds pins to match against, and then it matches against them. You want to test each independently.)

okhttp/src/main/java/com/squareup/okhttp/CertificatePinner.java
artem-zinnatullin: Done, added comment from `OkHostnameVerifier`, I think that APIs used for "security things" should be well documented, especially `CertificatePinner` because with incorrect `hostname` definition it'll just skip check and MiTM attack will be possible, also, you can create an issue for `hostname` validation in `CertificatePinner.Builder#add()` it'll allow to prevent incorrect usage of `CertificatePinner`

okhttp/src/main/java/com/squareup/okhttp/CertificatePinner.java
artem-zinnatullin: Fixed (agree), thanks for pointing that

okhttp/src/main/java/com/squareup/okhttp/CertificatePinner.java
artem-zinnatullin: Done (agree)

okhttp/src/main/java/com/squareup/okhttp/CertificatePinner.java
artem-zinnatullin: Done -> extracted method `findMatchingPins(hostname)`, added several tests for it. I left tests for `check()` with wildcard pattern, because I think that `CertificatePinner` should be well tested.

Thank you for preventing me from making API worse :+1:

okhttp-tests/src/test/java/com/squareup/okhttp/TestUtil.java
```diff
@@ -15,4 +16,8 @@ private TestUtil() {
     }
     return result;
   }
+
+  public static boolean isEqualCollections(Collection<?> a, Collection<?> b) {
```
swankjesse: This isn't well defined. For example, this says that the following lists are equal: [A, B, C], [B, A, A]. I think you want to fix the call site to do set equality, which is well defined.

Perhaps add this helper method:

```
public <T> Set<T> setOf(T... elements) {
  return setOf(Arrays.asList(elements)); 
}
public <T> Set<T> setOf(Iterable<T> elements) {
  return new LinkedHashSet<>(elements); 
}
```

And then you can do this:

```
assertEquals(setOf(keypairACertificate1PinBase64, keypairBCertificate1PinBase64), setOf(matchedPins));
```

This isn't off the top of my head; I've run into this before!

swankjesse: Or even better, just change matchingPins to return a `Set<ByteString>` instead of a `List<ByteString>`. The cost difference is negligible, and semantically a set is more correct. (you don't have duplicates; you don't care about order).

(Please use `LinkedHashSet`; I'm addicted to determinism)

okhttp/src/main/java/com/squareup/okhttp/CertificatePinner.java
```diff
@@ -180,9 +215,23 @@ private static ByteString sha1(X509Certificate x509Certificate) {
...
+     *                 <p>Wildcard pattern rules:
```
swankjesse: Wanna extract this out to a top-level comment, at the top of the class? This is interesting stuff and hiding it in the `@param` doc is buring a cool feature.

Also want to use `<ol><li>` instead of `<p>1.` ? That's what we do elsewhere.

okhttp/src/main/java/com/squareup/okhttp/CertificatePinner.java
```diff
@@ -180,9 +215,23 @@ private static ByteString sha1(X509Certificate x509Certificate) {
...
+     *                 <p>1. Asterisk {@code *} is only permitted in the left-most
+     *                 domain name label and must be the only character in that label
+     *                 (i.e., must match the whole left-most label). For example,
+     *                 {@code *.example.com} is permitted, while {@code *a.example.com},
+     *                 {@code a*.example.com}, {@code a*b.example.com}, {@code a.*.example.com}
+     *                 are not permitted.
+     *                 <p>2. Asterisk {@code *} cannot match across domain name labels.
```
swankjesse: Great doc though. Thankyou

swankjesse:
I'm eager to merge this, but I have a few more ideas to make it even stronger. If you're getting tired of the back & forth, just say so & I'll merge!

okhttp-tests/src/test/java/com/squareup/okhttp/TestUtil.java
artem-zinnatullin: Damn, I just woke up with thoughts about this problem  of comparing collections, I'll change `List` to the `Set`. 

Yeah, I asked Jake some time ago why he use `LinkedHashSet` (Linked versions of collections in general)  instead of `HashSet`,  he said same thing about determinism. Agree.

artem-zinnatullin: Done!

okhttp/src/main/java/com/squareup/okhttp/CertificatePinner.java
artem-zinnatullin: Done!

artem-zinnatullin:
Rebased and squashed into one commit

okhttp/src/main/java/com/squareup/okhttp/CertificatePinner.java
```diff
@@ -91,8 +92,28 @@
  *       .build();
  * }</pre>
  *
- * Pinning is per-hostname. To pin both {@code publicobject.com} and {@code
- * www.publicobject.com}, you must configure both hostnames.
+ * Pinning is per-hostname and/or per-wildcard pattern. To pin both
+ * {@code publicobject.com} and {@code www.publicobject.com}, you must
+ * configure both hostnames.
+ *
+ * <p>Wildcard pattern rules:
```
swankjesse: This is wonderful.

swankjesse:
Thanks!

PR #1601, opened 2015-05-04
HttpURL parsing.

swankjesse:
This includes everything but character encoding. The implementation
is built around one oversized method that runs top to bottom; I found
this was easier to follow than the switch statement recommended by the
URL standard.

### DIFF
--- a/okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
@@ -32,70 +32,111 @@ import java.util.Set;
  * TODO: discussion on this vs. java.net.URL vs. java.net.URI
  */
 public final class HttpUrl {
-  private HttpUrl(Builder builder) {
-    throw new UnsupportedOperationException();
+  /** Either "http" or "https". */
+  private final String scheme;
+
+  /** Encoded username. */
+  private final String username;
+
+  /** Encoded password. */
+  private final String password;
+
+  /** Encoded hostname. */
+  // TODO(jwilson): implement punycode.
+  private final String host;
+
+  /** Either 80, 443 or a user-specified port. In range [1..65535]. */
+  private final int port;
+
+  /** Encoded path. */
+  private final String path;
+
+  /** Encoded query. */
+  private final String query;
+
+  /** Encoded fragment. */
+  private final String fragment;
+
+  /** Canonical URL. */
+  private final String url;
+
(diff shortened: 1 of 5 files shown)


okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
```diff
@@ -32,70 +32,107 @@
...
   }
 
   /**
-   * Returns the decoded (potentially non-ASCII) hostname. The returned string may contain non-ASCII
-   * characters and is <strong>not suitable</strong> for DNS lookups; for that use {@link
-   * #encodedHost}. For example, this may return {@code ☃.net} which is a user-displayable IDN that
-   * cannot be used for DNS lookups without encoding.
+   * Returns the host address suitable for use with {@link InetAddress#getAllByName(String)}. May
+   * be a regular host name ({@code android.com}), an IPv4 address ({@code 127.0.0.1}), an IPv6
+   * address ({@code ::1}; note that there are no square braces), or an encoded IDN ({@code
```
JakeWharton: missing closing parenthesis on IPv6 example

swankjesse: Fixed.

okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
```diff
@@ -228,75 +281,412 @@ public Builder encodedPassword(String encodedPassword) {
...
+
+      return new HttpUrl(scheme, username, effectivePassword, host, effectivePort, effectivePath,
+          query, fragment, url.toString());
+    }
+
+    HttpUrl parse(HttpUrl base, String input) {
+      int pos = 0;
+      int limit = input.length();
+      pos += leadingAsciiWhitespace(input, pos, limit);
+      limit = trailingAsciiWhitespace(input, pos, limit);
```
JakeWharton: Both of these methods are one-shots. Why not change behavior so you can do:
```java
int pos = skipLeadingAsciiWhitespace(input);
int limit = skipTrailingAsciiWhitespace(input, pos);
```

JakeWharton: The only reason I commented was the non-symmetric `

swankjesse: Oooh, I like that better.

JakeWharton:
LGTM

swankjesse:

PR #1602, opened 2015-05-04
Implement percent encoding.

swankjesse:
This works for all URL components except for host names. Those are
subject to other types of encoding.

### DIFF
--- a/okhttp-tests/src/test/java/com/squareup/okhttp/UrlComponentEncodingTester.java
+++ b/okhttp-tests/src/test/java/com/squareup/okhttp/UrlComponentEncodingTester.java
@@ -43,11 +43,11 @@ class UrlComponentEncodingTester {
     map.put(       0x6, Encoding.PERCENT); // Acknowledgment
     map.put(       0x7, Encoding.PERCENT); // Bell
     map.put((int) '\b', Encoding.PERCENT); // Backspace
-    map.put((int) '\t', Encoding.PERCENT); // Horizontal Tab
-    map.put((int) '\n', Encoding.PERCENT); // Line feed
+    map.put((int) '\t', Encoding.SKIP);    // Horizontal Tab
+    map.put((int) '\n', Encoding.SKIP);    // Line feed
     map.put(       0xb, Encoding.PERCENT); // Vertical Tab
-    map.put((int) '\f', Encoding.PERCENT); // Form feed
-    map.put((int) '\r', Encoding.PERCENT); // Carriage return
+    map.put((int) '\f', Encoding.SKIP);    // Form feed
+    map.put((int) '\r', Encoding.SKIP);    // Carriage return
     map.put(       0xe, Encoding.PERCENT); // Shift Out
     map.put(       0xf, Encoding.PERCENT); // Shift In
     map.put(      0x10, Encoding.PERCENT); // Data Link Escape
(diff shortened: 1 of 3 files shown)


swankjesse:

okhttp-tests/src/test/java/com/squareup/okhttp/UrlComponentEncodingTester.java
```diff
@@ -280,28 +282,31 @@ public String encode(int codePoint) {
...
       @Override public String decodedValue(HttpUrl url) {
-        return url.path().substring(1);
+        String path = url.path();
+        return path.substring(2, path.length() - 2);
       }
     },
     QUERY {
       @Override public String urlString(String value) {
-        throw new UnsupportedOperationException("TODO");
+        return "http://example.com/?a" + value + "z/";
```
JakeWharton: You don't need the trailing slash here and I think it only serves to confuse.

okhttp-tests/src/test/java/com/squareup/okhttp/UrlComponentEncodingTester.java
```diff
@@ -280,28 +282,31 @@ public String encode(int codePoint) {
...
       @Override public String decodedValue(HttpUrl url) {
-        throw new UnsupportedOperationException("TODO");
+        String query = url.query();
+        return query.substring(1, query.length() - 2);
       }
     },
     FRAGMENT {
       @Override public String urlString(String value) {
-        throw new UnsupportedOperationException("TODO");
+        return "http://example.com/#a" + value + "z/";
```
JakeWharton: You don't need the trailing slash here and I think it only serves to confuse.

okhttp-tests/src/test/java/com/squareup/okhttp/UrlComponentEncodingTester.java
```diff
@@ -280,28 +282,31 @@ public String encode(int codePoint) {
     },
     PATH {
       @Override public String urlString(String value) {
-        return "http://example.com/" + value;
+        return "http://example.com/a" + value + "z/";
```
JakeWharton: I guess technically you don't need it here either... but it at least makes sense here.

okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
```diff
@@ -661,21 +665,22 @@ private static int delimiterOffset(String input, int pos, int limit, String deli
...
+    private String password(String input, int pos, int limit) {
+      return encode(input, pos, limit, " \"':;<=>@[]\\^`{}|");
     }
 
     private static String host(String input, int pos, int limit) {
       return input.substring(pos, limit); // TODO: encode
     }
 
-    private static int port(String input, int pos, int limit) {
+    private int port(String input, int pos, int limit) {
```
JakeWharton: Why the static change in all these methods? I didn't see anything that referenced a field.

JakeWharton:
LGTM

okhttp-tests/src/test/java/com/squareup/okhttp/UrlComponentEncodingTester.java
swankjesse: Fixed.

okhttp-tests/src/test/java/com/squareup/okhttp/UrlComponentEncodingTester.java
swankjesse: Fixed.

okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
swankjesse: I'd experimented with making `utf8Buffer` a field of the builder, rather than a throwaway object used by `encode()`. Changed my mind but didn't put the statics back. I don't feel strongly either way!

PR #1607, opened 2015-05-05
Add toMultimap() method to Headers class

StephenBarlow:


okhttp/src/main/java/com/squareup/okhttp/Headers.java
```diff
@@ -130,6 +131,20 @@ public Builder newBuilder() {
     return result.toString();
   }
 
+  public Map<String, List<String>> toMultimap() {
+    HashMap<String, List<String>> result = new HashMap<String, List<String>>();
+    for (int i = 0, size = size(); i < size; i++) {
+      if (result.containsKey(name(i))) {
```
JakeWharton: You can avoid the multiple map lookups here by doing:
```java
String name = name(i);
List<String> values = result.get(name);
if (values == null) {
  values = new ArrayList<>(2);
  result.put(name, values);
}
values.add(value(i));
```

okhttp/src/main/java/com/squareup/okhttp/Headers.java
```diff
@@ -130,6 +131,20 @@ public Builder newBuilder() {
     return result.toString();
   }
 
+  public Map<String, List<String>> toMultimap() {
+    HashMap<String, List<String>> result = new HashMap<String, List<String>>();
```
JakeWharton: As a general rule we always use `LinkedHashMap`, and in this case you'll get predictable iteration order which corresponds to the first occurrence of each header.

okhttp/src/main/java/com/squareup/okhttp/Headers.java
```diff
@@ -130,6 +131,20 @@ public Builder newBuilder() {
     return result.toString();
   }
 
+  public Map<String, List<String>> toMultimap() {
+    LinkedHashMap<String, List<String>> result = new LinkedHashMap<String, List<String>>();
+    for (int i = 0, size = size(); i < size; i++) {
+      String name = name(i);
+      List<String> values = result.get(name);
+        if (values == null) {
```
swankjesse: nit: I don't think this needs to be indented

okhttp/src/main/java/com/squareup/okhttp/Headers.java
```diff
@@ -130,6 +131,20 @@ public Builder newBuilder() {
     return result.toString();
   }
 
+  public Map<String, List<String>> toMultimap() {
+    LinkedHashMap<String, List<String>> result = new LinkedHashMap<String, List<String>>();
```
swankjesse: nit: by convention we use interface type on LHS, like this:

```
Map<String, List<String>> result = new LinkedHashMap<>();
```

swankjesse:
Sorry for the parade of nitpicks. One last one – do you mind doing a simple test case? HeadersTest is already ready to go.

swankjesse:
Oh and one one more nitpick: please use http://rebaseandsqua.sh so we get only one commit in our git history!

okhttp-tests/src/test/java/com/squareup/okhttp/internal/http/HeadersTest.java
```diff
@@ -302,4 +302,14 @@
     } catch (IllegalArgumentException expected) {
     }
   }
+
+  @Test public void toMultimapGroupsHeaders() {
+    Headers headers = Headers.of(
+        "cache-control", "no-cache",
```
swankjesse: dig.

okhttp/src/main/java/com/squareup/okhttp/Headers.java
```diff
@@ -130,6 +131,20 @@ public Builder newBuilder() {
     return result.toString();
   }
 
+  public Map<String, List<String>> toMultimap() {
+    Map<String, List<String>> result = new LinkedHashMap<String, List<String>>();
+    for (int i = 0, size = size(); i < size; i++) {
```
swankjesse: Nice.

swankjesse:
Thank you!

PR #1612, opened 2015-05-08
IDNA mapping. Converts Σ.com to σ.com.

swankjesse:


okhttp/src/main/java/com/squareup/okhttp/IdnaMappingTable.java
```diff
@@ -0,0 +1,394 @@
...
+ * <p>This documentation refers to "unsigned short". Java's unsigned short type is named 'char',
+ * but none of the arrays in this class contain UTF-16 code points and should not be interpreted
+ * as text.
+ */
+final class IdnaMappingTable {
+  private static final BitPackingType[] BIT_PACKING_TYPE_VALUES = BitPackingType.values();
+  private static final Status[] STATUS_VALUES = Status.values();
+  private static final Idna2008Status[] IDNA_2008_STATUS_VALUES = Idna2008Status.values();
+  public static final IdnaMappingTable INSTANCE;
+  static {
```
JakeWharton: Not sure if this is possible, but it would be nice if this was class loaded when OkHttp's thread canonicalizes an `HttpUrl` instance rather than on the caller's thread. Resource loading is woefully slow on Android and you might cause a stutter on the main thread, potentially at app launch if there's background fetches to refresh data.

okhttp/src/main/java/com/squareup/okhttp/IdnaMappingTable.java
```diff
@@ -0,0 +1,394 @@
...
+    for (char c : rangeStarts) {
+      out.writeShort(c);
+    }
+    for (char c : rangeValues) {
+      out.writeShort(c);
+    }
+
+    Buffer replacementsUtf8 = new Buffer().writeUtf8(replacements);
+    out.writeInt((int) replacementsUtf8.size());
+    out.write(replacementsUtf8, replacementsUtf8.size());
```
JakeWharton: could just be `writeAll` not that you really gain much...

okhttp/src/main/java/com/squareup/okhttp/IdnaMappingTable.java
```diff
@@ -0,0 +1,394 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import okio.Buffer;
+import okio.BufferedSink;
+import okio.BufferedSource;
+import okio.Okio;
+
+/**
+ * The <a href="http://www.unicode.org/reports/tr46/#IDNA_Mapping_Table">IDNA mapping table</a>
```
JakeWharton: How frequently does this document update? There's no instructions or convenience mechanism for generating a new copy of the binary file. Perhaps a file in the tests?

okhttp/src/main/java/com/squareup/okhttp/IdnaMappingTable.java
swankjesse: My goal is to put a small shim in front of this so that this class only gets loaded if your URL contains characters outside of `[A-Za-z0-9.-]`. Though I know it's tough to prevent classes from being loaded!

okhttp/src/main/java/com/squareup/okhttp/IdnaMappingTable.java
JakeWharton: Seems like the `readText` method (and friends) could belong there as well since that would be the only file in which the text format was referenced. #caloriecounting

okhttp/src/main/java/com/squareup/okhttp/IdnaMappingTable.java
swankjesse: Yes! Think there's much utility to bulk APIs for arrays? To potentially speed up `out.writeShort()` calls in the above loops.

okhttp-tests/src/test/java/com/squareup/okhttp/IdnaMappingTableTest.java
```diff
@@ -0,0 +1,71 @@
...
+package com.squareup.okhttp;
+
+import okio.Buffer;
+import okio.BufferedSource;
+import okio.Okio;
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+
+public final class IdnaMappingTableTest {
```
JakeWharton: You are missing the required snowman character test...

JakeWharton:
LGTM

swankjesse:
I'm going to hold on this for now in favor of the IDN 2003 spec already offered by Java 6.

Issue #1613, opened 2015-05-09
HttpUrl IPv6 parsing doesn't fail on too many leading 0s


swankjesse:
This URL is malformed but parses successfully: `http://[::00001]`.

For reference, here's a (more) correct implementation that I wrote. I'm not committing it to save code.

```
    private static InetAddress decodeIpv6(String input, int pos, int limit) {
      byte[] address = new byte[16];
      int b = 0;
      int compress = -1;

      for (int i = pos; i < limit; ) {
        if (b == address.length) return null; // Too many groups.

        // Read a delimiter.
        if (i + 2 <= limit && input.regionMatches(i, "::", 0, 2)) {
          // Compression "::" delimiter, which is anywhere in the input, including its prefix.
          if (compress != -1) return null; // Multiple "::" delimiters.
          i += 2;
          b += 2;
          compress = b;
          if (i == limit) break;
        } else if (b != 0) {
          // Group separator ":" delimiter.
          if (input.regionMatches(i, ":", 0, 1)) {
            i++;
          } else if (input.regionMatches(i, ".", 0, 1)) {
            return decodeIpv4(input, pos, limit); // Not the delimiter we expected, but not fatal.
          } else {
            return null; // Wrong delimiter.
          }
        }

        // Read a group, one to four hex digits.
        int value = 0;
        int groupOffset = i;
        for (; i < limit; i++) {
          char c = input.charAt(i);
          int hexDigit = decodeHexDigit(c);
          if (hexDigit == -1) break;
          value = (value << 4) + hexDigit;
        }
        int groupLength = i - groupOffset;
        if (groupLength == 0 || groupLength > 4) return null; // Group is the wrong size.

        // Success. Assign the group's value to our byte array.
        address[b++] = (byte) ((value >>> 8) & 0xff);
        address[b++] = (byte) (value & 0xff);
      }

      // All done. If compression happened, we need to move bytes to the right place in the
      // address. Here's a sample:
      //
      //      input: "1111:2222:3333::7777:8888"
      //     before: { 11, 11, 22, 22, 33, 33, 00, 00, 77, 77, 88, 88, 00, 00, 00, 00  }
      //   compress: 6
      //          b: 10
      //      after: { 11, 11, 22, 22, 33, 33, 00, 00, 00, 00, 00, 00, 77, 77, 88, 88 }
      //
      if (b != address.length) {
        if (compress == -1) return null; // Address didn't have compression or enough groups.
[...]

PR #1614, opened 2015-05-09
Parse IPv6 addresses.

swankjesse:
This leans on InetAddress to do the work. It's not perfect, but it's
also not a lot of code. https://github.com/square/okhttp/issues/1613

### DIFF
--- a/okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
@@ -17,11 +17,14 @@ package com.squareup.okhttp;
 
 import com.squareup.okhttp.internal.Util;
 import java.io.IOException;
+import java.net.IDN;
 import java.net.InetAddress;
 import java.net.URI;
 import java.net.URL;
+import java.net.UnknownHostException;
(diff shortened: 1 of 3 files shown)


okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
```diff
@@ -731,12 +736,20 @@ private static String host(String input, int pos, int limit) {
       return idnDecoded;
     }
 
-    private static String decodeIpv6(String input, int pos, int limit) {
-      return input.substring(pos, limit); // TODO(jwilson) implement IPv6 decoding.
+    private static InetAddress decodeIpv6(String input, int pos, int limit) {
+      try {
+        return InetAddress.getByName(input.substring(pos, limit));
```
JakeWharton: Move substring to call site? Unless there's some other value I'm not seeing (on mobile).

JakeWharton:
LGTM. Comment is matter of taste and optional.

swankjesse:
I'm gonna merge as-is. If we ever switch to the handrolled parser I made, we can save one allocation.

swankjesse:

PR #1617, opened 2015-05-09
Cleanup some HttpUrl tests.

swankjesse:
We don't need to compare our results against major browsers anymore;
we're passing more than they are.

The list of failing tests is still too large, but should be easy to
churn through.


okhttp-tests/src/test/java/com/squareup/okhttp/HttpUrlTest.java
```diff
@@ -275,7 +308,7 @@
     assertEquals(a2, HttpUrl.parse("http://[1::]").host());
   }
 
-  @Ignore
+  @Ignore // java.net.InetAddress isn't as strict as it should be.
```
JakeWharton: `@Ignore("java.net.InetAddress isn't as strict as it should be")`

okhttp-tests/src/test/java/com/squareup/okhttp/HttpUrlTest.java
```diff
@@ -226,6 +226,39 @@
...
+  @Test public void hostnameMultipleCharacterMapping() throws Exception {
+    // Map the single character telephone symbol (℡) to the string "tel".
+    assertEquals("tel", HttpUrl.parse("http://\u2121").host());
+  }
+
+  @Test public void hostnameMappingLastMappedCodePoint() throws Exception {
+    assertEquals("xn--pu5l", HttpUrl.parse("http://\uD87E\uDE1D").host());
+  }
+
+  @Ignore // The java.net.IDN implementation doesn't ignore characters that it should.
```
JakeWharton: `@Ignore("The java.net.IDN implementation doesn't ignore characters that it should.")`

swankjesse:

PR #1618, opened 2015-05-09
Do IPv6 parsing natively.

swankjesse:
Avoid any risk of doing DNS during HTTP URL parsing.

Closes https://github.com/square/okhttp/issues/1613

### DIFF
--- a/okhttp-tests/src/test/java/com/squareup/okhttp/WebPlatformUrlTest.java
+++ b/okhttp-tests/src/test/java/com/squareup/okhttp/WebPlatformUrlTest.java
@@ -59,46 +59,37 @@ public final class WebPlatformUrlTest {
       "Parsing: <http://f:999999/c> against <http://example.org/foo/bar>",
       "Parsing: <#β> against <http://example.org/foo/bar>",
       "Parsing: <http://www.google.com/foo?bar=baz# »> against <about:blank>",
-      "Parsing: <http://[www.google.com]/> against <about:blank>",
       "Parsing: <http://192.0x00A80001> against <about:blank>",
-      "Parsing: <http://user:pass@/> against <about:blank>",
-      "Parsing: <http:/:@/www.example.com> against <about:blank>",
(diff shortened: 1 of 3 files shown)


okhttp-tests/src/test/java/com/squareup/okhttp/WebPlatformUrlTest.java
```diff
@@ -59,46 +59,37 @@
       "Parsing: <http://f:999999/c> against <http://example.org/foo/bar>",
       "Parsing: <#β> against <http://example.org/foo/bar>",
       "Parsing: <http://www.google.com/foo?bar=baz# »> against <about:blank>",
-      "Parsing: <http://[www.google.com]/> against <about:blank>",
```
swankjesse: This failing test was one of the motivating cases to do IPv6 in OkHttp itself. When we have a hostname in `[square braces]` we need to confirm that it's an IPv6 address otherwise that's invalid. How do we check if it's an IPv6 address? Either we naively check for at least one `:`, or we fully parse it.

swankjesse:

okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
```diff
@@ -736,17 +737,118 @@ private static String host(String input, int pos, int limit) {
...
+        // We've successfully read a group. Assign its value to our byte array.
+        address[b++] = (byte) ((value >>> 8) & 0xff);
+        address[b++] = (byte) (value & 0xff);
+      }
+
+      // All done. If compression happened, we need to move bytes to the right place in the
+      // address. Here's a sample:
+      //
+      //      input: "1111:2222:3333::7777:8888"
+      //     before: { 11, 11, 22, 22, 33, 33, 00, 00, 77, 77, 88, 88, 00, 00, 00, 00  }
```
JakeWharton: uuuuubernit: double space before `}`

PR #1619, opened 2015-05-09
Compose URLs from parts.

swankjesse:


swankjesse:

swankjesse:
There's some asymmetry here...

The builder has two sets of methods: `username()` and `encodedUsername()`.
The object has a different two sets of methods: `username()` and `decodeUsername()`.

Here's where it's surprising:

```
   HttpUrl url = new HttpUrl.Builder().scheme("http").host("host").username("give 100%").build();
   assertEquals("give%20100%25", url.username());
   assertEquals("give 100%", url.decodeUsername());
```
On the builder, the only difference between `username()` and `encodedUsername()` is that the former escapes `%`; the latter assumes its already escaped.

On the object, we don't decode until you explicitly request it.

Most of the time this doesn't matter, but I'd don't like that it's surprising. Any thoughts?

swankjesse:
One option is to do 2 renames:
 * `HttpUrl.username()` → `HttpUrl.encodedUsername()` 
 * `HttpUrl.decodeUsername()` → `HttpUrl.username()`

Then we'd have no methods with a _decode_ prefix. I think I'll do this in follow-up, unless anyone has opinions!

JakeWharton:
My opinions are aligned with yours.

PR #1621, opened 2015-05-10
Use lists to model paths and queries.

swankjesse:
This is actually a little trickier than strings, partly because
we now need code to go in both directions, and partly because
paths that end with '/' are an important but awkward special
case.

The goal is that this will make accessing query parameters much
easier.

### DIFF
--- a/okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
@@ -62,7 +62,7 @@ import okio.Buffer;
  *   f: images
  * }</pre>
  *
- * In addition to composing URLs from their component parts, and decomposing URLs into their
+ * In addition to composing URLs from their component parts and decomposing URLs into their
  * component parts, this class implements relative URL resolution: what address you'd reach by
  * clicking a relative link on a specified page. For example: <pre>   {@code
  *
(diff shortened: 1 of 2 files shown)


swankjesse:

JakeWharton:
Needs a rebase.

okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
```diff
@@ -31,6 +31,8 @@
 /**
  * A <a href="https://url.spec.whatwg.org/">URL</a> with an {@code http} or {@code https} scheme.
  *
+ * TODO: parts of a URL
```
JakeWharton: Probably your conflict right here.

okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
```diff
@@ -245,7 +320,19 @@ public HttpUrl resolve(String link) {
...
 
   public Builder newBuilder() {
-    return new Builder(this);
+    Builder result = new Builder();
+    result.scheme = scheme;
+    result.username = username;
+    result.password = password;
+    result.host = host;
+    result.port = port;
+    result.pathSegments.clear();
```
JakeWharton: Redundant?

swankjesse: Nope, needed 'cause it defaults to `""`.

JakeWharton: Ah, missed the ctor line.

swankjesse:
Rebased!

JakeWharton:
LGTM :shipit:

PR #1623, opened 2015-05-11
Docs on URL components.

swankjesse:


okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
```diff
@@ -29,13 +29,108 @@
...
+ *   }
+ * }</pre>
+ *
+ * which prints: <pre>   {@code
+ *
+ *   q: cute #puppies
+ *   f: images
+ * }</pre>
+ *
+ * In addition to composing URLs from their component parts, and decomposing URLs into their
```
JakeWharton: I think the comma isn't needed

okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
```diff
@@ -29,13 +29,108 @@
...
+ *   https://www.youtube.com/watch?v=cbP2N1BQdYc
+ * }</pre>
+ *
+ * <h3>What's in a URL?</h3>
+ *
+ * A URL has several components.
+ *
+ * <h4>Scheme</h4>
+ * Sometimes referred to as <i>protocol</i>, A URL's scheme describes what mechanism should be used
+ * to retrieve the resource. Although URLs have many schemes (mailto, file, ftp), this class only
```
JakeWharton: These schemes are not wrapped on code while http and https are. Normalize?

JakeWharton:
Nice oxford comma usage!

swankjesse:

PR #1625, opened 2015-05-11
Implement query parameter management.

swankjesse:
The tricky part of this is '+' and ' ' management.

### DIFF
--- a/okhttp-tests/src/test/java/com/squareup/okhttp/HttpUrlTest.java
+++ b/okhttp-tests/src/test/java/com/squareup/okhttp/HttpUrlTest.java
@@ -20,9 +20,12 @@ import com.squareup.okhttp.UrlComponentEncodingTester.Encoding;
 import java.net.URI;
 import java.net.URL;
 import java.util.Arrays;
+import java.util.Collections;
+import java.util.LinkedHashSet;
 import org.junit.Ignore;
 import org.junit.Test;
 
+import static java.util.Collections.singletonList;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.fail;
 
@@ -414,7 +417,7 @@ public final class HttpUrlTest {
     new UrlComponentEncodingTester()
         .override(Encoding.IDENTITY, '?', '`')
         .override(Encoding.PERCENT, '\'')
-        .override(Encoding.SKIP, '#')
+        .override(Encoding.SKIP, '#', ' ')
         .skipForUri('%', '\\', '^', '`', '{', '|', '}')
         .test(Component.QUERY);
   }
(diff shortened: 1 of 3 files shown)


okhttp-tests/src/test/java/com/squareup/okhttp/HttpUrlTest.java
```diff
@@ -612,4 +613,70 @@
...
+    HttpUrl base = HttpUrl.parse("http://host/");
+    HttpUrl url = base.newBuilder().addQueryParameter("a+=& b", "c+=& d").build();
+    assertEquals("http://host/?a%2B%3D%26+b=c%2B%3D%26+d", url.toString());
+    assertEquals("c+=& d", url.queryParameterValue(0));
+    assertEquals("a+=& b", url.queryParameterName(0));
+    assertEquals("c+=& d", url.queryParameter("a+=& b"));
+    assertEquals(Collections.singleton("a+=& b"), url.queryParameterNames());
+    assertEquals(Collections.singletonList("c+=& d"), url.queryParameterValues("a+=& b"));
+    assertEquals(1, url.querySize());
+    assertEquals("a+=& b=c+=& d", url.decodeQuery()); // Ambiguous! (Though working as designed.)
```
JakeWharton: Does this method have a use case? Or is it just for parity?

okhttp-tests/src/test/java/com/squareup/okhttp/HttpUrlTest.java
```diff
@@ -612,4 +613,70 @@
...
+    assertEquals("c =& d", url.queryParameter("a =& b"));
+  }
+
+  @Test public void composeQueryRemoveQueryParameter() throws Exception {
+    HttpUrl url = HttpUrl.parse("http://host/").newBuilder()
+        .addQueryParameter("a+=& b", "c+=& d")
+        .removeAllQueryParameters("a+=& b")
+        .build();
+    assertEquals("http://host/", url.toString());
+    assertEquals(null, url.queryParameter("a+=& b"));
```
JakeWharton: What does query size return for this case?

okhttp-tests/src/test/java/com/squareup/okhttp/HttpUrlTest.java
```diff
@@ -612,4 +613,70 @@
...
+        .removeAllEncodedQueryParameters("a+=& b")
+        .build();
+    assertEquals("http://host/", url.toString());
+    assertEquals(null, url.queryParameter("a =& b"));
+  }
+
+  @Test public void composeQuerySetQueryParameter() throws Exception {
+    HttpUrl url = HttpUrl.parse("http://host/").newBuilder()
+        .addQueryParameter("a+=& b", "c+=& d")
+        .setQueryParameter("a+=& b", "ef")
```
JakeWharton: Duplicate this test and do two calls to add and ensure repeated names are allowed.

okhttp-tests/src/test/java/com/squareup/okhttp/HttpUrlTest.java
JakeWharton: I guess it returns 0 from seeing the code. Should it return -1 when empty to differentiate between `http://host/` and `http://host/?` cases?

okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
```diff
@@ -362,37 +365,64 @@ public String decodeQuery() {
...
-    throw new UnsupportedOperationException(); // TODO(jwilson).
+    if (queryNamesAndValues == null) return null;
+    String encodedName = canonicalize(name, QUERY_COMPONENT_ENCODE_SET, false, true);
+    for (int i = 0, size = queryNamesAndValues.size(); i < size; i += 2) {
+      if (encodedName.equals(queryNamesAndValues.get(i))) {
+        String value = queryNamesAndValues.get(i + 1);
+        return value != null ? percentDecode(value, true) : null;
+      }
+    }
+    return null;
```
JakeWharton: This is ambiguous between "not found" and "found but value was null".

When is the value `null`? `http://host/?foo`? `http://host/?foo=`? I would think the last one would be empty string. Not sure what the behavior of the former would be.

okhttp-tests/src/test/java/com/squareup/okhttp/HttpUrlTest.java
swankjesse: As far as I can tell, there's applications that don't want to cut the query up into name/value pairs, and instead want a single string. This is for those applications.

okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
```diff
@@ -569,48 +599,75 @@ public Builder encodedQuery(String encodedQuery) {
...
+      if (queryNamesAndValues == null) return this;
+      String nameToRemove = canonicalize(name, QUERY_COMPONENT_ENCODE_SET, false, true);
+      removeAllCanonicalQueryParameters(nameToRemove);
+      return this;
     }
 
     public Builder removeAllEncodedQueryParameters(String encodedName) {
-      if (encodedName == null) throw new IllegalArgumentException("encodedName == null");
-      throw new UnsupportedOperationException(); // TODO(jwilson)
+      if (encodedName == null) throw new IllegalArgumentException("name == null");
```
JakeWharton: Wrong param name in exception.

okhttp-tests/src/test/java/com/squareup/okhttp/HttpUrlTest.java
JakeWharton: Wouldn't that be `query()` though?

JakeWharton:
I'd be interested in how the following behaves:
 * `http://host/?foo&bar&baz`
 * `http://host/?foo=&bar=&baz=`
 * `http://host/?foo[]=1&foo[]=2&foo=[3]`

okhttp-tests/src/test/java/com/squareup/okhttp/HttpUrlTest.java
swankjesse: no, `query()` is still encoded. (Until I flip the defaults, at which point this becomes `query()` and that become `encodedQuery()`.)

okhttp-tests/src/test/java/com/squareup/okhttp/HttpUrlTest.java
swankjesse: Fixed as discussed, plus tests.

okhttp-tests/src/test/java/com/squareup/okhttp/HttpUrlTest.java
swankjesse: Done.

okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
swankjesse: If there's an equals sign, the value is non-null.

okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
swankjesse: Good catch.

swankjesse:
Please take a look! I still need to address `+` vs `%20` when encoding spaces.

JakeWharton:
Also a rebase!

swankjesse:
Rebased.

JakeWharton:
LGTM. These tests are great.

swankjesse:

PR #1626, opened 2015-05-11
Enhance Javadocs on OkHttpClient.setSocketFactory

unguiculus:
I'd suggest enhancing the Javadocs to make it explicit how a Socket Factory is used. Thanks.


swankjesse:
Thanks!

PR #1629, opened 2015-05-12
Rename username to encodedUsername and decodeUsername to username

swankjesse:
Similarly for all other URL components.

Basically the decoded form now gets the shorter name.

### DIFF
--- a/okhttp-tests/src/test/java/com/squareup/okhttp/UrlComponentEncodingTester.java
+++ b/okhttp-tests/src/test/java/com/squareup/okhttp/UrlComponentEncodingTester.java
@@ -213,7 +213,7 @@ class UrlComponentEncodingTester {
     String encoded = encoding.encode(codePoint);
     String urlString = component.urlString(encoded);
     HttpUrl url = HttpUrl.parse(urlString);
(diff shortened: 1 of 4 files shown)


okhttp-tests/src/test/java/com/squareup/okhttp/UrlComponentEncodingTester.java
```diff
@@ -298,7 +298,7 @@ public String encode(int codePoint) {
         return "http://" + value + "@example.com/";
       }
       @Override public String decodedValue(HttpUrl url) {
```
JakeWharton: "decoded value" calling "encoded username" raises naming flags. Rename these methods too to avoid confusion?

swankjesse: Fixed.

swankjesse: (this name was completely broken, by the way. New names are better.)

JakeWharton:
LGTM DO WANT :shipit:

swankjesse:

PR #1630, opened 2015-05-12
Always encode space as %20.

swankjesse:
In queries this means transcoding '+' to '%2B' also.


okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
```diff
@@ -1286,7 +1286,7 @@ static String percentDecode(String encoded, int pos, int limit, boolean query) {
     return encoded.substring(pos, limit);
   }
 
-  static void percentDecode(Buffer out, String encoded, int pos, int limit, boolean query) {
+  static void percentDecode(Buffer out, String encoded, int pos, int limit) {
```
JakeWharton: The death of a boolean is always a good thing!

swankjesse:

PR #1631, opened 2015-05-12
Use HttpUrl in Request.

swankjesse:
This makes URL validation more eager than it is today.

### DIFF
--- a/okhttp-tests/src/test/java/com/squareup/okhttp/CallTest.java
+++ b/okhttp-tests/src/test/java/com/squareup/okhttp/CallTest.java
@@ -62,7 +62,6 @@ import okio.GzipSink;
 import okio.Okio;
 import org.junit.After;
 import org.junit.Before;
-import org.junit.Ignore;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.TemporaryFolder;
@@ -73,7 +72,6 @@ import static com.squareup.okhttp.internal.Internal.logger;
 import static java.net.CookiePolicy.ACCEPT_ORIGINAL_SERVER;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNotSame;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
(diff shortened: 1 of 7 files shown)


okhttp/src/main/java/com/squareup/okhttp/Request.java
```diff
@@ -108,22 +104,21 @@ public CacheControl cacheControl() {
...
+        + url
         + ", tag="
         + (tag != this ? tag : null)
         + '}';
   }
 
   public static class Builder {
-    private String urlString;
-    private URL url;
+    private HttpUrl httpUrl;
```
JakeWharton: nit: would probably keep this named `url` for parity with the rest of the file

okhttp/src/main/java/com/squareup/okhttp/Request.java
```diff
@@ -29,45 +26,44 @@
...
-    this.urlString = builder.urlString;
+    this.url = builder.httpUrl;
     this.method = builder.method;
     this.headers = builder.headers.build();
     this.body = builder.body;
     this.tag = builder.tag != null ? builder.tag : this;
-    this.url = builder.url;
+  }
+
+  public HttpUrl httpUrl() {
```
JakeWharton: Unfortunate forced name. 3.0 kills `url()` and `uri()` and renames this `url()` maybe?

okhttp/src/main/java/com/squareup/okhttp/Request.java
```diff
@@ -29,45 +26,44 @@
...
-      throw new RuntimeException("Malformed URL: " + urlString, e);
-    }
+    URL result = javaNetUrl;
+    return result != null ? result : (javaNetUrl = url.url());
   }
 
   public URI uri() throws IOException {
     try {
-      URI result = uri;
-      return result != null ? result : (uri = Platform.get().toUriLenient(url()));
```
JakeWharton: Can this `Platform` API be deleted as well? Or is it still used internally?

swankjesse: Killed. FYI @nfuller

okhttp/src/main/java/com/squareup/okhttp/Request.java
swankjesse: Yes, definitely.

okhttp/src/main/java/com/squareup/okhttp/Request.java
swankjesse: Fixed.

okhttp-tests/src/test/java/com/squareup/okhttp/CallTest.java
```diff
@@ -127,40 +126,46 @@
...
-    executeSynchronously(request2)
-        .assertCode(200)
-        .assertSuccessful()
-        .assertBody("abc");
+    assertEquals(httpUrl, request.httpUrl());
+
+    executeSynchronously(request).assertSuccessful();
+  }
+
+  @Test public void eagerlyValidateRequestUrl() throws Exception {
```
JakeWharton: This seems like a dupe of invalid scheme test. The rest is valid I think...

JakeWharton:
LGTM

okhttp-tests/src/test/java/com/squareup/okhttp/CallTest.java
swankjesse: Yup, you're right. Previously it was different cases because `java.net.URL` knows `ftp`  but not `foo`.

okhttp/src/main/java/com/squareup/okhttp/internal/http/AuthenticatorAdapter.java
```diff
@@ -20,6 +20,8 @@
 import com.squareup.okhttp.Credentials;
 import com.squareup.okhttp.Request;
 import com.squareup.okhttp.Response;
+import com.squareup.okhttp.internal.Internal;
```
JakeWharton: Unused. Failed the build...

JakeWharton:
```
  MainTest.simple:32 expected:<http://example.com[]> but was:<http://example.com[/]>
  MainTest.contentTypeHeader:66 expected:<http://example.com[]> but was:<http://example.com[/]>
  MainTest.put:39 expected:<http://example.com[]> but was:<http://example.com[/]>
  MainTest.userAgent:82 expected:<http://example.com[]> but was:<http://example.com[/]>
  MainTest.referer:74 expected:<http://example.com[]> but was:<http://example.com[/]>
  MainTest.dataPut:56 expected:<http://example.com[]> but was:<http://example.com[/]>
  MainTest.dataPost:47 expected:<http://example.com[]> but was:<http://example.com[/]>
```

swankjesse:

PR #1639, opened 2015-05-13
Adds upgrade failed callback to web sockets

pwerry:
This, or some variant of it, is necessary for a consumer of the API can deal with non-101 http status codes from the server via [RFC2616](https://tools.ietf.org/html/rfc2616) (also specified in the WebSocket specs [RFC6455 Section 4.1](https://tools.ietf.org/html/rfc6455#section-4.1))

An alternative implementation might be to augment `onFailure()` to supply the `Response` and `Reason`, but this approach makes the upgrade path explicit. 

commit 8ce30f297f81c19d44ff0de2f7d31c3c07cdb0c8
Author: Peter Werry <pwerry@gmail.com>
Date:   Tue May 12 12:01:49 2015 +0100

    Update websocket tests for new onUpgradeFailed() callback

commit bd4db0ced6ad1944bbc111b23a2a7ff9dbb2085d
Author: Peter Werry <pwerry@gmail.com>
Date:   Sat May 9 14:57:33 2015 +0100

    Adds onUpgradeFailed() WebSocketListener callback

### DIFF
--- a/okhttp-ws-tests/src/test/java/com/squareup/okhttp/ws/AutobahnTester.java
+++ b/okhttp-ws-tests/src/test/java/com/squareup/okhttp/ws/AutobahnTester.java
@@ -29,6 +29,8 @@ import java.util.concurrent.atomic.AtomicReference;
 import okio.Buffer;
 import okio.BufferedSource;
 
+import static com.squareup.okhttp.ws.WebSocket.UpgradeFailureReason;
+
 /**
  * Exercises the web socket implementation against the
  * <a href="http://autobahn.ws/testsuite/">Autobahn Testsuite</a>.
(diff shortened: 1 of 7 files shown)


swankjesse:
Just curious: when would this happen in practice? Ie. do real applications need to differentiate?

JakeWharton:
We talked about this at some point. A potential solution would also be to deliver the nullable `Response` to `onFailure`. I'm really hesitant about yet another callback... we have a lot as is.

swankjesse:
+1 to optional response to onFailure().

xeizmendi:
I'll give an example of why you would need to receive the HTTP response from the server:
You have an application that uses long living tokens to connect to the websocket and identify a particular session. When the application connects the server can return different errors like token expired, token invalid, application needs upgrade, etc. Then the application can decide what to do based on the HTTP response.
It's one of the advantages of the websockets, you can use the upgrade step (by setting headers) as an authorisation/authentication step and only upgrade after this is accomplished.

pwerry:
Jake thanks for getting back. I agree about the added callback being a bit clunky. I'd be happy to bin this PR and issue a new one with a nullable response to onFailure(). Is this the flow you guys would prefer? Or would you rather I amend to this one to maintain the discussion?

JakeWharton:
Going to look at this tonight because I'd really like this to make it into 2.4 which is coming _very_ soon.

PR #1641, opened 2015-05-13
Make matching of TlsVersion optionally stricter

nfuller:
On Android there are 4 connection specs configured:
1) TLSv1.2, TLSv1.1, TLSv1.0, SSLv3
2) TLSv1.1, TLSv1.0, SSLv3
3) TLSv1.0, SSLv3
4) SSLv3

(Typically spec 4 is not tried because SSLv3 is not
enabled by default).

If only TLSv1.0 is enabled then OkHttp will
attempt 3 connections (1, 2 and 3 above), with exactly
the same enabled protocols enabled (just TLSv1.0).

This change adds an addition restriction: the first
protocol listed *must* be supported for the connection
negotiation to be attempted so only spec 3 would be
attempted after this change.

Note, this does not change default behavior.

OkHttp by default tries twice (MODERN_TLS + COMPATIBLE_TLS).
If only TLSv1.0 is enabled: it will still try to connect
twice with TLSv1.0.

### DIFF
--- a/okhttp/src/main/java/com/squareup/okhttp/ConnectionSpec.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/ConnectionSpec.java
@@ -78,6 +78,12 @@ public final class ConnectionSpec {
   /** Used if tls == true. The TLS protocol versions to use. */
   private final String[] tlsVersions;
 
+  /**
+   * Used if tls == true. Whether the first tlsVersion must be enabled for
+   * {@link #isCompatible(javax.net.ssl.SSLSocket)} to return true.
+   */
+  private final boolean requireFirstTlsVersion;
+
   final boolean supportsTlsExtensions;
 
   private ConnectionSpec(Builder builder) {
@@ -85,6 +91,7 @@ public final class ConnectionSpec {
     this.cipherSuites = builder.cipherSuites;
     this.tlsVersions = builder.tlsVersions;
     this.supportsTlsExtensions = builder.supportsTlsExtensions;
+    this.requireFirstTlsVersion = builder.requireFirstTlsVersion;
   }
 
   public boolean isTls() {
@@ -188,6 +195,9 @@ public final class ConnectionSpec {
     }
 
     String[] enabledProtocols = socket.getEnabledProtocols();
+    if (requireFirstTlsVersion && !contains(enabledProtocols, tlsVersions[0])) {
+      return false;
+    }
     boolean requiredProtocolsEnabled = nonEmptyIntersection(tlsVersions, enabledProtocols);
     if (!requiredProtocolsEnabled) {
       return false;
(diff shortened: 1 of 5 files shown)


okhttp/src/main/java/com/squareup/okhttp/ConnectionSpec.java
```diff
@@ -78,6 +78,12 @@
   /** Used if tls == true. The TLS protocol versions to use. */
   private final String[] tlsVersions;
 
+  /**
+   * Used if tls == true. Whether the first tlsVersion must be enabled for
+   * {@link #isCompatible(javax.net.ssl.SSLSocket)} to return true.
+   */
+  private final boolean requireFirstTlsVersion;
```
swankjesse: Shouldn't this just be the default?

swankjesse: Hmmm . . . I think I see that it can't be, because then we might skip MODERN_TLS entirely just because TLS_1_2 is absent.

swankjesse:
This works, but it seems like more flexibility than should be strictly necessary. 

Could we simplify by just remembering what parameters we most recently attempted, and never trying the same configuration twice in a row?

swankjesse:
(that's more work internally, but a smaller API. I prefer smaller APIs!)

nfuller:
Took a quick run at doing as you suggested. Storing the sets of protocols already attempted in ConnectionSpecSelector is fine. However....

I had to add a parameter to ConnectionSpec.isCompatible() to take a List<Set<String>> parameter containing the list of (set of) protocols that that have already been tried. This could be achieved with an overload if needed to keep API compatibility.

We currently try to predict whether we'll be able to do fallback to avoid connecting a socket and discovering that there's nothing to do. This leads to a bit of a chicken-and-egg problem with the current APIs:

In ConnectionSpecSelector.configureSecureSocket(sslSocket) we would need:

connectionSpec = pick next connectionSpec, where connectionSpec.isCompatible(sslSocket, protocolsTried) is true.
boolean isFallbackPossible = isFallbackPossible(sslSocket, <protocolsTried + (set X)>);
connectionSpec.configureSocket(sslSocket, isFallbackRequest);
set X = sslSocket.getEnabledProtocols();
protocolsTried.add(set X);

"set X" is the set of protocols that will be / is being used for the current attempt.
To get set X we need to be able to predict the set of protocols we're going to try next. This would suggest an additional "intersect the ConnectionSpec protocols and the socket's enabled protocols" step, which duplicates logic in ConnectionSpec.

We could move isCompatible() and apply() to ConnectionSpecSelector and make ConnectionSpec completely "dumb" again. That means modifying the API further so I stopped.

swankjesse:
Sorry for sitting on this for forever. Can we fix this without introducing additional configuration?

swankjesse:
Closing for now because I don't want the new API. I do think fixing this is a good idea ’cause it makes connecting both faster & safer.

PR #1645, opened 2015-05-14
Add support for validating all pinned certificates

deronbrown:
Give the developer the option of validating that all pinned certificates are present in the returned peer certificate chain. This is influenced by AFNetworking's [AFSecurityPolicy](https://github.com/AFNetworking/AFNetworking/blob/master/AFNetworking/AFSecurityPolicy.m#L280).


swankjesse:
What happens when the webserver wants to change its certificates? Do all clients break?

swankjesse:
I think this is a dangerous policy and wouldn't want to encourage anyone to deploy it. What am I missing?

swankjesse:
Ahhh . . . I think what AFNetworking is offering is something different: whether the pins check the leaf certificates, or the leaf nodes plus the signing certificates. This is not a very useful constraint. What would be a situation where it would be more secure? Or more convenient? I can't think of any.

deronbrown:
Yeah, I may have been misinterpreting the two implementations. I'll have to go back later and re-read them.

PR #1649, opened 2015-05-15
APIs to set and remove path segments.

swankjesse:


okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
```diff
@@ -692,6 +696,38 @@ public Builder addEncodedPathSegment(String encodedPathSegment) {
       return this;
     }
 
+    public Builder setPathSegment(int index, String pathSegment) {
+      if (pathSegment == null) throw new IllegalArgumentException("pathSegment == null");
```
JakeWharton: Missing index bounds check. Or if you are falling back to List to do the work then tests are missing.

okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
```diff
@@ -993,33 +1029,36 @@ private void resolvePath(String input, int pos, int limit) {
...
         encodedPathSegments.add(segment);
       }
-
       if (addTrailingSlash) {
         encodedPathSegments.add("");
       }
     }
 
+    private boolean isDot(String input) {
+      return input.equalsIgnoreCase(".") || input.equalsIgnoreCase("%2e");
```
JakeWharton: Ignore case on dot?

JakeWharton: Seems intentional and I'm curious.

JakeWharton:
LGTM!

okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
swankjesse: I'll add a test!

okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
swankjesse: Seemed simpler to keep all cases symmetric. Probably unnecessary. I eyeballed it both ways.

I'll change it.

Issue #1215, opened 2014-12-21
Pre-encoded name and/or value strings for form URL bodies.


JakeWharton:
Retrofit allows specifying pre-encoded values for both name and value. Putting this capability in OkHttp would allow a Retrofit v2 to completely remove its knowledge of form URL encoding.

The API for this is pretty awful, not sure if we can come up with something better.
```java
void add(String name, boolean encodeName, String value, boolean encodeValue);
```

swankjesse:
Punting to 2.4, wherein I'm going to add our own URL class.

PR #1650, opened 2015-05-15
New FormEncodingBuilder.addEncoded() API.

swankjesse:
This is a behavior change: previously we'd encode '+' as plus
and encode characters that don't need it, including ','. Now
we use the same rules as HttpUrl.

### DIFF
--- a/okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
@@ -255,12 +255,12 @@ import okio.Buffer;
 public final class HttpUrl {
   private static final char[] HEX_DIGITS =
       { '0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'A', 'B', 'C', 'D', 'E', 'F' };
-  private static final String USERNAME_ENCODE_SET = " \"':;<=>@[]^`{}|/\\?#";
-  private static final String PASSWORD_ENCODE_SET = " \"':;<=>@[]\\^`{}|/\\?#";
-  private static final String PATH_SEGMENT_ENCODE_SET = " \"<>^`{}|/\\?#";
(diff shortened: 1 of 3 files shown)


okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
```diff
@@ -1417,13 +1417,13 @@ static void canonicalize(StringBuilder out, String input, int pos, int limit,
...
-          out.append(HEX_DIGITS[(b >> 4) & 0xf]);
-          out.append(HEX_DIGITS[b & 0xf]);
+          out.writeByte('%');
+          out.writeByte(HEX_DIGITS[(b >> 4) & 0xf]);
+          out.writeByte(HEX_DIGITS[b & 0xf]);
         }
       } else {
         // This character doesn't need encoding. Just copy it over.
-        out.append((char) codePoint);
+        out.writeUtf8CodePoint((char) codePoint);
```
JakeWharton: I think you can lose the cast here. Otherwise you're going int -> char -> int.

swankjesse: Yeah, actually the code there now will be incorrect when I fix fragments to not encode non-ascii chars.

JakeWharton: Actually I think it's optimized out...

JakeWharton: Nope!
```
int codePoint = 1;
int foo = (char) codePoint;
```
```
0: iconst_1
1: istore_1
2: iload_1
3: i2c
4: istore_2
```

JakeWharton:
LGTM

JakeWharton:
Closes #1215

PR #1658, opened 2015-05-17
Fix a race closing MockSpdyPeer.

swankjesse:
It would have crashed if close() was called before readAndWriteFrames() started.
That's fixed now.


JakeWharton:
LGTM

PR #1663, opened 2015-05-19
Accept ws: URLs in Request.Builder.

swankjesse:
I contemplated supporting them in HttpUrl, but then realized I'd
either need to document that behavior (which is weird behavior)
or return ws from the scheme (which seems arbitrary.)

Doing it in Request seems less bad: we support websockets requests.

### DIFF
--- a/okhttp/src/main/java/com/squareup/okhttp/Request.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/Request.java
@@ -145,6 +145,14 @@ public final class Request {
 
     public Builder url(String url) {
       if (url == null) throw new IllegalArgumentException("url == null");
+
+      // Silently replace websocket URLs with HTTP URLs.
+      if (url.regionMatches(true, 0, "ws:", 0, 3)) {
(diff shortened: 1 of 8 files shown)


okhttp/src/main/java/com/squareup/okhttp/Request.java
```diff
@@ -145,6 +145,14 @@ public Builder url(HttpUrl url) {
 
     public Builder url(String url) {
       if (url == null) throw new IllegalArgumentException("url == null");
+
+      // Silently replace websocket URLs with HTTP URLs.
```
JakeWharton: There's a corresponding chunk in `WebSocketCall` that can be deleted.

JakeWharton:
LGTM. Not sure what else to do anyways...

okhttp/src/main/java/com/squareup/okhttp/Request.java
swankjesse: Fixed!

PR #1665, opened 2015-05-20
Forbid response bodies on HTTP 204 and 205 responses.

### LINKED ISSUE #1664: Throw ProtocolException on non-empty 204s and 205s
Currently Retrofit is guarding against this manually. The HTTP client should do it.
https://github.com/square/retrofit/pull/845/files#diff-826f81c41efbd1fcc363bccd4dbc4f17R143

JakeWharton:
Closes #1664.

### DIFF
--- a/okhttp/src/main/java/com/squareup/okhttp/internal/http/HttpEngine.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/internal/http/HttpEngine.java
@@ -899,7 +899,15 @@ public final class HttpEngine {
         bufferedRequestBody.close();
       }
 
-      return readNetworkResponse();
+      Response response = readNetworkResponse();
+
+      int code = response.code();
+      if ((code == 204 || code == 205) && response.body().contentLength() > 0) {
+        throw new ProtocolException(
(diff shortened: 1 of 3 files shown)


okhttp/src/main/java/com/squareup/okhttp/internal/http/HttpEngine.java
```diff
@@ -899,7 +899,15 @@ public void readResponse() throws IOException {
         bufferedRequestBody.close();
       }
 
-      return readNetworkResponse();
+      Response response = readNetworkResponse();
+
+      int code = response.code();
+      if ((code == 204 || code == 205) && response.body().contentLength() > 0) {
```
JakeWharton: I wanted to put this closer to the parsing of this information, but since `HttpUrlConnection` specifically allows this behavior I opted for here. Open to suggestions if there's a more appropriate place.

swankjesse: Yeah, this doesn't feel like the right place for it. This has nothing to do with interceptors! Perhaps better in readNetworkResponse(), guarded by `callerWritesRequestBody` ? (That's code for `HttpUrlConnection`.)

PR #1668, opened 2015-05-21
Add optional Response to WS failure callback.

JakeWharton:
This also removes the explicit Request from WS open callback. It can be accessed by calling 'request()' on the supplied Response.

Closes #1639.


pwerry:
Thanks for this Jake! Was just about to push essentially the exact same change. We're already running it from a 2.3 fork with good results.

PR #1683, opened 2015-06-03
make OkApacheClient.execute() method react to httpHost paramater

Ponyets:
Some code, especially some library code such as thrift, put host info in HttpHost and only path info in HttpRequest, that makes OkApacheClient got an MalformedURLException when calling Request.url() method.

### DIFF
--- a/okhttp-apache/src/main/java/com/squareup/okhttp/apache/OkApacheClient.java
+++ b/okhttp-apache/src/main/java/com/squareup/okhttp/apache/OkApacheClient.java
@@ -10,6 +10,8 @@ import com.squareup.okhttp.ResponseBody;
 import java.io.IOException;
 import java.net.InetSocketAddress;
 import java.net.Proxy;
+import java.net.URI;
+import java.net.URISyntaxException;
 import org.apache.http.Header;
 import org.apache.http.HttpEntity;
 import org.apache.http.HttpEntityEnclosingRequest;
@@ -39,12 +41,24 @@ import static org.apache.http.HttpVersion.HTTP_1_1;
  * and others.
  */
 public final class OkApacheClient implements HttpClient {
-  private static Request transformRequest(HttpRequest request) {
+  private static Request transformRequest(HttpHost httpHost, HttpRequest request) {
     Request.Builder builder = new Request.Builder();
 
     RequestLine requestLine = request.getRequestLine();
     String method = requestLine.getMethod();
-    builder.url(requestLine.getUri());
+    String url = requestLine.getUri();
+    if (httpHost != null) {
+      try {
+        URI uri = URI.create(url);
+        URI uriWithHost = new URI(httpHost.getSchemeName(), uri.getUserInfo(),
+            httpHost.getHostName(), httpHost.getPort(), uri.getPath(),
+            uri.getQuery(), uri.getFragment());
+        url = uriWithHost.toString();
+      } catch (URISyntaxException e) {
+        throw new IllegalArgumentException(e.getMessage(), e);
+      }
+    }
+    builder.url(url);
 
     String contentType = null;
     for (Header header : request.getAllHeaders()) {
@@ -168,7 +182,7 @@ public final class OkApacheClient implements HttpClient {
 
   @Override public HttpResponse execute(HttpHost host, HttpRequest request, HttpContext context)
       throws IOException {
-    Request okRequest = transformRequest(request);
+    Request okRequest = transformRequest(host, request);
     Response okResponse = client.newCall(okRequest).execute();
     return transformResponse(okResponse);
   }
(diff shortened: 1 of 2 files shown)


okhttp-apache/src/main/java/com/squareup/okhttp/apache/OkApacheClient.java
```diff
@@ -39,12 +39,22 @@
...
+  private static Request transformRequest(HttpHost httpHost, HttpRequest request) {
     Request.Builder builder = new Request.Builder();
 
     RequestLine requestLine = request.getRequestLine();
     String method = requestLine.getMethod();
-    builder.url(requestLine.getUri());
+    String url = requestLine.getUri();
+    if (httpHost != null) {
+      try {
+        URI uri = URI.create(url);
```
swankjesse: Wanna use our new `HttpUrl` hotness for construction here? That'll support more URLs.

okhttp-apache/src/main/java/com/squareup/okhttp/apache/OkApacheClient.java
```diff
@@ -39,12 +39,22 @@
...
+        URI uriWithHost = new URI(httpHost.getSchemeName(), uri.getUserInfo(), httpHost.getHostName(),
+            httpHost.getPort(), uri.getPath(), uri.getQuery(), uri.getFragment());
+        url = uriWithHost.toString();
+      } catch (Exception ignored) {
```
swankjesse: Eeep, what danger is this?

okhttp-apache/src/main/java/com/squareup/okhttp/apache/OkApacheClient.java
```diff
@@ -8,8 +8,8 @@
 import com.squareup.okhttp.Response;
 import com.squareup.okhttp.ResponseBody;
 import java.io.IOException;
-import java.net.InetSocketAddress;
-import java.net.Proxy;
+import java.net.*;
```
swankjesse: nit: we don't wildcard imports in our projects.

swankjesse:
(CLA is good. From 2013!)

okhttp-apache/src/main/java/com/squareup/okhttp/apache/OkApacheClient.java
Ponyets: Tried, but `HttpUrl` do not support URI without scheme or authority

Ponyets:
I don't know why CI build always failed. Building is success on my local environment. 
I can't open the CI web page.

okhttp-apache/src/main/java/com/squareup/okhttp/apache/OkApacheClient.java
```diff
@@ -39,12 +41,22 @@
...
     String method = requestLine.getMethod();
-    builder.url(requestLine.getUri());
+    String url = requestLine.getUri();
+    if (httpHost != null) {
+      try {
+        URI uri = URI.create(url);
+        URI uriWithHost = new URI(httpHost.getSchemeName(), uri.getUserInfo(), httpHost.getHostName(),
+            httpHost.getPort(), uri.getPath(), uri.getQuery(), uri.getFragment());
+        url = uriWithHost.toString();
+      } catch (URISyntaxException ignored) {
```
swankjesse: I think we want to re-throw an exception here. Maybe an IllegalArgumentException? Otherwise the HttpHost will be ignored if it's malformed, hiding bugs.

okhttp-apache/src/test/java/com/squareup/okhttp/apache/OkApacheClientTest.java
```diff
@@ -260,6 +261,16 @@
     assertEquals(text, EntityUtils.toString(entity, UTF_8));
   }
 
+  @Test public void httpHost() throws Exception {
```
swankjesse: Wouldn't this test pass if you didn't have HTTP host support?

okhttp-apache/src/test/java/com/squareup/okhttp/apache/OkApacheClientTest.java
```diff
@@ -260,6 +261,16 @@
...
+    server.enqueue(new MockResponse());
+    HttpHost host = new HttpHost(server.getHostName(), server.getPort());
+    HttpGet get = new HttpGet(server.getUrl("/test").getFile());
```
swankjesse: I think for this test to be right you need to pass `"/test"` without the `server.getUrl()` call.

swankjesse: ie. this test would have passed in the code without your fix.

swankjesse:
Please reopen if this is still an issue.

Issue #831, opened 2014-05-17
ETag and If-None-Match


arturdryomov:
I have troubles with the ETag-based caching. I am using OkHttp 1.5.4 and Retrofit 1.5.1 with `HttpResponseCache` set. There are logs below for better explanation.

The request with the `Session` header set to a token **A**. OkHttp hits the network, which is great.

```
---> HTTP GET https://localhost/shares
Session: TOKEN_A
Accept: application/json
User-Agent: Application/1.0.0 (Android 4.4.2; Nexus 4) Size/4.7 Resolution/1280x768
---> END HTTP (0-byte body)
<--- HTTP 200 https://localhost/shares (402ms)
: HTTP/1.1 200 OK
Content-Length: 136
Content-Type: application/json
Date: Sat, 17 May 2014 11:07:20 GMT
Etag: "677210f4c22a13203551e2b92a20684528a23c87"
Last-Modified: Sat, 17 May 2014 04:07:20 GMT
OkHttp-Received-Millis: 1400324839589
OkHttp-Response-Source: NETWORK 200
OkHttp-Sent-Millis: 1400324839194
<--- END HTTP (136-byte body)
```

The second request with the `Session` header set to a token **A**. OkHttp hits the cache, which is great as well.

```
---> HTTP GET https://localhost/shares
Session: TOKEN_A
Accept: application/json
User-Agent: Application/1.0.0 (Android 4.4.2; Nexus 4) Size/4.7 Resolution/1280x768
---> END HTTP (0-byte body)
GC_FOR_ALLOC freed 415K, 5% free 9588K/10036K, paused 19ms, total 19ms
<--- HTTP 200 https://localhost/shares (46ms)
: HTTP/1.1 200 OK
Content-Length: 136
Content-Type: application/json
Date: Sat, 17 May 2014 11:07:20 GMT
Etag: "677210f4c22a13203551e2b92a20684528a23c87"
Last-Modified: Sat, 17 May 2014 04:07:20 GMT
OkHttp-Received-Millis: 1400324839589
OkHttp-Response-Source: CACHE 200
OkHttp-Sent-Millis: 1400324839194
<--- END HTTP (136-byte body)
```

The third request with the `Session` header set to token **B**. OkHttp hits the cache again, which is not so great, because in this case the server returns a different response with a different ETag actually.

```
---> HTTP GET https://localhost/shares
Session: TOKEN_B
Accept: application/json
[...]

PR #1689, opened 2015-06-08
Use cache finally

yegorius:
I've been having the same problem as discussed in #831
I noticed that `cache` and `internalCache` are both null for `OkHttpClient` instance in `HttpEngine::sendRequest()`.
Also call `.setUseCaches(true)` to jump over the check in `HttpURLConnectionImpl::newHttpEngine()` line 358.

### DIFF
--- a/okhttp/src/main/java/com/squareup/okhttp/OkHttpClient.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/OkHttpClient.java
@@ -342,7 +342,7 @@ public class OkHttpClient implements Cloneable {
 
   public OkHttpClient setCache(Cache cache) {
     this.cache = cache;
-    this.internalCache = null;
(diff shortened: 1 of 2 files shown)


okhttp/src/main/java/com/squareup/okhttp/OkHttpClient.java
```diff
@@ -342,7 +342,7 @@ InternalCache internalCache() {
 
   public OkHttpClient setCache(Cache cache) {
     this.cache = cache;
-    this.internalCache = null;
+    this.internalCache = cache.internalCache;
```
yegorius: This one is really nasty :smile:

swankjesse:
I don't follow this at all. Could you provide a failing test that this fixes?

swankjesse:
Ping on a failing test case.

PR #1702, opened 2015-06-15
Expose WebSocket.closeConnection() to public API

pwerry:
The reason for doing this is to expose an API to the client that allows it to force the connection to close. The current `WebSocket.close()` method (correctly) attempts to send a `WebSocket` close frame before shutting down the backing `Connection`. But if the `Connection` is in an SSL timeout state (for example, when Android is connected to a WiFi router dropping a ton of WAN packets), then this function will hang and there will be no way to force the `WebSocket` to terminate. Giving the client the power to force a `Connection` shutdown allows it to set up connect and shutdown timers coupled with reachability and make the right decisions.

### DIFF
--- a/okhttp-ws/src/main/java/com/squareup/okhttp/ws/WebSocket.java
+++ b/okhttp-ws/src/main/java/com/squareup/okhttp/ws/WebSocket.java
@@ -15,6 +15,7 @@
  */
 package com.squareup.okhttp.ws;
 
+import com.squareup.okhttp.Connection;
 import java.io.IOException;
 import okio.Buffer;
 import okio.BufferedSink;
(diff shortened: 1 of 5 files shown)


okhttp-ws/src/main/java/com/squareup/okhttp/ws/WebSocket.java
```diff
@@ -65,4 +66,16 @@
...
+   * Closes the underlying {@link Connection connection} backing the WebSocket.
+   * <p>
+   * The correct way to close a WebSocket is to call {@link WebSocket#close close()}. Calling this
+   * method will force the WebSocket to close its backing {@link Connection connection}, skipping
+   * the WebSocket close frame.
+   * See <a href="http://tools.ietf.org/html/rfc6455#section-5.5.1">Close Frame</a> for details
+   *
+   * @throws IOException
+   */
+  void closeConnection() throws IOException;
```
swankjesse: `disconnect()` ?

JakeWharton:
A potential alternative: we could expose a `Timeout` instance so that you could set timeouts or deadlines for things like this which would automatically close the underlying connection.

okhttp-ws/src/main/java/com/squareup/okhttp/ws/WebSocket.java
pwerry: `disconnect()` is probably a better naming convention. In this instance I simply exposed a function from the underlying implementation. Happy to change

pwerry:
Jake exposing a `Timeout` would certainly solve the majority of cases, and would improve safety, so that's something I'd support. How would you expect that API to look? We could expose a configuration through the `WebSocketCall#create` API, but that would mean the `Timeout` wouldn't be call-time configurable. Clients will have a better understanding of the behaviour they require at the moment of shutdown than at configuration time. A timeout used locally during the `close()` process might be more appropriate.

PR #1708, opened 2015-06-17
Rename 'spdy' references to 'framed' where appropriate.

JakeWharton:


okhttp/src/main/java/com/squareup/okhttp/internal/framed/Http2.java
```diff
@@ -467,7 +467,7 @@ private void writeContinuationFrames(int streamId, long byteCount) throws IOExce
     @Override public synchronized void rstStream(int streamId, ErrorCode errorCode)
         throws IOException {
       if (closed) throw new IOException("closed");
-      if (errorCode.spdyRstCode == -1) throw new IllegalArgumentException();
+      if (errorCode.httpCode == -1) throw new IllegalArgumentException();
```
JakeWharton: Bug! (Would have been a programmer error if triggered though)

swankjesse:
FYI @ejona86 this will break you when we release OkHttp 2.5!

ejona86:
@swankjesse, thanks for the heads up. I look forward to the cleaner names :)

Issue #1694, opened 2015-06-10
HttpUrl#canonicalize() is not escape `;`(semicolon) character 


keima:
I'm using OkHttp for sending html tags (contains style attributes).
In 2.2.0, okhttp is running perfectly.
But I upgrade it to 2.4.0, server received corrupted html.
This server treats '&' and ';' as separator.
(hoge=fuga&foo=bar -> {hoge: fuga, foo: bar} , hoge=fuga;foo=bar -> {hoge: fuga, foo: bar} )
http://www.w3.org/TR/1999/REC-html401-19991224/appendix/notes.html#h-B.2.2

Result of investigation, I found that incompatible changes from 2.2.0 to 2.4.0.

As 2.2.0, value of `FormEncodingBuilder.add(key,value)` is escaped `URLEncoder.encode(value, "UTF-8")`.
(hoge=fu;ga&foo=bar -> {hoge: fu%3Bga, foo: bar})

But 2.4.0, value is escaped `HttpUrl#canonicalize()`.
This method (or, `HttpUrl.QUERY_COMPONENT_ENCODE_SET`) is not treated ';'. So value data is still alive semicolon!
(hoge=fu;ga&foo=bar -> {hoge: fu, ga:null , foo: bar})
(sorry, ga:null is no confidences... In any way, hoge's value is corrupted.)

I think, Add semicolon to `HttpUrl.QUERY_COMPONENT_ENCODE_SET`.
https://github.com/square/okhttp/blob/master/okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java#L262
But I don't try yet and there are no confidences for me :(

swankjesse:
OkHttp's behavior is consistent with web browsers.

keima:
If OkHttp's behavior is (largely) similar to modern web browsers, I hope it should follow [RFC 3986](http://www.ietf.org/rfc/rfc3986.txt) and [W3C HTML5 recommendation](http://www.w3.org/TR/html5/forms.html#url-encoded-form-data).

There are various views about what is escaped ( `(` is escaped or not ), 
In all documents which I read, semicolon is handled as escape.
Therefore, I think OkHttp's `FormEncodingBuilder ` should be escape `;` just like `&`.

If you say OkHttp don't treat semicolon as escaped, I want to know why it is.

swankjesse:
Try this in your browser's JavaScript console:
```
  new URL("http://host/path?fun;ga").href
```

swankjesse:
[...]

PR #1710, opened 2015-06-18
Add ';' to HttpUrl.QUERY_COMPONENT_ENCODE_SET

GUIpsp:
This fixes a real world issue found on GAE.

### DIFF
--- a/okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
@@ -259,7 +259,7 @@ public final class HttpUrl {
   static final String PASSWORD_ENCODE_SET = " \"':;<=>@[]\\^`{}|/\\?#";
   static final String PATH_SEGMENT_ENCODE_SET = " \"<>^`{}|/\\?#";
   static final String QUERY_ENCODE_SET = " \"'<>#";
-  static final String QUERY_COMPONENT_ENCODE_SET = " \"'<>#&=";
+  static final String QUERY_COMPONENT_ENCODE_SET = " \"'<>#&=;";
   static final String FRAGMENT_ENCODE_SET = "";
 
   /** Either "http" or "https". */


GUIpsp:
Also, I think having different encoded sets for different parts of the URL is kinda silly, and we should just encode everything the [RFC](http://www.faqs.org/rfcs/rfc3986.html) says we should encode, but that's better left for another patch. This is just a quick fix.

JakeWharton:
The HttpUrl class is not implemented to the URI RFC.

swankjesse:
What's the bug?

GUIpsp:
> The HttpUrl class is not implemented to the URI RFC.

Any reason why it shouldn't?

> What's the bug?

Try posting the equivalent of [this](http://pyeval.appspot.com/index.html) form with a script that contains `;` (say, `print ";"`). A browser will urlencode the `;` and it will work. OKHttp won't, and it will fail.

That form `post`s to `/exec` with `input` set to the python code which returns and id. You then need to `get` `/exec` with `id` set to the ID returned by the `post` to fetch the result of the evaluation.

swankjesse:
Got it. This is about form encoding, which we're already planning on fixing.

swankjesse:
Duplicate of
https://github.com/square/okhttp/issues/1694

GUIpsp:
@swankjesse Neat, but this PR also fixes the issue, and I already signed the CLA.

swankjesse:
Yeah, but this changes the behavior of URL canonicalization, which we don't want. We only want to change form encoding. If you'd like to submit a PR that does that, please do!

GUIpsp:
Is the only char that needs to be added `;`?

swankjesse:
@GUIpsp unsure. You could write a little program to see which characters Chrome and Firefox and Safari and IE escape in forms.

GUIpsp:
From [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent), `encodeURIComponent escapes all characters except the following: alphabetic, decimal digits, - _ . ! ~ * ' ( )`. This seems to match the [WHATWG's recommendations](https://html.spec.whatwg.org/multipage/forms.html#application/x-www-form-urlencoded-encoding-algorithm) `If the byte is in the range 0x2A, 0x2D, 0x2E, 0x30 to 0x39, 0x41 to 0x5A, 0x5F, 0x61 to 0x7A leave the byte as is.` `Otherwise encode`. However the current code present in okhttp uses a blacklist, not a whitelist. Is it worth it to implement a whitelist-like system for forms?

swankjesse:
Yup, OkHttp 2.5.0's behavior is completely wrong, and it's my mistake!

I confirmed with Firefox & Chrome; both do percent-encoding of everything beyond 0x7f which isn't what we're doing.

https://gist.github.com/swankjesse/e46017775bb30f7e52c8

PR #1714, opened 2015-06-21
Fix FormEncodingBuilder to be consistent with browsers.

swankjesse:
Right now we're using the same encode set as the URL query; that's a mistake.


okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
```diff
@@ -256,10 +256,11 @@
   private static final char[] HEX_DIGITS =
       { '0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'A', 'B', 'C', 'D', 'E', 'F' };
   static final String USERNAME_ENCODE_SET = " \"':;<=>@[]^`{}|/\\?#";
-  static final String PASSWORD_ENCODE_SET = " \"':;<=>@[]\\^`{}|/\\?#";
```
swankjesse: The `\\` was duplicated!

PR #1715, opened 2015-06-21
Add HttpUrl methods to MWS and update non-HUC usages.

JakeWharton:


okhttp-tests/src/test/java/com/squareup/okhttp/CacheTest.java
```diff
@@ -601,7 +600,7 @@ private void testClientPrematureDisconnect(TransferKind transferKind) throws IOE
     server.enqueue(new MockResponse()
         .setBody("B"));
 
-    URL url = server.getUrl("/?foo=bar");
+    HttpUrl url = server.url("/").newBuilder().addQueryParameter("foo", "bar").build();
```
JakeWharton: Note the change that had to be made here. The "path" parameter for this method is strictly path.

benchmarks/src/main/java/com/squareup/okhttp/benchmarks/NettyHttpClient.java
```diff
@@ -102,7 +101,7 @@
       }
     }
     if (httpChannel == null) {
-      Channel channel = bootstrap.connect(url.getHost(), Util.getEffectivePort(url))
```
swankjesse: WIN

okhttp-tests/src/test/java/com/squareup/okhttp/CacheTest.java
swankjesse: Dig.

PR #1717, opened 2015-06-21
[mockwebserver] add custom method

the4thfloor:
Add custom method to the list of accepted HTTP methods.

We need this to add extra methods to mock our UPnP server.
See http://upnp.org/specs/arch/UPnP-arch-DeviceArchitecture-v1.1.pdf chapter 4.1 Unicast eventing

Signed-off-by: Ralph Bergmann <ralph@the4thfloor.eu>

### DIFF
--- a/mockwebserver/src/main/java/com/squareup/okhttp/mockwebserver/MockWebServer.java
+++ b/mockwebserver/src/main/java/com/squareup/okhttp/mockwebserver/MockWebServer.java
@@ -47,6 +47,7 @@ import java.security.cert.CertificateException;
 import java.security.cert.X509Certificate;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Locale;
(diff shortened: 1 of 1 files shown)


swankjesse:
This is a weird API, especially since you're putting the behavior on `MockWebServer` but the validation really belongs on `MockResponse`.

I think instead we probably just have to give up on knowing in advance what all of the HTTP methods are. We can hardcode hard rules for a few special methods: `GET`, `HEAD` etc.. For other methods, if there's a `Content-Length` or `Transfer-Encoding` header, then there's a body. Otherwise we'll assume no body.

Does that work for you? Do you want to submit a PR for that?

the4thfloor:
But this validation was already there. I added a way to add extra HTTP methods.

At the end I need SUBSCRIBE and UNSUBSCRIBE both without body.

swankjesse:
I'm still not a fan of the static configuration, and MockWebServer definitely shouldn't need to be configured based on what client expectations are. Can you send a PR to implement policy based on `Content-Length` and `Transfer-Encoding` headers?

PR #1720, opened 2015-06-21
Set default timeouts to ten seconds.

swankjesse:
Closes https://github.com/square/okhttp/issues/1719

### DIFF
--- a/okhttp-tests/src/test/java/com/squareup/okhttp/OkHttpClientTest.java
+++ b/okhttp-tests/src/test/java/com/squareup/okhttp/OkHttpClientTest.java
@@ -57,6 +57,13 @@ public final class OkHttpClientTest {
     Authenticator.setDefault(DEFAULT_AUTHENTICATOR);
   }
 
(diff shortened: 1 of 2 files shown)


okhttp-tests/src/test/java/com/squareup/okhttp/OkHttpClientTest.java
```diff
@@ -57,6 +57,13 @@
     Authenticator.setDefault(DEFAULT_AUTHENTICATOR);
   }
 
+  @Test public void timeoutDefaults() {
+    OkHttpClient client = new OkHttpClient();
+    assertEquals(10_000, client.getConnectTimeout());
```
swankjesse: These timeouts are the first to get cleaned up in OkHttp 3.x

swankjesse: (the `int` type, and lack of time unit in the API)

PR #1732, opened 2015-06-27
Split the web socket callback interface into two.

JakeWharton:
This separates the connection callbacks from the message callbacks. The goal here is making the `WebSocket` type standalone to be passed around before setting up the message callbacks. The new API on WebSocket, start(), must be called within the client's connect timeout for this to work.

This is a potential solution for https://github.com/square/retrofit/issues/924#issuecomment-115956713


okhttp-ws/src/main/java/com/squareup/okhttp/ws/WebSocketCallback.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+package com.squareup.okhttp.ws;
+
+import com.squareup.okhttp.Response;
+import java.io.IOException;
+
+public interface WebSocketCallback {
+  /**
+   * Called when the request has successfully been upgraded to a web socket. You <b>must</b> call
+   * {@link WebSocket#start(WebSocket.Listener) start()} on the {@code WebSocket} with a listener.
```
swankjesse: This interface feels labor intensive. Let's do a Hangout & you can better explain to me what the moving parts are!

JakeWharton: :+1:

swankjesse:
Is this ready for another look?

JakeWharton:
No changes needed in OkHttp.

PR #1756, opened 2015-07-19
Add policy for disconnecting request body halfway.

### LINKED ISSUE #1755: MockWebServer SocketPolicy interrupting request
SocketPolicy has DISCONNECT_AFTER_REQUEST. Is there a way to force the socket to disconnect during the request, i.e. while the request is being read but before it is finished?

JakeWharton:
This also fixes a bug where specifying a policy to disconnect the response body would erroneously also apply to the request body.

Closes #1755.

### DIFF
--- a/mockwebserver/src/test/java/com/squareup/okhttp/mockwebserver/MockWebServerTest.java
+++ b/mockwebserver/src/test/java/com/squareup/okhttp/mockwebserver/MockWebServerTest.java
@@ -21,6 +21,7 @@ import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
+import java.io.OutputStream;
 import java.net.HttpURLConnection;
 import java.net.SocketTimeoutException;
 import java.net.URL;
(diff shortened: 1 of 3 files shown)


JakeWharton:
The number of bytes you can write after socket close varies wildly, of course. So I'm not sure how to write a deterministic test for this. On my machine it always fails within a few bytes, but on Travis CI i've seen up to 10,000 single-byte writes go through before it notices.

swankjesse:
Is this different enough from `DISCONNECT_AT_START` to exist?

JakeWharton:
I guess ask on the linked issue.

mockwebserver/src/test/java/com/squareup/okhttp/mockwebserver/MockWebServerTest.java
```diff
@@ -255,7 +256,28 @@
...
+
+    int i;
+    for (i = 0; i < 20000; i++) {
+      try {
+        out.write('a');
+      } catch (IOException e) {
+        break;
+      }
+    }
+    assertEquals(50f, i / 200f, 2f); // Disconnected at 50% +/- 2%
```
swankjesse: definitely should be at least 50% right?

JakeWharton: Yeah. I certainly can change the test to that. The problem is that on the first go it wrote all 20,000 single-byte writes with no failure.

JakeWharton:
In any case, there's an important bug fix in here. I'll submit that separately.

JakeWharton:
Bug fix is #1758. Closing this until more info on the linked issue.

PR #1759, opened 2015-07-21
Use the same logic for permitting request body as client.

JakeWharton:


mockwebserver/src/main/java/com/squareup/okhttp/mockwebserver/MockWebServer.java
```diff
@@ -629,19 +630,9 @@ private RecordedRequest readRequest(Socket socket, BufferedSource source, Buffer
...
-        || request.startsWith("TRACE ")
-        || request.startsWith("CONNECT ")) {
-      if (hasBody) {
-        throw new IllegalArgumentException("Request must not have a body: " + request);
-      }
-    } else if (!request.startsWith("POST ")
-        && !request.startsWith("PUT ")
-        && !request.startsWith("PATCH ")
-        && !request.startsWith("DELETE ")) { // Permitted as spec is ambiguous.
-      throw new UnsupportedOperationException("Unexpected method: " + request);
```
JakeWharton: This behavior disappears... not sure if it matters.

swankjesse:
Win.

PR #1760, opened 2015-07-21
make the DiskLruCache's initialize() public through the Cache interface

halizton:
The DiskLruCache’s cache does lazy initialization. The time spent in
initialization varies depending on the actual cache size and journal
file size.

On very low end device, the initialization time could be as long as
1000 ms once the cache reaches 100M for example on a 4.0 device. This
has significant performance impact on the very 1st cache hit.

So the proposed solution is to give the developer an opportunity to
explicitly do the initialization early on and in a back ground thread.
Without this change, the developer would have to call any method from
the Cache class to warm up the cache implicitly.

### DIFF
--- a/okhttp/src/main/java/com/squareup/okhttp/Cache.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/Cache.java
@@ -270,6 +270,23 @@ public final class Cache {
   }
 
   /**
+   * Initialize the cache. This will include reading the journal files from
+   * the storage and building up the necessary in-memory cache information.
+   * <p>
+   * The initialization time may vary depending on the journal file size and
+   * the current actual cache size. The application needs to be aware of calling
+   * this function during the initialization phase and preferrably in a background
+   * worker thread.
+   * <p>
+   * Note that if the application chooses to not call this method to initialize
+   * the cache. By default, the okhttp will perform lazy initialization upon the
+   * first usage of the cache.
+   */
+  public void initialize() throws IOException {
+    cache.initialize();
+  }
+
+  /**
    * Closes the cache and deletes all of its stored values. This will delete
    * all files in the cache directory including files that weren't created by
    * the cache.
--- a/okhttp/src/main/java/com/squareup/okhttp/internal/DiskLruCache.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/internal/DiskLruCache.java
@@ -197,8 +197,7 @@ public final class DiskLruCache implements Closeable {
     this.executor = executor;
   }
 
-  // Visible for testing.
-  void initialize() throws IOException {
+  public synchronized void initialize() throws IOException {
     assert Thread.holdsLock(this);
 
     if (initialized) {


okhttp/src/main/java/com/squareup/okhttp/internal/DiskLruCache.java
```diff
@@ -198,7 +198,7 @@ public void run() {
   }
 
   // Visible for testing.
```
JakeWharton: This comment is no longer necessary.

okhttp/src/main/java/com/squareup/okhttp/Cache.java
```diff
@@ -270,6 +270,19 @@ private void abortQuietly(DiskLruCache.Editor editor) {
   }
 
   /**
+   * Initialize the cache. This will include reading the journal files from
+   * the storage and building up the necessary in-memory cache information.
+   *
+   * The initialization time may vary depending on the journal file size and
```
JakeWharton: Need a `<p>` before this paragraph or Javadoc will join them together.

okhttp/src/main/java/com/squareup/okhttp/Cache.java
```diff
@@ -270,6 +270,19 @@ private void abortQuietly(DiskLruCache.Editor editor) {
...
+   * the current actual cache size. The application needs to be aware of calling
+   * this function during the initialization phase and preferrably in a background
+   * worker thread.
+   */
+  public void initialize() throws IOException {
+    cache.initialize();
```
JakeWharton: The method that you are calling here requires a lock on the `cache` object so this call would need surrounded with a `synchronized` block.

okhttp/src/main/java/com/squareup/okhttp/internal/DiskLruCache.java
halizton: accepted and will change.

okhttp/src/main/java/com/squareup/okhttp/Cache.java
halizton: accepted and will change.

okhttp/src/main/java/com/squareup/okhttp/Cache.java
halizton: should we add the synchronized key word to the DiskLruCache::initialize() method or surround the initialize() here with synchronized(cache) ?

Looks like the reason the DiskLruCache::initialize() does not have synchronized keyword is because it is internal to the DiskLruCache class and all the callers of this method have already had the synchronized blocks.

JakeWharton: Yeah it can be added to the method signature.

halizton: cool, will change that way.

okhttp/src/main/java/com/squareup/okhttp/Cache.java
```diff
@@ -270,6 +270,19 @@ private void abortQuietly(DiskLruCache.Editor editor) {
   }
 
   /**
+   * Initialize the cache. This will include reading the journal files from
```
swankjesse: nit: also add a note that if the cache isn't initialized manually, it'll self-initialize on its first request?

swankjesse:
Works for me. One small comment, and also please use http://rebaseandsqua.sh.

swankjesse:
CLA is good!

okhttp/src/main/java/com/squareup/okhttp/Cache.java
halizton: Sure, will apply this.

halizton:
Thanks, swankjesse, I used rebaseandsqua.sh to squash the 3 commits into one.

swankjesse:
Thanks!

PR #1765, opened 2015-07-24
Fold MockWebServerRule into MockWebServer.

JakeWharton:
This avoid needing to duplicate the methods on the rule which historically was a subset of everything.

JUnit is now a required dependency.


swankjesse:
Would it be simpler to flatten the hierarchy?

swankjesse:
LGTM

JakeWharton:
There was a reason I didn't about 6 months ago. I'll try it and see if I
can rediscover why.

JakeWharton:
PTAL!

mockwebserver/src/main/java/com/squareup/okhttp/mockwebserver/MockWebServer.java
```diff
@@ -127,44 +129,61 @@
...
   @Deprecated
   public URL getUrl(String path) {
-    try {
-      return sslSocketFactory != null
-          ? new URL("https://" + getHostName() + ":" + getPort() + path)
-          : new URL("http://" + getHostName() + ":" + getPort() + path);
-    } catch (MalformedURLException e) {
-      throw new AssertionError(e);
-    }
+    return url(path).url();
```
JakeWharton: Winning

mockwebserver/src/main/java/com/squareup/okhttp/mockwebserver/MockWebServer.java
```diff
@@ -177,8 +196,8 @@ public HttpUrl url(String path) {
         .scheme(sslSocketFactory != null ? "https" : "http")
         .host(getHostName())
         .port(getPort())
-        .encodedPath(path)
-        .build();
+        .build()
+        .resolve(path);
```
JakeWharton: This allows values like `/?foo=bar` to work.

swankjesse: Also winning

PR #1767, opened 2015-07-25
Remove IOException throws from source() and close() on ResponseBody.

### LINKED ISSUE #1696: Evaluate removing IOException from ResponseBody close
Propagating the already-rare exception that can be thrown when closing the underlying `Source` of a `ResponseBody` is not actionable and puts the caller in an awkward position. Is there any use to this exception? Or should we simply swallow it for the caller's convenience since it's what they are going to do anyway?

JakeWharton:
`source()` throwing was the result of legacy integrations which now eagerly open their streams. This eliminates the need for this method to throw as well as a few downstream methods like `byteStream()` and `charStream()`.

`close()` throwing puts the caller in an awkward position. They have already read the body (or are ignoring it) and just want to ensure resource cleanup happens. If this fails, however, and the `IOException` was throw there is no reasonable action for them to take other than ignoring. As such, we take the liberty of ignoring it for them.

Closes #1696.

### DIFF
--- a/okhttp-android-support/src/main/java/com/squareup/okhttp/internal/huc/JavaApiConverter.java
+++ b/okhttp-android-support/src/main/java/com/squareup/okhttp/internal/huc/JavaApiConverter.java
@@ -469,26 +469,19 @@ public final class JavaApiConverter {
   /**
    * Creates an OkHttp Response.Body containing the supplied information.
    */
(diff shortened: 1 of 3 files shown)


JakeWharton:
Ugh. I broke something in the least understood part (for me) of the library. Will dig later.

PR #1774, opened 2015-07-28
Hide canceled exception

minakov:


swankjesse:
This doesn't actually cancel the call.

minakov:
What reason to execute canceled AsyncCall?

Issue #1459, opened 2015-02-25
Make Cache class non-final for purposes of extending


JaredBanyard:
It would be great if the Cache class was non-final and built in a way that promoted extending for the purposes of encrypting, restructuring, or enhancing.

I know this stack overflow says that you guys may be changing the Caching mechanism frequently and that its not a good idea to roll our own cache:
http://stackoverflow.com/questions/22373763/encrypting-okhttps-httpresponsecache

Thoughts?

swankjesse:
We're about to redo the current cache to support cache-push on HTTP/2. The cache is going to be very tricky code: we want to use it as the center of our request joining and canceling system.

It'll make this change impossible if the cache is pluggable.

That said, I'm sympathetic to your request. What features do you want, and are you willing to submit pull requests? Writing code like this for reuse is a lot of work, but if the feature is compelling enough, we could include it t in the core.

JaredBanyard:
Well primarily, I am looking to pass all read/writes to the DiskLruCache through encryption/decryption streams.  Facebook has made a surprisingly efficient library called Conceal that must have had an enormous amount of manpower to develop. It is extremely fast and I have successfully implemented and integrated a file encryption class that utilizes Conceal to wrap DiskLruCache.  It is extremely fast and works on every device I've seen.  So if there is an opportunity to expose the raw input/output streams to DiskLruCache in your new implementation, it should be possible to hook in libraries like this.  I would gladly contribute if you all think this is possible.

swankjesse:
@JaredBanyard we could probably get by with a small indirection to the file system.
```
interface Filesystem {
  Source source(String path);
  Sink sink(String path);
  Sink appendingSink(String path);
  void delete(String path);
  void atomicRename(String from, String to);
}
```
Behind-the-scenes you could do whatever. Encrypt per-file, keep things in-memory, in-memory cache. We'd start with this as a non-public API in `internal` (subject to API changes between OkHttp releases), but it'd give you the indirection you need without real cost on OkHttp itself.

@JakeWharton am I a crazy person for pondering the idea of putting `FileSystem` into Okio?

JakeWharton:
Probably! You either have too many methods (`delete` and `rename`) or too few (`exists`, `length`, etc.).

JaredBanyard:
[...]

PR #1781, opened 2015-08-01
Expose internal APIs for pluggable file systems.

swankjesse:
We aren't yet ready to make FileSystem a public type, but I don't mind
making it _almost_ available via an internal API for those brave enough
to try that.

Also migrate all of our tests to use the in memory file system. It's simpler.

https://github.com/square/okhttp/issues/1459


okhttp/src/main/java/com/squareup/okhttp/Cache.java
```diff
@@ -165,7 +165,11 @@
   private int requestCount;
 
   public Cache(File directory, long maxSize) {
-    cache = DiskLruCache.create(FileSystem.SYSTEM, directory, VERSION, ENTRY_COUNT, maxSize);
+    this(directory, maxSize, FileSystem.SYSTEM);
+  }
+
+  Cache(File directory, long maxSize, FileSystem fileSystem) {
```
swankjesse: This is the new seam that allows daring application developers to plug in their own file system wrappers.

swankjesse: internal-only, but if it proves useful we can make it public later (possibly in okio).

Issue #1779, opened 2015-08-01
Cannot cancel async post/get requests.


kdpdev:
Cannot cancel async POST/GET requests. To reproduse it i use simple
node.js server, that accepts all incomming requests and just log
it to console.
Steps to reproduce:
1. Make sure that a machine with the Node.js server is available
   for a device with OkHttp client. (Server is not started!).
2. Create POST request from OkHttp client.
3. Wait a little (3-5 ceconds).
4. Cancel request.
5. Start Node.js server and track its output.
6. Node.js accept POST request and process it.
7. Wait for connection timeout period and got onFailure call on
   OkHttp client.

I.e. cancelled and not finished request comes to server and server
handle it.

The problem is in okhttp/internal/http/HttpEngine.java file
Lines 247-251:
.....
if (connection == null) {
  connect();
}

transport = Internal.instance.newTransport(connection, this);
....

While server is not started we are inside connect method until
connection timeout. If we try to cancel request at this time,
the disconnect method is called:

public void disconnect() {
  if (transport != null) {
    try {
      transport.disconnect(this);
    } catch (IOException ignored) {
    }
  }
}

It does nothing as transport is not created yet (we are inside
connect method). After starting server the connect method
returns and next code correctly send request to server.

For post requests there is workaround: wrap RequestBody with
cancellable body that throws IOException("Cancelled") inside
its methods in cancelled state. But it is not convinient. For
get GET requests there is no any workaround.

PR #1782, opened 2015-08-01
Make call canceling more reliable.

swankjesse:
We had a bug where the socket-being-connected wasn't being closed when the
application used Call.cancel(). The problem is that the SocketConnector model
assumes the Connection doesn't want a Socket instance until it's fully
connected.

This moves the SocketConnector code back into Connection, removes a lot of
nested try/catch blocks, and assigns a Socket instance as soon as its created.

This also likely fixes some bugs where sockets weren't being closed when
an IOException or RouteException was thrown during connection. Now we always
close at the top level of connect() unless the connection is successful.

https://github.com/square/okhttp/issues/1779


swankjesse:
FYI @nfuller @kdpdev

JakeWharton:
Git was not kind to this diff.

Issue #1044, opened 2014-09-04
Gracefully work around invalid URLs


hrj:
With the default HttpURLStreamHandler, URLs are automatically encoded correctly.
For example, this URL: `http://host/abc|xyz` works fine with the default handler.

But OkHttp throws an exception:

```

java.io.IOException: Illegal character in path at index 15: http://host/abc|xyz
at com.squareup.okhttp.Request.uri(Request.java:65)
at com.squareup.okhttp.internal.http.HttpEngine.networkRequest(HttpEngine.java:579)
at com.squareup.okhttp.internal.http.HttpEngine.sendRequest(HttpEngine.java:212)
at com.squareup.okhttp.internal.huc.HttpURLConnectionImpl.execute(HttpURLConnectionImpl.java:420)
at com.squareup.okhttp.internal.huc.HttpURLConnectionImpl.getResponse(HttpURLConnectionImpl.java:371)
at com.squareup.okhttp.internal.huc.HttpURLConnectionImpl.getResponseCode(HttpURLConnectionImpl.java:466)
```

swankjesse:
I'm punting this to 2.4. The focus of that release will be our own URL API. I think can probably do something that's better-enough from `java.net.URL` and `java.net.URI` that it justifies its own existence.

Handling this kind of thing is just one example.

swankjesse:
HttpUrl has this now. We still need to offer HttpUrl as a mechanism that you can use with the rest of OkHttp.

swankjesse:
I'm going to actually wire this into HttpEngine in 2.5.

PR #1784, opened 2015-08-03
Use HttpUrl internally.

swankjesse:
Closes https://github.com/square/okhttp/issues/1044


okhttp/src/main/java/com/squareup/okhttp/internal/http/HttpEngine.java
```diff
@@ -1091,14 +1082,14 @@ public Request followUpRequest() throws IOException {
 
         String location = userResponse.header("Location");
         if (location == null) return null;
-        URL url = new URL(userRequest.url(), location);
+        HttpUrl url = userRequest.httpUrl().resolve(location);
 
         // Don't follow redirects to unsupported protocols.
-        if (!url.getProtocol().equals("https") && !url.getProtocol().equals("http")) return null;
+        if (url == null) return null;
```
JakeWharton: :+1:

JakeWharton:
Lovely

Issue #891, opened 2014-06-03
Non-ASCII Header Encoding Issues


JakeWharton:
When passing headers parameters containing a 'é' it is received on server side as 'Ã©'.
It might seem wird to pass such chars as arguments but I can't modify server side. 
I know this behaviour is specific to retrofit  because it worked with the old client using HttpUrlConnection and it also works with Postman chrome plugin.

via @cmorin6 from https://github.com/square/retrofit/issues/516

yyjdelete:
Android use the below method in com.android.internal.os.RuntimeInit to set default UA, and okhttp-urlconnection direct use the UA as default http head without encode it. 
```java
    /**
     * Returns an HTTP user agent of the form
     * "Dalvik/1.1.0 (Linux; U; Android Eclair Build/MASTER)".
     */
    private static String getDefaultUserAgent() {
        StringBuilder result = new StringBuilder(64);
        result.append("Dalvik/");
        result.append(System.getProperty("java.vm.version")); // such as 1.1.0
        result.append(" (Linux; U; Android ");

        String version = Build.VERSION.RELEASE; // "1.0" or "3.4b5"
        result.append(version.length() > 0 ? version : "1.0");

        // add the model for the release build
        if ("REL".equals(Build.VERSION.CODENAME)) {
            String model = Build.MODEL;
            if (model.length() > 0) {
                result.append("; ");
                result.append(model);
            }
        }
        String id = Build.ID; // "MASTER" or "M4-rc20"
        if (id.length() > 0) {
            result.append(" Build/");
            result.append(id);
        }
        result.append(")");
        return result.toString();
    }
```
As an result, when I use okhttp-urlconnection in my apk, my server recevied the below request from an device with non-ASCII MODEL, which is invalid.

![Model with non-ASCII char](https://cloud.githubusercontent.com/assets/1819074/8028576/47bf0c38-0de3-11e5-9c7b-ed93933ed693.png)

yyjdelete:
AnyOne?

swankjesse:
Yep, we should sanitize the User-Agent.

swankjesse:
Working plan: throw an exception on non-ASCII headers in `Headers.Builder`.

vvelmurugan:
@swankjesse Why the add method is not throwing illegalArgumentException, for us developers to catch and handle the error and mostly importantly to know that this method will throw this exception. Not many will look into the source and find that the header values are sanitised strictly.

swankjesse:
[...]

PR #1785, opened 2015-08-03
Be strict on invalid characters in request headers.

swankjesse:
This includes potential security problems (newline characters) as well as
simple non-ASCII characters including international characters and emoji.

Closes https://github.com/square/okhttp/issues/891


swankjesse:
FYI @nfuller this is a behavior change!

Issue #1788, opened 2015-08-04
Changing HttpUrl's scheme should fix default ports


swankjesse:
Right now if I create a Builder starting with `http://example.com/` and change the scheme to `https`, it retains port 80. This is a mistake.

JakeWharton:
Fixed by #1789.

PR #1789, opened 2015-08-05
Changing the scheme will update the default port

MichaelEvans:
Addresses #1788 

Not really sure this is the best solution, but my idea was to check the port when you're building the Url, and if the port is a default (which would be assigned from parsing a http(s)), use the default port for the scheme instead. 

I'd be glad to make changes to my approach if you had something else in mind!

### DIFF
--- a/okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
@@ -570,7 +570,12 @@ public final class HttpUrl {
     result.encodedUsername = encodedUsername();
     result.encodedPassword = encodedPassword();
     result.host = host;
-    result.port = port;
+    // If we're set to a default port, unset it, in case of a scheme change.
+    if (port == defaultPort(scheme)) {
+      result.port = -1;
+    } else {
+      result.port = port;
+    }
     result.encodedPathSegments.clear();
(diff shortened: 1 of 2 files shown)


okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
```diff
@@ -831,6 +831,8 @@ public Builder encodedFragment(String encodedFragment) {
     public HttpUrl build() {
       if (scheme == null) throw new IllegalStateException("scheme == null");
       if (host == null) throw new IllegalStateException("host == null");
+      // If we're set to a default port, update it to match the default of the scheme.
+      if (port == 80 || port == 443) port = defaultPort(scheme);
```
swankjesse: I _think_ we want to set the port to -1 when setting the port to what would be its default value. That way you can still get HTTPS on 80 and HTTP on 443 as long as you set the scheme first.

MichaelEvans:
I think this is more in line with what you had in mind? (And the tests pass this time, which is a plus :stuck_out_tongue:)

Lemme know if you think there are other changes needed.

swankjesse:
Awesome. I just need you to complete square's CLA.
https://github.com/square/okhttp/blob/master/CONTRIBUTING.md

MichaelEvans:
Done!

Issue #1790, opened 2015-08-05
Define policy on timeouts and retries


swankjesse:
@nfuller:

> Here: [HttpEngine.isRecoverable](https://github.com/square/okhttp/blob/master/okhttp/src/main/java/com/squareup/okhttp/internal/http/HttpEngine.java#L440)
> The catch should be treating SocketTimeoutException (which, you'll recall, is subclass of InterruptedIOException) as a "retryable". InterruptedIOException should probably still be non-retryable In this one case. The reasoning being that timeouts can be the result of misconfigured networking, but other InterruptedIOExceptions may be a call to Thread.interrupt() (e.g. issue #1146). Other cases still look like SocketTimeoutException should be treated the same as InterruptedIOException (to me) so shouldn't need to be changed, except to point out that the logic is no longer common.

I think we should figure out what we want the policy to be, and then implement that exactly. And I think we probably want the policy to be that timeouts are short (say _5_ or _10_ seconds by default) and retries are automatic.

We need to figure out how we count various things: do 3 SSL negotiations count as a single attempt? And how many attempts to we make to a webservice with 5 IP addresses?

Until now this hasn't really been decided, and it should be! Plus we should figure out what knobs to expose.

Related: https://github.com/square/okhttp/issues/1146
And of course this: https://twitter.com/mumak/status/628489427489288192

nfuller:
Initially I'd like to just put the behavior back the way it was (intended to be).

The Android connectivity detection code uses the HttpURLConnection and assumes (rightly or wrongly) that a certain subset of "routes" are tried (where "route" is a combination of proxy/direct, ip addresses).

Possibly related: We Android bugs raised occasionally where people have complained that HttpURLConnection.setConnectTimeout() and setReadTimeout() actually means we can wait N times those values because those timeouts are observed per socket. It's hard for apps to set a "correct" timeout in terms of "time that will actually elapse" because an arbitrary URL may resolve to an arbitrary number of socket connections (N).

However, I've mostly pushed back on any modification because it will break existing apps and any other interpretation of the URL connection timeout is also prone to problems, particularly when misconfigured networks can just blackhole connection attempts and we always run into the underlying socket timeout. For true flexibility developers would probably need separate "HTTP connect timeout" and "Socket connect timeout" settings.

nfuller:
I have updated the commit to address most comments. I'm not sure what to do about IPv6 addresses with scopes.

swankjesse:
[...]

Issue #1146, opened 2014-11-11
Request retrying on InterruptedIOException


DanielNovak:
OkHttp 1.5.3 added basic support for Thread.interrupt() in commit https://github.com/square/okhttp/commit/664b65d18e164a571a2c8498fe372d3f934499c3. This throws an InterruptedIOException but this is then handled by OkHttp's http retry engine in the catch block of HttpURLConnectionImpl.execute() method **and the request is retried** - which I think should not happen, because I deliberately interrupted the Thread and IO should stop (based on the commit from @swankjesse). I put logging inside OkHttp to verify that the same request is retried under some circumstances after calling Thread.interrupt() and the execute() catch block is indeed handling the InterruptedIOException.

PR #1793, opened 2015-08-06
Partially revert connection timeout behavior

nfuller:
Related to issue #1790.

If there is a *connection* timeout the next route should
be tried. Timeouts during a request/response should
probably not be retried.

The timeout/interrupt behavior was changed in PR #1257
for issue #1146. This modified both connection and
request/response behavior. The issue reported was actually
about the use of Thread.interrupt(), not timeouts, but the
behavior change modified both behavior for interrupt and
timeout.

PR #1388 split the handling of exceptions so that separate
code now handles connection errors from that which handles
request/response errors. It faithfully kept the behavior
from PR #1257.

The modification of the timeout behavior during *connection*
caused problems on Android. Now that the two types of error
handling are split it's possible to deal with them
differently and return the connection behavior to trying
the next route. This does not modify behavior during
request/response.

A test has been added for timeouts. It does not seem to be
possible to interrupt a Socket.connect() call with
Thread.interrupt() on either OpenJDK or Android so a
placeholder has been implemented. If it ever is implemented
a Thread.interrupt() during connect will not be retried.

### DIFF
--- a/okhttp/src/main/java/com/squareup/okhttp/internal/ConnectionSpecSelector.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/internal/ConnectionSpecSelector.java
@@ -91,13 +91,18 @@ public final class ConnectionSpecSelector {
     // Any future attempt to connect using this strategy will be a fallback attempt.
     isFallback = true;
 
-    // TODO(nfuller): This is the same logic as in HttpEngine.
+    if (!isFallbackPossible) {
+      return false;
+    }
+
     // If there was a protocol problem, don't recover.
     if (e instanceof ProtocolException) {
       return false;
     }
 
-    // If there was an interruption or timeout, don't recover.
+    // If there was an interruption or timeout (SocketTimeoutException), don't recover.
+    // For the socket connect timeout case we do not try the same host with a different
+    // ConnectionSpec: we assume it is unreachable.
     if (e instanceof InterruptedIOException) {
       return false;
     }
@@ -115,13 +120,11 @@ public final class ConnectionSpecSelector {
       // e.g. a certificate pinning error.
       return false;
     }
-    // TODO(nfuller): End of common code.
 
 
     // On Android, SSLProtocolExceptions can be caused by TLS_FALLBACK_SCSV failures, which means we
     // retry those when we probably should not.
-    return ((e instanceof SSLHandshakeException || e instanceof SSLProtocolException))
-        && isFallbackPossible;
+    return (e instanceof SSLHandshakeException || e instanceof SSLProtocolException);
   }
 
   /**
--- a/okhttp/src/main/java/com/squareup/okhttp/internal/http/HttpEngine.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/internal/http/HttpEngine.java
@@ -40,6 +40,7 @@ import java.io.InterruptedIOException;
 import java.net.CookieHandler;
 import java.net.ProtocolException;
 import java.net.Proxy;
+import java.net.SocketTimeoutException;
 import java.security.cert.CertificateException;
 import java.util.Date;
 import java.util.List;
@@ -430,15 +431,15 @@ public final class HttpEngine {
 
     IOException ioe = e.getLastConnectException();
 
-    // TODO(nfuller): This is the same logic as in ConnectionSpecSelector
     // If there was a protocol problem, don't recover.
     if (ioe instanceof ProtocolException) {
       return false;
     }
 
-    // If there was an interruption or timeout, don't recover.
+    // If there was an interruption don't recover, but if there was a timeout
+    // we should try the next route (if there is one).
     if (ioe instanceof InterruptedIOException) {
-      return false;
+      return ioe instanceof SocketTimeoutException;
     }
 
     // Look for known client-side or negotiation errors that are unlikely to be fixed by trying
@@ -454,7 +455,6 @@ public final class HttpEngine {
       // e.g. a certificate pinning error.
       return false;
     }
-    // TODO(nfuller): End of common code.
 
     // An example of one we might want to retry with a different route is a problem connecting to a
     // proxy and would manifest as a standard IOException. Unless it is one we know we should not


mockwebserver/src/main/java/com/squareup/okhttp/mockwebserver/MockWebServer.java
```diff
@@ -164,6 +165,18 @@ public int getPort() {
 
   public String getHostName() {
     maybeStart();
+    if (inetSocketAddress.getAddress() instanceof Inet6Address) {
+      Inet6Address address = (Inet6Address) inetSocketAddress.getAddress();
+      String hostName = inetSocketAddress.getHostName();
```
swankjesse: Does this rally happen?! I think this wants to be a separate PR.

okhttp-tests/src/test/java/com/squareup/okhttp/DelegatingSocket.java
```diff
@@ -0,0 +1,200 @@
...
+import java.net.InetAddress;
+import java.net.Socket;
+import java.net.SocketAddress;
+import java.net.SocketException;
+import java.nio.channels.SocketChannel;
+
+/**
+ * An {@link Socket} that delegates all calls.
+ */
+public abstract class DelegatingSocket extends Socket {
```
swankjesse: There isn't a better way to test this? I nolove classes like this, particularly because self-use doesn't work.

okhttp-tests/src/test/java/com/squareup/okhttp/internal/http/ThreadInterruptTest.java
```diff
@@ -45,6 +59,14 @@
   // The size of the socket buffers in bytes.
   private static final int SOCKET_BUFFER_SIZE = 256 * 1024;
 
+  // 203.0.113.255 reserved for documentation.
```
swankjesse: That's awesome.

okhttp-tests/src/test/java/com/squareup/okhttp/internal/http/ThreadInterruptTest.java
```diff
@@ -45,6 +59,14 @@
...
+  private static final byte[] IPV4_BLACKHOLE_ADDRESS =
```
swankjesse: Shouldn't the type of this be `InetAddress` ?

okhttp-tests/src/test/java/com/squareup/okhttp/internal/http/ThreadInterruptTest.java
```diff
@@ -45,6 +59,14 @@
...
+      new byte[] { (byte) 203, 0, 113, (byte) 255 };
+
+  // 2001:db8::1 reserved for documentation.
+  public static final byte[] IPV6_BLACKHOLE_ADDRESS =
```
swankjesse: Simliarly

okhttp-tests/src/test/java/com/squareup/okhttp/internal/http/ThreadInterruptTest.java
```diff
@@ -70,6 +92,35 @@ protected void configureSocket(Socket socket) throws IOException {
     });
   }
 
+  @Test @Ignore public void interruptConnect() throws Exception {
```
swankjesse: Is this true? What about when building an tunnel for HTTPS through an HTTP proxy? That'll fire during InputStream.read().

okhttp-tests/src/test/java/com/squareup/okhttp/internal/http/ThreadInterruptTest.java
```diff
@@ -133,4 +184,103 @@ private void interruptLater(final int delayMillis) {
...
+  }
+
+  private static InetAddress getBlackholeAddress(InetAddress localAddress)
+      throws UnknownHostException {
+    byte[] addressBytes =
+        localAddress instanceof Inet4Address ? IPV4_BLACKHOLE_ADDRESS : IPV6_BLACKHOLE_ADDRESS;
+    return InetAddress.getByAddress(addressBytes);
+  }
+
+  private static class TimeoutCountingSocketFactory extends DelegatingSocketFactory {
```
swankjesse: I think you'll want to change DelegatingSocketFactory so that configureSocket returns a Socket. That way we define these 5 overloads only once!

okhttp-tests/src/test/java/com/squareup/okhttp/DelegatingSocket.java
swankjesse: On further inspection, we're already doing this for lots of things.

swankjesse: ... can you make DelegatingSSLSocket extend this?

okhttp-tests/src/test/java/com/squareup/okhttp/DelegatingSocket.java
```diff
@@ -0,0 +1,200 @@
...
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.net.InetAddress;
+import java.net.Socket;
+import java.net.SocketAddress;
+import java.net.SocketException;
+import java.nio.channels.SocketChannel;
+
+/**
+ * An {@link Socket} that delegates all calls.
```
swankjesse: nit: An -> A

swankjesse:
LGTM

mockwebserver/src/main/java/com/squareup/okhttp/mockwebserver/MockWebServer.java
nfuller: It did happen for me in the test I'd written because it grabs the first non-loopback NIC it finds and then grabs the first local IP address. On my Linux machine it was using IPv6 and everything broke.

i.e.
inetSocketAddress.getHostName() returns "<ipv6 address>%2"

Calling MockWebServer.getUrl() exploded. Looks like scopes are generally not expected. HttpUrl throws IllegalArgumentException and if we got through that URL would throw MalformedURLException unless the % was escaped (and nothing expects the scope later either). Even whether the % should be escaped appears to be a bit controversial.

[I couldn't think of a better way of forcing a connection timeout as loopback addresses can't be "blackholed" with an unroutable address.]

As I said: not sure what to do about it and I agree a second PR once the way forward is clearer.

okhttp-tests/src/test/java/com/squareup/okhttp/DelegatingSocket.java
nfuller: > ... can you make DelegatingSSLSocket extend this?

Difficult without multiple inheritance because SSLSocket is concrete.

okhttp-tests/src/test/java/com/squareup/okhttp/internal/http/ThreadInterruptTest.java
nfuller: To switch I think it requires that we handle the UnknownHostException exception from getByAddress(): if we just throw it the class might fail to initialize if we don't have both IPv6 and IPv4. We could set it to null in that case, I suppose.

okhttp-tests/src/test/java/com/squareup/okhttp/internal/http/ThreadInterruptTest.java
nfuller: As above.

okhttp-tests/src/test/java/com/squareup/okhttp/internal/http/ThreadInterruptTest.java
nfuller: In this case we know that's not happening because of the way the MockWebServer is configured, but you're right that is not tested.

A lot of the complexity of this change comes about from trying to force blocking behavior so we can then timeout or interrupt it. Not very happy with the test but it seemed better than no test and confirmed the code was broken before and better after.

Forego the test instead?

okhttp-tests/src/test/java/com/squareup/okhttp/internal/http/ThreadInterruptTest.java
nfuller: Done for this and others.

swankjesse:
Travis suggests the new test is flaky.
```
Tests run: 4, Failures: 1, Errors: 0, Skipped: 1, Time elapsed: 1.195 sec <<< FAILURE! - in com.squareup.okhttp.internal.http.ThreadInterruptTest
timeoutConnect(com.squareup.okhttp.internal.http.ThreadInterruptTest)  Time elapsed: 0.011 sec  <<< FAILURE!
java.lang.AssertionError: expected:<1> but was:<0>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:743)
	at org.junit.Assert.assertEquals(Assert.java:118)
	at org.junit.Assert.assertEquals(Assert.java:555)
	at org.junit.Assert.assertEquals(Assert.java:542)
	at com.squareup.okhttp.internal.http.ThreadInterruptTest$TimeoutCountingSocketFactory.assertTimeoutCount(ThreadInterruptTest.java:233)
	at com.squareup.okhttp.internal.http.ThreadInterruptTest.timeoutConnect(ThreadInterruptTest.java:122)
```

nfuller:
I am not surprised!

okhttp-tests/src/test/java/com/squareup/okhttp/internal/http/ThreadInterruptTest.java
swankjesse: Yeah.

nfuller:
I have removed the test.

PR #1794, opened 2015-08-06
Fixes #1787

### LINKED ISSUE #1787: Attempt to invoke virtual method 'void java.net.Socket.close()' on a null object reference
I try to check fix for Cannot cancel async post/get requests. #1779 and got null pointer exception.
Trace:
java.lang.NullPointerException: Attempt to invoke virtual method 'void java.net.Socket.close()' on a null object reference
            at com.squareup.okhttp.Connection.closeIfOwnedBy(Connection.java:147)
            at com.squareup.okhttp.OkHttpClient$1.closeIfOwnedBy(OkHttpClient.java:75)
            at com.squareup.okhttp.internal.http.HttpEngine.disconnect(HttpEngine.java:577)
            at com.squareup.okhttp.Call.cancel(Call.java:122)

Steps to reproduce are described in #1779. But there is one detail:
When i am trying to connect to not started server, i got onFailure as expected. But a little bit later i invoke call.cancel() for that failed call.

The code in Connection.java (lines 174-175) closes the socket and makes it nullable:
        Util.closeQuietly(socket);
        socket = null;

felipecsl:
Not sure this is the right way to fix it (probably not), but just in case...

### DIFF
--- a/okhttp/src/main/java/com/squareup/okhttp/Connection.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/Connection.java
@@ -144,7 +144,9 @@ public final class Connection {
     }
 
     // Don't close() inside the synchronized block.
-    socket.close();
+    if (socket != null) {
+      socket.close();
+    }
   }
 
   void connect(int connectTimeout, int readTimeout, int writeTimeout, Request request,


swankjesse:
It's not the complete long-term fix, but it mitigates it. Thanks!

Issue #1651, opened 2015-05-15
IllegalStateException - Http2$Writer.headers


juliocotta:
I found this crash in crashlytics Android 5.0 - XT1097
```java
java.lang.IllegalStateException
       at com.squareup.okhttp.internal.spdy.Http2$Writer.headers(Http2.java:444)
       at com.squareup.okhttp.internal.spdy.Http2$Writer.synStream(Http2.java:410)
       at com.squareup.okhttp.internal.spdy.SpdyConnection.newStream(SpdyConnection.java:266)
       at com.squareup.okhttp.internal.spdy.SpdyConnection.newStream(SpdyConnection.java:242)
       at com.squareup.okhttp.internal.http.SpdyTransport.writeRequestHeaders(SpdyTransport.java:89)
       at com.squareup.okhttp.internal.http.HttpEngine$NetworkInterceptorChain.proceed(HttpEngine.java:814)
       at com.squareup.okhttp.internal.http.HttpEngine.readResponse(HttpEngine.java:684)
       at com.squareup.okhttp.Call.getResponse(Call.java:272)
       at com.squareup.okhttp.Call$ApplicationInterceptorChain.proceed(Call.java:228)
       at com.squareup.okhttp.Call.getResponseWithInterceptorChain(Call.java:199)
       at com.squareup.okhttp.Call.access$100(Call.java:34)
       at com.squareup.okhttp.Call$AsyncCall.execute(Call.java:162)
       at com.squareup.okhttp.internal.NamedRunnable.run(NamedRunnable.java:33)
       at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1112)
       at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:587)
       at java.lang.Thread.run(Thread.java:818)
```

swankjesse:
Are you able to reproduce? Seems impossible!

juliocotta:
When the error happens (happened 7 times in crashlytics log) I have a lot of threads (100 or more) with the name like "OkHttp https://maps.googleapis.com/maps/api/geocode/json?&latlng=-31.011166%2C-50.188923" with the latlng changing a bit (The app is tracking the user location)...the bunch of requests may be some bug in my code, but the error itself does not seems to be. 
Each thread is in the same point when the error happens:
```java
thread
       at libcore.io.Posix.android_getaddrinfo(Posix.java)
[...]

PR #1799, opened 2015-08-10
Don't throw IllegalStateException after an IOException.

swankjesse:
We were a little too aggressive in checking our preconditions. If a previous
write failed, we don't want the following write to crash. (It'll throw an
IOException anyway.)

Closes https://github.com/square/okhttp/issues/1651

### DIFF
--- a/okhttp-tests/src/test/java/com/squareup/okhttp/internal/framed/Http2ConnectionTest.java
+++ b/okhttp-tests/src/test/java/com/squareup/okhttp/internal/framed/Http2ConnectionTest.java
@@ -17,6 +17,7 @@ package com.squareup.okhttp.internal.framed;
 
 import com.squareup.okhttp.internal.Util;
 import java.io.IOException;
+import java.net.Socket;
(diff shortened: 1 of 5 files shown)


adriancole:
Good idea and easy to understand test.

:shipit:

okhttp-tests/src/test/java/com/squareup/okhttp/internal/framed/Http2ConnectionTest.java
```diff
@@ -427,6 +429,37 @@ private Buffer data(int byteCount) {
     assertEquals(CANCEL, rstStream.errorCode);
   }
 
+  /**
+   * When writing a set of headers fails due to an {@code IOException}, make sure the writer is left
+   * in a consistent state so the next writer also gets an {@code IOException} also instead of
+   * something worse (like an {@link IllegalStateException}.
```
JakeWharton: ubbber nit: missing closing paren

PR #1808, opened 2015-08-17
OAuth 2.0 authorization header support

vicmosin:
Added OAuth 2.0 authorization support based on RFC6750 (http://tools.ietf.org/html/rfc6750)


swankjesse:
I don't think we gain anything from the additional tests. The new API method isn't very interesting. Is OAuth the only thing that uses bearer tokens?

Issue #1684, opened 2015-06-03
EBDAF Exception


pguilbot:
I use OkHTTP for my android app.

I get a lot of crashes related to an Http issue :
java.lang.AssertionError: libcore.io.ErrnoException: getsockname failed: EBADF (Bad file number)

Coming from : 
at com.squareup.okhttp.internal.http.SocketConnector.connectTls(SocketConnector.java:103)
or
at com.squareup.okhttp.internal.Util.closeQuietly(Util.java:105)
or
at com.squareup.okhttp.internal.Platform$Android.connectSocket(Platform.java:190)

Is there a solution to catch this exception ?

Thanks

swankjesse:
@nfuller any ideas?

iNoles:
Could be https://code.google.com/p/android/issues/detail?id=54072 ?

nfuller:
@pguilbot  Going to need more information. What version of Android is affected? I've found similar examples from ~2012 (against ICS). Are you doing things from multiple threads?

Can you send the full stack traces, too? Feel free  to obfuscate your own app's parts of the stack.

The one @iNoles also looks like a possibility. Hard to tell without more info, though.

There's also https://code.google.com/p/android/issues/detail?id=81700, which apparently I own. OkHttp doesn't use nio, though, from what I can see so unless it has a common cause it's probably not that.

swankjesse:
@pguilbot if you could provide a URL that reproduces this, or a test case, or anything we can make forward progress & I'll reopen this issue. For now we're stalled on insufficient information.

pguilbot:
Hi all, 
Thanks for your fast reply.

I think that the @iNoles issue is definitely a possibility.

Here are some full stack traces causing this issue : 

------------------------

java.lang.AssertionError: libcore.io.ErrnoException: getsockname failed: EBADF (Bad file number)
       at libcore.io.IoBridge.getSocketLocalPort(IoBridge.java:649)
       at libcore.io.IoBridge.closeSocket(IoBridge.java:202)
       at java.net.PlainSocketImpl.close(PlainSocketImpl.java:162)
       at java.net.Socket.close(Socket.java:319)
       at org.apache.harmony.xnet.provider.jsse.OpenSSLSocketImpl.close(OpenSSLSocketImpl.java:928)
       at org.apache.harmony.xnet.provider.jsse.OpenSSLSocketImpl.startHandshake(OpenSSLSocketImpl.java:425)
[...]

PR #1817, opened 2015-08-25
Recover from Android 4.2.2 EBADF crashes.

swankjesse:
Originally submitted by @pguilbot.

See also https://github.com/square/okhttp/issues/1684

### DIFF
--- a/okhttp/src/main/java/com/squareup/okhttp/internal/Util.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/internal/Util.java
@@ -84,6 +84,8 @@ public final class Util {
     if (socket != null) {
       try {
         socket.close();
+      } catch (AssertionError e) {
+        if (!isAndroidGetsocknameError(e)) throw e;
       } catch (RuntimeException rethrown) {
         throw rethrown;
       } catch (Exception ignored) {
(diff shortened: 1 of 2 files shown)


okhttp/src/main/java/com/squareup/okhttp/internal/Util.java
```diff
@@ -277,4 +279,13 @@ public static String toHumanReadableAscii(String s) {
     }
     return s;
   }
+
+  /**
+   * Returns true if {@code e} is due to a firmware bug fixed after Android 4.2.2.
+   * https://code.google.com/p/android/issues/detail?id=54072
+   */
+  public static boolean isAndroidGetsocknameError(AssertionError e) {
```
JakeWharton: This method name makes my left eye twitch.

JakeWharton:
LGTM

okhttp/src/main/java/com/squareup/okhttp/internal/Util.java
swankjesse: It's deliberately ugly, named after the ugliness it bandaids.

swankjesse:
@pguilbot I'm doing a round of releases tonight.

Issue #1827, opened 2015-08-28
HttpURLConnection behavior change for badly formed URLs


nfuller:
Prior to the move to HttpUrl the code would throw UnknownHostException for (e.g.) hosts with spaces in them. After the move, an IllegalArgumentException is thrown.

[I have a fix coming: I am creating this issue for tracking purposes]

nfuller:
Thanks!

PR #1828, opened 2015-08-28
Throw UnknownHostException again from HttpURLConnection

nfuller:
This is to address issue #1827.

### DIFF
--- a/okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
@@ -561,7 +561,9 @@ public final class HttpUrl {
 
   /** Returns the URL that would be retrieved by following {@code link} from this URL. */
   public HttpUrl resolve(String link) {
-    return new Builder().parse(this, link);
+    Builder builder = new Builder();
+    Builder.ParseResult result = builder.parse(this, link);
+    return result == Builder.ParseResult.SUCCESS ? builder.build() : null;
   }
 
   public Builder newBuilder() {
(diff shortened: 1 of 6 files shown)


okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
```diff
@@ -584,11 +586,13 @@ public Builder newBuilder() {
   }
 
   /**
-   * Returns a new {@code OkUrl} representing {@code url} if it is a well-formed HTTP or HTTPS URL,
-   * or null if it isn't.
+   * Returns a new {@code HttpUrl} representing {@code url} if it is a well-formed HTTP or HTTPS
```
swankjesse: Thanks!

okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
```diff
@@ -599,6 +603,29 @@ public static HttpUrl get(URL url) {
...
   }
 
+  /**
+   * Returns a new {@code HttpUrl} representing {@code url} if it is a well-formed HTTP or HTTPS
+   * URL, or throws an exception if it isn't.
+   *
+   * @throws MalformedURLException if there was a non-host related URL issue
+   * @throws UnknownHostException if the host was invalid
+   */
+  static HttpUrl getChecked(URL url) throws MalformedURLException, UnknownHostException {
```
swankjesse: Mind making this method take a `String`? Then the caller can call `url.toString()`, which potentially makes this slightly more applicable.

swankjesse:
LGTM! one small suggestion.

nfuller:
Made the change: URL is now String.

swankjesse:
Nice!

PR #1829, opened 2015-08-29
Change ambiguous wording for timeouts

mgp:
The previous wording made it sound like any connection that was ongoing
for more than 10 seconds (e.g. a download of a large file) would be
disconnected.


swankjesse:
Yes, much better!

PR #1835, opened 2015-09-02
Test code refactoring

nfuller:
From comments in a different PR.

Modify the various Delegating*SocketFactory implementations
to return a *Socket from configureSocket() rather than assume
configuration just modifies the object supplied. Improves
flexibility and reduces the need to override multiple methods
in some places.


swankjesse:
Thanks!

PR #1845, opened 2015-09-09
Fix encoding set for query parameters

iagreen:
Added additional encoding characters that cause URI to throw an `URISyntaxException` when used in a query parameter.

fixes this issue (reported on retrofit, but belongs on okhttp) https://github.com/square/retrofit/issues/1058


swankjesse:
I'm a bit embarrassed that no tests needed to change here! I believe the behavior here changes what's on the wire, and is different from what web browsers do. I'm tempted to fix the encoding _only_ on conversion to URI. Does that work for you? We'd make the URI conversion method smarter.

iagreen:
Agreed, it does change what is on the wire.  A quick survey of web browsers shows all the characters I added are put on the wire as is, except for Firefox, which encodes the backtick. 

Were you thinking something like the above for a smarter URI conversion?

swankjesse:
This is perfect.

PR #1847, opened 2015-09-10
Avoid extra MediaType.parse() calls by adding charset to MediaType in FormEncodingBuilder

JasperG:
Add charset to MediaType in order to avoid it being added after calling .build()

RequestBody.create() always checks and adds utf-8 charset if undefined in MediaType causing an extra MediaType.parse() call for each use of FormEncodingBuilder


swankjesse:
Are you sure? Note that the overload we're calling takes a `ByteString`.

JasperG:
You're absolutely right; Sorry about that.

PR #1848, opened 2015-09-11
use application/x-www-form-urlencoded in okcurl

yschimke:
This is the correct default mimetype for POST requests

see [POST#Use_for_submitting_web_forms](https://en.wikipedia.org/wiki/POST_(HTTP)#Use_for_submitting_web_forms)


swankjesse:
Thank you!

PR #1849, opened 2015-09-13
Consider digits in HttpUrl scheme as valid

felipecsl:
Even though `HttpUrl` only supports http and https URLs, it should still consider a URI like `image640://480.png` as valid, even though it's unsupported. This changes `schemeDelimiterOffset()` to correctly account for digits in the scheme.

Specs: https://tools.ietf.org/html/rfc3986#page-17


swankjesse:
thanks!

Issue #1833, opened 2015-09-02
HttpURLConnection: Behavior change around non-LDH ASCII characters in host names


nfuller:
Creating to have a discussion. If I create a URL containing ASCII codes besides letters, digits and hyphens in the authority/host I get a failure. e.g.

    String host = "host\u0001";
    HttpUrl expected = HttpUrl.parse("http://" + host + "/");
    assertEquals(host, expected.host());

The ASCII code 0x01 in the host is not transformed (underneath, because IDN.toAscii(host) is used). 

After the switch to HttpUrl internally, this can leads to problems with the host header with HttpURLConnectionImpl. The host header checking is currently more strict than the host parsing.

In HttpEngine this code fails with an IllegalArgumentException here:

    if (request.header("Host") == null) {
      result.header("Host", Util.hostHeader(request.httpUrl()));
    }

Previous behavior was to encode the host: e.g. "host\u0001" -> "host%01"

Ideally from a "least likely to break apps" PoV we would keep the old behavior.
The new behavior seems likely to have an impact because it involves throwing a runtime exception.
I could see us throwing an UnknownHostException instead but I'd need to check the implications. (For automatic following of redirects I can see us refusing to follow the redirect if the Location is found to be incorrect).

https://url.spec.whatwg.org/#host-parsing suggests the OkHttp implementation is correct, but https://url.spec.whatwg.org/#host-syntax suggests codes like 0x01 are not valid.

The whatwg spec also suggests that an encoded value (like %01) would break the host-parsing rules, but probably not the host-syntax.

Any suggestions?

swankjesse:
Right now we have this inside `canonicalizeHost`:

      // Confirm that the decoded result doesn't contain any illegal characters.
      int length = idnDecoded.length();
      if (delimiterOffset(idnDecoded, 0, length, "\u0000\t\n\r #%/:?@[\\]") != length) {
        return null;
      }

I see no harm in strengthening this to forbid all control characters and 0x7f. Would that solve the problem?

nfuller:
I started down that road yesterday but backed off when I saw the list of existing chars came from the WHATWG spec.
I shall resurrect the change and see if there are any problems. I think it will result in the UnknownHostException / not following redirect behavior mentioned above, which constitutes a behavior change but not one that I can (currently) see causing too many problems.

swankjesse:
Yeah, I think it's better behavior.

swankjesse:
@nfuller any further action worth taking here?

[...]

PR #1851, opened 2015-09-14
Throw UnknownHostException not IllegalArgumentException for bad host …

nfuller:
…chars

Before this change HttpURLConnection would parse the bad ASCII
characters like \u0001, but fail when setting them in the host
header.

After this change HttpURLConnection throws UnknownHostException.
Earlier versions of OkHttp would escape the bad characters, but
this behavior has not been retained.

Fix for issue #1833.

This change involves a public API change to Headers (addHeaderLenient)
to enable building of MockResponse objects containing response headers
that OkHttp considers invalid, and a behavior change for HttpUrl, and
consequently HttpURLConnectionImpl.

### DIFF
--- a/okhttp-tests/src/test/java/com/squareup/okhttp/URLConnectionTest.java
+++ b/okhttp-tests/src/test/java/com/squareup/okhttp/URLConnectionTest.java
@@ -3212,6 +3212,25 @@ public final class URLConnectionTest {
     }
   }
 
+  @Test public void urlRedirectToHostWithNul() throws Exception {
+    String redirectUrl = "http://host\u0000/";
+    server.enqueue(new MockResponse().setResponseCode(302)
+        .addHeaderLenient("Location", redirectUrl));
+
+    HttpURLConnection urlConnection = client.open(server.getUrl("/"));
+    assertEquals(302, urlConnection.getResponseCode());
+    assertEquals(redirectUrl, urlConnection.getHeaderField("Location"));
+  }
+
+  @Test public void urlWithBadAsciiHost() throws Exception {
+    URLConnection urlConnection = client.open(new URL("http://host\u0001/"));
+    try {
+      urlConnection.getInputStream();
+      fail();
+    } catch (UnknownHostException expected) {
+    }
+  }
+
   /** Returns a gzipped copy of {@code bytes}. */
   public Buffer gzip(String data) throws IOException {
     Buffer result = new Buffer();
--- a/okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
@@ -1238,17 +1238,7 @@ public final class HttpUrl {
         throw new AssertionError();
       }
 
-      // Do IDN decoding. This converts {@code ☃.net} to {@code xn--n3h.net}.
-      String idnDecoded = domainToAscii(percentDecoded);
-      if (idnDecoded == null) return null;
-
-      // Confirm that the decoded result doesn't contain any illegal characters.
-      int length = idnDecoded.length();
-      if (delimiterOffset(idnDecoded, 0, length, "\u0000\t\n\r #%/:?@[\\]") != length) {
-        return null;
-      }
-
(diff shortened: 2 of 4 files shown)


okhttp/src/main/java/com/squareup/okhttp/Headers.java
```diff
@@ -250,7 +250,8 @@ public Builder add(String name, String value) {
      * Add a field with the specified value without any validation. Only
      * appropriate for headers from the remote peer or cache.
      */
-    Builder addLenient(String name, String value) {
+    // public for MockWebServer to allow simulation of bad server responses.
+    public Builder addLenient(String name, String value) {
```
JakeWharton: You can go through `Internal.addLenient` to avoid having to put this in the public API.

nfuller: Thanks. Fixed.

okhttp-tests/src/test/java/com/squareup/okhttp/URLConnectionTest.java
```diff
@@ -3212,6 +3212,25 @@ private void zeroLengthPayload(String method)
...
+    String redirectUrl = "http://host\u0000/";
+    server.enqueue(new MockResponse().setResponseCode(302)
+        .addHeaderLenient("Location", redirectUrl));
+
+    HttpURLConnection urlConnection = client.open(server.getUrl("/"));
+    assertEquals(302, urlConnection.getResponseCode());
+    assertEquals(redirectUrl, urlConnection.getHeaderField("Location"));
+  }
+
+  @Test public void urlHostWithNonLdhAscii() throws Exception {
```
swankjesse: Ldh?

okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
```diff
@@ -1243,14 +1243,32 @@ private static String canonicalizeHost(String input, int pos, int limit) {
...
-      if (delimiterOffset(idnDecoded, 0, length, "\u0000\t\n\r #%/:?@[\\]") != length) {
+      if (containsInvalidHostnameCharacters(idnDecoded)) {
         return null;
       }
 
       return idnDecoded;
     }
 
+    private static boolean containsInvalidHostnameCharacters(String idnDecoded) {
+      for (int i = 0; i < idnDecoded.length(); i++) {
```
swankjesse: wanna use a codepoint loop instead of a char loop? There's a few other examples in this file.

swankjesse: Should this method use a whitelist rather than a blacklist?

okhttp-tests/src/test/java/com/squareup/okhttp/URLConnectionTest.java
nfuller: letter, decimal, hyphen. I'd not seen it before, but it is thrown around in IDN errors and unicode docs (e.g. https://tools.ietf.org/html/rfc5890) so I assumed I was just being ignorant. Happy to rename.

okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
swankjesse: (probably not right now, but who has a hostname of `~` ?)

swankjesse:
LGTM

okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
nfuller: Seems unnecessary to switch to unicode codepoints: the input is guaranteed to be ASCII at this point. I've rearranged the code to try to make that clearer.

FYI - delimiterOffset, the call I replaced, still uses char-by-char comparison. I didn't look to see if that's valid.

Happy to change it if you feel passionate about it.

Regarding blacklist Vs whitelist: could do. I was trying to keep it as close as possible to the WHATWG spec.

swankjesse:
Works for me!

Issue #1824, opened 2015-08-27
MockWebServer#disconnectHalfway() test fails when URLConnection used in tests is OkHttp


nfuller:
On Android the disconnectHalfway() test fails due to a ProtocolException.

    java.net.ProtocolException: unexpected end of stream
    	at com.android.okhttp.internal.http.HttpConnection$FixedLengthSource.read(HttpConnection.java:421)
    	at com.android.okhttp.okio.RealBufferedSource$1.read(RealBufferedSource.java:338)
    	at com.squareup.okhttp.mockwebserver.MockWebServerTest.disconnectHalfway(MockWebServerTest.java:267)

When run against OpenJDK 7 it passes.
The difference is because Android uses OkHttp to supply the URLConnection implementation, and OkHttp treats this as a ProtocolException rather than a simple "end of stream". I believe the ProtocolException has been the behavior for a while and seems reasonable if the server has promised more bytes than were received.

If the goal of the test is to confirm the DISCONNECT_DURING_RESPONSE_BODY policy is operating correctly I suspect the answer here is to accept both behaviors.

I can produce the change but want to check you agree before I do.

swankjesse:
Fixed with https://github.com/square/okhttp/pull/1852

PR #1852, opened 2015-09-15
Make the MockWebServer tests more Android compatible.

nfuller:
There are a few cases in OkHttp where API calls don't exist on
Android, but this is one that we can work around with any loss
of understanding.

The same test also behaves differently on Android due to
differences between the OpenJDK and Android's
HttpURLConnection implementation.

See issue #1824.


swankjesse:
Neat!

PR #1855, opened 2015-09-16
Close ResponseBody in sample

NightlyNexus:


samples/simple-client/src/main/java/com/squareup/okhttp/sample/OkHttpContributors.java
```diff
@@ -26,16 +27,19 @@ public static void main(String... args) throws Exception {
     OkHttpClient client = new OkHttpClient();
 
     // Create request for remote resource.
-    Request request = new Request.Builder()
-        .url(ENDPOINT)
+    Request request = new Request.Builder() //
```
swankjesse: please no `//`

samples/simple-client/src/main/java/com/squareup/okhttp/sample/OkHttpContributors.java
```diff
@@ -26,16 +27,19 @@ public static void main(String... args) throws Exception {
...
     // Execute the request and retrieve the response.
     Response response = client.newCall(request).execute();
 
     // Deserialize HTTP response to concrete type.
-    Reader body = response.body().charStream();
-    List<Contributor> contributors = GSON.fromJson(body, CONTRIBUTORS.getType());
+    ResponseBody body = response.body();
+    Reader charStream = body.charStream();
+    List<Contributor> contributors = GSON.fromJson(charStream, CONTRIBUTORS.getType());
+    charStream.close();
```
swankjesse: you don't need both of these `close()` calls

swankjesse: probably `body.close()` is enough

NightlyNexus: I know I sound dumb, but I don't really understand.
Can you ELI5 for my benefit?

swankjesse: There’s no need to call `close()` on the charStream. Closing the body does that for you.

NightlyNexus: Right, but I don't understand how.
ResponseBody closes the BufferedSource, but how does it close the Reader for the char stream?

JakeWharton: The `Reader` is just wrapping the `BufferedSource` and it holds no resources itself.

NightlyNexus: Oh, yep.  Now, I see that, and it makes obvious sense.  Thanks for the help!

swankjesse:
Thanks!

PR #1856, opened 2015-09-16
feat: Need expanding 'QUERY_COMPONENT_ENCODE_SET' values in 'okhttp/H…

KENJU:
Is it possible that `QUERY_ENCODE_SET` or `QUERY_COMPONENT_ENCODE_SET` does not support `| (%7C)` like `FORM_ENCODE_SET`?  

Is it possible that `QUERY_ENCODE_SET` or `QUERY_COMPONENT_ENCODE_SET` does not support `| (%7C)` like `FORM_ENCODE_SET`?  Or would you tell me why because I might missing the basic knowledge of URL Encoding...

Thanks,

```java

  static final String USERNAME_ENCODE_SET = " \"':;<=>@[]^`{}|/\\?#";
  static final String PASSWORD_ENCODE_SET = " \"':;<=>@[]^`{}|/\\?#";
  static final String PATH_SEGMENT_ENCODE_SET = " \"<>^`{}|/\\?#";
  static final String QUERY_ENCODE_SET = " \"'<>#";
  static final String QUERY_COMPONENT_ENCODE_SET = " \"'<>#&=";
  static final String FORM_ENCODE_SET = " \"':;<=>@[]^`{}|/\\?#&!$(),~";
  static final String FRAGMENT_ENCODE_SET = "";

```

### DIFF
--- a/okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
@@ -259,7 +259,7 @@ public final class HttpUrl {
   static final String PASSWORD_ENCODE_SET = " \"':;<=>@[]^`{}|/\\?#";
   static final String PATH_SEGMENT_ENCODE_SET = " \"<>^`{}|/\\?#";
   static final String QUERY_ENCODE_SET = " \"'<>#";
-  static final String QUERY_COMPONENT_ENCODE_SET = " \"'<>#&=";
+  static final String QUERY_COMPONENT_ENCODE_SET = " \"'<>|#&=";
   static final String CONVERT_TO_URI_ENCODE_SET = "^`{}|\\";
   static final String FORM_ENCODE_SET = " \"':;<=>@[]^`{}|/\\?#&!$(),~";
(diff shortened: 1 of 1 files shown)


swankjesse:
This isn't what web browsers do. What's the problem you're seeing?

KENJU:
Actually I had a problem when passing param including '|' at first. My OKHttp version was at first 2.4, and then tried after updating to 2.5.

```java

// below code throws IllegalStateException
builder.addQueryParameter(entry.getKey(), entry.getValue());

```

So, after looking at this [issue #1044](https://github.com/square/okhttp/issues/1044),
I wonder why encoding `|` still throws `IllegalStateException`.

However, after that I found `addEncodedQueryParameter()` method and passed encoded url with `URLEncoder.encode()` like below,
it gets the correct response.

```java

String encodedData = URLEncoder.encode(entry.getValue(), "UTF-8");
builder.addEncodedQueryParameter(entry.getKey(), encodedData);

```

Maybe do I misunderstand how to use okhttp? Sorry for having disturbed you...

swankjesse:
@kenju try the code in master? I think this has been fixed since 2.5.

Issue #1801, opened 2015-08-10
AsyncCall NPE's if an interceptor throws before the call is attempted.


GrahamFirst:
AsyncCall throws an uncaught NullPointerException [here](https://github.com/square/okhttp/blob/master/okhttp/src/main/java/com/squareup/okhttp/Call.java#L175) because engine is null, having never been initialized, if the call was never attempted.  For example, this can happen if an ("application-level") interceptor aborted the call by throwing an IOException before the call was ever attempted.

I suggest that it could simply pass a null request to the callback in this case instead.

ansman:
I'm getting the same problem, here's the stacktrace:
```
java.lang.NullPointerException: Attempt to invoke virtual method 'com.squareup.okhttp.Request com.squareup.okhttp.internal.http.HttpEngine.getRequest()' on a null object reference
    at com.squareup.okhttp.Call$AsyncCall.execute(Call.java:175)
    at com.squareup.okhttp.internal.NamedRunnable.run(NamedRunnable.java:33)
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1112)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:587)
    at java.lang.Thread.run(Thread.java:818)
```

PR #1861, opened 2015-09-18
Add test for AsyncCall failure in application Interceptor

NightlyNexus:
A failing test for https://github.com/square/okhttp/issues/1801
I am not sure this is the right place for the test, though.

### DIFF
--- a/okhttp-tests/src/test/java/com/squareup/okhttp/CallTest.java
+++ b/okhttp-tests/src/test/java/com/squareup/okhttp/CallTest.java
@@ -697,6 +697,17 @@ public final class CallTest {
     }
   }
 
+  @Test public void asyncCallEngineInitialized () {
+    OkHttpClient client = new OkHttpClient();
+    client.interceptors().add(new Interceptor() {
+
+      @Override public Response intercept(Chain chain) throws IOException {
+        throw new IOException();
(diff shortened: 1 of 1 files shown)


okhttp-tests/src/test/java/com/squareup/okhttp/DispatcherTest.java
```diff
@@ -24,6 +25,17 @@
     dispatcher.setMaxRequestsPerHost(10);
   }
 
+  @Test public void asyncCallEngineInitialized () {
```
swankjesse: nit: space before `()`

okhttp-tests/src/test/java/com/squareup/okhttp/DispatcherTest.java
```diff
@@ -24,6 +25,17 @@
...
+    final OkHttpClient client = new OkHttpClient();
```
swankjesse: nit: we don't `final` on locals unless absolutely necessary

swankjesse:
Probably `CallTest` is the best place for it. It's where all the tests go.

NightlyNexus:
Closing, see: https://github.com/square/okhttp/pull/1870

PR #1863, opened 2015-09-18
Allow interceptor to change method

cketti:
Changing the request method from e.g. `GET` to `POST` in a network interceptor lead to the body not being written to the transport. That's because `HttpEngine#permitsRequestBody()` was using the original request to determine whether sending a request body was allowed.


swankjesse:
Nice!

PR #1870, opened 2015-09-24
Make AsyncCall handle interceptors that throws an IOException

### LINKED ISSUE #1801: AsyncCall NPE's if an interceptor throws before the call is attempted.

ansman:
This closes #1801

### DIFF
--- a/okhttp-tests/src/test/java/com/squareup/okhttp/CallTest.java
+++ b/okhttp-tests/src/test/java/com/squareup/okhttp/CallTest.java
@@ -696,6 +696,20 @@ public final class CallTest {
     }
   }
 
+  // https://github.com/square/okhttp/issues/1801
+  @Test public void asyncCallEngineInitialized() throws Exception {
+    OkHttpClient c = new OkHttpClient();
+    c.interceptors().add(new Interceptor() {
+      @Override public Response intercept(Chain chain) throws IOException {
+        throw new IOException();
+      }
+    });
+    Request request = new Request.Builder().url(server.url("/")).build();
+    c.newCall(request).enqueue(callback);
+    RecordedResponse response = callback.await(request.httpUrl());
+    assertEquals(request, response.request);
+  }
+
   @Test public void reusedSinksGetIndependentTimeoutInstances() throws Exception {
     server.enqueue(new MockResponse());
     server.enqueue(new MockResponse());
--- a/okhttp/src/main/java/com/squareup/okhttp/Call.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/Call.java
@@ -172,7 +172,8 @@ public class Call {
           // Do not signal the callback twice!
           logger.log(Level.INFO, "Callback failure for " + toLoggableString(), e);
         } else {
-          responseCallback.onFailure(engine.getRequest(), e);
+          Request request = engine == null ? originalRequest : engine.getRequest();
+          responseCallback.onFailure(request, e);
         }
       } finally {
         client.getDispatcher().finished(this);


okhttp/src/main/java/com/squareup/okhttp/Call.java
```diff
@@ -172,7 +172,8 @@ Call get() {
           // Do not signal the callback twice!
           logger.log(Level.INFO, "Callback failure for " + toLoggableString(), e);
         } else {
-          responseCallback.onFailure(engine.getRequest(), e);
+          Request request = engine == null ? null : engine.getRequest();
+          responseCallback.onFailure(request, e);
```
swankjesse: We've already called `onFailure()` when we get here.

swankjesse: hang on, I'm reading this wrong.

okhttp/src/main/java/com/squareup/okhttp/Call.java
```diff
@@ -172,7 +172,8 @@ Call get() {
           // Do not signal the callback twice!
           logger.log(Level.INFO, "Callback failure for " + toLoggableString(), e);
         } else {
-          responseCallback.onFailure(engine.getRequest(), e);
+          Request request = engine == null ? null : engine.getRequest();
```
swankjesse: use `originalRequest` ?

ansman: Doh, of course. I'll change so it uses original request if the engine is `null`

ansman: Implemented as requested, let me know if I should squash the commits or not.

okhttp-tests/src/test/java/com/squareup/okhttp/CallTest.java
```diff
@@ -697,6 +697,19 @@ private void postBodyRetransmittedAfterAuthorizationFail(String body) throws Exc
     }
   }
 
+  @Test public void asyncCallEngineInitialized () throws Exception {
+    OkHttpClient client = new OkHttpClient();
```
swankjesse: For what it's worth, this shadows another variable called `client`.

okhttp-tests/src/test/java/com/squareup/okhttp/CallTest.java
```diff
@@ -697,6 +697,19 @@ private void postBodyRetransmittedAfterAuthorizationFail(String body) throws Exc
     }
   }
 
+  @Test public void asyncCallEngineInitialized () throws Exception {
```
swankjesse: ultranit: no ` ` before `()`.

okhttp-tests/src/test/java/com/squareup/okhttp/CallTest.java
```diff
@@ -697,6 +697,19 @@ private void postBodyRetransmittedAfterAuthorizationFail(String body) throws Exc
...
+    OkHttpClient client = new OkHttpClient();
+    client.interceptors().add(new Interceptor() {
+      @Override public Response intercept(Chain chain) throws IOException {
+        throw new IOException();
+      }
+    });
+    Request request = new Request.Builder().url(server.url("/")).build();
+    client.newCall(request).enqueue(callback);
+    RecordedResponse response = callback.await(request.httpUrl());
+    assertEquals(request, response.request);
```
swankjesse: Nice test!

swankjesse:
LGTM, just two tiny nitpicks. Sorry for sitting on this; GitHub doesn't email me when you push new code on a PR.

swankjesse:
Also- please http://rebaseandsqua.sh.

swankjesse:
Please complete Square's contributors agreement.
https://docs.google.com/forms/d/13WR8m5uZ2nAkJH41k7GdVBXAAbzDk00vxtEYjd6Imzg/viewform?formkey=dDViT2xzUHAwRkI3X3k5Z0lQM091OGc6MQ&ndplr=1

ansman:
@swankjesse 
 * A comment to reference #1801 was made before the test
 * Extraneous space was removed
 * Client variable shadowing was fixed
 * Rebased and squased

swankjesse:
Sweet. Thank you!

GrahamFirst:
Awesome to see this fix!  Any idea when it will be released?

swankjesse:
@GrahamFirst I'm hoping to finish 2.6 in 2015.

PR #1874, opened 2015-09-25
Add logging interceptor for simple request and response logging.

JakeWharton:
https://github.com/square/retrofit/pull/1114


okhttp-logging-interceptor/README.md
```diff
@@ -0,0 +1,44 @@
+Logging Interceptor
+===================
+
+An [OkHttp interceptor][1] which logs HTTP request and response data.
```
swankjesse: Add a note that `HEADERS` and `BODY` levels shouldn't be used in production? Otherwise secrets are leaked. (`Authorization`, `Cookie`, etc.)

swankjesse:
Brilliant. LGTM.

okhttp-logging-interceptor/README.md
JakeWharton: Added:

**Warning**: The logs generated by this interceptor when using the `HEADERS` or `BODY` levels has
the potential to leak sensitive information such as "Authorization" or "Cookie" headers and the
contents of request and response bodies. This data should only be logged in a controlled way or in
a non-production environment.

swankjesse: Thanks!

PR #1875, opened 2015-09-26
Represent WebSocket payload using RequestBody and ResponseBody.

JakeWharton:
This actually breaks the `WebSocketEcho` sample because echo.websocket.org is non-spec complaint (as far as I can tell). Because all messages are technically streamed now, we send an empty close frame which it omits from the frames it echoes back causing a `ProtocolException` by `WebSocketReader` (correctly, I might add).

### DIFF
--- a/okhttp-ws-tests/src/test/java/com/squareup/okhttp/internal/ws/RealWebSocketTest.java
+++ b/okhttp-ws-tests/src/test/java/com/squareup/okhttp/internal/ws/RealWebSocketTest.java
@@ -15,6 +15,8 @@
  */
 package com.squareup.okhttp.internal.ws;
 
+import com.squareup.okhttp.MediaType;
+import com.squareup.okhttp.RequestBody;
 import com.squareup.okhttp.ws.WebSocketRecorder;
 import java.io.IOException;
 import java.net.ProtocolException;
@@ -30,8 +32,8 @@ import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
 
-import static com.squareup.okhttp.ws.WebSocket.PayloadType.BINARY;
-import static com.squareup.okhttp.ws.WebSocket.PayloadType.TEXT;
+import static com.squareup.okhttp.ws.WebSocket.BINARY;
+import static com.squareup.okhttp.ws.WebSocket.TEXT;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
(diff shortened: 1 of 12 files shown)


JakeWharton:
I have a plan to fix this behavior but there's still a rare chance (1 in `okio.Segment.SIZE`) that it could happen. In my opinion, the spec clearly allows this behavior so they're broken, not us.

okhttp-ws-tests/src/test/java/com/squareup/okhttp/internal/ws/RealWebSocketTest.java
```diff
@@ -80,32 +82,49 @@
...
-    sink.writeUtf8("lo!").flush();
-    sink.close();
+    RequestBody message = new RequestBody() {
+      @Override public MediaType contentType() {
+        return TEXT;
+      }
+
+      @Override public void writeTo(BufferedSink sink) throws IOException {
+        sink.writeUtf8("Hel").flush();
+        client.sendPing(new Buffer().writeUtf8("Pong?"));
```
swankjesse: Crazy

okhttp-ws/src/main/java/com/squareup/okhttp/ws/WebSocket.java
```diff
@@ -15,36 +15,28 @@
...
-    /** UTF8-encoded text data. */
-    TEXT,
-    /** Arbitrary binary data. */
-    BINARY
-  }
+  // Using a "utf-8" charset here is an optimization for use by RequestBody.create().
+  /** A {@link MediaType} indicating that text frames should be used when sending a message. */
+  MediaType TEXT = MediaType.parse("application/vnd.okhttp.websocket+text; charset=utf-8");
+  /** A {@link MediaType} indicating that binary frames should be used when sending a message. */
+  MediaType BINARY = MediaType.parse("application/vnd.okhttp.websocket+binary; charset=utf-8");
```
swankjesse: UTF-8 ?!

swankjesse: Really like `MediaType` here, by the way

swankjesse:
LGTM!

okhttp-ws-tests/src/test/java/com/squareup/okhttp/internal/ws/RealWebSocketTest.java
swankjesse: (in a good way!)

okhttp-ws/src/main/java/com/squareup/okhttp/ws/WebSocket.java
JakeWharton: `RequestBody.create(BINARY, "Hello")` will append a UTF-8 charset and prevent `equals` or `==` from working without.

swankjesse: How do you feel about this?
```
RequestBody.create(BINARY, ByteString.encodeUtf8("Hello"))
```

JakeWharton: That's what I do in the tests, my concern was for naive consumers.

I'm realizing that it's actually a non-issue since we currently default to binary frames unless the media type matches `TEXT` exactly. So even if you do the wrong thing, you'll still get binary. We currently don't throw on invalid media type usage, but it's something we could do, and it would make using a string with `BINARY` something that has the potential to trip up people.

swankjesse: Yeah, the code that adds the charset when it applies one can be surprising. If we do ever need to gracefully tolerate this, the easiest way is to compare `type()` and `subtype()` instead of the entire `MediaType` object. (Though that's cumbersome I admit).

JakeWharton: I'm going to change it to be strict on `type()` and `subtype()` but lenient on the rest. This will enforce that you used `TEXT` or `BINARY` but still allow the `RequestBody.create(BINARY, "hi!");` case.

swankjesse: Works for me.

PR #1876, opened 2015-09-27
Introduce intermediate storage for frame contents.

JakeWharton:
This eliminates the need to always send a frame for every `FrameSink#write` and to always send an empty frame for `FrameSink#close`. Now, we only emit bytes once Okio reports complete segments.

### DIFF
--- a/okhttp-ws-tests/src/test/java/com/squareup/okhttp/internal/ws/RealWebSocketTest.java
+++ b/okhttp-ws-tests/src/test/java/com/squareup/okhttp/internal/ws/RealWebSocketTest.java
@@ -247,13 +247,6 @@ public final class RealWebSocketTest {
         clientListener.assertClose(1000, "Hello!");
 
         try {
-          sink.writeUtf8("lo!").emit(); // No writing to the underlying sink.
-          fail();
-        } catch (IOException e) {
-          assertEquals("closed", e.getMessage());
-          sink.buffer().clear();
-        }
-        try {
           sink.flush(); // No flushing.
           fail();
         } catch (IOException e) {
@@ -320,7 +313,7 @@ public final class RealWebSocketTest {
     clientListener.assertClose(1000, "Bye!");
   }
 
-  @Test public void protocolErrorBeforeCloseSendsClose() {
+  @Test public void protocolErrorBeforeCloseSendsClose() throws IOException {
     server2client.write(ByteString.decodeHex("0a00")); // Invalid non-final ping frame.
 
     client.readMessage(); // Detects error, send close.
--- a/okhttp-ws-tests/src/test/java/com/squareup/okhttp/internal/ws/WebSocketWriterTest.java
+++ b/okhttp-ws-tests/src/test/java/com/squareup/okhttp/internal/ws/WebSocketWriterTest.java
@@ -31,6 +31,8 @@ import org.junit.runners.model.Statement;
 
 import static com.squareup.okhttp.internal.ws.WebSocketProtocol.OPCODE_BINARY;
 import static com.squareup.okhttp.internal.ws.WebSocketProtocol.OPCODE_TEXT;
+import static com.squareup.okhttp.internal.ws.WebSocketProtocol.PAYLOAD_BYTE_MAX;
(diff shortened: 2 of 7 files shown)


okhttp-ws-tests/src/test/java/com/squareup/okhttp/internal/ws/RealWebSocketTest.java
```diff
@@ -247,13 +247,6 @@
         clientListener.assertClose(1000, "Hello!");
 
         try {
-          sink.writeUtf8("lo!").emit(); // No writing to the underlying sink.
```
JakeWharton: Message writes are single-threaded by policy and thus don't take locks to check whether the larger writer is closed. If this write was larger than a segment it would fail, however.

okhttp-ws-tests/src/test/java/com/squareup/okhttp/internal/ws/WebSocketWriterTest.java
```diff
@@ -71,15 +73,30 @@
...
+  @Test public void closeFlushes() throws IOException {
     BufferedSink sink = Okio.buffer(serverWriter.newMessageSink(OPCODE_TEXT));
 
     sink.writeUtf8("Hel").flush();
     assertData("010348656c");
 
     sink.writeUtf8("lo").close();
-    assertData("00026c6f");
-    assertData("8000");
+    assertData("80026c6f");
```
JakeWharton: No more empty close frames!

okhttp-ws/src/main/java/com/squareup/okhttp/internal/ws/WebSocketWriter.java
```diff
@@ -156,54 +160,53 @@ public Sink newMessageSink(int formatOpcode) {
...
-        writeAllMasked(source, byteCount);
-      } else {
-        sink.write(source, byteCount);
-      }
+    int b1 = 0;
+    if (isClient) {
+      b1 |= B1_FLAG_MASK;
+      random.nextBytes(maskKey);
+    }
+    if (byteCount <= PAYLOAD_MAX) {
```
swankjesse: would it make sense to rename this internally to `PAYLOAD_BYTE_MAX` ? Weird that `SHORT` is the 2nd biggest size

okhttp-ws/src/main/java/com/squareup/okhttp/internal/ws/WebSocketProtocol.java
```diff
@@ -71,11 +71,13 @@
    * Maximum length of frame payload. Larger payloads, if supported, can use the special values
```
swankjesse: This says _if supported_ . . . below it looks like we're assuming these will always be supported?

okhttp-ws/src/main/java/com/squareup/okhttp/internal/ws/WebSocketWriter.java
JakeWharton: Yes it would.

okhttp-ws/src/main/java/com/squareup/okhttp/internal/ws/WebSocketWriter.java
```diff
@@ -156,54 +160,53 @@ public Sink newMessageSink(int formatOpcode) {
...
     frameSink.isFirstFrame = true;
+    frameSink.closed = false;
+
     return frameSink;
   }
 
-  private void writeFrame(int formatOpcode, Buffer source, long byteCount,
-      boolean isFirstFrame, boolean isFinal) throws IOException {
-    if (closed) throw new IOException("closed");
+  private void writeMessageFrameSynchronized(int formatOpcode, long byteCount, boolean isFirstFrame,
```
swankjesse:
```
    assert Thread.holdsLock(this);
```
?

swankjesse:
LGTM

I like this a lot better

okhttp-ws/src/main/java/com/squareup/okhttp/internal/ws/WebSocketProtocol.java
JakeWharton: Added a clarification. Some frame types (control frames) do not support larger payloads.

PR #1877, opened 2015-09-27
Honor (and thus fail on) reserved close codes defined in the spec.

JakeWharton:


JakeWharton:
This brings us to only one failure and 8 non-strict conformance.

okhttp-ws-tests/src/test/java/com/squareup/okhttp/internal/ws/WebSocketReaderTest.java
```diff
@@ -351,6 +363,44 @@
...
+    } catch (ProtocolException e) {
+      assertEquals("Code must be in range [1000,5000): 5000", e.getMessage());
+    }
+  }
+
+  @Test public void closeReservedSetThrows() throws IOException {
+    data.write(ByteString.decodeHex("880203ec")); // Close with code 1004
+    data.write(ByteString.decodeHex("880203ed")); // Close with code 1005
+    data.write(ByteString.decodeHex("880203ee")); // Close with code 1006
+    for (int i = 1012; i <= 2999; i++) {
```
swankjesse: Well then!

Issue #1872, opened 2015-09-25
HttpUrl.uri() not encoding enough to create valid URI instances.


JakeWharton:
```java
HttpUrl url = new Builder() //
    .scheme("http")
    .host("example.com")
    .addPathSegment("data=[out:json];node[\"name\"~\"Karlsruhe\"][\"place\"~\"city|village|town\"];out body;")
    .build();
System.out.println(url);
System.out.println(url.uri());
```
```
http://example.com/data=[out:json];node[%22name%22~%22Karlsruhe%22][%22place%22~%22city%7Cvillage%7Ctown%22];out%20body;
Exception in thread "main" java.lang.IllegalStateException: not valid as a java.net.URI: http://example.com/data=[out:json];node[%22name%22~%22Karlsruhe%22][%22place%22~%22city%7Cvillage%7Ctown%22];out%20body;
```

JakeWharton:
I tried looking into this but gave up. At the very least, adding `[]` to the encode set for URI fixed the issue, but I'm not sure that it's the correct approach to re-encode the entire URL using a global encode set as URI has different encoding sets that it's verifying against for the different parts of the URL. Plus I'm worried (but did not confirm) that this wouldn't play well with IPv6.

We might need a more sophisticated strategy of encoding piecewise before handing to URI either through its `create` method or perhaps one of its constructors.

markrietveld:
This is similar to https://github.com/square/okhttp/issues/1836. According to [this](https://github.com/square/okhttp/issues/1061), doing cookies in OkHttp removes the last dependent on `java.net.URI`, which would avoid this issue

iagreen:
There is a test case matching your case above that is expecting failure on '[' in paths (https://github.com/square/okhttp/blob/master/okhttp-tests/src/test/java/com/squareup/okhttp/HttpUrlTest.java#L933).  The other PR was directed at queries, not paths, but it does seem sketchy to do a global re-encode.  [URI's multi-parameter](http://docs.oracle.com/javase/7/docs/api/java/net/URI.html#URI%28java.lang.String,%20java.lang.String,%20java.lang.String,%20int,%20java.lang.String,%20java.lang.String,%20java.lang.String%29) constructor does piecewise encoding.

whyte624:
When PR will be accepted and released? Thanks!

PR #1885, opened 2015-09-29
use URI constructor for encoding

iagreen:
fixes - https://github.com/square/okhttp/issues/1872
1) Use the [URI's multi-parameter](http://docs.oracle.com/javase/7/docs/api/java/net/URI.html#URI%28java.lang.String,%20java.lang.String,%20java.lang.String,%20int,%20java.lang.String,%20java.lang.String,%20java.lang.String%29) constructor does piecewise encoding. 
2) Removed test (toUriForbiddenCharacter) that expects '[' to be invalid in the path segment since it will be encoded now.
3) Updated the `invalidHost` test to expect a different exception.  It appears that URI errors earlier in the process (URI creation) using this constructor.
4) reverts previous fix (https://github.com/square/okhttp/pull/1845)


swankjesse:
Thank you. And sorry for the delay. I might rework this further, but this is a much better starting place than where we are now.

PR #1894, opened 2015-10-02
adding ability to get query & post params from the request

marenzo:
as some clients passing parameters via query uri, or just in the body (as form-urlencoded request),
there should be option to get the value from the request.
I've added this ability to RecordedRequest


mockwebserver/src/main/java/com/squareup/okhttp/mockwebserver/RecordedRequest.java
```diff
@@ -121,7 +138,111 @@ public TlsVersion getTlsVersion() {
...
+   * Returns all query params
+   */
+  public Map<String, List<String>> getQueryParams() {
+    return queryParams;
+  }
+
+  /**
+   * Returns single value for query param
+   */
+  public String getQueryParam(String param) {
```
swankjesse: For these I think we want to just expose the request URL as an `HttpUrl`

mockwebserver/src/main/java/com/squareup/okhttp/mockwebserver/RecordedRequest.java
```diff
@@ -121,7 +138,111 @@ public TlsVersion getTlsVersion() {
...
+    return query_pairs;
+  }
+
+  /**
+   * split valid query string
+   */
+  private Map<String, List<String>> parseParams(String body) {
+    Map<String, List<String>> queryParams;
+    try {
+      queryParams = Collections.unmodifiableMap(splitQuery(body));
```
swankjesse: nit: this is an unmodifiable map, but the lists in it are modifiable

swankjesse:
I'm a bit reluctant to support POST so intimately in RecordedRequest. I think it'd be better to expose a mechanism to get the body as bytes, and a separate API to decode that.

PR #1936, opened 2015-10-19
Remove the need for an explicit Content-Length header to log response body.

### LINKED ISSUE #1935: HttpLoggingInterceptor does not log response body when Transfer-Encoding is set to chunked
OkHttp 2.6.0-SNAPSHOT

[`HttpLoggingInterceptor`](https://github.com/square/okhttp/blob/master/okhttp-logging-interceptor/src/main/java/com/squareup/okhttp/logging/HttpLoggingInterceptor.java) is not logging the response body of a `POST` request where the response is sent using `Transfer-Encoding: chunked` instead of setting a `Content-Length`.

Stepping through the code I see that [`responseBody.contentLength()`](https://github.com/square/okhttp/blob/master/okhttp-logging-interceptor/src/main/java/com/squareup/okhttp/logging/HttpLoggingInterceptor.java#L207) `== -1` (which it eventually grabs from `RealResponseBody` using the `Content-Length` header), even though evaluating [`buffer.clone().readString(charset)`](https://github.com/square/okhttp/blob/master/okhttp-logging-interceptor/src/main/java/com/squareup/okhttp/logging/HttpLoggingInterceptor.java#L209) still returns the full response body.

I'm not sure why the server is chunking a json response, but should the logger still work as intended in this case?

JakeWharton:
This allows chunked responses with no specified length to be logged and have their size reported correctly.

Closes #1935.

### DIFF
--- a/okhttp-logging-interceptor/src/main/java/com/squareup/okhttp/logging/HttpLoggingInterceptor.java
+++ b/okhttp-logging-interceptor/src/main/java/com/squareup/okhttp/logging/HttpLoggingInterceptor.java
@@ -157,6 +157,7 @@ public final class HttpLoggingInterceptor implements Interceptor {
         logger.log(headers.name(i) + ": " + headers.value(i));
       }
 
+      String endMessage = "--> END " + request.method();
       if (logBody && hasRequestBody) {
         Buffer buffer = new Buffer();
         requestBody.writeTo(buffer);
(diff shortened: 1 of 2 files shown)


okhttp-logging-interceptor/src/main/java/com/squareup/okhttp/logging/HttpLoggingInterceptor.java
```diff
@@ -193,10 +191,11 @@ public void setLevel(Level level) {
         logger.log(headers.name(i) + ": " + headers.value(i));
       }
 
+      String endMessage = "<-- END HTTP";
       if (logBody) {
         BufferedSource source = responseBody.source();
         source.request(Long.MAX_VALUE); // Buffer the entire body.
-        Buffer buffer = source.buffer().clone();
+        Buffer buffer = source.buffer();
```
JakeWharton: Sneaking this in this PR. We already were defensively cloning for the read.

PR #1942, opened 2015-10-23
Change encoding/decoding rules for +.

ErikHellman:
- Don't do percent encoding of + in query string.
- Always decode + to a ' ' (space).

### DIFF
--- a/okhttp-tests/src/test/java/com/squareup/okhttp/HttpUrlTest.java
+++ b/okhttp-tests/src/test/java/com/squareup/okhttp/HttpUrlTest.java
@@ -28,9 +28,7 @@ import org.junit.Ignore;
 import org.junit.Test;
 
 import static java.util.Collections.singletonList;
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNull;
-import static org.junit.Assert.fail;
+import static org.junit.Assert.*;
 
 public final class HttpUrlTest {
   @Test public void parseTrimsAsciiWhitespace() throws Exception {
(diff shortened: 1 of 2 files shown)


okhttp-tests/src/test/java/com/squareup/okhttp/HttpUrlTest.java
```diff
@@ -1005,8 +1012,8 @@
   @Test public void composeQueryWithEncodedComponents() throws Exception {
     HttpUrl base = HttpUrl.parse("http://host/");
     HttpUrl url = base.newBuilder().addEncodedQueryParameter("a+=& b", "c+=& d").build();
-    assertEquals("http://host/?a%20%3D%26%20b=c%20%3D%26%20d", url.toString());
-    assertEquals("c =& d", url.queryParameter("a =& b"));
+    assertEquals("http://host/?a+%3D%26%20b=c+%3D%26%20d", url.toString());
+    assertEquals("c+=& d", url.queryParameter("a+=& b"));
```
swankjesse: queryParameter takes the _unencoded_ value, and you're querying with the encoded value. This is incorrect.

swankjesse: Doesn't this look like a regression that'll break other users?

okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
```diff
@@ -1472,7 +1472,12 @@ static String percentDecode(String encoded, int pos, int limit) {
     }
 
     // Fast path: no characters in [pos..limit) required decoding.
-    return encoded.substring(pos, limit);
+    String result = encoded.substring(pos, limit);
+    // + become ' ' for queries
+    if(isQuery) {
+      result = result.replace('+', ' ');
```
swankjesse: yuck. I don't like that we have to loop through the characters of the string a 3rd time. Can't `percentDecode` do this job too?

swankjesse:
I don't think this is going to work.

JakeWharton:
Obviated by #1995.

Issue #1736, opened 2015-06-29
No retry/failover on Android with setConnectTimeout() set


matejsnoha:
When using OkHttpClient with setConnectTimeout() set to 10 seconds on Android, I get a SocketTimeoutException after 10 seconds if it picks an unreachable IP (another IP is reachable). 

When I do not call setConnectTimeout(), the request runs for a few minutes (seems to be different on different Android versions) while waiting for the unreachable IP and then correctly fails over to the reachable IP and returns response.

On desktop the behaviour is correct, the request always returns response from the one available IP and in a very short time.

My setup:
DNS name with 2 IPv4 A addresses, 1 is unreachable.
OkHttp 2.4.0
Android 4.1.1 and 5.0

    java.net.SocketTimeoutException: failed to connect to host/unreachable-ip (port 443) after 10000ms
    at libcore.io.IoBridge.connectErrno(IoBridge.java:159)
    at libcore.io.IoBridge.connect(IoBridge.java:112)
    at java.net.PlainSocketImpl.connect(PlainSocketImpl.java:192)
    at java.net.PlainSocketImpl.connect(PlainSocketImpl.java:459)
    at java.net.Socket.connect(Socket.java:842)
    at com.squareup.okhttp.internal.Platform$Android.connectSocket(Platform.java:190)
    at com.squareup.okhttp.internal.http.SocketConnector.connectRawSocket(SocketConnector.java:160)
    at com.squareup.okhttp.internal.http.SocketConnector.connectTls(SocketConnector.java:79)
    at com.squareup.okhttp.Connection.connect(Connection.java:143)
    at com.squareup.okhttp.Connection.connectAndSetOwner(Connection.java:185)
    at com.squareup.okhttp.OkHttpClient$1.connectAndSetOwner(OkHttpClient.java:128)
    at com.squareup.okhttp.internal.http.HttpEngine.nextConnection(HttpEngine.java:341)
    at com.squareup.okhttp.internal.http.HttpEngine.connect(HttpEngine.java:330)
    at com.squareup.okhttp.internal.http.HttpEngine.sendRequest(HttpEngine.java:248)
    at com.squareup.okhttp.Call.getResponse(Call.java:273)
    at com.squareup.okhttp.Call$ApplicationInterceptorChain.proceed(Call.java:230)
    at com.squareup.okhttp.Call.getResponseWithInterceptorChain(Call.java:201)
    at com.squareup.okhttp.Call.execute(Call.java:81)

[...]

PR #1961, opened 2015-11-02
Prefer IPv4 over IPv6

thest1:
Prefer IPv4 over IPv6 because IPv4 has higher chances to be available. (see #1736)

### DIFF
--- a/okhttp/src/main/java/com/squareup/okhttp/internal/Network.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/internal/Network.java
@@ -17,6 +17,7 @@ package com.squareup.okhttp.internal;
 
 import java.net.InetAddress;
 import java.net.Inet4Address;
+import java.util.ArrayList;
 import java.net.UnknownHostException;
 
 /**
--- a/okhttp/src/main/java/com/squareup/okhttp/internal/Network.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/internal/Network.java
@@ -26,7 +26,19 @@ public interface Network {
   Network DEFAULT = new Network() {
     @Override public InetAddress[] resolveInetAddresses(String host) throws UnknownHostException {
       if (host == null) throw new UnknownHostException("host == null");
-      return InetAddress.getAllByName(host);
+      InetAddress[] ips_arr = InetAddress.getAllByName(host);
+      //Prefer IPv4 over IPv6 because IPv4 has higher chances to be available
+      //That's important because we'll probably have to wait for timeout before trying the second IP
+      ArrayList<InetAddress> ips_all = new ArrayList<InetAddress>(ips_arr.length);
+      ArrayList<InetAddress> ips_6 = new ArrayList<InetAddress>(ips_arr.length);
+      for(InetAddress ip : ips_arr) {
+          if (ip instanceof Inet4Address)
+              ips_all.add(ip);
+          else
+              ips_6.add(ip);
+      }
+      ips_all.addAll(ips_6);
+      return ips_all.toArray(ips_arr);
     }
   };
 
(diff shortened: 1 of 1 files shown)


swankjesse:
This seems to contradict recent documentation claiming that IPv6 is faster.
https://code.facebook.com/posts/1192894270727351/ipv6-it-s-time-to-get-on-board-/

thest1:
Thank you Jesse. Makes sense. So I'll have to make it just an option for those who have IPv6 problems.

swankjesse:
@thest the other issue is that if you're suffering IPv6 issues, you should fix that in the network or in the OS, not in the application. It's a bad idea for every application to paper over lower-level bugs!

thest1:
@swankjesse That's right. But most users can't handle that on their own. It's all about home wifi routers, user has no one to help there. They can see that all browsers work fine (due to Happy Eyeballs). So they can't imagine there's something wrong about their network. They're pretty sure my application is the one that's broken. Pretty sad. Anyway thank you for your feedback.

warpspin:
Re-ordering the IP addresses by using Internal.instance.setNetwork to set a custom Network interface implementation is also what we have been doing for some time now, till Happy Eyeballs hopefully arrives one day in OkHttp. IPv4 in fact IS slower, depending on the ISP, especially on mobile, probably due to carrier grade NAT, but this is still preferable for now to 10 second delays per request on broken configurations. Our application still suffers from a bunch of 1 star reviews from users, usually with broken home networks.

swankjesse:
Suppose we make the DNS lookup pluggable in 2.6. Then you don't need an internal API to get this behavior?

warpspin:
At the moment, we're overriding resolveInetAddresses with a mechanism similar to @thest1's - in addition to re-ordering IP addresses for OkHttp, we also have hardcoded fallbacks for some host names, in case the resolving is completely bugged, so well, unless your pluggable DNS lookup is as powerful as the current Network-based mechanism, I'd probably still use internal APIs to achieve the result. If, by "pluggable DNS lookup" you basically mean promoting the current Network mechanism or something equally powerful to an official feature outside the internal.* package, I'm all for it, though.

```java
public class DnsClient implements Network {
    DnsClient() {}

    @Override
    public InetAddress[] resolveInetAddresses(final String host) throws UnknownHostException {
        return resolve(host, false );
    }

    private static InetAddress[] resolve(final String host, boolean preferIPv6) throws UnknownHostException {
        if (host == null) throw new UnknownHostException("host == null");
        InetAddress[] addresses;
        try {
            addresses = InetAddress.getAllByName(host);
        } catch( UnknownHostException e ) {
            addresses = fallbackResolver(host);
            if( addresses == null || addresses.length == 0 ) throw e;
        }
        if( addresses.length == 1 && host.endsWith(".spin.de") &&
                (addresses[0].isSiteLocalAddress() || addresses[0].isLinkLocalAddress())) {
            InetAddress[] tmpAddresses = fallbackResolver(host);
            if( tmpAddresses != null && tmpAddresses.length > 0 ) return tmpAddresses;
        }
        InetAddress[] result = new InetAddress[addresses.length];
        int i = 0;
        for( InetAddress addr : addresses ) {
            if( preferIPv6 ^ addr instanceof Inet4Address) {
                result[i++] = addr;
            }
        }
        for( InetAddress addr : addresses ) {
            if( !(preferIPv6 ^ addr instanceof Inet4Address)) {
                result[i++] = addr;
            }
        }
        return result;
    }

    // some more stuff, not only for okhttp...
}
```

swankjesse:
@warpspin yep, this is about promoting the Network classes.

thest1:
@warpspin Yeah, right now I do the same with Internal.instance.setNetwork(). Thanks for sharing your experience. Soon I'll see how it works for me in production.

swankjesse:
As promised . . . https://github.com/square/okhttp/pull/1967

swankjesse:
Closing this, I'm hoping the pluggable DNS API is sufficient!

PR #1967, opened 2015-11-04
Make the DNS service pluggable.

swankjesse:
It was already pluggable as the internal class named Network. Now it's
exposed as a public interface called DNS.


okhttp/src/main/java/com/squareup/okhttp/Address.java
```diff
@@ -30,44 +30,57 @@
  * this is the server's hostname and port. If an explicit proxy is requested (or
  * {@linkplain Proxy#NO_PROXY no proxy} is explicitly requested), this also includes
  * that proxy information. For secure connections the address also includes the
- * SSL socket factory and hostname verifier.
+ * SSL socket factory, hostname verifier and certificate pinner.
```
JakeWharton: oxford :grinning:

okhttp/src/main/java/com/squareup/okhttp/Dns.java
```diff
@@ -0,0 +1,46 @@
+/*
+ * Copyright (C) 2012 Square, Inc.
```
JakeWharton: It's a shame you can't force git to honor a rename + change. I assume this started its life as a `Network.java` rename, hence the 2012.

okhttp/src/main/java/com/squareup/okhttp/Dns.java
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * A domain name service that resolves IP addresses for host names. Most applications will use the
+ * {@linkplain #SYSTEM system DNS service}, which is the default. Some applications may provide
+ * their own implementation to use a different DNS server, to prefer IPv6 addresses, to prefer IPv4
+ * addresses, or to force a specific known IP address.
+ *
+ * <p>Implementations of this interface must be safe for concurrent use.
+ */
+public interface Dns {
+  Dns SYSTEM = new Dns() {
```
JakeWharton: This could use a one-sentence Javadoc, perhaps just speaking to the implementation. If only because `OkHttpClient#setDns` links you directly to this constant.

JakeWharton:
LGTM :shipit:

Issue #1635, opened 2015-05-13
Don't percent-encode non-ASCII characters in fragments


swankjesse:
Sequences like `#β` shouldn't be percent-encoded.

(Follow up to https://github.com/square/okhttp/issues/1486)

PR #1970, opened 2015-11-05
Don't percent-encode non-ASCII characters in fragments

swankjesse:
Fixes https://github.com/square/okhttp/issues/1635

### DIFF
--- a/okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
@@ -673,25 +673,25 @@ public final class HttpUrl {
 
     public Builder username(String username) {
       if (username == null) throw new IllegalArgumentException("username == null");
-      this.encodedUsername = canonicalize(username, USERNAME_ENCODE_SET, false, false);
+      this.encodedUsername = canonicalize(username, USERNAME_ENCODE_SET, false, false, true);
       return this;
     }
 
     public Builder encodedUsername(String encodedUsername) {
       if (encodedUsername == null) throw new IllegalArgumentException("encodedUsername == null");
-      this.encodedUsername = canonicalize(encodedUsername, USERNAME_ENCODE_SET, true, false);
+      this.encodedUsername = canonicalize(encodedUsername, USERNAME_ENCODE_SET, true, false, true);
       return this;
     }
 
     public Builder password(String password) {
       if (password == null) throw new IllegalArgumentException("password == null");
-      this.encodedPassword = canonicalize(password, PASSWORD_ENCODE_SET, false, false);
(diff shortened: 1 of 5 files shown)


okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
```diff
@@ -774,14 +774,15 @@ public Builder encodedPath(String encodedPath) {
...
+          ? queryStringToNamesAndValues(canonicalize(query, QUERY_ENCODE_SET, false, true, false))
           : null;
       return this;
     }
 
     public Builder encodedQuery(String encodedQuery) {
       this.encodedQueryNamesAndValues = encodedQuery != null
-          ? queryStringToNamesAndValues(canonicalize(encodedQuery, QUERY_ENCODE_SET, true, true))
+          ? queryStringToNamesAndValues(
+          canonicalize(encodedQuery, QUERY_ENCODE_SET, true, true, false))
```
JakeWharton: Shouldn't this be indented more?

okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
```diff
@@ -1537,21 +1540,23 @@ static int decodeHexDigit(char c) {
    *
    * @param alreadyEncoded true to leave '%' as-is; false to convert it to '%25'.
    * @param query true if to encode ' ' as '+', and '+' as "%2B".
+   * @param fragment true to leave non-ASCII codepoints as-is; false to encode them.
    */
   static String canonicalize(String input, int pos, int limit, String encodeSet,
-      boolean alreadyEncoded, boolean query) {
+      boolean alreadyEncoded, boolean query, boolean fragment) {
```
JakeWharton: This describes the consumption location, not the actual behavior. This method knows nothing about fragments. Maybe call it `encodeToAscii`? or `retainNonAscii`? I don't know... naming is hard.

JakeWharton: (Granted it's in a file that only concerns itself with urls)

JakeWharton:
lgtm

okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
swankjesse: It's not how IntelliJ does it. Fixed manually.

okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java
swankjesse: I inverted the boolean & used `asciiOnly`. That's semantic.

PR #1977, opened 2015-11-07
Handle UnsupportedCharsetException in HttpLoggingInterceptor

rfc2822:
I got this runtime exception because I have used `HttpLoggingInterceptor` in an app, and it was quite unexpected. Maybe it's wise to point this possibility out in the docs.

Another approach would be that `MediaType.charset()` catches the exception and returns a default value. However, I don't think that's appropriate because the data (which are propbably in another charset, an unknown one) would become misinterpreted and corrupted. Unsupported/unknown charsets are an error condition and should be treated as such.

See also #1978.

### DIFF
--- a/okhttp-logging-interceptor/src/main/java/com/squareup/okhttp/logging/HttpLoggingInterceptor.java
+++ b/okhttp-logging-interceptor/src/main/java/com/squareup/okhttp/logging/HttpLoggingInterceptor.java
@@ -29,6 +29,7 @@ import com.squareup.okhttp.ResponseBody;
 import com.squareup.okhttp.internal.Platform;
 import java.io.IOException;
 import java.nio.charset.Charset;
+import java.nio.charset.UnsupportedCharsetException;
 import java.util.concurrent.TimeUnit;
 import okio.Buffer;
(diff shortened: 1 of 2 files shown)


okhttp-logging-interceptor/src/main/java/com/squareup/okhttp/logging/HttpLoggingInterceptor.java
```diff
@@ -199,9 +200,12 @@ public void setLevel(Level level) {
...
         Charset charset = UTF8;
         MediaType contentType = responseBody.contentType();
-        if (contentType != null) {
-          charset = contentType.charset(UTF8);
-        }
+        if (contentType != null)
+          try {
+            charset = contentType.charset(UTF8);
+          } catch (UnsupportedCharsetException e) {
+            // unsupported charset, assume UTF8 nevertheless
```
swankjesse: Should we log that the media type is probably broken? I'd love to just log something like “couldn't decode the response body; charset is likely malformed”.

Otherwise developers aren't motivated to fix their broken servers.

swankjesse:
Ping!

PR #1982, opened 2015-11-08
Connection allocations model.

swankjesse:
This is intended to replace some of the state machines in HttpConnection, HttpEngine,
and ConnectionPool. With this we'll keep track of all streams in flight, not just
the connections currently idle.

Note that I've sketched this out into a new Connection.java class. In a follow up
change I'll attempt to combine this with the existing Connection class.

### DIFF
--- a/okhttp-tests/src/test/java/com/squareup/okhttp/internal/allocations/ConnectionTest.java
+++ b/okhttp-tests/src/test/java/com/squareup/okhttp/internal/allocations/ConnectionTest.java
@@ -1,0 +1,206 @@
+/*
+ * Copyright (C) 2015 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.okhttp.internal.allocations;
+
+import com.squareup.okhttp.ConnectionPool;
+import org.junit.Test;
+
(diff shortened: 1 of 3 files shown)


okhttp/src/main/java/com/squareup/okhttp/internal/allocations/Connection.java
```diff
@@ -0,0 +1,231 @@
...
+
+    public void rescind() {
+      StreamAllocation streamAllocation = get();
+      if (streamAllocation != null) {
+        streamAllocation.rescinded = true;
+      }
+    }
+  }
+
+  public static class Stream {
```
swankjesse: (this is a placeholder)

okhttp-tests/src/test/java/com/squareup/okhttp/internal/allocations/ConnectionTest.java
```diff
@@ -0,0 +1,207 @@
...
+  }
+
+  @Test public void reserveReturnsNullAfterNoNewStreams() throws Exception {
+    Connection connection = new Connection(connectionPool);
+    connection.setAllocationLimit(1);
+
+    Connection.StreamAllocation a = connection.reserve("a");
+    Connection.Stream a1 = a.newStream("a1");
+
+    // Setting the limit to 0 prevents new reservations.
```
JakeWharton: Is this comment slightly mismatched? This sets the limit as a side-effect with different semantics and is different than actually setting the limit with the `setAllocationCount` API.

swankjesse: Yeah, bad comment. Good catch. (Originally I used a single int, but that didn't handle all the things it needed to handle.)

okhttp/src/main/java/com/squareup/okhttp/ConnectionPool.java
```diff
@@ -75,7 +76,7 @@
   private final int maxIdleConnections;
   private final long keepAliveDurationNs;
 
-  private final LinkedList<Connection> connections = new LinkedList<>();
+  private final Deque<Connection> connections = new ArrayDeque<>();
```
JakeWharton: Nice!

JakeWharton:
LGTM

PR #1984, opened 2015-11-09
Refactor CertificatePinner into an interface and a SHA1-implementation

timcooijmans:
There are plenty of other certificate-pinning implementations besides the SHA1 based chain-pinning method used. This pull request allows developers to provide their own certificate-pinning by refactoring the CertificatePinner into an interface and keeps backwards compatibility by providing the previously used CertificatePinner as Sha1CertificatePinner.


JakeWharton:
Since this is a breaking API change it would have to wait until 3.0, if
indeed it's something we want to add support for.

timcooijmans:
I guess I can refactor this into a non-API breaking change by keeping CertificatePinner as the implementation class and using another name for the interface. Would that be an option?

swankjesse:
What feature is OkHttp’s certificate pinner missing? I'm extremely reluctant to make this feature even more fussy than it is already.

timcooijmans:
There are several reasons, some wishes that I currently have:
- Use SHA256 hashes instead of SHA1
- Pinning all certificates in the chain (currently if one of the certificates in the chain is in the "pins"-list the certificate-chain has a valid "pin")
- Pinning certain combinations of certificates.
- Only pin a specific part of the chain (only the root)
- Base pinning on something else then the public-key.
- Use HTTP Public Key Pinning (https://en.wikipedia.org/wiki/HTTP_Public_Key_Pinning)

swankjesse:
Gotcha. We're currently tracking the pins format used by Chrome. Have they switched to SHA256 yet? When they do, we'll support that in OkHttp.

For now, I recommend building your own pinning in a network interceptor. That code runs after the TLS handshake but before any data is transmitted.

PR #1996, opened 2015-11-14
Break apart close states. Try to send close on IOE.

### LINKED ISSUE #1931: IOException when writing to Websocket sink causes writing to fail forever
This change: [https://github.com/square/okhttp/commit/40a6dc7dd5ff6843b82f904516181182d512d0ca] contains these lines in `public void sendMessage(RequestBody message)`:

```java 
    BufferedSink sink = Okio.buffer(writer.newMessageSink(formatOpcode));
    message.writeTo(sink);
    sink.close();
```

Perhaps rephrase as:
```java
    BufferedSink sink = Okio.buffer(writer.newMessageSink(formatOpcode));
    try {
      message.writeTo(sink);
    } finally {
      sink.close();
    }
```

If an `IOException` is thrown here the sink will never close yet the websocket will remain in a valid state. This causes subsequent executions of `writer.NewMessageSink(formatOpcode)` to fail because the writer is still active.

JakeWharton:
This breaks apart the states used to track sending a close message and whether or not the connection has been closed. Prior to this, certain threading orders would allow multiple calls to close() which would attempt to remove ownership of the connection, the second of which would fail.

Closes #1931.

### DIFF
--- a/okhttp-ws-tests/src/test/java/com/squareup/okhttp/internal/ws/RealWebSocketTest.java
+++ b/okhttp-ws-tests/src/test/java/com/squareup/okhttp/internal/ws/RealWebSocketTest.java
@@ -18,16 +18,19 @@ package com.squareup.okhttp.internal.ws;
 import com.squareup.okhttp.MediaType;
 import com.squareup.okhttp.RequestBody;
 import com.squareup.okhttp.ws.WebSocketRecorder;
+import java.io.Closeable;
 import java.io.IOException;
 import java.net.ProtocolException;
 import java.util.Random;
-import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.Executor;
-import java.util.concurrent.Executors;
-import java.util.concurrent.TimeUnit;
 import okio.Buffer;
 import okio.BufferedSink;
+import okio.BufferedSource;
 import okio.ByteString;
+import okio.Okio;
+import okio.Sink;
+import okio.Source;
+import okio.Timeout;
 import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
@@ -36,6 +39,8 @@ import static com.squareup.okhttp.ws.WebSocket.BINARY;
 import static com.squareup.okhttp.ws.WebSocket.TEXT;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotEquals;
+import static org.junit.Assert.assertSame;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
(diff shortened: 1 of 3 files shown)


JakeWharton:
Sorry there's a lot here. I can probably split it into two PRs if that helps.

okhttp-ws/src/main/java/com/squareup/okhttp/internal/ws/RealWebSocket.java
```diff
@@ -46,8 +47,9 @@
   private volatile boolean writerSentClose;
```
JakeWharton: This no longer requires locking. Only the application thread reads and writes to it for deciding whether to throw an `IllegalStateException`. Other threads look at the value as an indicator of whether they should try to write a close frame, but they'll ultimately catch the `IOException` if they end up missing the correct value.

okhttp-ws/src/main/java/com/squareup/okhttp/internal/ws/RealWebSocket.java
```diff
@@ -45,9 +46,10 @@
   /** True after calling {@link #close(int, String)}. No writes are allowed afterward. */
   private volatile boolean writerSentClose;
   /** True after a close frame was read by the reader. No frames will follow it. */
-  private volatile boolean readerSentClose;
-  /** Lock required to negotiate closing the connection. */
-  private final Object closeLock = new Object();
+  private boolean readerSentClose;
```
JakeWharton: This no longer requires locking or even `volatile`. It's value is only read and written on the reader thread.

okhttp-ws-tests/src/test/java/com/squareup/okhttp/internal/ws/RealWebSocketTest.java
```diff
@@ -207,7 +219,92 @@
...
+      fail();
+    } catch (IOException e) {
+      assertSame(brokenException, e);
+    }
+
+    // A failed write prevents further use of the WebSocket instance.
+    try {
+      client.sendPing(new Buffer().writeUtf8("Ping!"));
+      fail();
+    } catch (IllegalStateException e) {
```
swankjesse: not sure I like the consequences here. One client throwing an IOException can cause another client to get a runtime exception?

swankjesse: ... I think the only defensible exception here is another IOException, perhaps `IOException("stream not writable")`

swankjesse: ... rule of thumb: if it's a programming bug, IllegalStateException. If it's not, IOException. Here I think it isn't a programming bug.

okhttp-ws-tests/src/test/java/com/squareup/okhttp/internal/ws/RealWebSocketTest.java
```diff
@@ -218,12 +315,26 @@
...
     }
+  }
+
+  @Test public void serverCloseThenWritingCloseThrows() throws IOException {
+    server.close(1000, "Hello!");
+    client.readMessage();
+    clientListener.assertClose(1000, "Hello!");
+
     try {
       client.close(1000, "Bye!");
```
swankjesse: should this just fail silently? I have no idea what the application layer would want

swankjesse: For reference, our framed stuff throws if there's an IOException writing the actual bytes, but silently ignores close on already closed. Looks like this is the same. (And that's as good a policy as any!)

okhttp-ws/src/main/java/com/squareup/okhttp/internal/ws/RealWebSocket.java
swankjesse: Do multiple threads share a single client socket socket? My guess is that the UI will enqueue messages to be sent asynchronously, so a single write that fails shouldn't cause other writes to be treated as programming errors?

okhttp-ws/src/main/java/com/squareup/okhttp/internal/ws/RealWebSocket.java
swankjesse: Win.

okhttp-ws/src/main/java/com/squareup/okhttp/internal/ws/RealWebSocket.java
```diff
@@ -129,51 +124,77 @@ public boolean readMessage() {
...
-    writer.writeClose(code, reason);
-
-    if (performClose) {
-      close();
+    try {
+      writer.writeClose(code, reason);
+    } catch (IOException e) {
+      if (connectionClosed.compareAndSet(false, true)) {
+        // Try to close the connection without masking the original exception.
+        try {
```
swankjesse: `closeQuietly(this)` ?

swankjesse: nope, not a public method!

swankjesse:
Structure looks good but I'm unhappy with the exceptions. Thoughts?

okhttp-ws-tests/src/test/java/com/squareup/okhttp/internal/ws/RealWebSocketTest.java
JakeWharton: There is only one client. If it gets IO then subsequent interactions are a
programming bug.

okhttp-ws/src/main/java/com/squareup/okhttp/internal/ws/RealWebSocket.java
JakeWharton: Writes are single threaded and serial. If the calling code calls close,
there's no reason for it to interact with the WebSocket again.

Even if writes are queued, a single thread will be processing them and it
will know whether it ever called close or not.

swankjesse:
Merged cause the stuff I'm worried about is a corner case, possibly academic. I serve the right to make a follow up PR!

PR #1999, opened 2015-11-15
Force consumers to call WS#close on IO error.

### LINKED ISSUE #1931: IOException when writing to Websocket sink causes writing to fail forever

JakeWharton:
This actually un-fixes #1931, but it formalizes the consumer behavior (and ensures that it will behave correctly).

### DIFF
--- a/okhttp-ws-tests/src/test/java/com/squareup/okhttp/internal/ws/RealWebSocketTest.java
+++ b/okhttp-ws-tests/src/test/java/com/squareup/okhttp/internal/ws/RealWebSocketTest.java
@@ -40,7 +40,6 @@ import static com.squareup.okhttp.ws.WebSocket.TEXT;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotEquals;
-import static org.junit.Assert.assertSame;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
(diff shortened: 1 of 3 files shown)


okhttp-ws-tests/src/test/java/com/squareup/okhttp/internal/ws/RealWebSocketTest.java
```diff
@@ -279,28 +253,16 @@
...
-      client.sendPing(new Buffer().writeUtf8("Ping!"));
+      client.sendMessage(RequestBody.create(TEXT, "Hello!"));
       fail();
     } catch (IllegalStateException e) {
-      assertEquals("closed", e.getMessage());
+      assertEquals("must call close()", e.getMessage());
     }
-  }
-
-  @Test public void socketClosedDuringCloseKillsWebSocket() throws IOException {
```
JakeWharton: This case is already covered above.

PR #2004, opened 2015-11-16
OPTIONS allows body.

JakeWharton:
> If the OPTIONS request includes an entity-body (as indicated by the presence of Content-Length or Transfer-Encoding), then the media type MUST be indicated by a Content-Type field. Although this specification does not define any use for such a body, future extensions to HTTP might use the OPTIONS body to make more detailed queries on the server. A server that does not support such an extension MAY discard the request body.


swankjesse:
LGTM

JakeWharton:
```
java.lang.AssertionError: Request + Response: 2643ms
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at com.squareup.okhttp.mockwebserver.MockWebServerTest.throttleResponse(MockWebServerTest.java:238)
```

PR #2008, opened 2015-11-18
Stop returning HTTP/1.0 from Connection.getProtocol().

swankjesse:
This was updating the protocol as a side-effect of an HTTP/1.0 response.
This made the protocol field mutable, and the code more difficult to
trace.

One consequence of this change is that OkHttp will attempt HTTP/1.1 for
all requests, even if a server returns HTTP/1.0 from the first response.
This is closer to our implementation anyway.

### DIFF
--- a/okhttp/src/main/java/com/squareup/okhttp/Connection.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/Connection.java
@@ -78,12 +78,11 @@ public final class Connection {
   private final Route route;
 
   private Socket socket;
-  private boolean connected = false;
+  private Handshake handshake;
+  private Protocol protocol;
   private HttpConnection httpConnection;
   private FramedConnection framedConnection;
-  private Protocol protocol = Protocol.HTTP_1_1;
   private long idleStartTimeNs;
-  private Handshake handshake;
   private int recycleCount;
 
   /**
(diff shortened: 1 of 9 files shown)


okhttp/src/main/java/com/squareup/okhttp/Connection.java
```diff
@@ -444,25 +447,17 @@ boolean isFramed() {
...
-  void setProtocol(Protocol protocol) {
-    if (protocol == null) throw new IllegalArgumentException("protocol == null");
-    this.protocol = protocol;
+    return protocol != null ? protocol : Protocol.HTTP_1_1;
   }
 
   void setTimeouts(int readTimeoutMillis, int writeTimeoutMillis)
       throws RouteException {
-    if (!connected) throw new IllegalStateException("setTimeouts - not connected");
+    if (protocol == null) throw new IllegalStateException("setTimeouts - not connected");
```
JakeWharton: unrelated: What's up with this exception message? The trace will have the method name.

JakeWharton:
LGTM. Unused import in `Internal.java`.

JakeWharton:
Would be great if we had a tool to do line or PR comments of checkstyle violations with links to automatically correct them and update your branch. I can dream...

okhttp/src/main/java/com/squareup/okhttp/Connection.java
swankjesse: it's a bad message. Likely very old. Fixed!

swankjesse: https://android-review.googlesource.com/#/c/63821/1/src/main/java/com/squareup/okhttp/Connection.java

Issue #2001, opened 2015-11-15
Flaky test: CallTest.cancelTagImmediatelyAfterEnqueue()


swankjesse:
```
Nov 15, 2015 3:36:02 AM com.squareup.okhttp.mockwebserver.MockWebServer$3 acceptConnections
INFO: MockWebServer[48241] done accepting connections: Socket closed
Tests run: 109, Failures: 1, Errors: 0, Skipped: 1, Time elapsed: 27.112 sec <<< FAILURE! - in com.squareup.okhttp.CallTest
cancelTagImmediatelyAfterEnqueue(com.squareup.okhttp.CallTest)  Time elapsed: 15.007 sec  <<< FAILURE!
java.lang.AssertionError: Timed out waiting for response to http://localhost:60088/a
	at com.squareup.okhttp.RecordingCallback.await(RecordingCallback.java:63)
	at com.squareup.okhttp.CallTest.cancelTagImmediatelyAfterEnqueue(CallTest.java:1535)
```

PR #2013, opened 2015-11-19
Fix flakiness in CallTest.cancelTagImmediatelyAfterEnqueue()

swankjesse:
There's a race where the HTTP engine has been created, but not yet connected
and the cancel isn't as immediate as it could be. Fix the test by making sure
there's a response to be retrieved.

With the upcoming connection pool changes we might make the implementation
more robust.

Closes: https://github.com/square/okhttp/issues/2001


JakeWharton:
lgtm

Issue #2012, opened 2015-11-19
HttpLoggingInterceptor relative or absolute HttpUrl logging


austynmahoney:
Should the `HttpLoggingInterceptor` log the fully qualified `HttpUrl`?

Currently it [builds the output](https://github.com/square/okhttp/blob/master/okhttp-logging-interceptor/src/main/java/com/squareup/okhttp/logging/HttpLoggingInterceptor.java#L224) using `url.encodedPath()`, which removes the hostname.

### Example
`https://store.google.com/category/phones` => `--> GET /category/phones HTTP/1.1`

JakeWharton:
It mimcs the HTTP request line. Put it as a network interceptor and you
should see a host header.

austynmahoney:
Just tried it as a network interceptor and got the same behavior.

austynmahoney:
Thanks, I see it now. Still trying to get used to the nuances of this logger compared to the old Retrofit one.

W-quan:
It can desige to choose by user?

PR #2015, opened 2015-11-19
Add ability to log full urls.

runningcode:
Add getter for log level.

### DIFF
--- a/okhttp-logging-interceptor/src/main/java/com/squareup/okhttp/logging/HttpLoggingInterceptor.java
+++ b/okhttp-logging-interceptor/src/main/java/com/squareup/okhttp/logging/HttpLoggingInterceptor.java
@@ -17,7 +17,6 @@ package com.squareup.okhttp.logging;
 
 import com.squareup.okhttp.Connection;
 import com.squareup.okhttp.Headers;
-import com.squareup.okhttp.HttpUrl;
 import com.squareup.okhttp.Interceptor;
 import com.squareup.okhttp.MediaType;
 import com.squareup.okhttp.OkHttpClient;
@@ -52,7 +51,7 @@ public final class HttpLoggingInterceptor implements Interceptor {
      * <p>
      * Example:
      * <pre>{@code
-     * --> POST /greeting HTTP/1.1 (3-byte body)
+     * --> POST http://example.com:8080/greeting HTTP/1.1 (3-byte body)
      *
      * <-- HTTP/1.1 200 OK (22ms, 6-byte body)
      * }</pre>
@@ -63,7 +62,7 @@ public final class HttpLoggingInterceptor implements Interceptor {
      * <p>
      * Example:
      * <pre>{@code
-     * --> POST /greeting HTTP/1.1
+     * --> POST http://example.com:8080/greeting HTTP/1.1
      * Host: example.com
      * Content-Type: plain/text
      * Content-Length: 3
@@ -81,7 +80,7 @@ public final class HttpLoggingInterceptor implements Interceptor {
      * <p>
      * Example:
      * <pre>{@code
-     * --> POST /greeting HTTP/1.1
+     * --> POST http://example.com:8080/greeting HTTP/1.1
      * Host: example.com
      * Content-Type: plain/text
      * Content-Length: 3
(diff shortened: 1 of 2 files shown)


runningcode:
I'm not sure if this is better or adding a new `Level` which logs the url is better.
This was very helpful to me in debugging my retrofit 2.0 upgrade and I imagine it could be useful for others as wells since the handling of the baseUrl in retrofit 2.0 has changed.

JakeWharton:
Similar to #2012.

okhttp-logging-interceptor/src/main/java/com/squareup/okhttp/logging/HttpLoggingInterceptor.java
```diff
@@ -123,11 +123,22 @@ public HttpLoggingInterceptor(Logger logger) {
...
   public void setLevel(Level level) {
     this.level = level;
   }
 
+  public Level getLevel() {
+    return level;
+  }
+
+  /** If true, the full url will be logged otherwise just the path is logged **/
+  public void logFullUrl(boolean logFullUrl) {
```
JakeWharton: should have `set` prefix

okhttp-logging-interceptor/src/main/java/com/squareup/okhttp/logging/HttpLoggingInterceptor.java
```diff
@@ -220,9 +231,13 @@ private static String protocol(Protocol protocol) {
     return protocol == Protocol.HTTP_1_0 ? "HTTP/1.0" : "HTTP/1.1";
   }
 
-  private static String requestPath(HttpUrl url) {
-    String path = url.encodedPath();
-    String query = url.encodedQuery();
-    return query != null ? (path + '?' + query) : path;
+  private String requestPath(HttpUrl url) {
+    if (logFullUrl) {
```
JakeWharton: This logic is backwards. Also you don't need an `else` when you `return` inside an `if`
```java
if (logFullUrl) {
  return url.toString();
}

// other things
}
```

JakeWharton:
Don't know how I feel about this option. It's really low value. If this actually a problem I'd rather just log the full URL 100% of the time.

swankjesse:
Yeah, I'd prefer to always log the full URL.

runningcode:
Changed it to log full urls. :)

okhttp-logging-interceptor/src/main/java/com/squareup/okhttp/logging/HttpLoggingInterceptor.java
```diff
@@ -145,7 +149,7 @@ public void setLevel(Level level) {
     Connection connection = chain.connection();
     Protocol protocol = connection != null ? connection.getProtocol() : Protocol.HTTP_1_1;
     String requestStartMessage =
-        "--> " + request.method() + ' ' + requestPath(request.httpUrl()) + ' ' + protocol(protocol);
+        "--> " + request.method() + ' ' + requestUrl(request.httpUrl()) + ' ' + protocol(protocol);
```
JakeWharton: You can just eliminate the method call altogether. The string concatenation will take care of calling `.toString()`.

okhttp-logging-interceptor/src/test/java/com/squareup/okhttp/logging/HttpLoggingInterceptorTest.java
```diff
@@ -66,7 +66,7 @@
     client.newCall(request().build()).execute();
 
     assertEquals(2, logs.size());
-    assertEquals("--> GET / HTTP/1.1", logs.get(0));
+    assertEquals("--> GET http://" + server.getHostName() + ":" + server.getPort() + "/ HTTP/1.1", logs.get(0));
```
JakeWharton: You can use `server.url("/")` here instead of building the URL yourself.

austynmahoney:
@JakeWharton:  As a note from #2012, you suggested adding the logger as a `NetworkInterceptor` to get the Host to show up. When I do that, the body doesn't get decoded before the interceptor logs it, so you get garbled output if it is gzipped.

I'd also prefer the full URL all the time, so you don't have to use it as a `NetworkInterceptor`.

PR #2019, opened 2015-11-21
Implement SOCKS5 proxy for remote DNS resolving.

rnauber:
Hi,
I like yo to merge this patch that enables remote DNS resolution through a SOCKS5 proxy. One use case is to support .onion name resolution through the TOR SOCKS5 proxy. The patch includes a minimal implementation and a test case.

Thank you,
Richard

### DIFF
--- a/okhttp-tests/src/test/java/com/squareup/okhttp/SocksProxy.java
+++ b/okhttp-tests/src/test/java/com/squareup/okhttp/SocksProxy.java
@@ -16,6 +16,7 @@
 package com.squareup.okhttp;
 
 import com.squareup.okhttp.internal.NamedRunnable;
+import com.squareup.okhttp.internal.SocksProtocol;
 import com.squareup.okhttp.internal.Util;
 import java.io.IOException;
 import java.net.InetAddress;
@@ -41,15 +42,9 @@ import okio.Okio;
  * See <a href="https://www.ietf.org/rfc/rfc1928.txt">RFC 1928</a>.
  */
 public final class SocksProxy {
-  private static final int VERSION_5 = 5;
-  private static final int METHOD_NONE = 0xff;
-  private static final int METHOD_NO_AUTHENTICATION_REQUIRED = 0;
-  private static final int ADDRESS_TYPE_IPV4 = 1;
-  private static final int ADDRESS_TYPE_DOMAIN_NAME = 3;
-  private static final int COMMAND_CONNECT = 1;
-  private static final int REPLY_SUCCEEDED = 0;
-
   private static final Logger logger = Logger.getLogger(SocksProxy.class.getName());
+
+  public final String HOSTNAME_THAT_ONLY_THE_PROXY_KNOWS = "onlyProxyCanResolveMe.org";
 
   private final ExecutorService executor = Executors.newCachedThreadPool(
       Util.threadFactory("SocksProxy", false));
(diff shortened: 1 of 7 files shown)


swankjesse:
Doesn't the JDK’s built-in socks proxy do this for us already? I'm uncomfortable shipping a complete socks proxy client in OkHttp without a substantial number of new tests

swankjesse:
I'm going to close this for now because I'm not prepared to take on the maintenance cost of including our own socks proxy client. You can get this behavior by configuring OkHttp to use a custom SocketFactory.

rnauber:
Hi Jesse, 
I definitely see your point. The JDK  includes a nice SOCKS-client implementation that is should be sufficient for most use cases (I checked with java-1.7.0-openjdk-amd64/jre/lib/rt.jar!/java/net/SocksSocketImpl.class). But when using okHttp on Android, somehow only a very limited SOCKS client (SOCKS4 only, no remote DNS) is available. So maybe I can rework this patch to use the Platform mechanism to make this SOCKS implementation Android-only. 
Bolting on Socks5 support through a custom SocketFactory would mean that each Android app has to duplicate the effort of a custom implementation for a very common task (e.g. connecting through Tor). I do not think that is desirable...
So maybe I reduce the scope of this patch (Android only) and up the testing (what test cases do you like to see?) and so the overall impact of this patch should be very limited. What do you say?

See you,
Richard

swankjesse:
Interested in creating a standalone GitHub project, like SocksSocketFactory or TorSocketFactory?

I actually would like to embed such a thing in OkHttp, but only after it gets some real-world exercise: which features do proxies expect? Is version negotiation necessary? Which versions?

And also, perhaps Android itself wants a Socks client. With `ProxySelector` you could automatically use the Proxy for all `.onion` URLs.

rnauber:
> Interested in creating a standalone GitHub project, like SocksSocketFactory or TorSocketFactory?
>I actually would like to embed such a thing in OkHttp, but only after it gets some real-world exercise: which features do proxies expect? Is version negotiation necessary? Which versions?

Well, I could spawn a new project for this. But I think I would limit the scope to SOCKS5 only, no authentication for now. 
But if you want a proven, full-fledged implementation, maybe you could add a dependency to https://github.com/ravn/jsocks . It supports SOCKS4+5, authentication and so on... 
Both would be fine with me.

>And also, perhaps Android itself wants a Socks client. With ProxySelector you could automatically use the Proxy for all .onion URLs.

Well my usecase involves exactly that:
I liked a calendar app (https://gitlab.com/bitfireAT/davdroid) to support my self hosted owncloud server ( on a tor hidden service because of the easy setup without domain name registration, dyndns, etc.). So I added a CustProxySelector to this app, that automatically routes the tld .onion through tor: 

```
public class CustProxySelector extends ProxySelector {
    //...
    @Override
    public List<Proxy> select(URI uri) {
        //.onion host, use Orbot socks proxy
            if (uri.getHost().endsWith(".onion")) {
                List<Proxy> proxyList = new ArrayList<>();
                Proxy proxy = new Proxy(Proxy.Type.SOCKS, new InetSocketAddress("127.0.0.1", 9050));
                proxyList.add(proxy);
                return proxyList;
            }
//...

public class HttpClient extends OkHttpClient {
//...
    protected HttpClient(Logger log, Context context) {
//...
      setProxySelector( new CustProxySelector(this.log,ProxySelector.getDefault()));
//...
```

But this required patching okHttp (that DavDroid uses) because:
1. okHttp resolves the hostname locally and that will not work with .onion
( see okhttp/src/main/java/com/squareup/okhttp/internal/http/RouteSelector.java:183 )
2. the Android socket implementation does not provide SOCKS5 support
which can be fixed by either
 *  adding a handcrafted SOCKS5 socket implementation (that is what the first path includes)
 * adding a dependency on something like jsocks
 *  fixing the socket implementation in Android directly (which seems quite unrealistic)

swankjesse:
I'm thinking the easiest thing is for you to make a custom `SocketFactory` that calls through the minimal socks5 code in this PR. OkHttp will happily use that, and it can do Socks however it prefers.

rnauber:
Ok, I am fine with that, so the broken SOCKS implementation of Android needs to be fixed elsewhere. And for the remaining piece of the puzzle (deferring to resolve the hostname in RouteSelector.java:183), I created a separate PR: #2034

See you,
Richard

Issue #1906, opened 2015-10-08
null byte in User-Agent header after updating to okhttp 2.5.0


brennantaylor:
This is a very intermittent issue that I've only been able to see on the Moto G.

In my client I set a header with `User-Agent: fake` but `MockWebServer` reads it strangely.

````text
java.lang.IllegalArgumentException: Unexpected char 0x00 at 5 in header value: fake�fake
	at com.squareup.okhttp.Headers$Builder.checkNameAndValue(Headers.java:295)
	at com.squareup.okhttp.Headers$Builder.add(Headers.java:245)
	at com.squareup.okhttp.mockwebserver.MockWebServer$FramedSocketHandler.readRequest(MockWebServer.java:890)
	at com.squareup.okhttp.mockwebserver.MockWebServer$FramedSocketHandler.receive(MockWebServer.java:859)
	at com.squareup.okhttp.internal.framed.FramedConnection$Reader$1.execute(FramedConnection.java:648)
	at com.squareup.okhttp.internal.NamedRunnable.run(NamedRunnable.java:33)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1112)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:587)
	at java.lang.Thread.run(Thread.java:818)

````

swankjesse:
Very interesting, this suggests a bug in the HPACK encoder. I assume you're using OkHttp as the client also?

brennantaylor:
Yes and it is selecting h2. I'm still narrowing it down.

brennantaylor:
It looks like I may be reusing a request builder to initiate another request after the first succeeded.

swankjesse:
Oh, very interesting.

brennantaylor:
Also not that intermittent it's just a race condition for tripping the recursive `doRequest`. I can reproduce easily now.

swankjesse:
Awesome. Wanna send a test case? Something that loops 1000 times is okay!

brennantaylor:
Will do.

swankjesse:
@brennantaylor ping!

brennantaylor:
Yeah sorry I forgot. I will get to this soon.

PR #2020, opened 2015-11-21
Fix HTTP/2 bug where we were using SPDY/3 style header concatenation.

swankjesse:
When multiple headers have the same value in SPDY/3, they are concantenated
and separated by \u0000.

When multiple headers have the same value in HTTP/2, they are each written
independently.

This fixes the problem and rearranges the code to share less behavior than
previously. It gets us closer to being able to drop SPDY/3.

Closes https://github.com/square/okhttp/issues/1906

### DIFF
--- a/okhttp/src/main/java/com/squareup/okhttp/internal/http/FramedTransport.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/internal/http/FramedTransport.java
@@ -47,24 +47,60 @@ import static com.squareup.okhttp.internal.framed.Header.TARGET_SCHEME;
 import static com.squareup.okhttp.internal.framed.Header.VERSION;
 
 public final class FramedTransport implements Transport {
+  private static final ByteString CONNECTION = ByteString.encodeUtf8("connection");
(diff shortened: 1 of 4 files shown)


okhttp/src/main/java/com/squareup/okhttp/internal/http/FramedTransport.java
```diff
@@ -196,7 +235,33 @@ private static String joinOnNull(String first, String second) {
...
     StatusLine statusLine = StatusLine.parse(version + " " + status);
     return new Response.Builder()
-        .protocol(protocol)
+        .protocol(Protocol.SPDY_3)
+        .code(statusLine.code)
+        .message(statusLine.message)
+        .headers(headersBuilder.build());
+  }
+
+  /** Returns headers for a name value block containing a SPDY response. */
```
JakeWharton: `s|SPDY|HTTP/2|`

JakeWharton:
LGTM. Feel bad for only a single comment on a comment...

PR #2030, opened 2015-11-24
Write some new connection pool tests.

swankjesse:
These use the Call APIs. They're high-level so they won't break with the
implementation rewrite.

### DIFF
--- a/okhttp-tests/src/test/java/com/squareup/okhttp/NewConnectionPoolTest.java
+++ b/okhttp-tests/src/test/java/com/squareup/okhttp/NewConnectionPoolTest.java
@@ -1,0 +1,230 @@
+/*
+ * Copyright (C) 2015 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


okhttp-tests/src/test/java/com/squareup/okhttp/NewConnectionPoolTest.java
```diff
@@ -0,0 +1,230 @@
...
+import org.junit.Ignore;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TestRule;
+import org.junit.rules.Timeout;
+
+import static org.junit.Assert.assertEquals;
+
+public final class NewConnectionPoolTest {
+  @Rule public final TestRule timeout = new Timeout(30_000);
```
JakeWharton: Are we using JUnit 4.12? There should be an overload that takes a long and a `TimeUnit`.

JakeWharton:
High-level and behavioral tests for the win. Trying to move Retrofit to more of these.

PR #2033, opened 2015-11-24
Support 204 and 205 'No Content' replies when application interceptor.

### LINKED ISSUE #2029: HttpLoggingInterceptor crashes when 204 not content
If the server returns a no content response, `HttpLoggingInterceptor` is crashing for me with this stack trace:

```java
java.io.EOFException
   at okio.RealBufferedSource.require(RealBufferedSource.java:64)
   at okio.RealBufferedSource.request(RealBufferedSource.java:71)
   at com.squareup.okhttp.logging.HttpLoggingInterceptor.intercept(HttpLoggingInterceptor.java:197)
   at com.squareup.okhttp.Call$ApplicationInterceptorChain.proceed(Call.java:223)
   at com.squareup.okhttp.Call.getResponseWithInterceptorChain(Call.java:196)
   at com.squareup.okhttp.Call.execute(Call.java:79)
```

JakeWharton:
Closes #2029.

### DIFF
--- a/okhttp-logging-interceptor/src/main/java/com/squareup/okhttp/logging/HttpLoggingInterceptor.java
+++ b/okhttp-logging-interceptor/src/main/java/com/squareup/okhttp/logging/HttpLoggingInterceptor.java
@@ -27,6 +27,7 @@ import com.squareup.okhttp.RequestBody;
 import com.squareup.okhttp.Response;
 import com.squareup.okhttp.ResponseBody;
 import com.squareup.okhttp.internal.Platform;
+import com.squareup.okhttp.internal.http.HttpEngine;
 import java.io.IOException;
 import java.nio.charset.Charset;
 import java.util.concurrent.TimeUnit;
@@ -194,7 +195,7 @@ public final class HttpLoggingInterceptor implements Interceptor {
       }
 
       String endMessage = "<-- END HTTP";
-      if (logBody) {
+      if (logBody && HttpEngine.hasBody(response)) {
         BufferedSource source = responseBody.source();
         source.request(Long.MAX_VALUE); // Buffer the entire body.
         Buffer buffer = source.buffer();
--- a/okhttp-logging-interceptor/src/test/java/com/squareup/okhttp/logging/HttpLoggingInterceptorTest.java
+++ b/okhttp-logging-interceptor/src/test/java/com/squareup/okhttp/logging/HttpLoggingInterceptorTest.java
@@ -41,23 +41,32 @@ public final class HttpLoggingInterceptorTest {
   @Rule public final MockWebServer server = new MockWebServer();
 
   private final OkHttpClient client = new OkHttpClient();
-  private final List<String> logs = new ArrayList<>();
-  private HttpLoggingInterceptor interceptor;
+  private String host;
+
+  private final LogRecorder networkLogs = new LogRecorder();
+  private final HttpLoggingInterceptor networkInterceptor =
+      new HttpLoggingInterceptor(networkLogs);
+
+  private final LogRecorder applicationLogs = new LogRecorder();
+  private final HttpLoggingInterceptor applicationInterceptor =
+      new HttpLoggingInterceptor(applicationLogs);
+
+  private void setLevel(Level level) {
+    networkInterceptor.setLevel(level);
+    applicationInterceptor.setLevel(level);
+  }
 
   @Before public void setUp() {
-    HttpLoggingInterceptor.Logger logger = new HttpLoggingInterceptor.Logger() {
-      @Override public void log(String message) {
-        logs.add(message);
-      }
-    };
-    interceptor = new HttpLoggingInterceptor(logger);
-    client.networkInterceptors().add(interceptor);
+    client.networkInterceptors().add(networkInterceptor);
+    client.interceptors().add(applicationInterceptor);
     client.setConnectionPool(null);
+
+    host = server.getHostName() + ":" + server.getPort();
   }
 
   @Test public void setLevelShouldPreventNullValue() {
     try {
-      interceptor.setLevel(null);
+      applicationInterceptor.setLevel(null);
       fail();
     } catch (NullPointerException expected) {
       assertEquals("level == null. Use Level.NONE instead.", expected.getMessage());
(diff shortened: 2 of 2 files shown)


okhttp-logging-interceptor/src/main/java/com/squareup/okhttp/logging/HttpLoggingInterceptor.java
```diff
@@ -195,22 +195,26 @@ public HttpLoggingInterceptor setLevel(Level level) {
...
       if (logBody) {
-        BufferedSource source = responseBody.source();
-        source.request(Long.MAX_VALUE); // Buffer the entire body.
-        Buffer buffer = source.buffer();
-
-        Charset charset = UTF8;
-        MediaType contentType = responseBody.contentType();
-        if (contentType != null) {
-          charset = contentType.charset(UTF8);
+        if (response.code() == 204 || response.code() == 205) {
```
swankjesse: I don't think this is the right fix. Client code shouldn't have to know anything about particular response codes, instead it should just consume the empty body.

okhttp-logging-interceptor/src/test/java/com/squareup/okhttp/logging/HttpLoggingInterceptorTest.java
```diff
@@ -66,218 +83,406 @@
...
-    assertTrue(Pattern.matches("OkHttp-Sent-Millis: \\d+", logs.get(10)));
-    assertTrue(Pattern.matches("OkHttp-Received-Millis: \\d+", logs.get(11)));
-    assertEquals("<-- END HTTP", logs.get(12));
+    assertLogs(applicationLogs)
+        .hasMessageEqual("--> GET / HTTP/1.1")
+        .hasMessageEqual("--> END GET")
+        .hasMessageMatch("<-- HTTP/1\\.1 200 OK \\(\\d+ms\\)")
+        .hasMessageEqual("Content-Length: 6")
+        .hasMessageEqual("Content-Type: text/plain; charset=utf-8")
+        .hasMessageEqual("OkHttp-Selected-Protocol: http/1.1")
```
swankjesse: I consider these to be leaked implementation details

okhttp-logging-interceptor/src/test/java/com/squareup/okhttp/logging/HttpLoggingInterceptorTest.java
```diff
@@ -66,218 +83,406 @@
...
+    }
+
+    LogAssert hasMessageEqual(String expected) {
+      assertTrue("No more messages found", index < logs.size());
+      String actual = logs.get(index++);
+      assertEquals(expected, actual);
+      return this;
+    }
+
+    LogAssert hasMessageMatch(String pattern) {
```
swankjesse: hasMessageMatching ?

swankjesse: Don't love that the headers & order of headers is now something that can break this test

okhttp-logging-interceptor/src/main/java/com/squareup/okhttp/logging/HttpLoggingInterceptor.java
JakeWharton: Without this you get EOF because of transparent Gzip being unable to consume its header.

okhttp-logging-interceptor/src/test/java/com/squareup/okhttp/logging/HttpLoggingInterceptorTest.java
swankjesse: (I guess that this is an OkHttp test, so it doesn't really hurt if we change things up)

okhttp-logging-interceptor/src/main/java/com/squareup/okhttp/logging/HttpLoggingInterceptor.java
JakeWharton: Only when used as an application interceptor though.

okhttp-logging-interceptor/src/test/java/com/squareup/okhttp/logging/HttpLoggingInterceptorTest.java
```diff
@@ -41,23 +41,40 @@
   @Rule public final MockWebServer server = new MockWebServer();
 
   private final OkHttpClient client = new OkHttpClient();
-  private final List<String> logs = new ArrayList<>();
-  private HttpLoggingInterceptor interceptor;
+  private String host;
+
+  private final List<String> networkLogs = new ArrayList<>();
+  private final HttpLoggingInterceptor networkInterceptor =
+      new HttpLoggingInterceptor(new HttpLoggingInterceptor.Logger() {
```
swankjesse: Maybe the LogAssert and Logger should be the same instance?

okhttp-logging-interceptor/src/test/java/com/squareup/okhttp/logging/HttpLoggingInterceptorTest.java
JakeWharton: So I had this originally and changed it to "Match" so that it would line up vertically (with "Equal") in the calling code. Other options for lining up would be "Matching" + "EqualsTo".

okhttp-logging-interceptor/src/main/java/com/squareup/okhttp/logging/HttpLoggingInterceptor.java
swankjesse: Got it. You want com.squareup.okhttp.internal.http.HttpEngine#hasBody

okhttp-logging-interceptor/src/test/java/com/squareup/okhttp/logging/HttpLoggingInterceptorTest.java
JakeWharton: It's legacy, and exposed in API now. Kill it?

JakeWharton: Should I manually strip these from being logged?

swankjesse: Maybe in a separate change, we can drop these from everywhere.

swankjesse: https://github.com/square/okhttp/issues/2035

okhttp-logging-interceptor/src/test/java/com/squareup/okhttp/logging/HttpLoggingInterceptorTest.java
JakeWharton: Good idea. Done.

okhttp-logging-interceptor/src/test/java/com/squareup/okhttp/logging/HttpLoggingInterceptorTest.java
JakeWharton: > Don't love that the headers & order of headers is now something that can break this test

Hmm..  I can change the tests to call the `Interceptor` API directly instead of doing these integration tests.

okhttp-logging-interceptor/src/main/java/com/squareup/okhttp/logging/HttpLoggingInterceptor.java
JakeWharton: Cool. Updated.

swankjesse:
LGTM!

PR #2034, opened 2015-11-24
Remote DNS resolving for connections through a SOCKS proxy

rnauber:
This small patch ensures that a connection through a SOCKS proxy will have its hostname resolved through the proxy as well. It includes a test case for that behaviour.


swankjesse:
Awesome. Mind completing Square’s contributor’s agreement ?
https://github.com/square/okhttp/blob/master/CONTRIBUTING.md

rnauber:
Done! 
Thank you Jesse, it was a pleasure contributing to okHttp...

swankjesse:
Awesome. FYI, I made a few small style changes on the merge. Nothing interesting, but worth looking at if you're gonna send us more PRs.

Issue #2035, opened 2015-11-25
Strip OkHttp-Sent-Millis, OkHttp-Sent-Millis


swankjesse:
These are leaked implementation details. They’re useful for the response cache but they shouldn't exist elsewhere in the API.

panzerdev:
Why do you want to remove those headers? I think their useful for tracking the speed of the connections in an Interceptor. 
I use them for this at least.

swankjesse:
You can do the same thing in your interceptor by calling `System.nanoTime()` !

panzerdev:
Sure! By now I was just counting on them beeing there. But I guess the build will break wehen you remove them, because I use the constants for finding them.

swankjesse:
Yeah, they’re a leaked implementation detail. Best to not rely on ’em!

PR #2037, opened 2015-11-25
Remove synthetic protocol header from native API.

JakeWharton:
Also formalize HttpURLConnection-only 'Response-Source' synthetic header to be on par with the protocol header (which is retained for this legacy interface).

Refs #2035


swankjesse:
Nice. LGTM

PR #2040, opened 2015-11-25
separate socketTimeout and readTimeout

### LINKED ISSUE #2028: SocketTimeoutException throw immediately when next request send in keep-alive mode
Assume that keep-alive time set to 5 minute(default-time for default connection pool) and also i set read-timeout for 60 second. when first request send, a connection created on connection pool, response received and naturally connection keep open for 5 minute.

**but for example after 58 second from first response received (2-3 seconds remain for read-timeout) i send another request, after few seconds (before request received by server) i get SocketTimeoutException immediately !**

i think in keep-alive mode read-timeout interval must reset after send new request on persistent connection to avoid this situation. Or set read-timeout on request not connection.

asg-mojtaba:
this change help to resolve #2028 by separate socketTimeout and readTimeout.

### DIFF
--- a/okhttp-tests/src/test/java/com/squareup/okhttp/ConnectionPoolTest.java
+++ b/okhttp-tests/src/test/java/com/squareup/okhttp/ConnectionPoolTest.java
@@ -48,8 +48,8 @@ public final class ConnectionPoolTest {
     Internal.initializeInstanceForTests();
   }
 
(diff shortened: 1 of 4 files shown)


swankjesse:
Please start with a failing test case. If that’s convincing, only then will we proceed to implement a fix.

asg-mojtaba:
You are right, this is my mistake , issue #2028 closed.
thanks man!

PR #2042, opened 2015-11-25
Use bogus pin that doesn’t throw an exception

0xced:
"sha1/BOGUSPIN" throws an `IllegalArgumentException` because `BOGUSPIN` is not valid base64.


swankjesse:
(That's the idea)

swankjesse:
Actually, no, I see what you're saying. Good catch.

swankjesse:
thanks!

Issue #1592, opened 2015-04-29
Call.cancel shouldn't offend strict mode


swankjesse:
We should document that it doesn't block, and get our friends at AOSP to fix strict mode to not issue a false positive warning.

nfuller:
Do you have a stack trace? Is it one of these cases where it calls a potentially blocking POSIX call that, in practice, will never block with the way the sockets are configured?

swankjesse:
It's related to this issue, which has come up again: https://github.com/square/okhttp/issues/869

nfuller:
I'm not up on conscrypt. From a quick skim of the code: when it closes it looks like it sets the socket into blocking mode, then calls SSL_shutdown, which is documented to potentially block. If that's correct it looks like the call to block guard is at least correct with the current implementation.

What makes you think it's a false positive?

swankjesse:
Hmmm . . .  . What I want is to quickly & violently break a socket that isn't needed anymore. This isn't intended to be a clean shutdown that flushes queued bytes; instead it just severs the network. Perhaps for SSL we need to break the underlying socket only, which shouldn't require any blocking.

nfuller:
The block guard code for close() appears to be smart enough to check for linger and only complain if they have it set. AFAIK, the HTTP sockets shouldn't have it set, so it will be classed as non-blocking.

Wavesonics:
I've got a stack if it helps:

android.os.NetworkOnMainThreadException
            at android.os.StrictMode$AndroidBlockGuardPolicy.onNetwork(StrictMode.java:1145)
            at com.android.org.conscrypt.OpenSSLSocketImpl.close(OpenSSLSocketImpl.java:1009)
            at com.squareup.okhttp.Connection.closeIfOwnedBy(Connection.java:142)
            at com.squareup.okhttp.OkHttpClient$1.closeIfOwnedBy(OkHttpClient.java:71)
            at com.squareup.okhttp.internal.http.HttpConnection.closeIfOwnedBy(HttpConnection.java:136)
            at com.squareup.okhttp.internal.http.HttpTransport.disconnect(HttpTransport.java:135)
            at com.squareup.okhttp.internal.http.HttpEngine.disconnect(HttpEngine.java:507)
            at com.squareup.okhttp.Call.cancel(Call.java:122)
            ...

iNoles:
@Wavesonics you supposed to run any network call on background thread.

JakeWharton:
@iNoles that is not the issue

[...]

PR #2048, opened 2015-11-30
Change async cancel to cancel the raw socket only.

swankjesse:
Previously we could close an SSL socket which does synchronous I/O. This
made it unreasonable to cancel a call on a UI thread.

Closes: https://github.com/square/okhttp/issues/1592


JakeWharton:
:+1:

philipgiuliani:
Great to see this one fixed!! :+1: :boom: :tada:

PR #2052, opened 2015-12-01
Fewer buffers in FramedConnections.

swankjesse:
Previously we had a BufferedSource in the RealConnection and another
independent BufferedSource in the FramedConnection. We only need one.


JakeWharton:
The streak is alive!

swankjesse:
The streak is exhausting!

Issue #1484, opened 2015-03-11
Max concurrent streams for HTTP/2


swankjesse:
Not implemented yet. We need policy, both for synchronous and asynchronous calls into a socket that's hit its limit.

Ankit-tyagi-05:
Does max concurrent streams is configurable at OkHttpClient level?
I don't find any way to configure this at client side with OkHttp

swankjesse:
I made a feature request for that here:
https://github.com/square/okhttp/issues/5950

PR #2055, opened 2015-12-02
Honor the max concurrent streams setting.

swankjesse:
Closes https://github.com/square/okhttp/issues/1484

### DIFF
--- a/mockwebserver/src/main/java/com/squareup/okhttp/mockwebserver/MockResponse.java
+++ b/mockwebserver/src/main/java/com/squareup/okhttp/mockwebserver/MockResponse.java
@@ -17,6 +17,7 @@ package com.squareup.okhttp.mockwebserver;
 
 import com.squareup.okhttp.Headers;
 import com.squareup.okhttp.internal.Internal;
+import com.squareup.okhttp.internal.framed.Settings;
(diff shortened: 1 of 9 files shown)


mockwebserver/src/main/java/com/squareup/okhttp/mockwebserver/MockResponse.java
```diff
@@ -242,6 +244,20 @@ public MockResponse withPush(PushPromise promise) {
   }
 
   /**
+   * When {@link MockWebServer#setProtocols(java.util.List) protocols} include
```
JakeWharton: `{@linkplain`

mockwebserver/src/main/java/com/squareup/okhttp/mockwebserver/MockResponse.java
```diff
@@ -242,6 +244,20 @@ public MockResponse withPush(PushPromise promise) {
...
+   * {@linkplain com.squareup.okhttp.Protocol#HTTP_2}, this pushes {@code
```
JakeWharton: Either `{@link` or add "HTTP/2" inside the `{@linkplain`

JakeWharton:
LGTM

Issue #2050, opened 2015-11-30
No "Content-Length" header if missing RequestBody.contentLength()


Malachiasz:
I had request using RequestBody object. I did not Override contentType() method, but set "Content-Length" header manually with:

Request request = new Request.Builder()
                    .url(uploadUri)
                    .post(requestBody)
                    .addHeader("Content-Length", String.valueOf(fileSize))
                    .build();

But my header was removed by OkHttp because of this code in Call class:

    Response getResponse(Request request, boolean forWebSocket) throws IOException {
        RequestBody body = request.body();
        if(body != null) {
            Builder followUpCount = request.newBuilder();
            MediaType response = body.contentType();
            if(response != null) {
                followUpCount.header("Content-Type", response.toString());
            }

            long followUp = body.contentLength();
            if(followUp != -1L) {
                followUpCount.header("Content-Length", Long.toString(followUp));
                followUpCount.removeHeader("Transfer-Encoding");
            } else {
                followUpCount.header("Transfer-Encoding", "chunked");
                followUpCount.removeHeader("Content-Length");
            }

            request = followUpCount.build();
        }

So basically how it works now is that the lib checks if body has contentLength and if it doesn't then it removes header set by me. I would expect exception being thrown, or header not removed.

swankjesse:
This is working as designed. Honoring the headers’ content length is problematic because it interferes with transformations like compression.

So better question: can you work with the current behavior?

Malachiasz:
Sure I can. It's just prone for errors, because a developer may not notice that the library overrides his header. Minimal fix would be to document this behavior in the signature of Request.addHeader().

swankjesse:
Thanks @chaitanyajun12

PR #2059, opened 2015-12-02
Added documentation #2050

chaitanyajun12:
#2050 Added a one liner for the overriding of the Content-Length header.


swankjesse:
Thanks! Please rebase and squash. http://rebaseandsqua.sh!

chaitanyajun12:
Rebase and squashing done. Thanks.

swankjesse:
Merged manually. Note that I tweaked your copy & formatting. In particular I dropped the word 'compression' because OkHttp doesn't do request compression out of the box, and your copy slightly suggests that. (We _do_ do response compression, so we have to be careful not to confuse our users!)

chaitanyajun12:
Ok. Thanks for the correction.

PR #2062, opened 2015-12-04
Don't log gzipped data. It isn't human readable.

swankjesse:


okhttp-logging-interceptor/src/main/java/com/squareup/okhttp/logging/HttpLoggingInterceptor.java
```diff
@@ -237,6 +237,11 @@ public Level getLevel() {
     return response;
   }
 
+  private boolean logBody(Headers headers) {
+    String contentEncoding = headers.get("Content-Encoding");
+    return contentEncoding == null || contentEncoding.equalsIgnoreCase("identity");
```
JakeWharton: `"identity".equalsIgnoreCase(headers.get("Content-Encoding"))` :trollface:

okhttp-logging-interceptor/src/main/java/com/squareup/okhttp/logging/HttpLoggingInterceptor.java
```diff
@@ -179,7 +179,7 @@ public Level getLevel() {
       }
 
       String endMessage = "--> END " + request.method();
-      if (logBody && hasRequestBody) {
+      if (logBody && hasRequestBody && logBody(request.headers())) {
```
JakeWharton: I'm wondering if this should be wrapped around line 193 with an `else` to say "(gzip body omitted)" or something.

JakeWharton: This would be especially nice since we use the buffered body to get its size, which is otherwise unavailable due to lack of a header.

okhttp-logging-interceptor/src/main/java/com/squareup/okhttp/logging/HttpLoggingInterceptor.java
swankjesse: that's the opposite behavior from the one that we want for null.

JakeWharton: Hah. I was just trolling anyway, didn't really look closely.

okhttp-logging-interceptor/src/main/java/com/squareup/okhttp/logging/HttpLoggingInterceptor.java
swankjesse: Fixed. I also restructured this a bit so we can suppress logging for non-text bodies. (That's a separate PR.)

swankjesse: `<-- END HTTP (encoded body omitted)`

JakeWharton:
lgtm

PR #2066, opened 2015-12-06
Fix SSLSocket configuration

0xced:
Trying to connect to a server supporting *only* the `"SSL_RSA_WITH_3DES_EDE_CBC_SHA"` cipher suite on Android 5.0 (API Level 21) and later fails with the following exception:

    javax.net.ssl.SSLHandshakeException: Connection closed by peer
    	at com.android.org.conscrypt.NativeCrypto.SSL_do_handshake(Native Method)
    	at com.android.org.conscrypt.OpenSSLSocketImpl.startHandshake(OpenSSLSocketImpl.java:302)
    	at com.squareup.okhttp.internal.io.RealConnection.connectTls(RealConnection.java:184)
    	at com.squareup.okhttp.internal.io.RealConnection.connectSocket(RealConnection.java:141)
    	at com.squareup.okhttp.internal.io.RealConnection.connect(RealConnection.java:104)
    	at com.squareup.okhttp.internal.http.StreamAllocation.findConnection(StreamAllocation.java:176)
    	at com.squareup.okhttp.internal.http.StreamAllocation.findHealthyConnection(StreamAllocation.java:124)
    	at com.squareup.okhttp.internal.http.StreamAllocation.newStream(StreamAllocation.java:93)
    	at com.squareup.okhttp.internal.http.HttpEngine.connect(HttpEngine.java:283)
    	at com.squareup.okhttp.internal.http.HttpEngine.sendRequest(HttpEngine.java:224)
    	at com.squareup.okhttp.Call.getResponse(Call.java:285)
    	at com.squareup.okhttp.Call$ApplicationInterceptorChain.proceed(Call.java:243)
    	at com.squareup.okhttp.Call.getResponseWithInterceptorChain(Call.java:205)
    	at com.squareup.okhttp.Call.execute(Call.java:80)

It used to work on Android 4.4 and earlier.

In Android 5.0, `"SSL_RSA_WITH_3DES_EDE_CBC_SHA"` (`CipherSuite.TLS_RSA_WITH_3DES_EDE_CBC_SHA`) [is not enabled by default anymore][1] but is still supported.

When computing the supported `ConnectionSpec` for a given `SSLSocket`, the `cipherSuites` are intersected with the *enabled* cipher suites of the socket. By intersecting with the *supported* cipher suites instead, we ensure that the socket is properly configured with the cipher suites defined in the connection spec.

  [1]: http://developer.android.com/reference/javax/net/ssl/SSLEngine.html

### DIFF
--- a/okhttp/src/main/java/com/squareup/okhttp/ConnectionSpec.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/ConnectionSpec.java
@@ -137,15 +137,15 @@ public final class ConnectionSpec {
   }
 
   /**
-   * Returns a copy of this that omits cipher suites and TLS versions not enabled by {@code
+   * Returns a copy of this that omits cipher suites and TLS versions not supported by {@code
    * sslSocket}.
    */
   private ConnectionSpec supportedSpec(SSLSocket sslSocket, boolean isFallback) {
     String[] cipherSuitesIntersection = cipherSuites != null
-        ? Util.intersect(String.class, cipherSuites, sslSocket.getEnabledCipherSuites())
+        ? Util.intersect(String.class, cipherSuites, sslSocket.getSupportedCipherSuites())
         : sslSocket.getEnabledCipherSuites();
     String[] tlsVersionsIntersection = tlsVersions != null
-        ? Util.intersect(String.class, tlsVersions, sslSocket.getEnabledProtocols())
+        ? Util.intersect(String.class, tlsVersions, sslSocket.getSupportedProtocols())
         : sslSocket.getEnabledProtocols();
 
     // In accordance with https://tools.ietf.org/html/draft-ietf-tls-downgrade-scsv-00
@@ -162,14 +162,14 @@ public final class ConnectionSpec {
 
   /**
    * Returns {@code true} if the socket, as currently configured, supports this connection spec.
-   * In order for a socket to be compatible the enabled cipher suites and protocols must intersect.
+   * In order for a socket to be compatible the supported cipher suites and protocols must intersect.
    *
    * <p>For cipher suites, at least one of the {@link #cipherSuites() required cipher suites} must
-   * match the socket's enabled cipher suites. If there are no required cipher suites the socket
+   * match the socket's supported cipher suites. If there are no required cipher suites the socket
    * must have at least one cipher suite enabled.
    *
    * <p>For protocols, at least one of the {@link #tlsVersions() required protocols} must match the
-   * socket's enabled protocols.
+   * socket's supported protocols.
    */
   public boolean isCompatible(SSLSocket socket) {
     if (!tls) {
@@ -177,12 +177,12 @@ public final class ConnectionSpec {
     }
 
     if (tlsVersions != null
-        && !nonEmptyIntersection(tlsVersions, socket.getEnabledProtocols())) {
+        && !nonEmptyIntersection(tlsVersions, socket.getSupportedProtocols())) {
       return false;
     }
 
     if (cipherSuites != null
-        && !nonEmptyIntersection(cipherSuites, socket.getEnabledCipherSuites())) {
+        && !nonEmptyIntersection(cipherSuites, socket.getSupportedCipherSuites())) {
       return false;
     }
 


swankjesse:
The current behavior is deliberate. Instead you need to fix your SSLSocketFactory to enable it there.

swankjesse:
And of course, fix your server.

0xced:
OK, I see now that this was changed in f695ea7f23e1df67aa28756da47b39b0a42cddfd by @nfuller and I understand the motivation.

But now I’m left wondering: what’s the purpose of the `ConnectionSpec` class? By intersecting with the socket’s cipher suites and protocols, we can only match or reduce the number of supported cipher suites and protocols. So why define everything in the ConnectionSpec if we are eventually just using the enabled values defined by the socket? This is very misleading, I spent a huge amount of time to realise that OkHttp was dropping `"SSL_RSA_WITH_3DES_EDE_CBC_SHA"` even though it was in the connection spec.

What would have helped me a lot for my particular issue is a log telling me that `"SSL_RSA_WITH_3DES_EDE_CBC_SHA"` was dropped because it was not enabled on the socket. Would you be interested in a pull request implementing such a log? If yes, should I use `Platform.get().log()`, `Platform.get().logW()` or something else?

swankjesse:
Your make a strong argument. @nfuller, willing to reverse this?

nfuller:
I would rather not reverse the use of "enabled" vs "supported". Hopefully compelling arguments below.

Improved logging: I agree the connection code is highly complex and sometimes working out how client ended up connecting (or not) through all the various routes, IPs, protocol fallbacks and ciphers negotiation can be very difficult: I tend to use the debugger for this and it's painful. Not sure on the OkHttp "logging for debug" strategy - I've not seen much code for it: it tends to just be informative exception messages.

Two sections below: context / motivation for the choice when I implemented it followed by a "how can we make everybody happy?" section.

----

This is expanding on what I wrote in the change that @0xced referenced, so may be redundant, but there's a more detail.

There is probably a tension between "OkHttp as an app library" Vs "OkHttp as an Android platform component". My focus has been the latter.

A person's stance may depend on whether they expect the library/app or the platform to have the best available information (about what ciphers/protocols are still good). I'm mostly reacting to what I've seen and the unintended side effects of what was previously implemented.

By platform I mean "the SSLSocket implementation". On Android, the SSLSocket implementation can be the one that came with the device (potentially old info), or it can be one installed by using the dynamic security provider from Google Play Services (near best available info).

When a cipher or protocol becomes discredited, the SSLSocket developers don't just remove support. They leave support for people that must still use it, but they disable it by default for the general case. On flip side with new ciphers / protocols they also tend to be supported but not enabled until the developers are confident in the impl.

OkHttp can have "worse info" than the SSLSocket in the scenario where there's an old APK with an app-bundled okhttp running on a newer Android release, or in the "Android-bundled" case (where OkHttp is being used to provide the HttpURLConnection) with a Play Services-provided SSLSocket.

So, sometimes OkHttp has the latest/best info, sometimes the SSLSocket. 

I expect app developers are going to be thinking "clearly my copy of OkHttp in the APK is going to be newer than the version of Android, and I'll keep taking the latest version of OkHttp and releasing my app regularly" but are not thinking so much about how things might age, the use of OkHttp in the platform or assumptions that might break when SSLSocket can "leap frog" the rest of the platform and the app.

The current approach leads to a compromise: the intersection of both sets of info is used and only the ones that OkHttp and the platform agree are "good" are used.

Using the "enabled" set tends to result in a "subtraction effect" (both must agree otherwise something is disabled / not enabled). Using "supports" was having a more "additive" effect (okhttp could override the platform's information leading (in some scenarios) to "not good" ciphers being re-enabled.

-----

Moving forward...

For use of OkHttp in the Android platform I think using the "enabled" set is probably still best and I think it's a safe default for the general case given some of the scenarios outlined above.

Clearly I have a bias because of my Android platform role / PoV. For OkHttp use in apps, I think developers should probably have the ability to override this behavior when it doesn't work for them _and_ they know what they are doing and understand the consequences.

In an earlier version of one of the ConnectionSpec changes I think I controlled this behavior with a flag - i.e. should I use the enabled or the supported set when working out which to enable on a socket? I was nervous about foisting the "use enabled" behavior on the general case but I think @swankjesse  had legitimate concerns about expanding the public API so we went with making it the general case.

FYI - at one stage it was looking like ConnectionSpec may embody all the necessary logic about this selection (ConnectionSpec.apply(SSLSocket, boolean) / ConnectionSpec.isCompatible(SSLSocket) in the current Android copy) which could lead the way to making ConnectionSpec pluggable with a strategy or subclassable for high levels of customization. This would probably have a disproportionate impact on the public API. Not sure if the code has been rewritten since I last looked, though. Suspect a "enabled" vs "supports" switch somewhere would satisfy most if not all usecases.

swankjesse:
Thanks for the thorough explanation @nfuller. I think the current policy makes the most sense going forward. (We may want to mention that in our documentation.)

rakshitsoni02:
Getting same error 
06-09 16:15:02.722 24776-24860/org.trakie.application W/System.err: javax.net.ssl.SSLException: Connection closed by peer
06-09 16:15:03.201 24776-24860/org.trakie.application W/System.err: at org.apache.harmony.xnet.provider.jsse.NativeCrypto.SSL_do_handshake(Native Method)
06-09 16:15:03.250 24776-24860/org.trakie.application W/System.err: at org.apache.harmony.xnet.provider.jsse.OpenSSLSocketImpl.startHandshake(OpenSSLSocketImpl.java:371)
06-09 16:15:03.280 24776-24860/org.trakie.application W/System.err: at okhttp3.internal.io.RealConnection.connectTls(RealConnection.java:195)
06-09 16:15:03.301 24776-24860/org.trakie.application W/System.err: at okhttp3.internal.io.RealConnection.connectSocket(RealConnection.java:148)
06-09 16:15:03.380 24776-24860/org.trakie.application W/System.err: at okhttp3.internal.io.RealConnection.connect(RealConnection.java:111)
06-09 16:15:03.420 24776-24860/org.trakie.application W/System.err: at okhttp3.internal.http.StreamAllocation.findConnection(StreamAllocation.java:188)
06-09 16:15:03.460 24776-24860/org.trakie.application W/System.err: at okhttp3.internal.http.StreamAllocation.findHealthyConnection(StreamAllocation.java:127)
06-09 16:15:03.490 24776-24860/org.trakie.application W/System.err: at okhttp3.internal.http.StreamAllocation.newStream(StreamAllocation.java:97)
06-09 16:15:03.640 24776-24860/org.trakie.application W/System.err: at okhttp3.internal.http.HttpEngine.connect(HttpEngine.java:289)
06-09 16:15:03.670 24776-24860/org.trakie.application W/System.err: at okhttp3.internal.http.HttpEngine.sendRequest(HttpEngine.java:241)
06-09 16:15:03.830 24776-24860/org.trakie.application W/System.err: at okhttp3.RealCall.getResponse(RealCall.java:240)
06-09 16:15:03.870 24776-24860/org.trakie.application W/System.err: at okhttp3.RealCall$ApplicationInterceptorChain.proceed(RealCall.java:198)
06-09 16:15:03.980 24776-24860/org.trakie.application W/System.err: at okhttp3.RealCall.getResponseWithInterceptorChain(RealCall.java:160)
06-09 16:15:04.000 24776-24860/org.trakie.application W/System.err: at okhttp3.RealCall.execute(RealCall.java:57)

0xced:
For the record, here is the socket factory I used to workaround this issue: [CompatibilitySSLSocketFactory.java](https://gist.github.com/0xced/e188c7cd32e67dfe0b9c33c7fc26d910)

I did not know that OkHttp is also an _Android platform component_ but from the point of view of _OkHttp as an app library_ this workaround still looks really awkward one year later. 😕

Anyway, I now deleting that custom socket factory from my codebase since the server I'm talking to has been upgraded to a modern TLS configuration. 🎉

PR #2067, opened 2015-12-06
Restore time-based eviction.

swankjesse:
We still don't implement a limit on the number of open connections.
That needs to come afterwards, and will be difficult because we'll
need to add policy on which connections to evict.

### DIFF
--- a/okhttp/src/main/java/com/squareup/okhttp/ConnectionPool.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/ConnectionPool.java
@@ -24,6 +24,10 @@ import java.util.ArrayList;
 import java.util.Deque;
 import java.util.Iterator;
 import java.util.List;
+import java.util.concurrent.Executor;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
 
 /**
  * Manages reuse of HTTP and SPDY connections for reduced network latency. HTTP
@@ -57,7 +61,8 @@ public final class ConnectionPool {
     String keepAlive = System.getProperty("http.keepAlive");
     String keepAliveDuration = System.getProperty("http.keepAliveDuration");
     String maxIdleConnections = System.getProperty("http.maxConnections");
-    long keepAliveDurationMs = keepAliveDuration != null ? Long.parseLong(keepAliveDuration)
+    long keepAliveDurationMs = keepAliveDuration != null
+        ? Long.parseLong(keepAliveDuration)
         : DEFAULT_KEEP_ALIVE_DURATION_MS;
     if (keepAlive != null && !Boolean.parseBoolean(keepAlive)) {
       systemDefault = new ConnectionPool(0, keepAliveDurationMs);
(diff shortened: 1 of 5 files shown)


okhttp/src/main/java/com/squareup/okhttp/internal/io/RealConnection.java
```diff
@@ -67,7 +67,7 @@
    * The application layer socket. Either an {@link SSLSocket} layered over {@link #rawSocket}, or
    * {@link #rawSocket} itself if this connection does not use SSL.
    */
-  private Socket socket;
+  public Socket socket;
```
swankjesse: excessive visibility on internal classes . . . ugh. Thoughts?

okhttp/src/main/java/com/squareup/okhttp/ConnectionPool.java
```diff
@@ -166,4 +210,64 @@ public void evictAll() {
...
+    } else if (inUseConnectionCount != 0) {
+      // All connections are in use. It'll be at least the keep alive duration 'til we run again.
+      return keepAliveDurationNs;
+    } else {
+      // No connections, idle or in use.
+      return -1;
+    }
+  }
+
+  void setCleanupRunnableForTest(Runnable cleanupRunnable) {
```
swankjesse: also inconsistent!

swankjesse:
(test failure is a flake)

okhttp/src/main/java/com/squareup/okhttp/ConnectionPool.java
```diff
@@ -68,16 +73,52 @@
...
+   * zero threads, permitting this pool to be garbage collected.
+   */
+  private final Executor executor = new ThreadPoolExecutor(
+      0 /* corePoolSize */, 1 /* maximumPoolSize */, 60L /* keepAliveTime */, TimeUnit.SECONDS,
+      new LinkedBlockingQueue<Runnable>(), Util.threadFactory("OkHttp ConnectionPool", true));
+
   /** The maximum number of idle connections for each address. */
   private final int maxIdleConnections;
   private final long keepAliveDurationNs;
+  private Runnable cleanupRunnable = new Runnable() {
```
JakeWharton: `final`

JakeWharton: Ah, I see.

okhttp/src/main/java/com/squareup/okhttp/ConnectionPool.java
JakeWharton: You could use a packaged-scoped third constructor as an alternative. I don't really care either way.

okhttp/src/main/java/com/squareup/okhttp/internal/io/RealConnection.java
JakeWharton: I don't have a strong opinion. It's bad practice when widespread, but I don't think it's terrible to do it when you otherwise can't access the internal state that needs verified. You could put a `public Socket getSocket()` on this but that changes nothing in actuality.

okhttp/src/main/java/com/squareup/okhttp/ConnectionPool.java
swankjesse: Originally that's what I had, but sadly the constructors can't delegate for a surprising reason.

PR #2071, opened 2015-12-07
moved all body validations from Request.Builder to Request.validate()

### LINKED ISSUE #2069: Interceptor chain should be run before the request is checked for method and body requirements
Since I can use Interceptors to completely change the request, it is kind of odd that Builder.method(String method, RequestBody body) is run before the chain of Interceptors. I can quite easily do some processing in an interceptor that either:

* changes an invalid request into a valid one
* changes a valid request into an invalid one

It seems to me that this check should be executed after the Interceptor chain has had its chance to run and fix/break things.

todorus:
fixes #2069

Builder validates the Request before the Interceptor chain. The Interceptor chain can easily rewrite a request to become valid and vice versa, which makes this validation better suited after the chain has made its changes to the Request.

### DIFF
--- a/okhttp-tests/src/test/java/com/squareup/okhttp/CallTest.java
+++ b/okhttp-tests/src/test/java/com/squareup/okhttp/CallTest.java
@@ -219,10 +219,30 @@ public final class CallTest {
   @Test public void getWithRequestBody() throws Exception {
     server.enqueue(new MockResponse());
 
+    Request request = new Request.Builder()
+        .url(server.url("/"))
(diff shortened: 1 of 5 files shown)


swankjesse:
I think requests need to be valid at every step of the way. It's very weird to send invalid requests & assume an interceptor will repair 'em.

todorus:
@swankjesse and is it weird to assume an Intercepter can break them?

Rolf-Smit:
I totally agree with @todorus.

Better check them at the end of the road, than at the beginning and assuming the rest of the road won't break the request.

swankjesse:
Currently they're validated on each step.

PR #2077, opened 2015-12-09
Detect leaked connections.

swankjesse:
This is already suggesting some bugs in OkHttp itself.


JakeWharton:
Took a while, but I got through it!

PR #2078, opened 2015-12-09
fix #1919: PROPFIND request should not redirect to GET request

### LINKED ISSUE #1919: redirect for PROFIND request issues GET request
see code in `com.squareup.okhttp.internal.http.HttpEngine#followUpRequest`

```java
      case HTTP_MULT_CHOICE:
      case HTTP_MOVED_PERM:
      case HTTP_MOVED_TEMP:
      case HTTP_SEE_OTHER:
        // Does the client allow redirects?
        if (!client.getFollowRedirects()) return null;

        String location = userResponse.header("Location");
        if (location == null) return null;
        HttpUrl url = userRequest.httpUrl().resolve(location);

        // Don't follow redirects to unsupported protocols.
        if (url == null) return null;

        // If configured, don't follow redirects between SSL and non-SSL.
        boolean sameScheme = url.scheme().equals(userRequest.httpUrl().scheme());
        if (!sameScheme && !client.getFollowSslRedirects()) return null;

        // Redirects don't include a request body.
        Request.Builder requestBuilder = userRequest.newBuilder();
        if (HttpMethod.permitsRequestBody(userRequest.method())) {
          requestBuilder.method("GET", null);
          requestBuilder.removeHeader("Transfer-Encoding");
          requestBuilder.removeHeader("Content-Length");
          requestBuilder.removeHeader("Content-Type");
        }
```

Thus, any http body is lost and out of a PROFIND request I get a GET request instead.

IMHO we need a more flexible solution for coming up with a redirected request. For PROPFIND, the redirected request should again be PROPFIND and the HTTP body should be maintained.

rburgst:
fix for issue #1919

### DIFF
--- a/okhttp/src/main/java/com/squareup/okhttp/internal/http/HttpMethod.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/internal/http/HttpMethod.java
@@ -41,6 +41,11 @@ public final class HttpMethod {
         || method.equals("LOCK");     // (WebDAV) body: create lock, without body: refresh lock
   }
 
+  public static boolean redirectsToGET(String method) {
+    // all requests but propfind should redirect to a GET request
+    return !method.equals("PROPFIND");
+  }
+
   private HttpMethod() {
   }
 }
--- a/okhttp/src/main/java/com/squareup/okhttp/internal/http/HttpMethod.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/internal/http/HttpMethod.java
@@ -41,8 +41,8 @@ public final class HttpMethod {
         || method.equals("LOCK");     // (WebDAV) body: create lock, without body: refresh lock
   }
 
-  public static boolean redirectsToGET(String method) {
-    // all requests but propfind should redirect to a GET request
+  public static boolean redirectsToGet(String method) {
+    // All requests but PROPFIND should redirect to a GET request.
     return !method.equals("PROPFIND");
   }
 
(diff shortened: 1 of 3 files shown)


okhttp/src/main/java/com/squareup/okhttp/internal/http/HttpMethod.java
```diff
@@ -41,6 +41,11 @@ public static boolean permitsRequestBody(String method) {
         || method.equals("LOCK");     // (WebDAV) body: create lock, without body: refresh lock
   }
 
+  public static boolean redirectsToGET(String method) {
```
swankjesse: nit: `GET` –> `Get` ?

okhttp/src/main/java/com/squareup/okhttp/internal/http/HttpMethod.java
```diff
@@ -41,6 +41,11 @@ public static boolean permitsRequestBody(String method) {
...
+    // all requests but propfind should redirect to a GET request
```
swankjesse: http://nedbatchelder.com/blog/201401/comments_should_be_sentences.html

rburgst: will fix this but the lines above in the same file did not really adhere to that, therefore, I didn't know that this was in the code style.

rburgst:
implemented the review comments

swankjesse:
Great! Thanks.

swankjesse:
Merged manually!

PR #2080, opened 2015-12-10
Fix some stream leaks.

swankjesse:
There's a few places where OkHttp could leak a connection that needed to be
closed. With our new connection pool model this is easier to find. This fixes
two specific problems:
 - too many redirects doesn't release the last used connection
 - interceptors that throw runtime exceptions don't release the connection

There are likely more situations. I have hacked together a small little test
harness to make finding these leaks easier; that's not included in this PR.

### DIFF
--- a/okhttp/src/main/java/com/squareup/okhttp/Call.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/Call.java
@@ -281,9 +281,11 @@ public class Call {
         throw new IOException("Canceled");
       }
 
+      boolean releaseConnection = true;
       try {
         engine.sendRequest();
         engine.readResponse();
+        releaseConnection = false;
       } catch (RequestException e) {
         // The attempt to interpret the request failed. Give up.
         throw e.getCause();
@@ -291,6 +293,7 @@ public class Call {
         // The attempt to connect via a route failed. The request will not have been sent.
         HttpEngine retryEngine = engine.recover(e);
         if (retryEngine != null) {
+          releaseConnection = false;
           engine = retryEngine;
           continue;
         }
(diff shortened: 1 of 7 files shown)


JakeWharton:
LGTM so far,
```

Tests in error: 
  ConnectionReuseTest.silentRetryWhenIdempotentRequestFailsOnReusedConnection:162 Â» IllegalState
  CallTest.cleartextCallsFailWhenCleartextIsDisabled:1002 Â» IllegalState release...
  CallTest.postBodyRetransmittedOnFailureRecovery:1097 Â» IllegalState released
  CallTest.recoverWhenRetryOnConnectionFailureIsTrue:890->executeSynchronously:2139 Â» IllegalState
  CallTest.failingCallsDoNotInterfereWithConnection:2007->access$200:86->makeFailingCall:2131 Â» IllegalState
```

swankjesse:
Uh oh

swankjesse:
... This is the problem when I get CallTest working, then move on to InterceptorTest, and don't circle back. The fix for the test failures is in Call, which needs to not release the stream if a recovery is being made.

okhttp/src/main/java/com/squareup/okhttp/Call.java
```diff
@@ -291,6 +293,7 @@ Response getResponse(Request request, boolean forWebSocket) throws IOException {
         // The attempt to connect via a route failed. The request will not have been sent.
         HttpEngine retryEngine = engine.recover(e);
         if (retryEngine != null) {
+          releaseConnection = false;
```
JakeWharton: Ah, GitHub's collapsing of unchanged code hid the `continue`s.

JakeWharton:
LGTM, double-rainbow test flake on the PR.
```
Tests in error: 
  ConnectionReuseTest.staleConnectionNotReusedForNonIdempotentRequest:184 » SocketTimeout
```
```
Failed tests: 
  Spdy3ConnectionTest.serverSendsSettingsToClient:285 expected:<10> but was:<0>
```

swankjesse:
Yeah, the flaky tests are angry!

swankjesse:
I think this is forward progress, but there’s more to do here in simplification.

Issue #2064, opened 2015-12-04
Model API FormEncoding type.


JakeWharton:
We have `MultipartBuilder` and `FormEncodingBuilder` but they produce an opaque `RequestBody`, instead of a subclass that offers APIs to inspect the composite parts.

PR #2082, opened 2015-12-11
Promote Multipart to a top-level type with a nested builder.

JakeWharton:
This creates a model-like API in which individual parts can be read. It also offers the posibility of using this model to parse multipart in the future.

Refs #2064.


swankjesse:
LGTM

PR #2085, opened 2015-12-12
Add URL to HTTP response logging

adstro:
Add the URL to the HTTP response when it is logged in the same manner
that the URL is added to the request logging. The added logging adds
some additional clarity when reading logs that contain many
request/response pairs.


swankjesse:
Thoughts @JakeWharton?

JakeWharton:
If we're going to do this, there's no reason to put "HTTP/1.1" anymore (to mimic the on-wire data) and I would argue the status code should come before the URL.

adstro:
@swankjesse @JakeWharton I am happy to update the code with the changes outlined above if the idea of having the URL in the response seems like a good one. I have just found that having the URL somewhere in the response makes it a bit easier to search through logs when troubleshooting.

swankjesse:
@adstro awesome, please do!

adstro:
@swankjesse Made the changes and pushed them. The travis build seems to be failing eventhough all the tests pass on my machine and when they run under my travis-ci account. I don't think I can trigger another travis build unless I commit another change (which would just be to trigger the build).

JakeWharton:
Squashed, rebased, and merged. Thanks!

PR #2094, opened 2015-12-14
Use the cipher suite enum rather than the string.

swankjesse:


JakeWharton:
LGTM
```
src/test/java/com/squareup/okhttp/CacheTest.java:[262,53] incompatible types: com.squareup.okhttp.CipherSuite cannot be converted to java.lang.String
```
(FYI I rebased this branch for you, you'll need to do it locally as well or pull the new SHA from the remote)

PR #2101, opened 2015-12-15
Convert form-encoding body to model-based version.

### LINKED ISSUE #2064: Model API FormEncoding type.

JakeWharton:
Update the Multipart version as well to match some conventions.

Closes #2064.


JakeWharton:
This currently exposes the user-supplied String in the model API. We probably should behave more like HttpUrl.

okhttp/src/main/java/com/squareup/okhttp/FormEncodingBody.java
```diff
@@ -0,0 +1,137 @@
...
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.BitSet;
+import java.util.List;
+import okio.Buffer;
+import okio.BufferedSink;
+
+import static com.squareup.okhttp.HttpUrl.FORM_ENCODE_SET;
+
+public final class FormEncodingBody extends RequestBody {
```
swankjesse: `Form` ?

okhttp/src/main/java/com/squareup/okhttp/FormEncodingBody.java
```diff
@@ -0,0 +1,137 @@
...
+  private static final MediaType CONTENT_TYPE =
+      MediaType.parse("application/x-www-form-urlencoded");
+
+  private final List<String> names;
+  private final List<String> values;
+  private final BitSet encoded;
+
+  public FormEncodingBody(List<String> names, List<String> values, BitSet encoded) {
```
swankjesse: private

okhttp/src/main/java/com/squareup/okhttp/FormEncodingBody.java
```diff
@@ -0,0 +1,137 @@
...
+
+  public List<String> names() {
+    return names;
+  }
+
+  public String name(int index) {
+    return names.get(index);
+  }
+
+  public List<String> values() {
```
swankjesse: We’ll probably want some map-like access also

swankjesse: Is getting values as a list useful?

okhttp/src/main/java/com/squareup/okhttp/FormEncodingBody.java
```diff
@@ -0,0 +1,137 @@
...
+    this.values = Util.immutableList(values);
+    this.encoded = (BitSet) encoded.clone();
+  }
+
+  /** The number of key-value pairs in this form-encoded body. */
+  public int size() {
+    return names.size();
+  }
+
+  public List<String> names() {
```
swankjesse: Set?

swankjesse: (I think of this as map keys)

okhttp/src/main/java/com/squareup/okhttp/FormEncodingBody.java
```diff
@@ -0,0 +1,137 @@
...
+    return names;
+  }
+
+  public String name(int index) {
+    return names.get(index);
```
swankjesse: this needs to decode?

okhttp/src/main/java/com/squareup/okhttp/FormEncodingBody.java
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  public List<String> values() {
+    return values;
+  }
+
+  public String value(int index) {
+    return values.get(index);
```
swankjesse: needs to decode?

okhttp/src/main/java/com/squareup/okhttp/FormEncodingBody.java
swankjesse: Maybe just omit this?

swankjesse:
LGTM

I think we might want to follow-up with handy map-like access to the parameters

okhttp/src/main/java/com/squareup/okhttp/FormEncodingBody.java
JakeWharton: Nope. It's an ordered list of pairs. Repeated keys are valid!

JakeWharton: Although perhaps we want a Headers-like approach here.

JakeWharton:
Renamed to `FormBody`, encode on the way in, JIT decode on the way out, expose encoded accessors.

PR #2105, opened 2015-12-15
Remove IOException from source() and streams.

### LINKED ISSUE #1696: Evaluate removing IOException from ResponseBody close

JakeWharton:
Closes #1696.

### DIFF
--- a/okhttp-android-support/src/main/java/com/squareup/okhttp/internal/huc/JavaApiConverter.java
+++ b/okhttp-android-support/src/main/java/com/squareup/okhttp/internal/huc/JavaApiConverter.java
@@ -474,10 +474,9 @@ public final class JavaApiConverter {
    * Creates an OkHttp Response.Body containing the supplied information.
    */
   private static ResponseBody createOkBody(final Headers okHeaders,
-      final CacheResponse cacheResponse) {
+      final CacheResponse cacheResponse) throws IOException {
+    final BufferedSource body = Okio.buffer(Okio.source(cacheResponse.getBody()));
     return new ResponseBody() {
-      private BufferedSource body;
-
       @Override
       public MediaType contentType() {
         String contentTypeHeader = okHeaders.get("Content-Type");
@@ -488,11 +487,7 @@ public final class JavaApiConverter {
       public long contentLength() {
         return OkHeaders.contentLength(okHeaders);
       }
-      @Override public BufferedSource source() throws IOException {
-        if (body == null) {
-          InputStream is = cacheResponse.getBody();
-          body = Okio.buffer(Okio.source(is));
-        }
+      @Override public BufferedSource source() {
         return body;
       }
     };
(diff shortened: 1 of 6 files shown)


okhttp-android-support/src/main/java/com/squareup/okhttp/internal/huc/JavaApiConverter.java
```diff
@@ -727,7 +718,11 @@ public Object getContent(Class[] classes) throws IOException {
 
     @Override
     public InputStream getInputStream() throws IOException {
-      throw throwResponseBodyAccessException();
+      return new InputStream() {
+        @Override public int read() throws IOException {
+          throw throwResponseBodyAccessException();
```
JakeWharton: I'm most unsure about the changes in this file because the larger semantics of this are unclear to me. As you can see, this is the _only_ usage which necessitated the `throws IOException` declaration. Nothing in OkHttp core needs it.

swankjesse: Let’s change this to be an UnsupportedOperationException?

swankjesse: ... it already is!

okhttp-android-support/src/test/java/com/squareup/okhttp/internal/huc/ResponseCacheTest.java
```diff
@@ -1866,8 +1866,9 @@ private void assertGatewayTimeout(HttpURLConnection connection) throws IOExcepti
         HttpURLConnection httpURLConnection = (HttpURLConnection) connection;
         assertEquals(server.url("/").url(), uri.toURL());
         assertEquals(200, httpURLConnection.getResponseCode());
+        InputStream is = httpURLConnection.getInputStream();
         try {
-          httpURLConnection.getInputStream();
```
swankjesse: I think this might have just worked?

JakeWharton: It failed on `fail()` without.

swankjesse: what about just having `getInputStream` throw UOE ?

JakeWharton: That fails 100s of tests since it is now eagerly opened above. The alternative is a `Source` that lazily opens the underlying `InputStream` on first access and then forwards all calls.

okhttp-android-support/src/main/java/com/squareup/okhttp/internal/huc/JavaApiConverter.java
swankjesse: Got it. It’s because now `createOkBody` delegates here.

swankjesse:
So eager to nuke the twisty maze of HttpURLConnection support.

Issue #2106, opened 2015-12-15
Nuke Response#uri() from public API.


JakeWharton:
`Response#uri()` is a weird one! With `HttpUrl`, one can call `url().uri()` to get a `URI`, but the semantics of `Request#uri()` are that it throws `IOException` when the underlying conversion throws `IllegalStateException`. This is an unfortunate behavior to expose as public API.

PR #2115, opened 2015-12-16
Drop Request.uri(). Just use Request.url().uri().

swankjesse:
There's a mostly-academic corner case on URLs like http://host/%xx that
have a malformed escape sequence. This changes the exceptions on those
URLs from checked to unchecked.

I think we may want a separate change to HttpUrl to fix those URLs to
be encoded as http://host/%25xx which avoids the problem altogether.

Closes: https://github.com/square/okhttp/issues/2106

### DIFF
--- a/okhttp/src/main/java/com/squareup/okhttp/Request.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/Request.java
@@ -16,7 +16,6 @@
 package com.squareup.okhttp;
 
 import com.squareup.okhttp.internal.http.HttpMethod;
-import java.io.IOException;
 import java.net.URI;
 import java.net.URL;
 import java.util.List;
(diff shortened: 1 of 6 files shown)


JakeWharton:
Sounds good to me.

okhttp/src/main/java/com/squareup/okhttp/Request.java
```diff
@@ -47,15 +46,6 @@ public HttpUrl url() {
     return url;
   }
 
-  public URI uri() throws IOException {
-    try {
-      URI result = javaNetUri;
```
ChristianBecker: `javaNetUri` is now unused and can be dropped or am I missing something?

swankjesse: Yes!

PR #2118, opened 2015-12-16
Rename com.squareup.okhttp to okhttp3

swankjesse:
Maven group changes will come in follow up.


JakeWharton:
LGTM

swankjesse:
Ugh, IntelliJ struggles a bit retaining static imports on changes like this one. Fixed most of 'em manually.

okhttp/src/main/java/okhttp3/internal/http/CacheStrategy.java
```diff
@@ -59,12 +57,12 @@ public static boolean isCacheable(Response response, Request request) {
...
       case HTTP_REQ_TOO_LONG:
       case HTTP_NOT_IMPLEMENTED:
-      case HTTP_PERM_REDIRECT:
+      case StatusLine.HTTP_PERM_REDIRECT:
       // These codes can be cached unless headers forbid it.
       break;
 
       case HTTP_MOVED_TEMP:
-      case HTTP_TEMP_REDIRECT:
+      case StatusLine.HTTP_TEMP_REDIRECT:
```
JakeWharton: Two missed here.

PR #2124, opened 2015-12-17
Change groupid to com.squareup.okhttp3.

swankjesse:


mockwebserver/pom.xml
```diff
@@ -14,18 +14,18 @@
 
   <dependencies>
     <dependency>
-      <groupId>com.squareup.okhttp</groupId>
+      <groupId>com.squareup.okhttp3</groupId>
```
JakeWharton: If you care to, all these dependencies on sibling artifacts can use `${project.groupId}`.

swankjesse: Seems like a win. Will save me a few minutes in 2025 when we do okhttp4 !

JakeWharton:
LGTM after that change

mockwebserver/pom.xml
swankjesse: Done!

PR #2126, opened 2015-12-17
One authenticate method on the authenticator interface.

swankjesse:
Instead of having one instance of Authenticator, and that interface
defines two methods, we now have two instances of Authenticator and
that interface defines one method.

This prevents leaking request-specific configuration (the authenticator)
into shared connection configuration (the proxy authenticator).

### DIFF
--- a/okhttp/src/main/java/okhttp3/OkHttpClient.java
+++ b/okhttp/src/main/java/okhttp3/OkHttpClient.java
@@ -15,14 +15,6 @@
  */
 package okhttp3;
 
-import okhttp3.internal.Internal;
-import okhttp3.internal.InternalCache;
-import okhttp3.internal.RouteDatabase;
-import okhttp3.internal.Util;
-import okhttp3.internal.http.AuthenticatorAdapter;
-import okhttp3.internal.http.StreamAllocation;
-import okhttp3.internal.io.RealConnection;
-import okhttp3.internal.tls.OkHostnameVerifier;
 import java.net.CookieHandler;
 import java.net.MalformedURLException;
 import java.net.Proxy;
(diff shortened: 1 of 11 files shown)


okhttp/src/main/java/okhttp3/OkHttpClient.java
```diff
@@ -400,6 +402,22 @@ public Authenticator getAuthenticator() {
   }
 
   /**
+   * Sets the authenticator used to respond to challenges from proxy servers. Use {@link
+   * #setAuthenticator} to set the authenticator for origin servers.
+   *
+   * <p>If unset, the {@link java.net.Authenticator#setDefault system-wide default} authenticator
```
JakeWharton: linkplain

okhttp/src/main/java/okhttp3/OkHttpClient.java
```diff
@@ -384,11 +386,11 @@ public CertificatePinner getCertificatePinner() {
...
 
   /**
-   * Sets the authenticator used to respond to challenges from the remote web
-   * server or proxy server.
+   * Sets the authenticator used to respond to challenges from origin servers. Use {@link
+   * #setProxyAuthenticator} to set the authenticator for proxy servers.
    *
-   * <p>If unset, the {@link java.net.Authenticator#setDefault system-wide default}
-   * authenticator will be used.
+   * <p>If unset, the {@link java.net.Authenticator#setDefault system-wide default} authenticator
```
JakeWharton: linkplain

JakeWharton:
LGTM

okhttp/src/main/java/okhttp3/OkHttpClient.java
swankjesse: Fixed.

okhttp/src/main/java/okhttp3/OkHttpClient.java
swankjesse: Fixed.

PR #2128, opened 2015-12-18
Querying the response content-length never does I/O.

JakeWharton:


swankjesse:
... And if it ever had to, it could return -1.

PR #2133, opened 2015-12-19
twitter auth for okcurl

yschimke:
Early review to see if this fits within the okhttp repo and get comments on the approach etc.  

I briefly checked the code against two similar Apache licensed joauth clients

https://github.com/twitter/hbc
https://github.com/yusuke/twitter4j

Works for a few simple GET requests and POST with form bodies.  

Currently only reads the credentials from existing https://github.com/twitter/twurl config file ~/.twurlrc

*Todo in this pull request*

- tests
- sensible error handling e.g. .twurlrc missing
- javadoc
- give example supported requests, also useful to find edge cases

*Later features*

- support alternate credentials store e.g. Keychain on osx
- Implement authentication flow using "OkHttp Test Client" by default and optionally other clients.  e.g. open browser for user login and prompt user to enter the PIN.


swankjesse:
I think this is interesting, but it’s far beyond the scope of okcurl. Maybe create your own GitHub project for this feature?

yschimke:
What about the Authentication Interceptor?  I'm happy to default to putting it into an okhttp-twitter module in a new repo.

Would you be against adding generic hooks to okcurl to allow using authenticators defined by external projects?  There are some improvements I'd like to add to okcurl that are generic, even if you have different backends to talk to.

I'm used to have tools like this for testing APIs, okcurl seems like a promising starting point.

swankjesse:
How'd you feel about configuring an OkHttpClient instance just add you like it, then having okcurl use that instance? We'd just need a mechanism to manage the command line parameters with a prepared OkHttpClient instance.

yschimke:
Sounds good.  Some of the other changes I was thinking of adding were support in okcurl for the Cache and putting the right JARs on the boot classpath for SPDY or HTTP/2 support based on your environment.

Essentially I'd like to be able reproduce the behaviour that mobile clients would see from the command line.

Are these significant enough that I should do them in a separate okcurl, let's call it "greatcurl" for now. :)

PR #2137, opened 2015-12-20
Attempt an alternate route when a connect times out.

swankjesse:
This changes the new allocations model to keep the same route unless it
has been specifically found to have connectivity problems. Previously we
attempted to null out the route selector when a route succeeded; with
this we do the opposite by nulling out a route when it fails.

Closes: https://github.com/square/okhttp/issues/1736

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/internal/http/RecordingProxySelector.java
+++ b/okhttp-tests/src/test/java/okhttp3/internal/http/RecordingProxySelector.java
@@ -28,9 +28,9 @@ import java.util.List;
 import static org.junit.Assert.assertEquals;
 
 public final class RecordingProxySelector extends ProxySelector {
-  final List<URI> requestedUris = new ArrayList<>();
-  List<Proxy> proxies = new ArrayList<>();
-  final List<String> failures = new ArrayList<>();
+  public final List<Proxy> proxies = new ArrayList<>();
+  public final List<URI> requestedUris = new ArrayList<>();
+  public final List<String> failures = new ArrayList<>();
 
   @Override public List<Proxy> select(URI uri) {
     requestedUris.add(uri);
--- a/okhttp/src/main/java/okhttp3/internal/http/Http1xStream.java
+++ b/okhttp/src/main/java/okhttp3/internal/http/Http1xStream.java
@@ -16,6 +16,9 @@
 
 package okhttp3.internal.http;
 
+import java.io.EOFException;
+import java.io.IOException;
+import java.net.ProtocolException;
 import okhttp3.Headers;
 import okhttp3.Request;
 import okhttp3.Response;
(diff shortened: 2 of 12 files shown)


okhttp-tests/src/test/java/okhttp3/internal/http/RecordingProxySelector.java
```diff
@@ -28,9 +28,9 @@
 import static org.junit.Assert.assertEquals;
 
 public final class RecordingProxySelector extends ProxySelector {
-  final List<URI> requestedUris = new ArrayList<>();
-  List<Proxy> proxies = new ArrayList<>();
-  final List<String> failures = new ArrayList<>();
+  public List<Proxy> proxies = new ArrayList<>();
```
JakeWharton: nit: `final`

okhttp/src/main/java/okhttp3/internal/http/Http1xStream.java
```diff
@@ -406,9 +397,8 @@ public FixedLengthSource(long length) throws IOException {
     @Override public void close() throws IOException {
       if (closed) return;
 
-      if (bytesRemaining != 0
-          && !Util.discard(this, DISCARD_STREAM_TIMEOUT_MILLIS, MILLISECONDS)) {
-        unexpectedEndOfInput();
+      if (bytesRemaining != 0 && !Util.discard(this, DISCARD_STREAM_TIMEOUT_MILLIS, MILLISECONDS)) {
+        endOfInput(false);
```
JakeWharton: Shouldn't this be
```java
endOfInput(Util.discard(this, ..., ...));
```
so that successful consumption of remaining bytes allows connection re-use?

okhttp/src/main/java/okhttp3/internal/http/Http1xStream.java
```diff
@@ -463,14 +453,14 @@ private void readChunkSize() throws IOException {
...
-        endOfInput();
+        endOfInput(true);
       }
     }
 
     @Override public void close() throws IOException {
       if (closed) return;
       if (hasMoreChunks && !Util.discard(this, DISCARD_STREAM_TIMEOUT_MILLIS, MILLISECONDS)) {
-        unexpectedEndOfInput();
+        endOfInput(false);
```
JakeWharton: Ditto

JakeWharton:
LGTM

okhttp-tests/src/test/java/okhttp3/internal/http/RecordingProxySelector.java
swankjesse: Fixed.

okhttp/src/main/java/okhttp3/internal/http/Http1xStream.java
swankjesse: When it discards successfully, exhausting the stream does the right thing.

okhttp/src/main/java/okhttp3/internal/http/Http1xStream.java
swankjesse: Ditto.

okhttp/src/main/java/okhttp3/internal/http/Http1xStream.java
JakeWharton: Is there a test for that? Could we add one? Only because it seems prone to breakage as highly-underused behavior (the automatic exhausting, that is).

swankjesse: Yeah, there’s several tests for it.

PR #2139, opened 2015-12-22
Add TlsVersion to the Handshake.

swankjesse:
This is awkward because it adds another field to the cache. I've
decided to try to keep old values in the cache working. We will
probably do a change that'll disrupt the cache later, but that
is invasive and we don't need that yet.

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/RecordedResponse.java
+++ b/okhttp-tests/src/test/java/okhttp3/RecordedResponse.java
@@ -86,6 +86,7 @@ public final class RecordedResponse {
 
   public RecordedResponse assertHandshake() {
     Handshake handshake = response.handshake();
+    assertNotNull(handshake.tlsVersion());
     assertNotNull(handshake.cipherSuite());
     assertNotNull(handshake.peerPrincipal());
     assertEquals(1, handshake.peerCertificates().size());
--- a/okhttp/src/main/java/okhttp3/Handshake.java
+++ b/okhttp/src/main/java/okhttp3/Handshake.java
@@ -16,7 +16,6 @@
 
 package okhttp3;
 
-import okhttp3.internal.Util;
 import java.security.Principal;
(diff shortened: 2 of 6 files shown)


JakeWharton:
This is currently unused, right? I don't see anywhere that passes an instance beyond the cache (which reads it).

LGTM otherwise.

swankjesse:
Yep, I could add tests to confirm that this is available beyond the cache.

okhttp-tests/src/test/java/okhttp3/RecordedResponse.java
```diff
@@ -86,6 +86,7 @@ public RecordedResponse assertBody(String expectedBody) {
 
   public RecordedResponse assertHandshake() {
     Handshake handshake = response.handshake();
+    assertNotNull(handshake.tlsVersion());
```
swankjesse: new!

okhttp/src/main/java/okhttp3/Handshake.java
```diff
@@ -49,6 +51,10 @@ public static Handshake get(SSLSession session) {
     if (cipherSuiteString == null) throw new IllegalStateException("cipherSuite == null");
     CipherSuite cipherSuite = CipherSuite.forJavaName(cipherSuiteString);
 
+    String tlsVersionString = session.getProtocol();
+    if (tlsVersionString == null) throw new IllegalStateException("tlsVersion == null");
+    TlsVersion tlsVersion = TlsVersion.forJavaName(tlsVersionString);
```
JakeWharton: Oh, derp. It's right here. I was missing this usage of it (and propagation to the ctor).

okhttp-tests/src/test/java/okhttp3/RecordedResponse.java
JakeWharton: :+1: thanks. I commented where I failed to :eyes: the usage.

PR #2149, opened 2015-12-24
Sort imports and wrap documentation at 100 columns.

swankjesse:
This is a mechanical refactoring. The only thing I had trouble with is IntelliJ's
weak support for `<ul>` tags in Javadoc.

### DIFF
--- a/checkstyle.xml
+++ b/checkstyle.xml
@@ -82,7 +82,15 @@
     <module name="ParenPad"/>
     <module name="TypecastParenPad"/>
     <module name="WhitespaceAfter"/>
-    <module name="WhitespaceAround"/>
+    <module name="WhitespaceAround">
+      <property name="tokens"
+          value="ASSIGN, BAND, BAND_ASSIGN, BOR, BOR_ASSIGN, BSR, BSR_ASSIGN, BXOR, BXOR_ASSIGN,
+          COLON, DIV, DIV_ASSIGN, DO_WHILE, EQUAL, GE, GT, LAND, LCURLY, LE, LITERAL_CATCH,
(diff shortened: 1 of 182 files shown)


swankjesse:
This exceeded GitHub’s easy-to-review threshold so I moved `samples` and `mockwebserver` to a 2nd PR: https://github.com/square/okhttp/pull/2150

checkstyle.xml
```diff
@@ -82,7 +82,15 @@
     <module name="ParenPad"/>
     <module name="TypecastParenPad"/>
     <module name="WhitespaceAfter"/>
-    <module name="WhitespaceAround"/>
+    <module name="WhitespaceAround">
+      <property name="tokens"
```
swankjesse: Our IntelliJ code style doesn’t put a space before the `}` in this annotation:

```
  @Option(name = {"-X", "--request"}, description = "Specify request command to use")
  public String method;
```

So here I manually remove `RCURLY` to make checkstyle happy. Sigh!

okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
```diff
@@ -49,8 +48,8 @@
     NONE,
     /**
      * Logs request and response lines.
-     * <p>
```
swankjesse: When `<p>` is on its own line, IntelliJ thinks the line break is accidental and it collapses everything into a single (wrapped) line. With a blank line between the paragraphs, IntelliJ autoformatting preserves that blank line.

PR #2155, opened 2015-12-24
New CookieJar interface.

swankjesse:
No good first-party implementation yet.

There is an implementation in okhttp-urlconnection that delegates to the java.net
CookieHandler. That should be good enough for AOSP.

### DIFF
--- a/okhttp-urlconnection/src/main/java/okhttp3/internal/JavaNetCookieJar.java
+++ b/okhttp-urlconnection/src/main/java/okhttp3/internal/JavaNetCookieJar.java
@@ -1,0 +1,86 @@
+/*
+ * Copyright (C) 2015 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okhttp3.internal;
+
+import java.io.IOException;
+import java.net.CookieHandler;
+import java.util.List;
+import java.util.Map;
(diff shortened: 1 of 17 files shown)


okhttp-urlconnection/src/main/java/okhttp3/internal/JavaNetCookieJar.java
```diff
@@ -0,0 +1,86 @@
...
+import java.net.CookieHandler;
+import java.util.List;
+import java.util.Map;
+import java.util.logging.Level;
+import okhttp3.CookieJar;
+import okhttp3.Headers;
+import okhttp3.HttpUrl;
+
+/** A cookie jar that delegates to a {@link java.net.CookieHandler}. */
+public final class JavaNetCookieJar implements CookieJar {
```
swankjesse: Should this be a public type in `okhttp3` ? It’d probably be deprecated, but it’ll give cookie users _something_ while they wait for us to actually implement our own cookie jar.

I don’t want to make this implementation the default because it has to use awkward types like the “multimap” to get anything done.

okhttp/src/main/java/okhttp3/internal/http/HttpEngine.java
```diff
@@ -498,17 +496,13 @@ private Request networkRequest(Request request) throws IOException {
...
-      // This is mostly to stay close to the RI; it is unlikely any of the headers above would
-      // affect cookie choice besides "Host".
-      Map<String, List<String>> headers = OkHeaders.toMultimap(result.build().headers(), null);
-
-      Map<String, List<String>> cookies = cookieHandler.get(request.url().uri(), headers);
-
-      // Add any new cookies to the request.
-      OkHeaders.addCookies(result, cookies);
+    CookieJar cookieJar = client.getCookieJar();
+    if (cookieJar != CookieJar.NO_COOKIES) {
```
JakeWharton: Nice optimization.

okhttp-urlconnection/src/main/java/okhttp3/internal/JavaNetCookieJar.java
JakeWharton: Yeah very tough call. It's not strictly part of the HttpURLConnection stuff, but it also somewhat is.

The only other alternative I can think of is package scoped here and standalone version in recipes/samples.

JakeWharton:
LGTM

PR #2163, opened 2015-12-27
Fixed URLs to 'Full Source' for Get and Post examples in index.html.

twbarber:
Links no longer lead to 'Not Found' message.


swankjesse:
Thanks!

PR #2164, opened 2015-12-28
Add newBuilder overload for resolving directly to a builder.

### LINKED ISSUE #2063: Ability to either call resolve() on a HttpUrl.Builder -or- an overload of resolve() that returns an HttpUrl.Builder

JakeWharton:
Closes #2063.

### DIFF
--- a/okhttp/src/main/java/okhttp3/HttpUrl.java
+++ b/okhttp/src/main/java/okhttp3/HttpUrl.java
@@ -597,11 +597,13 @@ public final class HttpUrl {
     return fragment;
   }
 
-  /** Returns the URL that would be retrieved by following {@code link} from this URL. */
+  /**
+   * Returns the URL that would be retrieved by following {@code link} from this URL, or null if
+   * the resulting URL is not well-formed.
+   */
   public HttpUrl resolve(String link) {
-    Builder builder = new Builder();
-    Builder.ParseResult result = builder.parse(this, link);
-    return result == Builder.ParseResult.SUCCESS ? builder.build() : null;
+    Builder builder = newBuilder(link);
+    return builder != null ? builder.build() : null;
   }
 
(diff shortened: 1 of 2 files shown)


okhttp/src/main/java/okhttp3/HttpUrl.java
```diff
@@ -620,6 +622,17 @@ public Builder newBuilder() {
   }
 
   /**
+   * Returns a builder for the URL that would be retrieved by following {@code link} from this URL,
+   * allowing it to be manipulated further. Equivalent and more efficient than calling
+   * {@link #resolve(String) resolve(link).newBuilder()}.
```
swankjesse: Not sure I love the circular definition!

okhttp/src/main/java/okhttp3/HttpUrl.java
```diff
@@ -597,11 +597,13 @@ public String fragment() {
     return fragment;
   }
 
-  /** Returns the URL that would be retrieved by following {@code link} from this URL. */
+  /**
+   * Returns the URL that would be retrieved by following {@code link} from this URL. Equivalent
```
swankjesse: It was missing before, but maybe mention that this returns `null` if the link cannot be resolved?

swankjesse:
LGTM

okhttp/src/main/java/okhttp3/HttpUrl.java
JakeWharton: Removed. Wasn't really attached to it.

okhttp/src/main/java/okhttp3/HttpUrl.java
JakeWharton: Added to both `resolve` and `newBuilder` with the same wording as `parse`.

Issue #2165, opened 2015-12-28
ConnectionPoolTest.inUseConnectionsNotEvicted() is flaky


swankjesse:
Also `leakedAllocation()`.

```
Tests run: 5, Failures: 0, Errors: 2, Skipped: 0, Time elapsed: 0.093 sec <<< FAILURE! - in okhttp3.ConnectionPoolTest
inUseConnectionsNotEvicted(okhttp3.ConnectionPoolTest)  Time elapsed: 0.009 sec  <<< ERROR!
java.lang.NullPointerException: null
	at okhttp3.internal.http.StreamAllocation.routeDatabase(StreamAllocation.java:204)
	at okhttp3.internal.http.StreamAllocation.<init>(StreamAllocation.java:90)
	at okhttp3.ConnectionPoolTest.inUseConnectionsNotEvicted(ConnectionPoolTest.java:84)

leakedAllocation(okhttp3.ConnectionPoolTest)  Time elapsed: 0 sec  <<< ERROR!
java.lang.NullPointerException: null
	at okhttp3.internal.http.StreamAllocation.routeDatabase(StreamAllocation.java:204)
	at okhttp3.internal.http.StreamAllocation.<init>(StreamAllocation.java:90)
	at okhttp3.ConnectionPoolTest.allocateAndLeakAllocation(ConnectionPoolTest.java:175)
	at okhttp3.ConnectionPoolTest.leakedAllocation(ConnectionPoolTest.java:164)
```

https://github.com/square/okhttp/pull/2162

swankjesse:
Fixed with this:
```
  static {
    Internal.initializeInstanceForTests();
  }
```

PR #2167, opened 2015-12-28
Finish implementing RFC 6265 cookies.

swankjesse:


swankjesse:
Interestingly, the tests found a difference between JDK7 and JDK8. On JDK7 `example.com.` is converted to `example.com` by the IDN mechanism. (We don’t want this.)

So big win for having Travis running multiple versions of the JDK.

swankjesse:
Failure is https://github.com/square/okhttp/issues/2165

okhttp/src/main/java/okhttp3/internal/Util.java
```diff
@@ -371,4 +387,50 @@ public static int delimiterOffset(String input, int pos, int limit, char delimit
...
+    for (int i = 0; i < hostnameAscii.length(); i++) {
+      char c = hostnameAscii.charAt(i);
+      // The WHATWG Host parsing rules accepts some character codes which are invalid by
+      // definition for OkHttp's host header checks (and the WHATWG Host syntax definition). Here
+      // we rule out characters that would cause problems in host headers.
+      if (c <= '\u001f' || c >= '\u007f') {
+        return true;
+      }
+      // Check for the characters mentioned in the WHATWG Host parsing spec:
+      // U+0000, U+0009, U+000A, U+000D, U+0020, "#", "%", "/", ":", "?", "@", "[", "\", and "]"
```
JakeWharton: Not seeing the first four items from this list in the check.

JakeWharton: Though this existed before...

swankjesse: (the first four are in `0x0000...0x001f` above.)

swankjesse: These two methods moved from HttpUrl.

JakeWharton:
:+1:

PR #2169, opened 2015-12-28
Dropped ProxySelectors.

iNoles:
This is for OkHttp 3.x backwards-incompatible API changes.


swankjesse:
Uh oh. I think I want to keep proxy selectors in OkHttp 3 actually. I fear we need em to deal with the proxy configured in the OS.

PR #2173, opened 2015-12-29
Give cookies a builder. Constrain dates to <= Dec 31, 9999.

swankjesse:


okhttp/src/main/java/okhttp3/Cookie.java
```diff
@@ -85,7 +107,7 @@ public boolean persistent() {
 
   /**
    * Returns the time that this cookie expires, in the same format as {@link
-   * System#currentTimeMillis()}. This is {@link Long#MAX_VALUE} if the cookie is not persistent, in
+   * System#currentTimeMillis()}. This is December 31, 9999 if the cookie is not persistent, in
```
JakeWharton: How is calling code supposed to check against this value? Shouldn't the constant be on this type?

okhttp/src/main/java/okhttp3/Cookie.java
```diff
@@ -403,7 +427,120 @@ private static String parseDomain(String s) {
...
+
+    public Builder expiresAt(long expiresAt) {
+      if (expiresAt <= 0) expiresAt = Long.MIN_VALUE;
+      if (expiresAt > HttpDate.MAX_DATE) expiresAt = HttpDate.MAX_DATE;
+      this.expiresAt = expiresAt;
+      this.persistent = true;
+      return this;
+    }
+
+    public Builder domain(String domain, boolean hostOnly) {
```
JakeWharton: Worth providing two overloads instead of a boolean?

swankjesse: I was definitely tempted. Can do.

okhttp/src/main/java/okhttp3/Cookie.java
swankjesse: The calling code should just check persistent or not.

swankjesse: (ie. `persistent()`)

JakeWharton: Oh, got it.

JakeWharton:
LGTM

PR #2176, opened 2015-12-30
Change CookieJar to deal with Cookies, not Headers.

swankjesse:
There's a big consequence to this for compatible with RFC 2965 Cookies. In particular,
this drops our ability to support 'Set-Cookie2' headers, and some features used there
including quoted attributes like 'Max-Age="25"'. This is the right move for interop with
browsers, and likely to make things better for application developers, but some
people who strictly followed the RFC 2965 spec will be broken. (That spec was never
seriously adopted anywhere, which is the entire motivation of RFC 6265.)

The upside is the CookieJar interface is now much more straightforward. This is
particularly good going forward, and for clients who don't have to worry about the
strangeness of things like $Version=1 in the RFC 2965 spec.

### DIFF
--- a/okhttp-testing-support/src/main/java/okhttp3/RecordingCookieJar.java
+++ b/okhttp-testing-support/src/main/java/okhttp3/RecordingCookieJar.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright (C) 2015 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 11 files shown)


swankjesse:
More build fails due to incompatibilities in cookie handing in Java 7 vs. Java 8. I’m starting to feel vindicated in getting OkHttp away from those APIs.

okhttp-testing-support/src/main/java/okhttp3/RecordingCookieJar.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+
+import static org.junit.Assert.assertEquals;
+
+public final class RecordingCookieJar implements CookieJar {
+  private final List<List<Cookie>> requestCookies = new ArrayList<>();
+  private final List<List<Cookie>> responseCookies = new ArrayList<>();
```
JakeWharton: Not that it _really_ matters, but both of these would be better served as `ArrayDeque`s

swankjesse: Yeah, that’s more semantic. Fixed.

adriancole:
Thanks so much. I did feel the more RFC-correct impl would feel deterring for developers unfamiliar or unaffected by nuances around that. This implementation is far more accessible.. much appreciated!

PR #2177, opened 2015-12-31
Refactor towards an OkHttpClient.Builder.

swankjesse:
Still some work to do on method naming, but this was the interesting part.
Also very much need to rename OkUrlFactory in tests to urlFactory, otherwise
the client.client stuff becomes madness.


okhttp-ws/src/main/java/okhttp3/ws/WebSocketCall.java
```diff
@@ -66,9 +66,10 @@ public static WebSocketCall create(OkHttpClient client, Request request) {
 
     // Copy the client. Otherwise changes (socket factory, redirect policy,
     // etc.) may incorrectly be reflected in the request when it is executed.
```
JakeWharton: comment should be removed

JakeWharton:
LGTM

Issue #2151, opened 2015-12-24
java.util.NoSuchElementException


sfsheng0322:
java.util.NoSuchElementException
	at com.squareup.okhttp.internal.http.RouteSelector.next(RouteSelector.java:79)
	at com.squareup.okhttp.internal.http.StreamAllocation.findConnection(StreamAllocation.java:164)
	at com.squareup.okhttp.internal.http.StreamAllocation.findHealthyConnection(StreamAllocation.java:126)
	at com.squareup.okhttp.internal.http.StreamAllocation.newStream(StreamAllocation.java:95)
	at com.squareup.okhttp.internal.http.HttpEngine.connect(HttpEngine.java:283)
	at com.squareup.okhttp.internal.http.HttpEngine.sendRequest(HttpEngine.java:224)
	at com.squareup.okhttp.Call.getResponse(Call.java:286)
	at com.squareup.okhttp.Call$ApplicationInterceptorChain.proceed(Call.java:243)
	at com.squareup.okhttp.Call.getResponseWithInterceptorChain(Call.java:205)
	at com.squareup.okhttp.Call.access$100(Call.java:35)
	at com.squareup.okhttp.Call$AsyncCall.execute(Call.java:171)
	at com.squareup.okhttp.internal.NamedRunnable.run(NamedRunnable.java:33)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1112)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:587)
	at java.lang.Thread.run(Thread.java:818)

Can you help me to find why it is happened?
Thanks!

swankjesse:
Two possibilities:

 * OkHttp has a bug where we’re attempting to recover when we shouldn't be because we’ve exhausted all of the available routes.
 * OkHttp has a bug where you’ve got a custom DNS implementation that returned zero IP addresses for a host and we didn't reject that.

swankjesse:
Please try OkHttp 2.7.1 which has a targeted fix for this problem.

swankjesse:
I believe we have this fixed with 2.7.1. Will reopen if that fix wasn't sufficient.

sfsheng0322:
2.6 is Ok, I will try 2.7.1 later, thanks very much! Merry Christmas！

PR #2179, opened 2015-12-31
Don't do health checks on brand new connections.

swankjesse:
This is imperfect, but it should save some unnecessary work and will
hopefully prevent RouteSelector from attempting a route when none is
available.

https://github.com/square/okhttp/issues/2151

### DIFF
--- a/okhttp/src/main/java/com/squareup/okhttp/internal/http/StreamAllocation.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/internal/http/StreamAllocation.java
@@ -123,11 +123,21 @@ public final class StreamAllocation {
       int writeTimeout, boolean connectionRetryEnabled, boolean doExtensiveHealthChecks)
       throws IOException, RouteException {
     while (true) {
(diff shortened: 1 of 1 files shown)


swankjesse:
Note that this is a PR for OkHttp 2.7. Once merged I’ll cut a 2.7.1.

tankjj:
I found this patch may cause endless loop:
1. on a mobile phone, user sends one request, in function StreamAllocation::findHealthyConnection(), it returns a candidate
2. if this candidate successCount is 0, it is returned without health check.(candidate.isHealthy())
3. switch the phone's netowrk from 4G to wifi, so the socket is broken, the candidate is unhealthy.
4. to send request, SocketException(socket is closed) is thrown to RetryAndFollowUpInterceptor, and will have one more retry.
5.  StreamAllocation::findHealthyConnection() returns the same candidate.
...
so it causes endless loop.

swankjesse:
Yikes. Can fix!

swankjesse:
https://github.com/square/okhttp/issues/5547

PR #2180, opened 2016-01-01
Don't do health checks on brand new connections.

swankjesse:
This is imperfect, but it should save some unnecessary work and will
hopefully prevent RouteSelector from attempting a route when none is
available.

https://github.com/square/okhttp/issues/2151


swankjesse:
I’m going to just merge this cherry-pick! Original PR is https://github.com/square/okhttp/pull/2179

JakeWharton:
SGTM

PR #2183, opened 2016-01-01
Restore no-args constructor for OkHttpClient.

swankjesse:
This is a popular API and keeping it working is handy.


mboukadir:
I just want to draw your attention to the fact that putting the constructor okhttpClient public that allows developers to write:
`return new OkhttpClient().newBuilder().addInterceptor(headerInterceptor).build()`

I discovered this piece of code in a project that is already in production.

yschimke:
@mboukadir This API is already there and changing it would be a big breaking change.  Also if you follow the advice, the cost of creating that initial client is so tiny anyway.

But just generally that line is equivalent to something that *IS* a good pattern

```
sharedClient = new OkhttpClient()
...
perServiceClient = shared.newBuilder().addInterceptor(headerInterceptor).build()
```

PR #2185, opened 2016-01-01
Detect API use errors earlier.

swankjesse:
Previously we would accept invalid input, but crash when that invalid input
was used.

### DIFF
--- a/okhttp-android-support/src/main/java/okhttp3/internal/huc/JavaApiConverter.java
+++ b/okhttp-android-support/src/main/java/okhttp3/internal/huc/JavaApiConverter.java
@@ -633,7 +633,7 @@ public final class JavaApiConverter {
       if (position < 0) {
         throw new IllegalArgumentException("Invalid header index: " + position);
       }
-      if (position == 0) {
+      if (position == 0 || position > response.headers().size()) {
         return null;
       }
       return response.headers().name(position - 1);
@@ -647,6 +647,9 @@ public final class JavaApiConverter {
       }
       if (position == 0) {
         return StatusLine.get(response).toString();
+      }
+      if (position > response.headers().size()) {
+        return null;
       }
       return response.headers().value(position - 1);
     }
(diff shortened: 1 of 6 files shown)


okhttp-android-support/src/main/java/okhttp3/internal/huc/JavaApiConverter.java
```diff
@@ -648,6 +648,9 @@ public String getHeaderField(int position) {
       if (position == 0) {
         return StatusLine.get(response).toString();
       }
+      if (position > response.headers().size()) {
```
swankjesse: OkHttp gets strict, but HttpURLConnection stays loose!

okhttp-tests/src/test/java/okhttp3/URLConnectionTest.java
```diff
@@ -35,6 +35,7 @@
 import java.net.URL;
 import java.net.URLConnection;
 import java.net.UnknownHostException;
+import java.security.KeyManagementException;
```
JakeWharton: Unused

okhttp/src/main/java/okhttp3/Headers.java
```diff
@@ -75,20 +75,12 @@ public int size() {
...
   public String name(int index) {
-    int nameIndex = index * 2;
-    if (nameIndex < 0 || nameIndex >= namesAndValues.length) {
-      return null;
-    }
-    return namesAndValues[nameIndex];
+    return namesAndValues[index * 2];
   }
 
   /** Returns the value at {@code index} or null if that is out of range. */
```
JakeWharton: Doc needs updated.

okhttp/src/main/java/okhttp3/Headers.java
```diff
@@ -75,20 +75,12 @@ public int size() {
 
   /** Returns the field at {@code position} or null if that is out of range. */
```
JakeWharton: Doc needs updated.

JakeWharton:
LGTM

okhttp/src/main/java/okhttp3/Headers.java
swankjesse: Good good. Fixed

PR #2192, opened 2016-01-04
Make OkHttp OSGi ready

lburgazzoli:
This PR adds basic support for OSGi to okhttp jars by adding the proper meta-data to jar's MANIFEST.

Notes:
- okhttp-urlconnection is not well supported as this artifact has the same packages as main okhttp one but as far as I understand OkHttp as now a better replacement
- it would have be better if okhttp-ws had okhttp3.ws.internal instead of okhttp3.internal.ws so that there is a clear distinction between okhttp-ws internal and okhttp ones

### DIFF
--- a/okhttp-apache/pom.xml
+++ b/okhttp-apache/pom.xml
@@ -9,6 +9,7 @@
     <version>3.0.0-SNAPSHOT</version>
   </parent>
 
+  <packaging>bundle</packaging>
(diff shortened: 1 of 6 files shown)


swankjesse:
Mind squashing into a single commit? http://rebaseandsqua.sh

lburgazzoli:
Done

PR #2195, opened 2016-01-04
Handle unknown content lengths.

### LINKED ISSUE #2194: HttpLoggingInterceptor reports -1-byte body when set to Level.BASIC and response is chunked
The body length of the responses printed by HttpLoggingInterceptor all report `-1-byte body` when set to `Level.BASIC` and the response is `chunked`. They are reported correctly when the logging level is set to `Level.BODY`.

Example:
`HTTP/1.1 200 OK (133ms, -1-byte body)`

JakeWharton:
Closes #2194.

### DIFF
--- a/okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
+++ b/okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
@@ -204,9 +204,11 @@ public final class HttpLoggingInterceptor implements Interceptor {
     long tookMs = TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - startNs);
 
     ResponseBody responseBody = response.body();
+    long contentLength = responseBody.contentLength();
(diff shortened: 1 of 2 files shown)


austynmahoney:
Should it read the chunked response and then print the body length instead of just printing unknown? Seems weird to get a different result if the logger is on `BASIC` or `BODY`. Now every one of my responses just prints unknown length (better than `-1` I guess).

I tried my hand at this fix, but it got messy and I'm not sure if we really want to read the entire response when using `BASIC`.

JakeWharton:
> I'm not sure if we really want to read the entire response when using BASIC.

Yeah I don't think we want to. We enable BASIC in production and route the logs to an in-memory logger. Forcing the entire body to be read and buffered in memory just to get it's size would seriously harm performance.

> Should it read the chunked response.

I initially had it printing the contents of the `Transfer-Encoding` header as a fallback (so you'd get "chunked body", "gzip body", etc.). I removed it because the HTTP spec has a lot of rules about the header that I didn't feel like figuring out right now.

I don't think we'd be opposed to something smarter like that so long as it behaved well within the spec.

PR #2197, opened 2016-01-04
Javadoc fix for immutable Interceptors in OkHttpClient

### LINKED ISSUE #2196: okHttpClient.interceptors() not returning a modifiable list
http://square.github.io/okhttp/3.x/okhttp/okhttp3/OkHttpClient.html#interceptors-- should return a modifiable list according to documentation 

> Returns a modifiable list of interceptors that observe the full span of each call: from before the connection is established (if any) until after the response source is selected (either the origin server, cache, or both).

but it is returning an unmodifiable one.

I'm using version 3.0.0-RC1

austynmahoney:
Fixes #2196


swankjesse:
Thanks!

Issue #2202, opened 2016-01-06
`JavaNetCookieJar` parses cookies incorrectly


floating-cat:
After debugged several times I found:
https://github.com/square/okhttp/blob/7f4ce6fdd75d142819d612196dd45aff0795d682/okhttp-urlconnection/src/main/java/okhttp3/JavaNetCookieJar.java#L87
```
/**
 * Convert a request header to OkHttp's cookies via {@link HttpCookie}. That extra step handles
 * multiple cookies in a single request header, which {@link Cookie#parse} doesn't support.
 */
private List<Cookie> decodeHeaderAsJavaNetCookies(HttpUrl url, String header) {
   List<HttpCookie> javaNetCookies;
  try {
    javaNetCookies = HttpCookie.parse(header);
...
```
misuses [`HttpCookie.parse(String)`](http://developer.android.com/reference/java/net/HttpCookie.html#parse(java.lang.String)).
```
public static List<HttpCookie> parse (String header)

Constructs a cookie from a string. The string should comply with set-cookie
or set-cookie2 header format as specified in RFC 2965. Since set-cookies2
syntax allows more than one cookie definitions in one header, the returned
object is a list.
```

Current the header string was composed of several cookies which can't be parse with `HttpCookie.parse(String)`.
`decodeHeaderAsJavaNetCookies(HttpUrl, String)`'s doc comment details the same thing.

Edit: details more about this issue.

floating-cat:
`HttpCookie.parse(String)` only return a `List` with one item if the header is *set-cookie* or a `List` with more items if the header is *set-cookie2*.

floating-cat:
Also see https://android.googlesource.com/platform/libcore/+/f7ab2bc37debba91864bfec6572a3e7bbe994c58/luni/src/main/java/java/net/CookieManager.java#144
```
return Collections.singletonMap("Cookie", Collections.singletonList(result.toString()));
```

swankjesse:
I’ve added a test that tries to reproduce the problem, but doesn’t. Could you send a test that does?
https://github.com/square/okhttp/pull/2207/files

PR #2207, opened 2016-01-07
Test that JavaNetCookieJar can handle multiple cookies.

swankjesse:
See https://github.com/square/okhttp/issues/2202

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/internal/http/CookiesTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/internal/http/CookiesTest.java
@@ -163,6 +163,28 @@ public class CookiesTest {
     assertEquals("a=android; b=banana", request.getHeader("Cookie"));
   }
 
+  @Test public void receiveAndSendMultipleCookies() throws Exception {
+    MockWebServer server = new MockWebServer();
+    server.enqueue(new MockResponse()
+        .addHeader("Set-Cookie", "a=android")
+        .addHeader("Set-Cookie", "b=banana"));
+    server.enqueue(new MockResponse());
+    server.start();
+
+    CookieManager cookieManager = new CookieManager(null, ACCEPT_ORIGINAL_SERVER);
+    client = client.newBuilder()
+        .cookieJar(new JavaNetCookieJar(cookieManager))
+        .build();
+
+    get(urlWithIpAddress(server, "/"));
+    RecordedRequest request1 = server.takeRequest();
+    assertNull(request1.getHeader("Cookie"));
+
+    get(urlWithIpAddress(server, "/"));
+    RecordedRequest request2 = server.takeRequest();
+    assertEquals("a=android; b=banana", request2.getHeader("Cookie"));
+  }
+
   @Test public void testRedirectsDoNotIncludeTooManyCookies() throws Exception {
     MockWebServer redirectTarget = new MockWebServer();
     redirectTarget.enqueue(new MockResponse().setBody("A"));


JakeWharton:
Looks like your test fails on JDK 7

swankjesse:
Updated the PR. Forgot that Java 7 hates `localhost` and needs `localhost.local` for the cookie domain.

JakeWharton:
LGTM

okhttp-tests/src/test/java/okhttp3/internal/http/CookiesTest.java
```diff
@@ -163,6 +163,28 @@ public void testNetscapeResponse() throws Exception {
...
 
+  @Test public void receiveAndSendMultipleCookies() throws Exception {
+    MockWebServer server = new MockWebServer();
+    server.enqueue(new MockResponse()
+        .addHeader("Set-Cookie", "a=android")
+        .addHeader("Set-Cookie", "b=banana"));
+    server.enqueue(new MockResponse());
+    server.start();
+
+    CookieManager cookieManager = new CookieManager(null, ACCEPT_ORIGINAL_SERVER);
```
floating-cat: This `CookieManager`'s implementation is difference between openJDK and the Android JDK.
openJDK's `CookieManager#get(URI, Map<String, List<String>>)`[1] return a map whose cookie's entry is a List contains each cookie. But the Android JDKs `CookieManager#get(URI, Map<String, List<String>>)`[2] return a map whose cookie's entry is a List with only one string. This string was composed of several cookies.
So If you write a test for openJDK, this test can pass because
```
    // copied from javaNetCookieJar.java#L70
    List<Cookie> cookies = null;
    for (Map.Entry<String, List<String>> entry : cookieHeaders.entrySet()) {
      String key = entry.getKey();
      if (("Cookie".equalsIgnoreCase(key) || "Cookie2".equalsIgnoreCase(key))
          && !entry.getValue().isEmpty()) {
        for (String header : entry.getValue()) {
          if (cookies == null) cookies = new ArrayList<>();
          cookies.addAll(decodeHeaderAsJavaNetCookies(url, header));
        }
      }
    }
```
use `for each` to invoke `decodeHeaderAsJavaNetCookies(HttpUrl, String)` for each cookie. But Android's won't, because Android's only invoke `for each` once, and use `decodeHeaderAsJavaNetCookies(HttpUrl, String)` to parse string which was composed of several cookies.

I guess this test only works in Android if Android N use openJDK's implementation.

[1] http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/8u40-b25/java/net/CookieManager.java/#195
[2] https://android.googlesource.com/platform/libcore/+/f7ab2bc37debba91864bfec6572a3e7bbe994c58/luni/src/main/java/java/net/CookieManager.java#120

floating-cat: A sample test case:
```java
  // pass
  @Test public void testMultipleCookiesParseForOpenJdk() throws Exception {
    // similar to current OpenJDk's CookieManager implementation
    Map<String, List<String>> map = new HashMap<>();
    List<String> list = new ArrayList<>();
    list.add("a=android");
    list.add("b=banana");
    map.put("Cookie", list);
    assertEquals(2 ,parseMultiCookies(map).size());
  }

  // fail
  @Test public void testMultipleCookiesParseForAndroidJDK() throws Exception {
    // similar to current Android JDK's CookieManager implementation
    List<String> list = new ArrayList<>();
    list.add("a=android");
    list.add("b=banana");
    StringBuilder result = new StringBuilder();
    result.append(list.get(0));
    for (int i = 1; i < list.size(); i++) {
      result.append("; ").append(list.get(i));
    }
    Map<String, List<String>> map = Collections.singletonMap("Cookie",
            Collections.singletonList(result.toString()));

    assertEquals(2 ,parseMultiCookies(map).size());
  }

  // similar logic to current JavaNetCookieJar's implementation
  private List<HttpCookie> parseMultiCookies(Map<String, List<String>> map){
    List<HttpCookie> cookies = null;
    for (Map.Entry<String, List<String>> entry : map.entrySet()) {
      String key = entry.getKey();
      if (("Cookie".equalsIgnoreCase(key) || "Cookie2".equalsIgnoreCase(key)) 
              && !entry.getValue().isEmpty()) {
        for (String header : entry.getValue()) {
          if (cookies == null) cookies = new ArrayList<>();
          cookies.addAll(HttpCookie.parse(header));
        }
      }
    }

    return cookies;
  }
```

swankjesse: Got it. Can fix!

PR #2211, opened 2016-01-08
Update okhttp-logging-interceptor README.md

moltak:
Change how to use on okhttp-loging-interceptor's README.md


swankjesse:
Thanks!

Issue #1467, opened 2015-03-04
Hostname cannot be verified in case SSL certificate is issued for IP address


mnasyrov:
OkHttp v2.2.0.

Following exception happens when OkHttp tries to connect to a server using IP address:
```
Caused by: java.io.IOException: Hostname 192.168.1.1 not verified:
    certificate: sha1/5iMPVYXxMn4eoyIuLZCZPWL6myM=
    DN: CN=192.168.1.1
    subjectAltNames: []
	at com.squareup.okhttp.Connection.upgradeToTls(Connection.java:260)
	at com.squareup.okhttp.Connection.connect(Connection.java:158)
	at com.squareup.okhttp.Connection.connectAndSetOwner(Connection.java:174)
	at com.squareup.okhttp.OkHttpClient$1.connectAndSetOwner(OkHttpClient.java:120)
	at com.squareup.okhttp.internal.http.RouteSelector.next(RouteSelector.java:131)
	at com.squareup.okhttp.internal.http.HttpEngine.connect(HttpEngine.java:312)
	at com.squareup.okhttp.internal.http.HttpEngine.sendRequest(HttpEngine.java:235)
	at com.squareup.okhttp.Call.getResponse(Call.java:262)
	at com.squareup.okhttp.Call$ApplicationInterceptorChain.proceed(Call.java:219)
	at com.squareup.okhttp.Call.getResponseWithInterceptorChain(Call.java:192)
	at com.squareup.okhttp.Call.execute(Call.java:79)
	at retrofit.client.OkClient.execute(OkClient.java:53)
	at retrofit.RestAdapter$RestHandler.invokeRequest(RestAdapter.java:326)
	... 67 common frames omitted
```

A server has a self-signed SSL certificate which was generated for server's IP addresses using following commands:
```
$ openssl req -x509 -newkey rsa:2048 -keyout server.key.pem -out server.cert.pem -days 365 -subj /CN=192.168.1.1
$ openssl rsa -in server.key.pem -out server.key.pem
```

As I see in current source codes, OkHostnameVerifier class tries to verify a hostname using only IP addresses in "alternate names" section of SSL certificate. I think it must try to verify a hostname using CN records before.

swankjesse:
Pull requests welcome on this. They should include a test!

ilansas:
+1 on this

swankjesse:
Pull requests welcome on this. They should include a test!

wzpan:
+1

[...]

PR #2214, opened 2016-01-08
Fix OkHostnameVerifier cannot verify SSL certificate issued for IP address

topwu:
Fix OkHostnameVerifier cannot verify SSL certificate issued for IP address, a test is provided.

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/HostnameVerifierTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/HostnameVerifierTest.java
@@ -1,0 +1,283 @@
+/*
+ * Copyright (C) 2013 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


swankjesse:
https://wiki.mozilla.org/CA:Problematic_Practices#Certificates_referencing_hostnames_or_private_IP_addresses

> “It is not standards compliant for printable ASCII representations of IP addresses to be placed in any certificate field that is intended to hold DNS names, including the subject common name and the DNSName field of the Subject Alternative Names extension. There is a place in a certificate specifically intended to be where IP (v4 or v6) addresses may be placed. It is in the Subject Alternative Names extension. The SubjectAltNames extension has places for both additional DNS names and for IP addresses. The place for IP addresses takes them in binary form, not in printable ASCII (e.g. dotted decimal) form.”

swankjesse:
Sorry, this behavior is by design.

topwu:
I though this is also fix the issue [#1467](https://github.com/square/okhttp/issues/1467#issuecomment-152850929)? 
Also [StrictHostnameVerifier](http://grepcode.com/file_/repo1.maven.org/maven2/com.ning/metrics.action/0.2.1/org/apache/http/conn/ssl/StrictHostnameVerifier.java/?v=source) in apache also support IP address.

swankjesse:
It's probably an oversight on the bug. Maybe we can just close it. The subjectAltNames is the right place for IP addresses, not the CN. If Apache does it wrong, that's their bug.

Issue #2216, opened 2016-01-08
SSLContext will be called if ConnectionSpec.CLEARTEXT is only set


iNoles:
When I called like

```java
private static final OkHttpClient client = new OkHttpClient().newBuilder()
            .connectionSpecs(Collections.singletonList(ConnectionSpec.CLEARTEXT))
            .addInterceptor(new GzipRequestInterceptor())
            .build();
```

the private constructor of OkHttpClient will call
```java
if (builder.sslSocketFactory != null) {
      this.sslSocketFactory = builder.sslSocketFactory;
    } else {
      try {
        SSLContext sslContext = SSLContext.getInstance("TLS");
        sslContext.init(null, null, null);
        this.sslSocketFactory = sslContext.getSocketFactory();
      } catch (GeneralSecurityException e) {
        throw new AssertionError(); // The system has no TLS. Just give up.
      }
    }
```

It would be waste of CPU times if any developers are not planning to use it. Is there is a reason to call SSLContext by default?

swankjesse:
That’s an optimization I hadn’t considered! Will accept a pull request for that.

PR #2218, opened 2016-01-09
Fix for #2216, Add Optmizations for ClearText to avoid SSLContext

iNoles:


okhttp/src/main/java/okhttp3/OkHttpClient.java
```diff
@@ -159,7 +159,16 @@ private OkHttpClient(Builder builder) {
     this.cache = builder.cache;
     this.internalCache = builder.internalCache;
     this.socketFactory = builder.socketFactory;
-    if (builder.sslSocketFactory != null) {
+
+    boolean isTLS = true;
+    for (ConnectionSpec spec : connectionSpecs) {
+      if (connectionSpecs.equals(DEFAULT_CONNECTION_SPECS)) {
+        break;
+      }
```
JakeWharton: This check shouldn't be inside the for loop, unless you meant it to be something else.

okhttp/src/main/java/okhttp3/OkHttpClient.java
```diff
@@ -159,7 +159,16 @@ private OkHttpClient(Builder builder) {
...
+      isTLS = spec.isTls();
+    }
+
+    if (builder.sslSocketFactory != null || !isTLS) {
```
JakeWharton: It's slightly unfortunate that you always need to loop even when a factory was specified. Trading one inefficiency for another (small) one.

iNoles:
I am trying to find a better way for it. 

Something like
```java
boolean isTLS = true;
if (!connectionSpecs.equals(DEFAULT_CONNECTION_SPECS)) {
    for (ConnectionSpec spec : connectionSpecs) {
        isTLS = spec.isTls();
    }
}
```

okhttp/src/main/java/okhttp3/OkHttpClient.java
```diff
@@ -159,7 +159,16 @@ private OkHttpClient(Builder builder) {
...
     this.internalCache = builder.internalCache;
     this.socketFactory = builder.socketFactory;
-    if (builder.sslSocketFactory != null) {
+
+    boolean isTLS = true;
+    for (ConnectionSpec spec : connectionSpecs) {
+      if (connectionSpecs.equals(DEFAULT_CONNECTION_SPECS)) {
+        break;
+      }
+      isTLS = spec.isTls();
```
swankjesse: this needs to be `isTls = isTls || spec.isTls()` because TLS and non-TLS can be mixed.

swankjesse:
One thing I don’t like about this is that it optimizes a case I actively want to discourage. Cleartext connections are bad, and I don’t love the idea of having code to support making the worse thing better.

iNoles:
Do I need to rebase this?

swankjesse:
Nope, looks good. I do request you run http://rebaseandsqua.sh to squish this into a single commit.

PR #2220, opened 2016-01-09
Logging Interceptor - Use charset specified by the request body content type

jgilfelt:
It's read but never assigned.


JakeWharton:
Yikes. We should write a test for this to ensure the declared charset is used. Add one of you're feeling icerly generous, otherwise I'll whip one up later or tomorrow.

Issue #2219, opened 2016-01-09
java.lang.UnsupportedOperationException while adding interceptor


outofdate:
3.0.0-RC1
This chunk of code is throwing exception. 
So i found it out when migrating to okhttp3. 

``` java
 client.interceptors().add(new Interceptor() {
            @Override
            public okhttp3.Response intercept(Chain chain) throws IOException {
                Request request = chain.request();
                Request.Builder requestBuilder = request.newBuilder()
                        .addHeader("Authorization", authPrefs.getString(Properties.ApiProperties.TOKEN, "null"))
                        .method(request.method(), request.body());

                request = requestBuilder.build();
                return chain.proceed(request);
            }
        });
```
Version 2.7.0 works fine for me.

artem-zinnatullin:
This part of OkHttp is now immutable, please use OkHttpClient.Builder and
add interceptor when you create OkHttpClient instance.

Something like this:

new OkHttpClient.Builder().addInterceptor(...).build()

On Sat, Jan 9, 2016, 14:19 outofdate <notifications@github.com> wrote:

> 300-RC1
> This chunk of code is throwing exception
> So i found it out when migrating to okhttp3
>
>  clientinterceptors()add(new Interceptor() {
>             @Override
>             public okhttp3Response intercept(Chain chain) throws IOException {
>                 Request request = chainrequest();
>                 RequestBuilder requestBuilder = requestnewBuilder()
>                         addHeader("Authorization", authPrefsgetString(PropertiesApiPropertiesTOKEN, "null"))
>                         method(requestmethod(), requestbody());
>
>                 request = requestBuilderbuild();
>                 return chainproceed(request);
>             }
>         });
>
> Version 270 works fine for me
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/square/okhttp/issues/2219>.
>
-- 

@artem_zin

Issue #2206, opened 2016-01-06
okhttp3:logging-interceptor:3.0.0-RC1 wile adding to retrofit2 okhttp3 client


silva96:
Hi, im using this gradle plugins
```
    compile 'com.squareup.retrofit2:retrofit:2.0.0-beta3'
    compile 'com.squareup.retrofit2:converter-gson:2.0.0-beta3'
    compile 'com.squareup.okhttp3:logging-interceptor:3.0.0-RC1'
```

This is how I create my service:

```
        OkHttpClient client = new OkHttpClient();
        HttpLoggingInterceptor logging = new HttpLoggingInterceptor();
        logging.setLevel(HttpLoggingInterceptor.Level.BODY);
        client.interceptors().add(logging);
        Retrofit retrofit = new Retrofit.Builder()
                .baseUrl(BASE_URL)
                .addConverterFactory(GsonConverterFactory.create())
                .client(client)
                .build();
        APIService service = retrofit.create(APIService.class);

```

It fails when `client.interceptors().add(logging);`

With this strack trace:

```
01-06 09:30:27.944 27005-27005/com.silva.benjamin.ninetyninecoins E/AndroidRuntime: FATAL EXCEPTION: main
                                                                                    Process: com.silva.benjamin.ninetyninecoins, PID: 27005
                                                                                    java.lang.UnsupportedOperationException
                                                                                        at java.util.Collections$UnmodifiableCollection.add(Collections.java:928)
                                                                                        at com.silva.benjamin.ninetyninecoins.util.Helper.service(Helper.java:22)
                                                                                        at com.silva.benjamin.ninetyninecoins.fragments.ConsoleFragment.renderData(ConsoleFragment.java:60)
                                                                                        at com.silva.benjamin.ninetyninecoins.fragments.ConsoleFragment.onCreateView(ConsoleFragment.java:45)
                                                                                        at android.support.v4.app.Fragment.performCreateView(Fragment.java:1962)
                                                                                        at android.support.v4.app.FragmentManagerImpl.moveToState(FragmentManager.java:1067)
                                                                                        at android.support.v4.app.FragmentManagerImpl.moveToState(FragmentManager.java:1248)
                                                                                        at android.support.v4.app.BackStackRecord.run(BackStackRecord.java:738)
                                                                                        at android.support.v4.app.FragmentManagerImpl.execPendingActions(FragmentManager.java:1613)
[...]

PR #2221, opened 2016-01-09
Provide interceptors as list with read-only methods

artem-zinnatullin:
Looks like a lot of projects and their developers will face `UnsupportedOperationException` at runtime after switching from v2 to v3: #2219, #2206.

Would be great if we could turn it into a compile time error.

### DIFF
--- a/okhttp/src/main/java/okhttp3/UnmodifiableList.java
+++ b/okhttp/src/main/java/okhttp3/UnmodifiableList.java
@@ -1,0 +1,47 @@
+package okhttp3;
+
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+
+public final class UnmodifiableList<T> implements Iterable<T> {
+
+  private final List<T> realList;
+
+  public UnmodifiableList(List<T> realList) {
+    this.realList = new ArrayList<>(realList);
+  }
+
+  /**
+   * Returns the number of elements in this list.  If this list contains
+   * more than <tt>Integer.MAX_VALUE</tt> elements, returns
+   * <tt>Integer.MAX_VALUE</tt>.
+   *
+   * @return the number of elements in this list
+   */
+  public int size() {
+    return realList.size();
+  }
+
+  /**
+   * Returns the element at the specified position in this list.
+   *
+   * @param index index of the element to return
+   * @return the element at the specified position in this list
+   * @throws IndexOutOfBoundsException if the index is out of range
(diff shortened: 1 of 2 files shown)


okhttp/src/main/java/okhttp3/UnmodifiableList.java
```diff
@@ -0,0 +1,46 @@
+package okhttp3;
+
+import java.util.Iterator;
+import java.util.List;
+
+public final class UnmodifiableList<T> implements Iterable<T> {
```
artem-zinnatullin: Not a fan of new public class for the project, but hadn't found a better solution yet. But, looks like it can be used not only for interceptors but also for other things too both internally and for the public API.

okhttp/src/main/java/okhttp3/UnmodifiableList.java
```diff
@@ -0,0 +1,46 @@
...
+
+  private final List<T> realList;
+
+  public UnmodifiableList(List<T> realList) {
+    this.realList = realList;
```
JakeWharton: Missing a copy. You're just caching a mutable list that can be modified after wrapping.

okhttp/src/main/java/okhttp3/UnmodifiableList.java
JakeWharton: Opposed in general to this. This class doesn't carry its weight in an API
that we try very hard to minimize.

artem-zinnatullin: Yeah. Initially, I wanted to change `List` to `Iterable` but then found usages with `size()` and `get(index)` in the project, though they can be carefully rewritten.

Maybe you'll find such class useful for other cases when OkHttp uses/provides lists? (It will remove whole class of runtime problems).

Feel free to close this PR if you see a better solution (or it's not a problem).

okhttp/src/main/java/okhttp3/UnmodifiableList.java
artem-zinnatullin: Right, fixed

swankjesse:
Not a fan.

swankjesse:
I suspect if we did this we’d get the same bug reports, only that the new list class doesn't define an `add()` method.

I don't want the first thing in our shiny new API to be an ugly bridge that only serves people coming from our old mutable API.

swankjesse:
Neat idea though!

artem-zinnatullin:
>shiny new API

!

okay :)

PR #2224, opened 2016-01-11
Change default mock reason.

### LINKED ISSUE #2222: MockResponse clas doesn't set the right status string when setStatusCode() is called with non 200 status code
MockResponse class doesn't set the right status string when setStatusCode() is called with non 200 status code

example: 
```
responded: HTTP/1.1 401 OK
```
Ideally the status code should be set as following.
expected:
```
responded: HTTP/1.1 401 UNAUTHORIZED
```
Here is the [implementation of the method](https://github.com/square/okhttp/blob/c9080f14f37f8d85d1c6da648e843a70ee474436/mockwebserver/src/main/java/okhttp3/mockwebserver/MockResponse.java#L71) which should change.

```
  public MockResponse setResponseCode(int code) {
    return setStatus("HTTP/1.1 " + code + " OK");
  }
```

JakeWharton:
Closes #2222.


swankjesse:
LGTM

JakeWharton:
PTAL

PR #2244, opened 2016-01-14
Tweak error copy, add a couple tests

pforhan:


okhttp/src/main/java/okhttp3/Headers.java
```diff
@@ -199,6 +195,26 @@ public static Headers of(Map<String, String> headers) {
...
+        throw new IllegalArgumentException(String.format(
+            "Unexpected char %#04x at %d in header name: %s", (int) c, i, name));
+      }
+    }
+    if (value == null) throw new IllegalArgumentException("value == null");
+    for (int i = 0, length = value.length(); i < length; i++) {
+      char c = value.charAt(i);
+      if (c <= '\u001f' || c >= '\u007f') {
+        throw new IllegalArgumentException(String.format(
+            "Unexpected char %#04x at %d in %s value: %s", (int) c, i, name, value));
```
pforhan: This line is the only one I actually care about... I just want to see the header name in this error message.

okhttp/src/main/java/okhttp3/Headers.java
```diff
@@ -188,9 +186,7 @@ public static Headers of(Map<String, String> headers) {
       }
       String name = header.getKey().trim();
       String value = header.getValue().trim();
-      if (name.length() == 0 || name.indexOf('\0') != -1 || value.indexOf('\0') != -1) {
-        throw new IllegalArgumentException("Unexpected header: " + name + ": " + value);
-      }
+      checkNameAndValue(name, value);
```
pforhan: It seems like this is right, stuff should use Builder.addLenient() if they don't want validation.  But was there another reason not to fully validate?

pforhan:
Darn builds are tuncated so I don't know what the actual failures are.  I'm guessing someone *does* use of() expecting lenient behavior.

pforhan:
It is MultipartBodyTest failing on stringEscapingIsWeird.  I'll just drop those changes.

pforhan:
Build passes locally, not sure how to tell travis to retry.

PR #2251, opened 2016-01-16
Use listeners to observe allocation limit changes.

swankjesse:
This removes some awkward code where the synchronized connection pool is calling
a synchronized methond on a connection. Nested synchronized is gross.

### DIFF
--- a/okhttp/src/main/java/okhttp3/ConnectionPool.java
+++ b/okhttp/src/main/java/okhttp3/ConnectionPool.java
@@ -117,9 +117,7 @@ public final class ConnectionPool {
   RealConnection get(Address address, StreamAllocation streamAllocation) {
     assert (Thread.holdsLock(this));
     for (RealConnection connection : connections) {
-      // TODO(jwilson): this is awkward. We're already holding a lock on 'this', and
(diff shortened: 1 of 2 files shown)


okhttp/src/main/java/okhttp3/ConnectionPool.java
```diff
@@ -119,7 +119,7 @@ RealConnection get(Address address, StreamAllocation streamAllocation) {
     for (RealConnection connection : connections) {
       // TODO(jwilson): this is awkward. We're already holding a lock on 'this', and
       //     connection.allocationLimit() may also lock the FramedConnection.
```
JakeWharton: Remove comment!

okhttp/src/main/java/okhttp3/internal/io/RealConnection.java
```diff
@@ -347,6 +347,16 @@ public boolean isHealthy(boolean doExtensiveChecks) {
     return true;
   }
 
+  /** Refuse incoing streams. */
```
JakeWharton: "incoming"

JakeWharton:
LGTM

PR #2253, opened 2016-01-16
Updated gson recipe to close response body.

dave-r12:


JakeWharton:
Hmm... we should really turn off animal sniffer for non-library modules. The use of try-with-resources is tripping up its enforcement of using Java 6 APIs only.

dave-r12:
Oh, should I change it to just be a call to close() instead?

JakeWharton:
Yeah. That would be easiest.

swankjesse:
CLA is good! Thanks Dave.

PR #2255, opened 2016-01-16
Handle UnsupportedCharsetException in HttpLoggingInterceptor

dave-r12:
This is an attempt to push #1977 over the finish line.

### DIFF
--- a/okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
+++ b/okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
@@ -17,6 +17,7 @@ package okhttp3.logging;
 
 import java.io.IOException;
 import java.nio.charset.Charset;
+import java.nio.charset.UnsupportedCharsetException;
 import java.util.concurrent.TimeUnit;
 import okhttp3.Connection;
 import okhttp3.Headers;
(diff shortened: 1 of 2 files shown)


okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
```diff
@@ -227,7 +228,15 @@ public Level getLevel() {
         Charset charset = UTF8;
         MediaType contentType = responseBody.contentType();
         if (contentType != null) {
-          charset = contentType.charset(UTF8);
+          try {
+            charset = contentType.charset(UTF8);
+          } catch (UnsupportedCharsetException e) {
+            logger.log("");
+            logger.log("Couldn't decode the response body; charset is likely malformed.");
```
swankjesse: maybe include the MediaType so the unlucky user can see `UFT-8` or whatever ?

swankjesse:
LGTM!

okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
JakeWharton: If they got this far they'll see the header logged above.

swankjesse: Yeah, good point!

Issue #2269, opened 2016-01-21
Provide a finer grained way to disallow cleartext traffic


chadbrubaker:
As a follow up to #1058 we've found in deploying the cleartext traffic opt-out that the all or nothing nature of the current approach is too coarse grained to be enabled in many applications. For example the most common issue we run into is that a developer wants to disallow cleartext traffic to their secure domains while still being able to connect to arbitrary external non-sensitive URLs in cleartext.

It would be great if there was a way to approve cleartext traffic at a hostname level. This could take the form of a callback, perhaps a more generic checkRequestPermitted(Url destination, String method) throws ... which could be used for other types of filtering?

I'm happy to provide a patch if the approach sounds sensible.

chadbrubaker:
Actually it seems that Interceptors would allow exactly what we need, but unfortunately we are stuck with the HttpUrlConnection factory which does not have support them and is deprecated. Would it be worth hooking them up in HttpUrlConnection?

JakeWharton:
Interceptors cannot work with HttpURLConnection because of how it deals
with request bodies (bytes are pushed in rather than pulled).

swankjesse:
@chadbrubaker today is the day to upgrade to OkHttp’s Request/Response API. You won’t regret it!

chadbrubaker:
This is for Android, so I don't think's something we can do.

nfuller:
To be clear, this is for the HttpURLConnection in the Android platform: Chad would like hooks that allow connections to be vetoed/failed based on a per-application policy for host names. Including redirects. So, he was hoping for some callbacks he could use that trigger before a connection is made.

swankjesse:
There is no support for this in the HttpURLConnection APIs.

swankjesse:
(You could try to install a `URLStreamHandlerFactory` that rejects `http:`, but even that won’t be able to handle redirects.)

chadbrubaker:
Correct, there's no API support for this in HttpURLConnection, the goal here is to allow it to respect android's NetworkSecurityPolicy.isCleartextTrafficPermitted, or similar out of band policy. I can go into the why about having such a policy if you'd like.

I've thrown together a couple of quick patches to show what I mean. The simplest can be found in pr #2275.
This isn't as clean as putting into OkHttpClient, but clients of OkHttpClient can already use the Inteceptor functionality to honor the NetworkSecurityPolicy so adding another just seems like cruft. This limits the impact to the deprecated okhttp-urlconnection.

[...]

PR #2275, opened 2016-01-23
Add basic URLFilter for HttpURLConnection

chadbrubaker:
This optional filter can be used to allow for blocking HTTP traffic to
certain URLs, for example in order to prevent accessing sensitive
content over HTTP instead of HTTPS.

This behavior is already possible for clients of OkHttpClient, this patch allows legacy clients using the HttpUrlConnection APIs to do basic blocking.

See #2269

### DIFF
--- a/okhttp-urlconnection/src/main/java/okhttp3/OkUrlFactory.java
+++ b/okhttp-urlconnection/src/main/java/okhttp3/OkUrlFactory.java
@@ -21,6 +21,7 @@ import java.net.URL;
 import java.net.URLConnection;
 import java.net.URLStreamHandler;
 import java.net.URLStreamHandlerFactory;
+import okhttp3.internal.URLFilter;
 import okhttp3.internal.huc.HttpURLConnectionImpl;
 import okhttp3.internal.huc.HttpsURLConnectionImpl;
 
@@ -31,6 +32,7 @@ import okhttp3.internal.huc.HttpsURLConnectionImpl;
  */
 public final class OkUrlFactory implements URLStreamHandlerFactory, Cloneable {
   private OkHttpClient client;
+  private URLFilter urlFilter;
 
   public OkUrlFactory(OkHttpClient client) {
     this.client = client;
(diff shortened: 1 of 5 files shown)


swankjesse:
Sorry, this use case is too obscure to warrant extra API. If you really need this behavior, write your own URLStreamHandler that delegates to OkHttp’s UrlStreamHandler.

okhttp-urlconnection/src/test/java/okhttp3/OkUrlFactoryTest.java
```diff
@@ -148,6 +149,30 @@ public void setInstanceFollowRedirectsFalse() throws Exception {
     assertResponseCode(connection, 302);
   }
 
+  @Test
+  public void testURLFilter() throws Exception {
```
swankjesse: Wanna test redirect from HTTPS to HTTP ? I think we have some tests like that lying around somewhere!

okhttp-urlconnection/src/main/java/okhttp3/OkUrlFactory.java
```diff
@@ -97,4 +103,8 @@ HttpURLConnection open(URL url, Proxy proxy) {
       }
     };
   }
+
+  public interface URLFilter {
+      void checkURLPermitted(URL url) throws IOException;
```
swankjesse: nit: we’re 2-space indents throughout this project

swankjesse:
Is there a way we can make this UrlFilter interface non-public and expose this behavior only to AOSP?

chadbrubaker:
We can make it package private to keep it AOSP only.

chadbrubaker:
I made the OkUrlFactory.setUrlFilter method package private, that works fine for AOSP but will keep it from being publicly usable. The interface is still public since it is referenced by HttpURLConnectionImpl but I could move it into okhttp3.internal as well.

swankjesse:
Awesome, yes please do.

chadbrubaker:
Done

chadbrubaker:
Fixed an unused import that I missed when moving the interface to internal.

swankjesse:
Neat! Thanks for jumping through all the hoops for me!

chadbrubaker:
Thanks! Sorry for not being clearer on the usage from the start.

PR #2284, opened 2016-01-27
Override the equals() and hashCode() methods of Cookie

franmontiel:
Now it is possible to use Cookies in Collections that depend on the hash function. (Useful when implementing an in-memory CookieJar)

Both methods takes into account the name, domain a path values as mentioned in the RFC6265 specs (https://tools.ietf.org/html/rfc6265#section-4.1.2):
"If the user agent receives a new cookie with the same cookie-name,domain-value, and path-value as a cookie that it has already stored, the existing cookie is evicted and replaced with the new cookie."

### DIFF
--- a/okhttp/src/main/java/okhttp3/Cookie.java
+++ b/okhttp/src/main/java/okhttp3/Cookie.java
@@ -559,4 +559,24 @@ public final class Cookie {
 
     return result.toString();
   }
+
+  @Override
+  public boolean equals(Object o) {
+    if (!(o instanceof Cookie)) return false;
+
+    Cookie cookie = (Cookie) o;
+    return cookie.name().equalsIgnoreCase(name)
+            && cookie.domain().equalsIgnoreCase(domain)
+            && cookie.path().equalsIgnoreCase(path);
+  }
+
+  @Override
+  public int hashCode() {
+    int hash = 7;
+    hash = 31 * hash + name.toLowerCase().hashCode();
+    hash = 31 * hash + domain.toLowerCase().hashCode();
+    hash = 31 * hash + path.toLowerCase().hashCode();
+
+    return hash;
+  }
 }
--- a/okhttp/src/main/java/okhttp3/Cookie.java
+++ b/okhttp/src/main/java/okhttp3/Cookie.java
@@ -565,17 +565,29 @@ public final class Cookie {
     if (!(o instanceof Cookie)) return false;
 
     Cookie cookie = (Cookie) o;
-    return cookie.name().equalsIgnoreCase(name)
-            && cookie.domain().equalsIgnoreCase(domain)
-            && cookie.path().equalsIgnoreCase(path);
+    return cookie.name.equals(name)
+            && cookie.value.equals(value)
+            && cookie.domain.equals(domain)
+            && cookie.path.equals(path)
+            && cookie.expiresAt == expiresAt
+            && cookie.secure == secure
+            && cookie.httpOnly == httpOnly
+            && cookie.persistent == persistent
+            && cookie.hostOnly == hostOnly;
   }
 
   @Override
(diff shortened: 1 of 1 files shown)


swankjesse:
Not sure if this is the right move. As-is, two cookies with different expiration dates are equal, but they certainly aren’t interchangeable. Instead how do you feel about implementing a basic field-by-field equals that considers all fields and doesn’t do any `equalsIgnoreCase()` stuff ?


   private final String name;
   private final String name;
   private final String value;
   private final String value;
   private final long expiresAt;
   private final long expiresAt;
   private final String domain;
   private final String domain;
   private final String path;
   private final String path;
   private final boolean secure;
   private final boolean secure;
   private final boolean httpOnly;
   private final boolean httpOnly;
 
 
   private final boolean persistent; // True if 'expires' or 'max-age' is present.
   private final boolean persistent; // True if 'expires' or 'max-age' is present.
   private final boolean hostOnly; // True unless 'domain' is present.
   private final boolean hostOnly; // True unless 'domain' is present.

franmontiel:
About replacing `equalsIgnoreCase()` with `equals()` you're totally right I didn't know that URLs are case-sensitive and cookie names too.

About the equals() function taking into account all cookie fields...
Reading the RFC6265 [4.1.2](https://tools.ietf.org/html/rfc6265#section-4.1.2) (3rd paragraph) and [5.3](https://tools.ietf.org/html/rfc6265#section-5.3)(11th point) I understood that a stored cookie should be updated when a new one with the same name, domain and path is received. So in my opinion these three fields are the ones that identifies a cookie the same as a id field identifies an entity no matter what his other fields are.

swankjesse:
If you think of it as a `Map<Key, Value>`, the name+domain+path are the _key_ and the cookie is the _value_. It doesn’t really work to have the cookie as the key since you can’t update it.

franmontiel:
I was trying to create a persistent CookieJar with in-memory cache that uses a `Set<Cookie>` to avoid cookie duplication.

If I'm not wrong, the discussion here seems to be a little bit theoretical: Should equals take into account all fields or should only take into account fields that matter for the identification of the object?

swankjesse:
It's really weird if your Set claims that it contains the cookie `foo=bar` when it contains the key `foo=quux` because the cookies value isn't considered in equality.

The only correct way to implement equals is that two values are equal if they're interchangeable. Two cookies with different values are not interchangeable.

franmontiel:
So in your opinion the "cookie overriding equality" (same name, domain and path) should be checked outside the Cookie and let the equal() be agnostic about business rules like that taking into account all fields, I'm I correct?

swankjesse:
Correct!

franmontiel:
In case you consider useful to have equals and hashcode implemented I've changed both methods in order to take into account all cookie fields.

For my use case I will not rely in equals when checking if one cookie should be overwritten by another one. Thanks for your advice.

okhttp/src/main/java/okhttp3/Cookie.java
```diff
@@ -559,4 +559,36 @@ public Cookie build() {
...
+
+  @Override
+  public int hashCode() {
+    int hash = 7;
+    hash = 31 * hash + name.hashCode();
+    hash = 31 * hash + value.hashCode();
+    hash = 31 * hash + domain.hashCode();
+    hash = 31 * hash + path.hashCode();
+    hash = hash + Long.hashCode(expiresAt);
+    hash = hash + Boolean.hashCode(secure);
```
swankjesse: This API doesn’t exist on Android 2.3.

swankjesse:
Wanna add a test case?

swankjesse:
LGTM.

swankjesse:
Merged manually. Thank you!

Issue #1699, opened 2015-06-11
CertificatePinner pinning against public key of root cert


anirudhgargsc:
Servers often send a certificate chain without the self-signed root cert at the end of the chain, because clients usually contain that cert in their trust store.
For instance, running CertificatePinner with a bogus pin on www.google.com returns the following chain:
>sha1/QO5igIZiHh7a+KVqv8Yc5acwgzM=: CN=www.google.com,O=Google Inc,L=Mountain View,ST=California,C=US
>sha1/Q9rWMO5T+KmAym79hfRqo3mQ4Oo=: CN=Google Internet Authority G2,O=Google Inc,C=US
>sha1/wHqYaI2J+6sFZAwRfap9ZbjKzE4=: CN=GeoTrust Global CA,O=GeoTrust Inc.,C=US

If we look at the certificate chain sent by www.google.com through OpenSSL, it shows that that last cert is not self signed, and is in fact signed by Equifax Secure CA which is the root in this case. But because the Equifax Secure CA cert is not a part of the chain, CertificatePinner doesn't check its public key.

So in scenarios like these, where the certificate chain does not contain the root cert, is there any way to pin against the public key of the root cert?

Thanks!

swankjesse:
Yeah, that's a good idea. We should fix `CertificatePinner` to use the CA cert if it's used for trust.

anirudhgargsc:
Yea I think it will be useful to have the option of pinning against all the certs in the chain and the issuer root if not in the chain.
Are you guys planning on adding this in a future release?

swankjesse:
Yes, we'll add this. It's a bug that we don't have it.

swankjesse:
I suspect I might need hacks or reflection to extract the trusted certificate! I’m anxiously awaiting a better answer here: http://stackoverflow.com/questions/31920719/getting-the-trusted-x509certificate-for-a-java-sslsocket

swankjesse:
I think we're going to need to capture the TrustManager when we create the initial SSLContext, and keep that around so we can use it to look up the trusted cert. It's a hassle!

anirudhgargsc:
Yea that makes sense. Doesn't seem like there's a straightforward way to do this. What a headache.

PR #2285, opened 2016-01-27
Teach OkHttp to lookup the X509TrustManagerFactory.

swankjesse:
Right now there's problems with our ability to confirm certificate pins with
root certificates because we don't actually know which certificate authority
was used.

This unblocks that work.

This code is particularly fragile and gross. There's no API to get the
TrustManager from an SSLSocketFactory. We can potentially change the way
that we compute the initial SSLSocketFactory, but this approach is most
general.

This is motivated by https://github.com/square/okhttp/issues/1699

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/DelegatingSSLSocketFactory.java
+++ b/okhttp-tests/src/test/java/okhttp3/DelegatingSSLSocketFactory.java
@@ -34,51 +34,43 @@ public class DelegatingSSLSocketFactory extends SSLSocketFactory {
     this.delegate = delegate;
   }
 
-  @Override
-  public SSLSocket createSocket() throws IOException {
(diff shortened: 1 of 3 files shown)


swankjesse:
Manually tested on OpenJDK 7, OpenJDK 8, Android 2.3, Android 4.1, and Android 6.0.

kungfoo:
Hmm, the problem is I use okhttp as a transitive dependency of retrofit. Is there a snapshot build?

swankjesse:
Snapshots are here:
https://oss.sonatype.org/content/repositories/snapshots/com/squareup/okhttp3/okhttp/3.3.0-SNAPSHOT/

You'll have to configure your build tool to import snapshot builds.

kungfoo:
Thanks, found them and that seems to work. I think I can even just leave it and use the code in `defaultTrustManager()`.

PR #2299, opened 2016-01-29
Fix throwing with HTTP2 unknown settings frame id

### LINKED ISSUE #2282: http2 readSettings() throws invalid parameter id when unknown settings encountered
We are using OkHttp2 to talk to a service configured using IIS. It works fine with Android 4.x, but on Android 5.0 and above, the connection is always closed by the client. 

The issue turns out to be that the frame returned back from the service contains a setting  with id 16, which is not defined in https specification, thus OkHttp's http2 readSettings() method throws invalid parameter  exception. 

According to the RFC (https://tools.ietf.org/html/rfc7540#section-6.5.2), unknown parameter should be ignored, so would it be better to ignore and continue in this case instead of throwing the exception?

NightlyNexus:
Fixes #2282.

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/internal/framed/Http2Test.java
+++ b/okhttp-tests/src/test/java/okhttp3/internal/framed/Http2Test.java
@@ -18,6 +18,7 @@ package okhttp3.internal.framed;
 import java.io.IOException;
 import java.util.Arrays;
 import java.util.List;
+import java.util.concurrent.atomic.AtomicInteger;
 import okhttp3.internal.Util;
 import okio.Buffer;
 import okio.BufferedSink;
@@ -268,7 +269,7 @@ public class Http2Test {
     }
   }
 
-  @Test public void readSettingsFrameInvalidSettingId() throws IOException {
+  @Test public void readSettingsFrameUnknownSettingId() throws IOException {
     writeMedium(frame, 6); // 2 for the code and 4 for the value
     frame.writeByte(Http2.TYPE_SETTINGS);
     frame.writeByte(Http2.FLAG_NONE);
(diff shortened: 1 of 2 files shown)


NightlyNexus:
Oops, forgot a test.

JakeWharton:
Looks like there's already one that you can update: `readSettingsFrameInvalidSettingId`

NightlyNexus:
(I have no idea what I'm doing.)
Does this look at all reasonable?

okhttp-tests/src/test/java/okhttp3/internal/framed/Http2Test.java
```diff
@@ -277,10 +277,13 @@ public void pushPromise(int streamId, int promisedStreamId, List<Header> headerB
...
 
     try {
-      fr.nextFrame(new BaseTestHandler());
-      fail();
+      fr.nextFrame(new BaseTestHandler() {
+        @Override public void settings(boolean clearPrevious, Settings settings) {
+          assert settings.get(7) == 1;
+        }
+      });
     } catch (IOException e) {
```
JakeWharton: You can just drop this entire try/catch block`

okhttp-tests/src/test/java/okhttp3/internal/framed/Http2Test.java
```diff
@@ -277,10 +277,13 @@ public void pushPromise(int streamId, int promisedStreamId, List<Header> headerB
     frame.writeInt(1);
 
     try {
-      fr.nextFrame(new BaseTestHandler());
-      fail();
+      fr.nextFrame(new BaseTestHandler() {
+        @Override public void settings(boolean clearPrevious, Settings settings) {
+          assert settings.get(7) == 1;
```
JakeWharton: I would look at other tests to see how they do this. I personally would use an AtomicInteger outside the class, set its value in the callback here, and assert its value below (using `assertEquals()`, not `assert`).

NightlyNexus: Oh, good idea. It looks like some of the tests in `Http2Test` don't do that, but I'll do it here. Thanks!

Issue #1693, opened 2015-06-10
OkHttp should provide an option to automatically retry connections that return http status 408


ahanhine:
I am experiencing intermittent http 408 status errors when using OkHttp 2.4.0 or 2.3.0 and Android 5.0 or M Preview. When I downgrade to OkHttp 2.2.0 the problem does not occur.

Http specification states that:
408 Request Timeout
The client did not produce a request within the time that the server was prepared to wait. The client MAY repeat the request without modifications at any later time.

If we add following to HttpEngine.followUpRequest()
case HTTP_CLIENT_TIMEOUT:
  return userRequest.newBuilder().build();
The problem does not occur any more.

Could there be an option in OkHttp to allow retry when http status 408 is returned?

swankjesse:
Dumb question: if it timed out on the first attempt, why would it succeed on attempt #2?

ahanhine:
The 408 response usually comes in milliseconds, for example 15ms. This leads me to think that the socket connection has been used previously and the server at some point has decided that the connection is no longer valid and written the status 408 to sockets stream. When our application or in this case OkHttp, tries to reuse that socket, the status 408 is returned almost instantly. Also retrying the request, usually suceeds.

swankjesse:
Got it. Is this a quirk of your particular application server? I've never heard of this.

ahanhine:
It could be a a quirk, because the problem is not present in all our test environments. My coworker studied the issue and found similar problems reported to for example web browsers. I'll try to attach links to those when I get a chance.

ahanhine:
Here are the issues that seem similar:
https://bugzilla.mozilla.org/show_bug.cgi?id=248827
https://code.google.com/p/chromium/issues/detail?id=377581

swankjesse:
Strange, yeah it looks like this is a thing we should handle. Thanks for the links.

ahanhine:
No broblem, thanks for the great library.

dave-r12:
Are you still in favor of supporting this? At first glance, it looks small in scope, mind if I grab it?

swankjesse:
Do it!

dave-r12:
Should the HttpURLConnection API be ignored? Using the default JDK implementation, it returns the 408 response.

swankjesse:
I think what you have for HttpURLConnection is right. We’re not particular about being consistent with the JDK’s implementation, it’s stuck in 1995.

dave-r12:
Closed with #2302 ?

PR #2302, opened 2016-01-30
Handle 408 responses by retrying the request

dave-r12:
First draft for #1693. In my testing, both Chrome and Firefox repeat the request. Wasn't sure on best approach for the cache tests.

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/CallTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/CallTest.java
@@ -1400,6 +1400,41 @@ public final class CallTest {
     assertEquals("GET /page2 HTTP/1.1", page2.getRequestLine());
   }
 
+  @Test public void getClientRequestTimeout() throws Exception {
+    enqueueRequestTimeoutResponses();
+
+    Response response = client.newCall(new Request.Builder()
+        .url(server.url("/")).build()).execute();
+
+    assertEquals("Body", response.body().string());
+  }
+
+  private void enqueueRequestTimeoutResponses() {
+    server.enqueue(new MockResponse()
+        .setSocketPolicy(SocketPolicy.DISCONNECT_AT_END)
+        .setResponseCode(HttpURLConnection.HTTP_CLIENT_TIMEOUT)
+        .setHeader("Connection", "Close")
+        .setBody("You took too long!"));
+    server.enqueue(new MockResponse().setBody("Body"));
+  }
+
+  @Test public void requestBodyRetransmittedOnClientRequestTimeout() throws Exception {
+    enqueueRequestTimeoutResponses();
+
+    Response response = client.newCall(new Request.Builder()
+        .url(server.url("/"))
+        .post(RequestBody.create(MediaType.parse("text/plain"), "Hello"))
+        .build()).execute();
+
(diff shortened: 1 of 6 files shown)


okhttp/src/main/java/okhttp3/internal/http/HttpEngine.java
```diff
@@ -940,6 +941,17 @@ public Request followUpRequest() throws IOException {
...
+      case HTTP_CLIENT_TIMEOUT:
+        // 408's are rare in practice, but some servers like HAProxy use this response code. The
+        // spec says that we may repeat the request without modifications. Modern browsers also
+        // repeat the request (even non-idempotent ones.)
+        boolean retryableBody = requestBodyOut == null || requestBodyOut instanceof RetryableSink;
+        if (callerWritesRequestBody && !retryableBody) {
+          return null;
+        }
+
+        return userRequest.newBuilder().build();
```
JakeWharton: Is a copy needed here? You can't just return the original `userRequest` directly?

dave-r12: Gotcha, will change

okhttp-tests/src/test/java/okhttp3/CallTest.java
```diff
@@ -1400,6 +1400,41 @@ private void postBodyRetransmittedAfterAuthorizationFail(String body) throws Exc
...
+  @Test public void getClientRequestTimeout() throws Exception {
+    enqueueRequestTimeoutResponses();
+
+    Response response = client.newCall(new Request.Builder()
+        .url(server.url("/")).build()).execute();
+
+    assertEquals("Body", response.body().string());
+  }
+
+  private void enqueueRequestTimeoutResponses() {
```
swankjesse: Mostly off topic . . . though it causes duplication, I’m generally a fan of just inlining methods like these. For production code DRY is good. But for tests, I prefer to keep tests self-contained over keeping ’em not-repetitive.

That said, it's matters of degree. Lots of our test suite is built with supporting classes - even MockWebServer started out as a humble test utility here.

okhttp-tests/src/test/java/okhttp3/URLConnectionTest.java
```diff
@@ -2327,6 +2327,53 @@ private void testRedirect(boolean temporary, String method) throws Exception {
...
+    assertEquals("Body", readAscii(connection.getInputStream(), Integer.MAX_VALUE));
+
+    RecordedRequest request1 = server.takeRequest();
+    assertEquals("Hello", request1.getBody().readUtf8());
+
+    RecordedRequest request2 = server.takeRequest();
+    assertEquals("Hello", request2.getBody().readUtf8());
+  }
+
+  @Test public void streamedBodyWithClientRequestTimeout() throws Exception {
```
swankjesse: Nice.

swankjesse:
This is awesome.

okhttp-tests/src/test/java/okhttp3/CallTest.java
dave-r12: Will do, I'll keep that in mind going forward

PR #2304, opened 2016-01-31
New CertificateAuthorityCouncil class to normalize certificate chains.

swankjesse:
The goal of this is to get the root CA certificate into the certificate
chain, so that it can be considered by the certificate pinner. The work
to integrate CertificateAuthorityCouncil with CertificatePinner will
come in a follow-up PR.

See: https://github.com/square/okhttp/issues/1699

### DIFF
--- a/mockwebserver/src/main/java/okhttp3/internal/SslContextBuilder.java
+++ b/mockwebserver/src/main/java/okhttp3/internal/SslContextBuilder.java
@@ -13,29 +13,19 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 package okhttp3.internal;
 
 import java.io.IOException;
 import java.io.InputStream;
-import java.math.BigInteger;
 import java.net.InetAddress;
 import java.net.UnknownHostException;
 import java.security.GeneralSecurityException;
-import java.security.KeyPair;
-import java.security.KeyPairGenerator;
 import java.security.KeyStore;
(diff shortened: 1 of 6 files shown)


okhttp/src/main/java/okhttp3/internal/tls/CertificateAuthorityCouncil.java
```diff
@@ -0,0 +1,120 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one or more
```
swankjesse: (this is the Conscrypt copyright)

mockwebserver/src/main/java/okhttp3/internal/SslContextBuilder.java
```diff
@@ -79,17 +61,19 @@ public static synchronized SSLContext localhost() {
...
 
   public SSLContext build() throws GeneralSecurityException {
-    char[] password = "password".toCharArray();
-
-    // Generate public and private keys and use them to make a self-signed certificate.
-    KeyPair keyPair = generateKeyPair();
-    X509Certificate certificate = selfSignedCertificate(keyPair, "1");
+    // Generate a self-signed cert for the server to serve and the client to trust.
+    HeldCertificate heldCertificate = new HeldCertificate.Builder()
+        .serialNumber("1")
```
JakeWharton: nit: not needed

PR #2317, opened 2016-02-05
Defer known-length payload writes when within a segment of completion.

### LINKED ISSUE #2270: [Websocket] Configure frame size to avoid payload splitting.
Before version 2.5 of okhttp-ws, client was passing a **Buffer** and it was sent as specified by the client.

On version 2.6, the implementation migrated to use **RequestBody** and **ResponseBody**. By doing this all payload sent to server are truncated at 2048 (size of the **Segment**). If the payload is bigger than 2048, then the implementation sends the remaining data as a continuation frame.

According to [RFC](https://tools.ietf.org/html/rfc6455#page-27), there is nothing wrong by splitting the payload into multiple frames.

But in my opinion this splitting is unnecessary and it should have an option to define the max desired frame size to avoid splitting.

JakeWharton:
This allows both really small bodies and very large buffered bodies to be written as a single frame.

Closes #2270.

### DIFF
--- a/okhttp-ws-tests/src/test/java/okhttp3/internal/ws/WebSocketWriterTest.java
+++ b/okhttp-ws-tests/src/test/java/okhttp3/internal/ws/WebSocketWriterTest.java
@@ -18,6 +18,7 @@ package okhttp3.internal.ws;
 import java.io.EOFException;
 import java.io.IOException;
 import java.util.Random;
+import okhttp3.RequestBody;
 import okio.Buffer;
 import okio.BufferedSink;
 import okio.ByteString;
@@ -35,6 +36,7 @@ import static okhttp3.internal.ws.WebSocketProtocol.PAYLOAD_BYTE_MAX;
 import static okhttp3.internal.ws.WebSocketProtocol.PAYLOAD_SHORT_MAX;
 import static okhttp3.internal.ws.WebSocketProtocol.toggleMask;
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
 public final class WebSocketWriterTest {
@@ -61,7 +63,7 @@ public final class WebSocketWriterTest {
   private final WebSocketWriter clientWriter = new WebSocketWriter(true, data, random);
 
   @Test public void serverTextMessage() throws IOException {
-    BufferedSink sink = Okio.buffer(serverWriter.newMessageSink(OPCODE_TEXT));
+    BufferedSink sink = Okio.buffer(serverWriter.newMessageSink(OPCODE_TEXT, -1));
 
     sink.writeUtf8("Hel").flush();
     assertData("010348656c");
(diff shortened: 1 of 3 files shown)


okhttp-ws-tests/src/test/java/okhttp3/internal/ws/WebSocketWriterTest.java
```diff
@@ -73,8 +75,65 @@
...
+    Buffer body = bytes.clone();
+    sink.write(body.readByteString(20_000));
+    sink.write(body.readByteString(20_000));
+    sink.write(body.readByteString(20_000));
+    sink.write(body.readByteString(20_000));
+    sink.write(body.readByteString(20_000));
+    sink.close();
+
+    assertData("017e4800");
+    assertData(bytes.readByteArray(18_432));
```
swankjesse: curious - what causes this specific value?

JakeWharton: 2048 * 9. It's from `completeSegmentByteCount()`

okhttp-ws/src/main/java/okhttp3/internal/ws/WebSocketWriter.java
```diff
@@ -234,6 +236,20 @@ private void writeMaskedSynchronized(BufferedSource source, long byteCount) thro
...
+        }
+        if (isFirstFrame && buffer.size() > contentLength - 2048 /* segment size */) {
+          // This is either a larger buffered write or a very small write. Defer flushing to the
+          // underlying socket since close() is about to be called and will flush all the content.
+          return;
+        }
+      }
+
       long emitCount = buffer.completeSegmentByteCount();
       if (emitCount > 0) {
```
swankjesse: I think the decision as to whether to emit or not should be here?

swankjesse: Doesn’t really matter, it’s just weird that both this expression and the expression on 246 are deciding whether to emit.

okhttp-ws-tests/src/test/java/okhttp3/internal/ws/WebSocketWriterTest.java
swankjesse: Ah, yeah this is a little bit of emergent behavior. Awkwardly the segment count is different on master.

okhttp-ws/src/main/java/okhttp3/internal/ws/WebSocketWriter.java
JakeWharton: Yeah I had a few iterations before I landed on what's in this PR. I'll rework it so there's no early return.

okhttp-ws-tests/src/test/java/okhttp3/internal/ws/WebSocketWriterTest.java
swankjesse: (And in the future might be even more dynamic!)

swankjesse:
LGTM

Net win, though I think there’s still opportunity to make okio + test cases smarter.

okhttp-ws-tests/src/test/java/okhttp3/internal/ws/WebSocketWriterTest.java
JakeWharton: Yep. The only problem is that if I make it too dynamic it will no longer be testing the desired behavior. I suspect this won't be the only test to break if we change the segment size, but at least it will be explicitly on an Okio upgrade.

I'll think about it what we can do about it.

okhttp-ws/src/main/java/okhttp3/internal/ws/WebSocketWriter.java
JakeWharton: Fixed. I removed the write-to-much check since it wasn't working correctly anyway.

okhttp-ws/src/main/java/okhttp3/internal/ws/WebSocketWriter.java
```diff
@@ -234,8 +236,13 @@ private void writeMaskedSynchronized(BufferedSource source, long byteCount) thro
 
       buffer.write(source, byteCount);
 
+      // Determine if this is a buffered write which we can defer until close() flushes.
```
swankjesse: Yep, I like this better.

Issue #2321, opened 2016-02-07
Very High CPU time on NimbleDroid


iNoles:
NimbeDroid runs only Cold Startup on Android 4.4. This should be a regression from OkHttp 3.0

1533ms okhttp3.OkHttpClient constructor (okhttp3.OkHttpClient$Builder)
    - 1260ms com.android.conscrypt.TrustManagerImpl.getAcceptedIssuers
    - 120ms okhttp3.internal.tls.CertificateAuthorityCouncil constructor
    - 115ms javax.net.ssl.SSLContext.init
    -  65ms  LinkedHashmap.get
    -  54ms com.android.conscrypt.OpenSSLX509Certificate.getSubjectX500Principal

swankjesse:
Aww, shucks. Thanks for reporting this.

swankjesse:
I think we fix by only doi g this work if a CertificatePinner is in use.

We can possibly also move the work to Platform which might be able to get Android’s own certificate cleaner to do that work.

swankjesse:
Released as 3.1.1 and 2.7.4.

PR #2324, opened 2016-02-07
Don't build a trust root index on Android.

swankjesse:
We can just cheat and use reflection to use Conscrypt's trust root index
directly. This results in a substantial savings in app startup - 500
milliseconds or more.

Closes: https://github.com/square/okhttp/issues/2321

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/tls/CertificateChainCleaner.java
+++ b/okhttp/src/main/java/okhttp3/internal/tls/CertificateChainCleaner.java
@@ -17,56 +17,43 @@
 package okhttp3.internal.tls;
 
 import java.security.GeneralSecurityException;
-import java.security.PublicKey;
 import java.security.cert.Certificate;
 import java.security.cert.X509Certificate;
 import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Deque;
 import java.util.Iterator;
-import java.util.LinkedHashMap;
 import java.util.List;
-import java.util.Map;
 import javax.net.ssl.SSLPeerUnverifiedException;
-import javax.security.auth.x500.X500Principal;
 
 /**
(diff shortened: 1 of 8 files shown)


okhttp/src/main/java/okhttp3/internal/tls/CertificateChainCleaner.java
```diff
@@ -17,56 +17,43 @@
...
+ * <p>Use of the chain cleaner is necessary to omit unexpected certificates that aren't relevant to
+ * the TLS handshake and to extract the trusted CA certificate for the benefit of certificate
+ * pinning.
  *
  * <p>This class includes code from <a href="https://conscrypt.org/">Conscrypt's</a> {@code
  * TrustManagerImpl} and {@code TrustedCertificateIndex}.
  */
-public final class CertificateAuthorityCouncil {
-  private final Map<X500Principal, List<X509Certificate>> subjectToCaCerts = new LinkedHashMap<>();
+public final class CertificateChainCleaner {
```
swankjesse: Most of this change is splitting the data (trust root index) from behavior (cleaning certificates).

okhttp/src/main/java/okhttp3/internal/tls/CertificateChainCleaner.java
```diff
@@ -78,8 +65,8 @@ public CertificateAuthorityCouncil(X509Certificate... caCerts) {
       // If this cert has been signed by a trusted CA cert, we're done. Add the trusted CA
       // certificate to the end of the chain, unless it's already present. (That would happen if the
       // first certificate in the chain is itself a self-signed and trusted CA certificate.)
-      X509Certificate caCert = findByIssuerAndSignature(toVerify);
-      if (caCert != null && verifySignature(toVerify, caCert)) {
```
swankjesse: (was already being verified)

Issue #2323, opened 2016-02-07
Unable to extract the trust manager


AlexLardschneider:
Hi,
after upgrading from OkHttp 3.0.1 to 3.1.0 I get the following stack trace and crash after calling build() to create a OkHttpClient.

```
java.lang.IllegalStateException: Unable to extract the trust manager on okhttp3.internal.Platform$Android@1d8cf999, sslSocketFactory is class com.google.android.gms.org.conscrypt.KitKatPlatformOpenSSLSocketAdapterFactory
                                                                        at okhttp3.OkHttpClient.<init>(OkHttpClient.java:187)
                                                                        at okhttp3.OkHttpClient.<init>(OkHttpClient.java:60)
                                                                        at okhttp3.OkHttpClient$Builder.build(OkHttpClient.java:719)
```

This issue was not present in 3.0.1.

Here's where the crash happens:
```
OkHttpClient.Builder builder = new OkHttpClient.Builder()
                    .connectTimeout(10, TimeUnit.SECONDS)
                    .writeTimeout(10, TimeUnit.SECONDS)
                    .readTimeout(30, TimeUnit.SECONDS);

            OkHttpClient client = builder.build();
```

swankjesse:
Ahh, yeah I gotta fix this for the GMS TLS stack. Will fix.

swankjesse:
Released as 3.1.1 and 2.7.4.

artem-zinnatullin:
Just re-checked with 3.1.1, unfortunately, still doesn't work under Robolectric but works on actual Android (API 23).

```
Caused by: java.lang.IllegalStateException: Unable to extract the trust manager on okhttp3.internal.Platform$Android@20bcec4c, sslSocketFactory is class sun.security.ssl.SSLSocketFactoryImpl
	at okhttp3.OkHttpClient.<init>(OkHttpClient.java:186)
	at okhttp3.OkHttpClient.<init>(OkHttpClient.java:60)
	at okhttp3.OkHttpClient$Builder.build(OkHttpClient.java:718)
```

marco-duckma:
Same thing here, robolectric tests are now crashing after upgrade from 3.0.0-RC1

kkocel:
Same here on robolectric tests:

`java.lang.IllegalStateException: Unable to extract the trust manager on okhttp3.internal.Platform$Android@19f0acf3, sslSocketFactory is class sun.security.ssl.SSLSocketFactoryImpl
[...]

PR #2325, opened 2016-02-08
Support the GMS security provider when searching for the trust manager.

swankjesse:
Closes https://github.com/square/okhttp/issues/2323


vangliana:
It would be good if a release is made once this gets merged. This issue is wrecking havoc with any test relying on Robolectric runner

JakeWharton:
@vangliana You can use the previous release, the SNAPSHOT from Sonatype's snapshot repository, or JitPack until a release is made.

vangliana:
@JakeWharton previous release is no good since okhttp3.OkHttpClient modifier is final which breaks my mocks. I'll use the snapshot instead.

JakeWharton:
Mocking `OkHttpClient` is an anti-pattern. You can mock or fake `Call.Factory` instead.

PR #2335, opened 2016-02-10
Add HttpUrl.Builder.addPathSegments()

NightlyNexus:


NightlyNexus:
I'm probably missing things here, especially in regards to `addEncodedPathSegments()`.

okhttp/src/main/java/okhttp3/HttpUrl.java
```diff
@@ -772,6 +772,23 @@ public Builder addPathSegment(String pathSegment) {
...
     }
 
+    public Builder addPathSegments(String groupedPathSegments) {
+      if (groupedPathSegments == null) {
+        throw new IllegalArgumentException("groupedPathSegments == null");
+      }
+      int startIndex = 0;
+      int endIndex = groupedPathSegments.indexOf('/');
+      while (endIndex != -1) {
+        String pathSegment = groupedPathSegments.substring(startIndex, endIndex);
```
swankjesse: Oooh, you can avoid the substring call. Push will do that for you automatically.

okhttp/src/main/java/okhttp3/HttpUrl.java
```diff
@@ -772,6 +772,23 @@ public Builder addPathSegment(String pathSegment) {
       return this;
     }
 
+    public Builder addPathSegments(String groupedPathSegments) {
+      if (groupedPathSegments == null) {
+        throw new IllegalArgumentException("groupedPathSegments == null");
+      }
+      int startIndex = 0;
+      int endIndex = groupedPathSegments.indexOf('/');
+      while (endIndex != -1) {
```
swankjesse: With `Util.delimiterOffset()` it should be straightforward to call `push()` only once within the loop. Something like

```
int offset = 0;
do {
  int segmentEnd = delimiterOffset(...)
  push(groupedPathSegments, ...);
  offset = segmentEnd + 1;
} while (offset <= groupedPathSegments.length());
```
I think one interesting test case will be when the input ends with `/` and I think the resulting URL should then end with `/`. We should probably also confirm that the code works if the input starts with `/`.

swankjesse: (starting with `/` means there’s going to be an empty segment in the output, like `http://example.com//x`, which is weird but legal)

okhttp/src/main/java/okhttp3/HttpUrl.java
```diff
@@ -772,6 +772,23 @@ public Builder addPathSegment(String pathSegment) {
       return this;
     }
 
+    public Builder addPathSegments(String groupedPathSegments) {
```
swankjesse: rename `groupedPathSegments` to just `pathSegments` ? Awkward to have the method’s name (path segments) disagree with its parameter

okhttp/src/main/java/okhttp3/HttpUrl.java
swankjesse: (The code at the bottom of resolvePath is a good place to start)

okhttp/src/main/java/okhttp3/HttpUrl.java
NightlyNexus: Yup, that was dumb of me haha. fixing.
I am a bit nervous about the similarity in the method names. There would be no cost using the "plural" method, just the possibility of hiding bad input.

okhttp/src/main/java/okhttp3/HttpUrl.java
NightlyNexus: Nice! I should have seen that.

okhttp/src/main/java/okhttp3/HttpUrl.java
NightlyNexus: Ah, yeah, clever.
Cool, going to look into those cases.

okhttp-tests/src/test/java/okhttp3/HttpUrlTest.java
```diff
@@ -801,6 +801,28 @@
     assertEquals(3, HttpUrl.parse("http://host/a/b/c").pathSize());
   }
 
+  @Test public void addPathSegments() throws Exception {
```
swankjesse: Nice tests!

NightlyNexus:
Merged as https://github.com/square/okhttp/commit/de0d48866fe25f2d2643b936c9b04d14841c9536

Issue #2341, opened 2016-02-13
Certificate pinning crash with prelease GMS


tadfisher:
Since we shipped 2.7.4 in a hotfix release, we're receiving many reports of this crash, confined completely to the Nexus 5X, Nexus 6, and Nexus 6P on Android 6.0.1.

Any pointers would be appreciated.

```
Fatal Exception: java.lang.NullPointerException: Attempt to invoke virtual method 'java.security.cert.X509Certificate java.security.cert.TrustAnchor.getTrustedCert()' on a null object reference
       at com.squareup.okhttp.internal.tls.AndroidTrustRootIndex.findByIssuerAndSignature(AndroidTrustRootIndex.java:43)
       at com.squareup.okhttp.internal.tls.CertificateChainCleaner.clean(CertificateChainCleaner.java:68)
       at com.squareup.okhttp.internal.io.RealConnection.connectTls(RealConnection.java:207)
       at com.squareup.okhttp.internal.io.RealConnection.connectSocket(RealConnection.java:149)
       at com.squareup.okhttp.internal.io.RealConnection.connect(RealConnection.java:112)
       at com.squareup.okhttp.internal.http.StreamAllocation.findConnection(StreamAllocation.java:184)
       at com.squareup.okhttp.internal.http.StreamAllocation.findHealthyConnection(StreamAllocation.java:126)
       at com.squareup.okhttp.internal.http.StreamAllocation.newStream(StreamAllocation.java:95)
       at com.squareup.okhttp.internal.http.HttpEngine.connect(HttpEngine.java:281)
       at com.squareup.okhttp.internal.http.HttpEngine.sendRequest(HttpEngine.java:224)
       at com.squareup.okhttp.Call.getResponse(Call.java:286)
       at com.squareup.okhttp.Call$ApplicationInterceptorChain.proceed(Call.java:243)
       at com.banksimple.http.ApiInterceptor.intercept(ApiInterceptor.java:50)
       at com.squareup.okhttp.Call$ApplicationInterceptorChain.proceed(Call.java:232)
       at com.squareup.okhttp.Call.getResponseWithInterceptorChain(Call.java:205)
       at com.squareup.okhttp.Call.execute(Call.java:80)
       at com.banksimple.http.BaseApiClient.execute(BaseApiClient.java:94)
       at com.banksimple.http.BaseApiClient.execute(BaseApiClient.java:106)
[...]

PR #2353, opened 2016-02-21
Always build full chains with CertificateChainCleaner.

swankjesse:
We had a bug on Android where the trust root index would return trusted
intermediate certificates when we called findTrustAnchorByIssuerAndSignature().
This would lead to a partial chain, which would fail the pin check
unnecessarily.

Also permit that method to return null, as it does when the certificate
is not trusted.

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/CertificateChainCleanerTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/CertificateChainCleanerTest.java
@@ -152,17 +152,17 @@ public final class CertificateChainCleanerTest {
         council.clean(list(certB, certUnnecessary, certA, root)));
   }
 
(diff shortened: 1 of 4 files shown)


swankjesse:
See https://github.com/square/okhttp/issues/2341

tadfisher:
:beers:

JakeWharton:
Hurt my brain to read through these tests.

scompt:
@JakeWharton Would you consider backporting this and releasing it as 2.7.5. We weren't planning to update to OkHttp 3 quite yet, but we're starting to see the crash from #2341. The patch applies cleanly except for the new `CertificatePinnerChainValidationTest` test.

swankjesse:
Can do.

Issue #1539, opened 2015-03-30
CertificatePinner uses sha1 hashes


Wick3r:
This is not my main domain, so i am maybe wrong with this:

Using a sha1 hash is a security issue, since sha1 should not be used for signing certificates anymore, so it should not for pinning.

Sincerely

swankjesse:
We're using Chromium's format.
http://src.chromium.org/viewvc/chrome/trunk/src/net/http/transport_security_state_static.certs

If you can convince them to change, we'll change too. But for now the tangible benefits of consistency with chrome outweigh the attack vector. Remember, pinning is a 2nd layer of security; the attacker still needs to compromise a CA.

tbroyer:
FYI, HPKP uses sha256: https://tools.ietf.org/html/rfc7469
Given the authors, I suspect Chrome will update soon ;-)

bottiger:
@swankjesse I would also like to place a vote for sha256 support since Google have already begun phasing out sha1

http://googleonlinesecurity.blogspot.dk/2014/09/gradually-sunsetting-sha-1.html
https://www.digicert.com/sha-2-ssl-certificates.htm

mhoogenboom:
Will the conversion to SHA256 hashing be reconsidered in the near future?

swankjesse:
Let's do it.

PR #2356, opened 2016-02-21
Implement support for SHA-256 pins.

swankjesse:
Closes: https://github.com/square/okhttp/issues/1539

### DIFF
--- a/okhttp/src/main/java/okhttp3/CertificatePinner.java
+++ b/okhttp/src/main/java/okhttp3/CertificatePinner.java
@@ -17,19 +17,15 @@ package okhttp3;
 
 import java.security.cert.Certificate;
 import java.security.cert.X509Certificate;
+import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.LinkedHashMap;
-import java.util.LinkedHashSet;
+import java.util.Collections;
 import java.util.List;
-import java.util.Map;
-import java.util.Set;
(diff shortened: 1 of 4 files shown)


okhttp/src/main/java/okhttp3/CertificatePinner.java
```diff
@@ -183,68 +194,103 @@ public void check(String hostname, Certificate... peerCertificates)
...
   Builder newBuilder() {
     return new Builder(this);
   }
 
-  /**
-   * Returns the SHA-1 of {@code certificate}'s public key. This uses the mechanism Moxie
-   * Marlinspike describes in <a href="https://github.com/moxie0/AndroidPinning">Android
-   * Pinning</a>.
-   */
+  /** Returns the SHA-256 of {@code certificate}'s public key. */
```
JakeWharton: Should we document that this along the lines of that the method returns the latest pin hash format supported by this class. Technically you're making a breaking API change here, but I think we can gloss over it with documentation to something like that effect.

JakeWharton:
LGTM

okhttp/src/main/java/okhttp3/CertificatePinner.java
swankjesse: I added a note that it was different in 3.1.2 and earlier. I don't expect it to matter in practice.

Issue #2368, opened 2016-02-24
ResponseBody is null when a WebSocket call fails (e.g. http 500)


iocanel:
We are using okhttp in https://github.com/fabric8io/kubernetes-client. 

In some cases the websocket calls we create fail with http 500 + json message that contains a detailed description to the problem.

Unfortunately, the response cannot be read from the Response body or message and we are left in the dark.

Test Case to reproduce the problem: https://gist.github.com/iocanel/b3a9cb6d62e5aabe5426

Thanks

JakeWharton:
Somewhat related to #2264 as well. Will look.

iocanel:
FWIW I forgot to mention that I get this with versions 2.7.2, 2.7.3 and 2.7.4 (haven't tried older ones).

dave-r12:
This should be fixed with OkHttp 3.3.1, see #2374 and #2583.

Issue #2264, opened 2016-01-20
NPE when logging-interceptor:3.0.1with okhttp-ws:3.0.1 


whylee259:
I'm using okhttp-ws and logging-interceptor is added to okhttp client.

I think NPE is caused because `response.body()` returns alway null.

```
01-20 14:28:45.033 12757-12904/? E/AndroidRuntime: FATAL EXCEPTION: OkHttp Dispatcher
  java.lang.NullPointerException
    at okhttp3.logging.HttpLoggingInterceptor.intercept(HttpLoggingInterceptor.java:207)
    at okhttp3.RealCall$ApplicationInterceptorChain.proceed(RealCall.java:187)
    at okhttp3.RealCall.getResponseWithInterceptorChain(RealCall.java:160)
    at okhttp3.RealCall.access$100(RealCall.java:30)
    at okhttp3.RealCall$AsyncCall.execute(RealCall.java:127)
    at okhttp3.internal.NamedRunnable.run(NamedRunnable.java:33)
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1112)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:587)
    at java.lang.Thread.run(Thread.java:841)
```

JakeWharton:
I think we define that response.body() should never return null, so this is
more likely a bug in the okhttp-ws module than the logger.

JakeWharton:
Fixed by https://github.com/square/okhttp/pull/2831

PR #2374, opened 2016-02-28
bug fix 2368

sourabh-goyal:
The reason of bug was due to unhandled condition of non 101 during
websocket response readNetworkResponse() method  in http Engine. Also response body was being closed in createWebSocketCall when response code was non 101. I ran ws related testcases after making my changes.
This fix solves problem for issue #2368 and possibly #2264 as well.

### DIFF
--- a/okhttp-ws/src/main/java/okhttp3/ws/WebSocketCall.java
+++ b/okhttp-ws/src/main/java/okhttp3/ws/WebSocketCall.java
@@ -115,7 +115,7 @@ public final class WebSocketCall {
 
   private void createWebSocket(Response response, WebSocketListener listener) throws IOException {
     if (response.code() != 101) {
-    	//Util.closeQuietly(response.body());   
+//Util.closeQuietly(response.body());
       throw new ProtocolException("Expected HTTP 101 response but was '"
           + response.code()
           + " "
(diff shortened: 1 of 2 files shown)


okhttp/src/main/java/okhttp3/internal/http/HttpEngine.java
```diff
@@ -731,6 +731,10 @@ private Response readNetworkResponse() throws IOException {
       networkResponse = networkResponse.newBuilder()
           .body(httpStream.openResponseBody(networkResponse))
           .build();
+    } else if (networkResponse.code() != 101) {
```
JakeWharton: This block has the same contents as the one above it, so just merge the conditionals with an `||`.

okhttp-ws/src/main/java/okhttp3/ws/WebSocketCall.java
```diff
@@ -115,7 +115,7 @@ public void cancel() {
 
   private void createWebSocket(Response response, WebSocketListener listener) throws IOException {
     if (response.code() != 101) {
-      Util.closeQuietly(response.body());
+//Util.closeQuietly(response.body());
```
JakeWharton: Just delete this. There's no value in keeping it around.

karlicoss:
Hi guys, is this getting merged anywhere soon?

dave-r12:
I think this one looks ok. Was there anything else? Maybe just needs a test case?

karlicoss:
@JakeWharton can you merge this?

swankjesse:
Resolved with #2583. Thanks!

Issue #2351, opened 2016-02-18
Omit unreadable bodies from logging interceptor


alexrwegener:
It would be nice to only log the bodies of human readable content-types. With image heavy applications also using OkHttp as Picasso's downloader using Level.BODY produces lots of noise. 

Content-Type: image/jpeg
Content-Length: 57634
Accept-Ranges: bytes
Access-Control-Allow-Origin: *
Date: Wed, 17 Feb 2016 15:14:45 GMT
Etag: "5c21b2b-e122-52b7fec4f1540"
Last-Modified: Thu, 11 Feb 2016 15:08:29 GMT
Server: ECS (ord/269B)
X-Cache: HIT
OkHttp-Sent-Millis: 1455722085753
OkHttp-Received-Millis: 1455722085767
������JFIF����H��H��������C��			
����C
... etc x 1000

If this is something you all would be open to fixing I'd love to help :)

swankjesse:
We probably need a simple heuristic to detect plaintext.

Here’s my proposal:
1. Decode the body as UTF-8
2. Count replacement characters (U+FFFD) and other non-printable control characters (U+0000, etc.) in the first 16 codepoints of the decoded body.
3. If the count is ≥ 2, we consider it non-text and don’t print.

Checking 2 of 16 is arbitrary. Ideally it matches well with this:
https://en.wikipedia.org/wiki/List_of_file_signatures

swankjesse:
http://stackoverflow.com/questions/7110750/how-do-popular-source-control-systems-differentiate-binary-files-from-text-files
http://eli.thegreenplace.net/2011/10/19/perls-guess-if-file-is-text-or-binary-implemented-in-python

swankjesse:
Thanks @alexrwegener !

PR #2376, opened 2016-02-28
Detect plaintext bodies

alexrwegener:
My attempt at fixing https://github.com/square/okhttp/issues/2351

### DIFF
--- a/okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
+++ b/okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
@@ -192,7 +192,7 @@ public final class HttpLoggingInterceptor implements Interceptor {
         }
 
         logger.log("");
-        if (isPlainText(buffer.clone())) {
+        if (isPlainText(buffer)) {
           logger.log(buffer.readString(charset));
         } else {
           logger.log("binary body omitted.");
--- a/okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
+++ b/okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
@@ -273,7 +273,8 @@ public final class HttpLoggingInterceptor implements Interceptor {
   }
 
   private boolean isNonPrintableCharacter(int codePoint) {
-    return codePoint == '\ufffd' || ('\u0000' <= codePoint && codePoint <= '\u001f') || ('\u007f' <= codePoint && codePoint <= '\u009f');
+    return codePoint == '\ufffd' || ('\u0000' <= codePoint && codePoint <= '\u001f')
+        || ('\u007f' <= codePoint && codePoint <= '\u009f');
   }
 
   private boolean bodyEncoded(Headers headers) {
--- a/okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
+++ b/okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
@@ -194,12 +194,12 @@ public final class HttpLoggingInterceptor implements Interceptor {
         logger.log("");
         if (isPlainText(buffer)) {
           logger.log(buffer.readString(charset));
+          logger.log("--> END " + request.method()
+              + " (" + requestBody.contentLength() + "-byte body)");
         } else {
-          logger.log("binary body omitted.");
+          logger.log("--> END " + request.method() + " (binary "
+              + requestBody.contentLength() + "-byte body omitted)");
         }
-
-        logger.log("--> END " + request.method()
-            + " (" + requestBody.contentLength() + "-byte body)");
       }
     }
 
@@ -243,13 +243,15 @@ public final class HttpLoggingInterceptor implements Interceptor {
           }
         }
 
+        if (!isPlainText(buffer)) {
+          logger.log("");
+          logger.log("<-- END HTTP (binary " + buffer.size() + "-byte body omitted)");
+          return response;
+        }
+
         if (contentLength != 0) {
           logger.log("");
-          if (isPlainText(buffer)) {
-            logger.log(buffer.clone().readString(charset));
-          } else {
-            logger.log("binary body omitted.");
-          }
+          logger.log(buffer.clone().readString(charset));
         }
 
         logger.log("<-- END HTTP (" + buffer.size() + "-byte body)");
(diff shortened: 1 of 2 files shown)


okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
```diff
@@ -192,7 +192,11 @@ public Level getLevel() {
         }
 
         logger.log("");
-        logger.log(buffer.readString(charset));
+        if (isPlainText(buffer.clone())) {
```
JakeWharton: This is a very expensive thing to do for every body when we're only checking the first 16 chars.

okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
```diff
@@ -251,6 +259,27 @@ public Level getLevel() {
     return response;
   }
 
+  private boolean isPlainText(Buffer buffer) throws IOException {
```
swankjesse: Please add a plaintext description of the heuristic here?

okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
```diff
@@ -251,6 +259,27 @@ public Level getLevel() {
...
+    Buffer out = new Buffer();
+    long bytes = buffer.size() < 64 ? buffer.size() : 64;
```
swankjesse: What is 64 for here?

okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
```diff
@@ -251,6 +259,27 @@ public Level getLevel() {
...
+    buffer.copyTo(out, 0, bytes);
```
swankjesse: You can probably avoid creating an entire additional buffer, just use the original buffer and call `getByte(i)`

okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
```diff
@@ -251,6 +259,27 @@ public Level getLevel() {
...
+    int charCount = 0;
+    for (int i = 0; i < 16; i++) {
+      if (out.exhausted()) {
+        break;
+      }
+      if (isNonPrintableCharacter(out.readUtf8CodePoint())) {
+        charCount++;
```
swankjesse: Maybe just `return false` here? One non-ASCII byte is probably enough

okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
```diff
@@ -251,6 +259,27 @@ public Level getLevel() {
...
+      }
+    }
+    return charCount < 2;
```
swankjesse: ... and this would be `return true`

okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
swankjesse: Actually on second thought, we want to look at code points and not bytes, so your approach is legit

okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
swankjesse: Ahh, that’s because 16 codepoints * 4 bytes each = 64

okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
```diff
@@ -251,6 +259,27 @@ public Level getLevel() {
...
+  }
+
+  private boolean isNonPrintableCharacter(int codePoint) {
+    return codePoint == '\ufffd' || ('\u0000' <= codePoint && codePoint <= '\u001f')
```
swankjesse: nice

swankjesse: Maybe just use Character.isISOControl() and don’t worry about `\ufffd` ? That could plausibly show up in plaintext anyway right?

okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
swankjesse: nit: got a better name than `bytes` ? Usually we use that to refer to a `byte[]`. Maybe `byteCount` ?

okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
```diff
@@ -192,7 +192,11 @@ public Level getLevel() {
         }
 
         logger.log("");
-        logger.log(buffer.readString(charset));
+        if (isPlainText(buffer)) {
+          logger.log(buffer.readString(charset));
+        } else {
+          logger.log("binary body omitted.");
```
swankjesse: any way you can move the log so we only log a single line in this case, as on line 201 ? Maybe something like on 183?

swankjesse: 183 has this:
```
 logger.log("--> END " + request.method() + " (encoded body omitted)");
```
Maybe this wants to be something like:
```
 logger.log("--> END " + request.method() + " (binary " + requestBody.contentLength() + "-byte body omitted)");
```

swankjesse:
Neat

okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
alexrwegener: I agree with you. I think it's better to err on the side of logging the body. I included it because I noticed error conditions in `readUtf8CodePoint()` return `\ufffd`.

dave-r12:
Squash the commits?

swankjesse:
Squashed & merged manually. Note that I added an extra test case & discovered a potential problem when the input contains a partial UTF-8 sequence.
https://github.com/square/okhttp/commit/e0c2795bae8391c28be39662e8b22aa5f63860db#diff-c3a8fab52de096becf5cf3f7af032e76R647

PR #2382, opened 2016-03-03
Replace IAE with NPE for null parameters

mattlogan:
I replaced IAEs with NPEs for null parameter where a null value is prohibited, and updated existing tests.

There are a couple situations where `IllegalArgumentException` is thrown when a value inside a non-null collection parameter is null. For example, `Headers.of(Map<String, String> headers)` (line 186 in `Headers`) checks for null keys and/or values and throws an IAE. I left this as is, specifically because Effective Java tells us to use NPE when a "parameter value is null where prohibited" and use IAE when "non-null parameter value is inappropriate." Since the `Map` parameter itself is non-null and it's value (it's keys and values) is inappropriate, I think this should be kept as an IAE. I used the same reasoning for `Headers.of(String... namesAndValues)`.

Another tricky situation is in `Request.Builder.method(String method, RequestBody body)`, where we had the following check:

```
if (body == null && HttpMethod.requiresRequestBody(method)) {
  throw new IllegalArgumentException("method " + method + " must have a request body.");
}
```

If we stay true to the "[throw NPE] when a parameter value is null where prohibited" rule, then this should be changed to NPE, which I did change.

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/RequestTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/RequestTest.java
@@ -175,7 +175,7 @@ public final class RequestTest {
   }
 
   @Test public void headerForbidsControlCharacters() throws Exception {
-    assertForbiddenHeader(null);
+    assertNullHeader(null);
     assertForbiddenHeader("\u0000");
     assertForbiddenHeader("\r");
     assertForbiddenHeader("\n");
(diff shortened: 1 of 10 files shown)


okhttp-tests/src/test/java/okhttp3/RequestTest.java
```diff
@@ -175,7 +175,7 @@
   }
 
   @Test public void headerForbidsControlCharacters() throws Exception {
-    assertForbiddenHeader(null);
+    assertNullHeader(null);
```
JakeWharton: You can just delete this line and convert the new method into a test named `headerForbidsNullArguments`

okhttp/src/main/java/okhttp3/Request.java
```diff
@@ -229,14 +229,13 @@ public Builder patch(RequestBody body) {
...
-        throw new IllegalArgumentException("method == null || method.length() == 0");
-      }
+      if (method == null) throw new NullPointerException("method == null");
+      if (method.length() == 0) throw new IllegalArgumentException("method.length() == 0");
       if (body != null && !HttpMethod.permitsRequestBody(method)) {
         throw new IllegalArgumentException("method " + method + " must not have a request body.");
       }
       if (body == null && HttpMethod.requiresRequestBody(method)) {
-        throw new IllegalArgumentException("method " + method + " must have a request body.");
+        throw new NullPointerException("method " + method + " must have a request body.");
```
JakeWharton: I would argue that this should remain IAE since null is conditionally accepted based on other arguments.

okhttp-tests/src/test/java/okhttp3/RequestTest.java
mattlogan: Good idea

okhttp/src/main/java/okhttp3/Request.java
mattlogan: I'll defer to you on this one! My own interpretation of the Effective Java rules led me down the NPE path, but I agree that an NPE here would be confusing to the consumer.

swankjesse:
Merged manually!

Issue #2384, opened 2016-03-04
Log is not showing problem


morder:
Problem in file
> okhttp/okhttp/src/main/java/okhttp3/internal/Platform.java

method 
```
 @Override public void log(String message) {
      // Split by line, then ensure each line can fit into Log's maximum length.
      for (int i = 0, length = message.length(); i < length; i++) {
        int newline = message.indexOf('\n', i);
        newline = newline != -1 ? newline : length;
        do {
          int end = Math.min(newline, i + MAX_LOG_LENGTH);
          Log.d("OkHttp", message.substring(i, end));
          i = end;
        } while (i < newline);
      }
    }
```

if our request contains something like this

> --575082a3-3c3f-437a-9196-5f8451383ad5\r\n
Content-Disposition: form-data; name="api_key"\r\n
Content-Transfer-Encoding: binary\r\n
Content-Type: text/plain; charset=UTF-8\r\n
Content-Length: 31\r\n
\r\n
A06BHPUumm6FyYKz0xtC1LUqfkmwzAG\r\n
--575082a3-3c3f-437a-9196-5f8451383ad5\r\n
Content-Disposition: form-data; name="mobile"\r\n
Content-Transfer-Encoding: binary\r\n

no one line is showing because of \r\n in the end of each line
so problem have to be, if we have any \r or \n in printed line

I think we can fix it like this
```
Log.d("OkHttp", message.substring(i, end).replace("\r", "\\r").replace("\n", "\\n"));
```

swankjesse:
I think that if you want this, you should avoid OkHttp’s log method.

PR #2385, opened 2016-03-04
Log is not showing problem

morder:
fix for this issue 
https://github.com/square/okhttp/issues/2384


swankjesse:
I think that making newline separators visible introduces more problems than it fixes.

morder:
it just a logs

swankjesse:
I don’t think this is a good fit for OkHttp’s log method.

PR #2388, opened 2016-03-05
Gracefully recover from a failure to rebuild the DiskLruCache journal.

dave-r12:
I think I have something functional, but it's dependent on the answer to this question: if the rebuild fails, should we attempt to re-open the journal writer?

For now I said 'no' and treated the rebuild action as an atomic operation. That means writes are not allowed so we don't leak files. Reads are allowed and same with removals. If an entry is removed while in this state, and the cache is then closed, on a subsequent open the metadata will be inaccurate. To cope with this, an entry will be removed if we discover it's cache files are missing.

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/internal/DiskLruCacheTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/internal/DiskLruCacheTest.java
@@ -697,6 +697,199 @@ public final class DiskLruCacheTest {
     }
   }
 
+  @Test public void rebuildJournalFailurePreventsEditors() throws Exception {
+    while (executor.jobs.isEmpty()) {
+      set("a", "a", "a");
+      set("b", "b", "b");
+    }
+
+    // Cause the rebuild action to fail.
+    fileSystem.setFaultyRename(new File(cacheDir, DiskLruCache.JOURNAL_FILE_BACKUP), true);
+    executor.jobs.removeFirst().run();
+
+    // Don't allow edits under any circumstances.
+    assertNull(cache.edit("a"));
+    assertNull(cache.edit("c"));
+    DiskLruCache.Snapshot snapshot = cache.get("a");
+    assertNull(snapshot.edit());
+    snapshot.close();
+  }
+
+  @Test public void rebuildJournalFailureIsRetried() throws Exception {
+    while (executor.jobs.isEmpty()) {
+      set("a", "a", "a");
+      set("b", "b", "b");
+    }
+
+    // Cause the rebuild action to fail.
+    fileSystem.setFaultyRename(new File(cacheDir, DiskLruCache.JOURNAL_FILE_BACKUP), true);
+    executor.jobs.removeFirst().run();
+
+    // The rebuild is retried on cache hits and on cache edits.
+    DiskLruCache.Snapshot snapshot = cache.get("b");
+    snapshot.close();
+    assertNull(cache.edit("d"));
+    assertEquals(2, executor.jobs.size());
+
+    // On cache misses, no retry job is queued.
+    assertNull(cache.get("c"));
+    assertEquals(2, executor.jobs.size());
+
+    // Let the rebuild complete successfully.
+    fileSystem.setFaultyRename(new File(cacheDir, DiskLruCache.JOURNAL_FILE_BACKUP), false);
+    executor.jobs.removeFirst().run();
+    assertJournalEquals("CLEAN a 1 1", "CLEAN b 1 1");
+  }
+
+  @Test public void rebuildJournalFailureWithInFlightEditors() throws Exception {
+    while (executor.jobs.isEmpty()) {
+      set("a", "a", "a");
+      set("b", "b", "b");
+    }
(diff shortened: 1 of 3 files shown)


dave-r12:
@swankjesse @JakeWharton any feedback?

swankjesse:
Sorry for radio silence on this. Was hoping to take a look on Monday and life happened! Will get to it ASAP!

okhttp-tests/src/test/java/okhttp3/internal/DiskLruCacheTest.java
```diff
@@ -697,6 +697,199 @@ private void createNewCacheWithSize(int maxSize) throws IOException {
...
+
+    // Cause the rebuild action to fail.
+    fileSystem.setFaultyRename(new File(cacheDir, DiskLruCache.JOURNAL_FILE_BACKUP), true);
+    executor.jobs.removeFirst().run();
+
+    // The rebuild is retried on cache hits and on cache edits.
+    DiskLruCache.Snapshot snapshot = cache.get("b");
+    snapshot.close();
+    assertNull(cache.edit("d"));
+    assertEquals(2, executor.jobs.size());
```
swankjesse: Off topic - I wonder if we want to prevent enqueueing when it’s already enqueued. Not related to this PR.

okhttp-tests/src/test/java/okhttp3/internal/DiskLruCacheTest.java
```diff
@@ -697,6 +697,199 @@ private void createNewCacheWithSize(int maxSize) throws IOException {
...
+    setString(commitEditor, 1, "c");
+    commitEditor.commit();
+    assertValue("c", "c", "c");
+
+    abortEditor.abort();
+
+    cache.close();
+    createNewCache();
+
+    // Although 'c' successfully committed above, the journal wasn't available to issue a CLEAN op.
```
swankjesse: Nice comment

okhttp-tests/src/test/java/okhttp3/internal/DiskLruCacheTest.java
```diff
@@ -697,6 +697,199 @@ private void createNewCacheWithSize(int maxSize) throws IOException {
...
+    }
+
+    // Cause the rebuild action to fail.
+    fileSystem.setFaultyRename(new File(cacheDir, DiskLruCache.JOURNAL_FILE_BACKUP), true);
+    executor.jobs.removeFirst().run();
+
+    assertTrue(cache.remove("a"));
+    assertAbsent("a");
+
+    // Let the rebuild complete successfully.
```
swankjesse: Off topic – I wonder if it ever heals after breaking

okhttp-tests/src/test/java/okhttp3/internal/DiskLruCacheTest.java
```diff
@@ -697,6 +697,199 @@ private void createNewCacheWithSize(int maxSize) throws IOException {
...
+
+    cache.close();
+    createNewCache();
+
+    // The journal will have no record that 'a' was removed. It will have an entry for 'a', but when
+    // it tries to read the cache files, it will find they were deleted. Once it encounters an entry
+    // with missing cache files, it should remove it from the cache entirely.
+    assertEquals(4, cache.size());
+    assertNull(cache.get("a"));
+    assertEquals(2, cache.size());
```
swankjesse: This is really neat

okhttp/src/main/java/okhttp3/internal/DiskLruCache.java
```diff
@@ -460,8 +463,12 @@ private synchronized Editor edit(String key, long expectedSequenceNumber) throws
     if (entry != null && entry.currentEditor != null) {
       return null; // Another edit is in progress.
     }
-    if (mostRecentTrimFailed) {
-      // Prevent new writes so the cache doesn't grow any further and retry the clean up operation.
+    if (mostRecentTrimFailed || mostRecentRebuildFailed) {
+      // The OS has become our enemy! If the trim job failed, it means we are storing more data than
```
swankjesse: I love this phrase

swankjesse:
Thanks! And sorry for missing a week on this. I don't intend to make that a habit!

okhttp-tests/src/test/java/okhttp3/internal/DiskLruCacheTest.java
dave-r12: Do you mean on devices that hit this problem? I was trying to research the conditions that would cause the OS to _not_ allow file deletions, it seems really bizarre. I didn't come up with anything concrete.

swankjesse: Yeah. I have no idea whether deletion failures are temporary or permanent.

PieterAelse:
@swankjesse any estimates on a release date? If I'm correct there's no release yet where this fix is in.

swankjesse:
No estimates for releases. We typically like to release OkHttp every 2–3 months so we’re due.

PR #2390, opened 2016-03-08
Ensure HTTP 1XX response codes work properly

### LINKED ISSUE #1337: 1XX response codes should work properly
Making this work for uploads is worth doing. We should do it, being mindful of the consequences for http/2.0.

ameya-pandilwar:
Fixes #1337

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/http/Http1xStream.java
+++ b/okhttp/src/main/java/okhttp3/internal/http/Http1xStream.java
@@ -37,6 +37,7 @@ import okio.Timeout;
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
 import static okhttp3.internal.Util.checkOffsetAndCount;
 import static okhttp3.internal.http.StatusLine.HTTP_CONTINUE;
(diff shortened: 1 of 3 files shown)


okhttp/src/main/java/okhttp3/internal/http/Http1xStream.java
```diff
@@ -189,7 +190,7 @@ public void writeRequest(Headers headers, String requestLine) throws IOException
             .message(statusLine.message)
             .headers(readHeaders());
 
-        if (statusLine.code != HTTP_CONTINUE) {
+        if (statusLine.code != HTTP_CONTINUE || statusLine.code != HTTP_PROCESSING) {
```
swankjesse: Shouldn't this be `&&` instead of `||`?

swankjesse:
Test case please.

dave-r12:
This one looks almost complete. I'll try to finish it up with a test case.

swankjesse:
Let’s start a new PR from scratch if we pick up this work.

PR #2391, opened 2016-03-08
PROPFIND redirects should maintain request body

rburgst:
the fact that all redirects strip the request body is somewhat unusual.
In WEBDAV the body contains restrictions such as the depth of how much
data you want to receive. Stripping this off will produce an unlimited
request with unlimited depth (where the initial request had depth 1).

* havent checked other WEBDAV methods though whether they require similar
  treatment.

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/CallTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/CallTest.java
@@ -1503,17 +1503,21 @@ public final class CallTest {
     assertEquals("Hello", request2.getBody().readUtf8());
   }
 
-  @Test public void propfindRedirectsToPropfind() throws Exception {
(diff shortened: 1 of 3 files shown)


swankjesse:
Sorry for sitting on this for forever. I’m just going through old pull requests and I hadn’t realized we’d been completely silent on this one. Sorry about that.

Reading through the change it looks like a good change. It’s weird that Webdav works that way, but it doesn’t really complicate OkHttp either so it doesn’t bother me.

@rburgst next steps – I’m assuming you’ve probably grown tired of waiting and have moved on. If you haven’t, please rebase this change and I’ll merge. Quickly. Otherwise I’ll take a look at rebasing it.

(One other thing is that since we rearranged HttpURLConnection it’s now much safer to replay the request body. Previously that was impossible in some situations, which would have complicated things.)

rburgst:
rebased the patch. Would still appreciate the merge :)

Issue #1711, opened 2015-06-20
Client Certificates


tofutim:
Is there a way of using Client Certificates with okhttp?

swankjesse:
Probably. But I haven't done it, and we don't have any guides on how to do it, or test cases that keep it working. If you'd like to do the legwork of figuring that out, I'd love to see a sample.

Most of the work is going to go in the SSL socket factory code, which OkHttp mostly just delegates to.

tofutim:
Ok, I don't know much about Java, but let me see what I can find.

tofutim:
In OkHttpClient.java in okhttp/src/main
```
  private synchronized SSLSocketFactory getDefaultSSLSocketFactory() {
    if (defaultSslSocketFactory == null) {
      try {
        SSLContext sslContext = SSLContext.getInstance("TLS");
        sslContext.init(null, null, null);
        defaultSslSocketFactory = sslContext.getSocketFactory();
      } catch (GeneralSecurityException e) {
        throw new AssertionError(); // The system has no TLS. Just give up.
      }
    }
    return defaultSslSocketFactory;
  }
```

Hmm, what does this mean: 

"This code avoids that by defaulting to an OkHttp-created SSL context.
   * The drawback of this approach is that apps that customize the global SSL
   * context will lose these customizations."

sslContext.init looks like
```
public final void init(KeyManager[] km,
        TrustManager[] tm,
        SecureRandom random)
```
According to http://chariotsolutions.com/blog/post/https-with-client-certificates-on/, the KeyManager is what will provides the client certificate. Given a pfx,
```
keyStore = KeyStore.getInstance("PKCS12");
fis = new FileInputStream(certificateFile);<br>keyStore.load(fis, clientCertPassword.toCharArray());
```
```
KeyManagerFactory kmf = KeyManagerFactory.getInstance("X509");
kmf.init(keyStore, clientCertPassword.toCharArray());
KeyManager[] keyManagers = kmf.getKeyManagers();
SSLContext sslContext = SSLContext.getInstance("TLS");
sslContext.init(keyManagers, null, null);
```

tofutim:
Maybe something like below. Also could be extended to allow for specifying your own ca.

```
private synchronized SSLSocketFactory getDefaultSSLSocketFactory(
		  byte[] pksData, String password) {
	    if (defaultSslSocketFactory == null) {
	      try {
	    	InputStream stream = new ByteArrayInputStream(pksData);	    	
[...]

PR #2393, opened 2016-03-08
[WIP] Use key store from default SSLContext, if SSLSocketFactory is not set

tomfitzhenry:
Related: issue #1711.

### DIFF
--- a/mockwebserver/src/main/java/okhttp3/internal/SslContextBuilder.java
+++ b/mockwebserver/src/main/java/okhttp3/internal/SslContextBuilder.java
@@ -116,7 +116,7 @@ public final class SslContextBuilder {
     return sslContext;
   }
 
-  private KeyStore newEmptyKeyStore(char[] password) throws GeneralSecurityException {
+  public static KeyStore newEmptyKeyStore(char[] password) throws GeneralSecurityException {
     try {
       KeyStore keyStore = KeyStore.getInstance(KeyStore.getDefaultType());
       InputStream in = null; // By convention, 'null' creates an empty key store.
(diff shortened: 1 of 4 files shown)


swankjesse:
Is this still a work in progress?

tomfitzhenry:
Yes. It'll be finished when OkHttp uses the default KeyStore, or if that
isn't a desirable fix, the test will be changed to use the default
SSLSocketFactory, in which case the test will start passing.

Either way it'll be a test that documents the recommended way to do client
certificates in OkHttp.

There's some tidying to do too. I opened the PR, despite it being
incomplete, for early feedback and in case others found it useful. I can
remove it, if having a broken PR is inconvenient.

swankjesse:
Could you turn this into a sample? We've got a samples directory where this would fit neatly.

Any way you can change your code to configure the keystore programmatically instead of with system properties? The problem with system properties is that they're global, which is unlikely to work in applications that make multiple kinds of HTTP requests.

tomfitzhenry:
Thanks for prompt (modulo time zone) responses.

Yes, I can make it a sample instead of a test case. Why do you prefer it be
a sample, rather than a test, OOC?

With regards to programmatically configuring keystores, do you mean my
application code, or the code in the pull request?

If the former then the code that uses OkHttp is sonar-runner-api (used by
maven-sonar-plugin). Users expect to be able to pass client certificates
using the standard javax.net.ssl.keyStore system property, when executing
mvn.

Or do you mean the latter, and for example to add clientCertificate method
to OkHttp.Builder? I could. My concern with that is that other applications
that use OkHttp will not support client certificates (this tends to not be
considered by tool developers), making it hard for client certificate users.

What do you mean by "multiple kinds of HTTP requests"? I'll have a think
what I expect from a HTTP library API re client certificates, as if I had
multiple client certificates, and I wanted things like connection pooling
to work in some sane way. It's not immediately obvious what a good API for
that is.

swankjesse:
> Why do you prefer it be a sample, rather than a test, OOC?

If the goal is documentation, samples feel less contrived. We expect end users to cut & paste samples into their projects. Samples can also hit webservers on the public Internet.


> With regards to programmatically configuring keystores, do you mean my application code, or the code in the pull request?

Yes, application code. I don’t think system properties are a good idea.


> Users expect to be able to pass client certificates using the standard javax.net.ssl.keyStore system property, when executing mvn.

If the Maven plugin supports this, then it should have its own configuration mechanism. Even just reading the supported system properties and configuring OkHttp with them should be sufficient.

swankjesse:
Closing as stale. Happy to accept again if you’d like to resubmit without system properties!

Issue #2288, opened 2016-01-27
Network interceptor connection() might be null


swankjesse:
For example, if looking up the connection after a request is completed, and it has both `Connection: close` and an empty response body.

See https://github.com/facebook/stetho/issues/346

swankjesse:
(We should fix network interceptors to _never_ return null for the connection, even after the response has been returned.)

arriolac:
Related to this crash?
```
 E  java.lang.NullPointerException: Attempt to invoke virtual method 'long okhttp3.ResponseBody.contentLength()' on a null object reference
                         E      at okhttp3.logging.HttpLoggingInterceptor.intercept(HttpLoggingInterceptor.java:207)
                         E      at okhttp3.RealCall$ApplicationInterceptorChain.proceed(RealCall.java:187)
                         E      at okhttp3.RealCall.getResponseWithInterceptorChain(RealCall.java:160)
                         E      at okhttp3.RealCall.access$100(RealCall.java:30)
                         E      at okhttp3.RealCall$AsyncCall.execute(RealCall.java:127)
                         E      at okhttp3.internal.NamedRunnable.run(NamedRunnable.java:33)
                         E      at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1113)
                         E      at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:588)
                         E      at java.lang.Thread.run(Thread.java:818)
```

dave-r12:
Closed with #2403.

swankjesse:
Looks like there’s still a case where null is coming through to the network interceptor.

swankjesse:
(Details in the Stetho report.)

dave-r12:
Am I missing something? RealCall doesn't have a connection() method. ApplicationInterceptorChain does and it returns null. Stetho is using a NetworkInterceptor.

dave-r12:
I responded on the stetho issue, I believe that person is referring to the ApplicationInterceptorChain class. Returning null for connection() is deliberate. The wiki mentions connection() is available with network interceptors only. Let me know if I'm misunderstanding something.

swankjesse:
Ah cool. Thanks!

PR #2403, opened 2016-03-13
Ensure network interceptors always have access to the underlying connection.

dave-r12:
This is for #2288 - not entirely sure if this is what you intended.


swankjesse:
Yes, this is perfect. Thanks!

PR #2405, opened 2016-03-15
Add ability to specify X509TrustManager via builder

artem-zinnatullin:
Addresses #2323.

This is WIP and so on. Initially I though that interface will be needed, but since trust manager required at `OkHttpClient` creation time, we can accept it right in the builder.

Also, allows user to save some reflection calls if they have direct reference to the trust manager.

Probably this is totally not the direction you would like to go, so feel free to close.

---
BTW, maybe better to allow user pass `Platform` implementation? This will open so many possibilities to support any kind of environment without modifying the library.

### DIFF
--- a/okhttp/src/main/java/okhttp3/OkHttpClient.java
+++ b/okhttp/src/main/java/okhttp3/OkHttpClient.java
@@ -181,7 +181,9 @@ public class OkHttpClient implements Cloneable, Call.Factory {
       }
     }
     if (sslSocketFactory != null && builder.trustRootIndex == null) {
-      X509TrustManager trustManager = Platform.get().trustManager(sslSocketFactory);
+      X509TrustManager trustManager = builder.x509TrustManager == null
+          ? Platform.get().trustManager(sslSocketFactory)
+          : builder.x509TrustManager;
       if (trustManager == null) {
         throw new IllegalStateException("Unable to extract the trust manager on " + Platform.get()
             + ", sslSocketFactory is " + sslSocketFactory.getClass());
@@ -353,6 +355,7 @@ public class OkHttpClient implements Cloneable, Call.Factory {
     int connectTimeout;
     int readTimeout;
     int writeTimeout;
+    X509TrustManager x509TrustManager;
 
     public Builder() {
       dispatcher = new Dispatcher();
@@ -711,6 +714,11 @@ public class OkHttpClient implements Cloneable, Call.Factory {
 
     public Builder addNetworkInterceptor(Interceptor interceptor) {
       networkInterceptors.add(interceptor);
+      return this;
+    }
+
+    public Builder x509TrustManager(X509TrustManager x509TrustManager) {
+      this.x509TrustManager = x509TrustManager;
       return this;
     }
 


okhttp/src/main/java/okhttp3/OkHttpClient.java
```diff
@@ -714,6 +717,11 @@ public Builder addNetworkInterceptor(Interceptor interceptor) {
       return this;
     }
 
+    public Builder x509TrustManager(X509TrustManager x509TrustManager) {
```
artem-zinnatullin: Not sure how to test this properly, looks like it'll require mock a lot of objects

artem-zinnatullin:
Any feedback…? #2427 has similar points about overridable `x509TrustManager` and  providing `Platform` implementation.

swankjesse:
We want something similar to this. I’m not yet sure how I want the API to look.

artem-zinnatullin:
If you don't mind I can investigate the ability of accepting `Platform` implementation, it looks like a better and more flexible solution that will allow OkHttp work almost in any environment without changes in the library.

// Thinking further: you may then finally split platform dependencies and ship `platform-android`/etc jars separately without need in reflection for simple actions, so many possibilities!

JakeWharton:
We don't need reflection currently, it's more of a legacy implementation.
Retrofit's Platform behavior is reflection-free and we'll get around to
porting it to OkHttp at some point.

JakeWharton:
I'm also very against a separate artifact required for correct use on
Android and I'll need convincing that Platform is worth exposing. There's a
lot of engine parts in there.

swankjesse:
Definitely no plans to make Platform user-visible. Platform is where the sausage is made.

artem-zinnatullin:
Well, doesn't `Platform` look like Converters/Adapters API in Retrofit 2? I mean, pluggability of those APIs and your point about maintaining third-party convertors/adapters outside of the library.

At the moment [`Platform`](https://github.com/square/okhttp/pull/2405#issuecomment-200114196) has  bindings to: ~default JDK, `Android`, `JdkWithJettyBootPlatform`. #2427 needs integration with IBM JDK, some fix was recently required for Robolectric and who knows what could be needed next?

I don't have strong opinion about moving Android bindings to separate jar, totally up to you. But it will give direct access to `android.util.Log` at least haha. Also, probably @nfuller will find something that they will be able to do better with direct access to Android SDK classes (for those who uses external version of OkHttp), idk.

Anyway, all current `Platform` implementations and default `findPlatform()` mechanism may be left in the library as it's now and you can just give an api to set custom `Platform` implementation so you won't need to care about weird environments.

// btw, making call to `findPlatform()` skippable may speedup startup of thousands Android apps, not too much, but some amount of milliseconds probably.

If my words don't make sense — please close the PR.

JakeWharton:
> Well, doesn't Platform look like Converters/Adapters API in Retrofit 2?

No. Those are interceptors, authenticators, etc. in OkHttp.

> At the moment Platform has bindings to: ~default JDK, Android, JdkWithJettyBootPlatform. #2427 needs integration with IBM JDK, some fix was recently required for Robolectric and who knows what could be needed next?

The fixes are not because of Platform. That just happens to be where the new code path using non-exposed APIs happens to live. That's what this whole PR is meant to address.

> I don't have strong opinion about moving Android bindings to separate jar, totally up to you. But it will give direct access to android.util.Log at least haha. Also, probably @nfuller will find something that they will be able to do better with direct access to Android SDK classes (for those who uses external version of OkHttp), idk.

See above comment about Retrofit's `Platform`. This isn't hard to get.

> Anyway, all current Platform implementations and default findPlatform() mechanism may be left in the library as it's now and you can just give an api to set custom Platform implementation so you won't need to care about weird environments.
>
> btw, making call to findPlatform() skippable may speedup startup of thousands Android apps, not too much, but some amount of milliseconds probably.

Two things here:

 * It's mostly an irrelevant amount of milliseconds. This PR addresses the thing which is 100x-1000x slower than `findPlatform()`.
 * OkHttpClient doesn't need to be initialized synchronously. It should be deferred until first use. I have a blog post on this coming next week.

swankjesse:
We’ve got the best parts of this here:
https://github.com/square/okhttp/pull/2446

Issue #2414, opened 2016-03-18
[Bug report] Proxy authenticate over http.


smatei:
I have successfully tested https requests with the code

```
   String run(String url) throws IOException {
    String proxy = "xxxxxxxxxxx";
    String user = "xxxxxxx";
    String password = "xxxxxxxxx";

    Authenticator proxyAuthenticator = new Authenticator() {
      public Request authenticate(Route route, Response response) throws IOException {
           String credential = Credentials.basic(user, password);
           return response.request().newBuilder()
               .header("Proxy-Authorization", credential)
               .build();
      }
    };
    
    OkHttpClient client = new OkHttpClient.Builder()
        .protocols(Arrays.asList(Protocol.HTTP_2, Protocol.HTTP_1_1, Protocol.SPDY_3))
        .connectTimeout(60, TimeUnit.SECONDS)
        .writeTimeout(60, TimeUnit.SECONDS)
        .readTimeout(60, TimeUnit.SECONDS)
        .proxy(new Proxy(Proxy.Type.HTTP, new InetSocketAddress(proxy, 80)))
        .proxyAuthenticator(proxyAuthenticator)
        .build();
    
    Request request = new Request.Builder()
        .url(url)
        .header("User-Agent", "Mozilla/5.0 (Windows NT 6.2; WOW64; rv:43.0) Gecko/20100101 Firefox/43.0")
        .build();

    Response response = client.newCall(request).execute();
    System.out.println("response " + response.code());
    return response.body().string();
  }

  public static void main(String[] args) throws IOException {
    testok example = new testok();
    String response = example.run("https://www.google.com");
    System.out.println(response);
  }
```

The authentication works just fine, I get 200 response code. Now I want to get 

`String response = example.run("http://www.bing.com");`

instead of 

`String response = example.run("https://www.google.com");`

I get response 407. In the HttpEngine class, followUpRequest method, 

```
     switch (responseCode) {
      case HTTP_PROXY_AUTH:
        Proxy selectedProxy = route != null
            ? route.proxy()
            : client.proxy();
        if (selectedProxy.type() != Proxy.Type.HTTP) {
          throw new ProtocolException("Received HTTP_PROXY_AUTH (407) code while not using proxy");
        }
        // fall-through
      case HTTP_UNAUTHORIZED:
        return client.authenticator().authenticate(route, userResponse);`
```
[...]

PR #2417, opened 2016-03-19
Fix a regression in authenticating HTTP proxies.

swankjesse:
We were using the regular authenticator, not the proxy authenticator.

Closes: https://github.com/square/okhttp/issues/2414


JakeWharton:
LGTM. Test failures?

swankjesse:
We had a broken test! Fixed that too.

PR #2421, opened 2016-03-20
Recover more gracefully when an editor is detached.

swankjesse:
We had a bug where we could have two editors for the same entry. This
would occur when the cache was cleared, or if the entry was otherwise
evicted while the edit was in progress.

Previously the two editors would corrupt each other, and potentially
the cache's size. With this change the detached editor is limited and
harmless.

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/internal/DiskLruCacheTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/internal/DiskLruCacheTest.java
@@ -1536,6 +1536,80 @@ public final class DiskLruCacheTest {
     assertNull(cache.get("a"));
   }
 
+  @Test public void noSizeCorruptionAfterCreatorDetached() throws Exception {
+    // Create an editor for k1. Detach it by clearing the cache.
+    DiskLruCache.Editor editor = cache.edit("k1");
+    setString(editor, 0, "a");
+    setString(editor, 1, "a");
+    cache.evictAll();
+
+    // Create a new value in its place.
+    set("k1", "bb", "bb");
+    assertEquals(4, cache.size());
+
+    // Committing the detached editor should not change the cache's size.
+    editor.commit();
+    assertEquals(4, cache.size());
+    assertValue("k1", "bb", "bb");
+  }
+
+  @Test public void noSizeCorruptionAfterEditorDetached() throws Exception {
+    set("k1", "a", "a");
+
+    // Create an editor for k1. Detach it by clearing the cache.
+    DiskLruCache.Editor editor = cache.edit("k1");
+    setString(editor, 0, "bb");
+    setString(editor, 1, "bb");
+    cache.evictAll();
+
+    // Create a new value in its place.
+    set("k1", "ccc", "ccc");
+    assertEquals(6, cache.size());
+
+    // Committing the detached editor should not change the cache's size.
+    editor.commit();
+    assertEquals(6, cache.size());
+    assertValue("k1", "ccc", "ccc");
+  }
+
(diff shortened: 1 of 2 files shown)


okhttp-tests/src/test/java/okhttp3/internal/DiskLruCacheTest.java
```diff
@@ -1536,6 +1536,78 @@ private void createNewCacheWithSize(int maxSize) throws IOException {
...
+    setString(editor, 0, "a");
+    setString(editor, 1, "a");
+    cache.evictAll();
+
+    // Create a new value in its place.
+    set("k1", "bb", "bb");
+
+    // Committing the detached editor should not change the cache's size.
+    editor.commit();
+    assertEquals(4, cache.size());
```
JakeWharton: Can we assert this before as well so that "should not change" is materialized in the test?

okhttp-tests/src/test/java/okhttp3/internal/DiskLruCacheTest.java
```diff
@@ -1536,6 +1536,78 @@ private void createNewCacheWithSize(int maxSize) throws IOException {
     assertNull(cache.get("a"));
   }
 
+  @Test public void noSizeCorruptionAfterCreatorDetached() throws Exception {
+    // Create an editor for k1. That editor's write privileges away by clearing the cache.
```
JakeWharton: Comment missing a verb

JakeWharton: "Take that..." ?

dave-r12:
Nice find!

JakeWharton:
lgtm

okhttp-tests/src/test/java/okhttp3/internal/DiskLruCacheTest.java
swankjesse: Good call. Done x 2.

okhttp-tests/src/test/java/okhttp3/internal/DiskLruCacheTest.java
swankjesse: Fixed.

mowenGithub:
I have the same bug with okhttp:3.12.1，has it fixed in 3.12.x version？

OkHttp DiskLruCache java.util.NoSuchElementException

java.util.LinkedHashMap$LinkedHashIterator.nextNode(LinkedHashMap.java:762)
java.util.LinkedHashMap$LinkedValueIterator.next(LinkedHashMap.java:788)
okhttp3.internal.cache.DiskLruCache.trimToSize(DiskLruCache.java:684)
okhttp3.internal.cache.DiskLruCache$1.run(DiskLruCache.java:177)
java.util.concurrent.ThreadPoolExecutor.processTask(ThreadPoolExecutor.java:1187)
java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1152)
java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
java.lang.Thread.run(Thread.java:784)

yschimke:
@mowenGithub  Do you mean 3.12.10?  Please let us know if you can still reproduce with 3.12.10.

mowenGithub:
> @mowenGithub Do you mean 3.12.10? Please let us know if you can still reproduce with 3.12.10.

No, I used 3.12.6,I will try 3.12.10 and reply the result. Thank you!

Mryushen:
I have the same bug with okhttp:3.14.9

PR #2437, opened 2016-03-25
Add some useful badges with automatic version detection: Maven and Javadoc

valery1707:


swankjesse:
Not a big fan of badges. Thanks anyway.

PR #2439, opened 2016-03-25
Fix javadoc on setProtocols method.

IgorGanapolsky:


swankjesse:
What’s wrong with the current doc?

swankjesse:
(maybe we want `{@linkplain ...}` instead of `{@link ...}`)

IgorGanapolsky:
I believe the curly brace was at the wrong place for @link reference in
that method.

JakeWharton:
+1 to `{@linkplain ..}`. The `}` was in the correct place.

IgorGanapolsky:
Updated to linkplain.

PR #2442, opened 2016-03-26
http age calculations wrong because of phone time fall back

lightai:
Handle the situation: user adjust system time(fall back),we don't request network timely


swankjesse:
Just so I’m clear: is this an attempt to deal with users overriding their system clocks to set the time into the future, then setting it back?

lightai:
yes. this situation can lead to okhttp use local cache longer time than expected

swankjesse:
I don’t think this is the right solution to the problem. It’s difficult to know whether the time is currently wrong, or if it was wrong at the time of the response. And this change means that small adjustments to the system time can break the cache. Typically small adjustments are the only ones we expect to see.

PR #2444, opened 2016-03-28
Igor static.analysis

IgorGanapolsky:
Reformat code.

### DIFF
--- a/mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.java
+++ b/mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.java
@@ -17,10 +17,6 @@
 
 package okhttp3.mockwebserver;
 
-import org.junit.rules.TestRule;
-import org.junit.runner.Description;
-import org.junit.runners.model.Statement;
-
 import java.io.IOException;
 import java.net.InetAddress;
 import java.net.InetSocketAddress;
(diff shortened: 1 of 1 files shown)


mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.java
```diff
@@ -310,7 +314,7 @@ public void start(int port) throws IOException {
    * @param port the port to listen to, or 0 for any available port. Automated tests should always
    * use port 0 to avoid flakiness when a specific port is unavailable.
    */
-  public void start(InetAddress inetAddress, int port) throws IOException {
+  private void start(InetAddress inetAddress, int port) throws IOException {
```
artem-zinnatullin: This is public API

IgorGanapolsky: The InetAddress param method is exposed to public?

swankjesse:
Not sure what the goal is here.

IgorGanapolsky:
@swankjesse You don't believe in static analysis?

swankjesse:
I’m sorry, your PR appears to just reformat the code. I don't believe in reformatting the code.

IgorGanapolsky:
There were a couple of methods that were throwing exceptions for no
reason.  And a couple of generic types that do not need to be there.  And
imports that weren't organized into proper blocks.

JakeWharton:
Imports should not be put into blocks.

As for the other things, submit individual PRs which change only the
problematic section.

Issue #2453, opened 2016-03-31
[Question] Proxy connection always fail with unexpected end of stream.


smatei:
Hi,

On some proxies (squid proxy) OKHttpClient works well. On others (also squid proxies), for an unknown reason, I **always** get the **"unexpected end of stream**" error (Caused by: java.io.EOFException: \n not found: size=0 content). For all the browsers and other http clients (java, apache, jetty), the proxy works. 

`Exception in thread "main" java.io.IOException: unexpected end of stream
	at okhttp3.internal.http.Http1xStream.readResponse(Http1xStream.java:215)
	at okhttp3.internal.http.Http1xStream.readResponseHeaders(Http1xStream.java:127)
	at okhttp3.internal.http.HttpEngine.readNetworkResponse(HttpEngine.java:732)
	at okhttp3.internal.http.HttpEngine.access$2(HttpEngine.java:726)
	at okhttp3.internal.http.HttpEngine$NetworkInterceptorChain.proceed(HttpEngine.java:714)
	at okhttp3.internal.http.HttpEngine.readResponse(HttpEngine.java:566)
	at okhttp3.RealCall.getResponse(RealCall.java:245)
	at okhttp3.RealCall$ApplicationInterceptorChain.proceed(RealCall.java:200)
	at okhttp3.RealCall.getResponseWithInterceptorChain(RealCall.java:162)
	at okhttp3.RealCall.execute(RealCall.java:59)
	at testok.run(testok.java:54)
	at testok.main(testok.java:61)
Caused by: java.io.EOFException: \n not found: size=0 content=...
	at okio.RealBufferedSource.readUtf8LineStrict(RealBufferedSource.java:196)
	at okhttp3.internal.http.Http1xStream.readResponse(Http1xStream.java:191)
	... 11 more`

I have tried to debug it using the OKHttp and okio sources, by sniffing with Wireshark, but no luck. Somehow the server response is missing. How can I debug this?

The code that I use is quite simple.

`    Authenticator proxyAuthenticator = new Authenticator() {
      public Request authenticate(Route route, Response response) throws IOException {
           String credential = Credentials.basic(user, password);
           return response.request().newBuilder()
               .header("Proxy-Authorization", credential)
               .build();
      }
    };
    
    OkHttpClient client = new OkHttpClient.Builder()
        .protocols(Arrays.asList(Protocol.HTTP_2, Protocol.HTTP_1_1, Protocol.SPDY_3))
[...]

Issue #2435, opened 2016-03-25
Preemptive proxy authentication


smatei:
Hi,

Is there a way to make a basic proxy authentication preemptive? By default, the request is done in RealCall.getResponse - engine.sendRequest(), then HttpEngine.followUpRequest is called that detects responseCode HTTP_PROXY_AUTH and then the authentication is done

return client.proxyAuthenticator().authenticate(route, userResponse);

(https://github.com/square/okhttp/pull/2417, https://github.com/square/okhttp/pull/2414)

Then another sendRequest is called that is successful.

Is there a way to force authenticate from the first sendRequest? Because I know my proxy needs authentication, I don't need the server to tell me this.

In apache http client I can set authentication preemptive. 

aClient.getParams().setAuthenticationPreemptive(true);

Do  you have such a thing?

Stefan Matei

swankjesse:
Nothing at the moment. We should offer something here.

smatei:
Please add it as a feature request.

Stefan Matei

smatei:
Hi, 

Figured out how to authenticate directly. Added the "Proxy-Authorization" header to my original request and everything is working perfect.

`String credential = Credentials.basic(user, password);
    
    Request request = new Request.Builder()
        .url(url)
        .header("User-Agent", "Mozilla/5.0 (Windows NT 6.2; WOW64; rv:43.0) Gecko/20100101 Firefox/43.0")
        .header("Proxy-Authorization", credential)
        .build();`

swankjesse:
Considering the APIs for this. We already have this:
```
interface Authenticator {
  Request authenticate(Route route, Response response);
```
I’m tempted by this:
```
class AbstractAuthenticator implements Authenticator {
  Request authenticate(Request request);
  Request authenticate(Route route, Response response);
```
Certainly the name is terrible. But I like having both proactive and reactive authentication in one interface. And we can use it for both proxies and direct requests.

jsoref:
@swankjesse: Did you mean `interface` for your `AbstractAuthenticator` as opposed to `class`?
ala:
`public interface AbstractAuthenticator extends Authenticator {`

bsia:
Is it possible to do pre-emptive proxy authentication via interceptors currently?

gupta11:
So using OkHttp how do we connect through a proxy on the wifi network? All suggestions I have tried so far have failed. I am using retrofit and okhttp3.

nicktrav:
Bumping this.

[...]

Issue #2426, opened 2016-03-22
Support of "Connection close" headers for HTTPS through proxy


axel3rd:
Hi, 

This issue could be linked to #1517, but is here specific to https through proxy (So I allow myself to open a new one).

**Synthesis** : The client seems not respect `Connection: close` header come from a proxy when no preemptive authentication on a HTTPS connection.

Consider this snippet code (simple HTTP GET through a proxy) :
```
        Authenticator.setDefault(new Authenticator() {
            protected PasswordAuthentication getPasswordAuthentication() {
                return new PasswordAuthentication(USER, PASSWORD.toCharArray());
            }
        });

        Proxy PROXY = new Proxy(Type.HTTP, new InetSocketAddress(PROXY_URL, 80));
        OkHttpClient okHttpClient = new OkHttpClient();
        okHttpClient.setProxy(PROXY);
        Request request = new Request.Builder().url("http://www.google.org/").build();
        Response response = okHttpClient.newCall(request).execute();
        System.out.println(response.code());
        response.body().close();
```
The call works fine and the network trace (wireshark) is : 
```
GET http://www.google.org/ HTTP/1.1
Host: www.google.org
Connection: Keep-Alive
Accept-Encoding: gzip
User-Agent: okhttp/2.6.0

HTTP/1.1 407 Proxy Authentication Required
Mime-Version: 1.0
Date: Tue, 22 Mar 2016 19:26:29 CET
Content-Type: text/html
Proxy-Authenticate: Basic realm="[deleted]"
Connection: close
Proxy-Connection: close
Content-Length: 1723

[HTML content]
```
And **in a second TCP stream** : 
```
GET http://www.google.org/ HTTP/1.1
Proxy-Authorization: Basic [deleted]=
Host: www.google.org
Connection: Keep-Alive
Accept-Encoding: gzip
User-Agent: okhttp/2.6.0

HTTP/1.1 200 OK
[correct headers]
```

Now consider the same snippet code on HTTPS => **https://www.google.org/**, the call fails with Java stackstrace : 
```
java.io.IOException: unexpected end of stream on Connection{www.google.org:443, proxy=HTTP @ [deleted] hostAddress=[deleted] cipherSuite=none protocol=null} (recycle count=0)
	at com.squareup.okhttp.internal.http.HttpConnection.readResponse(HttpConnection.java:211)
 [...]
Caused by: java.io.EOFException: \n not found: size=0 content=...
[...]

PR #2458, opened 2016-04-01
preemptive authentication

smatei:
Added preemptive authentication methods to the Authenticator interface.

https://github.com/square/okhttp/issues/2435
https://github.com/square/okhttp/issues/2453

Called them in the RealConnection.createTunnel method for the https requests, and HttpEngine.networkRequest for the http requests.

Now my Authenticator looks like this:

`    Authenticator proxyAuthenticator = new Authenticator() {
      public Request authenticate(Route route, Response response) throws IOException {
         // no longer need to implement, as I will not receive a 407 any more
         return null;
      }

      @Override
      public Request authenticatePreemptive(Route route, Request request)
          throws IOException
      {
        String credential = Credentials.basic(user, password);
        return request.newBuilder()
            .header("Proxy-Authorization", credential)
            .build();
      }
      
      @Override
      public boolean isPreemptive()
      {
        return true;
      }
    };`

This works for me. No need for two requests any more, one with 407 response and one with authenticate header.

By the way, even without my preemptive mechanism, if the user/password is wrong, I get 

Exception in thread "main" java.net.ProtocolException: Too many follow-up requests: 21

The exception is not so friendly, maybe you can improve this.

### DIFF
--- a/okhttp/src/main/java/okhttp3/Authenticator.java
+++ b/okhttp/src/main/java/okhttp3/Authenticator.java
@@ -52,6 +52,16 @@ public interface Authenticator {
     @Override public Request authenticate(Route route, Response response) {
       return null;
     }
+
+	@Override
+	public Request authenticate(Route route, Request request) throws IOException {
+		return null;
+	}
+
+	@Override
+	public boolean isPreemptive() {
+		return false;
+	}
   };
 
   /**
@@ -59,4 +69,8 @@ public interface Authenticator {
    * response}. Returns null if the challenge cannot be satisfied.
    */
   Request authenticate(Route route, Response response) throws IOException;
+  
+  Request authenticate(Route route, Request request) throws IOException;
+  
+  boolean isPreemptive();  
 }
--- a/okhttp/src/main/java/okhttp3/Authenticator.java
+++ b/okhttp/src/main/java/okhttp3/Authenticator.java
@@ -54,7 +54,7 @@ public interface Authenticator {
     }
 
 	@Override
-	public Request authenticate(Route route, Request request) throws IOException {
+	public Request authenticatePreemptive(Route route, Request request) throws IOException {
 		return null;
 	}
 
@@ -70,7 +70,15 @@ public interface Authenticator {
    */
   Request authenticate(Route route, Response response) throws IOException;
   
-  Request authenticate(Route route, Request request) throws IOException;
+  /**
+   * Return a request that is a copy of the original request plus authentication header.
+   * isPreemptive method must return true.
+   * 
+   */
+  Request authenticatePreemptive(Route route, Request request) throws IOException;
   
+  /**
+   *  If preemptive authentication is needed, implement this return true.
+   */
   boolean isPreemptive();  
 }
--- a/okhttp/src/main/java/okhttp3/Authenticator.java
+++ b/okhttp/src/main/java/okhttp3/Authenticator.java
@@ -69,16 +69,16 @@ public interface Authenticator {
    * response}. Returns null if the challenge cannot be satisfied.
    */
   Request authenticate(Route route, Response response) throws IOException;
-  
+
   /**
    * Return a request that is a copy of the original request plus authentication header.
    * isPreemptive method must return true.
-   * 
+   *
    */
   Request authenticatePreemptive(Route route, Request request) throws IOException;
-  
+
   /**
    *  If preemptive authentication is needed, implement this return true.
    */
-  boolean isPreemptive();  
+  boolean isPreemptive();
 }
--- a/okhttp/src/main/java/okhttp3/Authenticator.java
+++ b/okhttp/src/main/java/okhttp3/Authenticator.java
@@ -53,15 +53,15 @@ public interface Authenticator {
       return null;
     }
 
-	@Override
-	public Request authenticatePreemptive(Route route, Request request) throws IOException {
-		return null;
-	}
+    @Override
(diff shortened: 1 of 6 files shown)


okhttp/src/main/java/okhttp3/Authenticator.java
```diff
@@ -52,11 +52,33 @@
...
   };
 
   /**
    * Returns a request that includes a credential to satisfy an authentication challenge in {@code
    * response}. Returns null if the challenge cannot be satisfied.
    */
   Request authenticate(Route route, Response response) throws IOException;
+
+  /**
+   * Return a request that is a copy of the original request plus authentication header.
```
yschimke: Not all authentication needs to be in the form of headers.  e.g. an oauth token as part of the query params.

okhttp/src/main/java/okhttp3/Authenticator.java
```diff
@@ -52,11 +52,33 @@
...
+   * isPreemptive method must return true.
+   *
+   */
+  Request authenticatePreemptive(Route route, Request request) throws IOException;
```
yschimke: These two methods will break binary compatibility for existing implementors. Since OkHttp only allows a single Authenticator, but may be servicing multiple hosts concurrently, maybe the best compromise is an extended interface?

okhttp/src/main/java/okhttp3/internal/http/HttpEngine.java
```diff
@@ -499,6 +500,13 @@ public static boolean hasBody(Response response) {
    * content types the application is interested in.
    */
   private Request networkRequest(Request request) throws IOException {
+    if (!request.isHttps()) {
+      Authenticator authenticator = client.proxyAuthenticator();
```
yschimke: Is this mean to be using proxyAuthenticator here? Where do you do preemptive authentication with client.authenticator()?

yschimke:
Hope you don't mind drive-by comments.  A big +1 to this feature, I've implemented using an Interceptor, and this looks like it will be cleaner.

okhttp/src/main/java/okhttp3/Authenticator.java
smatei: @yschimke All I need is basic proxy authentication. Maybe it would make more sense to add another interface ProxyAuthenticator that extends Authenticator. Anyway, my comment should be added to the implementation. I am thinking of adding BasicAuthenticator class.

swankjesse:
I like the idea but I don’t think we can safely change the Authenticator interface. How about a boolean option on OkHttpClient that configures whether the ProxyAuthenticator is pre-emptive?

yschimke:
Can't an authenticator be both? You may have different authentication rules in play for the different hosts.

You could configure separate HttpClient instances but that seems to break down in the case of redirects anyway.

smatei:
@swankjesse @yschimke 
I like the idea of a preemptive setting for OKHttpClient and leave the Authenticator interface untouched. I will do this tomorrow.

okhttp/src/main/java/okhttp3/internal/http/HttpEngine.java
smatei: @yschimke I only need it for proxy. Did not bother think about regular authenticator. I like @swankjesse suggestion about the setting in the client. I will implement it.

swankjesse: I think we can offer the option for proxies initially, and maybe other clients later.

swankjesse: Uh oh, the current API won’t really work for pre-emptive auth unless we generate a fake response. Hmm.

smatei: @swankjesse that's why I needed a second authenticate method, one with request as parameter. A fake response would not be a nice implementation. Maybe a mixed solution with preemptive setting in client, and adding the authenticate method with request parameter to the interface. It does not necessarily be Authenticator, maybe we can extend Authenticator by adding ProxyAuthenticator having this method.

a-k-g:
I would like this feature as well. Since we can't change the Authenticator interface, a clean option could be -
1. Add a `PreemptiveAuthenticator` interface similar to `Authenticator`, which has the following method
    ```
Request authenticatePreemptively(Route route, Request request)
    ```
2. Add `preemptiveProxyAuth` boolean flag, and `preemptiveProxyAuthenticator` property (similar to the existing proxyAuthenticator)

Following the pattern in 2, if needed someone could also implement a `preemptiveAuth` boolean flag and `preemptiveAuthenticator` to do pre-emptive authorization for the non-proxy case.

If people are happy with this idea, I could give it a go.

smatei:
Hi @a-k-g,

Did not have the time to finish the preemptive auth (only created a workaround). Please give it a go!

Best Regards,
Stefan Matei

a-k-g:
I just stumbled over https://github.com/square/okhttp/issues/2426 and it seems like it might unblock some people and make pre-emptive auth less important.

solvingJ:
Is there any workaround for this until this pullrequest is merged?  We just ran into this and it's a significant issue for us.

a-k-g:
@solvingJ - which version of OkHttp are you on? https://github.com/square/okhttp/issues/2426 should fix the problem for most uses.

solvingJ:
Currently using 3.3.1 via maven, which is coming in from another dependency.  I think this is actually a result of a nuance in our environment, so please disregard.  @a-k-g  If you would be willing to take a question about our issue offline, please send me an email.

Issue #2461, opened 2016-04-04
OutOfMemory error on uploading file


Anonomousxxx:
Hello
I'm implementing a service for uploading files, which requries uploading huge files to server
I do it in chunks which requires a lot of requests.
After some time I always get outofmemory exeception:
```
 Throwing OutOfMemoryError "Failed to allocate a 2060 byte allocation with 1328 free bytes and 1328B until OOM"
04-04 13:02:27.608 7536-10064/by.set.pibox E/art: Throwing OutOfMemoryError "Failed to allocate a 3904 byte allocation with 0 free bytes and 0B until OOM" (recursive case)
04-04 13:02:27.751 7536-10064/by.set.pibox E/art: "pool-15-thread-1" prio=5 tid=96 Runnable
04-04 13:02:27.752 7536-10064/by.set.pibox E/art:   | group="main" sCount=0 dsCount=0 obj=0x32c0aee0 self=0xe13af000
04-04 13:02:27.752 7536-10064/by.set.pibox E/art:   | sysTid=10064 nice=0 cgrp=default sched=0/0 handle=0xdd40e880
04-04 13:02:27.752 7536-10064/by.set.pibox E/art:   | state=R schedstat=( 32306483502 2875351514 12061 ) utm=2248 stm=982 core=3 HZ=100
04-04 13:02:27.752 7536-10064/by.set.pibox E/art:   | stack=0xdb821000-0xdb823000 stackSize=1036KB
04-04 13:02:27.752 7536-10064/by.set.pibox E/art:   | held mutexes= "mutator lock"(shared held)
04-04 13:02:27.752 7536-10064/by.set.pibox E/art:   native: #00 pc 0000485c  /system/lib/libbacktrace_libc++.so (UnwindCurrent::Unwind(unsigned int, ucontext*)+23)
04-04 13:02:27.752 7536-10064/by.set.pibox E/art:   native: #01 pc 00003005  /system/lib/libbacktrace_libc++.so (Backtrace::Unwind(unsigned int, ucontext*)+8)
04-04 13:02:27.752 7536-10064/by.set.pibox E/art:   native: #02 pc 00243911  /system/lib/libart.so (art::DumpNativeStack(std::__1::basic_ostream<char, std::__1::char_traits<char> >&, int, char const*, art::mirror::ArtMethod*)+68)
[...]

PR #2462, opened 2016-04-04
Clarify behavior of ResponseBody.string() in its javadoc

artem-zinnatullin:
See https://github.com/square/okhttp/issues/2461#issuecomment-205260542 and bunch of other issues.

### DIFF
--- a/okhttp/src/main/java/okhttp3/ResponseBody.java
+++ b/okhttp/src/main/java/okhttp3/ResponseBody.java
@@ -105,8 +105,8 @@ public abstract class ResponseBody implements Closeable {
    * header is either absent or lacks a charset, this will attempt to decode the response body as
    * UTF-8. Closes {@link ResponseBody} automatically.
    *
-   * <p>Warning: this method loads whole response body into memory and may cause
-   * {@link OutOfMemoryError}.
+   * <p>Please keep in mind that this method loads whole response body into memory, in case of
+   * big response body it may lead to {@link OutOfMemoryError}.
    */
   public final String string() throws IOException {
     return new String(bytes(), charset().name());
--- a/okhttp/src/main/java/okhttp3/ResponseBody.java
+++ b/okhttp/src/main/java/okhttp3/ResponseBody.java
@@ -103,7 +103,10 @@ public abstract class ResponseBody implements Closeable {
   /**
    * Returns the response as a string decoded with the charset of the Content-Type header. If that
    * header is either absent or lacks a charset, this will attempt to decode the response body as
-   * UTF-8.
+   * UTF-8. Closes {@link ResponseBody} automatically.
+   *
+   * <p>Warning: this method loads whole response body into memory and may cause
+   * {@link OutOfMemoryError}.
    */
   public final String string() throws IOException {
     return new String(bytes(), charset().name());


okhttp/src/main/java/okhttp3/ResponseBody.java
```diff
@@ -103,7 +103,10 @@ public final Reader charStream() {
   /**
    * Returns the response as a string decoded with the charset of the Content-Type header. If that
    * header is either absent or lacks a charset, this will attempt to decode the response body as
-   * UTF-8.
+   * UTF-8. Closes {@linkplain ResponseBody} automatically.
```
JakeWharton: should not be `{@linkplain}`

okhttp/src/main/java/okhttp3/ResponseBody.java
```diff
@@ -103,7 +103,10 @@ public final Reader charStream() {
...
+   *
+   * <p>Warning: this method loads whole response body into memory and may cause
+   * {@linkplain OutOfMemoryError}.
```
JakeWharton: should not be `{@linkplain}`

okhttp/src/main/java/okhttp3/ResponseBody.java
artem-zinnatullin: Can you please explain when you prefer `@link` and when `@linkplain`?

According to http://www.tutorialspoint.com/java/java_documentation.htm it looks like style preference rather than functional, right?

artem-zinnatullin:
Changed `{@linkplain}` to `{@link}`

okhttp/src/main/java/okhttp3/ResponseBody.java
JakeWharton: link: code
linkplain: english

okhttp/src/main/java/okhttp3/ResponseBody.java
```diff
@@ -103,7 +103,10 @@ public final Reader charStream() {
   /**
    * Returns the response as a string decoded with the charset of the Content-Type header. If that
    * header is either absent or lacks a charset, this will attempt to decode the response body as
-   * UTF-8.
+   * UTF-8. Closes {@link ResponseBody} automatically.
+   *
+   * <p>Warning: this method loads whole response body into memory and may cause
```
swankjesse: I think this warning needs to be qualified.

artem-zinnatullin: Very sorry, but not sure I understand what exactly do you want to qualify… 

// Asked my US friend-coworker to help because I thought that it was lack of my understanding (English), but he is also not sure what needs to be qualified.

swankjesse: The warning is extremely scary, and doesn’t mention that it will only trigger an OutOfMemoryError if the response is very large. I’m worried about people not using this method when they should because of this scary warning.

artem-zinnatullin: Oh, got it, yeah my bad.

Rewritten it without word `Warning` and more as a thing to keep in mind, hope it's okay now and will help users.

Also, made a separate commit so you'll be able to use new cool GitHub features like ["Pick up where you left off"](https://github.com/blog/2123-more-code-review-tools) and ["Squash your commits"](https://github.com/blog/2141-squash-your-commits) :walking:

swankjesse:
Merged manually!

PR #2467, opened 2016-04-07
Accommodate tunneling proxies that close the connection after an auth challenge.

dave-r12:
A fix for #2426. This was mainly a rejiggering of the existing connection code.

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/io/RealConnection.java
+++ b/okhttp/src/main/java/okhttp3/internal/io/RealConnection.java
@@ -19,6 +19,7 @@ package okhttp3.internal.io;
 import java.io.IOException;
 import java.lang.ref.Reference;
 import java.net.ConnectException;
+import java.net.ProtocolException;
 import java.net.Proxy;
 import java.net.Socket;
 import java.net.SocketTimeoutException;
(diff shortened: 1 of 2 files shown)


okhttp/src/main/java/okhttp3/internal/io/RealConnection.java
```diff
@@ -132,9 +133,52 @@ public void connect(int connectTimeout, int readTimeout, int writeTimeout,
...
+      }
+
+      connectSocket(connectTimeout, readTimeout, writeTimeout, connectionSpecSelector);
+      tunnelRequest = createTunnel(readTimeout, writeTimeout, tunnelRequest, url);
+
+      if (tunnelRequest == null) break; // Tunnel successfully created.
+
+      // The proxy decided to close the connection after an auth challenge. We need to create a new
+      // connection, but this time with the auth credentials.
+      closeQuietly(rawSocket);
```
swankjesse: nit: also set rawSocket to null?

swankjesse:
LGTM

This is brilliant. Very nice.

okhttp/src/main/java/okhttp3/internal/io/RealConnection.java
dave-r12: Yes, done.

axel3rd:
@dave-r12 / @swankjesse : Many thanks to you guys  :satisfied:

PR #2468, opened 2016-04-09
Bundle for okcurl and local launcher

yschimke:
Support for running okcurl locally with a single command from a git checkout.

n.b. This script is currently designed around OSX, e.g. finds JDK8 and if found includes the alpn bootclasspath.

Also builds a bundle including the alpn-boot, which is suitable for Homebrew e.g. https://github.com/yschimke/homebrew-tap/blob/okcurl/okcurl.rb

    brew install yschimke/okcurl

### DIFF
--- a/okcurl/README.md
+++ b/okcurl/README.md
@@ -5,3 +5,28 @@ _A curl for the next-generation web._
 
 OkCurl is an OkHttp-backed curl clone which allows you to test OkHttp's HTTP engine (including
 SPDY and HTTP/2) against web servers.
+
+    $ ./okcurl/okcurl --frames https://graph.facebook.com/robots.txt
+    << 0x00000000    30 SETTINGS      
+    >> 0x00000000     0 SETTINGS      ACK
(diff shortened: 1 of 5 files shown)


swankjesse:
Not sure how I feel about the script building the application before it runs. It would be simpler, and potentially more efficient, to keep run and build independent.

JakeWharton:
The okcurl jar is already a proper executable. Isn't that enough?

yschimke:
It only works with http/2 if you have bootclasspath set and run JDK 8.  

I'm happy for you guys to drop this PR, I was trying to help get towards a situation that okcurl is a viable standalone testing tool. There are still issues with the PR, like its OSX focused.

But it is really useful for testing http/2 against real world services.  That's how I found the issue with the http/2 race condition, wondering why I was not seeing the expected series of frames against facebook and twitter.

swankjesse:
Yeah, that was a nice find. Next steps is to hook up ALPN on the Java 9 EA so that setup is easier!

PR #2470, opened 2016-04-10
Example of using an European National ID card for client auth

yschimke:
This example works with the Estonian National ID Card. 

Output from s_server test server

```
Client certificate
Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number:
            ...
        Signature Algorithm: sha256WithRSAEncryption
        Issuer: C=EE, O=AS Sertifitseerimiskeskus, CN=ESTEID-SK 2011/emailAddress=pki@sk.ee
        Validity
            Not Before: ...
            Not After : ...
        Subject: C=EE, O=ESTEID, OU=authentication, CN=...
```

### DIFF
--- a/samples/guide/src/main/java/okhttp3/recipes/CustomTrust.java
+++ b/samples/guide/src/main/java/okhttp3/recipes/CustomTrust.java
@@ -171,7 +171,7 @@ public final class CustomTrust {
    * not use custom trusted certificates in production without the blessing of your server's TLS
    * administrator.
    */
-  private X509TrustManager trustManagerForCertificates(InputStream in)
+  static X509TrustManager trustManagerForCertificates(InputStream in)
       throws GeneralSecurityException {
     CertificateFactory certificateFactory = CertificateFactory.getInstance("X.509");
     Collection<? extends Certificate> certificates = certificateFactory.generateCertificates(in);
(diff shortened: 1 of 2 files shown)


samples/guide/src/main/java/okhttp3/recipes/CustomTrust.java
```diff
@@ -171,7 +171,7 @@ private InputStream trustedCertificatesInputStream() {
    * not use custom trusted certificates in production without the blessing of your server's TLS
    * administrator.
    */
-  private X509TrustManager trustManagerForCertificates(InputStream in)
+  static X509TrustManager trustManagerForCertificates(InputStream in)
```
swankjesse: (these samples are intended to be completely standalone)

samples/guide/src/main/java/okhttp3/recipes/OpenScClientAuthentication.java
```diff
@@ -0,0 +1,83 @@
...
+import javax.net.ssl.KeyManager;
+import javax.net.ssl.KeyManagerFactory;
+import javax.net.ssl.SSLContext;
+import javax.net.ssl.SSLSocketFactory;
+import javax.net.ssl.TrustManager;
+import javax.net.ssl.X509TrustManager;
+import okhttp3.OkHttpClient;
+import okhttp3.Request;
+import okhttp3.Response;
+import sun.security.pkcs11.SunPKCS11;
```
swankjesse: this is not cross platform right?

swankjesse:
I don’t think this is useful as an example, but it’s certainly useful. Maybe create a standalone GitHub project?

yschimke:
Yeah, I've got it in my own project already.  I thought it was likely that it falls in between OpenSC and various clients, with no good home as a reference.

Issue #2469, opened 2016-04-09
Possible race condition with http2


yschimke:
I haven't looked into this yet, but am seeing different behaviour for the same operation

```
$ okcurl --frames https://graph.facebook.com/robots.txt
<< 0x00000000    30 SETTINGS      
<< 0x00000000     4 WINDOW_UPDATE 
>> CONNECTION 505249202a20485454502f322e300d0a0d0a534d0d0a0d0a
>> 0x00000000     0 SETTINGS      ACK
>> 0x00000000     6 SETTINGS      
>> 0x00000000     4 WINDOW_UPDATE 
>> 0x00000003    76 HEADERS       END_STREAM|END_HEADERS
<< 0x00000000     0 SETTINGS      ACK
<< 0x00000003     4 WINDOW_UPDATE 
<< 0x00000003   174 HEADERS       END_HEADERS
<< 0x00000003    26 DATA          END_STREAM
User-agent: *
Disallow: /
>> 0x00000000     8 GOAWAY       
``` 

```
$ okcurl --frames https://graph.facebook.com/robots.txt
<< 0x00000000    30 SETTINGS      
>> 0x00000000     0 SETTINGS      ACK
<< 0x00000000     4 WINDOW_UPDATE 
>> CONNECTION 505249202a20485454502f322e300d0a0d0a534d0d0a0d0a
>> 0x00000000     6 SETTINGS      
>> 0x00000000     4 WINDOW_UPDATE 
>> 0x00000003    76 HEADERS       END_STREAM|END_HEADERS
<< 0x00000000    34 GOAWAY        
>> CONNECTION 505249202a20485454502f322e300d0a0d0a534d0d0a0d0a
<< 0x00000000    30 SETTINGS      
<< 0x00000000     4 WINDOW_UPDATE 
>> 0x00000000     0 SETTINGS      ACK
>> 0x00000000     6 SETTINGS      
>> 0x00000000     4 WINDOW_UPDATE 
>> 0x00000003    76 HEADERS       END_STREAM|END_HEADERS
<< 0x00000000     0 SETTINGS      ACK
<< 0x00000003     4 WINDOW_UPDATE 
<< 0x00000003   172 HEADERS       END_HEADERS
<< 0x00000003    26 DATA          END_STREAM
User-agent: *
Disallow: /
>> 0x00000000     8 GOAWAY   
```

yschimke:
[...]

PR #2473, opened 2016-04-10
Don't start the reader thread until after the connection preface.

swankjesse:
This is slightly more work than ideal because our tests don't bother with the
connection preface. That makes the tests both simpler, and further from reality.
The workaround is a package-private method to keep the tests working as they
are currently.

Closes https://github.com/square/okhttp/issues/2469

### DIFF
--- a/mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.java
+++ b/mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.java
@@ -449,6 +449,7 @@ public final class MockWebServer implements TestRule {
               .protocol(protocol)
               .listener(framedSocketListener)
               .build();
+          framedConnection.start();
           openFramedConnections.add(framedConnection);
           openClientSockets.remove(socket);
           return;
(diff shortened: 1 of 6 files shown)


mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.java
```diff
@@ -449,6 +449,7 @@ public void processConnection() throws Exception {
               .protocol(protocol)
               .listener(framedSocketListener)
               .build();
+          framedConnection.start();
```
swankjesse: Another bug!

swankjesse: We weren’t sending a connection preface. Though this didn’t matter for SPDY/3, it’s non-compliant for HTTP/2. We didn’t catch it because OkHttp doesn’t enforce this.

adriancole:
Good tradeoff on the package private method for tests.

LGTM

mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.java
dave-r12: Oh, right. I assume it's no big deal, but should we create an issue to enforce it?

dave-r12:
Agree w/ @adriancole, LGTM.

mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.java
swankjesse: Good call.
https://github.com/square/okhttp/issues/2474

PR #2478, opened 2016-04-12
support httpdns,verify hostname in header(if set) not get in url

lizhangqu:
If host is set in header,then replace host by the value in header(named headerHost).
When using https and httpdns,if get the host from url it will get an Exception (javax.net.ssl.SSLPeerUnverifiedException).
What is the httpdns ,see https://www.dnspod.cn/httpdns.
See more in recipes named HttpDns in okhttp-sample.

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/http/HttpEngine.java
+++ b/okhttp/src/main/java/okhttp3/internal/http/HttpEngine.java
@@ -16,11 +16,32 @@
  */
 package okhttp3.internal.http;
 
-import okhttp3.*;
+import okhttp3.Address;
+import okhttp3.CertificatePinner;
+import okhttp3.Connection;
+import okhttp3.Cookie;
+import okhttp3.CookieJar;
+import okhttp3.Headers;
+import okhttp3.HttpUrl;
+import okhttp3.Interceptor;
+import okhttp3.MediaType;
+import okhttp3.OkHttpClient;
+import okhttp3.Protocol;
+import okhttp3.Request;
+import okhttp3.Response;
+import okhttp3.ResponseBody;
+import okhttp3.Route;
 import okhttp3.internal.Internal;
 import okhttp3.internal.InternalCache;
 import okhttp3.internal.Version;
-import okio.*;
+import okio.Buffer;
+import okio.BufferedSink;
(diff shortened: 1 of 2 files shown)


swankjesse:
I don’t think this is worth supporting in the core library. But you can do what you like in interceptors. You might be able to use an application interceptor to rewrite the address OkHttp connects to, and another network interceptor to rewrite what host is sent once connected.

swankjesse:
Scratch that, the network interceptor won’t let you rewrite the URL host. But it will let you set a custom Host header.

lizhangqu:
Sorry,My English is poor.

When it connects by **domain** under the **https** protocol,there is **no problom**.
But if i connects by **ip**（this ip is get from **httpdns** before request,don't use **lcoaldns**）under the https,the ssl verify will get the host **from url**,now the host is ip,will throw an Exception named **javax.net.ssl.SSLPeerUnverifiedException**.

Sample:

If I want to connet the url [https://www.baidu.com](https://www.baidu.com),now using okhttp to connect ,there is no problem.

If I use httpdns to get the domain's ip before request,like this（also is a request）
```
http://119.29.29.29/d?dn=www.baidu.com
```

assert  the ip is 61.135.169.121.

I should replace the domain host by the ip(61.135.169.121) in application interceptors ,and will set a Header named Host,the header value is the domain(www.baidu.com)，now the request is  https://61.135.169.121 with a Host header(www.baidu.com),the ssl verify will get the host from url ,it means the host is ip(61.135.169.121) now,but in https certs,there is no ip,so throws javax.net.ssl.SSLPeerUnverifiedException.But if get the host from header(www.baidu.com) , now will get verified.Because there is a domain(www.baidu.com) in https certs.


Also if i use ip to connet,like this.
```
OkHttpClient okHttpClient = new OkHttpClient.Builder().build();
        .url("https://220.181.112.244")
        .get()
        .build();
Response execute = okHttpClient.newCall(request).execute();

System.out.println(execute.body().string());
```

throws an Exception too.

swankjesse:
Can you implement your own `Dns` class?

swankjesse:
http://square.github.io/okhttp/3.x/okhttp/okhttp3/Dns.html

Issue #2394, opened 2016-03-09
Default behaviour of retryOnConnectionFailure is dangerous for POST requests


ktchernov:
Default behaviour of OkHttpClient.Builder to set `retryOnConnectionFailure` may result in POST requests being silently resent on network timeouts.

While this may be typically alright for GET requests, POST/PATCH/DELETE requests modify data, and re-submitting them without knowing whether they have been received at the other end.

ktchernov:
This is also something that is a significant change in OkHttp v3 over v2. It would be good to include this in `CHANGES.md`

swankjesse:
This has been our behavior forever.

ktchernov:
I have investigated this further, and there is a change in behaviour between v2 and v3 here, which is in the way SocketConnectionTimeout is handled. In v2 the request was always dropped, in v3 the connection is retried unless `retryOnConnectionFailure` is set.

Compare https://github.com/square/okhttp/blob/parent-3.0.0-RC1/okhttp/src/main/java/okhttp3/internal/http/StreamAllocation.java line 336 to https://github.com/square/okhttp/blob/parent-2.7.5/okhttp/src/main/java/com/squareup/okhttp/internal/http/StreamAllocation.java line 358

This means that if the client sent a POST request, then the server received it but the response got lost due to network issues, the client will resend it again. While in v2 it would fail.

ktchernov:
This is the change - https://github.com/square/okhttp/commit/3b44fa37675c1d68298283ae4b8b45cd5771888d and https://github.com/square/okhttp/issues/1736

dave-r12:
@ktchernov could you write a test case? There are great examples in CallTest, https://github.com/square/okhttp/blob/master/okhttp-tests/src/test/java/okhttp3/CallTest.java

renaudcerrato:
Along with https://github.com/square/okhttp/issues/2443, that's exactly what caused my development backend to explode: the multipart upload tooks slightly more time than my read timeout, causing Retrofit to silently retry over and over...

JakeWharton:
Retrofit doesn't retry.

renaudcerrato:
> Retrofit doesn't retry.

I meaned the underlying OkHttp, used by Retrofit 2.

ktchernov:
[...]

PR #2479, opened 2016-04-13
Further constrain when we recover from failed routes.

swankjesse:
Don't recover if we encounter a read timeout after sending the request, but
do recover if we encounter a timeout building a connection.

See https://github.com/square/okhttp/issues/2394


JakeWharton:
LGTM!

PR #2492, opened 2016-04-18
add fastjson's sample

wenshao:


swankjesse:
Thanks, but I don't think this sample adds much beyond our existing samples. Our goal is not to show every possible JSON library.

wenshao:
in china, fastjson more popular than gson, this example may help users okhttp combination fastjson use.
like this issue : https://github.com/alibaba/fastjson/issues/490

swankjesse:
We'll probably change the example to use Moshi instead of Gson, since it has massive support for Okio. If fastjson ever supports Okio natively, we'll at least add it to our Works With OkHttp wiki page.

wenshao:
I see you participate gson project, I invite you to participate in our fastjson project, which is Alibaba's open source project, which has [Java SE version](https://github.com/alibaba/fastjson) and [Android version](https://github.com/alibaba/fastjson/tree/android), fastjson have the [best performance](https://github.com/alibaba/fastjson/wiki/Benchmark_1_2_11), API is also easier to use and more the important point is that it is China's most popular. I hope more people outside China can use it.

swankjesse:
Consider contributing to this project:
https://github.com/cowtowncoder/java-json-performance-benchmarks

wenshao:
I've submitted to this project (https://github.com/eishay/jvm-serializers), it is more serialization library authors used as a reference. https://github.com/eishay/jvm-serializers/pull/66 , it is the latest version of the pull request, there 1.2.11 version of the performance test data.

wenshao:
https://github.com/cowtowncoder/java-json-performance-benchmarks/pull/2

PR #2493, opened 2016-04-19
Fix CertificatePinner javadoc to use a Builder

positron:
The provided example wouldn't compile because OkHttpClient doesn't have
a `setCertificatePinner()` method.


swankjesse:
Thanks!

PR #2494, opened 2016-04-20
Java 9 support for ALPN without alpn-boot

yschimke:
Use SSLParameters.setApplicationProtocols(String[]) and SSLSocket.getApplicationProtocol()

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/Platform.java
+++ b/okhttp/src/main/java/okhttp3/internal/Platform.java
@@ -28,6 +28,7 @@ import java.net.Socket;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.logging.Level;
+import javax.net.ssl.SSLParameters;
 import javax.net.ssl.SSLSocket;
 import javax.net.ssl.SSLSocketFactory;
 import javax.net.ssl.X509TrustManager;
@@ -57,6 +58,8 @@ import static okhttp3.internal.Internal.logger;
  * unstable.
  *
  * Supported on OpenJDK 7 and 8 (via the JettyALPN-boot library).
+ *
+ * Supported on OpenJDK 9 via SSLParameters and SSLSocket features.
  *
  * <h3>Trust Manager Extraction</h3>
  *
@@ -123,6 +126,16 @@ public class Platform {
     System.out.println(message);
   }
 
+  public static List<String> alpnProtocolNames(List<Protocol> protocols) {
+    List<String> names = new ArrayList<>(protocols.size());
+    for (int i = 0, size = protocols.size(); i < size; i++) {
+      Protocol protocol = protocols.get(i);
+      if (protocol == Protocol.HTTP_1_0) continue; // No HTTP/1.0 for ALPN.
+      names.add(protocol.toString());
+    }
+    return names;
+  }
+
   /** Attempt to match the host runtime to a capable Platform implementation. */
   private static Platform findPlatform() {
     // Attempt to find Android 2.3+ APIs.
@@ -155,6 +168,18 @@ public class Platform {
           getAlpnSelectedProtocol, setAlpnProtocols);
     } catch (ClassNotFoundException ignored) {
       // This isn't an Android runtime.
+    }
+
+
+    // Find JDK 9 new methods
+    try {
+      Method setProtocolMethod =
+          SSLParameters.class.getMethod("setApplicationProtocols", String[].class);
+      Method getProtocolMethod = SSLSocket.class.getMethod("getApplicationProtocol");
+
+      return new Jdk9Platform(setProtocolMethod, getProtocolMethod);
+    } catch (NoSuchMethodException ignored) {
+      // pre JDK 9
     }
 
     // Find Jetty's ALPN extension for OpenJDK.
@@ -274,6 +299,54 @@ public class Platform {
   }
 
   /**
+   * OpenJDK 9+.
+   */
+  private static final class Jdk9Platform extends Platform {
+    private final Method setProtocolMethod;
+    private final Method getProtocolMethod;
+
+    public Jdk9Platform(Method setProtocolMethod, Method getProtocolMethod) {
+      this.setProtocolMethod = setProtocolMethod;
+      this.getProtocolMethod = getProtocolMethod;
+    }
+
+    @Override
+    public void configureTlsExtensions(SSLSocket sslSocket, String hostname,
+                                       List<Protocol> protocols) {
+      try {
+        SSLParameters sslParameters = sslSocket.getSSLParameters();
+
+        List<String> names = alpnProtocolNames(protocols);
+
+        setProtocolMethod.invoke(sslParameters,
+            new Object[]{names.toArray(new String[names.size()])});
+
+        sslSocket.setSSLParameters(sslParameters);
+      } catch (IllegalAccessException | InvocationTargetException e) {
+        throw new AssertionError();
+      }
+    }
+
+    @Override
+    public String getSelectedProtocol(SSLSocket socket) {
+      try {
+        String protocol = (String) getProtocolMethod.invoke(socket);
+
+        // SSLSocket.getApplicationProtocol returns "" if application protocols values will not
+        // be used. Observed if you didn't specify SSLParameters.setApplicationProtocols
+        if (protocol == null || protocol.equals("")) {
+          return null;
+        }
+
+        return protocol;
+      } catch (IllegalAccessException | InvocationTargetException e) {
+        throw new AssertionError();
+      }
+    }
+  }
+
+
+  /**
    * OpenJDK 7+ with {@code org.mortbay.jetty.alpn/alpn-boot} in the boot class path.
    */
   private static class JdkWithJettyBootPlatform extends Platform {
@@ -294,12 +367,8 @@ public class Platform {
 
     @Override public void configureTlsExtensions(
         SSLSocket sslSocket, String hostname, List<Protocol> protocols) {
-      List<String> names = new ArrayList<>(protocols.size());
-      for (int i = 0, size = protocols.size(); i < size; i++) {
-        Protocol protocol = protocols.get(i);
-        if (protocol == Protocol.HTTP_1_0) continue; // No HTTP/1.0 for ALPN.
-        names.add(protocol.toString());
-      }
+      List<String> names = alpnProtocolNames(protocols);
+
       try {
         Object provider = Proxy.newProxyInstance(Platform.class.getClassLoader(),
             new Class[] {clientProviderClass, serverProviderClass}, new JettyNegoProvider(names));


okhttp/src/main/java/okhttp3/internal/Platform.java
```diff
@@ -176,6 +193,13 @@ private static Platform findPlatform() {
     return new Platform();
   }
 
+  /**
+   * Is Java 9.  Does not include Java 10 or later.
```
swankjesse: Typically this isn’t how we detect the host platform. Instead it’s generally more reliable to look for signatures like the presence of methods like `setApplicationProtocols()`. Otherwise we regress on Java 10 for example.

okhttp/src/main/java/okhttp3/internal/Platform.java
```diff
@@ -274,6 +298,55 @@ public Android(Class<?> sslParametersClass, OptionalMethod<Socket> setUseSession
   }
 
   /**
+   * OpenJDK 9+.
+   */
+  private static class Jdk9Platform extends Platform {
```
swankjesse: nit: final

okhttp/src/main/java/okhttp3/internal/Platform.java
```diff
@@ -58,6 +59,8 @@
  *
  * Supported on OpenJDK 7 and 8 (via the JettyALPN-boot library).
  *
+ * Supported on OpenJDK9 via SSLParameters and SSLSocket features.
```
swankjesse: I believe they spell it with a space between `OpenJDK` and `9`.

okhttp/src/main/java/okhttp3/internal/Platform.java
```diff
@@ -274,6 +298,55 @@ public Android(Class<?> sslParametersClass, OptionalMethod<Socket> setUseSession
...
+        this.getProtocolMethod = SSLSocket.class.getMethod("getApplicationProtocol");
+      } catch (NoSuchMethodException e) {
+        throw new AssertionError();
+      }
+    }
+
+    @Override public void configureTlsExtensions(SSLSocket sslSocket, String hostname,
+        List<Protocol> protocols) {
+      try {
+        SSLParameters p = sslSocket.getSSLParameters();
```
swankjesse: nit: give this a proper name, like `sslParameters`

okhttp/src/main/java/okhttp3/internal/Platform.java
```diff
@@ -274,6 +298,55 @@ public Android(Class<?> sslParametersClass, OptionalMethod<Socket> setUseSession
...
+      } catch (IllegalAccessException | InvocationTargetException e) {
+        throw new AssertionError();
+      }
+    }
+
+    @Override public String getSelectedProtocol(SSLSocket socket) {
+      try {
+        String protocol = (String) getProtocolMethod.invoke(socket);
+
+        if (protocol != null && protocol.equals("")) {
```
swankjesse: is `""` a documented case? That’s disgusting

swankjesse: Indeed:

>     null if it has not yet been determined if application protocols might be used for this connection, an empty String if application protocols values will not be used, or a non-empty application protocol String if a value was successfully negotiated.

swankjesse: I think this line might want to be `if (protocol == null || protocol.equals(""))` . . . either of these cases is symmetric

okhttp/src/main/java/okhttp3/internal/Platform.java
swankjesse: In this case I think you wanna do the method lookup that you have in `Jdk9Platform` and ignore failures in the catch clause if that misses

swankjesse:
Awesome. A few very small nitpicks & we’re good to go.

swankjesse:
If you haven’t yet completed Square’s CLA, please do that now.

yschimke:
I have previously submitted the CLA.  let me know and I'll do it again if you can't find it or similar.

okhttp/src/main/java/okhttp3/internal/Platform.java
yschimke: 1 tangential comment, OptionalMethod is broken in JDK9 because it uses the concrete type which can be offlimits instead of the public supertype that declares the method.

okhttp/src/main/java/okhttp3/internal/Platform.java
yschimke: Should the other instances be final also? or another PR?

yschimke:
I'll do further testing tonight after work before I'm happy for it to merge

okhttp/src/main/java/okhttp3/internal/Platform.java
```diff
@@ -274,6 +298,58 @@ public Android(Class<?> sslParametersClass, OptionalMethod<Socket> setUseSession
...
+
+          sslSocket.setSSLParameters(sslParameters);
+        } catch (IllegalAccessException | InvocationTargetException e) {
+          throw new AssertionError();
+        }
+      }
+    }
+
+    @Override public String getSelectedProtocol(SSLSocket socket) {
+      if (this.getProtocolMethod != null) {
```
swankjesse: this is always true. There’s no way to get an instance of this class with a null getProtocolMethod

okhttp/src/main/java/okhttp3/internal/Platform.java
```diff
@@ -274,6 +298,58 @@ public Android(Class<?> sslParametersClass, OptionalMethod<Socket> setUseSession
...
+    private Method getProtocolMethod = null;
+
+    public Jdk9Platform(Method setProtocolMethod, Method getProtocolMethod) {
+      this.setProtocolMethod = setProtocolMethod;
+      this.getProtocolMethod = getProtocolMethod;
+    }
+
+    @Override public void configureTlsExtensions(SSLSocket sslSocket, String hostname,
+        List<Protocol> protocols) {
+      if (this.setProtocolMethod != null) {
```
swankjesse: this is always true. There’s no way to get an instance of this class with a null setProtocolMethod

okhttp/src/main/java/okhttp3/internal/Platform.java
```diff
@@ -274,6 +298,58 @@ public Android(Class<?> sslParametersClass, OptionalMethod<Socket> setUseSession
   }
 
   /**
+   * OpenJDK 9+.
+   */
+  private static final class Jdk9Platform extends Platform {
+    private Method setProtocolMethod = null;
```
swankjesse: `final`

swankjesse:
LGTM! Please squash your changes

okhttp/src/main/java/okhttp3/internal/Platform.java
```diff
@@ -157,6 +170,17 @@ private static Platform findPlatform() {
...
 
+
+    // Find JDK 9 new methods
+    try {
+      Method setProtocolMethod =
+          SSLParameters.class.getMethod("setApplicationProtocols", String[].class);
+      Method getProtocolMethod = SSLSocket.class.getMethod("getApplicationProtocol");
+
+      return new Jdk9Platform(setProtocolMethod, getProtocolMethod);
+    } catch (NoSuchMethodException e) {
```
swankjesse: nit: rename `e` to `ignored` please.

okhttp/src/main/java/okhttp3/internal/Platform.java
```diff
@@ -274,6 +298,58 @@ public Android(Class<?> sslParametersClass, OptionalMethod<Socket> setUseSession
...
+        }
+      }
+    }
+
+    @Override public String getSelectedProtocol(SSLSocket socket) {
+      if (this.getProtocolMethod != null) {
+        try {
+          String protocol = (String) getProtocolMethod.invoke(socket);
+
+          // SSLSocket.getApplicationProtocol returns "" if application protocols values will not be used
```
swankjesse: Our style checker doesn't like this one:
```
/home/travis/build/square/okhttp/okhttp/src/main/java/okhttp3/internal/Platform.java:335: Line is longer than 100 characters (found 105).
```

yschimke:
Tested ok

Java 9

```
$ $(/usr/libexec/java_home -v 9)/bin/java -jar okcurl/target/okcurl-3.3.0-SNAPSHOT-jar-with-dependencies.jar --frames https://graph.facebook.com/robots.txt
>> CONNECTION 505249202a20485454502f322e300d0a0d0a534d0d0a0d0a
>> 0x00000000     6 SETTINGS      
>> 0x00000000     4 WINDOW_UPDATE 
<< 0x00000000    30 SETTINGS      
>> 0x00000000     0 SETTINGS      ACK
<< 0x00000000     4 WINDOW_UPDATE 
>> 0x00000003    76 HEADERS       END_STREAM|END_HEADERS
<< 0x00000000     0 SETTINGS      ACK
<< 0x00000003     4 WINDOW_UPDATE 
<< 0x00000003   173 HEADERS       END_HEADERS
<< 0x00000003    26 DATA          END_STREAM
User-agent: *
Disallow: /
>> 0x00000000     8 GOAWAY   
```     

Java 8

```
$ $(/usr/libexec/java_home -v 1.8)/bin/java -jar okcurl/target/okcurl-3.3.0-SNAPSHOT-jar-with-dependencies.jar --frames https://graph.facebook.com/robots.txt
User-agent: *
Disallow: /
```

Java 8 + alpn-boot

```
$ $(/usr/libexec/java_home -v 1.8)/bin/java -Xbootclasspath/p:../../.m2/repository/org/mortbay/jetty/alpn/alpn-boot/8.1.7.v20160121/alpn-boot-8.1.7.v20160121.jar -jar okcurl/target/okcurl-3.3.0-SNAPSHOT-jar-with-dependencies.jar --frames https://graph.facebook.com/robots.txt
>> CONNECTION 505249202a20485454502f322e300d0a0d0a534d0d0a0d0a
>> 0x00000000     6 SETTINGS      
>> 0x00000000     4 WINDOW_UPDATE 
<< 0x00000000    30 SETTINGS      
>> 0x00000000     0 SETTINGS      ACK
<< 0x00000000     4 WINDOW_UPDATE 
>> 0x00000003    76 HEADERS       END_STREAM|END_HEADERS
<< 0x00000000     0 SETTINGS      ACK
<< 0x00000003     4 WINDOW_UPDATE 
<< 0x00000003   174 HEADERS       END_HEADERS
<< 0x00000003    26 DATA          END_STREAM
User-agent: *
Disallow: /
>> 0x00000000     8 GOAWAY    
```

swankjesse:
Wonderful. Thankyou!

okhttp/src/main/java/okhttp3/internal/Platform.java
```diff
@@ -123,6 +126,16 @@ public void log(String message) {
     System.out.println(message);
   }
 
+  public static List<String> alpnProtocolNames(List<Protocol> protocols) {
+    List<String> names = new ArrayList<>(protocols.size());
+    for (int i = 0, size = protocols.size(); i < size; i++) {
```
nremond: May I ask why it's not `for (Protocol protocol: protocols) {...}`

iNoles: it is from https://speakerdeck.com/jakewharton/eliminating-code-overhead-square-hq-2015

JakeWharton: Saves an Iterator allocation and all its indirection from the underlying data.

nremond: I see. Thanks for the info !

PR #2496, opened 2016-04-21
retain logger instance to actually log things

yschimke:
This is nasty.  I'm open to suggestions but frame logging doesn't work without this.

https://docs.oracle.com/javase/8/docs/api/java/util/logging/LogManager.html#getLogger-java.lang.String-


swankjesse:
Yep. What a terrible API that is.

swankjesse:
Thanks!

Issue #2481, opened 2016-04-13
AsyncCall in RealCall sets the thread name to the url, exposing sensitive data


Floens:
The AsyncCall runnable inside the RealCall class extends NamedRunnable, and sets the current thread name to the full url of the call.
This can expose sensitive information, for example when API's require a token in the url. I'm having the problem that I log the thread name with log4j2 and it now logs the full url.
I suggest changing it to `toLoggableString()`.

Thanks

swankjesse:
I thought we had all of these but we must have missed one. Can fix.

dave-r12:
Closed with #2499.

PR #2499, opened 2016-04-21
Redact the full url in AsyncCall thread name.

dave-r12:
We don't want to accidentally leak sensitive information that may be
embedded in the url.

For #2481.


quom:
Is there any way to prevent this behaviour? I can't differentiate between which thread a response comes back on when both are to the same host.

swankjesse:
Use `Thread.setNameI()` in your callback.

PR #2502, opened 2016-04-22
Use Moshi in OkHttp.

swankjesse:
As discussed here: https://github.com/square/okhttp/pull/2492

### DIFF
--- a/samples/simple-client/src/main/java/okhttp3/sample/OkHttpContributors.java
+++ b/samples/simple-client/src/main/java/okhttp3/sample/OkHttpContributors.java
@@ -1,8 +1,8 @@
 package okhttp3.sample;
 
-import com.google.gson.Gson;
-import com.google.gson.reflect.TypeToken;
-import java.io.Reader;
+import com.squareup.moshi.JsonAdapter;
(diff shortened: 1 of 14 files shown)


samples/simple-client/src/main/java/okhttp3/sample/OkHttpContributors.java
```diff
@@ -13,10 +13,9 @@
 
 public class OkHttpContributors {
   private static final String ENDPOINT = "https://api.github.com/repos/square/okhttp/contributors";
-  private static final Gson GSON = new Gson();
-  private static final TypeToken<List<Contributor>> CONTRIBUTORS =
-      new TypeToken<List<Contributor>>() {
-      };
+  private static final Moshi MOSHI = new Moshi.Builder().build();
+  private static final JsonAdapter<List<Contributor>> CONTRIBUTORS_JSON_ADAPTER = MOSHI.adapter(
+      Types.newParameterizedType(List.class, Contributor.class));
```
dave-r12: Nice, I haven't done much with this aspect of Moshi but like the API.

PR #2518, opened 2016-04-27
Unconditional FileNotFoundException for response code >= 400 violates…

ericatkin:
… RFC7231.

6.5.  Client Error 4xx

   The 4xx (Client Error) class of status code indicates that the client
   seems to have erred.  Except when responding to a HEAD request, the
   server SHOULD send a representation containing an explanation of the
   error situation, and whether it is a temporary or permanent
   condition.  These status codes are applicable to any request method.
   **User agents SHOULD display any included representation to the user.**

6.6.  Server Error 5xx

   The 5xx (Server Error) class of status code indicates that the server
   is aware that it has erred or is incapable of performing the
   requested method.  Except when responding to a HEAD request, the
   server SHOULD send a representation containing an explanation of the
   error situation, and whether it is a temporary or permanent
   condition.  **A user agent SHOULD display any included representation
   to the user.**  These response codes are applicable to any request
   method.

### DIFF
--- a/okhttp-urlconnection/src/main/java/okhttp3/internal/huc/HttpURLConnectionImpl.java
+++ b/okhttp-urlconnection/src/main/java/okhttp3/internal/huc/HttpURLConnectionImpl.java
@@ -17,7 +17,6 @@
 
 package okhttp3.internal.huc;
 
-import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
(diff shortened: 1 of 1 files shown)


swankjesse:
This is specified by Oracle in their HttpUrlConnection definition. I agree that this design is awkward.

Your best bet is to use OkHttp’s native API.

ericatkin:
Thanks. Can you point me at this definition?

swankjesse:
https://docs.oracle.com/javase/7/docs/api/java/net/HttpURLConnection.html

ericatkin:
Thank you, but I had examined that document before. I don't see any requirement in there that >= 400 responses discard the response body. In fact, they link the RFC, which as documented above says quite the opposite. I appreciate your suggestion to use OkHttp natively and am likely to go that route anyway, but I hoped this change would improve the usefulness of the HttpUrlConnection API which has historically been so messy, or at least understand why it was implemented in this way. I consider the current behavior broken but I won't press it from here.

ericatkin:
I see it now under getErrorStream(). Thanks again.

PR #2520, opened 2016-04-27
split out platform and test them based on active profile

yschimke:
Want to be able to add some tests for JDK 9, but only run when we have a profile active.

Also fix a bug with JDK 9 profile activation.

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/internal/Jdk9PlatformTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/internal/Jdk9PlatformTest.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright (C) 2016 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okhttp3.internal;
+
+import org.junit.Test;
+
+import static okhttp3.internal.PlatformTest.getPlatform;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assume.assumeTrue;
+
+public class Jdk9PlatformTest {
+  @Test
+  public void buildsWhenJdk9() {
+    assumeTrue(getPlatform().equals("jdk9"));
+
+    assertNotNull(Jdk9Platform.buildIfSupported());
+  }
+
+  @Test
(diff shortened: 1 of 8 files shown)


yschimke:
BTW Just in case the angst in your tweet about high cost of guiding PRs by new contributors relate to this, I'm perfectly happy hearing that the PR is a bad idea full stop or some very high level changes that are needed.

But I was trying to get to a point where okhttp can build and run tests with JDK7, 8 or 9, currently it fails badly with 9.

swankjesse:
Nope, I think this PR is a good idea. But it’ll take me some time to understand and I haven’t yet invested a minute. Thanks for your patience.

okhttp/src/main/java/okhttp3/internal/JdkWithJettyBootPlatform.java
```diff
@@ -0,0 +1,136 @@
...
+import java.lang.reflect.Proxy;
+import java.util.List;
+import java.util.logging.Level;
+import javax.net.ssl.SSLSocket;
+import okhttp3.Protocol;
+
+import static okhttp3.internal.Internal.logger;
+
+/**
+ * OpenJDK 7+ with {@code org.mortbay.jetty.alpn/alpn-boot} in the boot class path.
```
swankjesse: mind changing this from `OpenJDK 7+` to `OpenJDK 7 or OpenJDK 8`

okhttp/src/main/java/okhttp3/internal/Android.java
```diff
@@ -0,0 +1,164 @@
...
+import java.net.InetSocketAddress;
+import java.net.Socket;
+import java.util.List;
+import javax.net.ssl.SSLSocket;
+import javax.net.ssl.SSLSocketFactory;
+import javax.net.ssl.X509TrustManager;
+import okhttp3.Protocol;
+
+/** Android 2.3 or better. */
+class Android extends Platform {
```
swankjesse: rename to AndroidPlatform? Now that it’s standalone that’s not as obvious

okhttp-tests/src/test/java/okhttp3/internal/Jdk9PlatformTest.java
```diff
@@ -0,0 +1,27 @@
+package okhttp3.internal;
```
swankjesse: Please add a Square 2016 copyright to all new files

okhttp-tests/src/test/java/okhttp3/internal/Jdk9PlatformTest.java
```diff
@@ -0,0 +1,27 @@
...
+import org.junit.Test;
+
+import static okhttp3.internal.PlatformTest.getPlatform;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assume.assumeTrue;
+
+public class Jdk9PlatformTest {
+  @Test
+  public void testBuildsWhenJdk9() {
```
swankjesse: nit: we don’t typically prefix tests with `test`, that’s a relic from old JUnit

swankjesse:
Seems reasonable enough. Please address the feedback above, and squash into a single commit.

Sorry for the slow response.

Issue #2505, opened 2016-04-22
Logging - Please let the underlying log implementation open


SeameX:
Feature Request

Please remove these sout/syso and implement a logging framework. 
**Let the underlying log implementation open!**

As Developer i want to use my own logging implementation and i dont want to have sout's

Best regards

Stefan

PR #2540, opened 2016-05-08
Make logging more consistent throughout OkHttp.

swankjesse:
Avoid using System.out.

Use the best logging implementation on the host platform. On Java this is
java.util.logging. On Android it's Android.util.Log.

Closes https://github.com/square/okhttp/issues/2505


JakeWharton:
lgtm

PR #2542, opened 2016-05-08
Adding MockHTTPDispatcher, MockRequest and MockRequestBuilder to make mocking requests easier

ajitsing:
MockHTTPDispatcher has following features:
1) Mocking a request
2) Number of times a request is served
3) Responding to request on demand
4) Interrupt dispatcher

### DIFF
--- a/mockwebserver/src/main/java/okhttp3/mockwebserver/MockHTTPDispatcher.java
+++ b/mockwebserver/src/main/java/okhttp3/mockwebserver/MockHTTPDispatcher.java
@@ -39,10 +39,11 @@ public class MockHTTPDispatcher extends Dispatcher {
     }
 
     /**
-     * Mocks a request
+     * Mocks a request.
      */
     public MockHTTPDispatcher mock(MockRequest mockRequest) throws IOException {
-        MockResponse mockResponse = new MockResponse().addHeader("Content-Type", mockRequest.responseContentType.toString());
+        MockResponse mockResponse = new MockResponse()
+                .addHeader("Content-Type", mockRequest.responseContentType.toString());
         if (mockRequest.response != null)
             mockResponse.setBody(mockRequest.response);
         else if (mockRequest.responseStream != null)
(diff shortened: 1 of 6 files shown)


swankjesse:
I don’t understand. What's the goal? Why is mocking requests necessary?

ajitsing:
Thanks @swankjesse for the quick response.

I will explain the goal of this pull request with an example. In our android project we are using mockWebServer to mock the responses of our request in the integration tests. The idea is, we don't want to hit the actual services to get the response in tests.
Now in order to do that, we created our own MockHTTPDispatcher which takes a MockRequest and Returns a MockResponse when the request is served. There are tests which makes the requests in different ways. e.g

1) There is an Android Activity which makes an API call on start. To test that activity using android instrumentation test, we need to mock the API call using MockHTTPDispatcher and render the page ASAP.

2) Before making the API call, I want to test that there is a loader present on the screen. For which I have  to put the dispatcher on hold(using MockRequest#respondOnDemand) until I see the loader. And after I verify the necessary steps before making the call, I tell the dispatcher to continue with serving the request (using MockRequest#respondNow).

3) There are cases when I want to mock request to return an image back in response. In those cases I can construct the request by passing an InputStream as a MockResponse (using MockRequest#setResponseStream(InputStream))

4) In few scenarios I want to test the refresh functionality of a page. In that case I want to know the number of times the request to service API was made. In this case the request will be executed twice, 
            i) on load of page 
           ii) on click of refresh button. 
MockHTTPDispatcher provides functionality to check, how many times a request was executed.

I hope this helps. In case you need any specific clarification, Please let me know.
Thanks.

swankjesse:
Cool! I think this makes a significant scope increase in what MockWebServer does. We don’t yet provide any dispatching policy, and this would introduce that. Which is fine, but I’m reluctant to include it in the core open source project.

Would you be open to creating this as a standalone open source project? I suspect there’s a lot you can do with defining dispatching rules and behaviors and MockWebServer just wants to be the dumb engine beneath it all.

swankjesse:
See also https://github.com/orhanobut/mockwebserverplus (and others)

swankjesse:
Closing because although this is Good Stuff I’m not prepared to grow the scope of MockWebServer.

PR #2544, opened 2016-05-08
Promote sent and received timestamps to fields on Response.

swankjesse:
I'm currently keeping these as longs. That's easy an unambiguous, but it also
feels rather primitive.

This was more work than I expected, but it doesn't seem particularly risky.
It's also potentially more efficient, since there's fewer places where we're
converting from string to long and vice versa.

Closes: https://github.com/square/okhttp/issues/2035

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/CallTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/CallTest.java
@@ -112,13 +112,25 @@ public final class CallTest {
   }
 
   @Test public void get() throws Exception {
-    server.enqueue(new MockResponse().setBody("abc").addHeader("Content-Type: text/plain"));
+    server.enqueue(new MockResponse()
+        .setBody("abc")
+        .clearHeaders()
+        .addHeader("content-type: text/plain")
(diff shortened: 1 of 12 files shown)


okhttp-tests/src/test/java/okhttp3/CallTest.java
```diff
@@ -1177,31 +1189,48 @@ private void postBodyRetransmittedAfterAuthorizationFail(String body) throws Exc
...
     HttpUrl url = server.url("/");
+    long request1SentAt = System.currentTimeMillis();
     executeSynchronously("/", "Accept-Language", "fr-CA", "Accept-Charset", "UTF-8")
         .assertCode(200)
         .assertBody("A");
+    long request1ReceivedAt = System.currentTimeMillis();
     assertNull(server.takeRequest().getHeader("If-None-Match"));
 
     // Hit that stored response. It's different, but Vary says it doesn't matter.
+    Thread.sleep(10); // Make sure the timestamps are unique.
```
JakeWharton: You're only recording one

PR #2552, opened 2016-05-12
Add a simple version dynamic table of HPACK to the writer client.

mostleg:
This will automatically add all the headers into the dynamic table of HPACK if it is not in the static table.

I didn't write new unit test for checking the new added logic, e.g., the eviction of the dynamic table. This is just for a quick review. If the algorithm I used looks good, I'll add them.

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/framed/Hpack.java
+++ b/okhttp/src/main/java/okhttp3/internal/framed/Hpack.java
@@ -19,6 +19,7 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


okhttp/src/main/java/okhttp3/internal/framed/Hpack.java
```diff
@@ -367,27 +368,125 @@ ByteString readByteString() throws IOException {
   }
 
   static final class Writer {
+    private static final byte SEPARATED_TOKEN = ':';
+    private static final int SETTINGS_HEADER_TABLE_SIZE = 4096;
+
     private final Buffer out;
+    private final Map<ByteString, Integer> headerStringToDynamicIndex =
+        new HashMap<ByteString, Integer>();
```
swankjesse: nit: can you please do a LinkedHashMap? Nondeterministic code is one of my pet peeves.

okhttp/src/main/java/okhttp3/internal/framed/Hpack.java
```diff
@@ -367,27 +368,125 @@ ByteString readByteString() throws IOException {
...
+
+    private int headerTableSizeSetting;
+    private int maxDynamicTableByteCount;
+    // Visible for testing.
+    Header[] dynamicTable = new Header[8];
```
swankjesse: I’m surprised this is is 8 . . . I guess I need to learn more about HPACK!

okhttp/src/main/java/okhttp3/internal/framed/Hpack.java
```diff
@@ -367,27 +368,125 @@ ByteString readByteString() throws IOException {
...
+    // Array is populated back to front, so new entries always have lowest index.
```
swankjesse: interesting!

okhttp/src/main/java/okhttp3/internal/framed/Hpack.java
```diff
@@ -367,27 +368,125 @@ ByteString readByteString() throws IOException {
...
+        // determine how many headers need to be evicted.
+        for (int j = dynamicTable.length - 1; j >= nextHeaderIndex && bytesToRecover > 0; j--) {
+          bytesToRecover -= dynamicTable[j].hpackSize;
+          dynamicTableByteCount -= dynamicTable[j].hpackSize;
+          headerCount--;
+          entriesToEvict++;
+        }
+        System.arraycopy(dynamicTable, nextHeaderIndex + 1, dynamicTable,
+            nextHeaderIndex + 1 + entriesToEvict, headerCount);
+        for (Map.Entry<ByteString, Integer> p : headerStringToDynamicIndex.entrySet()) {
```
swankjesse: This is simple but I wonder if there’s a datastructure better suited to this. I’m thinking about Android and the amount of boxing and unboxing that happens here.

(I guess we shouldn’t worry about micro-optimizations until we know the code is working!)

okhttp/src/main/java/okhttp3/internal/framed/Hpack.java
```diff
@@ -367,27 +368,125 @@ ByteString readByteString() throws IOException {
...
+      if (delta > maxDynamicTableByteCount) {
+        clearDynamicTable();
+        return;
+      }
+
+      // Evict headers to the required length.
+      int bytesToRecover = (dynamicTableByteCount + delta) - maxDynamicTableByteCount;
+      evictToRecoverBytes(bytesToRecover);
+
+      if (headerCount + 1 > dynamicTable.length) { // Need to grow the dynamic table.
```
swankjesse: Ahh, this is why `8` above is okay. Makes sense now. I’m just giving you my raw reactions to this code as I read it!

okhttp/src/main/java/okhttp3/internal/framed/Hpack.java
```diff
@@ -367,27 +368,125 @@ ByteString readByteString() throws IOException {
...
-          writeByteString(name);
-          writeByteString(headerBlock.get(i).value);
+          ByteString headerString = getHeaderString(headerBlock.get(i));
+          Integer dynamicIndex = headerStringToDynamicIndex.get(headerString);
+          if (dynamicIndex != null) {
+            // Indexed Header.
+            writeInt(dynamicTable.length - dynamicIndex + STATIC_HEADER_TABLE.length, PREFIX_7_BITS,
+                0);
+          } else {
+            // Literal Header Field with Incremental Indexing — New Name
```
swankjesse: Neat.

okhttp/src/main/java/okhttp3/internal/framed/Hpack.java
```diff
@@ -367,27 +368,125 @@ ByteString readByteString() throws IOException {
...
+            out.writeByte(0x40);
+            writeByteString(name);
+            writeByteString(value);
+            insertIntoDynamicTable(headerBlock.get(i));
```
swankjesse: nit: we probably want to change all of this code to call HeaderBlock.get(i) only 1x.

swankjesse:
Thanks much, and sorry for the delay on the response.

Yeah, this is good stuff. I think you’ve done a very nice job of starting with a simple, obviously-correct implementation. In another PR we can follow up and try to optimize the allocations etc. to avoid generating garbage or doing unnecessary work. 

Next steps – there are many many possible edge cases and things to test here. Table doubling, eviction, static vs. dynamic, etc. Would you like to proceed to add those? Despite many of the comments I made above, I don’t think we should try to micro-optimize this quite yet. It’s going to be easier to merge this with the implementation you have and then figure that stuff out later. Heck we might opt for LRU if HPACK can do that, and so any optimization done now would be counterproductive for that.

mostleg:
Thank you. Sure I'm pleased to proceed to add the tests.

okhttp/src/main/java/okhttp3/internal/framed/Hpack.java
mostleg: OK, I'll change it.

okhttp/src/main/java/okhttp3/internal/framed/Hpack.java
mostleg: En. Actually the code are all imitate the one in this file for the dynamic table in the reader, haha.

okhttp/src/main/java/okhttp3/internal/framed/Hpack.java
mostleg: I see. Will change it.

okhttp/src/main/java/okhttp3/internal/framed/Hpack.java
mostleg: Yes. Maybe think about it later.

okhttp/src/main/java/okhttp3/internal/framed/Hpack.java
mostleg: Do you mean the minus? changed.

mostleg:
I made some changes based on the comments above, and add some unit tests.

swankjesse:
Sorry for dragging on this. I hope to take a looks soon.

mostleg:
No problem, please review it when you have time.

swankjesse:
This pull request is fantastic. I spent the weekend learning about HPACK (I didn’t write our HPACK code) and this is a simple change that makes sense.

I’m going to take this and add a few small changes. For example, I think this code might be simplified with ArrayDeque instead of the manual arrays. I’ll make those changes on top of your changes, and I’ll create a PR with both.

Thanks!

mostleg:
Thanks Jesse!

swankjesse:
My changes are too slow. Let’s merge this as-is and I can come back and tweak the minor details later.

mostleg:
I see. Thanks! When will we release it?

swankjesse:
No commitments on release dates. Historically we like to do every 90 days or so.

Issue #2409, opened 2016-03-15
IllegalStateException: stream != null


hamberluo:
```
Fatal Exception: java.lang.IllegalStateException: stream != null
       at okhttp3.internal.http.StreamAllocation.findConnection(StreamAllocation.java:155)
       at okhttp3.internal.http.StreamAllocation.findHealthyConnection(StreamAllocation.java:127)
       at okhttp3.internal.http.StreamAllocation.newStream(StreamAllocation.java:97)
       at okhttp3.internal.http.HttpEngine.connect(HttpEngine.java:289)
       at okhttp3.internal.http.HttpEngine.sendRequest(HttpEngine.java:241)
       at okhttp3.RealCall.getResponse(RealCall.java:240)
       at okhttp3.RealCall$ApplicationInterceptorChain.proceed(RealCall.java:198)
       at okhttp3.RealCall.getResponseWithInterceptorChain(RealCall.java:160)
       at okhttp3.RealCall.access$100(RealCall.java:30)
       at okhttp3.RealCall$AsyncCall.execute(RealCall.java:127)
       at okhttp3.internal.NamedRunnable.run(NamedRunnable.java:32)
       at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1112)
       at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:587)
       at java.lang.Thread.run(Thread.java:841)
```

dave-r12:
I was researching this one, but haven't found anything yet. @hamberluo do you know if you're using Http1 or Http2 when this exception happened?

dave-r12:
I've not been able to reproduce this one. I believe the cause would be related to a response not being closed prior to making a follow up request. The only thing I came up with is to create a 'misbehaving' network interceptor that wrapped the response, but didn't close it properly.

swankjesse:
Thanks for investigating this Dave. I also tried to isolate this, but my investigation wasn’t very deep! I ended up with this, which discovered that cache bug I fixed last week.
https://gist.github.com/swankjesse/ae7315037362e284edc7

swankjesse:
(I’d like to eventually tidy that up and check it in. If you want to take it over, please do!)

dave-r12:
Oh interesting, I'll have a look.

hamberluo:
@dave-r12 i'm using Http1

dave-r12:
[...]

PR #2555, opened 2016-05-13
Don't crash when a stream hasn't been closed.

swankjesse:
Closes: https://github.com/square/okhttp/issues/2409


dave-r12:
Good catch. I was thinking this was more a user mistake. One upside with the current behavior is that it's fail fast. The exception message is not helpful without some knowledge of the internals, but it is thrown at the source. This change will potentially make tracking down the response body leak a bit more difficult. A possible middle ground, log a warning about a leaked response body?

swankjesse:
Another approach: https://github.com/square/okhttp/pull/2559

PR #2558, opened 2016-05-14
SslClient/Builder and tests mostly running in JDK 9

yschimke:
This is a big change and there are still a few tests failing on JDK9 that I need to look into.  But it seems to be working on JDK7.

open questions

- How much logic should be in SslClient.Builder.  e.g. should builder allow you to provide a whole trustManager? can you create a new builder from an existing SslClient and how that works given SslContext has the existing trustManager internally.

src/main changes
- Remove most deprecated uses of OkHttpClient.Builder.sslSocketFactory(SSLSocketFactory)
- introduce SslClient/Builder for building SSLContext and TrustManager
- JDK9Platform.getTrustManager fails early and noisily
- OkHttpClient.Builder helper method to take SslClient

src/test changes
- change SslContextBuilder to use SslClient
- on JDK9 disable a few UrlConnectionCacheTest tests and test it fails once 
- change lots of tests that used deprecated methods

pom changes 
- revert the JDK9 profile fix that wasn't

```
$ JAVA_HOME=`/usr/libexec/java_home -v 1.7` mvn help:active-profiles -pl okhttp | grep jdk
 - alpn-when-jdk7 (source: com.squareup.okhttp3:parent:3.3.0-SNAPSHOT)
$ JAVA_HOME=`/usr/libexec/java_home -v 1.8` mvn help:active-profiles -pl okhttp | grep jdk
 - alpn-when-jdk8 (source: com.squareup.okhttp3:parent:3.3.0-SNAPSHOT)
$ JAVA_HOME=`/usr/libexec/java_home -v 9` mvn help:active-profiles -pl okhttp | grep jdk
 - jdk9 (source: com.squareup.okhttp3:parent:3.3.0-SNAPSHOT)
```


yschimke:
Failing tests in JDK9

```
  Http2ConnectionTest.peerHttp2ServerLowersInitialWindowSize:122 expected:<65535> but was:<1684>
  CertificatePinnerChainValidationTest.unrelatedPinnedLeafCertificateInChain:207 java.security.KeyStoreException: Key protection  algorithm not found: java.security.KeyStoreException: Certificate chain is not valid
  CertificatePinnerChainValidationTest.unrelatedPinnedIntermediateCertificateInChain:281 java.security.KeyStoreException: Key protection  algorithm not found: java.security.KeyStoreException: Certificate chain is not valid
```

okhttp/src/main/java/okhttp3/internal/tls/DefaultSecurity.java
```diff
@@ -0,0 +1,59 @@
...
+import javax.net.ssl.SSLContext;
+import javax.net.ssl.SSLSocketFactory;
+import javax.net.ssl.TrustManager;
+import javax.net.ssl.TrustManagerFactory;
+import javax.net.ssl.X509TrustManager;
+
+/**
+ * Methods to get the default SSL implementations.
+ */
+public final class DefaultSecurity {
```
swankjesse: Wanna move this into `Util` instead? We’re trying to limit the total size of our `.jar`/`.dex` file and extra classes have a (negligible) impact on that

okhttp/src/main/java/okhttp3/internal/Jdk9Platform.java
```diff
@@ -68,6 +70,15 @@ public String getSelectedProtocol(SSLSocket socket) {
     }
   }
 
+  @Override public X509TrustManager trustManager(SSLSocketFactory sslSocketFactory) {
+    // not supported due to access checks on Jdk9
+    // java.lang.reflect.InaccessibleObjectException: Unable to make member of class
+    // sun.security.ssl.SSLSocketFactoryImpl accessible:  module java.base does not export
+    // sun.security.ssl to unnamed module @xxx
+    throw new UnsupportedOperationException(
+        "clientBuilder.sslSocketFactory(SSLSocketFactory) unsupported on JDK9");
```
swankjesse: nit: “unsupported on JDK9” –> “not supported on JDK 9+” ?

benchmarks/src/main/java/okhttp3/benchmarks/ApacheHttpClient.java
```diff
@@ -42,9 +41,9 @@
     super.prepare(benchmark);
     ClientConnectionManager connectionManager = new PoolingClientConnectionManager();
     if (benchmark.tls) {
-      SSLContext sslContext = SslContextBuilder.localhost();
+      SslContextBuilder sslContextBuilder = SslContextBuilder.localhost();
```
swankjesse: nit: I’d prefer to just immediately call `build()` and not increase the visibility of the mutable `builder` thing

benchmarks/src/main/java/okhttp3/benchmarks/OkHttp.java
```diff
@@ -40,15 +39,15 @@
         .build();
 
     if (benchmark.tls) {
-      SSLContext sslContext = SslContextBuilder.localhost();
-      SSLSocketFactory socketFactory = sslContext.getSocketFactory();
+      SslContextBuilder sslContextBuilder = SslContextBuilder.localhost();
```
swankjesse: Ahh, but this makes sense. . . . should we have a different name than `SslContextBuilder` ? Throughout OkHttp `Builder` almost always means a mutable thing that builds an immutable thing. This feels awkward here.

(not necessarily new with your change, I’m just reflecting on the API!)

mockwebserver/src/main/java/okhttp3/internal/SslContextBuilder.java
```diff
@@ -38,10 +41,14 @@
  * instances where possible.
  */
 public final class SslContextBuilder {
```
swankjesse: . . . .with this change I think this isn’t building an SSLContext, it’s instead building like, a `SSLContext` and a `TrustManager[]` array. Would it be offensive to come up with a new word to represent that pair of things? Maybe like `SslClient` and `SslClient.Builder` ? The `SslClient` is the immutable thing with two fields (sslContext, trustmanagers) and the `SslClient.Builder` is its builder?

okhttp/src/main/java/okhttp3/internal/tls/DefaultSecurity.java
swankjesse: (though I do really like this API. I can be convinced either way.)

okhttp-android-support/src/test/java/okhttp3/internal/huc/CacheAdapterTest.java
```diff
@@ -116,7 +116,7 @@
     };
     setInternalCache(new CacheAdapter(responseCache));
     client = client.newBuilder()
-        .sslSocketFactory(sslContext.getSocketFactory())
+        .sslSocketFactory(sslContextBuilder.socketFactory(), sslContextBuilder.trustManager())
```
swankjesse: ... if we make `SslClient` or whatever, these can be public final member fields instead of methods, which feels neater

okhttp-tests/src/test/java/okhttp3/DelegatingSSLSocket.java
```diff
@@ -280,4 +284,55 @@ public DelegatingSSLSocket(SSLSocket delegate) {
   @Override public void setPerformancePreferences(int connectionTime, int latency, int bandwidth) {
     delegate.setPerformancePreferences(connectionTime, latency, bandwidth);
   }
+
+  // Java 9 methods
```
swankjesse: Nice.

okhttp-tests/src/test/java/okhttp3/URLConnectionTest.java
```diff
@@ -566,17 +568,18 @@ private void doUpload(TransferKind uploadKind, WriteKind writeKind) throws Excep
     assertNotNull(httpsConnection.getCipherSuite());
   }
 
-  @Test public void connectViaHttpsReusingConnections() throws IOException, InterruptedException {
-    server.useHttps(sslContext.getSocketFactory(), false);
+  @Test public void connectViaHttpsReusingConnections()
+      throws IOException, InterruptedException, GeneralSecurityException {
```
swankjesse: nit: or just `throws Exception`

okhttp-tests/src/test/java/okhttp3/URLConnectionTest.java
```diff
@@ -696,8 +698,9 @@ private void doUpload(TransferKind uploadKind, WriteKind writeKind) throws Excep
    *
    * http://code.google.com/p/android/issues/detail?id=13178
    */
-  @Test public void connectViaHttpsToUntrustedServer() throws IOException, InterruptedException {
-    server.useHttps(sslContext.getSocketFactory(), false);
+  @Test public void connectViaHttpsToUntrustedServer()
+      throws IOException, InterruptedException, GeneralSecurityException {
```
swankjesse: same

okhttp-tests/src/test/java/okhttp3/internal/http/RouteSelectorTest.java
```diff
@@ -78,6 +79,9 @@
   private final RecordingProxySelector proxySelector = new RecordingProxySelector();
   private RouteDatabase routeDatabase = new RouteDatabase();
 
+  public RouteSelectorTest() throws GeneralSecurityException {
```
swankjesse: ?

okhttp-tests/src/test/java/okhttp3/internal/tls/CertificatePinnerChainValidationTest.java
```diff
@@ -37,6 +37,11 @@
 import static org.junit.Assert.fail;
 
 public final class CertificatePinnerChainValidationTest {
+  //static {
```
swankjesse: ?

okhttp-urlconnection/src/test/java/okhttp3/UrlConnectionCacheTest.java
```diff
@@ -1399,22 +1405,46 @@ private RecordedRequest assertClientSuppliedCondition(MockResponse seed, String
     assertEquals("B", readAscii(urlFactory.open(server.url("/").url())));
   }
 
+  public static String getPlatform() {
```
swankjesse: nit: move this to the bottom?

okhttp-urlconnection/src/test/java/okhttp3/UrlConnectionCacheTest.java
```diff
@@ -1399,22 +1405,46 @@ private RecordedRequest assertClientSuppliedCondition(MockResponse seed, String
...
+    return System.getProperty("okhttp.platform", "platform");
+  }
+
+  @Test(expected = UnsupportedOperationException.class)
```
swankjesse: I loathe expected exceptions as implemented in JUnit. They are clumsy for many reasons:
 - you can’t be confident that it threw on the last line of the test
 - you can’t assert the contents of the exception
 - you can’t assert the state of the world after the exception is thrown

Please just use `try { call(); fail(); } catch (UnsupportedOperationException expected) {}`

okhttp/src/main/java/okhttp3/internal/Jdk9Platform.java
```diff
@@ -36,14 +38,14 @@ public Jdk9Platform(Method setProtocolMethod, Method getProtocolMethod) {
 
   @Override
   public void configureTlsExtensions(SSLSocket sslSocket, String hostname,
-                                     List<Protocol> protocols) {
+      List<Protocol> protocols) {
```
swankjesse: thanks for fixing these

swankjesse:
# LGTM (with feedback!)

Despite the massive number of pedantic nitpicks I made, I think this is a great improvement to the library and I’m really happy you made it!

I think the most interesting feedback I have is whether the name `SslContextBuilder` is still appropriate. I think we want a new name for the SSLContext + TrustManagers pair. I suggest `SslClient` but you might have something better. Then the builder can be an inner class of that thing.

If you wanna make that change please do. Otherwise I’m happy to take over your PR, address my obsessive nitpicks, and merge it.

yschimke:
I'm happy to make the changes you suggest.  I'll move DefaultSecurity in Util and go with SslClient.Builder. Let's see what that looks like and iterate again if needed.  No rush on this, JDK 9 is a way off, so thanks for the timely feedback.

okhttp-android-support/src/test/java/okhttp3/internal/huc/CacheAdapterTest.java
yschimke: ok, so if you are suggesting moving this new Builder to okhttp/src/main then it also kills off DefaultSecurity or extra methods in Util.  Is that what you mean?

swankjesse: Two different things.

1. Split SSLContextBuilder into two classes: SSLClient and SSLClient.Builder.
2. Move the new methods in DefaultSecurity to Util. (Though I don’t feel strongly here.)

yschimke:
I think the build failure is unrelated to the branch

```
        [ skipping 6144K ]
  6144K ,,,,,,,, ,,,,,,,, ,,,,,,,, ,,,,,,,, ........ ........  5%  266M 2s
  9216K ........ ........ ........ ........ ........ ........  6% 45.4M 3s
 12288K ........ ........ ........ ........ ........ ........  8% 50.4M 3s
 15360K ........ ........ ........ ........ ........ ........ 10% 70.8M 3s
 18432K ........ ........ ........ ........ ........ ........ 12% 48.5M 3s
 21504K ........ ........ ........ ........ ........ ........ 13% 58.9M 3s
 24576K ........ ........ ........ ........                   15% 59.4M=0.3s
2016-05-26 05:57:15 (55.2 MB/s) - Connection closed at byte 27262976. Giving up.
download failed
Oracle JDK 8 is NOT installed.
dpkg: error processing oracle-java8-installer (--configure):
 subprocess installed post-installation script returned error exit status 1
E: Sub-process /usr/bin/dpkg returned an error code (1)
apt-get.diagnostics
apt-get install failed
```

JakeWharton:
I retried and looks like JDK 7 is working, but yeah JDK 8 is broken. I'll retry it in the morning, or if you push any more commits it'll obviously automatically retry.

yschimke:
My biggest open design questions is how much logic to put in SslClient.Builder?

e.g. should builder allow you to provide a whole trustManager? 
can you create a new builder from an existing SslClient and how that works given SslContext has the existing trustManager internally.

swankjesse:
@yschimke I was thinking it would mostly just be a rearrangement of what we’re currently doing with SSLContextBuilder. That class is almost perfect, except now it has the wrong name because its responsibilities have grown.

yschimke:
That's what I was aiming for, but I drew the line at supporting SslContextBuilder.localhost which depends on the mockwebserver class HeldCertificate.

swankjesse:
I manually merged this! In the process I made a few changes:

I took SslClient out of the public API. We don’t want to be in the business of building SSL APIs. We’ve done this before with certificate pinning and it’s a giant obligation to do it right. Instead we’ll keep the clumsy two-part APIs we have currently. This harms users because they need a bunch of SSL boilerplate if they want to configure their own SSL. But it also means that the users are responsible for their SSL code, not OkHttp. We can revisit this decision later if necessary.

I folded SslContextBuilder into SslClient. No big deal now that it isn’t public API.

PR #2559, opened 2016-05-14
Another approach to handling strange interceptor bodies.

swankjesse:
The previous approach Just Works if users have fancy or interesting interceptors
that genuinely need to swap the response body in an interceptor and keep the
original body.

One problem with that solution is that although it gives expert users a powerful
way to separate response bodies, it also allows normal users to accidentally leak
response bodies in their interceptors.

This is an alternate solution that forbids the expert use case and requires that
closing the response body stream also closes the underlying socket stream. It
throws an exception if that implicit contract is not honored.

I'm fine with either solution but think we should consider both.

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/ConnectionReuseTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/ConnectionReuseTest.java
@@ -16,9 +16,7 @@
 package okhttp3;
 
 import java.io.IOException;
-import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.List;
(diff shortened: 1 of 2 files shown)


JakeWharton:
> This is an alternate solution that forbids the expert use case

Or just requires them to buffer which is probably also a reasonable tradeoff.

This lgtm if you want to move forward with it.

swankjesse:
Yeah, buffering is good enough.

dave-r12:
Oh, gotcha, I guess those are 2 competing cases. 👍  for the alternative approach.

Issue #2568, opened 2016-05-21
sample code doesn't work


realpg:
File:
> okhttp/samples/guide/src/main/java/okhttp3/recipes/CustomTrust.java
Line 55:
`.sslSocketFactory(sslSocketFactory, trustManager)`

It seems okhttp has already changed the api.

PR #2569, opened 2016-05-21
fix issue #2568

realpg:
fix issue #2568
sslSocketFactory() only need 1 parameter of SSLSocketFactory


swankjesse:
The example has the new, unreleased API. We'll release it soon.

Issue #2543, opened 2016-05-08
Recover from REFUSED_STREAM in HTTP/2


swankjesse:
OkHttp promises to persevere when there’s trouble, but it doesn’t recover from `REFUSED_STREAM` in HTTP/2. We should transparently retry when an HTTP/2 server returns a `REFUSED_STREAM` error.

We should set `noNewStreams = true` on the connection that refused the stream. That is a good enough policy that’s in spirit of the HTTP/2 spec: it’ll prevent the connection from being used for any future streams. Then when we retry we’ll get a different connection and with any luck that one will accept the stream.

We need a special case if the refused stream’s streamId is 1. That will let us recover when [Nginx refuses request bodies sent before the settings ack](https://trac.nginx.org/nginx/ticket/959). In that case we should retry on the same physical connection, which we can assume has since ACK’d the settings. It’s still a little bit racy, but unlikely to be a problem in practice. More discussion on this workaround on [issue 2506](https://github.com/square/okhttp/issues/2506) and on the [http-wg list](https://lists.w3.org/Archives/Public/ietf-http-wg/2016AprJun/0240.html).

zyfmix:
can you tell me how to set noNewStreams = true in my code  below:
`
public class CoamAuthInterceptor implements Interceptor {
    @Override
    public Response intercept(Chain chain) throws IOException {

        Request originalRequest = chain.request();

        Response originalResponse = chain.proceed(originalRequest);

        
        Timber.i("HttpURLConnection:=>=>=>Start + HttpURLConnection.HTTP_STATE...");
        if (!originalResponse.isSuccessful()) {

        }
        
        return originalResponse;
    }
}
`
Thank you very much!

swankjesse:
After getting a response from the http-wg list, our recovery policy can be simpler: attempt one retry on the same connection, and then permanently fail the request.

swankjesse:
(And do that one retry on a different connection if the current connection is shutting down.)

dave-r12:
@swankjesse if you didn't already discover this, one other thing I noted when running experiments with nginx is that even if the timing works out, it's possible OkHttp will never send the request. I traced it to this line, https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/internal/framed/FramedConnection.java#L278

I think the headers will need to be flushed prior to writing the data frames.

rfc2822:
[...]

PR #2571, opened 2016-05-21
Recover from REFUSED_STREAM errors in HTTP/2.

swankjesse:
This implements the following policy:

 - If a REFUSED_STREAM error is received, OkHttp will retry the same stream
   on the same socket 1x.
 - If any other error is received, or an additional REFUSED_STREAM error is
   received, OkHttp will retry on a different route if one exists.

We may want to follow up by going through HTTP/2 error codes and deciding
a per-code retry policy, but this should be good enough for now.

Closes: https://github.com/square/okhttp/issues/2543

### DIFF
--- a/mockwebserver/src/main/java/okhttp3/mockwebserver/MockResponse.java
+++ b/mockwebserver/src/main/java/okhttp3/mockwebserver/MockResponse.java
@@ -38,6 +38,7 @@ public final class MockResponse implements Cloneable {
   private TimeUnit throttlePeriodUnit = TimeUnit.SECONDS;
 
   private SocketPolicy socketPolicy = SocketPolicy.KEEP_OPEN;
+  private int http2ErrorCode = -1;
 
   private long bodyDelayAmount = 0;
   private TimeUnit bodyDelayUnit = TimeUnit.MILLISECONDS;
@@ -202,6 +203,20 @@ public final class MockResponse implements Cloneable {
 
   public MockResponse setSocketPolicy(SocketPolicy socketPolicy) {
     this.socketPolicy = socketPolicy;
+    return this;
+  }
+
+  public int getHttp2ErrorCode() {
+    return http2ErrorCode;
+  }
+
+  /**
+   * Sets the <a href="https://tools.ietf.org/html/rfc7540#section-7">HTTP/2 error code</a> to be
+   * returned when resetting the stream. This is only valid with {@link
+   * SocketPolicy#RESET_STREAM_AT_START}.
+   */
+  public MockResponse setHttp2ErrorCode(int http2ErrorCode) {
+    this.http2ErrorCode = http2ErrorCode;
     return this;
   }
 
--- a/mockwebserver/src/main/java/okhttp3/mockwebserver/SocketPolicy.java
+++ b/mockwebserver/src/main/java/okhttp3/mockwebserver/SocketPolicy.java
@@ -88,5 +88,11 @@ public enum SocketPolicy {
    * Don't respond to the request but keep the socket open. For testing read response header timeout
    * issue.
    */
-  NO_RESPONSE
+  NO_RESPONSE,
+
+  /**
+   * Fail HTTP/2 requests without processing them by sending an {@linkplain
+   * MockResponse#getHttp2ErrorCode() HTTP/2 error code}.
+   */
+  RESET_STREAM_AT_START
 }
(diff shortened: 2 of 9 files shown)


okhttp/src/main/java/okhttp3/internal/http/HttpEngine.java
```diff
@@ -371,6 +386,38 @@ public HttpEngine recover(IOException e, boolean routeException) {
     return recover(e, routeException, requestBodyOut);
   }
 
+  private boolean isRecoverable(IOException e, boolean routeException) {
```
swankjesse: (This code is just moved)

okhttp-tests/src/test/java/okhttp3/internal/framed/HttpOverSpdyTest.java
```diff
@@ -582,6 +583,99 @@ protected HttpOverSpdyTest(Protocol protocol) {
...
   }
 
+  @Test public void recoverFromOneRefusedStreamReusesConnection() throws Exception {
+    server.enqueue(new MockResponse()
+        .setSocketPolicy(SocketPolicy.RESET_STREAM_AT_START)
+        .setHttp2ErrorCode(ErrorCode.REFUSED_STREAM.httpCode));
+    server.enqueue(new MockResponse()
+        .setBody("abc"));
+
+    Call call = client.newCall(new Request.Builder()
```
dave-r12: What happens if you use the `SingleInetAddressDns` here? I was testing with nginx 1.9.15 and I noticed the refuse stream stack trace. I think it's because I was using the IP address directly, but still investigating.

swankjesse: You saw something I didn’t see! I turned that on and of course the test started to fail. I fixed it by changing `hasMoreRoutes()` to return true if the current route is still valid.

dave-r12: Oh, gotcha. I couldn't get those tests to run in intellij, but just got it working by upgrading my JDK version.

dave-r12:
Nice, OkHttp can now persevere in yet another hostile scenario. The only thing I had left in my notes - if we wanted this to work with nginx 1.9.15 specifically, I believe we'll need to flush the header frames. Right now, when OkHttp tries to write the data frames, it blocks indefinitely because the window size is 0. The header frames are buffered, so never get sent to nginx and we never receive a window update. See https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/internal/framed/FramedConnection.java#L278

I'll defer to Jake.

mockwebserver/src/main/java/okhttp3/mockwebserver/MockResponse.java
```diff
@@ -205,6 +206,15 @@ public MockResponse setSocketPolicy(SocketPolicy socketPolicy) {
     return this;
   }
 
+  public int getHttp2ErrorCode() {
+    return http2ErrorCode;
+  }
+
+  public MockResponse setHttp2ErrorCode(int http2ErrorCode) {
```
JakeWharton: "FramedErrorCode"?

JakeWharton: +Javadoc explaining that this is used only for `SocketPolicy. RESET_STREAM_AT_START`

mockwebserver/src/main/java/okhttp3/mockwebserver/SocketPolicy.java
```diff
@@ -88,5 +88,11 @@
    * Don't respond to the request but keep the socket open. For testing read response header timeout
    * issue.
    */
-  NO_RESPONSE
+  NO_RESPONSE,
+
+  /**
+   * Fail HTTP/2 requests without processing them by sending an {@link
```
JakeWharton: linkplain

JakeWharton:
only minor nits, lgtm

mockwebserver/src/main/java/okhttp3/mockwebserver/MockResponse.java
swankjesse: The code is defined by HTTP/2. Definitely needs doc. Added.

mockwebserver/src/main/java/okhttp3/mockwebserver/SocketPolicy.java
swankjesse: Fixed.

PR #2572, opened 2016-05-21
Better documentation and examples on closing response bodies.

swankjesse:
Also implement Closeable on Response to make it easier to do the right thing.


JakeWharton:
Lgtm. I'm going to guess animal sniffer is breaking on the samples module. We should just disable it for that module.

PR #2574, opened 2016-05-22
Flush streams if the window size is zero.

swankjesse:
Otherwise we can deadlock, with OkHttp waiting for the server to send a larger
window and the server not knowing that there's a stream that even wants it.

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/internal/framed/Http2ConnectionTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/internal/framed/Http2ConnectionTest.java
@@ -177,6 +177,50 @@ public final class Http2ConnectionTest {
     assertEquals(newMaxFrameSize, connection.frameWriter.maxDataLength());
   }
 
+  /**
+   * Webservers may set the initial window size to zero, which is a special case because it means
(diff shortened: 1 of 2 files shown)


okhttp-tests/src/test/java/okhttp3/internal/framed/Http2ConnectionTest.java
```diff
@@ -177,6 +177,50 @@
...
+  /**
+   * Webservers may set the initial window size to zero, which is a special case because it means
+   * that we have to flush headers immediately before any request body can be sent.
+   * https://github.com/square/okhttp/issues/2543
+   */
+  @Test public void peerSetsZeroFlowControl() throws Exception {
+    peer.setVariantAndClient(HTTP_2, true);
+
+    // Write the mocking script.
+    peer.sendFrame().settings(new Settings().set(INITIAL_WINDOW_SIZE, 0, 0));
```
dave-r12: 👍 the test infrastructure here is awesome.

dave-r12:
LGTM - OkHttp now adapts to the (somewhat awkward) nginx 1.9.15!

PR #2583, opened 2016-05-27
Retain response body when web socket returns non-101 response code.

dave-r12:
This change also fixes an edge case where a server returns a response
that allows a follow up request. Previously, this would cause an
exception because the body was being ignored.


dave-r12:
This builds on #2374, I added some test cases.

swankjesse:
Nice.

Issue #2579, opened 2016-05-25
LoggingInterceptor doesn't log iCalendar files, WebDAV XML response bodies etc. because they are considered binary


rfc2822:
Since okhttp 3.3, `HttpLoggingInterceptor` makes a difference between plain-text and binary bodies and doesn't log binary bodies.

However, I want `HttpLoggingInterceptor` to always log the bodies. Requested iCalendar (plain text) files, WebDAV XML response bodies etc. are considered to be binary be the interceptor and thus dismissed for logging. Extending `HttpLoggingInterceptor` and overwriting `isPlaintext()` is not possible because it's static.

What's the intended way to make sure `HttpLoggingInterceptor` is always logging the bodies?

JakeWharton:
You can copy/paste it into your project and modify to your needs. The provided interceptor isn't designed to solve everyone's use-case, more just the common ones.

swankjesse:
I’m surprised that these files are considered to be binary. Could you paste a gist with a sample file?

rfc2822:
Unfortunately, Github doesn't allow to attach `.vcf` or `.ics` files. So I have zipped the files: [sample.zip](https://github.com/square/okhttp/files/287739/sample.zip)

I guess they're considered binary because they contain control characters within the first 16 characters?

If I remember correctly, I have also seen WebDAV XML responses which were treated binary, probably for the same reason.

What about making `isPlaintext()` non-static, so that users can overwrite it?

swankjesse:
Turns out it’s just the newline characters. I’ll get this fixed and cut a 3.3.1. Sigh.

PR #2586, opened 2016-05-28
Fix HttpLoggingInterceptor to be cool with newlines.

swankjesse:
Unfortunately our plaintext detector treats \r and \n as non-plaintext
characters, and they're completely fine.

Closes: https://github.com/square/okhttp/issues/2579


artem-zinnatullin:
:+1:

Issue #1998, opened 2015-11-14
IllegalArgumentException in Headers$Builder.checkNameAndValue() for content-disposition filename response header


andymutton:
The fix for https://github.com/square/okhttp/issues/891 is causing IllegalArgumentException to be thrown for non ASCII _response_ headers. Most notably, content-disposition, and the filename part. This is basically causing filenames with all the best letters from all the best countries (ü,ä,î,â etc) to crash our app.

The failure mode for us is semi-rare for a variety of reasons, but I imagine this would be far worse for others.

The RFC on the "filename" parameter allows any ISO-8859-1 character: http://tools.ietf.org/html/rfc6266#page-5 so the current behaviour definitely seems to be incorrect.

It makes sense to check request headers. I'm not convinced that it's reasonable for response headers, given that there are plenty of occasions where you just don't control the full stack. Actually, even the unit test that was added in the commit https://github.com/square/okhttp/commit/a57aa43c57819e06fede3f89a653eb6f059449a8 (responseHeaderParsingIsLenient) seems to imply that this is what was intended - must just be a bug.

It would be great if OkHttp could be a little less strict in production situations, but that's sort of unrelated.

Possible fix would be addLenient at https://github.com/square/okhttp/blob/master/okhttp/src/main/java/com/squareup/okhttp/internal/http/FramedTransport.java#L192 but I don't know the ins and outs of this code with enough confidence to make those changes.

java.lang.IllegalArgumentException: Unexpected char 0xf6 at 21 in header value: inline;filename="Andy��IsSad.png"
at com.squareup.okhttp.Headers$Builder.checkNameAndValue(Headers.java:295)
	at com.squareup.okhttp.Headers$Builder.add(Headers.java:245)
	at com.squareup.okhttp.internal.http.FramedTransport.readNameValueBlock(FramedTransport.java:192)
	at com.squareup.okhttp.internal.http.FramedTransport.readResponseHeaders(FramedTransport.java:104)
	at com.squareup.okhttp.internal.http.HttpEngine.readNetworkResponse(HttpEngine.java:906)
	at com.squareup.okhttp.internal.http.HttpEngine.access$300(HttpEngine.java:92)
	at com.squareup.okhttp.internal.http.HttpEngine$NetworkInterceptorChain.proceed(HttpEngine.java:891)
	at com.squareup.okhttp.internal.http.HttpEngine.readResponse(HttpEngine.java:749)
	at com.squareup.okhttp.Call.getResponse(Call.java:268)
[...]

PR #2587, opened 2016-05-28
Don't crash reading non-ASCII characters from the server or the cache.

swankjesse:
This doesn't completely support ISO-8859-1 headers; instead they will most likely
be mangled when they are decoded as UTF-8. If we decide we absolutely must support
ISO-8859-1 here we can do that in another change. (I'm not currently enthusiastic
about this idea.)

But this does prevent OkHttp from crashing when it encounters non-ASCII characters
in headers for HTTP/2, SPDY, and cached responses.

Closes: https://github.com/square/okhttp/issues/1998

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/CacheTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/CacheTest.java
@@ -2320,6 +2320,31 @@ public final class CacheTest {
     assertEquals("v2", server.takeRequest().getHeader("If-None-Match"));
   }
 
+  @Test public void combinedCacheHeadersCanBeNonAscii() throws Exception {
+    server.enqueue(new MockResponse()
+        .addHeader("Last-Modified: " + formatDate(-1, TimeUnit.HOURS))
+        .addHeader("Cache-Control: max-age=0")
+        .addHeaderLenient("Alpha", "α")
+        .addHeaderLenient("β", "Beta")
+        .setBody("abcd"));
+    server.enqueue(new MockResponse()
+        .addHeader("Transfer-Encoding: none")
+        .addHeaderLenient("Gamma", "Γ")
+        .addHeaderLenient("Δ", "Delta")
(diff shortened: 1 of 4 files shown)


BlaineRideCell:
I am still seeing this issue with OkHttp, version 3.6, and Retrofit 2.2.0. Here is my stacktrace: 

`03-13 12:08:17.730 18865-19003/com.my.app.dev E/AndroidRuntime: FATAL EXCEPTION: OkHttp Dispatcher
                                                                                 Process:com.my.app.dev, PID: 18865
                                                                                 java.lang.IllegalArgumentException: Unexpected char 0x10c at 0 in x-locale value: Česká republika
                                                                                     at okhttp3.Headers$Builder.checkNameAndValue(Headers.java:320)
                                                                                     at okhttp3.Headers$Builder.set(Headers.java:300)
                                                                                     at okhttp3.Request$Builder.header(Request.java:164)
                                                                                     at com.ridecell.rest.RetrofitClient$1.intercept(RetrofitClient.java:125)
                                                                                     at okhttp3.RealCall.getResponseWithInterceptorChain(RealCall.java:179)
                                                                                     at okhttp3.RealCall$AsyncCall.execute(RealCall.java:129)
                                                                                     at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1133)
                                                                                     at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:607)
                                                                                     at java.lang.Thread.run(Thread.java:761)
`

manjunath143:
@swankjesse 
I just tried passing ISO-8859-1 non-ASCII string as header. 
     String header="";
                try {
                    String s="ääääääääääääääääääää";
                    header=new String(s.getBytes("UTF-8"), "ISO-8859-1")
                }catch (UnsupportedEncodingException e){
                    e.getStackTrace();
               }

But it does not work. Just curious to know if okhttp supports this ?

PR #2588, opened 2016-05-28
refactor Android specific ChainCleaner code under AndroidPlatform

yschimke:
A take it or leave it PR.  Close if there are good reasons not to make this split.

When I was changing the SslClient code this stuck out.  Seems nicer is all Android specific implementation code can be hidden behind Platform.  I don't use okhttp in Android, so the more cleanly separated the better for my debugging.

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/tls/BasicCertificateChainCleaner.java
+++ b/okhttp/src/main/java/okhttp3/internal/tls/BasicCertificateChainCleaner.java
@@ -1,0 +1,99 @@
+package okhttp3.internal.tls;
+
+import java.security.GeneralSecurityException;
+import java.security.cert.Certificate;
+import java.security.cert.X509Certificate;
+import java.util.ArrayDeque;
+import java.util.ArrayList;
+import java.util.Deque;
+import java.util.Iterator;
+import java.util.List;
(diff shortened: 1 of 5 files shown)


okhttp/src/main/java/okhttp3/internal/tls/BasicCertificateChainCleaner.java
```diff
@@ -0,0 +1,99 @@
+package okhttp3.internal.tls;
```
swankjesse: ultranit: Square copyright

swankjesse:
LGTM!

okhttp/src/main/java/okhttp3/internal/tls/TrustRootIndex.java
```diff
@@ -29,7 +29,7 @@
 
 public abstract class TrustRootIndex {
   /** Returns the trusted CA certificate that signed {@code cert}. */
-  abstract X509Certificate findByIssuerAndSignature(X509Certificate cert);
+  public abstract X509Certificate findByIssuerAndSignature(X509Certificate cert);
```
yschimke: calling out this is public now

swankjesse:
Squashed & merged.

swankjesse:
Thanks!

PR #2589, opened 2016-05-28
samples closing cleanly

yschimke:
Fix AsynchronousGet.java which was hanging and add an example for HTTP/2 that closes cleanly

### DIFF
--- a/samples/guide/src/main/java/okhttp3/recipes/AsynchronousGet.java
+++ b/samples/guide/src/main/java/okhttp3/recipes/AsynchronousGet.java
@@ -50,6 +50,9 @@ public final class AsynchronousGet {
         }
       }
     });
+
+    // required for clean shutdown after using enqueue
+    client.dispatcher().executorService().shutdown();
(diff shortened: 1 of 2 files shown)


swankjesse:
I don’t love it, but I have a tough time expressing why & how. Mostly this will be copy-pasted by developers into their own apps that don’t need this behavior, and will potentially suffer worse performance as a consequence.

yschimke:
I agree, but both these issues hit me.  In the http/2 example, it's likely to hit desktop users when they happen to install JDK 9 and start getting http/2 connections.

For the first example, the current sample doesn't close cleanly.

Other options

1) Some fix to threading e.g. daemon threads or cleanup after some time?
2) an API for clients to close, potentially with failures instead of slowness.
3) add an FAQ instead

swankjesse:
Adding a FAQ seems simplest to me.

Daemon threads could be problematic. Sometimes the only thread you care about is the one that's doing HTTP.

PR #2594, opened 2016-05-31
Remove methods from FramedConnection that are unused

nfuller:
While investigating an Android issue with these methods
I found they were not used. Tidying up.


swankjesse:
Good find.

PR #2596, opened 2016-06-01
Add alpn version support for all Java 8 versions

teresakrause:
Currently the ALPN version used for Java 8 builds is hard coded for a version only compatible with jdk8 updates 71-74. This PR adds separate maven activated profiles for each jdk8 release up to 92, setting “alpn.jdk8.version” based on the jdk minor version. I pulled the jdk-alpn version pairs from this table:
http://www.eclipse.org/jetty/documentation/current/alpn-chapter.html#alpn-versions.

It would be nice to be able to use okhttp with any build of Java 8, and it’s simple to add new profiles for jdk8 releases to the pom file while still supporting older releases.


swankjesse:
Neat!

PR #2597, opened 2016-06-01
WIP: Changes to help reason about how idle time works

nfuller:
Jesse,

This is *not* intended for submission and is against branch 2.7: I want to check a few things with you. I found the ConnectionPool / StreamAllocation / RealConnection arrangement in 2.7 hard to reason about. Apparently there is a bug in the OkHttp on N (~2.5) due to the use of Long.MAX_VALUE to initialize the old equivalent of idleAtNanos.

The arithmetic to work out whether a connection was idle would fail if System.nanoTime() returned a value > 0 and < keepAliveNanos. Generally I think the use of any absolute "can never happen" or "in far future" timestamp is probably bogus when dealing with nanoTime(): no guarantees about sign or magnitude, only really useful for elapsed time or ordering comparisons.

I _believe_ that bug has gone away with subsequent commits which now check whether but I found it hard to prove to myself. This change includes additional docs and assertions. Do you agree with them?

It's possible that streamCount should also be guarded by the connection pool monitor, though that's not currently the case.

Not sure if an equivalent change would transfer to OkHttp 3: a lot of it looks similar.

(Ignore the pom.xml change too)

### DIFF
--- a/okhttp-tests/src/test/java/com/squareup/okhttp/ConnectionPoolTest.java
+++ b/okhttp-tests/src/test/java/com/squareup/okhttp/ConnectionPoolTest.java
@@ -48,7 +48,7 @@ public final class ConnectionPoolTest {
     ConnectionPool pool = new ConnectionPool(Integer.MAX_VALUE, 100L, TimeUnit.NANOSECONDS);
     pool.setCleanupRunnableForTest(emptyRunnable);
 
-    RealConnection c1 = newConnection(pool, routeA1, 50L);
+    RealConnection c1 = newIdleConnection(pool, routeA1, 50L /* idleAtNanos */);
 
     // Running at time 50, the pool returns that nothing can be evicted until time 150.
     assertEquals(100L, pool.cleanup(50L));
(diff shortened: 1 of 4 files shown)


okhttp-tests/src/test/java/com/squareup/okhttp/ConnectionPoolTest.java
```diff
@@ -187,16 +179,46 @@ private void awaitGarbageCollection() throws InterruptedException {
...
+    RealConnection connection = createStubbedConnection(route);
+    synchronized (pool) {
+      allocateAndLeakAllocation(pool, connection);
+      pool.put(connection);
+    }
+    return connection;
+  }
+
+  /** Use a helper method so there's no hidden reference remaining on the stack. */
+  private void allocateAndLeakAllocation(ConnectionPool pool, RealConnection connection) {
```
swankjesse: Inline this? The comment about the stack will apply to `newLeakedConnection()` also

okhttp/src/main/java/com/squareup/okhttp/internal/http/StreamAllocation.java
```diff
@@ -173,9 +173,9 @@ private RealConnection findConnection(int connectTimeout, int readTimeout, int w
 
     Route route = routeSelector.next();
     RealConnection newConnection = new RealConnection(route);
-    acquire(newConnection);
 
     synchronized (connectionPool) {
+      acquire(newConnection);
```
swankjesse: Nice find.

swankjesse:
Wanna submit a PR to master?

swankjesse:
PR for master: https://github.com/square/okhttp/pull/2763

PR #2601, opened 2016-06-02
Add support for an idle callback to Dispatcher.

### LINKED ISSUE #2600: Make Dispatcher easier to hook up to Espresso
Espresso wants two things:
* the ability to synchronously query whether something is idle
* the ability to invoke a callback when transitioning to idle

`Dispatcher` should be able to support the first case using `runningCallsCount() == 0`.

For the second, we can consume a `Runnable` to be invoked on idle. I think ideally this would be a setter instead of a constructor param so that `OkHttpClient` instances can be observed without having to mutate them and set a new `Dispatcher`.

JakeWharton:
Closes #2600.

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/DispatcherTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/DispatcherTest.java
@@ -1,6 +1,7 @@
 package okhttp3;
 
 import java.io.IOException;
+import java.io.InterruptedIOException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -11,10 +12,12 @@ import java.util.Set;
 import java.util.concurrent.AbstractExecutorService;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 import okhttp3.RealCall.AsyncCall;
 import org.junit.Before;
 import org.junit.Test;
 
+import static java.util.concurrent.TimeUnit.SECONDS;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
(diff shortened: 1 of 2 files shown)


okhttp/src/main/java/okhttp3/Dispatcher.java
```diff
@@ -106,6 +107,13 @@ public synchronized int getMaxRequestsPerHost() {
     return maxRequestsPerHost;
   }
 
+  /**
+   * Set a callback to be invoked when the number of
```
swankjesse: ...

swankjesse: Probably worthwhile to note that this is fired before the response body has been consumed.

okhttp/src/main/java/okhttp3/Dispatcher.java
```diff
@@ -173,6 +185,10 @@ synchronized void executed(RealCall call) {
   /** Used by {@code Call#execute} to signal completion. */
   synchronized void finished(Call call) {
     if (!runningSyncCalls.remove(call)) throw new AssertionError("Call wasn't in-flight!");
+
+    if (runningCallsCount() == 0 && idleCallback != null) {
```
swankjesse: It’ll make the code a little uglier, but I think it’s probably worthwhile to rearrange this so we call the callback outside of the `synchronized` block. As is if the callback takes a while to run then no calls can be enqueued which is bad.

swankjesse: . . . my rule of `synchronized` is that you never do anything blocking in a `synchronized` block. (The FramedConnection writer is one exception to this rule.)

okhttp/src/main/java/okhttp3/Dispatcher.java
JakeWharton: Hah. Whoops. I was working on this during builds...

JakeWharton:
PTAL

okhttp-tests/src/test/java/okhttp3/DispatcherTest.java
```diff
@@ -213,6 +215,38 @@
...
+            ready.countDown();
+            return chain.proceed(chain.request());
+          }
+        })
+        .build();
+
+    Thread t1 = makeSynchronousCall(client.newCall(newRequest("http://a/3")));
+    ready.await(5, SECONDS);
+    executor.finishJob("http://a/2");
+    assertFalse(idle.get());
```
swankjesse: Is this line racy? Not sure how we prevent the 3rd request from completing earlier

JakeWharton: Oops yeah. The other test has a CDL waiting inside the interceptor.

JakeWharton: Fixed

okhttp/src/main/java/okhttp3/Dispatcher.java
```diff
@@ -106,6 +107,22 @@ public synchronized int getMaxRequestsPerHost() {
...
+   * Set a callback to be invoked each time the dispatcher becomes idle (when the number of running
+   * calls returns to zero).
+   *
+   * <p>Note: The time at which a {@linkplain Call call} is considered idle is different depending
+   * on whether it was run {@linkplain Call#enqueue(Callback) asynchronously} or
+   * {@linkplain Call#execute() synchronously}. Asynchronous calls become idle after the
+   * {@link Callback#onResponse onResponse} or {@link Callback#onFailure onFailure} callback has
+   * returned. Synchronous calls become idle once {@link Call#execute() execute()} returns. This
+   * means that if you are doing synchronous calls the network layer will not truly be idle until
+   * each returned {@link Response} has been closed.
```
swankjesse: ultranit: .... until every returned

okhttp/src/main/java/okhttp3/Dispatcher.java
```diff
@@ -170,9 +181,29 @@ synchronized void executed(RealCall call) {
...
+  }
+
   /** Used by {@code Call#execute} to signal completion. */
-  synchronized void finished(Call call) {
-    if (!runningSyncCalls.remove(call)) throw new AssertionError("Call wasn't in-flight!");
+  void finished(RealCall call) {
+    finished(runningSyncCalls, call, false);
+  }
+
+  private <T> void finished(Deque<T> calls, T call, boolean promoteCalls) {
```
swankjesse: perfect

swankjesse:
LGTM!

PR #2614, opened 2016-06-10
Let CipherSuites have arbitrary names.

15characterlimi:
This allows Handshake to accept CipherSuite names that do not have a
corresponding predefined instance in CipherSuite.java, including ones
that conscrypt uses today or might use in future.

The following OkHttp 3 APIs are not changed by this CL:

 - ConnectionSpec still uses Strings to represent CipherSuites for now.

 - CipherSuites are interned (value equality is based on object
   equality) in order to not break clients that use ==.

 - CipherSuite.forJavaName("TLS_" + s) is supported as an alias for
   CipherSuite.forJavaName("SSL_" + s) for the 19 specific values of s.
   This means that for 19 specific Strings s, but only for those:
     CipherSuite.forJavaName("TLS_"s).javaName().equals("SSL_" + s)

This CL does introduce these API and behavioral changes:

 - Since CipherSuite is no longer an enum, it no longer extends Enum,
   no longer has name() or valueOf() methods, and no longer implements
   Serializable.

Change-Id: I9a544e1c0a51848d492883b3c9b4579899e2744a


okhttp/src/main/java/okhttp3/CipherSuite.java
```diff
@@ -29,356 +32,421 @@
  * <p>See also <a href="https://android.googlesource.com/platform/external/conscrypt/+/master/src/main/java/org/conscrypt/NativeCrypto.java">NativeCrypto.java</a>
  * from conscrypt, which lists the cipher suites supported by Android.
  */
-public enum CipherSuite {
+public final class CipherSuite {
+  // for interning - must be at top of file
+  private static final ConcurrentMap<String, CipherSuite> INSTANCES = new ConcurrentHashMap<>();
```
swankjesse: (I might later look at replacing this with a copy-on-write LinkedHashMap; ConcurrentMap might be overkill for this seldom-edited case. As is this is fine.)

okhttp/src/main/java/okhttp3/CipherSuite.java
```diff
@@ -29,356 +32,421 @@
  * <p>See also <a href="https://android.googlesource.com/platform/external/conscrypt/+/master/src/main/java/org/conscrypt/NativeCrypto.java">NativeCrypto.java</a>
  * from conscrypt, which lists the cipher suites supported by Android.
  */
-public enum CipherSuite {
+public final class CipherSuite {
+  // for interning - must be at top of file
```
swankjesse: Nit: we follow this: http://nedbatchelder.com/blog/201401/comments_should_be_sentences.html

okhttp/src/main/java/okhttp3/CipherSuite.java
```diff
@@ -29,356 +32,421 @@
...
+  // public static final CipherSuite TLS_RSA_WITH_ARIA_128_CBC_SHA256 = of("TLS_RSA_WITH_ARIA_128_CBC_SHA256", 0xc03c, 6209, MAX_VALUE, MAX_VALUE);
+  // public static final CipherSuite TLS_RSA_WITH_ARIA_256_CBC_SHA384 = of("TLS_RSA_WITH_ARIA_256_CBC_SHA384", 0xc03d, 6209, MAX_VALUE, MAX_VALUE);
+  // public static final CipherSuite TLS_DH_DSS_WITH_ARIA_128_CBC_SHA256 = of("TLS_DH_DSS_WITH_ARIA_128_CBC_SHA256", 0xc03e, 6209, MAX_VALUE, MAX_VALUE);
+  // public static final CipherSuite TLS_DH_DSS_WITH_ARIA_256_CBC_SHA384 = of("TLS_DH_DSS_WITH_ARIA_256_CBC_SHA384", 0xc03f, 6209, MAX_VALUE, MAX_VALUE);
+  // public static final CipherSuite TLS_DH_RSA_WITH_ARIA_128_CBC_SHA256 = of("TLS_DH_RSA_WITH_ARIA_128_CBC_SHA256", 0xc040, 6209, MAX_VALUE, MAX_VALUE);
+  // public static final CipherSuite TLS_DH_RSA_WITH_ARIA_256_CBC_SHA384 = of("TLS_DH_RSA_WITH_ARIA_256_CBC_SHA384", 0xc041, 6209, MAX_VALUE, MAX_VALUE);
+  // public static final CipherSuite TLS_DHE_DSS_WITH_ARIA_128_CBC_SHA256 = of("TLS_DHE_DSS_WITH_ARIA_128_CBC_SHA256", 0xc042, 6209, MAX_VALUE, MAX_VALUE);
+  // public static final CipherSuite TLS_DHE_DSS_WITH_ARIA_256_CBC_SHA384 = of
```
swankjesse: Do we really need this? I think we can get away without this whitelist.

swankjesse: Neat

swankjesse: Dog.

okhttp-tests/src/test/java/okhttp3/CipherSuiteTest.java
```diff
@@ -0,0 +1,116 @@
...
+      // expected
+    }
+  }
+
+  @Test public void hashCode_usesIdentityHashCode() {
+    for (CipherSuite cipherSuite : Arrays.asList(
+        TLS_RSA_EXPORT_WITH_RC4_40_MD5,
+        TLS_RSA_WITH_AES_128_CBC_SHA256
+    )) {
+      assertEquals(cipherSuite.toString(),
```
swankjesse: probably way simpler to just make two assertEquals calls vs. having a loop

okhttp-tests/src/test/java/okhttp3/CipherSuiteTest.java
```diff
@@ -0,0 +1,116 @@
...
+    )) {
+      assertEquals(cipherSuite.javaName, cipherSuite.toString());
+    }
+  }
+
+  /**
+   * Tests backwards compatibility for OkHttp 3.2's behavior of allowing the prefix
+   * "TLS_" as an alias for 19 specific "SSL_" instances.
+   */
+  @Test public void forJavaName_fromLegacyEnumName() {
```
swankjesse: I don’t think we need this.

okhttp-tests/src/test/java/okhttp3/CipherSuiteTest.java
```diff
@@ -0,0 +1,116 @@
+/*
+ * Copyright (C) 2016 Google Inc.
```
swankjesse: Can you assign copyright of this file to Square?

swankjesse: (I believe Square contributions to AOSP and other Google projects are given Google copyrights)

swankjesse:
I like this!

swankjesse:
Just a few nitpicks to address & then we can merge.

okhttp/src/main/java/okhttp3/CipherSuite.java
15characterlimi: I actually changed this now to use a WeakHashMap guarded by a ReentrantReadWriteLock.

The reason I'd like weak references is that the values from which we construct CipherSuite instances may come from a (potentially misbehaving) server, so keeping any value ever mentioned in strongly referenced memory may not be a good idea. The new code only keeps those CipherSuite instances around that are either strongly referenced themselves or whose javaName is strongly referenced (e.g. those that occur as String literals). I've added a test for this as well which calls Runtime.gc().

Re copy-on-write LinkedHashMap: Why copy-on-write? This map is actually modified many times during <clinit> so this copy-on-write would make this O(n^2) without additional care/boilerplate. Re the "Linked" part, why Linked? This Map is never iterated over so the iteration order shouldn't matter.

Let me know if you have any concerns about my "WeakHashMap guarded by a ReentrantReadWriteLock" approach.

okhttp/src/main/java/okhttp3/CipherSuite.java
15characterlimi: Done.

okhttp/src/main/java/okhttp3/CipherSuite.java
15characterlimi: I assume this comment was meant to refer to the whitelist logic in lines 362-399? Done, happy to drop this logic - I don't actually care about this logic personally, I only added it for the benefits of OkHttp's other clients which may not expect this detail of OkHttp's behavior to change in a minor version. But, happy to drop this backwards compatibility since it's fine with you. Done.

okhttp-tests/src/test/java/okhttp3/CipherSuiteTest.java
15characterlimi: Done.

okhttp-tests/src/test/java/okhttp3/CipherSuiteTest.java
15characterlimi: Where in AOSP is that? I thought that generally the copyright line was assigned to "The Android Open Source Project":
https://android.googlesource.com/platform/libcore/+/master/luni/src/main/java/java/net/AddressCache.java
just like Guava's copyright lines are for "The Guava authors".

Perhaps OkHttp's copyright line should refer to "The OkHttp authors" or it should include an AUTHORS file? My understanding is that the Copyright line identifies the author of the file but does not impact Square's or anyone else's ability to reuse this code under the Apache 2 license. Is that correct? If so, a copyright line of "Square, Inc" would be simply incorrect since you've already accepted contributions by many external contributors?

Is that correct / how much do you care / would a copyright line of either "Google Inc" or "The OkHttp authors" (perhaps with an AUTHORS file) be fine with you?

swankjesse: I’d like to have copyright for contributions to Square projects be assigned to Square. If you’re uncomfortable with this, that’s okay, but I’m going to decline this PR. We can still implement this functionality, but I’ll ask somebody else to do so.

swankjesse: ... I talked to our lawyers and they explained to me that this is fine. I’m thinking copyright works like property and you can assign us ownership here, and you can’t. So this is merely attribution. I’m cool with that, and I apologize for slowing this down and for getting this wrong.

15characterlimi: Google's preference would be an AUTHORS file (initially listing Square and Google) or alternatively to change the copyright on all files to "Copyright OkHttp Project Authors" since either of those would be easier to maintain and more indicative of others' contributions over time. Is that something you'd be comfortable with?

swankjesse: Not yet. Things like that trigger new incentives and I'm not prepared to deal with the consequences.

okhttp-tests/src/test/java/okhttp3/CipherSuiteTest.java
15characterlimi: Removed the backwards-compatibility support for the 19 aliases that used to be supported on previous versions of OkHttp. Adjusted the test.

okhttp/src/main/java/okhttp3/CipherSuite.java
15characterlimi: Reverted back to ConcurrentMap because WeakHashMap doesn't work when the value objects have a strong reference to the keys.

swankjesse:
Thanks!

Issue #2609, opened 2016-06-07
Certificate pinner doesn't work if host name has capital letters


jkauppin:
If certificate pinner is set for host "www.HOST.com" if doesn't apply to the calls that has endpoint "www.host.com".

The CertificatePinner.findMatchingPins() should ignore the case because http doesn't care if the name is in uppercase or not.

Tested with OkHttp 2.7.5.

artem-zinnatullin:
Why not pass lowercased hostname to the CertificatePinner?

jkauppin:
Yes, that was the fix on the app. However that feels unsafe since a small typo in the host will disable the certificate pinning. From http point of view capital letters has no meaning.

swankjesse:
Yeah, we should fix this.

PR #2617, opened 2016-06-12
Changed CertificatePinner to ignore case when matching pins.

dave-r12:
Closes: https://github.com/square/okhttp/issues/2609

### DIFF
--- a/okhttp/src/main/java/okhttp3/CertificatePinner.java
+++ b/okhttp/src/main/java/okhttp3/CertificatePinner.java
@@ -237,8 +237,11 @@ public final class CertificatePinner {
   }
 
   static final class Pin {
+    private static final String WILDCARD = "*.";
     /** A hostname like {@code example.com} or a pattern like {@code *.example.com}. */
     final String pattern;
+    /** The canonical hostname, i.e. {@code EXAMPLE.com} becomes {@code example.com}. */
+    final String canonicalHostname;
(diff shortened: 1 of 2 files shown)


swankjesse:
I wonder if a better strategy is to fully canonicalized the host, possibly by roundtripping it through HttpUrl.

```
HttpUrl.parse("http://" + hostWithoutWildcard). host()
```
That'd catch punycode also.
The

dave-r12:
Gotcha, that does sound like a better strategy. I'll see if I can rework this.

okhttp/src/main/java/okhttp3/CertificatePinner.java
```diff
@@ -237,15 +237,21 @@ static ByteString sha256(X509Certificate x509Certificate) {
...
+    /** The canonical hostname, i.e. {@code EXAMPLE.com} becomes {@code example.com}. */
+    final String canonicalHostname;
     /** Either {@code sha1/} or {@code sha256/}. */
     final String hashAlgorithm;
     /** The hash of the pinned certificate using {@link #hashAlgorithm}. */
     final ByteString hash;
 
     Pin(String pattern, String pin) {
       this.pattern = pattern;
+      this.canonicalHostname = pattern.startsWith(WILDCARD)
```
swankjesse: Nice.

swankjesse:
This is very nice. As an added bonus it’ll blow up if the string is bogus.

Issue #270, opened 2013-08-08
Analytics API


swankjesse:
We should add verbose logging for the response cache. Folks occasionally are puzzled when the cache doesn't work. Let's make it easy for them to figure out why.

It would also be handy for debug panels and optimizers to monitor bytes in and out, connections created, etc.

karthikrg:
@swankjesse We've tried using interceptors but they don't allow us to analyze a connection at the granularity we want to. We are looking for more low level details like DNS lookup time, SSL negotiation time etc. Exposing HttpEngine with hooks for collecting this info would be very valuable to us.

swankjesse:
With OkHttp 2.6 you'll be able to provide your own instrumented DNS for that, and you can already provide your own instrumented SSL Engine to time that. So what we really need is just a nice API to put it all together.

swankjesse:
Relevant: https://github.com/raskasa/metrics-okhttp

swankjesse:
Updated sketch:
```
package okhttp3;

import java.net.InetAddress;
import java.util.List;

public class EventListener {
  public static final EventListener NULL_EVENT_LISTENER = new EventListener() {
  };

  void dnsStart(String domainName) {
  }

  void dnsEnd(String domainName, List<InetAddress> inetAddressList) {
  }

  void requestHeadersStart(Call call) {
  }

  void requestHeadersEnd(Call call) {
  }
}
```

swankjesse:
https://www.w3.org/TR/navigation-timing/#sec-navigation-timing-interface

adriancole:
I like that it has a couple important boundaries. Because it is a class not
an interface, we can always add more later as needed without breaking
people.

yschimke:
FWIW I am using a Dns decorator and a SSLSocketFactory decorator to trace the DNS time and connect + handshake estimate.  So these are possible without dedicated APIs.  Separating TCP connect and SSL Handshake would probably require an API change, or this metrics API.

adriancole:
this might also help get out of a problem I've run into with interceptors https://stackoverflow.com/questions/41040727/propagating-state-even-when-backlogged

adriancole:
actually if dnsStart having a reference to the Call that caused it might be handy.. that or some guarantee that dnsStart is on the same thread as some other callback ex requestAccepted

swankjesse:
I think you’re right – hooking up analytics will be the best way to make Zipkin work.

mjbenedict:
[...]

PR #2637, opened 2016-06-17
Add hook to gather network metrics

cmanton:
If enabled, send an event log for each http
request with latency, http status code and
if an exception was thrown.

The idea here is to instrument each  http connection
so that statistically significant data can be gathered
from all consumers about underlying network health.

It should be minimally intrusive when disabled, and
emit a single event log per connection.

Android is using an older version ok2.5 thus the ancient version request.

### DIFF
--- a/okhttp-urlconnection/src/main/java/com/squareup/okhttp/internal/huc/HttpURLConnectionImpl.java
+++ b/okhttp-urlconnection/src/main/java/com/squareup/okhttp/internal/huc/HttpURLConnectionImpl.java
@@ -230,7 +230,18 @@ public class HttpURLConnectionImpl extends HttpURLConnection {
       throw new ProtocolException("This protocol does not support input");
     }
 
-    HttpEngine response = getResponse();
+    final long startTimeMs = System.currentTimeMillis();
+    HttpEngine response;
+    boolean exceptionThrown = true;
+    int responseCode = 0;
+    try {
+      response = getResponse();
+      responseCode = getResponseCode();
+      exceptionThrown = false;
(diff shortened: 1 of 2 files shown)


swankjesse:
Neat! I think we want to do this, but I’m not quite sure where the implementation should go.

If this is for AOSP or another Android fork, I recommend just merging it into your branch. As is it’s awkward to accept this change; it’s on a dead branch that has no ongoing development or releases planned.

Please see this old issue, where we’re thinking about how measurement should work and where it should live.

cmanton:
Thanks for the feedback...I primarily wanted another set of eyes on
gathering these metrics.

The Observer pattern you sketched out in that thread would be a handy
little way to get this information...we could get latency data as a
function of the underlying network layer.

Admittedly, Android is on an older branch and I'm unsure where the core
team is heading...I'm looking to gather these types of metrics to
understand the underlying network performance for the various connectivity
options available to Android.

I'll look into patching this into our branch, but will keep an eye on
future okhttp analytic development in this area.

Thanks again,

Chris
--

swankjesse:
One more possibility – with some recent changes we’re close to supporting interceptors on HttpURLConnection. That makes some interesting things possible for metrics.

cmanton:
That would be great...actually, the interceptors was my first try...I wired
it all up and then wondered why they weren't being triggered ;)

The intent of the interceptors aligns nicely with gathering those types of
metrics.

C
--

swankjesse:
No further action to take here. Stay tuned for interceptors on HttpURLConnection.

Issue #2134, opened 2015-12-19
Consider building a big fancy fake clock infrastructure for OkHttp


swankjesse:
At the moment tests flake because we want to put an upper bound on how long a given task will take, and sometimes the machines we get for free in Travis pause for 7 or 8 seconds.

We have the option of building a fake clock mechanism in OkHttp to make our tests less flaky, but that abstraction has a cost.

​Everything that calls `Object.wait()`, `Thread.sleep()`, `System.nanoTime()` would new operate through the fake clock mechanism:
- mockwebserver throttling
- connection pool sleeping
- discarding the end of a closed stream
- okio async timeouts

The most interesting part of this is definitely Okio . . .  at the moment we use both the socket’s native timeouts, and Okio’s application-layer timeouts. With this change we would need to be a bit clever to make sure native timeouts worked as they’re supposed to. Somehow.

​I think the real challenge is how to expose the fake clock so that it could be accessed by test cases, Okio, & OkHttp. Do application developers want access to such a thing when they’re writing their own code on top of Okio & OkHttp?

JakeWharton:
Retrofit's `MockRetrofit` would want this, if exposed as a public API. And in our consuming Android app we already have this abstraction, and you know about our server one which seems _too_ fancy...

swankjesse:
The TaskRunner being introduced in 4.3 addresses some of this.

swankjesse:
Tracking as #5673.

PR #2638, opened 2016-06-19
A Pipe.

swankjesse:
This is something we need to implement HttpURLConnection atop the native
OkHttp API. The application thread wants to be active, writing to its
OutputStream. But in OkHttp's API, this work wants to happen in the
RequestBody.writeTo() callback. To make this work we give the HttpUrlConnection
OutputStream the sink end of the pipe, and the RequestBody the source end
of the pipe. It's a bit of plumbing, but the benefit is that backpressure,
timeouts, and failures will propagate as expected.


okhttp/src/main/java/okhttp3/internal/io/Pipe.java
```diff
@@ -0,0 +1,182 @@
...
+        waitNanos = timeout.deadlineNanoTime() - start;
+      } else {
+        waitNanos = timeoutNanos;
+      }
+
+      // Attempt to wait that long. This will break out early if the monitor is notified.
+      long elapsedNanos = 0L;
+      if (waitNanos > 0L) {
+        long waitMillis = waitNanos / 1000000L;
+        monitor.wait(waitMillis, (int) (waitNanos - waitMillis * 1000000L));
```
JakeWharton: This API is so gross

okhttp/src/main/java/okhttp3/internal/io/Pipe.java
```diff
@@ -0,0 +1,182 @@
...
+  }
+
+  final class PipeSource implements Source {
+    final Timeout timeout = new Timeout();
+
+    @Override public long read(Buffer sink, long byteCount) throws IOException {
+      synchronized (buffer) {
+        if (sourceClosed) throw new IllegalStateException("closed");
+
+        while (buffer.size() == 0) {
```
JakeWharton: Could be `buffer.exhausted()`. Doesn't matter though.

okhttp/src/main/java/okhttp3/internal/io/Pipe.java
```diff
@@ -0,0 +1,182 @@
...
+ *
+ * <p>When the sink is closed, source reads will continue to complete normally until the buffer has
+ * been exhausted. At that point reads will return -1, indicating the end of the stream. But if the
+ * source is closed first, writes to the sink will immediately fail with an {@link IOException}.
+ */
+public final class Pipe {
+  private final long maxBufferSize;
+  private final Buffer buffer = new Buffer();
+  private boolean sinkClosed;
+  private boolean sourceClosed;
```
JakeWharton: All four of these should be package scoped to avoid synthetic accessors since they are accessed by the nested classes.

JakeWharton:
Really good tests. LGTM except for synthetic accessors make me 😢 .

swankjesse:
Thanks for the quick review. I wanna see how this works in practice. There may be a time to promote it up to Okio, but not yet.

swankjesse:
Sigh.
```
  WaitUntilNotifiedTest.notified:51->assertElapsed:142 expected:<250.0> but was:<389.63476399704814>
```

JakeWharton:
MWS tests and Mock Retrofit tests have the same problem. Maybe time for an
internal Clock abstraction over currentTimeMillis()/nanoTime(), sleep(),
and wait()?

swankjesse:
That’d definitely help. But figuring out where that API lives is quite awkward.

swankjesse:
https://github.com/square/okhttp/issues/2134

PR #2639, opened 2016-06-19
Reimplement HttpURLConnection on the native OkHttp API.

swankjesse:


okhttp/src/main/java/okhttp3/internal/http/OneShotRequestBody.java
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okhttp3.internal.http;
+
+public interface OneShotRequestBody {
```
dave-r12: I understand the intent here, thoughts on calling this `UnrepeatableRequestBody`?

okhttp-urlconnection/src/main/java/okhttp3/internal/huc/HttpURLConnectionImpl.java
```diff
@@ -408,120 +395,46 @@ private String defaultUserAgent() {
...
       }
-
-      StreamAllocation streamAllocation = httpEngine.close();
-      if (!httpEngine.sameConnection(followUp.url())) {
-        streamAllocation.release();
-        streamAllocation = null;
+    } else {
+      executed = true;
+      try {
+        onResponse(call, call.execute());
```
dave-r12: For my own curiosity, it looks like this now supports async and sync call execution. What motivated one over the other?

dave-r12:
This is really interesting and clever, nice.

okhttp/src/main/java/okhttp3/internal/http/OneShotRequestBody.java
swankjesse: Yes, I like that name a lot better. The other problem with OneShotRequestBody is I kept typing `OSRB` into IntelliJ and getting `OutputStreamRequestBody` !

okhttp-urlconnection/src/main/java/okhttp3/internal/huc/HttpURLConnectionImpl.java
swankjesse: I’d prefer to use synchronous exclusively ’cause it uses a single thread. But there’s two cases where that wasn’t feasible:
 - When the application calls `connect()` we need to make an actual connection exist, with a TLS handshake, and return control to the calling thread *before* transmitting the request body. The `Call` API doesn’t let us do this, so I cheat by using an extra thread.
 - When the application wants to stream the request body, we again need to return control to the caller in a place where the `Call` API won’t let us.

okhttp-urlconnection/src/main/java/okhttp3/internal/huc/HttpURLConnectionImpl.java
```diff
@@ -322,81 +319,71 @@ public void setInstanceFollowRedirects(boolean followRedirects) {
...
-    URL url = getURL();
-    HttpUrl httpUrl = Internal.instance.getHttpUrlChecked(url.toString());
-    Request.Builder builder = new Request.Builder()
-        .url(httpUrl)
-        .method(method, placeholderBody);
-    Headers headers = requestHeaders.build();
-    for (int i = 0, size = headers.size(); i < size; i++) {
-      builder.addHeader(headers.name(i), headers.value(i));
+    if (requestHeaders.get("User-Agent") == null) {
+      requestHeaders.add("User-Agent", defaultUserAgent());
```
JakeWharton: Is this needed? Shouldn't the normal `Call` codepath add this header as well?

okhttp-urlconnection/src/main/java/okhttp3/internal/huc/HttpURLConnectionImpl.java
```diff
@@ -322,81 +319,71 @@ public void setInstanceFollowRedirects(boolean followRedirects) {
...
-      engineClient = client.newBuilder()
-          .cache(null)
-          .build();
+    if (!getUseCaches()) {
+      clientBuilder.cache(null);
     }
 
-    return new HttpEngine(engineClient, request, bufferRequestBody, true, false, streamAllocation,
-        requestBody, priorResponse);
+    return call = clientBuilder.build().newCall(request);
```
JakeWharton: It's a bit weird to run the `Call` on a different instance than the one which created it.

okhttp-urlconnection/src/main/java/okhttp3/internal/huc/HttpURLConnectionImpl.java
```diff
@@ -651,4 +563,72 @@ private void setProtocols(String protocolsString, boolean append) {
     this.fixedContentLength = contentLength;
     super.fixedContentLength = (int) Math.min(contentLength, Integer.MAX_VALUE);
   }
+
+  @Override public synchronized void onFailure(Call call, IOException e) {
+    this.callFailure = e;
+    notifyAll();
```
JakeWharton: Should we worry about application code potentially synchronizing and waiting on this instance? The latter seems unlikely but the former might deadlock something?

okhttp-urlconnection/src/main/java/okhttp3/internal/huc/OutputStreamRequestBody.java
```diff
@@ -0,0 +1,105 @@
...
+    this.timeout = sink.timeout();
+    this.expectedContentLength = expectedContentLength;
+
+    // An output stream that writes to sink. If expectedContentLength is not -1, then this expects
+    // exactly that many bytes to be written.
+    this.outputStream = new OutputStream() {
+      private long bytesReceived;
+
+      @Override public void write(int b) throws IOException {
+        write(new byte[] {(byte) b}, 0, 1);
```
JakeWharton: Brutal... but then again who cares about people who use this method?

JakeWharton:
lgtm

okhttp-urlconnection/src/main/java/okhttp3/internal/huc/HttpURLConnectionImpl.java
swankjesse: Sadly the default user agents are different!

```
  private String defaultUserAgent() {
    String agent = System.getProperty("http.agent");
    return agent != null ? Util.toHumanReadableAscii(agent) : Version.userAgent();
  }
```

okhttp-urlconnection/src/main/java/okhttp3/internal/huc/HttpURLConnectionImpl.java
swankjesse: Yeah, though this client is derived from that one. I’d like to follow up to re-enable interceptors. Stay tuned.

okhttp-urlconnection/src/main/java/okhttp3/internal/huc/HttpURLConnectionImpl.java
swankjesse: Yeah, good call. I’ll take a private lock.

okhttp-urlconnection/src/main/java/okhttp3/internal/huc/OutputStreamRequestBody.java
swankjesse: Yeah. What a mess.

JakeWharton:
\o/

PR #2645, opened 2016-06-21
First pass of deleting code only needed by the old HttpURLConnection.

swankjesse:
There's still a lot of refactoring and cleanup possible, but this is an easy first step.


okhttp-tests/src/test/java/okhttp3/internal/io/WaitUntilNotifiedTest.java
```diff
@@ -142,6 +142,6 @@ private double now() {
    * -50..+450 milliseconds.
    */
   private void assertElapsed(double duration, double start) {
-    assertEquals(duration, now() - start + 200d, 250.0);
+    assertEquals(duration, now() - start - 200d, 250.0);
```
swankjesse: dumb bug!

JakeWharton:
LGTM!

okhttp/src/main/java/okhttp3/internal/http/HttpEngine.java
```diff
@@ -356,17 +293,13 @@ public Connection getConnection() {
...
    */
-  public HttpEngine recover(IOException e, boolean routeException, Sink requestBodyOut) {
+  public HttpEngine recover(IOException e, boolean routeException) {
     streamAllocation.streamFailed(e);
 
     if (!client.retryOnConnectionFailure()) {
       return null; // The application layer has forbidden retries.
     }
 
-    if (requestBodyOut != null && !(requestBodyOut instanceof RetryableSink)) {
```
dave-r12: Is this safe to remove? Am thinking about the case of StreamedRequestBody with multiple Routes, and the first route fails.

swankjesse: That's still handled, just not here. It's duplication!

dave-r12: Here's what I mean, it may not be valid, although it does seem to pass with 3.3.1 and 3.2. If it is valid, I think it happened during the HttpURLConnection -> native transition.

```java
  @Test public void streamedBodyNotRepeated() throws Exception {
    server.enqueue(new MockResponse().setSocketPolicy(SocketPolicy.DISCONNECT_AFTER_REQUEST));

    Dns dns = new FakeDns().addresses(Arrays.asList(InetAddress.getByName("localhost"),
        InetAddress.getByName("localhost")));
    OkHttpClient client = new OkHttpClient.Builder()
        .dns(dns)
        .build();
    factory = new OkUrlFactory(client);
    HttpURLConnection connection = factory.open(server.url("/").url());
    connection.setDoOutput(true);
    connection.setChunkedStreamingMode(100);
    OutputStream os = connection.getOutputStream();
    os.write("OutputStream is no fun.".getBytes("UTF-8"));
    os.close();

    try {
      connection.getResponseCode();
      fail();
    }

    assertEquals(1, server.getRequestCount());
  }
```

swankjesse: Yes, good catch. Previously we’d fail early and now we fail late. Lemme fix.

PR #2647, opened 2016-06-22
Don't retry streamed HTTP request bodies.

swankjesse:


swankjesse:
We had this check for redirects but not for failure recovery.

JakeWharton:
lgtm. tests are failing in weird ways...

dave-r12:
👍

PR #2648, opened 2016-06-22
Begin to fold together parts of RealCall and HttpEngine.

swankjesse:
Previously we had an awkward, arbitrary separation because RealCall contained
the stuff that wasn't in HttpURLConnection, and HttpEngine contained everything
that was shared.

It was also awkward because HttpEngine could be interrupted in various parts
of the HttpURLConnection flow: after connecting, while transmitting the request
body, etc. With this change we no longer need to handle API calls while we're
in these intermediate states, which means we can reduce the scope of certain
things from fields to local variables.

There's still a way to go here but this is more easy wins.

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/http/HttpEngine.java
+++ b/okhttp/src/main/java/okhttp3/internal/http/HttpEngine.java
@@ -124,57 +124,19 @@ public final class HttpEngine {
    */
   private boolean transparentGzip;
 
-  /**
-   * True if the request body must be completely buffered before transmission; false if it can be
-   * streamed. Buffering has two advantages: we don't need the content-length in advance and we can
-   * retransmit if necessary. The upside of streaming is that we can save memory.
-   */
-  public final boolean bufferRequestBody;
-
-  /**
-   * The original application-provided request. Never modified by OkHttp. When follow-up requests
-   * are necessary, they are derived from this request.
-   */
-  private final Request userRequest;
-
-  /**
-   * The request to send on the network, or null for no network request. This is derived from the
-   * user request, and customized to support OkHttp features like compression and caching.
-   */
-  private Request networkRequest;
-
-  /**
-   * The cached response, or null if the cache doesn't exist or cannot be used for this request.
-   * Conditional caching means this may be non-null even when the network request is non-null. Never
-   * modified by OkHttp.
-   */
-  private Response cacheResponse;
-
-  /**
-   * The user-visible response. This is derived from either the network response, cache response, or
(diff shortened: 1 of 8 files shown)


okhttp/src/main/java/okhttp3/internal/http/HttpEngine.java
```diff
@@ -131,50 +131,20 @@
    */
   public final boolean bufferRequestBody;
 
-  /**
-   * The original application-provided request. Never modified by OkHttp. When follow-up requests
-   * are necessary, they are derived from this request.
-   */
-  private final Request userRequest;
```
swankjesse: ... all of these are now locals in the combined `sendRequest() / readResponse()` method which is called `proceed()`.

okhttp/src/main/java/okhttp3/internal/http/HttpEngine.java
```diff
@@ -637,17 +581,14 @@ public void readResponse() throws IOException {
 
       httpStream.writeRequestHeaders(request);
 
-      //Update the networkRequest with the possibly updated interceptor request.
-      networkRequest = request;
```
swankjesse: This is slightly a bugfix. If Previously we’d pin the networkRequest in the deepest call to `proceed()`. Now we look at the interceptor’s returned `Response` and take it’s originating request.

okhttp/src/main/java/okhttp3/internal/http/HttpEngine.java
dave-r12: Cool - I noted the class level javadocs still reference sendRequest and readResponse.

okhttp/src/main/java/okhttp3/internal/http/HttpEngine.java
dave-r12: Apologies, I don't completely understand this. I think I see how the behavior is slightly different, but am not sure what the implications are.

dave-r12:
Good stuff 👍

okhttp/src/main/java/okhttp3/internal/http/HttpEngine.java
swankjesse: Ahh, can fix. Follow up!

okhttp/src/main/java/okhttp3/internal/http/HttpEngine.java
swankjesse: Basically we used to snapshot in the network request at the time of the network I/O. Now we just accept whatever network request the network interceptors return.

PR #2650, opened 2016-06-22
IllegalArgumentException

kaiqiangzhang:
Unhandled exception java.lang.IllegalArgumentException: method POST must have a request body.


swankjesse:
The solution here is for you to supply the request body. Null and empty are different things and conflating them is dangerous!

PR #2655, opened 2016-06-24
Add encode for asterisk in form fields.

Shyish:
According to [RFC3986 point 2.2](https://tools.ietf.org/html/rfc3986#page-12) and [RFC1738 point 2.2](http://www.rfc-editor.org/rfc/rfc1738.txt)

> URIs include components and subcomponents that are delimited by
   characters in the "reserved" set.  These characters are called
   "reserved" because they may (or may not) be defined as delimiters by
   the generic syntax, by each scheme-specific syntax, or by the
   implementation-specific syntax of a URI's dereferencing algorithm.

Although:

> [...] only alphanumerics, the special characters "$-_.+!*'(),", and
   reserved characters used for their reserved purposes may be used
   unencoded within a URL.

Therefore as far as I understood is optional to encode some characters ("$-_.+!*'(),"), therefore following the logic that `HttpUrl.FORM_ENCODE_SET` already includes almost all of this characters, I think it would make sense to be consistent within this convention and thus include the asterisk symbol as well.

Please let me know if you agree with this and if the changes made are fine.
Thank you in advance.

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/FormBodyTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/FormBodyTest.java
@@ -152,7 +152,7 @@ public final class FormBodyTest {
     assertEquals("%27", formEncode(39));
     assertEquals("%28", formEncode(40));
     assertEquals("%29", formEncode(41));
-    assertEquals("*", formEncode(42));
+    assertEquals("%2A", formEncode(42));
     assertEquals("%2B", formEncode(43));
     assertEquals("%2C", formEncode(44));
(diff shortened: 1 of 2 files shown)


swankjesse:
What do Chrome and Firefox do?

Shyish:
AFAIK Chrome is sending unencoded *, but the point is that this is merely a matter of consistency since encoding is optional for certain characters, so if you're already encoding some of them, why not all?

swankjesse:
Our goal is consistency with Chrome and Firefox. Encoding characters for the heck of it is not prudent.

PR #2657, opened 2016-06-25
Collapse interceptors into a single chain.

swankjesse:
Instead of having a chain of application interceptors and a separate chain
of network interceptors, this makes a single heterogeneous chain. It makes
the OkHttp core and network caller into interceptors.

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/http/HttpEngine.java
+++ b/okhttp/src/main/java/okhttp3/internal/http/HttpEngine.java
@@ -103,9 +103,6 @@ public final class HttpEngine {
   private final Response priorResponse;
   private HttpStream httpStream;
 
-  /** The time when the request headers were written, or -1 if they haven't been written yet. */
-  long sentRequestMillis = -1;
-
   /**
    * True if this client added an "Accept-Encoding: gzip" header field and is therefore responsible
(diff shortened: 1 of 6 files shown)


okhttp/src/main/java/okhttp3/internal/http/HttpEngine.java
```diff
@@ -273,16 +264,7 @@ private static Response stripBody(Response response) {
...
-   * Called immediately before the transport transmits HTTP request headers. This is used to observe
-   * the sent time should the request be cached.
-   */
-  public void writingRequestHeaders() {
-    if (sentRequestMillis != -1) throw new IllegalStateException();
-    sentRequestMillis = System.currentTimeMillis();
-  }
-
-  boolean permitsRequestBody(Request request) {
+  static boolean permitsRequestBody(Request request) {
```
JakeWharton: Inline?

Looks like you might have a caller still using this as an instance method too.

JakeWharton:
lgtm!

okhttp/src/main/java/okhttp3/internal/http/HttpEngine.java
swankjesse: Fixed.

PR #2658, opened 2016-06-25
Rename HttpURLConnectionImpl to OkHttpURLConnection.

swankjesse:
This is an internal class so nobody should see the name. But having a
completely different name will make it much easier to diagnose future
stacktraces because the internal implementation has changed so much
recently.

Also share fewer internal details with the rest of OkHttp. In particular
remove the OkHttp-Selected-Protocol and OkHttp-Response-Source headers
from the core.


JakeWharton:
Lgtm

PR #2659, opened 2016-06-25
Fix OkHttpURLConnection to not swallow unexpected exceptions.

swankjesse:
Previously we would sometimes delegate to the asynchronous implementation,
and that would prevent exceptions from reaching their destination.

### DIFF
--- a/okhttp-urlconnection/src/main/java/okhttp3/internal/huc/OkHttpURLConnection.java
+++ b/okhttp-urlconnection/src/main/java/okhttp3/internal/huc/OkHttpURLConnection.java
@@ -100,7 +100,7 @@ public final class OkHttpURLConnection extends HttpURLConnection implements Call
   // These fields are guarded by lock.
   private final Object lock = new Object();
   private Response response;
-  private IOException callFailure;
+  private Throwable callFailure;
   Response networkResponse;
   boolean connectPending = true;
   Proxy proxy;
@@ -129,7 +129,7 @@ public final class OkHttpURLConnection extends HttpURLConnection implements Call
           lock.wait(); // Wait 'til the network interceptor is reached or the call fails.
         }
         if (callFailure != null) {
-          throw callFailure;
+          throw propagate(callFailure);
         }
       } catch (InterruptedException e) {
         throw new InterruptedIOException();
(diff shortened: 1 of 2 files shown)


swankjesse:
. . . the underlying problem is that `Callback.onFailure()` expects an IOException. If any other kind of exception prevents a response from being delivered, `Callback.onFailure()` is never called. This is generally legit – the program has crashed! But in this case it’s surprising that a synchronous call never receives a response.

okhttp-urlconnection/src/main/java/okhttp3/internal/huc/OkHttpURLConnection.java
```diff
@@ -587,6 +588,32 @@ private void setProtocols(String protocolsString, boolean append) {
...
   }
 
+  static final class UnexpectedException extends IOException {
+    static final Interceptor INTERCEPTOR = new Interceptor() {
+      @Override public Response intercept(Chain chain) throws IOException {
+        try {
+          return chain.proceed(chain.request());
+        } catch (IOException e) {
+          throw e;
+        } catch (Error | RuntimeException e) {
```
dave-r12: Good catch, literally! I can't say I've caught an Error before, I assume that's ok.

okhttp-urlconnection/src/main/java/okhttp3/internal/huc/OkHttpURLConnection.java
```diff
@@ -587,6 +588,32 @@ private void setProtocols(String protocolsString, boolean append) {
...
+      }
+    };
+
+    public UnexpectedException(Throwable cause) {
+      super(cause);
+    }
+  }
+
+  /** Throws {@code throwable} as either an IOException, RuntimeException, or Error. */
+  private static IOException propagate(Throwable throwable) throws IOException {
```
dave-r12: Should this to be a void return type?

okhttp-urlconnection/src/main/java/okhttp3/internal/huc/OkHttpURLConnection.java
swankjesse: Yeah, it’s most useful to catch an AssertionError . . . it’ll get rethrown shortly and hopefully not caught!

okhttp-urlconnection/src/main/java/okhttp3/internal/huc/OkHttpURLConnection.java
swankjesse: The unused return type exists to make the callsites happy . . . 
https://publicobject.com/2008/05/calling-method-that-always-throws.html

dave-r12: Oh, good read, thanks.

dave-r12:
Nice, lgtm.

PR #2661, opened 2016-06-25
Fix tests to share a common Dispatcher.

swankjesse:
Otherwise we create too many threads and crash Travis.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -253,7 +253,7 @@
               <groupId>org.apache.maven.plugins</groupId>
               <artifactId>maven-surefire-plugin</artifactId>
               <configuration>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -253,7 +253,7 @@
               <groupId>org.apache.maven.plugins</groupId>
               <artifactId>maven-surefire-plugin</artifactId>
               <configuration>
-                <argLine>-Xbootclasspath/p:${bootclasspathPrefix}</argLine>
+                <argLine>-Xbootclasspath/p:${bootclasspathPrefix} -Xms512m -Xmx512m</argLine>
```
swankjesse: This is the real fix!

PR #2663, opened 2016-06-26
throw ConnectException in connection phase

asg-mojtaba:
In many cases it is important to detect whether the sent request is failed because of connection error or not.

For example, In a payment application. When we send a payment request to our servers if we get a connection error it is not necessary to attempt for reversing the payment request but in other cases we need to do a reverse phase.

okhttp throws `IOException` for both kind of connection errors or other kind of errors such as "Socket Read/Write Timeout"; So it is not possible to detect whether it is connection error or something like socket read error. 

In order to fix this problem I made some changes in `AndroidPlatform` & `Platform` class for throwing `ConnectException` in connection phase.

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/AndroidPlatform.java
+++ b/okhttp/src/main/java/okhttp3/internal/AndroidPlatform.java
@@ -20,6 +20,7 @@ import java.io.IOException;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
+import java.net.ConnectException;
 import java.net.InetSocketAddress;
 import java.net.Socket;
 import java.security.cert.Certificate;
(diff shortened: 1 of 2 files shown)


swankjesse:
This doesn't work in general. For example, if you're using an HTTP proxy or load balancer, you may connect just fine and still not reach the target server. Similarly, this makes it difficult to differentiate timeout problems from other unreachability problems.

swankjesse:
Have you considered using an idempotence token?

asg-mojtaba:
I agree with you about proxy and load balancer, but having a non-complete solution is better than nothing.

Maybe someone fix this issue generally but anyway i think its reasonable to throw different exception type in connection phase

asg-mojtaba:
This will at least solve my problem (and those that not using proxy & etc.)

swankjesse:
There's still things we can do to make this better but I’m reluctant to rewrite exception types.

swankjesse:
In the interim you can do this with interceptors. If a network interceptor is reached, you probably have a connection.

asg-mojtaba:
I throw ConnectException that is a IOException so it's just a sign and it's reasonable to throw related exception. 

Anyway it's up to you

asg-mojtaba:
Okhttp must first responsible for what's goes wrong in first step.(proxy and load balancer issues is not important than direct connection issue)

Apache http client do it right. If you have a connect exception why don't notify client ? Through Exception type or what else !

swankjesse:
Are you _sure_ Apache does it right? What happens on pooled connections?

asg-mojtaba:
Yes i'm sure. It throw ConnectionException in case of connection exception.

PR #2664, opened 2016-06-26
Split HttpEngine into 3 interceptors.

swankjesse:
BridgeInterceptor does some basic header management and implements gzip
on responses.

CacheInterceptor implements HTTP response caching.

ConnectInterceptor takes the StreamAllocation that's already allocated
and uses it to create an HttpStream.


okhttp-tests/src/test/java/okhttp3/ConnectionReuseTest.java
```diff
@@ -188,6 +188,9 @@
     assertEquals("a", responseA.body().string());
     assertEquals(0, server.takeRequest().getSequenceNumber());
 
+    // Give the socket a chance to become stale.
+    Thread.sleep(250);
```
swankjesse: (this was a flaky test)

okhttp/src/main/java/okhttp3/internal/http/ConnectInterceptor.java
```diff
@@ -0,0 +1,46 @@
...
+  public ConnectInterceptor(OkHttpClient client) {
+    this.client = client;
+  }
+
+  @Override public Response intercept(Chain chain) throws IOException {
+    RealInterceptorChain realChain = (RealInterceptorChain) chain;
+    Request request = realChain.request();
+    StreamAllocation streamAllocation = realChain.streamAllocation();
+
+    // We need the network to satisfy this request. Possibly for validating a conditional GET.
```
dave-r12: Hmm, would a better spot for this be in the cache interceptor?

swankjesse: I think it's just a distracting comment. The intention is to suggest that we need the network, even though this response might use the cache anyway.

dave-r12:
This fits together nicely. I also like how the responsibilities are separated out. Lgtm.

okhttp/src/main/java/okhttp3/internal/http/HttpEngine.java
```diff
@@ -1,508 +0,0 @@
...
-import static okhttp3.internal.Util.closeQuietly;
-import static okhttp3.internal.Util.discard;
-import static okhttp3.internal.Util.hostHeader;
-import static okhttp3.internal.http.StatusLine.HTTP_CONTINUE;
-
-/**
- * Handles a single HTTP request/response pair. The request and response may be served by the HTTP
- * response cache, by the network, or by both in the event of a conditional GET.
- */
-public final class HttpEngine {
```
JakeWharton: Rest in pieces.

PR #2666, opened 2016-06-27
Run IntelliJ's analysis on OkHttp.

swankjesse:
This made a few hundred suggestions, many of which are thoughtful but
I'm deliberately ignoring anyway.

The following fixes are good ideas.


JakeWharton:
lgtm

PR #2668, opened 2016-06-27
Cleanup refactoring: Move 3 methods into DelegatingHttpsURLConnection

15characterlimi:
These three methods were delegating to delegate; their implementation
was duplicated across both subclasses of DelegatingHttpsURLConnection.

This change moves them into the base class in order to
  - avoid duplication
  - separate the responsibility of delegation (base class) from the
    responsibility of change to behavior (subclasses).

### DIFF
--- a/okhttp-android-support/src/main/java/okhttp3/internal/huc/JavaApiConverter.java
+++ b/okhttp-android-support/src/main/java/okhttp3/internal/huc/JavaApiConverter.java
@@ -859,18 +859,6 @@ public final class JavaApiConverter {
     @Override public SSLSocketFactory getSSLSocketFactory() {
       throw throwRequestSslAccessException();
     }
-
-    @Override public long getContentLengthLong() {
-      return delegate.getContentLengthLong();
-    }
-
-    @Override public void setFixedLengthStreamingMode(long contentLength) {
-      delegate.setFixedLengthStreamingMode(contentLength);
-    }
-
-    @Override public long getHeaderFieldLong(String field, long defaultValue) {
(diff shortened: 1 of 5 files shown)


15characterlimi:
Huh, seems like this doesn't build - I'm looking into it, sorry about the noise.

15characterlimi:
I think this requires:
 a) adding @IgnoreJRERequirement to the DelegatingHttpsURLConnection base class,
 b) adding a dependency on animal-sniffer-annotations to okhttp.

Question 1: Any problem with either of a-b)?
Question 2: Any idea why the animal-sniffer check didn't complain about those three methods when they were in the subclasses (CacheHttpsURLConnection, OkHttpsURLConnection)? Does it have anything to do with those classes being final?

swankjesse:
I think both (a) and (b) are fine. We’re already doing similarly in Okio. Note that the dependency there is `optional`; that prevents it from being required downstream.
https://github.com/square/okio/blob/master/okio/pom.xml

I’m not quite sure what’s going on with animal-sniffer. It’s a clever hack; I wouldn’t put too much trust in it!

15characterlimi:
Okay, did both (a) and (b) now. Build now succeeding.

swankjesse:
Merged manually. (I needed to add `<optional>true</optional>` to the new dep.)

swankjesse:
https://github.com/square/okhttp/commit/f5099ce11f9b9df4eaf2000b80343868f1799176

15characterlimi:
Thanks! - sorry about the hassle with the manual merge.

PR #2673, opened 2016-06-28
Do not allow remote peer to arbitrarily size the HPACK decoder dynamic table.

dave-r12:
It's my understanding this setting is meant to signal the amount of memory the _encoder_ can use, not the _decoder_. I think we'll want to respect this value with the new HPACK writer. I was going to see if I can incorporate that change but wanted to make sure I was reading the spec right.

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/internal/framed/Http2ConnectionTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/internal/framed/Http2ConnectionTest.java
@@ -27,6 +27,7 @@ import okio.BufferedSource;
 import okio.Okio;
 import okio.Source;
 import org.junit.After;
+import org.junit.Ignore;
 import org.junit.Test;
 
 import static okhttp3.TestUtil.headerEntries;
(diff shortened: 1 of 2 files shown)


swankjesse:
I’m also a bit unsure on how this is supposed to work. What’s the next step?

dave-r12:
I was hoping you'd have the answer. I'll research this further, including other HTTP2 client implementations and maybe even post to the mailing list.

swankjesse:
Sadly my HPACK knowledge is pretty weak. I’m not the author of that code, though I have done some code reviews & read the spec. Definitely seems like a possible problem though.

dave-r12:
I found this on the mailing list https://lists.w3.org/Archives/Public/ietf-http-wg/2015OctDec/0088.html. The last message in the thread:
> The way negotiation works:
- Decoder side sets the upper bound via SETTINGS_HEADER_TABLE_SIZE.
- Encoder side sets the actual size via dynamic table updates (inside
  HPACK bitstream) within limits set by decoder.

From chrome HPACK Encoder, https://chromium.googlesource.com/chromium/src/+/master/net/spdy/hpack/hpack_encoder.h#52
>   // Called upon a change to SETTINGS_HEADER_TABLE_SIZE. Specifically, this
  // is to be called after receiving (and sending an acknowledgement for) a
  // SETTINGS_HEADER_TABLE_SIZE update from the remote decoding endpoint.

I think this change is safe. As a next step, I can attempt to implement the handling code in the encoder (writer). My feeling is that it will mostly be a reshuffling of the logic already in the decoder (reader).

adriancole:
I wrote this code, and I agree it looks like you've fixed a bug based on your sources and also the spec 4.2.

https://tools.ietf.org/html/rfc7541

The test, if anything, should show the encoder limiting up to the peer's table size setting.

swankjesse:
Sweet!

Issue #2628, opened 2016-06-15
Reading cache headers crashes when header contains utf-8 chars


mpost:
When a `Cache` is set on a okhttp3 instance and a connection to a server is created that uses utf-8 chars in its header response the library crashes with the following exception:

```java
  java.lang.IllegalArgumentException: Unexpected char 0xfffd at 67 in If-None-Match value: "562949954286466-404-Fri Jun 03 2016 17:43:42 GMT+0200 (Mitteleurop�ische Sommerzeit)"
   at okhttp3.Headers$Builder.checkNameAndValue(Headers.java:320)
   at okhttp3.Headers$Builder.set(Headers.java:300)
   at okhttp3.Request$Builder.header(Request.java:163)
   at okhttp3.internal.http.CacheStrategy$Factory.getCandidate(CacheStrategy.java:235)
   at okhttp3.internal.http.CacheStrategy$Factory.get(CacheStrategy.java:168)
   at okhttp3.internal.http.HttpEngine.sendRequest(HttpEngine.java:207)
   at okhttp3.RealCall.getResponse(RealCall.java:235)
   at okhttp3.RealCall$ApplicationInterceptorChain.proceed(RealCall.java:193)
   at okhttp3.RealCall.getResponseWithInterceptorChain(RealCall.java:155)
   at okhttp3.RealCall.execute(RealCall.java:57)
```

The problem can be experienced on `3.3.1` as well as an `3.4.0-SNAPSHOT`.

I think an attempt to fix similar issues has been made in the pull request https://github.com/square/okhttp/pull/2587 by @swankjesse.

swankjesse:
Yep. Yuck. Can fix.

PR #2684, opened 2016-07-01
Permit non-ASCII ETag headers.

swankjesse:
These are currently being encoded in UTF-8, and the charset is not
otherwise specified, so the result is not necessarily going to
succeed at caching.

But this avoids crashing when the server returns an unexpected ETag.

Closes: https://github.com/square/okhttp/issues/2628


JakeWharton:
lgtm

Issue #2640, opened 2016-06-20
isCleartextTrafficPermitted is broken


swankjesse:
If the app is configured to require TLS for a single host, we do the wrong thing and require TLS for all hosts.

PR #2686, opened 2016-07-02
Always pass a host to NetworkSecurityPolicy.isCleartextTrafficPermitted

swankjesse:
Previously we were misinterpretting which hosts this method applied to.
Suppose an Android app was configured to require TLS for bank.com and
not for any other address. The NetworkSecurityPolicy.isCleartextTrafficPermitted()
method would return false because cleartext traffic wasn't universally
permitted. And OkHttp would incorrectly forbid cleartext communication
to other hosts like puppies.com.

Closes: https://github.com/square/okhttp/issues/2640


dave-r12:
Lgtm.

PR #2688, opened 2016-07-02
Java9 tests

yschimke:
ConnectionReuseTest

builder.sslSocketFactory(sslSocketFactory2) always fails on JDK9, so using a non-deprecated form

URLConnectionTest

enable SSL in the JDK9 specific tests to avoid ClassCastException

Http2ConnectionTest

settings value is unpredictable due to async nature of test

CertificatePinnerChainValidationTest

fails before serve can be started due to JDK9 moving to PKCS12

### DIFF
--- a/okhttp-android-support/src/test/java/okhttp3/internal/huc/ResponseCacheTest.java
+++ b/okhttp-android-support/src/test/java/okhttp3/internal/huc/ResponseCacheTest.java
@@ -84,6 +84,7 @@ import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
+import static org.junit.Assume.assumeFalse;
 
 /**
  * Tests the interaction between OkHttp and {@link ResponseCache}. Based on okhttp3.CacheTest with
@@ -269,6 +270,8 @@ public final class ResponseCacheTest {
   }
 
   @Test public void secureResponseCaching() throws IOException {
+    assumeFalse(getPlatform().equals("jdk9"));
+
     server.useHttps(sslClient.socketFactory, false);
     server.enqueue(new MockResponse()
         .addHeader("Last-Modified: " + formatDate(-1, TimeUnit.HOURS))
@@ -1983,6 +1986,8 @@ public final class ResponseCacheTest {
   }
 
   @Test public void cacheReturnsInsecureResponseForSecureRequest() throws IOException {
+    assumeFalse(getPlatform().equals("jdk9"));
+
     server.useHttps(sslClient.socketFactory, false);
     server.enqueue(new MockResponse().setBody("ABC"));
     server.enqueue(new MockResponse().setBody("DEF"));
@@ -2187,5 +2192,9 @@ public final class ResponseCacheTest {
     OkHttpClient.Builder builder = urlFactory.client().newBuilder();
     Internal.instance.setCache(builder, internalCache);
     urlFactory.setClient(builder.build());
+  }
+
+  private String getPlatform() {
+    return System.getProperty("okhttp.platform", "platform");
   }
 }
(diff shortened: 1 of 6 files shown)


okhttp-tests/src/test/java/okhttp3/ConnectionReuseTest.java
```diff
@@ -253,11 +253,9 @@
     response.body().close();
 
     // This client shares a connection pool but has a different SSL socket factory.
-    SSLContext sslContext2 = SSLContext.getInstance("TLS");
-    sslContext2.init(null, null, null);
-    SSLSocketFactory sslSocketFactory2 = sslContext2.getSocketFactory();
+    SslClient sslClient2 = new SslClient.Builder().build();
```
swankjesse: Nice!

okhttp-tests/src/test/java/okhttp3/internal/tls/CertificatePinnerChainValidationTest.java
```diff
@@ -197,8 +199,20 @@
...
-        .certificateChain(rogueCertificate.keyPair, rogueCertificate.certificate, compromisedIntermediateCa.certificate, goodCertificate.certificate, rootCa.certificate)
+
+    SslClient.Builder sslBuilder = new SslClient.Builder();
+
+    // Test setup fails on JDK9
+    // java.security.KeyStoreException: Certificate chain is not valid
+    // at sun.security.pkcs12.PKCS12KeyStore.setKeyEntry
+    // http://openjdk.java.net/jeps/229
+    // http://hg.openjdk.java.net/jdk9/jdk9/jdk/file/2c1c21d11e58/src/share/classes/sun/security/pkcs12/PKCS12KeyStore.java#l596
+    if (getPlatform().equals("jdk9")) {
```
swankjesse: Should this workaround just go in SslClient.Builder? Seems simpler and avoids distracting from the tests?

swankjesse:
LGTM! Please squash your changes and I’ll merge. (http://rebaseandsqua.sh for a quick script).

okhttp-tests/src/test/java/okhttp3/internal/tls/CertificatePinnerChainValidationTest.java
yschimke: The reason I didn't is that having the tests run with the default KeyStore provider (PKCS12 in JDK 9+) seems generally good.  These specific tests want to make a server that is invalid and only JKS allows that.

okhttp-android-support/src/test/java/okhttp3/internal/huc/ResponseCacheTest.java
```diff
@@ -269,6 +270,8 @@ private void testResponseCaching(TransferKind transferKind) throws IOException {
   }
 
   @Test public void secureResponseCaching() throws IOException {
+    assumeFalse(getPlatform().equals("jdk9"));
```
vanniktech: `"jdk9".equals(getPlatform())` would not throw a NPE when `getPlatform()` returns null

yschimke: It's a good idiomatic improvement. In this case getPlatform() can't be null because of it's implementation System.getProperty("okhttp.platform", "platform"), but your point is valid.

For now, I'd rather keep this inline with other similar usages, and change all in one go if we want to.

```
$ git grep getPlatform...equals | awk -F / '{print $NF}'
ResponseCacheTest.java:    assumeFalse(getPlatform().equals("jdk9"));
ResponseCacheTest.java:    assumeFalse(getPlatform().equals("jdk9"));
URLConnectionTest.java:    assumeTrue(getPlatform().equals("jdk9"));
Jdk9PlatformTest.java:    assumeTrue(getPlatform().equals("jdk9"));
Jdk9PlatformTest.java:    assumeTrue(getPlatform().equals("jdk9"));
JdkWithJettyBootPlatformTest.java:    assumeTrue(getPlatform().equals("jdk-with-jetty-boot"));
CertificatePinnerChainValidationTest.java:    if (getPlatform().equals("jdk9")) {
CertificatePinnerChainValidationTest.java:    if (getPlatform().equals("jdk9")) {
UrlConnectionCacheTest.java:    assumeFalse(getPlatform().equals("jdk9"));
UrlConnectionCacheTest.java:    assumeFalse(getPlatform().equals("jdk9"));
```

swankjesse:
It’s really handy when writing say, `CHANGELOG.md` to have a simple history that only includes what was actually accepted.

PR #2689, opened 2016-07-03
Removed ability to change the HPACK decoder header table size setting.

dave-r12:
Previously, we exposed this method and called it when receiving the
header table size setting from a remote peer. We concluded this was not
the intent of the spec and removed that code.


dave-r12:
This is now dead code, and if we truly want to support this we'll need additional code (i.e. sending a settings frame, waiting for ACK, etc.)

swankjesse:
Dig!

Issue #2667, opened 2016-06-27
New HttpURLConnection shouldn’t block on the Dispatcher


swankjesse:
At the moment we exercise the Dispatcher from within HttpURLConnection. This has some pretty low limits that could cause unexpected low amounts of concurrency in some apps.

Alternatives are to just bump the dispatcher limit for HttpURLConnection calls, somehow.

PR #2692, opened 2016-07-04
Don't share the OkHttpClient's Dispatcher in HttpURLConnection.

swankjesse:
Any limits enforced here would be awkward because there's an application
thread blocked on the work anyway.

Closes: https://github.com/square/okhttp/issues/2667


dave-r12:
Makes sense, lgtm.

Issue #2701, opened 2016-07-06
Make TreeMap<>() with String.CASE_INSENSITIVE_ORDER in Headers.toMultimap()


a-k-g:
Feature request: Improve behavior of Headers.toMultimap - It seems that in commit https://github.com/square/okhttp/commit/bab8943f2b5c8bc44a64067ada556aa62fcec3a7 , the API for `Headers.toMultimap` has changed slightly. Earlier it used to preserve the case for the headers it was initialized with. Since this commit it has started to lower the case for the header names. I see why the change might make sense but it's breaking consumers who either made assumptions that case is preserved, or were using constants from `com.google.common.net.HttpHeaders` or `javax.ws.rs.core`. 

My FR is for this method to create a `TreeMap<>()` with `String.CASE_INSENSITIVE_ORDER` so that consumers of this multimap don't get affected. The proposed enhancement would also make the consumption of headers simple. It would enable us to directly use constants from `com.google.common.net.HttpHeaders` or `javax.ws.rs.core` both of which use Train-Case, example `Content-Type` for `javax.ws.rs.core.CONTENT_TYPE`.

PR #2702, opened 2016-07-06
Use TreeMap in Headers.toMultimap()

a-k-g:
For https://github.com/square/okhttp/issues/2701


swankjesse:
I see your pull request and I raise you one more:
https://github.com/google/guava/pull/2518

I don’t think it matters much either way, but we trade one form of information loss (the case of headers) for another form (the order of headers).

swankjesse:
Thanks!

PR #2708, opened 2016-07-08
Update Javadoc example for HttpUrl.encodedPath()

jrodbx:
Per https://github.com/square/okhttp/pull/1629


swankjesse:
These two will be the same, right?

jrodbx:
yup!  Looking at the [history](https://github.com/square/okhttp/blame/master/okhttp/src/main/java/okhttp3/HttpUrl.java#L540), https://github.com/square/okhttp/pull/1629 only changed the method name, while https://github.com/square/okhttp/pull/1648 only changed implementation details.

PR #2713, opened 2016-07-09
Allow to clone calls

### LINKED ISSUE #2709: Allow to clone a call
I created the following Kotlin extension method on `Call` to use calls with RxJava.

```
fun Call.asObservable(): Observable<Response> {
    return Observable.create { subscriber ->
        subscriber.add(Subscriptions.create { cancel() })

        try {
            val response = execute()
            if (!subscriber.isUnsubscribed) {
                subscriber.onNext(response)
                subscriber.onCompleted()
            }
        } catch (error: IOException) {
            if (!subscriber.isUnsubscribed) {
                subscriber.onError(error)
            }
        }
    }
}
```

It works fine normally, but if a request fails and I use a `.retry()` in the chain, it throws an exception because the call can't be reused. Then I saw Retrofit has a `.clone()` method on its `Call` class for the RxJava's adapter (I guess for a similar scenario).

Though I don't know if it makes sense to include a clone, because the original call would never be executed or canceled.

My current solution is to always create a new call inside the observable, but then I need a reference to the client and the request, and I would like to abstract that.

inorichi:
Fixes #2709.

Implementing this was pretty much straightforward, but I don't know if I missed something.

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/CallTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/CallTest.java
@@ -619,6 +619,42 @@ public final class CallTest {
     assertEquals("SyncApiTest", server.takeRequest().getHeader("User-Agent"));
   }
 
+  @Test public void legalToExecuteTwiceCloning() throws Exception {
+    server.enqueue(new MockResponse().setBody("abc"));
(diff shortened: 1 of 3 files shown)


swankjesse:
LGTM!

PR #2714, opened 2016-07-09
Rename FramedConnection to Http2Connection.

swankjesse:
Also rename HttpStream to HttpCodec. This is the interface implemented
for both HTTP/1.1 and HTTP/2. The HTTP/2 codec creates a stream when
it is used.


okhttp-hpacktests/src/test/java/okhttp3/internal/http2/HpackRoundTripTest.java
```diff
@@ -13,11 +13,12 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package okhttp3.internal.framed;
+package okhttp3.internal.http2;
 
 import java.util.Collection;
-import okhttp3.internal.framed.hpackjson.Case;
-import okhttp3.internal.framed.hpackjson.Story;
+import okhttp3.internal.framed.Hpack;
```
JakeWharton: Missed some?

PR #2715, opened 2016-07-10
Fix a major bug in our HPACK encoder.

swankjesse:
Running the HPACK regression suite shows that in some cases we may
use the wrong dynamic index after the dynamic table has been resized.

This fixes the problem and simplifies the implementation to avoid
maps with Integer values, which were difficult to reason about and
easy to get out-of-sync with the main table.

### DIFF
--- a/okhttp-hpacktests/pom.xml
+++ b/okhttp-hpacktests/pom.xml
@@ -8,7 +8,7 @@
   <parent>
     <groupId>com.squareup.okhttp3</groupId>
     <artifactId>parent</artifactId>
-    <version>3.3.0-SNAPSHOT</version>
+    <version>3.5.0-SNAPSHOT</version>
   </parent>
 
   <artifactId>okhttp-hpacktests</artifactId>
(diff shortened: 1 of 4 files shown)


okhttp-hpacktests/pom.xml
```diff
@@ -8,7 +8,7 @@
   <parent>
     <groupId>com.squareup.okhttp3</groupId>
     <artifactId>parent</artifactId>
-    <version>3.3.0-SNAPSHOT</version>
```
swankjesse: This was what made us vulnerable to the regression. We weren’t running HPACK tests against the current code; we were running them against a known working version.

swankjesse:
Note the branch is okhttp_34! I’ll also put this in master once it’s good!

okhttp-hpacktests/pom.xml
dave-r12: That was unfortunate. Are there any direct (or indirect) impediments to running these with the normal build? I don't see it in the parent pom.xml.

swankjesse: It’s currently disabled due to its dependence on git submodules. We could just snapshot the sample data though.

PR #2716, opened 2016-07-10
Fix a major bug in our HPACK encoder.

swankjesse:
Running the HPACK regression suite shows that in some cases we may
use the wrong dynamic index after the dynamic table has been resized.

This fixes the problem and simplifies the implementation to avoid
maps with Integer values, which were difficult to reason about and
easy to get out-of-sync with the main table.


swankjesse:
Self-merging. This is taking an already-reviewed PR from okhttp_34 and applying it to master.
https://github.com/square/okhttp/pull/2715

Issue #2677, opened 2016-06-29
Recover gracefully instead of throwing IllegalCharsetNameException


Biswangluo:
Wrong data from server 

# 1.
`<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
<html xmlns="http://www.w3.org/1999/xhtml"><head>
<meta content="text/html; charset=utf-8" http-equiv="Content-Type">
<title>系统发生错误</title>
<style type="text/css">`

#2
557 4920-5047/arkui.live W/System.err: java.nio.charset.IllegalCharsetNameException: 'utf-8'
06-29 12:47:49.557 4920-5047/arkui.live W/System.err:     at java.nio.charset.Charset.checkCharsetName(Charset.java:197)
06-29 12:47:49.557 4920-5047/arkui.live W/System.err:     at java.nio.charset.Charset.forName(Charset.java:295)
06-29 12:47:49.557 4920-5047/arkui.live W/System.err:     at okhttp3.MediaType.charset(MediaType.java:103)
06-29 12:47:49.557 4920-5047/arkui.live W/System.err:     at okhttp3.ResponseBody.charset(ResponseBody.java:159)
06-29 12:47:49.557 4920-5047/arkui.live W/System.err:     at okhttp3.ResponseBody.string(ResponseBody.java:154)
06-29 12:47:49.557 4920-5047/arkui.live W/System.err:     at arkui.live.dao.LoadDataUtil$1.onResponse(LoadDataUtil.java:82)
06-29 12:47:49.557 4920-5047/arkui.live W/System.err:     at okhttp3.RealCall$AsyncCall.execute(RealCall.java:133)
06-29 12:47:49.557 4920-5047/arkui.live W/System.err:     at okhttp3.internal.NamedRunnable.run(NamedRunnable.java:32)
06-29 12:47:49.557 4920-5047/arkui.live W/System.err:     at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1112)
06-29 12:47:49.557 4920-5047/arkui.live W/System.err:     at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:587)
06-29 12:47:49.557 4920-5047/arkui.live W/System.err:     at java.lang.Thread.run(Thread.java:818)

swankjesse:
[...]

PR #2717, opened 2016-07-11
Be lenient if a character set is 'single-quoted'.

swankjesse:
It's what Chrome does. Firefox and Safari ignore it. I'd love to ignore
it, but in our API that really just means the caller needs to handle an
InvalidCharsetNameException and nobody is expecting that.

Closes: https://github.com/square/okhttp/issues/2677


JakeWharton:
Lgtm

Issue #2549, opened 2016-05-11
okhttp Cookie strips off leading . from domain


karthikrg:
`  private static String parseDomain(String s) {
    if (s.endsWith(".")) {
      throw new IllegalArgumentException();
    }
    if (s.startsWith(".")) {
      s = s.substring(1);
    }
    String canonicalDomain = domainToAscii(s);
    if (canonicalDomain == null) {
      throw new IllegalArgumentException();
    }
    return canonicalDomain;
  }`

Cookie seems to strip off the leading . from the domain; this in turn causes Android's CookiePolicy#ACCEPT_ORIGINAL_SERVER to drop the cookie when used in conjunction with JavaNetCookieStore and a CookieManager. Currently I am getting around this by using a custom interceptor to manually parse the header and directly persist cookies into CookieManager.

Why don't we preserve the domain as is from the header?

swankjesse:
Yup, this is a problem when converting to a format that works with CookieManager. (It isn't a problem with our cookie class itself.)

boxme:
@swankjesse HttpCookie#domainMatches in CookiePolicy#ACCEPT_ORIGINAL_SERVER has a slightly different set of matching rules compared to Cookie#domainMatch. Is there a reason for this?

I'm trying to understand the rationale to stripping off the leading . from the domain.

Thanks!

swankjesse:
It's a bug! Can fix.

PR #2722, opened 2016-07-13
Inject a leading . for better matching under JavaNetCookieJar

swankjesse:
Closes: https://github.com/square/okhttp/issues/2549

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/CookiesTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/CookiesTest.java
@@ -23,6 +23,7 @@ import java.net.HttpCookie;
 import java.net.HttpURLConnection;
 import java.net.InetAddress;
 import java.net.URI;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


JakeWharton:
lgtm

sisel:
LGTM,
seems it's not released yet , look forward to next release.

michaelzengke:
So, this change suddenly reverts from always RFC 6265 to always RFC 2965. 

However, Cookie#parseDomain(String) always removes the leading ".", meaning Cookie impl chooses RFC 6265.

Are they contradicting each other?

Looks like either Cookie#parseDomain(String) should be agnostic to the upstream's RFC compliance, or CookieJar#saveFromResponse(HttpUrl, List<Cookie>) should be CookieJar#saveFromResponse(HttpUrl, List<String>) such that CookieJar impl can have its own choice of RFC compliance.

swankjesse:
Here’s how I think of it:
 * All of OkHttp prefers RFC 6265 and uses that.
 * All of java.net prefers RFC 2965 and uses that.

The `JavaNetCookieJar` class is a bridge from OkHttp to java.net, so it formats cookies as RFC 2965 wants ’em. I’d like to also [someday](https://github.com/square/okhttp/issues/2890) make a `RealCookieJar` or something that doesn’t delegate to obsolete code in java.net.

raskasa:
@swankjesse Should your explanation be added to the class-level Javadoc for, or as a code comment within, `JavaNetCookieJar`?

michaelzengke:
I get your point. The problem here is, within our server farm, some cookies are issued per RFC 2965, some per RFC 6265. Our apps out in the market won't survive if okhttp3 modifies the domain in any way, add or remove the leading ".".

After all, CookieJar is designed such that users can choose various cookie handling strategy, isn't it?

If the interface is designed as CookieJar#saveFromResponse(HttpUrl, List), the choice is in the user's hands. Otherwise, today, when Cookie is passed down, whether the original domain has or has not the leading "." is no longer known to the user.

swankjesse:
You probably want to fix your server to use RFC 6265 exclusively. RFC 2965 is a dead-end spec that was never widely deployed and isn’t well supported by browsers.

PR #2726, opened 2016-07-14
Rebuild journal to work with an unwritable disk when it is full

boxme:
#746 
Rebuild the journal directly on  `journalFile` when the disk is full and `journalFileTmp` becomes unwritable. 

From my understanding, this will allow `rebuildJournal()` to work when the disk is full because the original content of `journalFile` will be overridden. Doing it on other files such as `journalFileBackup` will not work because it is usually already without content and does not contribute significantly to the occupied disk space.


swankjesse:
Unfortunately we’re in the process of rewriting this entire class. I don’t think this is a good time to make changes to DiskLruCache, particularly changes that are potentially risky.

boxme:
@swankjesse Ok! Then I shall close this then. I should have asked for more details before embarking on a bug fix

PR #2727, opened 2016-07-15
Huffman encode string literals to save bytes in HPACK encoder.

dave-r12:
This change was inspired by my investigation of the Chrome HPACK Encoder: https://chromium.googlesource.com/chromium/src/net/+/master/spdy/hpack/hpack_encoder.cc#145

In my testing locally, both Chrome and Firefox use Huffman when it saves bytes.

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/http2/Hpack.java
+++ b/okhttp/src/main/java/okhttp3/internal/http2/Hpack.java
@@ -370,6 +370,7 @@ final class Hpack {
     private static final int SETTINGS_HEADER_TABLE_SIZE_LIMIT = 16384;
 
     private final Buffer out;
+    private final boolean useCompression;
 
     /**
      * In the scenario where the dynamic table size changes multiple times between transmission of
@@ -389,12 +390,13 @@ final class Hpack {
     int dynamicTableByteCount = 0;
 
     Writer(Buffer out) {
-      this(SETTINGS_HEADER_TABLE_SIZE, out);
+      this(SETTINGS_HEADER_TABLE_SIZE, true, out);
     }
 
-    Writer(int headerTableSizeSetting, Buffer out) {
+    Writer(int headerTableSizeSetting, boolean useCompression, Buffer out) {
       this.headerTableSizeSetting = headerTableSizeSetting;
       this.maxDynamicTableByteCount = headerTableSizeSetting;
+      this.useCompression = useCompression;
       this.out = out;
     }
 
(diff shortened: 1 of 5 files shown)


okhttp/src/main/java/okhttp3/internal/http2/Hpack.java
```diff
@@ -370,6 +370,7 @@ ByteString readByteString() throws IOException {
     private static final int SETTINGS_HEADER_TABLE_SIZE_LIMIT = 16384;
 
     private final Buffer out;
+    private final boolean useCompression;
```
dave-r12: Turning on Huffman encoding by default broke a number of tests because most of the string literals can be encoded, so I pulled it out into a boolean. It wasn't clear to me if there was any value in updating all those tests to now use the Huffman encoding, rather than the raw literals.

swankjesse: Yep. I can see it both ways. No strong feelings.

okhttp/src/main/java/okhttp3/internal/http2/Hpack.java
```diff
@@ -510,8 +512,16 @@ void writeInt(int value, int prefixMask, int bits) {
     }
 
     void writeByteString(ByteString data) throws IOException {
-      writeInt(data.size(), PREFIX_7_BITS, 0);
-      out.write(data);
+      if (useCompression && Huffman.get().encodedLength(data) < data.size()) {
```
swankjesse: I’m curious about the runtime cost of calling `encodedLength()` and then `encode()`. . .

It might be more efficient in aggregate to always do a full encode and to just measure the buffer’s length. That’s going to be faster in the fast, common case when Huffman is useful, right?

swankjesse: though I haven’t looked to see how fast `encodedLength()` is vs. `encode()`.

okhttp/src/main/java/okhttp3/internal/http2/Hpack.java
```diff
@@ -510,8 +512,16 @@ void writeInt(int value, int prefixMask, int bits) {
...
+        Buffer huffmanBuffer = new Buffer();
```
swankjesse: If ever we want to really micro-optimize this, we can probably prevent this allocation by reusing the buffer. Not yet!

okhttp/src/main/java/okhttp3/internal/http2/Huffman.java
```diff
@@ -102,22 +103,22 @@ void encode(byte[] data, OutputStream out) throws IOException {
...
     if (n > 0) {
       current <<= (8 - n);
       current |= (0xFF >>> n);
-      out.write((int) current);
+      sink.writeByte((int) current);
     }
   }
 
-  int encodedLength(byte[] bytes) {
+  int encodedLength(ByteString bytes) {
```
swankjesse: ahh, this does look really fast

Issue #2191, opened 2016-01-04
Feature Request: set deadline to okio


danvinokour:
Hi,

There is no elegant way to set timeout to the whole process of the request.
ReadTimeout affects only until data is ready to be read and is passed to okio as timeout which as i understand effects is timeout between each byte being read.
It would be great if i could set a timeout to the whole request process which probably means that okio's deadline is  a good parameter to play with.

ericcj:
this would be great

swankjesse:
Duplicates https://github.com/square/okhttp/issues/2840

PR #2729, opened 2016-07-16
Okio deadline for HTTP and HTTP2 stream

boxme:
#2191 

Set a deadline timeout for each HTTP2 stream. DeadlineTimeout will start when the request header is written and will stop the moment `Http2Stream#receiveFin()` is triggered. 

This should allow the entire process of the request to be kept within the allocated deadline time.

Edit:
A timeout using the expected deadline duration is created in `Http1Codec` which will ensure that the time duration at which a request is made to getting a response is within the allocated deadline time

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/OkHttpClientTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/OkHttpClientTest.java
@@ -47,6 +47,7 @@ public final class OkHttpClientTest {
     assertEquals(10_000, client.connectTimeoutMillis());
     assertEquals(10_000, client.readTimeoutMillis());
     assertEquals(10_000, client.writeTimeoutMillis());
+    assertEquals(10_000, client.requestDeadlineMillis());
   }
 
   @Test public void timeoutValidRange() {
@@ -64,6 +65,10 @@ public final class OkHttpClientTest {
     } catch (IllegalArgumentException ignored) {
     }
     try {
+      builder.requestDeadline(1, TimeUnit.NANOSECONDS);
+    } catch (IllegalArgumentException ignored) {
+    }
+    try {
       builder.connectTimeout(365, TimeUnit.DAYS);
     } catch (IllegalArgumentException ignored) {
     }
@@ -73,6 +78,10 @@ public final class OkHttpClientTest {
     }
     try {
       builder.readTimeout(365, TimeUnit.DAYS);
+    } catch (IllegalArgumentException ignored) {
+    }
+    try {
+      builder.requestDeadline(365, TimeUnit.DAYS);
     } catch (IllegalArgumentException ignored) {
     }
   }
--- a/okhttp-tests/src/test/java/okhttp3/OkHttpClientTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/OkHttpClientTest.java
@@ -54,34 +54,42 @@ public final class OkHttpClientTest {
     OkHttpClient.Builder builder = new OkHttpClient.Builder();
     try {
       builder.connectTimeout(1, TimeUnit.NANOSECONDS);
+      fail("Connect timeout value is too small and check is not working");
     } catch (IllegalArgumentException ignored) {
     }
     try {
       builder.writeTimeout(1, TimeUnit.NANOSECONDS);
+      fail("Write timeout value is too small and check is not working");
     } catch (IllegalArgumentException ignored) {
     }
     try {
       builder.readTimeout(1, TimeUnit.NANOSECONDS);
+      fail("Read timeout value is too small and check is not working");
     } catch (IllegalArgumentException ignored) {
     }
     try {
       builder.requestDeadline(1, TimeUnit.NANOSECONDS);
+      fail("Request deadline value is too small and check is not working");
     } catch (IllegalArgumentException ignored) {
     }
     try {
       builder.connectTimeout(365, TimeUnit.DAYS);
+      fail("Connect timeout value is too large and check is not working");
     } catch (IllegalArgumentException ignored) {
     }
     try {
       builder.writeTimeout(365, TimeUnit.DAYS);
+      fail("Write timeout value is too large and check is not working");
     } catch (IllegalArgumentException ignored) {
     }
     try {
       builder.readTimeout(365, TimeUnit.DAYS);
+      fail("Read timeout value is too large and check is not working");
     } catch (IllegalArgumentException ignored) {
     }
     try {
       builder.requestDeadline(365, TimeUnit.DAYS);
+      fail("Request deadline value is too large and check is not working");
     } catch (IllegalArgumentException ignored) {
     }
   }
--- a/okhttp/src/main/java/okhttp3/OkHttpClient.java
+++ b/okhttp/src/main/java/okhttp3/OkHttpClient.java
@@ -199,6 +199,7 @@ public class OkHttpClient implements Cloneable, Call.Factory {
   final int connectTimeout;
   final int readTimeout;
   final int writeTimeout;
+  final int requestDeadline;
 
   public OkHttpClient() {
     this(new Builder());
@@ -244,6 +245,7 @@ public class OkHttpClient implements Cloneable, Call.Factory {
     this.connectTimeout = builder.connectTimeout;
     this.readTimeout = builder.readTimeout;
     this.writeTimeout = builder.writeTimeout;
+    this.requestDeadline = builder.requestDeadline;
   }
 
   private X509TrustManager systemDefaultTrustManager() {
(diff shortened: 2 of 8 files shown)


okhttp-tests/src/test/java/okhttp3/OkHttpClientTest.java
```diff
@@ -64,6 +65,10 @@
     } catch (IllegalArgumentException ignored) {
     }
     try {
+      builder.deadlineTimeout(1, TimeUnit.NANOSECONDS);
```
dave-r12: Should there be some `fail()` calls here?

okhttp/src/main/java/okhttp3/internal/http1/Http1Codec.java
```diff
@@ -18,6 +18,8 @@
 import java.io.EOFException;
 import java.io.IOException;
 import java.net.ProtocolException;
+import java.net.SocketTimeoutException;
+
```
dave-r12: No spaces between imports. I think IntelliJ adds them by default, you'll want to change that.

okhttp/src/main/java/okhttp3/internal/http1/Http1Codec.java
```diff
@@ -74,12 +77,15 @@
...
   private static final int STATE_READING_RESPONSE_BODY = 5;
   private static final int STATE_CLOSED = 6;
+  private static final int STATE_CANCEL = 7;
 
   /** The client that configures this stream. May be null for HTTPS proxy tunnels. */
   private final OkHttpClient client;
   /** The stream allocation that owns this stream. May be null for HTTPS proxy tunnels. */
   private final StreamAllocation streamAllocation;
 
+  private final Http1ConnectionTimeout connectionTimeout;
```
dave-r12: Should this be called `connectionDeadline` or something with deadline in it? I think there is a difference between a timeout and a deadline.

okhttp/src/main/java/okhttp3/internal/http1/Http1Codec.java
```diff
@@ -491,4 +515,19 @@ private void readChunkSize() throws IOException {
       closed = true;
     }
   }
+
+
```
dave-r12: Extra newline here.

okhttp/src/main/java/okhttp3/internal/http1/Http1Codec.java
```diff
@@ -491,4 +515,19 @@ private void readChunkSize() throws IOException {
...
+  private class Http1ConnectionTimeout extends AsyncTimeout {
+    @Override protected void timedOut() {
+      state = STATE_CANCEL;
```
dave-r12: I believe this gets called on a separate thread, does `state` need to be made thread safe?

boxme:
@dave-r12 Hi! I have updated the commits based on your reviews. 

Specifically for making `state` in `Http1Codec` thread safe, I am using `volatile` to keep the synchronisation cost low, while ensuring the visibility of its edits. 

Since the `state` does not depend on its current value to edit or is involved in an invariant with other variables, I think this should be valid.

okhttp/src/main/java/okhttp3/OkHttpClient.java
```diff
@@ -512,6 +522,21 @@ public Builder writeTimeout(long timeout, TimeUnit unit) {
     }
 
     /**
+     * Sets the default deadline timeout for new connections. A value of 0 means no timeout,
+     * otherwise values must be between 1 and {@link Integer#MAX_VALUE} when converted
+     * to milliseconds.
+     */
+    public Builder deadlineTimeout(long timeout, TimeUnit unit) {
```
swankjesse: I don’t think this is the right name. In Okio we have specific words for deadlines and timeouts:
http://square.github.io/okio/1.x/okio/okio/Timeout.html

> Timeouts specify the maximum time to wait for a single operation to complete. Timeouts are typically used to detect problems like network partitions. For example, if a remote peer doesn't return any data for ten seconds, we may assume that the peer is unavailable.
> 
> Deadlines specify the maximum time to spend on a job, composed of one or more operations. Use deadlines to set an upper bound on the time invested on a job. For example, a battery-conscious app may limit how much time it spends pre-loading content.

swankjesse: Is this the _call_ timeout?

okhttp/src/main/java/okhttp3/internal/http1/Http1Codec.java
```diff
@@ -74,22 +76,31 @@
...
 
   public Http1Codec(OkHttpClient client, StreamAllocation streamAllocation, BufferedSource source,
       BufferedSink sink) {
     this.client = client;
     this.streamAllocation = streamAllocation;
     this.source = source;
     this.sink = sink;
+    this.connectionDeadline = new Http1ConnectionDeadline();
+    if (client != null) {
+      // Setting to AsyncTimeout#timeout() instead of AsyncTimeout#deadline() because the
```
swankjesse: Yeah, this is the confusion I see here.

okhttp/src/main/java/okhttp3/internal/http1/Http1Codec.java
```diff
@@ -491,4 +516,18 @@ private void readChunkSize() throws IOException {
       closed = true;
     }
   }
+
+  private class Http1ConnectionDeadline extends AsyncTimeout {
+    @Override protected void timedOut() {
+      state = STATE_CANCEL;
```
swankjesse: I don’t think this is safe concurrency wise. Instead we want to manually break the connection by calling cancel().

boxme:
@swankjesse 
To avoid confusion, I remove the keyword `timeout` from the request's deadline. Instead, I renamed it to `requestDeadline`. For example, `OkHttpClient#requestDeadlineMillis()`

In addition, I created an abstract class `HttpRequestDeadline` and have deadlines for both HTTP1 and HTTP2 extends from it to ensure that the separation is clear from `StreamTimeout`

Lastly, I have `Http1RequestDeadline#timeOut()` calls cancel() to close the socket, which at the same time will close both the sink and source. Any further attempt to read/write from them will throw an IOException. I am not entirely sure if this will be safe concurrency wise, hence I hope to get your thoughts on this.

Thanks!

swankjesse:
I’m still not sure about how to expose this as an API. Here’s an alternative:

```
public interface Call {
  ...
  Timeout timeout();
```

This feels more like how Okio typically works, which I like. We could also offer the default timeout on OkHttpClient.Builder:

```
class OkHttpClient {
  ...
  class Builder {
    public Builder connectTimeout(long timeout, TimeUnit unit) { ... }
    public Builder readTimeout(long timeout, TimeUnit unit) { ... }
    public Builder writeTimeout(long timeout, TimeUnit unit) { ... }
    public Builder callTimeout(long timeout, TimeUnit unit) { ... }
  }
}
```

I’m curious – what’s your use case here? Just trying to fix a bug from our queue or do you actually need this fix?

JakeWharton:
FWIW I really like the exposure in Call and would mirror that in Retrofit

boxme:
>I’m curious – what’s your use case here? Just trying to fix a bug from our queue or do you actually need this fix?

A little bit of both. Was previously trying to set an absolute timeout for the entire call through the `OkHttpClient#Builder`. Also trying to fix stuff up from the queue to contribute to a library I use extensively at work.

I chose the latter, albeit with a different name ` Builder#requestDeadline()`. Reason being that this is currently where all the timeouts are being set.

Alternatively, as you have suggested, `Call#timeout()` will allow each specific call request to have their own deadline, which I like for its flexibility. If this is chosen, then I am thinking of having `CallServerInterceptor` to take in the timeout and then set it into `HttpCodec`.

What do you think?
> I’m still not sure about how to expose this as an API. Here’s an alternative:
```
public interface Call {
  Timeout timeout();
```
This feels more like how Okio typically works, which I like. We could also offer the default timeout on OkHttpClient.Builder:
```
class OkHttpClient {
  class Builder {
    public Builder connectTimeout(long timeout, TimeUnit unit) { ... }
    public Builder readTimeout(long timeout, TimeUnit unit) { ... }
    public Builder writeTimeout(long timeout, TimeUnit unit) { ... }
    public Builder callTimeout(long timeout, TimeUnit unit) { ... }
  }
}
```

swankjesse:
At the moment I’m not completely convinced that this is a real problem that needs to be fixed. If an application is seeing bytes make forward progress, but slowly, do they ever want the client to kill the request?

boxme:
@swankjesse A quick check around me also shows that most ppl are not expecting the client to kill the request when the bytes are seen making forward progress. This is indeed a good to have and not must have. 

Will go on to close this away. Thanks!
> At the moment I’m not completely convinced that this is a real problem that needs to be fixed. If an application is seeing bytes make forward progress, but slowly, do they ever want the client to kill the request?

PR #2731, opened 2016-07-19
Allow users to provide a custom FileSystem to Cache

awans:
Hey y'all,

We wrote an EncryptedSource/EncryptedSink to use in our request cache, and wrote an EncryptedFilesystem that wraps FileSystem.SYSTEM in that layer of encryption.

Let me know if you think there's a better way to encrypt our Request Cache instead!

Andrew

### DIFF
--- a/okhttp/src/main/java/okhttp3/Cache.java
+++ b/okhttp/src/main/java/okhttp3/Cache.java
@@ -178,7 +178,7 @@ public final class Cache implements Closeable, Flushable {
     this(directory, maxSize, FileSystem.SYSTEM);
   }
 
-  Cache(File directory, long maxSize, FileSystem fileSystem) {
+  public Cache(File directory, long maxSize, FileSystem fileSystem) {
     this.cache = DiskLruCache.create(fileSystem, directory, VERSION, ENTRY_COUNT, maxSize);
   }
 


JakeWharton:
The only potentially bad thing here is that we're exposing an internal type in a non-internal public API.

swankjesse:
This is really cool. We’re about to start building DiskLruCache2, which will hopefully support the same concept. But it’ll likely require us to grow the FileSystem API – it’ll need some random access to file contents.

Would it make sense to offer encryption as a built-in feature of the cache? Rather than exposing the complete FileSystem abstraction?

awans:
If the suggestion is some sort of interceptor-style interface to inject custom encryption code, or some clever way to provide a wrapping source/sink, then that sounds great.

If the suggestion is that DiskLruCache has encryption and key management baked into it, I'd be worried about the key management part of that—I expect many apps will have idiosyncratic requirements around key storage, which unfortunately imposes idiosyncratic requirements on encryption. For example, the Android Keystore API prior to M doesn't support generating or storing symmetric keys..

In regards to random access, the pattern I've seen for encrypted files is to split it into blocks and encrypt each block separately. That imposes the requirement that all reads and writes know about their position within the file, or at least can seek to find the beginning of a block by looking for some sort of sigil.

swankjesse:
Got it. So the ideal API will let you transform a source and a sink for persistence. Something like this:

```
interface Codec {
  Sink encode(Sink s);
  Source decode(Source s);
}
```

awans:
That SGTM.

Issue #2464, opened 2016-04-06
[bug] 407 proxy error is not reported when credentials are wrong, instead "Too many follow-up requests"


smatei:
Hello,

When connecting through a proxy with authentication, the first request is done without authentication. 
(I have implemented a preemptive authentication here https://github.com/square/okhttp/pull/2458 in order to send the first request with authentication). 407 response is received, the authenticate method from authenticator is called. If the **user/password credentials are wrong**, we get 407 again, and we make the call again with wrong credentials, and again, and again.

`Exception in thread "main" java.net.ProtocolException: Too many follow-up requests: 21
    at okhttp3.RealCall.getResponse(RealCall.java:289)
    at okhttp3.RealCall$ApplicationInterceptorChain.proceed(RealCall.java:198)
    at okhttp3.RealCall.getResponseWithInterceptorChain(RealCall.java:160)
    at okhttp3.RealCall.execute(RealCall.java:57)
    at testok.run(testok.java:72)
    at testok.main(testok.java:79)
`

If 407 is received and we did send "authentication" header in the request, the request should return 407 without so many useless redirects. 

For the first call (without header) 407 means "try to authenticate". For the second call (with header) 407 should mean "failed to authenticate". **And exit.**

Best Regards,
Stefan Matei

swankjesse:
Our expectation is that your Authenticator will check `Response.priorResponse()`, but we could do a lot to make this more obvious.

smatei:
Hi @swankjesse,

Do you mean like this?


`

        case HTTP_PROXY_AUTH:
        Proxy selectedProxy = route != null
            ? route.proxy()
            : client.proxy();
        if (selectedProxy.type() != Proxy.Type.HTTP) {
          throw new ProtocolException("Received HTTP_PROXY_AUTH (407) code while not using proxy");
        }       

        if (userResponse.priorResponse() == null)
        {
        	return client.proxyAuthenticator().authenticate(route, userResponse);
        }
        else if (userResponse.priorResponse().code() == 407)
        {        	
        	return null;
        }

`

This breaks the retry loop and returns 407 at the second retry.

smatei:
Hi @swankjesse,

I have found a simple solution. Change signature of followUpRequest  so that it contains the request also. If HTTP_PROXY_AUTH and the request already contained 'Proxy-Authorization' header, the authentication is useless and we can return.

`


    private Request followUpRequest(Request request, Response userResponse)
    .....
    switch (responseCode) {
[...]

PR #2732, opened 2016-07-19
Fix for bug 2464 - "Too many follow-up requests" on 407 or 401 when wrong credentials

smatei:
Fix for bug https://github.com/square/okhttp/issues/2464 - "Too many follow-up requests" on 407 or 401 when wrong credentials. If we find in request Authorization or Proxy-Authorization headers, stop the authenticate loop.


swankjesse:
Unfortunately this is backwards incompatible. The standard workaround is to do this in your `Authenticator` class.

PR #2741, opened 2016-07-22
Remove some dead code pointed out by running coverage.

swankjesse:
AOSP doesn't use SystemPropertiesConnectionPool so it's useless. It also doesn't
need to configure protocols through request properties.


JakeWharton:
LGTM. Looks like Travis CI updated to JDK 1.8.0_101 so we need new ALPN jars.

Issue #2742, opened 2016-07-22
Don’t leak MockWebServer ports across tests


swankjesse:
https://github.com/google/j2objc/commit/1158bbd903df87802c9bfe21655ae47e053b3e57

swankjesse:
We’re already doing this in MockWebServer#shutdown.

PR #2743, opened 2016-07-23
Check for MockWebServer shutdown result across the tests

boxme:
#2742 

Check that the MockWebServer is shutdown successfully across the tests to prevent port leakage

### DIFF
--- a/mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.java
+++ b/mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.java
@@ -159,10 +159,15 @@ public final class MockWebServer implements TestRule {
         try {
           base.evaluate();
         } finally {
+          boolean shutDownSuccess = true;
           try {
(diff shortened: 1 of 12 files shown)


swankjesse:
What is the behavior change this PR attempts to make? As far as I can tell it throws an AssertionError if the executor fails to return after 5 minutes in both cases.

boxme:
This is my bad. I thought the link reference on the #2742 indicated that you wish to stop the test when the executor fails to return after 5 seconds. I completely missed out on the fact that AssertError was already thrown from the original code.

Will proceed to close this.

swankjesse:
Unfortunately our issues list is missing a lot of context on what each issue specifically wants. Some of the issues say “do X” but they really mean, “do X and then see what the consequences are and then decide whether we actually want X”. This is bad news for contributors like you ’cause you don’t have that context. Best advice is to avoid spending too much time on any given issue without checking that you’re heading in the right direction.

Issue #2682, opened 2016-06-29
DiskLruCache2


swankjesse:
### Goals
- Read while writing.
- Avoid I/O in synchronized blocks.
- Only one file per entry.
### API sketch

```
class DiskLruCache2 {
  Entry get(String key);
  Entry put(String key, ByteString metadata, Source upstream);
  // ... plus methods for sizing, trimming, iterating, etc.

  static class Entry {
    ByteString metadata();
    Source newSource();
  }
}
```
### Implementation sketch

Each entry that is being written has these sources and sinks:
- The upstream source passed into `put()`. This has a logical cursor starting at 0, representing the next byte to read.
- The file sink. Data read from upstream is written to the file sink. Its cursor starts at 0 and is incremented when a flush completes successfully. Its cursor represents that last byte that is safe to read from the file.
- The sources created by calls to `newSource()`. On each attempt to read it will read from the file (if its cursor is lower than the file cursor) or from the network otherwise. If it reads from the network it is responsible for writing to the file.

Readers cooperate with synchronization, with only one reading from upstream at a time. We should also add a buffer to contain up to _N_ bytes preceding the upstream source’s cursor. Readers may prefer to read from this buffer instead of from the file for better performance. Also this makes it possible to reduce flushes to the file system.
### Notes
- What happens when a FileInputStream reads the last byte of a file? Does it eagerly EOF? Or does it continue to read bytes if further bytes are later written to the file?
- The current DiskLruCache has `CLEAN` and `DIRTY` entries in its journal. We should do likewise. An entry can be made clean after the upstream source on a put reaches EOF. This requires a last flush to the file.
- We want to be able to `put()` a new entry when an existing entry is already in place. This is necessary when a conditional GET misses.
- We may also want to be able to `put()` new metadata _without_ updating the rest of the cache. This might motivate putting the metadata ByteString at the end of the file.

digitalbuddha:
Since 2 only comes once, I wanted to chime in with  high level  requests that would help a data loading library we are currently working on. 

I'd like to be able to construct a `DataStore<Request,Response>`  The store would have the following interface:
```
Response get(request)//return from memory if present else disk if present else network
Response fresh(request)//return from network
notify()//listen for all responses (not just 1 key)
```

I'm not sure if it makes sense to encapsulate the memory cache inside of OKHTTP since you would want to cache the inflated models for speed of retrieval. This can't be done without passing a parser into the client

[...]

PR #2744, opened 2016-07-24
FileOperator, a utility for doing random access I/O with Okio.

swankjesse:
This is just the basics that I need for DiskLruCache2. It's not a
particularly general purpose API and at the moment there are no
plans to expose it as such.

https://github.com/square/okhttp/issues/2682

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/internal/cache2/FileOperatorTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/internal/cache2/FileOperatorTest.java
@@ -1,0 +1,200 @@
+/*
+ * Copyright (C) 2016 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okhttp3.internal.cache2;
+
+import java.io.File;
+import java.io.IOException;
+import java.io.RandomAccessFile;
+import java.util.Random;
+import okio.Buffer;
+import okio.BufferedSink;
+import okio.BufferedSource;
+import okio.ByteString;
+import okio.Okio;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
+
+public final class FileOperatorTest {
+  @Rule public final TemporaryFolder tempDir = new TemporaryFolder();
+
+  private File file;
+  private RandomAccessFile randomAccessFile;
+
+  @Before public void setUp() throws Exception {
+    file = tempDir.newFile();
+    randomAccessFile = new RandomAccessFile(file, "rw");
+  }
+
+  @After public void tearDown() throws Exception {
+    randomAccessFile.close();
+  }
+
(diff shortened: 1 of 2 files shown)


okhttp/src/main/java/okhttp3/internal/cache2/FileOperator.java
```diff
@@ -0,0 +1,98 @@
...
+  public void read(long pos, Buffer sink, long byteCount) throws IOException {
+    if (byteCount < 0) throw new IndexOutOfBoundsException();
+
+    while (byteCount > 0L) {
+      try {
+        // Read up to byteCount bytes.
+        byteBuffer.limit((int) Math.min(BUFFER_SIZE, byteCount));
+        fileChannel.read(byteBuffer, pos);
+        int bytesRead = byteBuffer.position();
+        if (bytesRead == -1) throw new EOFException();
```
dave-r12: Am wondering, should bytesRead be the value returned from `fileChannel.read(..)`? The invariant on Buffer says 0 <= position.

swankjesse: My forthcoming calling code always knows how many bytes are available, so it really wants a complete read.

Channels and input streams and sources all return as soon as they can because the next call could block.

dave-r12: Ok, so it sounds like the condition will never happen with the calling code. My reading of the `java.nio.Buffer#position()` suggests the value will always be >= 0, but if the calling code will never hit EOF I guess it doesn't matter.

dave-r12:
Cool, looking forward to DiskLruCache2 take form.

okhttp/src/main/java/okhttp3/internal/cache2/FileOperator.java
swankjesse: Ahhh, I see what you’re saying. Yep, this is my bad.

swankjesse: Both `fileChannel.read(byteBuffer, pos);` and `byteBuffer.position();` should have the same value, _unless_ it’s -1. This got oversimplified away while I was working.

swankjesse: Fixed.

okhttp/src/main/java/okhttp3/internal/cache2/FileOperator.java
```diff
@@ -0,0 +1,96 @@
...
+  private final byte[] byteArray = new byte[BUFFER_SIZE];
+  private final ByteBuffer byteBuffer = ByteBuffer.wrap(byteArray);
+  private final FileChannel fileChannel;
+
+  public FileOperator(FileChannel fileChannel) {
+    this.fileChannel = fileChannel;
+  }
+
+  /** Write {@code byteCount} bytes from {@code source} to the file at {@code pos}. */
+  public void write(long pos, Buffer source, long byteCount) throws IOException {
```
JakeWharton: Surprised to see `Buffer` here. Would have expected
```java
public BufferedSink write(long pos)
```

okhttp-tests/src/test/java/okhttp3/internal/cache2/FileOperatorTest.java
```diff
@@ -0,0 +1,200 @@
...
+    operator.read(62, buffer, 6);
+    operator.read(4, buffer, 19);
+    operator.write(80, buffer, buffer.size());
+
+    assertEquals(snapshot(), ByteString.encodeUtf8(""
+        + "god creates dinosaurs. "
+        + "god destroys dinosaurs. "
+        + "god creates man. "
+        + "man destroys god. "
+        + "man creates dinosaurs. "));
```
JakeWharton: amazing

JakeWharton:
lgtm

okhttp/src/main/java/okhttp3/internal/cache2/FileOperator.java
swankjesse: Yeah, or even just `Sink` that fits more neatly. We should definitely consider that if we ever build a public API for this.

PR #2750, opened 2016-07-25
Relay. This is a single entry in the upcoming DiskLruCache2.

swankjesse:
At the moment there are some things still incomplete:
 - Timeouts. Not yet tested, and no great implementation plan.
 - Thread interrupts. This has been quite problematic for
   the original DiskLruCache. We should test and support this
   in Relay.
 - File system problems. If the host file system is full or not
   accepting writes, we should fallback gracefully. This may
   be difficult.
 - Metadata updates. At the moment the metadata is permanent
   and immutable.

That said, many nice features are already implemented:
 - Metadata and stream data in one file.
 - No locks held during file I/O or network I/O.
 - Read from the buffer, the file, or the socket in that order.
 - Reference counting on open sources. It is assumed that the
   creator of each Relay will take a source, and that guarantees
   that the count will always reach 0.

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/cache2/Relay.java
+++ b/okhttp/src/main/java/okhttp3/internal/cache2/Relay.java
@@ -1,0 +1,361 @@
+/*
+ * Copyright (C) 2016 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okhttp3.internal.cache2;
+
+import java.io.File;
+import java.io.IOException;
+import java.io.RandomAccessFile;
+import okio.Buffer;
+import okio.ByteString;
+import okio.Source;
+import okio.Timeout;
+
+import static okhttp3.internal.Util.closeQuietly;
+
+/**
+ * Replicates a single upstream source into multiple downstream sources. Each downstream source
+ * returns the same bytes as the upstream source. Downstream sources may read data either as it
+ * is returned by upstream, or after the upstream source has been exhausted.
+ *
+ * <p>As bytes are returned from upstream they are written to a local file. Downstream sources read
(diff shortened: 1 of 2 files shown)


okhttp/src/main/java/okhttp3/internal/cache2/Relay.java
```diff
@@ -0,0 +1,360 @@
...
+
+        // The buffer has the data we need. Read from there and return immediately.
+        long bytesToRead = Math.min(byteCount, upstreamPos - sourcePos);
+        buffer.copyTo(sink, sourcePos - bufferPos, bytesToRead);
+        sourcePos += bytesToRead;
+        return bytesToRead;
+      }
+
+      // Read from the file.
+      if (source == SOURCE_FILE) {
```
dave-r12: Nice

okhttp/src/main/java/okhttp3/internal/cache2/Relay.java
```diff
@@ -0,0 +1,360 @@
...
+
+    /** The next byte to read. This is always less than or equal to {@code upstreamPos}. */
+    private long sourcePos;
+
+    /**
+     * Selects where to find the bytes for a read and read them. This is one of three sources.
+     *
+     * <h3>Upstream:</h3>
+     * In this case the current thread is assigned as the upstream reader. We read bytes from
+     * upstream and copy them to both the file and to the buffer. Finally we release the release
```
dave-r12: Minor detail, 

> Finally we release the release

dave-r12:
I'm still absorbing some bits and pieces but the logic is straightforward given the problem this is solving. Good stuff.

swankjesse:
After a tiny amount of soul searching I decided what to do about timeouts. I’ll do the work in a follow up PR.

Whenever a thread reads from upstream it’ll first configure upstream’s timeouts from its own timeouts. So if a thread has a 3 second timeout it’ll use that to read upstream, even if other waiting threads have longer or shorter timeouts.

I figure this is pretty reasonable because in practice all of the threads in the same HTTP client reading the same resource will inevitably have the same timeouts anyway.

dave-r12:
Any reason to hold this one up? LGTM.

swankjesse:
I pitched this design to Jake before I wrote the code so I’m hoping to get his eyes on it too. He’s been tied up but should be back soon.

okhttp/src/main/java/okhttp3/internal/cache2/Relay.java
swankjesse: Oooh fixed.

JakeWharton:
lgtm. this is dense. had to give it two passes.

swankjesse:
Yeah! More exciting stuff to come. Thanks @dave-r12 and @jakewharton for the reviews!

Issue #1838, opened 2015-09-04
Websocket creates threads with url as name, which captures tokens in urls in crashlytics


osama-lionheart:
We use okhttp-ws to communicate with our server, and as an optimization step, we pass an `oauth2` token in the websocket url to the server (to avoid making an extra auth request every time we connect), and since we use crashlytics as a reporting tool for our exceptions, we noticed that oauth2 tokens are captured in the url as the name of thread that the websocket creates.

swankjesse:
Yup, we should partially redact the URL before using it to name a thread. We already have some code like this in `Call`. If you'd like to send a pull request, please do!

JakeWharton:
Unless I'm missing something, it looks like `Call` does no redaction of the URL. It just calls `toString()` on the `HttpUrl` from the original request.

iNoles:
Looks like Jake is right. https://github.com/square/okhttp/blob/master/okhttp-ws/src/main/java/okhttp3/ws/WebSocketCall.java#L160

JakeWharton:
I was referring to the non-websocket `Call`: https://github.com/square/okhttp/blob/9920c57e7a8c0d49ec2732be899b3e27c215854a/okhttp/src/main/java/okhttp3/RealCall.java#L99

boxme:
I made a simple pull request to add a method to redact the url in Http, which is then used in WebSocketCall to hide the captured tokens

JakeWharton:
Fixed by https://github.com/square/okhttp/pull/2751

PR #2751, opened 2016-07-26
Method redact the url in HttpUrl

boxme:
#1838 

Add HttpUrl#redactedUrl() to display only the domain name of the url

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/HttpUrlTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/HttpUrlTest.java
@@ -29,6 +29,7 @@ import org.junit.Test;
 
 import static java.util.Collections.singletonList;
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.fail;
 
@@ -122,6 +123,15 @@ public final class HttpUrlTest {
     assertEquals(null, base.newBuilder("ht+tp://b"));
     assertEquals(null, base.newBuilder("ht-tp://b"));
     assertEquals(null, base.newBuilder("ht.tp://b"));
+  }
+
+  @Test public void redactedUrl() {
+    HttpUrl baseWithPasswordAndUsername = HttpUrl.parse("http://username:password@host/a/b#fragment");
+    HttpUrl baseWithUsernameOnly = HttpUrl.parse("http://username@host/a/b#fragment");
+    HttpUrl baseWithPasswordOnly = HttpUrl.parse("http://password@host/a/b#fragment");
+    assertEquals(HttpUrl.parse("http://host/..."), baseWithPasswordAndUsername.redact());
+    assertEquals(HttpUrl.parse("http://host/..."), baseWithUsernameOnly.redact());
+    assertEquals(HttpUrl.parse("http://host/..."), baseWithPasswordOnly.redact());
   }
 
   @Test public void resolveNoScheme() throws Exception {
--- a/okhttp/src/main/java/okhttp3/HttpUrl.java
+++ b/okhttp/src/main/java/okhttp3/HttpUrl.java
@@ -842,6 +842,17 @@ public final class HttpUrl {
   }
 
   /**
+   * Returns the HttpUrl with the username, password, path, query, and fragment stripped.
+   * Example: http://username:password@example.com/path returns http://example.com/...
+   */
+  public HttpUrl redact() {
+    Builder builder = newBuilder("/...");
+    builder.username("");
+    builder.password("");
+    return builder.build();
+  }
+
+  /**
    * Returns the URL that would be retrieved by following {@code link} from this URL, or null if
    * the resulting URL is not well-formed.
    */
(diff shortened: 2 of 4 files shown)


okhttp-tests/src/test/java/okhttp3/HttpUrlTest.java
```diff
@@ -124,6 +122,12 @@
     assertEquals(null, base.newBuilder("ht.tp://b"));
   }
 
+  @Test public void redactedUrl() {
+    HttpUrl base = HttpUrl.parse("http://host/a/b#fragment");
```
JakeWharton: Add username and password to URL

okhttp-tests/src/test/java/okhttp3/HttpUrlTest.java
```diff
@@ -28,9 +28,7 @@
 import org.junit.Test;
 
 import static java.util.Collections.singletonList;
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNull;
-import static org.junit.Assert.fail;
+import static org.junit.Assert.*;
```
JakeWharton: Revert. We don't allow star imports.

okhttp/src/main/java/okhttp3/HttpUrl.java
```diff
@@ -842,6 +842,13 @@ public String fragment() {
   }
 
   /**
+   * Returns the redacted URL that would only show the domain name.
+   */
+  public HttpUrl redactedUrl() {
```
JakeWharton: Not sure this has any value being a `HttpUrl` since you can't really do anything with it. I would return `String` and just proactively `toString()` here.

okhttp-tests/src/test/java/okhttp3/HttpUrlTest.java
boxme: My bad for missing out on this case. Will remove the username and password from the redacted url if they exist.

okhttp/src/main/java/okhttp3/RealCall.java
```diff
@@ -152,8 +152,8 @@ private String toLoggableString() {
     return string + " to " + redactedUrl();
   }
 
-  HttpUrl redactedUrl() {
-    return originalRequest.url().resolve("/...");
+  String redactedUrl() {
```
NightlyNexus: Remove the now-redundant `toString()` at line 106?

boxme: Fixed! And thanks for catching that!

okhttp/src/main/java/okhttp3/HttpUrl.java
```diff
@@ -842,6 +842,19 @@ public String fragment() {
   }
 
   /**
+   * Returns the redacted URL that would only show the domain name.
+   */
+  public String redactedUrl() {
+    Builder builder = newBuilder("/...");
+    if (builder != null) {
```
swankjesse: why would builder be null?

okhttp/src/main/java/okhttp3/HttpUrl.java
```diff
@@ -842,6 +842,19 @@ public String fragment() {
   }
 
   /**
+   * Returns the redacted URL that would only show the domain name.
+   */
+  public String redactedUrl() {
```
swankjesse: should this be just `redact()` ?

okhttp/src/main/java/okhttp3/HttpUrl.java
```diff
@@ -842,6 +842,19 @@ public String fragment() {
   }
 
   /**
+   * Returns the redacted URL that would only show the domain name.
```
swankjesse: Not sure about this comment. It suggests the `/...` won’t be present . . . maybe say `Returns a URL like http://example.com/... with the username, password, path, query, and fragment stripped` ?

okhttp/src/main/java/okhttp3/HttpUrl.java
swankjesse: Should this return an HttpUrl or a String? could be either

boxme: For this particular use case we are trying to retrieve the redacted Url and therefore I decided to go with return a String and keeping the name `redactedUrl()`

But if we decide to return a HttpUrl instead, then `redact()` will be more appropriate. What do you think?

okhttp/src/main/java/okhttp3/HttpUrl.java
boxme: I realise that builder will never be null because the url will have to be valid in order to create a HttpUrl out of its builder.

Fixed

okhttp/src/main/java/okhttp3/HttpUrl.java
boxme: Modified the comment for more clarity.

swankjesse:
I like where this is going. Sorry for dragging on this, but I’m particular about APIs!

Two last requests:
 - change the method name to `redact()` ?
 - change the return type to `HttpUrl` ?

Do those & we’ll merge. Thanks for your patience!

boxme:
@swankjesse Fixed! 

I totally understand the importance of APIs, so thanks for correcting me to keep up to great standards you set here!

swankjesse:
Thanks.

Issue #2757, opened 2016-07-28
Connections should be shared between OkHttpClients when using shared ConnectionPool


15characterlimi:
What kind of issue is this?

 - [X] Bug report.

## tl;dr
CertificateChainCleaner / CertificatePinner don't override equals() even though Platform.buildCertificateChainCleaner() constructs new (not interned) instances. This breaks connection pooling across separately constructed OkHttpClient instances with shared ConnectionPool.

## Failing tests
   - [Gist with a single failing test (adopted from a part of URLConnectionTest)](https://github.com/15characterlimi/okhttp/commit/119873249d7414964a0ea24fac79081ad7c7b7d9)
   - [Draft commit that adds two failing tests](https://github.com/15characterlimi/okhttp/commit/119873249d7414964a0ea24fac79081ad7c7b7d9)

## Detail

 - A few places (namely OkHttpClient(Builder), and both overloads of 
   OkHttpClient.Builder.sslSocketFactory()s), call
   CertificateChainCleaner.get(X509TrustManager) to initialize a
   OkHttpClient's CertificateChainCleaner, and
 - CertificateChainCleaner.get(X509TrustManager) always returns a new
   CertificateChainCleaner instance, and
 - CertificateChainCleaner.equals() is based on Object identity, and
 - CertificatePinner.withCertificateChainCleaner() returns a new CertificatePinner
   instance if the CertificateChainCleaner instances are different, and
 - Address.equals() depends on equal(this.certificatePinner, that.certificatePinner), and
 - ConnectionPool.get() only reuses a Connection if address.equals(connection.route().address)

The issue should be easy to work around once one finds the root cause, but in case I feel inspired to pull together a fix for you, what's your view on how this should be fixed?

I suggest that CertificateChainCleaner and CertificatePinner should implement value equality (i.e. override Object.equals()). That CertificatePinner.withCertificateChainCleaner() uses != to compare object instances (namely CertificateChainCleaner instances) seems like a major antipattern to me.

Alternatively, CertificateChainCleaner instances could be interned.

Both of those approaches would rely on X509TrustManager.equals() to work in order to figure out when a new CertificateChainCleaner instance is required; but from what see, X509TrustManager seems to be a singleton in practice so its equals() should not be a problem. 

I know the docs suggest using a OkHttpClient singleton instance as the base for all connections as well as additional OkHttpClient.Builders, however it seems like poor behavior from the library to rely on this in order for ConnectionPool instances to be shared.

[...]

PR #2760, opened 2016-07-29
Fix Connection sharing across equivalent OkHttpClients

15characterlimi:
Fix Connection sharing across equivalent OkHttpClients

Fixes https://github.com/square/okhttp/issues/2757

The OkHttpClient(Builder) constructor and both overloads of
OkHttpClient.Builder.sslSocketFactory() set a new CertificateChainCleaner
instance on the OkHttpClient.

Before this CL, CertificateChainCleaner, CertificatePinner and
TrustRootIndex used object rather than value identity semantics. This
caused Address.equals() to not recognize when addresses were equivalent,
which broke connection pooling across OkHttpClient instances.

This CL changes CertificateChainClaner, CertificatePinner and
TrustRootIndex to use value equality by overriding equals() and hashCode().

CertificatePinner.pins and BasicTrustRootIndex.subjectToCaCerts's values
have changed from List to Set because their order shouldn't matter.
This CL uses LinkedHashSet, although HashSet should also work.

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/CertificateChainCleanerTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/CertificateChainCleanerTest.java
@@ -21,6 +21,10 @@ import java.security.cert.X509Certificate;
 import java.util.ArrayList;
 import java.util.List;
 import javax.net.ssl.SSLPeerUnverifiedException;
+import javax.net.ssl.SSLSocketFactory;
+import javax.net.ssl.X509TrustManager;
+
+import okhttp3.internal.platform.Platform;
 import okhttp3.internal.tls.CertificateChainCleaner;
 import okhttp3.internal.tls.HeldCertificate;
 import org.junit.Test;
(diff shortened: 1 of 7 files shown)


swankjesse:
I’m looking at this. Does this only happen if you build your OkHttpClient from scratch? Ie. is it a problem if you just use `newBuilder()` ? And if so, is the calling code actually creating new clients from scratch rather than using `newBuilder()` ? That’s also unexpected.

15characterlimi:
I believe one way to cause it is to call .sslSocketFactory() on a OkHttpClient.Builder, because that results in a new CertificateChainCleaner instance which is then != the previous instance at the time the ConnectionPool checks Address.equals().

So if you say:
OkHttpClient clientA = new OkHttpClient.Builder()
            .cache(cache)
             .connectionPool(connectionPool)
             .cookieJar(cookieJar)
             .sslSocketFactory(clientSocketFactory, sslClient.trustManager)
             .hostnameVerifier(hostnameVerifier)
             .build();
OkHttpClient clientB = clientA.newBuilder().sslSocketFactory(clientSocketFactory, sslClient.trustManager).build();

then I believe clientA and clientB will not share connections, even though I haven't specifically checked that case. I've checked that the test which I supplied would fail without my fix, though.

I believe it will be hard to guarantee object identity for equivalent CertificateChainCleaners. That's why my pull request https://github.com/square/okhttp/pull/2760 fixes the issue by overriding equals() and hashCode() for the relevant classes.

swankjesse:
Makes sense!

swankjesse:
Merged manually. Apologies for the delay I took some time off!

PR #2763, opened 2016-07-30
Guard allocation changes by the connection pool lock.

swankjesse:
As suggested here: https://github.com/square/okhttp/pull/2597/files


nfuller:
Sorry for not doing this. Crazy busy.

Issue #2127, opened 2015-12-17
Leaked connection warning could be more actionable


swankjesse:
We shouldn't print the same warning multiple times.
It should include a stacktrace.

swankjesse:
Also maybe `Response.body().closeQuietly()` to make that less of a hassle.

JakeWharton:
We talked about removing `IOException` from `close()` since it's not useful

JakeWharton:
#2129

swankjesse:
Plus, we’re seeing this when I run `Crawler` with the new HttpURLConnection frontend. So I can use it myself!

dave-r12:
I was tracking one of these down recently, would you be ok with capturing a stack trace at the time of execute/enqueue?

swankjesse:
Yeah, I think that's exactly what we should be doing.

PR #2765, opened 2016-08-01
Provide capability to capture the Call stack trace on leaked connections.

dave-r12:
Also plugged some leaks discovered after running the tests with this
change in place.

https://github.com/square/okhttp/issues/2127

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/CallTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/CallTest.java
@@ -45,7 +45,9 @@ import java.util.concurrent.SynchronousQueue;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.logging.Level;
 import java.util.logging.Logger;
+import java.util.logging.SimpleFormatter;
 import javax.net.ServerSocketFactory;
 import javax.net.ssl.SSLHandshakeException;
 import javax.net.ssl.SSLPeerUnverifiedException;
@@ -79,6 +81,7 @@ import org.junit.rules.TestRule;
 import org.junit.rules.Timeout;
 
 import static java.net.CookiePolicy.ACCEPT_ORIGINAL_SERVER;
+import static okhttp3.TestUtil.awaitGarbageCollection;
 import static okhttp3.TestUtil.defaultClient;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
@@ -342,6 +345,7 @@ public final class CallTest {
 
     Response response = client.newCall(request).execute();
     assertEquals(200, response.code());
+    response.body().close();
 
     RecordedRequest recordedRequest1 = server.takeRequest();
     assertEquals("POST", recordedRequest1.getMethod());
@@ -2567,6 +2571,78 @@ public final class CallTest {
         .assertCode(200)
         .assertHeader("abc", "def")
         .assertBody("");
+  }
+
+  @Test public void leakedResponseBodyLogsStackTrace() throws Exception {
+    server.enqueue(new MockResponse()
+        .setBody("This gets leaked."));
+
+    client = new OkHttpClient.Builder()
+        .connectionPool(new ConnectionPool(0, 10, TimeUnit.MILLISECONDS))
+        .build();
+
+    Request request = new Request.Builder()
+        .url(server.url("/"))
+        .build();
+
+    Level original = logger.getLevel();
+    logger.setLevel(Level.FINE);
+    logHandler.setFormatter(new SimpleFormatter());
+    try {
+      client.newCall(request).execute(); // Ignore the response so it gets leaked then GC'd.
+      awaitGarbageCollection();
+
+      String message = logHandler.take();
+      assertTrue(message.contains("WARNING: A connection to " + server.url("/") + " was leaked."
+          + " Did you forget to close a response body?"));
+      assertTrue(message.contains("okhttp3.RealCall.execute("));
+      assertTrue(message.contains("okhttp3.CallTest.leakedResponseBodyLogsStackTrace("));
+    } finally {
+      logger.setLevel(original);
+    }
+  }
+
+  @Test public void asyncLeakedResponseBodyLogsStackTrace() throws Exception {
+    server.enqueue(new MockResponse()
+        .setBody("This gets leaked."));
+
+    client = new OkHttpClient.Builder()
+        .connectionPool(new ConnectionPool(0, 10, TimeUnit.MILLISECONDS))
+        .build();
+
+    Request request = new Request.Builder()
+        .url(server.url("/"))
+        .build();
+
+    Level original = logger.getLevel();
+    logger.setLevel(Level.FINE);
+    logHandler.setFormatter(new SimpleFormatter());
+    try {
+      final CountDownLatch latch = new CountDownLatch(1);
+      client.newCall(request).enqueue(new Callback() {
+        @Override public void onFailure(Call call, IOException e) {
+          fail();
+        }
+
+        @Override public void onResponse(Call call, Response response) throws IOException {
+          // Ignore the response so it gets leaked then GC'd.
+          latch.countDown();
+        }
+      });
+      latch.await();
+      // There's some flakiness when triggering a GC for objects in a separate thread. Adding a
+      // small delay appears to ensure the objects will get GC'd.
+      Thread.sleep(200);
+      awaitGarbageCollection();
+
+      String message = logHandler.take();
+      assertTrue(message.contains("WARNING: A connection to " + server.url("/") + " was leaked."
+          + " Did you forget to close a response body?"));
+      assertTrue(message.contains("okhttp3.RealCall.enqueue("));
+      assertTrue(message.contains("okhttp3.CallTest.asyncLeakedResponseBodyLogsStackTrace("));
+    } finally {
+      logger.setLevel(original);
+    }
   }
 
   private void makeFailingCall() {
--- a/okhttp-tests/src/test/java/okhttp3/TestUtil.java
+++ b/okhttp-tests/src/test/java/okhttp3/TestUtil.java
@@ -43,4 +43,15 @@ public final class TestUtil {
     Arrays.fill(array, c);
     return new String(array);
   }
+
+  /**
+   * See FinalizationTester for discussion on how to best trigger GC in tests.
+   * https://android.googlesource.com/platform/libcore/+/master/support/src/test/java/libcore/
+   * java/lang/ref/FinalizationTester.java
+   */
+  public static void awaitGarbageCollection() throws InterruptedException {
+    Runtime.getRuntime().gc();
+    Thread.sleep(100);
+    System.runFinalization();
+  }
 }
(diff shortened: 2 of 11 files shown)


okhttp/src/main/java/okhttp3/RealCall.java
```diff
@@ -65,6 +66,16 @@ protected RealCall(OkHttpClient client, Request originalRequest) {
     }
   }
 
+  private void captureCallStackTrace() {
+    StackTraceElement[] stackTrace = Thread.currentThread().getStackTrace();
```
JakeWharton: This is going to be expensive for Android to do.

dave-r12: Oh, gotcha. Are there any good ways of getting that information? I can also do some research on this.

swankjesse: Hmmm...

Android’s libcore has CloseGuard:
https://android.googlesource.com/platform/libcore/+/master/dalvik/src/main/java/dalvik/system/CloseGuard.java

Here’s how it’s used:
https://android.googlesource.com/platform/libcore/+/android-4.3_r1/luni/src/main/java/java/io/FileInputStream.java

Is there a reasonable way to detect if we’re in a development build? Maybe we can add this to Platform?
https://stuff.mit.edu/afs/sipb/project/android/docs/reference/android/os/Debug.html#isDebuggerConnected()

JakeWharton: Sadly, not that I'm aware of.

swankjesse: Also this: http://stackoverflow.com/a/6865049/40013

swankjesse: One other option that’s tempting . . . what if we asked `java.util.logging` whether the logger for say, `okhttp3.OkHttpClient` was enabled at `Level.INFO` ?

dave-r12: Checking the log level seems reasonable. Should it be `INFO` or `FINE`? A quick experiment on Android was showing me `INFO` was evaluating to true.

dave-r12:
I altered the approach. It's still a WIP, I was hoping to get feedback.

okhttp/src/main/java/okhttp3/internal/platform/Platform.java
```diff
@@ -133,6 +133,20 @@ public boolean isCleartextTrafficPermitted(String hostname) {
     return true;
   }
 
+  public Object captureAllocationSite() {
+    if (!logger.isLoggable(Level.FINE)) return null;
```
dave-r12: Use the log level for non-Android to detect whether to capture the allocation site.

okhttp/src/main/java/okhttp3/internal/platform/Platform.java
```diff
@@ -133,6 +133,20 @@ public boolean isCleartextTrafficPermitted(String hostname) {
...
+    return new Throwable("response.body().close()");
```
dave-r12: Get rid of `Thread.currentThread().getStackTrace()`. That's probably the worst way of getting that information.

okhttp/src/main/java/okhttp3/internal/platform/AndroidPlatform.java
```diff
@@ -132,6 +135,34 @@ public AndroidPlatform(Class<?> sslParametersClass, OptionalMethod<Socket> setUs
     }
   }
 
+  @Override public Object captureAllocationSite() {
+    if (missingCloseGuard.get()) return null;
+
+    try {
+      Class<?> closeGuardClass = Class.forName("dalvik.system.CloseGuard");
```
dave-r12: On Android, use CloseGuard? A hack, but all the other ways seem like hacks too.

swankjesse: Yeah, sigh. I like this approach ’cause it’s likely to do the right thing in development vs. production. It’s still a bit of a hack, but defensible.

okhttp-tests/src/test/java/okhttp3/CallTest.java
```diff
@@ -2569,6 +2573,76 @@ private RequestBody requestBody(final boolean chunked, final long size, final in
...
+        .url(server.url("/"))
+        .build();
+
+    Level original = logger.getLevel();
+    logger.setLevel(Level.FINE);
+    logHandler.setFormatter(new SimpleFormatter());
+    try {
+      client.newCall(request).execute(); // Ignore the response so it gets leaked then GC'd.
+      awaitGarbageCollection();
+      Thread.sleep(200); // Delay just a bit longer to make sure we catch the leaked response.
```
swankjesse: should this be in `awaitGarbageCollection()` ?

okhttp-tests/src/test/java/okhttp3/CallTest.java
```diff
@@ -2569,6 +2573,76 @@ private RequestBody requestBody(final boolean chunked, final long size, final in
...
+
+        @Override public void onResponse(Call call, Response response) throws IOException {
+          // Ignore the response so it gets leaked then GC'd.
+          latch.countDown();
+        }
+      });
+      latch.await();
+      Thread.sleep(200);
+      awaitGarbageCollection();
+      Thread.sleep(200);// Delay just a bit longer to make sure we catch the leaked response.
```
swankjesse: same

okhttp-tests/src/test/java/okhttp3/TestUtil.java
```diff
@@ -43,4 +43,15 @@ public static String repeat(char c, int count) {
     Arrays.fill(array, c);
     return new String(array);
   }
+
+  /**
+   * See FinalizationTester for discussion on how to best trigger GC in tests.
+   * https://android.googlesource.com/platform/libcore/+/master/support/src/test/java/libcore/
+   * java/lang/ref/FinalizationTester.java
+   */
+  public static void awaitGarbageCollection() throws InterruptedException {
```
swankjesse: Dig.

okhttp/src/main/java/okhttp3/internal/connection/StreamAllocation.java
```diff
@@ -340,4 +343,17 @@ public boolean hasMoreRoutes() {
   @Override public String toString() {
     return address.toString();
   }
+
+  public static final class StreamAllocationReference extends WeakReference<StreamAllocation> {
```
swankjesse: Thumbs up

okhttp/src/main/java/okhttp3/internal/connection/StreamAllocation.java
```diff
@@ -73,6 +73,7 @@
   public final Address address;
   private Route route;
   private final ConnectionPool connectionPool;
+  private final Object callStackTrace;
```
swankjesse: nit: callExecutionSite ?

okhttp/src/main/java/okhttp3/RealCall.java
```diff
@@ -65,6 +66,10 @@ protected RealCall(OkHttpClient client, Request originalRequest) {
     }
   }
 
+  private void captureCallStackTrace() {
```
swankjesse: Lots of different words for the same thing: StackTrace, AllocationSite, ExecutionSite. I like ExecutionSite or StackTrace, but we should be consistent.

okhttp/src/main/java/okhttp3/internal/http/RetryAndFollowUpInterceptor.java
```diff
@@ -164,13 +169,14 @@ public StreamAllocation streamAllocation() {
       }
 
       if (followUp.body() instanceof UnrepeatableRequestBody) {
+        streamAllocation.release();
```
swankjesse: hmmm . . . . I wonder if the right fix here is to close something?

okhttp/src/main/java/okhttp3/internal/platform/Platform.java
```diff
@@ -133,6 +133,20 @@ public boolean isCleartextTrafficPermitted(String hostname) {
...
+  }
+
+  public void reportLeakedAllocation(String message, Object allocationSite) {
+    Throwable throwable = (Throwable) allocationSite;
+    if (allocationSite == null) {
+      message += " To see the where this connection was allocated, set the okhttp3.OkHttpClient "
```
swankjesse: Neat!

swankjesse: Maybe be very explicit? Give a full example?

```
Logger.getLogger(OkHttpClient.class.getName()).setLevel(Level.FINE);
```

swankjesse:
This is a bit more involved than I’d anticipated but it’s all good stuff. Just a few questions above.

okhttp/src/main/java/okhttp3/internal/platform/AndroidPlatform.java
dave-r12: Agree. I opened an issue on AOSP to ask for a way to hook into this. I'm guessing it will get closed, but I figured it wouldn't hurt to ask.

okhttp/src/main/java/okhttp3/RealCall.java
dave-r12: Apologies - I wanted to make sure the approach was ok before I went back and started renaming things. I'll clean this up.

okhttp/src/main/java/okhttp3/internal/http/RetryAndFollowUpInterceptor.java
dave-r12: The response body is getting closed a few lines up. I'll look further into this to make sure.

okhttp/src/main/java/okhttp3/internal/platform/AndroidPlatform.java
JakeWharton: Link so we can star it to watch?

swankjesse:
(and LGTM!)

okhttp-tests/src/test/java/okhttp3/CallTest.java
dave-r12: I removed these.

okhttp/src/main/java/okhttp3/internal/http/RetryAndFollowUpInterceptor.java
dave-r12: I think this is the right move. As I mentioned, a few lines up the response body gets closed. That boils down to a call to StreamAllocation#streamFinished which nulls out the codec. It will not release the connection because StreamAllocation#release has not yet been called. So this call to release will return the connection to the pool. Let me know if you think that's inaccurate.

okhttp/src/main/java/okhttp3/internal/platform/Platform.java
dave-r12: Good idea, done.

dave-r12:
I think this should be in good shape:
- Made naming consistent, I went with stack trace.
- I changed how AndroidPlatform accesses CloseGuard so that it caches the method lookups. 
- Tested this on API 9, 23 and 24 and all seems well.

okhttp/src/main/java/okhttp3/internal/http/RetryAndFollowUpInterceptor.java
swankjesse: SGTM!

swankjesse:
Nice. I like that you got to find a missing release along the way.

PR #2767, opened 2016-08-03
response.body() should never be null

yankeppey:
This PR solves the issue #2264. 
Response.body() should never be null, let's return an empty body instead.


swankjesse:
I don’t think this is the right fix. It’s causing a response body to exist for a websocket.

yankeppey:
@swankjesse this fix was based on this @JakeWharton's comment: https://github.com/square/okhttp/issues/2264#issuecomment-173095613

swankjesse:
I think the right fix is to not look for the body on a websocket response.

JakeWharton:
Sorry I forgot there was a PR for this. We implemented it in https://github.com/square/okhttp/pull/2831 and it will be in 3.5.0. Thanks for taking the time to contribute though!

PR #2776, opened 2016-08-06
Adds utility to create a MockResponse from resource file

dherges:
Hi guys,

noticed that I am writing the following code over and over again in various test cases.

Usually, we have static files in ``src/test/resources`` that we use to create mock responses. Maybe it's helpful to have that utility in the library.

wdyt? better naming? should there be a test for the two methods?

cheers! 🌙

### DIFF
--- a/mockwebserver/src/main/java/okhttp3/mockwebserver/MockResponse.java
+++ b/mockwebserver/src/main/java/okhttp3/mockwebserver/MockResponse.java
@@ -15,6 +15,8 @@
  */
 package okhttp3.mockwebserver;
 
+import java.io.IOException;
+import java.io.InputStream;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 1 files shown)


swankjesse:
I think this is completely reasonable but it expands the scope of what you’re potentially using MockWebServer for. We aren’t ready to consider it a general-purpose webserver, and this starts to address that use case.

Have you considered this?
https://github.com/orhanobut/mockwebserverplus

dherges:
Thanks for the hint. That is an even better solution :)

Feel free to close this PR.

slukes:
Hello,

I just came here to make an almost identical pull request.  Whilst I can see what you mean about MockWebServerPlus already meeting this need - I have today seen in other google code projects code almost identical methods to what is proposed here.

Would you be willing to reconsider?

swankjesse:
We’re considering doing an OkHttpServer possibly in 2017, but it’s a ton of work. Until then we don’t want to open the door to that at all. Even simple things like building responses from Files expand the scope of the project and we aren’t ready for that.

dherges:
@swankjesse when you're starting with OkHttpServer can you get in touch with me? I could be tempted to contribute something to it :)

PR #2789, opened 2016-08-13
Force close websocket

jwendell:


okhttp/src/main/java/okhttp3/internal/connection/StreamAllocation.java
```diff
@@ -201,7 +201,7 @@ private RealConnection findConnection(int connectTimeout, int readTimeout, int w
 
   public void streamFinished(boolean noNewStreams, HttpCodec codec) {
     synchronized (connectionPool) {
-      if (codec == null || codec != this.codec) {
+      if (codec != this.codec) {
```
JakeWharton: This seems completely unrelated to the change. Please submit it as a separate PR.

okhttp-ws/src/main/java/okhttp3/internal/ws/RealWebSocket.java
```diff
@@ -166,6 +169,9 @@ public void sendPong(Buffer payload) throws IOException {
 
     try {
       writer.writeClose(code, reason);
+
+      // Wait graciously 5 seconds for server close response before shutdown
+      this.source.timeout().timeout(5, TimeUnit.SECONDS);
```
JakeWharton: I'm tempted to re-use the connect timeout value here (think: disconnect timeout?). @swankjesse thoughts?

JakeWharton:
Can you include a test for this behavior?

okhttp-ws/src/main/java/okhttp3/internal/ws/RealWebSocket.java
swankjesse: What bad thing happens if we close eagerly?

JakeWharton: Probably nothing terrible, although I wouldn't be surprised if someone eventually complained.

jwendell:
It's been harder than I thought... Writing a test for this has been a pain. Could you guys shed some light here?

I changed MockWebServer to not reply any message from the client. But the client seems to not disconnect. Only when the server side closes the connection (MockWebServer.shutdown()) the client does its close stuff... I tried to put a delay (Thread.sleep) in the client (the test method itself) so that after 5 seconds the client sends the close() but it didn't work...

jimmidyson:
Can we get this resurrected? I don't mind submitting a new PR with these changes in now that master has moved on a bit?

jwendell:
@jimmidyson Feel free to continue the work I began. Like I said, writing the test was the difficult part for me. The patch itself is working fine.

jimmidyson:
Things have moved on in 3.4 branch now so the patch may be a bit more involved. For me (as I'm sure for you) the issue was due to a bug in Kubernetes 1.3, which is fixed in 1.4 & OpenShift so might just note the incompatibility instead...

swankjesse:
With recent changes to our API I think this is no longer the right place for this behavior to live. In particular we should force the close after writing a close, not after reading it. I’m going to fix the bug, but the codebase has shifted enough since this PR was sent that a new fix makes more sense.

Thanks for sending the PR, and sorry for not getting it in!

Issue #2795, opened 2016-08-16
Feature : Allow direct connection to unresolved hostname


dryware-fr:
**Use case :** Take an okhttp3 client configured for using a socket factory, for example one given by orchid tor library. Since orchid expects only DIRECT connections with unresolved hostnames, that dns call always will fail. Plus, the client forces us to provide a DNS or let it use the SYSTEM default.

**Related issues :** #2631

In orchid, it is up to the socket to open up a route to the designed host, using tor protocol name resolution.

Maybe you could just add a flag to let us disable that particular dns call and use unresolved address whenever we see fit. Something like that in RouteSelector.resetNextInetSocketAddress : 

```
/** Prepares the socket addresses to attempt for the current proxy or host. */
  private void resetNextInetSocketAddress(Proxy proxy) throws IOException {

/* ... */

    if (proxy.type() == Proxy.Type.SOCKS || proxy.useTor()) { // <== this is the change
      inetSocketAddresses.add(InetSocketAddress.createUnresolved(socketHost, socketPort));
    } else {
      // Try each address for best behavior in mixed IPv4/IPv6 environments.
      List<InetAddress> addresses = address.dns().lookup(socketHost); // <== this is the offending line
      for (int i = 0, size = addresses.size(); i < size; i++) {
        InetAddress inetAddress = addresses.get(i);
        inetSocketAddresses.add(new InetSocketAddress(inetAddress, socketPort));
      }
    }

/* ... */

  }
```

I looked for alternate ways, but the Android JVM Socket only accepts already resolved InetSocketAddress. That prevents people from using the usual SOCKS proxy way, as the Socket will never try to resolve the hostname without asking the proxy. Other JVM implementation don't behave like that.

PR #2796, opened 2016-08-16
Added an option to disable hostname resolution on connecting (for use…

dryware-fr:
… with orchid tor client)

See issue #2795

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/AddressTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/AddressTest.java
@@ -34,18 +34,18 @@ public final class AddressTest {
 
   @Test public void equalsAndHashcode() throws Exception {
     Address a = new Address("square.com", 80, dns, socketFactory, null, null, null,
-        authenticator, null, protocols, connectionSpecs, proxySelector);
+        authenticator, null, protocols, connectionSpecs, proxySelector, false);
     Address b = new Address("square.com", 80, dns, socketFactory, null, null, null,
-        authenticator, null, protocols, connectionSpecs, proxySelector);
+        authenticator, null, protocols, connectionSpecs, proxySelector, false);
     assertEquals(a, b);
     assertEquals(a.hashCode(), b.hashCode());
   }
 
   @Test public void differentProxySelectorsAreDifferent() throws Exception {
     Address a = new Address("square.com", 80, dns, socketFactory, null, null, null,
-        authenticator, null, protocols, connectionSpecs, new RecordingProxySelector());
+        authenticator, null, protocols, connectionSpecs, new RecordingProxySelector(), false);
     Address b = new Address("square.com", 80, dns, socketFactory, null, null, null,
-        authenticator, null, protocols, connectionSpecs, new RecordingProxySelector());
+        authenticator, null, protocols, connectionSpecs, new RecordingProxySelector(), false);
     assertFalse(a.equals(b));
   }
 }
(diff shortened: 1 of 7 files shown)


swankjesse:
Would it work to create a `Dns` implementation that used a placeholder for the socket addresses? Perhaps `127.0.0.1` ?

swankjesse:
Doesn’t tor use the SOCKS proxy anyway?

dryware-fr:
It is possible to connect to the tor network with orchid via the exposed SOCKS proxy.

The thing is that, to use a SOCKS proxy, you provide the proxy to the socket instance with an unresolved InetSocketAddress as destination (with the hostname in it) and it works great with Oracle or openjdk JVMs. The Socket will then connect to the proxy and build the request without any problem.

But on Android dalvikvm, the socket implementation accepts only already resolved InetSocketAddress(es) or fails with an Unresolved host exception. So that makes using the SOCKS proxy on Android impossible. An alternative would be implementing a HTTP proxy, which is supported on Android, but this is a lot of effort.

The simplest way to overcome this is to use the provided SocketFactory, which handles perfectly well any tor connection. And that's when I hit the systematic hostname resolution.

I didn't understood what you had in mind with the local dns implementation but Onion addresses aren't resolved like usual Internet hostnames (a special raw query is sent by orchid on the tor socket at creation time to resolve the hostname and find the path to reach it).

dryware-fr:
Since it really is tor specific, i thought using a specific flag to handle that particular context without impacting the rest of the library would be a nice way to go. It makes OkHttp client more polyvalent without making the library any heavier.

That particular issue happens with Apache HttpClient too and many others. So that would be helpful for every lad wanting to integrate tor in an Android application without depending on an external app (the guardian project pushes forward the use of Orbot app to open a proxy but this is really not stable).

swankjesse:
I'm just wondering whether as a workaround  you could just fake out the DNS.

dryware-fr:
I have a feeling it wouldn't change anything. The socket address must be created unresolved, like for using a SOCKS proxy, except that you get the socket from the socket factory so you don't need the proxy settings.

If you tune the dns, the lookup will only return InetAddresses that doesn't hold the unresolved hostname and so, they'll contain no information. So, in order to do that, you'll have to change the Dns interface to make it capable of returning InetSocketAddresses and not only InetAddresses, plus modifying branches in code when you could use existing ones.

I have a feeling this is more a matter of which branch to take in that particular case than modifying the client features (which works great btw :D) that's why I went for a flag which looks really simple and works.

swankjesse:
If you use `InetAddress.getByAddress(String, byte[])` you can get an `InetAddress` instance that has a hostname (like `foo.onion`) and a placeholder IP address (like `0.0.0.0`). It knows both!

The flag is simple, but also potentially distracting. It suggests OkHttp supports tor, which we don’t really. I’d love to figure out how to get AOSP fixed so the same SOCKS proxy works everywhere.

swankjesse:
(Is there a tracking bug for that? Seems like a real AOSP problem!)

dryware-fr:
It is indeed an AOSP problem, but they voluntarily changed the way it was in the standard JREs.

I thought that extending the library support was a good way to gain users from your side, since that "android feature" affects all available http clients.

dryware-fr:
But even then, it's your call. Tell me if you want me to close this request.

swankjesse:
I’m going to close this because I think offering it as an API might cause user confusion. Doing the hack DNS thing is sad, but it works and I expect that situation to be rare.

PR #2797, opened 2016-08-16
Allow Cache-Control durations to be longs

mlc:
Sometimes, when providing a Cache-Control duration in milliseconds or a smaller unit, you want to be able to specify something greater than `INT_MAX`. This patch enables that, without affecting the code which clamps the final result to `INT_MAX` seconds.


JakeWharton:
This is a binary-incompatible change.

mlc:
that's true, of course.

If you think it's necessary or useful, I'd be happy to add overloads which 
match the existing signatures and cast up to long.

swankjesse:
It's a good idea but probably not worth the cost to migrate.

PR #2806, opened 2016-08-22
Add mvn profile for ALPN version 8.1.9.v20160720 for jdk8 102

boxme:
Added a new profile for jdk8 102 and found that there is also a new version release for org.mortbay.jetty.alpn


swankjesse:
Thanks!

Issue #2815, opened 2016-08-28
Refresh Expiration on Cache Entry on 304 Not Modified Response - 


stevendelapena:
When receiving a 304 response, the cache entry is not updated, reasonably so since the response does not  contain a body. However, since it is not updated, it remains "expired". This means that every subsequent call to this same url will hit the network bypassing the cache until it is updated.  The HTTP spec is vague on how this should be dealt with.

My suggestion is the cache strategy should respect the max-age or expires header sent from the 304 response. Use this to update the cached entry's max-age or expires. This will save network calls for subsequent requests until it is expired again.

swankjesse:
I think we do update the cached headers in response to a 304. Could you write a test case to demonstrate your problem?

stevendelapena:
Seems you are correct and do (as you pointed me to https://github.com/square/okhttp/blob/master/okhttp-tests/src/test/java/okhttp3/CacheTest.java#L1825 in the SO at http://stackoverflow.com/questions/36835550/android-okhttp-how-to-handle-etag/36853783?noredirect=1#comment65736650_36853783 ). This may be an Android specific issue. I'll put together a test using the AndroidJUnit runner and see if it replicates the issue. If so, I'll post back here. One thing I know for sure is that the OkHttp-Sent-Millis and OkHttp-Received-Millis values in the cache entry are never updated from a 304 response. If my understanding is correct, these are used to determine the freshness of the cache. If they never get updated, the cache will always be stale and never used. The odd thing is that the other header entries do get updated from a 304, just not these two.

brillenheini:
Thanks for fixing this! Do you have plans for a release including this fix? This problem is causing lots of unnecessary requests for us.

swankjesse:
Yep. Planning on doing 3.5.0 very soon.

PR #2818, opened 2016-08-29
Fix conditional cache hits to update timestamp fields.

swankjesse:
Closes: https://github.com/square/okhttp/issues/2815


JakeWharton:
Timing tests without a fake clock 😢 

LGTM

cplu:
Could I ask when this fix will be published? or what version in plan

Issue #2707, opened 2016-07-08
Change ResponseBody.charStream() to use Okio's Reader factory.


JakeWharton:
Requires https://github.com/square/okio/issues/239 and allows proper BOM handling.

PR #2819, opened 2016-08-29
Make string() function consider the BOM

danielgindi:
So `response.body().string(true)` will return a string without the BOM.

This helps avoid a lot of boilerplate code just for handling BOMs, in every app that uses OkHttp.

### DIFF
--- a/okhttp/src/main/java/okhttp3/ResponseBody.java
+++ b/okhttp/src/main/java/okhttp3/ResponseBody.java
@@ -162,7 +162,7 @@ public abstract class ResponseBody implements Closeable {
    * possibility for your response.
    */
   public final String string() throws IOException {
-    return new String(bytes(), charset().name());
+    return Util.getStringConsideringBOM(bytes(), charset());
   }
 
   private Charset charset() {
--- a/okhttp/src/main/java/okhttp3/internal/Util.java
+++ b/okhttp/src/main/java/okhttp3/internal/Util.java
@@ -48,6 +48,18 @@ public final class Util {
   /** A cheap and type-safe constant for the UTF-8 Charset. */
   public static final Charset UTF_8 = Charset.forName("UTF-8");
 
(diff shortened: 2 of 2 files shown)


JakeWharton:
I think #2707 will solve this automatically. All calls to `string()` will be BOM-free once that is implemented.

JakeWharton:
A BOM, by its nature and definition, shouldn't be exposed to the consumer in the char world. Only the byte world (aka `source()` or `byteStream()`).

danielgindi:
Tell that to Sun... :)

It's been a month, I thought maybe there are complexities that delay the merge, so I've added a simple function to handle this at least for the `string()`.

If the other one is getting merged for sure, feel free to close this. I hope that this will at least help people coming here looking for a temporary solution, it's a copy-paste function...

swankjesse:
I think we want this API. I’m not completely convinced we want to do it here and not in Okio, but this is a good start.

Can you confirm that this is 100% original code written by you, and complete our CLA?

danielgindi:
Yes, this is completely written by me. Not too complex, I think :)

danielgindi:
And done!

JakeWharton:
I don't want this API. This is providing the user a choice when there is no actual choice needed. Just make `string()` do the Right Thing™. If you want a BOM-containing string then use `source().readUtf8()`. `charStream()` also should do the Right Thing™.

swankjesse:
Yup, I'm convinced. Let's just eat the BOM in string() and charStream(). If you want it, you do your own InputStreamReader.

danielgindi:
Yeah, I thought so too, just didn't want to do introduce any breaking changes. Then I realized that if anyone has considered BOM in their code, they must have done so conditionally.

I've modified the code and rebased.

danielgindi:
I've just fixed a typo in my commit, but I see that you have a bigger solution prepared here for chartStream as well.

JakeWharton:
Just a heads up I submitted #2823 earlier this morning.

PR #2823, opened 2016-08-30
Make string() and charStream() BOM-aware.

### LINKED ISSUE #2707: Change ResponseBody.charStream() to use Okio's Reader factory.

JakeWharton:
Closes #2707. Closes #2819.

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/ResponseBodyTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/ResponseBodyTest.java
@@ -1,0 +1,412 @@
+/*
+ * Copyright (C) 2016 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okhttp3;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.io.Reader;
+import java.util.concurrent.atomic.AtomicBoolean;
+import okio.Buffer;
+import okio.BufferedSource;
+import okio.ByteString;
+import okio.ForwardingSource;
+import okio.Okio;
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+public final class ResponseBodyTest {
+  @Test public void stringEmpty() throws IOException {
+    ResponseBody body = body("");
+    assertEquals("", body.string());
+  }
+
+  @Test public void stringLooksLikeBomButTooShort() throws IOException {
+    ResponseBody body = body("000048");
+    assertEquals("\0\0H", body.string());
+  }
+
+  @Test public void stringDefaultsToUtf8() throws IOException {
+    ResponseBody body = body("68656c6c6f");
+    assertEquals("hello", body.string());
+  }
+
+  @Test public void stringExplicitCharset() throws IOException {
+    ResponseBody body = body("00000068000000650000006c0000006c0000006f", "utf-32be");
+    assertEquals("hello", body.string());
+  }
+
+  @Test public void stringBomOverridesExplicitCharset() throws IOException {
+    ResponseBody body = body("0000ffff00000068000000650000006c0000006c0000006f", "utf-8");
+    assertEquals("hello", body.string());
+  }
+
+  @Test public void stringBomUtf8() throws IOException {
+    ResponseBody body = body("efbbff68656c6c6f");
+    assertEquals("hello", body.string());
+  }
+
+  @Test public void stringBomUtf16Be() throws IOException {
+    ResponseBody body = body("feff00680065006c006c006f");
+    assertEquals("hello", body.string());
+  }
+
+  @Test public void stringBomUtf16Le() throws IOException {
+    ResponseBody body = body("fffe680065006c006c006f00");
+    assertEquals("hello", body.string());
+  }
+
+  @Test public void stringBomUtf32Be() throws IOException {
+    ResponseBody body = body("0000ffff00000068000000650000006c0000006c0000006f");
+    assertEquals("hello", body.string());
+  }
+
+  @Test public void stringBomUtf32Le() throws IOException {
+    ResponseBody body = body("ffff000068000000650000006c0000006c0000006f000000");
+    assertEquals("hello", body.string());
+  }
+
+  @Test public void stringClosesUnderlyingSource() throws IOException {
+    final AtomicBoolean closed = new AtomicBoolean();
+    ResponseBody body = new ResponseBody() {
+      @Override public MediaType contentType() {
+        return null;
+      }
+
+      @Override public long contentLength() {
+        return 5;
+      }
+
+      @Override public BufferedSource source() {
+        Buffer source = new Buffer().writeUtf8("hello");
+        return Okio.buffer(new ForwardingSource(source) {
+          @Override public void close() throws IOException {
+            closed.set(true);
+            super.close();
+          }
+        });
+      }
+    };
+    assertEquals("hello", body.string());
+    assertTrue(closed.get());
+  }
+
+  @Test public void readerEmpty() throws IOException {
+    ResponseBody body = body("");
+    assertEquals("", exhaust(body.charStream()));
+  }
+
+  @Test public void readerLooksLikeBomButTooShort() throws IOException {
+    ResponseBody body = body("000048");
+    assertEquals("\0\0H", exhaust(body.charStream()));
+  }
+
+  @Test public void readerDefaultsToUtf8() throws IOException {
+    ResponseBody body = body("68656c6c6f");
+    assertEquals("hello", exhaust(body.charStream()));
+  }
+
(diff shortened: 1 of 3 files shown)


okhttp/src/main/java/okhttp3/ResponseBody.java
```diff
@@ -145,24 +145,31 @@ public final InputStream byteStream() {
   /**
    * Returns the response as a character stream decoded with the charset of the Content-Type header.
    * If that header is either absent or lacks a charset, this will attempt to decode the response
-   * body as UTF-8.
+   * body in accordance to its BOM or UTF-8.
    */
-  public final Reader charStream() {
+  public final Reader charStream() throws IOException {
```
JakeWharton: Source incompatible but binary compatible. Would prefer to defer this work until first interaction of the returned `Reader` but that seems like a lot of machinery.

swankjesse: Might be worthwhile to do the work. I’m imagining applications that call `charStream()` and then pass that stream to another thread to do the work.

swankjesse: also reading the first byte could take an arbitrarily long amount of time . . . if the server on the other end is doing something fancy (say, gzip) then it will need to batch a whole bunch of data before it emits any.

JakeWharton: Yeah. I worked hard to make calling `byteStream()` and `charStream()` not do any I/O so I'd be sad to see it regress.

JakeWharton: Well not that hard... but let me give this a shot.

JakeWharton: Updated

okhttp/src/main/java/okhttp3/ResponseBody.java
```diff
@@ -145,24 +145,26 @@ public final InputStream byteStream() {
...
+   * header is either absent or lacks a charset, this will attempt to decode the response body in
+   * accordance to its BOM or UTF-8. Closes {@link ResponseBody} automatically.
    *
    * <p>This method loads entire response body into memory. If the response body is very large this
    * may trigger an {@link OutOfMemoryError}. Prefer to stream the response body if this is a
    * possibility for your response.
    */
   public final String string() throws IOException {
-    return new String(bytes(), charset().name());
+    BufferedSource source = source();
```
JakeLoo: Should there be a close on the source?

JakeWharton: Fixed

okhttp/src/main/java/okhttp3/ResponseBody.java
```diff
@@ -215,4 +221,35 @@ public static ResponseBody create(
...
+      if (delegate == null) {
+        Charset charset = Util.bomAwareCharset(source, this.charset);
+        delegate = this.delegate = new InputStreamReader(source.inputStream(), charset);
+      }
+      return delegate.read(cbuf, off, len);
+    }
+
+    @Override public void close() throws IOException {
+      closed = true;
+      if (delegate != null) {
```
swankjesse: this needs an else clause to close `source`. Otherwise creating a BomAwareReader and immediately closing it is problematic

okhttp-tests/src/test/java/okhttp3/ResponseBodyTest.java
```diff
@@ -0,0 +1,139 @@
...
+    assertEquals("hello", body.string());
+  }
+
+  @Test public void stringBomUtf16Le() throws IOException {
+    ResponseBody body = body("FFFE680065006c006c006f00");
+    assertEquals("hello", body.string());
+  }
+
+  @Test public void stringBomUtf32Be() throws IOException {
+    ResponseBody body = body("0000FFFF00000068000000650000006c0000006c0000006f");
```
swankjesse: supernit: a bit weird to mix uppercase and lowercase hex in the same string. (I like lowercase hex)

okhttp-tests/src/test/java/okhttp3/ResponseBodyTest.java
```diff
@@ -0,0 +1,139 @@
...
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import okio.BufferedSource;
+import okio.ByteString;
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+
+public final class ResponseBodyTest {
+  @Test public void stringDefaultsToUtf8() throws IOException {
```
swankjesse: tests for empty string, empty reader, string with leading BOM-like prefix (say `fe20` ?)

swankjesse: Thankfully BufferedSource makes it easy to not accidentally read 4 bytes to skip a BOM that might not exist.

okhttp/src/main/java/okhttp3/ResponseBody.java
```diff
@@ -145,24 +145,30 @@ public final InputStream byteStream() {
   /**
    * Returns the response as a character stream decoded with the charset of the Content-Type header.
    * If that header is either absent or lacks a charset, this will attempt to decode the response
-   * body as UTF-8.
+   * body in accordance to its BOM or UTF-8.
```
swankjesse: worth linking the Wikipedia article on BOMs? Not necessarily a well-known term?

swankjesse:
Yep, this is exactly what we want. Nice change.

swankjesse:
LGTM

okhttp/src/main/java/okhttp3/ResponseBody.java
JakeWharton: I'll add close tests too

okhttp-tests/src/test/java/okhttp3/ResponseBodyTest.java
JakeWharton: mixing hand-written and copy/paste! i'll normalize lower.

okhttp-tests/src/test/java/okhttp3/ResponseBodyTest.java
JakeWharton: Updated

okhttp/src/main/java/okhttp3/ResponseBody.java
JakeWharton: Added

okhttp-tests/src/test/java/okhttp3/ResponseBodyTest.java
```diff
@@ -0,0 +1,405 @@
...
+
+public final class ResponseBodyTest {
+  @Test public void stringEmpty() throws IOException {
+    ResponseBody body = body("");
+    assertEquals("", body.string());
+  }
+
+  @Test public void stringLooksLikeBomButTooShort() throws IOException {
+    ResponseBody body = body("ef48");
+    assertEquals("�H", body.string());
```
swankjesse: supernit: this might read easier as `\uffef`

okhttp-tests/src/test/java/okhttp3/ResponseBodyTest.java
```diff
@@ -0,0 +1,405 @@
...
+
+  @Test public void readerEmpty() throws IOException {
+    ResponseBody body = body("");
+    // BufferedReader.readLine is documented to return null if the end of stream is reached.
+    assertNull(new BufferedReader(body.charStream()).readLine());
+  }
+
+  @Test public void readerLooksLikeBomButTooShort() throws IOException {
+    ResponseBody body = body("ef48");
+    assertEquals("�H", new BufferedReader(body.charStream()).readLine());
```
swankjesse: same

swankjesse: nit: `readLine()` when you probably want `readString`

okhttp-tests/src/test/java/okhttp3/ResponseBodyTest.java
```diff
@@ -0,0 +1,405 @@
...
+    ResponseBody body = body("");
+    BufferedSource source = body.source();
+    assertTrue(source.exhausted());
+    assertEquals("", source.readUtf8());
+  }
+
+  @Test public void sourceSeesBom() throws IOException {
+    ResponseBody body = body("efbbff68656C6C6F");
+    BufferedSource source = body.source();
+    assertEquals(0xEF, source.readByte() & 0xFF);
```
swankjesse: nit: more uppercase hexes

okhttp/src/main/java/okhttp3/ResponseBody.java
```diff
@@ -137,32 +137,44 @@ public final InputStream byteStream() {
       Util.closeQuietly(source);
     }
     if (contentLength != -1 && contentLength != bytes.length) {
-      throw new IOException("Content-Length and stream length disagree");
+      throw new IOException("Content-Length ("
+          + contentLength
```
swankjesse: nice!

swankjesse:
LGTM!

okhttp-tests/src/test/java/okhttp3/ResponseBodyTest.java
JakeWharton: Hmm I don't think `readString()` exists anywhere in java.io. It's why everyone turns to commons-io or Guava.... or Okio!

JakeWharton: Added an `exhaust(Reader)` helper

JakeWharton:
Hmm weird string decoding problems on Java 7. Will look later.

danielgindi:
Looks good to me too. Tested all BOMs - working correctly. 

When is a release expected? ;-)

danielgindi:
Okay so I guess I've tested everything *but* BOM8.
The bom you've written in the new PR was sadly not copied correctly, and has a typo.

#2994

Issue #2474, opened 2016-04-10
Require connection preface on HTTP/2 connections


swankjesse:
Currently we aren’t strict on this.

https://github.com/square/okhttp/pull/2473/files/8996934748a48fe04f7dde56069042dfecb64c91#r59138454

boxme:
Wouldn't the following in `RealConnection` already ensure that connection preface is being enforced on HTTP/2 connections?
```
if (protocol == Protocol.HTTP_2) {
      socket.setSoTimeout(0); // Framed connection timeouts are set per-stream.

      Http2Connection http2Connection = new Http2Connection.Builder(true)
          .socket(socket, route.address().url().host(), source, sink)
          .listener(this)
          .build();
      http2Connection.start();
```

Or do you have the idea to enforce it on the `Http2Writer` level?

swankjesse:
There's a few places where we need to fail the connection if the peer doesn't send it. In MockWebServer for one example.

PR #2825, opened 2016-08-31
ConnectException in MockWebServer if no connection preface is received for HTTP2 from a client

boxme:
#2474

A ConnectionException will be thrown in the MockWebServer if it does not receive a connection preface from it's client when a HTTP2 connection is established.


swankjesse:
Thanks for the PR. I need to investigate our APIs to see how this should fit in. It’s a bit of a special case because it occurs before any HTTP request is made.

swankjesse:
Thanks! Obsoleted by this PR:
https://github.com/square/okhttp/pull/3032/files

PR #2833, opened 2016-09-02
Be more liberal adding headers to the HPACK encoder dynamic table.

dave-r12:
Previously, if a header name was in the static table we'd avoid adding it
to the dynamic table. Since most common headers are in the static table,
we were missing the chance to save on bytes sent.

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/internal/http2/HpackTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/internal/http2/HpackTest.java
@@ -925,6 +925,77 @@ public final class HpackTest {
     assertEquals(expected, actual);
   }
 
+  @Test public void staticTableIndexedHeader() throws IOException {
+    hpackWriter.writeHeaders(headerEntries(":method", "GET"));
+    assertBytes(0x82);
+    assertEquals(0, hpackWriter.headerCount);
+  }
+
+  @Test public void dynamicTableIndexedHeader() throws IOException {
+    hpackWriter.writeHeaders(headerEntries("custom-key", "custom-header"));
+    assertBytes(0x40,
+        10, 'c', 'u', 's', 't', 'o', 'm', '-', 'k', 'e', 'y',
+        13, 'c', 'u', 's', 't', 'o', 'm', '-', 'h', 'e', 'a', 'd', 'e', 'r');
+    assertEquals(1, hpackWriter.headerCount);
+
+    hpackWriter.writeHeaders(headerEntries("custom-key", "custom-header"));
+    assertBytes(0xbe);
+    assertEquals(1, hpackWriter.headerCount);
+  }
+
+  @Test public void doNotIndexPseudoHeaders() throws IOException {
+    hpackWriter.writeHeaders(headerEntries(":method", "PUT"));
+    assertBytes(0x02, 3, 'P', 'U', 'T');
+    assertEquals(0, hpackWriter.headerCount);
+
+    hpackWriter.writeHeaders(headerEntries(":method", "PUT"));
+    assertBytes(0x02, 3, 'P', 'U', 'T');
+    assertEquals(0, hpackWriter.headerCount);
+  }
+
+  @Test public void incrementalIndexingWithAuthorityPseudoHeader() throws IOException {
+    hpackWriter.writeHeaders(headerEntries(":authority", "foo.com"));
+    assertBytes(0x41, 7, 'f', 'o', 'o', '.', 'c', 'o', 'm');
+    assertEquals(1, hpackWriter.headerCount);
+
+    hpackWriter.writeHeaders(headerEntries(":authority", "foo.com"));
+    assertBytes(0xbe);
+    assertEquals(1, hpackWriter.headerCount);
+
+    // If the :authority header somehow changes, it should be re-added to the dynamic table.
+    hpackWriter.writeHeaders(headerEntries(":authority", "bar.com"));
+    assertBytes(0x41, 7, 'b', 'a', 'r', '.', 'c', 'o', 'm');
+    assertEquals(2, hpackWriter.headerCount);
+
+    hpackWriter.writeHeaders(headerEntries(":authority", "bar.com"));
+    assertBytes(0xbe);
+    assertEquals(2, hpackWriter.headerCount);
+  }
+
+  @Test public void incrementalIndexingWithStaticTableIndexedName() throws IOException {
+    hpackWriter.writeHeaders(headerEntries("accept-encoding", "gzip"));
+    assertBytes(0x50, 4, 'g', 'z', 'i', 'p');
+    assertEquals(1, hpackWriter.headerCount);
+
+    hpackWriter.writeHeaders(headerEntries("accept-encoding", "gzip"));
+    assertBytes(0xbe);
+    assertEquals(1, hpackWriter.headerCount);
+  }
+
+  @Test public void incrementalIndexingWithDynamcTableIndexedName() throws IOException {
+    hpackWriter.writeHeaders(headerEntries("foo", "bar"));
+    assertBytes(0x40, 3, 'f', 'o', 'o', 3, 'b', 'a', 'r');
+    assertEquals(1, hpackWriter.headerCount);
+
+    hpackWriter.writeHeaders(headerEntries("foo", "bar1"));
+    assertBytes(0x7e, 4, 'b', 'a', 'r', '1');
+    assertEquals(2, hpackWriter.headerCount);
+
+    hpackWriter.writeHeaders(headerEntries("foo", "bar1"));
+    assertBytes(0xbe);
+    assertEquals(2, hpackWriter.headerCount);
+  }
+
   private Hpack.Reader newReader(Buffer source) {
     return new Hpack.Reader(4096, source);
   }
--- a/okhttp/src/main/java/okhttp3/internal/http2/Hpack.java
+++ b/okhttp/src/main/java/okhttp3/internal/http2/Hpack.java
@@ -152,7 +152,6 @@ final class Hpack {
     }
 
     private void clearDynamicTable() {
-      headerList.clear();
       Arrays.fill(dynamicTable, null);
       nextHeaderIndex = dynamicTable.length - 1;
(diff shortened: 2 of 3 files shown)


okhttp/src/main/java/okhttp3/internal/http2/Hpack.java
```diff
@@ -152,7 +154,6 @@ private void adjustDynamicTableByteCount() {
     }
 
     private void clearDynamicTable() {
-      headerList.clear();
```
dave-r12: I triggered an edge case here in the HPACK decoder. If the server HPACK encoder sent a header intended to be added to the dynamic table, and it's length was larger than the entire dynamic table, we'd clear the header list. My understanding is that we should never clear the header list, even if a header wipes out all entries in the dynamic table.

To see it in action, re-add this line and run HpackTest#largeHeaderValue test case.

okhttp/src/main/java/okhttp3/internal/http2/Hpack.java
```diff
@@ -463,29 +464,37 @@ void writeHeaders(List<Header> headerBlock) throws IOException {
         smallestHeaderTableSizeSetting = Integer.MAX_VALUE;
         writeInt(maxDynamicTableByteCount, PREFIX_5_BITS, 0x20);
       }
-      // TODO: implement index tracking
-      for (int i = 0, size = headerBlock.size(); i < size; i++) {
-        Header header = headerBlock.get(i);
+
+      for (Header header : headerBlock) {
```
JakeWharton: Switch back to indexed for here? We try to use them everywhere in libraries to avoid the allocation and indirection that comes with `Iterator`.

okhttp/src/main/java/okhttp3/internal/http2/Hpack.java
```diff
@@ -511,6 +520,36 @@ void writeInt(int value, int prefixMask, int bits) {
...
+      return -1;
+    }
+
+    /** Converts a raw array index in the dynamic table array into the HPACK index address space. */
+    int normalizeDynamicTableIndex(int arrayIndex) {
+      return arrayIndex - nextHeaderIndex + STATIC_HEADER_TABLE.length;
+    }
+
+    int getHeaderNameIndex(ByteString name) {
+      if (NAME_TO_FIRST_INDEX.containsKey(name)) return NAME_TO_FIRST_INDEX.get(name) + 1;
```
JakeWharton: Call `get()` and check for `null` to avoid having to do double hashing and lookup.

okhttp/src/main/java/okhttp3/internal/http2/Hpack.java
dave-r12: Ahh, ok. Will fix.

okhttp/src/main/java/okhttp3/internal/http2/Hpack.java
dave-r12: Got it, done.

okhttp-tests/src/test/java/okhttp3/internal/http2/HpackTest.java
```diff
@@ -925,6 +925,77 @@ private void checkReadThirdRequestWithHuffman() {
...
+    assertEquals(1, hpackWriter.headerCount);
+
+    hpackWriter.writeHeaders(headerEntries("accept-encoding", "gzip"));
+    assertBytes(0xbe);
+    assertEquals(1, hpackWriter.headerCount);
+  }
+
+  @Test public void incrementalIndexingWithDynamcTableIndexedName() throws IOException {
+    hpackWriter.writeHeaders(headerEntries("foo", "bar"));
+    assertBytes(0x40, 3, 'f', 'o', 'o', 3, 'b', 'a', 'r');
```
swankjesse: these tests are really readable and good

okhttp/src/main/java/okhttp3/internal/http2/Hpack.java
```diff
@@ -511,6 +521,37 @@ void writeInt(int value, int prefixMask, int bits) {
...
+      // unnecessary to waste cycles looking at them.
+      for (int i = 1; i < 7; i++) {
+        if (equal(STATIC_HEADER_TABLE[i], header)) return i + 1;
+      }
+
+      return -1;
+    }
+
+    /** Converts a raw array index in the dynamic table array into the HPACK index address space. */
+    int normalizeDynamicTableIndex(int arrayIndex) {
```
swankjesse: should the method name better match what it does? toHpackIndex ?

okhttp/src/main/java/okhttp3/internal/http2/Hpack.java
```diff
@@ -511,6 +521,37 @@ void writeInt(int value, int prefixMask, int bits) {
       out.writeByte(value);
     }
 
+    int getHeaderIndex(Header header) {
+      int dynamicHeaderIndex = Util.indexOf(dynamicTable, header);
+      if (dynamicHeaderIndex != -1) return normalizeDynamicTableIndex(dynamicHeaderIndex);
+
+      // Only search a subset of the static header table. Most entries have an empty value, so it's
+      // unnecessary to waste cycles looking at them.
```
swankjesse: nice

okhttp/src/main/java/okhttp3/internal/http2/Hpack.java
```diff
@@ -511,6 +521,37 @@ void writeInt(int value, int prefixMask, int bits) {
...
+    /** Converts a raw array index in the dynamic table array into the HPACK index address space. */
+    int normalizeDynamicTableIndex(int arrayIndex) {
+      return arrayIndex - nextHeaderIndex + STATIC_HEADER_TABLE.length;
+    }
+
+    int getHeaderNameIndex(ByteString name) {
+      Integer staticIndex = NAME_TO_FIRST_INDEX.get(name);
+      if (staticIndex != null) return staticIndex + 1;
+
+      for (int i = 0, size = dynamicTable.length; i < size; i++) {
```
swankjesse: More stuff to consider, for now or for later...

Rather than searching the entire dynamic table (which has nulls) we could loop from `i = nextHeaderIndex + 1`  through the size. That avoids the null check here.

And we could do likewise in the code that computes `dynamicIndex` also.
```
          int dynamicIndex = Util.indexOf(dynamicTable, header);
```
Seems like this code was mostly written to do the job, and we haven’t really looked over it to make it optimal.

Also while I’m here, I was looking in `evictToRecoverBytes()` and I think the bounds of that loop are wrong.
```
        for (int j = dynamicTable.length - 1; j >= nextHeaderIndex && bytesToRecover > 0; j--) {
```
Since `nextHeaderIndex` should always be null, that check should either be `>` instead of `>=`, or even better – we can remove the check completely. As written there’s no way it ever matters.

swankjesse: (I can handle this in a follow-up, or you can, or whatever!)

swankjesse: Also worth considering, given an infinite amount of time...

Right now we loop over the static & dynamic tables up to two times:
 - look for an exact match
 - look for a name match

What would happen if we looped 1x, and returned the ”best match” – the exact index if it’s found, or the header index otherwise. Figuring out a way to return that unambiguously without allocation is painful (negative vs. positive?) and might make it not worth the effort.

swankjesse:
LGTM

swankjesse:
Also – I _think_ our HPACK test suite is still disabled. Might be worth running it again for good measure.

okhttp/src/main/java/okhttp3/internal/http2/Hpack.java
dave-r12: Good points. I'm going to change the implementation a bit so it only requires a single pass. I will also incorporate skipping the known null elements.

Good catch on the `evictToRecoverBytes`. I'll leave that out of this PR.

dave-r12:
Yes, I've been running them quite often. I'm also running some manual tests against other HTTP/2 servers.

okhttp/src/main/java/okhttp3/internal/http2/Hpack.java
dave-r12: Yes - I ended up inlining this into the main for loop.

okhttp/src/main/java/okhttp3/internal/http2/Hpack.java
```diff
@@ -463,29 +462,56 @@ void writeHeaders(List<Header> headerBlock) throws IOException {
...
+            break;
+          }
+        }
+
+        if (headerIndex == -1) {
+          for (int j = nextHeaderIndex + 1, length = dynamicTable.length; j < length; j++) {
+            if (Util.equal(dynamicTable[j], header)) {
+              headerIndex = j - nextHeaderIndex + STATIC_HEADER_TABLE.length;
+              break;
+            } else if (headerNameIndex == -1 && Util.equal(dynamicTable[j].name, name)) {
```
swankjesse: One option to possibly consider:

```
  for (int j = nextHeaderIndex + 1, length = dynamicTable.length; j < length; j++) {
    if (Util.equal(dynamicTable[j].name, name)) {
      if (Util.equal(dynamicTable[j].value, value)) {
        headerIndex = j - nextHeaderIndex + STATIC_HEADER_TABLE.length;
        break;
      } else {
        headerNameIndex = j - nextHeaderIndex + STATIC_HEADER_TABLE.length;
      }
    }
```
Ie. test if the name matches only once.

swankjesse:
# LGTM! I’m going to merge as-is ’cause this is a nice improvement. There are some things to follow up on but no need to hold this.

swankjesse:
(for my own use, possible followups are fixing bounds checks on `evictToRecoverBytes()` and only checking the name once in `writeHeaders()`).

okhttp/src/main/java/okhttp3/internal/http2/Hpack.java
dave-r12: Good call, will follow up.

dave-r12:
Hehe, apologies for all the churn here.

PR #2834, opened 2016-09-04
Modify for loop to more performance sensitive

shaunkawano:
I have modified one of the for-loops to improve its performance.

Reference: https://speakerdeck.com/jakewharton/eliminating-code-overhead-square-hq-2015


shaunkawano:
I will close this as this is dup of #2835

PR #2835, opened 2016-09-04
Modify for-loop more performance sensitively

shaunkawano:
I happen to find one for-loop that can be written more performance sensitively, based on this: 

https://speakerdeck.com/jakewharton/eliminating-code-overhead-square-hq-2015


swankjesse:
Thanks.

iNoles:
This one used 
```java
for (int i = 0; i < size; i++) {
```
than
```java
for (int i = 0, size = list.size(); i < size; i++) {
```

I thought first one wasted more memory than 2nd one.

swankjesse:
The second form is generally favorable. In this case the first is okay because we avoid making two calls to `list.size()`.

swankjesse:
... because we use size() both in the loop and also to pre-size an `ArrayList.`

PR #2850, opened 2016-09-13
Make WebSocketCall an abstract class.

JakeWharton:
Ideally it would be an interface like Call, but the static factory method needs to hang somewhere.


swankjesse:
Is it time to promote websockets into OkHttp proper?

JakeWharton:
Heh... see Slack

JakeWharton:
#2852

PR #2852, opened 2016-09-13
Move WebSockets from separate artifact to main.

JakeWharton:
The factory method has moved to OkHttpClient and WebSocketCall has changed to an interface.


dave-r12:
Nice!

artem-zinnatullin:
Sorry for being late to the party, but why?

swankjesse:
It makes the implementation simpler.

artem-zinnatullin:
But isn't it kind of wrong that HTTP client includes implementation of
another protocol (yes they're related to each other, but still different
things)?

Or you're trying to support everything network related as modern browsers
such as Chrome offer?

Another point is that most of OkHttp users use library for um, well, HTTP,
adding WebSockets support will simply bloat library for many users (hope
ProGuard/etc will help).

JakeWharton:
HTTP/1.1 and HTTP/2 are different but related protocols as well. Both are
included.

swankjesse:
It's not a slippery slope.

artem-zinnatullin:
Ok! Though comparison with HTTP/2 is not very accurate, most of HTTP/1
users will eventually and probably automatically migrate to /2 which is
completely not true for HTTP (1 or 2) with WebSockets.

PR #2853, opened 2016-09-13
Expose Part headers and body.

### LINKED ISSUE #2845: cannot deal with request body by Interceptors when upload files & parameters by MultipartBody
What kind of issue is this?

 - [ ] Feature Request. 

1. function: upload files & paramters, but need encrypt the request body(just string parameters), so I try to use interceptors to solve it.

2. but I have no idea how to separate the (from)request body(not files) from "public Response intercept(Chain chain)",  it can get the Parts list from request, but cann't get every request body from part.

JakeWharton:
Closes #2845.

### DIFF
--- a/okhttp/src/main/java/okhttp3/MultipartBody.java
+++ b/okhttp/src/main/java/okhttp3/MultipartBody.java
@@ -257,12 +257,20 @@ public final class MultipartBody extends RequestBody {
       return create(Headers.of("Content-Disposition", disposition.toString()), body);
     }
 
-    private final Headers headers;
-    private final RequestBody body;
+    final Headers headers;
+    final RequestBody body;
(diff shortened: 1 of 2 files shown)


okhttp/src/main/java/okhttp3/MultipartBody.java
```diff
@@ -257,13 +257,21 @@ public static Part createFormData(String name, String filename, RequestBody body
       return create(Headers.of("Content-Disposition", disposition.toString()), body);
     }
 
-    private final Headers headers;
-    private final RequestBody body;
+    final Headers headers;
+    final RequestBody body;
```
JakeWharton: Eliminating synthetic accessor methods

PR #2858, opened 2016-09-16
Corrected javadoc of the Cookie persistent() method

franmontiel:
It was claiming that returns true if the cookie expires and the end of the session and really is the opposite behavior.


swankjesse:
Whoa that’s embarrassing. Thanks!

PR #2860, opened 2016-09-17
Minor memory optimization for Util.immutableList

kilink:
I didn't doing any kind of formal analysis, but simply running the tests showed that several calls to Utils.immutableList yielded either the empty list instance or a singleton list, instead of allocating an ArrayList + UnmodifiedList.

Definitely a micro-optimization, but seems harmless at worst.

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/internal/connection/RouteSelectorTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/internal/connection/RouteSelectorTest.java
@@ -24,7 +24,6 @@ import java.net.SocketAddress;
 import java.net.URI;
 import java.net.UnknownHostException;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.List;
 import java.util.NoSuchElementException;
 import javax.net.SocketFactory;
@@ -72,7 +71,7 @@ public final class RouteSelectorTest {
   private HostnameVerifier hostnameVerifier;
 
   private final Authenticator authenticator = Authenticator.NONE;
-  private final List<Protocol> protocols = Arrays.asList(Protocol.HTTP_1_1);
+  private final List<Protocol> protocols = Util.immutableList(Protocol.HTTP_1_1);
   private final FakeDns dns = new FakeDns();
   private final RecordingProxySelector proxySelector = new RecordingProxySelector();
   private RouteDatabase routeDatabase = new RouteDatabase();
--- a/okhttp/src/main/java/okhttp3/internal/Util.java
+++ b/okhttp/src/main/java/okhttp3/internal/Util.java
@@ -226,11 +226,16 @@ public final class Util {
 
   /** Returns an immutable copy of {@code list}. */
   public static <T> List<T> immutableList(List<T> list) {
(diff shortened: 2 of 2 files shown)


swankjesse:
We used to do things like this all the time. But I’ve learned that in practice it can make things slower! https://shipilev.net/blog/2015/black-magic-method-dispatch/

kilink:
That's an interesting article, but it seems to be focused on the performance benchmark of virtual method dispatch, so I'm at a loss as to how it applies here (the added calls to isEmpty / size?). I don't think anything here would affect the inline-ability of the (static) method itself, and I doubt that calls to this method are on any particular critical path.

By the way, the change is similar to what Guava's ImmutableList.copyOf() and Java 9's List.of() do for empty and single-item lists / arrays.

I ran some benchmarks since I was curious about the performance impact, and I found that this variant is actually faster is all cases. Admittedly I'm no expert at benchmarking, but it seems like the change is harmless at worst, and saves a bit of memory at best. I'm fine if you think memory savings just aren't worth it for whatever reason, but I'm not convinced that performance is at all an issue.

Thanks!

swankjesse:
The dispatched method is List.size() and List.get(). You won’t see it in benchmarks because they don’t usually capture dispatch performance. Guava’s methods may have been a mistake. Lemme follow up with them to confirm.

kilink:
For Guava, I was thinking of the overload of [copyOf](https://github.com/google/guava/blob/master/guava/src/com/google/common/collect/ImmutableList.java#L247) that takes an Iterator, which uses hasNext / next to special case for 0 or 1 elements.  The Iterable overload will dispatch to this method if the Iterable isn't also a Collection. I assume the overhead here is comparable to calling isEmpty / size / get, since there are a few instanceof checks and the Iterator allocation + hasNext() and next() calls.

Their [copyOf](https://github.com/google/guava/blob/master/guava/src/com/google/common/collect/ImmutableList.java#L266) overload that takes an array checks the array length to special case for 0 or 1 elements, like this variant (although they use a switch).

There is no immutable list factory method in Java 9 that takes a Collection/List as a parameter, instead the only comparable one is the [one](http://hg.openjdk.java.net/jdk9/jdk9/jdk/file/32d957185656/src/java.base/share/classes/java/util/List.java#l1029) that takes an array of elements. That one also special cases for 0 or 1 elements by checking the array length. Note that the Java 9 factory methods were [influenced by Guava](http://openjdk.java.net/jeps/269).

I'm not trying to make an argument from authority by citing those examples, but it is interesting to look at similar implementations. It seems for the array vararg case it's a clear win, but it would probably be bad to have special behavior in one method but not the other.

Tangent: to write a fair benchmark, couldn't I create a bunch of different sized lists using various implementations (ArrayList, LinkedList, emptyList, SingletonList, etc), then randomly choose one and pass it into Utils.immutableList?

swankjesse:
I dug up the blog post that shapes my thinking on this.
http://smallwig.blogspot.ca/2010/03/little-optimization-that-couldnt.html

swankjesse:
And here's the Guava discussion:
https://github.com/google/guava/issues/1268

PR #2862, opened 2016-09-18
Eliminate synthetic accessor methods, an allocation in web sockets.

JakeWharton:


okhttp/src/main/java/okhttp3/internal/ws/RealWebSocket.java
```diff
@@ -52,38 +55,37 @@
...
+    this.replyExecutor = replyExecutor;
     this.listener = listener;
+    this.url = url;
 
     writer = new WebSocketWriter(isClient, sink, random);
-    reader = new WebSocketReader(isClient, source, new FrameCallback() {
-      @Override public void onMessage(ResponseBody message) throws IOException {
-        listener.onMessage(message);
-      }
+    reader = new WebSocketReader(isClient, source, this);
```
JakeWharton: Here's the allocation elimination. Moved the listener to just be on the enclosing type.

PR #2863, opened 2016-09-18
Remove connection read/write timeouts for web sockets.

### LINKED ISSUE #1930: OkHttpClient default timeouts not reasonable for websockets
A change [https://github.com/square/okhttp/commit/0de14e992c228fd9de9fe78417788131bee00902] was made to set the default timeouts for read/write to 10 seconds. Having a read timeout doesn't really make sense for a websocket connection. 

Workaround - `okHttpClient.setReadTimeout(0, TimeUnit.NANOSECONDS);`

JakeWharton:
Closes #1930.

### DIFF
--- a/okhttp/src/main/java/okhttp3/RealWebSocketCall.java
+++ b/okhttp/src/main/java/okhttp3/RealWebSocketCall.java
@@ -19,6 +19,7 @@ import java.io.IOException;
 import java.net.ProtocolException;
 import java.security.SecureRandom;
 import java.util.Collections;
+import java.util.List;
 import java.util.Random;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.LinkedBlockingDeque;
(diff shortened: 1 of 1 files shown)


jimmidyson:
Sorry to revisit a merged PR but how should we handle remote servers going down with infinite read timeouts? There seems no way other than tuning TCP stack for established connections to be timed out in this case?

swankjesse:
@jimmidyson good point! Wanna open an issue and we’ll investigate what other user agents do?

JakeWharton:
You could send pings on regular intervals, at the rate at which you care
about timeouts (e.g., every 30 seconds).

jimmidyson:
@JakeWharton Yeah that depends on target server handling that correctly though, right?

JakeWharton:
Well if it doesn't handle a ping then it's not spec compliant. Even if it
doesn't handle pings you can fake them as messages and expect responses
with a timeout to indicate the server disappeared. This is recommended by a
lot of web socket APIs (e.g., Slack).

jimmidyson:
@JakeWharton Ah now that is not such a bad idea. A periodic dummy request that you discard the response from but just use to satisfy read timeout (if ping isn't supported - I didn't know that was part of the spec - thanks for that tip).

jimmidyson:
Raised #2876.

PR #2865, opened 2016-09-18
Web socket API and implementation improvements.

### LINKED ISSUE #1701: When server send message with Mask bit on, it goes to onFailure callback, not close the websocket
According to [rfc6455, chapter 5.1](http://tools.ietf.org/html/rfc6455#section-5.1), "A server MUST NOT mask any frames that it sends to the client.  A client MUST close a connection if it detects a masked frame".

But when my app server respond to my client close message, it send a masked frame, and the WebSocketClient didn't close the connection, in fact, it call the onFailure callback, but still with the connection on!

I'm using the version 2.4.0

### LINKED ISSUE #2455: RejectedExecutionException in RealWebSocket.onClose()
I'm getting a lot of RejectedExecutionException from line okhttp3.internal.ws.RealWebSocket$1.onClose (RealWebSocket.java:82) while using the engine.io/socket.io library. 

I assume that ping response runnable that is writing the pong by locking the thread, is staying in the pool long enough to create a problem when server requests socket close in the meantime? Since only one runnable can be in the pool at the time and so RejectedExecutionException is thrown.

https://github.com/square/okhttp/commit/8a1e6ceec92963423b4cc9f5edded6363fc2c43f#diff-84851fc7b3d535627088311c7ca1951dR70

Library version 3.0.1

Sorry for not writing the test case, I spent the whole day so far trying to find the issue...

JakeWharton:
* onFailure callback's exception type widened to Throwable. This allows runtime exceptions from other callbacks to be passed along.
* Ensure the connection is closed properly for all failures.
* Fix and document the threading inside RealWebSocket. This ensures the listener is always called on the same thread and replies always happen from the correct thread.

Closes #2455. Closes #1701.

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/WebSocketCallTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/WebSocketCallTest.java
@@ -18,26 +18,30 @@ package okhttp3;
 import java.io.IOException;
 import java.net.ProtocolException;
 import java.util.Random;
-import java.util.concurrent.CountDownLatch;
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicReference;
+import java.util.logging.Logger;
 import okhttp3.internal.tls.SslClient;
+import okhttp3.internal.ws.EmptyWebSocketListener;
 import okhttp3.internal.ws.WebSocketRecorder;
 import okhttp3.mockwebserver.MockResponse;
 import okhttp3.mockwebserver.MockWebServer;
 import okio.Buffer;
 import org.junit.After;
+import org.junit.Ignore;
 import org.junit.Rule;
 import org.junit.Test;
 
+import static okhttp3.WebSocket.BINARY;
 import static okhttp3.WebSocket.TEXT;
+import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.fail;
 
 public final class WebSocketCallTest {
-  @Rule public final MockWebServer server = new MockWebServer();
+  @Rule public final MockWebServer webServer = new MockWebServer();
 
   private final SslClient sslClient = SslClient.localhost();
-  private final WebSocketRecorder listener = new WebSocketRecorder();
+  private final WebSocketRecorder clientListener = new WebSocketRecorder("client");
+  private final WebSocketRecorder serverListener = new WebSocketRecorder("server");
   private final Random random = new Random(0);
   private OkHttpClient client = new OkHttpClient.Builder()
       .addInterceptor(new Interceptor() {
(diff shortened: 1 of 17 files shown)


JakeWharton:
Sorry there's a lot here. Once you start pulling a thread...

okhttp/src/main/java/okhttp3/internal/ws/RealWebSocket.java
```diff
@@ -22,92 +22,207 @@
...
+ * <li><b>Reader</b>: This is the only thread allowed to access {@link #reader}. Methods from
+ * {@link FrameCallback} will happen on this thread as a result. This is the only thread that
+ * should invoke methods on the {@link #readerListener}.</li>
+ * <li><b>Replier</b>: Invoked on {@link #replier} to write responses from reading
+ * frames. Contends with the "Sender" thread for access to {@link #writer}.</li>
+ * <li><b>Sender</b>: Methods from {@link WebSocket} will happen on this thread. Contends with the
+ * "Replier" thread</li>
+ * </ul>
+ * Instance variables have prefixes matching the thread names based on the thread on which they can
+ * be accessed. A prefix of "writer" indicates both "Reader" and "Replier" threads can access.
```
dave-r12: Should this read: "writer" indicates both "_Sender_" and "Replier" ...?

okhttp-tests/src/test/java/okhttp3/WebSocketCallTest.java
```diff
@@ -50,132 +54,281 @@
...
+    enqueueWebSocket();
 
-    awaitWebSocket();
-    listener.assertTextMessage("Hello, WebSockets!");
+    serverListener.assertOpen();
+    serverListener.assertClose(1001, "");
+    clientListener.assertFailure(e);
+  }
+
+  @Ignore("AsyncCall currently lets runtime exceptions propagate.")
```
swankjesse: yup, good test

okhttp/src/main/java/okhttp3/WebSocketListener.java
```diff
@@ -18,27 +18,34 @@
...
+ *
+ * <h2>Lifecycle Rules</h2>
+ * <ul>
+ * <li>Either {@link #onOpen} or {@link #onFailure} will be called first depending on if the web
+ * socket was successfully opened or if there was an error connecting to the server or parsing its
+ * response.</li>
+ * <li>After {@link #onOpen} is called, {@link #onFailure} can be called at any time. No more
+ * callbacks will follow a call to {@link #onFailure}.</li>
+ * <li>After {@link #onOpen} is called, {@link #onMessage} and {@link #onPong} will be called for
+ * each message and pong frame, respectively. Note: {@link #onPong} may be called while {@link
```
swankjesse: hang on, are there multiple callback threads?

swankjesse: Ahh, because it’s re-entrant. That’s awkward!

okhttp/src/main/java/okhttp3/internal/Util.java
```diff
@@ -463,4 +463,11 @@ public static Charset bomAwareCharset(BufferedSource source, Charset charset) th
     }
     return charset;
   }
+
+  /** Re-throws {@code t} if it is a fatal exception which should not be handled. */
+  public static void throwIfFatal(Throwable t) {
+    if (t instanceof VirtualMachineError) throw (VirtualMachineError) t;
+    if (t instanceof ThreadDeath) throw (ThreadDeath) t;
+    if (t instanceof LinkageError) throw (LinkageError) t;
```
swankjesse: OutOfMemoryError ?

okhttp/src/main/java/okhttp3/internal/ws/RealWebSocket.java
```diff
@@ -22,92 +22,207 @@
...
-    replyExecutor.execute(new NamedRunnable("OkHttp %s WebSocket Pong Reply", url) {
-      @Override protected void execute() {
-        peerPing(buffer);
-      }
-    });
+  /**
+   * Read a single message from the web socket and deliver it to the listener. Returns false when
+   * no more messages can be read.
+   */
+  final boolean readMessage() {
```
swankjesse: readFrame ?

okhttp/src/main/java/okhttp3/internal/ws/RealWebSocket.java
```diff
@@ -22,92 +22,207 @@
...
     }
   }
 
+  ////// SENDER THREAD (aka user thread)
+
   @Override public final void sendMessage(RequestBody message) throws IOException {
     if (message == null) throw new NullPointerException("message == null");
-    if (writerSentClose) throw new IllegalStateException("closed");
-    if (writerWantsClose) throw new IllegalStateException("must call close()");
+    if (senderSentClose) throw new IllegalStateException("closed");
```
swankjesse: sender’s a better word here

okhttp/src/main/java/okhttp3/internal/Util.java
JakeWharton: That's a VME subclass.

okhttp/src/main/java/okhttp3/internal/ws/RealWebSocket.java
JakeWharton: This reads a single message which may comprise one to many frames.

JakeWharton: Actually it reads a single control frame _or_ a message that may be one to many frames with control frames potentially interleaved... so `readFrame()` isn't terrible it's just not always strictly correct. Maybe `readFrames()` + some documentation?

okhttp/src/main/java/okhttp3/WebSocketListener.java
JakeWharton: Yep! Not sure what to do otherwise. A queue would skew timings if you were using ping/pongs to measure.

okhttp/src/main/java/okhttp3/internal/ws/RealWebSocket.java
JakeWharton: I changed to `processNextFrame` which is the name of the method on `WebSocketReader` that it is proxying to.

JakeWharton:
Thanks for review! Smaller PRs in the future I promise.

PR #2868, opened 2016-09-19
Do not disable TLSv1.3 by default

### LINKED ISSUE #2867: Support TLSv1.3
TLSv1.3 is coming. okhttp needs to be adjusted to not disable TLSv1.3 in the underlying TLS stack, or to at least let the user of okhttp enable TLSv1.3 if the underlying TLS stack supports it.

At the moment, okhttp prevents the use of TLSv1.3 even if the underlying TLS stack offers it. My understanding is that all that needs to change in okhttp is: (1) add TLS_1_3 enum to TlsVersion, and (2) add TlsVersion.1_3 to ConnectionSpec.MODERN_TLS.

klyubin:
This adjusts TLS protocol selection logic to leave TLSv1.3 enabled if
it's enabled in the underlying TLS stack. Prior to this change,
TLSv1.3 was disabled unless the user of okhttp explicitly enabled it
via a custom ConnectionSpec.

Fixes #2867

### DIFF
--- a/okhttp/src/main/java/okhttp3/ConnectionSpec.java
+++ b/okhttp/src/main/java/okhttp3/ConnectionSpec.java
@@ -31,7 +31,7 @@ import static okhttp3.internal.Util.intersect;
  * connection.
  *
  * <p>The TLS versions configured in a connection spec are only be used if they are also enabled in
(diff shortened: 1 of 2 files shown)


swankjesse:
Thanks!

Issue #2881, opened 2016-09-25
Details about ConnectException


Tolriq:
What kind of issue is this?

 - [X] Feature Request. Start by telling us what problem you’re trying to solve. Often a solution
       already exists! Don’t send pull requests to implement new features without first getting our
       support. Sometimes we leave features out on purpose to keep the project small.

I've got a few users that logs ConnectException from okhttp , but the underlying detail is hidden by okhttp implementation.

https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/internal/connection/RealConnection.java#L189

Is there a way to get more details like returning underlying error ?
Most of the time I get proper IOException with details like timeout / refused that help to give clue to the user.
But with this one I have no idea what the root cause can be.

swankjesse:
Send a pull request to set the cause?

Tolriq:
@swankjesse #2882

PR #2882, opened 2016-09-25
Keep the cause when sending ConnectException

Tolriq:
Proposed fix for Issue #2881 

Keep the cause when sending ConnectException


swankjesse:
Thanks! Merged manually.

Tolriq:
@swankjesse for my culture as I'm still java noob.

Any reason for temp object versus the cast ?

I hesitated but went with what is shown in Javadoc for initCause.

swankjesse:
The only difference is in style. I don’t like single-line `if` statements to be wrapped.

Tolriq:
Ok thanks.

PR #2887, opened 2016-09-29
Add stripQuotes in JavaNetCookieJar

leeight:


leeight:
@swankjesse

swankjesse:
I spent an hour tonight studying the spec, the implementation, and the alternative implementations. Here’s what I learned:

 * Quotes aren’t special characters in modern cookies. It’s a bug in `java.net.HttpCookie` code that quotes are added.
 * Quotes are added if the cookie parser believes it’s looking at a version=1 cookie. In practice this is any cookie that has a max-age attribute. It’s also version=1 if you call `new HttpCookie()`.
 * There’s no good way to know if quotes were added by `java.net.HttpCookie` of if they were already present in the cookie from the server.

I think our best long-term action is to ditch `JavaNetCookieJar` and write our own interface for storing cookies. The baggage in making the `java.net.HttpCookie` work well is just too high.

In the short term, this PR isn’t great. If we offer this option people will use it, and they won’t understand it. This is totally an expert option, and we have no good guidance on how to configure this. What would our Javadoc say, _“Toggle this if your cookies aren’t working.”_ Especially since this setting might work for one origin server and not for another.

I’m going to reject this PR and recommend you use this replacement instead:
https://gist.github.com/swankjesse/e283f2ddf264abd9a0d45bc80e7ec2d8

We need to build a successor to JavaNetCookieJar that doesn’t have this problem. We’ll track that work here: https://github.com/square/okhttp/issues/2890

manquius:
Using jetty 2.22.2 if I send a cookie with a single double quote I get an error 500
curl -v  http://localhost:8080/myApp/ -b A="\""

Could you please confirm if this was fixed on a newer version?

Issue #2756, opened 2016-07-28
okhttp 3.4.1will go endless loop,when there is a IOException("shutdown")


leokelly:
here is how it go endless loop:
in file RetryAndFollowUpInterceptor.java 

```
public Response intercept(Chain chain) throws IOException {

Request request = chain.request();
streamAllocation = new StreamAllocation(
    client.connectionPool(), createAddress(request.url()));
int followUpCount = 0;
Response priorResponse = null;
while (true) {
  if (canceled) {
    streamAllocation.release();
    throw new IOException("Canceled");
  }

  Response response = null;
  boolean releaseConnection = true;
  try {
    response = ((RealInterceptorChain) chain).proceed(request, streamAllocation, null, null);
    releaseConnection = false;
  } catch (RouteException e) {
    // The attempt to connect via a route failed. The request will not have been sent.
    if (!recover(e.getLastConnectException(), true, request)) throw e.getLastConnectException();
    releaseConnection = false;
    continue;
  } catch (IOException e) {
    // An attempt to communicate with a server failed. The request may have been sent.
    if (!recover(e, false, request)) throw e;
    releaseConnection = false;
    continue;
  } finally {
    // We're throwing an unchecked exception. Release any resources.
    if (releaseConnection) {
      streamAllocation.streamFailed(null);
      streamAllocation.release();
    }
  }
```

when there is a  IOException("shutdown") when execute 
`response = ((RealInterceptorChain) chain).proceed(request, streamAllocation, null, null);`
it will go into catch block,and judge if can recover,`recover(e, false, request)` unfortunately recover() return true,so it will continue and go endless loop

swankjesse:
Can you provide a test case to demonstrate?

mopsalarm:
I had the same issue today too. The loop ran endlessly for minutes until I killed the app. Maybe put an upper bound to the number of executions or at least slow down after 10 iterations?

swankjesse:
We gotta make sure shutdown connections get removed from the pool.

dave-r12:
@swankjesse were you planning to pick this one up? If not I can look into it.

swankjesse:
Please do!

fabioCollini:
I have a similar issue using http2, I tried to debug the code and I have found this TODO in method `isHealthy` of `RealConnection` class:

```
if (framedConnection != null) {
  return true; // TODO: check framedConnection.shutdown.
}
```

The problem seems to be related to this TODO.

[...]

PR #2889, opened 2016-10-03
Improve management of HTTP/2 connections that are shutdown.

dave-r12:
Closes: https://github.com/square/okhttp/issues/2756

### DIFF
--- a/mockwebserver/src/main/java/okhttp3/mockwebserver/MockResponse.java
+++ b/mockwebserver/src/main/java/okhttp3/mockwebserver/MockResponse.java
@@ -43,6 +43,9 @@ public final class MockResponse implements Cloneable {
   private long bodyDelayAmount = 0;
   private TimeUnit bodyDelayUnit = TimeUnit.MILLISECONDS;
 
+  private long shutdownDelay = 0;
+  private TimeUnit shutdownDelayUnit = TimeUnit.MILLISECONDS;
+
   private List<PushPromise> promises = new ArrayList<>();
   private Settings settings;
   private WebSocketListener webSocketListener;
@@ -251,6 +254,21 @@ public final class MockResponse implements Cloneable {
 
   public long getBodyDelay(TimeUnit unit) {
     return unit.convert(bodyDelayAmount, bodyDelayUnit);
+  }
+
+  /**
+   * Set the time to {@code delay} prior to shutting down an HTTP/2 connection. This is only valid
+   * with {@link SocketPolicy#DISCONNECT_AT_END}. Use this to simulate a graceful HTTP/2
+   * connection shutdown.
+   */
+  public MockResponse setShutdownDelay(long delay, TimeUnit unit) {
+    shutdownDelay = delay;
+    shutdownDelayUnit = unit;
+    return this;
+  }
+
+  public long getShutdownDelay(TimeUnit unit) {
+    return unit.convert(shutdownDelay, shutdownDelayUnit);
   }
 
   /**
--- a/mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.java
+++ b/mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.java
@@ -369,6 +369,7 @@ public final class MockWebServer implements TestRule, Closeable {
             logger.info(MockWebServer.this + " done accepting connections: " + e.getMessage());
             return;
           }
+          logger.info("accepted connection on port " + socket.getPort());
           SocketPolicy socketPolicy = dispatcher.peek().getSocketPolicy();
           if (socketPolicy == DISCONNECT_AT_START) {
             dispatchBookkeepingRequest(0, socket);
@@ -876,6 +877,17 @@ public final class MockWebServer implements TestRule, Closeable {
       if (logger.isLoggable(Level.INFO)) {
         logger.info(MockWebServer.this + " received request: " + request
             + " and responded: " + response + " protocol is " + protocol.toString());
+      }
+
+      if (response.getSocketPolicy() == DISCONNECT_AT_END) {
+        Http2Connection connection = stream.getConnection();
+        connection.shutdown(ErrorCode.NO_ERROR);
+        try {
+          Thread.sleep(response.getShutdownDelay(TimeUnit.MILLISECONDS));
+        } catch (InterruptedException e) {
+          throw new InterruptedIOException();
+        }
+        connection.close();
       }
     }
 
--- a/mockwebserver/src/main/java/okhttp3/mockwebserver/SocketPolicy.java
+++ b/mockwebserver/src/main/java/okhttp3/mockwebserver/SocketPolicy.java
@@ -37,7 +37,10 @@ public enum SocketPolicy {
   KEEP_OPEN,
 
   /**
-   * Close the socket after the response. This is the default HTTP/1.0 behavior.
+   * Close the socket after the response. This is the default HTTP/1.0 behavior. For HTTP/2
+   * connections, send a <a href="https://tools.ietf.org/html/rfc7540#section-6.8">GOAWAY frame</a>
+   * immediately after the response, then close the connection. The connection will be closed after
+   * delaying for {@link MockResponse#getShutdownDelay(java.util.concurrent.TimeUnit)}.
    *
    * <p>See {@link SocketPolicy} for reasons why this can cause test flakiness and how to avoid it.
    */
--- a/okhttp-tests/src/test/java/okhttp3/URLConnectionTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/URLConnectionTest.java
@@ -3530,6 +3530,40 @@ public final class URLConnectionTest {
     assertEquals(0, dispatcher.runningCallsCount());
   }
 
+  @Test public void streamedBodyIsRetriedOnHttp2Shutdown() throws Exception {
+    enableProtocol(Protocol.HTTP_2);
+    server.enqueue(new MockResponse()
+        .setSocketPolicy(SocketPolicy.DISCONNECT_AT_END)
+        .setShutdownDelay(1, TimeUnit.SECONDS)
+        .setBody("abc"));
+    server.enqueue(new MockResponse()
+        .setBody("def"));
+
+    HttpURLConnection connection1 = urlFactory.open(server.url("/").url());
+    connection1.setChunkedStreamingMode(4096);
+    connection1.setRequestMethod("POST");
+    connection1.connect(); // Establish healthy HTTP/2 connection, but don't write yet.
+
+    // Send a separate request which will trigger a GOAWAY frame on the healthy connection.
+    HttpURLConnection connection2 = urlFactory.open(server.url("/").url());
+    assertContent("abc", connection2);
+
+    // Delay to ensure the GOAWAY frame is processed.
+    Thread.sleep(250);
+
+    OutputStream os = connection1.getOutputStream();
+    os.write(new byte[] { '1', '2', '3' });
+    os.close();
+    assertContent("def", connection1);
+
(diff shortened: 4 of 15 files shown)


mockwebserver/src/main/java/okhttp3/mockwebserver/MockResponse.java
```diff
@@ -43,6 +43,9 @@
   private long bodyDelayAmount = 0;
   private TimeUnit bodyDelayUnit = TimeUnit.MILLISECONDS;
 
+  private long shutdownDelay = 1;
```
swankjesse: Typically we just store milliseconds in the fields. The setters can take (duration, unit) and we can convert to milliseconds.

mockwebserver/src/main/java/okhttp3/mockwebserver/MockResponse.java
```diff
@@ -254,6 +257,21 @@ public long getBodyDelay(TimeUnit unit) {
...
+   * with {@link SocketPolicy#GRACEFUL_SHUTDOWN_AT_END}. Use this to simulate a graceful HTTP/2
+   * connection shutdown.
+   */
+  public MockResponse setShutdownDelay(long delay, TimeUnit unit) {
+    shutdownDelay = delay;
+    shutdownDelayUnit = unit;
+    return this;
+  }
+
+  public long getShutdownDelay(TimeUnit unit) {
```
swankjesse: ... the upside of just storing milliseconds in the field is that accessors don’t need to do this awkward thing, where precision can be lost.

... though this _is_ consistent with java.util.concurrent...
https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/Delayed.html#getDelay(java.util.concurrent.TimeUnit)

swankjesse: This is pretty curious stuff to me. If we do this, we might want to add symmetrical accessors to OkHttpClient for its timeouts

okhttp-tests/src/test/java/okhttp3/URLConnectionTest.java
```diff
@@ -3530,6 +3530,39 @@ private void zeroLengthPayload(String method)
...
+        .setBody("abc"));
+    server.enqueue(new MockResponse()
+        .setBody("def"));
+
+    HttpURLConnection connection1 = urlFactory.open(server.url("/").url());
+    connection1.setChunkedStreamingMode(4096);
+    connection1.setRequestMethod("POST");
+    connection1.connect(); // Establish healthy HTTP/2 connection, but don't write yet.
+
+    // Send a separate request which will trigger a GOAWAY frame on the healthy connection.
```
swankjesse: neat

swankjesse: We’d do this in CallTest by making the 2nd call in the 1st call’s callback, or in its network interceptor

okhttp/src/main/java/okhttp3/internal/http/RetryAndFollowUpInterceptor.java
```diff
@@ -219,6 +220,10 @@ private boolean recover(IOException e, boolean routeException, Request userReque
     return true;
   }
 
+  private boolean isBodyUnrepeatable(IOException e, RequestBody body) {
```
swankjesse: I’m a little reluctant to put these two checks into a single method. The method name suggests that it has nothing to do with which exception was thrown.

I think the goal is to detect that request sending never started, which is good. Maybe we could do that by renaming `routeException` in the calling code to `requestSendStarted` ? Previously that was 1:1 with route exceptions and now we’ve got this extra case that seems similar.

mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.java
```diff
@@ -877,6 +878,17 @@ private FramedSocketHandler(Socket socket, Protocol protocol) {
         logger.info(MockWebServer.this + " received request: " + request
             + " and responded: " + response + " protocol is " + protocol.toString());
       }
+
+      if (response.getSocketPolicy() == GRACEFUL_SHUTDOWN_AT_END) {
```
swankjesse: Dig.

mockwebserver/src/main/java/okhttp3/mockwebserver/SocketPolicy.java
```diff
@@ -94,5 +94,16 @@
...
+  /**
+   * For HTTP/2 connections, send a
+   * <a href="https://tools.ietf.org/html/rfc7540#section-6.8">GOAWAY frame</a> immediately after
+   * the response, then close the connection. The connection will be closed after delaying for
+   * {@link MockResponse#getShutdownDelay(java.util.concurrent.TimeUnit)}. Use this to simulate a
+   * graceful HTTP/2 connection shutdown.
+   *
+   * <p>See {@link SocketPolicy} for reasons why this can cause test flakiness and how to avoid it.
+   */
+  GRACEFUL_SHUTDOWN_AT_END
```
swankjesse: Should this just be `DISCONNECT_AT_END,` when the delay is included?

okhttp/src/main/java/okhttp3/internal/connection/StreamAllocation.java
```diff
@@ -298,6 +299,8 @@ public void streamFailed(IOException e) {
           noNewStreams = true;
           route = null;
         }
+      } else if (e instanceof ConnectionShutdownException) {
+        noNewStreams = true;
```
swankjesse: Nice.

swankjesse: I’m wondering if the case below which is special cased for multiplexed connections actually shouldn’t be special cased that way? Maybe instead it should be like this?

```
  noNewStreams = !(e instanceof StreamResetException);
```

swankjesse: Or maybe even this:
```
noNewStreams = !(e instanceof StreamResetException)
    || ((StreamResetException) e).errorCode != ErrorCode.CANCEL;
```
ie. by default everything kills the connection _except_ for stream canceling, which we know is safe.

mockwebserver/src/main/java/okhttp3/mockwebserver/MockResponse.java
dave-r12: I like the idea of storing the milliseconds. I do not have a good reason for doing it this way other than that's how the body delay works. Any strong preference?

mockwebserver/src/main/java/okhttp3/mockwebserver/SocketPolicy.java
dave-r12: Good thought. So a "graceful" shutdown would be a combination of `DISCONNECT_AT_END` + setting the shutdown delay? Then there would be no need for this value.

okhttp-tests/src/test/java/okhttp3/URLConnectionTest.java
dave-r12: Gotcha, I will add a test case to CallTest using an interceptor that stalls, similar to the OkHttpURLConnection.

okhttp/src/main/java/okhttp3/internal/connection/StreamAllocation.java
dave-r12: I've thought about what you said, but I'm still uncertain on this. One counterexample I came up with, what if an app layer network interceptor throws an unchecked exception? Should that kill the connection? I believe the current code will preserve an HTTP/2 connection (I like that), but kill an HTTP/1.1 connection.

dave-r12:
I made revisions attempting to incorporate the feedback. Hopefully it's in better shape, but will likely need some adjustments.

mockwebserver/src/main/java/okhttp3/mockwebserver/MockResponse.java
swankjesse: Nope!

swankjesse: (this is fine.)

okhttp-tests/src/test/java/okhttp3/URLConnectionTest.java
```diff
@@ -3530,6 +3530,40 @@ private void zeroLengthPayload(String method)
...
+    os.write(new byte[] { '1', '2', '3' });
+    os.close();
+    assertContent("def", connection1);
+
+    RecordedRequest request1 = server.takeRequest();
+    assertEquals(0, request1.getSequenceNumber());
+
+    RecordedRequest request2 = server.takeRequest();
+    assertEquals("123", request2.getBody().readUtf8());
+    assertEquals(0, request2.getSequenceNumber());
```
swankjesse: Nice.

swankjesse:
Travis reports a failure in `CertificatePinnerChainValidationTest.pinIntermediatePresentInChain`.

dave-r12:
Ahh weird, seems to run locally. I'm going to try a rebuild.

swankjesse:
On the Travis failures, I think `CertificatePinnerChainValidationTest` had old expectations for how `DISCONNECT_AT_END` would work, and those are invalid now.

How about instead, remove the `DISCONNECT_AT_END` thing from the response and then call this instead of sleeping?
```
    // Force a fresh connection for the next request.
    client.connectionPool().evictAll();
```

That way we get a fresh connection (which we want) without fighting the HTTP/2 stuff too hard.

dave-r12:
Thanks (and apologies if this was causing noise.) I'm trying to get a handle on why the tests are failing. One of them is intuitive: it's killing the connection between the end of request1 and the start of request2. But there is a separate failure that isn't as clear. I think it's a combination of multiple threads + closing the socket + trying to write to the closed socket. It's a bit of pain because I can't reproduce it locally! If you don't mind, I'm going to continue looking at this.

My fear is that someone who uses `DISCONNECT_AT_END` with HTTP/2 and does not set a shutdown delay will experience this exact behavior.

swankjesse:
Definitely. Tracking as #2918. My guess is that it’s a bug in our client!

dave-r12:
I was able to reproduce the problem with a vanilla Java application outside of OkHttp. The behavior I was seeing seemed incorrect and my knowledge of the JDK code is limited. I opened a bug here to track that issue: https://bugs.openjdk.java.net/browse/JDK-8168513. I'll wait to see what their thoughts are.

swankjesse:
Wow, good find.

swankjesse:
After some digging I was able to reproduce this problem without enabling SSL.

This problem occurs whenever a server closes its output stream before it has consumed everything from the client’s input stream. I was able to reproduce this with a simple Java server and `nc`.

dave-r12:
Interesting, was it linux? macOS? The example I came up with only seemed to trigger on linux, and only when TLS was in the mix. Commenting out the SSL stuff appeared to make the problem go away. Or rather, it allowed the 'client' to drain the receive buffer prior to getting a connection reset.

I've continued to study this but it's been slow going. Part of me now believes linux is doing the right thing. I found this in RFC793:
> If the RST bit is set then, any outstanding RECEIVEs and SEND
        should receive "reset" responses.  All segment queues should be
        flushed.  Users should also receive an unsolicited general
        "connection reset" signal.  Enter the CLOSED state, delete the
        TCB, and return.

Anyways, I'm tempted to abandon this approach. I'd like to get a fix in, but the nondeterminism with the changes here makes me nervous.

swankjesse:
I’ve stolen the tests from this PR and put them here:
https://github.com/square/okhttp/pull/2957

The last thing we should do is recover gracefully when a connection is shutdown. Currently we don’t do that very well. I’d like to do that in a separate PR, possibly borrowing the new exception types defined here.

PR #2893, opened 2016-10-04
RFC: frameLogger

yschimke:
I'd like to be able to customise how to deal with frame events.  But logging to string is a bit restrictive.

Putting this up as a dreadful strawman to see if there is any prevailing thoughts here?


swankjesse:
We’ve used `internal` as a lousy SPI for the file system. That’s ugly, but workable. Doing similarly for analytics events might be a good way to help that interface to mature.

What if the interface took frame data and was responsible for calling into the logger - or not? That'd be interesting at least.

yschimke:
> What if the interface took frame data and was responsible for calling into the logger - or not? That'd be interesting at least.

I'm happy to put up a PR for discussion following that approach, and leave the logging implementation as the default.  I'd put the interface and the set method behind the internal packages.

swankjesse:
@yschimke SGTM!

swankjesse:
I’ll close this. Please open a separate PR when your FrameLogger is ready.

PR #2898, opened 2016-10-10
Allow customize what types of body can be logged based on Content-Type

mingchen:
Allow customize what types of body can be logged based on Content-Type. 

The client can define its own list of Content-Type want to be list, default only log text and json.

The old implementation may log binary content, for example I have multipart content type, the first part is text, the second part is a binary file, it will log the binary in this case. With this new enhancement it will not log multipart content.


swankjesse:
I’m pretty reluctant on this. Our logger is not intended to be a general purpose system; instead it’s just a starter example. On this change we’re opening ourselves to maintaining a registry of human-readable content types; that’s a decent amount of work.

Instead I think it’s better to build a separate, fancy logger, possibly as a standalone project. For example, [OkLog](https://github.com/simonpercic/OkLog) is pretty cool, and supports some fancy things. If you’d like to do your own deluxe logging interceptor I’m sure you could get others interested.

mingchen:
I think there is no need maintenance the human readable content types in okhttp. To allow user log specific content type, the constructer can also accept content types that user want to log.

The reason I submit this pull request I use this logger and I have a multipart data, the last part contents binary data and it not detected by properly. I hope this pull request can solve this kind of this issue, do not print binary data in logcat.

Issue #2902, opened 2016-10-12
WebSocket API Review


swankjesse:
I went through the WebSocket API and wrote down my thoughts. Nothing too earth-shattering, just some ideas.
1. How should WebSockets interact with the Dispatcher? In particular should open web sockets count towards `maxRequestsPerHost` ? (I _think_ yes, but we should explicitly call this out.)
2. When can I cancel a WebSocketCall? Once it’s open is that equivalent to calling `WebSocket#close()` ?
3. Does `isCanceled()` return true after a WebSocket is closed?
4. ~~Do we throw an IllegalStateException if you write the socket in `onOpen()` ? What about  `sendPing()` and `close()` ?~~ #2905
5. It looks like pings and the corresponding pongs are very loosely coupled. Any idea how strict the spec is? Seems weird to call out that pongs can come unsolicited. I’m really unsure about what to do with pings and pongs. My first instinct is to hide them from the API and instead offer something to send pings automatically for you on a schedule
6. I wonder if there’s a better name for `WebSocketListener`. With `Call` we have `Callback` and the names are cute. Is `WebSocketCallback` terrible?
7. An obnoxious idea – would it ever make sense to combine the `WebSocketCall` and `WebSocket` types? I’m getting here via the (slight) redundancy between `cancel()` and `close()`. Obviously `sendMessage()` would need to be a bit more clever to do something reasonable if it were invoked before `onOpen()` is received. Though it’s interesting to think about “fast open” where a web socket client can enqueue messages before we have confirmation that it’s all going to connect.
8. ~~Should it be `sendMessage(RequestBody)` or just `send(RequestBody)` or just `message(RequestMessage)` ? Maybe instead of `sendPing(ByteString)` it should be just `ping(ByteString)` or even `ping()` ?~~ #2904

JakeWharton:
1. Sure
2. Before `open` only. We could send a "going away" close if it was opened.
3. No. Only on user cancel currently.
4. Ideally we disallow writes on the reader thread entirely. currently you can write in onOpen and any listener callback
5. We can hide the API iff we ship with auto-pings. i'm still willing to bet exposing it is the first issue filed after the release though.
6. I think we decided listener because it was long-lived instead of a callback which is a single invocation
7. Wasn't the initial version like this? I can't remember. I think it would ISE if you tried to do something before connected. My biggest concern is that no matter what the application has to wait for the open callback so why not prevent them from sending anything with the type system instead of with runtime checks.
[...]

PR #2904, opened 2016-10-12
Remove inconsistent 'send' prefix from message/ping methods.

JakeWharton:
Refs #2902.


AlexBonel:
@JakeWharton Just a question for sake of understanding. I don't get the omission of **send** from `message`.  `message` is not a verb and in this context  it is implied that a message is literally *sent* through WebSocket. Moreover, often using of nouns in methods signatures could make the reader think that this is a function which returns some value and doesn't perform any actions.

`ping` on the other hand is much more self descriptive.

PR #2905, opened 2016-10-12
Forbid writes from reader thread.

JakeWharton:
Refs #2902.


okhttp/src/main/java/okhttp3/internal/ws/RealWebSocket.java
```diff
@@ -272,10 +286,13 @@ private void replyToReaderError(final Throwable t) {
   }
 
   /** Send an unsolicited pong with the specified payload. */
-  final void sendPong(ByteString payload) throws IOException {
+  public final void pong(ByteString payload) throws IOException {
```
JakeWharton: Note: not actually a public API since this is internal

PR #2914, opened 2016-10-14
Generic body for both Request and Responses

mescortes:


okhttp-apache/src/main/java/okhttp3/apache/OkApacheClient.java
```diff
@@ -80,8 +79,14 @@ private static HttpResponse transformResponse(Response response) {
...
     BasicHttpResponse httpResponse = new BasicHttpResponse(HTTP_1_1, code, message);
 
-    ResponseBody body = response.body();
-    InputStreamEntity entity = new InputStreamEntity(body.byteStream(), body.contentLength());
+    Body body = response.body();
+    long contentLength = 0;
+    try {
+      contentLength = body.contentLength();
+    } catch (IOException e) {
+      e.printStackTrace();
```
JakeWharton: 👀

okhttp/src/main/java/okhttp3/Body.java
```diff
@@ -0,0 +1,263 @@
...
+    this.readableContent = readableContent;
+  }
+
+  public MediaType contentType() {
+    return contentType;
+  }
+
+  public void writeTo(BufferedSink sink) throws IOException {
+    if (writableContent == null) throw new UnsupportedOperationException("Not writable");
+    sink.writeAll(new Buffer().write(writableContent));
```
JakeWharton: You shouldn't need the additional `Buffer`. Just `sink.writeAll(writeableContent)`.

okhttp/src/main/java/okhttp3/Body.java
```diff
@@ -0,0 +1,263 @@
...
+  public Body() {
+    this(null, -1, null, null);
+  }
+
+  public Body(MediaType contentType) {
+    this(contentType, -1, null, null);
+  }
+
+  public Body(MediaType contentType, byte[] content) {
+    this(contentType, content.length, content, new Buffer().write(content));
```
JakeWharton: This currently makes `byte[]` bodies one-shot types instead of being able to use them multiple times.

okhttp/src/main/java/okhttp3/Body.java
```diff
@@ -0,0 +1,263 @@
...
+  }
+
+  public Body(MediaType contentType, Buffer content) {
+    this(contentType, content.size(), content.readByteArray(), content);
```
JakeWharton: The fourth argument becomes an empty buffer because you exhaust it creating the third argument.

okhttp/src/main/java/okhttp3/Body.java
```diff
@@ -0,0 +1,263 @@
...
+   */
+  public final byte[] bytes() throws IOException {
+    long contentLength = contentLength();
+    if (contentLength > Integer.MAX_VALUE) {
+      throw new IOException("Cannot buffer entire body for content length: " + contentLength);
+    }
+
+    byte[] bytes;
+    try {
+      bytes = source().readByteArray();
```
JakeWharton: This method requires a readable source but could support writeable bodies by writing to `new Buffer()` and then calling `readByteArray()`.

okhttp/src/main/java/okhttp3/Body.java
```diff
@@ -0,0 +1,263 @@
...
+
+  public Body(MediaType contentType, BufferedSource content) {
+    this(contentType, -1, null, content);
+  }
+
+  public Body(MediaType contentType, long contentLength, BufferedSource content) {
+    this(contentType, contentLength, null, content);
+  }
+
+  public Body(MediaType contentType, long contentLength, byte[] writableContent,
```
JakeWharton: Way too many constructors _and_ static factories. I liked the old setup where static factories mean we handle everything for you from known types and subclasses opt-in to handling everything that's required.

okhttp-android-support/src/main/java/okhttp3/internal/huc/JavaApiConverter.java
```diff
@@ -505,51 +504,30 @@ static String extractStatusLine(Map<String, List<String>> javaResponseHeaders)
...
-        String s = urlConnection.getHeaderField("Content-Length");
-        return stringToLong(s);
-      }
-
-      @Override public BufferedSource source() {
-        return body;
-      }
-    };
+    return new Body(urlConnection.getContentType() != null
+        ? MediaType.parse(urlConnection.getContentType())
```
swankjesse: I think I’d prefer these to be computed lazily, not eagerly

swankjesse: ... if the media type is invalid, and nobody attempts to parse it, no big deal. This is relevant in particular for this: `text/plain; charset=utf-8; charset=ascii`

okhttp-tests/src/test/java/okhttp3/MultipartBodyTest.java
```diff
@@ -192,19 +191,13 @@
...
-      }
-
-      @Override public MediaType contentType() {
-        return null;
+        super(null, new Buffer().writeUtf8(body));
       }
 
-      @Override public void writeTo(BufferedSink sink) throws IOException {
-        sink.writeUtf8(body);
+      @Override public long contentLength() throws IOException {
```
swankjesse: this one’s a bit interesting

okhttp-tests/src/test/java/okhttp3/MultipartBodyTest.java
```diff
@@ -192,19 +191,13 @@
...
+        return -1;
```
swankjesse: supernit: throughout this project I’ve been trying to suffix longs like `-1L`

okhttp/src/main/java/okhttp3/Body.java
```diff
@@ -0,0 +1,263 @@
...
+import static okhttp3.internal.Util.UTF_8;
+
+/**
+ * The body for HTTP requests or responses.
+ */
+public class Body implements Closeable {
+  public static final byte[] EMPTY_ARRAY = new byte[0];
+  public static final Body EMPTY = new Body(null, EMPTY_ARRAY);
+
+  protected final MediaType contentType;
```
swankjesse: not sure how I feel about the visibility of these. Protected is weird.

okhttp/src/main/java/okhttp3/Body.java
swankjesse: I think we probably want `byte[]` to be one-shot for reading and many-shots for writing

okhttp/src/main/java/okhttp3/Body.java
swankjesse: Yeah, I wonder if we should even try to make streaming bodies do anything clever here.

Maybe there’s actually 3 kinds?
 - streaming readable
 - streaming writable
 - fully-buffered

okhttp/src/main/java/okhttp3/Cache.java
```diff
@@ -733,41 +733,20 @@ private static int readInt(BufferedSource source) throws IOException {
...
     private final DiskLruCache.Snapshot snapshot;
-    private final BufferedSource bodySource;
-    private final String contentType;
-    private final String contentLength;
 
-    public CacheResponseBody(final DiskLruCache.Snapshot snapshot,
-        String contentType, String contentLength) {
+    public CacheResponseBody(final DiskLruCache.Snapshot snapshot, String contentType,
+        String contentLength) {
+      super(contentType != null ? MediaType.parse(contentType) : null,
```
swankjesse: I nolove this. I think the other form with abstract methods was easier to reason about.

okhttp/src/main/java/okhttp3/ReadableBody.java
```diff
@@ -0,0 +1,41 @@
...
+public interface ReadableBody extends Closeable {
+  MediaType contentType();
+
+  long contentLength() throws IOException;
+
+  BufferedSource source();
+
+  byte[] bytes() throws IOException;
+
+  Body peek(long byteCount) throws IOException;
```
swankjesse: Not sure how I feel about this one either.

samples/guide/src/main/java/okhttp3/recipes/PostStreaming.java
```diff
@@ -1,4 +1,5 @@
 /*
+
```
swankjesse: nit

okhttp/src/main/java/okhttp3/Body.java
mescortes: If the body has `static` content we should allow multiple read/write.

I think I have a better implementation for this. Let me give it another try ;-)

okhttp/src/main/java/okhttp3/Body.java
mescortes: Agree. I will change it.

swankjesse:
Another PR is forthcoming.

Issue #2780, opened 2016-08-08
Support for WWW-Authenticate headers which does not strictly follow RFC 2617


alexeyvasilyev:
OkHttp expects WWW-Authenticate header to have **realm** parameter going right after **Digest** token in HttpHeaders.parseChallenges(Headers responseHeaders, String challengeHeader) method, e.g.
`WWW-Authenticate: Digest realm="DS-2CD2432F-IW", nonce="4d6a6c4551305a43515449364e6d497a4d6a63314e546b3d", qop="auth", stale="FALSE"`

 However some web servers (e.g. widely used Hikvision IP cameras) can send parameters in different order, e.g.

`WWW-Authenticate: Digest qop="auth", realm="DS-2CD2432F-IW", nonce="4d6a6c4551305a43515449364e6d497a4d6a63314e546b3d", stale="FALSE"`

This situation is not handled properly in OkHttp.

UNIT TEST:

`

    private static void test() {
        // Strict RFC 2617 header
        Headers headers = new Headers.Builder()
                .add("WWW-Authenticate", "Digest realm=\"DS-2CD2432F-IW\", nonce=\"4d6a6c4551305a43515449364e6d497a4d6a63314e546b3d\", qop=\"auth\", stale=\"FALSE\"").build();

        Assert.assertTrue(HttpHeaders.parseChallenges(headers, "WWW-Authenticate").size() == 1);

        // Not strict RFC 2617 header. Generates error in OkHttp 3.4.1.
        headers = new Headers.Builder()
                .add("WWW-Authenticate", "Digest qop=\"auth\", realm=\"DS-2CD2432F-IW\", nonce=\"4d6a6c4551305a43515449364e6d497a4d6a63314e546b3d\", stale=\"FALSE\"").build();
        Assert.assertTrue(HttpHeaders.parseChallenges(headers, "WWW-Authenticate").size() == 1);
    }
`

alexeyvasilyev:
Got the same issue with widely used IP Webcam app version 1.12.4r
https://play.google.com/store/apps/details?id=com.pas.webcam

GET /shot.jpg HTTP/1.1
User-Agent: Android/6.0 (6.0 - 5X) com.alexvas.dvr.pro
Host: 10.0.1.4:8080
Connection: Keep-Alive
Accept-Encoding: gzip

HTTP/1.1 401 Unauthorized
Content-Length: 0
WWW-Authenticate: Digest qop="auth", realm="IP Webcam", nonce="1470841389"

[...]

PR #2921, opened 2016-10-17
Parse authentication headers with wrong order

drlue:
HttpHeaders.parseChallenges(...) changed to parse also wrong ordered authentication headers.

Authentication headers with wrongly ordered parameters not following RFC 2617 were not recognized.
See: https://github.com/square/okhttp/issues/2780

Probably you can find this usefull. If you are already working on a more feature rich Challenges version just forget the pull request :-)

Thanks for the great library.
regards
Lukas

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/http/HttpHeaders.java
+++ b/okhttp/src/main/java/okhttp3/internal/http/HttpHeaders.java
@@ -38,8 +38,7 @@ import static okhttp3.internal.http.StatusLine.HTTP_CONTINUE;
 
 /** Headers and utilities for internal use by OkHttp. */
 public final class HttpHeaders {
-  private static final Pattern AUTHENTICATION_HEADER_PATTERN
-          = Pattern.compile("(.*?) .*?realm=\"(.*?)\"", Pattern.CASE_INSENSITIVE);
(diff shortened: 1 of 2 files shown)


okhttp/src/main/java/okhttp3/internal/http/HttpHeaders.java
```diff
@@ -35,6 +38,9 @@
 
 /** Headers and utilities for internal use by OkHttp. */
 public final class HttpHeaders {
+  private static final Pattern AUTHENTICATION_HEADER_PATTERN
+          = Pattern.compile("(.*?) .*?realm=\"(.*?)\"", Pattern.CASE_INSENSITIVE);
```
swankjesse: It’s academic but I think this regex is too loose. Is this possible?

```
Digest underrealm="a" realm="b"
```

swankjesse: I do like how much smaller this code is though. Nice.

drlue: Yes your definitively right, please see the new commit.

swankjesse:
Merged manually after tweaking the regex.

swankjesse:
Thanks @drlue !

PR #2925, opened 2016-10-18
Fixed incorrect comment from #2883

### LINKED ISSUE #2883: Bug: Response.body() returning null
According to [this comment](https://github.com/square/okhttp/blob/6014ab9752cff344a0baf57eb7626d98ca1d3401/okhttp/src/main/java/okhttp3/Response.java#L168) the `Response.body()` method may never return null. 

As can be seen in this [gist](https://gist.github.com/JemarJones/fcce225de6a4d59025c4011f4f06395b), there are cases in which this method can return null.

I think this comment either needs to be updated to be more accurate, or the functionality should be updated to match the claim made in the comment.

JemarJones:
As in the commit message, this fixes #2883.

### DIFF
--- a/okhttp/src/main/java/okhttp3/Response.java
+++ b/okhttp/src/main/java/okhttp3/Response.java
@@ -165,7 +165,7 @@ public final class Response implements Closeable {
   }
 
   /**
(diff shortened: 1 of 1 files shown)


swankjesse:
Thanks! Merged manually.

PR #2926, opened 2016-10-19
Cache POST requests with OkHttp

romantic011:
i  have got a question,how can i cache post request with OkHttp,


swankjesse:
Thanks for your question. This issue tracker is only for bug reports with test cases and feature requests. Please ask usage questions on Stack Overflow.
https://stackoverflow.com/questions/tagged/okhttp

romantic011:
tks

PR #2935, opened 2016-10-22
A WebSockets sample that uses the Slack API.

swankjesse:
Currently it just prints WebSockets messages as they arrive, never sending any.
In a follow-up I'd like to do that, and also start to use this to figure out
where and how our WebSockets client can be improved.

https://github.com/square/okhttp/issues/2902

### DIFF
--- a/samples/slack/src/main/java/okhttp3/slack/OAuthSessionFactory.java
+++ b/samples/slack/src/main/java/okhttp3/slack/OAuthSessionFactory.java
@@ -1,0 +1,126 @@
+/*
+ * Copyright (C) 2016 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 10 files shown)


samples/slack/src/main/java/okhttp3/slack/RtmSession.java
```diff
@@ -0,0 +1,89 @@
...
+  public void open(String accessToken) throws IOException {
+    if (webSocketCall != null) throw new IllegalStateException();
+
+    RtmStartResponse rtmStartResponse = slackApi.rtmStart(accessToken);
+    webSocketCall = slackApi.rtm(rtmStartResponse.url);
+    webSocketCall.enqueue(this);
+  }
+
+  // TODO(jwilson): can I read the response body? Do I have to?
+  //                the body from slack is a 0-byte-buffer
```
JakeWharton: We remove the body. What's passed is a 0-byte one so that  we don't violate the contract that it's non-null.

samples/slack/src/main/java/okhttp3/slack/OAuthSessionFactory.java
```diff
@@ -0,0 +1,126 @@
...
+ * <p>Clients should call {@link #start}, {@link #newAuthorizeUrl} and {@link #close} in that order.
+ * Clients may request multiple sessions.
+ */
+public final class OAuthSessionFactory extends Dispatcher implements Closeable {
+  private final SecureRandom secureRandom = new SecureRandom();
+
+  private final SlackApi slackApi;
+  private MockWebServer mockWebServer;
+
+  /** Guarded byt this. */
```
raskasa: Typo here?

swankjesse: Yep! Fixed!

Issue #2940, opened 2016-10-27
Bug Report: validateCloseCode() throws IllegalArgumentException for CLOSE_NO_STATUS_CODE


jparkie:
When the server closes by sending a Close control frame, and there is no body, I received the following exception:

```
java.lang.IllegalArgumentException: Code 1005 is reserved and may not be used.
at okhttp3.internal.ws.WebSocketProtocol.validateCloseCode(WebSocketProtocol.java:114)
at okhttp3.internal.ws.WebSocketWriter.writeClose(WebSocketWriter.java:101)
at okhttp3.internal.ws.RealWebSocket$2.execute(RealWebSocket.java:190)
at okhttp3.internal.NamedRunnable.run(NamedRunnable.java:32)
at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1113)
at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:588)
at java.lang.Thread.run(Thread.java:818)
```

I found this confusing because as stated in https://tools.ietf.org/html/rfc6455#section-5.5.1:

```
The Close frame MAY contain a body (the "Application data" portion of
the frame) that indicates a reason for closing, such as an endpoint
shutting down, an endpoint having received a frame too large, or an
endpoint having received a frame that does not conform to the format
expected by the endpoint.  If there is a body, the first two bytes of
the body MUST be a 2-byte unsigned integer (in network byte order)
representing a status code with value /code/ defined in Section 7.4.
```

Furthermore, as stated in https://tools.ietf.org/html/rfc6455#section-7.1.6:

```
As defined in Sections 5.5.1 and 7.4, a Close control frame may
contain a status code indicating a reason for closure
```

My understanding is that what my server sent was valid. 

Searching 1005, https://tools.ietf.org/html/rfc6455#section-7.4.1 states the following:

```
1005 is a reserved value and MUST NOT be set as a status code in a
Close control frame by an endpoint.  It is designated for use in
applications expecting a status code to indicate that no status
code was actually present.
```

I understand OkHttp should validate my server did not send 1005, but I sent nothing. Searching the code https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/internal/ws/WebSocketReader.java#L192-L205 assumes 1005 without checking first I sent nothing. I believe this is contrary to the behaviour outline in https://tools.ietf.org/html/rfc6455#section-7.1.5

```
[...]

PR #2941, opened 2016-10-27
Allow empty close control frames for WebSockets according to rfc6455

jparkie:
Closes: https://github.com/square/okhttp/issues/2940


swankjesse:
Hey @jparkie I’m finally looking at this, right after making big changes to our websocket internals. I changed things so that callers are now responsible sending close frames.

I’m thinking about how that impacts this change. I think what we want is this:

 * If the application layer calls `close()` with 1005, we interpret that to mean `0` — we just send a close frame without a code.
 * We encourage applications to always use 1000 when they close, rather than echoing the code from the peer.

PR #2955, opened 2016-11-02
Make HTTP/2's DISCONNECT_AT_END more graceful.

swankjesse:
Rather than immediately and violently closing the HTTP/2 stream,
we initiate the shutdown sequence for a graceful disconnect.

This fixes StreamAllocation to avoid creating new streams on
shutdown connections. That behavior is not well tested, however.
That will come by integrating dave-r12's PR:

See https://github.com/square/okhttp/pull/2889


swankjesse:
@dave-r12 this is the miniature version of #2889. We don’t attempt to violently close the socket, and all of our nondeterminism and strangeness goes away.

If this passes tests (I really hope it does) then I’d like you to take the rest of #2889 and apply it on-top!

dave-r12:
Gotcha - did you want to integrate this first? Or want me to apply directly on this PR?

swankjesse:
Lemme merge this as-is and you can rebase!

PR #2957, opened 2016-11-03
Tests for management of HTTP/2 connections that are shutdown.

swankjesse:
See: https://github.com/square/okhttp/issues/2756


okhttp-tests/src/test/java/okhttp3/URLConnectionTest.java
```diff
@@ -3555,6 +3555,37 @@ private void zeroLengthPayload(String method)
     assertEquals(0, dispatcher.runningCallsCount());
   }
 
+  @Ignore // TODO: recover gracefully when a connection is shutdown.
+  @Test public void streamedBodyIsRetriedOnHttp2Shutdown() throws Exception {
```
swankjesse: FYI @dave-r12 this is your commit, in my pull request!

PR #2959, opened 2016-11-05
New non-blocking API for websockets.

swankjesse:
Currently this uses a placholder name, NewWebSocket. I'd like to get this
reviewed, expand it to cover MockWebServer's needs, and then I'll delete
the current blocking API.

Still undecided is which APIs to add - if any - to expose the state of the
web socket.

https://github.com/square/okhttp/issues/2902

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/NewWebSocketTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/NewWebSocketTest.java
@@ -1,0 +1,347 @@
+/*
+ * Copyright (C) 2014 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okhttp3;
+
+import java.io.IOException;
+import java.net.ProtocolException;
+import java.util.Random;
+import java.util.logging.Logger;
+import okhttp3.internal.tls.SslClient;
+import okhttp3.internal.ws.NewWebSocketRecorder;
+import okhttp3.internal.ws.RealNewWebSocket;
+import okhttp3.internal.ws.WebSocketRecorder;
+import okhttp3.mockwebserver.MockResponse;
+import okhttp3.mockwebserver.MockWebServer;
+import okio.ByteString;
+import org.junit.After;
+import org.junit.Ignore;
+import org.junit.Rule;
+import org.junit.Test;
+
+import static okhttp3.TestUtil.defaultClient;
+import static okhttp3.WebSocket.TEXT;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.fail;
+
+public final class NewWebSocketTest {
+  @Rule public final MockWebServer webServer = new MockWebServer();
+
+  private final SslClient sslClient = SslClient.localhost();
+  private final NewWebSocketRecorder clientListener = new NewWebSocketRecorder("client");
+  private final WebSocketRecorder serverListener = new WebSocketRecorder("server");
+  private final Random random = new Random(0);
+  private OkHttpClient client = defaultClient().newBuilder()
+      .addInterceptor(new Interceptor() {
+        @Override public Response intercept(Chain chain) throws IOException {
+          Response response = chain.proceed(chain.request());
+          assertNotNull(response.body()); // Ensure application interceptors never see a null body.
+          return response;
+        }
+      })
+      .build();
+
+  @After public void tearDown() {
+    clientListener.assertExhausted();
+  }
+
+  @Test public void textMessage() throws IOException {
+    webServer.enqueue(new MockResponse().withWebSocketUpgrade(serverListener));
+    NewWebSocket client = enqueueClientWebSocket();
+
+    clientListener.assertOpen();
+    serverListener.assertOpen();
+
+    client.send("Hello, WebSockets!");
+    serverListener.assertTextMessage("Hello, WebSockets!");
+  }
+
+  @Test public void binaryMessage() throws IOException {
+    webServer.enqueue(new MockResponse().withWebSocketUpgrade(serverListener));
+    RealNewWebSocket client = enqueueClientWebSocket();
+
+    clientListener.assertOpen();
+    serverListener.assertOpen();
+
+    client.send(ByteString.encodeUtf8("Hello!"));
+    serverListener.assertBinaryMessage(new byte[] {'H', 'e', 'l', 'l', 'o', '!'});
+  }
+
+  @Test public void nullStringThrows() throws IOException {
+    webServer.enqueue(new MockResponse().withWebSocketUpgrade(serverListener));
+    RealNewWebSocket client = enqueueClientWebSocket();
+
+    clientListener.assertOpen();
+    try {
+      client.send((String) null);
+      fail();
+    } catch (NullPointerException e) {
+      assertEquals("text == null", e.getMessage());
+    }
+  }
+
+  @Test public void nullByteStringThrows() throws IOException {
+    webServer.enqueue(new MockResponse().withWebSocketUpgrade(serverListener));
+    RealNewWebSocket client = enqueueClientWebSocket();
+
+    clientListener.assertOpen();
+    try {
+      client.send((ByteString) null);
+      fail();
+    } catch (NullPointerException e) {
+      assertEquals("bytes == null", e.getMessage());
+    }
+  }
+
+  @Test public void serverMessage() throws IOException {
+    webServer.enqueue(new MockResponse().withWebSocketUpgrade(serverListener));
+    enqueueClientWebSocket();
+
+    clientListener.assertOpen();
+    WebSocket server = serverListener.assertOpen();
+
+    server.message(RequestBody.create(TEXT, "Hello, WebSockets!"));
+    clientListener.assertTextMessage("Hello, WebSockets!");
+  }
+
+  @Test public void throwingOnOpenFailsImmediately() {
+    webServer.enqueue(new MockResponse().withWebSocketUpgrade(serverListener));
+
+    final RuntimeException e = new RuntimeException();
+    clientListener.setNextEventDelegate(new NewWebSocket.Listener() {
+      @Override public void onOpen(NewWebSocket webSocket, Response response) {
+        throw e;
+      }
+    });
+    enqueueClientWebSocket();
+
+    serverListener.assertOpen();
+    serverListener.assertExhausted();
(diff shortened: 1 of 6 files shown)


okhttp/src/main/java/okhttp3/NewWebSocket.java
```diff
@@ -0,0 +1,150 @@
...
+   * Immediately and violently release resources held by this websocket, discarding any enqueued
+   * messages. This does nothing if the websocket has already been closed or canceled.
+   */
+  void cancel();
+
+  interface Factory {
+    NewWebSocket newWebSocket(Request request, Listener listener);
+  }
+
+  abstract class Listener {
```
swankjesse: (this should be a top-level class in the finished result)

okhttp/src/main/java/okhttp3/NewWebSocket.java
```diff
@@ -0,0 +1,150 @@
...
+ *   <li><strong>Canceled:</strong> the websocket connection failed. Messages that were successfully
+ *       enqueued by either peer may not have been transmitted to the other.
+ * </ul>
+ *
+ * Note that the state progression is independent for each peer. Arriving at a gracefully-closed
+ * state indicates that a peer has sent all of its outgoing messages and received all of its
+ * incoming messages. But it does not guarantee that the other peer will successfully receive all of
+ * its incoming messages.
+ */
+public interface NewWebSocket {
```
swankjesse: (this should be WebSocket, but that name’s currently taken!)

okhttp/src/main/java/okhttp3/NewWebSocket.java
```diff
@@ -0,0 +1,150 @@
...
+  /** Returns the original request that initiated this websocket. */
+  Request request();
+
+  /**
+   * Returns the size in bytes of all messages enqueued to be transmitted to the server. This
+   * doesn't include framing overhead. It also doesn't include any bytes buffered by the operating
+   * system or network intermediaries. This method returns 0 if no messages are waiting
+   * in the queue. If may return a nonzero value after the websocket has been canceled; this
```
JakeWharton: "If"

okhttp/src/main/java/okhttp3/NewWebSocket.java
```diff
@@ -0,0 +1,150 @@
...
+   * indicates that enqueued messages were not transmitted.
+   */
+  long queueSize();
+
+  /**
+   * Attempts to enqueue {@code text} to be UTF-8 encoded and sent as a the data of a text (type
+   * {@code 0x1}) message.
+   *
+   * <p>This method returns true if the message was enqueued. Messages that would overflow the
+   * outgoing message buffer will be rejected and trigger a {@link #close graceful shutdown} of
```
JakeWharton: Linkplain

okhttp/src/main/java/okhttp3/NewWebSocket.java
```diff
@@ -0,0 +1,150 @@
...
+   * <p>This method returns immediately.
+   */
+  boolean send(String text);
+
+  /**
+   * Attempts to enqueue {@code bytes} to be sent as a the data of a binary (type {@code 0x2})
+   * message.
+   *
+   * <p>This method returns true if the message was enqueued. Messages that would overflow the
+   * outgoing message buffer will be rejected and trigger a {@link #close graceful shutdown} of
```
JakeWharton: Linkplain

JakeWharton:
Not sure about onClosing and onClosed. What's the use case?

swankjesse:
There’s two different shutdown events that occur:
 * server stops sending messages
 * client queue is drained and socket is released

When the server stops sending messages, you might want to open a new websocket to a new server.

When the client queue is drained... you know onFailure isn't coming.

dlew:
Regarding `onClosing()` and `onClosed()`, either the client or the server initiates the close. If it's the client, then I actively *do not* want any more messages. In a way, then, it's actually more important for me to be able to query the `WebSocket` for its current state (i.e. open, closing, closed) than it is for me to get a notification of `onClosing`.

Is `onClosing` only limited to server closes?

okhttp/src/main/java/okhttp3/NewWebSocket.java
```diff
@@ -0,0 +1,150 @@
...
+  interface Factory {
+    NewWebSocket newWebSocket(Request request, Listener listener);
+  }
+
+  abstract class Listener {
+    /**
+     * Invoked when a websocket has been accepted by the remote peer and may begin transmitting
+     * messages.
+     */
+    public void onOpen(NewWebSocket websocket, Response response) {
```
dlew: I don't think this needs to return the `NewWebSocket` anymore since you get it with the `Factory`.

dlew: Oh hmm, all the methods return `NewWebSocket`. Is that necessary? Would someone really use a single `Listener` for multiple websocket connections?

okhttp/src/main/java/okhttp3/NewWebSocket.java
```diff
@@ -0,0 +1,150 @@
...
+ *
+ * Upon normal operation each websocket progresses through a sequence of states:
+ *
+ * <ul>
+ *   <li><strong>Connecting:</strong> the initial state of each websocket. Messages may be enqueued
+ *       but they won't be transmitted until the websocket is open.
+ *   <li><strong>Open:</strong> the websocket has been accepted by the remote peer and is fully
+ *       operational. Messages in either direction are enqueued for immediate transmission.
+ *   <li><strong>Closing:</strong> one of the peers on the websocket has initiated a graceful
+ *       shutdown. The websocket will continue to transmit already-enqueued messages but will refuse
```
dlew: Again, I'm not sure if "closing" includes the client calling `close()`. But if it does, then the client probably doesn't want it to keep transmitting already-enqueued messages.

dlew:
One thing I think is lacking right now is any ability to query the current state of the websocket. The listener gives you enough hooks to keep track of it yourself, but it'd be easier if you could just ask `NewWebSocket` if it's connecting, open, closing, closed or canceled.

okhttp/src/main/java/okhttp3/NewWebSocket.java
swankjesse: I can see it both ways. I definitely don’t feel strongly!

swankjesse:
Thanks for the feedback. Offering up the state machine is something I initially included, then pulled out, and might put back in.

dlew:
Often times I'm dealing with two competing state machines - the state of the websocket and the app's interactions with it.

For example, if my app wants to subscribe to a new channel, it first asks to connect to websockets. If it's not connected, then we connect and subscribe. If it's already connected, then we just send a channel subscription message. With the current API I'd have to track the websocket state even for this standard interaction.

swankjesse:
I thought about that, but I suspect you want to keep the state machines independent...

You keep a `Set<ChannelName> intendedSubscriptions`, and whenever you see `closing()` or `failed()` you schedule something to connect a new WebSocket.

dlew:
What about when I modify `intendedSubscriptions`? How do I know whether to initiate a new connection or shut it down (based on adding/removing subscriptions)?

okhttp-tests/src/test/java/okhttp3/NewWebSocketTest.java
```diff
@@ -0,0 +1,347 @@
...
+import org.junit.Rule;
+import org.junit.Test;
+
+import static okhttp3.TestUtil.defaultClient;
+import static okhttp3.WebSocket.TEXT;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.fail;
+
+public final class NewWebSocketTest {
```
swankjesse: copy paste, with interesting deltas called out

okhttp-tests/src/test/java/okhttp3/NewWebSocketTest.java
```diff
@@ -0,0 +1,347 @@
...
+    final RuntimeException e = new RuntimeException();
+    clientListener.setNextEventDelegate(new NewWebSocket.Listener() {
+      @Override public void onOpen(NewWebSocket webSocket, Response response) {
+        throw e;
+      }
+    });
+    enqueueClientWebSocket();
+
+    serverListener.assertOpen();
+    serverListener.assertExhausted();
```
swankjesse: in the other test we fail more gracefully:
```
    serverListener.assertClose(1001, "");
```
I don’t think this is likely to work reliably in practice, because the server may have sent further messages and closing the reader stream early will prevent the writer stream from writing the close frame.

okhttp-tests/src/test/java/okhttp3/NewWebSocketTest.java
```diff
@@ -0,0 +1,347 @@
...
+    webServer.enqueue(new MockResponse().withWebSocketUpgrade(serverListener));
+    RealNewWebSocket client = enqueueClientWebSocket();
+
+    clientListener.assertOpen();
+    WebSocket server = serverListener.assertOpen();
+
+    client.send("abc");
+    serverListener.assertTextMessage("abc");
+
+    server.message(RequestBody.create(TEXT, "def"));
```
swankjesse: The other test uses pings here. I don’t have ’em so I just used normal messages.

okhttp/src/main/java/okhttp3/internal/ws/RealNewWebSocket.java
```diff
@@ -0,0 +1,488 @@
...
+      writerRunning = true;
+      executor.execute(writerRunnable);
+    }
+  }
+
+  /**
+   * Attempts to remove a single frame from a queue and send it. This prefers to write urgent pongs
+   * before less urgent messages and close frames. For example it's possible that a caller will
+   * enqueue messages followed by pongs, but this sends pongs followed by messages. Pongs are always
+   * written in the order they were enqueued.
```
swankjesse: This pongs-have-precedence is probably overkill. The blocking API would just use `synchronized` to pick which order the frames were written in, which could starve pongs. Not sure what WebSockets says but HTTP/2 was particular that pongs were supposed to be shortcutted to the front-of-the-queue.

swankjesse: Ooooh, and that's something I've broken. I think the fix may be to make the writer help out here.

okhttp/src/main/java/okhttp3/NewWebSocket.java
swankjesse: Fixed all.

okhttp/src/main/java/okhttp3/internal/ws/RealNewWebSocket.java
```diff
@@ -0,0 +1,488 @@
...
+    // Receive frames until there are no more.
+    WebSocketReader reader = new WebSocketReader(streams.client, streams.source, this);
+    listener.onOpen(this, response);
+    while (receivedCloseCode == -1) {
+      // This method call results in one or more onRead* methods being called on this thread.
+      reader.processNextFrame();
+    }
+  }
+
+  @Override public void onReadMessage(ResponseBody body) throws IOException {
```
JakeWharton: Future optimization: skip the `ResponseBody` wrapper. Either with a `boolean` + `BufferedSource` or two callbacks with either `String` or `ByteString` already.

swankjesse: Definitely.

PR #2969, opened 2016-11-11
Add @NotNull annotation to Interceptor

edenman:
(so Kotlin knows that params don't have to be optionals)

### DIFF
--- a/okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
+++ b/okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
@@ -34,6 +34,7 @@ import okhttp3.internal.http.HttpHeaders;
 import okhttp3.internal.platform.Platform;
 import okio.Buffer;
 import okio.BufferedSource;
+import org.jetbrains.annotations.NotNull;
 
 import static okhttp3.internal.platform.Platform.INFO;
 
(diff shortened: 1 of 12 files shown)


swankjesse:
Weird. I think `@NotNull` should be the default. I’m not eager to change everything to match this.

digitalbuddha:
will Kotlin recognize `@ParametersAreNonnullByDefault`? I believe you can add it at it to a `package-info.java`

edenman:
Yeah, unfortunately Kotlin can't assume `@NotNull` by default and there doesn't seem to be a way to tell it that aside from one-off annotations.  AFAICT from [this](https://kotlinlang.org/docs/reference/java-interop.html#nullability-annotations), anyway.  I found a request for `@ParametersAreNonnullByDefault` support [here](https://youtrack.jetbrains.com/issue/KT-10942) but it hasn't had any activity or discussion yet. 

Not the end of the world if you don't want to accept this PR, i just thought it was worth proposing since I'm having to do a null check in my kotlin impl of an `Interceptor`.

JakeWharton:
You can change the parameter type to non-nullable (since it's a platform
type and not a truly nullable type) and the null check will be emitted
automatically. Kotlin just defaults platform types to nullable when you use
the IDE to generate method implementation stubs for you.

edenman:
oh, nice!  ignore me

PR #2973, opened 2016-11-13
Always use LinkedHashMap. Never HashMap.

swankjesse:


yegeniy:
Why? http://stackoverflow.com/a/22839050/2916086 ?

swankjesse:
Nondeterminism is the worst. It causes butterfly effects through an app. My rule of “Never Use HashMap” and it’s never been a performance problem in practice.

yegeniy:
Interesting. Thanks @swankjesse

Issue #2696, opened 2016-07-05
Websocket status code 1005 & 1006


b4stien:
As per RFC 6455, status code 1005 or 1006 should be used when the websocket abnormally closed (ref: https://tools.ietf.org/html/rfc6455#section-7.1.5 and https://tools.ietf.org/html/rfc6455#section-7.4.1).

1006 in particular would be handy to detect that the network (or the server on the other side) crashed.

JakeWharton:
Interesting. For both of these cases we trigger `onFailure` with the appropriate exception. I presume you're saying that we should instead be calling `onClose` on the listener and sending them in a close frame back to the server.

JakeWharton:
The spec is pretty clear on this. Working on it.

PR #2989, opened 2016-11-18
Migrate MockWebServer to the async web sockets API.

swankjesse:


okhttp-tests/src/test/java/okhttp3/internal/ws/RealWebSocketTest.java
```diff
@@ -83,418 +76,223 @@
...
     assertTrue(client.processNextFrame());
     clientListener.assertTextMessage("Hello!");
     assertFalse(client.processNextFrame());
-    clientListener.assertClose(1000, "Bye!");
+    clientListener.assertClosing(1000, "Bye!");
   }
 
-  @Test public void protocolErrorBeforeCloseSendsClose() {
-    server2client.raw().write(ByteString.decodeHex("0a00")); // Invalid non-final ping frame.
+  @Test public void protocolErrorBeforeCloseSendsClose() throws IOException {
```
JakeWharton: Test name is wrong. Doesn't send close anymore.

okhttp-tests/src/test/java/okhttp3/internal/ws/RealWebSocketTest.java
```diff
@@ -83,418 +76,223 @@
...
-    server.processNextFrame();
-    serverListener.assertClose(1000, "Hello!");
+    serverListener.assertClosing(1000, "Hello!");
 
     serverListener.assertExhausted(); // Client should not have sent second close.
   }
 
-  @Test public void networkErrorReportedAsCloseNotFailure() {
-    server2client.close();
+  @Test public void networkErrorReportedAsCloseNotFailure() throws IOException {
```
JakeWharton: Test name is wrong. Doesn't send close anymore.

JakeWharton: Also I'm pretty sure this behavior was the result of someone filing an issue.

swankjesse: Indeed. https://github.com/square/okhttp/issues/2696

I think I disagree with that change. The WebSocket RFC doesn’t anticipate an `onFailure()` callback, and I think we should either keep `onFailure()` and use it universally, or go all-in on `onClose()` and synthesize codes 1005 and 1006 instead of supplying exceptions.

PR #2992, opened 2016-11-20
Gracefully recover from an HTTP/2 connection shutdown at start of request.

dave-r12:
This was work started while addressing #2756, and is very much an edge case. I got sidetracked with the issue around violently and abruptly closing a socket and wanted to close the loop here.


swankjesse:
Awesome. Thanks for closing the loop on this.

kenyee:
Did this make it into 3.5?

swankjesse:
@kenyee easiest way to find out is to check out the 3.5 tag and look for this commit.

kenyee:
oops...was looking under branches (no branch for 3.5)...it's in 3.5 👍

PR #2994, opened 2016-11-20
Correct UTF8 BOM is EFBBBF

danielgindi:


dave-r12:
Good catch.

swankjesse:
Squashed & merged manually. Thanks!

Issue #2892, opened 2016-10-04
HTTP/2: empty data-frame without Content-Length generated for requests without request body


rfc2822:
Let's assume this code for generating an `OPTIONS` request:

```
Response response = httpClient.newCall(new Request.Builder()
                .method("OPTIONS", null)
                .url(location)
                .build()).execute();
```

With HTTP/1.1, it just generates a request without body, which will be accepted and processed by the server.

However, when HTTP/2 is active, okhttp 3.4.1 generates an empty DATA frame without `Content-Length`:

```
davdroid.DavResourceFinder: [HttpClient$1] --> OPTIONS https://yourserver/katana/public/server.php/principals/bitfire/ http/1.1
davdroid.DavResourceFinder: [HttpClient$1] --> END OPTIONS
davdroid.dav4android: [dav4android.BasicDigestAuthHandler] Adding Basic authorization header for https://yourserver/katana/public/server.php/principals/bitfire/
okhttp3.internal.framed.Http2$FrameLogger: [okhttp3.internal.framed.Http2$Writer] >> 0x0000001f   218 HEADERS       END_HEADERS
[!!!] okhttp3.internal.framed.Http2$FrameLogger: [okhttp3.internal.framed.Http2$Writer] >> 0x0000001f     0 DATA          END_STREAM
okhttp3.internal.framed.Http2$FrameLogger: [okhttp3.internal.framed.Http2$Reader] << 0x0000001f     7 HEADERS       END_HEADERS
```

Servers like Apache 2.4 reject this request because the `OPTIONS` request should either have
1. no body – and thus no data frame, or
2. a body, and thus `Content-Length`.

Because `null` is specified as the request body, okhttp shouldn't send a data frame.

For further information, see https://forums.bitfire.at/topic/1124/411-length-error-davdroid-1-0-9-2-owncloud/10. If I can provide more information, please let me know.

swankjesse:
Can fix!

PR #2995, opened 2016-11-21
Avoid sending empty HTTP/2 data frames when there is no request body.

dave-r12:
Closes: https://github.com/square/okhttp/issues/2892


swankjesse:
Build failure is unrelated.
https://github.com/square/okhttp/issues/2997

swankjesse:
(We might want a test for this.)

dave-r12:
Agree on the testing. Did you have anything in mind? I didn't see a good way for adding a test as-is. One idea I was playing with was adding a listener for inbound HTTP/2 frames, and then checking no data frames were sent. It turned into a fair bit of change, so I withdrew. If you're receptive to the idea, I can follow up with something more concrete.

swankjesse:
Yeah, good point. We lack visibility here. In RecordedRequest we track the chunk sizes for chunked encoding as list of integers. Maybe we populate that with frame sizes for HTTP/2?

dave-r12:
Alright, I'll look into that, thanks.

PR #2996, opened 2016-11-21
Fix some Autobahn tests.

swankjesse:
Bump the queue limit to 16 MiB for Autobahn tests. This is the ceiling they
use and necessary to pass their tests.

Move to port 9099. Port 9001 conflicts with supervisord.

Throw the right exceptions on close codes, at the right times.


okhttp/src/main/java/okhttp3/internal/ws/WebSocketProtocol.java
```diff
@@ -103,21 +103,21 @@ static void toggleMask(byte[] buffer, long byteCount, byte[] key, long frameByte
     }
   }
 
-  static void validateCloseCode(int code, boolean argument) throws ProtocolException {
```
swankjesse: this was awkward because it was throwing a checked exception

PR #3000, opened 2016-11-22
Apply timeouts within individual frames.

swankjesse:


swankjesse:
They're asymmetric. Reads need to not time out even if the peer isn't writing new frames. Writes need to time out if the peer isn't reading existing frames. I can write tests for that!

Issue #2997, opened 2016-11-21
URLConnectionTest.authenticateWithChunkedStreaming is flaky


swankjesse:
```
authenticateWithChunkedStreaming(okhttp3.URLConnectionTest)  Time elapsed: 0.021 sec  <<< ERROR!
java.io.FileNotFoundException: http://localhost:48167/
	at okhttp3.internal.huc.OkHttpURLConnection.getInputStream(OkHttpURLConnection.java:253)
	at okhttp3.URLConnectionTest.testAuthenticateWithStreamingPost(URLConnectionTest.java:1533)
	at okhttp3.URLConnectionTest.authenticateWithChunkedStreaming(URLConnectionTest.java:1508)
```

PR #3001, opened 2016-11-22
Fix flakiness in authenticateWithChunkedStreaming().

swankjesse:
We were prefering to return the networkResponse before the request
had completely failed. Instead we should return the network response
only on failure, and only for APIs that return failed responses.

Closes: https://github.com/square/okhttp/issues/2997


dave-r12:
Good find 👍

PR #3005, opened 2016-11-24
Automatically ping the peer on a user-specified interval.

swankjesse:
This changes RealWebSocket from using a shared ThreadPoolExecutor to
using a private ScheduledExecutorService. This make some of the code
simpler but it means that applications will hold 1 writer thread for
each websocket, even if it is mostly not writing.

Testing automatic pings is awkward. This refactors much of
RealWebSocketTest to make it more natural to reinitialize the
configuration.

### DIFF
--- a/okhttp/src/main/java/okhttp3/OkHttpClient.java
+++ b/okhttp/src/main/java/okhttp3/OkHttpClient.java
@@ -201,6 +201,7 @@ public class OkHttpClient implements Cloneable, Call.Factory, WebSocket.Factory 
   final int connectTimeout;
   final int readTimeout;
   final int writeTimeout;
+  final int pingInterval;
 
   public OkHttpClient() {
     this(new Builder());
@@ -246,6 +247,7 @@ public class OkHttpClient implements Cloneable, Call.Factory, WebSocket.Factory 
     this.connectTimeout = builder.connectTimeout;
     this.readTimeout = builder.readTimeout;
     this.writeTimeout = builder.writeTimeout;
+    this.pingInterval = builder.pingInterval;
   }
 
   private X509TrustManager systemDefaultTrustManager() {
@@ -287,6 +289,11 @@ public class OkHttpClient implements Cloneable, Call.Factory, WebSocket.Factory 
   /** Default write timeout (in milliseconds). */
   public int writeTimeoutMillis() {
     return writeTimeout;
+  }
+
+  /** Web socket ping interval (in milliseconds). */
+  public int pingIntervalMillis() {
+    return pingInterval;
   }
 
   public Proxy proxy() {
@@ -429,6 +436,7 @@ public class OkHttpClient implements Cloneable, Call.Factory, WebSocket.Factory 
     int connectTimeout;
     int readTimeout;
     int writeTimeout;
+    int pingInterval;
 
     public Builder() {
       dispatcher = new Dispatcher();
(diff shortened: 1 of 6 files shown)


okhttp/src/main/java/okhttp3/internal/ws/RealWebSocket.java
```diff
@@ -74,22 +66,19 @@
...
-   * True if {@link #writerRunnable} is active. Because writing is single-threaded we only enqueue
-   * it if it isn't already enqueued.
-   */
-  private boolean writerRunning;
-
   /** Null until this web socket is connected. Note that messages may be enqueued before that. */
   private WebSocketWriter writer;
 
+  /** Null until this web socket is connected. Used for writes, pings, and close timeouts. */
+  private ScheduledThreadPoolExecutor executor;
```
dave-r12: Did you mean to use the concrete type here? It doesn't make much difference, just used to seeing the interface type.

okhttp/src/main/java/okhttp3/OkHttpClient.java
```diff
@@ -513,15 +513,32 @@ public Builder readTimeout(long timeout, TimeUnit unit) {
...
+     * enforced on the acknowledging pongs.
+     *
+     * <p>The default value of 0 disables client-initiated pings.
+     */
+    public Builder pingInterval(long interval, TimeUnit unit) {
+      pingInterval = checkDuration("interval", interval, unit);
       return this;
     }
 
+    private int checkDuration(String name, long duration, TimeUnit unit) {
```
JakeWharton: nit: could be static

okhttp/src/main/java/okhttp3/OkHttpClient.java
```diff
@@ -289,6 +291,11 @@ public int writeTimeoutMillis() {
     return writeTimeout;
   }
 
+  /** Web socket ping interval (in milliseconds). */
```
JakeWharton: HTTP/2 as well? Future?

okhttp/src/main/java/okhttp3/internal/ws/RealWebSocket.java
JakeWharton: Emits invokevirtural calls instead of invokeinterface? :trollface: 

We're not there yet, are we? No...

swankjesse: `ScheduledExecutorService` for the win. I think for a while there was a method in the concrete class that I thought I needed.

okhttp/src/main/java/okhttp3/OkHttpClient.java
```diff
@@ -513,15 +513,32 @@ public Builder readTimeout(long timeout, TimeUnit unit) {
...
-      if (millis == 0 && timeout > 0) throw new IllegalArgumentException("Timeout too small.");
-      writeTimeout = (int) millis;
+      writeTimeout = checkDuration("timeout", timeout, unit);
+      return this;
+    }
+
+    /**
+     * Sets the interval between web socket pings initiated by this client. Use this to
+     * automatically send web socket ping frames until either the web socket fails or it is closed.
+     * This keep the connection alive and may detect connectivity failures early. No timeouts are
```
swankjesse: Fixed!

okhttp/src/main/java/okhttp3/OkHttpClient.java
swankjesse: I think so.

PR #3008, opened 2016-11-27
Remove creation of synthetic accessor methods.

vanniktech:
There are quite a few of them. Any of them were there by design?


dave-r12:
Any before/after counts? Someone should create a maven plugin to catch these during build time.

JakeWharton:
Last time I checked (about a year ago) OkHttp had something like 250
synthetic accessor methods.

vanniktech:
IntelliJ inspections showed me I think 316 accessor methods.

swankjesse:
Thanks.

Issue #2788, opened 2016-08-13
Websocket isn't robust enough with servers that don't reply a close message


jwendell:
If a websocket server doesn't reply the close message, the client will never close the connection.

PR #3013, opened 2016-12-01
If a graceful close doesn't work, cancel the websocket.

swankjesse:
Closes: https://github.com/square/okhttp/issues/2788

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/ws/RealWebSocket.java
+++ b/okhttp/src/main/java/okhttp3/internal/ws/RealWebSocket.java
@@ -23,6 +23,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Random;
 import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.ScheduledThreadPoolExecutor;
 import okhttp3.Call;
 import okhttp3.Callback;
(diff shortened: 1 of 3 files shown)


okhttp/src/main/java/okhttp3/internal/ws/RealWebSocket.java
```diff
@@ -56,6 +57,12 @@
    */
   private static final long MAX_QUEUE_SIZE = 16 * 1024 * 1024; // 16 MiB.
 
+  /**
+   * The maximum amount of time after the client calls {@link #close} to wait for a graceful
+   * shutdown. If the server doesn't respond the websocket will be canceled.
+   */
+  private static final long CANCEL_AFTER_CLOSE_MILLIS = 60 * 1000;
```
swankjesse: Not currently configurable. Strong feelings?

okhttp/src/main/java/okhttp3/internal/ws/RealWebSocket.java
```diff
@@ -99,6 +106,12 @@
   /** True if we've enqueued a close frame. No further message frames will be enqueued. */
   private boolean enqueuedClose;
 
+  /**
+   * When executed this will cancel this websocket. This future itself should be canceled if that is
+   * unnecessary because the web socket is already closed or canceled.
+   */
+  private ScheduledFuture<?> cancelFuture;
```
swankjesse: Need to keep this so that we can release the ScheduledExecutorService thread as eagerly as possible.

okhttp/src/main/java/okhttp3/internal/ws/RealWebSocket.java
JakeWharton: Use the connect timeout? [Dis]connect timeout, am I right? Otherwise even read timeout seems appropriate.

okhttp/src/main/java/okhttp3/internal/ws/RealWebSocket.java
```diff
@@ -356,7 +370,12 @@ public synchronized boolean pong(ByteString payload) {
     return true;
   }
 
-  @Override public synchronized boolean close(final int code, final String reason) {
+  @Override public boolean close(int code, String reason) {
+    return close(code, reason, CANCEL_AFTER_CLOSE_MILLIS);
+  }
+
+  public synchronized boolean close(
```
JakeWharton: private?

PR #3015, opened 2016-12-01
Change HttpURL.redact() to return a String.

swankjesse:


okhttp/src/main/java/okhttp3/HttpUrl.java
```diff
@@ -842,14 +842,16 @@ public String fragment() {
...
+   * Returns a string with containing this URL with its username, password, query, and fragment
+   * stripped, and its path replaced with {@code /...}. For example, redacting {@code
+   * http://username:password@example.com/path} returns {@code http://example.com/...}.
    */
-  public HttpUrl redact() {
-    Builder builder = newBuilder("/...");
-    builder.username("");
-    builder.password("");
-    return builder.build();
+  public String redact() {
```
JakeWharton: This is unreleased, right?

swankjesse: Exactly.

swankjesse:
I snuck in a fix for a flaky test in my last PR.

PR #3019, opened 2016-12-01
Support NPN protocol for android.

duanbo1983:
The NPN protocol is not supported in OkHttp, however, there are still quite a few servers live with it.

And Google Play Service has already provided support to NPN on Android. And the adoption of Google Play Service in Android world is pretty high (except devices from China), meaning it could benefit a lot users.

So, this change is to support NPN support by using [ProviderInstaller](https://developers.google.com/android/reference/com/google/android/gms/security/ProviderInstaller).

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/platform/AndroidPlatform.java
+++ b/okhttp/src/main/java/okhttp3/internal/platform/AndroidPlatform.java
@@ -16,22 +16,25 @@
 package okhttp3.internal.platform;
 
 import android.util.Log;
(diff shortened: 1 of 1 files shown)


swankjesse:
We dropped NPN a long time ago in favor of ALPN.

swankjesse:
NPN is completely obsolete.
https://bugs.chromium.org/p/chromium/issues/detail?id=587472

hzsweers:
@duanbo1983 sounds like we should have a meeting with some of our backend folks...

PR #3025, opened 2016-12-07
Introduce SocketRecorder: a means to inspect all bytes written/read to a socket.

dave-r12:
I recently made a change in https://github.com/square/okhttp/pull/2995 and did not include a test case. The primary reason was a lack of visibility in MWS. The goal of this PR is to provide that visibility.

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/internal/SocketRecorder.java
+++ b/okhttp-tests/src/test/java/okhttp3/internal/SocketRecorder.java
@@ -1,0 +1,206 @@
+/*
+ * Copyright (C) 2016 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okhttp3.internal;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.net.Socket;
+import java.util.Deque;
+import java.util.concurrent.LinkedBlockingDeque;
+import javax.net.ssl.SSLSocket;
+import javax.net.ssl.SSLSocketFactory;
+import okhttp3.DelegatingSSLSocket;
+import okhttp3.DelegatingSSLSocketFactory;
+import okio.Buffer;
+import okio.ByteString;
+
+/** Records all bytes written and read from a socket and makes them available for inspection. */
+public final class SocketRecorder {
+  private final Deque<RecordedSocket> recordedSockets = new LinkedBlockingDeque<>();
+
+  /** Returns an SSLSocketFactory whose sockets will record all transmitted bytes. */
+  public SSLSocketFactory sslSocketFactory(SSLSocketFactory delegate) {
+    return new DelegatingSSLSocketFactory(delegate) {
+      @Override protected SSLSocket configureSocket(SSLSocket sslSocket) throws IOException {
+        RecordedSocket recordedSocket = new RecordedSocket();
+        recordedSockets.add(recordedSocket);
+        return new RecordingSSLSocket(sslSocket, recordedSocket);
+      }
+    };
+  }
+
+  public RecordedSocket takeSocket() {
+    return recordedSockets.remove();
+  }
+
+  /** A bidirectional transfer of unadulterated bytes over a socket. */
+  public static final class RecordedSocket {
+    private final Buffer bytesWritten = new Buffer();
+    private final Buffer bytesRead = new Buffer();
+
+    synchronized void byteWritten(int b) {
+      bytesWritten.writeByte(b);
+    }
+
+    synchronized void byteRead(int b) {
+      bytesRead.writeByte(b);
+    }
+
+    synchronized void bytesWritten(byte[] bytes, int offset, int length) {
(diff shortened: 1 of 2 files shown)


okhttp-tests/src/test/java/okhttp3/internal/ws/RealWebSocketTest.java
```diff
@@ -175,6 +176,8 @@
     server.listener.assertClosing(1000, "Hi!");
     client.listener.assertClosed(1000, "Hello!");
     server.listener.assertClosed(1000, "Hi!");
+
+    client.closeCalled.await(10, TimeUnit.SECONDS);
     assertTrue(client.closed);
```
dave-r12: I was getting an intermittent failure here. I believe the reason is that there is a race setting closed to true.

swankjesse: Yup. Totally believable!

dave-r12:
Gotcha, yeah I think your assessment is spot-on.

swankjesse:
I had another, weirder idea on how to potentially accomplish this. Here’s a rough sketch of the layers between client and server in our tests:

```
OkHttpClient <=> TLS client <=> TCP socket <=> TLS server <=> MockWebServer
```

It would be annoying, but not _that_ annoying, do to a SocketRecorder that MITMs the entire connection. It would either terminate TLS internally as a true MITM, or hook into `SocketFactory`.

```
OkHttpClient <=> TLS client <=> TCP socket <=> TLS server <=> MITM <=> TLS client <=> TCP socket <=> TLS server <=> MockWebServer
```

Doing this would allow us to capture, say, a `ByteString` for everything transmitted by client and by server. We could process such things with `Http2Reader` to decode frames from a recorded conversation.

I don’t quite know if this is 100 lines of code or 500 lines of code. But adding a TCP-layer MITM could be handy in a few places in our tests.

dave-r12:
Cool - I like that it's more general _and_ more useful. Is that something you'd prefer to work directly? If not, I'm happy to give it a go. I think I get the gist, but may take me a few iterations to get it right.

swankjesse:
Give it a go!

dave-r12:
I've created a functional prototype of a possible MITM. It's still very much a WIP, but I was hopeful to get feedback early so I don't veer off the intended path.

I spent some time toying with a custom SSLSocketFactory that would wrap the sockets and record the input/output stream calls. That worked, except it seems to cause an issue with Jetty ALPN. Specifically, the ALPN.put() call appears to require the actual SSLSocket that will perform the handshake. If you ALPN.put() an SSLSocket instance that is not the instance you do the handshake with, it breaks. (At least my reading of the code.) This can be worked around, but my feeling was that it would require some logic in OkHttp itself.

swankjesse:
Ahh, yeah see. You end up proxying the ALPN also.

dave-r12:
I experimented more with the pure MITM, and was curious what a more generic example might look like. While doing that, I discovered it may take a fair amount of duplicated code to properly implement it. For example, how are multiple routes handled? What happens if a route fails? How do you handle follow up requests? I think these can all be accounted for, but again it seemed to me like a fair amount of duplicate code. Alternatively, it could just be constrained to handle simple cases.

With that, I went back to the wrapping (SSL)SocketFactory. I think this is a better approach because you get all that stuff in OkHttp for free (multiple route selection, connection failures, follow up requests, etc.) The only thing was the Jetty ALPN issue, which can be worked around.

Let me know if you think either of these provide some value.

swankjesse:
Oooh, I like them both. Am I reading this right that the PR has both the MITM approach and the single-socket recorder, and they’re independent?

dave-r12:
Ok. Yes you are reading that right. I ended up spending some time on both approaches. I'm leaning towards the SocketRecorder at this point. I'll remove the ConnectionRecorder (MITM) and clean up what's left with SocketRecorder.

swankjesse:
Dig. I think the MITM code is neat. Maybe we’ll revive it someday. But for this task SocketRecorder is a better fit.

okhttp-tests/src/test/java/okhttp3/internal/SocketRecorder.java
```diff
@@ -0,0 +1,206 @@
...
+    private final OutputStream outputStream;
+
+    RecordingSSLSocket(SSLSocket delegate, RecordedSocket recordedSocket) {
+      super(delegate);
+      inputStream = new RecordingInputStream(delegate, recordedSocket);
+      outputStream = new RecordingOutputStream(delegate, recordedSocket);
+    }
+
+    @Override public void startHandshake() throws IOException {
+      // Intercept the handshake to properly configure TLS extensions with Jetty ALPN. Jetty ALPN
```
swankjesse: It’s too bad this is so much work. It’s invisible to the user though, so I’m cool with it.

okhttp-tests/src/test/java/okhttp3/internal/http2/HttpOverHttp2Test.java
```diff
@@ -760,6 +766,70 @@ private void noRecoveryFromErrorWithRetryDisabled(ErrorCode errorCode) throws Ex
...
+    buffer.write(recordedSocket.bytesWritten());
+
+    RecordingHandler handler = new RecordingHandler();
+    Http2Reader reader = new Http2Reader(buffer, false);
+    reader.readConnectionPreface(null);
+    while (reader.nextFrame(false, handler)) {
+    }
+
+    assertEquals(1, handler.headerFrameCount);
+    assertTrue(handler.dataFrames.isEmpty());
```
swankjesse: this is awesome

okhttp-tests/src/test/java/okhttp3/internal/http2/HttpOverHttp2Test.java
```diff
@@ -760,6 +766,70 @@ private void noRecoveryFromErrorWithRetryDisabled(ErrorCode errorCode) throws Ex
     assertEquals("bar", pushedRequest.getHeader("foo"));
   }
 
+  @Test public void noDataFramesSentWithNullRequestBody() throws Exception {
```
swankjesse: I love that we have a test for this now.

swankjesse:
I’m very happy with where this landed.

PR #3033, opened 2016-12-11
Fix flakiness in RealWebSocketTest.clientAndServerCloseClosesConnection

swankjesse:


okhttp-tests/src/test/java/okhttp3/internal/ws/RealWebSocketTest.java
```diff
@@ -175,6 +175,7 @@
     server.listener.assertClosing(1000, "Hi!");
     client.listener.assertClosed(1000, "Hello!");
     server.listener.assertClosed(1000, "Hi!");
+    client.webSocket.awaitTermination(5, TimeUnit.SECONDS);
```
dave-r12: Does this establish a _happens-before_ on `client.closed`?

JakeWharton: Is the reason this is needed is because `closed = true` doesn't happen until after the listener callback?

swankjesse: Exactly.

PR #3037, opened 2016-12-12
Preferred to use the "Host" header of request as host name when create address.

hanpfei:
Use okhttp with HttpDns library.
Even if we set "Host" header for request correctly, 
okhttp still cannot send correct SNI to server when for https.

Preferred to use the "Host" header of request as host name when create address to fix the issue.

Signed-off-by: hanpfei <hanpfei@gmail.com>


swankjesse:
This isn’t going to work generally – things like HTTP/2 and HTTP proxies also need the URL’s hostname. Instead I think the fix you want is to customize the `Dns`.

hanpfei:
OK. Thanks a lot!

Issue #675, opened 2014-04-04
Expect 100-continue does not appear to work


ifedorenko:
This maybe my misunderstanding of HttpURLConnection API or HTTP protocol itself, but I can't get okhttp do Expect 100-continue handshake as described in rfc [1]. Identical code using standard HttpURLConnection implementation appears to work. [2] has is an example that shows both okhttp and jdk tests I did

[1] http://www.w3.org/Protocols/rfc2616/rfc2616-sec8.html
[2] https://github.com/ifedorenko/bug430772/blob/master/src/test/java/bug430772/Expect100ContinueTest.java

swankjesse:
I started looking into this. Seems relatively straightforward. We need to add some special support in HTTP/2, since the `100: Continue` comes back as a 2nd set of headers there.
http://httpwg.org/specs/rfc7540.html#rfc.section.8.1.3

PR #3057, opened 2016-12-23
Support 'Expect: 100-continue' as a request header.

swankjesse:
This will stall reading the response body until the server returns
an intermediate '100 Continue' response.

Closes: https://github.com/square/okhttp/issues/675

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/http2/Http2Codec.java
+++ b/okhttp/src/main/java/okhttp3/internal/http2/Http2Codec.java
@@ -40,6 +40,7 @@ import okio.Okio;
 import okio.Sink;
 import okio.Source;
 
+import static okhttp3.internal.http.StatusLine.HTTP_CONTINUE;
 import static okhttp3.internal.http2.Header.RESPONSE_STATUS;
 import static okhttp3.internal.http2.Header.TARGET_AUTHORITY;
 import static okhttp3.internal.http2.Header.TARGET_METHOD;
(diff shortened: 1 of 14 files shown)


okhttp/src/main/java/okhttp3/internal/http2/Http2Codec.java
```diff
@@ -158,6 +168,25 @@ public Http2Codec(
         .headers(headersBuilder.build());
   }
 
+  /**
+   * Returns the code of the last ":status" header in {@code headerBlock}, or -1 if there is no
+   * status header in the block. Returning the last status works around a corner case where the
+   * response headers contain both a "100 Continue" status line and its follow-up "200 OK" status
+   * line.
+   */
+  private int statusCode(List<Header> headerBlock) throws IOException {
```
swankjesse: Yeah, good call. You can see I went around in circles a bit here. I’ve changed it some more, and addressed my TODO about how to handle multiple blocks of headers in the normal case. It’s a bit ugly (null as sentinel!) but it’s well-hidden.

okhttp/src/main/java/okhttp3/internal/http2/Http2Codec.java
```diff
@@ -135,22 +145,31 @@ public Http2Codec(
...
   public static Response.Builder readHttp2HeadersList(List<Header> headerBlock) throws IOException {
-    String status = null;
-
+    StatusLine statusLine = null;
     Headers.Builder headersBuilder = new Headers.Builder();
     for (int i = 0, size = headerBlock.size(); i < size; i++) {
-      ByteString name = headerBlock.get(i).name;
+      Header header = headerBlock.get(i);
+
+      // If there were multiple header blocks they will be delimited by nulls. Discard existing
```
dave-r12: Just to clarify: the scenario you're envisioning here is the server sends back-to-back header blocks, i.e. one with '100 Continue' followed by '200 OK', and that happens prior to the stalled request having time to read the first header block?

okhttp/src/main/java/okhttp3/internal/http2/Http2Codec.java
```diff
@@ -135,22 +145,31 @@ public Http2Codec(
...
+      // header blocks if the existing header block is a '100 Continue' intermediate response.
+      if (header == null) {
+        if (statusLine != null && statusLine.code == HTTP_CONTINUE) {
+          statusLine = null;
+          headerBlock.clear();
```
dave-r12: Should this be: `headersBuilder = new Headers.Builder();`?

okhttp/src/main/java/okhttp3/internal/http2/Http2Codec.java
swankjesse: Exactly.

okhttp/src/main/java/okhttp3/internal/http2/Http2Codec.java
swankjesse: yup. Ooooh, good catch!

Issue #373, opened 2013-12-21
Multi HTTP/2 Connections connecting to a single domain


lingmingyb:
Hi, 

Under a multi thread frame we may send several requests synchronously to a single domain. In OkHttp(SPDY) according to the codes I found that multi SPDY connections maybe set up to the same remote (I`m keeping tracking version 1.1.0, so not sure if the newest version has fixed it, seems not from the codes.) 

I do not think it's a correct behavior for clients so just wanna open the topic here. I've got a patch for it and will make a pull request if you guys confirm this issue.

Thanks

yschimke:
I wonder whether client protocol (h2 vs http1) could be specific per domain as well as a default per client.  This could allow high volume apps to configure h2 only for their api and then the blocking could be done predictably.  I don't think any of the major sites are going to turn off h2 now.

I think we saw this in production and were scratching our heads why we had more connections than we predicted.

PR #3080, opened 2017-01-03
Release duplicated multiplexed connections.

swankjesse:
If we make concurrent requests to an HTTP/2 server, close all but
the first connection. Creating multiple connections and then later
releasing them is somewhat pessimistic; it's also much safer for
awkward cases like connect attempts being canceled.

Closes: https://github.com/square/okhttp/issues/373

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/connection/StreamAllocation.java
+++ b/okhttp/src/main/java/okhttp3/internal/connection/StreamAllocation.java
@@ -15,6 +15,7 @@
  */
 package okhttp3.internal.connection;
 
+import java.io.Closeable;
 import java.io.IOException;
 import java.lang.ref.Reference;
 import java.lang.ref.WeakReference;
@@ -28,6 +29,8 @@ import okhttp3.internal.http.HttpCodec;
 import okhttp3.internal.http2.ConnectionShutdownException;
 import okhttp3.internal.http2.ErrorCode;
 import okhttp3.internal.http2.StreamResetException;
+
+import static okhttp3.internal.Util.closeQuietly;
 
 /**
  * This class coordinates the relationship between three entities:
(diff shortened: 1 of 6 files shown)


okhttp/src/main/java/okhttp3/internal/connection/StreamAllocation.java
```diff
@@ -170,32 +172,44 @@ private RealConnection findConnection(int connectTimeout, int readTimeout, int w
...
-    RealConnection newConnection = new RealConnection(connectionPool, selectedRoute);
+
+    RealConnection result = new RealConnection(connectionPool, selectedRoute);
 
     synchronized (connectionPool) {
-      acquire(newConnection);
-      Internal.instance.put(connectionPool, newConnection);
-      this.connection = newConnection;
+      acquire(result);
+      Internal.instance.put(connectionPool, result);
```
dave-r12: I'm wondering: should the put call occur after the deduplicate check? I think there is a possible (highly unlikely) set of interleavings that cause the `connections.allocations.size()` to be greater than 1, causing an ISE during the releaseAndAcquire.

swankjesse: Yeah good thinking. Here's the moving parts currently being balanced:

 * Keeping `streamAllocation.connection` in sync with `connection.allocations`.
 * Assigning `streamAllocation.connection` before the connect completes so that it can be canceled in flight. 
 * Not leaking anything, and detecting when we do.

I’d like to try and force the bad interleavings, maybe with a torture test. I'll see how that goes.

dave-r12: Sorry, it may be more trouble than it's worth. Here were my notes.

Test program: https://gist.github.com/dave-r12/d029c81e927b56a23bcc4b26a9f53b4c

Given connections (C1, C2, C3):
- Break points on StreamAllocation: 121, 176, 184.
- Let C1 proceed.
- Let C2 proceed.
- Let C2 proceed (now in the pool but stalled on connect).
- Let C1 proceed.
- Let C1 proceed (C1 is now the primary H2 connection).
- Step over C2 (now connected, allocationLimit is upped for H2).
- Step over C3 (it found the connection in the pool from C2.)
- Let C2 proceed.

swankjesse: Oooh, yeah that’ll break things. So we allow a stream allocation to hold a connection before it’s in the pool? That’s not too terrible. I think this code needs a pass of simplification – it’s too subtle.

dave-r12: Yes, I think the ideal time to share the connection is after learning it's H2 and it's not a duplicate. I hear ya on the other things that need to be balanced as well. I'll think through that a bit more.

👍  for the pass of simplification. It all seems necessary, but if there is a way to simplify that would be a win.

swankjesse: I’ve added code to defer pooling. I think it’s a net win.

I haven’t yet done the simplifying step. I’ll do that in follow-up.

PR #3089, opened 2017-01-07
Make the crawler a little better.

swankjesse:
Name threads by their URLs. Don't follow all fragments of the same URL.


swankjesse:
https://github.com/square/okhttp/issues/3090

PR #3096, opened 2017-01-09
Website only: Update first example to use try with resources

rbygrave:
This changes the first example on the website from using try/finally to using try with resources.

This fixes 2 issues with my previous PR for the first website example.
- The actual linked GetExample.java uses try with resources (not try finally so my change didn't match the linked GetExample.java)
- My try/finally example had issues.  It didn't initialise the Response properly and close if non-null etc

Apologies for getting this wrong in the first PR. I think this fixes those 2 issues in my previous PR and I think it is nice if the first example code seen is as simple and clean (and correct) as possible (I think try with resources is a bit cleaner/nicer than try/finally).

### DIFF
--- a/website/index.html
+++ b/website/index.html
@@ -78,11 +78,8 @@ String run(String url) throws IOException {
       .url(url)
       .build();
 
-  try {
-    Response response = client.newCall(request).execute();
+  try (Response response = client.newCall(request).execute()) {
     return response.body().string();
-  } finally {
-    response.close();
   }
 }
 </pre>


rbygrave:
> users are building with tools that don't worry try with resources

Yes. I'll push up a separate 3rd example with a separate try/finally example.  It uses `Util.closeQuietly()` which you may consider internal and wish to discourage (so maybe this isn't quite right) but I'll put it up for review.

swankjesse:
Yep. closeQuietly is in `internal` and not for our users!

PR #3097, opened 2017-01-09
Website and Guide only: Add try/finally example (closing Response)

rbygrave:
Note this example uses `okhttp3.internal.Util.closeQuietly()` ... to close potentially null `Response`.  I'm not sure if you want to encourage that or not.

### DIFF
--- a/samples/guide/src/main/java/okhttp3/guide/GetTryFinallyExample.java
+++ b/samples/guide/src/main/java/okhttp3/guide/GetTryFinallyExample.java
@@ -3,7 +3,6 @@ package okhttp3.guide;
 import okhttp3.OkHttpClient;
 import okhttp3.Request;
 import okhttp3.Response;
-import okhttp3.internal.Util;
 
 import java.io.IOException;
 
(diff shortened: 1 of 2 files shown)


website/index.html
```diff
@@ -107,6 +107,33 @@ <h3 id="examples">Examples</h3>
 }
 </pre>
 
+            <h4>Try/Finally example</h4>
```
swankjesse: I think this might be a distraction from our otherwise spartan docs. Wanna just fix the first example to use try/finally?

samples/guide/src/main/java/okhttp3/guide/GetTryFinallyExample.java
```diff
@@ -0,0 +1,37 @@
...
+import okhttp3.Request;
+import okhttp3.Response;
+import okhttp3.internal.Util;
+
+import java.io.IOException;
+
+/**
+ * Example using try/finally (typically when try with resources is not available).
+ */
+public class GetTryFinallyExample {
```
swankjesse: This might be overkill.

swankjesse:
Sorry for sending you in circles on this. I really want our docs to show off OkHttp’s API, and not tour the madness that is writing defensive Java code. Maybe the standalone try block is sufficient.

Gross sample code is gross.

yschimke:
Closing after a year

Issue #3115, opened 2017-01-17
Android init crash after add cache in some devices.


SethFeng:
Crash stack:
```
OkHttp Dispatcher
java.lang.IllegalStateException
cache is closed
okhttp3.internal.DiskLruCache.void checkNotClosed()(SourceFile:644)
okhttp3.internal.DiskLruCache.okhttp3.internal.DiskLruCache$Editor edit(java.lang.String,long)(SourceFile:458)
okhttp3.internal.DiskLruCache.okhttp3.internal.DiskLruCache$Editor edit(java.lang.String)(SourceFile:452)
okhttp3.Cache.okhttp3.internal.http.CacheRequest put(okhttp3.Response)(SourceFile:244)
okhttp3.Cache.access$000(SourceFile:135)
okhttp3.Cache$1.okhttp3.internal.http.CacheRequest put(okhttp3.Response)(SourceFile:147)
okhttp3.internal.http.HttpEngine.void maybeCache()(SourceFile:438)
okhttp3.internal.http.HttpEngine.void readResponse()(SourceFile:683)
okhttp3.RealCall.okhttp3.Response getResponse(okhttp3.Request,boolean)(SourceFile:244)
okhttp3.RealCall$ApplicationInterceptorChain.okhttp3.Response proceed(okhttp3.Request)(SourceFile:201)
okhttp3.logging.HttpLoggingInterceptor.okhttp3.Response intercept(okhttp3.Interceptor$Chain)(SourceFile:212)
okhttp3.RealCall$ApplicationInterceptorChain.okhttp3.Response proceed(okhttp3.Request)(SourceFile:190)
okhttp3.RealCall.okhttp3.Response getResponseWithInterceptorChain(boolean)(SourceFile:163)
okhttp3.RealCall.access$100(SourceFile:30)
okhttp3.RealCall$AsyncCall.void execute()(SourceFile:127)
okhttp3.internal.NamedRunnable.void run()(SourceFile:32)
java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1113)
java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:588)
java.lang.Thread.run(Thread.java:818)
```

log(may be useful):
```
[...]

PR #3120, opened 2017-01-18
Added test for cache initialization failure and recovery.

dave-r12:
https://github.com/square/okhttp/issues/3115


swankjesse:
I’m getting to be very happy with the fake file system abstraction. Should Okio inherit this?

swankjesse:
(sorry, off topic. Let’s discuss [here](https://github.com/square/okio/issues/282).)

PR #3123, opened 2017-01-19
Connection coalescing - feature/help wanted

yschimke:
Was playing around with implementing Connection Coalescing as described here.  DNS and subject alternative name match.

Also allowed for manual configs e.g. for 1st party domains.

a few questions

1) Is this a useful feature for okhttp3, potentially moved from an interceptor into OkHttpClient?
2) Are there ways to access the subject alternative names from the ConnectionPool? without code changes, or are those code changes appropriate?

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/ConnectionCoalescingTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/ConnectionCoalescingTest.java
@@ -1,0 +1,100 @@
+package okhttp3;
+
+import java.io.IOException;
+import java.util.List;
+import okhttp3.coalescing.ConnectionInterceptor;
+import okhttp3.coalescing.ConnectionTarget;
+import org.junit.Before;
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+
+public class ConnectionCoalescingTest {
+  private OkHttpClient client;
+  private ConnectionTarget target = new ConnectionTarget();
+
+  @Before
+  public void setup() {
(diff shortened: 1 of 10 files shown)


swankjesse:
This is pretty cool! I think the most _um_ OkHttp way to do this would be to make `Connection.isEligible()` return true when connections can be coalesced. This means relaxing the current code from this:

```
  /** Returns true if this connection can carry a stream allocation to {@code address}. */
  public boolean isEligible(Address address) {
    return allocations.size() < allocationLimit
        && address.equals(route().address())
        && !noNewStreams;
  }
```

to something like this:

```
  /** Returns true if this connection can carry a stream allocation to {@code address}. */
  public boolean isEligible(Address address) {
    return allocations.size() < allocationLimit
        && canCarryAddress(handshake(), route(), address)
        && !noNewStreams;
  }
```

What’s also kind of neat is that the recent HTTP/2 connection deduplication should also work here.

yschimke:
That sounds good, I'll try that.  should I target a clean change in okhttp?  or keep this as an application interceptor.  It can't work as network interceptor as it changes the url.

The downsides to being an application interceptor are 
1) new url is used in cache
2) network interceptors may not kick in based on updated host e.g. auth

swankjesse:
I think this would work better without the interceptor. I think you should be able to make it work for subject alt names like `*.twitter.com`. You'll just need to consult the handshake. And maybe rework the code that does hostname matching so we don't have to parse the certificate multiple times.

yschimke:
will reopen a new PR when its ready

PR #3125, opened 2017-01-20
Use Socket instead of Closeable for compatibility.

### LINKED ISSUE #3124: IncompatibleClassChangeError from okhttp3.internal.Util.closeQuietly
Using 3.6.0-SNAPSHOT, we are seeing `IncompatibleClassChangeError` exceptions thrown from `okhttp3.internal.Util.closeQuietly()` on Android versions prior to 4.4. It can be reliably reproduced, but it's challenging to write a test, as we're only seeing the error consistently when the server socket closes unexpectedly. So I've described steps to reproduce here.

Here's the stack:
```
Caused by: java.lang.IncompatibleClassChangeError: interface not implemented
at okhttp3.internal.Util.closeQuietly(Util.java:100)
at okhttp3.internal.connection.StreamAllocation.streamFailed(StreamAllocation.java:332)
at okhttp3.internal.http.RetryAndFollowUpInterceptor.recover(RetryAndFollowUpInterceptor.java:209)
at okhttp3.internal.http.RetryAndFollowUpInterceptor.intercept(RetryAndFollowUpInterceptor.java:132)
at okhttp3.logging.HttpLoggingInterceptor.intercept(HttpLoggingInterceptor.java:212)
at okhttp3.RealCall.getResponseWithInterceptorChain(RealCall.java:179)
at okhttp3.RealCall.execute(RealCall.java:63)
```

To reproduce the problem, we hosted the following php (borrowed from [http://stackoverflow.com/a/141026/315702](http://stackoverflow.com/a/141026/315702)) on a server:

```
<?php
ob_end_clean();
header("Connection: close");
ignore_user_abort(true); // just to be safe
ob_start();
echo('Text the user will see');
$size = ob_get_length();
header("Content-Length: $size");
ob_end_flush(); // Strange behaviour, will not work
flush(); // Unless both are called !
sleep(30);
echo('Text user will never see');
?>
```

This is a boiled-down version of our okhttp call:

```
Context context = aParameter;
String body = anotherParameter;
File cacheDirectory = context.getDir("ts_okhttp_cache", Context.MODE_PRIVATE);
int cacheSize = 5 * 1024 * 1024; // 5 MiB
Cache fileCache = new Cache(cacheDirectory, cacheSize);
OkHttpClient.Builder clientBuilder = new OkHttpClient.Builder();
clientBuilder.cache(fileCache);
OkHttpClient client = clientBuilder.build();
MediaType jsonMediaType = MediaType.parse("application/json; charset=utf-8");
Request.Builder requestBuilder = new Request.Builder();
requestBuilder.url(urlString);
requestBuilder.post(RequestBody.create(jsonMediaType, body));
return client.newCall(requestBuilder.build()).execute();
```

Every time we make a call to that php page, using Android emulators running Android 4.3 or earlier, we get the `IncompatibleClassChangeError` exception. Just guessing, but it seems like this could be due to a lack of bytecode backward compatibility with Java 6, where `java.net.Socket` doesn't implement `Closeable`.

When we switch to okhttp 3.5.0, the problem goes away. (Same goes for 3.4.1.)

A note for any other developers who have run into this and are puzzled, because they don't think they're using 3.6.0-SNAPSHOT... make sure to check the dependencies of your app. You probably have a third-party library that's pulling in okhttp:+. In our case, it was the okhttp extension for exoplayer, and we were able to block the dependency with the following change to our build.gradle file:

```
compile('com.google.android.exoplayer:extension-okhttp:r2.0.4') {
  exclude module: 'okhttp'
}
```

JakeWharton:
Versions prior to Android 4.4 (aka Java 6) did not have Socket implementing Closeable.

Closes #3124.

### DIFF
--- a/okhttp/src/main/java/okhttp3/ConnectionPool.java
+++ b/okhttp/src/main/java/okhttp3/ConnectionPool.java
@@ -16,8 +16,8 @@
  */
 package okhttp3;
 
-import java.io.Closeable;
 import java.lang.ref.Reference;
+import java.net.Socket;
 import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Deque;
@@ -130,7 +130,7 @@ public final class ConnectionPool {
    * Replaces the connection held by {@code streamAllocation} with a shared connection if possible.
    * This recovers when multiple multiplexed connections are created concurrently.
    */
-  Closeable deduplicate(Address address, StreamAllocation streamAllocation) {
+  Socket deduplicate(Address address, StreamAllocation streamAllocation) {
     assert (Thread.holdsLock(this));
     for (RealConnection connection : connections) {
       if (connection.isEligible(address)
--- a/okhttp/src/main/java/okhttp3/OkHttpClient.java
+++ b/okhttp/src/main/java/okhttp3/OkHttpClient.java
@@ -15,10 +15,10 @@
  */
 package okhttp3;
 
-import java.io.Closeable;
 import java.net.MalformedURLException;
 import java.net.Proxy;
 import java.net.ProxySelector;
+import java.net.Socket;
 import java.net.UnknownHostException;
 import java.security.GeneralSecurityException;
 import java.security.KeyStore;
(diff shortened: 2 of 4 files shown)


JakeWharton:
Build failure on JDK 8 is unrelated. Fix @ https://github.com/square/okhttp/pull/3126

civilordergone:
When will RetroFit contain an updated POM.xml to point to this patch?

JakeWharton:
First of all this was never released so there's nothing broken. Second, you
can always specify the OkHttp version explicitly in your build system to
pull in newer versions.

PR #3127, opened 2017-01-21
Recover from a severe cache initialization failure.

dave-r12:
Closes: https://github.com/square/okhttp/issues/3115


swankjesse:
Dig!

PR #3128, opened 2017-01-21
Connection coalescing

yschimke:
As described here

https://hpbn.co/optimizing-application-delivery/#eliminate-domain-sharding
https://tools.ietf.org/html/rfc7540#section-9.1.1

when the following is true
- HTTP/2 so we don't overload non multiplexed connections
- HTTPS since we need subjectAltNames (n.b. rfc7540 defines as safe to use over plaintext, based purely on DNS) 
- desired url host is part of the subjectAltNames for an existing connection
- One of the DNS results for desired host match the established connection
- HostnameVerifier is *not* overriden
- ConnectionPinner accepts the existing certificate chain

### Possibly

From https://tools.ietf.org/html/rfc7540#section-9.1.1

> A server that does not wish clients to reuse connections can indicate
> that it is not authoritative for a request by sending a 421
> (Misdirected Request) status code in response to the request (see
> Section 9.1.2).


okhttp/src/main/java/okhttp3/internal/connection/RealConnection.java
```diff
@@ -276,6 +280,9 @@ private void connectTls(ConnectionSpecSelector connectionSpecSelector) throws IO
             + "\n    subjectAltNames: " + OkHostnameVerifier.allSubjectAltNames(cert));
       }
 
+      X509Certificate cert = (X509Certificate) unverifiedHandshake.peerCertificates().get(0);
+      supportedHosts = new HashSet<>(OkHostnameVerifier.allSubjectAltNames(cert));
```
swankjesse: nit: https://publicobject.com/2016/02/08/linkedhashmap-is-always-better-than-hashmap/

yschimke:
Without new APIs

Auto coalescing would have to be enabled by default.  This seems to be what Firefox/Chrome do, are you ok with that?

Also documented here https://hpbn.co/optimizing-application-delivery/#eliminate-domain-sharding

Interesting discussion [here](https://community.akamai.com/thread/4327-does-connection-coalesce-work-when-receiving-resources-from-multiple-edge-servers) also points to the certs used by google

```
$ openssl s_client -connect google.com:443 | openssl x509 -noout -text | grep DNS
depth=2 /C=US/O=GeoTrust Inc./CN=GeoTrust Global CA
                DNS:*.google.com, DNS:*.android.com, DNS:*.appengine.google.com, DNS:*.cloud.google.com, DNS:*.gcp.gvt2.com, DNS:*.google-analytics.com, DNS:*.google.ca, DNS:*.google.cl, DNS:*.google.co.in, DNS:*.google.co.jp, DNS:*.google.co.uk, DNS:*.google.com.ar, DNS:*.google.com.au, DNS:*.google.com.br, DNS:*.google.com.co, DNS:*.google.com.mx, DNS:*.google.com.tr, DNS:*.google.com.vn, DNS:*.google.de, DNS:*.google.es, DNS:*.google.fr, DNS:*.google.hu, DNS:*.google.it, DNS:*.google.nl, DNS:*.google.pl, DNS:*.google.pt, DNS:*.googleadapis.com, DNS:*.googleapis.cn, DNS:*.googlecommerce.com, DNS:*.googlevideo.com, DNS:*.gstatic.cn, DNS:*.gstatic.com, DNS:*.gvt1.com, DNS:*.gvt2.com, DNS:*.metric.gstatic.com, DNS:*.urchin.com, DNS:*.url.google.com, DNS:*.youtube-nocookie.com, DNS:*.youtube.com, DNS:*.youtubeeducation.com, DNS:*.ytimg.com, DNS:android.clients.google.com, DNS:android.com, DNS:developer.android.google.cn, DNS:g.co, DNS:goo.gl, DNS:google-analytics.com, DNS:google.com, DNS:googlecommerce.com, DNS:urchin.com, DNS:www.goo.gl, DNS:youtu.be, DNS:youtube.com, DNS:youtubeeducation.com

$ openssl s_client -connect graph.facebook.com:443 2>&1 | openssl x509 -noout -text | grep DNS
                DNS:*.facebook.com, DNS:facebook.com, DNS:*.facebook.net, DNS:*.fb.com, DNS:*.fbcdn.net, DNS:*.fbsbx.com, DNS:*.m.facebook.com, DNS:*.messenger.com, DNS:*.xx.fbcdn.net, DNS:*.xy.fbcdn.net, DNS:*.xz.fbcdn.net, DNS:fb.com, DNS:messenger.com
```

The manual mode probably came from the choice of my first test host

```
$ openssl s_client -connect twitter.com:443 2>&1 | openssl x509 -noout -text | grep DNS
                DNS:twitter.com, DNS:www.twitter.com
$ openssl s_client -connect api.twitter.com:443 2>&1 | openssl x509 -noout -text | grep DNS
                DNS:api.twitter.com
$ openssl s_client -connect pbs.twimg.com:443 2>&1 | openssl x509 -noout -text | grep DNS
                DNS:*.twimg.com, DNS:twimg.com, DNS:si0.twimg.com, DNS:gu.twimg.com, DNS:g2.twimg.com, DNS:v.cdn.vine.co, DNS:mtc.cdn.vine.co, DNS:cdn.api.twitter.com, DNS:platform.twitter.com, DNS:preview.cdn.twitter.com, DNS:p.twimg.com, DNS:o.twimg.com, DNS:si1.twimg.com, DNS:si2.twimg.com, DNS:si3.twimg.com, DNS:si4.twimg.com, DNS:si5.twimg.com, DNS:cdn-dev.api.twitter.com, DNS:pbs.twimg.com, DNS:jp.twimg.com, DNS:abs.twimg.com, DNS:hca.twimg.com, DNS:ea.twimg.com, DNS:widgets.twimg.com, DNS:cdn.syndication.twimg.com, DNS:dnt.twimg.com, DNS:g.twimg.com, DNS:widgets.platform.twitter.com, DNS:ton.twimg.com, DNS:v.twimg.com, DNS:vmtc.twimg.com, DNS:tailfeather.twimg.com, DNS:ma.twimg.com, DNS:status.twitter.com, DNS:cdn.syndication.twitter.com, DNS:cdn.digits.com, DNS:video.twimg.com, DNS:pbs-ec.twimg.com
```

So in conclusion: ditch the manual modes and implement always on, or a single boolean.

n.b. It would also remove the domain fronting support, without your own application interceptor.  But that is probably advanced enough that its outside the default okhttp remit.

Thoughts?

yschimke:
n.b. is it safe to assume DNS caching makes repeated calls performant?  I know the JVM usually caches pretty aggressively and setting networkaddress.cache.ttl to 60 is also common.

What about on Android?

swankjesse:
I’m not very confident that this change will be a big win performance-wise. I assume most coalescing will impact a small number of hostnames, and a few CDNs will impact a large number.

So I think this feature should only land if it can be done safely with minimal code & API changes. That means picking a simple policy (like Chrome’s or Firefox’s) and limiting configuration complexity.

swankjesse:
(The current PR looks great for those requirements: a few local changes is a win.)

swankjesse:
On DNS we should query each hostname only once if we can. And we definitely shouldn’t query DNS while holding a lock on the connection pool.

yschimke:
The DNS issue is awkward since we may not need to do a DNS lookup if there is an existing connection.  So by the time we expect that we need a DNS lookup we are inside the connectionPool lock.

swankjesse:
Yeah, I think that’s a bit annoying to code but the behavior is what we want.

yschimke:
It sounds reasonable and addresses some of the concerns I had, but enough change that I wanted to confirm the logic before implementing.  I think this might be for next weekend.

Thanks for the reviews and feedback.

yschimke:
Sorry to be chatty, what is the expected behaviour for proxies?

- only use coalescing for direct connections
- allow with SOCKS/HTTP Connect proxies*
- other?

* the rules seem mostly unchanged, however instead of the SOCKS5 or HTTP proxy being able to DNS lookup a DNS name, okhttp will be trying an additional DNS lookup possibly without being able to resolve.

swankjesse:
We don’t have enough information to coalesce connections through our HTTP or SOCKS proxy. The client knows the proxy’s IP but not the origin server’s, and IP address knowledge is required to coalesce connections.

Confirming whether Chrome & Firefox do something here is interesting.

yschimke:
I'll start with direct only.

I already asked the question for Firefox - https://twitter.com/yschimke/status/825648492743950336

okhttp/src/main/java/okhttp3/Address.java
```diff
@@ -46,6 +52,8 @@
   final SSLSocketFactory sslSocketFactory;
   final HostnameVerifier hostnameVerifier;
   final CertificatePinner certificatePinner;
+  private Map<String, List<InetAddress>> cachedDnsResults =
```
yschimke: Happy for suggestions about this approach.  it effectively leaks memory as we retain the dns results.  I could clear them or find another way.

yschimke:
Currently fails on

RealInterceptorChain:
```
   private boolean sameConnection(HttpUrl url) {
    return url.host().equals(connection.route().address().url().host())
        && url.port() == connection.route().address().url().port();
  }
```

I'll work on a fix and test

yschimke:
@swankjesse I believe this is functionally complete and works nicely for a number of sites that automatically redirect okhttp clients from www.x.com to mobile.x.com.

I realise it is a big change and requires a thorough review including the security aspects, but I'm pretty sure there are also some basic things I'll need to cleanup.

swankjesse:
Awesome. Gimme a couple days to review this.

yschimke:
@swankjesse Curious if you have a chance to look in the next week. Or even a "this will never ship" :)

swankjesse:
I’m on it. I’ve had a couple busy weeks.

okhttp/src/main/java/okhttp3/Address.java
```diff
@@ -195,4 +207,25 @@ public CertificatePinner certificatePinner() {
     result.append("}");
     return result.toString();
   }
+
+  public boolean isCoalescable() {
```
swankjesse: ... we use `Internal` to hack around visibility

okhttp/src/main/java/okhttp3/Address.java
```diff
@@ -195,4 +207,25 @@ public CertificatePinner certificatePinner() {
...
+    return url().isHttps()
+        && this.proxy() == null
+        && hostnameVerifier == OkHostnameVerifier.INSTANCE;
+  }
+
+  public List<InetAddress> dnsLookup(String hostname) throws UnknownHostException {
```
swankjesse: This doesn’t seem like the right place. The side effect is implicit but it should probably be explicit

swankjesse: In particular, why isn’t `hostname` just the hostname of this address? That’s surprising

okhttp/src/main/java/okhttp3/internal/connection/RealConnection.java
```diff
@@ -104,6 +108,10 @@
 
   /** Nanotime timestamp when {@code allocations.size()} reached zero. */
   public long idleAtNanos = Long.MAX_VALUE;
+  private List<String> subjectAlternativeNames;
+
+  private Map<String, Boolean> supportedHostCache =
```
swankjesse: probably wants Javadoc . . . supports _what_ ?
Also final.
Also `Map<String, Boolean>` is kind of a weird type; usually that’s a `Set<String>`...

okhttp/src/main/java/okhttp3/internal/connection/RealConnection.java
```diff
@@ -276,6 +284,9 @@ private void connectTls(ConnectionSpecSelector connectionSpecSelector) throws IO
             + "\n    subjectAltNames: " + OkHostnameVerifier.allSubjectAltNames(cert));
       }
 
+      X509Certificate cert = (X509Certificate) unverifiedHandshake.peerCertificates().get(0);
+      subjectAlternativeNames = OkHostnameVerifier.allSubjectAltNames(cert);
```
swankjesse: ideally we compute this once before line 278, then reuse the parsed subjectAlternativeNames in the call to `verify()`

swankjesse: ... again, not necessarily for this PR. But add a `TODO` comment? Seems wasteful to parse 2x. 

(calling directly through the hostname verifier is annoying because HostnameVerifier is a public type)

okhttp/src/main/java/okhttp3/internal/connection/RealConnection.java
```diff
@@ -376,8 +387,78 @@ private Request createTunnelRequest() {
...
-        && address.equals(route().address())
-        && !noNewStreams;
+        && !noNewStreams
+        && canCarryAddress(address);
+  }
+
+  private boolean canCarryAddress(Address address) {
+    boolean directMatch = address.equals(route().address());
+
+    if (directMatch) {
```
swankjesse: nit: inline this?

okhttp/src/main/java/okhttp3/internal/connection/RealConnection.java
```diff
@@ -376,8 +387,78 @@ private Request createTunnelRequest() {
...
+      return true;
+    }
+
+    if (address.isCoalescable() && this.isCoalescable() && address.equalsNonHost(route.address())) {
```
swankjesse: As I read it, there’s four requirements here:

 * address.isCoalescable
 * this.isCoalescable
 * address.equalsNonHost(route.address())
 * supportsHost()

Should this just be one expression?

okhttp/src/main/java/okhttp3/internal/connection/RealConnection.java
```diff
@@ -376,8 +387,78 @@ private Request createTunnelRequest() {
...
+  private boolean supportsHost(Address address) {
+    String addressHost = address.url().host();
+
+    Boolean supportsHost = supportedHostCache.get(addressHost);
+
+    if (supportsHost != null) {
+      return supportsHost;
+    }
+
+    // attempt to coalesce secure HTTP/2 connections by subjectAltNames
```
swankjesse: nit: http://nedbatchelder.com/blog/201401/comments_should_be_sentences.html

okhttp/src/main/java/okhttp3/internal/connection/RealConnection.java
```diff
@@ -376,8 +387,78 @@ private Request createTunnelRequest() {
...
+    // currently does not apply to HTTP/1.1 because reuse assumptions are not the same
+    // e.g. app may use domain sharding for performance
+    boolean hostMatch = false;
+    for (String subjectAlternativeName : subjectAlternativeNames) {
+      if (OkHostnameVerifier.verifyHostname(addressHost, subjectAlternativeName)) {
```
swankjesse: this is pretty neat

okhttp/src/main/java/okhttp3/internal/connection/RealConnection.java
```diff
@@ -376,8 +387,78 @@ private Request createTunnelRequest() {
...
+        hostMatch = true;
+        break;
+      }
+    }
+    if (!hostMatch) {
+      supportedHostCache.put(addressHost, false);
```
swankjesse: I’m tempted to drop the cache. I think it’s a bit weird to cache local in-memory computation

okhttp/src/main/java/okhttp3/internal/connection/RealConnection.java
```diff
@@ -376,8 +387,78 @@ private Request createTunnelRequest() {
...
+      return false;
+    }
+
+    List<InetAddress> ips = address.cachedDnsResults(addressHost);
+    if (ips == null) {
+      // called on first pass without pre-emptive DNS lookup
+      return false;
+    }
+
+    boolean result = ips.contains(route.socketAddress().getAddress());
```
swankjesse: oh interesting, this case is pretty tricky

yschimke:
a big +1 to any changes you want to do on your branch.  Thanks for the thorough review, it's a pretty easy change to let something painful slip through.

okhttp/src/main/java/okhttp3/internal/connection/RealConnection.java
yschimke: I want to cache positive and negative results.  mainly because the previous lookup was exact match, mostly reference equality to find the right connection.  I wanted to avoid doing all the work on the 2nd, 3rd etc uses of a coalesced connection.

okhttp/src/main/java/okhttp3/internal/connection/RealConnection.java
yschimke: Yep, I guess caching should come from a benchmark with realistic traffic pattern

swankjesse:
I extracted & adjusted the APIs for your FakeDns changes:
https://github.com/square/okhttp/pull/3206

swankjesse:
It’s long overdue, but I spent my afternoon revising your change, and have pushed my update here. The commit is under your name; you deserve credit for doing all of the work to figure out how to slot this in.
https://github.com/square/okhttp/pull/3207/files

I dropped all caching of the the subject alternative names, and of matching and non-matching hosts. This is potentially slower than your PR, but it’s also a smaller change. If we want to bring this caching we can do it in a separate PR. I suspect we might be better off optimizing HostnameVerifier vs. caching its output.

I also changed when & where DNS resolution is performed. Prior to any connection coalescing this is what StreamAllocation does when allocating a connection:

1. Create an Address as a cache key for a connection.
2. Look in the connection pool for a matching connection. If found, return.
3. Do a potentially slow DNS lookup.
4. Do a potentially slow TCP connect.
5. Add the connection into the pool, de-duplicating it if possible.

I’ve changed the process to add a new step after DNS resolution and before TCP connect.
3.5. Take another look through the connection pool. This 2nd pass supports coalescing.

Overall the new pull request is simpler but less optimized than this PR. If necessary we should optimize it further in follow-up changes.

okhttp/src/main/java/okhttp3/Address.java
```diff
@@ -152,20 +156,24 @@ public CertificatePinner certificatePinner() {
...
+    return this.dns.equals(that.dns)
+        && this.proxyAuthenticator.equals(that.proxyAuthenticator)
+        && this.protocols.equals(that.protocols)
+        && this.connectionSpecs.equals(that.connectionSpecs)
+        && this.proxySelector.equals(that.proxySelector)
+        && equal(this.proxy, that.proxy)
+        && equal(this.sslSocketFactory, that.sslSocketFactory)
+        && equal(this.hostnameVerifier, that.hostnameVerifier)
+        && equal(this.certificatePinner, that.certificatePinner)
+        && this.url().port() == that.url().port();
```
yschimke: I don't think it matters, but should this also check url().scheme() as well?

This is built as 

https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/Address.java#L54-L58

Issue #1468, opened 2015-03-04
Charset support for Basic Auth


swankjesse:
http://datatracker.ietf.org/doc/draft-ietf-httpauth-basicauth-update/?include_text=1

swankjesse:
Done!

Issue #2911, opened 2016-10-13
Allow UTF-8 for Credentials.basic (as defined in RFC 7617)


rfc2822:
Currently, `Credentials.basic` uses ISO-8859-1 encoding for credentials, which seems to be correct as far as I understand it. However, many people say that encoding is not defined that exactly and [expect UTF-8](https://github.com/nextcloud/server/issues/1731), which [causes problems](https://forums.bitfire.at/topic/1230/problem-davdroid-1-3-2-2-on-android-5-0-nextcloud-10-0-1-on-ubuntu-16-04-on-my-own-vserver-is-not-working/7). Other renowned clients like curl also use UTF-8, at least with UTF-8 system locale.

As a solution, RFC 7617 has been published, which allows servers to request a specific character set (and the only one allowed is UTF-8). So, `Credentials.basic` could have a parameter `boolean utf8`, which allows to switch between ISO-8859-1 (default) and UTF-8 (RFC 7617) behaviour.

If you're interested in this, I can try to make a patch.

swankjesse:
Tracking as https://github.com/square/okhttp/issues/1468

If you’d like to prepare a pull request please do, but also please don’t write too much code. We.don’t generally like big PRs.

rfc2822:
I'd even switch to UTF-8 completely. The ISO-8859-1 is a raw guess and not really specified in the standard, so it might work or it might not, according to the server. Every client and server uses another encoding. The only way which will always comply with RFC 7617 is always using UTF-8 (which I have now chosen for dav4android/davdroid). What do you think about it?

swankjesse:
Seems like it could be backwards-incompatible.

PR #3134, opened 2017-01-24
Add charset support for Credentials.basic

erikghonyan:
Issue #2911, #1468

### DIFF
--- a/okhttp/src/main/java/okhttp3/Credentials.java
+++ b/okhttp/src/main/java/okhttp3/Credentials.java
@@ -15,7 +15,7 @@
  */
 package okhttp3;
 
-import java.io.UnsupportedEncodingException;
+import java.nio.charset.Charset;
 import okio.ByteString;
 
 /** Factory for HTTP authorization credentials. */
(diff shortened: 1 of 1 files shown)


swankjesse:
I don’t think this is consistent with the spec.
https://datatracker.ietf.org/doc/rfc7617/

erikghonyan:
Well, if I understand correctly the spec doesn't really force any charset.

>    The original definition of this authentication scheme failed to
>    specify the character encoding scheme used to convert the user-pass
>    into an octet sequence.  In practice, most implementations chose
>    either a locale-specific encoding such as ISO-8859-1 ([ISO-8859-1]),
>    or UTF-8 ([RFC3629]).  For backwards compatibility reasons, this
>    specification continues to leave the default encoding undefined, as
>    long as it is compatible with US-ASCII (mapping any US-ASCII
>    character to a single octet matching the US-ASCII character code).

And according to the following paragraph the only reason not to switch to UTF-8 completely is because there are sites that don't use UTF-8 and won't work with UTF-8.

> B.3.  Why not simply switch the default encoding to UTF-8?
> 
>    There are sites in use today that default to a local character
>    encoding scheme, such as ISO-8859-1 ([ISO-8859-1]), and expect user
>    agents to use that encoding.  Authentication on these sites will stop
>    working if the user agent switches to a different encoding, such as
>    UTF-8.
> 
>    Note that sites might even inspect the User-Agent header field
>    ([RFC7231], Section 5.5.3) to decide which character encoding scheme
>    to expect from the client.  Therefore, they might support UTF-8 for
>    some user agents, but default to something else for others.  User
>    agents in the latter group will have to continue to do what they do
>    today until the majority of these servers have been upgraded to
>    always use UTF-8.

So if the server supports UTF-8, shouldn't we allow the clients to use that charset? The patch doesn't change the default charset, it's still ISO-8859-1, but if the clients want to use UTF-8 they can.

swankjesse:
Yep, you’re right. I’m sorry, I was misreading this thinking it’s the clients that specify a `charset` parameter. I’m wrong; that’s the server’s job.

Any thoughts on whether this should take a `Charset` instead of a `String` to identify the charset? I like the strongly-typed one myself because it discourages clients from passing through garbage from the incoming `Challenge` header.

erikghonyan:
Actually that's what I wanted to do at first! I like strongly-typed code too, just didn't want to make too many changes :) Updated the patch.

swankjesse:
Thanks!

erikghonyan:
@swankjesse I think we should close #1468

swankjesse:
Should we add `charset` to `Challenge` ?

erikghonyan:
Here you go #3144

PR #3137, opened 2017-01-27
HUC regression test: disconnect in the middle of connecting.

15characterlimi:
Prior to commit 084b06b48bae2b566bb1be3415b6c847d8ea3682 (OkHttp 3.4.0),
a HttpURLConnection.disconnect() in the middle of connecting caused an
infinite loop. HUC does not claim to be thread safe, but concurrent
disconnect should generally be supported (it is supported by HttpEngine)
and this bug occurred even when the disconnect() happened in the same
thread, e.g. through the CookieJar.

The infinite loop prior to that CL occurred because
 - The loop in HttpURLConnectionImpl.java:418 did not check for the
   disconnected state,
 - {StreamAllocation,RetryAndFollowUpInterceptor}.recover() returns
   true for canceled StreamAllocations/calls, but
   StreamAllocation.newStream() immediately fails if already canceled.

This bug was specific to HUC and did not affect the Call API because
RetryAndFollowUpInterceptor's (and, in earlier OkHttp versions,
Call.getResponse()'s) infinite loop did check the canceled case
before proceeding.

The bug existed since at least OkHttp 2.7.5, likely introduced by
commit c358656c8799d30fd422448153e99a5dd37e298a (OkHttp 2.7.0).

The new test asserts that a disconnect() while constructing the
cookie headers leads to the new connection being aborted. It would
also be permissible for the connection to succeed (as long as it
doesn't deadlock like it used to), but the stricter test seems
reasonable to ensure that any behavior change is deliberate.
A disconnect() while accessing the Cache FileSystem has the same
effect, but this is not covered by the added test because a
new FileSystem wrapper class that delegates all calls would
create code maintenance overhead.

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/URLConnectionTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/URLConnectionTest.java
@@ -15,11 +15,14 @@
  */
 package okhttp3;
 
+import java.io.File;
+import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.net.Authenticator;
 import java.net.ConnectException;
+import java.net.CookieHandler;
 import java.net.CookieManager;
 import java.net.HttpRetryException;
 import java.net.HttpURLConnection;
@@ -35,6 +38,7 @@ import java.net.URI;
 import java.net.URL;
 import java.net.URLConnection;
 import java.net.UnknownHostException;
+import java.nio.charset.Charset;
 import java.security.KeyStore;
 import java.security.cert.CertificateException;
 import java.security.cert.X509Certificate;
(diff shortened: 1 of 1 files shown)


15characterlimi:
FYI I don't know why the Travis CI failed. I observe a similar "The forked VM terminated without properly saying goodbye" error locally, even when "mvn clean verify"ing the master branch, but assumed it was just an issue with my environment. I'll try to look into it next week.

okhttp-tests/src/test/java/okhttp3/URLConnectionTest.java
```diff
@@ -1154,6 +1161,33 @@ private void initResponseCache() throws IOException {
     in.close();
   }
 
+  @Test public void disconnectDuringConnect_cookieJar() throws Exception {
+    final AtomicReference<HttpURLConnection> connectionHolder = new AtomicReference<>();
+    class DisconnectingCookieJar implements CookieJar {
+      @Override public void saveFromResponse(HttpUrl url, List<Cookie> cookies) { }
+      @Override
+      public List<Cookie> loadForRequest(HttpUrl url) {
+        connectionHolder.get().disconnect();
```
swankjesse: that’s a fun way to do it

swankjesse:
Failure is superscary but not related to this particular test.
https://github.com/square/okhttp/issues/3138

PR #3139, opened 2017-01-28
junit 5 compatibility for MockWebServer

yschimke:
Junit 5 support for MockWebServer via extending a known base class.

http://junit.org/junit5/docs/current/user-guide/#migrating-from-junit4-rulesupport

### DIFF
--- a/mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.java
+++ b/mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.java
@@ -76,9 +76,7 @@ import okio.ByteString;
 import okio.Okio;
 import okio.Sink;
 import okio.Timeout;
-import org.junit.rules.TestRule;
-import org.junit.runner.Description;
-import org.junit.runners.model.Statement;
+import org.junit.rules.ExternalResource;
 
 import static okhttp3.internal.Util.closeQuietly;
 import static okhttp3.mockwebserver.SocketPolicy.DISCONNECT_AFTER_REQUEST;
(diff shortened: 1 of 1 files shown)


mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.java
```diff
@@ -398,6 +379,14 @@ public synchronized void shutdown() throws IOException {
     }
   }
 
+  public synchronized void after() {
```
swankjesse: protected

swankjesse:
What the heck is wrong with the JUnit API designers. Breaking APIs like this is not a thing.

JakeWharton:
This is a breaking API changes so we can't merge until OkHttp 4.x. And even when that happens, I'd prefer separate artifacts for JUnit 4 and 5. Making MockWebServer depend directly on JUnit 4 was a huge mistake on my part.

swankjesse:
I don’t think this is a breaking change in OkHttp. ExternalResource still implements TestRule.

JakeWharton:
Ah, cool.

mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.java
```diff
@@ -142,7 +140,7 @@
 
   private boolean started;
 
-  private synchronized void maybeStart() {
+  protected synchronized void before() {
```
JakeWharton: Missing `@Override`?

mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.java
JakeWharton: Missing `@Override`?

yschimke:
Yeah, it looks worse than it is.  ExternalResource has been around since 4.7 and just removes some boilerplate.

I tried upgrading my code to junit5 and this was the only thing I was hitting. This isn't urgent but seems fully backwards compatible.

yschimke:
And no actual dependency on junit 5 anywhere.

JakeWharton:
Yeah I was reviewing on mobile so I jumped to a few conclusions. This does
look good.

swankjesse:
Thanks!

PR #3141, opened 2017-01-29
Confirm top-level domain cookie matching.

dave-r12:
I'm attempting to make my way through RFC 6265 and stumbled on this. Apparently, there is no algorithmic way to figure this out, so "_the list_" was created https://publicsuffix.org/.

I'm wondering if any action should be taken here. I came up with 3 options:

1. Don't bother: most applications know the origin server, so there is minimal risk.
2. Make it optional: Give users the option of adding a separate dependency that maintains this list. That way, users who don't use the functionality won't incur the cost, but those who do have an option.
3. Include it: This will increase the size of the library and add a size cost to those who don't use the functionality.

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/CookieTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/CookieTest.java
@@ -228,6 +228,13 @@ public final class CookieTest {
     assertFalse(cookie.matches(HttpUrl.parse("http://square.com")));
   }
 
+  @Test public void topLevelDomainMatches() {
+    Cookie cookie = Cookie.parse(url, "a=b; domain=com");
+    assertTrue(cookie.matches(HttpUrl.parse("http://example.com")));
+    assertTrue(cookie.matches(HttpUrl.parse("http://www.example.com")));
+    assertTrue(cookie.matches(HttpUrl.parse("http://square.com")));
+  }
+
   /** Ignore an optional leading `.` in the domain. */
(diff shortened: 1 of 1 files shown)


swankjesse:
Yikes. That definitely isn't good. I think we include the list.

dave-r12:
I just discovered Guava has this. I've seen the package in passing but never knew what it was for: https://github.com/google/guava/tree/master/guava/src/com/google/thirdparty/publicsuffix

swankjesse:
I'm trying to think about the best way to do this. The list is pretty huge and most users won't benefit from it. Ugh.

If we make it optional nobody will use it when they probably should. If we build it in everyone pays for it.

Hmmm.

swankjesse:
The publicsuffix file is 188K. Stripping comments shrinks it to 92 KiB. Zipping shrinks that to 32 KiB. That’s about 10% of okhttp.jar, which is sad, but acceptable.

Loading it on demand might work – probably if the `domain` attribute is specified in a cookie and is a suffix of the URL’s domain, then we confirm that it isn’t a public suffix.

If we end up loading it into memory it’d be nice to not create tons and tons of strings that will probably never be needed. Building a single 92 KiB `byte[]` that we binary search is potentially interesting.

Still, ugh.

dave-r12:
Agreed. Having the list may also be helpful for the cookie jar, to bucket cookies by domain. I'll see if I can implement what you described here.

PR #3143, opened 2017-01-29
Change MediaType's failure mode to not crash on charset problems.

swankjesse:
As-is it throws unchecked exceptions on unexpected charsets. This is a problem
because it can cause a misbehaving webserver to crash the client.

I don't expect this to break existing clients; returning 'null' has always
been a possibility; it's just returned in more cases.

### DIFF
--- a/okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
+++ b/okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
@@ -18,7 +18,6 @@ package okhttp3.logging;
 import java.io.EOFException;
 import java.io.IOException;
 import java.nio.charset.Charset;
-import java.nio.charset.UnsupportedCharsetException;
 import java.util.concurrent.TimeUnit;
 import okhttp3.Connection;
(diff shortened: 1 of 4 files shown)


okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
```diff
@@ -241,15 +240,7 @@ public Level getLevel() {
         Charset charset = UTF8;
         MediaType contentType = responseBody.contentType();
         if (contentType != null) {
-          try {
```
swankjesse: @JakeWharton FYI I had to remove this ’cause we no longer differentiate between unknown and absent. Thoughts?

JakeWharton: I don't have a strong opinion. A reasonable fallback seems fine to me. @dave-r12 added the original code here. Doesn't look like was motivated by a particular issue.

dave-r12: 👍

PR #3144, opened 2017-01-29
Add charset support for RFC 7617 challenges

### LINKED ISSUE #1468: Charset support for Basic Auth

erikghonyan:
Fixes #1468

### DIFF
--- a/okhttp/src/main/java/okhttp3/Challenge.java
+++ b/okhttp/src/main/java/okhttp3/Challenge.java
@@ -15,16 +15,23 @@
  */
 package okhttp3;
 
+import java.nio.charset.Charset;
 import okhttp3.internal.Util;
 
-/** An RFC 2617 challenge. */
+/** An RFC 7617 challenge. */
 public final class Challenge {
   private final String scheme;
   private final String realm;
+  private Charset charset;
 
   public Challenge(String scheme, String realm) {
     this.scheme = scheme;
     this.realm = realm;
+  }
+
+  public Challenge withCharset(Charset charset) {
+    this.charset = charset;
+    return this;
   }
 
   /** Returns the authentication scheme, like {@code Basic}. */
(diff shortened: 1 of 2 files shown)


okhttp/src/main/java/okhttp3/Challenge.java
```diff
@@ -17,14 +17,16 @@
...
 
-/** An RFC 2617 challenge. */
+/** An RFC 7617 challenge. */
 public final class Challenge {
   private final String scheme;
   private final String realm;
+  private final String charset;
 
-  public Challenge(String scheme, String realm) {
+  public Challenge(String scheme, String realm, String charset) {
```
erikghonyan: Do we need a constructor with just scheme and realm parameters?

erikghonyan:
Should I add a test for this or should I send a separate pull request for that?

okhttp/src/main/java/okhttp3/Challenge.java
swankjesse: We definitely need to keep the original constructor for compatibility. I was thinking something like this:

```
Challenge c = new Challenge("basic", "WallyWorld");
c = c.withCharset(StandardCharsets.UTF_8);
```

Basically keep the public constructor as-is and add a public factory method that adds a charset to an existing challenge.

okhttp/src/main/java/okhttp3/internal/http/HttpHeaders.java
```diff
@@ -158,17 +158,27 @@ public static Headers varyHeaders(Headers requestHeaders, Headers responseHeader
...
         if (header.regionMatches(true, matcher.start(1), "realm", 0, 5)) {
-          String scheme = header.substring(0, index);
-          String realm = matcher.group(3);
-          if (realm != null) {
-            challenges.add(new Challenge(scheme, realm));
-            break;
-          }
+          realm = matcher.group(3);
+        } else if (header.regionMatches(true, matcher.start(1), "charset", 0, 7)) {
+          charset = matcher.group(3);
```
swankjesse: dig

erikghonyan: Can you provide more details please?

swankjesse:
Merged manually. Thanks for your patience, @erikghonyan !

swankjesse:
https://github.com/square/okhttp/commit/b3ecb0a6c9bf3a213377b889f2b3077a940363e2

PR #3147, opened 2017-01-31
Do not fall back to COMPATIBLE_TLS (TLSv1) by default.

15characterlimi:
Insecure TLS version fallback is a compatibility workaround to connect
to servers that do not implement TLS protocol version negotiation
correctly; it is vulnerable to man-in-the-middle-attacks (POODLE).
Support for TLS version fallback was removed in Firefox 37 (Mar 2015)
and Chrome 50 (Apr 2016). As of late 2015, fewer than 0.01% of web
servers relied on protocol fallback.

After this commit, OkHttp still supports TLS version fallback, but it
is now opt-in rather than opt-out.

Existing tests for fallback functionality were modified to opt-in
to the fallback; new tests in CallTest and URLConnectionTest
assert that no fallback happens by default (when not opted-in).

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/CallTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/CallTest.java
@@ -1030,6 +1030,18 @@ public final class CallTest {
     noRecoverWhenRetryOnConnectionFailureIsFalse();
   }
 
+  @Test public void tlsHandshakeFailure_noFallbackByDefault() throws Exception {
+    server.useHttps(sslClient.socketFactory, false);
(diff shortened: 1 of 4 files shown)


okhttp-tests/src/test/java/okhttp3/CallTest.java
```diff
@@ -1030,6 +1030,18 @@ private void postBodyRetransmittedAfterAuthorizationFail(String body) throws Exc
     noRecoverWhenRetryOnConnectionFailureIsFalse();
   }
 
+  @Test public void tlsHandshakeFailure_noFallbackByDefault() throws Exception {
```
swankjesse: dig

PR #3148, opened 2017-02-01
Port basic DistinguishedNameParserTest from AOSP

15characterlimi:
Port the test from http://r.android.com/183814 . This consists of two commits:
(1.) minimal changes to make the test compile and runnable, (2.) update
expectations to current behavior to make the test pass.

DistinguishedNameParser is an internal API, so this test is mostly to ensure
that any future behavior changes are deliberate.


swankjesse:
Neat.

PR #3149, opened 2017-02-01
Strawman class for EventListener

mjbenedict:
Related to the discussion in #270, this a class that could be used to add
Zipkin or other traces to OkHttp requests

### DIFF
--- a/okhttp/src/main/java/okhttp3/EventListener.java
+++ b/okhttp/src/main/java/okhttp3/EventListener.java
@@ -15,7 +15,8 @@
  */
 package okhttp3;
 
-import java.io.IOException;
+import java.net.InetAddress;
+import java.util.List;
 
 /**
  * Observes, modifies, and potentially short-circuits requests going out and the corresponding
@@ -23,51 +24,52 @@ import java.io.IOException;
  * or response.
  */
 public class EventListener {
-	public static final EventListener NULL_EVENT_LISTENER = new EventListener() {
-	};
+  public static final EventListener NULL_EVENT_LISTENER = new EventListener() {
+  };
 
-	void requestEnqueued(Call call) {
-	}
+  void requestEnqueued(Call call) {
+  }
 
-	void dnsStart(Call call, String domainName) {
-	}
+  void dnsStart(Call call, String domainName) {
+  }
 
-	void dnsEnd(Call call, String domainName, List<InetAddress> inetAddressList, Throwable throwable) {
-	}
+  void dnsEnd(Call call, String domainName, List<InetAddress> inetAddressList,
+      Throwable throwable) {
+  }
 
-	void connectStart(Call call, InetAddress address, int port) {
-	}
-	
-	void secureConnectStart(Call call) {
-	}
+  void connectStart(Call call, InetAddress address, int port) {
+  }
 
-	void secureConnectEnd(Call call, String cipherSuite, String protocol, Throwable throwable) {
-	}
+  void secureConnectStart(Call call) {
+  }
 
-	void connectEnd(Call call,  InetAddress address, int port, String protocol, Throwable throwable) {
-	}
+  void secureConnectEnd(Call call, String cipherSuite, String protocol, Throwable throwable) {
+  }
 
-	void requestHeadersStart(Call call) {
-	}
+  void connectEnd(Call call,  InetAddress address, int port, String protocol, Throwable throwable) {
+  }
 
-	void requestHeadersEnd(Call call, Throwable throwable) {
-	}
+  void requestHeadersStart(Call call) {
+  }
 
-	void requestBodyStart(Call call) {
-	}
+  void requestHeadersEnd(Call call, Throwable throwable) {
+  }
 
-	void requestBodyEnd(Call call, Throwable throwable) {
-	}
+  void requestBodyStart(Call call) {
+  }
 
-	void responseHeadersStart(Call call) {
-	}
+  void requestBodyEnd(Call call, Throwable throwable) {
+  }
 
-	void responseHeadersEnd(Call call, Throwable throwable) {
-	}
+  void responseHeadersStart(Call call) {
+  }
 
-	void responseBodyStart(Call call) {
-	}
+  void responseHeadersEnd(Call call, Throwable throwable) {
+  }
 
-	void responseBodyEnd(Call call, Throwable throwable) {
-	}
+  void responseBodyStart(Call call) {
+  }
+
+  void responseBodyEnd(Call call, Throwable throwable) {
+  }
 }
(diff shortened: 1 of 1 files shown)


mjbenedict:
NB: I did not define the proxy connect before/after methods.  I know its transparent to my needs and I *think* it doesn't matter to Zipkin as long as the correct remote endpoint is known, which the IP address will indicate.  Handling proxy authentication, especially given potential retries is non-trivial and I did want to work toward the minimal viable solution.

okhttp/src/main/java/okhttp3/EventListener.java
```diff
@@ -0,0 +1,75 @@
...
+/**
+ * Observes, modifies, and potentially short-circuits requests going out and the corresponding
+ * responses coming back in. Typically interceptors add, remove, or transform headers on the request
+ * or response.
+ */
+public class EventListener {
+  public static final EventListener NULL_EVENT_LISTENER = new EventListener() {
+  };
+
+  void requestEnqueued(Call call) {
```
JakeWharton: What if you use `.execute()`? I think we'll want a `requestStarted(Call call)` which is called when OkHttp starts processing the request. `requestEnqueued` wouldn't be called for synchronous requests.

mjbenedict: Yes agree that we want something that starts the clock in either way... I was going to do `requestStart` but that is easy to get confused with `requestHeaderStart` and `requestBodyStart`... I am happy to defer to you guys what to name it.

JakeWharton: I also think we'll want to call this `Enqueue` not `Enqueued` since we use `Start` / `End` and not `Started` / `Ended`.

swankjesse: I was thinking we’d only call this if the request was async

swankjesse: these all probably wanna be public

okhttp/src/main/java/okhttp3/EventListener.java
```diff
@@ -0,0 +1,75 @@
...
+      Throwable throwable) {
+  }
+
+  void connectStart(Call call, InetAddress address, int port) {
+  }
+
+  void secureConnectStart(Call call) {
+  }
+
+  void secureConnectEnd(Call call, String cipherSuite, String protocol, Throwable throwable) {
```
swankjesse: do we want the Handshake instead?

okhttp/src/main/java/okhttp3/EventListener.java
JakeWharton: The entire listener? Or just this method? The problem with the latter is that there's no universal start method for capturing when OkHttp starts processing.

mjbenedict: Do we want one method like `fetchStart` for both `.execute()` and the enqueued state?  That makes the most sense to me personally.

JakeWharton: I want an event that indicates that the request has started executing regardless of whether it was submitted with `execute()` or `enqueue()`, yes, but I also want one that indicates when `enqueue()` is called so that I can measure the queue time.

okhttp/src/main/java/okhttp3/EventListener.java
```diff
@@ -0,0 +1,80 @@
...
+
+import java.net.InetAddress;
+import java.util.List;
+
+/**
+ * Observes, modifies, and potentially short-circuits requests going out and the corresponding
+ * responses coming back in. Typically interceptors add, remove, or transform headers on the request
+ * or response.
+ */
+public class EventListener {
```
adriancole: maybe add javadoc for each callback, mentioning whether they are re-entrant or not and if they are potentially terminal.

For example, in the interceptors design, the application interceptor is called once per semantic call, and the network interceptor is called once (or maybe not when served from cache?)

This will help us understand the state model. For example, if we do a single-span for the operation, will fetchEnd always be called, even if dnsEnd has a throwable?

adriancole:
zipkin doesn't really care about proxy as it has no prior art around that.. I think this is more or less how detailed users want to be. For example, they could add tags or otherwise to model proxy if it explained latency. Users making their own instrumentation may already do this, I don't know..

regardless, I agree that things like proxy case can be added later if desired.

JakeWharton:
I would like proxy included since Chrome DevTools wants it as an event.

okhttp/src/main/java/okhttp3/EventListener.java
mjbenedict: Very insightful request, I am still wrapping my head around this a little.  These methods would almost have to be re-entrant, no?  If OkHttp had to serialize these across workers it would probably have a significant performance impact.

Ideally if a *Start is called I think it is reasonable that *End would be called.  For the most part I could see that working but presently I have a tough time imagining an implementation that guarantees a `fetchEnd` for every `fetchStart`

adriancole: I think with the factory, the request is simpler. Without a factory, the question about reentrance is scoped to a call parameter, which makes my head hurt to explain :P On that note, I think since there's a factory now, the call parameters probably need to be squashed (unless the call parameters refer to a "network request")

Back to reentrance.. for the same call, would we ever see one of these callbacks twice? Which callbacks are optional? Which callbacks obviate others?

For example, connection callbacks are asymmetric wrt a call. We'd expect a start, end and abend to be on different calls (due to keepalives etc). However, a fetch start and end seems something that we can guarantee is one-one with a call. At least factory.create, end would be something we could guarantee.

The other thing is which callback would the exception raised from that connection break go to? Would a wrapped variant of that exception go somewhere else (with a root cause of the connection exception)? Mainly interested in the state diagram, knowing the whole design is a work in progress.

swankjesse: I think we start with the naked interface as it is here, and then when we actually change OkHttp to call the interface we can fill in details like _“this method is always followed by a later call to dnsEnd“_ and _“this method is invoked exactly once per call”_

adriancole: sgtm good point.. got a little weedy too fast :P

mjbenedict: Yeah lots of details left... so what do we think the next steps are?  Would you like me to try to extend this strawman PR to start showing where we would call into the EventListener?

swankjesse:
I think this has the right shape. I’m gonna merge this as-is, and we can take it from here.

Next steps:
 * API to OkHttpClient to specify an event listener or event listener factory
 * Drop the `Call` argument
 * Wire in a single operation – maybe DNS resolution? – and write a test case

mjbenedict:
Thank you @swankjesse !  Please let me know if there is anything I can do to help in this effort.

PR #3159, opened 2017-02-13
Introduce PublicSuffixDatabase: a list of public suffixes by publicsuffix.org.

dave-r12:
This is a follow-up to #3141. I found writing this code to be tricky and I'm convinced there are ways to simplify it. I haven't run any sort of benchmarks or done profiling at this point.

Given the code complexity (at least for me) and the added size overhead, I'm reluctant on this one. I think it does provide some value. The 2 cases I'm thinking of: confirming whether a cookie can be set for a domain, and also to bucket cookies in a cookie jar to enforce policy (i.e. number of cookies per domain, etc.)


swankjesse:
Apologies for sitting on this for so long. Hoping to take a look this week.

dave-r12:
No worries, thanks @swankjesse.

okhttp/src/main/java/okhttp3/internal/PublicSuffixDatabase.java
```diff
@@ -0,0 +1,346 @@
...
+        int permutations = 1 << numberOfWildcards;
+        for (int i = 1; i < permutations; i++) {
+          for (int labelPos = firstLevel - 1; labelPos >= currentLevel; labelPos--) {
+            // Use the bit positions to indicate whether we should place a wildcard there. For
+            // example with foo.bar.com: 01, 10, 11. The second level is position 1, third is
+            // position 2, etc.
+            int labelBitPos = 1 << firstLevel - 1 - labelPos;
+            int useWildcard = labelBitPos & i;
+
+            labelsWithWildcard[labelPos] = useWildcard > 0
```
swankjesse: do we ever assert 31 or fewer labels?

okhttp/src/main/java/okhttp3/internal/PublicSuffixDatabase.java
```diff
@@ -0,0 +1,346 @@
...
+
+    String[] exactRuleLabels = exactMatch != null
+        ? exactMatch.split("\\.")
+        : EMPTY_RULE;
+
+    String[] wildcardRuleLabels = wildcardMatch != null
+        ? wildcardMatch.split("\\.")
+        : EMPTY_RULE;
+
+    return exactRuleLabels.length > wildcardRuleLabels.length
```
swankjesse: can they have the same number of labels? I’m surprised that wildcards win ties

okhttp/src/main/java/okhttp3/internal/PublicSuffixDatabase.java
```diff
@@ -0,0 +1,346 @@
...
+      labelsWithWildcard[domainLabelsUtf8Bytes.length - 1] =
+          domainLabelsUtf8Bytes[domainLabelsUtf8Bytes.length - 1];
+
+      int firstLevel = domainLabelsUtf8Bytes.length - 1;
+      outer: for (int currentLevel = 0; currentLevel < firstLevel; currentLevel++) {
+        int numberOfWildcards = firstLevel - currentLevel;
+        int permutations = 1 << numberOfWildcards;
+        for (int i = 1; i < permutations; i++) {
+          for (int labelPos = firstLevel - 1; labelPos >= currentLevel; labelPos--) {
+            // Use the bit positions to indicate whether we should place a wildcard there. For
```
swankjesse: My reading of this is that we’re generating a sequence of bit strings that indicate how many labels to include on a domain like `ccc.ddd.eee.com`.

This loop searches in this order:

```
1: [*]
01: [ddd, *]
10: [*, eee]
11: [*, *]
001: [ccc, ddd, *]
010: [ccc, *, eee]
100: [*, ddd, eee]
011: [ccc, *, *]
101: [* ddd, *]
110: [*, *, eee]
111: [*, *, *]
```

This is interesting ’cause length is significant: `11` is different from `011`. It’s also interesting because it’s not just binary counting – `100` precedes `011`.

I wonder if there’s a nice way to replace the three nested loops with a generator function. That’d be independently testable, and it might reduce the number of moving parts in this method. The problem with the generator of course is that it needs to return both a count and a mask, though you could cheat by applying the mask to the input.

```
byte[][] labelsWithWildcard = domainLabelsUtf8Bytes.clone();
String wildcardMatch = null;
do {
  int level = nextPermutation(domainLabelsUtf8Bytes, labelsWithWildcard);
  if (level == -1) break;
  wildcardMatch = binarySearchBytes(publicSuffixListBytes, labelsWithWildcard, currentLevel);
} while (wildcardMatch == null);
```

```
private int nextPermutation(byte[][] previousPermutation, byte[][] domainLabels) {
   ?
}
```

swankjesse: As-is this is correct and reasonable, and I don’t think making this change is necessarily a net improvement. Just something I’m reacting to.

okhttp/src/main/java/okhttp3/internal/PublicSuffixDatabase.java
```diff
@@ -0,0 +1,346 @@
...
+      mid++;
+
+      // Now look for the ending '\n'.
+      int end = 1;
+      while (bytesToSearch[mid + end] != '\n') {
+        end++;
+      }
+      int publicSuffixLength = (mid + end) - mid;
+
+      // Compare the bytes. Note that the file stores UTF-8 encoded bytes, so we must compare the
```
swankjesse: Holy cow, nothing in this is easy

okhttp/src/main/java/okhttp3/internal/PublicSuffixDatabase.java
```diff
@@ -0,0 +1,346 @@
...
+    byte[] publicSuffixExceptionListBytes = null;
+
+    InputStream is = PublicSuffixDatabase.class.getClassLoader().getResourceAsStream(
+        PUBLIC_SUFFIX_RESOURCE);
+
+    if (is != null) {
+      Source source = new GzipSource(Okio.source(is));
+      BufferedSource bufferedSource = null;
+      try {
+        bufferedSource = Okio.buffer(source);
```
swankjesse: combine declaration and assignment?

okhttp/src/main/java/okhttp3/internal/PublicSuffixDatabase.java
```diff
@@ -0,0 +1,346 @@
...
+
+    synchronized (this) {
+      this.publicSuffixListBytes = publicSuffixListBytes;
+      this.publicSuffixExceptionListBytes = publicSuffixExceptionListBytes;
+    }
+
+    readCompleteLatch.countDown();
+  }
+
+  /** Visible for testing. */
```
swankjesse: cute

okhttp-tests/src/test/java/okhttp3/internal/PublicSuffixDatabaseTest.java
```diff
@@ -0,0 +1,271 @@
...
+        .writeUtf8("*.square.jp\n")
+        .writeUtf8("example.com\n")
+        .writeUtf8("square.com\n");
+    publicSuffixDatabase.setListBytes(buffer.readByteArray(), exception.readByteArray());
+
+    assertNull(publicSuffixDatabase.getEffectiveTldPlusOne("example.com"));
+    assertNull(publicSuffixDatabase.getEffectiveTldPlusOne("foo.square.jp"));
+  }
+
+  @Test public void allPublicSuffixes() throws IOException {
```
swankjesse: neat

okhttp/src/main/java/okhttp3/internal/PublicSuffixDatabase.java
swankjesse: ... reexamining this, nextPermutation would need the previous level as well.

okhttp/src/test/java/okhttp3/PublicSuffixListGenerator.java
```diff
@@ -0,0 +1,91 @@
...
+      + File.separator + "main" + File.separator + "resources";
+
+  private static final ByteString EXCEPTION_RULE_MARKER = ByteString.encodeUtf8("!");
+
+  public static void main(String... args) throws IOException {
+    OkHttpClient client = new OkHttpClient.Builder().build();
+    Request request = new Request.Builder()
+        .url("https://publicsuffix.org/list/public_suffix_list.dat")
+        .build();
+    Response response = client.newCall(request).execute();
```
dave-r12: Ahh right, can do.

okhttp/src/main/java/okhttp3/internal/PublicSuffixDatabase.java
dave-r12: > My reading of this is that we’re generating a sequence of bit strings that indicate how many labels to include on a domain like ccc.ddd.eee.com.

Basically, yes. The bit string is used to know where to slot in a `*`. A value of 1 means inject a `*` at that label position, and a value of 0 means use the original label.

Thanks for digging into this one. The code in the original PR made me uncomfortable. Re-reading now made me cringe.

I like the idea of pulling into a generator function. I'd like to try incorporating that.

okhttp/src/main/java/okhttp3/internal/PublicSuffixDatabase.java
dave-r12: Ohh, interesting point. The specification mentions that a wildcard is for an entire level, i.e. it shouldn't be possible for `cash.square.com` and `*.square.com` to be in the list. Maybe there is a better way of expressing that?

okhttp/src/main/java/okhttp3/internal/PublicSuffixDatabase.java
```diff
@@ -0,0 +1,346 @@
...
+
+  // The lists are held as a large array of UTF-8 bytes. This is to avoid allocating lots of strings
+  // that will likely never be used. Each rule is separated by '\n'. Please see the
+  // PublicSuffixListGenerator class for how these lists are generated.
+  // Guarded by this.
+  private byte[] publicSuffixListBytes;
+  private byte[] publicSuffixExceptionListBytes;
+
+  /**
+   * Returns the effective top-level domain plus one (eTLD+1) by referencing the public suffix list.
```
dave-r12: Added this - it's slightly awkward because null is overloaded, either it's a public suffix or the list failed to load.

okhttp/src/main/java/okhttp3/internal/PublicSuffixDatabase.java
```diff
@@ -0,0 +1,346 @@
...
+    // Preserve the passed in domain format (Unicode vs. punycode), but canonicalize it. This
+    // satisfies the publicsuffix.org test cases, but it's unclear if this should be done generally.
+    String domainToReturn = canonicalDomain.equalsIgnoreCase(domain)
+        ? canonicalDomain // We were passed a punycode domain.
+        : unicodeDomain; // We were passed a Unicode domain.
+
+    String[] domainLabels = unicodeDomain.split("\\.");
+    String[] ruleLabels = findMatchingRule(domainLabels);
+    if (ruleLabels == null) {
+      // If no rules match, the prevailing rule is '*'.
```
dave-r12: Makes sense to me, done.

okhttp/src/main/java/okhttp3/internal/PublicSuffixDatabase.java
```diff
@@ -0,0 +1,346 @@
...
+      ruleLabels = new String[]{" "};
+    }
+
+    if (domainLabels.length == ruleLabels.length && ruleLabels[0].charAt(0) != EXCEPTION_MARKER) {
+      // The domain is a public suffix.
+      return null;
+    }
+
+    int labelsToInclude;
```
dave-r12: That's right, I renamed it to `firstLabelOffset`.

okhttp/src/main/java/okhttp3/internal/PublicSuffixDatabase.java
```diff
@@ -0,0 +1,346 @@
...
+    }
+
+    // Wildcard rules throw a wrench into the mix. They are not restricted to appear in the leftmost
+    // position. An example is: *.*.foo. Therefore, we must exhaust the permutations,
+    // substituting a wildcard in the various label positions. For example, on the domain
+    // foo.bar.com, we'll try: [*, com], [foo, *, com], [*, bar, com], [*, *, com].
+    // Note: since we've already tried an exact match, we leave out those permutations.
+    String wildcardMatch = null;
+    if (domainLabelsUtf8Bytes.length > 1) {
+      // It seems highly unlikely (impossible?) the rightmost label will ever be a wildcard.
```
dave-r12: 👍 I've added some assertions to the generator.

okhttp/src/main/java/okhttp3/internal/PublicSuffixDatabase.java
dave-r12: I wrestled with this a bit more. I backed out of the complicated implementation. The specification as defined is very flexible. Wildcards are not restricted to the leftmost position. In practice though, they are all in the leftmost position. In effect, the code is testing permutations that do not exist and essentially doing unnecessary work.

The code cheats now, and only considers the leftmost position. I've incorporated some assertions in the generator to confirm this. It will mean if the very flexible specification is ever exercised, the implementation will need to be revisited.

Let me know if you think this is a bad decision.

okhttp/src/main/java/okhttp3/internal/PublicSuffixDatabase.java
dave-r12: Good catch. This shouldn't be possible with the revised implementation.

okhttp/src/main/java/okhttp3/internal/PublicSuffixDatabase.java
dave-r12: Done.

dave-r12:
Also worth noting, integrating this change will likely mean ongoing maintenance of the public suffix list. My hope is that it just involves running the generator class to update the list. I'm happy to take on that responsibility in the short term. I'm thinking a once-a-month recurring task.

okhttp/src/main/java/okhttp3/internal/PublicSuffixDatabase.java
swankjesse: I think that's a great simplification. If ever weird wildcards are added we can probably just be conservative and discard the prefix before the wildcard.

swankjesse:
Yeah, we can do it as a part of the release process. We’re already maintaining the cipher suite list. It adds value and encourages our users to keep up to date so I'm in favor.

okhttp-tests/src/test/java/okhttp3/internal/publicsuffix/PublicSuffixDatabaseTest.java
```diff
@@ -0,0 +1,255 @@
...
+
+      String test = "foobar." + exception;
+      assertEquals(exception, publicSuffixDatabase.getEffectiveTldPlusOne(test));
+    }
+  }
+
+  /**
+   * These tests are provided by <a href="https://publicsuffix.org/list/">publicsuffix.org</a>.
+   */
+  @Test public void publicSuffixDotOrgTestCases() {
```
swankjesse: I love this

okhttp/src/main/java/okhttp3/internal/publicsuffix/PublicSuffixDatabase.java
```diff
@@ -0,0 +1,325 @@
...
+    // TODO: Domains that start with '.' will be null on JDK8, but not on JDK7.
+    String canonicalDomain = Util.domainToAscii(domain);
+    if (canonicalDomain == null || canonicalDomain.startsWith(".")) return null;
+
+    // Convert it back to Unicode because the list uses Unicode, not punycode form.
+    String unicodeDomain = IDN.toUnicode(canonicalDomain);
+
+    // Preserve the passed in domain format (Unicode vs. punycode), but canonicalize it. This
+    // satisfies the publicsuffix.org test cases, but it's unclear if this should be done generally.
+    String domainToReturn = canonicalDomain.equalsIgnoreCase(domain)
```
swankjesse: Possibly for follow-up: should we require callers provide punycode only?

okhttp/src/main/java/okhttp3/internal/publicsuffix/PublicSuffixDatabase.java
```diff
@@ -0,0 +1,325 @@
...
+        ? canonicalDomain // We were passed a punycode domain.
+        : unicodeDomain; // We were passed a Unicode domain.
+
+    String[] domainLabels = unicodeDomain.split("\\.");
+    String[] rule = findMatchingRule(domainLabels);
+
+    if (rule == null
+        || (domainLabels.length == rule.length && rule[0].charAt(0) != EXCEPTION_MARKER)) {
+      // We failed to read the list or the domain is a public suffix.
```
swankjesse: I’m tempted to suggest just throwing if we ever fail to read the list. I can’t think of a good reason for that to happen in practice. (And if somebody is accidentally stripping resources from their okhttp.jar file, they probably don’t want to lose this behavior)

okhttp/src/main/java/okhttp3/internal/publicsuffix/PublicSuffixDatabase.java
```diff
@@ -0,0 +1,325 @@
...
+    String exactMatch = null;
+    for (int i = 0; i < domainLabelsUtf8Bytes.length; i++) {
+      String rule = binarySearchBytes(publicSuffixListBytes, domainLabelsUtf8Bytes, i);
+      if (rule != null) {
+        exactMatch = rule;
+        break;
+      }
+    }
+
+    // In theory, wildcard rules are not restricted to having the wildcard in the leftmost position.
```
swankjesse: love this explanation

okhttp/src/main/java/okhttp3/internal/publicsuffix/PublicSuffixDatabase.java
```diff
@@ -0,0 +1,325 @@
...
+    // In practice, wildcards are always in the leftmost position. For now, this implementation
+    // cheats and does not attempt every possible permutation. Instead, it only considers wildcards
+    // in the leftmost position. We assert this fact when we generate the public suffix file. If
+    // this assertion ever fails we'll need to refactor this implementation.
```
swankjesse: yep, and we’ll see it ’cause the new test cases will also fail. Shouldn’t ever happen, but if it does I think we’re still good.

okhttp/src/test/java/okhttp3/internal/publicsuffix/PublicSuffixListGenerator.java
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  /**
+   * These assertions ensure the {@link PublicSuffixDatabase} remains correct. The specification is
+   * very flexible regarding wildcard rules, but this flexibility is not something currently used
+   * in practice. To simplify the implementation, we've avoided implementing the flexible rules in
+   * favor of supporting what's actually used in practice. That means if these assertions ever fail,
+   * the implementation will need to be revisited to support a more flexible rule.
+   */
+  private static void assertWildcardRule(String rule) {
```
swankjesse: This build-time defensiveness is perfect.

swankjesse:
Another flaky test in web sockets.
https://github.com/square/okhttp/issues/3228

okhttp/src/main/java/okhttp3/internal/publicsuffix/PublicSuffixDatabase.java
dave-r12: 👍 - that will get rid of this madness. It'll mean tweaking the publicsuffix.org test cases slightly. I'll follow-up with that.

swankjesse: Cool. Or just put something in the test case to do punycode before the call?

okhttp/src/main/java/okhttp3/internal/publicsuffix/PublicSuffixDatabase.java
dave-r12: K, good point. I'd wanna find that out in a test environment and failing silently doesn't do that. Will fix.

PR #3161, opened 2017-02-14
Add error-prone compiler and fix its found issues.

JakeWharton:


okhttp-tests/src/test/java/okhttp3/internal/platform/OptionalMethodTest.java
```diff
@@ -309,9 +311,10 @@ public void invokeOptionalNonPublic() throws Exception {
       OptionalMethod<T> optionalMethod, T base, Object... args) throws Exception {
     try {
       optionalMethod.invoke(base, args);
-      fail();
     } catch (Error expected) {
+      return;
     }
+    fail();
```
swankjesse: yep

JakeWharton:
Looks like error-prone only runs on JDK 8. Are we okay dropping openjdk7 as a compilation target?

JakeWharton:
As a bonus, though, error-prone is based on JDK 9 which means it should support the new `-release` flag for targeting older platforms...

swankjesse:
I'm happy to bump our minimum desktop JVM to JDK 8. Does that impact IntelliJ plugins?

JakeWharton:
No. It only uses the JDK 8 to compile but still targets Java 7 bytecode as
an output. Animal sniffer prevents us from using Java 8 APIs as well.

PR #3183, opened 2017-02-22
optimized cancel all flow in dispatcher

guptasourabh:
In case of cancelAll flow we are running the for loop 3 times from start , rather than we should run it 1 time and cancel the calls.


JakeWharton:
You haven't changed the number of iterations since you still have to manage with the 3 individual iterators which are looping over the 3 lists. All the single loop does is prevent having to call `hasNext()` on the iterators which the for-each loop was implicitly doing but at the cost of the code being much more complicated.

swankjesse:
Yeah I don't think this is any improvement in efficiency here justifies the additional lines of code.

PR #3185, opened 2017-02-22
code optimization

guptasourabh:
Changes done are -:
Duplicate code have been moved to Single place.

### DIFF
--- a/okhttp/src/main/java/okhttp3/CacheControl.java
+++ b/okhttp/src/main/java/okhttp3/CacheControl.java
@@ -1,6 +1,7 @@
 package okhttp3;
 
 import java.util.concurrent.TimeUnit;
+
 import okhttp3.internal.http.HttpHeaders;
 
 /**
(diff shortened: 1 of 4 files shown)


okhttp/src/main/java/okhttp3/TlsVersion.java
```diff
@@ -34,19 +34,7 @@
...
-        return TLS_1_2;
-      case "TLSv1.1":
-        return TLS_1_1;
-      case "TLSv1":
-        return TLS_1_0;
-      case "SSLv3":
-        return SSL_3_0;
-    }
-    throw new IllegalArgumentException("Unexpected TLS version: " + javaName);
+    return TlsVersion.valueOf(javaName);
```
JakeWharton: This is not the same.

okhttp/src/main/java/okhttp3/CacheControl.java
```diff
@@ -350,5 +337,11 @@ public Builder noTransform() {
     public CacheControl build() {
       return new CacheControl(this);
     }
+
+    private Builder setValueAndReturnBuilder(int valueToUpdate, int valueToConvert, TimeUnit timeUnit) {
+      long valueInSeconds = timeUnit.toSeconds(valueToConvert);
+      valueToUpdate = valueInSeconds > Integer.MAX_VALUE ? Integer.MAX_VALUE : (int) valueInSeconds;
```
JakeWharton: Java passes arguments by value, not reference, so this statement doesn't actually do anything.

PR #3186, opened 2017-02-22
duplicate code moved to single place

guptasourabh:
changes -:
Duplicate code moved to single place.
Removed redundant array null check in for loop and initialization of array size.

### DIFF
--- a/okhttp/src/main/java/okhttp3/CacheControl.java
+++ b/okhttp/src/main/java/okhttp3/CacheControl.java
@@ -1,6 +1,7 @@
 package okhttp3;
 
 import java.util.concurrent.TimeUnit;
+
 import okhttp3.internal.http.HttpHeaders;
 
 /**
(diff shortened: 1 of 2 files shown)


okhttp/src/main/java/okhttp3/CacheControl.java
```diff
@@ -350,5 +342,11 @@ public Builder noTransform() {
     public CacheControl build() {
       return new CacheControl(this);
     }
+
+    private int getMaxValueInSeconds(int valueToConvert, TimeUnit timeUnit) {
```
NightlyNexus: static.

okhttp/src/main/java/okhttp3/CacheControl.java
```diff
@@ -350,5 +342,11 @@ public Builder noTransform() {
...
+      long valueInSeconds = timeUnit.toSeconds(valueToConvert);
+      return valueInSeconds > Integer.MAX_VALUE
+              ? Integer.MAX_VALUE : (int) valueInSeconds;
```
NightlyNexus: `return Math.max(Integer.MAX_VALUE, (int) valueInSeconds);` if touching this code anyway?

JakeWharton: That behavior is not the same.
```
Math.min(Integer.MAX_VALUE, (int) (1L + Integer.MAX_VALUE))
===> -2147483648
```

NightlyNexus: oh, haha, right, yeah. 🐑

swankjesse:
Could you complete Square’s contributor agreement?

guptasourabh:
@NightlyNexus changes done
@swankjesse completed square contributor agreement.

swankjesse:
I know that small contributions like these are well-intentioned, but we’re more interested in more dramatic improvements: bug fixes, measurable optimizations, and documentation improvements.

As is it seems like the motivation of this contribution is to make a contribution, and that distracts us from other improvements.

Issue #3171, opened 2017-02-20
java.lang.IllegalArgumentException: Unexpected code point: d83d


iglaweb:
The app crashes while calling  'execute' method of Call.

```
java.lang.IllegalArgumentException: Unexpected code point: d83d
	at okio.Buffer.a(SourceFile:925)
	at okhttp3.HttpUrl.a(SourceFile:1851)
	at okhttp3.HttpUrl.a(SourceFile:1820)
	at okhttp3.HttpUrl.a(SourceFile:1867)
	at okhttp3.HttpUrl$Builder.a(SourceFile:1133)
	at retrofit2.RequestBuilder.b(SourceFile:159)
	at retrofit2.ParameterHandler$Query.a(SourceFile:109)
	at retrofit2.ServiceMethod.a(SourceFile:109)
	at retrofit2.OkHttpCall.g(SourceFile:178)
	at retrofit2.OkHttpCall.a(SourceFile:162)
	at retrofit2.ExecutorCallAdapterFactory$ExecutorCallbackCall.a(SourceFile:89)

```
I'am using okhttp 3.5.0 and okio 1.11.0. It's a bug or incorrect use of api? Better to handle it by 'catch'?

Most problem devices are HTC ONE, HTC Desire 626G, HTC Desire 820G, but there are another manufactures.

swankjesse:
Yikes. That's a nasty bug. Can fix.

PR #3189, opened 2017-02-24
Ignore illegal chars to avoid IllegalArgumentException

iglaweb:
This should fix the issue expressed here: [#3171](https://github.com/square/okhttp/issues/3171)


swankjesse:
Thanks for sending this. Hoping to look soon.

swankjesse:
Despite my silence I’ve been thinking about what the heck we should do here. I’m starting to think our best way forward is to change Okio to not throw on illegal characters; instead it should emit the replacement character.

I’m going to do some research to compare this strategy against the JDK and Guava. If it’s not dissimilar I’ll make a PR that does that in Okio and close this PR.

swankjesse:
Fixed upstream!
https://github.com/square/okio/pull/297

PR #3193, opened 2017-02-28
Add multiple query parameters

DeychAI:
New methods for adding multiple query parameters from map in HttpUrl.Builder


swankjesse:
Thanks for this contribution.

I’m reluctant to include it because it pulls OkHttp away from its goal of staying small. Doing this one thing is small and easy, but it motivates more changes, such as exposing query parameters as a `Map`. That’s potentially awkward because query parameters are non-unique.

I’m declining this for now. Not because this code is bad, but because it sets a precedent for seductive APIs that I want to avoid.

PR #3207, opened 2017-03-06
Connection coalescing

swankjesse:
As described here:

https://hpbn.co/optimizing-application-delivery/#eliminate-domain-sharding
https://tools.ietf.org/html/rfc7540#section-9.1.1

When the following is true
- HTTP/2 so we don't overload non multiplexed connections.
- HTTPS since we need subjectAltNames (n.b. RFC 7540 defines as safe to use over plaintext, based purely on DNS).
- Desired url host is part of the subjectAltNames for an existing connection.
- One of the DNS results for desired host match the established connection.
- HostnameVerifier is *not* overridden.
- ConnectionPinner accepts the existing certificate chain.

From https://tools.ietf.org/html/rfc7540#section-9.1.1

> A server that does not wish clients to reuse connections can indicate
> that it is not authoritative for a request by sending a 421
> (Misdirected Request) status code in response to the request (see
> Section 9.1.2).

### DIFF
--- a/okhttp/src/main/java/okhttp3/Address.java
+++ b/okhttp/src/main/java/okhttp3/Address.java
@@ -150,20 +150,9 @@ public final class Address {
   }
 
   @Override public boolean equals(Object other) {
-    if (other instanceof Address) {
-      Address that = (Address) other;
-      return this.url.equals(that.url)
-          && this.dns.equals(that.dns)
-          && this.proxyAuthenticator.equals(that.proxyAuthenticator)
-          && this.protocols.equals(that.protocols)
-          && this.connectionSpecs.equals(that.connectionSpecs)
-          && this.proxySelector.equals(that.proxySelector)
-          && equal(this.proxy, that.proxy)
-          && equal(this.sslSocketFactory, that.sslSocketFactory)
-          && equal(this.hostnameVerifier, that.hostnameVerifier)
-          && equal(this.certificatePinner, that.certificatePinner);
-    }
-    return false;
+    return other instanceof Address
+        && url.equals(((Address) other).url)
+        && equalsNonHost((Address) other);
   }
 
   @Override public int hashCode() {
@@ -179,6 +168,19 @@ public final class Address {
     result = 31 * result + (hostnameVerifier != null ? hostnameVerifier.hashCode() : 0);
     result = 31 * result + (certificatePinner != null ? certificatePinner.hashCode() : 0);
     return result;
+  }
+
+  boolean equalsNonHost(Address that) {
+    return this.dns.equals(that.dns)
+        && this.proxyAuthenticator.equals(that.proxyAuthenticator)
+        && this.protocols.equals(that.protocols)
+        && this.connectionSpecs.equals(that.connectionSpecs)
+        && this.proxySelector.equals(that.proxySelector)
+        && equal(this.proxy, that.proxy)
+        && equal(this.sslSocketFactory, that.sslSocketFactory)
+        && equal(this.hostnameVerifier, that.hostnameVerifier)
+        && equal(this.certificatePinner, that.certificatePinner)
+        && this.url().port() == that.url().port();
   }
 
   @Override public String toString() {
--- a/okhttp/src/main/java/okhttp3/internal/connection/RealConnection.java
+++ b/okhttp/src/main/java/okhttp3/internal/connection/RealConnection.java
@@ -44,6 +44,7 @@ import okhttp3.Protocol;
 import okhttp3.Request;
 import okhttp3.Response;
 import okhttp3.Route;
+import okhttp3.internal.Internal;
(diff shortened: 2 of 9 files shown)


swankjesse:
It's unlikely we'll release this in the next 4 weeks. But if there's a risk this change is broken we shouldn’t merge it.

I think SAN and IP address checks are sufficient for correctness. In a followup we should support more flexible IP checks, otherwise I don’t think this will get much coverage.

yschimke:
Yep, I haven't found anything in my real world testing, and I think the tests are fairly comprehensive, but I'm happy to add more tests to cover specific attack cases if we identify any others.  I'll try to test specifically with a server that responds with 421 responses if I can find a real world example.

But I'm happy that it follows the spec and is an implementation of something actively used by firefox and chrome already.

The other risk is badly configured servers e.g. an API server re-uses certificates but because of DNS ordering wasn't getting traffic.

okhttp/src/main/java/okhttp3/internal/connection/RealConnection.java
```diff
@@ -373,11 +374,64 @@ private Request createTunnelRequest() {
...
+    // 1. This connection must be HTTP/2.
+    if (http2Connection == null) return false;
+
+    // 2. The routes must share an IP address. This requires us to have a DNS address for both
+    // hosts, which only happens after route planning. We can't coalesce connections that use a
+    // proxy, since proxies don't tell us the origin server's IP address.
+    if (route == null) return false;
+    if (route.proxy().type() != Proxy.Type.DIRECT) return false;
+    if (this.route.proxy().type() != Proxy.Type.DIRECT) return false;
+    if (!this.route.socketAddress().equals(route.socketAddress())) return false;
```
yschimke: This appears to be even less aggressive at coalescing than either Firefox or Chrome.

see "The Firefox way", "The Chrome way"

Might be worth a comment

okhttp/src/main/java/okhttp3/Address.java
```diff
@@ -181,6 +170,19 @@ public CertificatePinner certificatePinner() {
...
+    return this.dns.equals(that.dns)
+        && this.proxyAuthenticator.equals(that.proxyAuthenticator)
+        && this.protocols.equals(that.protocols)
+        && this.connectionSpecs.equals(that.connectionSpecs)
+        && this.proxySelector.equals(that.proxySelector)
+        && equal(this.proxy, that.proxy)
+        && equal(this.sslSocketFactory, that.sslSocketFactory)
+        && equal(this.hostnameVerifier, that.hostnameVerifier)
+        && equal(this.certificatePinner, that.certificatePinner)
+        && this.url().port() == that.url().port();
```
yschimke: I don't think it matters, but should this also check url().scheme() as well?

This is built as 

https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/Address.java#L54-L58

yschimke:
By waiting for the Route, this seems much cleaner.  It seems like a great way to introduce this and see how effective it is and I will separately try to evaluate how likely real world sites are to gain from this e.g. sites that return IPv4/IPv6 for the main site, but IPv4 only for some image serving domains, or just a different but partially overlapping set of IPs etc.

okhttp/src/main/java/okhttp3/Address.java
swankjesse: The SSL socket factory should cover that.

okhttp/src/main/java/okhttp3/internal/connection/RealConnection.java
swankjesse: Actually I wanna do a follow-up so we're closer to the Chrome way. Otherwise I don't think we’ll intersect consistently.

swankjesse: ... possibly by extracting “alternative IPs” from the RoutePlanner and keeping those on the RealConnection. If you wanna do analysis on what the hit rate is, we should land that first. Wanna send a PR once this is merged?

swankjesse:
@yschimke is this good to merge?

yschimke:
A bit of a follow up on this, results are worse than I expected, probably in line with what you expected.  The change looks good however, and adopting the Chrome mode looks like a clean change if Route carries the alternatives ipaddresses to match against.

I was testing different modes on a branch https://github.com/square/okhttp/compare/master...yschimke:coalesce_test?expand=1

Noticeably the results change depending on where you are and whether you end up using Google Dns servers etc.  So my 5->3 for facebook became 5->4 from a different location (SF instead of London).

swankjesse:
Yep, I saw the same thing. Tracking here. (And don’t be discouraged by the _Icebox_ label; I just didn’t want to put much urgency on this.)
https://github.com/square/okhttp/issues/3226

PR #3208, opened 2017-03-07
HTTP/2 reject DATA frame on streamId is zero

chhsiao90:
This is my first PR to this repository!

When I had run some tests with tool that will check about http2 spec violation,
I found that when okhttp client didn't handle well when received DATA frame on streamId=0

##### Scenario
sending DATA(streamId=0) frame to okhttp client with a custom mock server

##### Expected
okhttp client should treat it as connection error with type PROTOCOL_ERROR

##### Actual
okhttp send RST_STREAM on streamId=0

##### Reference
[rfc7540#section-6.1](https://tools.ietf.org/html/rfc7540#section-6.1)
> DATA frames MUST be associated with a stream.  If a DATA frame is
   received whose stream identifier field is 0x0, the recipient MUST
   respond with a connection error (Section 5.4.1) of type
   PROTOCOL_ERROR.

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/internal/http2/Http2Test.java
+++ b/okhttp-tests/src/test/java/okhttp3/internal/http2/Http2Test.java
@@ -392,6 +392,23 @@ public final class Http2Test {
     });
   }
 
+  @Test public void dataFrameNotAssociateWithStream() throws IOException {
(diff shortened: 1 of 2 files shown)


swankjesse:
Thanks!

PR #3210, opened 2017-03-07
MockWebServer: Expose request URL as HttpUrl

aahlenst:
It might be helpful for example when dealing with lots of query parameters in tests because all the URL parsing facilities from `HttpUrl` can be reused.  I'm not sure that the suggested change is the best possible solution and I'm open for suggestions.


swankjesse:
Thanks! We’ve still got your CLA on file from 2013!

PR #3230, opened 2017-03-19
Ignore minified files when searching code

grimreaper:


JakeWharton:
What uses this file?

grimreaper:
It's used by ag, ack, and a few other grep like tools. Usually you don't need to search minified files. I guess this might be considered global configuration.

swankjesse:
Doesn't seem relevant to OkHttp.

grimreaper:
closed;
FWIW I was using this to ignore
website/static/bootstrap.min.js
website/static/bootstrap-combined.min.css
but I just added them to my personal global config

PR #3241, opened 2017-03-25
Change PublicSuffixDatabase to require canonical, punycode encoded domains.

dave-r12:


swankjesse:
Yep, definitely a nice improvement.

PR #3242, opened 2017-03-25
Restrict cookies with a domain that is a public suffix.

dave-r12:
https://github.com/square/okhttp/pull/3141

I ran some preliminary tests by creating a simple cookie jar, and executing the Crawler (which is fun.) I didn't see any major issues while doing this.

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/CookieTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/CookieTest.java
@@ -27,6 +27,7 @@ import org.junit.Test;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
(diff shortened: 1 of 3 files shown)


okhttp-tests/src/test/java/okhttp3/CookieTest.java
```diff
@@ -266,6 +267,26 @@
...
 
+  /**
+   * These public suffixes were selected by inspecting the publicsuffix.org list. It's possible they
+   * may change in the future. If this test begins to fail, please double check they are still
+   * present in the public suffix list.
+   */
+  @Test public void domainIsPublicSuffix() {
+    HttpUrl ascii = HttpUrl.parse("https://foo1.foo.bar.elb.amazonaws.com");
+    assertNotNull(Cookie.parse(ascii, "a=b; domain=foo.bar.elb.amazonaws.com"));
+    assertNull(Cookie.parse(ascii, "a=b; domain=bar.elb.amazonaws.com"));
```
swankjesse: this is really great

PR #3253, opened 2017-03-29
API to OkHttpClient to specify an event listener or event listener factory

mjbenedict:
Related to the discussion in #270, this stakes out an API  that could be used to add
Zipkin or other traces to OkHttp requests.

Step one from https://github.com/square/okhttp/pull/3149#issuecomment-277298784

### DIFF
--- a/okhttp/src/main/java/okhttp3/OkHttpClient.java
+++ b/okhttp/src/main/java/okhttp3/OkHttpClient.java
@@ -216,6 +216,7 @@ public class OkHttpClient implements Cloneable, Call.Factory, WebSocket.Factory 
   final int readTimeout;
   final int writeTimeout;
   final int pingInterval;
+  final EventListener eventListener;
   final EventListener.Factory eventListenerFactory;
 
   public OkHttpClient() {
@@ -234,6 +235,7 @@ public class OkHttpClient implements Cloneable, Call.Factory, WebSocket.Factory 
     this.cache = builder.cache;
     this.internalCache = builder.internalCache;
     this.socketFactory = builder.socketFactory;
+    this.eventListener = builder.eventListener;
     this.eventListenerFactory = builder.eventListenerFactory;
 
     boolean isTLS = false;
--- a/okhttp/src/main/java/okhttp3/OkHttpClient.java
+++ b/okhttp/src/main/java/okhttp3/OkHttpClient.java
@@ -216,7 +216,6 @@ public class OkHttpClient implements Cloneable, Call.Factory, WebSocket.Factory 
   final int readTimeout;
   final int writeTimeout;
   final int pingInterval;
-  final EventListener eventListener;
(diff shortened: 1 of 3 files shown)


adriancole:
Thanks for starting. I see only a couple files affected, so will comment on
what I see :)

Seems addX syntax would be used when more than one can be present. Current
impl shows one. Are you planning to provide for multiple event listeners or
is that just a naming glitch?

mjbenedict:
Thanks!  I intend to try to create PRs for what I see as the minimum viable thing and let you guys tell me what needs to be added/changed/fixed.  I am happy to change these to arrays and support multiple event listeners, that seems totally reasonable.

adriancole:
oh.. it wasn't a request to support multiple :) more a naming question
whether that was planned.

mjbenedict:
Yeah I can change this to `setEventListener` and `setEventListenerFactory` 👌

okhttp/src/main/java/okhttp3/RealCall.java
```diff
@@ -42,10 +43,17 @@
   private boolean executed;
 
   RealCall(OkHttpClient client, Request originalRequest, boolean forWebSocket) {
+    final EventListener.Factory eventListenerFactory = client.eventListenerFactory();
+
     this.client = client;
     this.originalRequest = originalRequest;
     this.forWebSocket = forWebSocket;
     this.retryAndFollowUpInterceptor = new RetryAndFollowUpInterceptor(client, forWebSocket);
+    if (null == eventListenerFactory) {
```
swankjesse: nit: we don’t typically do yoda expressions in OkHttp

okhttp/src/main/java/okhttp3/OkHttpClient.java
```diff
@@ -216,6 +216,8 @@ public void apply(ConnectionSpec tlsConfiguration, SSLSocket sslSocket, boolean
   final int readTimeout;
   final int writeTimeout;
   final int pingInterval;
+  final EventListener eventListener;
```
swankjesse: I’m unsure about using fields for both of these. Maybe we should have a ConstantFactory or something for the single value case, and when you set the listener that’s what you get?

okhttp/src/main/java/okhttp3/OkHttpClient.java
```diff
@@ -878,6 +896,16 @@ public Builder addNetworkInterceptor(Interceptor interceptor) {
       return this;
     }
 
+    public Builder addEventListener(EventListener listener) {
```
swankjesse: as-is this definitely needs to clear the factory

swankjesse: or set the factory to a simple implementation that always returns `listener`

swankjesse:
Awesome. Merged manually!

PR #3257, opened 2017-03-31
Update README.md

hermanliang:
update ProGuard rule

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ ProGuard
 
 If you are using ProGuard you might need to add the following option:
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -50,7 +50,7 @@ ProGuard
 
 If you are using ProGuard you might need to add the following option:
 ```
--dontwarn okhttp3.**
+-dontwarn okio.**
```
swankjesse: Do you need both? I don't proguard.

vanniktech: Since okhttp pulls in okio you need both.

PR #3262, opened 2017-04-01
Conscrypt Support

yschimke:
Not necessarily for landing, testing what would be required.  At minimum waiting for a final 1.0.0 release would be prudent.  5 test bugs still to resolve, not yet clear if they are in okhttp main, okhttp test assumptions or conscrypt.

Any feedback welcome.

Tested with JDK8 without alpn-boot in the bootclasspath

```
16:46 $ ./oksocial -i --debug https://graph.facebook.com/robots.txt
/Library/Java/JavaVirtualMachines/jdk1.8.0_111.jdk/Contents/Home/bin/java
16:46:36.433	Platform: ConscryptPlatform
...
protocol h2
tls TLS_1_2
cipher TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256
HTTP/1.1 200
```

I'm getting an error when shutting down the client, but I suspect that could be a conscrypt issue.


yschimke:
Still has 12 tests failing, so just a proof of concept

okhttp/src/main/java/okhttp3/internal/platform/ConscryptPlatform.java
```diff
@@ -0,0 +1,65 @@
...
+  @Override public Provider getProvider() {
+    return new OpenSSLProvider();
+  }
+
+  @Override public X509TrustManager trustManager(SSLSocketFactory sslSocketFactory) {
+    try {
+      Class<?> sslContextClass = OpenSSLSocketFactoryImpl.class;
+      SSLParametersImpl sp =
+          readFieldOrNull(sslSocketFactory, SSLParametersImpl.class, "sslParameters");
+      if (sp == null) return null;
```
duchuyctlk:
```java
return sp == null ? null : sp.getDefaultX509TrustManager();
```

This could be easier for reading 😄

swankjesse:
Cool! What should we do with this?

yschimke:
If i clean it up get remaining test passing. Is it something you would land? With conscrypt jar marked as provided maybe.

yschimke:
Also it feels like the first overlapping Platform. Is this something that should be opt in. Eg. 

clientBuilder.platform(new ConscryptPlatform())

yschimke:
Remaining failing tests:

  CallTest.cancelDuringHttpsConnect:1849->cancelDuringConnect:1868 » NullPointer
  HttpOverHttp2Test.emptyDataFrameSentWithEmptyBody:859 » IO Expected a connecti...
  HttpOverHttp2Test.noDataFramesSentWithNullRequestBody:827 » IO Expected a conn...
  ClientAuthTest.invalidClientAuthFails:186 » StreamReset stream was reset: PROT...
  ClientAuthTest.missingClientAuthFailsForNeeds:163 » ConnectionShutdown

swankjesse:
Nice to see this is all working. I’m still not exactly sure when to recommend this over the VM’s built-in TLS stack.

yschimke:
The client auth tests are failing similarly to the problem with Java 9, a different more generic exception than SSLHandshakeException.  

After this there are 3 failures, 1 of which I believe is a Conscrypt bug.  The other two I haven't worked out yet, they fail as it seems to fallback to HTTP/1.1 so preface is missing.

The build above is passing because it doesn't activate the conscrypt profile.

yschimke:
@swankjesse The biggest question for you I had was how to support an optional Platform?

a) based on presence of the conscrypt jar?
b) based on a system property
c) configured by code, e.g. "ConscryptPlatform.install()" or "Platform.override(new ConscryptPlatform)"

yschimke:
Remaining failing tests:

CallTest.cancelDuringHttpsConnect:1849->cancelDuringConnect:1868 » NullPointer
HttpOverHttp2Test.emptyDataFrameSentWithEmptyBody:859 » IO Expected a connecti...
HttpOverHttp2Test.noDataFramesSentWithNullRequestBody:827 » IO Expected a conn...
ClientAuthTest.invalidClientAuthFails:186 » StreamReset stream was reset: PROT...

swankjesse:
I’m unsure. I think it depends on what motivates Conscrypt users.

I think it should probably not be completely automatic, only because there are severe security and connectivity consequences if we get it wrong. That might be accidentally not using Conscrypt when we should, or using it when we shouldn’t.

yschimke:
Conscrypt is very new, pre-1.0.  But it seems like there is momentum, e.g. likely Netty adoption and also generally for HTTP/2 without bootclassloader hacks.  

What about if we keep the system property for now as it avoids new API and doesn't assume a transitive dependency on Conscrypt means you want it automagically enabled?

swankjesse:
Sounds good!

yschimke:
Remaining failing tests:

CallTest.cancelDuringHttpsConnect:1849->cancelDuringConnect:1868 » NullPointer

This has an open report with conscrypt as referenced above.  The HttpOverHttp2Test tests were failing because of SocketRecorder specifically supporting alpn-boot but not conscrypt.

yschimke:
I'm testing a PR on the conscrypt side, unfortunately it shifts the NPE to a 10 second timeout on connect cancellation.  So still blocked on this test at the moment.  Hopefully I can work with them to get a proper fix for Conscrypt 1.0 or 1.1.

yschimke:
The workaround I found, is to retain the SslSocket earlier and close it also.  I also needed to workaround SslSocket.getSession() returning SSL_NULL_WITH_NULL_NULL for failures.

But given I'm removing a warning comment, I'm not sure how safe this is for other Platforms...

yschimke:
After that it should be generally working, enough to beta test.  But will need a .0 release with https://github.com/google/conscrypt/pull/178

yschimke:
As per https://github.com/google/conscrypt/issues/193#issuecomment-303840234

Switching to OpenSSLEngineSocketImpl

But now facing a failing test I am debugging, CallTest.recoverFromTlsHandshakeFailure_tlsFallbackScsvEnabled

yschimke:
Latest master is just failing in ConnectionSpecSelectorTest

https://github.com/google/conscrypt/issues/198

The Engine socket doesn't support disconnected sockets, which seems like a pretty big API breakage https://github.com/google/conscrypt/issues/198

`sslContext.getSocketFactory().createSocket();`

yschimke:
That last issue is fixed upstream.  All tests passing for me, with a locally built master.  I think just waiting on a 1.0 release now before possible to consider landing this.

okhttp-tests/src/test/java/okhttp3/internal/http2/HttpOverHttp2Test.java
```diff
@@ -75,18 +76,21 @@
...
+  private static final Logger http2Logger = Logger.getLogger(Http2.class.getName());
+
   @Rule public final TemporaryFolder tempDir = new TemporaryFolder();
   @Rule public final MockWebServer server = new MockWebServer();
 
   private SslClient sslClient = SslClient.localhost();
   private HostnameVerifier hostnameVerifier = new RecordingHostnameVerifier();
   private OkHttpClient client;
   private Cache cache;
+  private TestLogHandler http2Handler = new TestLogHandler();
```
yschimke: I can revert this change, but I think I prefer it.

okhttp/pom.xml
```diff
@@ -18,6 +18,12 @@
       <artifactId>okio</artifactId>
     </dependency>
     <dependency>
+      <groupId>org.conscrypt</groupId>
+      <artifactId>conscrypt-openjdk-uber</artifactId>
+      <version>1.0.0.RC7</version>
```
yschimke: Need to update this, but can't before next RC

swankjesse:
Cool. Will take a look when I have time.

yschimke:
This is working and usable with conscrypt 1.0.0.RC8.  So could land when 1.0.0 is released.

swankjesse:
Awesome. I think supporting Conscrypt from their 1.0 is a great strategy.

yschimke:
You can test with 

```
$ brew install yschimke/tap/oksocial

$ oksocial -i --debug https://graph.facebook.com/robots.txt
08:14:45.781	url https://graph.facebook.com/robots.txt
08:14:45.784	Request Request{method=GET, url=https://graph.facebook.com/robots.txt, tag=null}
08:14:46.013	Dns (graph.facebook.com): graph.facebook.com/31.13.76.100, graph.facebook.com/2a03:2880:f001:4:face:b00c:0:4f5a
08:14:46.711	>> CONNECTION 505249202a20485454502f322e300d0a0d0a534d0d0a0d0a
08:14:46.712	>> 0x00000000     6 SETTINGS
08:14:46.712	>> 0x00000000     4 WINDOW_UPDATE
08:14:46.713	<< 0x00000000    30 SETTINGS
08:14:46.714	>> 0x00000000     0 SETTINGS      ACK
08:14:46.714	<< 0x00000000     4 WINDOW_UPDATE
08:14:46.741	>> 0x00000003   303 HEADERS       END_STREAM|END_HEADERS
08:14:47.164	<< 0x00000000     0 SETTINGS      ACK
08:14:47.166	<< 0x00000003     4 WINDOW_UPDATE
08:14:47.311	<< 0x00000003   172 HEADERS       END_HEADERS
08:14:47.312	<< 0x00000003    26 DATA          END_STREAM
08:14:47.325	OkHttp Platform: ConscryptPlatform
08:14:47.326	TLS Version: TLS_1_2
08:14:47.326	Protocol: h2
08:14:47.326	Cipher: TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256
08:14:47.326	Peer Principal: CN=*.facebook.com, O="Facebook, Inc.", L=Menlo Park, ST=California, C=US
HTTP/1.1 200
expires: Mon, 17 Jul 2017 07:14:47 GMT
cache-control: max-age=2592000
accept-ranges: bytes
content-type: text/plain; charset=
content-length: 26
date: Sat, 17 Jun 2017 07:14:47 GMT

User-agent: *
Disallow: /

08:14:47.330	>> 0x00000000     8 GOAWAY 
```

yschimke:
We may want to implement ConscryptPlatform.trustManager() and fetch via OpenSSLSocketFactoryImpl -> SSLParametersImpl

But I'm inclined not to as it is deprecated anyway

yschimke:
Bumped to RC9, and using Conscrypt.isAvailable()

dansanduleac:
Any blockers? This PR already works with the current version of Conscrypt.

JakeWharton:
We can't depend on pre-release software.

yschimke:
RC10 - main thing I'm trying to understand is how to implement a clean CertificateChainCleaner on the Conscrypt API.

yschimke:
I think we are stuck with the existing JVM building, so I've raised https://github.com/google/conscrypt/issues/336 to hopefully get a clean accessor for the trust manager

yschimke:
n.b. RC13 has been cut but not yet pushed to maven central, I'll test with that but its likely that 1.0 is imminent.

swankjesse:
Nice.

.travis.yml
```diff
@@ -23,6 +23,8 @@ branches:
 notifications:
   email: false
 
+script: "mvn clean test -P conscrypt"
```
yschimke: To remove o r convert to matrix project

yschimke:
1.0.0 released, I'll double check everything then hopefully land.

Are we ok with the gating to enable this?

swankjesse:
I’d like to give it a read through, but yes, I’m enthusiastic about this.

okhttp/src/main/java/okhttp3/internal/platform/Platform.java
```diff
@@ -183,6 +183,15 @@ public CertificateChainCleaner buildCertificateChainCleaner(SSLSocketFactory ssl
 
   /** Attempt to match the host runtime to a capable Platform implementation. */
   private static Platform findPlatform() {
+    if ("conscrypt".equals(System.getProperty("okhttp.platform"))) {
```
yschimke: It might be better to just check if Conscrypt is the first installed Provider?

okhttp-tests/src/test/java/okhttp3/ConscryptTest.java
```diff
@@ -0,0 +1,65 @@
+package okhttp3;
```
swankjesse: nit: copyright header?

okhttp-tests/src/test/java/okhttp3/ConscryptTest.java
```diff
@@ -0,0 +1,65 @@
...
+      CipherSuite.TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,
+      CipherSuite.TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256,
+      CipherSuite.TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256
+  };
+
+  private OkHttpClient client = buildClient();
+
+  private OkHttpClient buildClient() {
+    ConnectionSpec spec = new ConnectionSpec.Builder(true)
+        .cipherSuites(MANDATORY_CIPHER_SUITES) // Check we are using strong ciphers
```
swankjesse: yass

okhttp-tests/src/test/java/okhttp3/ConscryptTest.java
```diff
@@ -0,0 +1,65 @@
...
+    return new OkHttpClient.Builder().connectionSpecs(Arrays.asList(spec)).build();
+  }
+
+  @Before
+  public void setup() {
+    Assume.assumeTrue(getPlatform().equals("conscrypt"));
+  }
+
+  @Test
+  public void testMozilla() throws IOException {
```
swankjesse: I think these might not work in CI, or in CI environments that don’t have internet access.

okhttp/pom.xml
```diff
@@ -17,6 +17,12 @@
       <groupId>com.squareup.okio</groupId>
       <artifactId>okio</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.conscrypt</groupId>
+      <artifactId>conscrypt-openjdk-uber</artifactId>
+      <version>1.0.0</version>
+      <scope>provided</scope>
```
swankjesse: nice

okhttp/src/main/java/okhttp3/internal/connection/RealConnection.java
```diff
@@ -288,6 +288,7 @@ private void connectTls(ConnectionSpecSelector connectionSpecSelector) throws IO
       // Create the wrapper over the connected socket.
       sslSocket = (SSLSocket) sslSocketFactory.createSocket(
           rawSocket, address.url().host(), address.url().port(), true /* autoClose */);
+      socket = sslSocket;
```
swankjesse: motivation? if we keep this you’ll wanna remove the corresponding assignment on line 325.

as-is I think I like the previous behavior better. We don’t promote the SSL socket until we’ve handshaked. But that might leak now?

okhttp/src/main/java/okhttp3/internal/connection/RealConnection.java
```diff
@@ -306,7 +307,7 @@ private void connectTls(ConnectionSpecSelector connectionSpecSelector) throws IO
       Handshake unverifiedHandshake = Handshake.get(sslSocketSession);
 
       // Verify that the socket's certificates are acceptable for the target host.
-      if (!address.hostnameVerifier().verify(address.url().host(), sslSocket.getSession())) {
+      if (!address.hostnameVerifier().verify(address.url().host(), sslSocketSession)) {
```
swankjesse: nice

okhttp/src/main/java/okhttp3/internal/platform/ConscryptPlatform.java
```diff
@@ -0,0 +1,99 @@
+package okhttp3.internal.platform;
```
swankjesse: nit: copyright

okhttp/src/main/java/okhttp3/internal/platform/ConscryptPlatform.java
```diff
@@ -0,0 +1,99 @@
...
+public class ConscryptPlatform extends Platform {
+  private ConscryptPlatform() {
+  }
+
+  public Provider getProvider() {
+    return new OpenSSLProvider();
+  }
+
+  @Override public X509TrustManager trustManager(SSLSocketFactory sslSocketFactory) {
+    if (Conscrypt.isConscrypt(sslSocketFactory)) {
```
swankjesse: simpler to invert this and remove a layer of indentation?

```
if (!Conscrypt.isConscrypt(sslSocketFactory)) {
  return super.trustManager(sslSocketFactory);
}
```

?

okhttp/src/main/java/okhttp3/internal/platform/ConscryptPlatform.java
```diff
@@ -0,0 +1,99 @@
...
+            "clientBuilder.sslSocketFactory(SSLSocketFactory) not supported on Conscrypt", e);
+      }
+    } else {
+      return super.trustManager(sslSocketFactory);
+    }
+  }
+
+  @Override public void configureTlsExtensions(
+      SSLSocket sslSocket, String hostname, List<Protocol> protocols) {
+    if (Conscrypt.isConscrypt(sslSocket)) {
```
swankjesse: same

okhttp/src/main/java/okhttp3/internal/platform/ConscryptPlatform.java
```diff
@@ -0,0 +1,99 @@
...
+      // Enable SNI and session tickets.
+      if (hostname != null) {
+        Conscrypt.setUseSessionTickets(sslSocket, true);
+        Conscrypt.setHostname(sslSocket, hostname);
```
swankjesse: love these

okhttp/src/main/java/okhttp3/internal/platform/ConscryptPlatform.java
```diff
@@ -0,0 +1,99 @@
...
+      }
+
+      // Enable ALPN.
```
swankjesse: Nice

okhttp/src/main/java/okhttp3/internal/platform/ConscryptPlatform.java
```diff
@@ -0,0 +1,99 @@
...
+    } else {
+      super.configureTlsExtensions(sslSocket, hostname, protocols);
+    }
+  }
+
+  @Override public String getSelectedProtocol(SSLSocket sslSocket) {
+    if (Conscrypt.isConscrypt(sslSocket)) {
+      String alpnResult = Conscrypt.getApplicationProtocol(sslSocket);
+
+      return alpnResult != null ? alpnResult : null;
```
swankjesse: return alpnResult

okhttp/src/main/java/okhttp3/internal/platform/ConscryptPlatform.java
```diff
@@ -0,0 +1,99 @@
...
+
+      // Enable ALPN.
+      List<String> names = Platform.alpnProtocolNames(protocols);
+      Conscrypt.setApplicationProtocols(sslSocket, names.toArray(new String[0]));
+    } else {
+      super.configureTlsExtensions(sslSocket, hostname, protocols);
+    }
+  }
+
+  @Override public String getSelectedProtocol(SSLSocket sslSocket) {
```
swankjesse: `@Nullable`

okhttp/src/main/java/okhttp3/internal/platform/ConscryptPlatform.java
swankjesse: ALSO – should we fail if we’re using a Concrypt platform but not its SSL sockets? That’s a weird situation to be in and it might be more friendly to just explode?

okhttp/src/main/java/okhttp3/internal/platform/ConscryptPlatform.java
```diff
@@ -0,0 +1,99 @@
...
+ * Platform using Conscrypt (conscrypt.org) if system property
+ * okhttp.platform is set to conscrypt.
+ *
+ * Requires org.conscrypt:conscrypt-openjdk-uber on the classpath.
+ */
+public class ConscryptPlatform extends Platform {
+  private ConscryptPlatform() {
+  }
+
+  public Provider getProvider() {
```
swankjesse: private?

okhttp/src/main/java/okhttp3/internal/platform/Platform.java
swankjesse: oooh I like that option. Then we’re relying on the same mechanism the Java security infrastructure wants us to. I like that. Can be now or in follow-up.

swankjesse:
Great PR.

okhttp-tests/src/test/java/okhttp3/internal/http2/HttpOverHttp2Test.java
yschimke: Needed because I don't think Conscrypt mixes well with fake sockets.  The tests fallback to HTTP/1.1 and then fail.

yschimke:
Tested externally with JDK 8 *without* alpn-boot and installing the provider.

https://github.com/yschimke/oksocial/pull/311

```
OkHttp Platform: ConscryptPlatform
TLS Version: TLS_1_2
Protocol: h2
Cipher: TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256
```

yschimke:
@swankjesse do you want to land this for 3.10 or post for more testing before a final cut?  If we are happy with the gating, I'd be happy to include sooner rather than later.

swankjesse:
I think it's our banner feature for 3.10.

yschimke:
It's secondary to AndroidPlatform, so unavailable there. I don't know whether anyone would want to use it ahead over the optional GCM provider.

How much usage does oksocial have on non-Android?

yschimke:
1.0.1 just released

yschimke:
@swankjesse Do you want to do any final checks and merge?

okhttp-tests/src/test/java/okhttp3/ConscryptTest.java
```diff
@@ -0,0 +1,107 @@
...
+    try {
+      InetAddress.getByName("www.google.com");
+    } catch (UnknownHostException uhe) {
+      Assume.assumeNoException(uhe);
+    }
+  }
+
+  @Test
+  public void testMozilla() throws IOException {
+    assumeNetwork();
```
swankjesse: Neat

okhttp/src/main/java/okhttp3/internal/platform/ConscryptPlatform.java
```diff
@@ -0,0 +1,124 @@
...
+ * Platform using Conscrypt (conscrypt.org) if installed as the first Security Provider.
+ *
+ * Requires org.conscrypt:conscrypt-openjdk-uber on the classpath.
+ */
+public class ConscryptPlatform extends Platform {
+  private ConscryptPlatform() {
+  }
+
+  public static boolean isPreferredPlatform() {
+    // mainly to allow tests to run cleanly
```
swankjesse: Cool!

swankjesse: Is this redundant with Platform.isConscryptPreferred ?

yschimke:
I haven't been able to reproduce those errors yet locally (JDK8 + alpn-boot), and they have only failed in travis once. n.b. I'm pretty sure I've seen similar failures during the PR development as I temporarily made the tests more lenient at one point.

They are most brittle since they are real traffic and look at the actual frames which I assume can validly vary slightly if the test is not controlled again.  At the extreme we expect these tests to fail if the CI environment is slow enough to trigger timeouts.  But I'd rather them brittle and flaky over making them more tolerant without understanding it.

My suggestion for now is just monitoring if it happens again in travis.

yschimke:
Tried various things to reproduce the failure locally.  No luck.

```
  public static void main(String[] args) throws Exception {

    Logger.getLogger(MockWebServer.class.getName()).setLevel(Level.WARNING);

    for (int i = 0; i < 10000; i++) {
      HttpOverHttp2Test test = new HttpOverHttp2Test();
      try {
        test.tempDir.create();
        test.server.start();
        test.setUp();
        test.emptyDataFrameSentWithEmptyBody();
        System.err.print(".");
        if (i % 100 == 99) {
          System.err.println();
        }
        test.tearDown();
      } finally {
        test.tempDir.delete();
        test.server.shutdown();
      }
    }
  }
```

Going to land based on the Approval and see if it happens again.  At the moment it's effectively a once off build failure.

swankjesse:
Good call. The tests work for us, not vice-versa.

PR #3277, opened 2017-04-09
Prevent NPE in Reponse#close() when body is null

joschi:
The `Response#close()` method threw a `NullPointerException` if it was called and
the response body was `null`.

Since Response is `Closable` (and transitively `AutoClosable`), it should be possible
to use a `Response `object in a [try-with-resources](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html) block without thinking about whether
the `#close()` method would throw an exception in some cases.

**Examples:**

Before:

    Response response = client.newCall(call).execute();
    try {
      // do something with response
    } finally {
      if (response.body() != null) {
        response.close();
      }
    }

After:

    try (Response response = client.newCall(call).execute()) {
      // do something with response
    }

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/ResponseTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/ResponseTest.java
@@ -24,6 +24,8 @@ import okio.Timeout;
 import org.junit.Test;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
 import static org.junit.Assert.fail;
 
 public final class ResponseTest {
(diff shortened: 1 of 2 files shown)


swankjesse:
When is the after case possible? Responses returned from calls always have a body.

joschi:
@swankjesse According to the Javadoc for [`Response#body()`](https://square.github.io/okhttp/3.x/okhttp/okhttp3/Response.html#body--):
> This always returns null on responses returned from `cacheResponse`, `networkResponse`, and `priorResponse()`.

While it technically doesn't make sense to "close" the request on responses returned by these methods (or does it?), it will lead to a `NullPointerException` if the developer didn't check if `body()` returned `null`.

This pull request takes the surprise element out of calling `Response#close()`.

FWIW, I ran into this when writing a method handling `Response` objects (without knowing their origin) and closing them at the end.

swankjesse:
I just think it's a very rare situation. You're closing a response without attempting to read its body.

joschi:
@swankjesse In my opinion it's the question, whether a developer should face a `NullPointerException` when basically doing "the right thing", i. e. closing potentially open resources.

Especially when using try-with-resources, this situation isn't that rare.

It may be a personal issue, but I think

    try (Response response = ...) {
      // do something with response
    }

beats

    Response response = ...;
    try {
      // do something with response
    } finally {
      if (response.body() != null) {
        response.close();
      }
    }

in the question of readability (and laziness of the developer).

If you don't want to merge this pull request it's totally fine, but I'd recommend adding a Javadoc comment about the potential `NullPointerException` to `Response#close()` in that case.

swankjesse:
A Response from a Call is quite different from a Response from `priorResponse()`, etc. We cheat by using the same type for both, and for the most part this works well. 

Can you give me more context on what the `// do something with response` code is actually doing? I haven’t needed to write generic code that handles both `priorResponse()`and regular `Response` objects. The closer we bring them together the more likely callers will crash dereferencing the body.

Issue #3243, opened 2017-03-26
Expose public suffix info via an API


swankjesse:
We have the public suffix DB. Should we expose it with an API somewhere?

Perhaps something like this?

```
HttpUrl url = HttpUrl.parse("https://adwords.google.co.uk");
assertEquals("google.co.uk", url.topPrivateDomain());
```

(I’m borrowing the method name from Guava here.)

dave-r12:
That might be handy. I haven't gotten this far yet, but I was going to experiment with using the _top private domain_ to bucket cookies in the cookie jar. Pulling it directly off the HttpUrl would be nice.

dave-r12:
@swankjesse was going to look at this one, please let me know if you have something in progress.

swankjesse:
Please go ahead!

PR #3283, opened 2017-04-15
Expose public suffix info via an API in HttpUrl.

dave-r12:
Closes: https://github.com/square/okhttp/issues/3243

### DIFF
--- a/okhttp/src/main/java/okhttp3/HttpUrl.java
+++ b/okhttp/src/main/java/okhttp3/HttpUrl.java
@@ -27,12 +27,14 @@ import java.util.Collections;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
+import okhttp3.internal.publicsuffix.PublicSuffixDatabase;
 import okio.Buffer;
 
 import static okhttp3.internal.Util.delimiterOffset;
 import static okhttp3.internal.Util.domainToAscii;
 import static okhttp3.internal.Util.skipLeadingAsciiWhitespace;
 import static okhttp3.internal.Util.skipTrailingAsciiWhitespace;
+import static okhttp3.internal.Util.verifyAsIpAddress;
 
 /**
  * A uniform resource locator (URL) with a scheme of either {@code http} or {@code https}. Use this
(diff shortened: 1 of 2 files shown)


okhttp/src/main/java/okhttp3/HttpUrl.java
```diff
@@ -945,6 +947,50 @@ public static HttpUrl get(URI uri) {
...
+   *   <tr><th>URL</th><th>{@code isPublicSuffix()}</th></tr>
+   *   <tr><td>{@code http://com}</td><td>{@code true}</td></tr>
+   *   <tr><td>{@code http://co.uk}</td><td>{@code true}</td></tr>
+   *   <tr><td>{@code http://square}</td><td>{@code false}</td></tr>
+   *   <tr><td>{@code http://google.com}</td><td>{@code false}</td></tr>
+   *   <tr><td>{@code http://localhost}</td><td>{@code false}</td></tr>
+   *   <tr><td>{@code http://127.0.0.1}</td><td>{@code false}</td></tr>
+   * </table>
+   */
+  public boolean isPublicSuffix() {
```
swankjesse: My initial reaction to this is mixed. I don’t think it fits as an API on HttpUrl. If you’re doing domain name parsing for something, you’re unlikely to get `https://com/` as a URL.

But I agree that `topPrivateDomain()` is missing one use case, which is that it tells you whether the domain is _private_ but not that it is _routable_. But I also think there’s no way you can discover routability without attempting it. Heck, even `https://com/` could be reachable on some networks. And of course `https://localhost` is reachable.

So – I propose dropping this and focusing the docs in `topPrivateDomain()` on the security/ownership consequences of the name: it’s a unit of trust. Basically we won’t offer any APIs for estimating routability; for that you have to just make the HTTP call.

okhttp/src/main/java/okhttp3/HttpUrl.java
```diff
@@ -945,6 +947,50 @@ public static HttpUrl get(URI uri) {
...
+   *   <tr><th>URL</th><th>{@code topPrivateDomain()}</th></tr>
+   *   <tr><td>{@code http://google.com}</td><td>{@code "google.com"}</td></tr>
+   *   <tr><td>{@code http://adwords.google.co.uk}</td><td>{@code "google.co.uk"}</td></tr>
+   *   <tr><td>{@code http://square}</td><td>null</td></tr>
+   *   <tr><td>{@code http://co.uk}</td><td>null</td></tr>
+   *   <tr><td>{@code http://localhost}</td><td>null</td></tr>
+   *   <tr><td>{@code http://127.0.0.1}</td><td>null</td></tr>
+   * </table>
+   */
+  public String topPrivateDomain() {
```
swankjesse: Nice doc!

okhttp/src/main/java/okhttp3/HttpUrl.java
swankjesse: (and if I’m missing something, I still think it might want to be a static method like `HttpUrl.isPublicSuffix(String)`. Unsure though.

dave-r12: That makes sense. I'll try to incorporate that into the docs of `topPrivateDomain()`.

Issue #3228, opened 2017-03-19
WebSocketHttpTest.throwingOnMessageClosesImmediatelyAndFails is flaky


swankjesse:
https://s3.amazonaws.com/archive.travis-ci.org/jobs/212680155/log.txt

```
throwingOnMessageClosesImmediatelyAndFails(okhttp3.internal.ws.WebSocketHttpTest)  Time elapsed: 0.02 sec  <<< FAILURE!
java.lang.AssertionError: Remaining events: [Failure[java.io.EOFException]]
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at okhttp3.internal.ws.WebSocketRecorder.assertExhausted(WebSocketRecorder.java:165)
	at okhttp3.internal.ws.WebSocketHttpTest.throwingOnMessageClosesImmediatelyAndFails(WebSocketHttpTest.java:190)
```

PR #3284, opened 2017-04-15
Fix WebSocketHttpTest.throwingOnMessageClosesImmediatelyAndFails().

swankjesse:
The server will receive an EOFException when the client errors out.
But because that didn't usually happen immediately our test got away
with confirming there were no further events.

Also fix WebSocketHttpTest.readTimeoutAppliesToHttpRequest().

Closes: https://github.com/square/okhttp/issues/3228
Closes: https://github.com/square/okhttp/issues/3090

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/internal/ws/WebSocketRecorder.java
+++ b/okhttp-tests/src/test/java/okhttp3/internal/ws/WebSocketRecorder.java
@@ -16,6 +16,7 @@
 package okhttp3.internal.ws;
 
 import java.io.IOException;
+import java.util.Arrays;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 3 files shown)


okhttp-tests/src/test/java/okhttp3/internal/ws/WebSocketRecorder.java
```diff
@@ -183,15 +184,17 @@ public void assertFailure(Throwable t) {
...
     Object event = nextEvent();
     if (!(event instanceof Failure)) {
       throw new AssertionError("Expected Failure but was " + event);
     }
     Failure failure = (Failure) event;
     assertNull(failure.response);
     assertEquals(cls, failure.t.getClass());
-    assertEquals(message, failure.t.getMessage());
+    if (messages.length > 0) {
+      assertTrue(failure.t.getMessage(), Arrays.asList(messages).contains(failure.t.getMessage()));
```
dave-r12: 👍

PR #3285, opened 2017-04-15
Hide EventListener from the public API for OkHttp 3.7.

swankjesse:
It isn't wired up yet.


dave-r12:
Did you mean OkHttp 3.7?

swankjesse:
Yep. Whoops!

PR #3288, opened 2017-04-16
Update JavaDoc for WebSocketListener

adi1133:


JakeWharton:
I'm not sure this is necessary. Is there something specific that "peer" didn't convey when reading this? Also WebSockets can be used locally. For example Chrome talks WebSockets to the developer tools panel even though they live inside the same app.

adi1133:
My intention is to avoid ambiguity about which peer triggers this event (local or remote), the phrase "remote peer" is used in [L36](https://github.com/adi1133/okhttp/blob/f03d44de5558c8798a7cbb26bce1efa2130ad2ac/okhttp/src/main/java/okhttp3/WebSocketListener.java#L36) and [L22](https://github.com/adi1133/okhttp/blob/f03d44de5558c8798a7cbb26bce1efa2130ad2ac/okhttp/src/main/java/okhttp3/WebSocketListener.java#L36).

JakeWharton:
Makes sense. Thanks for clarifying.

swankjesse:
Thanks!

PR #3295, opened 2017-04-19
Master observer

JeffWegherInstart:
We have implemented a lower-level statistics gathering mechanism for OkHttp that gives us visibility into the performance of the various stages of the network performance.

The primary changes involved passing around a StatisticsData object to various methods within the Codecs and related classes.

Upon stream completion if an observer has been set on the Request object it will be given the statistics that were gathered.


swankjesse:
We're already working on something similar. It's called EventListener.
https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/EventListener.java

JeffWegherInstart:
That is certainly very similar but does not appear to include byte counts. Would you be interested in providing me some guidance on implementing our code into your interface (with the addition of the byte counts) ?

swankjesse:
We could capture and report byte counts and report them to our EventListener interface. We’d probably overcount due to compression in HTTP/2, and undercount due to TLS.

Perhaps a better way to count bytes is by doing a custom SocketFactory, since only it sees the truth.

JeffWegherInstart:
We started out with a custom SocketFactory and custom SSLSocketFactory. Android security issues prevented us from using it (which took a long time to prove out) which is why we wound up implementing the code directly inside of OkHttp.

Please take a look at the way I implemented the byte counting, it should be accurate.

JeffWegherInstart:
Although yes you are correct that we have no control over TLS and what it ultimately writes.

swankjesse:
Could you provide more info on the problems with the custom SocketFactory?

JeffWegherInstart:
We created an SSLSocketFactory that created a simple delegating SSLSocket.
We used reflection (in addition to examining the Android source code) to
make sure that we were delegating all methods and copying all field values
correctly through.

When the delegating SSLSocket goes in to the handshake, it fails and closes
the socket. It's down in the native C library at that point, I can't
remember specifically which function. The exact same code running on the
Mac works just fine.

So there is something in the security model that prevents using a
delegating SSLSocket.

JeffWegherInstart:
Hey guys, do you have any thoughts on this?

swankjesse:
We’re implementing similar behavior here:
https://github.com/square/okhttp/pull/3410

No plans to go forward with this PR.

JeffWegherInstart:
Currently EventListener has no concept of byte counts sent and received. Is there a plan for adding this?

swankjesse:
Unsure. Its difficult to be precise here, especially because we lack visibility into the TLS layer.

JeffWegherInstart:
So, I briefly considered (and prototyped) ways to count the raw bytes at
the TLS layer and it's very doable, however there's very little benefit to
what we're looking to accomplish. Aside from the data exchanged for the
handshake which we don't care about, the number of encrypted bytes will be
very close to the number of non-encrypted bytes.

All we need are the timings around sending the headers and body and
receiving the headers and body. My pull request includes a lot of code that
accurately counts the bytes of the non-encrypted data.

The EventListener interface could certainly work for us with a few tweaks
and we would like to modify the work we've done to fit within this
framework. Can we work together to ensure a suitable enhancement to the
EventListener interface?

Thanks,

Jeff

PR #3307, opened 2017-04-25
Fix for EOFException if response body is empty

stasys123:
When Level is set to BODY and the POST request response status is 200 and response body contains no content - EOFException is thrown. This code handles the situation and doe not try to print body if it is empty.


autonomousapps:
+1 to preventing exceptions while logging!

JakeWharton:
Can you provide a test case that demonstrates the problem?

stasys123:
Seems that my "fix" completely disables body logging...
Sorry but I don't have time make the test.
Stacktrace:
java.io.EOFException
     at okio.RealBufferedSource.require(RealBufferedSource.java:59)
     at okio.RealBufferedSource.request(RealBufferedSource.java:66)
     at okhttp3.logging.HttpLoggingInterceptor.intercept(HttpLoggingInterceptor.java:238)

Looks like I will have to make a wrapper for HttpLoggingInterceptor and just ignore EOFException.

Closing pr.

PR #3311, opened 2017-04-27
Log the name of the null header

Jawnnypoo:
This makes it much easier to know what is going on and know which header is causing problems (without having to go through some debugging steps)


swankjesse:
The build failure is unrelated. ALPN needs to be updated.
https://github.com/square/okhttp/pull/3315

swankjesse:
Merged manually. Needed a single char of whitespace to make checkstyle happy.

Jawnnypoo:
Sounds great! Thanks!

PR #3316, opened 2017-05-01
Update Call Test & Route Selector

ericaschulz:


okhttp-tests/src/test/java/okhttp3/CallTest.java
```diff
@@ -94,7 +94,7 @@
 import static org.junit.Assert.fail;
 
 public final class CallTest {
-  @Rule public final TestRule timeout = new Timeout(30_000, TimeUnit.MILLISECONDS);
+  @Rule public final TestRule timeout = new Timeout(3000_000, TimeUnit.MILLISECONDS);
```
swankjesse: Setting this back to 0 to merge.

okhttp/pom.xml
```diff
@@ -22,6 +22,16 @@
       <artifactId>android</artifactId>
       <scope>provided</scope>
     </dependency>
+    <dependency>
```
swankjesse: Removing this too. I think IntelliJ might have added it.

pom.xml
```diff
@@ -48,7 +48,7 @@
     <apache.http.version>4.2.2</apache.http.version>
     <bouncycastle.version>1.50</bouncycastle.version>
     <guava.version>16.0</guava.version>
-    <java.version>1.7</java.version>
+    <java.version>1.8</java.version>
```
swankjesse: And this!

swankjesse:
Merged! Thanks!

PR #3331, opened 2017-05-06
Import jsr305 and use it to mark @Nullable stuff.

swankjesse:
This library is a provided dependency and is not necessary to be included
by OkHttp users.

This currently covers the okhttp and okhttp-logging-interceptor modules
only. It does not cover the 'internal' package in OkHttp, or its test cases.
We can add those as needed.


swankjesse:
Wants to be rebased on
https://github.com/square/okhttp/pull/3332

Issue #3245, opened 2017-03-28
java.lang.NullPointerException throw with null exception


liujigang:
error stack:
```
1 java.net.Socket.getInputStream(Socket.java:894)
2 okio.Okio.okio.Source source(java.net.Socket)(Okio.java:220)
3 okhttp3.internal.connection.RealConnection.void connectSocket(int,int)(RealConnection.java:226)
4 okhttp3.internal.connection.RealConnection.void connect(int,int,int,boolean)(RealConnection.java:146)
5 okhttp3.internal.connection.StreamAllocation.okhttp3.internal.connection.RealConnection findConnection(int,int,int,boolean)(StreamAllocation.java:186)
6 okhttp3.internal.connection.StreamAllocation.okhttp3.internal.connection.RealConnection findHealthyConnection(int,int,int,boolean,boolean)(StreamAllocation.java:121)
7 okhttp3.internal.connection.StreamAllocation.okhttp3.internal.http.HttpCodec newStream(okhttp3.OkHttpClient,boolean)(StreamAllocation.java:100)
8 okhttp3.internal.connection.ConnectInterceptor.okhttp3.Response intercept(okhttp3.Interceptor$Chain)(ConnectInterceptor.java:42)
9 okhttp3.internal.http.RealInterceptorChain.okhttp3.Response proceed(okhttp3.Request,okhttp3.internal.connection.StreamAllocation,okhttp3.internal.http.HttpCodec,okhttp3.Connection)(RealInterceptorChain.java:92)
10 okhttp3.internal.http.RealInterceptorChain.okhttp3.Response proceed(okhttp3.Request)(RealInterceptorChain.java:67)
11 okhttp3.internal.cache.CacheInterceptor.okhttp3.Response intercept(okhttp3.Interceptor$Chain)(CacheInterceptor.java:93)
12 okhttp3.internal.http.RealInterceptorChain.okhttp3.Response proceed(okhttp3.Request,okhttp3.internal.connection.StreamAllocation,okhttp3.internal.http.HttpCodec,okhttp3.Connection)(RealInterceptorChain.java:92)
13 okhttp3.internal.http.RealInterceptorChain.okhttp3.Response proceed(okhttp3.Request)(RealInterceptorChain.java:67)
14 okhttp3.internal.http.BridgeInterceptor.okhttp3.Response intercept(okhttp3.Interceptor$Chain)(BridgeInterceptor.java:93)
[...]

PR #3346, opened 2017-05-11
Fix for NPE Crash due to a bug on Android 7.0 

githubsaturn:
Fix the crash reported in https://github.com/square/okhttp/issues/3245


swankjesse:
Thanks!

post2shyam:
Late though. But shouldn't android version be checked explicitly ? Since it's version 7 specific.

swankjesse:
@post2shyam that’s difficult to do well. Send a PR if you’d like to find out what I mean!

PR #3352, opened 2017-05-13
Fix Javadoc builds with the jsr305 dependency.

swankjesse:
Because multiple modules share a package we need to share the dependency on
everything in package-info.java.

### DIFF
--- a/okhttp-android-support/pom.xml
+++ b/okhttp-android-support/pom.xml
@@ -17,6 +17,11 @@
 
   <dependencies>
     <dependency>
+      <groupId>com.google.code.findbugs</groupId>
+      <artifactId>jsr305</artifactId>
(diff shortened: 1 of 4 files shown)


okhttp-android-support/pom.xml
```diff
@@ -17,6 +17,12 @@
 
   <dependencies>
     <dependency>
+      <groupId>com.google.code.findbugs</groupId>
+      <artifactId>jsr305</artifactId>
+      <version>3.0.2</version>
```
JakeWharton: You don't need to repeat the version in all these though, right? The dependency management in the root will take care of that

swankjesse: ooh, you’re right.

Issue #3325, opened 2017-05-05
NetworkSecurityPolicy is not honored on Android API 23


ycai2:
I tried to block cleartext HTTP requests from my app by setting `usesCleartextTraffic` in my manifest file. When I set `android:usesCleartextTraffic="false"`, the `isCleartextTrafficPermitted()` function in Android's `NetworkSecurityPolicy` is returning `false` as expected. 
However, as I use OKHTTP in my application, HTTP calls are _not_ blocked with this configuration. 

I looked into OKHTTP's code. The problem seems to have been introduced in this PR: https://github.com/square/okhttp/pull/2686.
As of that PR, `AndroidPlatform.java` tries to call `isCleartextTrafficPermitted()` with a string argument ([this spot](https://github.com/square/okhttp/blob/parent-3.7.0/okhttp/src/main/java/okhttp3/internal/platform/AndroidPlatform.java#L155)), even though passing an argument is only supported as of API 24 (see: [Android's docs](https://developer.android.com/reference/android/security/NetworkSecurityPolicy.html)).
It then throws `ClassNotFoundException`, goes to `super.isCleartextTrafficPermitted`, which always returns `true`.

This means any device on API 23 (Marshmallow) wouldn't be able to block any cleartext traffic.

While on API version 24 the PR makes the policy more granular, on 23 it makes blocking cleartext not work at all. So, it might make sense to have a fallback for when the method with an argument isn't found, that will call the method without an argument.

ycai2:
@swankjesse Please take a look at this.

ycai2:
@Egorand Is this going to be a PR?

Egorand:
@ycai2 it's been merged already: https://github.com/square/okhttp/pull/3353
The fix is available in 3.9.0-SNAPSHOT, could you please verify that it fixes the issue for you?

PR #3353, opened 2017-05-13
#3325: Add proper fallback to parameterless isCleartextTrafficPermitted() method on API 23

Egorand:
I've created a project to reproduce the issue and published it along with a bundled snapshot that contains the fix: https://github.com/Egorand/okhttp-android-tests


swankjesse:
Thanks Egor!

PR #3354, opened 2017-05-13
README: ProGuard changes for OkHttp 3.8.0

talklittle:


swankjesse:
Brilliant, thank you. Wanna do Retrofit also?

talklittle:
@swankjesse Thanks for the offer but I do not currently use Retrofit so it would be an untested config. I guess the place to document it would be https://github.com/square/retrofit/blob/master/website/index.html

JakeWharton:
It's the same two additions.

talklittle:
Go for it, I don't need to take credit for the PR on Retrofit.

Issue #3308, opened 2017-04-25
NoSuchElementException in RouteSelector


Jeff11:
okhttp 3.7.0
retrofit 2.2.0

All device brands and Android versions are reporting this issue. 

It happened for example after I tapped a notification, the app then makes a GET which ended in this exception. It didn't happen when I recreated the scenario. Device was Pixel, Android 7.1.2, on WLAN.

```
java.util.NoSuchElementException
  at okhttp3.internal.connection.RouteSelector.next(RouteSelector.java:77)
  at okhttp3.internal.connection.StreamAllocation.findConnection(StreamAllocation.java:171)
  at okhttp3.internal.connection.StreamAllocation.findHealthyConnection(StreamAllocation.java:121)
  at okhttp3.internal.connection.StreamAllocation.newStream(StreamAllocation.java:100)
  at okhttp3.internal.connection.ConnectInterceptor.intercept(ConnectInterceptor.java:42)
  at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:92)
  at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:67)
  at okhttp3.internal.cache.CacheInterceptor.intercept(CacheInterceptor.java:93)
  at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:92)
  at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:67)
  at okhttp3.internal.http.BridgeInterceptor.intercept(BridgeInterceptor.java:93)
  at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:92)
  at okhttp3.internal.http.RetryAndFollowUpInterceptor.intercept(RetryAndFollowUpInterceptor.java:120)
  at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:92)
  at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:67)
  at okhttp3.logging.HttpLoggingInterceptor.intercept(HttpLoggingInterceptor.java:211)
  at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:92)
  at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:67)
  at com.readystatesoftware.chuck.ChuckInterceptor.intercept(ChuckInterceptor.java:167)
[...]

PR #3370, opened 2017-05-23
Recover from a stale coalesced connection.

dave-r12:
https://github.com/square/okhttp/issues/3308

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/ConnectionCoalescingTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/ConnectionCoalescingTest.java
@@ -20,6 +20,7 @@ import java.net.InetAddress;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


dave-r12:
This seems like a corner case (maybe impossible?), and might be related to #3308.

okhttp-tests/src/test/java/okhttp3/ConnectionCoalescingTest.java
```diff
@@ -117,6 +118,33 @@
...
+            connection.set(chain.connection());
+            return chain.proceed(chain.request());
+          }
+        })
+        .build();
+    dns.set("san.com", Dns.SYSTEM.lookup(server.getHostName()).subList(0, 1));
+
+    assert200Http2Response(execute(url), server.getHostName());
+
+    // Simulate a stale connection in the pool.
```
swankjesse: Neat

swankjesse:
Nice find.

PR #3377, opened 2017-05-25
Avoid index out of bounds with Http/2 settings id in the experimental range

ericgribkoff:
Currently HTTP/2 setting ids in the [designated experimental range](http://httpwg.org/specs/rfc7540.html#iana-settings), `#f000` to `#ffff`, cause an index-out-of-bounds exception when consumed by `Http2Reader#readSettings`, as `readShort()` interprets the bytes as negative integers. This expands the out-of-bounds check in `Settings#set` to check for a negative index. Without the code change, the added test fails, demonstrating the bug.

I'm not fully familiar with OkHttp internals to know if this is the best approach for the fix, but it seemed better to add a PR with a repro case than just opening an issue. This popped up in gRPC Java (https://github.com/grpc/grpc-java/issues/3032), and extends the earlier fix from  https://github.com/square/okhttp/pull/2299.

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/internal/http2/Http2Test.java
+++ b/okhttp-tests/src/test/java/okhttp3/internal/http2/Http2Test.java
@@ -278,6 +278,21 @@ public final class Http2Test {
     assertEquals(settingValue.intValue(), 1);
   }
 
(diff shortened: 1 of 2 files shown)


swankjesse:
Thanks!

PR #3390, opened 2017-06-07
Update MockWebServer.java

Douvi:
reset the value `started` when server is shutdown


swankjesse:
We don’t support reusing instances of MockWebServer. If you want to start it again, create a new instance.

Issue #3324, opened 2017-05-05
Invalid Wildcard Certificate Pinning Check Using Substring Matching


ChaosData:
What kind of issue is this?

 - [x] Bug report.

Hi, 

The `CertificatePinner.Pin::matches` method [<https://github.com/square/okhttp/blob/parent-3.7.0/okhttp/src/main/java/okhttp3/CertificatePinner.java#L289>] does not validate the length of the "canonical" hostname against the length of the target hostname. Due to this, a pin hostname of `"*.foo.com"` will match `"bar.foo.com.notfoo.com"` (ofc the hashes would also need to match for someone to intercept a request ;). With the current matching scheme, it is necessary to perform a check to verify that the "canonical" hostname length matches the remaining length of the target hostname (e.g. the part after the first `'.'`).

PR #3399, opened 2017-06-10
Match the entire pattern on certificate pinner patterns.

swankjesse:
Closes: https://github.com/square/okhttp/issues/3324

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/CertificatePinnerTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/CertificatePinnerTest.java
@@ -275,4 +275,32 @@ public final class CertificatePinnerTest {
     List<Pin> expectedPin = Arrays.asList(new Pin("σkhttp.com", certA1Sha256Pin));
     assertEquals(expectedPin, certificatePinner.findMatchingPins("xn--khttp-fde.com"));
   }
+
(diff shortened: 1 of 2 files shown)


okhttp-tests/src/test/java/okhttp3/CertificatePinnerTest.java
```diff
@@ -275,4 +275,32 @@
...
+    assertEquals(Collections.emptyList(),
+        certificatePinner.findMatchingPins("a.example.com.notexample.com"));
+    assertEquals(Collections.emptyList(),
+        certificatePinner.findMatchingPins("example.com.notexample.com"));
+    assertEquals(Collections.emptyList(),
+        certificatePinner.findMatchingPins("notexample.com"));
+    assertEquals(Collections.emptyList(),
+        certificatePinner.findMatchingPins("example.com"));
+    assertEquals(Collections.emptyList(),
+        certificatePinner.findMatchingPins("a.b.example.com"));
```
artem-zinnatullin: Hm, isn't it a valid use case? Or there is another way to match subdomains of all levels?

swankjesse: Nope. This is consistent with the way subject-alt-name works on TLS certificates.

PR #3405, opened 2017-06-13
Disallow creating Responses with no body.

NightlyNexus:
via https://github.com/square/retrofit/issues/2362
Is there a valid use case for making synthetic cache responses, though?


swankjesse:
We need these for the responses in cacheResponse() and networkResponse().

For that particular problem we could confirm the responses returned from interceptors are complete.

JakeWharton:
Yes that's the check I was trying to allude to. Responses from user
interceptors should have non-null bodies (we can probably safely assume
ours are correct).

NightlyNexus:
oops, duh. i'll take some stabs at that tomorrow morning.

Issue #3404, opened 2017-06-13
[Bug report] publicsuffix.gz read crash 


in7in7:
It seems to occur very rarely
Perhaps glide has canceled it and it seems to be thread interrupted

use library
- okhttp3 3.7.0
- glide 3.7.0

OkHttp Log
06-12 14:29:36.716 17923 18274 W OkHttp  : Failed to read public suffix list
06-12 14:29:36.716 17923 18274 W OkHttp  : java.io.InterruptedIOException: thread interrupted
06-12 14:29:36.716 17923 18274 W OkHttp  : 	at okio.Timeout.throwIfReached(Timeout.java:146)
06-12 14:29:36.716 17923 18274 W OkHttp  : 	at okio.Okio$2.read(Okio.java:135)
06-12 14:29:36.716 17923 18274 W OkHttp  : 	at okio.RealBufferedSource.request(RealBufferedSource.java:66)
06-12 14:29:36.716 17923 18274 W OkHttp  : 	at okio.RealBufferedSource.require(RealBufferedSource.java:59)
06-12 14:29:36.716 17923 18274 W OkHttp  : 	at okio.GzipSource.consumeHeader(GzipSource.java:114)
06-12 14:29:36.716 17923 18274 W OkHttp  : 	at okio.GzipSource.read(GzipSource.java:73)
06-12 14:29:36.716 17923 18274 W OkHttp  : 	at okio.RealBufferedSource.request(RealBufferedSource.java:66)
06-12 14:29:36.716 17923 18274 W OkHttp  : 	at okio.RealBufferedSource.require(RealBufferedSource.java:59)
06-12 14:29:36.716 17923 18274 W OkHttp  : 	at okio.RealBufferedSource.readInt(RealBufferedSource.java:255)
06-12 14:29:36.716 17923 18274 W OkHttp  : 	at okhttp3.internal.publicsuffix.PublicSuffixDatabase.readTheList(PublicSuffixDatabase.java:288)
06-12 14:29:36.716 17923 18274 W OkHttp  : 	at okhttp3.internal.publicsuffix.PublicSuffixDatabase.findMatchingRule(PublicSuffixDatabase.java:109)
06-12 14:29:36.717 17923 18274 W OkHttp  : 	at okhttp3.internal.publicsuffix.PublicSuffixDatabase.getEffectiveTldPlusOne(PublicSuffixDatabase.java:82)
[...]

PR #3408, opened 2017-06-15
Treat reading the public suffix list as uninterruptible.

dave-r12:
Fixes: https://github.com/square/okhttp/issues/3404


swankjesse:
Nice!

PR #3410, opened 2017-06-16
Make EventListener public and begin implementing Dns events.

dave-r12:
I'm interested in this one. 

I found this challenging because of the cycle between EventListener and Call. Notably, it seems like RealCall wants a reference to EventListener but the EventListener factory wants the instance of the Call. The approach in this PR is also broken 'cause someone could `execute` the call from the EventListener.

In addition, I attempted to use the EventListener for a successful DNS lookup. It became clear that this will require passing around the reference to EventListener. Also because every method takes in a Call, I think it means passing around the Call as well.

I probably am missing something here, or maybe this can help refine the EventListener class.

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
@@ -1,0 +1,166 @@
+/*
+ * Copyright (C) 2017 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okhttp3;
+
+import java.io.IOException;
+import java.net.InetAddress;
+import java.net.UnknownHostException;
+import java.util.ArrayDeque;
+import java.util.Collections;
+import java.util.Deque;
+import java.util.List;
+import okhttp3.internal.SingleInetAddressDns;
+import okhttp3.mockwebserver.MockResponse;
+import okhttp3.mockwebserver.MockWebServer;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertSame;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+public final class EventListenerTest {
+  @Rule public final MockWebServer server = new MockWebServer();
+
+  private OkHttpClient client;
+  private final RecordingEventListener listener = new RecordingEventListener();
+
+  @Before public void setUp() {
+    client = new OkHttpClient.Builder()
+        .dns(new SingleInetAddressDns())
+        .eventListener(listener)
+        .build();
+  }
+
+  @Test public void successfulDnsLookup() throws IOException {
+    server.enqueue(new MockResponse());
+
+    Call call = client.newCall(new Request.Builder()
+        .url(server.url("/"))
+        .build());
+    Response response = call.execute();
+    assertEquals(200, response.code());
+    response.body().close();
+
+    DnsStart dnsStart = listener.expectNextEvent(DnsStart.class);
+    assertSame(call, dnsStart.call);
+    assertEquals("localhost", dnsStart.domainName);
+
+    DnsEnd dnsEnd = listener.expectNextEvent(DnsEnd.class);
+    assertSame(call, dnsEnd.call);
+    assertEquals("localhost", dnsEnd.domainName);
+    assertEquals(1, dnsEnd.inetAddressList.size());
+    assertNull(dnsEnd.throwable);
+  }
+
+  @Test public void failedDnsLookup() {
+    client = client.newBuilder()
+        .dns(new FakeDns())
+        .build();
+    Call call = client.newCall(new Request.Builder()
+        .url("http://fakeurl/")
+        .build());
+    try {
+      call.execute();
+      fail();
+    } catch (IOException expected) {
+    }
+
+    listener.expectNextEvent(DnsStart.class);
+
+    DnsEnd dnsEnd = listener.expectNextEvent(DnsEnd.class);
+    assertSame(call, dnsEnd.call);
+    assertEquals("fakeurl", dnsEnd.domainName);
+    assertNull(dnsEnd.inetAddressList);
+    assertTrue(dnsEnd.throwable instanceof UnknownHostException);
+  }
+
+  @Test public void emptyDnsLookup() {
+    Dns emptyDns = new Dns() {
+      @Override public List<InetAddress> lookup(String hostname) throws UnknownHostException {
+        return Collections.emptyList();
+      }
+    };
+
+    client = client.newBuilder()
+        .dns(emptyDns)
+        .build();
+    Call call = client.newCall(new Request.Builder()
+        .url("http://fakeurl/")
+        .build());
+    try {
+      call.execute();
+      fail();
+    } catch (IOException expected) {
+    }
+
+    listener.expectNextEvent(DnsStart.class);
+
+    DnsEnd dnsEnd = listener.expectNextEvent(DnsEnd.class);
+    assertSame(call, dnsEnd.call);
+    assertEquals("fakeurl", dnsEnd.domainName);
+    assertNull(dnsEnd.inetAddressList);
+    assertTrue(dnsEnd.throwable instanceof UnknownHostException);
+  }
+
+  static final class DnsStart {
+    final Call call;
+    final String domainName;
+
+    DnsStart(Call call, String domainName) {
+      this.call = call;
+      this.domainName = domainName;
+    }
+  }
+
+  static final class DnsEnd {
+    final Call call;
+    final String domainName;
+    final List<InetAddress> inetAddressList;
+    final Throwable throwable;
+
+    DnsEnd(Call call, String domainName, List<InetAddress> inetAddressList, Throwable throwable) {
+      this.call = call;
+      this.domainName = domainName;
+      this.inetAddressList = inetAddressList;
+      this.throwable = throwable;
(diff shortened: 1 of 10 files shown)


okhttp/src/main/java/okhttp3/RealCall.java
```diff
@@ -33,7 +33,10 @@
 final class RealCall implements Call {
   final OkHttpClient client;
   final RetryAndFollowUpInterceptor retryAndFollowUpInterceptor;
-  final EventListener eventListener;
+
+  // There is a cycle between the Call and EventListener that makes this awkward. This will be set
+  // after we create the call instance then create the event listener instance.
+  volatile EventListener eventListener;
```
swankjesse: Does it need to be volatile? I’m trying to imagine when it’d ever be concurrent

okhttp/src/main/java/okhttp3/RealCall.java
```diff
@@ -33,7 +33,10 @@
 final class RealCall implements Call {
   final OkHttpClient client;
   final RetryAndFollowUpInterceptor retryAndFollowUpInterceptor;
-  final EventListener eventListener;
+
+  // There is a cycle between the Call and EventListener that makes this awkward. This will be set
```
swankjesse: Nit: convert to Javadoc comment so it’s attached for refactoring?

okhttp/src/main/java/okhttp3/RealCall.java
swankjesse: (the RealCall itself is safely published later when we share it with more threads)

okhttp/src/main/java/okhttp3/RealCall.java
```diff
@@ -168,6 +173,8 @@ String redactedUrl() {
   }
 
   Response getResponseWithInterceptorChain() throws IOException {
+    retryAndFollowUpInterceptor.setEventListener(eventListener);
```
swankjesse: should we put the eventListener on the RealInterceptorChain? That might make it easier for code that needs access to it to get access to it

okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
```diff
@@ -0,0 +1,85 @@
...
+
+    assertEquals(Events.DnsStart, listener.eventSequence.poll());
+    assertEquals("localhost", listener.dnsStartDomainName);
+
+    assertEquals(Events.DnsEnd, listener.eventSequence.poll());
+    assertEquals("localhost", listener.dnsEndDomainName);
+    assertEquals(1, listener.dnsEndInetAddressList.size());
+  }
+
+  enum Events {
```
swankjesse: should DnsStart and DnsEnd be classes?

swankjesse: Might be useful as this thing grows in complexity

okhttp/src/main/java/okhttp3/RealCall.java
dave-r12: At first I didn't have this, then I wondered what would happen if someone did fancy things in the EventListener.Factory.create method, i.e. something concurrent.

Also, it's unclear how to handle the case of someone calling `execute` from within the EventListener.Factory.create method. Any thoughts?

okhttp/src/main/java/okhttp3/RealCall.java
dave-r12: Yeah, good idea. I'll see what that looks like.

okhttp/src/main/java/okhttp3/RealCall.java
swankjesse: We need to prevent OkHttp from dereferencing `eventListener` before it’s initialized. All of the ways we will access this field already have a happens-before relation; ie. the assignment will completes before`execute()` is called.

okhttp/src/main/java/okhttp3/RealCall.java
dave-r12: Can do.

okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
dave-r12: Yes -- I adjusted this, please let me know if this was along the lines you were thinking.

okhttp/src/main/java/okhttp3/RealCall.java
dave-r12: Imagine an EventListener like this:
```java
class MisbehavingEventListener implements EventListener.Factory {
  @Override public EventListener create(Call call) {
    try {
      call.execute();
    } catch (IOException e) {
    }
    return null;
  }
}
```

swankjesse: This code is single-threaded, so volatile and safe-publication aren’t relevant?

The `call.execute()` line is going to not have an event listener anyway because that isn’t known yet. So the `call.execute()` will throw when it gets to the point that it’s trying to use its event listener.

We should document that it is an error for an EventListener to perform any mutating operation on the Call – heck, we could potentially even enforce that. In either case, it’s definitely not sane for an event listener to execute a call; that’s not its responsibility.

okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
```diff
@@ -0,0 +1,111 @@
...
+    assertSame(call, dnsStart.call);
+    assertEquals("localhost", dnsStart.domainName);
+
+    DnsEnd dnsEnd = listener.expectNextEvent(DnsEnd.class);
+    assertSame(call, dnsEnd.call);
+    assertEquals("localhost", dnsEnd.domainName);
+    assertEquals(1, dnsEnd.inetAddressList.size());
+  }
+
+  static final class DnsStart {
```
swankjesse: dig

okhttp/src/main/java/okhttp3/internal/CallEventListener.java
```diff
@@ -0,0 +1,40 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okhttp3.internal;
+
+import okhttp3.Call;
+import okhttp3.EventListener;
+
+/** An {@link EventListener} scoped to a specific {@link Call}. */
+public final class CallEventListener {
```
swankjesse: I suspect you can probably get away without this. Your fix of moving the code that creates the `EventListener` out of the constructor was sufficient – it’s still not concurrent and so I think we’re good.

adriancole:
Thank you!

okhttp/src/main/java/okhttp3/internal/connection/RouteSelector.java
```diff
@@ -168,7 +172,11 @@ private void resetNextInetSocketAddress(Proxy proxy) throws IOException {
       inetSocketAddresses.add(InetSocketAddress.createUnresolved(socketHost, socketPort));
     } else {
       // Try each address for best behavior in mixed IPv4/IPv6 environments.
+      callEventListener.eventListener().dnsStart(callEventListener.call(), socketHost);
       List<InetAddress> addresses = address.dns().lookup(socketHost);
```
yschimke: Based on your own comments it looks like you just haven't bothered with the failure case yet.  But flagging that the empty check below should be a failure also.

okhttp/src/main/java/okhttp3/EventListener.java
```diff
@@ -38,7 +38,7 @@ public void dnsStart(Call call, String domainName) {
   }
 
   public void dnsEnd(Call call, String domainName, List<InetAddress> inetAddressList,
```
yschimke: inetAddressList is Nullable in failure case.

yschimke:
Wondering how Dns implementation dependant details could get handled e.g.

IPv4 vs IPv6 preferences?
Overrides?
More specific details e.g. dns server, cache status if using Netty Dns.

okhttp/src/main/java/okhttp3/RealCall.java
dave-r12: > This code is single-threaded, so volatile and safe-publication aren’t relevant?

Ahh, yeah my bad. I'm conflating 2 separate issues here.

**Issue 1**: The current API can (theoretically) be used improperly. 
Sounds like we're ok with just documenting this fact. Worst case is to enforce it.

**Issue 2**: Should EventListener have volatile.
Per your note above, I removed `volatile`.

okhttp/src/main/java/okhttp3/internal/CallEventListener.java
dave-r12: Ohh, actually I wasn't thinking of this as a concurrency thing. Right now, the `EventListener` requires us to pass in a `Call` to all the methods. I thought bundling them in a single reference would be an easier thing to work with. As an alternative, I believe we'll need to pass around the `Call` in addition to the `EventListener`. Would you rather pass around 2 references?

okhttp/src/main/java/okhttp3/internal/connection/RouteSelector.java
dave-r12: Yes, I wanted to get buy in on the initial ground work. Note taken.

okhttp/src/main/java/okhttp3/EventListener.java
dave-r12: Right, this will end up as `@Nullable`. I often favor doing the minimum amount of change to get a thing working, just haven't got to the failure case yet.

yschimke: This looks cool BTW, very happy to see this going in.

dave-r12:
I filled this in a bit more: I've added DNS failure scenarios and started to add javadocs to the EventListener class.

okhttp/src/main/java/okhttp3/EventListener.java
```diff
@@ -83,6 +93,16 @@ public void fetchEnd(Call call, Throwable throwable) {
   }
 
   public interface Factory {
+    /**
+     * Creates an instance of the {@link EventListener} for a particular {@link Call}. The returned
```
swankjesse: Nice docs, thanks

okhttp/src/main/java/okhttp3/EventListener.java
```diff
@@ -83,6 +93,16 @@ public void fetchEnd(Call call, Throwable throwable) {
...
+     * {@link EventListener} instance will be used during the lifecycle of the {@code call}.
+     *
+     * <p>This method is invoked after the {@code call} is created. See
+     * {@link OkHttpClient#newCall(Request)}.
+     *
+     * <p><strong>It is an error for implementations to issue any mutating operations on the
```
swankjesse: Yes!

swankjesse:
Great PR. I’m looking forward to using this!

PR #3413, opened 2017-06-17
Disallow Interceptor responses with no bodies.

NightlyNexus:
https://github.com/square/retrofit/issues/2362


okhttp/src/main/java/okhttp3/internal/http/RealInterceptorChain.java
```diff
@@ -102,6 +102,10 @@ public Response proceed(Request request, StreamAllocation streamAllocation, Http
       throw new NullPointerException("interceptor " + interceptor + " returned null");
     }
 
+    if (response.body() == null) {
+      throw new IllegalStateException("response must have a body");
```
JakeWharton: We probably want to include the offending interceptor here like above.

NightlyNexus: oops, 👍

swankjesse:
Nice!

PR #3414, opened 2017-06-17
Make logging in HttpLoggingInterceptor synchronized

amaksoft:
Logs produced by `HttpLoggingInterceptor` are quite unreadable if we make multiple async calls simultaneously. (e.g. using Observable.merge())

Log looks something like this:

OkHTTP: --> GET http://localhost:37535/ http/1.1
OkHTTP: Host: localhost:37535
OkHTTP: Connection: Keep-Alive
OkHTTP: Accept-Encoding: gzip
OkHTTP: User-Agent: okhttp/3.9.0-SNAPSHOT
OkHTTP: --> END GET
OkHTTP: <-- 200 OK http://localhost:37535/ (5ms)
OkHTTP: <-- 200 OK http://localhost:37535/ (1ms)
OkHTTP: Content-Length: 0
OkHTTP: <-- END HTTP (0-byte body)
OkHTTP: Content-Length: 0
OkHTTP: <-- END HTTP (0-byte body)
OkHTTP: <-- 200 OK http://localhost:37535/ (17ms)
OkHTTP: Content-Length: 0
OkHTTP: <-- 200 OK http://localhost:37535/ (23ms)
OkHTTP: <-- END HTTP (0-byte body)
OkHTTP: Content-Length: 0
OkHTTP: <-- END HTTP (0-byte body)

I suggest synchronizing over final `logger` field

Yes, it will slow interceptor chain down a bit. But i think if someone uses this interceptor for debugging purposes, he would prefer consistent logs over performance.

### DIFF
--- a/okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
+++ b/okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
@@ -156,51 +156,54 @@ public final class HttpLoggingInterceptor implements Interceptor {
     if (!logHeaders && hasRequestBody) {
       requestStartMessage += " (" + requestBody.contentLength() + "-byte body)";
     }
-    logger.log(requestStartMessage);
 
-    if (logHeaders) {
-      if (hasRequestBody) {
-        // Request body headers are only present when installed as a network interceptor. Force
-        // them to be included (when available) so there values are known.
-        if (requestBody.contentType() != null) {
-          logger.log("Content-Type: " + requestBody.contentType());
-        }
-        if (requestBody.contentLength() != -1) {
-          logger.log("Content-Length: " + requestBody.contentLength());
-        }
-      }
+    synchronized (logger) {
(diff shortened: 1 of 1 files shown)


swankjesse:
Thanks for the PR, but this is definitely not something we want. If you want to log interleaved requests you should use a strong tool, like [Charles](http://charlesproxy.com/), [Chuck](https://github.com/jgilfelt/chuck), or the HTTP debugger in Android Studio.

artem-zinnatullin:
Hm, intention of the PR is quite good, synchronization is definitely not a great solution though…

@swankjesse maybe reduce `logger.log()` calls to minimum by merging things like headers into one String? That will reduce chance of logs overlapping for concurrent requests.

One of the reasons why we have own HttpLogger is exactly this one — overlapping logs (stripping sensitive data is another one).

In many cases tools like Charles, Chuck or HTTP debugger in AS can't be used, for instance in production environment.

amaksoft:
We could start a new thread and do (synchronized) logging from there.

PR #3419, opened 2017-06-20
Added support for secure connect events.

dave-r12:

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
@@ -19,17 +19,22 @@ import java.io.IOException;
 import java.net.InetAddress;
 import java.net.UnknownHostException;
 import java.util.ArrayDeque;
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.Deque;
 import java.util.List;
 import okhttp3.internal.SingleInetAddressDns;
+import okhttp3.internal.tls.SslClient;
 import okhttp3.mockwebserver.MockResponse;
 import okhttp3.mockwebserver.MockWebServer;
+import okhttp3.mockwebserver.SocketPolicy;
 import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertSame;
 import static org.junit.Assert.assertTrue;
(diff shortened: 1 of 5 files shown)


okhttp/src/main/java/okhttp3/internal/connection/RealConnection.java
```diff
@@ -121,8 +123,8 @@ public static RealConnection testConnection(
     return result;
   }
 
-  public void connect(
-      int connectTimeout, int readTimeout, int writeTimeout, boolean connectionRetryEnabled) {
+  public void connect(int connectTimeout, int readTimeout, int writeTimeout,
+      boolean connectionRetryEnabled, Call call, EventListener eventListener) {
```
dave-r12: Initially I put these as final fields on RealConnection. A few tests failed and it made me realize it's probably not a great idea for a connection to have a strong reference to the call or the event listener.

okhttp/src/main/java/okhttp3/EventListener.java
```diff
@@ -54,11 +54,27 @@ public void dnsEnd(Call call, String domainName, @Nullable List<InetAddress> ine
...
+   */
   public void secureConnectStart(Call call) {
   }
 
-  public void secureConnectEnd(Call call, Handshake handshake,
-      Throwable throwable) {
+  /**
+   * Invoked immediately after a TLS connection was attempted.
+   *
+   * <p>If the {@link Call#request()} does not use TLS, this method will not be invoked.
```
yschimke: This also isn't invoked when a pooled connection is used.

Also if a duplicate connection is created then then it may be closed and the other connection used instead.

So there should probably be an event, whenever an existing connection is used (both cases above).

dave-r12: Oh right, I'll add some tests and docs for those cases.

> So there should probably be an event, whenever an existing connection is used

Unsure on that one.

yschimke: Without this event you don't know the handshake details when reusing a connection, and you don't know if the handshake changes when your new connection is closed as a dupe

okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
```diff
@@ -48,6 +54,37 @@
...
+        .url(server.url("/"))
+        .build());
+    Response response = call.execute();
+    assertEquals(200, response.code());
+    response.body().close();
+
+    List<Class<?>> expectedEvents = Arrays.asList(
+        DnsStart.class, DnsEnd.class,
+        SecureConnectStart.class, SecureConnectEnd.class);
+    assertEquals(expectedEvents, listener.recordedEventTypes());
```
swankjesse: nice

okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
```diff
@@ -110,15 +147,92 @@
...
+    SecureConnectStart secureStart = listener.findNextEvent(SecureConnectStart.class);
+    assertSame(call, secureStart.call);
+
+    SecureConnectEnd secureEnd = listener.findNextEvent(SecureConnectEnd.class);
+    assertSame(call, secureEnd.call);
+    assertNotNull(secureEnd.handshake);
+    assertNull(secureEnd.throwable);
+  }
+
+  @Test public void failedSecureConnect() {
```
swankjesse: dig

okhttp/src/main/java/okhttp3/EventListener.java
swankjesse: Maybe an event like `connectionFound()` that is called when a connection is attached to a call? That could be used on both fresh and recycled connections.

okhttp/src/main/java/okhttp3/internal/connection/RealConnection.java
swankjesse: yeah, I think I see why!

adriancole:
Exciting. Also love the polishing going on in the test class. Great work.

okhttp/src/main/java/okhttp3/EventListener.java
dave-r12: That sounds reasonable, I can draft that.

PR #3425, opened 2017-06-23
Add more detail to dns/secure event docs; confirm it with test cases.

dave-r12:
I'm hopeful this makes the docs and behavior more clear. Let me know if it muddies the waters.

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
@@ -24,6 +24,7 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.Deque;
 import java.util.List;
+import okhttp3.internal.DoubleInetAddressDns;
 import okhttp3.internal.SingleInetAddressDns;
 import okhttp3.internal.tls.SslClient;
 import okhttp3.mockwebserver.MockResponse;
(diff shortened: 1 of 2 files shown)


okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
```diff
@@ -95,17 +98,70 @@
...
+    assertFalse(recordedEvents.contains(DnsEnd.class));
+  }
+
+  @Test public void multipleDnsLookupsForSingleCall() throws IOException {
+    server.enqueue(new MockResponse()
+        .setResponseCode(301)
+        .setHeader("Location", "http://www.fakeurl:" + server.getPort()));
+    server.enqueue(new MockResponse());
+
+    FakeDns dns = new FakeDns();
```
swankjesse: Nice

okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
```diff
@@ -147,9 +203,9 @@
     } catch (IOException expected) {
     }
 
-    listener.findNextEvent(DnsStart.class);
+    assertNotNull(listener.removeUpToEvent(DnsStart.class));
```
swankjesse: should removeUpToEvent() throw if it isn’t found?

swankjesse: ... at least in tests I assume we’ll always know whether an event should be present?

dave-r12: Ahh yeah, that's probably a safer approach.

Issue #3426, opened 2017-06-23
About the use of method.setAccessible() in TrustRootIndex


hrj:
> What kind of issue is this?

I believe this is neither a Question / Bug / Feature request, but a discussion, so I hope this is the right place for it.

When upgrading [our code-base](https://github.com/UprootLabs/gngr) from okhttp-2.x to okhttp-3.x, we are seeing a security exception. It comes from [TrustRootIndex.java#L40](https://github.com/square/okhttp/blob/e791f48af63e3ed4c9a61ffe874173c3327e8ce9/okhttp/src/main/java/okhttp3/internal/tls/TrustRootIndex.java#L40)

The `method.setAccessible(true)` call requires `ReflectPermission("suppressAccessChecks")` which we don't enable for any of the protection domains, because it is a [potently dangerous permission](https://www.securecoding.cert.org/confluence/display/java/ENV03-J.+Do+not+grant+dangerous+combinations+of+permissions).

Hence, I am trying to solve this without increasing permissions. I have a couple of questions and ideas.

* What is the intent of the method `TrustRootIndex.get()`? I understand the concept of certificate chains at a conceptual level, but I am not sure what this code is trying to achieve.
* When the `findTrustAnchorByIssuerAndSignature` method is not found, the exception handler calls `trustManager.getAcceptedIssuers()` instead. So, is the access to hidden field just an optimisation with a less performant fallback?
* If it is merely an optimisation, can you fallback to the less optimal method when access is denied by catching the `AccessControlException`?
* If it is not just an optimisation, but is crucial to the operation of okhttp, then would you consider creating a separate protection domain for reflection code. (Simplest way is to create a separate library that only does one thing: change accessibility of a method).

yschimke:
@swankjesse at a minimum, shouldn't this code move behind the Platform interface and only be used on AndroidPlatform?

swankjesse:
Configure your OkHttpClient instance with this method:
```
OkHttpClient.Builder#sslSocketFactory(SSLSocketFactory, 509TrustManager)
```
That’ll prevent the TrustRootIndex reflection from being used. It’s sadly necessary to avoid a problem with certificate pinning. In particular we can’t correctly validate the pins without the trust roots, and the SSLSocketFactory doesn’t expose an API for that.

hrj:
@swankjesse We had tried using that method but still encountered the same problem. Here's a reproducible test-case: https://github.com/UprootStaging/security-manager-okhttp-test

[...]

PR #3428, opened 2017-06-24
Avoid android specific code for TrustRootIndex

yschimke:
Split out Android TrustRootIndex into AndroidPlatform only code, avoid executing on other platforms


yschimke:
Relates to https://github.com/square/okhttp/issues/3426

swankjesse:
Love this. Thanks!

okhttp/src/main/java/okhttp3/internal/platform/Platform.java
```diff
@@ -228,4 +229,8 @@ private static Platform findPlatform() {
 
     return null;
   }
+
+    public TrustRootIndex buildTrustRootIndex(X509TrustManager trustManager) {
```
PromanSEW: These three lines must be indented by 1 level left

PR #3430, opened 2017-06-24
New method for ReadTimeout on Interceptor & Modified CallTest

ericaschulz:


swankjesse:
Accidental close. Reopened here:
https://github.com/square/okhttp/pull/3432/files

PR #3432, opened 2017-06-27
New method, withReadTimeout(), on Interceptor.Chain

swankjesse:


NightlyNexus:
but, can you add methods to interfaces without breaking binary compatibility?
though, nobody has a reason to implement Chain (could be an abstract class with controlled constructor in 3.0).

dave-r12:
Good point.

okhttp/src/main/java/okhttp3/Interceptor.java
```diff
@@ -36,5 +37,9 @@
      * of network interceptors; for application interceptors this is always null.
      */
     @Nullable Connection connection();
+
+    int readTimeoutMillis();
```
JakeWharton: Are we going to follow this up with the others?

JakeWharton:
Since it's only source-incompatible, I think we can exercise such liberty.

okhttp/src/main/java/okhttp3/Interceptor.java
swankjesse: Yes

swankjesse:
I’m okay breaking compatibility here. I don’t anticipate user code implementing `Chain`, and I think the net benefit of this change far outweighs the compatibility problem.

I think we follow this up with other options that we can promote from OkHttpClient.

PR #3433, opened 2017-06-27
Switch lock with CAS operations

maximn:
Randomly browsing the code and saw a lock that can be switched very easily with CAS operation.


swankjesse:
I think this might be a pessimization. We’re replacing uncontested synchronization with an extra object allocation.

swankjesse:
CAS is good when there’s contention. But here there isn’t any.

PR #3434, opened 2017-06-28
Add support for connect start/end events.

dave-r12:
I've altered the signatures of the connect start and end methods. Notably:

- Changed from `InetAddress` and `port` to `InetSocketAddress`. Was just easier to pull this off the route, but don't have strong feelings on this one.
- Add `Proxy` to connectStart. I thought this might be useful for people using proxies.
- Changed to use `Protocol` on connectEnd. This is already a public type and felt easier to work with.

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
@@ -17,6 +17,8 @@ package okhttp3;
 
 import java.io.IOException;
 import java.net.InetAddress;
+import java.net.InetSocketAddress;
+import java.net.Proxy;
 import java.net.UnknownHostException;
 import java.util.ArrayDeque;
 import java.util.ArrayList;
@@ -25,15 +27,18 @@ import java.util.Collections;
 import java.util.Deque;
 import java.util.List;
 import okhttp3.internal.DoubleInetAddressDns;
+import okhttp3.internal.RecordingOkAuthenticator;
 import okhttp3.internal.SingleInetAddressDns;
 import okhttp3.internal.tls.SslClient;
 import okhttp3.mockwebserver.MockResponse;
 import okhttp3.mockwebserver.MockWebServer;
 import okhttp3.mockwebserver.SocketPolicy;
+import org.junit.After;
 import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
 
+import static okhttp3.TestUtil.defaultClient;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
@@ -45,16 +50,24 @@ import static org.junit.Assert.fail;
 public final class EventListenerTest {
   @Rule public final MockWebServer server = new MockWebServer();
 
-  private OkHttpClient client;
   private final SingleInetAddressDns singleDns = new SingleInetAddressDns();
   private final RecordingEventListener listener = new RecordingEventListener();
   private final SslClient sslClient = SslClient.localhost();
 
-  @Before public void setUp() {
-    client = new OkHttpClient.Builder()
+  private OkHttpClient client;
+  private SocksProxy socksProxy;
+
+  @Before public void setUp() throws IOException {
+    client = defaultClient().newBuilder()
         .dns(singleDns)
         .eventListener(listener)
         .build();
+  }
+
+  @After public void tearDown() throws Exception {
+    if (socksProxy != null) {
+      socksProxy.shutdown();
+    }
   }
 
   @Test public void successfulCallEventSequence() throws IOException {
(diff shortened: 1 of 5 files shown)


okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
```diff
@@ -43,18 +47,27 @@
 import static org.junit.Assert.fail;
 
 public final class EventListenerTest {
-  @Rule public final MockWebServer server = new MockWebServer();
```
JakeWharton: Why this change? Don't see any reason in the code, unless I missed something.

dave-r12: There is currently a dependency when using the `SocksProxy` (I added a test that uses it) to shutdown the MWS first. Alternatively, SocksProxy could be adjusted to behave like MWS by closing all open connections.

okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
```diff
@@ -212,8 +228,178 @@
...
+    socksProxy = new SocksProxy();
+    socksProxy.play();
+    Proxy proxy = socksProxy.proxy();
+
+    client = client.newBuilder()
+        .proxy(proxy)
+        .build();
+
+    Call call = client.newCall(new Request.Builder()
+        .url("http://" + SocksProxy.HOSTNAME_THAT_ONLY_THE_PROXY_KNOWS + ":" + server.getPort())
```
swankjesse: neat

okhttp/src/main/java/okhttp3/internal/connection/RealConnection.java
```diff
@@ -208,18 +217,23 @@ private void connectTunnel(int connectTimeout, int readTimeout, int writeTimeout
       rawSocket = null;
       sink = null;
       source = null;
+      eventListener.connectEnd(call, route.socketAddress(), null, null);
```
swankjesse: this is a failure? I wonder if we can get the exception in there somehow

okhttp/src/main/java/okhttp3/EventListener.java
```diff
@@ -61,7 +63,18 @@ public void dnsEnd(Call call, String domainName, @Nullable List<InetAddress> ine
...
+  /**
+   * Invoked just prior to initiating a socket connection.
+   *
+   * <p>This method will be invoked if no existing connection in the {@link ConnectionPool} can be
+   * reused.
+   *
+   * <p>This can be invoked more than 1 time for a single {@link Call}. For example, if the response
+   * to the {@link Call#request()} is a redirect to a different address, or a connection is retried.
+   *
+   * <p>{@code proxy} will be non-null if an HTTP or SOCKS proxy is used and null otherwise.
```
swankjesse: Should we use `Proxy.NO_PROXY` ? I don’t know

okhttp/src/main/java/okhttp3/internal/connection/RealConnection.java
dave-r12: This is an oddball case: we attempted to create a tunnel, and forced to authenticate. The server closed the connection so we must reconnect with the authentication details. The connection itself is perfectly fine, but we can't derive the protocol at this point.

okhttp/src/main/java/okhttp3/EventListener.java
dave-r12: Sure. Are you reluctant to include the proxy?

swankjesse: I think that having a proxy field is okay. When there is no proxy should we use `Proxy.NO_PROXY` or should we use null?

swankjesse: (elsewhere we use `null` to indicate “we haven’t selected a proxy” and `NO_PROXY` to indicate “we’ve decided to use no proxy”)

okhttp/src/main/java/okhttp3/internal/connection/RealConnection.java
swankjesse: aahh. Makes sense.

okhttp/src/main/java/okhttp3/EventListener.java
dave-r12: Gotcha - `NO_PROXY` makes sense to me. But I'll think about it a bit more.

okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
dave-r12: I've made a small adjustment to `SocksProxy` so that this can continue to be a `@Rule`.

okhttp/src/main/java/okhttp3/EventListener.java
dave-r12: To clarify why I added this in the first place:
I noticed that when using a SOCKS proxy, that simply passing an `InetSocketAddress` wasn't enough because the socket is transparently using the SOCKS proxy. That is why I thought adding the Proxy would be helpful.

I've changed this to be `NO_PROXY`. The intent of this (in my mind) is to communicate the proxy being used to connect. It's not intended to report the outcome of the proxy selection.

swankjesse:
Good stuff.

PR #3451, opened 2017-07-08
Added new method ConnectTimeout() on Interceptor.Chain

ericaschulz:


adriancole:
Looks really good. Design/doc q. Interceptor is a stable type users implement, but chain isnt. It is obvious to me why adding a method on chain is cool, and perhaps adding to the type: dont do anything silly like implementing your own chains as version bumps might break you. That all said, is there a good way to annotate or otherwise types that are exposed to users form consumption but not implementation? Chain seems like one of those types.

swankjesse:
Tough call. 

I like chain as an interface because it’s potentially handy to unit test interceptors. Probably the best forwards-compatible thing would have been to use `default` methods and no-op the new timeout methods. But that’s not a good fit ’cause we don’t have Java 8.

I’m open to suggestions on this.

adriancole:
>
> Tough call.
>
indeed

> I like chain as an interface because it’s potentially handy to unit test
> interceptors. Probably the best forwards-compatible thing would have been
> to use default methods and no-op the new timeout methods. But that’s not
> a good fit ’cause we don’t have Java 8.
>
agreed

> I’m open to suggestions on this.
>
Only thing that comes to mind is either a javadoc which could be
distracting or an annotation which describes the same. Value in the latter
is we could teach tools to ignore it wrt compatibility tests. It is beyond
the scope of this issue for sure, but something been on the mind.

JakeWharton:
I believe the JDK uses Javadoc to indicate interfaces that might get new
methods. There's ones used for annotation processors that come to mind, but
I can't remember exactly which ones from memory.

NightlyNexus:
>is there a good way to annotate or otherwise types that are exposed to users form consumption but not implementation?

An abstract class with a package-private constructor?

>I like chain as an interface because it’s potentially handy to unit test interceptors.

Yeah, for this specific case of the Chain, the controlled implementation might not be perfect.

PR #3456, opened 2017-07-13
Confirm we check all DNS ips against existing connections before creating a new one.

dave-r12:
I accidentally stumbled on this while working on the connection found event. This test was failing because the connection pool had 2 connections instead of 1.

Apparently, I recently added my hostname to `/etc/hosts` as `127.0.0.1`. This test will attempt to connect to `0.0.0.0` and it succeeds. From the comment, I _think_ the intent is to enumerate all the IP's from DNS and look for an existing connection, but I could be wrong.

What do you think?

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/ConnectionCoalescingTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/ConnectionCoalescingTest.java
@@ -17,9 +17,12 @@ package okhttp3;
 
 import java.io.IOException;
 import java.net.InetAddress;
+import java.net.InetSocketAddress;
+import java.net.Proxy;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


okhttp-tests/src/test/java/okhttp3/ConnectionCoalescingTest.java
```diff
@@ -245,6 +248,18 @@
     server.enqueue(new MockResponse().setResponseCode(200));
     server.enqueue(new MockResponse().setResponseCode(200));
 
+    final AtomicInteger count = new AtomicInteger();
+    EventListener eventListener = new EventListener() {
```
swankjesse: awesome

dave-r12:
Alright, I'll see if I can put something together.

dave-r12:
Closed in favor of #3462.

PR #3459, opened 2017-07-14
Include and implement connection found event.

dave-r12:


swankjesse:
That was straightforward!

okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
```diff
@@ -523,6 +525,84 @@
...
+        .setBody("ABC"));
+
+    Call call = client.newCall(new Request.Builder()
+        .url(server.url("/"))
+        .build());
+    Response response = call.execute();
+    assertEquals("ABC", response.body().string());
+
+    listener.removeUpToEvent(ConnectionFound.class);
+    listener.removeUpToEvent(ConnectionFound.class);
```
raskasa: @dave-r12 Just following along from the above 3 unit tests, I see that you haven't ended the test with any assertions on the `Call`.  Was that intentional?

(I follow this codebase for learning purposes, so just looking to understand a little better.)

dave-r12: Yes -- I'm sensitive to confirming the same thing over again in tests, as it adds cost to maintain but doesn't add value. As I was writing this I wanted to confirm the 2 events were emitted, not so much the details of the events.

But it's a valid point, this won't catch a regression where the second event is using the wrong `Call`.

raskasa: Makes sense.  Thanks for clarifying.

okhttp/src/main/java/okhttp3/internal/connection/StreamAllocation.java
```diff
@@ -106,10 +106,16 @@ public HttpCodec newStream(
     boolean connectionRetryEnabled = client.retryOnConnectionFailure();
 
     try {
+      RealConnection existingConnection = connection;
+
       RealConnection resultConnection = findHealthyConnection(connectTimeout, readTimeout,
           writeTimeout, connectionRetryEnabled, doExtensiveHealthChecks);
       HttpCodec resultCodec = resultConnection.newCodec(client, chain, this);
 
+      if (existingConnection != connection) {
```
yschimke: I think there is a bug here when existingConnection is null.  In that case we create a new connection and call connectionFound

dave-r12: That was intentional. It will get called for fresh and recycled connections. It was suggested by @swankjesse unless I misunderstood.

yschimke: For tracing, it's awkward because I could use the connectionFound as a connectionStart/End with zero cost.  So I will just need to filter out the connectionFound when it's the same host.

If you think about a zipkin trace, I want a zero cost dot when its using an existing connection.  https://user-images.githubusercontent.com/231923/28271719-373be6b2-6abe-11e7-921c-c15bde40e127.png

dave-r12: Do you have any reference code? At first glance it seems ok to me. For example, if we were able to use a recycled connection, connectStart/End would not be called. In that case we'd report a zero cost. Otherwise, we'd report each connectStart/End attempt.

yschimke:
Not a perfect example, but shows that you need to look at the inet address in order to cleanly ignore a connection found that you already traced in start/end.

```
public class CEL {
  public static void main(String[] args) throws IOException {
    OkHttpClient client = new OkHttpClient.Builder().eventListener(new EventListener() {
      @Override
      public void connectStart(Call call, InetSocketAddress inetSocketAddress, Proxy proxy) {
        System.out.println("connectStart " + inetSocketAddress);
      }

      @Override public void connectEnd(Call call, InetSocketAddress inetSocketAddress,
          @Nullable Protocol protocol, @Nullable Throwable throwable) {
        System.out.println("connectEnd " + inetSocketAddress + " " + throwable);
      }

      @Override public void connectionFound(Call call, Connection connection) {
        System.out.println("connectionFound " + connection.route().inetSocketAddress);
      }
    }).build();

    client.newCall(new Request.Builder().url("https://google.com/robots.txt").build())
        .execute().body().string();

    System.out.println("---");

    client.newCall(new Request.Builder().url("http://google.com/robots.txt").build())
        .execute().body().string();

    System.out.println("---");

    client.newCall(new Request.Builder().url("https://google.com/robots.txt").build())
        .execute().body().string();
  }
}
```

```
connectStart google.com/216.58.195.78:443
connectEnd google.com/216.58.195.78:443 null
connectionFound google.com/216.58.195.78:443
connectStart www.google.com/172.217.5.100:443
connectEnd www.google.com/172.217.5.100:443 null
connectionFound www.google.com/172.217.5.100:443
---
connectStart google.com/216.58.195.78:80
connectEnd google.com/216.58.195.78:80 null
connectionFound google.com/216.58.195.78:80
connectionFound www.google.com/172.217.5.100:443
---
connectionFound google.com/216.58.195.78:443
connectionFound www.google.com/172.217.5.100:443
```

okhttp/src/main/java/okhttp3/internal/connection/StreamAllocation.java
yschimke: Not a perfect example, but shows that you need to look at the inet address in order to cleanly ignore a connection found that you already traced in start/end.

```
public class CEL {
  public static void main(String[] args) throws IOException {
    OkHttpClient client = new OkHttpClient.Builder().eventListener(new EventListener() {
      @Override
      public void connectStart(Call call, InetSocketAddress inetSocketAddress, Proxy proxy) {
        System.out.println("connectStart " + inetSocketAddress);
      }

      @Override public void connectEnd(Call call, InetSocketAddress inetSocketAddress,
          @Nullable Protocol protocol, @Nullable Throwable throwable) {
        System.out.println("connectEnd " + inetSocketAddress + " " + throwable);
      }

      @Override public void connectionFound(Call call, Connection connection) {
        System.out.println("connectionFound " + connection.route().inetSocketAddress);
      }
    }).build();

    client.newCall(new Request.Builder().url("https://google.com/robots.txt").build())
        .execute().body().string();

    System.out.println("---");

    client.newCall(new Request.Builder().url("http://google.com/robots.txt").build())
        .execute().body().string();

    System.out.println("---");

    client.newCall(new Request.Builder().url("https://google.com/robots.txt").build())
        .execute().body().string();
  }
}
```

```
connectStart google.com/216.58.195.78:443
connectEnd google.com/216.58.195.78:443 null
connectionFound google.com/216.58.195.78:443
connectStart www.google.com/172.217.5.100:443
connectEnd www.google.com/172.217.5.100:443 null
connectionFound www.google.com/172.217.5.100:443
---
connectStart google.com/216.58.195.78:80
connectEnd google.com/216.58.195.78:80 null
connectionFound google.com/216.58.195.78:80
connectionFound www.google.com/172.217.5.100:443
---
connectionFound google.com/216.58.195.78:443
connectionFound www.google.com/172.217.5.100:443
```

yschimke: I'm happy to leave this at it's awkward but correct, so deal with it.  But just giving the feedback that it makes avoiding logging ONLY one of either new connection or reused awkward.

dave-r12: Gotcha -- I'm wondering, should we start building a nontrivial sample of an event listener in the samples directory? Maybe that can help us sort some of this out.

yschimke: Do you want to start with a zipkin tracer? https://github.com/yschimke/oksocial/pull/255

:)

PR #3462, opened 2017-07-16
Consider all routes when looking for candidate coalesced connections.

dave-r12:
Previously we'd only look at one route at a time. If DNS returned
results in random order it would lower our chances of finding a
coalesced connection. Now we consider all routes at once which hopefully
improves our chances.

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/internal/connection/RouteSelectorTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/internal/connection/RouteSelectorTest.java
@@ -47,6 +47,7 @@ import org.junit.Test;
 import static java.net.Proxy.NO_PROXY;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertSame;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
@@ -90,8 +91,15 @@ public final class RouteSelectorTest {
 
     assertTrue(routeSelector.hasNext());
     dns.set(uriHost, dns.allocate(1));
-    assertRoute(routeSelector.next(), address, NO_PROXY, dns.lookup(uriHost, 0), uriPort);
+    RouteSelector.Selection selection = routeSelector.next();
+    assertRoute(selection.next(), address, NO_PROXY, dns.lookup(uriHost, 0), uriPort);
     dns.assertRequests(uriHost);
+    assertFalse(selection.hasNext());
+    try {
+      selection.next();
+      fail();
+    } catch (NoSuchElementException expected) {
+    }
 
     assertFalse(routeSelector.hasNext());
     try {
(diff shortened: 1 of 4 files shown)


swankjesse:
(Sorry for the delay on this!)

okhttp/src/main/java/okhttp3/internal/connection/RouteSelector.java
```diff
@@ -69,35 +64,45 @@ public RouteSelector(Address address, RouteDatabase routeDatabase, Call call,
...
+      throw new NoSuchElementException();
+    }
+
+    // Compute the next set of routes to attempt.
+    List<Route> routes = new ArrayList<>();
+    while (hasNextProxy()) {
+      // Postponed routes are always tried last. For example, if we have 2 proxies and all the
+      // routes for proxy1 should be postponed, we'll move to proxy2. Only after we've exhausted
+      // all the good routes will we attempt the postponed routes.
+      Proxy lastProxy = nextProxy();
```
swankjesse: nit: rename to `proxy`? Weird to have a line where we assign `nextProxy()` to a field called `lastProxy()` !

okhttp/src/main/java/okhttp3/internal/connection/RouteSelector.java
```diff
@@ -220,27 +223,28 @@ static String getHostString(InetSocketAddress socketAddress) {
     return address.getHostAddress();
   }
 
-  /** Returns true if there's another socket address to try. */
-  private boolean hasNextInetSocketAddress() {
-    return nextInetSocketAddressIndex < inetSocketAddresses.size();
-  }
+  /** A set of selected Routes. */
```
swankjesse: Would it be sane to just use `Iterator<Route>` for this type?

swankjesse: answering my own question: no! Because `getAll()` needs to work

okhttp/src/main/java/okhttp3/internal/connection/StreamAllocation.java
```diff
@@ -181,21 +183,33 @@ private RealConnection findConnection(int connectTimeout, int readTimeout, int w
...
     synchronized (connectionPool) {
       if (canceled) throw new IOException("Canceled");
 
-      // Now that we have an IP address, make another attempt at getting a connection from the pool.
-      // This could match due to connection coalescing.
-      Internal.instance.get(connectionPool, address, this, selectedRoute);
-      if (connection != null) {
-        route = selectedRoute;
-        return connection;
+      if (newRouteSelection) {
```
swankjesse: This is great!

okhttp-tests/src/test/java/okhttp3/internal/connection/RouteSelectorTest.java
```diff
@@ -311,14 +349,70 @@
...
-    List<Route> routesWithFailedRoute = new ArrayList<>();
-    while (routeSelector.hasNext()) {
-      routesWithFailedRoute.add(routeSelector.next());
-    }
+    // The first selection prioritizes the non-failed routes.
+    RouteSelector.Selection selection2 = routeSelector.next();
+    assertEquals(regularRoutes.get(1), selection2.next());
+    assertFalse(selection2.hasNext());
+
+    // The second selection will contain all failed routes.
```
swankjesse: Dig

swankjesse:
I really, really like the `Selection` abstraction. It really makes the coalescing feel like a natural fit. Yay!

okhttp/src/main/java/okhttp3/internal/connection/RouteSelector.java
dave-r12: Will do.

PR #3475, opened 2017-07-23
Fill out documentation for EventListener

yschimke:
Fill out documentation clearing up the semantics of the EventListener API

Things to call out
- I've defined fetchStart/End to be from enqueue/execute to caller closing the Response body.
- There is no easy hook in request/response(Header|Body)(Start|End) except implicit ordering.  All useful classes e.g. StreamAllocation are internal.
- There is no event to signify when OkHttp really starts processing, or has handed off the response to the caller.

### DIFF
--- a/okhttp/src/main/java/okhttp3/EventListener.java
+++ b/okhttp/src/main/java/okhttp3/EventListener.java
@@ -136,14 +136,25 @@ public abstract class EventListener {
    * <p>This can be invoked more than 1 time for a single {@link Call}. For example, if the response
    * to the {@link Call#request()} is a redirect to a different address.
    */
-  public void connectionFound(Call call, Connection connection) {
+  public void connectionAcquired(Call call, Connection connection) {
+  }
+
+  /**
+   * Invoked after a connection has been released for the {@code call}.
+   *
+   * <p>This method is always invoked after {@link #connectionAcquired(Call, Connection)}.
+   *
+   * <p>This can be invoked more than 1 time for a single {@link Call}. For example, if the response
+   * to the {@link Call#request()} is a redirect to a different address.
+   */
+  public void connectionReleased(Call call, Connection connection) {
   }
 
   /**
    * Invoked just prior to sending request headers.
    *
    * <p>The connection is implicit, and will generally relate to the last
-   * {@link #connectionFound(Call, Connection)} event.
+   * {@link #connectionAcquired(Call, Connection)} event.
    *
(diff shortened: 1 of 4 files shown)


yschimke:
cc @dave-r12

okhttp/src/main/java/okhttp3/EventListener.java
```diff
@@ -128,33 +136,130 @@ public void connectEnd(Call call, InetSocketAddress inetSocketAddress,
...
 
+  /**
+   * Invoked just prior to sending a request body.  Will only be invoked for request allowing and
+   * having a request body to send.
+   *
+   * <p>The connection is implicit, and will generally relate to the last
+   * {@link #connectionAcquired(Call, Connection)} event.
+   *
+   * <p>This can be invoked more than 1 time for a single {@link Call}. For example, if the response
+   * to the {@link Call#request()} is a redirect to a different address.
```
swankjesse: Nit: this is a bit misleading. only `PROPFIND` redirects retain the body. Also HTTP 408s and when we recover from failures.

yschimke: "This will usually be invoked only 1 time for a single {@link Call}, exceptions are a limited set of cases including failure recovery."

okhttp/src/main/java/okhttp3/EventListener.java
```diff
@@ -128,33 +136,130 @@ public void connectEnd(Call call, InetSocketAddress inetSocketAddress,
...
+   */
   public void requestHeadersStart(Call call) {
   }
 
+  /**
+   * Invoked immediately after sending a request body.
```
dave-r12: Should this be after sending the request _headers_?

okhttp/src/main/java/okhttp3/EventListener.java
```diff
@@ -128,33 +136,130 @@ public void connectEnd(Call call, InetSocketAddress inetSocketAddress,
...
+   * {@link #connectionAcquired(Call, Connection)} event.
+   *
+   * <p>This will usually be invoked only 1 time for a single {@link Call},
+   * exceptions are a limited set of cases including failure recovery.
+   */
   public void responseBodyStart(Call call) {
   }
 
+  /**
+   * Invoked immediately after receiving a request body and completing reading it.
```
yschimke: missed this one also

yschimke:
n.b. the addition of proxy on connectEnd to allow matching cleanly against start

swankjesse:
Nice.

PR #3476, opened 2017-07-24
Complete EventListener

yschimke:
Call all methods in EventListener.
Adds proxy to connectEnd to allow matching to start event unambiguously.
Test EventListener fails fast for unmatched end event.
Make RecordingEventListener a top level class.

n.b. Ordering of ResponseBodyEvent needs to be addressed in another PR.  Currently it can be delayed by client not consuming output, even if all bytes have been read off the wire.

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
@@ -80,10 +80,11 @@ public final class EventListenerTest {
     assertEquals(200, response.code());
     response.body().close();
 
-    List<Class<?>> expectedEvents = Arrays.asList(
-        DnsStart.class, DnsEnd.class,
-        ConnectStart.class, ConnectEnd.class,
-        ConnectionFound.class);
+    List<Class<? extends CallEvent>> expectedEvents = Arrays.asList(FetchStart.class,
+        DnsStart.class, DnsEnd.class, ConnectStart.class, ConnectEnd.class,
+        ConnectionAcquired.class, RequestHeadersStart.class, RequestHeadersEnd.class,
+        ResponseHeadersStart.class, ResponseHeadersEnd.class, ResponseBodyStart.class,
+        ResponseBodyEnd.class, FetchEnd.class);
     assertEquals(expectedEvents, listener.recordedEventTypes());
   }
 
@@ -98,11 +99,12 @@ public final class EventListenerTest {
     assertEquals(200, response.code());
     response.body().close();
 
-    List<Class<?>> expectedEvents = Arrays.asList(
-        DnsStart.class, DnsEnd.class,
-        ConnectStart.class, SecureConnectStart.class,
+    List<Class<? extends CallEvent>> expectedEvents = Arrays.asList(FetchStart.class,
+        DnsStart.class, DnsEnd.class, ConnectStart.class, SecureConnectStart.class,
         SecureConnectEnd.class, ConnectEnd.class,
-        ConnectionFound.class);
+        ConnectionAcquired.class, RequestHeadersStart.class, RequestHeadersEnd.class,
+        ResponseHeadersStart.class, ResponseHeadersEnd.class, ResponseBodyStart.class,
+        ResponseBodyEnd.class, FetchEnd.class);
     assertEquals(expectedEvents, listener.recordedEventTypes());
   }
 
@@ -535,7 +537,7 @@ public final class EventListenerTest {
     assertEquals(200, response.code());
     response.body().close();
 
-    ConnectionFound connectionFound = listener.removeUpToEvent(ConnectionFound.class);
+    ConnectionAcquired connectionFound = listener.removeUpToEvent(ConnectionAcquired.class);
     assertSame(call, connectionFound.call);
     assertNotNull(connectionFound.connection);
   }
@@ -553,10 +555,10 @@ public final class EventListenerTest {
     Response response = call.execute();
     assertEquals("ABC", response.body().string());
 
-    listener.removeUpToEvent(ConnectionFound.class);
+    listener.removeUpToEvent(ConnectionAcquired.class);
 
     List<Class<?>> remainingEvents = listener.recordedEventTypes();
-    assertFalse(remainingEvents.contains(ConnectionFound.class));
+    assertFalse(remainingEvents.contains(ConnectionAcquired.class));
   }
 
   @Test public void pooledConnectionFound() throws IOException {
@@ -571,7 +573,7 @@ public final class EventListenerTest {
     assertEquals(200, response1.code());
     response1.body().close();
 
-    ConnectionFound connectionFound1 = listener.removeUpToEvent(ConnectionFound.class);
+    ConnectionAcquired connectionFound1 = listener.removeUpToEvent(ConnectionAcquired.class);
     listener.clearAllEvents();
 
     Call call2 = client.newCall(new Request.Builder()
@@ -581,7 +583,7 @@ public final class EventListenerTest {
     assertEquals(200, response2.code());
     response2.body().close();
 
-    ConnectionFound connectionFound2 = listener.removeUpToEvent(ConnectionFound.class);
+    ConnectionAcquired connectionFound2 = listener.removeUpToEvent(ConnectionAcquired.class);
     assertSame(connectionFound1.connection, connectionFound2.connection);
   }
 
@@ -599,8 +601,8 @@ public final class EventListenerTest {
     Response response = call.execute();
     assertEquals("ABC", response.body().string());
 
-    listener.removeUpToEvent(ConnectionFound.class);
-    listener.removeUpToEvent(ConnectionFound.class);
+    listener.removeUpToEvent(ConnectionAcquired.class);
+    listener.removeUpToEvent(ConnectionAcquired.class);
   }
 
   private void enableTlsWithTunnel(boolean tunnelProxy) {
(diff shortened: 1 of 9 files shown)


yschimke:
I'll try to finish off the TODOs this weekend, I'd like to get it in before the next release so it isn't a partial feature.  If I'm blocking anyone during the work week, please steal off me.

okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
```diff
@@ -611,88 +613,175 @@ private void enableTlsWithTunnel(boolean tunnelProxy) {
     server.useHttps(sslClient.socketFactory, tunnelProxy);
   }
 
-  static final class DnsStart {
+  static class CallEvent {
```
dave-r12: Hmm, not sure what I think of this. On the one hand it cuts down on the boilerplate and redundancy, but on the other hand we lose some type information on the params.

dave-r12:
Re: connection release event -- I'm unsure on the intended use case there.

okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
yschimke: There is no loss of type information as the fields are still toplevel.

The intent here was to implement as many of the asserts of matching pairs as part of the listener impl.  I'd like connectionReleased to fail fast if a matching connectionAcquired isn't already in place.  My goal was to incrementally implement the rules of the docs to make each new (edge case) unit test trivial to read and understand while having strong asserts.

yschimke:
@dave-r12 Connection start/end pairs are great to show the cost of initial connections to the servers.  But in HTTP/1.1 especially there is a window where this call has exclusive ownership of a connection, so it's a very top level thing even when we redirect http->https and reuse existing connections.

yschimke:
With connections shown.  n.b. I'll probably default to less detail in zipkin in my external PR, so this is just for demonstration.

image

image

yschimke:
@dave-r12 @swankjesse I'm uncertain where to record responseBodyEnd and fetchEnd.  I suspect in Http1Codec and Http2Codec as they are internal classes, so API changes won't be disruptive like ResponseBody.

But it seems awkward at the moment, particularly with the number of special cases in Http1Codec e.g. chunked encoding and how to handle exceptions.  

The current PR will sometimes complete request body too early but is calling it reliably.

okhttp/src/main/java/okhttp3/internal/http/CallServerInterceptor.java
```diff
@@ -82,26 +107,34 @@ public CallServerInterceptor(boolean forWebSocket) {
...
           .build();
     } else {
-      response = response.newBuilder()
-          .body(httpCodec.openResponseBody(response))
-          .build();
-    }
+      realChain.eventListener().responseBodyStart(realChain.call());
+      try {
+        response = response.newBuilder()
+            .body(httpCodec.openResponseBody(response))
```
swankjesse: I think we wanna move responseBodyStart() and responseBodyEnd() calls into the codec. Main reason is that if there’s exceptions reading the response body these will be the best places to signal the failure to responseBodyEnd.

okhttp-tests/src/test/java/okhttp3/RecordingEventListener.java
```diff
@@ -0,0 +1,374 @@
...
+    }
+    return eventTypes;
+  }
+
+  void clearAllEvents() {
+    eventSequence.clear();
+  }
+
+  private void logEvent(CallEvent e) {
+    CallEvent startEvent = e.closes();
```
swankjesse: This is neat

okhttp/src/main/java/okhttp3/internal/http/CallServerInterceptor.java
swankjesse: ALSO – I’d like the option of having request and response body events include the number of application-layer bytes in the message. Potentially useful for things like streaming JSON with gzip, where the response body size is pretty difficult to observe right now.

yschimke: Is it worth rethinking the EventListener and introduce parameter objects to allow additional fields in future without breaking compatibility?

Or marking the EventListener API as clearly experimental and likely to change?


```
  public void connectEnd(Call call, @Nullable ConnectEndDetails details, @Nullable Throwable throwable) {
  }

  public void responseBodyEnd(Call call, @Nullable ResponseBodyDetails details, @Nullable Throwable throwable) {
  }
```

yschimke: n.b. I moved the calls into the Codec, if it doesn't look right I can either

1) land as is and fix forward
2) revert to where I put it first time
3) fix after discussion

yschimke:
Other tests

$ ./oksocial -z https://httpbin.org/stream-bytes/1073741824 > /dev/null

image

$ ./oksocial -z http://httpbin.org/redirect-to?url=http%3A%2F%2Fgoogle.com%2Frobots.txt

image

$ ./oksocial -z https://httpbin.org/status/500
HTTP/1.1 500 INTERNAL SERVER ERROR

okhttp/src/main/java/okhttp3/internal/http/CallServerInterceptor.java
```diff
@@ -92,16 +117,16 @@ public CallServerInterceptor(boolean forWebSocket) {
       response = response.newBuilder()
           .body(httpCodec.openResponseBody(response))
           .build();
-    }
 
-    if ("close".equalsIgnoreCase(response.request().header("Connection"))
```
yschimke: fixing this

yschimke:
Suggest landing this as discussed and either marking as  experimental for next release or adopting a more forgiving pattern like parameter objects.

yschimke:
Anything blocking this? Event code is in a weird state if you have to release without this anyway.

dave-r12:
@yschimke the incomplete `EventListener` is my fault, although it was intentional. I was progressing through each event type so the CLs were smaller (I prefer small CLs to large CLs) and to bring focus on scrutinizing each event so everyone would be happy with the outcome. It wasn't clear to me how to make forward progress with this PR open. I didn't want to duplicate effort.

I can help get this completed.

yschimke:
@dave-r12 feel free to take over this PR, or drop it and start from scratch.  I was trying to use the events and getting blocked on the missing ones.

swankjesse:
Thanks @yschimke & @dave-r12. Sorry for the slow responses lately. Things have been pretty busy lately. I’ll look at this and try to get you both unblocked!

okhttp/src/main/java/okhttp3/internal/connection/RealConnection.java
```diff
@@ -267,11 +267,11 @@ private void establishProtocol(ConnectionSpecSelector connectionSpecSelector, Ca
     eventListener.secureConnectStart(call);
     try {
       connectTls(connectionSpecSelector);
+      eventListener.secureConnectEnd(call, handshake, null);
     } catch (Exception e) {
       eventListener.secureConnectEnd(call, null, e);
       throw e;
     }
-    eventListener.secureConnectEnd(call, handshake, null);
```
swankjesse: why move this? Only relevant if secureConnectEnd() throws

okhttp/src/main/java/okhttp3/internal/connection/StreamAllocation.java
```diff
@@ -301,6 +301,9 @@ private Socket deallocate(boolean noNewStreams, boolean released, boolean stream
       this.codec = null;
     }
     if (released) {
+      if (connection != null) {
+        eventListener.connectionReleased(call, connection);
```
swankjesse: I’m a little anxious about calling user code while holding the single global connection pool lock. Probably safer to call this without the lock

okhttp/src/main/java/okhttp3/internal/http/CallServerInterceptor.java
```diff
@@ -43,36 +43,61 @@ public CallServerInterceptor(boolean forWebSocket) {
...
-        responseBuilder = httpCodec.readResponseHeaders(true);
-      }
+      realChain.eventListener().requestBodyStart(realChain.call());
+      try {
+        // If there's a "Expect: 100-continue" header on the request, wait for a "HTTP/1.1 100
+        // Continue" response before transmitting the request body. If we don't get that, return
+        // what we did get (such as a 4xx response) without ever transmitting the request body.
+        if ("100-continue".equalsIgnoreCase(request.header("Expect"))) {
+          httpCodec.flushRequest();
+          // TODO event listener
```
swankjesse: ooh, tricky case

okhttp/src/main/java/okhttp3/internal/http1/Http1Codec.java
```diff
@@ -129,10 +130,36 @@ public Http1Codec(OkHttpClient client, StreamAllocation streamAllocation, Buffer
   }
 
   @Override public ResponseBody openResponseBody(Response response) throws IOException {
-    Source source = getTransferStream(response);
+    streamAllocation.eventListener.responseBodyStart(streamAllocation.call);
+    Source source = new EventListenerSource(getTransferStream(response));
     return new RealResponseBody(response.headers(), Okio.buffer(source));
   }
 
+  class EventListenerSource extends ForwardingSource {
```
swankjesse: I don’t think we want this as an extra object

swankjesse: I think it’s better without the extra abstraction

swankjesse:
I merged! None of my feedback is urgent; just style stuff.

okhttp/src/main/java/okhttp3/internal/http1/Http1Codec.java
swankjesse: https://github.com/square/okhttp/pull/3502

PR #3480, opened 2017-07-26
logging via SLF4J

takashima0411:
java.util.loggin is hard to use.
I want use SLF4J.


JakeWharton:
Any logger that you're using will have a j.u.logging bridge. I'm not sure
we should bother with the dependency here since it isn't even used in the
main client (and we generally have a high bar for dependencies).

PR #3489, opened 2017-07-31
JDK 8u141 & 8u144 Compatibility

packetpilot:
Per [this comment](https://github.com/jetty-project/jetty-alpn/commit/d58174d8924eda832f22606d68abaf879f4553fa#commitcomment-23304619) & reply from @sbordet, pom.xml was modified for jdk8u141, and tests succeeded.

An update today to 8u144 and subsequent test with u144's profile also successfully tested.

My completed CLA form was submitted about five minutes ago.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -552,5 +552,23 @@
         <alpn.jdk8.version>8.1.11.v20170118</alpn.jdk8.version>
       </properties>
     </profile>
+    <profile>
+      <id>alpn-when-jdk8_141</id>
+      <activation>
+        <jdk>1.8.0_141</jdk>
+      </activation>
+      <properties>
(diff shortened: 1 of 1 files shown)


swankjesse:
Thanks!

sbordet:
@packetpilot it is not correct to assume that because a version of Jetty's ALPN works for a specific JDK version, then it also works for a subsequent one, even if testing is successful.

This is because a newer JDK version may have fixed a security issue and using an older Jetty ALPN may work but not include the security fix present in the JDK.

You should always wait for an official statement from the Jetty Project (follow @JettyProject on Twitter)

Now, for this particular case, it is correct that Jetty's ALPN valid for 8u131 is also valid for 8u141 and 8u144.

Thanks !

Issue #3485, opened 2017-07-31
Proposal: Google Dynamic Security Provider Support


naturalwarren:
Currently we have a small fork of OkHttp that checks to see if [DSP](https://developer.android.com/training/articles/security-gms-provider.html) is present on Android devices. The implementation looks to see if the GMS `java.security.Provider` is installed when `AndroidPlatform.buildIfSupported()` is called:

```java
// Attempt to find Google Play Services Dynamic Security Provider.
Provider provider = Security.getProvider("GmsCore_OpenSSL");
if (provider != null) {
  getAlpnSelectedProtocol = GET_ALPN_SELECTED_PROTOCOL;
  setAlpnProtocols = SET_ALPN_PROTOCOLS;
} else {
  // Attempt to find Android 5.0+ APIs.
  ...
}
```

This would be pretty easy to add directly in a PR and would allow OkHttp to provide ALPN support for pre 5.x devices and facilitate HTTP/2 use on older devices. 

Thoughts?

swankjesse:
Send a pull request?

naturalwarren:
Will do, thanks Jesse.

naturalwarren:
Fixed in #3490.

yschimke:
Did we break this recently?

yschimke:
Supported cleanly via https://github.com/square/okhttp/pull/5243

Supports default Android, BYO Conscrypt and GMS

PR #3490, opened 2017-07-31
Check for DSP to enable ALPN.

naturalwarren:
Checks to see if DSP is present. If it is, ALPN can be used on Android 4.x devices which facilitates HTTP/2 use. 

Addresses #3485

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/platform/AndroidPlatform.java
+++ b/okhttp/src/main/java/okhttp3/internal/platform/AndroidPlatform.java
@@ -22,7 +22,6 @@ import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.net.InetSocketAddress;
 import java.net.Socket;
-import java.security.Provider;
 import java.security.Security;
 import java.security.cert.Certificate;
 import java.security.cert.TrustAnchor;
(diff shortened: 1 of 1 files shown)


okhttp/src/main/java/okhttp3/internal/platform/AndroidPlatform.java
```diff
@@ -40,6 +42,11 @@
 class AndroidPlatform extends Platform {
   private static final int MAX_LOG_LENGTH = 4000;
 
+  private static final OptionalMethod<Socket> getAlpnMethod
```
artem-zinnatullin: Is there a real need to hold these two guys as static final? 

Platform is usually built once, look like you can move this into `buildIfSupported()` as local variables.

naturalwarren: No there isn't. That's a good point, I'll move them down. Thanks Artem.

okhttp/src/main/java/okhttp3/internal/platform/AndroidPlatform.java
```diff
@@ -216,12 +218,24 @@ public static Platform buildIfSupported() {
       OptionalMethod<Socket> getAlpnSelectedProtocol = null;
       OptionalMethod<Socket> setAlpnProtocols = null;
 
-      // Attempt to find Android 5.0+ APIs.
-      try {
-        Class.forName("android.net.Network"); // Arbitrary class added in Android 5.0.
-        getAlpnSelectedProtocol = new OptionalMethod<>(byte[].class, "getAlpnSelectedProtocol");
-        setAlpnProtocols = new OptionalMethod<>(null, "setAlpnProtocols", byte[].class);
-      } catch (ClassNotFoundException ignored) {
+      final OptionalMethod<Socket> getAlpnMethod
```
yschimke: I think this method could be a lot cleaner if it was split up into two methods.

boolean isDefinitelyNewerAndroid() {
  return isGMS() || isAndroid5Plus(); 
}

As a desktop/server developer it isn't clear why this logic works without comments.

my 2c.

naturalwarren: I agree. Maybe name the method `supportsALPN()` instead of `isNewerAndroid()` though?

@swankjesse Any preference here?

naturalwarren: I like this. How about the method name `supportsALPN()` instead of `isNewerAndroid()` since that's what we're checking for.

Happy to make this change. Thanks @yschimke.

JakeWharton: Proper casing of that would be `supportAlpn`

naturalwarren: 0deb4c8

swankjesse:
Thanks!

PR #3502, opened 2017-08-05
Use endOfInput() to trigger the end of the response body.

swankjesse:
This removes the decorator and uses the existing superclass instead.


yschimke:
nice!

swankjesse:
Ugh, looks like this is racy. Won’t be after I rebase on this:
https://github.com/square/okhttp/pull/3503

PR #3503, opened 2017-08-05
Don't call EventListener methods while holding locks.

swankjesse:


swankjesse:
Ugh, this one’s flaky for a different reason.

yschimke:
@swankjesse if it's causing issues, I'm happy for my PR to be reverted and this to be completed in due course.  I was asking because it felt like a partial implementation was an awkward state, not to ram it in before ready.

But I'm also happy to see it land and start using it :)

swankjesse:
Yeah, let’s roll forward. I really like how this stuff looks.

okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
```diff
@@ -84,11 +87,16 @@
...
+    List<String> expectedEvents = Arrays.asList(
+        "FetchStart",
+          "DnsStart", "DnsEnd",
+          "ConnectionAcquired",
+            "ConnectStart", "ConnectEnd",
+            "RequestHeadersStart", "RequestHeadersEnd",
+            "ResponseHeadersStart", "ResponseHeadersEnd",
+            "ResponseBodyStart", "ResponseBodyEnd",
+          "ConnectionReleased",
+        "FetchEnd");
```
swankjesse: gratuitous formatting

swankjesse:
Alright, I figured out why it was passing locally and not on Travis. The response body is empty, and with the Http1 codec that response is terminated when it’s opened and the connection continues to be held. With the Http2 codec that response isn’t terminated until it’s consumed.

We need a follow-up PR to change the RetryAndFollowUpInterceptor to be in charge of ending the fetch. We should consider the fetch complete when the last byte of the last response body is consumed.

yschimke:
A big +1 to completing the fetch after the client reads the body.

Are you making this PR, if not let me know if you want me to make the change.

swankjesse:
Please do. I think we probably want to take advantage of the fact that we call `streamAllocation.release();` when we know there’s no further follow-ups coming. Maybe when we dispose of the connection after it’s been released we know the fetch is complete also.

okhttp-tests/src/test/java/okhttp3/RecordingEventListener.java
```diff
@@ -25,11 +25,19 @@
...
+import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
 
 final class RecordingEventListener extends EventListener {
   final Deque<CallEvent> eventSequence = new ArrayDeque<>();
 
+  final List<Object> forbiddenLocks = new ArrayList<>();
+
+  /** Confirm that the thread does not hold a lock on {@code lock} during the callback. */
+  public void forbidLock(Object lock) {
```
dave-r12: Cool 👍

okhttp/src/main/java/okhttp3/internal/connection/StreamAllocation.java
```diff
@@ -202,22 +205,34 @@ private RealConnection findConnection(int connectTimeout, int readTimeout, int w
...
+    }
 
-      // Create a connection and assign it to this allocation immediately. This makes it possible
-      // for an asynchronous cancel() to interrupt the handshake we're about to do.
-      route = selectedRoute;
-      refusedStreamCount = 0;
-      result = new RealConnection(connectionPool, selectedRoute);
-      acquire(result);
+    // We have a connection. Either a connected one from the pool, or one we need to connect.
+    eventListener.connectionAcquired(call, result);
```
dave-r12: Just noting in the case of a `Connection` we need to connect, the socket on the connection will be null.

okhttp/src/main/java/okhttp3/internal/http2/Http2Codec.java
```diff
@@ -195,28 +194,18 @@ public Http2Codec(OkHttpClient client, Interceptor.Chain chain, StreamAllocation
...
     StreamFinishingSource(Source delegate) {
       super(delegate);
     }
 
-    @Override public long read(Buffer sink, long byteCount) throws IOException {
-      try {
-        return super.read(sink, byteCount);
-      } catch (IOException ioe) {
-        completed = true;
-        streamAllocation.eventListener.responseBodyEnd(streamAllocation.call, ioe);
```
dave-r12: Did you mean to get rid of this? Seems like we won't call the `responseBodyEnd` on an IOException for HTTP/2.

dave-r12:
This is awesome. I found the scenarios around releasing a connection tricky to reason about but they all seem right. Do you think it'd be worth the effort to create test cases for the various cases?

okhttp/src/main/java/okhttp3/internal/connection/StreamAllocation.java
swankjesse: Yeah, good call.

okhttp/src/main/java/okhttp3/internal/http2/Http2Codec.java
swankjesse: I think this is a regression. I’ll do a follow-up PR with all of the failure cases, and that should shake this out.

swankjesse: https://github.com/square/okhttp/pull/3508

PR #3504, opened 2017-08-06
Handle h2 vs http/1.1 event ordering differences

yschimke:


yschimke:

okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
```diff
@@ -103,12 +104,47 @@
     assertEquals(200, response.code());
     response.body().close();
 
+    // ordering of response body end currently differs
+    List<String> expectedEvents;
+    if (response.protocol == Protocol.HTTP_2) {
```
dave-r12: Should we break this out into distinct test cases? We could disable HTTP/2 on the client in one case, and enforce it on the other.

okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
```diff
@@ -96,27 +98,58 @@
...
-    enableTlsWithTunnel(false);
-    server.enqueue(new MockResponse()
-        .setBody("abc"));
-
+  private void assertSuccessfulEventOrder() throws IOException {
     Call call = client.newCall(new Request.Builder()
-        .url(server.url("/"))
-        .build());
+            .url(server.url("/"))
+            .build());
```
swankjesse: nit: we do +4 on wrapped lines

swankjesse: If you’re using IntelliJ, it’s
https://github.com/square/java-code-styles/

yschimke: @swankjesse my bad, fresh Intellij install.  Is there an automated way to reformat?

PR #3505, opened 2017-08-06
Event docs

yschimke:


yschimke:

okhttp/src/main/java/okhttp3/EventListener.java
```diff
@@ -21,6 +21,31 @@
...
+ * <p>Nesting is as follows
+ * <ul>
+ * <li>call -> (dns -> connect -> secure connect)* -> request events</li>
+ * <li>call -> (connection acquire/release)*</li>
+ * </ul>
+ *
+ * <p>Request events are ordered: requestHeaders -> requestBody -> responseHeaders -> responseBody
+ *
+ * <p>Since connections may be reused, the dns and connect events may not be present for a call,
+ * or may be repeated in case of failure retries, even concurrently in case of happy eyeballs type
```
swankjesse: Dig. I hope you don’t mind I might go over this in a follow-up. Unlike code which is pretty objective, documentation is a thing that benefits from lots of copy-editing.

okhttp/src/main/java/okhttp3/EventListener.java
```diff
@@ -21,6 +21,31 @@
...
+ * scenarios. A redirect cross domain, or to use https may cause additional connection and request
+ * events.
+ *
+ * <p>All events must fast, without external locking, cannot throw exceptions,
```
swankjesse: must _execute_ fast?

okhttp/src/main/java/okhttp3/EventListener.java
yschimke: You don't need to ask or generally worry about offending me :)

PR #3506, opened 2017-08-06
Mark remaining @Nullable arguments in EventListener.

dave-r12:


swankjesse:
Travis fails are my own fault. Fix forthcoming.

PR #3510, opened 2017-08-07
Add ability to skip extensive health check for non-GETs

jnelson15:


swankjesse:
I don’t think we want this.

swankjesse:
A better solution is to use HTTP/2, which doesn’t need such hacks!

jnelson15:
Thanks @swankjesse, I'll see if I can use http/2 instead!

PR #3512, opened 2017-08-08
[WIP] HttpUrl - Punycode and Un-encoded params

yschimke:
Exploring how this would work and what edge cases there are.  As I accept user input in my own project, these end up being displayed in a very non-intuitive way.

Chrome will happily show 

https://www.google.com/search?q=рф
or 
http://ουτοπία.δπθ.gr

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/HttpUrlTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/HttpUrlTest.java
@@ -15,6 +15,7 @@
  */
 package okhttp3;
 
+import java.net.IDN;
 import java.net.MalformedURLException;
 import java.net.URI;
 import java.net.URL;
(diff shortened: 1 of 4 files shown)


okhttp/src/main/java/okhttp3/HttpUrl.java
```diff
@@ -972,6 +975,44 @@ static HttpUrl getChecked(String url) throws MalformedURLException, UnknownHostE
     return PublicSuffixDatabase.get().getEffectiveTldPlusOne(host);
   }
 
+  public static String displayHost(Builder.DisplayMode displayMode, String host) {
```
swankjesse: I don’t think DisplayMode is right as public API. It seems like an implementation detail to me.

okhttp/src/main/java/okhttp3/HttpUrl.java
```diff
@@ -972,6 +975,44 @@ static HttpUrl getChecked(String url) throws MalformedURLException, UnknownHostE
...
+        return isDisplaySafe(unicodeHost) ? unicodeHost : host;
+      case UNICODE:
+        return IDN.toUnicode(host, 0);
+      default:
+        return host;
+    }
+  }
+
+  /**
+   * Returns the unicode host if the host is considered IDN safe.
```
swankjesse: Provide an example?

swankjesse: ALSO – I believe in Chrome the display URLs take a Locale, and only use Unicode when the locales match. Should this do likewise?

okhttp/src/main/java/okhttp3/HttpUrl.java
```diff
@@ -1263,6 +1304,10 @@ public HttpUrl build() {
     }
 
     @Override public String toString() {
+      return toString(DisplayMode.RAW);
+    }
+
+    private String toString(DisplayMode displayMode) {
```
swankjesse: nit: package-private? As-is this generates a trampoline method

okhttp/src/main/java/okhttp3/HttpUrl.java
```diff
@@ -1295,7 +1340,12 @@ public HttpUrl build() {
 
       if (encodedQueryNamesAndValues != null) {
         result.append('?');
-        namesAndValuesToQueryString(result, encodedQueryNamesAndValues);
+        if (displayMode != DisplayMode.RAW) {
```
swankjesse: HMMM . . . these URLs are ambiguous aren’t they?

swankjesse: Do browsers do this?

okhttp-tests/src/test/java/okhttp3/HttpUrlTest.java
```diff
@@ -1507,4 +1511,97 @@
     assertNull(HttpUrl.parse("https://localhost").topPrivateDomain());
     assertNull(HttpUrl.parse("https://127.0.0.1").topPrivateDomain());
   }
+
+  @Test public void testDisplayStringPunyCodeNonIdentifier() {
```
swankjesse: nice tests

okhttp-tests/src/test/java/okhttp3/HttpUrlTest.java
```diff
@@ -1507,4 +1511,97 @@
...
+    assertReversible("http://ακρόπολητώρα.org/file.html", "ακρόπολητώρα.org");
+
+    assertReversible("http://☃.net/file.html", "");
+    // mixed characters in wikipedia
+    assertReversible("http://wіkіреdіа.org/file.html", "");
+
+    assertReversible("http://правительство.рф/search?q=рф", "правительство.рф");
+    assertReversible("http://en.wikipedia.org/search?q", "q");
+    assertReversible("http://en.wikipedia.org/search?q=a b", "q=a b");
+    // TODO how to handle?
```
swankjesse: What do browsers do?

okhttp-tests/src/test/java/okhttp3/HttpUrlTest.java
```diff
@@ -1507,4 +1511,97 @@
...
+    HttpUrl originalUrl = HttpUrl.parse(originalUrlString);
+
+    String urlToString = originalUrl.toString();
+    String urlToDisplayString = originalUrl.toDisplayString();
+
+    assertTrue("Couldn't find pattern '" + toStringCheck + "' in '" + urlToDisplayString + "'",
+        Pattern.compile(toStringCheck).matcher(urlToDisplayString).find());
+
+    HttpUrl reparsedUrl = HttpUrl.parse(urlToString);
+    HttpUrl reparsedDisplayUrl = HttpUrl.parse(urlToDisplayString);
```
swankjesse: I really like this approach

okhttp/src/main/java/okhttp3/HttpUrl.java
yschimke: I'll investigate.

okhttp/src/main/java/okhttp3/HttpUrl.java
yschimke: Yes, somewhat.  They seem to do some rewriting, but still maintain the reversibility.  I know this is tricky code, so I'm not in a rush to land this, but I think it's useful functionality.

yschimke:
Thanks for the review, my main question was whether this fits within OkHttp as I think it should.  It seems like if I can get it right you aren't against this.

swankjesse:
Yes, I definitely want this. Biggest challenge is trying to make it both usable and safe. Following the leads of Chrome, Firefox and Safari is the way to get there.

swankjesse:
By the way, I really like reversibility as a goal here. If it’s reversible we’re definitely on the right track.

yschimke:
Will reopen when I get back to this - been busy recently.

PR #3516, opened 2017-08-10
Fix occasional hang with http2 goaway frames

### LINKED ISSUE #3422: multithreaded http2 GET occasionally hang
When running multithreaded HTTP2 GET requests against nginx server, okhttp 3.8.1 occasionally hangs with all client threads stuck with the stacktrace like the one below and no `Http2Connection$ReaderRunnable` thread running. This is on OSX using Oracle java 1.8.0_131 jdk, if this makes any difference.

The problem may be related to server [limiting the number of requests using the same connection to 1000](http://nginx.org/en/docs/http/ngx_http_v2_module.html#http2_max_requests), then sending `GOAWAY` frame. But I am not sure how to setup a standalone test for this hypothesis and generally out of ideas how to troubleshoot this further.

```
"pool-1-thread-1" #27 prio=5 os_prio=31 tid=0x00007fdc1f9bd800 nid=0x8603 in Object.wait() [0x0000700002c18000]
   java.lang.Thread.State: WAITING (on object monitor)
	at java.lang.Object.wait(Native Method)
	at java.lang.Object.wait(Object.java:502)
	at okhttp3.internal.http2.Http2Stream.waitForIo(Http2Stream.java:577)
	at okhttp3.internal.http2.Http2Stream.takeResponseHeaders(Http2Stream.java:143)
	- locked <0x00000003c109ef90> (a okhttp3.internal.http2.Http2Stream)
	at okhttp3.internal.http2.Http2Codec.readResponseHeaders(Http2Codec.java:120)
	at okhttp3.internal.http.CallServerInterceptor.intercept(CallServerInterceptor.java:75)
	at okhttp3.internal.connection.ConnectInterceptor.intercept(ConnectInterceptor.java:45)
	at okhttp3.internal.cache.CacheInterceptor.intercept(CacheInterceptor.java:93)
	at okhttp3.internal.http.RetryAndFollowUpInterceptor.intercept(RetryAndFollowUpInterceptor.java:120)
	at okhttp3.logging.HttpLoggingInterceptor.intercept(HttpLoggingInterceptor.java:211)
	at okhttp3.RealCall.getResponseWithInterceptorChain(RealCall.java:185)
	at okhttp3.RealCall.execute(RealCall.java:69)
	at io.takari.generation.aether.connector.OkhttpContentLoader.get(OkhttpContentLoader.java:105)
...
```

ifedorenko:
Http2Connection was dropping headers of healthy streams
after GOAWAY frame.

Fixes #3422

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/internal/http2/HttpOverHttp2Test.java
+++ b/okhttp-tests/src/test/java/okhttp3/internal/http2/HttpOverHttp2Test.java
@@ -24,12 +24,16 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
+import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.SynchronousQueue;
+
 import javax.net.ssl.HostnameVerifier;
 import okhttp3.Cache;
 import okhttp3.Call;
+import okhttp3.Callback;
 import okhttp3.Cookie;
 import okhttp3.Credentials;
 import okhttp3.Headers;
@@ -976,6 +980,33 @@ public final class HttpOverHttp2Test {
 
     assertEquals(0, server.takeRequest().getSequenceNumber());
     assertEquals(0, server.takeRequest().getSequenceNumber());
+  }
+
+  @Test public void responseHeadersAfterGoaway() throws Exception {
+    server.enqueue(new MockResponse()
+        .setHeadersDelay(1, SECONDS)
+        .setBody("ABC"));
+    server.enqueue(new MockResponse()
+        .setSocketPolicy(SocketPolicy.DISCONNECT_AT_END)
+        .setBody("DEF"));
+
+    final BlockingQueue<String> bodies = new SynchronousQueue<>();
+    Callback callback = new Callback() {
+      @Override
+      public void onResponse(Call call, Response response) throws IOException {
+        bodies.add(response.body().string());
+      }
+      @Override
+      public void onFailure(Call call, IOException e) {
+        System.out.println(e);
+      }
+    };
+    client.newCall(new Request.Builder().url(server.url("/")).build()).enqueue(callback);
+    client.newCall(new Request.Builder().url(server.url("/")).build()).enqueue(callback);
+
+    assertEquals("DEF", bodies.poll(2, SECONDS));
+    assertEquals("ABC", bodies.poll(2, SECONDS));
+    assertEquals(2, server.getRequestCount());
   }
 
   /**
--- a/okhttp/src/main/java/okhttp3/internal/http2/Http2Connection.java
+++ b/okhttp/src/main/java/okhttp3/internal/http2/Http2Connection.java
@@ -605,8 +605,8 @@ public final class Http2Connection implements Closeable {
       }
       Http2Stream stream;
       synchronized (Http2Connection.this) {
-        // If we're shutdown, don't bother with this stream.
-        if (shutdown) return;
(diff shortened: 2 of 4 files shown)


ifedorenko:
Not sure why Travis build failed, all tests passed in Travis and `mvn clean verify` works for me locally too (osx, java 8)

```
Results :
Tests run: 350, Failures: 0, Errors: 0, Skipped: 6
```

JakeWharton:
Restarted build. The VM crashed for an unknown reason.

okhttp/src/main/java/okhttp3/internal/http2/Http2Connection.java
```diff
@@ -620,7 +620,7 @@ public Http2Connection build() {
           // Create a stream.
           final Http2Stream newStream = new Http2Stream(streamId, Http2Connection.this,
               false, inFinished, headerBlock);
-          lastGoodStreamId = streamId;
+          if (streamId > lastGoodStreamId) lastGoodStreamId = streamId;
```
swankjesse: I don’t think this works. The lastGoodStreamId refers to the _local_ last good stream, and this is the _remote_ last good stream

okhttp/src/main/java/okhttp3/internal/http2/Http2Connection.java
```diff
@@ -725,6 +725,7 @@ private void applyAndAckSettings(final Settings peerSettings) {
       synchronized (Http2Connection.this) {
         streamsCopy = streams.values().toArray(new Http2Stream[streams.size()]);
         shutdown = true;
+        Http2Connection.this.lastGoodStreamId = lastGoodStreamId;
```
swankjesse: could this cause the stream ID to become smaller?

okhttp/src/main/java/okhttp3/internal/http2/Http2Connection.java
swankjesse: Actually, scratch that. In this case both refer to remote stream IDs.

okhttp/src/main/java/okhttp3/internal/http2/Http2Connection.java
swankjesse: I think this is the line that’s most problematic. The local field `lastGoodStreamId` refers to remotely-initiated streams, and `goAway` is telling us an ID for locally-initiated streams

okhttp-tests/src/test/java/okhttp3/internal/http2/HttpOverHttp2Test.java
```diff
@@ -978,6 +982,33 @@ private void noRecoveryFromErrorWithRetryDisabled(ErrorCode errorCode) throws Ex
     assertEquals(0, server.takeRequest().getSequenceNumber());
   }
 
+  @Test public void responseHeadersAfterGoaway() throws Exception {
```
swankjesse: Nice test!

swankjesse:
I’m trying to understand exactly what’s going on here, and I’m having a bit of trouble mostly because I haven’t looked at this code for a long time.

Client sends request 1
Client sends request 3
Server sends response 3
Server sends GOAWAY shutdown, saying last client-initiated stream ID is 3
Server sends response headers 1
Client rejects response headers 1 because the connection is shutdown

I think the fix is to move this line in `headers()`

```
        if (shutdown) return;
```

down into the block that runs only if the stream is unknown:

```
        if (stream == null) {
          if (shutdown) return;
```

swankjesse:
... Just tested it, and that’s sufficient to get the new test to pass. If you’re cool with that, that’s what I’ll merge.

ifedorenko:
Yes, this is the scenario that causes 3.8.1 to hang and I confirmed your fix solves the problem when running against nginx. Thank you for looking into this.

swankjesse:
Merged manually. Thanks!

PR #3518, opened 2017-08-10
Allow using a charset other than UTF-8 for creating urlencoded form bodies.

niqueco:
This is useful when accessing a web server you don't control which receives form data encoded in a charset other than UTF-8 (e.g. ISO-8859-1).

To use this you pass an encoding to FormBody.Builder constructor. The constructor is used so it's clear you shouldn't switch encodings in the middle of building a form body.

The result is undefined is a character cannot be encoded in the specified charset (as the result of Java's String.getBytes(charset) is also undefined in that case.

### DIFF
--- a/okhttp/src/main/java/okhttp3/HttpUrl.java
+++ b/okhttp/src/main/java/okhttp3/HttpUrl.java
@@ -21,6 +21,7 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
 import java.net.UnknownHostException;
+import java.nio.charset.Charset;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -28,6 +29,7 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
 import javax.annotation.Nullable;
+import okhttp3.internal.Util;
 import okhttp3.internal.publicsuffix.PublicSuffixDatabase;
 import okio.Buffer;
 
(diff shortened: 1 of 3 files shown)


niqueco:
I can't see why travis build failed. I've run mvn verify locally and everything is fine (including the one test I've added to test this new addition).

JakeWharton:
Build failure is unrelated to your change. Have you rebased on the latest master? I can't tell if you just have an old ancestor commit or there's actually a new JDK we need to support with the ALPN jar.

niqueco:
This commit is based on the 3.8 branch. Should I rebase this to master?

yschimke:
@niqueco rebase onto master.  You can edit this PR, to target master after you have done the rebase.

niqueco:
Done. As I was rewriting the commit anyway I've included @JakeWharton 's formatting suggestions as well.

okhttp/src/main/java/okhttp3/HttpUrl.java
```diff
@@ -1875,12 +1885,19 @@ static void canonicalize(Buffer out, String input, int pos, int limit, String en
...
-        if (utf8Buffer == null) {
-          utf8Buffer = new Buffer();
+        if (encodedCharBuffer == null) {
+          encodedCharBuffer = new Buffer();
         }
-        utf8Buffer.writeUtf8CodePoint(codePoint);
-        while (!utf8Buffer.exhausted()) {
-          int b = utf8Buffer.readByte() & 0xff;
+
+        if (charset == null || charset.equals(Util.UTF_8)) {
```
swankjesse: nice

okhttp/src/main/java/okhttp3/HttpUrl.java
```diff
@@ -1875,12 +1885,19 @@ static void canonicalize(Buffer out, String input, int pos, int limit, String en
...
+          encodedCharBuffer.writeUtf8CodePoint(codePoint);
+        } else {
+          encodedCharBuffer.writeString(input.substring(i, i + Character.charCount(codePoint)),
```
swankjesse: nit: the overload `Buffer.writeString(String string, int beginIndex, int endIndex, Charset charset)` allows us to later avoid allocating a substring object

swankjesse: (though that optimization isn’t currently implemented)

swankjesse: Fixed it myself to unblock. GitHub is pretty handy!

swankjesse:
Thanks!

niqueco:
Thank you! =)

Issue #3521, opened 2017-08-14
multithreaded http2 GET occasionally throws IllegalStateException


ifedorenko:
I get infrequent exceptions like below when downloading artifacts from nginx on multiple threads over http2. The problem is likely related to (or at least made more severe by) nginx sending GOAWAY frames to force the client acquire new connection. I am using okhttp 3.9.0-SNAPSHOT with swankjesse's fix for #3516. I can provide nginx config and a small java app that triggers the problem semi-reliably but don't have proper standalone test.

```
java.lang.IllegalStateException
	at okhttp3.internal.connection.StreamAllocation.acquire(StreamAllocation.java:419)
	at okhttp3.ConnectionPool.get(ConnectionPool.java:126)
	at okhttp3.OkHttpClient$1.get(OkHttpClient.java:152)
	at okhttp3.internal.connection.StreamAllocation.findConnection(StreamAllocation.java:175)
	at okhttp3.internal.connection.StreamAllocation.findHealthyConnection(StreamAllocation.java:133)
	at okhttp3.internal.connection.StreamAllocation.newStream(StreamAllocation.java:112)
	at okhttp3.internal.connection.ConnectInterceptor.intercept(ConnectInterceptor.java:42)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:147)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:121)
	at okhttp3.internal.cache.CacheInterceptor.intercept(CacheInterceptor.java:93)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:147)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:121)
	at okhttp3.internal.http.BridgeInterceptor.intercept(BridgeInterceptor.java:93)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:147)
	at okhttp3.internal.http.RetryAndFollowUpInterceptor.intercept(RetryAndFollowUpInterceptor.java:125)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:147)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:121)
	at okhttp3.RealCall.getResponseWithInterceptorChain(RealCall.java:202)
	at okhttp3.RealCall.execute(RealCall.java:77)
[...]

PR #3525, opened 2017-08-15
Recover from an already-allocated connection that restricts new streams.

dave-r12:
This is an edge case that can occur with HTTP/2. Since multiple requests
use the same connection, it's possible for one request to flag the
connection as bad during a follow-up request.

https://github.com/square/okhttp/issues/3521

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/internal/http2/HttpOverHttp2Test.java
+++ b/okhttp-tests/src/test/java/okhttp3/internal/http2/HttpOverHttp2Test.java
@@ -29,7 +29,6 @@ import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.SynchronousQueue;
-
 import javax.net.ssl.HostnameVerifier;
 import okhttp3.Cache;
 import okhttp3.Call;
@@ -46,6 +45,7 @@ import okhttp3.RecordingHostnameVerifier;
 import okhttp3.Request;
 import okhttp3.RequestBody;
 import okhttp3.Response;
+import okhttp3.Route;
 import okhttp3.TestUtil;
 import okhttp3.internal.DoubleInetAddressDns;
 import okhttp3.internal.RecordingOkAuthenticator;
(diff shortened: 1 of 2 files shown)


dave-r12:
I couldn't come up with any good words to describe this. Will continue to think about it.

okhttp/src/main/java/okhttp3/internal/connection/StreamAllocation.java
```diff
@@ -257,6 +257,29 @@ private RealConnection findConnection(int connectTimeout, int readTimeout, int w
...
+  private RealConnection previousAllocatedConnection() {
+    assert (Thread.holdsLock(connectionPool));
+
+    RealConnection allocatedConnection = this.connection;
+    if (allocatedConnection == null) return null;
+
+    if (allocatedConnection.noNewStreams) {
+      // Our connection was flagged to disallow new streams. This could happen in HTTP/2 when
+      // multiple requests use the same connection. Discard it.
+      deallocate(false, true, true);
```
ifedorenko: I think this should be `deallocate(false, false, true)`, i.e. with `released=false`. Having the streamAllocation instance marked as released here results in `IllegalStateException: released` when request is retried after failures (see StreamAllocation#streamFailed and RetryAndFollowUpInterceptor.intercept).

dave-r12: Good catch, confirmed it with the test.

okhttp-tests/src/test/java/okhttp3/internal/http2/HttpOverHttp2Test.java
```diff
@@ -736,6 +736,67 @@ private void noRecoveryFromErrorWithRetryDisabled(ErrorCode errorCode) throws Ex
...
+        .setBody("DEF"));
+    server.enqueue(new MockResponse()
+        .setResponseCode(301)
+        .addHeader("Location", "/foo"));
+    server.enqueue(new MockResponse()
+        .setBody("ABC"));
+
+    final CountDownLatch latch = new CountDownLatch(1);
+    final BlockingQueue<String> responses = new SynchronousQueue<>();
+    okhttp3.Authenticator authenticator = new okhttp3.Authenticator() {
```
swankjesse: neat

okhttp-tests/src/test/java/okhttp3/internal/http2/HttpOverHttp2Test.java
```diff
@@ -736,6 +736,67 @@ private void noRecoveryFromErrorWithRetryDisabled(ErrorCode errorCode) throws Ex
...
+
+    // Now make the second request which will cause the HTTP/2 connection to be flagged as bad.
+    client.newCall(request).enqueue(callback);
+    String response2 = responses.take();
+    assertEquals("DEF", response2);
+
+    // Let the first request proceed.
+    latch.countDown();
+    String response3 = responses.take();
+    assertEquals("ABC", response3);
```
swankjesse: Maybe add this:

`assertEquals(0, server.takeRequest().sequenceNumber)`

?

okhttp/src/main/java/okhttp3/internal/connection/StreamAllocation.java
```diff
@@ -257,6 +257,29 @@ private RealConnection findConnection(int connectTimeout, int readTimeout, int w
...
+  private RealConnection previousAllocatedConnection() {
+    assert (Thread.holdsLock(connectionPool));
+
+    RealConnection allocatedConnection = this.connection;
+    if (allocatedConnection == null) return null;
+
+    if (allocatedConnection.noNewStreams) {
+      // Our connection was flagged to disallow new streams. This could happen in HTTP/2 when
+      // multiple requests use the same connection. Discard it.
+      deallocate(false, false, true);
```
swankjesse: ugh, this is the most annoying API to call. `deallocate` returns a socket and it’s the caller’s responsibility to close it, but only _after_ releasing the connection pool lock.

Which sucks here, because you can’t return a Socket and a RealConnection neatly

swankjesse: Maybe instead of calling this method `previousAllocatedConnection()` this method should instead be called say, `releaseIfNoNewStreams()` and that could return a `Socket` ?

swankjesse: (And then the caller would just use `this.connection` in all cases. It would either be null or not)

dave-r12: Ahh right, missed that. 👍  to `releaseIfNoNewStreams()`. I'll fix that.

okhttp-tests/src/test/java/okhttp3/internal/http2/HttpOverHttp2Test.java
dave-r12: Can do.

swankjesse:
Nice!

PR #3527, opened 2017-08-16
Fix snapshots link

PromanSEW:


JakeWharton:
Hmm the problem is that the original URL is what you need to use for a build system

PromanSEW:
@JakeWharton this link is not for browser navigation?
If not, it will be good to provide it as URL, not link

JakeWharton:
No. It's for build systems.

swankjesse:
Yeah, we don’t want people downloading these manually; that’s always going to spell trouble.

swankjesse:
Thanks though!

PromanSEW:
If you don't want that users download JARs manually, why you decorated this URL as link instead of plain text? -_-
I don't understand

PR #3532, opened 2017-08-20
EventListener payload bytes and header sizes in chars

yschimke:
Request bytes and Header chars logged to EventListener API.

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
@@ -883,6 +883,4 @@ public final class EventListenerTest {
         .build();
     server.useHttps(sslClient.socketFactory, tunnelProxy);
   }
-
-  // TODO add POST requests with body (streaming, empty, etc) and check length
 }
--- a/okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
@@ -24,13 +24,16 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
+import javax.annotation.Nullable;
 import okhttp3.RecordingEventListener.ConnectEnd;
 import okhttp3.RecordingEventListener.ConnectStart;
 import okhttp3.RecordingEventListener.ConnectionAcquired;
 import okhttp3.RecordingEventListener.DnsEnd;
 import okhttp3.RecordingEventListener.DnsStart;
 import okhttp3.RecordingEventListener.RequestBodyEnd;
+import okhttp3.RecordingEventListener.RequestHeadersEnd;
 import okhttp3.RecordingEventListener.ResponseBodyEnd;
+import okhttp3.RecordingEventListener.ResponseHeadersEnd;
 import okhttp3.RecordingEventListener.SecureConnectEnd;
 import okhttp3.RecordingEventListener.SecureConnectStart;
 import okhttp3.internal.DoubleInetAddressDns;
@@ -42,6 +45,9 @@ import okhttp3.mockwebserver.MockWebServer;
 import okhttp3.mockwebserver.SocketPolicy;
 import okio.Buffer;
 import okio.BufferedSink;
+import org.hamcrest.BaseMatcher;
+import org.hamcrest.Description;
+import org.hamcrest.Matcher;
 import org.junit.After;
 import org.junit.Before;
 import org.junit.Rule;
@@ -49,11 +55,14 @@ import org.junit.Test;
 
 import static java.util.Arrays.asList;
 import static okhttp3.TestUtil.defaultClient;
+import static org.hamcrest.CoreMatchers.any;
+import static org.hamcrest.CoreMatchers.equalTo;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertSame;
+import static org.junit.Assert.assertThat;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
@@ -119,12 +128,54 @@ public final class EventListenerTest {
     assertEquals(expectedEvents, listener.recordedEventTypes());
   }
 
+  private void assertBytesReadWritten(RecordingEventListener listener,
+      @Nullable Matcher<Long> requestHeaderBytes, @Nullable Matcher<Long> requestBodyBytes,
+      @Nullable Matcher<Long> responseHeaderBytes, @Nullable Matcher<Long> responseBodyBytes) {
+
+    if (requestHeaderBytes != null) {
+      RequestHeadersEnd responseHeadersEnd = listener.removeUpToEvent(RequestHeadersEnd.class);
+      // TODO implement correct bytes
+      //assertThat("request header bytes", responseHeadersEnd.bytesWritten, requestHeaderBytes);
+    }
+
+    if (requestBodyBytes != null) {
+      RequestBodyEnd responseBodyEnd = listener.removeUpToEvent(RequestBodyEnd.class);
+      assertThat("request body bytes", responseBodyEnd.bytesWritten, requestBodyBytes);
+    }
+
+    if (responseHeaderBytes != null) {
+      ResponseHeadersEnd responseHeadersEnd = listener.removeUpToEvent(ResponseHeadersEnd.class);
+      // TODO implement correct bytes
+      //assertThat("response header bytes", responseHeadersEnd.bytesRead, responseHeaderBytes);
+    }
+
+    if (responseBodyBytes != null) {
+      ResponseBodyEnd responseBodyEnd = listener.removeUpToEvent(ResponseBodyEnd.class);
+      assertThat("response body bytes", responseBodyEnd.bytesRead, responseBodyBytes);
+    }
+  }
+
+  private Matcher<Long> greaterThan(final long l) {
+    return new BaseMatcher<Long>() {
+      @Override public void describeTo(Description description) {
+        description.appendText("> " + l);
+      }
+
+      @Override public boolean matches(Object o) {
+        return ((Long)o) > l;
+      }
+    };
+  }
+
   @Test public void successfulEmptyH2CallEventSequence() throws IOException {
     enableTlsWithTunnel(false);
     server.setProtocols(Arrays.asList(Protocol.HTTP_2, Protocol.HTTP_1_1));
     server.enqueue(new MockResponse());
 
     assertSuccessfulEventOrder();
+
+    assertBytesReadWritten(listener, any(Long.class), null, greaterThan(0L),
+        equalTo(0L));
   }
 
   @Test public void successfulEmptyHttpsCallEventSequence() throws IOException {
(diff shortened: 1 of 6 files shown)


yschimke:
I still have some TODOs but working towards the header size as strings

dave-r12:
This looks like a good start.

okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
```diff
@@ -119,12 +128,55 @@ private void assertSuccessfulEventOrder() throws IOException {
...
   }
 
+  private void assertBytesReadWritten(RecordingEventListener listener,
+      @Nullable Matcher<Long> requestHeaderLength, @Nullable Matcher<Long> requestBodyBytes,
+      @Nullable Matcher<Long> responseHeaderLength, @Nullable Matcher<Long> responseBodyBytes) {
+
+    if (requestHeaderLength != null) {
+      RequestHeadersEnd responseHeadersEnd = listener.removeUpToEvent(RequestHeadersEnd.class);
+      // TODO implement correct bytes
+      assertThat("request header bytes", responseHeadersEnd.headerLength, requestHeaderLength);
```
dave-r12: Just wondering: I hadn't seen these used in other test cases, why the `assertThat`?

yschimke: I didn't know to avoid them.   :)

Specifically what I wanted here was an API for Predicate<Integer>, Matcher is pretty much that with self describing failure cases. Failures are like "Expected > 10L, but was 0L"

yschimke:
The remaining TODO is the correct bytes in the streaming request case.  It's awkward as the codec returns a Source

```
  /** Returns an output stream where the request body can be streamed. */
  Sink createRequestBody(Request request, long contentLength);
```

Easiest solution is to wrap the Sink returned.  I don't want to change request.body().writeTo() to return bytes written as it is public API.  I don't want HttpCodec to be stateful and store the bytes.  Or leave it unknown (-1) for streaming bodies, which defeats the point.

Any objection to wrapping Sink?

yschimke:
3 failing tests to work through

image

yschimke:
Now with request body bytes via a Counting sink

okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
```diff
@@ -42,22 +45,31 @@
...
+import org.hamcrest.Description;
+import org.hamcrest.Matcher;
 import org.junit.After;
 import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
 
 import static java.util.Arrays.asList;
 import static okhttp3.TestUtil.defaultClient;
+import static org.hamcrest.CoreMatchers.any;
```
swankjesse: I’m not 100% sure we have this dependency when OkHttp is consumed by AOSP

okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
```diff
@@ -119,12 +133,84 @@ private void assertSuccessfulEventOrder() throws IOException {
...
+
+    if (responseBodyBytes != null) {
+      ResponseBodyEnd responseBodyEnd = listener.removeUpToEvent(ResponseBodyEnd.class);
+      assertThat("response body bytes", responseBodyEnd.bytesRead, responseBodyBytes);
+    } else {
+      assertFalse("Found ResponseBodyEnd", listener.recordedEventTypes().contains("ResponseBodyEnd"));
+    }
+  }
+
+  private Matcher<Long> greaterThan(final long l) {
```
swankjesse: `l` is forbidden as variable name. Looks like `1`, especially on line 176.

okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
```diff
@@ -119,12 +133,84 @@ private void assertSuccessfulEventOrder() throws IOException {
...
+
   @Test public void successfulEmptyH2CallEventSequence() throws IOException {
     enableTlsWithTunnel(false);
     server.setProtocols(Arrays.asList(Protocol.HTTP_2, Protocol.HTTP_1_1));
     server.enqueue(new MockResponse());
 
-    assertSuccessfulEventOrder();
+    assertSuccessfulEventOrder(matchesProtocol(Protocol.HTTP_2));
+
+    assertBytesReadWritten(listener, any(Long.class), null, greaterThan(0L),
```
swankjesse: neat

okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
```diff
@@ -678,6 +773,10 @@ private void responseBodyFail(Protocol expectedProtocol) throws IOException {
         .url(server.url("/"))
         .build());
     Response response = call.execute();
+    if (expectedProtocol == Protocol.HTTP_2) {
+      // soft failure since client may not support depending on Platform
```
swankjesse: Our tests should _always_ support HTTP/2. We don’t run tests in environments where that is absent.

(Maybe we should?)

yschimke: I would argue that the tests should be runnable without HTTP/2 support.  But agree it is a separate topic.  I don't see the harm in making HTTP/2 tests shortcircuit if they can't support it.  It might be cleaner to check some new method Platform.getInstance().isHttp2Supported()

okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
yschimke: Should we find out first? or land this PR and then revert the dependency on hamcrest if needed?

yschimke: It's a core dependency of junit 4.12, so it would surprise me if it runs without

http://mvnrepository.com/artifact/junit/junit/4.12

swankjesse: Yep, it’s there. Whew. https://android.googlesource.com/platform/external/hamcrest/

okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
```diff
@@ -119,12 +133,84 @@ private void assertSuccessfulEventOrder() throws IOException {
...
+  }
+
+  private Matcher<Long> lessThan(final long value) {
+    return new BaseMatcher<Long>() {
+      @Override public void describeTo(Description description) {
+        description.appendText("< " + value);
+      }
+
+      @Override public boolean matches(Object o) {
+        return ((Long)o) < value;
```
swankjesse: supernit: space after cast

okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
swankjesse: yeah, for sure. We can use `assumeTrue()` for that.

okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
```diff
@@ -741,6 +841,60 @@ private void requestBodyFail() throws IOException {
...
+
+  @Test public void requestBodySuccessStreaming() throws IOException {
+    RequestBody requestBody = new RequestBody() {
+      @Override public MediaType contentType() {
+        return MediaType.parse("text/plain");
+      }
+
+      @Override public void writeTo(BufferedSink sink) throws IOException {
+        sink.write(new byte[8192]);
+        sink.flush();
```
swankjesse: supernit: not strictly necessary

okhttp/src/main/java/okhttp3/EventListener.java
```diff
@@ -193,11 +193,10 @@ public void requestHeadersStart(Call call) {
    *
    * <p>This method is always invoked after {@link #requestHeadersStart(Call)}.
    *
-   * <p>{@code throwable} will be null in the case of a successful attempt to send the headers.
-   *
-   * <p>{@code throwable} will be non-null in the case of a failed attempt to send the headers.
+   * @param headerLength the length in java characters of headers to be written.
```
swankjesse: I think we want to say “in bytes, prior to any encoding like HPACK” . . . because headers are ASCII-only, Java characters and bytes will be the same thing. And I like keeping the units for headers and bodies interchangeable.

swankjesse: ... so you can add ’em up!

okhttp/src/main/java/okhttp3/EventListener.java
```diff
@@ -269,11 +270,10 @@ public void responseBodyStart(Call call) {
    *
    * <p>This method is always invoked after {@link #requestBodyStart(Call)}.
    *
-   * <p>{@code throwable} will be null in the case of a successful attempt to send the body.
-   *
-   * <p>{@code throwable} will be non-null in the case of a failed attempt to send the body.
+   * @param bytesRead the length in bytes of the body read, including partial success.
+   * @param throwable null if response body was successfully received, non-null otherwise.
```
swankjesse: With `Callback` we report only `IOException` failures; everything else is a crash. Should we do likewise for event listeners? I suspect it might simplify our code to not attempt to send say, a `NullPointerException` caused by OkHttp crashing to the event listener.

Or in other words, if there’s an exception that isn’t an IOException our invariants are broken anyway, and we shouldn’t attempt to provide nice errors when we know we’re broken.

(This isn’t for this PR, but something to contemplate for a follow-up).

okhttp/src/main/java/okhttp3/internal/http/CallServerInterceptor.java
```diff
@@ -131,4 +145,28 @@ public CallServerInterceptor(boolean forWebSocket) {
 
     return response;
   }
+
+  private long headerLength(Headers headers) {
+    long length = 0;
+
+    for (int i = 0, size = headers.size(); i < size; i++) {
```
swankjesse: Should we add 4 bytes overhead per header for `: ` and `\r\n` ? I’m tempted to

okhttp/src/main/java/okhttp3/internal/http/CallServerInterceptor.java
```diff
@@ -131,4 +145,28 @@ public CallServerInterceptor(boolean forWebSocket) {
...
+      length += headers.name(i).length();
+      length += headers.value(i).length();
+    }
+
+    return length;
+  }
+
+  final class CountingSink extends ForwardingSink {
```
swankjesse: nice

okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
swankjesse: Backstory is that originally JUnit didn’t have the Hamcrest dependency, and we were trying to avoid adding it!

swankjesse:
I’ve merged with a handful of nitpicks and follow-ups. This is all good forward progress!

okhttp/src/main/java/okhttp3/internal/http/CallServerInterceptor.java
yschimke: Your call. I wouldn't bother as it seems wrong anyway for HTTP/2 with a binary hpack encoding.

okhttp/src/main/java/okhttp3/EventListener.java
yschimke: I like that suggestion a lot.  Throwable is awkward and we can differentiate expected exceptions from things are fundamentally broken.  It also matches the behaviour for call execute/enqueue.

https://square.github.io/okhttp/3.x/okhttp/okhttp3/Callback.html#onFailure-okhttp3.Call-java.io.IOException-

okhttp/src/main/java/okhttp3/internal/http/CallServerInterceptor.java
PromanSEW: Maybe it should be `static final class`?

yschimke: Thanks @PromanSEW https://github.com/square/okhttp/pull/3544

Issue #3531, opened 2017-08-19
RealWebSocket.close() throws IllegalArgumentException when code is 0


Egorand:
In the Javadoc for `WebSocket.close()`:

```
@param code Status code as defined by <a href="...">Section 7.4 of RFC 6455</a> or {@code 0}.
```

When doing the following:

```java
WebSocket webSocket = okHttpClient.newWebSocket(request, wsListener);
webSocket.close(0, "Bye!");
```

It throws:

```java
java.lang.IllegalArgumentException: Code must be in range [1000,5000): 0

	at okhttp3.internal.ws.WebSocketProtocol.validateCloseCode(WebSocketProtocol.java:120)
	at okhttp3.internal.ws.RealWebSocket.close(RealWebSocket.java:399)
	at okhttp3.internal.ws.RealWebSocket.close(RealWebSocket.java:395)
```

It's not clear what `0` would mean anyway, the [RFC](https://tools.ietf.org/html/rfc6455#section-7.4.2) states that this code is not used. Does it make sense to just fix the Javadoc?

JakeWharton:
It may have been changed after running the autobahn test suite. Doss the
commit history offer any answers? I can look later otherwise.

Egorand:
I found c8dbcce that adds `validateCloseCode(code)` to `close()`

yschimke:
WebSocketWriter handles this and tests expect it

```
if (code != 0) {
        validateCloseCode(code);
}
```

e.g. https://github.com/square/okhttp/blob/802c029b82672ea90534e8da1bd42100c194aca9/okhttp-tests/src/test/java/okhttp3/internal/ws/WebSocketWriterTest.java#L271

```
  @Test public void clientEmptyClose() throws IOException {
    clientWriter.writeClose(0, null);
    assertData("888060b420bb");
  }
```

yschimke:
I think we agreed to just change the docs https://github.com/square/okhttp/pull/3540

PR #3535, opened 2017-08-23
Re-add support for closing a websocket without a code

yschimke:


yschimke:
I believe a valid fix for https://github.com/square/okhttp/issues/3531

JakeWharton:
I feel like this behavior may have been the result of running the Autobahn websocket tests.

yschimke:
@JakeWharton Do you think updating the javadoc is preferable?

JakeWharton:
I tried to install the autobahn suite yesterday and failed because OSX. I'll try on my personal laptop this weekend to verify this change.

okhttp/src/main/java/okhttp3/internal/ws/RealWebSocket.java
```diff
@@ -396,7 +396,9 @@ synchronized boolean pong(ByteString payload) {
   }
 
   synchronized boolean close(int code, String reason, long cancelAfterCloseMillis) {
-    validateCloseCode(code);
+    if (code != 0) {
+      validateCloseCode(code);
```
swankjesse: I assume this behaves differently from making validateCloseCode accept 0 because that would be broken for incoming codes?

yschimke: 0 isn't a valid close code, but is acceptable input according to the *public* API `WebSocket.close`, so this seemed correct and safe.  advice welcome...

JakeWharton: Seems like it's just a documentation error then.

yschimke: I assumed this was an existing decision to mean 0 was something like a socket disconnect, without sending a close reason to the peer.

>    0-999
> 
>       Status codes in the range 0-999 are not used.

JakeWharton: Seems unlikely. The WebSocket code is at least 3 years old so I would trust the code far more than the docs.

yschimke: option b - https://github.com/square/okhttp/pull/3540

Issue #3103, opened 2017-01-11
HTTP/2 ignores user-specified Host header


moxie0:
- [X] Feature Request. Start by telling us what problem you’re trying to solve. Often a solution
       already exists! Don’t send pull requests to implement new features without first getting our
       support. Sometimes we leave features out on purpose to keep the project small.

I think the code [here](https://github.com/square/okhttp/blob/90021dcaea583b518fa92cdac58aa757dd6be111/okhttp/src/main/java/okhttp3/internal/http2/Http2Codec.java#L133) ignores the 'Host' header in a `Request` and pulls it straight out of the URI instead:

```
  public static List<Header> http2HeadersList(Request request) {
    Headers headers = request.headers();
    List<Header> result = new ArrayList<>(headers.size() + 4);
    result.add(new Header(TARGET_METHOD, request.method()));
    result.add(new Header(TARGET_PATH, RequestLine.requestPath(request.url())));
    result.add(new Header(TARGET_AUTHORITY, Util.hostHeader(request.url(), false))); // Optional.
    result.add(new Header(TARGET_SCHEME, request.url().scheme()));

    for (int i = 0, size = headers.size(); i < size; i++) {
      // header names must be lowercase.
      ByteString name = ByteString.encodeUtf8(headers.name(i).toLowerCase(Locale.US));
      if (!HTTP_2_SKIPPED_REQUEST_HEADERS.contains(name)) {
        result.add(new Header(name, headers.value(i)));
      }
    }
    return result;
  }
```

I assume this `Request` has gone through `BridgeInterceptor`, in which case it's either guaranteed to have a user-specified 'Host' header, or to have already been populated from the Uri, in which case I think it should be safe to replace:

```
    result.add(new Header(TARGET_AUTHORITY, Util.hostHeader(request.url(), false))); // Optional.
```

with:

```
    result.add(new Header(TARGET_AUTHORITY, request.header("Host"))); // Optional.
```

This might seem silly, but user-specified host headers are essential for those of us who depend on [domain fronting](https://www.bamsoftware.com/papers/fronting/) to circumvent censorship. Happy to submit a PR if this seems acceptable.

swankjesse:
Makes sense. Can do.

PR #3536, opened 2017-08-23
support dns

iceAnson:
okhttp 3.8.1 doesn't support https dns,when i replace https reqeuest header as some domain,it doesn't work,and i fix it .

### DIFF
--- a/okhttp/src/main/java/okhttp3/Address.java
+++ b/okhttp/src/main/java/okhttp3/Address.java
@@ -47,8 +47,11 @@ public final class Address {
   final @Nullable SSLSocketFactory sslSocketFactory;
   final @Nullable HostnameVerifier hostnameVerifier;
   final @Nullable CertificatePinner certificatePinner;
-
-  public Address(String uriHost, int uriPort, Dns dns, SocketFactory socketFactory,
+  String headerHost=null;
+  public String host() {
+    return headerHost;
+  }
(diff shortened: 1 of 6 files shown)


yschimke:
I'm a bit confused about this PR - can you explain more.

OkHttp supports development servers by replacing the DNS instance with one that responds with your actual target host e.g. https://github.com/yschimke/oksocial/blob/master/src/main/java/com/baulsupp/oksocial/network/DnsOverride.java

OkHttp also supports Domain fronting - see https://github.com/square/okhttp/issues/3103 and https://github.com/square/okhttp/pull/3108/files where you use the URL to connect to a specific host e.g. google or cloudflare, but the HOST header to make a tunneled secure request to a different site.

Can you explain what exactly you are trying to do?

iceAnson:
i try the DnsOveride,it has some problems:
1 it's no easy to extend, dns interfaces must be set when okhttp init;
2 DnsOveride 's method "lookup(hostname)" doens't invoke everytime for each request,okhttp cache that? and if this,how to do if the cache expired?
3 so many apps use http intercpetors handle http dns or https dns,i think okhttp can support this way.

yschimke:
Your Dns implementation can be mutable or extensible itself.  DNS is used for every request that doesn't have an existing pooled connection.  

I don't think this PR is likely to be accepted, being generally useful enough or ready to immediately land either.

PR #3540, opened 2017-08-26
Remove WebSocket API support for optional close code

yschimke:
close(0, null) still supported internally on WebSocketWriter.close, but no longer supported on public API.

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/internal/ws/RealWebSocketTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/internal/ws/RealWebSocketTest.java
@@ -81,7 +81,7 @@ public final class RealWebSocketTest {
 
   @Test public void clientCloseWith0Fails() throws IOException {
     try {
-    client.webSocket.close(0, null);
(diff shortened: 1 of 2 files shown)


okhttp-tests/src/test/java/okhttp3/internal/ws/RealWebSocketTest.java
```diff
@@ -79,6 +79,14 @@
     assertFalse(client.webSocket.send("Hello!"));
   }
 
+  @Test public void clientCloseWith0Fails() throws IOException {
+    try {
+      client.webSocket.close(0, null);
```
swankjesse: This needs a `fail()` after it.

yschimke: Thanks, good catch :(

yschimke: https://github.com/square/okhttp/pull/3545

PR #3541, opened 2017-08-26
More debug for possible but unlikely assertion failures

yschimke:
Excluding any *impossible* scenarios - retain root cause to aid users in debugging without raising an issue.


yschimke:
low-priority but might help with users seeing just AssertionError

yschimke:
I'll change to 

`throw new ((AssertionError) AssertionError("xxxx").initCause(e);`

okhttp/src/main/java/okhttp3/internal/platform/JdkWithJettyBootPlatform.java
```diff
@@ -60,7 +60,7 @@
     try {
       removeMethod.invoke(null, sslSocket);
     } catch (IllegalAccessException | InvocationTargetException ignored) {
-      throw new AssertionError();
+      throw new AssertionError("unable to remove", ignored);
```
swankjesse: nit: rename to `e` now that it isn’t ignored?

PR #3543, opened 2017-08-26
IOException instead of Throwable in EventListener API

yschimke:


yschimke:
@adriancole I took call.enqueue and Callback as my guide after @swankjesse suggested it

```
public interface Callback {
  /**
   * Called when the request could not be executed due to cancellation, a connectivity problem or
   * timeout. Because networks can fail during an exchange, it is possible that the remote server
   * accepted the request before the failure.
   */
  void onFailure(Call call, IOException e);
```

I tried testing it by making my EventListener throw a NPE and it actually just broke and I never received onFailure :( . I was hoping it was going to return a IOException that wrapped the NPE.

yschimke:
Consider this for discussion, a bullet in the chamber if you will.

adriancole:
If there's only one bullet in the chamber, well done I'd say! ;)

Issue #3546, opened 2017-08-27
Test Failure in MockWebServerTest


sneivandt:
I am experiencing a test failure in MockWebServerTest#requestUrlReconstructed. The test failure is because the `MockWebServer` is created with the host _localhost_ but the `HttpUrl` returned has the FQDN _localhost.localdomain_.

`Failed tests: 
  MockWebServerTest.requestUrlReconstructed:436 expected:<localhost[]> but was:<localhost[.localdomain]>`

PR #3547, opened 2017-08-27
Test Failure in MockWebServerTest

sneivandt:
Update `MockWebServerTest#requestUrlReconstructed` to address issue #3546.

### DIFF
--- a/mockwebserver/src/test/java/okhttp3/mockwebserver/MockWebServerTest.java
+++ b/mockwebserver/src/test/java/okhttp3/mockwebserver/MockWebServerTest.java
@@ -433,9 +433,8 @@ public final class MockWebServerTest {
     assertEquals("GET /a/deep/path?key=foo%20bar HTTP/1.1", request.getRequestLine());
 
     HttpUrl requestUrl = request.getRequestUrl();
-    String serverFQDN = InetAddress.getByName(server.getHostName()).getCanonicalHostName();
     assertEquals("http", requestUrl.scheme());
-    assertEquals(serverFQDN, requestUrl.host());
+    assertEquals(server.getHostName(), requestUrl.host());
     assertEquals(server.getPort(), requestUrl.port());
     assertEquals("/a/deep/path", requestUrl.encodedPath());
     assertEquals("foo bar", requestUrl.queryParameter("key"));
--- a/mockwebserver/src/test/java/okhttp3/mockwebserver/MockWebServerTest.java
+++ b/mockwebserver/src/test/java/okhttp3/mockwebserver/MockWebServerTest.java
@@ -23,6 +23,7 @@ import java.io.InputStreamReader;
 import java.io.OutputStream;
 import java.net.ConnectException;
 import java.net.HttpURLConnection;
+import java.net.InetAddress;
 import java.net.ProtocolException;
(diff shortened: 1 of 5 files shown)


sneivandt:
Well I see that the CI build failed. It does not look like the failure is related to the change which I have made. Is there something I should do to address this failure?

JakeWharton:
The failure is unrelated. I've restarted the build.

mockwebserver/src/test/java/okhttp3/mockwebserver/MockWebServerTest.java
```diff
@@ -432,8 +433,9 @@
     assertEquals("GET /a/deep/path?key=foo%20bar HTTP/1.1", request.getRequestLine());
 
     HttpUrl requestUrl = request.getRequestUrl();
+    String serverFQDN = InetAddress.getByName(server.getHostName()).getCanonicalHostName();
     assertEquals("http", requestUrl.scheme());
-    assertEquals(server.getHostName(), requestUrl.host());
+    assertEquals(serverFQDN, requestUrl.host());
```
swankjesse: I don’t see how this could fix anything. Doesn’t requestUrl.host() just use `server.getHostName()` ? Ie. how do these end up disagreeing?

sneivandt: `requestUrl.host()` is being set in RecordedRequest. The value of hostname is _localhost.localdomain_.

`String scheme = socket instanceof SSLSocket ? "https" : "http";`
`String hostname = socket.getInetAddress().getHostName();`
`int port = socket.getLocalPort();`
`this.requestUrl = HttpUrl.parse(String.format("%s://%s:%s%s", scheme, hostname, port, path));`

`socket` is coming from `MockWebServer#acceptConnections` and has the value `Socket[addr=/127.0.0.1,port=50196,localport=40867]`.

So it seems like the difference is that the host name for the MockWebServer is directly set to localhost but the RequestUrl is resolving the 127.0.0.1 to get the host name.

sneivandt: For your reference, this the the contents of /etc/hosts on my machine.

```
#
# /etc/hosts: static lookup table for host names
#

#<ip-address>	<hostname.domain.org>	<hostname>
127.0.0.1	localhost.localdomain	localhost
::1		localhost.localdomain	localhost
127.0.0.1       arch.localdomain        arch

# End of file

```

swankjesse: Gotcha. I think I wanna fix `MockWebServer.getHostName()` which seems to be the root cause of this problem.

How about this fix?

```
  public String getHostName() {
    before();
    return inetSocketAddress.getAddress().getCanonicalHostName();
  }
```

sneivandt:
It looks like the CI build may have failed for the same reasons it did for the first run. Perhaps it should be run again.

@swankjesse I did originally try the change you suggested. That change resulted in several other test failures. I am not very familiar with this code base so I hesitated to make that change. I have now made that change as well as some related changes to address the other impacted tests.

swankjesse:
Yeah, this is perfect. Thanks!

PR #3549, opened 2017-08-28
[for discussion] Catch RuntimeException and rewrap as IOException

yschimke:


yschimke:
It feels wrong that a bad listener can corrupt an OkHttp client, and callback never responds.  Without that response, there isn't even a clear error to report or fix.

swankjesse:
I think we just let the entire call crash. Most application code recovers gracefully from IOExceptions; we don’t want to recover from RuntimeExceptions.

yschimke:
@swankjesse call.execute throws a NPE, so you get the result.  But without the change in the PR, with call.enqueue, the RuntimeException just goes to the great UncaughtExceptionHandler in the sky and application code doesn't get any response.

Is that by design?

swankjesse:
Yeah, I think letting it go to the uncaught exception handler is best. I’d prefer to not try to gracefully recover from such problems. The only mitigation I’m in favor of is trying to avoid leaking resources in case that uncaught exception handler doesn’t terminate the process.

Issue #3514, opened 2017-08-09
Attempt to invoke interface method intercept on a null object reference


stitch9108:
Hi,

We are having a large number of crashes reported by Crashlytics with the following error. 
We have this issue with okhttp 3.5 and 3.8.
Maybe there is something wrong in our code.

Can anyone help us?

Thanks a lot

```
Fatal Exception: java.lang.NullPointerException: Attempt to invoke interface method 'okhttp3.Response okhttp3.Interceptor.intercept(okhttp3.Interceptor$Chain)' on a null object reference
       at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:92)
       at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:67)
       at okhttp3.RealCall.getResponseWithInterceptorChain(RealCall.java:179)
       at okhttp3.RealCall$AsyncCall.execute(RealCall.java:129)
       at okhttp3.internal.NamedRunnable.run(NamedRunnable.java:32)
       at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1133)
       at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:607)
       at java.lang.Thread.run(Thread.java:762)
```

swankjesse:
Looks like you’re adding null to your interceptor chain.

stitch9108:
Thanks for your answer @swankjesse 
However, I don't know where we could ne doing that.

Sorry but this is a large project and we are two new guys continuing the work of another company so we're still trying to understand everything.

From what I see, the addInterceptor method is only called using non null objects because it is created during the call as you can see bellow
``okHttpClientBuilder.addInterceptor(new Interceptor() {
                    @Override
                    public Response intercept(Interceptor.Chain chain) throws IOException {

                        HttpUrl url = chain.request().url().newBuilder()
                                .addQueryParameter(Constants.TOKEN_PARAMETER, TokenManager.getInstance().getCurrentToken()).build();

                        Request request = chain.request().newBuilder().url(url).build();

                        return chain.proceed(request);
                    }
                });``

Maybe there is something else that's done wrong. 

Thanks for the help!

Genleon:
@swankjesse, i have this problem too, how to fix it? thanks.
 `
        Interceptor addCookiesInterceptor = chain -> {
            Request.Builder builder = chain.request().newBuilder();
[...]

PR #3555, opened 2017-08-30
Crash early if any interceptor is null.

swankjesse:
Closes: https://github.com/square/okhttp/issues/3514

### DIFF
--- a/okhttp/src/main/java/okhttp3/CertificatePinner.java
+++ b/okhttp/src/main/java/okhttp3/CertificatePinner.java
@@ -226,7 +226,8 @@ public final class CertificatePinner {
   }
 
   /** Returns a certificate pinner that uses {@code certificateChainCleaner}. */
-  CertificatePinner withCertificateChainCleaner(CertificateChainCleaner certificateChainCleaner) {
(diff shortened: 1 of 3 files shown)


okhttp/src/main/java/okhttp3/CertificatePinner.java
```diff
@@ -226,7 +226,8 @@ public void check(String hostname, Certificate... peerCertificates)
   }
 
   /** Returns a certificate pinner that uses {@code certificateChainCleaner}. */
-  CertificatePinner withCertificateChainCleaner(CertificateChainCleaner certificateChainCleaner) {
+  CertificatePinner withCertificateChainCleaner(
+      @Nullable CertificateChainCleaner certificateChainCleaner) {
```
swankjesse: Snuck this in to fix a warning

JakeWharton:
Shouldn't we also have this in the builder interceptor methods so that you get an error (and stacktrace) pointing to where the problem was?

swankjesse:
Good call. I added it to `addInterceptor()` and `addNetworkInterceptor()`. I’m keeping the other new checks because they’re necessary if you manipulate the `List` directly.

PR #3557, opened 2017-08-30
New API, Chain.call() returns the in-flight call.

swankjesse:
Interceptors could use this to detect if the call has been canceled, or
cancel the call, or do other things.


naturalwarren:
Love it - thanks.

Issue #930, opened 2014-06-13
RFC2616 is Dead


adriancole:
There are new specs out, clarifying the behavior of http/1.1. Other projects such as Jetty are already incorporating these clarifications for all new releases. We should have a look and see where it helps us.
- [ ] RFC7230 - [HTTP/1.1: Message Syntax and Routing](http://tools.ietf.org/html/rfc7230) - low-level message parsing and connection management
- [ ] RFC7231 - [HTTP/1.1: Semantics and Content](http://tools.ietf.org/html/rfc7231) - methods, status codes and headers
- [ ] RFC7232 - [HTTP/1.1: Conditional Requests](http://tools.ietf.org/html/rfc7232) - e.g., If-Modified-Since
- [ ] RFC7233 - [HTTP/1.1: Range Requests](http://tools.ietf.org/html/rfc7233) - getting partial content
- [ ] RFC7234 - [HTTP/1.1: Caching](http://tools.ietf.org/html/rfc7234) - browser and intermediary caches
- [ ] RFC7235 - [HTTP/1.1: Authentication](http://tools.ietf.org/html/rfc7235) - a framework for HTTP authentication

Here's an [overview blog](https://www.mnot.net/blog/2014/06/07/rfc2616_is_dead).

swankjesse:
Yeah, we should definitely confirm that 307s work exactly as they're specified to.

floating-cat:
Hope these rfc2616 links (https://github.com/square/okhttp/search?q=rfc2616&type=Code&utf8=%E2%9C%93) will be replaced with RFC723X too.

swankjesse:
Looks like `CacheInterceptor#isEndToEnd` didn’t make it into RFC 7234. Probably worth changing behavior here.

yschimke:
related https://github.com/square/okhttp/pull/4270

yschimke:
Closing as dupe of other issues.

PR #3561, opened 2017-08-30
Update some comments to point at current RFCs

swankjesse:
https://github.com/square/okhttp/issues/930


swankjesse:
There’s lots of defensive code we can probably strip out. For example, we accept far more date formats than we should. I’d still like to get to that as a part of #930.

PR #3562, opened 2017-08-30
Failed, second connection and call.enqueue EventListener tests

yschimke:


okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
```diff
@@ -57,6 +58,7 @@
 import org.junit.Test;
 
 import static java.util.Arrays.asList;
+import static okhttp3.RecordingEventListener.*;
```
JakeWharton: Star imports are forbidden. Ensure you're using https://github.com/square/java-code-styles

okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
```diff
@@ -116,6 +118,59 @@
...
+        .url(server.url("/"))
+        .build());
+    try {
+      call.execute();
+      fail();
+    } catch (IOException expected) {
+      assertEquals("timeout", expected.getMessage());
+    }
+
+    List<String> expectedEvents = Arrays.asList("CallStart", "DnsStart", "DnsEnd",
```
yschimke: this is probably quite brittle... suggestions?

okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
yschimke: Yep, not sure why it happened.

image

JakeWharton:
Hmm check the imports tab? I use the SquareAndroid style, and I hate that
there's two of them!

okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
yschimke: Seems like an intellij bug with 
image

yschimke: TIL: inner classes can be imported with static or normal imports...

okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
```diff
@@ -116,6 +118,59 @@
     assertEquals(expectedEvents, listener.recordedEventTypes());
   }
 
+  @Test public void successfulCallEventSequenceForEnqueue() throws IOException,
```
swankjesse: nit: `throws Exception` is short and easy

okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
```diff
@@ -116,6 +118,59 @@
...
+      InterruptedException {
+    server.enqueue(new MockResponse()
+        .setBody("abc"));
+
+    Call call = client.newCall(new Request.Builder()
+        .url(server.url("/"))
+        .build());
+
+    final CountDownLatch l = new CountDownLatch(1);
```
swankjesse: Can’t have this. `l` is forbidden as a variable name, especially on a line that also has `1` !

okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
```diff
@@ -116,6 +118,59 @@
...
+    l.await();
+
+    List<String> expectedEvents = Arrays.asList("CallStart", "DnsStart", "DnsEnd",
+        "ConnectionAcquired", "ConnectStart", "ConnectEnd", "RequestHeadersStart",
+        "RequestHeadersEnd", "ResponseHeadersStart", "ResponseHeadersEnd", "ResponseBodyStart",
+        "ResponseBodyEnd", "ConnectionReleased", "CallEnd");
+    assertEquals(expectedEvents, listener.recordedEventTypes());
+  }
+
+  @Test public void failedCallEventSequence() throws IOException {
```
swankjesse: Nice

okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
swankjesse: I think this will work. The timeout will always happen reading the response headers.

The awkward thing is the order of ConnectionReleased/CallFailed. These are stable but I don’t love that you get a ConnectionReleased event when you’re expecting ResponseHeadersEnd. Hmm.

okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
```diff
@@ -151,6 +206,31 @@ private void assertSuccessfulEventOrder(Matcher<Response> responseMatcher) throw
     assertEquals(expectedEvents, listener.recordedEventTypes());
   }
 
+  @Test public void secondCallEventSequence() throws IOException {
```
swankjesse: Nice test

Issue #3498, opened 2017-08-04
MockWebServer discards request body if request contains "Expect: 100-Continue" header


Alexander--:
[This line](https://github.com/square/okhttp/blob/e172706/mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.java#L597) disables reading of request body to implement 100-Continue upload scheme. Unfortunately, the logic is backwards: this behaviour must triggered only if client sent "Expect: 100-Continue" *and* the SocketPolicy is `EXPECT_CONTINUE`, but is currently getting triggered for anyone, who sends "Expect: 100-Continue", causing their request body to be lost.

Here is a testcase for reproduction:

    @Test(timeout = 4000)
    public void unexpectedDevelopment() throws Exception {
        try (MockWebServer server = new MockWebServer()) {
            final String theBody = "xxxxxxxxxxxxxxxxxxxxxxxxxxxxx";

            server.start();

            server.enqueue(new MockResponse().setBody("Nice"));

            URLConnection conn = server.url("/").url().openConnection();
            conn.setDoOutput(true);
            conn.setRequestProperty("Expect", "100-Continue");

            try (OutputStream stream = conn.getOutputStream()) {
                stream.write(theBody.getBytes(StandardCharsets.UTF_8));
                stream.flush();
            }

            assertEquals("Nice", convertStreamToString(conn.getInputStream()));

            RecordedRequest request = server.takeRequest();

            assertEquals("POST", request.getMethod());
            assertEquals(theBody, request.getBody().readUtf8());
        }
    }
    
    private static String convertStreamToString(InputStream is) throws IOException {
        try (Scanner scanner = new Scanner(is, "UTF-8")) {
            String result = null;

            try {
                result = scanner.useDelimiter("\\A").next();
            } catch (java.util.NoSuchElementException e) {
                e.printStackTrace();
            }
            
            final IOException trueError = scanner.ioException();
            if (trueError != null) {
                throw trueError;
            }
        
            return result;
        }
    }

Removing "Expect" header works around the issue, but most real-world servers don't start to discard bodies in response to that header, they just ignore it. This bug makes it impossible to emulate those servers with MockWebServer.

swankjesse:
Can fix. Thanks!

PR #3563, opened 2017-08-30
Fix MockWebServer handling of 'Expect: 100 Continue'

swankjesse:
Previously we'd skip the request body unless the socket policy was
EXPECT_CONTINUE.

Closes: https://github.com/square/okhttp/issues/3498


okhttp-tests/src/test/java/okhttp3/internal/http/RecordingProxySelector.java
```diff
@@ -45,9 +45,8 @@ public void assertRequests(URI... expectedUris) {
 
   @Override public void connectFailed(URI uri, SocketAddress sa, IOException ioe) {
     InetSocketAddress socketAddress = (InetSocketAddress) sa;
-    failures.add(
-        Util.format("%s %s:%d %s", uri, socketAddress.getHostName(), socketAddress.getPort(),
```
swankjesse: calling `getHostName()` was doing DNS, which caused problems for CallTest.connectTimeoutsAttemptsAlternateRoute().

Issue #3361, opened 2017-05-17
can't use Cookie.Builder for ipv6 address


henry74918:
I've faced a exception when using ipv6 ip address with Cookie.
I'm using okhttp3 3.6.0 with PresistentCookieJar (https://github.com/franmontiel/PersistentCookieJar).
When try to connect directly using ivp6 address ex: 2001:b011:400a:156e:211:32ff:fe0f:c06b
I get a IllegalArgumentException: unexpected domain.

PresistentCookieJar using Cookie.Builder to create a cookie, set url domain into cookie.
Inside builder.domain(), it checks invalidHostname which can not contains "**:**"
```
  private static boolean containsInvalidHostnameAsciiCodes(String hostnameAscii) {
    for (int i = 0; i < hostnameAscii.length(); i++) {
      char c = hostnameAscii.charAt(i);
      // The WHATWG Host parsing rules accepts some character codes which are invalid by
      // definition for OkHttp's host header checks (and the WHATWG Host syntax definition). Here
      // we rule out characters that would cause problems in host headers.
      if (c <= '\u001f' || c >= '\u007f') {
        return true;
      }
      // Check for the characters mentioned in the WHATWG Host parsing spec:
      // U+0000, U+0009, U+000A, U+000D, U+0020, "#", "%", "/", ":", "?", "@", "[", "\", and "]"
      // (excluding the characters covered above).
      if (" #%/:?@[\\]".indexOf(c) != -1) {
        return true;
      }
    }
    return false;
  }
```
ref:
https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/internal/Util.java#L386

But ipv6 address contains **colon** anyway.
Is this a bug need to be fixed?
Thanks!

swankjesse:
Will take a look!

PR #3564, opened 2017-08-31
Fix IPv6 handling for cookies

swankjesse:
Closes: https://github.com/square/okhttp/issues/3361


okhttp/src/main/java/okhttp3/internal/Util.java
```diff
@@ -463,4 +480,153 @@ public static int checkDuration(String name, long duration, TimeUnit unit) {
   public static AssertionError assertionError(String message, Exception e) {
     return (AssertionError) new AssertionError(message).initCause(e);
   }
+
+  public static int decodeHexDigit(char c) {
```
swankjesse: all code below is moved from HttpUrl

swankjesse:
Closes: https://github.com/square/okhttp/issues/3361

PR #3566, opened 2017-08-31
Don't report connectionAcquired until actually connected.

swankjesse:
Currently our implementation acquires the connection early so that we have
something to close if the call is canceled. Event listeners are simpler if
they only get a connectionAcquired event once the connection has been
actually established.


yschimke:
LGTM! make more sense ordering wise also.

PR #3567, opened 2017-08-31
Fix MockWebServer websockets to be thread safe.

swankjesse:
Closing the sink while another thread is writing the sink is a
potential source for races.

Closes: https://github.com/square/okhttp/issues/3138

### DIFF
--- a/mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.java
+++ b/mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.java
@@ -465,8 +465,6 @@ public final class MockWebServer extends ExternalResource implements Closeable {
               + " didn't make a request");
         }
 
-        source.close();
-        sink.close();
         socket.close();
         openClientSockets.remove(socket);
       }
@@ -681,7 +679,6 @@ public final class MockWebServer extends ExternalResource implements Closeable {
     } catch (IOException e) {
       webSocket.failWebSocket(e, null);
     } finally {
-      closeQuietly(sink);
       closeQuietly(source);
     }
   }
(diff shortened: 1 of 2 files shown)


mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.java
```diff
@@ -681,7 +681,6 @@ private void handleWebSocketUpgrade(Socket socket, BufferedSource source, Buffer
     } catch (IOException e) {
       webSocket.failWebSocket(e, null);
     } finally {
-      closeQuietly(sink);
```
swankjesse: Above on line 671 we’re handing off the sink to another thread, so this operation was unsafe.

okhttp-tests/src/test/java/okhttp3/internal/ws/WebSocketHttpTest.java
```diff
@@ -322,7 +322,7 @@
             assertNull(chain.request().body());
             Response response = chain.proceed(chain.request());
             assertEquals("Upgrade", response.header("Connection"));
-            assertTrue("", response.body().source().exhausted());
+            assertTrue(response.body().source().exhausted());
```
swankjesse: The `""` was useless.

mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.java
JakeWharton: It's hard to tell from jumping around here in the PR, but doesn't `processConnection` also close both the `Sink` and the `Source`?

swankjesse: Oooh, good catch. I just removed those calls; they aren’t useful.

swankjesse: (Well, useful only in that they could cause segments to be recycled that wouldn’t be otherwise. Not worth writing more code for, especially since in normal operations the buffers will be empty.)

PR #3571, opened 2017-09-01
The WebSocketWriter is single-threaded.

swankjesse:
We were synchronizing to permit multiple writer threads. But that was a carry-over
from the previous design where we supported multiple writer threads. With the
current implementation only one thread writes at a time. This synchronization
was not necessary.

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/ws/WebSocketReader.java
+++ b/okhttp/src/main/java/okhttp3/internal/ws/WebSocketReader.java
@@ -46,8 +46,8 @@ import static okhttp3.internal.ws.WebSocketProtocol.toggleMask;
 
 /**
  * An <a href="http://tools.ietf.org/html/rfc6455">RFC 6455</a>-compatible WebSocket frame reader.
- * <p>
- * This class is not thread safe.
+ *
+ * <p>This class is not thread safe.
  */
 final class WebSocketReader {
   public interface FrameCallback {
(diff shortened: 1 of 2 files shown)


okhttp/src/main/java/okhttp3/internal/ws/WebSocketReader.java
```diff
@@ -46,8 +46,8 @@
 
 /**
  * An <a href="http://tools.ietf.org/html/rfc6455">RFC 6455</a>-compatible WebSocket frame reader.
- * <p>
- * This class is not thread safe.
+ *
+ * <p>This class is not thread safe.
```
JakeWharton: ๏_๏

swankjesse: Hmm. Apparently I’m inclined to do it the Google way, which is different than the way specified by Javadoc.

https://google.github.io/styleguide/javaguide.html#s7.1.2-javadoc-paragraphs
http://www.oracle.com/technetwork/java/javase/documentation/index-137868.html

Sigh.

JakeWharton: Yeah we talked about it once. I think we determined it was broken in an older Javadoc version, but I don't really remember. Also I don't actually care either way which is used. I'd prefer to be consistent and it looks like this style is used more throughout the library.

Issue #3365, opened 2017-05-18
[BUG] okhttp3 will raise "java.io.IOException: Content-Length (x) and stream length (0) disagree" while parseFrom a HEAD response


gaort:
with [rfc2616-sec9.4](https://tools.ietf.org/html/rfc2616#section-9.4):
>The HEAD method is identical to GET except that the server MUST NOT
   return a message-body in the response.

so, while parsing a HEAD response, the body-stream length will always be 0;

with [rfc2616-sec14.13](https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.13):
>The Content-Length entity-header field indicates the size of the entity-body, in decimal number of OCTETs, sent to the recipient or, in the case of the HEAD method, the size of the entity-body that would have been sent had the request been a GET.
>
 >      Content-Length    = "Content-Length" ":" 1*DIGIT
>An example is
>
 >     Content-Length: 3495

so, while parsing a HEAD response,  the "Content-Length" header may NOT be 0;

in okhttp3.ResponseBody#bytes(), line 140 nearby

    public final byte[] bytes() throws IOException {
        long contentLength = contentLength();
        if (contentLength > Integer.MAX_VALUE) {
            throw new IOException("Cannot buffer entire body for content length: " + contentLength);
        }

        BufferedSource source = source();
        byte[] bytes;
        try {
            bytes = source.readByteArray();
        } finally {
            Util.closeQuietly(source);
        }
        if (contentLength != -1 && contentLength != bytes.length) {
            throw new IOException("Content-Length ("
            + contentLength
            + ") and stream length ("
            + bytes.length
            + ") disagree");
        }
        return bytes;
    }

so, while okhttp3 receive a HEAD response, there will always raise a java.io.IOException

songkaiape:
I meet this issue too. Please help fix this. Thanks

travisspencer:
I get this in 3.14.3 once we turned on H2 on the server. **No other** changes except that.

yschimke:
@travisspencer We would need a way to reproduce against MockWebServer or externally to investigate.  Please report a bug with a repro, or consider a post to stackoverflow for help.

PR #3572, opened 2017-09-01
Improve handling of HEAD requests with a Content-Length specified.

swankjesse:
Rather than getting the content length from a header on demand, this gets it
at the time that the RealResponseBody is created. This means that HEAD requests
can have a non-zero Content-Length header while their bodies have no content.

Closes: https://github.com/square/okhttp/issues/3365

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/CallTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/CallTest.java
@@ -83,6 +83,7 @@ import org.junit.rules.Timeout;
 import static java.net.CookiePolicy.ACCEPT_ORIGINAL_SERVER;
 import static okhttp3.TestUtil.awaitGarbageCollection;
 import static okhttp3.TestUtil.defaultClient;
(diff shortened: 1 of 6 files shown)


okhttp-tests/src/test/java/okhttp3/CallTest.java
```diff
@@ -243,10 +244,9 @@
         .url(server.url("/"))
         .head()
         .build();
-    executeSynchronously(headRequest)
-        .assertCode(200)
-        .assertHeader("Content-Length", "100")
-        .assertBody("");
+    Response response = client.newCall(headRequest).execute();
+    assertEquals(200, response.code());
+    assertArrayEquals(new byte[0], response.body().bytes());
```
swankjesse: Here's the test. The string() method gets away without doing an extra length check.

JakeWharton: Derp. Blame mobile.

PR #3575, opened 2017-09-03
Fix typo in javadoc

KoFuk:


yschimke:
Not a typo

> A bracket means that end of the range is inclusive -- it includes the element listed. A parenthesis means that end is exclusive and doesn't contain the listed element. So for [first1, last1), the range starts with first1 (and includes it), but ends just before last1.
> 
> Assuming integers:
> 
> (0, 5) = 1, 2, 3, 4
> (0, 5] = 1, 2, 3, 4, 5
> [0, 5) = 0, 1, 2, 3, 4
> [0, 5] = 0, 1, 2, 3, 4, 5

PR #3576, opened 2017-09-03
Don't summarize the request and response headers.

swankjesse:
Just pass the Request and Response objects in directly.

### DIFF
--- a/okhttp/src/main/java/okhttp3/Headers.java
+++ b/okhttp/src/main/java/okhttp3/Headers.java
@@ -110,6 +110,23 @@ public final class Headers {
         : Collections.<String>emptyList();
   }
 
+  /**
+   * Returns the number of bytes required to encode these headers using HTTP/1.1. This is also the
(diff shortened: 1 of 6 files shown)


okhttp/src/main/java/okhttp3/Headers.java
```diff
@@ -110,6 +110,23 @@ public String value(int index) {
...
   }
 
+  /**
+   * Returns the number of bytes required to encode these headers using HTTP/1.1. This is also the
+   * approximate size of HTTP/2 headers before they are compressed with HPACK. This value is
+   * intended to be used as a metric: smaller headers are more efficient to encode and transmit.
+   */
+  public long byteCount() {
+    // Each header name has 2 bytes of overhead for ': ' and every header value has 2 bytes of
+    // overhead for '\r\n'.
```
JakeWharton: Is the extra line between headers and body accounted for somewhere?

swankjesse: Nope. Neither is the request line / status line.

PR #3591, opened 2017-09-09
google-oss-license support

iNoles:
Added support of google-oss-license that couldn't read from parent's pom files.

### DIFF
--- a/benchmarks/pom.xml
+++ b/benchmarks/pom.xml
@@ -12,6 +12,13 @@
   <artifactId>benchmarks</artifactId>
   <name>Benchmarks</name>
 
+  <licenses>
+    <license>
+      <name>Apache 2.0</name>
+      <url>http://www.apache.org/licenses/LICENSE-2.0.txt</url>
+    </license>
+  </licenses>
+
   <dependencies>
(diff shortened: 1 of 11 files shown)


yschimke:
This should work for tooling that is using the maven model, is there an example where this isn't working? Some popular license check tools?

```
$ mvn help:effective-pom | grep -A 2 license
  <licenses>
    <license>
      <name>Apache 2.0</name>
      <url>http://www.apache.org/licenses/LICENSE-2.0.txt</url>
    </license>
  </licenses>
  <scm>
    <connection>scm:git:https://github.com/square/okhttp.git/okhttp-urlconnection</connection>
```

JakeWharton:
Yeah I don't think we want this. Same PR was made to Retrofit. It's the new
[weird, bad] plugin in Play Services. I'd rather push it to them to fix.
The license is in the parent pom and they should resolve that and check.

yschimke:
@iNoles closing for now, but reopen if you want to discuss further.

JakeWharton:
I filed an issue internally about this. Going to wait to see what they say.

PR #3598, opened 2017-09-11
Okhttp

MS893:
It would be great if method canonicalize throw a NullPointerException when input string is null.

There is no check if a value is null whille buiding a body for a POST
RequestBody formBody = new FormBody.Builder() .add("value", value) .build();
if value is null, there is a :
java.lang.NullPointerException: Attempt to invoke virtual method 'int java.lang.String.length()' on a null object reference at okhttp3.HttpUrl.canonicalize(HttpUrl.java:1751)

### DIFF
--- a/benchmarks/pom.xml
+++ b/benchmarks/pom.xml
@@ -6,7 +6,7 @@
   <parent>
     <groupId>com.squareup.okhttp3</groupId>
     <artifactId>parent</artifactId>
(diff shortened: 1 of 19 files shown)


swankjesse:
Please revert the `pom.xml` changes.

yschimke:
Closing as I don't think there is a commit by the author in the stack.

Issue #2309, opened 2016-02-03
MockWebserver fails for empty body requests with "Transfer-Encoding: chunked" header


bclozel:
I'm using an HTTP client that sends a `"Transfer-Encoding: chunked"` HTTP request header for all requests, including with empty bodies. When testing that client with `MockWebServer`, I'm getting the following:

```
MockWebServer[54824] connection from /127.0.0.1 crashed
java.lang.IllegalArgumentException: Request must not have a body: GET /pojos HTTP/1.1
    at okhttp3.mockwebserver.MockWebServer.readRequest(MockWebServer.java:631)
    at okhttp3.mockwebserver.MockWebServer.access$1500(MockWebServer.java:99)
    at okhttp3.mockwebserver.MockWebServer$4.processOneRequest(MockWebServer.java:498)
    at okhttp3.mockwebserver.MockWebServer$4.processConnection(MockWebServer.java:460)
    at okhttp3.mockwebserver.MockWebServer$4.execute(MockWebServer.java:401)
    at okhttp3.internal.NamedRunnable.run(NamedRunnable.java:33)
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
    at java.lang.Thread.run(Thread.java:745)
```

Why is this enforced?
Is there a strong reason or something in the HTTP spec that qualifies those requests as invalid?
Could we relax that check if not necessary?

JakeWharton:
The RFC states that some servers might reject a GET with a body but that it otherwise has no defined semantics. Does the body have contents of `0\r\n`?

bclozel:
I guess you're right.
The [RFC states](https://tools.ietf.org/html/rfc7230#section-3.3) that:

> The presence of a message body in a request is signaled by a
   Content-Length or Transfer-Encoding header field.  Request message
   framing is independent of method semantics, even if the method does
   not define any use for a message body.

So in that case, I guess the client should at least send a `0\r\n` body to send an empty body if nothing has been sent.

Right now the client sends the following:

```
GET /greeting HTTP/1.1\r\n
transfer-encoding: chunked\r\n
host: localhost:50419\r\n
\r\n
0\r\n
\r\n
```

Is there something wrong with this request?

JakeWharton:
[...]

PR #3599, opened 2017-09-11
Issue2309

mdsiebler:
Hi, I am submitting a fix for the Issue 2309, having to do w/ GET requests that include a body. I also fixed a small checkstyle error in sample code.
https://github.com/square/okhttp/issues/2309


swankjesse:
I don’t think we should support GET requests that have a body.

mdsiebler:
ok fair enough

Issue #3537, opened 2017-08-23
Android O StrictMode: Untagged socket detected


hameno:
If we use StrictMode with VmPolicy detectAll() in Android O (Preview 6) we get errors in logcat like this:
```
08-23 14:41:03.439 E/StrictMode: null
                                 java.lang.Throwable: Untagged socket detected; use TrafficStats.setThreadSocketTag() to track all network usage
                                     at android.os.StrictMode.onUntaggedSocket(StrictMode.java:2010)
                                     at com.android.server.NetworkManagementSocketTagger.tag(NetworkManagementSocketTagger.java:78)
                                     at libcore.io.BlockGuardOs.tagSocket(BlockGuardOs.java:47)
                                     at libcore.io.BlockGuardOs.socket(BlockGuardOs.java:310)
                                     at libcore.io.IoBridge.socket(IoBridge.java:667)
                                     at java.net.PlainSocketImpl.socketCreate(PlainSocketImpl.java:116)
                                     at java.net.AbstractPlainSocketImpl.create(AbstractPlainSocketImpl.java:98)
                                     at java.net.Socket.createImpl(Socket.java:484)
                                     at java.net.Socket.getImpl(Socket.java:547)
                                     at java.net.Socket.setSoTimeout(Socket.java:1175)
                                     at okhttp3.internal.connection.RealConnection.connectSocket(RealConnection.java:221)
                                     at okhttp3.internal.connection.RealConnection.connect(RealConnection.java:149)
                                     at okhttp3.internal.connection.StreamAllocation.findConnection(StreamAllocation.java:195)
                                     at okhttp3.internal.connection.StreamAllocation.findHealthyConnection(StreamAllocation.java:121)
                                     at okhttp3.internal.connection.StreamAllocation.newStream(StreamAllocation.java:100)
                                     at okhttp3.internal.connection.ConnectInterceptor.intercept(ConnectInterceptor.java:42)
                                     at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:92)
                                     at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:67)
                                     at okhttp3.internal.cache.CacheInterceptor.intercept(CacheInterceptor.java:93)
                                     at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:92)
                                     at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:67)
[...]

PR #3600, opened 2017-09-11
RFC: example configure hook

yschimke:
Key line is really just 

`configure(address.socketFactory(), new Socket(proxy));`

Question is whether OkHttp should support a configure hook for creating raw sockets, that is hit in all cases?  

Currently a SocketFactory can be used to configure the socket created, except for when a SOCKS proxy is used.

```
    rawSocket = proxy.type() == Proxy.Type.DIRECT || proxy.type() == Proxy.Type.HTTP
        ? address.socketFactory().createSocket()
        : new Socket(proxy);
```

Configure hook options?

1) poor OptionalMethod like below.
2) an optional interface for SocketFactory supporting just a configureSocket method?

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/DelegatingSocketFactory.java
+++ b/okhttp-tests/src/test/java/okhttp3/DelegatingSocketFactory.java
@@ -58,7 +58,7 @@ public class DelegatingSocketFactory extends SocketFactory {
     return configureSocket(socket);
   }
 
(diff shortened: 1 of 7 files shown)


yschimke:
relates to https://github.com/square/okhttp/issues/3537

yschimke:
please excuse the shitty code, just for discussion.

yschimke:
Closing for now as it's not affecting me and SOCKS is a pretty obscure case

PR #3602, opened 2017-09-12
256kB header limit

yschimke:
Two main goals

- Don't allow a malicious server to flood a client
- Turn these hangs into clear failures, e.g. "no header line received within X characters: content starting 'XXXXXXXXX'"

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/CallTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/CallTest.java
@@ -1872,7 +1872,7 @@ public final class CallTest {
   }
 
   @Test public void httpWithExcessiveHeaders() throws IOException {
-    String longLine = "HTTP/1.1 200 " + stringFill('O', 100 * 1024) + "K";
+    String longLine = "HTTP/1.1 200 " + stringFill('O', 256 * 1024) + "K";
 
     server.setProtocols(Collections.singletonList(Protocol.HTTP_1_1));
 
--- a/okhttp-tests/src/test/java/okhttp3/CallTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/CallTest.java
@@ -1871,6 +1871,25 @@ public final class CallTest {
         .assertFailure("HTTP 205 had non-zero Content-Length: 39");
   }
 
+  @Test public void httpWithExcessiveHeaders() throws IOException {
+    String longLine = "HTTP/1.1 200 " + stringFill('O', 100 * 1024) + "K";
+
+    server.setProtocols(Collections.singletonList(Protocol.HTTP_1_1));
+
+    server.enqueue(new MockResponse()
+        .setStatus(longLine)
+        .setBody("I'm not even supposed to be here today."));
+
+    executeSynchronously("/")
+        .assertFailureMatches(".*unexpected end of stream on Connection.*");
+  }
+
+  private String stringFill(char fillChar, int length) {
+    char[] value = new char[length];
+    Arrays.fill(value, fillChar);
+    return new String(value);
+  }
+
   @Test public void canceledBeforeExecute() throws Exception {
     Call call = client.newCall(new Request.Builder().url(server.url("/a")).build());
     call.cancel();
--- a/okhttp/src/main/java/okhttp3/internal/http1/Http1Codec.java
+++ b/okhttp/src/main/java/okhttp3/internal/http1/Http1Codec.java
@@ -74,6 +74,7 @@ public final class Http1Codec implements HttpCodec {
   private static final int STATE_OPEN_RESPONSE_BODY = 4;
   private static final int STATE_READING_RESPONSE_BODY = 5;
   private static final int STATE_CLOSED = 6;
+  private static final int HEADER_LIMIT = Integer.getInteger("okhttp.headerlimit", 256 * 1024);
 
   /** The client that configures this stream. May be null for HTTPS proxy tunnels. */
   final OkHttpClient client;
@@ -83,7 +84,7 @@ public final class Http1Codec implements HttpCodec {
   final BufferedSource source;
   final BufferedSink sink;
   int state = STATE_IDLE;
-  private long headerLimit = 1024 * 100;
+  private long headerLimit = HEADER_LIMIT;
 
   public Http1Codec(OkHttpClient client, StreamAllocation streamAllocation, BufferedSource source,
       BufferedSink sink) {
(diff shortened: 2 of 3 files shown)


okhttp-tests/src/test/java/okhttp3/CallTest.java
```diff
@@ -1871,6 +1871,25 @@ private void postBodyRetransmittedAfterAuthorizationFail(String body) throws Exc
...
+    String longLine = "HTTP/1.1 200 " + stringFill('O', 100 * 1024) + "K";
+
+    server.setProtocols(Collections.singletonList(Protocol.HTTP_1_1));
+
+    server.enqueue(new MockResponse()
+        .setStatus(longLine)
+        .setBody("I'm not even supposed to be here today."));
+
+    executeSynchronously("/")
+        .assertFailureMatches(".*unexpected end of stream on Connection.*");
```
yschimke: I want to replace this with a clear message, since it isn't the server stopping sending, it is the lack of a completed UTF8 line \r\n

mdsiebler:
Hope this isn't out of line, but if you were to add a new limit on code which may be used for existing deployments, maybe it would be best to make it configurable?

yschimke:
@mdsiebler I don't think configurable with a public API is likely, maybe via a System property?

cc @swankjesse

yschimke:
I believe this PR will reduce support burden for users and maintainers of the library.  Errors will be more understandable, and less likely to show up as random hang.  The nasty thing about existing code is that the hang is within OkHttp code, so there is no easy workaround if it happens.

swankjesse:
I really like this. I think the only thing to consider adjusting is the exception, and that might not even be necessary.

Curious: what limits do Chrome & Firefox enforce? As long as we’re consistent with those I don’t think we’ll break anyone.

yschimke:
chrome 256kB https://bugs.chromium.org/p/chromium/issues/detail?id=535629

yschimke:
Considering HTTP/2 out of scope since the impl is different e.g. SETTINGS_MAX_HEADER_LIST_SIZE.  And HTTP/1.1 is the one susceptible to a corrupt stream.

yschimke:
One niggling doubt about the exception is that if it fails after 10Ks of header lines it will have a short limit in the message

```
java.io.IOException: unexpected end of stream on Connection{localhost:64784, proxy=DIRECT hostAddress=localhost/127.0.0.1:64784 cipherSuite=none protocol=http/1.1}
	at okhttp3.internal.http1.Http1Codec.readResponseHeaders(Http1Codec.java:205)
	at okhttp3.internal.http.CallServerInterceptor.intercept(CallServerInterceptor.java:88)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:147)
	at okhttp3.internal.connection.ConnectInterceptor.intercept(ConnectInterceptor.java:45)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:147)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:121)
	at okhttp3.internal.cache.CacheInterceptor.intercept(CacheInterceptor.java:93)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:147)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:121)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:147)
	at okhttp3.internal.http.RetryAndFollowUpInterceptor.intercept(RetryAndFollowUpInterceptor.java:125)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:147)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:121)
	at okhttp3.RealCall.getResponseWithInterceptorChain(RealCall.java:200)
	at okhttp3.RealCall.execute(RealCall.java:77)
	at okhttp3.CallTest.executeSynchronously(CallTest.java:3079)
	at okhttp3.CallTest.executeSynchronously(CallTest.java:3073)
	at okhttp3.CallTest.httpWithExcessiveHeaders(CallTest.java:1883)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
Caused by: java.io.EOFException: \n not found: limit=262144 content=485454502f312e3120323030204f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4f…
	at okio.RealBufferedSource.readUtf8LineStrict(RealBufferedSource.java:227)
	at okhttp3.internal.http1.Http1Codec.readHeaderLine(Http1Codec.java:212)
	at okhttp3.internal.http1.Http1Codec.readResponseHeaders(Http1Codec.java:189)
	... 35 more
```

okhttp-tests/src/test/java/okhttp3/CallTest.java
swankjesse: yeah

swankjesse: Should we fix in Okio?

okhttp/src/main/java/okhttp3/internal/http1/Http1Codec.java
```diff
@@ -74,6 +74,7 @@
   private static final int STATE_OPEN_RESPONSE_BODY = 4;
   private static final int STATE_READING_RESPONSE_BODY = 5;
   private static final int STATE_CLOSED = 6;
+  private static final int HEADER_LIMIT = Integer.getInteger("okhttp.headerlimit", 256 * 1024);
```
swankjesse: Not sure how I feel about making this configurable. In particular I don’t like the idea of anyone actually configuring this!

swankjesse:
Nice!

okhttp/src/main/java/okhttp3/internal/http1/Http1Codec.java
yschimke: Do you want to remove it.  I think OkHttp is an opinionated client library, not a toolkit that must be infinitely configurable.  And this shouldn't have a public API.

But as I expected someone will find this breaks their awkward API, this might be a workaround.

Do you want me to remove it?

swankjesse: Yeah, please do. No urgency though!

yschimke: https://github.com/square/okhttp/pull/3613

PR #3617, opened 2017-09-19
make readahead timeout configurable

ellisjoe:


swankjesse:
Why?

ellisjoe:
I'm regularly seeking around files that I know are larger than could be read in 100ms and I'd like to cancel the request rather than trying and failing to reuse the connection. Setting the readaheadTimeout to 0ms let's me handle reading ahead vs. reopening the connection above the okhttp layer.

swankjesse:
Gotcha. Does `Call.cancel()` work for you? That’s immediate.

ellisjoe:
so that's our current workaround, but it means I need to keep the `Call` around with my `Response`. It would be great if I could just pass around the response and have `close` on the body kill the connection immediately.

swankjesse:
Any chance you can upgrade your backend to HTTP/2 ? That way you can cancel instantly _and_ you don’t have to pay for a connection setup with each call.

I’m reluctant to add this feature because it guarantees degraded behavior for everyone that uses it.

ellisjoe:
It's an external service that I don't control, unfortunately. Does a `disableReadahead` setting to the client sound less error prone?

swankjesse:
The setting is leaking an implementation detail. I really don't like the idea of adding a setting that potentially wrecks your performance.

Have you considered using byte ranges in your requests? That would avoid the need to truncate a connection.

Otherwise I think Call.cancel is the best approach here.

ellisjoe:
Think we're going to have to continue to use `Call.cancel`. It would be nice if we could `cancel` the Response directly though.

swankjesse:
This will do it:
```
  ResponseBody cancelOnCloseBody(final Call call, final ResponseBody body) {
    return new ResponseBody() {
      @Override public MediaType contentType() {
        return body.contentType();
      }

      @Override public long contentLength() {
        return body.contentLength();
      }

      @Override public BufferedSource source() {
        return Okio.buffer(new ForwardingSource(body.source()) {
          @Override public void close() throws IOException {
            call.cancel();
            super.close();
          }
        });
      }
    };
  }
```

Issue #3227, opened 2017-03-18
Fail the WebSocket if a pong isn’t received within some reasonable period


swankjesse:
When we do automatic pings it’d be nice to fail the connection if the peer isn’t ponging back. As-is a server that blackholes pings is not sufficient to consider the connection healthy.

See also https://github.com/square/okhttp/issues/3197

ruijun:
@swankjesse 
 Hello, when to release 3.7?

ahulyk:
hi, any news on that?

dryganets:
Hi @ahulyk,

Thank you for your PR!
I'm not from the okhttp team, but we had the same problem and we are using older version of okhttp to be able to control ping/pong logic for now.

Here are some my thoughts regarding your PR:

- There should be an option to set the timeout. For different use-cases, reasonable time could be different.
- It doesn't matter if you received the pong or not it matters if you are getting any data through this socket. Ie I could receive/send a file through the socket and as a result, I could miss the pong packet or process it only once the limit has passed. I don't want to have the socket to be disconnected in such cases.

ivmos:
@dryganets Also interested in this PR but can you tell me which old okhttp version are you using as workaround? Thanks in advance.

colymore:
Also failling for me.

dryganets:
Pick any version below 3.5.0 as since this version WS is a part of OkHttp and have a slightly different API. It used to be a separate package before.

Version 3.4+ has ping method exposed, so it is possible to send pings to have a custom handling logic.
But you need to close the socket in case timeout is exceeded manually anyway as OkHttp expects sync socket behavior and sockets on android doesn't fail until its send buffer is full.

If I remember correctly, 3.5+ doesn't require you to close the response body in onMessage callback while versions before will crash if the response body is not closed correctly.

ivmos:
@dryganets thanks a lot

> It doesn't matter if you received the pong or not; it matters if you are getting any data through this socket. I.e., I could receive/send a file through the socket and as a result, I could miss the pong packet or process it only once the limit has passed. I don't want to have the socket to be disconnected in such cases.

Agree but then also ping should only be performed if elapsed time since last received byte happened enough time ago (e.g more than configured ping interval time). As we understand ping is redundant if we're already getting traffic through the websocket.. 

That said I guess PR is a good start compared to current behaviour...

ahulyk:
[...]

PR #3623, opened 2017-09-21
Fail the WebSocket if a pong isn’t received

ahulyk:
Fail the WebSocket if a pong isn’t received within some reasonable period #3227

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/ws/RealWebSocket.java
+++ b/okhttp/src/main/java/okhttp3/internal/ws/RealWebSocket.java
@@ -366,7 +366,7 @@ public final class RealWebSocket implements WebSocket, WebSocketReader.FrameCall
     @Override
     public synchronized void onWritePing() {
       if (pingSentPongNotReceived) {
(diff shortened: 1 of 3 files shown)


okhttp/src/main/java/okhttp3/internal/ws/RealWebSocket.java
```diff
@@ -131,6 +135,9 @@
   /** For testing. */
   int pongCount;
 
+  /** True if ping has been sent and pong is not received within some reasonable period.  */
+  private boolean pingSentPongNotRecived;
```
PromanSEW: private boolean pingSentPongNotRec**e**ived;

ahulyk: fixed

swankjesse:
Thanks for this PR. I’ve written a slightly different PR that accomplishes the same goal.

swankjesse:
https://github.com/square/okhttp/pull/3870

PR #3642, opened 2017-10-09
Set the sni by http head first

sandking:
Server Name Indication (SNI) is an extension to the TLS computer networking protocol by which a client indicates which hostname it is attempting to connect to at the start of the handshaking process .
This allows a server to present multiple certificates on the same IP address.

And now if we request by IPv4 address (like '127.0.0.1' ) instead of a regular host name( like  android.com ).
It can not work.

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/connection/RealConnection.java
+++ b/okhttp/src/main/java/okhttp3/internal/connection/RealConnection.java
@@ -265,7 +265,7 @@ public final class RealConnection extends Http2Connection.Listener implements Co
     }
 
     eventListener.secureConnectStart(call);
-    connectTls(connectionSpecSelector);
+    connectTls(connectionSpecSelector, call);
     eventListener.secureConnectEnd(call, handshake);
 
     if (protocol == Protocol.HTTP_2) {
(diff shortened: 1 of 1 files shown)


swankjesse:
Use the desired hostname in the URL, and if you need to configure where that is routed use the Dns interface.

sandking:
@swankjesse 

I  need request by ip instead of a regular host name .
The server (ip address) has present multiple certificates , It makes me should specify the hostname .

The Dns Interface can't set the sni .

yschimke:
Already supported, as @swankjesse suggests

https://github.com/square/okhttp/blob/1d9c158a96382facc3be672f6b41dc79fb5eb48b/okhttp/src/main/java/okhttp3/internal/platform/Platform.java

```
 * <h3>Server name indication (SNI)</h3>
 *
 * <p>Supported on Android 2.3+.
 *
 * Supported on OpenJDK 7+
 *
 * <h3>Session Tickets</h3>
 *
 * <p>Supported on Android 2.3+.
 *
 * <h3>Android Traffic Stats (Socket Tagging)</h3>
 *
 * <p>Supported on Android 4.0+.
```

This is an example override implementation https://github.com/yschimke/oksocial/blob/conscrypt/src/main/java/com/baulsupp/oksocial/network/DnsOverride.java

sandking:
@yschimke 

OK I got it , Thank you .

PR #3647, opened 2017-10-10
Update ProGuard suggested config.

### LINKED ISSUE #3582: OkHttp 3.9.0 IllegalStateException with Cookies and ProGuard
Sample project: https://github.com/talklittle/TestOkHttp3.9.0

Crashes with OkHttp 3.9.0; works with OkHttp 3.8.1

Crashes with `minifyEnabled true`; works with `minifyEnabled false`

I guess ProGuard is stripping away `src/main/resources/`.

```
java.lang.IllegalStateException: Unable to load publicsuffixes.gz resource from the classpath.
       at okhttp3.internal.publicsuffix.PublicSuffixDatabase.findMatchingRule(PublicSuffixDatabase.java:120)
       at okhttp3.internal.publicsuffix.PublicSuffixDatabase.getEffectiveTldPlusOne(PublicSuffixDatabase.java:83)
       at okhttp3.Cookie.parse(Cookie.java:311)
       at okhttp3.Cookie.parse(Cookie.java:219)
       at okhttp3.Cookie.parseAll(Cookie.java:445)
       at okhttp3.internal.http.HttpHeaders.receiveHeaders(HttpHeaders.java:201)
       at okhttp3.internal.http.BridgeInterceptor.intercept(BridgeInterceptor.java:95)
       at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:147)
       at okhttp3.internal.http.RetryAndFollowUpInterceptor.intercept(RetryAndFollowUpInterceptor.java:125)
       at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:147)
       at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:121)
       at okhttp3.RealCall.getResponseWithInterceptorChain(RealCall.java:200)
       at okhttp3.RealCall.execute(RealCall.java:77)
```

NightlyNexus:
Closes #3582.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ If you are using ProGuard you might need to add the following options:
 -dontwarn okhttp3.**
 -dontwarn okio.**
 -dontwarn javax.annotation.**
--keep class okhttp3.internal.publicsuffix.PublicSuffixDatabase
+-keepnames class okhttp3.internal.publicsuffix.PublicSuffixDatabase
 ```
 
 License
--- a/README.md
+++ b/README.md
@@ -54,6 +54,7 @@ If you are using ProGuard you might need to add the following options:
 -dontwarn okhttp3.**
 -dontwarn okio.**
 -dontwarn javax.annotation.**
+-keep class okhttp3.internal.publicsuffix.PublicSuffixDatabase
 ```
 
 License
--- a/README.md
+++ b/README.md
@@ -54,6 +54,7 @@ If you are using ProGuard you might need to add the following options:
 -dontwarn okhttp3.**
 -dontwarn okio.**
 -dontwarn javax.annotation.**
+# A resource is loaded with a relative path so the package of this class must be preserved.
 -keepnames class okhttp3.internal.publicsuffix.PublicSuffixDatabase
 ```
 


JakeWharton:
I thought it was the resource being stripped and not the class. The class is in the stacktrace.

NightlyNexus:
Yeah, you're right.
This rule does work (had this issue on the app I work on), but I guess it keeps too much, then.
I'll go look at the ProGuard docs to see if I can find the right config.

JakeWharton:
ProGuard doesn't strip resources according to its docs.

talklittle:
What I think is going on, is that the stacktrace in the original issue #3582 may have been copied from Crashlytics (I don't remember, but probably), and therefore the class names were deobfuscated using the ProGuard mapping. That's why the class names show up in the stacktrace as though they were untouched by ProGuard--it's misleading.

Adding the `-keep class` ProGuard rule originally suggested by @csshuai fixes the problem, because the package name **was** actually being obfuscated, so the resource directory didn't match the class package name.

JakeWharton:
Isn't the resource looked up with an absolute path though?

And if not, we only need to prevent the class from being obfuscated.
Preventing it from being removed in shrinking is too strict.

talklittle:
> Isn't the resource looked up with an absolute path though?

It had been, until OkHttp 3.9.0: 9bfe575960af8e7bb164642f2f513b5ccb39ff4f

> And if not, we only need to prevent the class from being obfuscated. Preventing it from being removed in shrinking is too strict.

:+1: Then it should be:

     -keepnames class okhttp3.internal.publicsuffix.PublicSuffixDatabase

Short for `-keep,allowshrinking` https://www.guardsquare.com/en/proguard/manual/usage#keepoptions

JakeWharton:
Ah, yep. Totally forgot about that!

Let's do the -keepnames and then we should be good.

NightlyNexus:
Ah, thank you! works for me.

PR #3668, opened 2017-10-27
We should always have our callbacks called

j-baker:
A common user error with OkHttp is to write an interceptor which might
throw an IllegalArgumentException or similar. This works fine when using
the 'execute()' interface, but with 'enqueue()' leads to serious
problems; callbacks might never be fired.

This leads to very subtle bugs. For example, the Retrofit Java 8 call
adapter returns a CompletableFuture, which might never be completed.

Here, we wrap every throwable with an IOException to ensure that the
callback fires.

Appreciate this might need some changes - throwing it up to get feedback.

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/CallTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/CallTest.java
@@ -36,8 +36,10 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.BrokenBarrierException;
 import java.util.concurrent.Callable;
 import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.CyclicBarrier;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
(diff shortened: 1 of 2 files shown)


NightlyNexus:
related: https://github.com/square/okhttp/pull/3549#issuecomment-325404271
Developer errors pretending to be IOExceptions seems like a recipe for even subtler bugs.

j-baker:
Hmm I'm confused. At the moment the bug is incredibly subtle - you have code which works in the execute() codepath, but if you ever use it with enqueue(), in failure cases it will cause thread pool starvation and a dead server.

If I tell the user that there was an error in their code, that feels a lot less subtle and more likely to be caught than just never calling their callback.

Explicitly here my issue is that the behavior is very different between the execute() and enqueue() codepaths - code that works fine with one causes thread pool starvation in the other (and generally only in the error case).

It's not clear that doing a Preconditions check in an interceptor is a Very Bad Idea for example.

Would a PR be accepted that also makes the execute() methods hang forever if people throw exceptions of the wrong type? That would unify the two behaviors, which would make the bug a lot less subtle.

swankjesse:
I think you just want to configure the uncaught exception handler. You can do this by configuring the Dispatcher’s ExecutorService, or globally for the entire process.

If you also want to defend against unchecked exceptions in interceptors, you could definitely write your own interceptor that does that.

PR #3674, opened 2017-11-05
Ensure encodedQuery doesn't include the URL's fragment

EricEdens:
Before updating the call to `delimiterOffset`, the test failed with: 

```
org.junit.ComparisonFailure: 
Expected :
Actual   :#fragment
 <Click to see difference>
```

I did a quick check of other calls to `delimiterOffset`, and those looked fine.

### DIFF
--- a/okhttp/src/main/java/okhttp3/HttpUrl.java
+++ b/okhttp/src/main/java/okhttp3/HttpUrl.java
@@ -611,7 +611,7 @@ public final class HttpUrl {
   public @Nullable String encodedQuery() {
     if (queryNamesAndValues == null) return null; // No query.
     int queryStart = url.indexOf('?') + 1;
-    int queryEnd = delimiterOffset(url, queryStart + 1, url.length(), '#');
+    int queryEnd = delimiterOffset(url, queryStart, url.length(), '#');
     return url.substring(queryStart, queryEnd);
   }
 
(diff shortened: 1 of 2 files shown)


EricEdens:
Not sure why the Travis tests are failing:

```
[ERROR] [ERROR] Some problems were encountered while processing the POMs:
[ERROR] 'build.plugins.plugin[org.apache.maven.plugins:maven-surefire-plugin].dependencies.dependency.version' for org.mortbay.jetty.alpn:alpn-boot:jar must be a valid version but is '${alpn.jdk8.version}'. @ 
```

`mvn clean verify` passed locally.

okhttp/src/main/java/okhttp3/HttpUrl.java
```diff
@@ -611,7 +611,7 @@ static void pathSegmentsToString(StringBuilder out, List<String> pathSegments) {
   public @Nullable String encodedQuery() {
     if (queryNamesAndValues == null) return null; // No query.
     int queryStart = url.indexOf('?') + 1;
-    int queryEnd = delimiterOffset(url, queryStart + 1, url.length(), '#');
```
swankjesse: ahhh, the +1 was here and on the line above. subtle.

swankjesse:
Nice find!

swankjesse:
Merging as-is. I need to add ALPN settings for the latest Java 8.

Issue #3595, opened 2017-09-11
[okhttp:3.8.0] SSLSocketFactory Crash with Android 4.1.2.


shailesh-mota:
I observed this crash often occurring with Samsung 4.1.2 devices for my app.
Crash occurs while building a OkHttpClient.

Crash : 
```
MyNetworkServiceRetryThread:null:java.lang.AssertionError:
at okhttp3.OkHttpClient.javax.net.ssl.SSLSocketFactory systemDefaultSslSocketFactory(javax.net.ssl.X509TrustManager)(SourceFile:292)
at okhttp3.OkHttpClient.void <init>(okhttp3.OkHttpClient$Builder)(SourceFile:250)
at a.b.c.net.MyNetworkService.okhttp3.OkHttpClient okhttp3.OkHttpClient$Builder.build()(SourceFile:3907)
```

Scenario : 
I have set the socket factory and trust manager for "TLS" something like this : 
```
TrustManagerFactory trustManagerFactory = TrustManagerFactory.getInstance(TrustManagerFactory.getDefaultAlgorithm());
trustManagerFactory.init((KeyStore) null);
TrustManager[] trustManagers = trustManagerFactory.getTrustManagers();
if (trustManagers.length != 1 || !(trustManagers[0] instanceof X509TrustManager)) {
    throw new IllegalStateException("Unexpected default trust managers:"+ Arrays.toString(trustManagers));
}
X509TrustManager trustManager = (X509TrustManager) trustManagers[0];

SSLContext sslContext = SSLContext.getInstance("TLS");
sslContext.init(null, new TrustManager[] { trustManager }, null);
SSLSocketFactory sslSocketFactory = sslContext.getSocketFactory();

OkHttpClient client = new OkHttpClient.Builder()
    .sslSocketFactory(sslSocketFactory, trustManager);
    .build();
```

However, I am providing my own `KeyStore`
so `trustManagerFactory.init((KeyStore) null);` above is not set to null.

Any help/workarounds ?

yschimke:
Can you easily reproduce with 3.9.0, it should give us additional information as we now log the root cause.

`throw assertionError("No System TLS", e); // The system has no TLS. Just give up.`

Also, what is the intention for the TrustManager and SSLContext code you show.  It looks pretty much like the default behaviour anyway.  But systemDefaultSslSocketFactory shouldn't be called if you actually configure your client as indicated above.

shailesh-mota:
[...]

PR #3696, opened 2017-11-22
Assertion error super careful

yschimke:
This can't happen but obviously does.  Be super tolerant as we want to always get the root cause on these dodgy phone JVMs.

https://github.com/square/okhttp/issues/3595#issuecomment-345655811

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/internal/UtilTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/internal/UtilTest.java
@@ -1,0 +1,16 @@
+package okhttp3.internal;
+
+import org.junit.Test;
+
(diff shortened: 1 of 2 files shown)


yschimke:
n.b. apparently can't check ae.getCause() != null

https://medium.com/square-corner-blog/no-cause-for-concern-rxjava-and-retrofit-throwing-a-tantrum-96c9e4ba8a6c

okhttp-tests/src/test/java/okhttp3/internal/UtilTest.java
```diff
@@ -0,0 +1,16 @@
+package okhttp3.internal;
```
swankjesse: Nit: paste a copyright header?

Issue #268, opened 2013-08-07
Content-Encoding is broken for HTTP 304 responses


swankjesse:
We're incorrectly attempting to unzip the empty response body.
https://code.google.com/p/android/issues/detail?id=58637

PR #3700, opened 2017-11-23
304 Not Modified overrides Content-Encoding and breaks cache

DracoBlue:
One of our clients uses the android version of okhttp and runs into the following issue:

Our api used to respond the content without gzip. The response is cached on the clients with corresponding etags. Now our api returns the content with "Content-Encoding: gzip", even in case of 304 Not Modified.

okhttp fails with:

```
java.io.EOFException
	at okio.RealBufferedSource.require(RealBufferedSource.java:60)
	at okio.RealBufferedSource.request(RealBufferedSource.java:67)
	at okio.RealBufferedSource.rangeEquals(RealBufferedSource.java:408)
	at okio.RealBufferedSource.rangeEquals(RealBufferedSource.java:392)
	at okhttp3.internal.Util.bomAwareCharset(Util.java:449)
	at okhttp3.ResponseBody.string(ResponseBody.java:174)
	at okhttp3.CacheTest.previouslyNotGzippedContentIsNotModifiedAndSpecifiesGzipEncoding(CacheTest.java:1048)
	at java.lang.reflect.Method.invoke(Method.java:483)
	at org.junit.rules.RunRules.evaluate(RunRules.java:20)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at com.intellij.junit4.JUnit4IdeaTestRunner.startRunnerWithArgs(JUnit4IdeaTestRunner.java:68)
	at com.intellij.rt.execution.junit.IdeaTestRunner$Repeater.startRunnerWithArgs(IdeaTestRunner.java:47)
	at com.intellij.rt.execution.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:242)
	at com.intellij.rt.execution.junit.JUnitStarter.main(JUnitStarter.java:70)
```

The first commit of this PR adds a failing test.

I found the reason. In https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/internal/cache/CacheInterceptor.java#L105 okhttp follows the spec at https://tools.ietf.org/html/rfc7234#section-4.3.4 but breaks the body of the cached response if the "Content-Encoding" actually only appears on 304 Not Modified but hasn't been available before. Thus my fix just removes the "Content-Encoding" response header in the merged cache entry IF it wasn't available before.

After review by @yschimke I added the following changes to the PR:

- "Content-Encoding" and "Content-Type" are now removed in the same way, like "Content-Length" was removed before, when merging cached headers with network headers.

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/cache/CacheInterceptor.java
+++ b/okhttp/src/main/java/okhttp3/internal/cache/CacheInterceptor.java
@@ -101,7 +101,10 @@ public final class CacheInterceptor implements Interceptor {
     // If we have a cache response too, then we're doing a conditional get.
     if (cacheResponse != null) {
       if (networkResponse.code() == HTTP_NOT_MODIFIED) {
-        Headers.Builder newResponseHeadersBuilder = combine(cacheResponse.headers(), networkResponse.headers()).newBuilder();
+        Headers.Builder newResponseHeadersBuilder = combine(
+            cacheResponse.headers(),
+            networkResponse.headers()
+        ).newBuilder();
 
         // If we have Content-Encoding IN the new response, but NOT in the old response our
         // response body will be broken that's why we ignore the new Content-Encoding header. See
--- a/okhttp/src/main/java/okhttp3/internal/cache/CacheInterceptor.java
+++ b/okhttp/src/main/java/okhttp3/internal/cache/CacheInterceptor.java
@@ -103,10 +103,14 @@ public final class CacheInterceptor implements Interceptor {
       if (networkResponse.code() == HTTP_NOT_MODIFIED) {
         Headers.Builder newResponseHeadersBuilder = combine(cacheResponse.headers(), networkResponse.headers()).newBuilder();
 
-        // If we have Content-Encoding IN the new response, but NOT in the old response our response body will be broken
-        // that's why we ignore the new Content-Encoding header. See https://github.com/square/okhttp/pull/3700 for more information.
-        if (!cacheResponse.headers().names().contains("Content-Encoding")) {
+        // If we have Content-Encoding IN the new response, but NOT in the old response our
+        // response body will be broken that's why we ignore the new Content-Encoding header. See
+        // https://github.com/square/okhttp/pull/3700 for more information.
+        String cachedContentEncoding = cacheResponse.headers().get("Content-Encoding");
+        if (cachedContentEncoding == null) {
           newResponseHeadersBuilder.removeAll("Content-Encoding");
+        } else {
+          newResponseHeadersBuilder.set("Content-Encoding", cachedContentEncoding);
         }
 
         Response response = cacheResponse.newBuilder()
--- a/okhttp/src/main/java/okhttp3/internal/cache/CacheInterceptor.java
+++ b/okhttp/src/main/java/okhttp3/internal/cache/CacheInterceptor.java
@@ -101,8 +101,16 @@ public final class CacheInterceptor implements Interceptor {
     // If we have a cache response too, then we're doing a conditional get.
     if (cacheResponse != null) {
       if (networkResponse.code() == HTTP_NOT_MODIFIED) {
+        Headers.Builder newResponseHeadersBuilder = combine(cacheResponse.headers(), networkResponse.headers()).newBuilder();
+
+        // If we have Content-Encoding IN the new response, but NOT in the old response our response body will be broken
+        // that's why we ignore the new Content-Encoding header. See https://github.com/square/okhttp/pull/3700 for more information.
+        if (!cacheResponse.headers().names().contains("Content-Encoding")) {
+          newResponseHeadersBuilder.removeAll("Content-Encoding");
+        }
+
         Response response = cacheResponse.newBuilder()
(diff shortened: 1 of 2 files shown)


DracoBlue:
This case is similar to #268

okhttp/src/main/java/okhttp3/internal/cache/CacheInterceptor.java
```diff
@@ -101,8 +101,16 @@ public CacheInterceptor(InternalCache cache) {
     // If we have a cache response too, then we're doing a conditional get.
     if (cacheResponse != null) {
       if (networkResponse.code() == HTTP_NOT_MODIFIED) {
+        Headers.Builder newResponseHeadersBuilder = combine(cacheResponse.headers(), networkResponse.headers()).newBuilder();
+
+        // If we have Content-Encoding IN the new response, but NOT in the old response our response body will be broken
+        // that's why we ignore the new Content-Encoding header. See https://github.com/square/okhttp/pull/3700 for more information.
+        if (!cacheResponse.headers().names().contains("Content-Encoding")) {
```
yschimke: names requires building a treeset, I think checking on headers() directly would be quicker.  Also it is probably better to unconditionally remove as even a change to Content-Encoding would break things.

DracoBlue: @yschimke good point!

I changed it to *remove* Content-Encoding from combined headers, if it was not present in the cached headers. And keep the cached header if it was present in the cached headers.

DracoBlue:
@yschimke pointed out that `Content-Type` could have the same issue. Should I add this workaround for `Content-Type`, too?

okhttp/src/main/java/okhttp3/internal/cache/CacheInterceptor.java
```diff
@@ -101,8 +101,23 @@ public CacheInterceptor(InternalCache cache) {
...
+        // https://github.com/square/okhttp/pull/3700 for more information.
+        String cachedContentEncoding = cacheResponse.headers().get("Content-Encoding");
+        if (cachedContentEncoding == null) {
+          newResponseHeadersBuilder.removeAll("Content-Encoding");
+        } else {
+          newResponseHeadersBuilder.set("Content-Encoding", cachedContentEncoding);
+        }
+
         Response response = cacheResponse.newBuilder()
-            .headers(combine(cacheResponse.headers(), networkResponse.headers()))
```
yschimke: Looking at this code more, combine is already single use and fairly specific to the use case of updating cached headers.  I suggest moving this logic into combine.

DracoBlue: @yschimke So like:

1. a list of ignored header fields as a parameter
2. ignoring those two headers by default within combine without extra parameters

what do you prefer?

DracoBlue: Ah I see "Content-Length" is handled similary at the moment.

DracoBlue: I moved "Content-Length" together with "Content-Type" and "Content-Encoding" into a new method isNotContentSpecificHeader and use this method in "combine" now to filter which header fields to ignore.

okhttp/src/main/java/okhttp3/internal/cache/CacheInterceptor.java
```diff
@@ -256,4 +254,14 @@ static boolean isEndToEnd(String fieldName) {
         && !"Transfer-Encoding".equalsIgnoreCase(fieldName)
         && !"Upgrade".equalsIgnoreCase(fieldName);
   }
+
+  /**
+   * Returns true if {@code fieldName} should be ignored, when combining cache headers with
+   * network headers.
+   */
+  static boolean isNotContentSpecificHeader(String fieldName) {
```
yschimke: nit: Could we negate this? it's used both ways but I don't like no double negatives

yschimke:
Looking good, if no feedback from anyone else after the small nit, then I'll merge.

DracoBlue:
good idea. Please check if the method comment fits, too.

NightlyNexus:
(if you want to be super picky about the comment... "therefor" is almost always spelled "therefore")

yschimke:
@DracoBlue have you signed the CLA? http://square.github.io/okhttp/#contributing

DracoBlue:
@yschimke I signed it with both my work email and private email

Issue #3429, opened 2017-06-24
Tunneling to https with an auth http proxy


drallieiv:
I tried using okhttp to connect to an https page using a http proxy

If I run a test, it seems to pass by RealConnection::createTunnel

The initial call responds with a 407 asking for auth and passes trough

```
        case HTTP_PROXY_AUTH:
          tunnelRequest = route.address().proxyAuthenticator().authenticate(route, response);
          if (tunnelRequest == null) throw new IOException("Failed to authenticate with proxy");

          if ("close".equalsIgnoreCase(response.header("Connection"))) {
            return tunnelRequest;
          }
          break;
```

but when I look at the headers I am getting : 
```
Proxy-Authenticate: Basic realm="proxy"
Proxy-Connection: close
Content-type: text/html; charset=utf-8
```

The header equal to close is Proxy-Connection and not Connection. And thus it does not goes trough a proxy.

If I do the same process using curl in verbose mode, I see that it makes a tunnel, like it should do when using http to call https.

The question is : 

Is the proxy returning a bad response, or should OkHttp be reading the right header (or both) ?

swankjesse:
Looks like [RFC 7230](https://tools.ietf.org/html/rfc7230#appendix-A.2) has advice on the `Proxy-Connection` header. We might be better to handle `Connection: close` when talking to a proxy. We can experiment to see what Firefox & Chrome do.

vetrovosk:
We are having the same issue with `squid/2.6.STABLE21`, it sends `Proxy-Connection: close` header with 407 response code, so we end up with

```
java.io.IOException: unexpected end of stream on null
	at okhttp3.internal.http1.Http1Codec.readResponseHeaders(Http1Codec.java:208)
	at okhttp3.internal.connection.RealConnection.createTunnel(RealConnection.java:375)
	at okhttp3.internal.connection.RealConnection.connectTunnel(RealConnection.java:218)
	at okhttp3.internal.connection.RealConnection.connect(RealConnection.java:159)
	at okhttp3.internal.connection.StreamAllocation.findConnection(StreamAllocation.java:257)
	at okhttp3.internal.connection.StreamAllocation.findHealthyConnection(StreamAllocation.java:135)
	at okhttp3.internal.connection.StreamAllocation.newStream(StreamAllocation.java:114)
	at okhttp3.internal.connection.ConnectInterceptor.intercept(ConnectInterceptor.java:42)
[...]

Issue #2514, opened 2016-04-26
User-Agent Header not used for TLS tunnel


psniranjan:
In case where HTTP tunnel is needed, the "User-Agent" header is reset and the default agent "okhttp/3.x.x" is used. This is causing the request being ignored as its considered to be originating from an unknown client.

Reference code from RealConnection.

```
 /**
   * Returns a request that creates a TLS tunnel via an HTTP proxy, or null if no tunnel is
   * necessary. Everything in the tunnel request is sent unencrypted to the proxy server, so tunnels
   * include only the minimum set of headers. This avoids sending potentially sensitive data like
   * HTTP cookies to the proxy unencrypted.
   */
  private Request createTunnelRequest() throws IOException {
    return new Request.Builder()
        .url(route.address().url())
        .header("Host", Util.hostHeader(route.address().url(), true))
        .header("Proxy-Connection", "Keep-Alive")
        .header("User-Agent", Version.userAgent()) // For HTTP/1.0 proxies like Squid.
        .build();
  }
```

yschimke:
I had a 5 minute look at this and it seems awkward to fix correctly given creating connections is somewhat detached from individual requests currently.  The user agent could be plumbed through, but it seems ugly.  Also a second http request with different user agent would still use the previous user agent from when the RealConnection is first established.

It does make me wonder whether OkHttpClient should support a user agent string which defaults to Version.userAgent(), but can be overridden in the client builder and eventually per request in headers. It seems natural to have as a client property and that could make it easier for RealConnection to use.

swankjesse:
We can fix this with new settings for preemptive authentication.

shwalabaka:
@psniranjan, @yschimke  Hello. Have the same problem (In case where HTTP tunnel is needed, the "User-Agent" header is reset and the default agent "okhttp/3.x.x" is used) in okhttp/3.7.0 . Did you solve this problem? And could you say how.

yschimke:
@shwalabaka - No, it wasn't affecting me.

Ashok-Varma:
@swankjesse 
1) any update on this bug (or) is there any work around for now to set User-Agent ??
2) do you guys have any stats for % of people who use proxy ??

JoshSolo:
Is this issue still on the roadmap? Last update is from 2018.

swankjesse:
@JoshSolo you can hack this in with an `Authenticator` by following the instructions on pre-emptive authentication:
[...]

PR #3712, opened 2017-12-08
feat: preemptive authentication (#2435)

LFZJun:


okhttp/src/main/java/okhttp3/Authenticator.java
```diff
@@ -28,24 +28,30 @@
...
- *    return response.request().newBuilder()
- *        .header("Authorization", credential)
- *        .build();
  * }</pre>
  *
  * <p>When authentication is requested by a proxy server, the response code is 407 and the
  * implementation should respond with a new request that sets the "Proxy-Authorization" header.
  * <pre>   {@code
  *
+ *  Request authenticate(Route route, Response response) {
```
yschimke: My reading of https://github.com/square/okhttp/issues/2435 was this would be not part of the interface, but a subclass that could be checked for.  This seems like a pretty big API breaking change as is.

LFZJun: Yes, You are right. You mean to add a method to avoid breaking change? It will be nice if the iusses can be solved in 3.10.0. etc. #2514 #3429 looks like familiar.

yschimke: If we prefer interfaces to an abstract class, then something like...

```
public interface PreemptiveAuthenticator extends Authenticator  {    
   /**
    * .... Returns null if no authentication.
    */
   @Nullable Request authenticate(Request request) throws IOException;
}
```

okhttp/src/main/java/okhttp3/Authenticator.java
```diff
@@ -62,11 +70,21 @@
...
+      return null;
+    }
+
     @Override public Request authenticate(Route route, Response response) {
       return null;
     }
   };
 
+  /**
+   * Returns a tunnel request if a TLS tunnel via an HTTP proxy. Returns a request before
```
yschimke: I don't really follow the docs here.  Is this just saying it can be used either for proxy authentication or request authentication?

yschimke:
I like this change because I think preemptive authentication should be the default when clients know  ahead of time it will be required.  I've implemented this using network interceptors and it's works except that the oauth2 refresh token flow means to do it right (and not expose to clients) you would need the network interceptor and an authenticator to retry.  So I do like this and wouldn't put up much of a fight. 

That said to do this right if we are implementing a more complete authenticator API I think we should consider the following features

1) Applying multiple authenticators e.g. a Twitter authenticator + Facebook authenticator
2) Which requests does a particular authenticator apply to?
3) synchronous token refresh flows.

my 2c, on the other hand it does vastly increase the scope of this change.

swankjesse:
Authenticator predates Interceptor. It's less capable and I'd like to consider deprecating Authenticator.

I'd like for our users to use interceptors instead. We should fix whatever we need to fix in interceptors to make that nice. Perhaps built-in Basic and Digest interceptors?

We also need a strategy for proxy auth, but that can be completely separate.

yschimke:
Do we land this PR or similar before that bigger support library?

swankjesse:
I don’t think we can land this because it adds a method to an interface, and that’s incompatible in all kinds of painful ways for our users.

yschimke:
sub interface is a possibility.  But that said, since it is possible with an interceptor, it does raise the question of why it's needed in the API you want to deprecate

PR #3716, opened 2017-12-10
Remove permitsRequestBody in HTTP Methods

EhsanMashhadi:
Remove permitsRequestBody in HTTP Methods to allow custom method with the body.
https://github.com/square/okhttp/issues/229

### DIFF
--- a/mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.java
+++ b/mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.java
@@ -60,7 +60,6 @@ import okhttp3.Response;
 import okhttp3.internal.Internal;
 import okhttp3.internal.NamedRunnable;
 import okhttp3.internal.Util;
-import okhttp3.internal.http.HttpMethod;
 import okhttp3.internal.http2.ErrorCode;
 import okhttp3.internal.http2.Header;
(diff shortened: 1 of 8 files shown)


EhsanMashhadi:
@swankjesse I have changed the code same as #229 but two tests are failed
```
URLConnectionTest.response308WithHead:2334->testRedirect:2373 expected:<[]> but was:<[Page 2]>
 URLConnectionTest.response307WithHead:2318->testRedirect:2373 expected:<[]> but was:<[Page 2]>
```

yschimke:
For further discussion: I don't think removing this check is the right approach.  It might be that we should blacklist instead of whitelisting the special cases.  

Overall given the support for WEBDAV verbs, supporting previously unknown HTTP verbs seems appropriate, so I'm not against the sentiment of the change.

EhsanMashhadi:
@yschimke  We are using lots of custom HTTP method in our projects, I have not seen any problem with using any custom verb with the body. Other rest clients like Alamofire in ios can do it now. Any plan for removing this restriction in a new version?

yschimke:
My point wasn't around unknown custom verbs.  It was that allowing body for a GET request seems like a step back.

yschimke:
I think https://github.com/square/okhttp/pull/3777 covers this

Issue #3719, opened 2017-12-12
Crash: Unexpected TLS version: NONE


anthony3444:
Hi there!

First at all thank you very much for your library!

I've integrated okhttp3 since couple months ago; there were no problem until yesterday with an user (Huawei, model MHA-L29, on Android 7.0):

```
Fatal Exception: java.lang.IllegalArgumentException: Unexpected TLS version: NONE
       at okhttp3.TlsVersion.forJavaName(TlsVersion.java:53)
       at okhttp3.Handshake.get(Handshake.java:56)
       at okhttp3.internal.connection.RealConnection.connectTls(RealConnection.java:300)
       at okhttp3.internal.connection.RealConnection.establishProtocol(RealConnection.java:268)
       at okhttp3.internal.connection.RealConnection.connect(RealConnection.java:160)
       at okhttp3.internal.connection.StreamAllocation.findConnection(StreamAllocation.java:256)
       at okhttp3.internal.connection.StreamAllocation.findHealthyConnection(StreamAllocation.java:134)
       at okhttp3.internal.connection.StreamAllocation.newStream(StreamAllocation.java:113)
       at okhttp3.internal.connection.ConnectInterceptor.intercept(ConnectInterceptor.java:42)
       at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:147)
       at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:121)
       at okhttp3.internal.cache.CacheInterceptor.intercept(CacheInterceptor.java:93)
       at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:147)
       at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:121)
       at okhttp3.internal.http.BridgeInterceptor.intercept(BridgeInterceptor.java:93)
       at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:147)
       at okhttp3.internal.http.RetryAndFollowUpInterceptor.intercept(RetryAndFollowUpInterceptor.java:125)
       at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:147)
       at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:121)
       at okhttp3.RealCall.getResponseWithInterceptorChain(RealCall.java:200)
[...]

PR #3721, opened 2017-12-13
avoid using invalid sessions

yschimke:
These checks are a bit redundant, and originally came from the conscrypt branch.  We should avoid using the session if there was a failure.

I think I'm checking effectively the same thing three times.

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/internal/http2/HttpOverHttp2Test.java
+++ b/okhttp-tests/src/test/java/okhttp3/internal/http2/HttpOverHttp2Test.java
@@ -22,13 +22,14 @@ import java.net.HttpURLConnection;
 import java.net.SocketTimeoutException;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.SynchronousQueue;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 import javax.net.ssl.HostnameVerifier;
 import okhttp3.Cache;
 import okhttp3.Call;
@@ -46,11 +47,11 @@ import okhttp3.Request;
 import okhttp3.RequestBody;
 import okhttp3.Response;
 import okhttp3.Route;
+import okhttp3.TestLogHandler;
 import okhttp3.TestUtil;
 import okhttp3.internal.DoubleInetAddressDns;
 import okhttp3.internal.RecordingOkAuthenticator;
 import okhttp3.internal.SingleInetAddressDns;
-import okhttp3.internal.SocketRecorder;
 import okhttp3.internal.Util;
 import okhttp3.internal.connection.RealConnection;
 import okhttp3.internal.tls.SslClient;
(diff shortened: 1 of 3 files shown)


yschimke:
relates to https://github.com/square/okhttp/issues/3719

yschimke:
@swankjesse apologies about the build failures, this particular change really depends on the SSL provider you use, so local IDE testing is tricky.  I'll fix the tests to avoid triggering these edge cases incidentally.

It will need careful review for such a small change, so we might just want to wait for the conscrypt branch to land to fix it.

okhttp-tests/src/test/java/okhttp3/internal/http2/HttpOverHttp2Test.java
```diff
@@ -876,64 +890,46 @@ private void noRecoveryFromErrorWithRetryDisabled(ErrorCode errorCode) throws Ex
...
+    List<String> logs = http2Handler.takeAll();
+    assertThat("log count", logs.size(), anyOf(equalTo(21), equalTo(22)));
+    assertThat("header logged", firstFrame(logs, "HEADERS"), containsString("HEADERS       END_HEADERS"));
+    assertThat("data logged", firstFrame(logs, "DATA"), containsString("0 DATA          END_STREAM"));
+  }
 
-    assertEquals(1, handler.headerFrameCount);
-    assertEquals(Collections.singletonList(0), handler.dataFrames);
+  private String firstFrame(List<String> logs, String type) {
+    for (String l: logs) {
```
swankjesse: nit: I really dislike `l` as a variable name because it looks like `i` and `I` and `1` and `|`. Perhaps `log` ?

okhttp-tests/src/test/java/okhttp3/internal/http2/HttpOverHttp2Test.java
```diff
@@ -876,64 +890,46 @@ private void noRecoveryFromErrorWithRetryDisabled(ErrorCode errorCode) throws Ex
...
+      if (l.contains(type)) {
+        return l;
```
swankjesse: (this in particular looks like `return 1;`

swankjesse:
Without having spent much time investigating this, I’m surprised that this SSL stuff is interfering with HTTP/2 framing. Those should be more independent?

okhttp-tests/src/test/java/okhttp3/internal/http2/HttpOverHttp2Test.java
yschimke: My bad, I've been guilty of this before.

yschimke:
Yeah, I wouldn't rush this in.  But it would be good to get right before Conscrypt 1.0 is released.  It seems like other providers can behave differently for sessions during connection and when errors/disconnects happen.  So I think we should tighten up OkHttp here.

yschimke:
Can you explain more about the HTTP/2 framing, it's possible my fixes in the tests are masking a real problem with these SSL providers.

swankjesse:
If we're doing things right TLS and HTTP/2 should be very independent of one another. Changing TLS setup shouldn't be able to impact framing because framing shouldn't even know TLS is present.

Is this two unrelated changes in one PR? Or is there a link between the TLS change and the test change that I'm missing?

yschimke:
I think the change is broken as is, let me simplify it further and try to avoid impacting any tests.

yschimke:
@swankjesse is this acceptable in this minimal form?

swankjesse:
Yes, perfect. I’d incorrectly inferred that the tests and the fix were related.

PR #3723, opened 2017-12-14
Update to new Gradle dependency configurations

shaishavgandhi05:
Signed-off-by: shaishavgandhi05 <shaishgandhi@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,6 @@ or Gradle:
 ```groovy
 implementation 'com.squareup.okhttp3:okhttp:3.9.1'
 ```
-Note : Use `api` instead of `implementation` if you're using OkHttp in a library and want to expose it's types as part of the public API.
 
 Snapshots of the development version are available in [Sonatype's `snapshots` repository][snap].
 
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@ or Gradle:
 ```groovy
 implementation 'com.squareup.okhttp3:okhttp:3.9.1'
 ```
+Note : Use `api` instead of `implementation` if you're using OkHttp in a library and want to expose it's types as part of the public API.
 
 Snapshots of the development version are available in [Sonatype's `snapshots` repository][snap].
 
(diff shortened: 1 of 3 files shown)


JakeWharton:
The use of `implementation` for OkHttp might not be accurate. For example, if we were to use this from Retrofit it would be incorrect as OkHttp types are exposed in Retrofit's API.

shaishavgandhi05:
True. Using `implementation` or `api` depends on the use case. I would be open to adding both
`implementation 'com.squareup.okhttp3:okhttp:3.9.1'` and `api 'com.squareup.okhttp3:okhttp:3.9.1'` both and adding a line explaining if you want okhttp types to be exposed beyond your module, you should use `api`.

JakeWharton:
Yeah it's an unfortunate side-effect of this new system. I'm not really sure what's best to do. Another potential approach is to only list the coordinates like
```
com.squareup.okhttp3:okhttp:3.9.0
```
and let people figure out how to apply that to Gradle, Maven, or whatever.

Anyone else have an opinion here? The problem is we have about 30 projects which have very similar READMEs so whatever we do it'd be nice to unify them in the new style.

yschimke:
I think @shaishavgandhi05 suggestion is the best one, use implementation as the only example, as most end users will use that.  Then a one line note referencing "api" for library maintainers to use is types are part of public API.

shaishavgandhi05:
I've kept `implementation` as it is and added a note about using OkHttp in a library and exposing its types in the public API.

NightlyNexus:
I'd like to eliminate the noisy boilerplate of READMEs that shows lines for build systems.
Having only the coordinates would be nice.

Unfortunately, I'm imagining hundreds of issues filed asking how to set every library up with Gradle and Maven.

Seems like `implementation` is good enough to be the copypasta.

swankjesse:
I removed the note about api vs. implementation, which seemed off-topic for OkHttp. Lets do the same for all of our READMEs. Yay!

PR #3741, opened 2017-12-27
Replaced explicit null checks by automatically generated

denis-zhdanov:
This *PR* applies [Traute](https://github.com/square/okhttp) *javac* plugin to the project. It generates *null*-checks based on configured annotations ([package-level ParametersAreNonnullByDefault](https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/package-info.java#L2)).  

Let's consider [Address](https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/Address.java) class as en example:  

*Source code*  

```java
public Address(String uriHost, int uriPort, Dns dns, SocketFactory socketFactory,
      @Nullable SSLSocketFactory sslSocketFactory, @Nullable HostnameVerifier hostnameVerifier,
      @Nullable CertificatePinner certificatePinner, Authenticator proxyAuthenticator,
      @Nullable Proxy proxy, List<Protocol> protocols, List<ConnectionSpec> connectionSpecs,
      ProxySelector proxySelector) {
    ...
}
```  

Resulting bytecode looks like it's compiled from the source below (note that no checks are generated for parameters marked by *Nullable*):  

```java
public Address(String uriHost, int uriPort, Dns dns, SocketFactory socketFactory,
      @Nullable SSLSocketFactory sslSocketFactory, @Nullable HostnameVerifier hostnameVerifier,
      @Nullable CertificatePinner certificatePinner, Authenticator proxyAuthenticator,
      @Nullable Proxy proxy, List<Protocol> protocols, List<ConnectionSpec> connectionSpecs,
      ProxySelector proxySelector) {
    if (uriHost == null) {
        throw new NullPointerException("uriHost == null");
    }
    if (dns == null) {
        throw new NullPointerException("dns == null");
    }
    if (socketFactory == null) {
        throw new NullPointerException("socketFactory == null");
    }
    if (proxyAuthenticator == null) {
        throw new NullPointerException("proxyAuthenticator == null");
    }
    if (protocols == null) {
        throw new NullPointerException("protocols == null");
    }
    if (connectionSpecs == null) {
        throw new NullPointerException("connectionSpecs == null");
    }
    if (proxySelector == null) {
        throw new NullPointerException("proxySelector == null");
    }
    ...
}
```  

*Bytecode*  

```
javap -c ./okhttp/target/classes/okhttp3/Address.class
...
  public okhttp3.Address(java.lang.String, int, okhttp3.Dns, javax.net.SocketFactory, javax.net.ssl.SSLSocketFactory, javax.net.ssl.HostnameVerifier, okhttp3.CertificatePinner, okhttp3.Authenticator, java.net.Proxy, java.util.List<okhttp3.Protocol>, java.util.List<okhttp3.ConnectionSpec>, java.net.ProxySelector);
    Code:
       0: aload_0
       1: invokespecial #1                  // Method java/lang/Object."<init>":()V
       4: aload_1
       5: ifnonnull     18
       8: new           #2                  // class java/lang/NullPointerException
      11: dup
      12: ldc           #3                  // String uriHost == null
      14: invokespecial #4                  // Method java/lang/NullPointerException."<init>":(Ljava/lang/String;)V
      17: athrow
      18: aload_3
      19: ifnonnull     32
      22: new           #2                  // class java/lang/NullPointerException
      25: dup
      26: ldc           #5                  // String dns == null
      28: invokespecial #4                  // Method java/lang/NullPointerException."<init>":(Ljava/lang/String;)V
      31: athrow
...
```

### DIFF
--- a/okhttp/src/main/java/okhttp3/EventListener.java
+++ b/okhttp/src/main/java/okhttp3/EventListener.java
@@ -95,7 +95,8 @@ public abstract class EventListener {
    *
    * <p>This method is invoked after {@link #dnsStart}.
    */
-  public void dnsEnd(@Nullable Call call, String domainName, @Nullable List<InetAddress> inetAddressList) {
+  public void dnsEnd(@Nullable Call call, String domainName,
+      @Nullable List<InetAddress> inetAddressList) {
   }
 
   /**
--- a/okhttp/src/main/java/okhttp3/EventListener.java
+++ b/okhttp/src/main/java/okhttp3/EventListener.java
@@ -87,7 +87,7 @@ public abstract class EventListener {
    * <p>If the {@link Call} is able to reuse an existing pooled connection, this method will not be
    * invoked. See {@link ConnectionPool}.
    */
-  public void dnsStart(Call call, String domainName) {
+  public void dnsStart(@Nullable Call call, String domainName) {
   }
 
   /**
@@ -95,7 +95,7 @@ public abstract class EventListener {
    *
    * <p>This method is invoked after {@link #dnsStart}.
    */
-  public void dnsEnd(Call call, String domainName, @Nullable List<InetAddress> inetAddressList) {
+  public void dnsEnd(@Nullable Call call, String domainName, @Nullable List<InetAddress> inetAddressList) {
   }
 
   /**
--- a/pom.xml
+++ b/pom.xml
@@ -141,6 +141,17 @@
             <forceJavacCompilerUse>true</forceJavacCompilerUse>
             <source>${java.version}</source>
             <target>${java.version}</target>
+            <compilerArgs>
+              <arg>-Xplugin:Traute</arg>
+              <arg>-Atraute.failure.text.parameter=$${PARAMETER_NAME} == null</arg>
+            </compilerArgs>
+            <annotationProcessorPaths>
+              <path>
+                <groupId>tech.harmonysoft</groupId>
+                <artifactId>traute-javac</artifactId>
+                <version>1.1.6</version>
+              </path>
+            </annotationProcessorPaths>
           </configuration>
           <dependencies>
             <dependency>
--- a/okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
+++ b/okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
@@ -126,7 +126,6 @@ public final class HttpLoggingInterceptor implements Interceptor {
 
   /** Change the level at which this interceptor logs. */
   public HttpLoggingInterceptor setLevel(Level level) {
-    if (level == null) throw new NullPointerException("level == null. Use Level.NONE instead.");
     this.level = level;
     return this;
   }
--- a/okhttp-logging-interceptor/src/test/java/okhttp3/logging/HttpLoggingInterceptorTest.java
+++ b/okhttp-logging-interceptor/src/test/java/okhttp3/logging/HttpLoggingInterceptorTest.java
@@ -102,7 +102,7 @@ public final class HttpLoggingInterceptorTest {
       applicationInterceptor.setLevel(null);
       fail();
     } catch (NullPointerException expected) {
-      assertEquals("level == null. Use Level.NONE instead.", expected.getMessage());
+      assertEquals("level == null", expected.getMessage());
     }
   }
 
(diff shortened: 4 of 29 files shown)


okhttp/src/main/java/okhttp3/EventListener.java
```diff
@@ -87,15 +87,15 @@ public void callStart(Call call) {
    * <p>If the {@link Call} is able to reuse an existing pooled connection, this method will not be
    * invoked. See {@link ConnectionPool}.
    */
-  public void dnsStart(Call call, String domainName) {
+  public void dnsStart(@Nullable Call call, String domainName) {
```
yschimke: Why are these now Nullable?

pom.xml
```diff
@@ -141,6 +141,17 @@
...
             <source>${java.version}</source>
             <target>${java.version}</target>
+            <compilerArgs>
+              <arg>-Xplugin:Traute</arg>
+              <arg>-Atraute.failure.text.parameter=$${PARAMETER_NAME} == null</arg>
+            </compilerArgs>
+            <annotationProcessorPaths>
+              <path>
+                <groupId>tech.harmonysoft</groupId>
+                <artifactId>traute-javac</artifactId>
```
yschimke: Docs seem to suggest this will mean compiling only with JDK8+

https://github.com/denis-zhdanov/traute/blob/master/core/javac/README.md

I'm not certain this is a problem, but certainly could be for any downstream build processes.

yschimke:
I like this change, but I don't know if it causes problems for the build tooling around the project.

okhttp/src/main/java/okhttp3/EventListener.java
denis-zhdanov: 1. There is a package-level *ParametersAreNonnullByDefault* for that class' package, so, *null*-checks are generated for all method parameters not marked by *Nullable*
2. `mvn clean package` shows that *null* is given there (at least in tests), hence, we get an *NPE*

pom.xml
denis-zhdanov: Yes, *Traute* is a *javac* plugin and corresponding *API* is available only since *java8*.  

There are clients who use something like `javac8 -source 1.7 -target 1.7` and *Traute* works perfectly for them. It doesn't work if you use *javac7* though.

denis-zhdanov:
@yschimke, thanks for the feedback!  

What type of build tooling problem do you have in mind?

yschimke:
It breaks the current status minimum requirements

"OkHttp supports Android 2.3 and above. For Java, the minimum requirement is 1.7."

But it works in travis, so it might be ok.  Not sure what other builds derive off this project, or if everyone consumes the built jars.

denis-zhdanov:
It depends on how do you compile the project - if that is done like `javac8 -source 1.7 -target 1.7` (which is the case for the *CI* - *java version "1.8.0_151"*), then everything is fine because the generated bytecode doesn't contain any java8+ - specific instructions). I.e. *javac8* tool is just used to generate *java7* bytecode and we enhance it in a way to generate additional *null*-checks.

yschimke:
Yep, understood.  I'll let @swankjesse or @JakeWharton answer definitively on this PR

pom.xml
yschimke: We should fix those tests

okhttp/src/main/java/okhttp3/EventListener.java
yschimke: We should fix those tests then

denis-zhdanov:
Thanks!

denis-zhdanov:
@swankjesse, thanks for the feedback!  
A couple of notes regarding that:  
* we're not just replacing existing explicit checks, we enforce existing nullability contract defined through the package-level annotations. That's the main benefit from my point of view
* am I right assuming that *OkHttp* is used as a library dependency, i.e. it's not built from source in, say, *AOSP*?

yschimke:
I assume it is this https://android.googlesource.com/platform/external/okhttp/

And built from source

denis-zhdanov:
Yeah, thanks, looks like it's built from the same maven setup, i.e. there shouldn't be a problem.  

My vision is that enforcing *null*-checks is a good addition to the project but of course it's up to you, guys, to decide though :)

yschimke:
@denis-zhdanov I assume it's in bazel or something similar e.g. https://android.googlesource.com/platform/external/okhttp/+/master/Android.bp

Is it worth splitting this PR into 2? one just focusing on using @Nullable correctly at all sites including tests.  Second one probably won't ship for using traute

denis-zhdanov:
@yschimke, I can do that if the change is going to be accepted :)

JakeWharton:
.bp is Blueprint which is an abstraction on Soong which is an abstraction on Ninja. Yay Android builds!

yschimke:
@denis-zhdanov Well, I expect that correcting the @nyllable sites will, but that the second won't be.

yschimke:
closing in it's current form, doesn't seem likely to land in this form.

Issue #3683, opened 2017-11-14
Adding QUIC entry to Protocol enum


ypyu2000:
We are adding QUIC (https://www.chromium.org/quic) support to the OKHttp by adding a QUIC interceptor. The QUIC interceptor can transmit request packets and response packets over QUIC between the OKHttp client and a QUIC proxy or a QUIC server. To integrate QUIC into existing okhttp codes, we plan to add QUIC(“quic”) to the Protocol enum in Protocol.java to be used in the QUIC interceptor. We can contribute back the QUIC interceptor to the OKHttp git if this addition can be approved. Can the committee approve this addition? Thanks.

swankjesse:
Sure. Who is “we” ?

naturalwarren:
The networking team at Uber developed the QUIC interceptor. It's pretty nifty! Not sure whether this kind of thing would live inside or outside of OkHttp. Instead of having the interceptor live inside OkHttp we could just add an API to support custom protocols. 

Once we have the Interceptor hosted on Git we'll ping this thread.

vinothchandar:
+ 1 for support for custom protocol api :) That would be awesome..

cfsbhawkins:
Where could one find this QUIC interceptor?

vinothchandar:
@cfsbhawkins Not open sourced as of now.

swankjesse:
Done!

esafirm-tiket:
Is the QUIC interceptor open sourced yet? Or can anyone help me find some resource to build one? Thanks

yschimke:
See Cronet example which should support Quic https://medium.com/the-react-native-log/using-cronet-in-your-mobile-app-7dda3a89c132

PR #3746, opened 2017-12-28
QUIC constant

yschimke:


yschimke:
Related to https://github.com/square/okhttp/issues/3683

In case we want to lang this in a next release, so interceptor doesn't have to wait for potentially another month.

But no rush either.

okhttp/src/main/java/okhttp3/Protocol.java
```diff
@@ -78,12 +85,15 @@ public static Protocol get(String protocol) throws IOException {
     if (protocol.equals(HTTP_1_1.protocol)) return HTTP_1_1;
     if (protocol.equals(HTTP_2.protocol)) return HTTP_2;
     if (protocol.equals(SPDY_3.protocol)) return SPDY_3;
+    if (protocol.equals(QUIC.protocol)) return QUIC;
```
swankjesse: I wonder if we need this. This method is used for ALPN, and QUIC doesn't use ALPN. In particular, the string "quic" isn't used on the we wire anywhere, is it?

okhttp/src/main/java/okhttp3/Protocol.java
```diff
@@ -78,12 +85,15 @@ public static Protocol get(String protocol) throws IOException {
...
     throw new IOException("Unexpected protocol: " + protocol);
   }
 
   /**
    * Returns the string used to identify this protocol for ALPN, like "http/1.1", "spdy/3.1" or
    * "h2".
+   *
+   * @link https://www.iana.org/assignments/tls-extensiontype-values/tls-extensiontype-values.xhtml
```
swankjesse: Neat. I'm surprised spdy/3.1 is absent.

okhttp/src/main/java/okhttp3/Protocol.java
yschimke: It's useful as API

https://github.com/yschimke/oksocial/blob/master/src/main/kotlin/com/baulsupp/oksocial/commands/CommandLineClient.kt#L450

swankjesse: Works for me.

Issue #3619, opened 2017-09-20
WebSocket returns false when sending large size string


hasretsariyer:
I try to send a string by websocket. The size of string is 18672379. send() method retuns false in this case. Sometimes, the socket is closed and onClose callback triggers.

If I send half of this string, it return true. Is there any limitation for sending string? If yes, how can I send large string?

My request is below:

`

     clientBuilder = new OkHttpClient.Builder();
     client = clientBuilder.build();
     request = new OkHttpRequest.Builder().url(_url).build();
     client.newWebSocket(request, listener);
`

swankjesse:
We don't allow more than 16 MiB of data to be enqueued on a websocket. This is a performance tradeoff. The assumption is that if you’re enqueueing that much then our websocket API is not a good fit for your use case.

hasretsariyer:
Ok. Is it documented? I searched a limitation of sending data for websocket. I couldn't find it.

swankjesse:
> “Messages that would overflow the outgoing message buffer will be rejected and trigger a graceful shutdown of this web socket.”

We do need to document the size of that buffer.

mrexodia:
Is there any way to change this limit? 16mib is very low...

swankjesse:
No way to change the limit at the moment. What’s your use case?

mrexodia:
My use case is sending state updates, where sometimes the state can get rather large (up to a few hundred mb depending on how out-of-sync the client and server are).

I honestly don't really see a good reason for any limit, since websocket doesn't have a limit and the protocol supports chunking/compression. Additionally the API here is super inconsistent, why would the client decide to disconnect if a limit is triggered instead of throwing some kind of exception so it can be handled gracefully?

For now I'll just patch the `MAX_QUEUE_SIZE` locally and see if that solves my issue. If you want I can add a `setMaxQueueSize` to the `WebSocket` class though, should just be a few lines and allow people to set their own limits.

swankjesse:
The limit is there because we haven’t done the Big Complicated Work to make our web sockets implementation work efficiently for large messages or large buffers.

In particular I made the design choice to not implement streaming of individual messages. Our messages are buffered in memory and that’s all we support! We don’t have mechanisms to reprioritize queued messages when there’s a big backlog. As is you can manually set your own limit, but you might find our code doesn’t handle a big backlog well.

mrexodia:
Thanks for the update, will see if things work out and report back at what point it gets unstable.

[...]

PR #3748, opened 2017-12-28
Document max websocket buffer

yschimke:


yschimke:
relates to https://github.com/square/okhttp/issues/3619

swankjesse:
I like. I think this might commit us to 16 MiB, which is fine. If we ever want to change that we'll make it user-configurable or something.

Issue #3670, opened 2017-10-30
Http2Connection does not handle integer overflow for stream id


nziebart:
We have a long-running service, and there was a period of about 1 minute where our okhttp client repeatedly threw this exception when attempting to make a remote call:

```
java.lang.IllegalArgumentException: reserved bit set: -2147483613
at okhttp3.internal.http2.Http2.illegalArgument(Http2.java:109)
at okhttp3.internal.http2.Http2Writer.frameHeader(Http2Writer.java:270)
at okhttp3.internal.http2.Http2Writer.headers(Http2Writer.java:306)
at okhttp3.internal.http2.Http2Writer.synStream(Http2Writer.java:127)
at okhttp3.internal.http2.Http2Connection.newStream(Http2Connection.java:230)
at okhttp3.internal.http2.Http2Connection.newStream(Http2Connection.java:205)
at okhttp3.internal.http2.Http2Codec.writeRequestHeaders(Http2Codec.java:106)
at okhttp3.internal.http.CallServerInterceptor.intercept(CallServerInterceptor.java:46)
at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:92)
at okhttp3.internal.connection.ConnectInterceptor.intercept(ConnectInterceptor.java:45)
```

After a few failures over the course of 1 minute, the client started working again. 

My guess is that we are overflowing an integer here, as the very first value that causes the error is `-2147483647` which is the minimum odd value for a signed integer.

The number (I believe it's the stream id) increased by 2 with each occurrence of the error - below are the stream ids from the last 5 failures before calls began succeeding again:
```
-2147483621
-2147483619
-2147483617
-2147483615
-2147483613
```

nziebart:
Looking at this more, it looks like we never check if `nextStreamId` will exceed `MAX_INT`. We probably need to detect this scenario and create a new connection.

swankjesse:
Wow, that’s impressive. I think we should put something in OkHttp to trigger a new connection when we cross, say, 50% of the threshold.

swankjesse:
This has been fixed!

PR #3752, opened 2017-12-30
Reset stream after a rollover

yschimke:


yschimke:
Unclear how to effectively test, Tested manually via starting at Integer.MAX_VALUE - 4

Three successive requests, showing the 3rd silently retrying.

image

image

image

yschimke:
relates to https://github.com/square/okhttp/issues/3670

yschimke:
I'm suspecting this isn't a full fix, but as I think it will tear down the connection in RetryAndFollowUpInterceptor

yschimke:
@swankjesse It seems better if StreamAllocation doesn't become aware of stream id logic.  Assuming this part stays in Http2Connection this seems exactly what shutdown is meant to achieve.

```
  /**
   * Degrades this connection such that new streams can neither be created locally, nor accepted
   * from the remote peer. Existing streams are not impacted. This is intended to permit an endpoint
   * to gracefully stop accepting new requests without harming previously established streams.
   */
  public void shutdown(ErrorCode statusCode) throws IOException {
```

yschimke:
@swankjesse I put up an [updated diff](https://github.com/square/okhttp/pull/3752/commits/cdf3228ee3a4f7ab0c7d57a7cac19bcc45720bc3), but do you want it moved to Stream Allocation still? To me that is awkward because is put HTTP/2 logic in that shared class.

swankjesse:
Yeah, I like this.

Issue #3665, opened 2017-10-25
OKHttp retry


jeffreyfan8:
When OKHTTP retries because of a 408, it does not respect the "Retry-After" header.  It looks like this is a standard header.  Could we add in this feature to respect the "Retry-After" header when retrying.

Thanks.

JakeWharton:
I would think if this value is non-zero we should be returning the response
directly and push the work of deciding to retry onto the application layer.

yschimke:
n.b. not clear that 408 with Retry-After is an official or common thing, but seems implicit.

PR #3753, opened 2017-12-30
Retry-After support 

yschimke:
Two effective changes 

- Support for 503, with 0 Retry-After meaning retry automatically
- Check with 408, and don't do automatic retry if there is a delay


yschimke:
relates to https://github.com/square/okhttp/issues/3665

yschimke:
One question is whether Retry-After should be more broadly supported, apart from these two cases

swankjesse:
Yeah, I could definitely see us supporting it for 4xx and 5xx responses generally. What do browsers do?

Regardless, this PR is great.

yschimke:
This PR is mostly me copying and pasting @dave-r12 code from a previous PR

Issue #3597, opened 2017-09-11
NullPointerException in HttpUrl.class


MS893:
IT IS NOT REALLY A BUG, BUT...

There is no check if a value is null whille buiding a body for a POST
`            RequestBody formBody = new FormBody.Builder()
                    .add("value", value)
                    .build();
`
if value is null, there is a :
java.lang.NullPointerException: Attempt to invoke virtual method 'int java.lang.String.length()' on a null object reference at okhttp3.HttpUrl.canonicalize(HttpUrl.java:1751)

My app doesn't stop after this crash and remains in a loop.

Would it be possible to throw a NullPointerException ?

swankjesse:
Yeah, definitely. Wanna send a pull request?

MS893:
Yes I do

yschimke:
Was this PR related to this? https://github.com/square/okhttp/pull/3598

swankjesse:
This was fixed.

MS893:
++++++++++ !

MarkRaoAndroid:
Okhttp request parameters can not be empty, check your parameters

PR #3755, opened 2018-01-01
NPE earlier in FormBody.Builder

yschimke:


yschimke:
relates to https://github.com/square/okhttp/issues/3597

PR #3762, opened 2018-01-05
OkHostnameVerifier: Don't fall back to CN verification.

15characterlimi:
The use of Common Name was deprecated in RFC 2818 (May 2000), section 3.1:

  Although the use of the Common Name is existing practice, it is
  deprecated and Certification Authorities are encouraged to use the
  dNSName instead.

In 2017, Chrome 58, Firefox 48, and Opera 45 web browsers removed this
fallback, with Chrome leaving it configurable for enterprise deployments
(see https://www.chromestatus.com/feature/4981025180483584).
Android is removing it in http://r.android.com/581382 .


15characterlimi:
Note: I haven't run any tests yet - I'm relying on your Travis integration to tell me whether the tests succeed. Hold on for their completion :)

15characterlimi:
Also, the test cases for CN now look a little weird, since they all say assertFalse(). I was considering extracting a helper method that gets a boolean parameter but ignores it, or just throwing those test cases away. Let me know your preference for how to address this.

15characterlimi:
Closing this pull request. I need to figure out how to fix ClientAuthTest, which currently generates certs with only a CN. Will reopen the pull request once I have figured this out (probably next week).

Issue #3628, opened 2017-09-25
100 (Continue) HTTP response should be handled gracefully regardless of whether it was requested/expected


stephentalley:
I'm using OkHttp 3.8.1 on Android to POST to a poorly-behaved server.

Even though the request does not include an "Expect: 100-continue" header, the web server returns a 100 Continue response. Rather than continuing to send the request body (or just ignoring the 100 response since the body had already been sent), then getting the actual (200) response, OkHttp stops there and sends back the 100 Continue response in my okhttp.Callback.

The [w3c definition](https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.1.1) of the 100 status code states:

> The client SHOULD continue by sending the remainder of the request or, if the request has already been completed, ignore this response.

and:

> A client MUST be prepared to accept one or more 1xx status responses prior to a regular response, even if the client does not expect a 100 Continue status message.

I have verified the correct behavior on three other HTTP clients (iOS, Windows, and curl(1)), who correctly ignore/continue when receiving an unexpected 100 Continue response.

In contrast, it looks like OkHttp only handles the 100 response iff it was explicitly requested via an "Expect: 100-continue" header.  From CallServerInterceptor.java:

    // If there's a "Expect: 100-continue" header on the request, wait for a "HTTP/1.1 100
    // Continue" response before transmitting the request body. If we don't get that, return
    // what we did get (such as a 4xx response) without ever transmitting the request body.
    if ("100-continue".equalsIgnoreCase(request.header("Expect"))) {
        httpCodec.flushRequest();
        realChain.eventListener().responseHeadersStart(realChain.call());
        responseBuilder = httpCodec.readResponseHeaders(true);
    }

Incidentally, the obvious workaround (including "Expect: 100-continue" in the request to trigger OkHttp's logic) didn't work for me because this particular server is further broken in that it complains about a mangled Expect: header. :expressionless:  It doesn't appear any other workaround is possible, though I'd love to be proven wrong.

See: https://stackoverflow.com/a/46382361/920849

stephentalley:
The key takeaway is that a server can choose to return 100 when the client doesn't request it, or return a non-100 response when even when the client does request it.

yschimke:
@stephentalley Thanks for the detailed writeup.  It sounds like a legitimate bug. Given the rarity of it and it being triggered by a badly behaving server, it *may* not be at the top of the pile to work on.  So if it's really impacting you we would love a PR and I'll happily review quickly.

[...]

PR #3766, opened 2018-01-06
[WIP] Handle 100 (Continue) HTTP response gracefully regardless of whether it was requested/expected

johan12345:
I created a failing test for issue #3628 (first commit) - or rather changed the previous `serverRespondsWithUnsolicited100Continue` test that asserts that the 100 response would just be returned to the user. I also tried to implement a solution, i.e. make OkHttp send the response body again if a response with code 100 was received (second commit).

When testing it with a real-world server in our app where the problem previously occurred, the issue also seems to be fixed.

**However**, the next HTTP request (in this case it's a GET) after the request where the server sent the 100 response gets a 403 Forbidden error. I suspect this has something to do with the HTTP connection being in an inconsistent state so that it cannot be reused. But I don't really understand why this should be the case here (the client has sent the request body and read the response, so the request should in principle be done, right?). Does someone have an idea about this?

/cc @yschimke


okhttp/src/main/java/okhttp3/internal/http1/Http1Codec.java
```diff
@@ -196,6 +196,9 @@ public void writeRequest(Headers headers, String requestLine) throws IOException
 
       if (expectContinue && statusLine.code == HTTP_CONTINUE) {
         return null;
+      } else if (statusLine.code == HTTP_CONTINUE) {
```
yschimke: I don't understand the need for this branch since it duplicates the lines below

okhttp/src/main/java/okhttp3/internal/http/CallServerInterceptor.java
```diff
@@ -99,11 +90,31 @@ public CallServerInterceptor(boolean forWebSocket) {
...
+
+      code = response.code();
+    }
+
     if (forWebSocket && code == 101) {
       // Connection is upgrading, but we need to ensure interceptors see a non-null response body.
       response = response.newBuilder()
-          .body(Util.EMPTY_RESPONSE)
-          .build();
+              .body(Util.EMPTY_RESPONSE)
```
yschimke: nit: inconsistent with formatting 3 lines below

okhttp/src/main/java/okhttp3/internal/http1/Http1Codec.java
johan12345: It sets the state to a different one (STATE_OPEN_REQUEST_BODY instead of STATE_OPEN_RESPONSE_BODY). Otherwise, sending the response body again would not work.

okhttp/src/main/java/okhttp3/internal/http/CallServerInterceptor.java
```diff
@@ -99,11 +90,31 @@ public CallServerInterceptor(boolean forWebSocket) {
         .responseHeadersEnd(realChain.call(), response);
 
     int code = response.code();
+    if (code == 100) {
+      // server sent a 100-continue even though we did not request one. Send the request body again.
```
yschimke: This doesn't seem to match the RFC which states

> The client ought to continue sending the request and discard the 100 response.

Specifically, why are we resending the request body? It seems like we should ignore the 100 and just try to read a later HTTP response?

okhttp/src/main/java/okhttp3/internal/http1/Http1Codec.java
yschimke: It's early, sorry!

okhttp/src/main/java/okhttp3/internal/http/CallServerInterceptor.java
johan12345: Good point. I will try if it works without resending the body.

johan12345:
@yschimke You're right, it was not necessary to re-send the request body, just to read the response again. I updated the PR accordingly.

In the case that the server just sends a 100 response, but not the actual response, the request now times out. I changed my commit to the CallTest so that this behavior is also tested.

johan12345:
Is there anything else I need to do here or can this be merged? Also, do you plan to release a new OkHttp version soon or should we build a JAR ourselves to fix this problem in our app in the meantime?

okhttp/src/main/java/okhttp3/internal/http/CallServerInterceptor.java
```diff
@@ -99,6 +99,25 @@ public CallServerInterceptor(boolean forWebSocket) {
         .responseHeadersEnd(realChain.call(), response);
 
     int code = response.code();
+    if (code == 100) {
```
swankjesse: In follow up we should probably check the entire range.

okhttp/src/main/java/okhttp3/internal/http/CallServerInterceptor.java
```diff
@@ -99,6 +99,25 @@ public CallServerInterceptor(boolean forWebSocket) {
         .responseHeadersEnd(realChain.call(), response);
```
swankjesse: I'm pondering the consequences of calling this 2x. It's going to crash listeners.

Can you change your fix to go before this listener call back? And we won't call back on the headers for the unsolicited 100; it's not worth the trouble.

swankjesse:
Sorry for the delay. I've been busy.

okhttp/src/main/java/okhttp3/internal/http/CallServerInterceptor.java
johan12345: So you mean allowing this for every 1xx status code instead of just 100? Yes, that should be possible.

okhttp/src/main/java/okhttp3/internal/http/CallServerInterceptor.java
johan12345: Okay, I changed this accordingly in 785e1d4

okhttp/src/main/java/okhttp3/internal/http/CallServerInterceptor.java
```diff
@@ -95,10 +95,26 @@ public CallServerInterceptor(boolean forWebSocket) {
         .receivedResponseAtMillis(System.currentTimeMillis())
         .build();
 
+    int code = response.code();
+    if (code == 100) {
+      // server sent a 100-continue even though we did not request one.
+      // try again to read the actual response
+      realChain.eventListener().responseHeadersStart(realChain.call());
```
swankjesse: we still have unbalanced calls to responseHeadersStart(). Mind dropping this call too?

johan12345: Oh, right. Fixed in 440531d.

yschimke:
@johan12345 did you fill in this contributor agreement?

https://github.com/square/okhttp/wiki/Contributing

johan12345:
@yschimke Yes, I think I did this already shortly after submitting the PR.

PR #3770, opened 2018-01-10
Fix locks

aaronjwood:
I came across a few instances of locking that I thought could be cleaned up a bit. There was one case where it looked like race conditions could still occur (the null check before synchronizing) so I adjusted the logic a bit there.


aaronjwood:
Looks like `stream` does need to be synchronized due to use the of monitors in that class so my change here is breaking the build https://github.com/square/okhttp/pull/3770/commits/f54726a10e9af4a38bb22808c45864fd32f8780c

I figured you'd want the null check inside of the synchronized block there. Can we just change `addBytesToWriteWindow` so that the call to `notifyAll` is wrapped inside a synchronized block?

swankjesse:
I don’t think these changes are safe. I trust you’ve ready our concurrency doc?
https://github.com/square/okhttp/wiki/Concurrency

aaronjwood:
Thanks, I missed that the first time around. Sounds like what I'm doing here is going against the concurrency model you guys have in place. I'll close this out.

PR #3776, opened 2018-01-13
Allow adding custom HTTP methods with specific behavior

cketti:
This allows adding custom HTTP methods to an `OkHttpClient` instance. See issue #229.

```java
OkHttpClient client = new OkHttpClient.Builder()
    .addMethod("CUSTOM", new HttpMethodBehavior.Builder()
        .invalidatesCache(false)
        .requiresRequestBody(false)
        .permitsRequestBody(true)
        .redirectsWithBody(true)
        .redirectsToGet(false)
        .build())
    .build();
```

Currently the behavior of (supported) methods is encoded in [`HttpMethod`](https://github.com/square/okhttp/blob/cd84d79c3574de81f23802f832f2ead6ad38a735/okhttp/src/main/java/okhttp3/internal/http/HttpMethod.java). With this change the behavior can be specified using `HttpMethodBehavior`. Mapping a method name to the corresponding behavior is done by `HttpMethods`.
Right now `Request.Builder.method()` checks if a method is allowed or required to provide a request body. With this change the check is deferred until the `Request` is used to create a new `Call` instance and we have access to the configured `HttpMethods`. Aside from that, this pull requests shouldn't introduce any behavior changes.

After working on this for a while I'm not convinced that much flexibility is really necessary. I'm guessing 99% of people running into issues with the current behavior would be helped if request bodies were allowed for all methods. Please don't feel bad for closing this pull request in favor of a simpler solution 😄


swankjesse:
Sure let's do the simpler one. Wanna send another?

cketti:
See #3777

PR #3777, opened 2018-01-14
Allow request bodies for all HTTP methods except GET and HEAD

### LINKED ISSUE #229: Support custom HTTP methods with bodies.

cketti:
Fixes #229

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/CallTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/CallTest.java
@@ -550,6 +550,25 @@ public final class CallTest {
     patch();
   }
 
+  @Test public void customMethodWithBody() throws Exception {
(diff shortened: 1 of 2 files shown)


swankjesse:
👍

PR #3788, opened 2018-01-19
Build Provider through the Platform

yschimke:
The SSL Provider may not always be the default one e.g. for Android (GMS) or Conscrypt, so move this logic in to the internal Provider API.

### DIFF
--- a/mockwebserver/src/main/java/okhttp3/internal/tls/SslClient.java
+++ b/mockwebserver/src/main/java/okhttp3/internal/tls/SslClient.java
@@ -80,7 +80,7 @@ public final class SslClient {
     private final List<X509Certificate> certificates = new ArrayList<>();
     private KeyPair keyPair;
     private String keyStoreType = KeyStore.getDefaultType();
-    private SSLContext sslContext = null;
+    private SSLContext sslContext;
 
     /**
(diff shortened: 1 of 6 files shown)


mockwebserver/src/main/java/okhttp3/internal/tls/SslClient.java
```diff
@@ -79,6 +80,7 @@ public static synchronized SslClient localhost() {
     private final List<X509Certificate> certificates = new ArrayList<>();
     private KeyPair keyPair;
     private String keyStoreType = KeyStore.getDefaultType();
+    private SSLContext sslContext = null;
```
swankjesse: nit: the `= null` has weird consequences and shouldn’t be done. The most surprising thing about it is that it shows up in the bytecode.

```
class X {
  String a = defaultValueOfA();
  String b = null;

  static String defaultValueOfA();
    b = "B";
    return "A";
  }
}
```

Regular programs don’t do this. But sometimes we accidentally write code that has side-effects like these. The `= null` doesn’t run when you’d expect it to, and so it yields surprising code.

My preference is to never explicitly assign fields to their natural default values (0, null, false).

PR #3790, opened 2018-01-19
optimize and add 'private'

jokermonn:
optimize List and add 'private'


swankjesse:
I don't think this does anything.

Issue #3780, opened 2018-01-17
Logging Gzipped responses in HttpLoggingInterceptor as a NetworkInterceptor


mattsheppard:
I'd like to be able to have HTTP Logging which includes redirect requests (i.e. it must be a networkInterceptor) and also logs the uncompressed version of any gzipped response bodies, which HttpLoggingInterceptor does not do if it's being used as a networkInterceptor.

If I were to implement support for gunzipping and logging response bodies when HttpLoggingInterceptor is used as a networkInterceptor is that a change you'd likely merge, or am I better off just forking my own HttpLoggingInterceptor?

Note: I saw https://github.com/square/okhttp/issues/1955 was a similar request, but it seems to have been closed because the original reporter was happy to run it a normal interceptor to get the gzipped body logged.

JakeWharton:
I think I'm okay with this (as a small part of the OkHttp team). The
biggest problem is that it adds more complexity, but I have convinced
myself it's worthwhile as I usually install the logging interceptor as a
network interceptor and miss out on transparent gzip-compressed responses.
I'd like to see both the compressed and uncompressed size in the log and
maybe even a percentage difference between the two.

I'd wait for Jesse to chime in though before starting any PR.

swankjesse:
Are the bodies of redirects ever interesting? In either case, sure, please send a PR.

mattsheppard:
> Are the bodies of redirects ever interesting

No, probably not - Maybe it's clearer if I say...

1. If I install HttpLoggingInterceptor as a networkInterceptor currently, I see the redirects but not gzipped body of the last (or any) response.

2. If I install HttpLoggingInterceptor as an interceptor currently, I see the decompressed version of the gzipped body of any responses which have them but no redirects.

I want the best of both worlds - Redirects and gzipped bodies both logged. I could install HttpLoggingInterceptor twice, but the resulting interleaving (or having two seperate logs) makes it hard to follow.

Anyhow, I've got a hacked up copy of HttpLoggingInterceptor which does the the gunzipping already, so I'll try to clean it up and submit it soon!

swankjesse:
Done!

PR #3792, opened 2018-01-21
Log gzipped bodies when HttpLoggingInterceptor is used as a NetworkInterceptor

mattsheppard:
Makes HttpLoggingInterceptor decompress and log the body of a response which is gzip encoded when used as a networkInterceptor (where previously it just logged that an encoded body was seen).

This is useful in the case where you want to log both any redirects which occur during the call and the response bodies. 

Fixes issue #3780

(My first time touching any of the okio stuff directly - very happy to hear if I've missed a better way to do something)

### DIFF
--- a/okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
+++ b/okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
@@ -248,7 +248,7 @@ public final class HttpLoggingInterceptor implements Interceptor {
             try {
                 gzippedResponseBody = new GzipSource(buffer.clone());
                 buffer = new Buffer();
-                gzippedResponseBody.read(buffer, Long.MAX_VALUE);
+                buffer.writeAll(gzippedResponseBody);
             } finally {
                 if (gzippedResponseBody != null) {
                     gzippedResponseBody.close();
--- a/okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
+++ b/okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
@@ -243,17 +243,17 @@ public final class HttpLoggingInterceptor implements Interceptor {
 
         Long gzippedLength = null;
         if ("gzip".equalsIgnoreCase(headers.get("Content-Encoding"))) {
-            gzippedLength = buffer.size();
-            GzipSource gzippedResponseBody = null;
-            try {
-                gzippedResponseBody = new GzipSource(buffer.clone());
-                buffer = new Buffer();
-                buffer.writeAll(gzippedResponseBody);
-            } finally {
-                if (gzippedResponseBody != null) {
-                    gzippedResponseBody.close();
-                }
+          gzippedLength = buffer.size();
+          GzipSource gzippedResponseBody = null;
+          try {
+            gzippedResponseBody = new GzipSource(buffer.clone());
+            buffer = new Buffer();
+            buffer.writeAll(gzippedResponseBody);
+          } finally {
+            if (gzippedResponseBody != null) {
+              gzippedResponseBody.close();
             }
+          }
         }
 
         Charset charset = UTF8;
--- a/okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
+++ b/okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
@@ -32,6 +32,7 @@ import okhttp3.internal.http.HttpHeaders;
 import okhttp3.internal.platform.Platform;
 import okio.Buffer;
 import okio.BufferedSource;
+import okio.GzipSource;
 
 import static okhttp3.internal.platform.Platform.INFO;
 
@@ -182,7 +183,7 @@ public final class HttpLoggingInterceptor implements Interceptor {
 
       if (!logBody || !hasRequestBody) {
         logger.log("--> END " + request.method());
-      } else if (bodyEncoded(request.headers())) {
+      } else if (bodyHasUnknownEncofing(request.headers())) {
         logger.log("--> END " + request.method() + " (encoded body omitted)");
       } else {
         Buffer buffer = new Buffer();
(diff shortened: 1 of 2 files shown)


okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
```diff
@@ -233,13 +234,28 @@ public Level getLevel() {
...
         Buffer buffer = source.buffer();
 
+        Long gzippedLength = null;
+        if ("gzip".equalsIgnoreCase(headers.get("Content-Encoding"))) {
+            gzippedLength = buffer.size();
+            GzipSource gzippedResponseBody = null;
+            try {
+                gzippedResponseBody = new GzipSource(buffer.clone());
+                buffer = new Buffer();
+                gzippedResponseBody.read(buffer, Long.MAX_VALUE);
```
mattsheppard: I think because of the source.request above, the read here is guaranteed to get everything, but it would be good to have confirmation of that from someone (or a pointer to a better way to go about it).

okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
```diff
@@ -182,7 +183,7 @@ public Level getLevel() {
 
       if (!logBody || !hasRequestBody) {
         logger.log("--> END " + request.method());
-      } else if (bodyEncoded(request.headers())) {
+      } else if (bodyHasUnknownEncofing(request.headers())) {
```
JakeWharton: nit: spelling

okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
JakeWharton: It is, but I don't think it's safe to rely on. Maybe try `buffer.writeAll(gzippedResponseBody)` instead?

okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
```diff
@@ -257,7 +273,12 @@ public Level getLevel() {
           logger.log(buffer.clone().readString(charset));
         }
 
-        logger.log("<-- END HTTP (" + buffer.size() + "-byte body)");
+        if (gzippedLength != null) {
+            logger.log("<-- END HTTP (" + buffer.size() + "-byte "
```
JakeWharton: Can you add a comma after "byte"?

mattsheppard: Yep - Done.

okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
mattsheppard: Oops - Thanks - Fixed.

okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
mattsheppard: Ah, that's what I'd been searching for on the source :)

Great - Done.

okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
```diff
@@ -233,13 +234,28 @@ public Level getLevel() {
 
       if (!logBody || !HttpHeaders.hasBody(response)) {
         logger.log("<-- END HTTP");
-      } else if (bodyEncoded(response.headers())) {
+      } else if (bodyHasUnknownEncoding(response.headers())) {
         logger.log("<-- END HTTP (encoded body omitted)");
       } else {
         BufferedSource source = responseBody.source();
```
yschimke: Can this be simplified using response.peekBody earlier?  And also logging the bytes in the unknown encoding case?

mattsheppard: I'll check it out and see if it improves things.

Re logging the bytes in the unknown case - Sure - You mean the count of them, not the encoded (and presumably base64d) raw content, correct?

yschimke: yep, nothing blocking from me. Just asking the questions.

mattsheppard: Gave it a shot - https://github.com/mattsheppard/okhttp/pull/1/files - I'm not convinced it's better, but maybe I'm missing the point of your idea.

Happy to just add the logging number of bytes in the unknown case if you like in any case.

okhttp-logging-interceptor/src/test/java/okhttp3/logging/HttpLoggingInterceptorTest.java
```diff
@@ -532,7 +532,7 @@ private void bodyGetNoBody(int code) throws IOException {
         .assertNoMoreLogs();
   }
 
-  @Test public void bodyResponseNotIdentityEncoded() throws IOException {
+  @Test public void bodyResponseGzipEncoded() throws IOException {
```
yschimke: Can you also check the normal response body is still valid in this test.  It's kind of the most critical thing for this test.

mattsheppard: Yep, good point - Done.

mattsheppard:
Just noticed the Travis build here failed - My quick reading is that it's not related to this change, and at a glance the latests builds on the upstream master are failing in the same way (https://travis-ci.org/square/okhttp/builds/334028584).

If I'm wrong and it's something I need to look into further, please let me know.

yschimke:
Yep - each JDK8 release requires matching to an alpn-boot version.  Hopefully the new 161 & 162 are still the old alpn-boot versions

https://github.com/square/okhttp/pull/3803

yschimke:
Passing now

okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
```diff
@@ -257,7 +273,12 @@ public Level getLevel() {
           logger.log(buffer.clone().readString(charset));
         }
 
-        logger.log("<-- END HTTP (" + buffer.size() + "-byte body)");
+        if (gzippedLength != null) {
+            logger.log("<-- END HTTP (" + buffer.size() + "-byte, "
+                + gzippedLength + "-gzipped-byte body)");
```
swankjesse: I love this

okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
```diff
@@ -257,7 +273,12 @@ public Level getLevel() {
           logger.log(buffer.clone().readString(charset));
```
swankjesse: small nit – we’re cloning the buffer here, but that isn’t necessary if it was gzipped. (The difference is negligible.)

PR #3796, opened 2018-01-24
Checking for a null socket input and output streams to prevent illegal argument exception in okio

JohnSolstice:
### Description
- Some devices occasionally have hardware/firmware issues with socket inputStream and outputStreams.  This change checks for a null socket inputStream and outputStream before using the Okio.sink and Okio.source methods.  This is necessary because 3rd party libraries that developers don't control might not be handling this exception correctly, causing their apps to crash.  It's best to just fix this issue at the source.

An example of the crash:
```
Fatal Exception: java.lang.IllegalArgumentException: out == null
       at okio.Okio.sink(SourceFile:69)
       at okio.Okio.sink(SourceFile:118)
       at okhttp3.internal.connection.RealConnection.connectSocket(SourceFile:251)
       at okhttp3.internal.connection.RealConnection.connect(SourceFile:158)
       at okhttp3.internal.connection.StreamAllocation.findConnection(SourceFile:256)
       at okhttp3.internal.connection.StreamAllocation.findHealthyConnection(SourceFile:134)
       at okhttp3.internal.connection.StreamAllocation.newStream(SourceFile:113)
       at okhttp3.internal.connection.ConnectInterceptor.intercept(SourceFile:42)
       at okhttp3.internal.http.RealInterceptorChain.proceed(SourceFile:147)
       at okhttp3.internal.http.RealInterceptorChain.proceed(SourceFile:121)
       at okhttp3.internal.cache.CacheInterceptor.intercept(SourceFile:93)
       at okhttp3.internal.http.RealInterceptorChain.proceed(SourceFile:147)
       at okhttp3.internal.http.RealInterceptorChain.proceed(SourceFile:121)
       at okhttp3.internal.http.BridgeInterceptor.intercept(SourceFile:93)
       at okhttp3.internal.http.RealInterceptorChain.proceed(SourceFile:147)
       at okhttp3.internal.http.RetryAndFollowUpInterceptor.intercept(SourceFile:125)
       at okhttp3.internal.http.RealInterceptorChain.proceed(SourceFile:147)
       at okhttp3.internal.http.RealInterceptorChain.proceed(SourceFile:121)
       at okhttp3.RealCall.getResponseWithInterceptorChain(SourceFile:200)
       at okhttp3.RealCall$AsyncCall.execute(SourceFile:147)
       at okhttp3.internal.NamedRunnable.run(SourceFile:32)
       at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1113)
       at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:588)
       at java.lang.Thread.run(Thread.java:818)```

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/connection/RealConnection.java
+++ b/okhttp/src/main/java/okhttp3/internal/connection/RealConnection.java
@@ -242,6 +242,11 @@ public final class RealConnection extends Http2Connection.Listener implements Co
       throw ce;
     }
 
+    // Protecting against rare occasion when input or output stream is null due to hardware/firmware issues
+    if (rawSocket.getInputStream() == null || rawSocket.getOutputStream() == null) {
+      throw new IOException("Socket's input or output stream is null");
+    }
+
     // The following try/catch block is a pseudo hacky way to get around a crash on Android 7.0
     // More details:
     // https://github.com/square/okhttp/issues/3245

### CODE REVIEW

**Comment by JakeWharton** [comment]
If this is true, we should put defensive code in `RealConnection.connectSocket` rather than catching an implementation detail exception from Okio.

**Comment by JohnSolstice** [comment]
@JakeWharton changed it to catch the exception in `RealConnection.connectSocket`

**Comment by JakeWharton** [comment]
We can detect the null `rawSocket` directly.

**Comment by JohnSolstice** [comment]
@JakeWharton okay, I see what you mean.  How's this?

File: okhttp/src/main/java/okhttp3/internal/connection/RealConnection.java [hunk 1]
```diff
@@ -242,6 +242,11 @@ private void connectSocket(int connectTimeout, int readTimeout, Call call,
       throw ce;
     }
 
+    // Making sure rawSocket is not null due to rare hardware/firmware issues with creating/sustaining a socket.
+    if (rawSocket == null) {
```
yschimke: I think this is too late as its used already at this point (lines 236, 238)

PromanSEW: This condtition must be before line 235

JohnSolstice: fixed

JohnSolstice:
Oh, hang on.  I made a mistake.  It's not the Socket that is null.  It's the Socket's InputStream and OutputStream.

swankjesse:
Yeah, good idea. But I think I'd prefer this fix to be in Okio.source(Socket) and Okio.sink(Socket). That'll catch more violators!

swankjesse:
Mind sending a PR there?

yschimke:
+1 to Okio

Also with this fix, I'm not sure it's guaranteed, e.g. behaviour to rawSocket.getInputStream could change in between two calls because of network changes etc.

JohnSolstice:
They currently do a check in okio, but they throw an InvalidArgumentException instead of an IOException.  You think I should make a MR that throws an IO instead?  I can try.  We'll see if they'll go for it.

JakeWharton:
They == the same people who are here!

JohnSolstice:
Well that makes it easy then :p

JohnSolstice:
I've put up a PR in okio https://github.com/square/okio/pull/337

JohnSolstice:
closing this in favor of the okio PR

PR #3799, opened 2018-01-26
Mark some EventListener parameters as non-null.

NightlyNexus:
These look like accidents, but it's very possible I missed some context on why these were nullable.


swankjesse:
(Build failure is due to a Java version bump in Travis.)

PR #3802, opened 2018-01-27
Do not count web sockets toward per-host limit.

### LINKED ISSUE #3801: Multiple websocket connections
```
    OkHttpClient client =
        new OkHttpClient.Builder().connectionPool(new ConnectionPool(10, 5, TimeUnit.MINUTES))
            .build();
    for (int i = 0; i < 10; i++) {
      client.newWebSocket(new Request.Builder().url("wss://echo.websocket.org").build(),
          new WebSocketListener() {
            @Override public void onOpen(WebSocket webSocket, Response response) {
              super.onOpen(webSocket, response);
              System.out.println("response=" + response);
            }
          });
    }
```
Above is the code to create 10 multiple websocket clients concurrently. However, it only opens 5 websocket connections.

JakeWharton:
Closes #3801.

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/RecordingWebSocketListener.java
+++ b/okhttp-tests/src/test/java/okhttp3/RecordingWebSocketListener.java
@@ -1,0 +1,45 @@
+/*
+ * Copyright (C) 2018 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


okhttp-tests/src/test/java/okhttp3/RecordingWebSocketListener.java
```diff
@@ -0,0 +1,45 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okhttp3;
+
+import javax.annotation.Nullable;
+import okio.ByteString;
+
+public final class RecordingWebSocketListener extends WebSocketListener {
```
JakeWharton: Named for people who go searching for it in the future to match RecordingCallback and RecordingEventListener.

swankjesse: Nice.

swankjesse:
Build failure is not related.

PR #3811, opened 2018-02-01
avoid resource leak if java.lang.Error occurred

igor-suhorukov:
use try with resource block to avoid resource leak

### DIFF
--- a/benchmarks/src/main/java/okhttp3/benchmarks/Benchmark.java
+++ b/benchmarks/src/main/java/okhttp3/benchmarks/Benchmark.java
@@ -185,9 +185,9 @@ public class Benchmark extends com.google.caliper.Benchmark {
 
     if (gzip) {
       Buffer gzipBody = new Buffer();
-      GzipSink gzipSink = new GzipSink(gzipBody);
-      gzipSink.write(body, body.size());
-      gzipSink.close();
+      try (GzipSink gzipSink = new GzipSink(gzipBody)) {
+        gzipSink.write(body, body.size());
+      }
       body = gzipBody;
       result.addHeader("Content-Encoding: gzip");
     }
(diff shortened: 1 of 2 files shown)


benchmarks/src/main/java/okhttp3/benchmarks/Benchmark.java
```diff
@@ -185,9 +185,9 @@ private MockResponse newResponse() throws IOException {
 
     if (gzip) {
       Buffer gzipBody = new Buffer();
-      GzipSink gzipSink = new GzipSink(gzipBody);
-      gzipSink.write(body, body.size());
-      gzipSink.close();
+      try (GzipSink gzipSink = new GzipSink(gzipBody)) {
```
swankjesse: This can’t fail in practice ’cause it’s all in memory.

mockwebserver/src/main/java/okhttp3/internal/http2/Http2Server.java
```diff
@@ -51,31 +51,32 @@ public Http2Server(File baseDirectory, SSLSocketFactory sslSocketFactory) {
...
-      Socket socket = null;
-      try {
-        socket = serverSocket.accept();
-
-        SSLSocket sslSocket = doSsl(socket);
-        String protocolString = Platform.get().getSelectedProtocol(sslSocket);
-        Protocol protocol = protocolString != null ? Protocol.get(protocolString) : null;
-        if (protocol != Protocol.HTTP_2) {
-          throw new ProtocolException("Protocol " + protocol + " unsupported");
+    try (ServerSocket serverSocket = new ServerSocket(8888)) {
```
swankjesse: We don’t use try-with-resources in our code ’cause it isn’t available on all of the platforms we support (Android 2.3+).

Issue #3793, opened 2018-01-22
Fatal Exception: ArrayIndexOutOfBoundsException in Hpack.java


AndrazP:
- Okhttp version: 3.9.1
- Retrofit version: 2.3.0

Received crash report from SM-G950F on Android 7.0.
Stacktrace
```
Fatal Exception: java.lang.ArrayIndexOutOfBoundsException: length=8; index=8
       at okhttp3.internal.http2.Hpack$Reader.getName(Hpack.java:265)
       at okhttp3.internal.http2.Hpack$Reader.readLiteralHeaderWithIncrementalIndexingIndexedName(Hpack.java:250)
       at okhttp3.internal.http2.Hpack$Reader.readHeaders(Hpack.java:195)
       at okhttp3.internal.http2.Http2Reader.readHeaderBlock(Http2Reader.java:196)
       at okhttp3.internal.http2.Http2Reader.readHeaders(Http2Reader.java:182)
       at okhttp3.internal.http2.Http2Reader.nextFrame(Http2Reader.java:129)
       at okhttp3.internal.http2.Http2Connection$ReaderRunnable.execute(Http2Connection.java:566)
       at okhttp3.internal.NamedRunnable.run(NamedRunnable.java:32)
       at java.lang.Thread.run(Thread.java:762)
```

yschimke:
Not sure if this is a purely client bug, but even assuming an invalid server response, this should fail a single connection (IOException), closing all streams and not kill the client instance or the app.

swankjesse:
This is fixed.

PR #3819, opened 2018-02-04
Check for invalid index to avoid IOOBE

yschimke:
Avoid any externally noticeable RuntimeExceptions due to bad server data

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/http2/Hpack.java
+++ b/okhttp/src/main/java/okhttp3/internal/http2/Hpack.java
@@ -221,7 +221,7 @@ final class Hpack {
         headerList.add(staticEntry);
       } else {
         int dynamicTableIndex = dynamicTableIndex(index - STATIC_HEADER_TABLE.length);
-        if (dynamicTableIndex < 0 || dynamicTableIndex > dynamicTable.length - 1) {
+        if (dynamicTableIndex < 0 || dynamicTableIndex >= dynamicTable.length) {
           throw new IOException("Header index too large " + (index + 1));
         }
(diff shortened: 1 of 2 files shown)


yschimke:
Addresses https://github.com/square/okhttp/issues/3793

yschimke:
Main discussion points

- Is this worth a bigger refactor e.g. check this whenever accessing the dynamic table
- Should it check the null range (start of dynamic table) also?

okhttp/src/main/java/okhttp3/internal/http2/Hpack.java
```diff
@@ -258,11 +258,16 @@ private void readLiteralHeaderWithIncrementalIndexingNewName() throws IOExceptio
...
     }
 
-    private ByteString getName(int index) {
+    private ByteString getName(int index) throws IOException {
       if (isStaticHeader(index)) {
         return STATIC_HEADER_TABLE[index].name;
       } else {
-        return dynamicTable[dynamicTableIndex(index - STATIC_HEADER_TABLE.length)].name;
+        int dynamicTableIndex = dynamicTableIndex(index - STATIC_HEADER_TABLE.length);
+        if (dynamicTableIndex < 0 || dynamicTableIndex > dynamicTable.length - 1) {
```
swankjesse: nit: more conventional to use `>=` instead of `>` and `- 1` ?

swankjesse:
We probably want to fuzz the decoder with near-real data?

yschimke:
Do you think it's better to fuzz the decoder and various components, in which case potentially finding theoretical but maybe inaccessible points caught by earlier components? Or to use something like a FuzzingSocket and test end to end?

PR #3821, opened 2018-02-04
Update deploy_snapshot.sh to compile tests and Version.java but not run

yschimke:


JakeWharton:
Why does this fix it? Maven is such pain...

yschimke:
@JakeWharton not certain it does, this build doesn't fail for me.  But based on Version being missing (a generated source)

http://maven.apache.org/plugins-archives/maven-surefire-plugin-2.12.4/examples/skipping-test.html

> You can also skip the tests via command line by executing the following command:
> 
> mvn install -DskipTests
> If you absolutely must, you can also use the maven.test.skip property to skip compiling the tests. maven.test.skip is honored by Surefire, Failsafe and the Compiler Plugin.
> 
> mvn install -Dmaven.test.skip=true

JakeWharton:
Yeah I understand what this changes but not why it fixes the problem.

yschimke:
This is a shot in the dark, but based on Version.java not being present, which I assume is because the right phases not running. It seems like a low effort thing to try. Just managed to repro using package instead of deploy.  Doesn't fix it after all.

yschimke:
More specifically the change I made fixes the Version.java issue, but the others remain.

yschimke:
Testing locally with

$ mvn clean package source:jar javadoc:jar install --settings=".buildscript/settings.xml" -Dmaven.javadoc.failOnError=false -DskipTests

PR #3832, opened 2018-02-08
set the Proxy-Authorization

zi6xuan:
When using http-proxy, and the request belongs to https, it can not be automatically authorized because the authorization code set by the user is not read correctly here


yschimke:
https://api.travis-ci.org/v3/job/338818673/log.txt

Results :

Failed tests: 
  CallTest.noProactiveProxyAuthorization:2771 expected null, but was:<password>
  URLConnectionTest.proxyConnectIncludesProxyHeadersOnly:1059 expected null, but was:<bar>

swankjesse:
Unfortunately this also has the side-effect of leaking proxy credentials to the origin server. We’re planning to do pre-emptive proxy auth, but I don’t think it belongs on the Request object.

PR #3833, opened 2018-02-08
multipart filter option

Aks-4125:
To avoid multipart data in logcat.

based on [this](https://stackoverflow.com/questions/48661623/okhttp-logging-interceptor-disable-multipart-data) & [this](https://stackoverflow.com/questions/38563394/okhttp-loggin-interceptor-in-multipart) problem

By setting `interceptor.setShowMultipartLog(false);` we can see actual request data in log cat.

How to apply?

       `HttpLoggingInterceptor interceptor= new HttpLoggingInterceptor ();
        logging.setShowMultipartLog(false); // set true to display multipart data (by default true)
        httpClient.addInterceptor(interceptor);`

Output in logcat:

D/OkHttp: --> POST api-end-point
D/OkHttp: Content-Type: application/json; charset=UTF-8
D/OkHttp: Content-Length: 7556444
D/OkHttp: Multipart length transmitted->7556444
D/OkHttp: --> END POST (7556444-byte body)

### DIFF
--- a/okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
+++ b/okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
@@ -125,6 +125,8 @@ public final class HttpLoggingInterceptor implements Interceptor {
 
   private volatile Level level = Level.NONE;
 
+  private boolean showMultipartLog = true;
+
   /** Change the level at which this interceptor logs. */
   public HttpLoggingInterceptor setLevel(Level level) {
     if (level == null) throw new NullPointerException("level == null. Use Level.NONE instead.");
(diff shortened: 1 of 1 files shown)


okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
```diff
@@ -136,6 +138,23 @@ public Level getLevel() {
     return level;
   }
 
+  /**
+   *
+   * @return check mutipart logging is enabled or not
```
yschimke: nit: typo mutipart

okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
```diff
@@ -136,6 +138,23 @@ public Level getLevel() {
...
+   */
+  public boolean isShowMultipartLog() {
+    return showMultipartLog;
+  }
+
+  /**
+   *
+   * @param showMultipartLogging set to false to display multipart length in integer instead of non-plain text (by default true)]
```
yschimke: Can you clean up these javadoc, put most of the comment to the body and less in the param and return.

Aks-4125: sure..

yschimke:
I'm not clear on this diff what makes it related to multipart response?  Doesn't it apply more generally?  I assume you want to avoid content that is ascii at the start but then becomes binary in a second part?

Please add some tests as well.

Aks-4125:
That does skip multipart response of type ascii/binary.

yschimke:
@Aks-4125 how can I help?

yschimke:
Closing for now, please reopen if it is still active.

PR #3838, opened 2018-02-10
Avoid extra details when using a proxy

yschimke:
These are not honoured for connection reuse, so seems strange to use initially.  AFAICT this will only impact Android, or JVM clients using a 3rd party ProxySelector impl


yschimke:
Not sure if this has ever impacted someone, but seems best to avoid if a PAC file is involved.

yschimke:
I'll add a test before it's worth landing.

swankjesse:
Thumbs up.

yschimke:
Test showed it is already handled as Address builds the URL

PR #3839, opened 2018-02-10
Minor cleanup

yschimke:


yschimke:
Minor low priority cleanup, while I'm working on something else.  Curious why the explicit for loop? To avoid calling size() multiple times?  I thought anecdotally, foreach was slightly faster than indexing back into the collection.

Close, if it's desired, I'm mainly just trying to scratch my curiosity.

JakeWharton:
All these for loops are over lists for which the for-each style is less efficient.

PR #3841, opened 2018-02-12
Small web socket optimizations.

JakeWharton:


JakeWharton:
`UnsafeCursor` usage for mask XORing is next...

okhttp/src/main/java/okhttp3/internal/ws/WebSocketReader.java
```diff
@@ -134,7 +133,7 @@ private void readHeader() throws IOException {
 
     int b1 = source.readByte() & 0xff;
 
-    isMasked = (b1 & B1_FLAG_MASK) != 0;
+    boolean isMasked = (b1 & B1_FLAG_MASK) != 0;
     if (isMasked == isClient) {
```
JakeWharton: This logic ensures `isMasked` and `isClient` were always the same.

JakeWharton:
Added more changes here that enable unsafe cursor usage.

PR #3855, opened 2018-02-17
RFC: ConnectionPool listener and list method

yschimke:
I'm trying to understand the behaviour of network connections under different scenarios on Android.  It is tougher without insight into exactly what happens during connectivity changes, e.g. switching between wifi, cell etc.

But just generally seems like a useful feature for teams trying to optimize connection behaviour via adjusting timeouts, as you change timeouts, being able to understand connection churn without wrapping Socket etc.

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/ConnectionPoolTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/ConnectionPoolTest.java
@@ -19,7 +19,9 @@ import java.net.InetSocketAddress;
 import java.net.Proxy;
 import java.net.ProxySelector;
 import java.net.Socket;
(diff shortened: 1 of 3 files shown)


yschimke:
Putting up to see whether this idea will fly or not.  Happy to use with dev builds and park for now if not considered wide enough use for a public api.

swankjesse:
I would like to have done this through the EventListener API, though I suppose it always needs a call and we won’t have a call when a connection is purged from the pool.

I’m not quite ready for this. I think it’s an interesting idea, but I worry that it’d be better to just branch & instrument OkHttp in source for the small number of users that need this.

Issue #3235, opened 2017-03-20
HttpUrl.Builder.addQueryParameter is not escaping curly braces which causes Tomcat 8 to reject request


arakelian:
OkHttp does not properly escape URL query parameters that contain curly braces.

Tomcat 7.0.73+, 8.0.39+, and 8.5.7+ will reject these requests with the following error:
```
Mar 20, 2017 2:47:59 PM org.apache.coyote.http11.Http11Processor service
INFO: Error parsing HTTP request header
 Note: further occurrences of HTTP header parsing errors will be logged at DEBUG level.
java.lang.IllegalArgumentException: Invalid character found in the request target. The valid characters are defined in RFC 7230 and RFC 3986
        at org.apache.coyote.http11.Http11InputBuffer.parseRequestLine(Http11InputBuffer.java:471)
        at org.apache.coyote.http11.Http11Processor.service(Http11Processor.java:667)
        at org.apache.coyote.AbstractProcessorLight.process(AbstractProcessorLight.java:66)
        at org.apache.coyote.AbstractProtocol$ConnectionHandler.process(AbstractProtocol.java:798)
        at org.apache.tomcat.util.net.NioEndpoint$SocketProcessor.doRun(NioEndpoint.java:1434)
        at org.apache.tomcat.util.net.SocketProcessorBase.run(SocketProcessorBase.java:49)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:615)
        at org.apache.tomcat.util.threads.TaskThread$WrappingRunnable.run(TaskThread.java:61)
        at java.lang.Thread.run(Thread.java:745)
```

These are the characters that Tomcat will now reject as part of the URI request line (see org.apache.tomcat.util.http.parser.HttpParser):
```
            if (IS_CONTROL[i] || i > 127 ||
                    i == ' ' || i == '\"' || i == '#' || i == '<' || i == '>' || i == '\\' ||
                    i == '^' || i == '`'  || i == '{' || i == '|' || i == '}') {
                IS_NOT_REQUEST_TARGET[i] = true;  // reject the character!
```

To solve issue, the characters included in HttpUrl.QUERY_COMPONENT_ENCODE_SET may need to be changed.  Here is current definition:
```
static final String QUERY_COMPONENT_ENCODE_SET = " \"'<>#&=";
```

These

[...]

PR #3863, opened 2018-02-19
Encode more characters when using addQueryParameter().

swankjesse:
Though it's good for OkHttp to retain the user's provided encoding, we
should be encoding more characters than we are when the user provides
us with unencoded data.

Closes: https://github.com/square/okhttp/issues/3235

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/HttpUrlTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/HttpUrlTest.java
@@ -553,6 +553,15 @@ public final class HttpUrlTest {
         .test(Component.QUERY);
   }
 
+  @Test public void queryValueCharacters() throws Exception {
+    new UrlComponentEncodingTester()
+        .override(Encoding.IDENTITY, '?', '`')
+        .override(Encoding.PERCENT, '\'')
+        .override(Encoding.SKIP, '#', '+')
+        .skipForUri('%', '\\', '^', '`', '{', '|', '}')
+        .test(Component.QUERY_VALUE);
+  }
+
   @Test public void fragmentCharacters() throws Exception {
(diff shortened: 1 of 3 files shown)


okhttp-tests/src/test/java/okhttp3/HttpUrlTest.java
```diff
@@ -1168,8 +1177,11 @@
         .host("host")
         .addQueryParameter("=[]:;\"~|?#@^/$%*", "a")
         .build();
-    assertEquals("http://host/?%3D[]:;%22~|?%23@^/$%25*=a", url.toString());
-    assertEquals("http://host/?%3D[]:;%22~%7C?%23@%5E/$%25*=a", url.uri().toString());
+    assertEquals("http://host/?%3D%5B%5D%3A%3B%22~%7C%3F%23%40%5E%2F%24%25*=a",
```
swankjesse: This is a big deal! Big behavior change.

okhttp-tests/src/test/java/okhttp3/HttpUrlTest.java
```diff
@@ -1168,8 +1177,11 @@
         .host("host")
         .addQueryParameter("=[]:;\"~|?#@^/$%*", "a")
         .build();
-    assertEquals("http://host/?%3D[]:;%22~|?%23@^/$%25*=a", url.toString());
-    assertEquals("http://host/?%3D[]:;%22~%7C?%23@%5E/$%25*=a", url.uri().toString());
+    assertEquals("http://host/?%3D%5B%5D%3A%3B%22%7E%7C%3F%23%40%5E%2F%24%25*=a",
```
swankjesse: This is a big deal! Big behavior change.

Issue #3035, opened 2016-12-12
Update cipher suites for June 2017


swankjesse:
We most recently updated the cipher suites in December 2016. We should update 'em again in June 2017.

When that's done, please open another tracking bug to update cipher suites in December 2017!

PR #3864, opened 2018-02-19
Remove CBC-mode ECDSA cipher suites

swankjesse:
This removes the following cipher suites as supported-by-default:
  TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA
  TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA

This tracks a similar change in Chromium 56:
  https://developers.google.com/web/updates/2016/12/chrome-56-deprecations#remove_cbc-mode_ecdsa_ciphers_in_tls

OkHttp's updated cipher suites, plus comparisons to Android, Java,
Chromium, Firefox, and Square's server-site configuration is here:
  https://tinyurl.com/okhttp-cipher-suites

Closes: https://github.com/square/okhttp/issues/3035

### DIFF
--- a/okhttp/src/main/java/okhttp3/ConnectionSpec.java
+++ b/okhttp/src/main/java/okhttp3/ConnectionSpec.java
@@ -54,9 +54,7 @@ public final class ConnectionSpec {
       // Note that the following cipher suites are all on HTTP/2's bad cipher suites list. We'll
       // continue to include them until better suites are commonly available. For example, none
       // of the better cipher suites listed above shipped with Android 4.4 or Java 7.
-      CipherSuite.TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA,
       CipherSuite.TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA,
-      CipherSuite.TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA,
       CipherSuite.TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA,
       CipherSuite.TLS_RSA_WITH_AES_128_GCM_SHA256,
       CipherSuite.TLS_RSA_WITH_AES_256_GCM_SHA384,


yschimke:
I'm curious about removing them from the compatible spec.  This list is the aspirational list, and when chrome changes them it does them in relation to two factors

1) Websites deploying better more modern TLS libraries e.g. patching openssl on the server.  This is a continual process.
2) The user agent e.g. Chrome itself adding support for new ciphers and dropping support for older ones to discourage their use and promote servers upgrading.

But with older versions of Android we are stuck with a static set of results for 2), so removing presumably just breaks some amount of older clients.

Not a blocker, I'd just like to understand the decision more.  e.g. why not drop the fallback to compatible and promote the Google Play provider when clients upgrade to OkHttp 3.10 instead?

swankjesse:
The linked explanation from Chromium explains it best: _“Although CBC-mode ciphers are still widely used with RSA, they are virtually nonexistent with ECDSA.”_

Removing a cipher suite that’s well-exercised is dangerous and needs to be done with much caution. Removing a cipher suite that doesn’t get used anyway is just reducing attack surface.

I’m assuming that because Chrome could remove these cipher suites without drama, we’ll be able to also. There’s a potential break for servers that only negotiate `SSL_RSA_WITH_3DES_EDE_CBC_SHA` and our two removed cipher suites. Such servers will be reachable in Chrome, but not in OkHttp on Android because Android disables `SSL_RSA_WITH_3DES_EDE_CBC_SHA` by default. I don’t expect this to be a problem in practice.

--

I’m pretty happy with how convergent things are. Android, Chrome, and Firefox are very similar today, but haven’t been historically. 

The JDK’s default cipher suites list is quite long, and includes some potentially-sketchy cipher suites. I haven’t investigated it too closely so I’m not sure if it’s vulnerable, but when [DHE-weaknesses](https://weakdh.org/) were announced Chromium, Firefox, and Android all removed DHE suites. The JDK hasn’t. (I wonder if that means that Apache HttpClient users have to curate cipher suites themselves?)

yschimke:
Thanks for humouring my curiosity

PR #3868, opened 2018-02-19
Conscrypt per factory [RFC]

yschimke:


samples/guide/src/main/java/okhttp3/recipes/CustomTrust.java
```diff
@@ -44,9 +44,7 @@ public CustomTrust() {
     SSLSocketFactory sslSocketFactory;
     try {
       trustManager = trustManagerForCertificates(trustedCertificatesInputStream());
-      SSLContext sslContext = SSLContext.getInstance("TLS");
```
swankjesse: Can’t call internal APIs in examples

samples/guide/src/main/java/okhttp3/recipes/CustomCipherSuites.java
```diff
@@ -55,7 +53,7 @@ public CustomCipherSuites() throws GeneralSecurityException {
         .build();
 
     X509TrustManager trustManager = defaultTrustManager();
-    SSLSocketFactory sslSocketFactory = defaultSslSocketFactory(trustManager);
+    SSLSocketFactory sslSocketFactory = Platform.get().getSslSocketFactory(trustManager);
```
swankjesse: No internal APIs in examples

yschimke:
I really like the client driven approach to passing a ConscryptSslSocketFactory to OkHttpClient.Builder, but in practice it is awkward because Platform is used as a JVM singleton and might have be Conscrypt, Jdk9, Platform or Jdk8Platform.

I think it only makes sense if we allow clients to select Platform per client instance.

swankjesse:
Maybe we should have a different Platform for different security providers? Maybe a method that retrieves a platform from a given SSLSocketFactory?

yschimke:
No short term fix here.  Can be worked around by apps using Conscrypt in two places, by setting one per connection and the other using global default.  For OkHttp this would mean supplying your own ssl socket factor that sets this.

PR #3869, opened 2018-02-19
Change conscrypt check [RFC]

yschimke:
Check Provider two ways

1) simple name "Conscrypt" but can be changed
2) class name, but can be obfuscated


yschimke:
Current check seems good enough and fails closed, which is probably what we want for now

PR #3870, opened 2018-02-20
Fail web sockets if the pong is not received

swankjesse:
There are several options for when to expect the pong response. I've
chosen to overload the ping interval. This seems reasonable because the
ping interval is the maximum amount of time that should elapse before
a connection failure is detected.

Closes: https://github.com/square/okhttp/issues/3227

### DIFF
--- a/okhttp/src/main/java/okhttp3/OkHttpClient.java
+++ b/okhttp/src/main/java/okhttp3/OkHttpClient.java
@@ -434,7 +434,7 @@ public class OkHttpClient implements Cloneable, Call.Factory, WebSocket.Factory 
    * Uses {@code request} to connect a new web socket.
    */
   @Override public WebSocket newWebSocket(Request request, WebSocketListener listener) {
-    RealWebSocket webSocket = new RealWebSocket(request, listener, new Random());
+    RealWebSocket webSocket = new RealWebSocket(request, listener, new Random(), pingInterval);
     webSocket.connect(this);
     return webSocket;
   }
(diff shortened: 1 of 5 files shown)


okhttp/src/main/java/okhttp3/OkHttpClient.java
```diff
@@ -570,8 +570,11 @@ public Builder writeTimeout(long timeout, TimeUnit unit) {
     /**
      * Sets the interval between web socket pings initiated by this client. Use this to
      * automatically send web socket ping frames until either the web socket fails or it is closed.
-     * This keeps the connection alive and may detect connectivity failures early. No timeouts are
-     * enforced on the acknowledging pongs.
+     * This keeps the connection alive and may detect connectivity failures.
+     *
+     * <p>If the server does not respond to each ping with a pong within {@code interval}, this
+     * client will assume that connectivity has been lost. When this happens the connection is
+     * canceled and its listener is {@link WebSocketListener#onFailure notified of the failure}.
```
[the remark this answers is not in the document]
swankjesse: Fixed.

okhttp/src/main/java/okhttp3/internal/ws/RealWebSocket.java
```diff
@@ -519,9 +528,21 @@ boolean writeOneFrame() throws IOException {
...
+    int receivedPongCount;
     synchronized (this) {
       if (failed) return;
       writer = this.writer;
+      sentPingCount = this.sentPingCount;
+      receivedPongCount = this.receivedPongCount;
+      this.sentPingCount++;
+    }
+
+    if (sentPingCount > receivedPongCount) {
```
yschimke: This ping strategy is simple and understandable, but I'm concerned it could break some existing clients with already established tight ping interval.  e.g. in 3.9.1 they used 50ms within a datacentre.  This ping strategy would then turn every occasional transient network blip into a connection loss.  Likewise on a mobile device a ping interval now needs to be > 3? 5? seconds, but even then may cause losses that were otherwise recoverable.

As it's an upgrade we should probably provide guidance during the upgrade of what a minimum interval should be.

An alternative strategy we used in RSocket was to allow a configurable number of outstanding pings before disconnection, so you can keep a low ping interval to detect problems sooner, but still recover from transient blips.  However it complicates things here greatly because of the ability of servers to coalesce pongs to the final response, so you would need to use the application data in the ping/pong to send a counter or similar.

yschimke: n.b. This change is effectively changing from ∞ # of outstanding pongs to 1, so it does seem possible we will catch some clients unaware in the upgrade.

Not a blocker from me, your call.

swankjesse: Yeah, this definitely needs to be considered by users as they upgrade. 

My expectation is that we'll detect a bunch of connectivity problems that were previously undetected. Hopefully we don't have false positives due to overly-aggressive ping intervals.

Issue #1019, opened 2014-08-08
Support “HTTP” URIs (HTTP Upgrade mechanism) 


JanAndersson:
As far as I understand this isn't supported in OkHttp. As it is part of the HTTP/2.0 standard I beleive this would be of great value to OkHttp...

yschimke:
I'm working with a backend that supports plaintext assumed http/2.  So I'd like to look into this, ideally both

- plainttext h2c via upgrade
- http/2 only for both http and https e.g. assumed without ALPN

Specifically this could relax OkHttpClient.Builder.protocols

```
     * @param protocols the protocols to use, in order of preference. The list must contain {@link
     * Protocol#HTTP_1_1}. 
```

swankjesse:
There's no spec for plaintext assumed HTTP/2. That's nonstandard and won't work with other tools.

yschimke:
Isn't this section "3.4 Starting HTTP/2 with Prior Knowledge"

http://http2.github.io/http2-spec/index.html#rfc.section.3.4

swankjesse:
Ah you're right. I still think it’s unwise to skip the upgrade path.

yschimke:
Unfortunately, if the server also assumes h2c over plaintext then the initial HTTP/1.1 request with upgrade header will cause the server to barf

yschimke:
Minus the plaintext Upgrade support which I'm still working on this turned out pretty easy to add assumed h2c mode, and http/2 only via ALPN.  Do you see any reason why this couldn't land once completed with proper tests etc?

https://github.com/yschimke/okhttp/pull/1/files

JaySWang:
Hi guys, do you have any plan to support h2c?

swankjesse:
None currently.

jpitz:
I have an h2c prior knowledge solution out for review. Changes to critical code paths was rather minimal and I was able to leverage much of the existing test infrastructure. See gh-3873

swankjesse:
We did HTTP/2 prior knowledge in 3.11. I’m not planning anything further here!

PR #3873, opened 2018-02-20
gh-1019: Implemented prior knowledge approach to h2c

jpitz:
Needs a unit test, but I wanted to get the idea out there of how easy this _could_ be.

An Address's List<Protocol> comes from the initial configuration. If someone specifies the H2C protocol in that list, then we assume prior knowledge since it requires users to specify the protocol in their configuration.

I tried to wire a boolean attribute through the constructor chain, but that got really gross. This seemed like an easier approach with a lot fewer changes in code.


swankjesse:
I’m reluctant to implement support for H2C because it’s a cleartext protocol and cleartext is dead to me.

jjshanks:
@swankjesse While understandable we have a pretty reasonable use case. [Transparent TLS via a service mesh](https://buoyant.io/2016/03/24/transparent-tls-with-linkerd/). We have clients that use Retrofit and Feign but can't use h2c to talk to localhost without this kind of change.

swankjesse:
Gotcha. So cleartext doesn’t leave the host. That’s not so bad.

okhttp/src/main/java/okhttp3/internal/connection/RealConnection.java
```diff
@@ -260,8 +260,16 @@ private void connectSocket(int connectTimeout, int readTimeout, Call call,
   private void establishProtocol(ConnectionSpecSelector connectionSpecSelector, Call call,
       EventListener eventListener) throws IOException {
     if (route.address().sslSocketFactory() == null) {
-      protocol = Protocol.HTTP_1_1;
       socket = rawSocket;
+
+      // this implementation takes the "prior knowledge" approach to the spec
+      // https://tools.ietf.org/html/rfc7540#section-3.4
+      if (route.address().protocols().contains(Protocol.H2C)) {
```
yschimke: I'd suggest we change 

https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/OkHttpClient.java#L854-L856

To allow H2C on it's own, but only on it's own, no fallback to HTTP/1.1.

yschimke:
I'm a big fan of this, have hit this a couple of times before during development and ended up using Netty.

yschimke:
To be clear, my main concern is this makes a client that is single purpose.  So we should probably only allow H2C as the single supported protocol for a client.  Otherwise this client instance will be *mostly* usable for https but broken for most external plaintext traffic.

It seems better to be really explicit that the client is single purpose.

okhttp/src/main/java/okhttp3/internal/connection/RealConnection.java
jpitz: Done. This ended up cleaning up some of the logic in the establishment.

okhttp/src/main/java/okhttp3/OkHttpClient.java
```diff
@@ -844,24 +846,30 @@ public Builder dispatcher(Dispatcher dispatcher) {
...
-      // Validate that the list has everything we require and nothing we forbid.
-      if (!protocols.contains(Protocol.HTTP_1_1)) {
-        throw new IllegalArgumentException("protocols doesn't contain http/1.1: " + protocols);
+      final Set<Protocol> uniqueProtocols = new HashSet<>(protocols);
+
+      if (uniqueProtocols.contains(Protocol.H2C) && uniqueProtocols.size() > 1) {
+        // when using h2c prior knowledge, no other protocol should be supported.
+        throw new IllegalArgumentException("protocols containing h2c cannot use other protocols:"
+                + uniqueProtocols);
+      } else if (!uniqueProtocols.contains(Protocol.HTTP_1_1)) {
```
jpitz: bug: this needs to check if uniqueProtocols does not contain H2c

okhttp/src/main/java/okhttp3/Protocol.java
```diff
@@ -61,6 +61,12 @@
    */
   HTTP_2("h2"),
 
+  /**
+   * Cleartext implementation of HTTP2. This enumeration exists for the "prior knowledge" upgrade
+   * semantic supported by the protocol.
+   */
+  H2C("h2c"),
```
jpitz: bug: this needs to be added to the static get method.

okhttp/src/main/java/okhttp3/OkHttpClient.java
yschimke: This is checked implicitly via the if case above

okhttp/src/main/java/okhttp3/OkHttpClient.java
```diff
@@ -844,24 +846,30 @@ public Builder dispatcher(Dispatcher dispatcher) {
      */
     public Builder protocols(List<Protocol> protocols) {
       // Create a private copy of the list.
-      protocols = new ArrayList<>(protocols);
-
-      // Validate that the list has everything we require and nothing we forbid.
-      if (!protocols.contains(Protocol.HTTP_1_1)) {
-        throw new IllegalArgumentException("protocols doesn't contain http/1.1: " + protocols);
+      final Set<Protocol> uniqueProtocols = new HashSet<>(protocols);
```
yschimke: @swankjesse has stated a preference for LinkedHashSet makes these a lot more predictable in tests, e.g. checking the exception message.  But generally you could probably just check the contains(H2C) and size > 1 on the list and fold in duplicates as bad input.  But this isn't super critical code path, so that's minor.

okhttp/src/main/java/okhttp3/OkHttpClient.java
```diff
@@ -844,24 +846,30 @@ public Builder dispatcher(Dispatcher dispatcher) {
...
         throw new IllegalArgumentException("protocols must not contain null");
       }
 
       // Remove protocols that we no longer support.
-      protocols.remove(Protocol.SPDY_3);
+      uniqueProtocols.remove(Protocol.SPDY_3);
 
       // Assign as an unmodifiable list. This is effectively immutable.
-      this.protocols = Collections.unmodifiableList(protocols);
+      this.protocols = Collections.unmodifiableList(new ArrayList<>(uniqueProtocols));
```
yschimke: I think changing the selected protocol because of hashing could affect the ALPN selection IIRC.  Not so much an issue without SPDY, but undesirable.

okhttp/src/main/java/okhttp3/OkHttpClient.java
jpitz: Good catch. It wasn't super clear if the list of protocols was treated as a prioritized. Adding to preserve ordering.

okhttp/src/main/java/okhttp3/OkHttpClient.java
jpitz: Yeah. I think that change went a little too far. Going to revert back to using just the list and suffer the minimal performance cost.

okhttp/src/main/java/okhttp3/OkHttpClient.java
jpitz: Actually it's not. In the event that someone specifies only H2C, then the second if check evaluates to false, triggering this else if. Caught it in one of the unit tests I was writing.

okhttp/src/main/java/okhttp3/Protocol.java
jpitz: Resolved

jpitz:
Do y'all prefer rebase or merge when conflicts arise? I didn't see a CONTRIBUTING guide anywhere.

I tend to do merges, but don't have a preference for either way.

JakeWharton:
We'll squash when merging a multi-commit PR so it's up to you.

jpitz:
Working checkpoint. I tested this against a service I have that support H2C out of box. Started to make modifications to the MockWebServer, but that might be a bit too ambitious. I tried putting in a shortcut, but that didn't seem to work. I'll try a few more things over the next couple of days

jpitz:
Alright, I managed to get the MockWebServer working and pulled the test cases from HttpOverHttp2Test up to an Http2TestBase that can be shared by both the cleartext and TLS flavors. I'm tempted to put up a second PR for this change. Any opposition or would you prefer to keep everything in this one?

okhttp-tests/src/test/java/okhttp3/internal/http2/HttpOverH2CTest.java
```diff
@@ -0,0 +1,85 @@
...
+import java.util.Arrays;
+import java.util.List;
+
+import static okhttp3.TestUtil.defaultClient;
+import static org.junit.Assert.assertEquals;
+
+/**
+ * Test H2C, the cleartext implementation of HTTP/2.
+ *
+ * @author jpitz
```
yschimke: strip author tags

okhttp/src/main/java/okhttp3/internal/http2/Http2Codec.java
```diff
@@ -124,6 +124,9 @@ public Http2Codec(OkHttpClient client, Interceptor.Chain chain, StreamAllocation
   @Override public Response.Builder readResponseHeaders(boolean expectContinue) throws IOException {
     List<Header> headers = stream.takeResponseHeaders();
     Response.Builder responseBuilder = readHttp2HeadersList(headers);
+    if (client.protocols().contains(Protocol.H2C)) {
```
yschimke: why set it here and not inside readHttp2HeadersList?

okhttp/src/main/java/okhttp3/internal/connection/RealConnection.java
```diff
@@ -260,9 +260,14 @@ private void connectSocket(int connectTimeout, int readTimeout, Call call,
 
   private void establishProtocol(ConnectionSpecSelector connectionSpecSelector,
       int pingIntervalMillis, Call call, EventListener eventListener) throws IOException {
-    if (route.address().sslSocketFactory() == null) {
-      protocol = Protocol.HTTP_1_1;
+    if (route.address().protocols().contains(Protocol.H2C)) {
```
yschimke: It seems like this check could be a subset of the "if (route.address().sslSocketFactory() == null) " case

meaning 0 cost in the more common https case, and clearer that this is a special case of clear text.  Thoughts?

okhttp/src/main/java/okhttp3/OkHttpClient.java
```diff
@@ -844,15 +844,25 @@ public Builder dispatcher(Dispatcher dispatcher) {
      * HTTP/1.1} only. If the server responds with {@code HTTP/1.0}, that will be exposed by {@link
      * Response#protocol()}.
      *
-     * @param protocols the protocols to use, in order of preference. The list must contain {@link
-     * Protocol#HTTP_1_1}. It must not contain null or {@link Protocol#HTTP_1_0}.
+     * @param protocols the protocols to use, in order of preference. If the list contains {@link
```
yschimke: Could we reword this to start with the normal case of HTTP_1_1.  As it is has a very H2C centric view of the world.

okhttp-tests/src/test/java/okhttp3/internal/http2/Http2TestBase.java
```diff
@@ -0,0 +1,1185 @@
...
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertThat;
+import static org.junit.Assert.fail;
+
+/**
+ * Base test cases for HTTP2 communication. These tests are executed for both H2C and HTTP2
+ * based requests. For implementation specific tests, include in the appropriate test case.
+ *
+ * This code originally lived in the HttpOverHttp2Test class, but was abstracted out for use
```
yschimke: I don't think this history is needed. We have git for that.

okhttp-tests/src/test/java/okhttp3/internal/http2/HttpOverH2CTest.java
```diff
@@ -0,0 +1,85 @@
...
+          .protocols(PROTOCOLS)
+          .build(),
+      "http",
+      Protocol.H2C
+    );
+    server.setProtocols(PROTOCOLS);
+  }
+
+  @Test
+  public void testProtocol() throws Exception {
```
yschimke: It feels like these tests from here onwards would be more natural in ProtocolTest (new), MockWebserverTest and OkHttpClientTest.  Thoughts?

okhttp-tests/src/test/java/okhttp3/internal/http2/HttpOverH2CTest.java
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * @author jpitz
+ */
+@SuppressWarnings("Duplicates")
+public final class HttpOverH2CTest extends Http2TestBase {
+  private static final List<Protocol> PROTOCOLS = Arrays.asList(Protocol.H2C);
+
+  @Before
+  public void setUp() throws Exception {
+    super.setUp(
```
yschimke: It feels like this could be handled without as much major surgery with a @Parameterized test.  Not sure which is generally favoured, but I try to avoid subclassing in case there is another dimension of freedom to the tests in the future e.g. Platform.

But interested to hear input from others and yours.

yschimke: It also means you can't go to a single test and run it from within Http2TestBase.

okhttp-tests/src/test/java/okhttp3/internal/http2/HttpOverH2CTest.java
jpitz: Ah. I looked for a ProtocolTest and there wasn't one. Didn't look again after handling the merge. I'll move them.

okhttp/src/main/java/okhttp3/OkHttpClient.java
jpitz: I'll take another crack. I started trying to word it that way, but ran into issues with clarity.

okhttp/src/main/java/okhttp3/internal/http2/Http2Codec.java
jpitz: That method was `public static` and so I would need to thread the value through the method signature, causing a breaking API change.

okhttp-tests/src/test/java/okhttp3/internal/http2/HttpOverH2CTest.java
jpitz: Huh.. This week I learned. Didn't know about Parameterized builds in JUnit. This is exactly what I am looking for.

okhttp/src/main/java/okhttp3/internal/http2/Http2Codec.java
jpitz: This could actually be cached on the class. No need to do the linear scan all the time.

jpitz:
@yschimke : master doesn't appear to have a ProtocolTest yet. Where is it?

okhttp/src/main/java/okhttp3/internal/http2/Http2Codec.java
jpitz: Alternatively, another attribute could be added to the OkHttpClient and that can be a cached referenced for the lifetime of the client

okhttp-tests/src/test/java/okhttp3/internal/http2/HttpOverHttp2Test.java
```diff
@@ -1147,6 +1182,8 @@ private int countFrames(List<String> logs, String message) {
    * <p>This test uses proxy tunnels to get a hook while a connection is being established.
    */
   @Test public void concurrentHttp2ConnectionsDeduplicated() throws Exception {
+    if (protocol == Protocol.H2C) return; // skip
```
yschimke: nit: using Assume is clearer in IDEs that it isn't running.

okhttp/src/main/java/okhttp3/internal/http2/Http2Codec.java
```diff
@@ -90,13 +90,17 @@
   final StreamAllocation streamAllocation;
   private final Http2Connection connection;
   private Http2Stream stream;
+  private final boolean useH2C;
```
yschimke: nit: seems clearer to just assign the Protocol here (HTTP_2 or H2C) and just pass in as a param to readHttp2HeadersList, instead of setting HTTP_2 and then based on this boolean overriding the value.

yschimke:
@jpitz You would be adding a ProtocolTest in this PR.  Seems like something that is just missing but is valid to test.

yschimke:
I'm keen to use this, thanks @jpitz for driving this through.

jpitz:
@yschimke : ah me adding the unit test wasn't clear from that message. I've gone through and added it. I've moved the protocol assignment into the readHeadersList function, but it changed the method signature. While I understand it to be library private and it's unlikely that anyone is using the method, it does pose some risk in the change should any external users be calling that method. I _could_ create a fallback method where the protocol is assumed to be HTTP_2.

jpitz:
Interesting, seems like adding assume causes the unit test to fail. I wonder if this is due to the fact that the testis now being run with the parameterized running vs the standard junit runner. My assumption would be that all Junit provided runners would behave the same, but that may not be true since they call out to it in their Javadoc: https://junit.org/junit4/javadoc/4.12/org/junit/Assume.html

"The default JUnit runner skips tests with failing assumptions. Custom runners may behave differently."

@yschimke : I'm going to go back to the simple if check and return.

jpitz:
Thats odd. Test ran fine locally. Will need to look into this later.

okhttp-tests/src/test/java/okhttp3/ProtocolTest.java
```diff
@@ -0,0 +1,34 @@
...
+    public void testGetKnown() throws IOException {
+        assertEquals(Protocol.HTTP_1_0, Protocol.get("http/1.0"));
+        assertEquals(Protocol.HTTP_1_1, Protocol.get("http/1.1"));
+        assertEquals(Protocol.SPDY_3, Protocol.get("spdy/3.1"));
+        assertEquals(Protocol.HTTP_2, Protocol.get("h2"));
+        assertEquals(Protocol.H2C, Protocol.get("h2c"));
+        assertEquals(Protocol.QUIC, Protocol.get("quic"));
+    }
+
+    @Test(expected = IOException.class)
```
yschimke: nit: I expect (no pun intended) Jesse will ask you to turn this into 

```
try {
  X; 
  fail();
assertEquals("xxxx", expected.getMessage()); 
}
```

okhttp-tests/src/test/java/okhttp3/OkHttpClientTest.java
```diff
@@ -169,4 +169,30 @@
...
+
+  @Test(expected = IllegalArgumentException.class)
+  public void testH2COkHttpClientConstructionFallback() {
+    // fallbacks are not allowed when using h2c prior knowledge
+    new OkHttpClient.Builder()
+            .protocols(Arrays.asList(Protocol.H2C, Protocol.HTTP_1_1))
+            .build();
+  }
+
+
```
yschimke: not double line

okhttp-tests/src/test/java/okhttp3/internal/http2/HttpOverHttp2Test.java
```diff
@@ -86,28 +89,60 @@
...
 /** Test how SPDY interacts with HTTP/2 features. */
+@RunWith(Parameterized.class)
 public final class HttpOverHttp2Test {
   private static final Logger http2Logger = Logger.getLogger(Http2.class.getName());
+  private static final SslClient sslClient = SslClient.localhost();
+
+  @Parameters
+  public static Collection<Object[]> data() {
+
+      return Arrays.asList(new Object[][] {
```
yschimke: indentation looks off

yschimke:
Looking good to me, only nits and formatting.

mockwebserver/src/test/java/okhttp3/mockwebserver/MockWebServerTest.java
```diff
@@ -456,4 +457,22 @@
     RecordedRequest request = server.takeRequest();
     assertEquals("request", request.getBody().readUtf8());
   }
+
+  @Test(expected = IllegalArgumentException.class)
```
yschimke: Same as previous comment, OkHttp tests favour explicit catch and check exception message over annotation expected tests.

jpitz:
@yschimke : done. any idea what's going on with the travis test? I'd prefer to have a green build before pushing forward

mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.java
```diff
@@ -420,7 +428,7 @@ public void processConnection() throws Exception {
             return;
           }
           socket = sslSocketFactory.createSocket(raw, raw.getInetAddress().getHostAddress(),
-              raw.getPort(), true);
+                  raw.getPort(), true);
```
swankjesse: nit: indent

mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.java
```diff
@@ -436,6 +444,10 @@ public void processConnection() throws Exception {
             protocol = protocolString != null ? Protocol.get(protocolString) : Protocol.HTTP_1_1;
           }
           openClientSockets.remove(raw);
+        } else if (protocols.contains(Protocol.H2C)) {
+          // force use of H2c
```
swankjesse: nit: I don’t think the comment is necessary!

okhttp-tests/src/test/java/okhttp3/OkHttpClientTest.java
```diff
@@ -169,4 +169,35 @@
...
+    try {
+      new OkHttpClient.Builder()
+              .protocols(Arrays.asList(Protocol.H2C, Protocol.H2C));
+      fail("When H2C is specified, no other protocol can be specified");
+    } catch (final IllegalArgumentException e) {
+      assertEquals("protocols containing h2c cannot use other protocols: [h2c, h2c]", e.getMessage());
+    }
+  }
+
+  @Test public void testH2COkHttpClientConstructionSuccess() {
```
swankjesse: nice.

okhttp-tests/src/test/java/okhttp3/OkHttpClientTest.java
```diff
@@ -169,4 +169,35 @@
...
+    final OkHttpClient okHttpClient = new OkHttpClient.Builder()
```
swankjesse: nit: `final` isn’t necessary

okhttp-tests/src/test/java/okhttp3/ProtocolTest.java
```diff
@@ -0,0 +1,34 @@
+package okhttp3;
```
swankjesse: nit: copyright header

okhttp-tests/src/test/java/okhttp3/ProtocolTest.java
swankjesse: Yes please.

okhttp-tests/src/test/java/okhttp3/ProtocolTest.java
```diff
@@ -0,0 +1,34 @@
...
+
+import org.junit.Test;
+
+import java.io.IOException;
+
+import static org.junit.Assert.assertEquals;
+
+public class ProtocolTest {
```
swankjesse: nit: 2-space indent please

okhttp-tests/src/test/java/okhttp3/internal/http2/HttpOverHttp2Test.java
```diff
@@ -86,28 +89,61 @@
 import static org.junit.Assert.fail;
 
 /** Test how SPDY interacts with HTTP/2 features. */
+@RunWith(Parameterized.class)
 public final class HttpOverHttp2Test {
   private static final Logger http2Logger = Logger.getLogger(Http2.class.getName());
+  private static final SslClient sslClient = SslClient.localhost();
+
+  @Parameters
```
swankjesse: neat

okhttp-tests/src/test/java/okhttp3/internal/http2/HttpOverHttp2Test.java
```diff
@@ -1147,6 +1183,8 @@ private int countFrames(List<String> logs, String message) {
    * <p>This test uses proxy tunnels to get a hook while a connection is being established.
    */
   @Test public void concurrentHttp2ConnectionsDeduplicated() throws Exception {
+    if (protocol == Protocol.H2C) return; // only run for http2
```
swankjesse: yeah, I think you want assumeTrue() here

okhttp/src/main/java/okhttp3/internal/connection/RealConnection.java
```diff
@@ -261,8 +261,15 @@ private void connectSocket(int connectTimeout, int readTimeout, Call call,
   private void establishProtocol(ConnectionSpecSelector connectionSpecSelector,
       int pingIntervalMillis, Call call, EventListener eventListener) throws IOException {
     if (route.address().sslSocketFactory() == null) {
-      protocol = Protocol.HTTP_1_1;
+      if (route.address().protocols().contains(Protocol.H2C)) {
```
swankjesse: nice

okhttp/src/main/java/okhttp3/internal/connection/RealConnection.java
```diff
@@ -271,16 +278,20 @@ private void establishProtocol(ConnectionSpecSelector connectionSpecSelector,
...
-          .build();
-      http2Connection.start();
+      startHttp2(pingIntervalMillis);
     }
   }
 
+  private void startHttp2(int pingIntervalMillis) throws IOException {
+    socket.setSoTimeout(0); // HTTP/2 connection timeouts are set per-stream.
+    http2Connection = new Http2Connection.Builder(true)
+            .socket(socket, route.address().url().host(), source, sink)
```
swankjesse: supernit: +4 on wrapped lines

okhttp/src/main/java/okhttp3/internal/http2/Http2Codec.java
```diff
@@ -90,13 +90,17 @@
...
+  private final Protocol protocol;
 
   public Http2Codec(OkHttpClient client, Interceptor.Chain chain, StreamAllocation streamAllocation,
       Http2Connection connection) {
     this.client = client;
     this.chain = chain;
     this.streamAllocation = streamAllocation;
     this.connection = connection;
+
+    // cache this so we don't do linear scans on every call
```
swankjesse: this comment is misleading. “linear scans” is on a list of a very small constant size.

swankjesse: (I think I’d just omit the comment)

okhttp-tests/src/test/java/okhttp3/internal/http2/HttpOverHttp2Test.java
yschimke: Can you change this so the IDE makes it clear what is running

`  @Parameters(name="{2}")`

image

yschimke:
Running fine locally in both IDE and "mvn test", so the JVM died during HttpOverHttp2 is baffling

`$ mvn test -Dtest=HttpOverHttp2Test -DfailIfNoTests=false`

I'll dig into why it's failing, particularly as I pushed you towards the parameterised test, which is a likely suspect.

yschimke:
Failing with

```
Uncaught exception in OkHttp thread "Thread-12"
java.util.concurrent.RejectedExecutionException: Task okhttp3.internal.http2.Http2Connection$6@57e4c31e rejected from java.util.concurrent.ThreadPoolExecutor@12e64786[Shutting down, pool size = 1, active threads = 0, queued tasks = 0, completed tasks = 1]
	at java.util.concurrent.ThreadPoolExecutor$AbortPolicy.rejectedExecution(ThreadPoolExecutor.java:2063)
	at java.util.concurrent.ThreadPoolExecutor.reject(ThreadPoolExecutor.java:830)
	at java.util.concurrent.ThreadPoolExecutor.execute(ThreadPoolExecutor.java:1379)
	at okhttp3.internal.http2.Http2Connection.pushResetLater(Http2Connection.java:907)
	at okhttp3.internal.http2.Http2Connection$ReaderRunnable.rstStream(Http2Connection.java:691)
	at okhttp3.internal.http2.Http2Reader.readRstStream(Http2Reader.java:242)
	at okhttp3.internal.http2.Http2Reader.nextFrame(Http2Reader.java:137)
	at okhttp3.internal.http2.Http2Connection$ReaderRunnable.execute(Http2Connection.java:608)
Last test to start was: serverSendsPushPromise_GET[h2c](okhttp3.internal.http2.HttpOverHttp2Test)
The command "mvn test -e -X -Dtest=HttpOverHttp2Test -DfailIfNoTests=false; cat okhttp-tests/target/surefire-reports/*" exited with 0.
```

yschimke:
It's something wacky with reuse of clients and mockwebserver instances and/or Http2Connections (these are used separately in both client and MockWebServer).  Anyway, the shutdown of pushExecutor causes uncaught exceptions which we barf on.

I suspect a convergence of

1) This PR now has multiple tests using push promises
2) The tests run slightly differently in travis CI and locally and Intellij
3) An existing timing bug with RejectedExecutionException thrown after shutdown of a connection
4) An existing test bug with MockWebserver or more likely client usage e.g. we are actually sharing connections across tests?

okhttp-tests/src/test/java/okhttp3/OkHttpClientTest.java
jpitz: habbit. we prefer having every thing be final.

okhttp-tests/src/test/java/okhttp3/ProtocolTest.java
jpitz: added

okhttp-tests/src/test/java/okhttp3/ProtocolTest.java
jpitz: fixed

okhttp-tests/src/test/java/okhttp3/internal/http2/HttpOverHttp2Test.java
```diff
@@ -84,30 +87,64 @@
...
+import static org.junit.Assume.assumeTrue;
 
 /** Test how SPDY interacts with HTTP/2 features. */
+@RunWith(Parameterized.class)
 public final class HttpOverHttp2Test {
   private static final Logger http2Logger = Logger.getLogger(Http2.class.getName());
+  private static final SslClient sslClient = SslClient.localhost();
+
+  @Parameters(name = "{2}")
+  public static Collection<Object[]> data() {
```
yschimke: This makes the client instances used across test runs.  Can we change this to

```
@Parameters(name = "{0}") 
public static Collection<Protocol> data() {
   return Arrays.asList(Protocol.H2C, Protocol.HTTP_2);
}
```

okhttp-tests/src/test/java/okhttp3/internal/http2/HttpOverHttp2Test.java
```diff
@@ -84,30 +87,64 @@
...
-  private SslClient sslClient = SslClient.localhost();
-  private HostnameVerifier hostnameVerifier = new RecordingHostnameVerifier();
   private OkHttpClient client;
   private Cache cache;
   private TestLogHandler http2Handler = new TestLogHandler();
   private Level previousLevel;
+  private String scheme;
+  private Protocol protocol;
+
+  public HttpOverHttp2Test(OkHttpClient client, String scheme, Protocol protocol) {
```
yschimke:
```
  public HttpOverHttp2Test(Protocol protocol) {
    this.protocol = protocol;
    this.client = protocol == Protocol.HTTP_2 ? buildClient() : buildH2cClient();
    this.scheme = protocol == Protocol.HTTP_2 ? "https" : "http";
  }

  private static OkHttpClient buildH2cClient() {
    return defaultClient().newBuilder()
        .connectionPool(new ConnectionPool())
        .protocols(Arrays.asList(Protocol.H2C))
        .build();
  }

  private static OkHttpClient buildClient() {
    return defaultClient().newBuilder()
        .protocols(Arrays.asList(Protocol.HTTP_2, Protocol.HTTP_1_1))
        .dns(new SingleInetAddressDns())
        .sslSocketFactory(sslClient.socketFactory, sslClient.trustManager)
        .hostnameVerifier(new RecordingHostnameVerifier())
        .connectionPool(new ConnectionPool())
        .build();
  }
```

then in tear down

```
  @After public void tearDown() throws Exception {
    Authenticator.setDefault(null);
    http2Logger.removeHandler(http2Handler);
    http2Logger.setLevel(previousLevel);

    client.connectionPool().evictAll();
  }
```

mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.java
```diff
@@ -223,6 +227,10 @@ public void setProtocols(List<Protocol> protocols) {
     this.protocols = protocols;
   }
 
+  public List<Protocol> protocols() {
+    return Collections.unmodifiableList(protocols);
```
swankjesse: when we set the protocols we call `Util.immutableList()` so I think unmodifiableList() is not necessary

okhttp-tests/src/test/java/okhttp3/OkHttpClientTest.java
```diff
@@ -169,4 +169,35 @@
...
     }
   }
+
+  @Test public void testH2COkHttpClientConstructionFallback() {
+    // fallbacks are not allowed when using h2c prior knowledge
+    try {
+      new OkHttpClient.Builder()
+              .protocols(Arrays.asList(Protocol.H2C, Protocol.HTTP_1_1));
+      fail("When H2C is specified, no other protocol can be specified");
+    } catch (final IllegalArgumentException e) {
```
swankjesse: nit: more finals

swankjesse: also I really like `expected` instead of `e` for these

yschimke:
This push failure happened on master https://api.travis-ci.org/v3/job/346125271/log.txt

So I suspect it is an existing problem, unrelated to this test now.  The reused static clients you had in an earlier version would have made it a lot more noticeable.

jpitz:
Alright, I believe all feedback has been address at this point. Let me know if theres anything else y'all need from me 😄

swankjesse:
Thanks @jpitz. Very nice first PR!

jpitz:
No problem. Happy to help! Let me know if any issues arise and would like me to assist.

Issue #3261, opened 2017-04-01
Ping to keep HTTP/2 connections alive


wanggang0813:
The default HTTP/2 connection idle timeout is 5 min, however, if the connection is  breakdown within 5 minutes. For example, LVS would erase idle connection at 90 seconds, this action would result in request read timeout.  Sometimes the read timeout would take 15-30 seconds long, it's not good  for user experience.

So, it's better to keep the connection alive within 5 min by sending ping frame to peer,  if  ping rtt is timeout or connection would be dead, close it. 

Best wishes!

wanggang0813:
Appreciated for your quick reply!

PR #3874, opened 2018-02-21
First step towards HTTP/2 automatic pings.

swankjesse:
This changes the executor services in HTTP/2 connection to split up
listener executors (of which there may be many threads) from writer
executors (of which there is one per connection).

This adds a scheduled executor job to send pings on an interval. We
don't yet fail the connection if pongs are not received.

For follow-up there's some work to remove exising APIs to send pings
through the HTTP/2 APIs. These are unused and make it awkard to
confirm that only the automatic pings are being replied to.

See https://github.com/square/okhttp/issues/3261

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/http2/Http2Connection.java
+++ b/okhttp/src/main/java/okhttp3/internal/http2/Http2Connection.java
@@ -27,6 +27,8 @@ import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.LinkedBlockingQueue;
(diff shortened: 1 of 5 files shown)


okhttp/src/main/java/okhttp3/internal/http2/Http2Connection.java
```diff
@@ -40,6 +42,7 @@
 import okio.ByteString;
 import okio.Okio;
 
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
```
PromanSEW: TimeUnit is already imported in L34 here

PR #3875, opened 2018-02-21
OkHttpsURLConnection.setSSLSocketFactory(null): throw.

15characterlimi:
OkHttpsURLConnection extends javax.net.ssl.HttpsURLConnection.
The super class documentation says that setSSLSocketFactory(null)
throws IllegalArgumentException, but this implementation didn't
do that.

This CL fixed OkHttpsURLConnection.setSSLSocketFactory(null) to
throw, as specified.

Test: not done locally, relying on Travis.

### DIFF
--- a/okhttp-urlconnection/src/main/java/okhttp3/internal/huc/OkHttpsURLConnection.java
+++ b/okhttp-urlconnection/src/main/java/okhttp3/internal/huc/OkHttpsURLConnection.java
@@ -58,6 +58,9 @@ public final class OkHttpsURLConnection extends DelegatingHttpsURLConnection {
   }
 
   @Override public void setSSLSocketFactory(SSLSocketFactory sslSocketFactory) {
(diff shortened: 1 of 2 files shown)


okhttp-urlconnection/src/main/java/okhttp3/internal/huc/OkHttpsURLConnection.java
```diff
@@ -58,6 +58,9 @@ public OkHttpsURLConnection(OkHttpURLConnection delegate) {
   }
 
   @Override public void setSSLSocketFactory(SSLSocketFactory sslSocketFactory) {
+    if (sslSocketFactory == null) {
+        throw new IllegalArgumentException("null sslSocketFactory");
```
PromanSEW: "sslSocketFactory == null"

swankjesse:
Merged manually!

PR #3877, opened 2018-02-22
Fix swapped name and value tests.

NightlyNexus:
`ofMapRejectsNullCharInName` was checking for null in the value, and `ofMapRejectsNullCharInValue` was checking for null in the name.


swankjesse:
👍

PR #3878, opened 2018-02-22
Fail HTTP/2 connections if the pong is not received.

swankjesse:
This removes the Ping abstraction from our http2 package. This was written as
if it were a public API, but never exposed to any callers but tests. Removing
it makes it easier to lock down how pings are used.

This also removes the NullServer class, replacing it with a new SocketPolicy
on MockWebServer. The new SocketPolicy, STALL_SOCKET_AT_START, allows us to
do TLS and negotiate an HTTP/2 connection without actually building a proper
Http2Connection and without responding to pings.

The behavior in HTTP/2 connections is modeled after our behavior in web sockets.
We count outgoing pings, incoming pongs, and confirm that the pongs are keeping
up. If later we make this policy more sophisticated we can track the changes
in both places.

Closes: https://github.com/square/okhttp/issues/3261


okhttp/src/main/java/okhttp3/internal/http2/Http2Connection.java
```diff
@@ -755,13 +741,13 @@ private void applyAndAckSettings(final Settings peerSettings) {
 
     @Override public void ping(boolean reply, int payload1, int payload2) {
       if (reply) {
-        Ping ping = removePing(payload1);
-        if (ping != null) {
-          ping.receive();
+        synchronized (Http2Connection.this) {
+          receivedPongCount++;
```
yschimke: Is there value in checking payload1 and payload2 in this case? e.g. whether it is a reply to a PING, or a standalone keepalive PING ACK?  Seems like that is a possibility, but I'm not very familiar with these in practice.

okhttp/src/main/java/okhttp3/internal/http2/Http2Connection.java
```diff
@@ -350,60 +351,56 @@ void writeWindowUpdateLater(final int streamId, final long unacknowledgedBytesRe
...
+  void writePing(boolean reply, int payload1, int payload2) {
+    if (!reply) {
+      int sentPingCount;
+      int receivedPongCount;
+      synchronized (this) {
+        sentPingCount = this.sentPingCount;
+        receivedPongCount = this.receivedPongCount;
+        this.sentPingCount++;
+      }
+      if (sentPingCount > receivedPongCount) {
```
yschimke: similar to the websocket case, do we ever see unrequested PING-ACKs from servers in the wild?

yschimke:
Nothing blocking, just flagging whether we should check the payload matches in ACKs and whether we should consider unrequested PING-ACKs

swankjesse:
I think we can change both this and websockets to track a boolean `awaitingPong` instead of counts. I'd like to do that in a follow up once there's parity between the two.

Checking payloads is possible but I don't think it's worth the extra bookkeeping.

okhttp/src/main/java/okhttp3/internal/http2/Http2Connection.java
swankjesse: No idea! Will address in follow up to track whether we're awaiting a pong rather than the counts.

PR #3885, opened 2018-02-23
changes of closing the stream on ioexception

CMahesh:
The followUpRequest IOException is catch and the streamallactor is closed and the exception is re throws the IOException

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/http/RetryAndFollowUpInterceptor.java
+++ b/okhttp/src/main/java/okhttp3/internal/http/RetryAndFollowUpInterceptor.java
@@ -157,7 +157,7 @@ public final class RetryAndFollowUpInterceptor implements Interceptor {
 
       Request followUp = null;
       try {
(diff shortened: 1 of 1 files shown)


okhttp/src/main/java/okhttp3/internal/http/RetryAndFollowUpInterceptor.java
```diff
@@ -155,7 +155,13 @@ public StreamAllocation streamAllocation() {
             .build();
       }
 
-      Request followUp = followUpRequest(response, streamAllocation.route());
+      Request followUp = null;
+      try {
+        followUp = followUpRequest(response, streamAllocation.route());
+      } catch (IOException e) {
+          streamAllocation.release();
```
yschimke: Formatting looks off

CMahesh:
@yschimke  ok sure. let me add a test

yschimke:
Do you need any help with the test?

PromanSEW:
This PR can be closed as of #3971

PR #3886, opened 2018-02-24
Release writer threads when the connection is shut down.

swankjesse:
Otherwise these threads stick around forever. Yikes.


okhttp/src/main/java/okhttp3/internal/http2/Http2Connection.java
```diff
@@ -320,32 +321,40 @@ void addBytesToWriteWindow(long delta) {
...
+          try {
+            writeSynReset(streamId, errorCode);
+          } catch (IOException e) {
+            failConnection();
+          }
         }
-      }
-    });
+      });
+    } catch (RejectedExecutionException ignored) {
```
swankjesse: GitHub isn’t doing a great job showing this diff. It’s just adding this catch clause everywhere we enqueue a job on an executor that might have been shutdown.

PR #3896, opened 2018-02-27
fix #3895: Set Conscrypt.setUseEngineSocket per SocketFactory

### LINKED ISSUE #3895: ConscryptPlatform platform should not globally configure Conscrypt.setUseEngineSocketByDefault
What kind of issue is this?

 - [x] Bug report.

`ConscryptPlatform.buildIfSupported()` globally sets `Conscrypt.setUseEngineSocketByDefault(true)`, which causes HttpUrlConnections used across the jvm to behave differently (connections external to okhttp, okhttp itself works fine). In my case I'm seeing socket timeouts causing tests to fail where there aren't problems with `setUseEngineSocketByDefault` disabled.

Proposed Fix:
We can invoke `Conscrypt.setUseEngineSocket(SSLSocketFactory, true)` on the `SSLSocketFactory` instance rather than setting global configuration.

cakofony:
This change avoids static configuration which may cause issues
for applications using both OkHttp and HttpsUrlConnections.

### DIFF
--- a/okhttp/src/main/java/okhttp3/OkHttpClient.java
+++ b/okhttp/src/main/java/okhttp3/OkHttpClient.java
@@ -256,6 +256,10 @@ public class OkHttpClient implements Cloneable, Call.Factory, WebSocket.Factory 
       this.certificateChainCleaner = CertificateChainCleaner.get(trustManager);
     }
 
+    if (sslSocketFactory != null) {
+      Platform.get().configureSslSocketFactory(sslSocketFactory);
+    }
+
     this.hostnameVerifier = builder.hostnameVerifier;
     this.certificatePinner = builder.certificatePinner.withCertificateChainCleaner(
         certificateChainCleaner);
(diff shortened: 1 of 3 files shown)


cakofony:
tagging @yschimke

yschimke:
I'm not against this, I did a similar PR but decided against landing it

https://github.com/square/okhttp/pull/3868/files

Partly because of this comment

https://github.com/google/conscrypt/pull/430#pullrequestreview-98607500

yschimke:
@cakofony what is the impact on you. Is it breaking usage, or are you concerned about performance? I’m mainly curious how it showed up.

cakofony:
@yschimke It causes HttpUrlConnections to block until socket timeout where I would expect them to complete successfully very quickly.

It doesn't seem to happen on all uses of HttpsUrlConnection, but fairly consistently for a simple api that loads my application version (one of the fastest calls in the app, goes from .5ms to 60s timeout, and fail)

I've not been able to produce a simple repro quite yet, but I can fix the problem on my stack by disabling okhttp conscrypt integration (or flipping the default false after the okhttp platform initializes).

yschimke:
I think the test failure in CI will be fixed by this https://github.com/square/okhttp/pull/3900

yschimke:
@cakofony have you submitted http://square.github.io/okhttp/#contributing

cakofony:
Yep!

PR #3905, opened 2018-03-02
DiskLruCache snapshot() method removing entries that are being cached

marcinbak:
When using DiskLruCache, hasNext() method of Iterator<Snapshot> returned by snapshot() method iterates through lruEntries, however if entry is not completed yet, it is removed (together with dirtyFiles) as cleanFiles don't exist yet - during saving response to disk. This caused sometimes file not being cached at all.


okhttp/src/main/java/okhttp3/internal/cache/DiskLruCache.java
```diff
@@ -753,6 +753,8 @@ private void validateKey(String key) {
 
           while (delegate.hasNext()) {
             Entry entry = delegate.next();
+            if (entry == null || !entry.readable) continue; // Entry during edit
```
yschimke: Can entry ever be null here?

Is it simpler to move this check inside of entry.snapshot()?

yschimke:
Can you add a reproducing test to DiskLruCacheTest?

okhttp/src/main/java/okhttp3/internal/cache/DiskLruCache.java
marcinbak: Not sure if entry can be ever null. I don't know the internal assumptions of OkHttp and based my code on line 42 in Snapshot get(String key) method, where same check is being done.

marcinbak:
Writing a test will be tricky, as this requires two concurrent operations to happen. I can give it a try if I find some free time.

yschimke:
At a glance it didn't look like concurrency was required e.g.


    DiskLruCache.Editor creator = cache.edit("k1");
    setString(creator, 0, "ABC");
    setString(creator, 1, "DE");
    assertNull(creator.newSource(0));
    assertNull(creator.newSource(1));
    creator.commit(); // readable = true here

yschimke:
@marcinbak ping

marcinbak:
@yschimke have been super busy past months with projects and will have no time till end of next week for sure. I'll come back and write tests in two weeks roughly. Setting up a reminder ;)

yschimke:
@marcinbak not a problem, no explanations needed.  Just wanted to know if the PR was still live.  Thanks, contributions always welcome.

InI4:
Just from reading this description. I am having a rare (one in several hundrets), non deterministic condition where Picasso (via okhttp 3.11.0) fails to give me a picture. Can this be connected?

marcinbak:
@InI4 yes, it might be related
@yschimke finally added a test, took some time, but hope it helps

yschimke:
Rebased through the kotlin changes 

https://github.com/square/okhttp/tree/marcinbak-master

marcinbak:
@yschimke jfyi, merged master to my PullRequest and fixed tests. Is there a chance you're going to merge it? :)

yschimke:
It's probably a 4.1 thing, but I'll defer to @swankjesse.  FWIW I was going to close, but decided to try fixing instead since delay is not on your side.

InI4:
Does this patch have a chance to go to 3.12.x? 
I understand, this is not really a deep security thing, but I still hope for my occasional Picasso quirks.

swankjesse:
Sure, let’s attempt to land this for all 3 of our release branches.

okhttp/src/main/java/okhttp3/internal/cache/DiskLruCache.kt
```diff
@@ -708,6 +708,8 @@ class DiskLruCache internal constructor(
 
           while (delegate.hasNext()) {
             val entry = delegate.next()
+            if (entry == null || !entry.readable) continue // Entry during edit
```
swankjesse: I think this null check is unnecessary?

yschimke: I'll cleanup.  And then cherry pick onto the other branches.

Issue #2424, opened 2016-03-21
Creating RequestBody from an InputStream


jianghaolu:
I was looking at RequestBody.java and I can't find an overload for create() like this:
```java
RequestBody.create(MediaType contentType, InputStream stream)'
```
Is there a reason not to implement this? This is pretty useful when piping from one stream into an upload job. The overload accepting a `File` is useful for files but that's not the only scenario where buffering in memory does not work.

Please shed some lights if I'm missing something. Thanks!

swankjesse:
In general we want to be able to replay a request body multiple times. But input streams can only be consumed once.

jianghaolu:
So if I provide a large file it will be read multiple times?

swankjesse:
Yep. For example, if you get a 308 redirect.

jianghaolu:
Or retries. Makes sense! Thanks.

MIFind:
Course of project , i need the overload ,  do you find answers?

JakeWharton:
The `Uri` is what you should create a `RequestBody` from then. As if it needs to be written twice it can have its stream opened twice.

bpappin:
@JakeWharton yes i agree, but there is no `RequestBody.create(MimeType, Uri)` that i can see.

The only way I can see to do it is to override RequetBody and source from the file descriptor.

JakeWharton:
Yep. That's the right way.

PR #3912, opened 2018-03-05
Add ability to use InputStream as source of RequestBody

yfouquet:
Following https://github.com/square/okhttp/issues/2424, support `InputStream` in `RequestBody#create`.

The implementation is taking feedbacks for last attempt: https://github.com/square/okhttp/pull/1038.
As retries may be required:
- we cannot close the input stream, it remains the responsability of the caller.
- `BufferedInputStream` is used (instead of any `InputStream`) because it can be read several times.

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/RequestTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/RequestTest.java
@@ -15,7 +15,10 @@
  */
 package okhttp3;
 
+import java.io.BufferedInputStream;
+import java.io.ByteArrayInputStream;
 import java.io.File;
+import java.io.FileInputStream;
 import java.io.FileWriter;
 import java.io.IOException;
 import java.net.URI;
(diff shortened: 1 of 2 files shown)


swankjesse:
I would prefer to not add this because the performance consequences are so unexpected. Similarly, the fact that we can’t automatically close this stream is going to be surprising to callers.

It needs to load the entire request body into memory. I’d prefer that if a caller is writing a file, they just use a separate FileInputStream on each call to `writeTo()`.

yschimke:
I really like this and think it is a great feature I could have used myself. It is, however, likely a source of support issues for the OkHttp maintainers.  Maybe it is better in a single purpose library to publish from your own github project?  The change seems pretty clean so it doesn't need to be part of okhttp core.

yfouquet:
Thanks for your quick feedbacks @swankjesse and @yschimke.
I'll find an alternative.

Issue #3915, opened 2018-03-06
Frequent cancelling the HTTP/2 requests will cause client never sends WINDOW_UPDATE frame


jifang:
Description: When downloading a lot of medium-sized files from Nginx HTTP/2 server, if requests are frequently canceled (RST_STREAM), eventually the connection will timeout and not be able to send/receive any data from the server.

On the server side configure the Nginx server to support HTTP/2 with a self-signed certificate. And host about 100 files with size about 500k each. 
On the client side, I built an Android test, using a single thread to download the file from file no.1 to file no.100 consequently. During each download, the request is canceled and then proceed to download next file. Eventually, the request will timeout. Depending on the setup the timeout may happen on different download. But once the testing environment is set, it always failed on the same file.

Server info: nginx/1.13.10 
Client info: Android 6.0 with OKHttp 3.10.0 

This issue is reproducible on multiple servers. For example, Go HTTP/2 server and AWS CloudFront.

A filtered (sid:75) server log shows that the header frame has been sent on 14:39:32 and then timed out after 60 seconds. Note, it is not always the header frame though. Sometimes it is the data frame not being sent out.
```
2018/03/05 14:39:32 [debug] 92282#0: *4 http2 frame type:1 f:5 l:14 sid:75
2018/03/05 14:39:32 [debug] 92282#0: *4 http2 HEADERS frame sid:75 depends on 0 excl:0 weight:16
2018/03/05 14:39:32 [debug] 92282#0: *4 http2 frame out: 00007FD4B082ABA0 sid:75 bl:1 len:122
2018/03/05 14:39:32 [debug] 92282#0: *4 http2 frame sent: 00007FD4B082ABA0 sid:75 bl:1 len:122
2018/03/05 14:40:32 [debug] 92282#0: *4 http2 send RST_STREAM frame sid:75, status:1
```
In the meantime, the client log shows the header is not received until 14:40:32
```
03-05 14:39:32.229 5979-5995/? V/VB_Network: Sending request https://10.0.0.186/files/37 on Connection{10.0.0.186:443, proxy=DIRECT hostAddress=/10.0.0.186:443 cipherSuite=TLS_RSA_WITH_AES_128_CBC_SHA protocol=h2}
                                             User-Agent: 
                                             Host: 10.0.0.186
                                             Connection: Keep-Alive
                                             Accept-Encoding: gzip
[...]

PR #3920, opened 2018-03-12
Do not alter connection flow-control window on SETTINGS frame.

dave-r12:
This might have been a difference between SPDY and HTTP/2. Reviewing the
HTTP/2 spec, only a WINDOWS_UPDATE frame can change the flow-control
window for a connection.


dave-r12:
I was doing some research on #3915 and noticed that after sending 16MiB of data on a stream the connection flow-control window was not 0.

I've read the spec a few times now and think this is right but please confirm. It seems the idea of shrinking the connection window size is not really accounted for.

yschimke:
That matches my reading of

http://http2.github.io/http2-spec/index.html#InitialWindowSize
and
http://http2.github.io/http2-spec/index.html#SettingValues

swankjesse:
Nice find. Surprised how long we got away with this.

Issue #3940, opened 2018-03-27
Why are cancel events in HTTP2 causing the connection to be closed?


oliveroneill:
https://github.com/square/okhttp/blob/878bde8fd6939b508adeb21e2911e390cfe27326/okhttp/src/main/java/okhttp3/internal/connection/StreamAllocation.java#L436

Only `REFUSED_STREAM` errors allow the reuse of a connection. I believe it should be acceptable for a cancelled stream to be able to reuse the connection. Am I misunderstanding what `ErrorCode.CANCEL` means?

swankjesse:
Looks like a bug!

yschimke:
Pull Requests are always welcome, we are happy to review as long as you sign the Contributor Agreement when you submit it.  n.b. If it's high enough priority and the PR doesn't look like it's coming, then someone else may fix instead.  These issues are not actively assigned to anyone.

And Thanks!

oliveroneill:
There is a test that uses `ErrorCode.CANCEL` to kill the connection. The test is `recoverFromConnectionNoNewStreamsOnFollowUp` in `HttpOverHttp2Test`. This makes me wonder if it's intended behaviour?

yschimke:
Completed PR - this appears closed already

PR #3949, opened 2018-04-01
Fix stream cancel error stopping reused connection

oliveroneill:
Fix issue #3940


oliveroneill:
Closing for now, as there still seems to be issues with streams being closed unnecessarily. May re-open later.

PR #3952, opened 2018-04-03
Use double-checked singleton patterns.

HaoFeiWang:
The efficiency of Double-Check is higher.

### DIFF
--- a/okhttp/src/main/java/okhttp3/Dispatcher.java
+++ b/okhttp/src/main/java/okhttp3/Dispatcher.java
@@ -42,7 +42,7 @@ public final class Dispatcher {
   private @Nullable Runnable idleCallback;
 
   /** Executes calls. Created lazily. */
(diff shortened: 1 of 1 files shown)


yschimke:
Is double checked locking safe in legacy Android phones?

Just curious: is this showing up in some profile as a hotspot?

swankjesse:
I don’t like this. The performance benefits are negligible and the readability costs are real.

artem-zinnatullin:
Hm, I like it tbh, only first call (realistically) needs synchronization, but current implementation synchronizes all the time. 

This PR will avoid synchronization in most of use-cases at cost of reading volatile which is not that bad.

But benchmark will be great.

swankjesse:
This increases the amount of code to write, read, compile, download, and execute. It needs to be motivated by a real problem.

I'm closing this. If you can show how this fixes a performance problem, we can reconsider.

Issue #3872, opened 2018-02-20
When Authenticator.authenticate() throws an exception the StreamAllocator isn't released


ubuntudroid:
First of all sorry for not providing a failing test here, but I struggled to do that for this case.

I was just implementing an Authenticator when I found that when `Authenticator.authenticate()` throws an exception (it has a catched `IOException` in its signature, so that seems to be an expected case) the `RetryAndFollowUpInterceptor` will never release its `StreamAllocation` - as opposed to when returning `null` in `authenticate()`.

Here are the according lines in the `RetryAndFollowUpInterceptor`:

```java
Request followUp = followUpRequest(response, streamAllocation.route());

if (followUp == null) {
    if (!forWebSocket) {
      streamAllocation.release();
    }
    return response;
}
```

Shouldn't there be a try catch around the call to `followUpRequest()` to handle that case and and then re-throw the exception?

CMahesh:
I am new to open source community. Can I work it ?

yschimke:
@CMahesh there is a contributor agreement to submit, http://square.github.io/okhttp/#contributing

CMahesh:
@yschimke  Thank you.. 👍  I have submitted the contributor agreement.  I will clone the repo and push the changes.

CMahesh:
okhttp3.internal.http.RetryAndFollowUpInterceptor

 Request followUp = null;
      try{
    	  followUp=followUpRequest(response, streamAllocation.route());
      }
      catch(IOException e){
    	  streamAllocation.release();
    	  throw e;
      }
      
I have made the changes after downloading the repo.
@yschimke  What is the procedure to commit the changes ??

yschimke:
There are a number of tutorials you can read

https://help.github.com/articles/creating-a-pull-request/

The goal is to get your changes showing up here https://github.com/square/okhttp/pulls

CMahesh:
https://github.com/square/okhttp/pull/3885

changes commited and pull request created.

ubuntudroid:
@yschimke as per #3971 this seems to be fixed and released. I guess this can be closed.

PR #3971, opened 2018-04-16
Release the connection when an Authenticator throws an exception.

dave-r12:
Closes: https://github.com/square/okhttp/issues/3872


dave-r12:
I stoles the changes from https://github.com/square/okhttp/pull/3885 ... and added some tests!

yschimke:
hmmm.... from https://github.com/square/okhttp/pull/3971/commits I thought this would keep the original author.  Not sure it did, was that because of squash merge?

swankjesse:
LGTM

dave-r12:
Ahh, yeah I tried to retain the original author. My bad.

PR #3977, opened 2018-04-21
alpn boot versions 171+172

yschimke:
No jetty release... yet.

Nothing obviously coming from http://www.oracle.com/technetwork/java/javase/8all-relnotes-2226344.html

Should fix travis build for now


yschimke:
Versions are ultimately correct https://github.com/jetty-project/jetty-alpn/blob/master/docs/version_mapping.properties

Issue #3980, opened 2018-04-26
Use SSLEngine.setApplicationProtocols() and SSLSocket.getApplicationProtocol() for Android Q


zjupure:
okhttp use some reflection to config the network, such as ALPN, when run in android P, system will logcat the warning in the logcat.

see https://developer.android.com/preview/restrictions-non-sdk-interfaces.html for Android P behavior change

```
Accessing hidden method Lcom/android/org/conscrypt/OpenSSLSocketImpl;->setAlpnProtocols([B)V (light greylist, reflection)
Accessing hidden method Lcom/android/org/conscrypt/OpenSSLSocketImpl;->getAlpnSelectedProtocol()[B (light greylist, reflection)
Accessing hidden method Lcom/android/org/conscrypt/TrustManagerImpl;-><init>(Ljava/security/KeyStore;)V (light greylist, reflection)
Accessing hidden method Ldalvik/system/CloseGuard;->get()Ldalvik/system/CloseGuard; (light greylist, reflection)
Accessing hidden method Ldalvik/system/CloseGuard;->open(Ljava/lang/String;)V (light greylist, reflection)
Accessing hidden method Ldalvik/system/CloseGuard;->warnIfOpen()V (light greylist, reflection)
```

We look through okhttp source, and found the reflection are all in `AndroidPlatform.java`.

okhttp has some solution or workaround not to reflet system non-open api?

swankjesse:
First two: https://issuetracker.google.com/issues/78796431

swankjesse:
For CloseGuard, we probably want to just delete the Android-specific overrides. We don’t particularly benefit from using CloseGuard instead of java.util.logging.Logger for configuration.

swankjesse:
@zjupure which version of OkHttp were you running? I couldn’t find the reflective call to `TrustManagerImpl;-><init>` in the current source tree.

af913337456:
the same problem hit me

AntoinePurnelle:
Strict mode now allows to have a stacktrace

```
StrictMode.setVmPolicy(new StrictMode.VmPolicy.Builder()
                .detectNonSdkApiUsage()
                .penaltyLog()
                .build());
```

```
06-06 19:52:56.935 8671-8671/? D/StrictMode: StrictMode policy violation: android.os.strictmode.NonSdkApiUsedViolation: Ldalvik/system/CloseGuard;->get()Ldalvik/system/CloseGuard;
[...]

PR #3983, opened 2018-04-30
Remove Android Hidden API of CloseGuard.

iNoles:
Fixed partial bugs from #3980


JakeWharton:
I don't think this is the right approach. We should file a bug requesting access to this functionality and only remove it if it gets promoted from light grey to dark grey (or black).

swankjesse:
Thanks for the PR, and sorry for sending you off in the wrong direction on it. I’m going to close this for now, but we might revisit when CloseGuard’s visibility changes.

PR #3990, opened 2018-05-07
Including support for detour in proxy case configured

diegocsilva:
- Included implementation of ProxySelector (AutomatiProxySelector), which consists in selecting a particular proxy according to the request URI.

- Included List in OkHttpClient to save hosts that do not need to use proxy in their requests (noProxyForHosts).

- Changed OkHttpClient build () to set an AutomatiProxySelector when there is a custom proxy set in the OkHttpClient and a noProxyForHosts List, then the implemented ProxySelector will automatically set between the custom Proxy (when the host is not in noProxyForHosts) and ProxySelector.getDefault () (when the host is in noProxyForHosts).

### DIFF
--- a/okhttp/src/main/java/okhttp3/AutomaticProxySelector.java
+++ b/okhttp/src/main/java/okhttp3/AutomaticProxySelector.java
@@ -1,20 +1,14 @@
 package okhttp3;
 
-import sun.net.spi.DefaultProxySelector;
-
 import java.io.IOException;
 import java.net.Proxy;
+import java.net.ProxySelector;
 import java.net.SocketAddress;
 import java.net.URI;
 import java.util.Collections;
 import java.util.List;
 
-public class AutomaticProxySelector extends DefaultProxySelector {
-
-    public AutomaticProxySelector(List<String> noProxyForHosts, List<Proxy> proxies) {
-        this.noProxyForHosts = noProxyForHosts;
-        this.proxies = proxies;
-    }
+public class AutomaticProxySelector extends ProxySelector {
 
     private List<String> noProxyForHosts;
     private List<Proxy> proxies;
@@ -43,5 +37,13 @@ public class AutomaticProxySelector extends DefaultProxySelector {
             }
         }
         return false;
+    }
+
+    public void setNoProxyForHosts(List<String> noProxyForHosts) {
+        this.noProxyForHosts = noProxyForHosts;
+    }
+
+    public void setProxies(List<Proxy> proxies) {
+        this.proxies = proxies;
     }
 }
--- a/okhttp/src/main/java/okhttp3/AutomaticProxySelector.java
+++ b/okhttp/src/main/java/okhttp3/AutomaticProxySelector.java
@@ -1,0 +1,47 @@
+package okhttp3;
+
+import sun.net.spi.DefaultProxySelector;
+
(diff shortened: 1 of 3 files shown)


yschimke:
Could you explain more the logic behind this change? I have two specific concerns with it

1) It causes additional OkHttp instances per call.  
2) IIUC this functionality is already supported behind the ProxySelector interface and ProxySelector.getDefault().  Android and Standard (http.nonProxyHosts) JVMs both take into account non proxy host configs.

diegocsilva:
Hello @yschimke , well let's see, this is an implementation that I thought after dealing with this problem https://github.com/OpenFeign/feign/issues/690, when using spring-cloud-feign, as in requests using feign I can not control the proxy deviation, because the class is built via AOP, I saw no other way than to create a producer for the OkHttpClient already configured before and pass the list with the HOSTS that do not need to be accessed through the proxy, from there in the my requests okhttp itself would be responsible for whether or not to use the proxy depending on the host. I thought of a way to improve and would like to work on this idea, if you have in mind any other way you would like to know.

yschimke:
Can you implement a custom ProxySelector? Apply any AOP to your ProxySelector implementation?

diegocsilva:
Hi @yschimke , I've changed my implementation idea according to your ProxySelector.getDefault () hint. The idea now is to select between the custom proxy defined in the OkHttpClient build and the ProxySelector.getDefault () according to the request host.

yschimke:
@diegocsilva My question was more, if you implement a ProxySelector, why do you also need to make changes to okhttp as well.  What is missing with the current API and functionality?

diegocsilva:
@yschimke my intention with this implementation is to increase the ease of use of api, I can actually implement a new ProxySelector and handle the use of direct proxy in my application, but I believe that my problem is also the problem of several other API users at try to implement a client with OkHttpClient for use in spring-cloud-feign. I could just ignore this, solve my problem with using proxy within my application, but I resolved to share this possible solution.

yschimke:
That seems like a good thing.  But could be achieved in multiple ways without changing core OkHttp classes

1) AutomaticProxySelector.java as an externally available library
2) AutomaticProxySelector.java as a useful optional addition to OkHttp
3) built in support in OkHttp.Builder as you've done

I'd probably argue for 1) because this isn't something most users on Android and the JVM should need.  The default implementations on those platforms already support hostname exclusions.  If they didn't then option 2) would be even more valuable.

Interested for input from others...

swankjesse:
I think AutomaticProxySelector is a better fit either as a standalone library, or perhaps as a file in our `recipes/` package. I don’t think it’s a good idea to put it in OkHttp itself, and I’m also anxious about the way this feature as-described interacts with other features of OkHttp.

For example, I could imagine users wanting a whitelist instead of a blacklist. Or policy depending on whether the URL is HTTP or HTTPS. Or wildcard hostnames.

Would you be interested in making this a standalone gist? Or perhaps contributing it to our examples?

Thanks

diegocsilva:
Hello @swankjesse , lately I have had many jobs, so it took me a long time to respond. On the question I am willing to contrubuir to the evolution of the OkHttp API.

swankjesse:
Thanks @diegocsilva. Perhaps start with a standalone gist? Or maybe an example for our samples directory?

PR #3996, opened 2018-05-11
Rename H2C to H2_PRIOR_KNOWLEDGE.

swankjesse:
The string h2c is used with cleartext upgrades. We're not doing those here,
so that identifier isn't appropriate.

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/CallTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/CallTest.java
@@ -1187,6 +1187,25 @@ public final class CallTest {
     }
   }
 
+  @Test public void httpsCallsFailWhenProtocolIsH2PriorKnowledge() throws Exception {
+    client = client.newBuilder()
+        .protocols(Collections.singletonList(Protocol.H2_PRIOR_KNOWLEDGE))
+        .build();
+
+    server.useHttps(sslClient.socketFactory, false);
+    server.enqueue(new MockResponse());
+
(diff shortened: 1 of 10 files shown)


okhttp-tests/src/test/java/okhttp3/CallTest.java
```diff
@@ -1187,6 +1187,25 @@ private void postBodyRetransmittedAfterAuthorizationFail(String body) throws Exc
     }
   }
 
+  @Test public void httpsCallsFailWhenProtocolIsH2PriorKnowledge() throws Exception {
```
dave-r12: Nice catch.

caojiajun:
while running okhttpclient 3.11.0 on my server, the http server is tomcat8 with h2c.
the okhttpclient will throw this exception ：
**okhttp3.internal.http2.StreamResetException: stream was reset: CANCEL**
and the tomcat will throw 
**org.apache.catalina.connector.ClientAbortException: java.nio.channels.ClosedChannelException**
or
**org.apache.catalina.connector.ClientAbortException: org.apache.coyote.CloseNowException: Connection [20], Stream [5], This stream is not writable**

Is there something wrong with the way I use it?
OkHttpClient okHttpClient = new OkHttpClient.Builder()
                        .protocols(Collections.singletonList(Protocol.H2_PRIOR_KNOWLEDGE))
                        .build();
inject the above okHttpClient  to the Feign Client

PR #4000, opened 2018-05-12
Add the ability to specify payload data for ping messages

tedlefter:
In RFC 6455, section 5.5.2 states that a ping message MAY include Application Data.
Adds the ability for the library user to specify the data he wishes to include in the ping message.


tedlefter:
Not sure how much control you want to provide users for this interaction. If it's sufficient to only specify a static payload for each ping, I can make the appropriate changes.

okhttp/src/main/java/okhttp3/PingPayloadProvider.java
```diff
@@ -0,0 +1,24 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
```
yschimke: I assume this should assign the Copyright (C) 2014 Square, Inc. as part of the CLA process.

okhttp/src/main/java/okhttp3/internal/ws/RealWebSocket.java
```diff
@@ -140,14 +142,24 @@
...
+      long pingIntervalMillis, @Nullable PingPayloadProvider pingPayloadProvider) {
     if (!"GET".equals(request.method())) {
       throw new IllegalArgumentException("Request must be GET: " + request.method());
     }
     this.originalRequest = request;
     this.listener = listener;
     this.random = random;
     this.pingIntervalMillis = pingIntervalMillis;
+    if(pingPayloadProvider == null) {
+      this.pingPayloadProvider = new PingPayloadProvider() {
```
yschimke: PingPayloadProvider.EMPTY?

yschimke:
Not sure if this PR will gain a lot of support, but if it goes ahead, should we also check the PONG payload matches?

tedlefter:
Thank you for the updates and suggestions.
I'm not sure I understood the last comment about checking the Pong payload.
From the code, it looks like we are sending a pong message with the payload we get in onReadPing()
Do you also mean to add tests around checking that we send a pong with the same payload we get from a ping message?

tedlefter:
From the RFC 6455 spec:
A Pong frame sent in response to a Ping frame must have identical "Application data" as found in the message body of the Ping frame being replied to.

I think the library already does that

yschimke:
On receiving a Ping, the same payload is bounced back by OkHttp.  So all good there.

However after sending a ping, no validation that the subsequent received pong matched the sent ping is done.  Also if this PR is about sending specific Ping payloads, it seems to imply that applications may want to receive the payloads also.

This seems like an incrementally useful change, but it raises two questions

1) Do we want to expose API around Pings for clients, rather than using them internally as apurely keepalive method.
2) If we have a public API for Pings, should it be more complete than this.

tedlefter:
I am not sure who would need to make/approve these decisions.
But, in my opinion, adding a more open API for ping and pongs makes the library more flexible and might cover more usages and useful cases.
One might want to use that for better logging and debugging.
Pong payloads might end up being useful.

As per the standard specs, the pongs must contain the same ping data, if the pong is being sent in the response for that ping, but a Pong frame MAY also be sent unsolicited. This serves as a unidirectional heartbeat. A response to an unsolicited Pong frame is not expected. https://tools.ietf.org/html/rfc6455#section-5.5.2
Because of this we might be hitting cases that a side might receive multiple unsolicited pongs, so failing if the Pong payload doesn't match with the previous Ping payload might add risks. It is not clear what an unsolicited pong payload might be.

I would be happy to expose Ping and Pong events in the WebSocketListener class if that is what we want to do.
If we do, I could also make the onPing() return value that would be the payload, instead of having the separate provider class. Let me know what you think about this.

yschimke:
My starting point would be, if we want an API for Pings it should be functionally complete, or should be able to move to that direction.  e.g. 

1) Allow specifying ping payload
2) Allow control of ping timing?
2) notification of pings, pongs (including unsolicited)

So I'm not super hyped on PingPayloadProvider which by name only handles point 1.

A secondary benefit of this change would be a default PingApi implementation may simplify the existing Websocket implementation by externalising that behaviour.

cc @swankjesse - do you see OkHttp having any public API for pings, or are you against that?

JakeWharton:
The ping API was explicitly removed from being publicly accessible before release. I don't think I like the abstraction as proposed since you still can initiate pings but can provide their data. I don't see what the use case there is.

swankjesse:
I think application data in pings is a mistake. OkHttp uses pings for it's own connectivity checks, and mixing in application data is a recipe for complexity.

I don't think pings are exposed by browser web socket APIs. This means that most servers won't do anything interesting with pings anyway.

PR #4014, opened 2018-05-20
[Early WIP] Websockets over HTTP/2

yschimke:
Towards https://datatracker.ietf.org/doc/draft-ietf-httpbis-h2-websockets/?include_text=1

Steps - working on step 2 currently

1. Introduce an API for UpgradeHandler
2. Refactor existing Websocket code to cleanly work with UpgradeHandler for HTTP/1.1
3. Introduce HTTP/1.1 upgrade flow
4. Introduce OkHttpClient.upgrade method 
5. Implement HTTP/2 upgrade flow

API changes

1. New public API UpgradeHandler and OkHttpClient.upgrade(UpgradeHandler) method
2. New public class Streams for "socket", two bidirectional streams, over HTTP/1.1 or HTTP/2

### DIFF
--- a/mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.java
+++ b/mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.java
@@ -57,6 +57,7 @@ import okhttp3.HttpUrl;
 import okhttp3.Protocol;
 import okhttp3.Request;
 import okhttp3.Response;
+import okhttp3.Streams;
 import okhttp3.internal.Internal;
 import okhttp3.internal.NamedRunnable;
 import okhttp3.internal.Util;
@@ -676,7 +677,7 @@ public final class MockWebServer extends ExternalResource implements Closeable {
         .build();
 
     final CountDownLatch connectionClose = new CountDownLatch(1);
-    RealWebSocket.Streams streams = new RealWebSocket.Streams(false, source, sink) {
+    Streams streams = new Streams(false, source, sink) {
       @Override public void close() {
         connectionClose.countDown();
       }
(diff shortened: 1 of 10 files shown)


yschimke:
Main API

```
public interface UpgradeHandler<T> {
  T connect(OkHttpClient client);

  boolean supportsHttp11();

  boolean supportsHttp2();

  /**
   * https://www.iana.org/assignments/http-upgrade-tokens/http-upgrade-tokens.xhtml
   */
  String connectProtocol();

  Request addUpgradeHeaders(Request upgradeRequest);

  void failConnect(Response response, IOException e) throws IOException;
}

```

JakeWharton:
The public API implications here are pretty devastating. Do we really need to alter the surface area so much?

yschimke:
I can implement it with all internal changes for now. But I’d love to be able to use with other external upgraded protocols provided by libraries. I can make that a completely separate later PR and discussion

JakeWharton:
I would prefer we decouple those efforts just for simplicity.

yschimke:
Yep, you had me at "The public API implications"

swankjesse:
I think it’d be great to get this working. I’m using a lot of web sockets in one of my current projects and this could make things a lot smoother.

How does the client indicate that we should attempt HTTP/2? Do we need to set policy? I think that’s the trickiest part of the whole business.

yschimke:
Good question. I was planning to attempt HTTP/2 once per client/host when there is existing connection. Then remember failure. Default to http/1.1 for new connections. 

I think getting in the basic support first is useful. 

OkHttp client API isn’t really helpful for per request or per host changes. 

Any suggestions?

swankjesse:
My preference is that _eventually_ it just works: if the server can do HTTP/2 web sockets we use ’em.

To get there I think we probably do a phased migration in:
 * initially it’s off by default with some hint to turn it on (maybe a header on the Request?)
 * later it’s on by default with some hint to turn it off (another header?)

In either case I think we get something like RetryAndFollowUpInterceptor to fallback to HTTP/1 if HTTP/2 doesn’t work.

PR #4018, opened 2018-05-22
update maven plugins and apache.httpcomponents/httpclient to 4.2.6

based2:
http://mvnrepository.com/artifact/org.apache.httpcomponents/httpclient

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -45,7 +45,7 @@
     <alpn.jdk7.version>7.1.2.v20141202</alpn.jdk7.version>
     <android.version>4.1.1.4</android.version>
     <animal.sniffer.version>1.15</animal.sniffer.version>
-    <apache.http.version>4.2.2</apache.http.version>
+    <apache.http.version>4.2.6</apache.http.version>
     <bouncycastle.version>1.50</bouncycastle.version>
     <guava.version>16.0</guava.version>
     <java.version>1.7</java.version>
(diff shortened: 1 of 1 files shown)


JakeWharton:
Reminder that mvnrepository is a third party index and not canonical for maven central. search.maven.org is where maven central lives.

based2:
https://archive.apache.org/dist/httpcomponents/httpclient/RELEASE_NOTES-4.5.x.txt

Release 4.2.3 -> ok, no 4.2.6
-------------------

HttpClient 4.2.3 (GA) is a bug fix release that addresses a number of issues reported since
release 4.2.2. This release also includes a thoroughly reworked NTLM authentication engine
which should result in a better compatibility with the newest Microsoft products.

Users of HttpClient 4.x are advised to upgrade.

yschimke:
The timezone bug raised by the new version of error prone seems practically unlikely but technically legit.  Only in a test thankfully.

```
[ERROR] /home/travis/build/square/okhttp/okhttp-urlconnection/src/test/java/okhttp3/UrlConnectionCacheTest.java:[1082,48] [InvalidTimeZoneID] Invalid time zone identifier. TimeZone.getTimeZone(String) will silently return GMT instead of the time zone you intended.
[ERROR]     (see http://errorprone.info/bugpattern/InvalidTimeZoneID)
```

swankjesse:
Thanks! Merged manually!

Issue #3850, opened 2018-02-15
Consider making HttpUrl.getChecked() public


grodin:
Using `HttpUrl.parse()` in Kotlin to parse constant urls that you know will parse is slightly annoying because of the nullable return. 

I've worked around it by writing a static method `parseHttpUrlOrThrow()` but this essentially duplicates the existing `HttpUrl.getChecked()`, which is unfortunately package private. It would be useful if this method was public.

JakeWharton:
We've talked about this for a long time. It's caused at least one bug in Retrofit and OkHttp thus far.

swankjesse:
Isn't it just `parse(url)!!` ?

JakeWharton:
Assuming you remember that, yes. In Java you don't. And in Kotlin it's a platform type so you sometimes are forced to.

JakeWharton:
Another advantage of a throwing version is that it's able to communicate useful information about the reason why it failed.

shaburov:
+1

yschimke:
oops PR hasn't landed yet

Issue #4030, opened 2018-05-27
Main.close() never close active connections


shaburov:
Main class close() method - **Close any persistent connections.**
```
  private void close() {
    client.connectionPool().evictAll(); // Close any persistent connections.
  }
```
ConnectionPool.evictAll() - **Close and remove all idle connections in the pool.**
```
  /**
   * Close and remove all idle connections in the pool.
   */
  public void evictAll() {
    List<RealConnection> evictedConnections = new ArrayList<>();
    synchronized (this) {
      for (Iterator<RealConnection> i = connections.iterator(); i.hasNext(); ) {
        RealConnection connection = i.next();
        if (connection.allocations.isEmpty()) {
          connection.noNewStreams = true;
          evictedConnections.add(connection);
          i.remove();
        }
      }
    }

    for (RealConnection connection : evictedConnections) {
      closeQuietly(connection.socket());
    }
  }
```
Condition `connection.allocations.isEmpty()` closes the connection only if there are no active streams in this connection

list of revisions:
1. **ConnectionPool** Rename evictAll() -> **closeIdle()**
2. Mark evictAll() method is Deprecated 
3. **ConnectionPool** Add method **closeAll()** - Close any persistent connections.
4. **Main** close() method call client.connectionPool().closeAll()
5. fix OkHttpClient javadoc
6. fix tests

Correct me if I'm wrong.

shaburov:
@swankjesse If I call the close () method in the Main class, I expect that all connections will be closed, not just those that are inactive.

swankjesse:
OkCurl’s close method is not public API.

yschimke:
It seems like we should have some API for closing all connections including active ones, as there are scenarios where callers are somewhat out of our control, and there are shared HTTP/2 connections that have inflight requests at all times.

PR #4032, opened 2018-05-27
#3850 modify "HttpUrl.getChecked()" method for public access + tests

shaburov:
Relates #3850

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/HttpUrlTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/HttpUrlTest.java
@@ -15,13 +15,9 @@
  */
 package okhttp3;
 
-import java.lang.reflect.Method;
-import java.lang.reflect.Modifier;
 import java.net.URLEncoder;
-import java.net.MalformedURLException;
 import java.net.URI;
 import java.net.URL;
-import java.net.UnknownHostException;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.LinkedHashSet;
@@ -32,7 +28,6 @@ import org.junit.Test;
 
 import static java.util.Collections.singletonList;
 import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.fail;
 
@@ -1304,60 +1299,62 @@ public final class HttpUrlTest {
     assertEquals(new URI("http://host/%25"), HttpUrl.parse("http://host/%").uri());
   }
 
-  @Test public void fromJavaNetUrl() throws Exception {
+  @Test public void parseFromJavaNetUrl() throws Exception {
     URL javaNetUrl = new URL("http://username:password@host/path?query#fragment");
-    HttpUrl httpUrl = HttpUrl.get(javaNetUrl);
-    assertEquals("http://username:password@host/path?query#fragment", httpUrl.toString());
+    HttpUrl httpUrl = HttpUrl.parse(javaNetUrl);
+    assertEquals("http://username:password@host/path?query#fragment", String.valueOf(httpUrl));
   }
 
-  @Test public void fromJavaNetUrlUnsupportedScheme() throws Exception {
+  @Test public void parseFromJavaNetUrlUnsupportedScheme() throws Exception {
     URL javaNetUrl = new URL("mailto:user@example.com");
-    assertEquals(null, HttpUrl.get(javaNetUrl));
+    assertNull(HttpUrl.parse(javaNetUrl));
   }
 
-  @Test public void fromUri() throws Exception {
+  @Test public void parseFromUri() throws Exception {
     URI uri = new URI("http://username:password@host/path?query#fragment");
-    HttpUrl httpUrl = HttpUrl.get(uri);
-    assertEquals("http://username:password@host/path?query#fragment", httpUrl.toString());
+    HttpUrl httpUrl = HttpUrl.parse(uri);
+    assertEquals("http://username:password@host/path?query#fragment", String.valueOf(httpUrl));
   }
 
-  @Test public void fromUriUnsupportedScheme() throws Exception {
+  @Test public void parseFromUriUnsupportedScheme() throws Exception {
     URI uri = new URI("mailto:user@example.com");
-    assertEquals(null, HttpUrl.get(uri));
+    assertNull(HttpUrl.parse(uri));
   }
 
-  @Test public void fromUriPartial() throws Exception {
+  @Test public void parseFromUriPartial() throws Exception {
     URI uri = new URI("/path");
-    assertEquals(null, HttpUrl.get(uri));
+    assertNull(HttpUrl.parse(uri));
   }
 
-  @Test public void fromJavaNetUrl_checked() throws Exception {
+  @Test public void getFromJavaNetUrl_checked() throws Exception {
     String namePass = "us.er-name:p@a!s s#w$o^r&d*";
     String encodedNamePass = URLEncoder.encode(namePass, "UTF-8");
     String url = "http://" + encodedNamePass + "@host/path?query#fragment";
-    HttpUrl httpUrl = HttpUrl.getChecked(url);
+    HttpUrl httpUrl = HttpUrl.get(url);
     assertEquals(url, httpUrl.toString());
   }
 
-  @Test public void fromJavaNetUrlUnsupportedScheme_checked() throws Exception {
+  @Test public void getFromJavaNetUrlUnsupportedScheme_checked() throws Exception {
     try {
-      HttpUrl.getChecked("mailto:user@example.com");
+      HttpUrl.get("mailto:user@example.com");
       fail();
-    } catch (MalformedURLException e) {
+    } catch (IllegalArgumentException e) {
+      assertEquals("Invalid URL: UNSUPPORTED_SCHEME for mailto:user@example.com", e.getMessage());
     }
   }
 
-  @Test public void fromJavaNetUrlBadHost_checked() throws Exception {
+  @Test public void getFromJavaNetUrlBadHost_checked() throws Exception {
     try {
-      HttpUrl.getChecked("http://hostw ithspace/");
+      HttpUrl.get("http://hostw ithspace/");
       fail();
-    } catch (UnknownHostException expected) {
+    } catch (IllegalArgumentException e) {
+      assertEquals("Invalid URL: INVALID_HOST for http://hostw ithspace/", e.getMessage());
     }
   }
 
   @Test public void getCheckedEmptyURL() throws Exception {
     try {
-      HttpUrl.getChecked("");
+      HttpUrl.get("");
       fail();
     } catch (IllegalArgumentException e) {
       assertEquals("URL must be specified.", e.getMessage());
@@ -1366,7 +1363,7 @@ public final class HttpUrlTest {
 
   @Test public void getCheckedNullURL() throws Exception {
     try {
-      HttpUrl.getChecked(null);
+      HttpUrl.get(null);
       fail();
     } catch (IllegalArgumentException e) {
       assertEquals("URL must be specified.", e.getMessage());
--- a/okhttp-tests/src/test/java/okhttp3/HttpUrlTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/HttpUrlTest.java
@@ -15,6 +15,9 @@
  */
 package okhttp3;
 
+import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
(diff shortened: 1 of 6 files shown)


okhttp-tests/src/test/java/okhttp3/internal/http2/HttpOverHttp2Test.java
```diff
@@ -1266,6 +1263,10 @@ private int countFrames(List<String> logs, String message) {
       }
     });
 
+    // Force a fresh idle connections
+    client.connectionPool().evictAll();
+    int activeConnections = client.connectionPool().connectionCount();
```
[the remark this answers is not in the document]
shaburov: https://github.com/square/okhttp/issues/4030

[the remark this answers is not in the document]
JakeWharton: Please separate this out into a separate PR.

okhttp-tests/src/test/java/okhttp3/HttpUrlTest.java
```diff
@@ -15,10 +15,9 @@
  */
 package okhttp3;
 
-import java.net.MalformedURLException;
-import java.net.URI;
-import java.net.URL;
-import java.net.UnknownHostException;
+import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
+import java.net.*;
```
JakeWharton: Revert. Star imports are forbidden.

okhttp-tests/src/test/java/okhttp3/HttpUrlTest.java
```diff
@@ -28,9 +27,7 @@
 import org.junit.Test;
 
 import static java.util.Collections.singletonList;
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNull;
-import static org.junit.Assert.fail;
+import static org.junit.Assert.*;
```
JakeWharton: Revert

okhttp-tests/src/test/java/okhttp3/HttpUrlTest.java
```diff
@@ -1328,24 +1325,26 @@
...
+    assertEquals(url, httpUrl.toString());
   }
 
-  @Test public void fromJavaNetUrlUnsupportedScheme_checked() throws Exception {
-    try {
-      HttpUrl.getChecked("mailto:user@example.com");
-      fail();
-    } catch (MalformedURLException e) {
-    }
+  @Test(expected = MalformedURLException.class)
```
JakeWharton: We forbid the use of `expected=` because it's ambiguous as to what test method body line threw (in multi-line bodies) and you cannot assert things like cause and message.

okhttp-tests/src/test/java/okhttp3/HttpUrlTest.java
```diff
@@ -1328,24 +1325,26 @@
...
-    try {
-      HttpUrl.getChecked("http://hostw ithspace/");
-      fail();
-    } catch (UnknownHostException expected) {
-    }
+  @Test(expected = UnknownHostException.class)
+  public void fromJavaNetUrlBadHost_checked() throws Exception {
+    HttpUrl.getChecked("http://hostw ithspace/");
+  }
+
```
JakeWharton: I don't think this test is necessary. We don't maintain binary compatibility through unit tests.

okhttp-tests/src/test/java/okhttp3/internal/http2/HttpOverHttp2Test.java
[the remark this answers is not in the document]
shaburov: @yschimke If this bug is confirmed, then it will be necessary to correct not only this test.
The problem is that evictAll() does not close the all active connections and this test is occasionally fail as it was always expected 1.
Without this change, mvn clean verify fail with an error in this test.
If I remove this change, then the build may not be stable.

okhttp/src/main/java/okhttp3/HttpUrl.java
```diff
@@ -918,7 +918,7 @@ public Builder newBuilder() {
    * @throws MalformedURLException if there was a non-host related URL issue
    * @throws UnknownHostException if the host was invalid
    */
-  static HttpUrl getChecked(String url) throws MalformedURLException, UnknownHostException {
+  public static HttpUrl getChecked(String url) throws MalformedURLException, UnknownHostException {
```
JakeWharton: I don't think we should just expose this with two checked exceptions. I would rather see a variant which throws IllegalArgumentException for all malformed input.

okhttp-tests/src/test/java/okhttp3/internal/http2/HttpOverHttp2Test.java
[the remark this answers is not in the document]
JakeWharton: It's not related to this desired API change so it should be separate.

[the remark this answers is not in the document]
shaburov: image

okhttp/src/main/java/okhttp3/HttpUrl.java
shaburov: IllegalArgumentException is RuntimeException.
Added check for null and empty string

okhttp-tests/src/test/java/okhttp3/internal/http2/HttpOverHttp2Test.java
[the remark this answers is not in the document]
JakeWharton: Yes. It is unrelated and should be a separate PR.

okhttp/src/main/java/okhttp3/HttpUrl.java
JakeWharton: I know. The desire in the issie is for a method that throws any exception, not necessarily a checked one. And I think a checked one is inappropriate for this case.

shaburov:
@JakeWharton @yschimke The notes are corrected. The tests were surprisingly successful.

okhttp-tests/src/test/java/okhttp3/HttpUrlTest.java
```diff
@@ -1348,6 +1355,29 @@
     }
   }
 
+  @Test public void getCheckedMethodAccess() throws Exception {
```
yschimke: I don't think this is a good precedent for checking we don't break API.  We can check this in other ways, but we could have hundreds of these tests and they have low value.

shaburov: @yschimke test deleted

okhttp-tests/src/test/java/okhttp3/HttpUrlTest.java
```diff
@@ -15,6 +15,9 @@
  */
 package okhttp3;
 
+import java.lang.reflect.Method;
```
yschimke: are these imports used?

okhttp/src/main/java/okhttp3/HttpUrl.java
```diff
@@ -915,10 +915,16 @@ public Builder newBuilder() {
    * Returns a new {@code HttpUrl} representing {@code url} if it is a well-formed HTTP or HTTPS
    * URL, or throws an exception if it isn't.
    *
+   * @param url string for check URL format
```
yschimke: Suggest: @param url The string to parse as a URL.

okhttp/src/main/java/okhttp3/HttpUrl.java
```diff
@@ -915,10 +915,16 @@ public Builder newBuilder() {
...
+   * @return instance of {@link HttpUrl}
+   *
    * @throws MalformedURLException if there was a non-host related URL issue
```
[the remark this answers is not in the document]
JakeWharton: Yeah. I don't think checked exceptions are useful here. The intent to use this method over `get` is enough to indicate that it has behave differences and requires try/catch if you desire to actually handle the failure locally.

okhttp/src/main/java/okhttp3/HttpUrl.java
```diff
@@ -915,10 +915,16 @@ public Builder newBuilder() {
...
    * @throws UnknownHostException if the host was invalid
+   * @throws IllegalArgumentException if URL not specified
    */
-  static HttpUrl getChecked(String url) throws MalformedURLException, UnknownHostException {
+  public static HttpUrl getChecked(String url)
```
swankjesse: Instead of making this public I prefer a new method, `HttpUrl.get(String)` that throws an `IllegalArgumentException` if the input is malformed.

With the right documentation our users should be able to remember that `parse()` returns null and `get()` throws on errors.

shaburov:
@JakeWharton @yschimke @swankjesse Please check the changes.
The last commit seems to satisfy all the comments and suggestions.

@yschimke I signed the CLA before pushed the first commit.

okhttp/src/main/java/okhttp3/HttpUrl.java
```diff
@@ -894,50 +894,41 @@ public Builder newBuilder() {
...
-   * Returns a new {@code HttpUrl} representing {@code url} if it is a well-formed HTTP or HTTPS
-   * URL, or null if it isn't.
+   * Returns a new {@code HttpUrl} parsed from the string value of the {@code object} if it is a
+   * well-formed HTTP or HTTPS URL, or null if it isn't.
+   *
+   * @param object for check URL format
+   * @return instance of {@link HttpUrl} or {@code null} in the case of a parse error
    */
-  public static @Nullable HttpUrl parse(String url) {
+  public static @Nullable HttpUrl parse(Object object) {
```
yschimke: Why Object as the param type?

shaburov: In our case, no matter what object pars URL, URI, InternetAddress or something else. In any case, whichever object we get, the very concept of parsing a string implies an object value of in a string.
I removed unnecessary methods that did not really make sense and duplicated the code.
~~get(URL)~~
~~get(URI)~~
etc.

okhttp/src/main/java/okhttp3/HttpUrl.java
```diff
@@ -894,50 +894,41 @@ public Builder newBuilder() {
...
-    switch (result) {
-      case SUCCESS:
-        return builder.build();
-      case INVALID_HOST:
-        throw new UnknownHostException("Invalid host: " + url);
-      case UNSUPPORTED_SCHEME:
-      case MISSING_SCHEME:
-      case INVALID_PORT:
-      default:
-        throw new MalformedURLException("Invalid URL: " + result + " for " + url);
```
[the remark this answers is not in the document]
shaburov: I understood you correctly.
get () - leave
getChacked () - return back.
If so, is it necessary to mark getChacked with Deprecated annotation?

okhttp/src/main/java/okhttp3/HttpUrl.java
```diff
@@ -894,50 +894,41 @@ public Builder newBuilder() {
...
    *
-   * @throws MalformedURLException if there was a non-host related URL issue
-   * @throws UnknownHostException if the host was invalid
+   * @param object for check URL format
+   * @return instance of {@link HttpUrl}
+   *
+   * @throws IllegalArgumentException if invalid URL specified
    */
-  static HttpUrl getChecked(String url) throws MalformedURLException, UnknownHostException {
+  public static HttpUrl get(Object object) throws IllegalArgumentException {
```
JakeWharton: I don't know where the comment went but I agree using `Object` here is super weird. We shouldn't be subverting the type system. If we want to support other types we should be supplying overloads.

okhttp/src/main/java/okhttp3/OkHttpClient.java
```diff
@@ -181,9 +179,8 @@ public void apply(ConnectionSpec tlsConfiguration, SSLSocket sslSocket, boolean
         tlsConfiguration.apply(sslSocket, isFallback);
       }
 
-      @Override public HttpUrl getHttpUrlChecked(String url)
-          throws MalformedURLException, UnknownHostException {
-        return HttpUrl.getChecked(url);
+      @Override public HttpUrl getHttpUrlChecked(String url) {
```
JakeWharton: We can eliminate this override and the method in the interface and call directly into the public method now.

swankjesse:
I think I’m on board with the theme here, but the API doesn’t feel right. Methods that accept `Object` are awkward because you don’t know what they really accept until runtime.

I’d like to create another PR that addresses the same goal with fresh code.

yschimke:
@shaburov Do you want to check https://github.com/square/okhttp/pull/4118 and see it meets your needs?

PR #4046, opened 2018-06-06
update conscrypt to 1.1.3

yschimke:
includes a shutdown NPE fix, and been a while since updates.


yschimke:
I don't think we have a rush on updating here generally, assuming semantic versioning but worth not getting too far behind.

PR #4052, opened 2018-06-08
Fix stream cancel error stopping reused connection

oliveroneill:
Fix issue #3940

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/internal/http2/HttpOverHttp2Test.java
+++ b/okhttp-tests/src/test/java/okhttp3/internal/http2/HttpOverHttp2Test.java
@@ -773,6 +773,47 @@ public final class HttpOverHttp2Test {
     assertEquals(1, server.takeRequest().getSequenceNumber()); // Reused connection.
   }
 
+  @Test public void recoverFromCancelReusesConnection() throws Exception {
+    // Set up two responses
+    server.enqueue(new MockResponse()
+            .setBodyDelay(10, TimeUnit.SECONDS)
+            .setBody("abc"));
+    server.enqueue(new MockResponse()
+            .setBody("def"));
+
+    client = client.newBuilder()
+            .dns(new DoubleInetAddressDns())
+            .build();
+
+    Call call1 = client.newCall(new Request.Builder()
+            .url(server.url("/"))
+            .build());
+    // Use async handler so that we can cancel before receiving the body
+    // We don't really care about the response since we'll cancel immediately,
+    // so we don't do anything on callback
+    Callback callback = new Callback() {
+      @Override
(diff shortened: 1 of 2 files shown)


yschimke:
The change looks well thought out.

However do we see CANCEL sent from server to client in real world usage? It isn't clear this is actually likely or intended from the spec.

The two cases where CANCEL is documented is

1. Client sends cancels on an existing stream to tell the server to stop processing (save compute resources/network).
2. Client sends cancel in response to a push promise to indicate it is undesired.

I also wonder whether unlimited retries of CANCEL is safe, if a server is implemented badly, we may infinitely loop.  What is the intention of a server sending a CANCEL to the client, and is that compatible with retrying?  How does the server know that a client initiated stream is no longer needed?

>    CANCEL (0x8):  Used by the endpoint to indicate that the stream is no
>       longer needed.

yschimke:
The code might be right, but I suggest changing your tests then. They test and indicate the opposite case. You could have a long running (delayed) response using setBodyDelay, cancel those requests before they complete and confirm that subsequent requests are on the same connection.

oliveroneill:
My mistake. Thanks for the help, I'm quite new to the codebase. I've changed the tests to send cancels from the client and squashed the commits. Hope that's ok.

yschimke:
Looking good, I think you just need a call to setBodyDelay or something similar so the tests are predictable, e.g. cancel really cancels.

oliveroneill:
Good point. I've updated my commit now. Thanks.

okhttp-tests/src/test/java/okhttp3/internal/http2/HttpOverHttp2Test.java
```diff
@@ -773,6 +773,47 @@ private static OkHttpClient buildHttp2Client() {
...
+            .dns(new DoubleInetAddressDns())
+            .build();
+
+    Call call1 = client.newCall(new Request.Builder()
+            .url(server.url("/"))
+            .build());
+    // Use async handler so that we can cancel before receiving the body
+    // We don't really care about the response since we'll cancel immediately,
+    // so we don't do anything on callback
+    Callback callback = new Callback() {
```
yschimke: You should be able to use execute, and just not read the body.  It should make the test simpler.

oliveroneill: Using execute causes SocketTimeoutExceptions. Does execute wait to read the body before returning?

swankjesse:
This is awesome. Thank you.

I merged this manually.

swankjesse:
https://github.com/square/okhttp/commit/8b17ab4e65b21c6480a38c383a16899f9ffabbcc

Issue #3973, opened 2018-04-19
Unexpected TLS version: NONE


lizhangqu:
These days so many crashes happened in our app. About 200+ Android devices has happened. And it's always happend on Andorid 8.1.0. The OkHttp's version I used is 3.10.0.

This is the crash's stack information
```
0 java.lang.IllegalArgumentException: Unexpected TLS version: NONE
1 okhttp3.TlsVersion.forJavaName(TlsVersion.java:46)
2 okhttp3.Handshake.get(Handshake.java:55)
3 okhttp3.internal.connection.RealConnection.connectTls(RealConnection.java:242)
4 okhttp3.internal.connection.RealConnection.establishProtocol(RealConnection.java:198)
5 okhttp3.internal.connection.RealConnection.buildConnection(RealConnection.java:174)
6 okhttp3.internal.connection.RealConnection.connect(RealConnection.java:114)
7 okhttp3.internal.connection.StreamAllocation.findConnection(StreamAllocation.java:193)
8 okhttp3.internal.connection.StreamAllocation.findHealthyConnection(StreamAllocation.java:129)
9 okhttp3.internal.connection.StreamAllocation.newStream(StreamAllocation.java:98)
10 okhttp3.internal.connection.ConnectInterceptor.intercept(ConnectInterceptor.java:42)
11 okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:92)
12 okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:67)
13 okhttp3.internal.cache.CacheInterceptor.intercept(CacheInterceptor.java:109)
14 okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:92)
15 okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:67)
16 okhttp3.internal.http.BridgeInterceptor.intercept(BridgeInterceptor.java:93)
17 okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:92)
18 okhttp3.internal.http.RetryAndFollowUpInterceptor.intercept(RetryAndFollowUpInterceptor.java:124)
19 okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:92)
20 okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:67)
[...]

PR #4055, opened 2018-06-10
Move conscrypt NONE checks internal to Handshake

yschimke:
Based on https://github.com/square/okhttp/issues/3973, looks like this can still happen

### DIFF
--- a/okhttp/src/main/java/okhttp3/Handshake.java
+++ b/okhttp/src/main/java/okhttp3/Handshake.java
@@ -50,6 +50,7 @@ public final class Handshake {
   public static Handshake get(SSLSession session) throws IOException {
     String cipherSuiteString = session.getCipherSuite();
     if (cipherSuiteString == null) throw new IllegalStateException("cipherSuite == null");
+    if ("SSL_NULL_WITH_NULL_NULL".equals(cipherSuiteString)) throw new IOException("cipherSuite == SSL_NULL_WITH_NULL_NULL");
     CipherSuite cipherSuite = CipherSuite.forJavaName(cipherSuiteString);
 
     String tlsVersionString = session.getProtocol();
--- a/okhttp/src/main/java/okhttp3/Handshake.java
+++ b/okhttp/src/main/java/okhttp3/Handshake.java
@@ -50,7 +50,9 @@ public final class Handshake {
   public static Handshake get(SSLSession session) throws IOException {
     String cipherSuiteString = session.getCipherSuite();
     if (cipherSuiteString == null) throw new IllegalStateException("cipherSuite == null");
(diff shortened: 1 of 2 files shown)


yschimke:
@JakeWharton @swankjesse @dave-r12 If this is happening in 8.1.0, it would be good to get the right fix in for 3.11.0 as adoption increases.  I'm not sure if this is it, or there is something else we should be doing to address #3973

yschimke:
Here I guess - https://github.com/google/conscrypt/blob/82dc6605fb3d06bf51be2e2f6097f3eee1f32d22/common/src/main/java/org/conscrypt/SSLNullSession.java

I don't think this part is really documented at all.  And since it's now baked into 8.1.0, there is no fix that makes our workaround not needed.

swankjesse:
I’d prefer to remove RealConnection.isValid() completely and moving both the cipher suite check and the version check into Handshake.get(). That way we only check once!

okhttp/src/main/java/okhttp3/Handshake.java
```diff
@@ -46,13 +47,14 @@ private Handshake(TlsVersion tlsVersion, CipherSuite cipherSuite,
     this.localCertificates = localCertificates;
   }
 
-  public static Handshake get(SSLSession session) {
+  public static Handshake get(SSLSession session) throws IOException {
```
swankjesse: Adding a checked exception isn’t source-compatible. I’m okay with that but we’ll need to call it out in the changelog.

swankjesse:
This is a nice improvement.

liufsd:
Hi, i found this issue: when got this issue, the host is null. that mean the SSLSession is error ?..?

```
@Aspect
public class OkhttpHandshakeTLSAspect {
    @Pointcut("execution(* okhttp3.Handshake.get(..))")
    public void handshakePoint() {
    }

    @Around("handshakePoint()")
    public Object onHandshakePointExc(final ProceedingJoinPoint joinPoint) throws Throwable {
        try {
//            throw new IllegalArgumentException("Unexpected TLS version: NONE");
            return joinPoint.proceed();
        } catch (IllegalArgumentException throwable) {
            if (TextUtils.equals(throwable.getMessage(), "Unexpected TLS version: NONE")){
                submitBIEvent(joinPoint);
                throw new IOException("Unexpected TLS version: NONE");
            } else {
                throw throwable;
            }
        }
    }

    private static void submitBIEvent(ProceedingJoinPoint joinPoint) {
        try {
            SSLSession sslSession = (SSLSession) joinPoint.getArgs()[0];
            Log.e("HandshakeTLSAspect", "PeerHost: " + sslSession.getPeerHost());
            BIReportHammer.create().buildEvent(RunTimeManager.instance().getApplicationContext(), "OKHTTP_TLS_ERROR")
                    .addExtData("host", sslSession.getPeerHost())  // host: null
                    .addExtData("port", sslSession.getPeerPort()) //port: -1
                    .commit();
        } catch (Exception e) {
            e.printStackTrace();
        }
    }
}

```

PR #4059, opened 2018-06-13
Add fromMultimap Headers factory.

NightlyNexus:
This is the other converter side of toMultimap.
This is very similar to the existing of(Map<String, String>) factory and is named fromMultimap to avoid the same name after erasure.

The [Cronet Android client API uses multimaps for headers](https://developer.android.com/guide/topics/connectivity/cronet/reference/org/chromium/net/UrlResponseInfo.html#getAllHeaders()), and this would be great for converting to okhttp3.Headers for an okhttp3.Response.

### DIFF
--- a/okhttp/src/main/java/okhttp3/Headers.java
+++ b/okhttp/src/main/java/okhttp3/Headers.java
@@ -234,7 +234,6 @@ public final class Headers {
   public static Headers of(Map<String, String> headers) {
     if (headers == null) throw new NullPointerException("headers == null");
 
-    // Make a defensive copy and clean it up.
     String[] namesAndValues = new String[headers.size() * 2];
     int i = 0;
     for (Map.Entry<String, String> header : headers.entrySet()) {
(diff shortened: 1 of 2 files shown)


okhttp/src/main/java/okhttp3/Headers.java
```diff
@@ -234,7 +234,6 @@ public static Headers of(String... namesAndValues) {
   public static Headers of(Map<String, String> headers) {
     if (headers == null) throw new NullPointerException("headers == null");
 
-    // Make a defensive copy and clean it up.
```
NightlyNexus: looks like this comment was accidentally copied from the implementation of `of(String...)`.

okhttp/src/main/java/okhttp3/Headers.java
```diff
@@ -254,6 +253,46 @@ public static Headers of(Map<String, String> headers) {
     return new Headers(namesAndValues);
   }
 
+  /**
+   * Returns headers for the header names and values in the {@link Map}.
+   */
+  public static Headers fromMultimap(Map<String, List<String>> headers) {
+    if (headers == null) throw new NullPointerException("headers == null");
+
+    List<String> namesAndValues = new ArrayList<>(headers.size() * 2);
```
swankjesse: would it be simpler to use the builder to reconstruct? You wouldn’t need to duplicate validation code

swankjesse: sadly presizing the array list will be wrong if any header has multiple values

NightlyNexus: >sadly presizing the array list will be wrong if any header has multiple values.
true, might be better to add some buffer or multiply by 3.

yeah, using the builder is nicer also because I can use only the public api.

NightlyNexus:
I'm writing a call factory that's implemented with Cronet, so I can try Cronet everywhere that uses OkHttp.
agree it'd be bad to encourage the use of maps as headers, of course.

I'm happy to keep the map->headers conversion in my code, since it seems I'm the first to ask for this.

yschimke:
@NightlyNexus Awesome, can I beta test your Call.Factory impl?

PR #4060, opened 2018-06-13
Skip UTF-8 decode of keys, defer UTF-8 decode of data

JakeWharton:


JakeWharton:
Haven't done actual performance testing on this, but I'm theorizing it should be faster at the expense of more code.

okhttp-sse/src/main/java/okhttp3/internal/sse/ServerSentEventReader.java
```diff
@@ -36,96 +40,119 @@ public ServerSentEventReader(BufferedSource source, Callback callback) {
...
-      long crlf = source.indexOfElement(CRLF);
-      if (crlf == -1L) {
+      long lineEnd = source.indexOfElement(CRLF);
+      if (lineEnd == -1L) {
         return false;
       }
 
-      if (crlf == 0L) {
-        skipCrAndOrLf();
+      switch (source.buffer().getByte(0)) {
```
swankjesse: worth trying select() ?

okhttp-sse/src/main/java/okhttp3/internal/sse/ServerSentEventReader.java
```diff
@@ -36,96 +40,119 @@ public ServerSentEventReader(BufferedSource source, Callback callback) {
...
           }
-          data.append(fieldValue);
           break;
 
-        case "id":
-          if (fieldValue.isEmpty()) {
-            fieldValue = null;
+        case 'e':
+          if (isKey(EVENT)) {
+            type = parseType(lineEnd);
```
swankjesse: should this be parseEvent ?

okhttp-sse/src/main/java/okhttp3/internal/sse/ServerSentEventReader.java
```diff
@@ -136,4 +163,24 @@ private void skipCrAndOrLf() throws IOException {
...
+   * trailing space. Returns the number of bytes skipped.
+   */
+  private long skipNameAndDivider(long length) throws IOException {
+    source.skip(length);
+
+    if (source.buffer().getByte(0) == ':') {
+      source.skip(1L);
+      length++;
+
+      if (source.buffer().getByte(0) == ' ') {
```
swankjesse: do we know this won’t EOF ?

okhttp-sse/src/main/java/okhttp3/internal/sse/ServerSentEventReader.java
JakeWharton: We can use select but I don't think it has many advantages. For one, we need 5 variants for each key. For example, `data` is actually `data: `, `data:`, `data\n`, `data\r\n`, and `data\r`. The use of the first char is a cheat since everything we care about starts with a different char: <b>d</b>ata, <b>e</b>vent, <b>i</b>d, <b>r</b>etry, <b>`\r`</b>/`\r\n`, or <b>`\n`</b>.

I started with select before moving to the char. I can try it again though.

JakeWharton: I guess I could match on just `data` and then do the check for the field name divider though!

swankjesse: nah, cause select will consume if it matches, and you don’t want to pick up datacos

JakeWharton: Ah yeah that's right. I had to do a bunch of math on the crlf position (moreso than this PR does) to handle the case where the key was a prefix of an unknown key.

PR #4065, opened 2018-06-15
Embed R8/ProGuard rules in the jar

JakeWharton:
These can be read automatically by R8


JakeWharton:
Pared these down to what I use. The ones we had in the README were a bit ridiculous.

PR #4074, opened 2018-06-18
Add EventSource type and first HTTP test.

JakeWharton:


okhttp-sse/src/main/java/okhttp3/EventSources.java
```diff
@@ -0,0 +1,33 @@
...
+    return new EventSource.Factory() {
+      @Override public EventSource newEventSource(Request request, EventSourceListener listener) {
+        RealEventSource eventSource = new RealEventSource(request, listener);
+        eventSource.connect(client);
+        return eventSource;
+      }
+    };
+  }
+
+  private EventSources() {
```
vanniktech: throw an AssertionError?

JakeWharton: This class will eventually disappear as it will be folded onto OkHttpClient. But I added for now.

okhttp-sse/src/main/java/okhttp3/internal/sse/RealEventSource.java
```diff
@@ -0,0 +1,92 @@
...
+
+  private Call call;
+
+  public RealEventSource(Request request, EventSourceListener listener) {
+    this.request = request;
+    this.listener = listener;
+  }
+
+  public void connect(OkHttpClient client) {
+    client = client.newBuilder()
```
yschimke: Why clear the event listener?  More so than websockets, these requests appear to fit the event listener model.

I was wondering if it's worth going the other direction and having an optional (sub-)interface that EventListeners can implement to get notified about SSE events.

JakeWharton: Just copy/paste. Can remove. It's still a suuuuper long-lived connection which might throw off metrics if you're not careful to exclude them.

okhttp-sse/src/main/java/okhttp3/internal/sse/RealEventSource.java
```diff
@@ -0,0 +1,92 @@
...
+
+    response = response.newBuilder()
+        .body(Util.EMPTY_RESPONSE)
+        .build();
+
+    try {
+      listener.onOpen(this, response);
+      while (reader.processNextEvent()) {
+      }
+    } catch (Exception e) {
```
yschimke: Exception here seem counter to the normal OkHttp opinionated IOException or GTFO approach :) . Do you want to swallow NPE etc here?

okhttp-sse/src/main/java/okhttp3/internal/sse/RealEventSource.java
swankjesse: Dig.

PR #4082, opened 2018-06-20
Additional error checking for SSE

yschimke:
@JakeWharton Not sure where this is at, feel free to ignore this if you have already covered.  But kicked the tyres of this with Facebook live video comments feed and 401 errors were coming back as immediate close instead of any sort of error.

Also for Kotlin clients, all params were nullable.

### DIFF
--- a/okhttp-sse/src/main/java/okhttp3/internal/sse/RealEventSource.java
+++ b/okhttp-sse/src/main/java/okhttp3/internal/sse/RealEventSource.java
@@ -20,13 +20,13 @@ import javax.annotation.Nullable;
 import okhttp3.Call;
 import okhttp3.Callback;
 import okhttp3.EventListener;
-import okhttp3.EventSource;
-import okhttp3.EventSourceListener;
 import okhttp3.MediaType;
 import okhttp3.OkHttpClient;
 import okhttp3.Request;
 import okhttp3.Response;
 import okhttp3.internal.Util;
+import okhttp3.sse.EventSource;
+import okhttp3.sse.EventSourceListener;
 import okio.BufferedSource;
 
 public final class RealEventSource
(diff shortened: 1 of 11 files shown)


okhttp-sse/src/main/java/okhttp3/package-info.java
```diff
@@ -0,0 +1,3 @@
+/** Support for server-sent events. */
+@javax.annotation.ParametersAreNonnullByDefault
+package okhttp3;
```
JakeWharton: I think the presence of this will break building on Android because it creates duplicate classes in the same package. Perhaps we should keep this stuff in `okhttp.sse` while developing?

okhttp-sse/src/main/java/okhttp3/internal/sse/RealEventSource.java
```diff
@@ -70,6 +82,11 @@ public void connect(OkHttpClient client) {
     listener.onClosed(this);
   }
 
+  private boolean isEventStream(@Nullable MediaType contentType) {
```
JakeWharton: nit: static

okhttp-sse/src/main/java/okhttp3/internal/sse/RealEventSource.java
```diff
@@ -54,6 +55,17 @@ public void connect(OkHttpClient client) {
     BufferedSource source = response.body().source();
     ServerSentEventReader reader = new ServerSentEventReader(source, this);
 
+    if (!response.isSuccessful()) {
+      listener.onFailure(this, null, response);
```
JakeWharton: You don't have to fix it in this PR, but I notice we never ensure the response body is closed in normal SSE operation or these new failure callbacks. We'll need a few try/finallys, but that's an easy follow-up.

okhttp-sse/src/main/java/okhttp3/package-info.java
yschimke: Also important for java 9 modules

okhttp-sse/src/main/java/okhttp3/internal/sse/RealEventSource.java
```diff
@@ -51,38 +51,41 @@ public void connect(OkHttpClient client) {
...
 
-    MediaType contentType = response.body().contentType();
-    if (!isEventStream(contentType)) {
-      listener.onFailure(this, new IllegalStateException("Invalid content-type: " + contentType), response);
-      return;
-    }
+      MediaType contentType = response.body().contentType();
+      if (!isEventStream(contentType)) {
+        listener.onFailure(this,
+            new IllegalStateException("Invalid content-type: " + contentType), response);
```
swankjesse: I think I prefer ProtocolException when it's the peer’s fault.

PR #4084, opened 2018-06-21
Request and Response body logs size limitation

wakwak3125:
I want to add a limitation of output log size.
Recently I made a massive size of post request and my app killed by OOM.

I think it is nice to have a limitation of output log size.

Thanks.


okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
```diff
@@ -270,7 +308,12 @@ public Level getLevel() {
 
         if (contentLength != 0) {
           logger.log("");
-          logger.log(buffer.clone().readString(charset));
+          if (buffer.size() <= getResponseBodyLogMax() || getResponseBodyLogMax() < 0) {
+            logger.log(buffer.clone().readString(charset));
+          } else {
+            logger.log("Too large to output logs. "
```
yschimke: Should we show the first limit bytes instead?

okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
```diff
@@ -136,6 +141,34 @@ public Level getLevel() {
     return level;
   }
 
+  /** Change the limitation of request body logs size. */
+  public HttpLoggingInterceptor setRequestBodyLogMax(long requestBodyLogMax) {
```
yschimke: Is it simpler to have a single limit for logging? How much value is there in having two with separate values?

I guess on sending you can get it over and done with and not hold onto the memory, higher chance you might received a slow response from the server and hold things in memory.

What's your thinking on this?

okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
```diff
@@ -136,6 +141,34 @@ public Level getLevel() {
...
+    if (requestBodyLogMax < 0) {
```
yschimke: <= -2 seems like an error

okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
wakwak3125: I don't have a strong opinion to separate values. It's just for convenience.
Simpler is also better for me, so I change it to the single value 👍

okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
```diff
@@ -136,6 +140,20 @@ public Level getLevel() {
     return level;
   }
 
+  /** Change the limitation of response and request body logs size. */
+  public HttpLoggingInterceptor setBodyLogMax(long bodyLogMax) {
+    if (bodyLogMax < 0) {
```
yschimke: I think we should be stricter here, throw for <= -2, it probably represents a coding error.

okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
```diff
@@ -197,7 +215,11 @@ public Level getLevel() {
 
         logger.log("");
         if (isPlaintext(buffer)) {
-          logger.log(buffer.readString(charset));
+          if (buffer.size() <= getBodyLogMax() || getBodyLogMax() < 0) {
```
yschimke: It is probably worth a comment that we still read the full body here, just display the first limit bytes, since memory pressure it still involves reading into a large array.  With an (public) API change this could be fixed, or alternatively by failing after limit bytes, catching and continuing.

Should we fix this also if the goal is "Recently I made a massive size of post request and my app killed by OOM."

okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
```diff
@@ -197,7 +215,11 @@ public Level getLevel() {
...
+            logger.log(buffer.readString(charset));
+          } else {
+            logger.log(buffer.readString(getBodyLogMax(), charset));
```
yschimke: It is probably worth a comment that we still read the full body here, just display the first limit bytes, since memory pressure it still involves reading into a large array.  With an (public) API change this could be fixed, or alternatively by failing after limit bytes, catching and continuing.

Should we fix this also if the goal is "Recently I made a massive size of post request and my app killed by OOM."

okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
```diff
@@ -270,7 +292,11 @@ public Level getLevel() {
 
         if (contentLength != 0) {
           logger.log("");
-          logger.log(buffer.clone().readString(charset));
+          if (buffer.size() <= getBodyLogMax() || getBodyLogMax() < 0) {
```
yschimke: So I still think you are reading the full buffer on line 241/263, so not sure this change is doing what you want.  Have you tested with your large post?

okhttp-logging-interceptor/src/test/java/okhttp3/logging/HttpLoggingInterceptorTest.java
```diff
@@ -743,6 +748,61 @@ private void bodyGetNoBody(int code) throws IOException {
...
+    }
+  }
+
+  @Test public void setBodyLogMax() throws Exception {
+    setBodyLogMax(5000);
+    assertEquals(5000, applicationInterceptor.getBodyLogMax());
+    assertEquals(5000, networkInterceptor.getBodyLogMax());
+  }
+
+  @Test public void logBodyLimitation() throws Exception {
```
yschimke: Ideally this test would verify that only the right amount is read into the buffer instead of just based on the output.  But not sure how easy that is.

yschimke:
I hope you don't mind the deep review.  This class comes up as a reported issue, it's simple useful logging, but it has problems like you have found.  If we are going to change this, which I think it worthwhile, we should fix the problems fully.

Thanks for working on the PR.

wakwak3125:
> I hope you don't mind the deep review.

I appreciate your review!

This stack trace is the actual one.
So, the main issue is converting large `byte[]` to `String` , isn't it?

My last pushed code entirely worked on my environment.
But i don't mind if you close this PR😀
I use own patched one.

```
Caused by: java.lang.OutOfMemoryError: Failed to allocate a 9128344 byte allocation with 5895952 free bytes and 5MB until OOM, max allowed footprint 201326592, growth limit 201326592
                         W      at java.lang.StringFactory.newStringFromBytes(StringFactory.java:178)
                         W      at java.lang.StringFactory.newStringFromBytes(StringFactory.java:209)
                         W      at okio.Buffer.readString(Buffer.java:620)
                         W      at okio.Buffer.readString(Buffer.java:603)
                         W      at okhttp3.logging.HttpLoggingInterceptor.intercept(HttpLoggingInterceptor.java:199)
                         W      at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:147)
                         W      at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:121)
                         W      at com.wantedly.android.namecard_scanner.rest.MyRestAdapter$10.intercept(MyRestAdapter.java:319)
                         W      at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:147)
                         W      at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:121)
                         W      at okhttp3.RealCall.getResponseWithInterceptorChain(RealCall.java:200)
                         W      at okhttp3.RealCall.execute(RealCall.java:77)
                         W      at com.google.firebase.perf.network.FirebasePerfOkHttpClient.execute(Unknown Source:17)
                         W      at retrofit2.OkHttpCall.execute(OkHttpCall.java:180)
                         W      at retrofit2.adapter.rxjava2.CallExecuteObservable.subscribeActual(CallExecuteObservable.java:41)
                         W      at io.reactivex.Observable.subscribe(Observable.java:10955)
                         W      at retrofit2.adapter.rxjava2.BodyObservable.subscribeActual(BodyObservable.java:34)
                         W      at io.reactivex.Observable.subscribe(Observable.java:10955)
                         W      at io.reactivex.internal.operators.observable.ObservableSubscribeOn$SubscribeTask.run(ObservableSubscribeOn.java:96)
                         W      at io.reactivex.Scheduler$DisposeTask.run(Scheduler.java:452)
                         W      at io.reactivex.internal.schedulers.ScheduledRunnable.run(ScheduledRunnable.java:61)
```

yschimke:
@wakwak3125 I'd prefer to work with you to get a fix in!  I think there are two problems, 1 reading a potentially large streaming request or response into byte[], and then converting it to string.  I hope we can solve both.

wakwak3125:
@yschimke Thanks! I try to solve.
I intend to add the function that limits the buffer writing size to the RequestBody class.

Like:
```java
@Override public void writeTo(BufferedSink sink, long byteCount) throws IOException {
  byte[] contentBytes = new byte[((int) byteCount)];
  for (int i = 0; i < byteCount; i++) {
    contentBytes[i] = content.getByte(i);
  }
  sink.write(contentBytes);
 }
```

About a ResponseBody, can we limit the buffer size here?
https://github.com/wakwak3125/okhttp/blob/1d733288c583e257055520af8c5344538d50f0f9/okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java#L264

This solution can solve both buffer size problem and read mega size String from byte[] and do not affect the src body.
Do you have any idea or better way to do this?

yschimke:
cc @JakeWharton @swankjesse anything changing public API or (alternatively) touching okio, they probably have strong opinions on.

okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
```diff
@@ -136,6 +140,17 @@ public Level getLevel() {
     return level;
   }
 
+  /** Change the limitation of response and request body logs size. */
+  public HttpLoggingInterceptor setBodyLogMax(long bodyLogMax) {
```
JakeWharton: I don't think "Log" needs to be in the method name here. Maybe `setMaxBodyLength`?

okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
yschimke: This seems like a good point to reduce Long.MAX_VALUE on the line I mention above, avoid additional delay in this interceptor.

wakwak3125:
@yschimke @JakeWharton @swankjesse 

Thank you for the reviewing!

For the summary of this discussion,
- Not only cut the log string but also buffer the body until limitation bytes then output the logs
- Use ForwardingSink to limit the body size

Is there any problem with this summary?

yschimke:
@wakwak3125 Just a heads-up that we will be cutting 3.11 soon.  This PR would be great to complete and add, as I think it hits a number of people when trying to add logging to real world apps.

GC-Xi:
I recently just had the OutOfMemoryException When Logging Large issue. I had to turn off logging for QA/Production build. Would be nice to have the fix in the next release.

wakwak3125:
I'm sorry for all related people to review my PR 🙇 

I could not work for this PR for my private situation still now.
I can start continue this PR in a couple of weeks.

yschimke:
@swankjesse Does BufferedSource.peek change the possibilities here?

Tolriq:
Just adding a side note about this need as I'm using a modified version of the logger for that same purpose.

My app can communicate with a very large variety of hosts that implement an API with more or less success or will small issues here and here.

Being able to log full request even when large in production is vital to be able to properly handle the small variations, and it helped me build special adapters for moshi and stuff like that. Those can't be seen in lab and tests.

Logging in prod is bad, but sometimes it's the only solution :(

yschimke:
Closing, we discussed a somewhat similar case on https://github.com/square/okhttp/pull/5173 and decided to not change these classes and instead ask people to maintain their own loggers that match their requirements.

yschimke:
@kachi1227 I understand the frustration, but we don't consider this a actively developed API.

Take a look at https://github.com/bartek-wesolowski/customizable-okhttp-logging-interceptor, it might suit your needs or accept a PR.

PR #4089, opened 2018-06-23
Android 4 and Java 1.7 prefer TLSv1.2 provider

yschimke:
Public (unpaid version) of Java 1.7u80 supports TLSv1.2 only via a different provider. 

Newer paid version u95 support a system property (jdk.tls.client.protocols), and u131 defaults to supporting TLSv1.2

https://www.java.com/en/configure_crypto.html#enableTLSv1_2


yschimke:
@swankjesse ping

swankjesse:
We should require Java 8 soon. Lambdas!

Issue #4094, opened 2018-06-26
Event "callEnd" not called if an Interceptor is added


dechanaud:
Given the "Events" isn't finalized on version 3.10.0, I hesitated to report that issue as a bug.

Issue : Event "callEnd" isn't called if an Interceptor (Network or not) is added.

Steps to reproduce
- Instantiate an HttpClient with an EventListenerFactory (logging all events)
- Perform a dummy but successful GET request
- Check that all events are triggered, especially "callEnd"

- Instantiate an HttpClient with an EventListenerFactory (logging all events) & an interceptor such as HttpLoggingInterceptor
- Perform a dummy but successful GET request
- Check that all events except "callEnd" are triggered (not verified for callFailed)

yschimke:
It's not any interceptor, seems like something that HttpLoggingInterceptor is doing

yschimke:
I haven't dug in, just confirmed the bug with a test we can commit with a fix.

dechanaud:
No problem ;) I went a bit further on my side.

If I disable the response logging, "callEnd" event is well triggered. It looks to be linked with the request.body() called within the intercept. I hope it helps you

dechanaud:
I think I got it.

HttpLoggingInterceptor : line 241 : source.request(Long.MAX_VALUE) is the guilty point.
I don't get exactly the purpose of this operation...

yschimke:
@dechanaud I think it tries to read until the end of the stream, so the clone is a full copy.

dechanaud:
Indeed. Without this call, "chunked" response aren't logged as we didn't get the end of the stream

dechanaud:
I just verified and it works fine with 3.11.0. Thanks guys ;)

PR #4095, opened 2018-06-27
HTTP Logging interceptor breaks EventListener

yschimke:
As reported here https://github.com/square/okhttp/issues/4094

### DIFF
--- a/okhttp-tests/pom.xml
+++ b/okhttp-tests/pom.xml
@@ -55,6 +55,12 @@
       <version>${project.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.squareup.okhttp3</groupId>
+      <artifactId>logging-interceptor</artifactId>
+      <version>3.11.0-SNAPSHOT</version>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 
   <build>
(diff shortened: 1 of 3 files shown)


yschimke:
The ordering of event with HttpLoggingInterceptor

```
streamFinished: released false
deallocate: released false
deallocate: released true
release releasing true
```

Without

```
deallocate: released true
release releasing false
streamFinished: released true
deallocate: released false
```

okhttp-tests/pom.xml
```diff
@@ -55,6 +55,12 @@
       <version>${project.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.squareup.okhttp3</groupId>
+      <artifactId>logging-interceptor</artifactId>
+      <version>3.11.0-SNAPSHOT</version>
```
swankjesse: should be `${project.version}`

okhttp-tests/pom.xml
```diff
@@ -55,6 +55,12 @@
       <version>${project.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.squareup.okhttp3</groupId>
```
PromanSEW: should be `${project.groupId}`, right?

PR #4115, opened 2018-07-06
Add ResponseBody factory overload for ByteString

JakeWharton:
We shouldn't punish people with a roundtrip through byte[] for using a superior type!


swankjesse:
We already have a RequestBody one for symmetry. Phew!

JakeWharton:
Yep. That was the first thing I checked!

PR #4116, opened 2018-07-06
Add parse overload which fails on invalid inputs

JakeWharton:
Doing this before a `HttpUrl.getOrThrow` because it's similar. I didn't update any call sites, but can do so if this method is something we want.

Refs #3850

### DIFF
--- a/okhttp/src/main/java/okhttp3/MediaType.java
+++ b/okhttp/src/main/java/okhttp3/MediaType.java
@@ -45,12 +45,15 @@ public final class MediaType {
   }
 
   /**
-   * Returns a media type for {@code string}, or null if {@code string} is not a well-formed media
-   * type.
+   * Returns a media type for {@code string}.
+   *
+   * @throws IllegalArgumentException if {@code string} is not a well-formed media type.
    */
-  public static @Nullable MediaType parse(String string) {
+  public static MediaType get(String string) {
     Matcher typeSubtype = TYPE_SUBTYPE.matcher(string);
-    if (!typeSubtype.lookingAt()) return null;
+    if (!typeSubtype.lookingAt()) {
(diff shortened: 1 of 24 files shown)


JakeWharton:
For OkHttp 4.x I'd want to reverse these. `parse` and `parseOrNull` and `HttpUrl.get` and `HttpUrl.getOrNull`.

okhttp/src/main/java/okhttp3/MediaType.java
```diff
@@ -45,20 +45,29 @@ private MediaType(String mediaType, String type, String subtype, @Nullable Strin
...
 
   /**
-   * Returns a media type for {@code string}, or null if {@code string} is not a well-formed media
-   * type.
+   * Returns a media type for {@code string}.
+   *
+   * @throws IllegalArgumentException if {@code string} is not a well-formed media type.
    */
-  public static @Nullable MediaType parse(String string) {
+  public static MediaType parseOrThrow(String string) {
```
swankjesse: I would like this method to be named `get()`:

 * parse(String s): Attempt to get s as a media type, returning null if it isn't one.
 * get(String s): Returns the media type for s, throwing if it isn’t one.

And we can do likewise for HttpUrl.

JakeWharton: Oh, yeah. Nice.

For whatever reason I thought `HttpUrl` used `get` already.

swankjesse: It does for URL and URI. They could have a different scheme (say ftp:) so these methods return null. For OkHttp 4.x I think we change these methods to throw instead of returning null.

JakeWharton:
Updated to `get()`, and updated all call-sites which use a string literal to use it.

swankjesse:
Yeah this is especially nice for Kotlin!!

Issue #3945, opened 2018-03-28
Okhttp clears the Thread interrupt status


DrMetallius:
We have a problem in our project due to Okhttp consuming the interrupted flag of the thread. #3107 describes the problem perfectly. The reply suggests that the thrown `InterruptedIOException` should be enough handle the case. It isn't because it is rethrown in two cases: when a timeout is reached and when the thread was actually interrupted. We have no way to actually find out whether my thread got interrupted while it was doing work in Okhttp or it is just a network problem. The only solution is to parse the error message, which is obviously very wonky. This applies to both `waitUntilNotified` and `throwIfReached` in `Timeout`.

**There should be a way to tell thread interruption apart from network problems**. I suggest changing `InterruptedIOException` to `InterruptedException` when the thread is interrupted, as this will actually make the code conform to the **interruption handling conventions which are currently violated as `InterruptedIOException` is not an `InterruptedException`**, it is an `IOException`. In the case if this is considered a breaking change, I suggest creating a subclass of the `InterruptedIOException` and throwing that instead on thread interruption.

swankjesse:
We should probably just mark the thread as interrupted before throwing the `InterruptedIOException`.

PR #4117, opened 2018-07-06
Retain interrupted state when throwing InterruptedIOException

swankjesse:
The drawbacks seem small; the callsite needs to handle interruption anyway
because the thread is prone to interruption.

And the upside is that a single interrupt should now be sufficient to break
out an in-flight OkHttp call.

Note that although we're fixing this, thread interruption is not well tested
in OkHttp. Most users should prefer Call.cancel(), which is well tested and
doesn't rely on the caller to know which threads OkHttp is using to make
the actual HTTP request.

Closes: https://github.com/square/okhttp/issues/3945

### DIFF
--- a/mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.java
+++ b/mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.java
@@ -19,7 +19,6 @@ package okhttp3.mockwebserver;
 
 import java.io.Closeable;
 import java.io.IOException;
-import java.io.InterruptedIOException;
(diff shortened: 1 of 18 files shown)


DrMetallius:
This is great news! Regarding using thread interruption for cancelling calls, that wasn't our use case at all. We were having problems because we wanted to cancel _our_ operations, but the interruption could get swallowed in the bowels of OkHttp before our code got the chance to detect it.

The same problem is in the Android's SQLite APIs, by the way. We have to run the operations on a separate thread so that our thread retains the interrupted state.

PR #4118, opened 2018-07-06
Add get(String) alternative which fails on invalid inputs

### LINKED ISSUE #3850: Consider making HttpUrl.getChecked() public

JakeWharton:
This is an unchecked-throwing version of parse(String).

Closes #3850. Closes #4032.

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/HttpUrlTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/HttpUrlTest.java
@@ -20,104 +20,126 @@ import java.net.URI;
 import java.net.URL;
 import java.net.UnknownHostException;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.LinkedHashSet;
 import okhttp3.UrlComponentEncodingTester.Component;
 import okhttp3.UrlComponentEncodingTester.Encoding;
 import org.junit.Ignore;
 import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
 
 import static java.util.Collections.singletonList;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.fail;
 
+@RunWith(Parameterized.class)
 public final class HttpUrlTest {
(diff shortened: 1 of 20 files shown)


okhttp/src/main/java/okhttp3/OkHttpClient.java
```diff
@@ -181,9 +179,8 @@ public void apply(ConnectionSpec tlsConfiguration, SSLSocket sslSocket, boolean
         tlsConfiguration.apply(sslSocket, isFallback);
       }
 
-      @Override public HttpUrl getHttpUrlChecked(String url)
-          throws MalformedURLException, UnknownHostException {
-        return HttpUrl.getChecked(url);
+      @Override public boolean isInvalidHttpUrlHost(IllegalArgumentException e) {
+        return e.getMessage().startsWith(HttpUrl.Builder.INVALID_HOST);
```
JakeWharton: So yeah this is kinda gross, but... whatever. It's only for HttpURLConnection use.

okhttp-tests/src/test/java/okhttp3/HttpUrlTest.java
```diff
@@ -20,292 +20,314 @@
...
-    assertEquals(null, HttpUrl.parse("ht1tp://host/"));
-    assertEquals(null, HttpUrl.parse("httpss://host/"));
+    assertEquals(parse("http://host/"), parse("http://host/"));
+    assertEquals(parse("http://host/"), parse("Http://host/"));
+    assertEquals(parse("http://host/"), parse("http://host/"));
+    assertEquals(parse("http://host/"), parse("HTTP://host/"));
+    assertEquals(parse("https://host/"), parse("https://host/"));
+    assertEquals(parse("https://host/"), parse("HTTPS://host/"));
+
+    assertInvalid("image640://480.png", "Expected URL scheme 'http' or 'https' but was 'image640'");
```
swankjesse: Oh nice.

Issue #3931, opened 2018-03-20
SslHandshakeException: Certificate Expire related exception is consumed 


bogdan-richard:
Currently, the following exception is thrown, when a certificate is expired (and used for communication) on Android:
_```
javax.net.ssl.SSLProtocolException: SSL handshake aborted: ssl=0x8d1cbf40: Failure in SSL library, usually a protocol error
error:10000415:SSL routines:OPENSSL_internal:SSLV3_ALERT_CERTIFICATE_EXPIRED (external/boringssl/src/ssl/tls_record.c:504 0x86a5d000:0x00000001)
```_

The exception is thrown when code reaches
[sslSocket.startHandshake()](https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/internal/connection/RealConnection.java#L313))
and when
[connectionFailed(IOException e)](https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/internal/connection/ConnectionSpecSelector.java#L90)
is called and returns true, so the connection is retried. After the retry the exception is changed to
`Caused by: javax.net.ssl.SSLProtocolException: SSL handshake terminated: ssl=0x7a3f655d40: Failure in SSL library, usually a protocol error                                                                                                                                                error:10000410:SSL routines:OPENSSL_internal:SSLV3_ALERT_HANDSHAKE_FAILURE (external/boringssl/src/ssl/s3_pkt.c:610 0x7a2d5bac40:0x00000001)`

So the original exception detail message is consumed, and the code of a http client won't get info, that the original problem was that the certificate expired. 

I do not know, why the thrown exception's cause for sslScoket.startHandsahke is not a CertificateException (I wasn't able to find the implementation of SslSocket on Android, maybe they use sun.security.ssl.SslSocketImpl), but I think connectionFailed(IOException e) in ConnectionSpecSelector we should analyze the detailmessage and search for the correct [Alert Message](https://tools.ietf.org/html/rfc5246#appendix-A.3)).

I can add more details, logs and explanation, maybe a pull a request, please let me know your thoughts.

yschimke:
a PR would be great, if you can reproduce it locally, e.g. some of the tests generate certificates, so hopefully that is possible all within a test.  But even if it's just reproducible against openssl running locally, we could iterate from there.

Anything like this that makes it easier for users to understand the root cause of failures would be very impactful.

yschimke:
[...]

PR #4120, opened 2018-07-06
Reverse suppression of connect exceptions

yschimke:
Throw the first exception, with all following exceptions directly connected to the first instead of as a linked list.


yschimke:
https://github.com/square/okhttp/issues/3931

okhttp/src/main/java/okhttp3/internal/http/RetryAndFollowUpInterceptor.java
```diff
@@ -128,7 +128,7 @@ public StreamAllocation streamAllocation() {
       } catch (RouteException e) {
         // The attempt to connect via a route failed. The request will not have been sent.
         if (!recover(e.getLastConnectException(), streamAllocation, false, request)) {
-          throw e.getLastConnectException();
```
swankjesse: Is this now unused?

yschimke: No, used 1 line above in determining whether error is recoverable.

swankjesse: Well I suppose my search wasn't exhaustive!

Issue #3923, opened 2018-03-14
Add an easy way to merge or add all Headers. Headers.Builder.addAll(Headers)


NightlyNexus:
I have a Headers constant I'd like to add to every request. Given `Utils.commonHeaders()`, I would like to turn this:

```java
Headers existingHeaders = request.headers();
Headers commonHeaders = Utils.commonHeaders();
Headers.Builder headersBuilder = existingHeaders.newBuilder();

for (Map.Entry<String, List<String>> commonHeader : commonHeaders.toMultimap().entrySet()) {
  String name = commonHeader.getKey();
  List<String> values = commonHeader.getValue();
  for (int i = 0, size = values.size(); i < size; i++) {
    headersBuilder.add(name, values.get(i));
  }
}

request = request.newBuilder().headers(headersBuilder.build()).build();
```

into this:

```java
Headers existingHeaders = request.headers();
Headers commonHeaders = Utils.commonHeaders();
Headers headers = existingHeaders.newBuilder().addAll(commonHeaders).build();

request = request.newBuilder().headers(headers).build();
```

I don't have a use case for merging/overwriting any headers, so I'd just like a `Headers.Builder.addAll(Headers)`.

Does this sound like something the Headers API should add?

yschimke:
+1 I'd argue in favour of making Headers implements Iterable<Header> and taking Iterable<Header> here.

my 2c: If we are tweaking APIs like this, it might be a good time to test usage of these APIs when combined with Java 8 streams and Kotlin functional styles.  It would be a shame to add more methods and then a second pass as we notice that usage in Android O+ and Kotlin is awkward.

PR #4121, opened 2018-07-06
Add headerBuilder.addAll(headers) method

yschimke:
For quickly adding common headers

https://github.com/square/okhttp/issues/3923


yschimke:
+1, I'll addLenient.  Although it feels like the fully correct fix is knowing the state and not allowing you to mix without validation somewhere.

Issue #4114, opened 2018-07-05
OkHttp advertises TLSv1.3 Support without any compatible cipher suites


yschimke:
See discussion on https://github.com/square/okhttp/pull/4109

yschimke:
Not a blocker, as JDK detects this and handles gracefully.  

With a mix of TLS 1.3 + 1.2 cipher suites

```
"ClientHello": {
  "client version"      : "TLSv1.2",
  "cipher suites"       : "[TLS_AES_128_GCM_SHA256(0x1301), TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384(0xC02C), TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256(0xC02B), TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384(0xC030), TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256(0xC02F)]",
    "supported_versions (43)": {
      "versions": [TLSv1.3, TLSv1.2]
    },
```

With TLS 1.2 cipher suites only

```
"ClientHello": {
  "client version"      : "TLSv1.2",
  "cipher suites"       : "[TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384(0xC02C), TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256(0xC02B), TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384(0xC030), TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256(0xC02F)]",
    "supported_versions (43)": {
      "versions": [TLSv1.2]
    },
```

yschimke:
Documented in the TLS 1.3 Spec

> Updates Affecting TLS 1.2
> 
>    This document defines several changes that optionally affect
>    implementations of TLS 1.2, including those which do not also support
>    TLS 1.3:
> 
>    -  A version downgrade protection mechanism is described in
>       Section 4.1.3.
> 
>    -  RSASSA-PSS signature schemes are defined in Section 4.2.3.
> 
>    -  The "supported_versions" ClientHello extension can be used to
>       negotiate the version of TLS to use, in preference to the
>       legacy_version field of the ClientHello.
> 
>    -  The "signature_algorithms_cert" extension allows a client to
>       indicate which signature algorithms it can validate in X.509
>       certificates

PR #4122, opened 2018-07-06
Remove TLSv1.3 as default

yschimke:
We specify TLSv1.3, but without any compatible cipher suites.  So makes sense to remove for 3.11 to avoid weirdness as JVMs supporting TLSv1.3 start showing up.  n.b. JDK 11 detects this and strips anyway.


yschimke:
https://github.com/square/okhttp/issues/4114

PR #4123, opened 2018-07-06
RequestBody handles content from an InputStream

ellisjoe:


ellisjoe:
The main reason for this request is that `UnrepeatableRequestBody` is clearly the thing we want when reading from InputStreams, but that's not obvious as there's no usage of it anywhere.

swankjesse:
I’ve been reluctant to add something like this. One problem is that we don’t expose a mechanism to close the input stream if we don’t end up transmitting the request, such as if the TLS handshake fails.

yschimke:
@swankjesse what about with a Callable<InputStream> or something so OkHttp doesn't take ownership until it's going to use it?  I guess that still requires clients implement some boilerplate code.

ellisjoe:
Alternatively, I'd be happy if something like `StreamedRequestBody` were public, which doesn't have the same resource ownership issue.

swankjesse:
@ellisjoe yeah I think you’re right. You can’t build that API yourself today because `UnrepeatableRequestBody` isn’t public. Lemme open a tracking issue for this. There’s a few different ways we can expose that.

swankjesse:
https://github.com/square/okhttp/issues/4134

Issue #3934, opened 2018-03-22
MockWebServer should have public APIs for HeldCertificate and SslClient


swankjesse:
Possibly with examples. Would be especially nice to have an example of mutual auth.

swankjesse:
* [ ] HeldCertificate: promote to `okhttp3.mockwebserver`
 * [ ] SslClient: new APIs for my certs vs. peer’s certs. Should accept a cert to trust.
 * [ ] MockWebServer: setWantClientAuth(true) on incoming SSL sockets
 * [ ] RecordedRequest: expose the server’s handshake as a new field

swankjesse:
Now that these are ready I’m tempted to take one more step – to move HeldCertificate and TlsNode into a new “okhttp3.tls” module.

It would allow us to remove bouncycastle as a MockWebServer dependency. (Anyone using TlsNode would still need bouncycastle). It would give us the option to later support non-test use cases, such as parsing PEM files and generating CSRs. And we’d also have the option to replace our current bouncycastle dependency with something focused on just our use-cases.

swankjesse:
Okay, that’s all done.

New next steps:
 * [ ] Be clear in the docs about whether this is for test code, or fix the defaults to be appropriate for production code
 * [ ] API to add the platform’s default trust roots (possibly opt-in or opt-out)
 * [ ] Try to find a replacement name for TlsNode

PR #4124, opened 2018-07-06
Support client authentication in MockWebServer

swankjesse:
Also expose the handshake in the RecordedResponse.

https://github.com/square/okhttp/issues/3934

### DIFF
--- a/mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.java
+++ b/mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.java
@@ -102,6 +102,10 @@ public final class MockWebServer extends ExternalResource implements Closeable {
     Internal.initializeInstanceForTests();
   }
 
+  private static final int CLIENT_AUTH_NONE = 0;
+  private static final int CLIENT_AUTH_REQUESTED = 1;
+  private static final int CLIENT_AUTH_REQUIRED = 2;
+
   private static final X509TrustManager UNTRUSTED_TRUST_MANAGER = new X509TrustManager() {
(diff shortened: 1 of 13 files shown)


mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.java
```diff
@@ -102,6 +102,10 @@
     Internal.initializeInstanceForTests();
   }
 
+  private static final int CLIENT_AUTH_NONE = 0;
+  private static final int CLIENT_AUTH_REQUESTED = 1;
+  private static final int CLIENT_AUTH_REQUIRED = 2;
```
JakeWharton: Enum?

okhttp-tests/src/test/java/okhttp3/internal/tls/ClientAuthTest.java
```diff
@@ -41,16 +37,11 @@
...
 import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
 public final class ClientAuthTest {
   @Rule public final MockWebServer server = new MockWebServer();
 
-  public enum ClientAuth {
-    NONE, WANTS, NEEDS
-  }
```
JakeWharton: It was before!

swankjesse: Yeah, I’m still in my 65K method limit mindset, where a private API doesn’t need the extra stuff.

PR #4128, opened 2018-07-08
Avoid conscrypt dep

yschimke:


swankjesse:
```
[ERROR] /home/travis/build/square/okhttp/okhttp-dnsoverhttps/src/test/java/okhttp3/dnsoverhttps/TestDohMain.java:[34,48] package org.conscrypt does not exist
```

okhttp-tests/pom.xml
```diff
@@ -41,7 +41,7 @@
     <dependency>
       <groupId>org.conscrypt</groupId>
       <artifactId>conscrypt-openjdk-uber</artifactId>
-      <version>1.0.1</version>
+      <version>1.1.4</version>
```
swankjesse: Do we need this version since it's already in the parent pom.xml?

okhttp/pom.xml
```diff
@@ -20,7 +20,7 @@
     <dependency>
       <groupId>org.conscrypt</groupId>
       <artifactId>conscrypt-openjdk-uber</artifactId>
-      <version>1.1.3</version>
+      <version>1.1.4</version>
```
swankjesse: Do we need this version since it's already in the parent pom.xml?

PR #4129, opened 2018-07-08
[Discussion] Kotlin Support library

yschimke:
What would Kotlin support include?  Just to kick off discussion for now.


dave-r12:
This is exciting. Eager to see where this goes.

okhttp-kotlin/src/main/kotlin/okhttp3/kotlin/coroutines.kt
```diff
@@ -0,0 +1,27 @@
...
+import okhttp3.Response
+import java.io.IOException
+
+suspend fun OkHttpClient.execute(request: Request): Response {
+    val call = this.newCall(request)
+    return call.await()
+}
+
+suspend fun Call.await(): Response {
+    return kotlinx.coroutines.experimental.suspendCancellableCoroutine { c ->
```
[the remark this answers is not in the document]
JakeWharton: We should stick with two spaces

yschimke:
Would like to merge this for testing in 3.12-SNAPSHOT

okhttp-kotlin/src/main/kotlin/okhttp3/kotlin/kotlin.kt
```diff
@@ -0,0 +1,23 @@
+package okhttp3.kotlin
+
+import okhttp3.HttpUrl
+import okhttp3.OkHttpClient
+import okhttp3.Request
+
+fun request(url: String? = null, init: Request.Builder.() -> Unit = {}) = Request.Builder().apply {
```
BugsBunnyBR: Is a top level fun desirable for the lib?

yschimke: It is top level within okhttp3.kotlin package.  But what would you suggest instead?

yschimke:
@swankjesse Would like to park this until 1.3 release with non-experimental continuations support and land.  Results from okio 2.0 will also be useful.

yschimke:
@swankjesse post Okio kotlin work...  What are your thoughts for official Kotlin support assuming no imminent plan for converting the codebase over. 

Given the Kotlin adoption in android, it does feel like something that should be available without an additional and potentially multiple conflicting 3rd party libraries.  But worth getting right, not rushing in.

swankjesse:
I’m quite torn actually.

I don’t like how the KTX approach means you expose both Java and Kotlin idioms in the public API. And I don’t like the situation we’ll be in when we do deep Kotlin support if we already have a KTX API.

My preference is that we track Okio’s approach. Migrate our API to Kotlin while enforceing binary compatibility, and call the project “OkHttp 4”. Probably ship OkHttp 4 with `okhttp3` package names, which is ugly but workable. And we could consider the project shippable as soon as we’ve migrated our public APIs; we’re under no obligation to migrate our `internal` package for the users to benefit.

Let’s stall this project until we’ve cut a quality-of-life release (or two) first. Personally I wanna finish whole-operation timeouts before anything destabilizing. We should assume that once we start this it’ll be a long time before most of our users are on it.

yschimke:
I'll keep this alive for now, to keep it building, but please don't consider this "licking the cookie" this is just a strawman.

Egorand:
Agree that we should go with Okio 2 approach: we planned to add KTX for Okio and it never happened.

yschimke:
This is useful for discussion, but maintaining the PR isn't, so closing.  I'm really excited for this to happen.

PR #4136, opened 2018-07-11
Validate the certificate's basic constraints.

swankjesse:
Also fix our calling code to have the smallest path lengths possible.
http://www.ietf.org/rfc/rfc3280.txt


okhttp-tls/src/main/java/okhttp3/tls/HeldCertificate.java
```diff
@@ -302,8 +302,14 @@ public Builder issuedBy(HeldCertificate issuedBy) {
     /**
      * Set this certificate to be a signing certificate, with up to {@code maxIntermediateCas}
      * intermediate signing certificates beneath it.
+     *
+     * <p>If unset, this certificate cannot not sign other certificates. If set to 0, this
```
yschimke: Should you list the default behaviour?

PR #4137, opened 2018-07-11
Fixup HeldCertificate so that it's worthy of production use.

swankjesse:
This changes the default from the insecure 1024-bit RSA to a secure 256-bit ECDSA
key. It adds a new option to use RSA keys for interoperating with older clients.

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/CallTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/CallTest.java
@@ -1086,8 +1086,8 @@ public final class CallTest {
     client = client.newBuilder()
         .hostnameVerifier(new RecordingHostnameVerifier())
         .dns(new SingleInetAddressDns())
-        // opt-in to fallback to COMPATIBLE_TLS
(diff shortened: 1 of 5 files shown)


okhttp-tests/src/test/java/okhttp3/CallTest.java
```diff
@@ -1137,7 +1137,8 @@ private void postBodyRetransmittedAfterAuthorizationFail(String body) throws Exc
 
     client = client.newBuilder()
         .hostnameVerifier(new RecordingHostnameVerifier())
-        .connectionSpecs(Arrays.asList(ConnectionSpec.MODERN_TLS, ConnectionSpec.COMPATIBLE_TLS))
+        // Attempt RESTRICTED_TLS then fall back to MODERN_TLS.
+        .connectionSpecs(Arrays.asList(ConnectionSpec.RESTRICTED_TLS, ConnectionSpec.MODERN_TLS))
```
yschimke: Is this a good point for a varargs method on connectionSpecs?

swankjesse: I don’t mind this API being a little awkward to use. Most users shouldn’t use it!

PR #4141, opened 2018-07-11
Rename TlsNode to HandshakeCertificates

swankjesse:


yschimke:
The name alone would suggest to me the result of a handshake, but it beats TlsNode.  And growing on me.

Issue #4087, opened 2018-06-22
Deluxe Tags


bric3:
As suggested by @yschimke in this https://github.com/square/okhttp/issues/270#issuecomment-399324485, I'm opening a new issue, regarding the analytics API to be able to handle _call_ contextual data. 

The idea of this issue is to have a way to give and access context per call / request. This is interesting for monitoring typically when the request is parameterized (via path params, etc.). Or to monitor certain kind of payload, e.g. separate metrics for the various grant type of an oauth endpoint.

----

Here's a verbatim of the comments in #270

* https://github.com/square/okhttp/issues/270#issuecomment-381157742

  > @swankjesse I would like to monitor request by their route as defined in Retrofit. However this appear to be difficult to capture this information in Retrofit, see square/retrofit#2732. For bootstrapping the idea I talked about using Request.tag to pass along this contextual route data, but I don't think this is wise regarding the tag usage documentation.
  > 
  > Do you think OkHttp could introduce another concept in that regard ?

* https://github.com/square/okhttp/issues/270#issuecomment-381291303

  > how are you planning to do the monitoring? For example, is it a timer? One thing you could do is add a synthetic request header that your "start" timer removes. There are probably other ways, too.

* https://github.com/square/okhttp/issues/270#issuecomment-381597941

  > @adriancole Actually I implemented an eventlistener that feeds prometheus collectors (gauge, summary, counters). Summary are the closest thing to timers if compared to micrometer. (But for reasons outside of this scope I'm using summary instead of histograms).
  > 
  > I considered this approach as well in retrofit, but since I cannot modify the request at the right moment / stage in retrofit I cannot capture the path of the retrofit annotation.

adriancole:
interesting things here are that call is delay started.. so for example, lifecycle would be more foreground with a feature like this.. for example, can you add to the context before the call is executed, or can you only add before it is executed? Is the state immutable after a certain point? If a call is created and enqueued, but doesn't actually end in a network request, does that monkey wrench any assumptions?

Also, sometimes folks want property propagation (explicit context), and other times they want implicit context (like MDC hooks). Properties are much easier.

Just things for teasing out the design.

bric3:
By the way I hacked thing on friday night. While this work I'm not satisfied by the overall design of my hack as it involves to tweak things at several places.

Regarding immutability in my hacky design, the object (`RequestMetadata`) set in the `tag` field is _mutable until read_.

[...]

PR #4142, opened 2018-07-12
Deluxe tags.

swankjesse:
This expands our single Object.class into a Map<Class<T>, T>.

Closes: https://github.com/square/okhttp/issues/4087

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/RequestTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/RequestTest.java
@@ -21,12 +21,14 @@ import java.io.IOException;
 import java.net.URI;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.UUID;
 import okhttp3.internal.Util;
 import okio.Buffer;
 import org.junit.Test;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertSame;
 import static org.junit.Assert.fail;
 
 public final class RequestTest {
@@ -123,7 +125,7 @@ public final class RequestTest {
     assertEquals(HttpUrl.get("http://localhost/api"), request.url());
   }
 
-  @Test public void newBuilderUrlResetsUrl() throws Exception {
+  @Test public void newBuilderUrlResetsUrl() {
     Request requestWithoutCache = new Request.Builder().url("http://localhost/api").build();
     Request builtRequestWithoutCache =
         requestWithoutCache.newBuilder().url("http://localhost/api/foo").build();
@@ -137,7 +139,7 @@ public final class RequestTest {
     assertEquals(HttpUrl.get("http://localhost/api/foo"), builtRequestWithCache.url());
   }
 
-  @Test public void cacheControl() throws Exception {
+  @Test public void cacheControl() {
     Request request = new Request.Builder()
         .cacheControl(new CacheControl.Builder().noCache().build())
         .url("https://square.com")
@@ -145,7 +147,7 @@ public final class RequestTest {
     assertEquals(Arrays.asList("no-cache"), request.headers("Cache-Control"));
   }
 
-  @Test public void emptyCacheControlClearsAllCacheControlHeaders() throws Exception {
+  @Test public void emptyCacheControlClearsAllCacheControlHeaders() {
     Request request = new Request.Builder()
         .header("Cache-Control", "foo")
         .cacheControl(new CacheControl.Builder().build())
@@ -154,13 +156,13 @@ public final class RequestTest {
     assertEquals(Collections.<String>emptyList(), request.headers("Cache-Control"));
   }
 
-  @Test public void headerAcceptsPermittedCharacters() throws Exception {
+  @Test public void headerAcceptsPermittedCharacters() {
     Request.Builder builder = new Request.Builder();
     builder.header("AZab09~", "AZab09 ~");
     builder.addHeader("AZab09~", "AZab09 ~");
   }
 
-  @Test public void emptyNameForbidden() throws Exception {
+  @Test public void emptyNameForbidden() {
     Request.Builder builder = new Request.Builder();
     try {
       builder.header("", "Value");
@@ -174,7 +176,7 @@ public final class RequestTest {
     }
   }
 
-  @Test public void headerForbidsNullArguments() throws Exception {
+  @Test public void headerForbidsNullArguments() {
     Request.Builder builder = new Request.Builder();
     try {
       builder.header(null, "Value");
@@ -198,7 +200,7 @@ public final class RequestTest {
     }
   }
 
-  @Test public void headerAllowsTabOnlyInValues() throws Exception {
+  @Test public void headerAllowsTabOnlyInValues() {
     Request.Builder builder = new Request.Builder();
     builder.header("key", "sample\tvalue");
     try {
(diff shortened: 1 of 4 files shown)


okhttp/src/main/java/okhttp3/Request.java
```diff
@@ -239,12 +262,29 @@ public Builder method(String method, @Nullable RequestBody body) {
...
+     * request using {@link Request#tag}. Use null to remove any existing tag assigned for {@code
+     * type}.
+     *
+     * <p>Use this API to attach timing, debugging, or other application data to a request so that
+     * you may read it in interceptors, event listeners, or callbacks.
      */
-    public Builder tag(Object tag) {
-      this.tag = tag;
+    public <T> Builder tag(Class<? extends T> type, @Nullable T tag) {
+      if (type == null) throw new IllegalArgumentException("type == null: " + type);
```
JakeWharton: NPE. No point in passing the value to exception

okhttp/src/main/java/okhttp3/Request.java
```diff
@@ -66,8 +69,24 @@ public Headers headers() {
...
   public Object tag() {
-    return tag;
+    return tag(Object.class);
+  }
+
+  /**
+   * Returns the tag attached with {@code type} as a key, or null if no tag is attached with that
+   * key.
+   */
+  public <T> T tag(Class<? extends T> type) {
```
JakeWharton: Nullable

okhttp-tests/src/test/java/okhttp3/RequestTest.java
```diff
@@ -242,6 +243,103 @@ private void assertForbiddenHeader(String s) {
...
+  }
+
+  @Test public void defaultTag() {
+    UUID tag = UUID.randomUUID();
+    Request request = new Request.Builder()
+        .url("https://square.com")
+        .tag(tag)
+        .build();
+    assertEquals(tag, request.tag());
+    assertEquals(tag, request.tag(Object.class));
```
JakeWharton: If I'm being super nit picky these should all be assertSame

okhttp/src/main/java/okhttp3/Request.java
```diff
@@ -66,8 +69,24 @@ public Headers headers() {
...
 
+  /**
+   * Returns the tag attached with {@code Object.class} as a key, or null if no tag is attached with
+   * that key.
+   *
+   * <p>Prior to OkHttp 3.11, this method never returned null if no tag was attached. Instead it
+   * returned either this request, or the request upon which this request was derived with {@link
+   * #newBuilder()}.
+   */
   public Object tag() {
```
JakeWharton: Nullable

okhttp/src/main/java/okhttp3/Request.java
```diff
@@ -239,12 +262,29 @@ public Builder method(String method, @Nullable RequestBody body) {
...
+     * Attaches {@code tag} to the request using {@code type} as a key. Tags can be read from a
+     * request using {@link Request#tag}. Use null to remove any existing tag assigned for {@code
+     * type}.
+     *
+     * <p>Use this API to attach timing, debugging, or other application data to a request so that
+     * you may read it in interceptors, event listeners, or callbacks.
      */
-    public Builder tag(Object tag) {
-      this.tag = tag;
+    public <T> Builder tag(Class<? extends T> type, @Nullable T tag) {
```
JakeWharton: My only complaint about this is that I can't easily store generic types or multiple of the same type. You can create wrapper or subclasses for statically known tagging, but you can dynamically add multiple. I don't know if that's something we care about probably being a `Map<String, Object>` where typing is the user problem.

okhttp/src/main/java/okhttp3/Request.java
```diff
@@ -239,12 +262,29 @@ public Builder method(String method, @Nullable RequestBody body) {
...
+      if (type == null) throw new IllegalArgumentException("type == null: " + type);
+
+      if (tag == null) {
+        tags.remove(type);
```
JakeWharton: Is this safe on empty map?

PromanSEW: No, `UnsupportedOperationException` is thrown for `Collections.emptyMap()`.
This method returns unmodifiable empty map.
This operation is safe only for empty `LinkedHashMap`

okhttp/src/main/java/okhttp3/Request.java
yschimke: Main use cases I see

- oauth tokens
- tracing context flags
- logging tag set (zerorated, video, userinteractive, experimentlabels)

I think for any of these I don't expect that there will be hundreds of call sites in an app setting these on the request, so client dealing with String -> Object is ok.

okhttp/src/main/java/okhttp3/Request.java
swankjesse: Fixed. Too much Kotlin lately!

okhttp/src/main/java/okhttp3/Request.java
swankjesse: Fixed x 2.

okhttp-tests/src/test/java/okhttp3/RequestTest.java
swankjesse: Good call. Fixed all.

okhttp/src/main/java/okhttp3/Request.java
swankjesse: Fixed.

okhttp/src/main/java/okhttp3/Request.java
swankjesse: It’s safe. I added this test.
```
  @Test public void removeAbsentTag() {
        .url("https://square.com")
        .tag(null)
        .build();
    assertNull(request.tag());
  }
```

okhttp/src/main/java/okhttp3/Request.java
swankjesse: I’m imagining similar enough use cases, but I’m assuming boxing strings isn’t much of a hassle. I also think we could leverage existing types in tags:

```
Request request = ...
retrofit2.Call<?> retrofitCall = request.tag(retrofit2.Call.class);
zipkin.Span span = request.tag(zipkin.Span.class);
...
```

adriancole:
main comment is I like the feature name.

okhttp/src/main/java/okhttp3/Request.java
adriancole: I like that we can do this.. quite a lot.

okhttp/src/main/java/okhttp3/Request.java
```diff
@@ -239,12 +262,29 @@ public Builder method(String method, @Nullable RequestBody body) {
...
-    public Builder tag(Object tag) {
-      this.tag = tag;
+    public <T> Builder tag(Class<? extends T> type, @Nullable T tag) {
+      if (type == null) throw new NullPointerException("type == null");
+
+      if (tag == null) {
+        tags.remove(type);
+      } else {
+        if (tags.isEmpty()) tags = new LinkedHashMap<>();
+        tags.put(type, tag);
```
stevecstian: Should we have a dynamic cast here to achieve runtime type safety?

swankjesse: Good idea. Done.

okhttp/src/main/java/okhttp3/Request.java
stevecstian: A tiny wrapper should help for generic type:
```java
interface MyListSupplier extends Supplier<List<String>> {}
...
List<String> myList = request.tag(ListSupplier.class).get();
...
```

JakeWharton: Assuming you have the flexibility to subtype and not wrap. And assuming your entries are statically known and not dynamic.

stevecstian: Yup

PR #4144, opened 2018-07-12
New API, HandshakeCertificates.addPlatformTrustedCertificates()

swankjesse:
Also rename HeldCertificate.Builder.issuedBy() to signedBy(). 'Issued' is the
word used by certificates; 'signed' is the word used by cryptographers. I prefer
'signed'.

### DIFF
--- a/okhttp-tls/src/main/java/okhttp3/tls/HandshakeCertificates.java
+++ b/okhttp-tls/src/main/java/okhttp3/tls/HandshakeCertificates.java
@@ -20,6 +20,7 @@ import java.security.KeyManagementException;
 import java.security.SecureRandom;
 import java.security.cert.X509Certificate;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import javax.net.ssl.KeyManager;
 import javax.net.ssl.SSLContext;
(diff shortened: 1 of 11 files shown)


okhttp-tls/src/main/java/okhttp3/tls/HandshakeCertificates.java
```diff
@@ -137,6 +140,24 @@ public Builder addTrustedCertificate(X509Certificate certificate) {
       return this;
     }
 
+    /**
+     * Add all of the host platform's trusted root certificates. This set varies by platform
+     * (Android vs. Java), by platform release (Android 4.4 vs. Android 9), and with user
+     * customizations.
+     *
+     * <p>Most TLS clients should that connect to hosts on the public Internet should call this
```
dave-r12: _Most TLS clients that connect_ ?

swankjesse: Fixed. I have much difficulty copy-editing my own work, especially when I write and rewrite this stuff many times before I’m happy with it.

okhttp/src/main/java/okhttp3/internal/Util.java
```diff
@@ -642,4 +647,20 @@ private static String inet6AddressToAscii(byte[] address) {
...
     return result.readUtf8();
   }
+
+  public static X509TrustManager platformTrustManager() {
+    try {
+      TrustManagerFactory trustManagerFactory = TrustManagerFactory.getInstance(
+          TrustManagerFactory.getDefaultAlgorithm());
+      trustManagerFactory.init((KeyStore) null);
+      TrustManager[] trustManagers = trustManagerFactory.getTrustManagers();
+      if (trustManagers.length != 1 || !(trustManagers[0] instanceof X509TrustManager)) {
```
yschimke: Curious about this common approach vs first compatible of Netty

https://github.com/netty/netty/blame/a137291ad161c970259d9a5d7cc206c026f08ee7/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslContext.java#L504-L510

Issue #4135, opened 2018-07-10
Infinite retry on FileNotFoundException in okhttp 3.6.0


deepanshu-hike:
I am getting infinte retries in okhttp on FileNotFoundException.

Trace:
1. In CallServerInterceptor's intercept(Chain chain), `request.body().writeTo(bufferedRequestBody);` is throwing FileNotFoundException.
2. Proceeding with interceptors in the chain, the above FileNotFoundException gets caught in RetryAndFollowUpInterceptor.
3. In RetryAndFollowUpInterceptor's intercept(Chain chain), `recover(e, requestSendStarted, request)` gives true. So the condition `if (!recover(e, requestSendStarted, request))` fails and the while loop continues forever.

This is causing infinite number of retries in case of FileNotFoundException.

superthib:
Still reproducing the issue with 3.12.
I'm using retrofit 2.5.0 to make a POST request, using the RequestBody.create(MediaType contentType, File file) method with a file that doesn't exist. The FileNotFoundException gets caught in RetryAndFollowUpInterceptor, which considers it as a kind of connection recoverable error, causing an infinite retry that floods the server (about 100 requests per second).

yschimke:
@superthib It's hard to be 100% certain, but I think 99.95% of the time we can rely on this exception being because of a bad request.

https://github.com/square/okhttp/pull/4409

PR #4151, opened 2018-07-14
Test for file not found in post body

yschimke:
https://github.com/square/okhttp/issues/4135

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/CallTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/CallTest.java
@@ -16,6 +16,7 @@
 package okhttp3;
 
 import java.io.File;
+import java.io.FileNotFoundException;
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


okhttp-tests/src/test/java/okhttp3/CallTest.java
```diff
@@ -3262,6 +3263,26 @@ private RequestBody requestBody(final boolean chunked, final long size, final in
...
   }
 
+  @Test public void postWithFileNotFound() throws Exception {
+    RequestBody body = new RequestBody() {
+      @Nullable @Override public MediaType contentType() {
+        return MediaType.get("application/octet-stream");
+      }
+
+      @Override public void writeTo(BufferedSink sink) throws IOException {
+        throw new FileNotFoundException();
```
yschimke: I might want to tighten this up that it is only checked once, not retried.

Issue #4152, opened 2018-07-14
RejectedExecutionException in client.dispatcher().enqueue


mstrelex:
Hi,

Getting the following exception after computer returns from sleep mode:

`Exception in thread "EventThread" java.util.concurrent.RejectedExecutionException: Task okhttp3.RealCall$AsyncCall@17c8d3 rejected from java.util.concurrent.ThreadPoolExecutor@7822a9d2[Shutting down, pool size = 1, active threads = 1, queued tasks = 0, completed tasks = 25]
	at java.util.concurrent.ThreadPoolExecutor$AbortPolicy.rejectedExecution(ThreadPoolExecutor.java:2063)
	at java.util.concurrent.ThreadPoolExecutor.reject(ThreadPoolExecutor.java:830)
	at java.util.concurrent.ThreadPoolExecutor.execute(ThreadPoolExecutor.java:1379)
	at okhttp3.Dispatcher.enqueue(Dispatcher.java:132)
	at okhttp3.RealCall.enqueue(RealCall.java:100)
	at okhttp3.internal.ws.RealWebSocket.connect(RealWebSocket.java:192)
	at okhttp3.OkHttpClient.newWebSocket(OkHttpClient.java:438)`

Any ideas?
Much obliged.

mstrelex:
Anyone ?

swankjesse:
Duplicate of https://github.com/square/okhttp/issues/3424

PR #4153, opened 2018-07-14
Handle RejectedExecutionException when enqueue fails

mstrelex:


mstrelex:
Fixed #4152

okhttp/src/main/java/okhttp3/RealCall.java
```diff
@@ -97,7 +97,11 @@ private void captureCallStackTrace() {
     }
     captureCallStackTrace();
     eventListener.callStart(this);
-    client.dispatcher().enqueue(new AsyncCall(responseCallback));
+    try {
+      client.dispatcher().enqueue(new AsyncCall(responseCallback));
+    } catch (Exception e) {
```
yschimke: Can you catch specific exceptions.

okhttp/src/main/java/okhttp3/RealCall.java
```diff
@@ -97,7 +97,11 @@ private void captureCallStackTrace() {
...
+      responseCallback.onFailure(RealCall.this, new IOException(e.getMessage()));
```
yschimke: Can you put a message, and chain the original exception e.g.

`new IOException("unable to enqueue to dispatcher", e);`

yschimke:
This seems like a good change, especially if people customise the dispatcher thread pool.

Please add a test.  You will also need to sign the CLA https://github.com/square/okhttp/wiki/Contributing

yschimke:
Also, can you add a test?

mstrelex:
Done, I will write a test later on if you don't mind.

mstrelex:
@yschimke Am I doing something wrong that causes checks to fail ? Thanks.

mstrelex:
It comes from https://github.com/socketio/socket.io-client-java/issues/539

yschimke:
@dave-r12 I disagree, we default to an infinitely sized executor so don't see this ourselves.  But okhttp public APIs allow specifying a tuned dispatcher with a specific (constrained) ExecutorService. And it's a public part of that API that job submissions can fail with this specific RuntimeException.  So we should handle this gracefully, and for OkHttp thats returning an IOException.  OkHttp throwing runtime exceptions tends to be fatal for android apps.

JakeWharton:
But it's not an IOException. That would be lying to the user.

yschimke:
How do we handle other cases when a request can't be satisfied on the client side before connecting remotely? We throw a fake 503 in other cases, when cache rules make the request impossible to fulfill.

But really the awkward thing for a caller is that given a valid but non default OkHttp client, a valid enqueue call can throw a RuntimeException which otherwise generally that indicates a bug or misuse, and sometimes leaves OkHttp in a undefined state.

mstrelex:
Have you come to any kind of a decision on this? Thanks.

swankjesse:
I think it’s a programming error to shutdown an executor service before you’re done using it. The RejectedExecutionException might already be the right choice.

yschimke:
It's not necessarily a shutdown executor, it could be a bounded one.  And reasons why I think we should handle it better than we currently do.

1) We suggest callers reuse a singleton Client.  So a caller to enqueue may not know how the client is configured.
2) Dispatcher with custom ExecutorService is part of the public API.  This usecase seems like a valid reason why you would specify a custom executor, to avoid unbounded memory growth.
3) We fail in surprising ways that may not be exposed to callers.  Leaving the call, event listeners and the client instance itself in an inconsistent state.

```
package com.baulsupp.oksocial

import okhttp3.Call
import okhttp3.Callback
import okhttp3.Dispatcher
import okhttp3.OkHttpClient
import okhttp3.Protocol
import okhttp3.Request
import okhttp3.Response
import java.io.IOException
import java.util.concurrent.ArrayBlockingQueue
import java.util.concurrent.BlockingQueue
import java.util.concurrent.RejectedExecutionException
import java.util.concurrent.RejectedExecutionHandler
import java.util.concurrent.ThreadPoolExecutor

fun main(args: Array<String>) {
  val queue: BlockingQueue<Runnable> = ArrayBlockingQueue(3)
  val rejectionHandler: RejectedExecutionHandler = ThreadPoolExecutor.AbortPolicy()
  val tp = ThreadPoolExecutor(1, 3, 5, TimeUnit.SECONDS, queue, rejectionHandler)

  val dispatcher = Dispatcher(tp)

  val client = OkHttpClient.Builder().dispatcher(dispatcher).protocols(listOf(Protocol.HTTP_1_1)).build()

  val delayRequest = Request.Builder().url("https://httpbin.org/delay/1").build()

  val callback = object : Callback {
    override fun onFailure(call: Call, e: IOException) {
      println("onFailure: ${e.message}")
    }

      println(response.code())
    }
  }

  try {
    client.newCall(delayRequest).enqueue(callback)
    client.newCall(delayRequest).enqueue(callback)
    client.newCall(delayRequest).enqueue(callback)
    client.newCall(delayRequest).enqueue(callback)
    client.newCall(delayRequest).enqueue(callback)
    client.newCall(delayRequest).enqueue(callback)
    client.newCall(delayRequest).enqueue(callback)
  } catch (ree: RejectedExecutionException) {
    println("Rejected: ${ree.message}")
  }
}
```  

n.b. Neither onFailure or Rejected is ever printed.

```
Exception in thread "pool-1-thread-1" java.util.concurrent.RejectedExecutionException: Task okhttp3.RealCall$AsyncCall@51c9d05e rejected from java.util.concurrent.ThreadPoolExecutor@23048844[Running, pool size = 3, active threads = 3, queued tasks = 2, completed tasks = 1]
	at java.util.concurrent.ThreadPoolExecutor$AbortPolicy.rejectedExecution(ThreadPoolExecutor.java:2063)
	at java.util.concurrent.ThreadPoolExecutor.reject(ThreadPoolExecutor.java:830)
	at java.util.concurrent.ThreadPoolExecutor.execute(ThreadPoolExecutor.java:1379)
	at okhttp3.Dispatcher.promoteCalls(Dispatcher.java:166)
	at okhttp3.Dispatcher.finished(Dispatcher.java:203)
	at okhttp3.Dispatcher.finished(Dispatcher.java:190)
	at okhttp3.RealCall$AsyncCall.execute(RealCall.java:164)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
200
200
200
200
200
200
```

yschimke:
I'm going to prod this further along by writing an illustrative test case I think we should aim to make pass.

dave-r12:
What about the JavaDocs on `Dispatcher`?

> Each dispatcher uses an {@link ExecutorService} to run calls internally. If you supply your own executor, it should be able to run {@linkplain #getMaxRequests the configured maximum} number of calls concurrently.

yschimke:
@dave-r12 fair point, it's still awkward to fail internally without ability to detect.

garun2536:
D

mstrelex:
Do you any concrete suggestions how to handle this?
You are saying the socket.io that uses okhttp should be responsible?

mstrelex:
Bump. Will appreciate a response.

dave-r12:
@mstrelex the way I'm reading this: socket io library is shutting down the executor then attempting to use it after shutdown. They shouldn't do that. Can we confirm that's the case?

mstrelex:
I do not see socket.io specifying a custom dispatcher or shutting down the executor.

swankjesse:
Obsoleted by https://github.com/square/okhttp/pull/4365

swankjesse:
Thanks @mstrelex ! I started by adding tests to your PR, and it ended up being a completely nastier change. Though you got us going in the right direction.

PR #4167, opened 2018-07-19
Platform internal nullable annotations

yschimke:


yschimke:
low-priority, but platform is subclassed, and shows up in usage in Intellij.

Issue #4200, opened 2018-08-09
[Feature Request] get RealConnection rawsocket localport


lxxgreat:
we use okhttp to implementing websocket, and want to get RealConnection rawsocket localport, but the connection pool's member variable is private. so can it be public or hava a public method ?

yschimke:
As mentioned on the PR, we should not expose private types.  But it would be useful to have some visibility into the ConnectionPool.  So I personally hope there is something we can do here.

This sort of change probably takes a number of review iterations and will need to prove it's value before getting accepted as it commits to a new public API.  i.e. it could be a frustrating experience.

swankjesse:
If you're looking for a workaround, you can provide a custom SocketFactory that remembers the sockets returned.

yschimke:
@swankjesse I want a similar feature to snapshot, and listen for connection pool additions and removals (facebook flipper and stetho type monitoring).  Have we considered public API here?

swankjesse:
I'm not sure. Our EventListener API is intended to help developers understand how their clients & servers interact. This would be strictly about OkHttp itself, which is a much narrower audience.

yschimke:
Won't fix.

PR #4201, opened 2018-08-09
Returns connections to get sub RealConnection info 

lxxgreat:
#4200


JakeWharton:
This exposes a non-public type

lxxgreat:
but we expose the connectionCount as followings: 
 
public synchronized int connectionCount() {
    return connections.size();
  }

yschimke:
@lxxgreat it's okhttp3.internal.connection.RealConnection that is the concern

Issue #4204, opened 2018-08-10
Client throws exception when instantiated under a restrictive SecurityManager


mgladdish:
What kind of issue is this?

 - [ ] Question. This issue tracker is not the place for questions. If you want to ask how to do
       something, or to understand why something isn't working the way you expect it to, use Stack
       Overflow. https://stackoverflow.com/questions/tagged/okhttp

 - [X] Bug report. If you’ve found a bug, spend the time to write a failing test. Bugs with tests
       get fixed. Here’s an example: https://gist.github.com/swankjesse/981fcae102f513eb13ed

 - [ ] Feature Request. Start by telling us what problem you’re trying to solve. Often a solution
       already exists! Don’t send pull requests to implement new features without first getting our
       support. Sometimes we leave features out on purpose to keep the project small.

If Java's SecurityManager is enabled, and has a restrictive policy in place, then the client will throw an exception when instantiated.

This is due to this line:
https://github.com/square/okhttp/blob/ec0cdf2970aa6104a481e35f4e2734448ade28bc/okhttp/src/main/java/okhttp3/internal/platform/Platform.java#L190

which throws a `java.security.AccessControlException` if the java policy applied does not have 

```permission java.util.PropertyPermission "okhttp.platform", "read";```

Given the client works fine without the property being set, I suggest it should catch `SecurityException` at this point and log a warning, then carry on without the property value.

swankjesse:
Want to send a pull request? I wonder if there's an easy way to test this.

By the way, why are you using a SecurityManager? I think it's a very bad API that adds a bunch of complexity without adding any security. Only reasonable use cases I'm aware of are in testing.

mgladdish:
I may not have the time soon - if it's still outstanding then, I may get the chance.

Testing it shouldn't be too bad - the main test suite should pass with a known policy in place, so a simple policy can be passed to the test runner's JVM. I presume there's another test that _does_ require this property to be set, so that will have to be run slightly differently I guess.

As for why a SecurityManager, we run a PaaS where we run other customers' code on our platform. Locking down the permissions of what that code can and cannot do is critical. It's not enough on its own, of course, but nothing is.

swankjesse:
@mgladdish got it. You should treat the Java security manager as advisory-only. There's a long history of successful attacks on the sandbox and no systematic fix. If your threat model assumes otherwise you may have a problem.

mgladdish:
Quite. Hence the "it's not enough on its own".

[...]

PR #4217, opened 2018-08-17
Support restrictive security manager

yschimke:
Addressing https://github.com/square/okhttp/issues/4204

Be tolerant of environments with security manager that restrict access to less critical resources like


yschimke:
I tried adding a test and it failed in Intellij because I couldn't install a SecurityManager in the test.  Let's close the issue as OkHttp isn't supported with a security manager, but it does seem like it should be able to work without this specific permission.  I suspect a reasonable policy might allow connections to some hosts but not all.  But I'm ok with closing this.

Issue #4225, opened 2018-08-22
MockWebServer recordedRequest.getRequestUrl() is null on ipv6 host


thevery:
1) On my linux build machine localhost is resolved to ipv6 address (`0:0:0:0:0:0:0:1`)
2) `okhttp3.mockwebserver.RecordedRequest#RecordedRequest` does not handle ipv6 properly because it  uses `HttpUrl.parse(String.format("%s://%s:%s%s", scheme, hostname, port, path));` - brackets are missing for ipv6. Easy fix is `HttpUrl.Builder` usage.
3) Unfortunately I cannot create a reliable test class, but I use `server.start(Inet4Address.getLocalHost(), 0);` as a workaround.

swankjesse:
That was pretty sloppy! Can fix.

PR #4235, opened 2018-08-26
MockWebserver fix for IPv6 addresses

yschimke:
surround IPv6 addresses with [] to make a valid URI

https://github.com/square/okhttp/issues/4225

### DIFF
--- a/mockwebserver/src/main/java/okhttp3/mockwebserver/RecordedRequest.java
+++ b/mockwebserver/src/main/java/okhttp3/mockwebserver/RecordedRequest.java
@@ -78,7 +78,7 @@ public final class RecordedRequest {
       }
 
       int port = socket.getLocalPort();
-      this.requestUrl = HttpUrl.get(String.format("%s://%s:%s%s", scheme, hostname, port, path));
+      this.requestUrl = HttpUrl.parse(String.format("%s://%s:%s%s", scheme, hostname, port, path));
     } else {
       this.requestUrl = null;
       this.method = null;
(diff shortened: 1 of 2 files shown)


mockwebserver/src/main/java/okhttp3/mockwebserver/RecordedRequest.java
```diff
@@ -61,12 +63,22 @@ public RecordedRequest(String requestLine, Headers headers, List<Integer> chunkS
       int methodEnd = requestLine.indexOf(' ');
       int pathEnd = requestLine.indexOf(' ', methodEnd + 1);
       this.method = requestLine.substring(0, methodEnd);
-      this.path = requestLine.substring(methodEnd + 1, pathEnd);
+      String path = requestLine.substring(methodEnd + 1, pathEnd);
+      if (!path.startsWith("/")) {
+        path = "/";
```
swankjesse: Curious: what's this for? Empty path?

yschimke: CONNECT method.  Maybe I should null out the request URI in this case?  Previously it was because of the HttpUrl.parse, but that didn't appear deliberate.

Thinking about it more, maybe MockWebserver shouldn't fail because it has a bad request path because it seems valid to reproduce a incorrectly implemented server.

yschimke:
@swankjesse pop quiz, why does it combine localPort and inet address.  It seems intuitively like inet address+port makes more sense, but obviously it works.

mockwebserver/src/main/java/okhttp3/mockwebserver/RecordedRequest.java
yschimke: with a "/" prefix?

swankjesse:
On your pop quiz, I think it's actually completely broken! And we only get away with it cause it's always localhost so local and remote are the same.

Should be local for both!

yschimke:
I'll submit follow on PRs for handling CONNECT and changing the port

swankjesse:
👍🏻

Issue #4208, opened 2018-08-13
IndexOutOfBoundsException in RealConnection


jaydenyu93:
got following crash in google pixel xl device with android p using okhttp3.11
Fatal Exception: java.lang.IndexOutOfBoundsException: Index: 0
       at java.util.Collections$EmptyList.get(Collections.java:4502)
       at okhttp3.internal.connection.RealConnection.connectTls(RealConnection.java:325)
       at okhttp3.internal.connection.RealConnection.establishProtocol(RealConnection.java:282)
       at okhttp3.internal.connection.RealConnection.connect(RealConnection.java:167)
       at okhttp3.internal.connection.StreamAllocation.findConnection(StreamAllocation.java:257)
       at okhttp3.internal.connection.StreamAllocation.findHealthyConnection(StreamAllocation.java:135)
       at okhttp3.internal.connection.StreamAllocation.newStream(StreamAllocation.java:114)
       at okhttp3.internal.connection.ConnectInterceptor.intercept(ConnectInterceptor.java:42)
       at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:147)
       at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:121)
       at okhttp3.internal.cache.CacheInterceptor.intercept(CacheInterceptor.java:93)
       at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:147)
       at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:121)
       at okhttp3.internal.http.BridgeInterceptor.intercept(BridgeInterceptor.java:93)
       at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:147)
       at okhttp3.internal.http.RetryAndFollowUpInterceptor.intercept(RetryAndFollowUpInterceptor.java:126)
       at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:147)
       at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:121)
       at okhttp3.RealCall.getResponseWithInterceptorChain(RealCall.java:200)
       at okhttp3.RealCall$AsyncCall.execute(RealCall.java:147)
       at okhttp3.internal.NamedRunnable.run(NamedRunnable.java:32)
[...]

PR #4236, opened 2018-08-26
Handle no peer certificates cleanly

yschimke:


yschimke:
https://github.com/square/okhttp/issues/4208

okhttp/src/main/java/okhttp3/internal/connection/RealConnection.java
```diff
@@ -320,9 +321,16 @@ private void connectTls(ConnectionSpecSelector connectionSpecSelector) throws IO
...
       Handshake unverifiedHandshake = Handshake.get(sslSocketSession);
 
+      List<Certificate> peerCertificates = unverifiedHandshake.peerCertificates();
+
       // Verify that the socket's certificates are acceptable for the target host.
       if (!address.hostnameVerifier().verify(address.url().host(), sslSocketSession)) {
-        X509Certificate cert = (X509Certificate) unverifiedHandshake.peerCertificates().get(0);
+        if (peerCertificates.isEmpty()) {
+          throw new SSLPeerUnverifiedException(
+              "Hostname " + address.url().host() + " not verified, no certificates provided");
```
swankjesse: Apparently there's a thing called “pre-shared keys” that make this case valid! Not sure how to check that though.

yschimke: Valid as in we should check if that is the case and allow those connections?  I'll look into it.

francaracuel:
Hi @yschimke, we got the same crash that appears here: https://github.com/square/okhttp/issues/4208.
Do you know if, finally, are you going to merge it to the master branch?
Thank you!

yschimke:
@francaracuel No, the proper fix was PSK support, and I haven't had a chance to test.

@swankjesse is there a clear handling that would fix this bug for now?

swankjesse:
I think we need a PSK test.

PSK won't have a regular handshake, and so if we support this exotic feature then **everyone** will need to nullcheck their handshakes even on HTTPS connections. Previously we could conveniently say all HTTPS calls had handshakes.

Kind of makes me want to figure out a way to make PSK look like something with a handshake.

yschimke:
I'd rather not fake the test based on how I understand PSK to work.  But conscrypt deprecated it because it doesn't fit with TLS 1.3 apparently, and I haven't found a nice clear sample of configuring it.

yuforia:
Please fix soon,  this has become a really big problem.
@swankjesse We are not using anything special like PSK, this just happens irregularly.
Is there a way to workaround the problem? How can the out of bound exception be caught and the operation of my program resumed?
Currently the "OkHttp Dispatcher" thread crashes and I don't know how to catch the exception and have to restart the whole program.

swankjesse:
@yuforia can you create a new issue with a test case that reproduces this?

PR #4242, opened 2018-08-29
Allow multiple unrequested 1xx responses

### LINKED ISSUE #4188: Multiple 1xx responses not handled properly
Hi,

Issue #1337 fixed the cases where OkHttp receives an unexpected 1xx response from the server. However this only works if the server sends a single 1xx unexpected response. Any additional unexpected 1xx responses are not handled properly. According to [RFC 7231](https://tools.ietf.org/html/rfc7231#section-6.2), clients should be able to cope with this:

> A client MUST be able to parse one or more 1xx responses received
prior to a final response, even if the client does not expect one. A
user agent MAY ignore unexpected 1xx responses.

v113:
This fixes #4188 

I am not sure about the test case (adding a policy just for the test), but could not figure out any other way to test it.

### DIFF
--- a/mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.java
+++ b/mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.java
@@ -78,6 +78,7 @@ import okio.Timeout;
 import org.junit.rules.ExternalResource;
 
 import static okhttp3.internal.Util.closeQuietly;
+import static okhttp3.mockwebserver.SocketPolicy.MULTIPLE_1XX;
 import static okhttp3.mockwebserver.SocketPolicy.CONTINUE_ALWAYS;
 import static okhttp3.mockwebserver.SocketPolicy.DISCONNECT_AFTER_REQUEST;
 import static okhttp3.mockwebserver.SocketPolicy.DISCONNECT_AT_END;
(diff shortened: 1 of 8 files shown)


mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.java
```diff
@@ -662,6 +663,23 @@ private RecordedRequest readRequest(Socket socket, BufferedSource source, Buffer
       sink.flush();
     }
 
+    if (socketPolicy == MULTIPLE_1XX) {
+      sink.writeUtf8("HTTP/1.1 100 Continue\r\n");
+      sink.writeUtf8("Content-Length: 0\r\n");
+      sink.writeUtf8("\r\n");
+      sink.flush();
+
+      sink.writeUtf8("HTTP/1.1 101 Switching Protocols\r\n");
```
swankjesse: I don’t think we should proceed to do HTTP after receiving an HTTP 101.

Instead should we just explicitly whitelist 100 and 102?

swankjesse: (The check for websocket support in particular makes me anxious.)

v113:
I just sent the form.

guillerodriguez:
Hi all. Will this be merged? It would fix #4188 .

swankjesse:
Needs attention. I think the HTTP early hints stuff is big & important and also might need to show up in our API. Needs real work.

yschimke:
Closing since the code is now in kotlin and last commit was 2018

Issue #4227, opened 2018-08-22
Shutdown webserver after N-th request


fred84:
Our application make several sequential http calls. We treat connection error differently for first and subsequent calls, e.g.
- when first request failed mark operation as failed
- when second request failed - schedule retry request later

We can test read timeouts with `MockResponse.setHeadersDelay` or `MockResponse.setSocketPolicy`. But I didn't find similar option for testing connection timeouts.

IMHO it will be handy to have `SocketPolicy.SHUTDOWN_SERVER_AFTER_REQUEST` for such cases.

swankjesse:
You might be able to use a custom Dispatcher to do this.

fred84:
Thanks for suggestion with `Dispatcher`. Current solution looks like:
```java
    @Test
    void shutdownAfterFirstRequest() {
        RestTemplate template = new RestTemplate();

        server.setDispatcher(new Dispatcher() {
            @Override
            public MockResponse dispatch(RecordedRequest request) {
                CompletableFuture.runAsync(() -> {
                    try {
                        server.shutdown();
                    } catch (IOException e) {
                        // ignoring for now
                    }
                });

                return new MockResponse().setBody("hello world");
            }
        });


        assertThat(
                template.getForObject(server.url("/").toString(), String.class),
                equalTo("hello world")
        );

        ResourceAccessException e = assertThrows(
                ResourceAccessException.class,
                () -> template.getForObject(server.url("/").toString(), String.class)
        );
        assertThat(
                e,
                hasCause(instanceOf(ConnectException.class))
        );
    }
```
Other option with SocketPolicy seems much easier (no additional threads created) and at first sight requires only small modification of `MockWebServer`. I've just added 
```java
        if (response.getSocketPolicy() == SHUTDOWN_SERVER) {
          shutdown();
        }
```
after writing http response in `MockWebServer::processOneRequest` and it worked fine.

I'll test it more and prepare PR if you think it worth adding such `SocketPolicy`.

swankjesse:
Please do. Maybe call it `SHUTDOWN_SERVER_AFTER_RESPONSE` ?

PR #4243, opened 2018-08-29
Shutdown MockWebServer after response

fred84:
New SocketPolicy to shutdown MockWebServer after response. More details in #4227

### DIFF
--- a/mockwebserver/src/main/java/okhttp3/mockwebserver/SocketPolicy.java
+++ b/mockwebserver/src/main/java/okhttp3/mockwebserver/SocketPolicy.java
@@ -32,8 +32,8 @@ package okhttp3.mockwebserver;
 public enum SocketPolicy {
 
   /**
-  * Shutdown {@link MockWebServer} after writing response.
-  */
+   * Shutdown {@link MockWebServer} after writing response.
+   */
   SHUTDOWN_SERVER_AFTER_RESPONSE,
 
   /**
--- a/mockwebserver/src/main/java/okhttp3/mockwebserver/SocketPolicy.java
+++ b/mockwebserver/src/main/java/okhttp3/mockwebserver/SocketPolicy.java
@@ -32,6 +32,11 @@ package okhttp3.mockwebserver;
 public enum SocketPolicy {
 
   /**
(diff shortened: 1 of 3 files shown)


mockwebserver/src/main/java/okhttp3/mockwebserver/SocketPolicy.java
```diff
@@ -31,6 +31,11 @@
  */
 public enum SocketPolicy {
 
+  /**
+  * Shutdown {@link MockWebServer} after writing response.
+  */
```
JakeWharton: nit: these two lines should be indented by one space

mockwebserver/src/test/java/okhttp3/mockwebserver/MockWebServerTest.java
```diff
@@ -452,6 +455,24 @@
...
+    HttpURLConnection connection = (HttpURLConnection) url.openConnection();
+    assertEquals(HttpURLConnection.HTTP_OK, connection.getResponseCode());
+
+    HttpURLConnection refusedConnection = (HttpURLConnection) url.openConnection();
+
+    try {
+      refusedConnection.getResponseCode();
+      fail("Second connection should be refused");
+    } catch (ConnectException e ) {
+      assertThat(e, hasMessage(containsString("refused")));
```
JakeWharton: We don't use hamcrest anywhere. If we adopted something new it'd probably be Truth. Perhaps just stick to `assertTrue(e.getMessage().contains("refused"))` for now.

mockwebserver/src/main/java/okhttp3/mockwebserver/SocketPolicy.java
fred84: +

mockwebserver/src/test/java/okhttp3/mockwebserver/MockWebServerTest.java
fred84: fixed

yschimke:
@fred84 if you haven't already, could you complete http://square.github.io/okhttp/#contributing

fred84:
@yschimke completed

PR #4247, opened 2018-09-01
Record proxy path also

yschimke:
For proxied requests, record more data to make the request identifiable.

### DIFF
--- a/mockwebserver/src/test/java/okhttp3/mockwebserver/RecordedRequestTest.java
+++ b/mockwebserver/src/test/java/okhttp3/mockwebserver/RecordedRequestTest.java
@@ -5,7 +5,6 @@ import java.net.Socket;
 import java.net.UnknownHostException;
 import java.util.Collections;
 import okhttp3.Headers;
-import okhttp3.HttpUrl;
 import okio.Buffer;
 import org.junit.Test;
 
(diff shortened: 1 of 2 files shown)


mockwebserver/src/test/java/okhttp3/mockwebserver/RecordedRequestTest.java
```diff
@@ -32,6 +31,16 @@ private FakeSocket(InetAddress inetAddress, int port) {
...
   }
 
+  @Test public void testIPv4Proxy() throws UnknownHostException {
+    Socket socket =
+        new FakeSocket(InetAddress.getByAddress("127.0.0.1", new byte[] { 127, 0, 0, 1 }), 80);
+
+    RecordedRequest request = new RecordedRequest("CONNECT android.com:443 HTTP/1.1", headers,
+        Collections.<Integer>emptyList(), 0, new Buffer(), 0, socket);
+
+    assertEquals("http://127.0.0.1/android.com:443", request.getRequestUrl().toString());
```
swankjesse: I don't think this is right. I would prefer `http://android.com/` here. It's incomplete, but it's all we know about the URL.

We might want to change the parameter in request line from `path` to something that applies in both cases.

yschimke:
Ok. I’ll record the data with enough fidelity that for a connect you can read the status line data separately.

PR #4252, opened 2018-09-05
Properly parse RFC 7235 authentication headers (#4249)

### LINKED ISSUE #4249: okhttp3.internal.http.HttpHeaders#parseChallenges does not handle authentication header with only scheme
~If you e. g. have `Proxy-Authenticate: Basic` which is a valid value, the method will not work properly, because of `int index = header.indexOf(' '); if (index == -1) continue;`, which means the method only works properly if there is a space in the header value.~ (For `Basic` auth `realm` indeed is required)

This also breaks `okhttp3.Response#challenges` and `okhttp3.JavaNetAuthenticator#authenticate`.

Vampire:
fixes #4249

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/HeadersTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/HeadersTest.java
@@ -22,15 +22,19 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import okhttp3.internal.Internal;
-import okhttp3.internal.Util;
 import okhttp3.internal.http.HttpHeaders;
 import okhttp3.internal.http2.Header;
 import okhttp3.internal.http2.Http2Codec;
 import org.junit.Test;
 
+import static java.util.Collections.emptyList;
+import static java.util.Collections.emptyMap;
+import static java.util.Collections.singletonList;
+import static java.util.Collections.singletonMap;
 import static okhttp3.TestUtil.headerEntries;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
@@ -174,7 +178,7 @@ public final class HeadersTest {
 
   @Test public void ofMapThrowsOnEmptyName() {
     try {
-      Headers.of(Collections.singletonMap("", "OkHttp"));
+      Headers.of(singletonMap("", "OkHttp"));
       fail();
     } catch (IllegalArgumentException expected) {
     }
@@ -182,24 +186,24 @@ public final class HeadersTest {
 
   @Test public void ofMapThrowsOnBlankName() {
     try {
-      Headers.of(Collections.singletonMap(" ", "OkHttp"));
+      Headers.of(singletonMap(" ", "OkHttp"));
       fail();
     } catch (IllegalArgumentException expected) {
     }
   }
 
   @Test public void ofMapAcceptsEmptyValue() {
-    Headers headers = Headers.of(Collections.singletonMap("User-Agent", ""));
+    Headers headers = Headers.of(singletonMap("User-Agent", ""));
     assertEquals("", headers.value(0));
   }
 
   @Test public void ofMapTrimsKey() {
-    Headers headers = Headers.of(Collections.singletonMap(" User-Agent ", "OkHttp"));
+    Headers headers = Headers.of(singletonMap(" User-Agent ", "OkHttp"));
     assertEquals("User-Agent", headers.name(0));
   }
 
   @Test public void ofMapTrimsValue() {
-    Headers headers = Headers.of(Collections.singletonMap("User-Agent", " OkHttp "));
+    Headers headers = Headers.of(singletonMap("User-Agent", " OkHttp "));
     assertEquals("OkHttp", headers.value(0));
   }
 
@@ -214,7 +218,7 @@ public final class HeadersTest {
 
   @Test public void ofMapRejectsNullCharInName() {
     try {
-      Headers.of(Collections.singletonMap("User-\u0000Agent", "OkHttp"));
+      Headers.of(singletonMap("User-\u0000Agent", "OkHttp"));
       fail();
     } catch (IllegalArgumentException expected) {
     }
@@ -222,7 +226,7 @@ public final class HeadersTest {
 
   @Test public void ofMapRejectsNullCharInValue() {
     try {
-      Headers.of(Collections.singletonMap("User-Agent", "Square\u0000OkHttp"));
+      Headers.of(singletonMap("User-Agent", "Square\u0000OkHttp"));
       fail();
     } catch (IllegalArgumentException expected) {
     }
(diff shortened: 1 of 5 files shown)


okhttp/src/main/java/okhttp3/Challenge.java
```diff
@@ -16,67 +16,97 @@
...
+  @Deprecated public Challenge(String scheme, String realm) {
     if (scheme == null) throw new NullPointerException("scheme == null");
     if (realm == null) throw new NullPointerException("realm == null");
-    if (charset == null) throw new NullPointerException("charset == null");
-    this.scheme = scheme;
-    this.realm = realm;
-    this.charset = charset;
+    this.scheme = scheme.toLowerCase();
+    Map<String, String> authParams = new HashMap<>();
+    authParams.put("realm", realm);
```
JakeWharton: Can use `Collections.singletonMap`

okhttp/src/main/java/okhttp3/Challenge.java
```diff
@@ -16,67 +16,97 @@
...
 
-  public Challenge(String scheme, String realm) {
-    this(scheme, realm, ISO_8859_1);
+  public Challenge(String scheme, Map<String, String> authParams) {
+    if (scheme == null) throw new NullPointerException("scheme == null");
+    this.scheme = scheme.toLowerCase();
+    if (authParams == null) {
+      this.authParams = emptyMap();
+    } else {
+      Map<String, String> newAuthParams = new HashMap<>();
```
JakeWharton: Please use `LinkedHashMap` to be consistent with the rest of the codebase

okhttp/src/main/java/okhttp3/Challenge.java
```diff
@@ -16,67 +16,97 @@
...
-  private final String realm;
-  private final Charset charset;
+  private final Map<String, String> authParams;
 
-  public Challenge(String scheme, String realm) {
-    this(scheme, realm, ISO_8859_1);
+  public Challenge(String scheme, Map<String, String> authParams) {
+    if (scheme == null) throw new NullPointerException("scheme == null");
+    this.scheme = scheme.toLowerCase();
+    if (authParams == null) {
```
JakeWharton: Nullable collections are a bit suspect. Do we want to allow this?

It's also worth noting that the lack of `@Nullable` annotations on this parameters means that the package-level `@NonNull` annotation takes precedence and Java callers and Kotlin callers enforcing JSR 305 _cannot_ pass null even if they wanted to.

okhttp/src/main/java/okhttp3/internal/http/HttpHeaders.java
```diff
@@ -144,53 +175,92 @@ public static Headers varyHeaders(Headers requestHeaders, Headers responseHeader
...
+      headerChallenges = new ArrayList<>();
 
-      // If a charset is provided, RFC 7617 says it must be "UTF-8".
-      if (charset != null) {
-        if (charset.equalsIgnoreCase("UTF-8")) {
-          challenge = challenge.withCharset(Util.UTF_8);
-        } else {
+      String[] challengeParts = PROXY_AUTHENTICATE_VALUE_SPLIT_PATTERN.split(header);
+      String authScheme = null;
+      Map<String, String> authParams = new HashMap<>();
```
JakeWharton: LinkedHashMap again

Vampire:
Thx for the comments, I'll look into them, but first I have to solve a bug in my algorithm that the tests revealed. :-(

Vampire:
Ok, bug fixed, travis is green, your comments are incorporated. :-)

okhttp-tests/src/test/java/okhttp3/HeadersTest.java
```diff
@@ -434,20 +481,285 @@
...
+        .add("WWW-Authenticate", "Other abc==").build();
+    challenges = HttpHeaders.parseChallenges(headers, "WWW-Authenticate");
+    assertEquals(1, challenges.size());
+    assertEquals("other", challenges.get(0).scheme());
+    assertNull(challenges.get(0).realm());
+    expectedAuthParams.clear();
+    expectedAuthParams.put(null, "abc==");
+    assertEquals(expectedAuthParams, challenges.get(0).authParams());
+
+    // Invalid token68 part and auth params.
```
yschimke: nit: I know the structure is existing, but each of these seem better suited as individual test methods rather than commented blocks.

yschimke:
LGTM - Change and tests look very thorough, but I only gave it a cursory review, not against the spec etc.

okhttp-tests/src/test/java/okhttp3/HeadersTest.java
Vampire: I totally agree, I just did like always with PRs and tried to adapt to the surroundings.
Want me to put all into separate methods, also the existing ones?

Vampire: Oh, I just noticed the test name which makes my tests in there misplaced anyway. :-D

okhttp-tests/src/test/java/okhttp3/HeadersTest.java
```diff
@@ -366,8 +370,14 @@
         .build();
     List<Challenge> challenges = HttpHeaders.parseChallenges(headers, "WWW-Authenticate");
     assertEquals(1, challenges.size());
-    assertEquals("Digest", challenges.get(0).scheme());
```
yschimke: I'm ok with this, but we should mention this as a potentially breaking change.  I hope most people using challenges are already using them case insensitively.

e.g. this will break

https://github.com/rburgst/okhttp-digest/blob/0205d0ef00273ce800bb012d4fc1cfc72857a3bc/src/test/java/com/burgstaller/okhttp/WrongOrderedAuthenticationHeaderTest.java#L27

okhttp/src/main/java/okhttp3/Challenge.java
```diff
@@ -16,67 +16,92 @@
...
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.Map.Entry;
 import javax.annotation.Nullable;
 
+import static java.util.Collections.singletonMap;
+import static java.util.Collections.unmodifiableMap;
 import static okhttp3.internal.Util.ISO_8859_1;
 
 /** An RFC 7617 challenge. */
```
yschimke: nit: We now refer to this elsewhere as a 7235 challenge.  7617 is Basic auth IIUC

okhttp/src/main/java/okhttp3/internal/http/HttpHeaders.java
```diff
@@ -144,53 +181,116 @@ public static Headers varyHeaders(Headers requestHeaders, Headers responseHeader
...
-      String charset = null;
-
-      Matcher matcher = PARAMETER.matcher(header);
-      for (int i = index; matcher.find(i); i = matcher.end()) {
-        if (header.regionMatches(true, matcher.start(1), "realm", 0, 5)) {
-          realm = matcher.group(3);
-        } else if (header.regionMatches(true, matcher.start(1), "charset", 0, 7)) {
-          charset = matcher.group(3);
-        }
+      List<Challenge> headerChallenges = CHALLENGES_CACHE.get(header);
```
yschimke: The cache makes sense given the repeated nature of the non-preemptive challenges.  But I don't think we should have it in an unbounded form.

I think we should either restrict it in some way e.g. TTL, size.  Or just accept the wasted effort, which is hopefully noise compared to the network cost.

okhttp/src/main/java/okhttp3/internal/http/HttpHeaders.java
```diff
@@ -144,53 +181,116 @@ public static Headers varyHeaders(Headers requestHeaders, Headers responseHeader
...
+   * No semantic validation is done, for example that {@code Basic} auth must have a
+   * {@code realm} auth param, this is up to the caller that interprets these challenges.
    */
   public static List<Challenge> parseChallenges(Headers responseHeaders, String challengeHeader) {
-    // auth-scheme = token
-    // auth-param  = token "=" ( token | quoted-string )
-    // challenge   = auth-scheme 1*SP 1#auth-param
-    // realm       = "realm" "=" realm-value
-    // realm-value = quoted-string
     List<Challenge> challenges = new ArrayList<>();
```
yschimke: TIOLI: is it worth caching and returning an immutable view of the List here?

okhttp-tests/src/test/java/okhttp3/HeadersTest.java
swankjesse: I’m curious what the benefit of forcing lowercase is? It costs an allocation.

okhttp-tests/src/test/java/okhttp3/HeadersTest.java
```diff
@@ -366,8 +370,14 @@
...
+    assertEquals("digest", challenges.get(0).scheme());
     assertEquals("myrealm", challenges.get(0).realm());
+    Map<String, String> expectedAuthParams = new HashMap<>();
```
swankjesse: nit: we never use HashMap and you shouldn’t either!
https://publicobject.com/2016/02/08/linkedhashmap-is-always-better-than-hashmap/

okhttp-tests/src/test/java/okhttp3/HeadersTest.java
```diff
@@ -366,8 +370,14 @@
...
+    expectedAuthParams.put("realm", "myrealm");
+    expectedAuthParams.put("nonce", "fjalskdflwejrlaskdfjlaskdjflaksjdflkasdf");
+    expectedAuthParams.put("qop", "auth");
+    expectedAuthParams.put("stale", "FALSE");
+    assertEquals(expectedAuthParams, challenges.get(0).authParams());
 
     // Not strict RFC 2617 header.
```
swankjesse: ... for what it’s worth, this really wants to be cut into several tests

okhttp-tests/src/test/java/okhttp3/HeadersTest.java
```diff
@@ -434,43 +482,254 @@
...
+    Headers headers = new Headers.Builder()
+        .add("WWW-Authenticate", " ,  , Basic realm=myrealm")
+        .build();
+    assertEquals(singletonList(new Challenge("Basic", singletonMap("realm", "myrealm"))),
+        HttpHeaders.parseChallenges(headers, "WWW-Authenticate"));
+  }
+
+  @Test public void spacesAroundKeyValueSeparator() {
+    Headers headers = new Headers.Builder()
+            .add("WWW-Authenticate", "Basic realm = \"myrealm\"")
```
swankjesse: nit: https://github.com/square/java-code-styles/

okhttp/src/main/java/okhttp3/Challenge.java
```diff
@@ -16,67 +16,92 @@
...
-  private final String realm;
-  private final Charset charset;
+  private final Map<String, String> authParams;
 
-  public Challenge(String scheme, String realm) {
-    this(scheme, realm, ISO_8859_1);
+  public Challenge(String scheme, Map<String, String> authParams) {
+    if (scheme == null) throw new NullPointerException("scheme == null");
+    if (authParams == null) throw new NullPointerException("authParams == null");
+    this.scheme = scheme.toLowerCase();
```
swankjesse: I don’t think this is worthwhile, and besides it won’t work in Turkey!
https://en.wikipedia.org/wiki/Dotted_and_dotless_I

okhttp/src/main/java/okhttp3/Challenge.java
```diff
@@ -16,67 +16,92 @@
...
+    Map<String, String> newAuthParams = new LinkedHashMap<>();
+    for (Entry<String, String> authParam : authParams.entrySet()) {
+      String key = (authParam.getKey() == null) ? null : authParam.getKey().toLowerCase();
+      newAuthParams.put(key, authParam.getValue());
+    }
+    this.authParams = unmodifiableMap(newAuthParams);
   }
 
-  private Challenge(String scheme, String realm, Charset charset) {
+  @Deprecated public Challenge(String scheme, String realm) {
```
swankjesse: why deprecate? this is still good, isn’t it?

swankjesse: won’t most callers want an easy two-argument overload?

okhttp/src/main/java/okhttp3/Challenge.java
```diff
@@ -16,67 +16,92 @@
...
-    this.realm = realm;
-    this.charset = charset;
+    this.scheme = scheme.toLowerCase();
+    this.authParams = singletonMap("realm", realm);
   }
 
   /** Returns a copy of this charset that expects a credential encoded with {@code charset}. */
-  public Challenge withCharset(Charset charset) {
-    return new Challenge(scheme, realm, charset);
+  @Deprecated public Challenge withCharset(Charset charset) {
```
swankjesse: why deprecate?

okhttp/src/main/java/okhttp3/Challenge.java
```diff
@@ -16,67 +16,92 @@
...
   public Charset charset() {
-    return charset;
+    String charset = authParams.get("charset");
+    if (charset != null) {
+      try {
+        return Charset.forName(charset);
+      } catch (Exception ignore) {
+      }
+    }
+    return ISO_8859_1;
```
swankjesse: is this the official default? I’d prefer ASCII if possible

okhttp/src/main/java/okhttp3/Challenge.java
```diff
@@ -16,67 +16,92 @@
...
-    result = 31 * result + charset.hashCode();
+    result = 31 * result + authParams.hashCode();
     return result;
   }
 
   @Override public String toString() {
-    return scheme
-        + " realm=\"" + realm + "\""
-        + " charset=\"" + charset + "\"";
+    return scheme + " authParams=\"" + authParams + "\"";
```
swankjesse: quoting the auth params seems weird now that they’re in a map

okhttp/src/main/java/okhttp3/Response.java
```diff
@@ -225,10 +225,16 @@ public boolean isRedirect() {
...
-   * is 401 unauthorized, this returns the "WWW-Authenticate" challenges. If the response code is
-   * 407 proxy unauthorized, this returns the "Proxy-Authenticate" challenges. Otherwise this
-   * returns an empty list of challenges.
+   * Returns the RFC 7235 authorization challenges appropriate for this response's code.
+   * If the response code is 401 unauthorized, this returns the "WWW-Authenticate" challenges.
+   * If the response code is 407 proxy unauthorized, this returns the "Proxy-Authenticate"
+   * challenges. Otherwise this returns an empty list of challenges.
+   * <p>
+   * If a challenge uses the {@code token68} variant instead of auth params,
+   * there is exactly one auth param in the challenge at key {@code null}.
```
swankjesse: key null doesn’t feel great, though I don’t have better ideas

okhttp/src/main/java/okhttp3/internal/http/HttpHeaders.java
```diff
@@ -38,10 +40,45 @@
...
+  private static final Pattern QUOTED_PAIR_PATTERN = Pattern.compile(QUOTED_PAIR_PATTERN_PART);
+
+  private static final Pattern PROXY_AUTHENTICATE_VALUE_SPLIT_PATTERN =
+          Pattern.compile(PROXY_AUTHENTICATE_VALUE_SPLIT_PATTERN_PART);
+  private static final Pattern WHITESPACE_SPLIT_PATTERN = Pattern.compile(" +");
+  private static final Pattern AUTH_PARAM_SPLIT_PATTERN =
+          Pattern.compile(OWS_PATTERN_PART + '=' + OWS_PATTERN_PART);
+  private static final Pattern QUOTED_STRING_AUTH_PARAM_AT_END_PATTERN =
+          Pattern.compile(TOKEN_PATTERN_PART + OWS_PATTERN_PART + '=' + OWS_PATTERN_PART
+                  + QUOTED_STRING_PATTERN_PART + '$');
```
swankjesse: The time to init this class just went up ... is there a way to do this with less code?

okhttp/src/main/java/okhttp3/internal/http/HttpHeaders.java
```diff
@@ -144,53 +181,116 @@ public static Headers varyHeaders(Headers requestHeaders, Headers responseHeader
   }
 
   /**
-   * Parse RFC 7617 challenges, also wrong ordered ones.
-   * This API is only interested in the scheme name and realm.
+   * Parse RFC 7235 challenges.
+   * If a challenge uses the {@code token68} variant instead of auth params,
```
swankjesse: this doc is duplicated from `Response.challenges()`. Since this class isn’t public API, maybe just omit it?

okhttp/src/main/java/okhttp3/internal/http/HttpHeaders.java
swankjesse: Yeah, I don’t want a ConcurrentHashMap here. The solution is to make decoding challenges fast & simple, not to cache it!

okhttp/src/main/java/okhttp3/internal/http/HttpHeaders.java
```diff
@@ -144,53 +181,116 @@ public static Headers varyHeaders(Headers requestHeaders, Headers responseHeader
...
    */
   public static List<Challenge> parseChallenges(Headers responseHeaders, String challengeHeader) {
-    // auth-scheme = token
-    // auth-param  = token "=" ( token | quoted-string )
-    // challenge   = auth-scheme 1*SP 1#auth-param
-    // realm       = "realm" "=" realm-value
-    // realm-value = quoted-string
     List<Challenge> challenges = new ArrayList<>();
     List<String> authenticationHeaders = responseHeaders.values(challengeHeader);
+headerLoop:
```
swankjesse: This all feels like a large solution chasing a small problem.

yschimke:
If there are real world examples of bugs with various proxies and services, it makes it more compelling to take on these changes. If the real world already works because noone really implements all this logic and in practice what we have is good enough, the taking on the additional complexity is a tradeoff.

@Vampire do you have any bug reports to show which servers etc are problematic?

okhttp-tests/src/test/java/okhttp3/HeadersTest.java
yschimke: I like this, but it's whether we want a small amount of upgrade pain for what is probably a small set of users.  I'm not as concerned about the allocation, nor turkish i characters :)

okhttp/src/main/java/okhttp3/Challenge.java
Vampire: You are totally correct.
It was not a pure 7617 challenge before already.
It was any 7235 challenge, but only had their algorithm, realm and charset properties while it now has all their auth params and can be used to implement any auth algorithm.
I changed the comment.

okhttp/src/main/java/okhttp3/internal/http/HttpHeaders.java
Vampire: So you want me to simply remove the cache?

okhttp-tests/src/test/java/okhttp3/HeadersTest.java
Vampire: Well, it's only in the tests, but sure, I'll replace it, no problem

okhttp-tests/src/test/java/okhttp3/HeadersTest.java
Vampire: Ok, I split the tests

okhttp-tests/src/test/java/okhttp3/HeadersTest.java
Vampire: You mean the indentation I guess? My bad, fixed.

okhttp/src/main/java/okhttp3/Challenge.java
Vampire: What do you mean?
The `toLowerCase`?
I don't think there will be auth schemes or auth param names in turkey, do you?
I personally would that accept that as case where one then would have to parse the challenge header themselves.

okhttp/src/main/java/okhttp3/Challenge.java
Vampire: Well, I'd have expected this to be only used by OkHttp internally mostly, but I can undeprecate it of course, was just a suggestion. :-)

Vampire: What about the `withCharset` that I also deprecated?

okhttp/src/main/java/okhttp3/Challenge.java
Vampire: Well, I thought it is better to use the constructer with charset right away. But I'll undeprecate it, also was just a suggestion.

okhttp/src/main/java/okhttp3/Challenge.java
Vampire: Well, it is what OkHttp already had as default.
Actually it depends on the concrete auth scheme what is the default.
For `Basic` it is the default afair.
Actually I'd say the default should be `null` as it is auth scheme specific, but that will break implementations of `Basic` auth that depend it to be set.

My first thought was to have `Challenge` without any `getRealm` and `getCharset` methods as both are genrally optional and only mandatory by auth scheme and then have a `BasicChallenge` subclass that adds those methods for basic auth. But I tried to do my changes backwards compatibly.

What do you think?

okhttp/src/main/java/okhttp3/Response.java
Vampire: Well, I had in mind either the token as key and a null value or a null key and the token as value or a separate method.
But as it is an exclusive or between the token and the auth params I thought this is the most elegant solution.
If you want me to change it to something else, or maybe a fixed key `token68`, just tell me.

okhttp/src/main/java/okhttp3/internal/http/HttpHeaders.java
Vampire: Well, the class init is only done once per JVM run, so I thought this is not too much of an issue, don't you?

okhttp-tests/src/test/java/okhttp3/HeadersTest.java
Vampire: I'd hope most users stick to the spec and compare case-insentively, so shouldn't hit too many I hope

Vampire:
> My current thought is to land this and then follow up myself to try to cut the implementation down to something I find palatable. Usually that kind of exercise either sheds some code or gives me a better appreciation for the problem being solved.
>
> Would you be cool with that?

Sure, if you come up with better code that passes the tests, be my guest. :-)

> do you have any bug reports to show which servers etc are problematic?

No, unfortunately not.
I just implemented / tested proxy support for Javacord and thus also tested authenticated proxies.
For that I sent valid authentication headers and found that OkHttp doesn't properly parse them as shown in the issue I referenced.
I implemented the proper parsing in Javacord itself currently to make it work with all proxies that might exist and stick to the RFC and thought I'd contribute it upstream, as I also was asked to in the issue. :-)

Vampire:
New push with the resolved conversations incorporated is done

okhttp/src/main/java/okhttp3/internal/http/HttpHeaders.java
Vampire: Ok, removed

okhttp/src/main/java/okhttp3/Challenge.java
swankjesse: The issue is that if you have an Android device configured with the default locale as Turkey then "BASIC".toLowerCase() returns "basıc". So you should never use that method without passing a parameter of Locale.US.

swankjesse: https://developer.android.com/reference/java/util/Locale#default_locale

Vampire: Oh, ok, I wasn't aware of that.
Escpecially as it doesn't happen here if I use turkish locale for the `toLowerCase` call.
I added the explicit US locale. now

yschimke: TIL!

yschimke:
I think there is an awkward venn diagram of

1) Things obviously valid because of RFC (must?)
2) Things observed from mainstream servers (FANG + Apache + Nginx etc)
3) Things technically valid because of RFC (should?)
4) Things observed with various one off servers e.g. 1 person github implementation of proxies in Ruby

These may conflict, and the priority of OkHttp is probably 1+2 and a simple fast understandable implementation.

It sounds like landing this and iterating on getting it as simple as possible and explicitly deciding to not support things from 3+4 is a valid way forward.  Based on the high quality of the tests you have added, the tests that start failing will be good documentation for sacrifices made.

Vampire:
Well, I'm usually all for proper spec support plus supporting broken implementations if necessary.

If you feel like not supporting valid cases for having simpler code, well it's your project and thus your decision. :-)

swankjesse:
Also, when in doubt we can see what Firefox & Chrome do. Though OkHttp isn’t a web browser, password challenges are browser-style authentication. That means we probably don’t need to support arbitrary schemes; instead Basıc & Digest are enough.

Server-to-server authentication should almost always be bearer tokens or client certificates and I assume will always be pre-emptive.

I assume server-to-proxy challenges are rarely used in practice.

yschimke:
@Vampire you could end up with a choice of implementing the RFC precisely but breaking some major servers.  And complexity to support a theoretical server that doesn't exist is probably the awkward point.

On broken servers, we have often closed okhttp issues suggesting they file issues against the dodgy server.

Vampire:
Well, it is only the challenge parsing and that can be done scheme-independently like it is done in my PR.
Which schemes to support e. g. by the `JavaNetAuthenticator` is a different topic.
But with the challenges parsed properly, any `okhttp3.Authenticator` can support any standard or non-standard authentication scheme, as all challgenges are available.
Of course a custom authenticator can also parse the headers themselves, but then there should not be a `challenges()` method in OkHttp, but a `basicChallenges()` and maybe a `digestChallenges()` method that only return the challenges in question. But even then it would probably be good to properly parse the headers like with multiple schemes and so on.

One could e. g. have a corporate proxy that supports `Negotiate` authentication providing SSO through the Windows login and so on.

Vampire:
> @Vampire you could end up with a choice of implementing the RFC precisely but breaking some major servers. And complexity to support a theoretical server that doesn't exist is probably the awkward point.

I don't think my changes will break anything, except maybe if someone did not compare the scheme name or auth param keys case-insensitively. Everything else should work like before, plus supporting more valid cases that weren't supported before.

swankjesse:
The discussion reminds me of [descriptive vs. prescriptive in language](https://en.m.wikipedia.org/wiki/Linguistic_prescription). I started my career as a prescriptivist, and have begrudgingly grown to be more of a descriptivist!

Specs and software grow apart. Specs are overly broad so they can support unknown future use cases. Software doesn’t follow specs strictly, sometimes to support use cases or constraints that were not anticipated.

yschimke:
I'd still suggest landing and then simplifying with good tests in place (Thanks @Vampire!), and possibly decide which to accept breaking.

Vampire:
Made some minor adjustments in naming and removed a comment that didn't apply anymore.

swankjesse:
Followup: https://github.com/square/okhttp/pull/4359

PR #4256, opened 2018-09-08
[Future] Okio 2.2.2

yschimke:
Basically asking the question about what happens if we find a bug in Okio.  Are we ready to take the hit of including the kotlin library (1mb) now, or will we defer and possibly release maintenance versions of okio?


JakeWharton:
It's not really a hit of any significance. If you care about size then you are using a shrinker and 99% of Kotlin will fall away. If you don't care about size then it could be 10KB, 1MB, or 10MB and not make a difference.

If Okio gets `BufferedSource.peek()` I will fight for the bump. Unless there's a feature we should probably just wait to stave off complaints.

okhttp/src/main/java/okhttp3/HttpUrl.java
```diff
@@ -1699,42 +1701,48 @@ static String canonicalize(String input, int pos, int limit, String encodeSet,
...
+            encodedCharBuffer.writeString(input, i, i + Character.charCount(codePoint), charset);
+          }
 
-        while (!encodedCharBuffer.exhausted()) {
-          int b = encodedCharBuffer.readByte() & 0xff;
-          out.writeByte('%');
-          out.writeByte(HEX_DIGITS[(b >> 4) & 0xf]);
-          out.writeByte(HEX_DIGITS[b & 0xf]);
+          while (!encodedCharBuffer.exhausted()) {
+            int b = encodedCharBuffer.readByte() & 0xff;
```
yschimke: Exception thrown here

yschimke:
@JakeWharton Yep, I'm happy to close this, just interesting to see the discussion.  Also the public API change was interesting as well.

yschimke:
FWIW i'm already running with Okio 2.0 and haven't noticed any issues

yschimke:
Also there is a decent niche for nice kotlin suspend aware IO functions.  Would be nice if Okio filled that spot as opposed to using two libraries.

yschimke:
Okio 2.1.0 now with 100% more peek

https://github.com/square/okio/blob/6d6b5158141cfd9753dcc417a90c55a7d5b3651c/okio/jvm/src/main/java/okio/BufferedSource.kt#L575-L594

yschimke:
@swankjesse @Egorand A few more changes needed based on public API changes

PromanSEW:
I don't like this change because of Kotlin Runtime which violates "lightweight" conception of OkHttp client

yschimke:
@PromanSEW no rush, just for discussion, but I assume it's coming at some point.  Okio already moved to Kotlin.

PromanSEW:
@yschimke I maintain the game which uses Appodeal, so I have 90k+ methods and 2 DEX files already (16 MB of code)
I use Okio 1.16.0 because I don't want to have additional 1 MB of Kotlin library code, unused by me at all.
Well, the game is heavy-weight application, so 1 MB is small change.
But, what about small lightweight apps, about 5-10 MB as example?
Please think about it. It is worth to have heavy dependency (which has 1 MB code vs OkHttp has about 420 KB of code itself) for using latest Okio, or not.
Only few developers use Kotlin now, so I think remaining won't like it if they found new dependency with HUGE amount of unused code and possible hidden problems (with ProGuard warnings / errors for example)

Egorand:
@PromanSEW Okio uses a very small number of methods in `kotlin-stdlib`, so if you're using a shrinker (which you should), the actual footprint will be well under 100 Kb. Also, 

> Only few developers use Kotlin now

If we're speaking about Android, then this is not true, according to Google's and JetBrains' stats. And the percentage of apps that are starting to use Kotlin will just grow over time.

PromanSEW:
@Egorand well, it's OK now

yschimke:
@PromanSEW also why I raised this PR for discussion, see the description at the top.  But ultimately unless upstream dependencies agree to have two active development branches, we will _eventually_ depend on a kotlin version of okio.

swankjesse:
OkHttp will have a Kotlin dependency by the end of 2019. Apps should use R8 or ProGuard to shrink artifacts as necessary.

yschimke:
I'm going to leave this around to track compatibility, not to land.

yschimke:
Closing this as I presume it can be wrapped into OkHttp 4

Issue #4238, opened 2018-08-26
route parameter of authenticate can be null but is not marked as @Nullable


technoir42:
Here is a test proving that `route` can be `null` (I'm using 3.11.0):

```java
import org.junit.Test;

import okhttp3.Authenticator;
import okhttp3.OkHttpClient;
import okhttp3.Request;
import okhttp3.Response;
import okhttp3.Route;

public class AuthenticatorTest {
    @Test
    public void routeCanBeNull() throws Exception {
        Authenticator authenticator = new Authenticator() {
            @Override
            public Request authenticate(Route route, Response response) {
                if (route == null) {
                    throw new NullPointerException("route is null");
                }
                return null;
            }
        };

        OkHttpClient client = new OkHttpClient.Builder()
                .authenticator(authenticator)
                .build();

        Request request = new Request.Builder()
                .url("https://httpbin.org/status/401")
                .build();

        client.newCall(request).execute(); // route is not null
        client.newCall(request).execute(); // route is null
    }
}
```

I'm not sure if it is a correct behaviour. If it is actually correct I'll make a PR adding `@Nullable` to `route`.

yschimke:
Yep, please submit a PR.  I've also hit this using another auth library, in my case also because Interceptor.Chain.connection is null, so an interceptor calling a Authenticator will never have a route.

swankjesse:
Related to https://github.com/square/okhttp/issues/3809

swankjesse:
I think we _could_ get a route if our internal APIs were a little better organized. The reason the HTTP bin one doesn’t have a route is because the response body is empty and we’ve gotten rid of things by the time we’re ready to authenticate.

My preference on this is to fix RetryAndFollowUpInterceptor, possibly with support from StreamAllocation, so that we have a route when we call into the Authenticator. That feels like a better fix than changing the type to be nullable.

yschimke:
It still makes it awkward if you want to reuse an Authenticator from an Interceptor for premptive auth.  Making route non @Nullable means it always need to be a network interceptor.  So I still think it's worth making route @Nullable

swankjesse:
Good point!

yschimke:
It's less of a good point, because you can't use it prior to sending the first request, but you can use it when implementing a app level interceptor.  I hit that this week with a JIRA server returning a 400 when an anonymous query request fails, and I'd like to reuse the authenticator.

MONIKA-KUMAR-JETHANI:
[...]

PR #4257, opened 2018-09-08
Nullable route annotation

yschimke:
Related to https://github.com/square/okhttp/issues/4238

Currently (by code structure) route may be null when used internally.  In addition as an API for handling authentication challenges, it makes sense that route may not be always available, even though it can be useful when determining when to apply a authentication method.


Ribesg:
What am I supposed to do while waiting for this to be released? Is there a workaround?

JakeWharton:
You can use the snapshot version.

Issue #1359, opened 2015-02-01
Socks 4 Proxy broken


dain:
OkHttp relies on the JVM implementation of SOCKS, which seems to be broken for SOCKS4.  When the JVM establishes the connection in `java.net.SocksSocketImpl#connect`, it first sends a SOCKS5 setup message:

``` java
        // This is SOCKS V5
        out.write(PROTO_VERS);
        out.write(2);
        out.write(NO_AUTH);
        out.write(USER_PASSW);
        out.flush();
```

and then reads a 2 byte response.  Then it checks if the first byte of the response is not `5`, and if so it tries a socks 4 message _using the same socket_.

The problem is since this above message is not a wellformed SOCKS4 message, the SOCKS4 server doesn't know how much data to read.  Also a SOCKS4 response is 8 bytes, and the code only drains 2 bytes.  This means that when it gets to the SOCKS4 connect, the socket is in an unknown state and always fails.

I'm not sure anything can be done other than hacking the JVM with reflection, or to write a new implementation of SOCKS 4 and 5.  

Regardless of the solution the broken SOCKS 4 support should be documented, so no one else has to dig into the terrible SOCKS code in the JVM.

swankjesse:
Any advice on how we should handle this in OkHttp?

 * We could implement SOCKS in the application layer, which is annoying, but not _that_ annoying. We already do it for our test server.
 * We could document that this is a problem. (Where would we do that?)

dain:
I'm not sure there is any way around this other than to implement SOCKS in OkHttp.  The protocol is pretty simple, and I have a testing SOCKS proxy implementation here https://github.com/airlift/airlift/blob/master/http-client/src/test/java/io/airlift/http/client/jetty/TestingSocksProxy.java. 

As for where to document this, I'd add it to the FAQ.  It would be good to test this on Android also, as it might correctly implement SOCKS.

swankjesse:
Yup. Our testing proxy is here: https://github.com/square/okhttp/blob/master/okhttp-tests/src/test/java/com/squareup/okhttp/SocksProxy.java

Who is using SOCKS4?

dain:
I only noticed it because my testing socks proxy only implemented SOCKS4.

swankjesse:
I'm putting this in the icebox for now. If/when somebody really wants SOCKS4, they can bring it up here and probably submit a pull request.

Vampire:
SOCKS 4 support in Java 8 is totally broken, pretty much as described here.
[...]

PR #4265, opened 2018-09-13
Try to resolve addresses with SOCKS proxies too, as SOCKS4 cannot work with unresolved addresses (#1359)

Vampire:
Fix for #1359 when used with Java 9 which fixed SOCKS4 support

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/internal/connection/RouteSelectorTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/internal/connection/RouteSelectorTest.java
@@ -91,7 +91,7 @@ public final class RouteSelectorTest {
 
     assertTrue(routeSelector.hasNext());
     dns.set(uriHost, dns.allocate(1));
-    RouteSelector.Selection selection = routeSelector.next();
+    RouteSelector.Selection selection = routeSelector.next(false);
     assertRoute(selection.next(), address, NO_PROXY, dns.lookup(uriHost, 0), uriPort);
     dns.assertRequests(uriHost);
     assertFalse(selection.hasNext());
@@ -103,7 +103,7 @@ public final class RouteSelectorTest {
 
     assertFalse(routeSelector.hasNext());
     try {
-      routeSelector.next();
+      routeSelector.next(false);
       fail();
     } catch (NoSuchElementException expected) {
     }
@@ -116,11 +116,11 @@ public final class RouteSelectorTest {
 
     assertTrue(routeSelector.hasNext());
     dns.set(uriHost, dns.allocate(1));
-    RouteSelector.Selection selection = routeSelector.next();
+    RouteSelector.Selection selection = routeSelector.next(false);
     Route route = selection.next();
     routeDatabase.failed(route);
     routeSelector = new RouteSelector(address, routeDatabase, null, EventListener.NONE);
-    selection = routeSelector.next();
+    selection = routeSelector.next(false);
     assertRoute(selection.next(), address, NO_PROXY, dns.lookup(uriHost, 0), uriPort);
     assertFalse(selection.hasNext());
 
@@ -132,7 +132,7 @@ public final class RouteSelectorTest {
 
     assertFalse(routeSelector.hasNext());
     try {
-      routeSelector.next();
+      routeSelector.next(false);
       fail();
     } catch (NoSuchElementException expected) {
     }
@@ -146,7 +146,7 @@ public final class RouteSelectorTest {
 
     assertTrue(routeSelector.hasNext());
     dns.set(proxyAHost, dns.allocate(2));
-    RouteSelector.Selection selection = routeSelector.next();
+    RouteSelector.Selection selection = routeSelector.next(false);
     assertRoute(selection.next(), address, proxyA, dns.lookup(proxyAHost, 0), proxyAPort);
     assertRoute(selection.next(), address, proxyA, dns.lookup(proxyAHost, 1), proxyAPort);
 
@@ -164,7 +164,7 @@ public final class RouteSelectorTest {
 
     assertTrue(routeSelector.hasNext());
     dns.set(uriHost, dns.allocate(2));
-    RouteSelector.Selection selection = routeSelector.next();
+    RouteSelector.Selection selection = routeSelector.next(false);
     assertRoute(selection.next(), address, NO_PROXY, dns.lookup(uriHost, 0), uriPort);
     assertRoute(selection.next(), address, NO_PROXY, dns.lookup(uriHost, 1), uriPort);
 
@@ -193,7 +193,7 @@ public final class RouteSelectorTest {
         EventListener.NONE);
     assertTrue(routeSelector.hasNext());
     dns.set(uriHost, dns.allocate(1));
-    RouteSelector.Selection selection = routeSelector.next();
+    RouteSelector.Selection selection = routeSelector.next(false);
     assertRoute(selection.next(), address, NO_PROXY, dns.lookup(uriHost, 0), uriPort);
     dns.assertRequests(uriHost);
 
@@ -208,7 +208,7 @@ public final class RouteSelectorTest {
 
     assertTrue(routeSelector.hasNext());
     dns.set(uriHost, dns.allocate(2));
-    RouteSelector.Selection selection = routeSelector.next();
+    RouteSelector.Selection selection = routeSelector.next(false);
     assertRoute(selection.next(), address, NO_PROXY, dns.lookup(uriHost, 0), uriPort);
     assertRoute(selection.next(), address, NO_PROXY, dns.lookup(uriHost, 1), uriPort);
 
@@ -230,7 +230,7 @@ public final class RouteSelectorTest {
     // First try the IP addresses of the first proxy, in sequence.
     assertTrue(routeSelector.hasNext());
     dns.set(proxyAHost, dns.allocate(2));
-    RouteSelector.Selection selection1 = routeSelector.next();
+    RouteSelector.Selection selection1 = routeSelector.next(false);
     assertRoute(selection1.next(), address, proxyA, dns.lookup(proxyAHost, 0), proxyAPort);
     assertRoute(selection1.next(), address, proxyA, dns.lookup(proxyAHost, 1), proxyAPort);
     dns.assertRequests(proxyAHost);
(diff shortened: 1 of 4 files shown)


Vampire:
> OkHttp is the wrong layer for this fix. It belongs in the underlying platform.

Sorry, but I don't quite agree.

> We can't resolve IP addresses here because in practice they don’t exist here. If they did, then what the heck is the point of the proxy?

Access control.
One of the typical use-case of SOCKS proxies afair is a firewall that controls who is allowed to connect where-to.

The problem with SOCKS4 (not SOCKS4a) is, that you **must** resolve the IP address locally.
The whole protocol is only one request and one response while the request is `0x4 0x1 <two byte port> <four byte numerical IP> <0-terminated username>`.
There is no way to use SOCKS4 with a hostname.

SOCKS4a added the possibility to specify an invalid IP, namely `0x0 0x0 0x0 <anthing besides 0x0>` and then append the 0-terminated hostname after the 0-terminated username.

SOCKS5 also added the possibility to only specify a hostname or to specify an IPv6 address.

But with SOCKS4, the hostname must be resolved on the client side. If only the proxy server can resolve the hostname, then either SOCKS4a or SOCKS5 has to be used.

> Imagine a proxy that provides access not to the broader Internet, but to a private internal network. The DNS mappings in that network are not visible to the OkHttpClient host. Heck, they may even contradict what OkHttp sees.

As written above, then SOCKS4 cannot be used in such an environment.

And think about the other way around, what if the client can resolve the hostname to an IP, but the proxy cannot do it?
Also a valid use-case that is supported with my changes, but not without them.

> It’s unfortunate that the JDK maintainers got this wrong.

I don't think they did, as the protocol hardly requires an IP being available and if the `InetSocketAddress` got intentionally created unresolved, who are they to now try to resolve it.

> To actually work around it we’d need to build our own complete SOCKS implementation.

Even then you have to resolve the IP address locally to use SOCKS4, as the protocol does not work without as described above.

> I don’t want to do that.

To quote yourself from the issue referenced above: `We could implement SOCKS in the application layer, which is annoying, but not that annoying.`
So you actually do consider doing it on application level if someone makes a PR for it (as you stated in a later comment).

Vampire:
Not really.
In the first naive implementation I did, it would, as the resolution attempt would throw an `UnknownHostException`.
Fortunately there are tests that cover this case, so I change to catch and rethrow the exception if it is not a SOCKS proxy.
For a SOCKS5 proxy where the client does not know the hostname, the UHE is caught and an unresolved `InetSocketAddress` created just like before.

So with my changes you have:

- SOCKS4 and only client can resolve the hostname => IP is resolved, everything works
- SOCKS4 and only proxy can resolve the hostname => unsupported as the protocol cannot handle it
- SOCKS4a and only client can resolve the hostname => IP is resolved, everything works
- SOCKS4a and only proxy can resolve the hostname => everything would work if the SOCKS socket factory would support SOCKS4a (which the standard JDK does not)
- SOCKS5 and only the client can resolve the hostname => IP is resolved, everything works
- SOCKS5 and only the proxy can resolve the hostname => hostname is sent, everything works

before it was:
- SOCKS4 => unsupported as no resolution attempt is done
- SOCKS4a and only client can resolve the hostname => unsupported as no resolution attempt is done
- SOCKS4a and only proxy can resolve the hostname => everything would work if the SOCKS socket factory would support SOCKS4a (which the standard JDK does not)
- SOCKS5 and only the client can resolve the hostname => unsupported as no resolution attempt is done
- SOCKS5 and only the proxy can resolve the hostname => hostname is sent, everything works

The only issue would be if client and proxy disagree on the IP for the hostname, as neither protocol can send both the hostname and the IP.
But you cannot support both, IP resolution on client side and covering that client and proxy may disagree on the IP for a hostname.

The only way would be to have a manual setting, so that the user can decide whether to try resolution on the client or not, but even then I'd recommend to default to do the resolution on the client as I think it is more common that they do not disagree if both can resolve the hostname.

One other way to handle it would be to only try the IP resolution if SOCKS4 is used, but the API for `Proxy` does not allow to determine which SOCKS version is to be used, unless you cast the proxy to `sun.net.SocksProxy` which is then bound to only work on JVMs that have this internal class and I thought it is more appropriate to just always try the IP resolution and just fall back to hostname mode if the IP couldn't be resolved.

swankjesse:
Lemme restate: For networks like tor there are serious consequences to getting this wrong.

If you do a DNS request for say `whistleblowing.onion` on a Tor + SOCKS5 proxy, you do not want your destination hostname to leak out with a local DNS request.

swankjesse:
Honest question: who uses SOCKS4? Shouldn’t we just encourage these people to get SOCKS5 or JDK9? I don’t want to add code or complexity to OkHttp unless it has a big positive impact.

yschimke:
My view is that at the worst we should expose internal addresses if a SOCKS4 proxy is used.  But private corporate addresses should be leaked like this by default.

Vampire:
Ok, here a new version, how do you like this one?
You now have to explicitly enable the new behavior.

swankjesse:
Could you motivate this problem more? Like, how do we instruct app developers when to set this flag?

I have a strong preference for leaving this to a standalone library that does a SOCKS4-friendly SocketFactory.

swankjesse:
More background on me being grumpy. I just read a rant about software size & complexity.

http://tonsky.me/blog/disenchantment/

Our job with OkHttp is to be very small and ship what satisfies 95% of the use cases. We leave the ugliest 5% of use cases out on purpose. If you're unlucky enough to need a SOCKS4 proxy in 2018 I don't really want to burden the rest of us with your problems! 

But also – if I'm dumb and misjudging how common this problem is, I want you to open my eyes!

Vampire:
> Could you motivate this problem more? Like, how do we instruct app developers when to set this flag?

Well, I thought I made it pretty clear in the JavaDoc of the `tryToResolveForSocksProxies` method: `You need to set this option to {@code true} if you either want to use a SOCKS5 proxy that cannot resolve the hostname but you can locally, or if you want to use a SOCKS4 proxy.`

> I have a strong preference for leaving this to a standalone library that does a SOCKS4-friendly SocketFactory.

It is not only the SOCKS4 that is not working,
but also the SOCKS5 if the proxy cannot resolve the hostname, but you could locally.
Currently you will just get an error response if the proxy cannot resolve the hostname.

> If you're unlucky enough to need a SOCKS4 proxy in 2018 I don't really want to burden the rest of us with your problems!

As stated above, it is not only the SOCKS4 that is not working, but also SOCKS5 where the proxy cannot resolve the hostname while the client could.

> But also – if I'm dumb and misjudging how common this problem is, I want you to open my eyes!

Well, I don't have any statistics on where SOCKS4 is still used, but it is a rather easy fix and small change as Java 9 does the rest already.
And I also have no statistics on SOCKS5 servers that cannot resolve the hostname as any installation of any SOCKS5 server could suffer from this problem.

> Could you get the JDK maintainers to fix this? They know which SOCKS version they’re talking to and can fix it for everyone in one step.

I don't think it is a JRE problem, especially the SOCKS5 one.
For SOCKS4 they could maybe try to resolve the address themselves, but actually then the `Dns` you set on the `OkHttpClient` is ignored as the JRE of course cannot use it.
With my patch the configured `Dns` would be used to resolve the hostname which is what I would expect as user.
Both for SOCKS4 and for SOCKS5 when the proxy cannot resolve the address.

swankjesse:
_You need to set this option to {@code true} if you either want to use a SOCKS5 proxy that cannot resolve the hostname but you can locally, or if you want to use a SOCKS4 proxy._

But how does the application developer know this? I don’t think how the SOCKS proxy is configured is the application’s concern; instead it’s the concern of the host OS.

What happens today if I’ve configured my OS to use a SOCKS4 proxy? I assume everything is broken, isn’t it? Like, does HttpURLConnection or Apache HTTP client work? I think the right solution to this problem isn’t to punt complexity to the application layer, but instead to get the user to get rid of their broken proxy.

Vampire:
You don't configure your OS to use a proxy, you configure Java to use a proxy.

And why are you keep just talking about SOCKS4?
SOCKS5 is also not working if you have a proxy server that is not able to resolve hostnames, which is a perfectly valid use-case.

An application developer will set this option if he knows that "leaking" names to the DNS server is not a problem.
So basically if someone develops a library that is possibly used for something like Tor, he will either not set this option or provide an option to set this option.
If someone develops a library that is unlikely to be used for something like Tor, he can set this option if he wants to support SOCKS5 proxies that cannot resolve hostnames and SOCKS4 proxies.
If someone develops an application that is not something like Tor, he can set this option if he wants to support SOCKS5 proxies that cannot resolve hostnames and SOCKS4 proxies.

So for example for Javacord which is a discord api wrapper library to build discord bots, I would set this option to true, as it is ok if the discord hostnames "leak" to the DNS server, as probably in almost any application besides a Tor client.


> Like, does HttpURLConnection or Apache HTTP client work?

Why shouldn't they?
The JRE supports SOCKS4 and SOCKS5 with resolved hostnames.
The problem is **not** the JRE.
The problem is that OkHttp explicitly creates an unresolved `InetSocketAddress` and thereby actively prevents using SOCKS4 or SOCKS5 with resolved hostnames.

So from a users perspective this is either a bug or an intentional rudeness in OkHttp.

Actually you admitted this yourself.
You actively prevent IP resolving to not "leak" the hostname to the DNS.
It is ok if you do something like that if you think this is a security issue, but only if this behavior can be configure, which my PR adds.
And the added logic is not really that fancy.

> I think the right solution to this problem isn’t to punt complexity to the application layer, but instead to get the user to get rid of their broken proxy.

The proxy is not broken, both, SOCKS4 and SOCKS5 without access to DNS are perfectly valid use-cases.

ramsestom:
> Honest question: who uses SOCKS4? Shouldn’t we just encourage these people to get SOCKS5 or JDK9? I don’t want to add code or complexity to OkHttp unless it has a big positive impact.

Many free proxies are still Socks4 proxies (they where probably setup into universities or companies years ago and never modified since then)


> _You need to set this option to {@code true} if you either want to use a SOCKS5 proxy that cannot resolve the hostname but you can locally, or if you want to use a SOCKS4 proxy._
> 
> But how does the application developer know this? I don’t think how the SOCKS proxy is configured is the application’s concern; instead it’s the concern of the host OS.

I really don't understand why this PR, that address a serious issue, has not been accepted yet. It just gives developpers an option to fix issues with some SOCKS4 and SOCKS5 proxies that wont work without it. And as stated, it is an **option** that needs to be set to true to be turned on. So it won't change the default okHttp behaviour nor break backward compatibility for developpers.  So why block a PR that is so usefull and just add up less than 50 lines of code?

yschimke:
Potentially relevant 

https://daniel.haxx.se/blog/2019/04/13/test-servers-for-curl/

swankjesse:
@ramsestom does this work for you?
https://gist.github.com/swankjesse/132ecc49bee853c3c335d33b9a95bef7

What are you using SOCKS4 for?

ramsestom:
I am using Socks4 in a proxyrotator that rely on free proxy lists (where many Socks proxies are still Socks4).
Did not test your hack of the SocketFactory but anyway it do not solve all issues you can have with valid Socks4 or Socks5 proxies (whereas Vampire's PR does so I implemented his PR in a fork of Okhttp for now and I don't see any valid argument for not accepting this PR officially)

PR #4269, opened 2018-09-14
Fix okHttp for RoboVM (and other Android-based runtimes)

### LINKED ISSUE #4266: 3.11.0 breaks on RoboVM
OkHttp 3.11.0 breaks on [RoboVM](http://robovm.mobidevelop.com/).

RoboVM is an AOT compiler for iOS. It uses a runtime class library that is "derived from Android" (i.e. based on Apache Harmony), but without any Android specific packages.

Because RoboVM's runtime class library is based on Apache Harmony, all the checks in AndroidPlatform.buildIfSupported() (which basically check for classes that are present in Android, but which are not actually Android-specific) pass, so OkHttp decides it is running on Android.

However, there are problems later when AndroidPlatform.getSSLContext() uses the Android-specific android.os.Build.VERSION class to decide whether TLS 1.2 is available.

This problem is only present in 3.11.0, as it was introduced by [this commit]( https://github.com/square/okhttp/commit/bdbe2dad8806a77bf018b87f4518df250fa67f89).

Here's the relevant RoboVM issue: https://github.com/MobiVM/robovm/issues/325

guillerodriguez:
Fixes #4266.

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/platform/AndroidPlatform.java
+++ b/okhttp/src/main/java/okhttp3/internal/platform/AndroidPlatform.java
@@ -431,7 +431,7 @@ class AndroidPlatform extends Platform {
   }
 
   @Override public SSLContext getSSLContext() {
-    boolean tryTls12 = true;
+    boolean tryTls12;
     try {
       tryTls12 = (Build.VERSION.SDK_INT >= 16 && Build.VERSION.SDK_INT < 22);
     } catch (NoClassDefFoundError e) {
--- a/okhttp/src/main/java/okhttp3/internal/platform/AndroidPlatform.java
+++ b/okhttp/src/main/java/okhttp3/internal/platform/AndroidPlatform.java
@@ -431,7 +431,16 @@ class AndroidPlatform extends Platform {
   }
 
   @Override public SSLContext getSSLContext() {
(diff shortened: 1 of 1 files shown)


okhttp/src/main/java/okhttp3/internal/platform/AndroidPlatform.java
```diff
@@ -431,7 +431,16 @@ public int hashCode() {
...
 
   @Override public SSLContext getSSLContext() {
-    if (Build.VERSION.SDK_INT >= 16 && Build.VERSION.SDK_INT < 22) {
+    boolean tryTls12 = true;
+    try {
+      tryTls12 = (Build.VERSION.SDK_INT >= 16 && Build.VERSION.SDK_INT < 22);
+    } catch (NoClassDefFoundError e) {
+      // Not a real Android runtime; probably RoboVM or MoE
+      // Try to load TLS 1.2 explicitly.
+      tryTls12 = true;
```
PromanSEW: This line can be removed

guillerodriguez: Ops, right. I actually just removed the initial assignment to make the intention more explicit.

Issue #4229, opened 2018-08-22
http/1.1, code=307, message=Temporary Redirect


vinaykrish505:
Response{protocol=http/1.1, code=307, message=Temporary Redirect, url=https://xxxxxxxxx/xxxxxx/}

Android studio with 
retrofit2 and okhttp3 llibrary latest versions 2.4.0 & 3.11.0 respectively. Getting the issue that redirecting. Previously Httpclient it is working but here in retrofit we dont have solution for it. 
please provide update as early as possible.

swankjesse:
Executable test case? Just want to be clear where OkHttp’s behavior is incorrect.

vinaykrish505:
Hello Team,

In one of my android projects, I used to work with the retrofit library and
Okhhtp client. In post request I'm getting response code 307 temporary url
redirection. But the same request in iOS is normally getting the
response from the server. So I digged for the solution and in the process,
I found that Okhttp is not allowing the 307(301 & 302 also) response to
load directly.

I'm requesting you guys that please enable the 307 response code to
redirect directly.




*Thanks & Regards*
 M. Vinay Krishna
 Mob : +91-9440170923

swankjesse:
Does calling OkHttpClient.Builder.followRedirects(false) help?

vinaykrish505:
Yeah, @Jesse I tried with followRedirects(true) but it is not worked for
me.




*Thanks & Regards*
 M. Vinay Krishna
 Mob : +91-9440170923

yschimke:
This works for my test client

```
$ okurl 'https://httpbin.org/redirect-to?status_code=307&url=https://api.twitter.com/robots.txt'
# Used for Google app indexing. See https://developers.google.com/app-indexing/webmasters/server
User-agent: Googlebot
Disallow:

User-agent: *
Disallow: /
```

swankjesse:
Unclear what action to take here.

Issue #3111, opened 2017-01-16
why not support 307 post REDIRECT


caopeng000:
- [ ] Feature Request. Start by telling us what problem you’re trying to solve. Often a solution
       already exists! Don’t send pull requests to implement new features without first getting our
       support. Sometimes we leave features out on purpose to keep the project small.

why not support 307 REDIRECT

case HTTP_TEMP_REDIRECT:
        // "If the 307 or 308 status code is received in response to a request other than GET
        // or HEAD, the user agent MUST NOT automatically redirect the request"
        if (!method.equals("GET") && !method.equals("HEAD")) {
          return null;
        }

Doing so can do
case HTTP_TEMP_REDIRECT:
   // Does the client allow redirects?
        if (!client.followRedirects()) return null;

        String location = userResponse.header("Location");
        if (location == null) return null;
        HttpUrl url = userRequest.url().resolve(location);

        // Don't follow redirects to unsupported protocols.
        if (url == null) return null;

        // If configured, don't follow redirects between SSL and non-SSL.
        boolean sameScheme = url.scheme().equals(userRequest.url().scheme());
        if (!sameScheme && !client.followSslRedirects()) return null;

        // Redirects don't include a request body.
        Request.Builder requestBuilder = userRequest.newBuilder();
        if (HttpMethod.permitsRequestBody(method)) {
          if (HttpMethod.redirectsToGet(method)) {
            requestBuilder.method("GET", null);
          } else {
            requestBuilder.method(method, null);
          }
          requestBuilder.removeHeader("Transfer-Encoding");
          requestBuilder.removeHeader("Content-Length");
          requestBuilder.removeHeader("Content-Type");
        }

        // When redirecting across hosts, drop all authentication headers. This
        // is potentially annoying to the application layer since they have no
        // way to retain them.
        if (!sameConnection(url)) {
          requestBuilder.removeHeader("Authorization");
        }

        return requestBuilder.url(url).build();

electrum:
The current behavior (and comment in the code) seems at odds with [RFC 7231 section 6.4.7](https://tools.ietf.org/html/rfc7231#section-6.4.7):

> The 307 (Temporary Redirect) status code indicates that ... the user agent MUST NOT change the request method if it performs an automatic redirection to that URI. The user agent MAY use the Location field value for automatic redirection.

electrum:
This is our current workaround (not great):
```java
try (Response response = client.newCall(request).execute()) {
[...]

Issue #2874, opened 2016-09-20
Reconsider which methods & status codes redirect


swankjesse:
From https://tools.ietf.org/html/rfc7231#appendix-B

```
   The set of request methods that are safe to automatically redirect is
   no longer closed; user agents are able to make that determination
   based upon the request method semantics.  The redirect status codes
   301, 302, and 307 no longer have normative requirements on response
   payloads and user interaction.  (Section 6.4)
```

See also https://github.com/square/okhttp/pull/944

swankjesse:
Stalled.

PR #4270, opened 2018-09-14
Complies with rfc7231 about statuses 307 and 308

vkravets:
Fix for #2874 #3111 and #4229

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/URLConnectionTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/URLConnectionTest.java
@@ -2395,19 +2395,9 @@ public final class URLConnectionTest {
       assertEquals("Page 2", response);
     } else if (method.equals("HEAD")) {
       assertEquals("", response);
-    } else {
-      // Methods other than GET/HEAD shouldn't follow the redirect
-      if (method.equals("POST")) {
-        assertTrue(connection.getDoOutput());
-        assertEquals("ABCD", page1.getBody().readUtf8());
-      }
-      assertEquals(1, server.getRequestCount());
-      assertEquals("This page has moved!", response);
-      return;
     }
 
-    // GET/HEAD requests should have followed the redirect with the same method
-    assertFalse(connection.getDoOutput());
+    assertFalse(method.equals("POST") && !connection.getDoOutput());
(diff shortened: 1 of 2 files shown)


swankjesse:
This is a straightforward change with very dramatic consequences. I’m agree that we should do this, but should be very cautious when rolling it out. 

One option is to give users a sample network interceptor that restores the previous behavior. It would probably just strip the Location header or something to prevent the redirect.

vkravets:
I'm not sure that intercepter will good idea for that... I want to propose solution with flag in the same way what is done for redirects, by default it will be false and old behavior will be working.... Otherwise new behavior of redirection for 307 and 308 will be used.

Main question how name such flag...
1. `setFollowAutomaticRedirect()` and `isFollowAutomaticRedirect()`
2. also we can check system variable, default behavior - old, if system variable specified - use new behavior...

As for me example of intercepter will more confuse end user instead of set the understanding of what this intercepter performs.

yschimke:
@swankjesse I don't like optional network interceptors if we think there is a better default.  Most users won't know to add the interceptor.  Should we just make the change as an opinionated user agent and have a way to disable to previous behaviour?

swankjesse:
Let's fix this and offer an interceptor to restore the previous behavior.

yschimke:
I'll take a look at the alternate fix unless @vkravets is interested in making the changes?

yschimke:
Realised since I intended to migrate the changes to Kotlin, it's worth keeping this open for now.

yschimke:
I'll take a look at this hopefully in next week.

yschimke:
@swankjesse How do you make big changes on a PR like this while retaining the author?

swankjesse:
@yschimke either `commit --amend` or use the git commit option that takes author metadata.

CoryARamirez:
@swankjesse is there an ETA as to when this will be complete?

swankjesse:
🔜

yschimke:
Also possibly relevant https://github.com/square/okhttp/issues/3111

yschimke:
Migrated current logic to kotlin in https://github.com/square/okhttp/pull/5990, with author retained

PR #4271, opened 2018-09-16
Refactor String#replaceAll - keep the compiled pattern

durigon:
If we repeatedly call String#replaceAll, we internally repeatedly call the regular expression pattern compilation every time as following:
```java
    public String replaceAll(String regex, String replacement) {
        return Pattern.compile(regex).matcher(this).replaceAll(replacement);
    }
```
The modifications are to keep the compiled pattern.
Therefore, compiling a relatively expensive regular expression pattern does not have to be done every time.

### DIFF
--- a/okhttp/src/main/java/okhttp3/HttpUrl.java
+++ b/okhttp/src/main/java/okhttp3/HttpUrl.java
@@ -26,6 +26,7 @@ import java.util.Collections;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.regex.Pattern;
 import javax.annotation.Nullable;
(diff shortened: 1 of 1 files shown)


JakeWharton:
This code path is meant to be rarely taken which is why we don't eagerly force all users to pay the cost of Pattern creation. Is this codepath being taken frequently in your codebase?

durigon:
@JakeWharton 

> This code path is meant to be rarely taken which is why we don't eagerly force all users to pay the cost of Pattern creation.
I understood your intentions. It seems that it is not good to orce all users to pay the cost of Pattern creation.

durigon:
I close this PR because it is an unlikely edge case.

PR #4272, opened 2018-09-17
add methods overload support

shellhub:
default `TimeUnit.MILLISECONDS` value can save coding time

### DIFF
--- a/okhttp/pom.xml
+++ b/okhttp/pom.xml
@@ -32,6 +32,12 @@
       <artifactId>jsr305</artifactId>
       <scope>provided</scope>
     </dependency>
+    <dependency>
+      <groupId>org.codehaus.mojo</groupId>
+      <artifactId>animal-sniffer-annotations</artifactId>
+      <version>1.17</version>
+      <scope>provided</scope>
+    </dependency>
   </dependencies>
 
   <build>
--- a/okhttp/src/main/java/okhttp3/OkHttpClient.java
+++ b/okhttp/src/main/java/okhttp3/OkHttpClient.java
@@ -19,6 +19,7 @@ import java.net.Proxy;
 import java.net.ProxySelector;
 import java.net.Socket;
 import java.security.GeneralSecurityException;
+import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
@@ -46,6 +47,7 @@ import okhttp3.internal.tls.OkHostnameVerifier;
 import okhttp3.internal.ws.RealWebSocket;
 import okio.Sink;
 import okio.Source;
+import org.codehaus.mojo.animal_sniffer.IgnoreJRERequirement;
 
 import static okhttp3.internal.Util.assertionError;
 import static okhttp3.internal.Util.checkDuration;
@@ -525,6 +527,20 @@ public class OkHttpClient implements Cloneable, Call.Factory, WebSocket.Factory 
     }
 
     /**
+     * Sets the default connect timeout for new connections. A value of 0 means no timeout,
+     * otherwise values must be between 1 and {@link Integer#MAX_VALUE} when converted to
+     * milliseconds.
+     *
+     * <p>The connectTimeout is applied when connecting a TCP socket to the target host.
+     * The default value is 10 seconds.
+     */
+    @IgnoreJRERequirement
+    public Builder connectTimeout(Duration duration) {
+      connectTimeout = checkDuration("timeout", duration.toMillis(), TimeUnit.MILLISECONDS);
+      return this;
+    }
+
+    /**
      * Sets the default read timeout for new connections. A value of 0 means no timeout, otherwise
      * values must be between 1 and {@link Integer#MAX_VALUE} when converted to milliseconds.
      *
@@ -536,6 +552,22 @@ public class OkHttpClient implements Cloneable, Call.Factory, WebSocket.Factory 
      */
     public Builder readTimeout(long timeout, TimeUnit unit) {
       readTimeout = checkDuration("timeout", timeout, unit);
+      return this;
+    }
+
+    /**
+     * Sets the default read timeout for new connections. A value of 0 means no timeout, otherwise
+     * values must be between 1 and {@link Integer#MAX_VALUE} when converted to milliseconds.
+     *
+     * <p>The read timeout is applied to both the TCP socket and for individual read IO operations
+     * including on {@link Source} of the {@link Response}. The default value is 10 seconds.
+     *
+     * @see Socket#setSoTimeout(int)
+     * @see Source#timeout()
+     */
+    @IgnoreJRERequirement
+    public Builder readTimeout(Duration duration) {
+      readTimeout = checkDuration("timeout", duration.toMillis(), TimeUnit.MILLISECONDS);
       return this;
     }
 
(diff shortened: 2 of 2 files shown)


yschimke:
@JakeWharton is that safe?

```
 * @since 1.8
 */
public final class Duration
```

JakeWharton:
Yep. You can only call it from Java 8.

shellhub:
@JakeWharton your comment is very useful to me. But many java function like `URLConnection.setConnectTimeout` taking parameter in milliseconds though implicit. What do you think?

JakeWharton:
I think those are bad APIs and that we should avoid continuing to make
their mistakes

shellhub:
@JakeWharton @yschimke I overload few methods  taking Duration though. It's ok?

okhttp/src/main/java/okhttp3/internal/Util.java
```diff
@@ -498,6 +499,15 @@ public static int checkDuration(String name, long duration, TimeUnit unit) {
     return (int) millis;
   }
 
+  public static int checkDuration(String name, Duration duration) {
+    if (duration == null) throw new NullPointerException("duration == null.");
+    long millis = duration.toMillis();
+    if (millis < 0) throw new IllegalArgumentException(name + " < 0");
+    if (millis > Integer.MAX_VALUE) throw new IllegalArgumentException(name + " too large.");
+    if (millis == 0) throw new IllegalArgumentException(name + " too small.");
```
yschimke: Doesn't this conflict with the javadocs for 0?

yschimke:
This would need tests, although might be best to resolve with @swankjesse 

my 2c, I think the accessors are less interesting for clients, and there is value in configuring using the typed methods.  So I'm happy to see this land.  Still surprised if this will compile given pom

    <java.version>1.7</java.version> 

and travis shows

[ERROR] /home/travis/build/square/okhttp/okhttp/src/main/java/okhttp3/internal/Util.java:504: Undefined reference: long java.time.Duration.toMillis()

okhttp/src/main/java/okhttp3/internal/Util.java
shellhub: no conflict :)

yschimke: > A value of 0 means no timeout, otherwise values must be between 1 and {@link Integer#MAX_VALUE} when converted to milliseconds.

shellhub: Yes, if call `Duration.ofXXX(0)` method, I think request won't success because of limited Broadband. So` throw new IllegalArgumentException(name + " too small.");`

shellhub:
@yschimke @JakeWharton @swankjesse I update `pom.xml`. Build successfully

yschimke:
@JakeWharton are you sure the compile with java 1.8 is safe?

JakeWharton:
You shouldn't be targeting 1.8. We've used JDK 8 forever which is enough to get the type onto the classpath.

shellhub:
> You shouldn't be targeting 1.8. We've used JDK 8 forever which is enough to get the type onto the classpath.

should I delete this? `IntellJ IDEA` report error if I delete. @JakeWharton . `Maven` can build success
``` xml
      <plugin>
        <artifactId>maven-compiler-plugin</artifactId>
        <configuration>
          <source>8</source>
          <target>8</target>
        </configuration>
      </plugin>
```

JakeWharton:
And the animal sniffer change, yes. It should just be a suppression.

shellhub:
I have deleted. Please review animal sniffer again！ Give some example code about animal sniffer if you can.Thank

shellhub:
should I close this `PR`? @yschimke  @JakeWharton

okhttp/pom.xml
```diff
@@ -48,6 +48,18 @@
...
+          <artifactId>animal-sniffer-maven-plugin</artifactId>
+          <version>1.17</version>
+          <configuration>
+              <signature>
+                  <groupId>org.codehaus.mojo.signature</groupId>
+                  <artifactId>java18</artifactId>
+                  <version>1.0</version>
+              </signature>
+          </configuration>
+      </plugin>
```
JakeWharton: This needs removed

okhttp/src/main/java/okhttp3/OkHttpClient.java
```diff
@@ -524,6 +525,19 @@ public Builder connectTimeout(long timeout, TimeUnit unit) {
...
 
+    /**
+     * Sets the default connect timeout for new connections. A value of 0 means no timeout,
+     * otherwise values must be between 1 and {@link Integer#MAX_VALUE} when converted to
+     * milliseconds.
+     *
+     * <p>The connectTimeout is applied when connecting a TCP socket to the target host.
+     * The default value is 10 seconds.
+     */
+    public Builder connectTimeout(Duration duration) {
```
JakeWharton: Should be annotated with `@IgnoreJRERequirement`

okhttp/src/main/java/okhttp3/OkHttpClient.java
```diff
@@ -539,6 +553,21 @@ public Builder readTimeout(long timeout, TimeUnit unit) {
...
+     * Sets the default read timeout for new connections. A value of 0 means no timeout, otherwise
+     * values must be between 1 and {@link Integer#MAX_VALUE} when converted to milliseconds.
+     *
+     * <p>The read timeout is applied to both the TCP socket and for individual read IO operations
+     * including on {@link Source} of the {@link Response}. The default value is 10 seconds.
+     *
+     * @see Socket#setSoTimeout(int)
+     * @see Source#timeout()
+     */
+    public Builder readTimeout(Duration duration) {
```
JakeWharton: `@IgnoreJRERequirement`

okhttp/src/main/java/okhttp3/OkHttpClient.java
```diff
@@ -553,6 +582,20 @@ public Builder writeTimeout(long timeout, TimeUnit unit) {
...
+    /**
+     * Sets the default write timeout for new connections. A value of 0 means no timeout, otherwise
+     * values must be between 1 and {@link Integer#MAX_VALUE} when converted to milliseconds.
+     *
+     * <p>The write timeout is applied for individual write IO operations.
+     * The default value is 10 seconds.
+     *
+     * @see Sink#timeout()
+     */
+    public Builder writeTimeout(Duration duration) {
```
JakeWharton: `@IgnoreJRERequirement`

okhttp/src/main/java/okhttp3/internal/Util.java
```diff
@@ -498,6 +499,14 @@ public static int checkDuration(String name, long duration, TimeUnit unit) {
     return (int) millis;
   }
 
+  public static int checkDuration(String name, Duration duration) {
```
JakeWharton: I don't think we need this method. Have the duration overloads call `sameMethod(duration.toMillis(), TimeUnit.MILLISECONDS)` so that they're just JDK 8 sugar on the JDK 6 methods.

okhttp/src/main/java/okhttp3/OkHttpClient.java
```diff
@@ -571,6 +614,24 @@ public Builder pingInterval(long interval, TimeUnit unit) {
...
+     *
+     * <p>If the server does not respond to each ping with a pong within {@code interval}, this
+     * client will assume that connectivity has been lost. When this happens on a web socket the
+     * connection is canceled and its listener is {@linkplain WebSocketListener#onFailure notified
+     * of the failure}. When it happens on an HTTP/2 connection the connection is closed and any
+     * calls it is carrying {@linkplain java.io.IOException will fail with an IOException}.
+     *
+     * <p>The default value of 0 disables client-initiated pings.
+     */
+    public Builder pingInterval(Duration duration) {
```
JakeWharton: `@IgnoreJRERequirement`

shellhub:
@JakeWharton IntelliJ IDEA & Maven Build fail.
> Cannot resolve symbol 'IgnoreJRERequirement'
image

JakeWharton:
We might not have the dependency: org.codehaus.mojo:animal-sniffer-annotations

okhttp/src/main/java/okhttp3/internal/Util.java
shellhub: Do you mean remove this method
``` java
  public static int checkDuration(String name, Duration duration) {
    if (duration == null) throw new NullPointerException("duration == null.");
    long millis = duration.toMillis();
    if (millis < 0) throw new IllegalArgumentException(name + " < 0");
    if (millis > Integer.MAX_VALUE) throw new IllegalArgumentException(name + " too large.");
    return (int) millis;
  }
```
and call like this
``` java
    @IgnoreJRERequirement
    public Builder writeTimeout(Duration duration) {
//      writeTimeout = checkDuration("timeout", duration);
      writeTimeout = checkDuration("timemeout", duration.toMillis(), TimeUnit.MILLISECONDS);
      return this;
    }
```

JakeWharton: Precisely

shellhub:
@JakeWharton @yschimke Updated. Please review again :)

JakeWharton:
Thank you!

PR #4280, opened 2018-09-21
feat: add request date in logs

mgohin:
If call fails with an exception, there is no date to get proper traceability


mgohin:
Oups, gonna fix Unit tests

PR #4286, opened 2018-09-23
Non Null return types

yschimke:
For discussion, this avoids a warning in kotlin clients IIUC.  The existing parameter solution doesn't work here.

```
@javax.annotation.ParametersAreNonnullByDefault
package okhttp3;
```

Advice?


swankjesse:
We need something like this:
https://stackoverflow.com/questions/7658353/is-parametersarenonnullbydefault-applies-to-method-return-values-too

yschimke:
Until we add that are you ok with adding these annotations? Or just visual noise?  It's really awkward on the kotlin side

swankjesse:
I think I’m gonna need an annotation like `@okhttp3.internal.EverythingIsNonNull` in each of our open source projects, starting with this one. It’s a small bit of boilerplate but definitely worth it.

yschimke:
How does that hook in with IDE and kotlin null checking support?

swankjesse:
Something like this:
https://github.com/uber-common/javax-extras/blob/master/javax-extras/src/main/java/com/uber/javaxextras/EverythingIsNonNullByDefault.java

okhttp-tls/src/main/java/okhttp3/tls/HeldCertificate.java
```diff
@@ -33,6 +33,7 @@
 import java.util.UUID;
 import java.util.concurrent.TimeUnit;
 import javax.security.auth.x500.X500Principal;
+import jdk.internal.jline.internal.Nullable;
```
swankjesse: Wrong import.

okhttp-tls/src/main/java/okhttp3/tls/HandshakeCertificates.java
```diff
@@ -28,6 +28,7 @@
 import javax.net.ssl.TrustManager;
 import javax.net.ssl.X509KeyManager;
 import javax.net.ssl.X509TrustManager;
+import jdk.internal.jline.internal.Nullable;
```
swankjesse: Wrong import.

okhttp-tls/src/main/java/okhttp3/tls/package-info.java
```diff
@@ -0,0 +1,5 @@
+/** An HTTP+HTTP/2 client for Android and Java applications. */
```
swankjesse: Omit this comment?

okhttp-urlconnection/src/main/java/okhttp3/package-info.java
```diff
@@ -0,0 +1,5 @@
+/** An HTTP+HTTP/2 client for Android and Java applications. */
```
swankjesse: Omit this comment.

okhttp/src/main/java/okhttp3/internal/annotations/EverythingIsNonNull.java
```diff
@@ -0,0 +1,41 @@
...
+ * local variables and fields.
+ *
+ * @see javax.annotation.ParametersAreNonnullByDefault
+ */
+@Documented
+@Nonnull
+@TypeQualifierDefault({
+    ElementType.FIELD,
+    ElementType.METHOD,
+    ElementType.LOCAL_VARIABLE,
```
swankjesse: I don't want to have to annotate local variables. I don't think they're worth locking down cause the nullability inferencer is usually sufficient.

okhttp/src/main/java/okhttp3/internal/annotations/EverythingIsNonNull.java
```diff
@@ -0,0 +1,41 @@
...
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import javax.annotation.Nonnull;
+import javax.annotation.meta.TypeQualifierDefault;
+
+/**
+ * Extends {@code ParametersAreNonnullByDefault} to also apply to Method results,
```
swankjesse: This is great.

okhttp/src/main/java/okhttp3/package-info.java
```diff
@@ -1,3 +1,5 @@
 /** An HTTP+HTTP/2 client for Android and Java applications. */
-@javax.annotation.ParametersAreNonnullByDefault
+@EverythingIsNonNull
```
PromanSEW: Use full name instead of import?

okhttp-tls/src/main/java/okhttp3/tls/HeldCertificate.java
yschimke: Thanks, should have picked up in my self review.  I was using Intellij inspections to flag the Nullability issues.

yschimke:
I'm not in a rush on this, but seems like a worthy thing for 3.12.  I'll fix issues and test a snapshot with a kotlin project.

yschimke:
So far no issues and some nice cleanups by using. e.g. mixed in with other upgrades.

https://github.com/yschimke/okurl/commit/e9563e89162cb3621b70b45690dbb9459e123543

swankjesse:
I'm gonna do likewise in our other projects!

okhttp-logging-interceptor/src/main/java/okhttp3/logging/package-info.java
```diff
@@ -1,3 +1,5 @@
 /** An OkHttp interceptor which logs HTTP request and response data. */
-@javax.annotation.ParametersAreNonnullByDefault
+@EverythingIsNonNull
```
yschimke: I'll fix these before landing.

Issue #3826, opened 2018-02-05
HttpLoggingInterceptor leaks Authorization Header


mnembrini:
Feature Request/Bug:

When using Level.HEADERS or Level.BODY all HTTP headers of request and response are printed. 
If someone is using these for e.g. tests running on a CI server, the logs will leak authorization tokens. 

There is a warning in the README about not using this in production, but I think it would be very easy to redact values of sensitive headers (I think "Authorization" and "Proxy-Authorization" are the only ones?)

I have already working code for this, but I wanted to know if this is something you would like to see upstreamed and how it should be configured: e.g. by adding more levels or adding a boolean *redactSensitiveHeaders* to the constructor.

JakeWharton:
A boolean isn't a very future-proof solution. We could accept a list. Or
add a builder where you could add individual headers.

PR #4287, opened 2018-09-23
Support header redaction in HttpLoggingInterceptor

amirlivneh:
This implements the enhancement discussed in https://github.com/square/okhttp/issues/3826.

### DIFF
--- a/okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
+++ b/okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
@@ -125,8 +125,9 @@ public final class HttpLoggingInterceptor implements Interceptor {
   }
 
   public static class Builder {
-    private Set<String> headersToRedact = new LinkedHashSet();
     private Logger logger = Logger.DEFAULT;
+    private Set<String> headersToRedact = new LinkedHashSet();
+    private Level level = Level.NONE;
 
     public Builder logger(Logger logger) {
       this.logger = logger;
@@ -138,6 +139,11 @@ public final class HttpLoggingInterceptor implements Interceptor {
       return this;
     }
 
+    public Builder level(Level level) {
+      this.level = level;
+      return this;
+    }
+
     public HttpLoggingInterceptor build() {
       return new HttpLoggingInterceptor(this);
     }
@@ -146,6 +152,7 @@ public final class HttpLoggingInterceptor implements Interceptor {
   public HttpLoggingInterceptor(Builder builder) {
     this.headersToRedact = builder.headersToRedact;
     this.logger = builder.logger;
+    this.level = builder.level;
   }
 
   private final Logger logger;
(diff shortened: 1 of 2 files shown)


amirlivneh:
There is a discussion on the use case and solution in https://github.com/square/okhttp/issues/3826. It looks like you added this enhancement to the Icebox milestone.

okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
```diff
@@ -119,10 +121,37 @@ public HttpLoggingInterceptor() {
 
   public HttpLoggingInterceptor(Logger logger) {
     this.logger = logger;
+    this.headersToRedact = new LinkedHashSet();
+  }
+
+  public static class Builder {
```
yschimke: Can you add a setter for level.

okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
```diff
@@ -285,6 +314,12 @@ public Level getLevel() {
     return response;
   }
 
+  private void logHeader(Headers headers, int i) {
+    String value =
+        headersToRedact.contains(headers.name(i).toLowerCase()) ? "<redacted>" : headers.value(i);
```
yschimke: Warning from the Ghost of Code Review Yet-to-Come:

https://github.com/square/okhttp/pull/4252#discussion_r216175280

yschimke:
@swankjesse I'm used to dev builds having the developers own credentials, so this makes sense to me.  It certainly doesn't seem like a bad thing and seems inline with other APIs we have.

okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
```diff
@@ -118,11 +122,19 @@ public HttpLoggingInterceptor() {
   }
 
   public HttpLoggingInterceptor(Logger logger) {
+    this(logger, Collections.<String>emptyList());
+  }
+
+  public HttpLoggingInterceptor(Logger logger, List<String> headersToRedact) {
```
swankjesse: Do you mind replacing this constructor with a method that adds a header to redact? I'm allergic to telescoping constructor parameters.

amirlivneh: I'm allergic to them either, but avoided such a method so that `headersToRedact` doesn't have to become synchronized. Do you prefer a synchronized set or do you think one is not necessary?

swankjesse: We could have our cake and eat it too.

```
private volatile Set<String> headersToRedact = Collections.emptySet();

public void redactHeader(String name) {
  Set<String> newHeadersToRedact = new TreeSet<>(CASE_INSENSITIVE_ORDER);
  newHeadersToRedact.addAll(headersToRedact);
  newHeadersToRedact.add(name);
  this.headersToRedact = newHeadersToRedact;
}
```

swankjesse:
# 💯

PR #4290, opened 2018-09-24
Remove unused okhttp3.internal.cache2.Relay

amirlivneh:
I have no context on this. I'm only creating this PR since Redex flagged this internal code as unreachable, and it hasn't been touched in over 2 years.

Feel free to ignore this if the code is kept intentionally.


swankjesse:
It's a work in progress!

amirlivneh:
Cool, I'll close it then.

yschimke:
is it better deleted and re-added as an open PR?  How dies this compare to okio BufferedSource.peek for the intended uses?

swankjesse:
Peek is for a single thread to read twice. Relay is for any number of threads to each read once.

PR #4291, opened 2018-09-27
Fix exception message

StefMa:
Just a minor change.
The param name is `file` - not content.
This may leads to confusing when observed in logs.

Another solution - to match with the other methods - would be to change the param name to `content` 🤷‍♂️ 
Please give me a hint when we should do it like this 👍


swankjesse:
Thanks!

PR #4292, opened 2018-09-27
Contribution Guidelines

yschimke:


yschimke:
n.b. I made changes to the Wiki before I realised this was also here.  I'll make any further changes to both places.

Issue #4283, opened 2018-09-22
Headers.of() doesn’t enforce ASCII-only


swankjesse:
Needs to call checkNameAndValue to confirm that invalid characters can’t sneak through.

swankjesse:
And maybe a proper supported mechanism to add non-ASCII headers because sometimes people need that.

amirlivneh:
I think this issue can be closed now.

PR #4296, opened 2018-09-29
A mechanism for adding non-ASCII headers

amirlivneh:
As suggested in https://github.com/square/okhttp/issues/4283#issuecomment-423703475.

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/HeadersTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/HeadersTest.java
@@ -353,6 +353,36 @@ public final class HeadersTest {
     }
   }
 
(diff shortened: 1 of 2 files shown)


swankjesse:
It's my preference that we only expose this via the builder, named `addUnsafeNonAscii()`. For this we should still require that header names pass strict validation.

swankjesse:
Thanks!

sxci:
https://github.com/square/okhttp/blob/parent-3.12.0/okhttp/src/main/java/okhttp3/MultipartBody.java#L259  
it looks like doesnot allow utf8 headers, Header.of  check strict validation.

PR #4297, opened 2018-09-29
Travis JDK 11 Experiment

yschimke:
Curious what breaks when building with JDK 11.  Also the oss parent pom is now deprecated.


swankjesse:
Something we need to look out for. SNI + TLS 1.3 + session resumption is broken in early versions of JDK 11. 
https://bugs.openjdk.java.net/browse/JDK-8211806

swankjesse:
Request peol

swankjesse:
Request people upgrade. It's fixed in 11.02.

pom.xml
```diff
@@ -290,6 +419,85 @@
...
+              <groupId>org.apache.maven.plugins</groupId>
+              <artifactId>maven-surefire-plugin</artifactId>
+              <configuration>
+                <argLine>-Xbootclasspath/p:${bootclasspathPrefix} -Xms512m -Xmx512m</argLine>
+              </configuration>
+              <dependencies>
+                <dependency>
+                  <groupId>org.mortbay.jetty.alpn</groupId>
+                  <artifactId>alpn-boot</artifactId>
+                  <version>${alpn.jdk7.version}</version>
```
yschimke: Bad merge, sorry

PR #4307, opened 2018-10-07
replacing 'ı' chars with 'i' in the value strings in 'checkValue' method

bozdemir84:
update to solve the exception 'ı' character in user agent "selenium/3.12.0 (java wındows)" in Turkish windows. 
Now Replacing 'ı' character with 'i'.


swankjesse:
This isn’t the right place for the fix. Find the code that’s calling `toLowercase()` and replace it with `toLowercase(Locale.US)`.

PR #4312, opened 2018-10-11
[WIP] Immutable Cache Control Extension Handling

StuStirling:


swankjesse:
Can you tell me where we got this wrong? My understanding is that we don't need conditions for the immutable directive.

StuStirling:
At the moment as soon as the immutable extensions is found in the `cache-control` header the request is never made to the server. After reading the [RFC 8246 spec](https://tools.ietf.org/html/rfc8246) more closely, the immutable extension should only be respected during the freshness lifetime of the response.

> The immutable extension only applies during the freshness lifetime of
   the stored response. Stale responses SHOULD be revalidated as they
   normally would be in the absence of the immutable extension.

If there is no freshness lifetime specified then it shouldn't use the cached response. My tests check that when a response that doesn't contain any `Last-Modified`, `E-Tag` or cache control directive, it doesn't use the cached response. Also the second test checks that it sends a conditional request after the response is stale.

I believe it is working correctly during the freshness lifetime at the moment, its just after it is stale it is not working and will never go to network unless you force it. My understanding is that immutable just means that if the request is made again during the response's freshness lifetime then it is an unconditional request.

swankjesse:
Right you are! Yep, we gotta fix.

swankjesse:
Fixed with https://github.com/square/okhttp/pull/4381

Issue #4267, opened 2018-09-13
Duplex Calls


swankjesse:
# Motivation

I’d like to build make it possible for gRPC clients to be built with OkHttp’s public API. The gRPC project already has an [OkHttp backend](https://github.com/grpc/grpc-java/tree/master/okhttp/src/main/java/io/grpc/okhttp), but it needs to rely on internal features.

There’s two things gRPC needs that OkHttp doesn’t offer:

 * Sending data on the request stream _after_ having received the response status and headers. This isn’t like HTTP because it isn’t request/response. It’s a bit more like WebSockets. It’s possible to do a two-way chat session with a single gRPC call.
 * Sending and receiving trailers. This is how gRPC cleanly signals a clean termination of a stream.

I’d like to build APIs on OkHttp to support these. At the moment gRPC would be the only use case for these APIs, but we wouldn’t label them as “gRPC”. The name I like best is “duplex” but I’d love to find something better.

# API

Here’s a sketch of how duplex calls would work.

### Caller Creates a Duplex Request

When you do this you can’t also specify a request body.

```
Request request = new Request.Builder()
    .url("https://api.github.com/markdown/raw")
    .duplexPost(...)
    .build();
```

The returned request has a null body, and returns `true` from a new `isDuplex()` method.

### Response Has An Outbound Stream

This class is structured like a `ResponseBody` that does blocking writes instead of blocking reads.

```
interface HttpSink extends Closeable {
  BufferedSink sink();
  void headers(Headers headers);
}
```

It’s exposed on the `ResponseBody` and you can access it after you `execute()` or in the callback to `enqueue()`.

```
try (Response response = client.newCall(request).execute()) {
  HttpSink sink = response.duplexSink();
  ...
}
```

I’m using the word `HttpSink` here instead of `DuplexRequestBody` because I think it might be possible to use the same type in MockWebServer when sending these things.

### Response Has Trailer Callbacks

There’s an interface to get callbacks when trailers arrive and a mechanism to register it on the response body.

```
interface HeadersListener {
  void onHeaders(Headers headers);
}
```

```
response.body().headersListener(new HeadersListener() {
  @Override public void onHeaders(Headers headers) {
    System.out.println(headers);
  }
});
```

I’d like to avoid having the word “trailers” in the API because the HTTP/2 wire protocol allows ’em to come anywhere in the stream. Also it would be bad to have a variable called `trailers` with a type of `Headers`.

### onHeaders Arrives On The User’s Thread

[...]

PR #4314, opened 2018-10-12
Send/Receive Headers via Http2 Frames

oldergod:
getting ready for #4267

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/internal/http2/Http2ConnectionTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/internal/http2/Http2ConnectionTest.java
@@ -18,6 +18,7 @@ package okhttp3.internal.http2;
 import java.io.IOException;
 import java.io.InterruptedIOException;
 import java.net.Socket;
+import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
@@ -33,6 +34,7 @@ import okio.BufferedSource;
 import okio.Okio;
 import okio.Sink;
 import okio.Source;
+import okio.Utf8;
 import org.junit.After;
 import org.junit.Rule;
 import org.junit.Test;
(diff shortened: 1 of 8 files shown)


okhttp-tests/src/test/java/okhttp3/internal/http2/Http2ConnectionTest.java
```diff
@@ -539,6 +541,143 @@
     assertTrue(Arrays.equals("c3po".getBytes("UTF-8"), requestData.data));
   }
 
+  static final class RecordingHeadersListener implements Header.Listener {
```
swankjesse: for follow-up we should move this into its own top-level file

swankjesse: though by that time it’ll be a `List<Headers>` not a `List<Header>`!

swankjesse:
```

[INFO] --- maven-checkstyle-plugin:2.17:checkstyle (default) @ okhttp ---
[INFO] Starting audit...
[ERROR] /home/travis/build/square/okhttp/okhttp/src/main/java/okhttp3/internal/http2/Http2Stream.java:391: Line is longer than 100 characters (found 103). [LineLength]
Audit done.
```

okhttp/src/main/java/okhttp3/internal/http2/Http2Stream.java
```diff
@@ -343,64 +357,84 @@ synchronized void receiveRstStream(ErrorCode errorCode) {
...
+              headers = null;
+
+            } else if (readBuffer.size() > 0) {
+              // Prepare to read bytes. Start by moving them to the caller's buffer.
+              readBytesDelivered = readBuffer.read(sink, Math.min(byteCount, readBuffer.size()));
+              unacknowledgedBytesRead += readBytesDelivered;
+
+              if (errorCodeToDeliver == null
+                  && unacknowledgedBytesRead
+                  >= connection.okHttpSettings.getInitialWindowSize() / 2) {
```
oldergod: that's tough indention

PR #4316, opened 2018-10-15
Added test to cover header drops with 100-continue

oldergod:
It doesn't seem to be cover in `CallTest`

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/HeadersTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/HeadersTest.java
@@ -16,6 +16,7 @@
 package okhttp3;
 
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.LinkedHashMap;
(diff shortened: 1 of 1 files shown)


okhttp-tests/src/test/java/okhttp3/HeadersTest.java
```diff
@@ -56,6 +57,23 @@
     assertEquals("HTTP/1.1", headers.value(0));
   }
 
+  @Test public void readNameValueBlockDropsHeadersUpTo100Continue() throws IOException {
```
swankjesse: Thought this test confirms that the implementation of readHttp2HeadersList does what it should, that method is a deep implementation detail that’s quite likely to get refactored in future. A more useful test would exercise the “user’s view” of these headers . . . that an actual HTTP/2 response gets the right headers.

oldergod: I think we need the implementation around `httpSink` to write this test in a more instrumental fashion.

PR #4320, opened 2018-10-16
Migrated Header to Headers

oldergod:
`Header` only lives inside `Hpack` now (because of the dynamic table thing)
Everything else is `Headers`

- [ ] Where to move the `Headers` listener? Inside Http2Stream? As itself within the `internal.okhttp2` package?
- [ ] Are we cool keeping `Header` for the dynamic table inside `Hpack`? Do we want to refactor that too?


okhttp/src/main/java/okhttp3/Headers.java
```diff
@@ -48,6 +49,15 @@
...
  */
 public final class Headers {
+  // TODO(oldergod) make them String ?
+  // Special header names defined in HTTP/2 spec.
+  public static final ByteString PSEUDO_PREFIX = ByteString.encodeUtf8(":");
+  public static final ByteString RESPONSE_STATUS = ByteString.encodeUtf8(":status");
+  public static final ByteString TARGET_METHOD = ByteString.encodeUtf8(":method");
+  public static final ByteString TARGET_PATH = ByteString.encodeUtf8(":path");
+  public static final ByteString TARGET_SCHEME = ByteString.encodeUtf8(":scheme");
+  public static final ByteString TARGET_AUTHORITY = ByteString.encodeUtf8(":authority");
```
JakeWharton: Don't think these make sense here. This API works on strings, not byte strings. Plus the values of these are never exposed publicly. What about on the `Http2` class?

okhttp/src/main/java/okhttp3/Headers.java
```diff
@@ -275,6 +285,21 @@ static void checkValue(String value, String name) {
     }
   }
 
+  public interface Listener {
```
JakeWharton: Doesn't seem like it should public API

okhttp/src/main/java/okhttp3/Headers.java
```diff
@@ -275,6 +285,21 @@ static void checkValue(String value, String name) {
...
+    void onHeaders(List<Headers> headers);
+  }
+
+  //// TODO(oldergod) maybe delete me? maybe not
+  public static Headers flattenList(List<Headers> headersBlocks) {
+    Builder builder = new Builder();
+    for (Headers headers : headersBlocks) {
+      if (headers == null) throw new IllegalStateException("WTF");
```
JakeWharton: Should be NPE and have a meaningful message

okhttp/src/main/java/okhttp3/Headers.java
oldergod: It's used inside `Http2Stream` to forward received headers blocks.

okhttp/src/main/java/okhttp3/Headers.java
oldergod: Looking at this now. There is some places there are `String` comparison, sometimes `ByteString` comparison; trying to find the right line to draw.

okhttp/src/main/java/okhttp3/Headers.java
JakeWharton: That's an internal class. This callback is unused by public API.

okhttp/src/main/java/okhttp3/Headers.java
oldergod: Gave it a shot under Http2; still had to be public to be used inside `MockWebServer`.

okhttp/src/main/java/okhttp3/Headers.java
swankjesse: We're trying to figure out where to put it for duplex calls, when users will be able to subscribe to headers updates.

okhttp/src/main/java/okhttp3/internal/http2/Hpack.java
```diff
@@ -32,7 +31,7 @@
 /**
  * Read and write HPACK v10.
```
yschimke: I assume the v10 was a draft

```suggestion
 * Read and write HPACK.
```

Issue #4258, opened 2018-09-09
Connection Leaks happening during WebSocket Operations


rohanKanojia:
Hi All,

Where exactly we are supposed to do response.close() in case of WebSocket operations? I don't see any in the example i.e [WebSocketEcho](https://github.com/square/okhttp/blob/master/samples/guide/src/main/java/okhttp3/recipes/WebSocketEcho.java) provided in okhttp repo.

We're trying to get rid of these OkHttp connection leaks generated during WebSocket operations in [fabric8io/kubernetes-client](https://github.com/fabric8io/kubernetes-client). For example, There is one websocket operation happening in [PodOperationsImpl](https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/PodOperationsImpl.java#L267)"
```
        try {
            URL url = new URL(URLUtils.join(getResourceUrl().toString(), sb.toString()));
            Request.Builder r = new Request.Builder().url(url).header("Sec-WebSocket-Protocol", "v4.channel.k8s.io").get();
            OkHttpClient clone = client.newBuilder().readTimeout(0, TimeUnit.MILLISECONDS).build();
            final ExecWebSocketListener execWebSocketListener = new ExecWebSocketListener(getConfig(), in, out, err, errChannel, inPipe, outPipe, errPipe, errChannelPipe, execListener);
            clone.newWebSocket(r.build(), execWebSocketListener);
            execWebSocketListener.waitUntilReady();
            return execWebSocketListener;
        } catch (Throwable t) {
            throw KubernetesClientException.launderThrowable(forOperationType("exec"), t);
        }
```
Which gets handled on [ExecWebSocketListener](https://github.com/fabric8io/kubernetes-client/blob/7ea9ce885f9fad72949db6e3f5cc8dceb8ce8766/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/ExecWebSocketListener.java#L212) , We're closing socket in [close(..)](https://github.com/fabric8io/kubernetes-client/blob/7ea9ce885f9fad72949db6e3f5cc8dceb8ce8766/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/ExecWebSocketListener.java#L138) callback here:
```
    private void closeWebSocketOnce(int code, String reason) {
      if (closed.get()) {
        return;
      }

      try {
        WebSocket ws = webSocketRef.get();
[...]

PR #4328, opened 2018-10-22
Okhttp issue4258

brokenjpl:
An attempt to fix the following issue: https://github.com/square/okhttp/issues/4258

The changes in this pull request satisfies the k8s client test mentioned in the issue above.

I have added a new method on the `Interceptor` inteface: `cleanup`.  This gives a chance to interceptors of cleaning up any resources once the interceptor chain has been processed. Unfortunately this is built with Java 7, so couldn't make use of a `default` method and implemented the new method on all Interceptors.  

I'm not super familiar with with okhttp, so I was hoping to start a discussion with this pull request. I assume there will be a few more changes to be applied once this is reviewed. Some questions which could affect the viability of the MR:

- Are users of this library often building their own Intercetpors? If that's the case, I'll need to come up with a different implementation to make this backwards compatible.
- The httpCodec already had a `close()` method, but this was called in a different codepath. Added the `cleanup()` method on HttpCodec as well, with the intention of releasing the StreamAllocation. Is this a viable implementation?


swankjesse:
What happens if we just call `StreamAllocation.release()` in the ConnectInterceptor?

okhttp/src/main/java/okhttp3/Interceptor.java
```diff
@@ -26,6 +26,7 @@
  */
 public interface Interceptor {
   Response intercept(Chain chain) throws IOException;
+  void cleanup();
```
yschimke: This will create a huge amount of pain for users.  There are lots of implementors of Interceptors.  So there is a massive barrier for this change to start.

Also cleanup() doesn't fit the lifecycle of an interceptor which may be used for millions of concurrent requests.  Interceptors already have a chance to cleanup as part of their internal logic in intercept.

brokenjpl: Sounds good, this was one of my concerns with this change.

brokenjpl:
@swankjesse My understanding is that we can't call streamAllocation.release() in the ConnnectInterceptor since the stream allocation is needed further down the chain as well.

I tried implementing something like this in ConnnectInterceptor, but a bunch of unit tests failed:

```
    Response response = realChain.proceed(request, streamAllocation, httpCodec, connection);
    streamAllocation.release();
    return response;
```
That's why I started going down the "cleanup" route. My thought was to simplify what Apache CXF does, something a bit similar with Interceptor 'Phases'.

I'm definitely opened to recommendations!

swankjesse:
I don’t think we can land this as-is. Backwards-compatible API changes are a non-starter for OkHttp.

Issue #3921, opened 2018-03-13
Add easy way to set date header.


NightlyNexus:
I want to put the date header (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Date) on all my requests.

```java
static final ThreadLocal<DateFormat> DATE_HEADER_FORMAT =
      new ThreadLocal<DateFormat>() {
        @Override protected DateFormat initialValue() {
          DateFormat format = new SimpleDateFormat("EEE, dd MMM yyyy HH:mm:ss 'GMT'", Locale.US);
          format.setTimeZone(TimeZone.getTimeZone("GMT"));
          return format;
        }
      };
...
requestBuilder.addHeader("Date", DATE_HEADER_FORMAT.get().format(new Date())
```

I'd like to remove this code in favor of perhaps `okhttp3.Headers.setDate(Date)`.

The downside is that it opens up discussion about other general headers that _could_ be added to the Headers API. Is this general-purpose enough?

swankjesse:
You'd also want overloads for the Request.Builder.

NightlyNexus:
Headers.Builder could be fine.
```java
Headers headers = request.headers();
request = request.newBuilder()
    .headers(headers.newBuilder().setDate(new Date()).build())
    .build();
```
or maybe that's too hidden, then?
Agree that it opens up a lot of questions about conveniences.

yschimke:
+1 I think OkHttp could usefully provide constants and formatters for more headers.  Would it be worth a sweep through https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers?

Date is also pretty common in responses.

JakeWharton:
So long as we provide a java.time overload to anything we add. Date is bad, bad news.

yschimke:
I thought java.time was Android O?

JakeWharton:
It was, yes.

jjshanks:
I did a pull request for this but left out the java.time stuff since that is newer and didn't want to break backward compatibility.

PR #4330, opened 2018-10-22
Add withDate() to header builder

jjshanks:
Fix for #3921

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/HeadersTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/HeadersTest.java
@@ -15,22 +15,19 @@
  */
 package okhttp3;
 
-import java.io.IOException;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Calendar;
-import java.util.Collections;
-import java.util.Date;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.TimeZone;
-
 import okhttp3.internal.Internal;
 import okhttp3.internal.http.HttpHeaders;
 import okhttp3.internal.http2.Header;
 import okhttp3.internal.http2.Http2Codec;
 import org.junit.Test;
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Date;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
 
 import static java.util.Collections.emptyList;
 import static java.util.Collections.emptyMap;
@@ -816,13 +813,44 @@ public final class HeadersTest {
         .byteCount());
   }
 
-  @Test public void withDate() {
-    Date lowerBound = new Date();
-    Calendar calendar = Calendar.getInstance(TimeZone.getTimeZone("GMT"));
-    calendar.add(Calendar.SECOND, 1);
-    Date upperBound = calendar.getTime();
-    Headers headers = new Headers.Builder().withDate().build();
-    Date actual = headers.getDate("Date");
-    assertTrue(actual.after(lowerBound) || actual.before(upperBound));
+  @Test public void addDate() {
+    Date expected = new Date(0);
+    Headers headers = new Headers.Builder()
+            .add("testDate", expected)
+            .build();
+    assertEquals("Thu, 01 Jan 1970 00:00:00 GMT", headers.get("testDate"));
+  }
+
+  @Test public void addDateNull() {
+    try {
+      Headers headers = new Headers.Builder()
+              .add("testDate", (Date) null)
+              .build();
+      fail("Should have complained about null value");
+    } catch (NullPointerException expected) {
+      assertEquals("value for name testDate == null",
+              expected.getMessage());
+    }
+  }
+
+  @Test public void setDate() {
+    Date expected = new Date(1000);
+    Headers headers = new Headers.Builder()
+            .add("testDate", new Date(0))
+            .set("testDate", expected)
+            .build();
+    assertEquals("Thu, 01 Jan 1970 00:00:01 GMT", headers.get("testDate"));
+  }
+
+  @Test public void setDateNull() {
+    try {
+      Headers headers = new Headers.Builder()
+              .set("testDate", (Date) null)
+              .build();
+      fail("Should have complained about null value");
+    } catch (NullPointerException expected) {
+      assertEquals("value for name testDate == null",
+              expected.getMessage());
+    }
   }
 }
--- a/okhttp-tests/src/test/java/okhttp3/HeadersTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/HeadersTest.java
@@ -24,6 +24,8 @@ import java.util.Date;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.TimeZone;
+
 import okhttp3.internal.Internal;
 import okhttp3.internal.http.HttpHeaders;
 import okhttp3.internal.http2.Header;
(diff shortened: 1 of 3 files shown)


okhttp-tests/src/test/java/okhttp3/HeadersTest.java
```diff
@@ -811,4 +815,14 @@
         .build()
         .byteCount());
   }
+
+  @Test public void withDate() {
+    Date lowerBound = new Date();
```
swankjesse: wanna just hardcode a date? Like `new Date(0)` and confirm it’s `1970-01-01T00:00:00Z`

okhttp/src/main/java/okhttp3/Headers.java
```diff
@@ -334,6 +339,17 @@ public Builder addAll(Headers headers) {
       return this;
     }
 
+    /**
+     * Add a Date header with the current time based on
+     * https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Date .
+     */
+    public Builder withDate() {
+      final DateFormat format = new SimpleDateFormat("EEE, dd MMM yyyy HH:mm:ss 'GMT'", Locale.US);
```
swankjesse: Call this instead? `okhttp3.internal.http.HttpDate#format`

okhttp/src/main/java/okhttp3/Headers.java
```diff
@@ -334,6 +339,17 @@ public Builder addAll(Headers headers) {
       return this;
     }
 
+    /**
+     * Add a Date header with the current time based on
+     * https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Date .
+     */
+    public Builder withDate() {
```
swankjesse: Thoughts on a different signature that makes the caller call `new Date()` ?

```
public Builder add(String name, Date value)
public Builder set(String name, Date value)
```

swankjesse: ... I think it’d be simpler to give callers a `DateInterceptor` than a `withDate()` method ’cause then they only need to configure it once.

okhttp-tests/src/test/java/okhttp3/HeadersTest.java
jjshanks: The builder would have to take a WallClock or some other time provider to do that in the current setup. I'll add an overload for withDate(Date date) so the test can use a constant date.

okhttp/src/main/java/okhttp3/Headers.java
jjshanks: Sure thing

okhttp/src/main/java/okhttp3/Headers.java
jjshanks: With regards to the add/set methods I think that would be fine, I was just trying to implement it as I understood it from the original feature request.

Agreed on the DateInterceptor as a good approach. This is typically how I would do it in my clients.

I'm happy to do it either or both ways.

swankjesse: Okay I think my preference is add/set methods that take a Date, and a sample interceptor in our Recipes or Docs somewhere.

I don’t want to provide a built-in Interceptor for this ’cause I want people to get their hands dirty and write their own interceptors. I think it’s actually the nicer way to use the library.

jjshanks: Sounds good, I will update the diff tomorrow with those changes.

yschimke: Could we also support the Java 8 Data/Time API? n.b. I haven't considered the correct types to map, and behaviour for timezones etc.

See https://github.com/square/okhttp/pull/4272/files for the required annotations.

okhttp/src/main/java/okhttp3/Headers.java
```diff
@@ -334,6 +339,17 @@ public Builder addAll(Headers headers) {
       return this;
     }
 
+    /**
+     * Add a Date header with the current time based on
+     * https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Date .
```
yschimke: Do we do this bit for server responses already?

>    A recipient with a clock that receives a response message without a
>    Date header field MUST record the time it was received and append a
>    corresponding Date header field to the message's header section if it
>    is cached or forwarded downstream.

okhttp/src/main/java/okhttp3/Headers.java
jjshanks: @yschimke do you need the Java 8 support in this diff or could it be a separate one?

jjshanks: @swankjesse Made those changes and add a recipe.

okhttp/src/main/java/okhttp3/Headers.java
jjshanks: I'm not sure. Feels like this would also be done via an interceptor though.

okhttp/src/main/java/okhttp3/Headers.java
```diff
@@ -276,6 +278,7 @@ static void checkValue(String value, String name) {
   }
 
   public static final class Builder {
+    final TimeZone gmt = TimeZone.getTimeZone("GMT");
```
swankjesse: unused?

okhttp/src/main/java/okhttp3/Headers.java
```diff
@@ -334,6 +337,28 @@ public Builder addAll(Headers headers) {
       return this;
     }
 
+    /**
+     * Add a header with the specified name and formated Date.
+     * Does validation of header names and values.
+     */
+    public Builder add(String name, Date value) {
+      if (value == null) throw new NullPointerException("value for name " + name + " == null");
+      String strValue = HttpDate.format(value);
```
swankjesse: nit: inline this? ie.

```
add(name, HttpDate.format(value));
```

okhttp/src/main/java/okhttp3/Headers.java
```diff
@@ -334,6 +337,28 @@ public Builder addAll(Headers headers) {
...
+      return this;
+    }
+
+    /**
+     * Set a field with the specified date. If the field is not found, it is added. If the field is
+     * found, the existing values are replaced.
+     */
+    public Builder set(String name, Date value) {
+      if (value == null) throw new NullPointerException("value for name " + name + " == null");
+      String strValue = HttpDate.format(value);
```
swankjesse: nit: inline this?

samples/guide/src/main/java/okhttp3/recipes/CurrentDateHeader.java
```diff
@@ -0,0 +1,45 @@
+package okhttp3.recipes;
```
swankjesse: nit: copyright header? You can borrow from another file in this directory

samples/guide/src/main/java/okhttp3/recipes/CurrentDateHeader.java
```diff
@@ -0,0 +1,45 @@
...
+import okhttp3.Headers;
+import okhttp3.Interceptor;
+import okhttp3.OkHttpClient;
+import okhttp3.Request;
+import okhttp3.Response;
+
+import java.io.IOException;
+import java.util.Date;
+
+public final class CurrentDateHeader {
```
swankjesse: this is awesome

okhttp/src/main/java/okhttp3/Headers.java
yschimke: @jjshanks For me, I think just add it as part of this PR.  I really don't want to be using Date() in my code :)

okhttp/src/main/java/okhttp3/Headers.java
jjshanks: done

okhttp/src/main/java/okhttp3/Headers.java
```diff
@@ -334,6 +335,26 @@ public Builder addAll(Headers headers) {
       return this;
     }
 
+    /**
+     * Add a header with the specified name and formated Date.
```
PromanSEW: Typo: forma**tt**ed

jjshanks: fixed

swankjesse:
Merged manually. Thanks!

PR #4339, opened 2018-10-25
Fail with clear message when missing scheme or host for HttpUrl.Builder.toString.

NightlyNexus:
ran into a place where I was missing the scheme in a logging call.

before:
missing host:
```
	at okhttp3.HttpUrl$Builder.toString(HttpUrl.java:1273)
	at okhttp3.HttpUrlTest.incompleteUrlComposition(HttpUrlTest.java:775)
```

missing scheme:
```
	at okhttp3.HttpUrl.defaultPort(HttpUrl.java:510)
	at okhttp3.HttpUrl$Builder.effectivePort(HttpUrl.java:1039)
	at okhttp3.HttpUrl$Builder.toString(HttpUrl.java:1285)
	at okhttp3.HttpUrlTest.incompleteUrlComposition(HttpUrlTest.java:782)
```

### DIFF
--- a/okhttp/src/main/java/okhttp3/HttpUrl.java
+++ b/okhttp/src/main/java/okhttp3/HttpUrl.java
@@ -1257,6 +1257,9 @@ public final class HttpUrl {
     }
 
     @Override public String toString() {
+      if (scheme == null) throw new IllegalStateException("scheme == null");
+      if (host == null) throw new IllegalStateException("host == null");
+
       StringBuilder result = new StringBuilder();
       result.append(scheme);
       result.append("://");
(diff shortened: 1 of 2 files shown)


okhttp/src/main/java/okhttp3/HttpUrl.java
```diff
@@ -1257,6 +1257,9 @@ public HttpUrl build() {
     }
 
     @Override public String toString() {
+      if (scheme == null) throw new IllegalStateException("scheme == null");
```
yschimke: I'm not convinced by this change, an incomplete HttpUrl.Builder is still valid, and should have a toString e.g. if debugging. #tioli

NightlyNexus: unrelated to this change, since this change just changes the error message, but, yeah, maybe toString should have some understandable message in a string here.

PR #4341, opened 2018-10-26
Allow Requests and Responses to communicate in duplex

oldergod:
Part of #4267

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/DuplexTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/DuplexTest.java
@@ -1,0 +1,187 @@
+/*
+ * Copyright (C) 2018 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okhttp3;
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
+import okhttp3.internal.RecordingHeadersListener;
+import okhttp3.internal.duplex.HeadersListener;
+import okhttp3.internal.duplex.HttpSink;
+import okhttp3.internal.duplex.MwsDuplexAccess;
+import okhttp3.mockwebserver.MockResponse;
+import okhttp3.mockwebserver.MockWebServer;
+import okhttp3.mockwebserver.RecordedRequest;
+import okhttp3.mockwebserver.internal.duplex.DuplexResponseBody;
+import okhttp3.tls.HandshakeCertificates;
+import okio.BufferedSource;
+import org.junit.Rule;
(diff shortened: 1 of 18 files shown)


okhttp-tests/src/test/java/okhttp3/DuplexTest.java
```diff
@@ -0,0 +1,183 @@
...
+
+    Response response = call.execute();
+    HttpSink httpSink = response.httpSink();
+    httpSink.sink().writeUtf8("hey\n");
+    httpSink.headers(Headers.of("a", "android"));
+    httpSink.sink().writeUtf8("whats going on\n");
+    httpSink.headers(Headers.of("b", "blackberry"));
+    httpSink.sink().close();
+
+    // check what the server received
```
oldergod: I wanted to test things on the test thread so moved it here but I'm not using the latchParty with this one. Not sure what's best

okhttp/src/main/java/okhttp3/internal/http2/Http2Stream.java
```diff
@@ -316,9 +316,6 @@ synchronized void receiveRstStream(ErrorCode errorCode) {
 
   public synchronized void setHeadersListener(Header.Listener headersListener) {
     this.headersListener = headersListener;
-    if (!headersQueue.isEmpty() && headersListener != null) {
```
yschimke: Whats the context behind this part of the change?  Does it still get accomplished elsewhere? Not needed?

okhttp/src/main/java/okhttp3/HttpSink.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+package okhttp3;
+
+import java.io.Closeable;
+import java.io.IOException;
+import okio.BufferedSink;
+
+public interface HttpSink extends Closeable {
+  BufferedSink sink();
+  void headers(Headers headers) throws IOException;
```
yschimke: This looks like you only intend to support HTTP/2, but in theory everything except trailers can work fine with HTTP/1.1.

If it's HTTP/2 should it be documented as such and discuss why?

If both, should the docs specify when trailers are allowed?  e.g. check the protocol?

swankjesse: As far as I know there’s no convention in HTTP/1 for servers to send the response headers before the request body has been completely received.

yschimke: Your call. But it does work the same. Browsers, proxies and client libraries all need to line up perfectly. https://serverfault.com/questions/888939/http-pipelining-producing-response-before-request-body-has-finished

okhttp/src/main/java/okhttp3/internal/http2/Http2Stream.java
oldergod: I don't actually remember why. Having headers but no listeners can lead the thread to `wait()` so I'd be to revert this change. Tests don't fail either way so I should address that too.
Do you recall @swankjesse ?

oldergod: Reading and setting listeners happen on the same thread on the client side, so there's not point in notifying here because server is not waiting for us anyway.

okhttp/src/main/java/okhttp3/HttpSink.java
oldergod: We added a comment to say we support only HTTP2 for now

okhttp/src/main/java/okhttp3/Response.java
```diff
@@ -136,6 +143,14 @@ public Headers headers() {
     return headers;
   }
 
+  /**
+   * Should work for any trailers actually right?
+   */
+  synchronized void headersListener(HeadersListener listener) {
```
swankjesse: this shouldn’t be synchronized!

Issue #4315, opened 2018-10-12
LoggingEventListener


swankjesse:
We should make one and make it very easy to install.

amirlivneh:
This can be useful for us too. I can send a PR for a basic implementation. Initially, it can have output similar to that of `HttpLoggingInterceptor`. It can then be incrementally improved to leverage all `EventListener`'s callbacks. The API can be something like:

```
import okhttp3.logging.LoggingEventListener.Factory;

LoggingEventListener.Factory loggingEventListenerFactory =
    new LoggingEventListener.Factory()
        .setLevel(LoggingEventListener.Level.HEADERS);

OkHttpClient client = new OkHttpClient.Builder()
      .eventListenerFactory(loggingEventListenerFactory)
      .build();
```

Alternatively, the API can be more concise and emphasize the feature's functionality over the implementation detail of using `EventListener`:

```
import okhttp3.logging.CallLogger;

CallLogger callLogger =
     new CallLogger().setLevel(CallLogger.Level.HEADERS);

OkHttpClient client = new OkHttpClient.Builder()
      .eventListenerFactory(callLogger.eventListenerFactory())
      .build();
```

To make it easier to install, we can have:
```
    OkHttpClient.Builder builder = new OkHttpClient.Builder();
    callLogger.install(builder);
```
But this may be redundant.

Let me know if one of these work.

yschimke:
A basic one would be really useful.  No objection if we stop there.

If we start thinking about anything more a few features I'd love to see included in this

1) Making it useful for us in getting bug reports e.g. output device details, handshake details.
2) Dynamically Filterable e.g. log requests to a set of site (Function<Request, Boolean>)
3) After the fact logging e.g. for long running requests, failed requests.

amirlivneh:
Any thoughts on the API options?

swankjesse:
Love it. Thanks @amirlivneh !

PR #4345, opened 2018-10-28
Add an EventListener-based logger

amirlivneh:
This adds a basic EventListener-based logger, as suggested in https://github.com/square/okhttp/issues/4315. This initial version and its tests are based on the battle-tested `HttpLoggingInterceptor`. It introduces the minimal set of changes necessary to make `CallLogger` on par with `HttpLoggingInterceptor`, while keeping the PR small and easy to review.

Once this is merged, we can incrementally enhance `CallLogger` to take advantage of `EventListener`'s power and granularity.

By naming the class `CallLogger`, I chose to de-emphasize the implementation detail of it using `EvenListener` under the hood and communicate its functionality. I'm open to suggestions on this and on the API in general.

As a bonus, the last commit leverages the new class to add a verbose option to okcurl.


pom.xml
```diff
@@ -30,7 +30,7 @@
     <module>okhttp-tls</module>
     <module>okhttp-urlconnection</module>
 
-    <module>okhttp-logging-interceptor</module>
+    <module>okhttp-logging</module>
```
yschimke: This and the is a public API change, it makes sense cleanup wise.  But it may be simpler to leave in the current location.  It also makes it possible for two versions of the existing logger to be on the classpath at the same time, because both maven/gradle dependencies can be included.

yschimke:
Since we already have the existing logger, the main win for me here is the different events for timing purposes.  So I'm less sold on a different spin on how to implement the existing logger.

Not against this, but I think this version causes more confusion than value that it adds.

amirlivneh:
The main motivation for me is getting more granular events, and not only for timing purposes, e.g. understand when a request causes a new connection to be established vs. reusing an existing one, etc.

The goal here was not to re-implement the existing logger, but to end up with a much better one, leveraging EventListener capabilities. To avoid a big PR and divide and conquer this, the first version introduces an MVP of the logger that is on par with the existing one. Subsequent PRs will leverage more EventListener callbacks for improved granularity.

Do you have a suggestion for doing it differently? What did you have in mind for the LoggingEventListener mentioned in https://github.com/square/okhttp/issues/4315?

pom.xml
JakeWharton: This change **requires** a redirection pom

amirlivneh: Let me know if you prefer me to simply revert the POM change or add a redirection POM.

yschimke:
I'm not against the approach of iterating, but my main hope is a listener that we can point bug reporters to that will collect data we need to hard to reproduce issues more easily e.g. fails on Samsung when hitting a particular internal site.

What you described sounds great, e.g. recording whether connection is reused.  Just the first code drop is functionally redundant.  For recording request response details, we can point people at the existing logging interceptor.

amirlivneh:
OK, that makes sense.

amirlivneh:
@yschimke, I can try another path to address your concerns, but want to make sure we're on the same page regarding the scope of the first version.

A initial PR can include logging from each one of `EventListener`'s callbacks, where the log simply includes the callback name and the parameters passed to it, relying on their `toString()` implementation.

Other features can include logging headers and possibly the request body. Let me know if the initial PR should include either of those, or any other feature.

yschimke:
@amirlivneh yep, that sounds reasonable.  Please include sample output from running it as a PR comment.

PR #4346, opened 2018-10-28
Cleanup HttpLoggingInterceptor

amirlivneh:


amirlivneh:
I reverted usage of try-with-resources to avoid breaking old Android APIs, leaving the other changes in place.

okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
```diff
@@ -326,7 +326,7 @@ static boolean isPlaintext(Buffer buffer) {
     }
   }
 
-  private boolean bodyHasUnknownEncoding(Headers headers) {
+  private static boolean bodyHasUnknownEncoding(Headers headers) {
```
swankjesse: why prefer static? I mean, it doesn’t _need_ state, but static is bad because we can’t later modify it to do so

amirlivneh: I habitually make methods static, when possible, mostly for clarity. Maybe it's just a bad habit. I do see your point on making it hard to modify it later.

PR #4349, opened 2018-10-30
ConscryptPlatform no longer overrides engine socket preference

cakofony:
This change allows OkHttp consumers to configure whether to use
the engine socket externally.

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/platform/ConscryptPlatform.java
+++ b/okhttp/src/main/java/okhttp3/internal/platform/ConscryptPlatform.java
@@ -112,11 +112,4 @@ public class ConscryptPlatform extends Platform {
       return null;
     }
   }
-
-  @Override
-  public void configureSslSocketFactory(SSLSocketFactory socketFactory) {
-    if (Conscrypt.isConscrypt(socketFactory)) {
-      Conscrypt.setUseEngineSocket(socketFactory, true);
-    }
-  }
 }


cakofony:
Thoughts @yschimke?

yschimke:
Has Conscrypt changed it's default now?  Or is this no longer switchable?

Not an issue for me as long as we test, can you include test evidence that it runs with -Pconscrypt.

We should probably conscrypt version bump to 1.4.0 from the current 1.3.0.

cakofony:
> can you include test evidence that it runs with -Pconscrypt.

Could you elaborate? Do existing CI tests not cover Conscrypt usage?

> Has Conscrypt changed it's default now?

The Conscrypt default has not changed. I'm more comfortable using the default socket implementation than the engine socket given my usage and testing with other clients, and [bugs in the non-default socket implementation](https://github.com/google/conscrypt/issues/433).

This way once conscrypt is comfortable updating the default, there's no change required to okhttp.

> We should probably conscrypt version bump to 1.4.0 from the current 1.3.0.

Sure, I've created #4352 with the version bump

yschimke:
Are there any cases where Conscrypt won't support a custom socket we provide? OkHttp allows this to be overridden, just thinking about what cases could break because of this?

cakofony:
Conscrypt uses one of two custom socket implementations: `ConscryptEngineSocket` and `ConscryptFileDescriptorSocket` (each of which has a java8 shim extension, but those defer to each of these base types). OkHttp currently forces the non-default engine socket instead of the default file descriptor socket. The file descriptor socket works more consistently in my experience.

Running tests locally with conscrypt fails with the following, however it fails in the same way on the master branch, so my development environment may be misconfigured. Apologies if I've missed something in the documentation!

```
[INFO] -------------------------------------------------------
[INFO]  T E S T S
[INFO] -------------------------------------------------------
[INFO] Running okhttp3.tls.HandshakeCertificatesTest
[ERROR] Tests run: 3, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.722 s <<< FAILURE! - in okhttp3.tls.HandshakeCertificatesTest
[ERROR] clientAndServer(okhttp3.tls.HandshakeCertificatesTest)  Time elapsed: 0.597 s  <<< ERROR!
java.util.concurrent.ExecutionException: javax.net.ssl.SSLHandshakeException: Handshake failed
	at okhttp3.tls.HandshakeCertificatesTest.clientAndServer(HandshakeCertificatesTest.java:95)
Caused by: javax.net.ssl.SSLHandshakeException: Handshake failed
Caused by: javax.net.ssl.SSLProtocolException:
SSL handshake aborted: ssl=0x7f9abf258708: Failure in SSL library, usually a protocol error
error:10000416:SSL routines:OPENSSL_internal:SSLV3_ALERT_CERTIFICATE_UNKNOWN (../ssl/tls_record.cc:587 0x7f9abf07e608:0x00000001)
```

yschimke:
I'd like to add a complete this table before landing this

|  | Standard | Custom Socket Factor (with FD) | Custom Socket Factor (without FD) |
| ------ | ------ | --- | --- |
| File Descriptor | Works | Need Test | Not working ?|
| Engine | Works, previously failed (https://github.com/google/conscrypt/issues/198) | Works | Works |

yschimke:
HandshakeCertificatesTest.clientAndServer fails regardless of engine/fd socket being configured.  Still investigating.

cakofony:
Some fixes here: https://github.com/square/okhttp/pull/4440

yschimke:
@cakofony Thanks for coming back to this, should we still land this for 3.13?

cakofony:
Running this change locally causes new tests to fail, so I'm fine closing this PR out!

PR #4357, opened 2018-11-02
Allow incomplete url builder tostring

yschimke:
toString is useful for debug even when incomplete


yschimke:
n.b. the builder.toString is used for the HttpUrl.toString representation, but those fields are non null and > 0 for any valid HttpUrl.

yschimke:
https://github.com/square/okhttp/pull/4361

PR #4359, opened 2018-11-03
Limit the use of regexes in the RFC 7235 challenge parser.

swankjesse:
Creating patterns dynamically to do comma-escaping is pretty awkward.
This is more code but I have more confidence in it.

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/http/HttpHeaders.java
+++ b/okhttp/src/main/java/okhttp3/internal/http/HttpHeaders.java
@@ -15,15 +15,15 @@
  */
 package okhttp3.internal.http;
 
+import java.io.EOFException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.TreeSet;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 import okhttp3.Challenge;
 import okhttp3.Cookie;
 import okhttp3.CookieJar;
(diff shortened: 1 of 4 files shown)


okhttp/src/main/java/okhttp3/internal/http/HttpHeaders.java
```diff
@@ -179,104 +145,168 @@ public static Headers varyHeaders(Headers requestHeaders, Headers responseHeader
...
+    while (!buffer.exhausted() && buffer.getByte(0) == b) {
+      count++;
+      buffer.readByte();
+    }
+    return count;
+  }
+
+  /**
+   * Reads a double-quoted string, unescaping quoted pairs like {@code \"} to the 2nd character in
+   * each sequence. Returns the unescaped string, or null if the buffer doesn't isn't prefixed with
```
yschimke: I don't like no double negatives

Vampire:
@swankjesse, @yschimke 
Are you aware that the tests only pass because one was ignored and for four the expectation was rewritten?
The current code does not do what the JavaDoc of the method states.
The JavaDoc of `okhttp3.Response#challenges` states `Invalid headers and challenges are ignored.`, while now some invalid things are accepted (the four tests with changed expectations).

swankjesse:
Yeah, I think it's a net improvement. We could definitely reject the entire line if any challenge is malformed, but I don't think that's necessary.

Vampire:
Well, I think my reasoning was, if a line does not match my overall regex, it cannot be determined reliably where to split and where the error really is, so it would be better to discard the line as documented in the JavaDoc.

You now even accept invalid challenges that have a token68 and auth params set.
But as the cases you changed are mostly edge cases of faulty implementations, I guess it is ok to have it like that.
Just wanted to make sure it is deliberate :-)

yschimke:
@Vampire Yes - I'm ok with that, see my reasoning here https://github.com/square/okhttp/pull/4252#issuecomment-419799502

Vampire:
Oh, sure, just wanted to make sure. :-)

PR #4362, opened 2018-11-03
Make scheme names case-sensitive again.

swankjesse:


swankjesse:
I don’t think this is necessarily the best behavior! But I don’t want to break compatibility here without a major release to go alongside.

Issue #2840, opened 2016-09-06
[Feature request] - Whole operation timeout


fededonna:
Hi, I'm working on an application that needs to check the entire request-response cycle time spent and set a max for that. It could be a good thing to add a timeout for the complete operation instead of one for connection, one for read and one for right.

Thanks in advance.

swankjesse:
Yep. Let's do something for this.

JakeWharton:
If you happen to be using [Retrofit][1] with its [RxJava][2] call adapter you can already get this for free:

```java
interface MyService {
  @GET("/user")
  Observable<User> getUser();
}
```
```java
MyService service = // ...
service.getUser()
  .timeout(2, SECONDS) // covers sending the request and waiting for the response
  .subscribe(..);
```

 [1]: https://github.com/square/retrofit
 [2]: https://github.com/ReactiveX/RxJava

fededonna:
I'm no using retrofit. I just want to create a client with a timeout for the whole operation.

tobinibot:
I’d also be interested in a whole operation timeout.

ericcj:
We successfully implemented a deadline for both the request and response body reading by subclassing Call in our application in production:  https://gist.github.com/ericcj/49118c877f471b3d8d53db545160e4ac

Looks like there have been a couple related implementations, one inside the HTTP codecs (https://github.com/square/okhttp/pull/2729) and a simple solution for just the body reading (https://github.com/square/okhttp/issues/1130).  

Much of it could also be implemented in an interceptor, except that we would have to close the socket instead of calling the more precise cancel method and it wouldn't cover the seemingly infinite retries on IOExceptions getting a connection in RetryAndFollowUpInterceptor which does while (true) until if (canceled) throw.  

Note that we also have a separate interceptor to timeout DNS (https://gist.github.com/ericcj/a4e5c99982d36d3348ff3c732fa51038) since even our deadline doesn't cover that as per https://github.com/square/okhttp/issues/95 although now using the new Dns interface would be preferable.

robbiepl:
Will this functionality be implemented?

robbiepl:
we waiting ... :))

fededonna:
@swankjesse any update regarding this issue? Thanks in advance.

swankjesse:
We can probably just hook up `AsyncTimeout` to call `Call.cancel()`. Heck, we could even make `Call` work like `Source` and `Sink` so it can be adjusted independent of the client.

```
[...]

PR #4363, opened 2018-11-03
Whole operation timeouts

swankjesse:
Strictly-speaking this change is backwards-incompatible because it adds
a new method to the Call interface. The method returns the call's timeout.

The trickiest part of this is signaling the end of the call, which
occurs after the last byte is consumed of the last follow up request,
or when the call fails. Fortunately this is made easier by borrowing
the sites used by EventListener, which already plots out where calls
end.

https://github.com/square/okhttp/issues/2840

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/WholeOperationTimeoutTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/WholeOperationTimeoutTest.java
@@ -1,0 +1,274 @@
+/*
+ * Copyright (C) 2018 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okhttp3;
+
+import java.io.IOException;
+import java.io.InterruptedIOException;
+import java.net.HttpURLConnection;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
+import okhttp3.mockwebserver.MockResponse;
+import okhttp3.mockwebserver.MockWebServer;
+import okio.BufferedSink;
+import org.junit.Rule;
+import org.junit.Test;
+
+import static okhttp3.TestUtil.defaultClient;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+public final class WholeOperationTimeoutTest {
+  /** A large response body. Smaller bodies might successfully read after the socket is closed! */
+  private static final String BIG_ENOUGH_BODY = TestUtil.repeat('a', 64 * 1024);
+
+  @Rule public final MockWebServer server = new MockWebServer();
+
+  private OkHttpClient client = defaultClient();
+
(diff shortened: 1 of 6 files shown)


okhttp/src/main/java/okhttp3/RealCall.java
```diff
@@ -72,20 +81,32 @@ static RealCall newRealCall(OkHttpClient client, Request originalRequest, boolea
...
       throw e;
     } finally {
       client.dispatcher().finished(this);
     }
   }
 
+  IOException timeoutExit(@Nullable IOException cause) {
+    if (!timeout.exit()) return cause;
+
+    InterruptedIOException e = new InterruptedIOException("timeout");
```
yschimke: Is it useful to include the timeout time in the exception? Would flag misconfiguration to potential bug reporters.

okhttp/src/main/java/okhttp3/RealCall.java
```diff
@@ -72,20 +81,32 @@ static RealCall newRealCall(OkHttpClient client, Request originalRequest, boolea
...
     } catch (IOException e) {
+      e = timeoutExit(e);
       eventListener.callFailed(this, e);
       throw e;
     } finally {
       client.dispatcher().finished(this);
     }
   }
 
+  IOException timeoutExit(@Nullable IOException cause) {
```
yschimke: Should this be marked as @Nullable?

okhttp/src/main/java/okhttp3/RealCall.java
```diff
@@ -72,20 +81,32 @@ static RealCall newRealCall(OkHttpClient client, Request originalRequest, boolea
       executed = true;
     }
     captureCallStackTrace();
+    timeout.enter();
     eventListener.callStart(this);
     try {
       client.dispatcher().executed(this);
       Response result = getResponseWithInterceptorChain();
       if (result == null) throw new IOException("Canceled");
       return result;
```
yschimke: Should you call timeout.exit() here on the happy path?

swankjesse: We don’t exit until the response body is exhausted, which could occur after this returns.

okhttp/src/main/java/okhttp3/RealCall.java
swankjesse: Yes, good call!

swankjesse: Yes, good call!

okhttp-tests/src/test/java/okhttp3/WholeOperationTimeoutTest.java
```diff
@@ -0,0 +1,272 @@
...
+        .url(server.url("/"))
+        .post(sleepingRequestBody(250))
+        .build();
+
+    Call call = client.newCall(request);
+    call.timeout().timeout(1000, TimeUnit.MILLISECONDS);
+    Response response = call.execute();
+    Thread.sleep(250);
+    response.body().source().readUtf8();
+    response.close();
```
yschimke: Can you firm up this test showing that the timeout doesn't fire? Not sure how to test via public API, but potentially via RealCall.timeout().  #tioli

okhttp/src/main/java/okhttp3/RealCall.java
swankjesse: Fixed.

okhttp-tests/src/test/java/okhttp3/WholeOperationTimeoutTest.java
swankjesse: Easy.
```
    assertFalse(call.isCanceled());
```

okhttp/src/main/java/okhttp3/RealCall.java
swankjesse: Ick, this is awkward because we don’t have a good toString() on timeout. Will fix in follow-up.

swankjesse: (and elsewhere that we do this!)

swankjesse: https://github.com/square/okhttp/issues/4367

swankjesse:
Yeah. I’m particularly happy with how reusable AsyncTimeout is.

PR #4365, opened 2018-11-04
Recover from executor shutdowns gracefully.

swankjesse:
This turns out to be pretty difficult because of the way our
dispatcher works.

Calls can be rejected either immediately when the user calls
enqueue(), or later when a queued call is promoted.

It's also awkward because we don't want to hold locks when
calling the user's callFailed() method.

### DIFF
--- a/okhttp/src/main/java/okhttp3/Dispatcher.java
+++ b/okhttp/src/main/java/okhttp3/Dispatcher.java
@@ -75,12 +75,14 @@ public final class Dispatcher {
    * <p>If more than {@code maxRequests} requests are in flight when this is invoked, those requests
    * will remain in flight.
    */
-  public synchronized void setMaxRequests(int maxRequests) {
+  public void setMaxRequests(int maxRequests) {
     if (maxRequests < 1) {
       throw new IllegalArgumentException("max < 1: " + maxRequests);
     }
-    this.maxRequests = maxRequests;
-    promoteCalls();
+    synchronized (this) {
+      this.maxRequests = maxRequests;
+    }
+    promoteAndExecute();
   }
 
   public synchronized int getMaxRequests() {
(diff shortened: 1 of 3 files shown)


okhttp/src/main/java/okhttp3/Dispatcher.java
```diff
@@ -75,12 +77,18 @@ public synchronized ExecutorService executorService() {
...
       throw new IllegalArgumentException("max < 1: " + maxRequests);
     }
-    this.maxRequests = maxRequests;
-    promoteCalls();
+    List<AsyncCall> promotedCalls;
+    synchronized (this) {
+      this.maxRequests = maxRequests;
+      promotedCalls = promoteCalls();
+    }
+    for (int i = 0, size = promotedCalls.size(); i < size; i++) {
```
yschimke: I wonder whether this can be simplfied by a promoteAndExecute() call used in multiple places?

okhttp/src/main/java/okhttp3/Dispatcher.java
```diff
@@ -126,13 +140,16 @@ public synchronized void setIdleCallback(@Nullable Runnable idleCallback) {
     this.idleCallback = idleCallback;
   }
 
-  synchronized void enqueue(AsyncCall call) {
-    if (runningAsyncCalls.size() < maxRequests && runningCallsForHost(call) < maxRequestsPerHost) {
+  void enqueue(AsyncCall call) {
+    synchronized (this) {
+      if (runningAsyncCalls.size() >= maxRequests
```
yschimke: It could be simpler to just add to readyAsyncCalls and then call something like a new promoteAndExecute()

Avoids duplicating this logic but at some marginal cost on the happy path.  So just a thought.

swankjesse: That worked nicely. Please consider taking another look!

PR #4369, opened 2018-11-05
Add APIs to configure the client's default call timeout.

swankjesse:
This isn't independently configured in an interceptor like the other
timeouts because it's already started by the time the interceptors run.

### DIFF
--- a/okhttp/src/main/java/okhttp3/OkHttpClient.java
+++ b/okhttp/src/main/java/okhttp3/OkHttpClient.java
@@ -532,7 +532,7 @@ public class OkHttpClient implements Cloneable, Call.Factory, WebSocket.Factory 
      * Sets the default timeout for complete calls. A value of 0 means no timeout, otherwise values
      * must be between 1 and {@link Integer#MAX_VALUE} when converted to milliseconds.
      *
-     * <p>The call timeout is spans the entire call: resolving DNS, connecting, writing the request
+     * <p>The call timeout spans the entire call: resolving DNS, connecting, writing the request
      * body, server processing, and reading the response body. If the call requires redirects or
      * retries all must complete within one timeout period.
      */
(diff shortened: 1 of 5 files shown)


okhttp/src/main/java/okhttp3/OkHttpClient.java
```diff
@@ -512,18 +521,46 @@ public Builder() {
...
       this.readTimeout = okHttpClient.readTimeout;
       this.writeTimeout = okHttpClient.writeTimeout;
       this.pingInterval = okHttpClient.pingInterval;
     }
 
+    /**
+     * Sets the default timeout for complete calls. A value of 0 means no timeout, otherwise values
+     * must be between 1 and {@link Integer#MAX_VALUE} when converted to milliseconds.
+     *
+     * <p>The call timeout is spans the entire call: resolving DNS, connecting, writing the request
```
yschimke:
```suggestion
     * <p>The call timeout spans the entire call: resolving DNS, connecting, writing the request
```

okhttp/src/main/java/okhttp3/OkHttpClient.java
```diff
@@ -512,18 +521,46 @@ public Builder() {
...
+    public Builder callTimeout(long timeout, TimeUnit unit) {
+      callTimeout = checkDuration("timeout", timeout, unit);
+      return this;
+    }
+
+    /**
+     * Sets the default timeout for complete calls. A value of 0 means no timeout, otherwise values
+     * must be between 1 and {@link Integer#MAX_VALUE} when converted to milliseconds.
+     *
+     * <p>The call timeout is spans the entire call: resolving DNS, connecting, writing the request
```
yschimke:
```suggestion
     * <p>The call timeout spans the entire call: resolving DNS, connecting, writing the request
```

PR #4374, opened 2018-11-06
Fix connection leaks on failed web socket upgrades.

swankjesse:
Closes: https://github.com/square/okhttp/issues/4258


ljnelson:
Given this fix, the `forWebSocket` instance variable is spurious now, yes?

PR #4375, opened 2018-11-06
Add some docs for Cache class

Jawnnypoo:
I was looking into the docs and could not see from the constructor what the size was in terms of. Turns out it is bytes. Hopefully this will help others save a step on figuring this out 😉

### DIFF
--- a/okhttp/src/main/java/okhttp3/Cache.java
+++ b/okhttp/src/main/java/okhttp3/Cache.java
@@ -385,7 +385,7 @@ public final class Cache implements Closeable, Flushable {
     return cache.size();
   }
 
-  /** Size of the cache (in bytes). */
+  /** Max size of the cache (in bytes). */
   public long maxSize() {
     return cache.getMaxSize();
   }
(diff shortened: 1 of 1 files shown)


okhttp/src/main/java/okhttp3/Cache.java
```diff
@@ -175,6 +175,12 @@
   private int hitCount;
   private int requestCount;
 
+  /**
+   * Create a cache which will reside in {@code directory}.
+   *
+   * @param directory a writable directory
+   * @param maxSize the maximum number of bytes this cache should use to store
```
JakeWharton: If you add `{@code maxSize}` in the sentence above you can drop these two `@param` tags. Maybe `Create a cache of at most {@code maxSize} bytes in {@code directory}.`?

Jawnnypoo: Sounds good! Updated

okhttp/src/main/java/okhttp3/Cache.java
```diff
@@ -379,6 +382,7 @@ public long size() throws IOException {
     return cache.size();
   }
 
+  /** Max size of the cache (in bytes). */
```
PromanSEW: `@return`?

PR #4376, opened 2018-11-07
Preemptive auth for proxy CONNECT

swankjesse:
All other forms of preemptive auth are nicely covered by interceptors,
but CONNECT doesn't get that luxury.

https://github.com/square/okhttp/issues/2435

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/CallTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/CallTest.java
@@ -397,7 +397,7 @@ public final class CallTest {
 
     String credential = Credentials.basic("jesse", "secret");
     client = client.newBuilder()
-        .authenticator(new RecordingOkAuthenticator(credential))
+        .authenticator(new RecordingOkAuthenticator(credential, null))
         .build();
 
     Response response = client.newCall(request).execute();
@@ -423,7 +423,7 @@ public final class CallTest {
 
     String credential = Credentials.basic("jesse", "secret");
     client = client.newBuilder()
-        .authenticator(new RecordingOkAuthenticator(credential))
+        .authenticator(new RecordingOkAuthenticator(credential, null))
         .build();
 
     executeSynchronously("/")
@@ -438,7 +438,7 @@ public final class CallTest {
 
     String credential = Credentials.basic("jesse", "secret");
     client = client.newBuilder()
-        .authenticator(new RecordingOkAuthenticator(credential))
+        .authenticator(new RecordingOkAuthenticator(credential, null))
         .build();
 
     try {
@@ -459,7 +459,7 @@ public final class CallTest {
         .setResponseCode(401)
         .setSocketPolicy(SocketPolicy.DISCONNECT_AT_END));
 
-    RecordingOkAuthenticator authenticator = new RecordingOkAuthenticator(null);
+    RecordingOkAuthenticator authenticator = new RecordingOkAuthenticator(null, null);
 
     client = client.newBuilder()
         .authenticator(authenticator)
@@ -1878,7 +1878,7 @@ public final class CallTest {
         .addHeader("Location: " + server2.url("/b")));
 
     client = client.newBuilder()
-        .authenticator(new RecordingOkAuthenticator(Credentials.basic("jesse", "secret")))
+        .authenticator(new RecordingOkAuthenticator(Credentials.basic("jesse", "secret"), null))
         .build();
 
     Request request = new Request.Builder().url(server.url("/a")).build();
@@ -2316,7 +2316,7 @@ public final class CallTest {
         .setBody(gzip("abcabcabc"))
         .addHeader("Content-Encoding: gzip"));
     client = client.newBuilder()
-        .authenticator(new RecordingOkAuthenticator("password"))
+        .authenticator(new RecordingOkAuthenticator("password", null))
         .build();
 
     executeSynchronously("/").assertBody("abcabcabc");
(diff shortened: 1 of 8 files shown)


swankjesse:
My goal is that the thing you try first might Just Work even if you don’t know it exists.

okhttp/src/main/java/okhttp3/internal/connection/RealConnection.java
```diff
@@ -416,14 +417,37 @@ private Request createTunnel(int readTimeout, int writeTimeout, Request tunnelRe
...
+
+    Response fakeAuthChallengeResponse = new Response.Builder()
+        .request(proxyConnectRequest)
+        .protocol(Protocol.HTTP_1_1)
+        .code(HttpURLConnection.HTTP_PROXY_AUTH)
+        .message("Preemptive Authenticate")
+        .body(Util.EMPTY_RESPONSE)
+        .sentRequestAtMillis(-1L)
+        .receivedResponseAtMillis(-1)
+        .header("Proxy-Authenticate", "OkHttp-Preemptive")
```
yschimke: It seems deliberate to not have a constant, this seems awkward.  How sure are you about this?

okhttp-tests/src/test/java/okhttp3/CallTest.java
```diff
@@ -2867,11 +2867,60 @@ private void cancelDuringConnect(String scheme) throws Exception {
...
 
+    RecordedRequest connect1 = server.takeRequest();
+    assertNull(connect1.getHeader("Proxy-Authorization"));
+
+    RecordedRequest connect2 = server.takeRequest();
+    assertEquals("password", connect2.getHeader("Proxy-Authorization"));
+  }
+
+  /** Confirm that we can send authentication information without being prompted first. */
+  @Test public void preemptiveProxyAuthentication() throws Exception {
```
yschimke: Might be worth adding a test showing a failed premptive proxy auth followed up by reactive attempt.

okhttp/src/main/java/okhttp3/Authenticator.java
```diff
@@ -19,17 +19,50 @@
...
+ * proxy.
+ *
+ * <p>Prior to sending any CONNECT request OkHttp always calls the proxy authenticator so that it
+ * may prepare preemptive authentication. OkHttp will call {@link #authenticate} with a fake {@code
+ * HTTP/1.1 407 Proxy Authentication Required} response that has a {@code Proxy-Authenticate:
+ * OkHttp-Preemptive} challenge. The proxy authenticator may return either either an authenticated
+ * request, or null to connect without authentication.
+ * <pre>   {@code
+ *    for (Challenge challenge : response.challenges()) {
+ *      // If this is preemptive auth, use preemptive credential!
```
yschimke: image

okhttp/src/main/java/okhttp3/internal/connection/RealConnection.java
```diff
@@ -416,14 +417,37 @@ private Request createTunnel(int readTimeout, int writeTimeout, Request tunnelRe
...
         .build();
+
+    Response fakeAuthChallengeResponse = new Response.Builder()
+        .request(proxyConnectRequest)
+        .protocol(Protocol.HTTP_1_1)
+        .code(HttpURLConnection.HTTP_PROXY_AUTH)
+        .message("Preemptive Authenticate")
+        .body(Util.EMPTY_RESPONSE)
+        .sentRequestAtMillis(-1L)
+        .receivedResponseAtMillis(-1)
```
yschimke: nit -1L

okhttp/src/main/java/okhttp3/Authenticator.java
```diff
@@ -71,12 +107,9 @@
...
    *
-   * The route is best effort, it currently may not always be provided even when logically
-   * available.  It may also not be provided when an Authenticator is re-used manually in
-   * an application interceptor, e.g. implementing client specific retries.
-   *
-   * @param route The route for evaluating how to respond to a challenge e.g. via intranet.
-   * @param response The response containing the auth challenges to respond to.
+   * <p>The route is best effort, it currently may not always be provided even when logically
+   * available. It may also not be provided when an authenticator is re-used manually in an
+   * application interceptor, such as wehn implementing client-specific retries.
```
yschimke:
```suggestion
   * application interceptor, such as when implementing client-specific retries.
```

yschimke:
@swankjesse looks great, my main nagging doubt in usage is how easy it is when dealing with variable proxies e.g. from a PAC file. (system or okhttp plugin based).

okhttp/src/main/java/okhttp3/internal/connection/RealConnection.java
swankjesse: Fixed

okhttp/src/main/java/okhttp3/Authenticator.java
swankjesse: Fixed!

okhttp-tests/src/test/java/okhttp3/CallTest.java
swankjesse: Good call. Done!

okhttp/src/main/java/okhttp3/internal/connection/RealConnection.java
swankjesse: My expectation is that calling apps are not actually going to look at this challenge. Instead they’re going to hardcode a credential here and not care whether it’s preemptive or reactive.

I documented how it fits together so they _could_ differentiate the two cases when necessary. Perhaps the most interesting question to me is – who uses reactive proxy auth? Maybe if it’s digest? Seems like a bunch of configuration pain.

PR #4377, opened 2018-11-07
changed readme of the mockwerserver to point to parent readme

Dmitry-Borodin:
There are two copies of download instructions for mockwebserver. In the MockWebServer folder and in a parent folder.
The one in parent folder has latest version, the one in server's folder asks to use last version for version number, which confuses me (where can I get one?).
Added reference to parent readme, to have single point of truth and up to date version in one place and avoid confusion.


swankjesse:
I think I prefer the current structure.

PR #4396, opened 2018-11-19
 Use BufferedSource.peek() in Response.peekBody().

NightlyNexus:


okhttp/src/main/java/okhttp3/Response.java
```diff
@@ -148,21 +148,16 @@ public Headers headers() {
...
-    Buffer result;
-    if (copy.size() > byteCount) {
-      result = new Buffer();
-      result.write(copy, byteCount);
-      copy.clear();
-    } else {
-      result = copy;
+    BufferedSource peeked = body.source().peek();
+    Buffer buffer = new Buffer();
+    while (byteCount > 0) {
```
JakeWharton: I think you can eliminate this loop with a call to `request(byteCount)` which will do it for you.

okhttp/src/main/java/okhttp3/Response.java
```diff
@@ -148,21 +148,16 @@ public Headers headers() {
...
-    // There may be more than byteCount bytes in source.buffer(). If there is, return a prefix.
-    Buffer result;
-    if (copy.size() > byteCount) {
-      result = new Buffer();
-      result.write(copy, byteCount);
-      copy.clear();
-    } else {
-      result = copy;
+    BufferedSource peeked = body.source().peek();
+    Buffer buffer = new Buffer();
```
swankjesse: maybe just

```
Buffer buffer = new Buffer();
peek.request(byteCount);
buffer.writeAll(peek, Math.min(byteCount, peek.buffer().size)
```
```

NightlyNexus:
yeah, that's best; thanks!

PR #4402, opened 2018-11-21
avoid NullPointerException

shellhub:
avoid `java.lang.NullPointerException`

### DIFF
--- a/okhttp/src/main/java/okhttp3/Headers.java
+++ b/okhttp/src/main/java/okhttp3/Headers.java
@@ -192,7 +192,7 @@ public final class Headers {
 
   private static String get(String[] namesAndValues, String name) {
     for (int i = namesAndValues.length - 2; i >= 0; i -= 2) {
-      if (name.equalsIgnoreCase(namesAndValues[i])) {
+      if (namesAndValues[i].equalsIgnoreCase(name)) {
         return namesAndValues[i + 1];
       }
     }
--- a/okhttp/src/main/java/okhttp3/Headers.java
+++ b/okhttp/src/main/java/okhttp3/Headers.java
@@ -192,7 +192,7 @@ public final class Headers {
 
   private static String get(String[] namesAndValues, String name) {
     for (int i = namesAndValues.length - 2; i >= 0; i -= 2) {
-      if (namesAndValues[i].equalsIgnoreCase(name)) {
+      if (name.equalsIgnoreCase(namesAndValues[i])) {
         return namesAndValues[i + 1];
       }
     }
@@ -310,6 +310,7 @@ public final class Headers {
     public Builder add(String name, String value) {
       checkName(name);
       checkValue(value, name);
+      removeAll(name);
       return addLenient(name, value);
     }
 


yschimke:
checkName seems overly strict here.  I suggest returning null on null, but allowing all other keys to not find a match and return null.

shellhub:
@yschimke  Thanks for your suggestion.  @swankjesse @JakeWharton do you have any suggestion?

swankjesse:
We should throw a NPE here. If you request a header with no name, that's a bug.

shellhub:
I thinks so, so I call `checkName(name)` function here. It will throw a NPE(with detail message `name==null`) when header name is `null `

yschimke:
But it also fails for an invalid header name.  Postel's law is a good default for APIs, particularly for a query API.

swankjesse:
Oooh I'm not sure. This is something where Kotlin would throw because it's a type error, not a dynamic error.

swankjesse:
Yeah, but in those cases is null special? I like fail-fast for unexpected nulls even if the action is non-destructive.

```
String authHeaderName;
switch (responseCode) {
  case 401: authHeaderName = "Authorization";
  case 407: authHeaderName = "Proxy-Authorization";
  default: authHeaderName = null;
}

String authHeader = headers.get(authHeaderName);
```

yschimke:
I'll defer to your wisdom here.

But remember you are adding extra constraints here to an existing public API.  The null is currently causing an unclear failure.  But headers.get("💩") currently returns null.

shellhub:
`💩` is invalid char, we throw `java.lang.IllegalArgumentException: Unexpected char 0xd83d at 0 in header name: 💩` instead of returning null

``` java
public static void main(String[] args) {
    Request request = new Request.Builder().url("https://www.google.com")
            .addHeader("\uD83D\uDCA9", "example value")
            .build();
}
```
> Exception in thread "main" java.lang.IllegalArgumentException: Unexpected char 0xd83d at 0 in header name: 💩

yschimke:
@shellhub your example calls addHeader, which is fine, we want to stop clients creating invalid requests.

But the PR is not about addHeader, its about get

```
  public static void main(String[] args) {
    Headers h = Headers.of();
    System.out.println(h.get("💩")); // null
  }
```

The more I think about it, this is a bad idea. It will throw an exception in previously working interceptors breaking some number of existing users for just checking if a non existent header is there.

yschimke:
No objection to a null check.

JakeWharton:
+1 to null check only. Nothing wrong with invalid headers being asked for.

shellhub:
@JakeWharton @swankjesse @yschimke I update code. I think this will avoid `java.lang.NullPointerException` when call `get(String name)`

swankjesse:
We gotta clarify whether this argument’s type is `@Nullable String` or `@NonNull String`. If/when we do to OkHttp what we’ve done to Okio & change the implementation language to Kotlin, this type choice will have teeth.

I think its bad to accept a wider type than we need. We want the compiler to help prevent callers from calling get() with an argument that'll never hit.

Issue #4386, opened 2018-11-14
callEnd called twice for redirect


yschimke:
```
$ okurl --tracing CONSOLE 'https://nghttp2.org/httpbin/redirect-to?url=http://www.twitter.com/robots.txt'
[0 ms] callStart: Request{method=GET, url=https://nghttp2.org/httpbin/redirect-to?url=http://www.twitter.com/robots.txt, tags={class java.lang.Object=TokenSet(name=default)}}
[5 ms] dnsStart: nghttp2.org
[139 ms] dnsEnd: [nghttp2.org/139.162.123.134, nghttp2.org/2400:8902:0:0:f03c:91ff:fe69:a454]
[169 ms] connectStart: nghttp2.org/139.162.123.134:443 DIRECT
[530 ms] secureConnectStart
[894 ms] secureConnectEnd
[908 ms] connectEnd: h2
[908 ms] connectionAcquired: Connection{nghttp2.org:443, proxy=DIRECT hostAddress=nghttp2.org/139.162.123.134:443 cipherSuite=TLS_AES_256_GCM_SHA384 protocol=h2}
[1094 ms] requestHeadersStart
[1114 ms] requestHeadersEnd
[1114 ms] responseHeadersStart
[1409 ms] responseHeadersEnd: Response{protocol=h2, code=302, message=, url=https://nghttp2.org/httpbin/redirect-to?url=http://www.twitter.com/robots.txt}
[1409 ms] responseBodyStart
[1410 ms] responseBodyEnd: byteCount=0
[1410 ms] connectionReleased
[1410 ms] callEnd
[1411 ms] dnsStart: www.twitter.com
[1428 ms] dnsEnd: [www.twitter.com/104.244.42.193, www.twitter.com/104.244.42.65]
[1428 ms] connectStart: www.twitter.com/104.244.42.193:80 DIRECT
[1443 ms] connectEnd: http/1.1
[1443 ms] connectionAcquired: Connection{www.twitter.com:80, proxy=DIRECT hostAddress=www.twitter.com/104.244.42.193:80 cipherSuite=none protocol=http/1.1}
[1519 ms] requestHeadersStart
[1519 ms] requestHeadersEnd
[1519 ms] responseHeadersStart
[1650 ms] responseHeadersEnd: Response{protocol=http/1.1, code=200, message=OK, url=http://www.twitter.com/robots.txt}
[1651 ms] responseBodyStart
[1813 ms] responseBodyEnd: byteCount=52
[...]

PR #4408, opened 2018-11-23
Test EventListener handling of redirected requests

amirlivneh:
Reproduces https://github.com/square/okhttp/issues/4386.

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
@@ -16,6 +16,7 @@
 package okhttp3;
 
 import java.io.IOException;
+import java.net.HttpURLConnection;
 import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.net.Proxy;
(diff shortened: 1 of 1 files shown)


okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
```diff
@@ -1082,4 +1083,46 @@ private void enableTlsWithTunnel(boolean tunnelProxy) {
...
+        "ResponseBodyEnd", "RequestHeadersStart", "RequestHeadersEnd", "ResponseHeadersStart",
+        "ResponseHeadersEnd", "ResponseBodyStart", "ResponseBodyEnd", "ConnectionReleased",
+        "CallEnd");
+    assertEquals(expectedEvents, listener.recordedEventTypes());
+  }
+
+  @Ignore("CallEnd emitted twice")
+  @Test
+  public void redirectUsingNewConnectionEventSequence() throws IOException {
+    MockWebServer otherServer = new MockWebServer();
```
swankjesse: ahhhh, that explains it.

swankjesse:
Thanks much

swankjesse:
We’re ending the call when the _connection_ is done, but we should be doing it when the _call_ is done.

https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/internal/connection/StreamAllocation.java#L43

PR #4409, opened 2018-11-23
Avoid retrying when client is missing body File

yschimke:
FileNotFound should be treated like an UnrepeatableRequestBody, all other IOExceptions are variable.  But FileNotFoundException is most likely to represent a bad client request.  Unfortunately this is more of a heuristic, given that a Socket IOException could be FileNotFound and a different route could return something else.


swankjesse:
Makes sense!

Issue #4218, opened 2018-08-17
EventListener can't access wire-level request in case of failure


amirlivneh:
I'm trying to log the URI of failed requests using an `EventListener`. The URI may be rewritten by Interceptors and I'm interested in the URI of the wire-level request.

In case of a successful request, the URI of the wire-level request can be obtained by by calling `Response.request()`. But if the request fails and there is no `Response` object, there seems to be no way to get this information; Only the original is accessible through `Call.request()`.

amirlivneh:
2 possible solutions are:
- Add a `Call.wireLevelRequest()` method. This is not strictly backward compatible, but may be acceptable.
- Change `EventListener.requestHeadersStart()`'s signature to `EventListener.requestHeadersStart(Call call, Request wireLevelRequest)`. This is not backward compatible, but there is a disclaimer in https://github.com/square/okhttp/wiki/Events#availability stating that this interface is subject to change.

If any of the above works, or if you can suggest a better solution, I can send a PR.

swankjesse:
If you need a quick & dirty fix you can use Call tags and interceptors to hack around this. Here’s a sketch:

```
public void callOkHttp() {
  Request request = new Request.Builder()
      .url(...)
      .tag(HackWireLevelRequestHolder.class, new HackWireLevelRequestHolder())
      .build();
  ...
}


class HackWireLevelRequestHolder {
  public static final Interceptor INTERCEPTOR = new Interceptor() {
    public Response intercept(Chain chain) throws IOException {
      Request request = chain.request();
      request.tag(HackWireLevelRequestHolder.class).wireLevelRequest = request;
      return chain.proceed(request);
    }
  }

  Request wireLevelRequest = null;
}
```

swankjesse:
Will this help?
https://github.com/square/okhttp/issues/4638

amirlivneh:
Yes, it can help, assuming that the wire-level request will be reachable from the parameters of the new methods.

yschimke:
Closing in favour of the request events.

yschimke:
Just hit this but without the ability to control the creation of the request.  Just realised I'm saved by the Call.Factory which I can trivially decorate to set the tag I need.

I don't actually need to check it, just use Call.Factory wrapper instead of an interceptor.

PR #4410, opened 2018-11-23
Test EventListener handling of requests modified by interceptors

amirlivneh:
Reproduces https://github.com/square/okhttp/issues/4218.

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
@@ -1132,9 +1132,8 @@ public final class EventListenerTest {
   public void requestModifiedByInterceptor() throws IOException {
     server.enqueue(new MockResponse());
 
-    client
-        .newBuilder()
-        .interceptors.add(
+    client = client.newBuilder()
+        .addInterceptor(
             new Interceptor() {
               @Override
               public Response intercept(Chain chain) throws IOException {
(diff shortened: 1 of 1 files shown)


okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
```diff
@@ -1125,4 +1126,28 @@ public void redirectUsingNewConnectionEventSequence() throws IOException {
         "CallEnd");
     assertEquals(expectedEvents, listener.recordedEventTypes());
   }
+
+  @Ignore("Header added by interceptor is not seen by EventListener.")
+  @Test
+  public void requestModifiedByInterceptor() throws IOException {
+    server.enqueue(new MockResponse());
+
+    client
```
swankjesse: You need to assign this to the client again.

amirlivneh: Good catch. Fixed.

swankjesse:
This test reproduces the reported problem but I still feel that this is working as it should.  That is, I don't think there's anything to fix here.

amirlivneh:
Yes, that's true. It's there to help the implementor of the enhancement.

swankjesse:
Got it. I’m going to close this but keep it linked to the issue. My preference is not to check in tests for enhancements ’cause it suggests the API is decided.

PR #4414, opened 2018-11-25
Change contributing guide in repo to match wiki

yschimke:
We could link, but much simpler to have each be standalone to ensure it gets read.


yschimke:
Copied from current content here - https://github.com/square/okhttp/wiki/Contributing

PR #4417, opened 2018-11-27
remove duplicate values

shellhub:


yschimke:
Why are you removing repeated headers?

shellhub:
I think we should remove repeated headers.  old values is useless
``` java
    public static void main(String[] args) {
                .url("https://www.google.com")
                .addHeader("name", "value1")
                .addHeader("name", "value2")
                .build();
        System.out.println(request.headers);
    }
```
above code output 
```
name:value1
name:value2
```
instead of 
```
name:value2
```

yschimke:
use set, this is deliberate and according to HTTP spec

PR #4428, opened 2018-12-03
Feat. from kuaikan

RocLiuYC:
1. support user-defined cache-key
2. add write disk progress listener

Signed-off-by: LiuYongChuan <lyc@kkworld.com>

### DIFF
--- a/okhttp/src/main/java/okhttp3/Cache.java
+++ b/okhttp/src/main/java/okhttp3/Cache.java
@@ -15,6 +15,8 @@
  */
 package okhttp3;
 
+import android.text.TextUtils;
+
 import java.io.Closeable;
 import java.io.File;
 import java.io.Flushable;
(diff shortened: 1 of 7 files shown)


okhttp/src/main/java/okhttp3/Cache.java
```diff
@@ -15,6 +15,8 @@
  */
 package okhttp3;
 
+import android.text.TextUtils;
```
yschimke: We can't use Android classes in core okhttp

okhttp/src/main/java/okhttp3/Request.java
```diff
@@ -217,6 +228,11 @@ public Builder headers(Headers headers) {
       return this;
     }
 
+    public Builder cacheKey(String cacheKey) {
```
yschimke: There is a high bar to public API additions or changes.  This should be discussed and would deserve it's own PR separate from the listener change.

I think cacheKey semantics alone would warrant a fair amount of discussion to get right if it is likely to be accepted.

okhttp/src/main/java/okhttp3/Call.java
```diff
@@ -57,6 +57,8 @@
    */
   Response execute() throws IOException;
 
+  Response execute(DiskCacheListener diskCacheListener) throws IOException;
```
yschimke: This API change feels way to "in your face", this is the key part of the okhttp client experience, and will now prompt users to choose between two options for what I presume is a niche feature.  Not all users use caches, and not all those will want a listener.

yschimke:
This PR isn't likely to get accepted because

1) It's a 2 distinct non mainstream feature
2) has no tests
3) change public api
4) Isn't really documented for discussion, e.g. why it's desired or critical.

PR #4430, opened 2018-12-03
Support SOCKS proxies on older Android versions

### LINKED ISSUE #2315: SOCKS proxy not working on Android platform
SOCKS proxy is **completely** broken on Android. When I set up SOCKS proxy in my app, all requests failed, when I wrote an instrumentation test, I found even create a default okhttp instance with a SOCKS proxy, it fails.

Here's the stack trace:
````
java.net.UnknownHostException: Host is unresolved: www.google.com
at java.net.Socket.connect(Socket.java:867)
at okhttp3.internal.Platform$Android.connectSocket(Platform.java:200)
at okhttp3.internal.io.RealConnection.connectSocket(RealConnection.java:137)
at okhttp3.internal.io.RealConnection.connect(RealConnection.java:108)
at okhttp3.internal.http.StreamAllocation.findConnection(StreamAllocation.java:188)
at okhttp3.internal.http.StreamAllocation.findHealthyConnection(StreamAllocation.java:127)
at okhttp3.internal.http.StreamAllocation.newStream(StreamAllocation.java:97)
at okhttp3.internal.http.HttpEngine.connect(HttpEngine.java:289)
at okhttp3.internal.http.HttpEngine.sendRequest(HttpEngine.java:241)
at okhttp3.RealCall.getResponse(RealCall.java:240)
at okhttp3.RealCall$ApplicationInterceptorChain.proceed(RealCall.java:198)
at okhttp3.RealCall.getResponseWithInterceptorChain(RealCall.java:160)
at okhttp3.RealCall.execute(RealCall.java:57)
at org.mariotaku.twidere.test.okhttp3.OkHttpClientTest.testSocksFunctionality(OkHttpClientTest.java:27)
at java.lang.reflect.Method.invoke(Native Method)
at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
at org.junit.runners.Suite.runChild(Suite.java:128)
at org.junit.runners.Suite.runChild(Suite.java:27)
at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
at org.junit.runner.JUnitCore.run(JUnitCore.java:115)
at android.support.test.internal.runner.TestExecutor.execute(TestExecutor.java:54)
at android.support.test.runner.AndroidJUnitRunner.onStart(AndroidJUnitRunner.java:240)
at android.app.Instrumentation$InstrumentationThread.run(Instrumentation.java:1879)
````

Sample code:

````java
import org.junit.Test;

import java.net.InetSocketAddress;
import java.net.Proxy;

import okhttp3.OkHttpClient;
import okhttp3.Request;

/**
 * Created by mariotaku on 16/2/5.
 */
public class OkHttpClientTest {

    @Test
    public void testSocksFunctionality() throws Exception {
        final Proxy proxy = new Proxy(Proxy.Type.SOCKS, InetSocketAddress.createUnresolved("127.0.0.1", 1080));
        final OkHttpClient client = new OkHttpClient.Builder()
                .proxy(proxy)
                .build();
        final Request request = new Request.Builder()
                .url("https://www.google.com/")
                .build();
        assertTrue(client.newCall(request).execute().isSuccessful());
    }

}

````

bemasc:
This is currently nonfunctional due to an interaction
between InetAddress.createUnresolved and the SOCKS 4 protocol.

Fixes #2315

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/connection/RouteSelector.java
+++ b/okhttp/src/main/java/okhttp3/internal/connection/RouteSelector.java
@@ -31,6 +31,7 @@ import okhttp3.Call;
 import okhttp3.EventListener;
 import okhttp3.HttpUrl;
 import okhttp3.Route;
+import okhttp3.internal.platform.Platform;
 import okhttp3.internal.Util;
 
 /**
@@ -176,7 +177,8 @@ public final class RouteSelector {
           + "; port is out of range");
     }
 
-    if (proxy.type() == Proxy.Type.SOCKS) {
+    if (proxy.type() == Proxy.Type.SOCKS && Platform.get().getSocksVersion() >= 5) {
+      // SOCKS 5 supports DNS resolution at the proxy.
       inetSocketAddresses.add(InetSocketAddress.createUnresolved(socketHost, socketPort));
     } else {
       eventListener.dnsStart(call, socketHost);
--- a/okhttp/src/main/java/okhttp3/internal/platform/AndroidPlatform.java
+++ b/okhttp/src/main/java/okhttp3/internal/platform/AndroidPlatform.java
@@ -454,4 +454,12 @@ class AndroidPlatform extends Platform {
       throw new IllegalStateException("No TLS provider", e);
     }
   }
+
+  @Override public int getSocksVersion() {
+    if (Build.VERSION.SDK_INT <= 23) {
+      // Android API 23 and lower only support SOCKS 4, not SOCKS 5.
+      return 4;
+    }
+    return super.getSocksVersion();
+  }
 }
--- a/okhttp/src/main/java/okhttp3/internal/platform/Jdk9Platform.java
+++ b/okhttp/src/main/java/okhttp3/internal/platform/Jdk9Platform.java
@@ -82,6 +82,12 @@ final class Jdk9Platform extends Platform {
         "clientBuilder.sslSocketFactory(SSLSocketFactory) not supported on JDK 9+");
   }
 
+  @Override public int getSocksVersion() {
+    // This setting controls the default SOCKS version.  See
+    // https://docs.oracle.com/javase/9/docs/api/java/net/doc-files/net-properties.html
+    return Integer.parseInt(System.getProperty("socksProxyVersion", "5"));
+  }
+
   public static Jdk9Platform buildIfSupported() {
     // Find JDK 9 new methods
     try {
--- a/okhttp/src/main/java/okhttp3/internal/platform/Platform.java
+++ b/okhttp/src/main/java/okhttp3/internal/platform/Platform.java
@@ -291,6 +291,12 @@ public class Platform {
   public void configureSslSocketFactory(SSLSocketFactory socketFactory) {
   }
 
+  public int getSocksVersion() {
+    // On JDK 7 and 8, SOCKS 5 is the only supported version.
+    // See https://bugs.openjdk.java.net/browse/JDK-8129444
+    return 5;
+  }
+
   @Override public String toString() {
     return getClass().getSimpleName();
   }


swankjesse:
FYI @Vampire !

I like the structure here. Biggest challenge is that if the platform only has SOCKS4 we end up leaking the DNS name to the DNS provider. For Tor this is bad!

yschimke:
@swankjesse could we have a global intent flag that indicates something along the lines of "public/corporate/secret" and use those to make decisions like this?  Without pushing this sort of awareness out to users accidentally combining features that leak details?

e.g. This can be ok if DOH is used at the same time.

Vampire:
Thanks for the mention.

I don't think this fix is the better one and would still suggest my fix to always resolve if SOCKS.

Some problems I see with this approach are:
- SOCKS5 where address resolution is only possible on the client but not on the proxy still does not work
- SOCKS5 proxy using a custom configured `Dns` instance does not work
- Determining SOCKS version depending on Android or JRE version is bad, as you could also use some socket factory that properly supports SOCKS4 or SOCKS5 even on the platform that don't properly support it natively (at least for Java this should be possible, not sure about Android)
- Determining the SOCKS version depending on the `socksProxyVersion` system property only works if you use the system properties and the default proxy selector of the JRE that honors those properties, but you can any time also have an own proxy selector or explicitly configured proxy that configures a SOCKS4 proxy.

Regarding leaking the hostname to the DNS for Tor as I said before the three possibilities I see are:
- Make the resolving only if a configuration is set like in the last version of my PR
- Make the resolving only if an experts-only system property is set, so the API is not cluttered with the configuration option and non-experts are not confused by this option but if someone asks for SOCKS4-support or SOCKS5-support with resolution on client, this experts-only property can be told
- Probably my favorite option: if you develop an app that considers "leaking" the hostname to DNS a security risk like Tor, then the app should set a custom `Dns` instance like `httpClientBuilder.dns(hostname -> Collections.emptyList());` which will make sure the hostname is never leaked to the DNS, as only the default `Dns` instance asks the system-configured DNS server.

bemasc:
> Probably my favorite option: if you develop an app that considers "leaking" the hostname to DNS a security risk like Tor, then the app should set a custom Dns instance like httpClientBuilder.dns(hostname -> Collections.emptyList()); which will make sure the hostname is never leaked to the DNS, as only the default Dns instance asks the system-configured DNS server.

That behavior would be fine with me.  I've opened a new PR (#4432) that goes in that direction.  (I'm happy to close either PR if we have consensus that the other is preferable.)

Vampire:
That's not really what I meant.
This way you can only use SOCKS4 and SOCKS5 with client-resolution if you configure a non-standard `Dns` which also is not too logic for a user.
What I meant is my PR in original version: https://github.com/Vampire/okhttp/commit/924c117530fba2f3370d414ecada59cbae59a64b
Or with a config option like in: https://github.com/Vampire/okhttp/commit/bddfa5cfa4fd95815e8c4008bec7f5f47c5c22ea
Or then with a system property or similar.

I'd still vote for my original version and something like `Tor` should do `httpClientBuilder.dns(hostname -> Collections.emptyList());`.

Vampire:
> FWIW, I disagree with this, especially for SOCKS5. When using the API, I expected that a custom Dns would be respected regardless of whether a proxy was in use. Having Builder.dns() silently fail when a SOCKS proxy is set was surprising to me.

What do you mean?
I totally agree to what you say and never said anything different.
Currently when a SOCKS proxy is configured, the `Dns` is never asked, but for SOCKS4 it fails and for SOCKS5 it relies on the proxy server being able to resolve the hostname as only the hostname is sent.
I also complained that SOCKS4 does not work and that SOCKS5 only works with resolution on the proxy and that both should work and also with a custom `Dns` being considered.
My changes take care of this.

bemasc:
> I totally agree to what you say and never said anything different.

OK, great.  I may have misunderstood.

> I also complained that SOCKS4 does not work and that SOCKS5 only works with resolution on the proxy and that both should work and also with a custom Dns being considered.

OK.  So there are four cases: (SOCKS 4 vs. SOCKS5) x (default DNS vs. custom DNS).  I think #4432 resolves 3 out of 4, leaving the case of SOCKS 4 and default DNS.

Based on your feedback, I added a small change to #4432 that returns a custom error message in this case.  This change doesn't make it "just work", but I think that's OK.  The few developers who care about proxy support on old versions of Android now have a reasonably well-documented way to support it, without risking accidental DNS leaks.

Vampire:
`UnknownHostException` can also be thrown due to other reasons so it is not the best indicator and your change only handles Android, not normal JRE usage.
What do you have against my solution that properly handles all cases and indeed "just works" properly out of the box with both, standard and custom DNS?

I really don't know why you try to reinvent the wheel I already provided.
My code is present, working and tested.
The only problem is that Jesse and Yuri refused to accept it.

Btw. your version with checking for the version being at least 5 has another problem that mine does not have.
It does not support SOCK4a which indeed can handle unresolved addresses by letting the proxy resolve the hostname.

bemasc:
@Vampire, either of your proposed solutions would solve my personal use case.  However, I can see why the maintainers might be reluctant to change the default SOCKS5 resolution behavior, or to add new public API methods for this unusual use case.  Perhaps a smaller change will be more acceptable.

Vampire:
Juts looked at objectively, my change is smaller than yours, more correct than yours, covering more cases than yours and works out-of-the-box.
The API option just was added as it was requested that if it gets accepted then it needs to be configurable.

Imho I would not add the configuration option and if someone wants to prevent "leakage" he has to do `httpClientBuilder.dns(hostname -> Collections.emptyList());`.
In case it gets accepted if configurable, the option in the API is one way, adding a system property that is considered another one.
I don't really care which of those three possibilities is accepted even if I have a personal preference, as long as one that works correctly and is obvious actually is accepted. ;-)

uselessness12:
Perfect

yschimke:
Closing out since it's pre kotlin.  Worth getting a positive signal before continuing with this PR.

PR #4435, opened 2018-12-07
Provided missing internal methods to test duplex calls

oldergod:


oldergod:
Hard to believe I broke some tests with my code. Hmm.

PR #4448, opened 2018-12-17
Handle security exception when testing if conscrypt should be used

zlogic:
In some containers or environments, a security policy may prevent reading unknown system properties.

As a result, `Platform#isConscryptPreferred()` is throwing an exception when running OkHttp in our container - because `okhttp.platform` is not on the whitelist of System Properties which are allowed to be read.

This pull request adds handling for such exceptions so that `OkHttpClient` initialization doesn't fail.

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/platform/Platform.java
+++ b/okhttp/src/main/java/okhttp3/internal/platform/Platform.java
@@ -20,6 +20,7 @@ import java.io.IOException;
 import java.lang.reflect.Field;
 import java.net.InetSocketAddress;
 import java.net.Socket;
+import java.security.AccessControlException;
 import java.security.NoSuchAlgorithmException;
 import java.security.Security;
 import java.util.ArrayList;
@@ -75,10 +76,10 @@ import okio.Buffer;
  * <p>Supported on Android 6.0+ via {@code NetworkSecurityPolicy}.
  */
 public class Platform {
+  private static final Logger logger = Logger.getLogger(OkHttpClient.class.getName());
   private static final Platform PLATFORM = findPlatform();
   public static final int INFO = 4;
   public static final int WARN = 5;
-  private static final Logger logger = Logger.getLogger(OkHttpClient.class.getName());
 
   public static Platform get() {
     return PLATFORM;
@@ -187,8 +188,12 @@ public class Platform {
 
   public static boolean isConscryptPreferred() {
     // mainly to allow tests to run cleanly
-    if ("conscrypt".equals(System.getProperty("okhttp.platform"))) {
-      return true;
+    try {
+      if ("conscrypt".equals(System.getProperty("okhttp.platform"))) {
+        return true;
+      }
+    } catch (AccessControlException ex) {
+      logger.log(Level.WARNING, "Cannot read the okhttp.platform property", ex);
     }
 
     // check if Provider manually installed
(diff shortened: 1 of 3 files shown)


okhttp/src/main/java/okhttp3/internal/platform/Platform.java
```diff
@@ -75,10 +76,10 @@
  * <p>Supported on Android 6.0+ via {@code NetworkSecurityPolicy}.
  */
 public class Platform {
+  private static final Logger logger = Logger.getLogger(OkHttpClient.class.getName());
```
swankjesse: Please revert this move.

zlogic: If the logger is initialized after `private static final Platform PLATFORM`, then `this.logger` will be null in `Platform.isConscryptPreferred()`.
So an NullPointerException will be thrown on line 196 when `Platform.isConscryptPreferred()` calls `logger.log(...)`.

Would you prefer to use another instance of the logger, or to not log this exception at all?

okhttp/src/main/java/okhttp3/internal/platform/Platform.java
```diff
@@ -187,8 +188,12 @@ public CertificateChainCleaner buildCertificateChainCleaner(SSLSocketFactory ssl
 
   public static boolean isConscryptPreferred() {
     // mainly to allow tests to run cleanly
-    if ("conscrypt".equals(System.getProperty("okhttp.platform"))) {
-      return true;
+    try {
+      if ("conscrypt".equals(System.getProperty("okhttp.platform"))) {
```
yschimke: should line 270 change also? anywhere we check any properties.

zlogic: OK, I've went through the code and updated other calls to System Properties as well.

zlogic: OK, I've went through the code and updated other calls to System Properties as well.

okhttp/src/main/java/okhttp3/internal/platform/Platform.java
yschimke: To unblock this and given our policy on public APIs how about using a literal string here and commenting why.

yschimke: Your explanation seems reasonable.  I suggest leaving it as you now have it.

okhttp/src/main/java/okhttp3/internal/platform/Platform.java
yschimke: What do you think about moving this pattern, attempt and swallow ACE to okhttp.internal.Util.getSystemProperty(String key, @Nullable String defaultValue).

zlogic: I agree, I've refactored the code and that made the pull request much cleaner.

okhttp/src/main/java/okhttp3/internal/platform/Platform.java
zlogic: By moving the try-catch block into `okhttp.internal.Util.getSystemProperty`, it's no longer needed to have the logger initialized in `Platform.isConscryptPreferred()`, so I reverted this change.

swankjesse:
This is great! Thanks.

PR #4452, opened 2018-12-18
Removed HttpSink and only expose just a sink instead

oldergod:
We learned that gRPC allows **headers \*data** for requests, and **headers \*data trailers** for responses. We don't need the `HttpSink` anymore; we can expose a sink to write to when we do bidirectional communications.

#4267


okhttp-tests/src/test/java/okhttp3/DuplexTest.java
```diff
@@ -0,0 +1,171 @@
...
+        .duplex("POST")
+        .build());
+
+    Response response = call.execute();
+    BufferedSink sink = response.sink();
+    sink.writeUtf8("hey\n");
+    sink.writeUtf8("whats going on\n");
+    sink.close();
+
+    // check what the server received
```
swankjesse: nit: https://nedbatchelder.com/blog/201401/comments_should_be_sentences.html

okhttp-tests/src/test/java/okhttp3/DuplexTest.java
```diff
@@ -0,0 +1,171 @@
...
+    BufferedSource requestBodySource = requestBodySourceRef.get();
+    assertEquals("hey", requestBodySource.readUtf8Line());
+    assertEquals("whats going on", requestBodySource.readUtf8Line());
+    assertTrue(requestBodySource.exhausted());
+  }
+
+  // TODO(oldergod) write tests for headers discarded with 100 Continue
```
swankjesse: yes!

PR #4462, opened 2018-12-22
HTTP/2 test for MockWebServer

amirlivneh:


swankjesse:
We're testing OkHttp with MockWebServer and MockWebServer with OkHttp. Doesn't feel right.

swankjesse:
(this functionality is already covered by OkHttp’s tests.)

PR #4463, opened 2018-12-24
Drop the benchmarks module

swankjesse:
Most of the things it's exercising are obsolete (old versions of Netty
and Apache HTTP), it's missing important alternatives (cronet, Jetty),
and Caliper has backwards-incompatible API changes.

I think it's better for us to delete this and possibly start fresh
in the future. Possibly in a separate standalone project.


amirlivneh:
@swankjesse, I’m considering implementing new benchmarks, possibly using jmh. Any thoughts on this? Specifically, should this be a separate project?

swankjesse:
Separate project has lots of upsides. You can more easily compare different OkHttp releases against each other. It also makes it easier to add various competing clients.

Happy to contribute there!

amirlivneh:
@swankjesse, do you have any lessons from the removed benchmarks that we should apply in the new project? E.g. do you recommend to stick initially to the same benchmark scenario?

swankjesse:
My favorite benchmarks are representative of real world needs. Stuff like:

- how long to go from cold start to a complete response? This rewards fast handshakes and code init time.
- sustained throughput of small requests with a lot of bandwidth but a fixed amount of memory. This rewards good pooling, minimizing GC, minimizing locking.

PR #4465, opened 2018-12-24
optimizing removeAll

shellhub:
remove header from top is faster


JakeWharton:
Er, is it? Both seem to be bad with an ArrayList.

shellhub:
I think it is better than before. I don't know how to improve

swankjesse:
As a rule, we only like optimizations motivated by real performance problems. This is a speculative fix for a speculative problem. I'm going to decline.

If you're suffering performance problems with OkHttp, please share benchmarks!

PR #4467, opened 2018-12-24
Add --http2-prior-knowledge option to okcurl

amirlivneh:


yschimke:
I started down the route of making okcurl more feature rich previously.  Got persuaded by @swankjesse that it wasn't really intended for that.

I've got a similar (brew installable) project https://github.com/yschimke/okurl that I use as a testbed for things like conscrypt or DoH support.

amirlivneh:
I see. I find it convenient to have okcurl in the same repo and able to exercise the library features, but I can try to use okurl instead.

yschimke:
@amirlivneh My point was less around that and more about not building okcurl inside the repo up with lots of features.

My other concern is whether it comes at the expense of good automated tests for okhttp-tests that cover all features.

Happy to defer to feedback from others, I don't have a strong opinion on this PR.

amirlivneh:
@yschimke, these are valid points. I've been using okcurl for reproducing issues and testing local changes against real servers, and was missing some of these options. I can achieve the same by making throwaway changes to okcurl or one of the samples. Customizing the request using throwaway Java code requires about the same effort as constructing a command line and can be more flexible.

I've sent this PR and the one adding --calll-timeout because these options seemed consistent with the ones already in okcurl.

I'm curious as to the intended purpose of okcurl and whether there is a rule of thumb for what goes into it.

Issue #4140, opened 2018-07-11
URLConnectionTest.testServerClosesOutput is flaky


swankjesse:
```
[ERROR] serverShutdownOutput(okhttp3.URLConnectionTest)  Time elapsed: 0.511 s  <<< FAILURE!
java.lang.AssertionError: expected:<0> but was:<1>
	at okhttp3.URLConnectionTest.testServerClosesOutput(URLConnectionTest.java:477)
	at okhttp3.URLConnectionTest.serverShutdownOutput(URLConnectionTest.java:434)
```

amirlivneh:
There are usually 2 requests sent to '/b' during the test:
#1 on the busted connection, attempting reuse
#2 on a new connection

On successful runs, the server records #1 and then #2 and adds them to the queue in that order.

On failed runs, the server starts reading request #1 before #2, but there is a context switch, and it finishes reading #2 before #1, recording them in that order. Since #1 is the last to be recorded, the test incorrectly uses it to assert that the sequence number is 0. Since #1 was the second to be received on the busted connection, it has a sequence number of 1 so the test fails.

A simple fix is to assert that either #1 or #2 has a sequence number of 0. I wonder if there is a cleaner way to assert that a new connection was used.

PR #4469, opened 2018-12-25
Fix flaky URLConnectionTest.serverShutdownOutput test

amirlivneh:
There are usually 2 requests sent to '/b' during the test:
#1 on the busted connection, attempting reuse
#2 on a new connection

On successful runs, the server recorded #1 and then #2 and added them to the queue in that order.

On failed runs, the server started reading request #1 before #2, but there was a context switch, and it finished reading #2 before #1, recording them in that order. Since #1 was the last to be recorded, the test incorrectly used it to assert that the sequence number was 0. Since #1 was the second to be received on the busted connection, it had a sequence number of 1 so the test failed.

The fix removes the assumption that #2 is the last to be recorded.

Fixes https://github.com/square/okhttp/issues/4140.


swankjesse:
Nice sleuthing!

Issue #4470, opened 2018-12-25
HttpOverHttp2Test.concurrentHttp2ConnectionsDeduplicated is flaky


amirlivneh:
```
java.lang.AssertionError: expected:<1> but was:<2>
	at okhttp3.internal.http2.HttpOverHttp2Test.concurrentHttp2ConnectionsDeduplicated(HttpOverHttp2Test.java:1422)
```

Example builds:
* https://travis-ci.org/square/okhttp/builds/471617400
* https://travis-ci.org/square/okhttp/builds/471293181

amirlivneh:
One scenario where the connection pool is not always cleared in `tearDown()` is after running `recoverFromMultipleCancelReusesConnection()`. This condition fails because `allocations` has 1 element:
https://github.com/square/okhttp/blob/5b37cda9e00626f43acf354df145fd452c3031f1/okhttp/src/main/java/okhttp3/ConnectionPool.java#L179
As a result, the connection is not removed from the pool and the next test starts with 1 connection in the pool.

PR #4478, opened 2018-12-28
Assert connection pool eviction in HttpOverHttp2Test.tearDown()

amirlivneh:
While investigating https://github.com/square/okhttp/issues/4470, I noticed there are sometimes connections left in the pool after calling ConnectionPool.evictAll() in HttpOverHttp2Test.tearDown(). As a result, tests that assume the connection pool is initially empty can appear flaky while the blame is in previous tests, making it hard to debug, and possibly hiding a connection leak bug.

So far I've noticed the new assertion fails (very rarely) when tearing down these tests:
* recoverFromMultipleCancelReusesConnection
* emptyDataFrameSentWithEmptyBody
* gzippedResponseBody

This change may increase flakiness in the short term, but will make it easier to debug, by blaming the right tests.


swankjesse:
Good call.

PR #4487, opened 2018-12-29
Fix NPE in HttpLoggingInterceptor when response has no body

flocsy:
When intercepting a WebSocket response.body() returns null

### DIFF
--- a/okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
+++ b/okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java
@@ -230,7 +230,7 @@ public final class HttpLoggingInterceptor implements Interceptor {
     long tookMs = TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - startNs);
 
     ResponseBody responseBody = response.body();
-    long contentLength = responseBody.contentLength();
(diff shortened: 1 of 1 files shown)


yschimke:
Is this linked to an observed issue? Or just good generally nullability handling?

flocsy:
> Is this linked to an observed issue? Or just good generally nullability handling?

It's from a real issue. I haven't open an issue in github though, but it happened in my real use-case:

        HttpLoggingInterceptor interceptor = new HttpLoggingInterceptor();
        interceptor.setLevel(HttpLoggingInterceptor.Level.BODY);

        client = getHttpClientBuilder(ignoreSslErrors)
                .readTimeout(0, TimeUnit.MILLISECONDS)
                .cookieJar(new JavaNetCookieJar(cookieManager))
                .addInterceptor(interceptor)
                .build();

I use this client everywhere in the app (it's ok for GET, POST requests), but in 1 place I use it like this:

        Request request = new Request.Builder().url(getBaseUrl() + "api/socket").build();
        webSocket = WebSocketCall.create(client, request);
        webSocket.enqueue(new WebSocketListener() {....}

All this code existed in github, I just added the interceptor for debugging reasons, and then I got NPE every time on the webSocket.enqueue()

JakeWharton:
Sounds like we need a test then

flocsy:
It seems that in the latest versions this doesn't happen, I was using okhttp-ws :)

PR #4488, opened 2018-12-29
Test call cancellation while request headers are sent

amirlivneh:
Reproduce an issue where canceling an HTTP/2 request while request headers are sent has no effect.

Other than being a real issue, this may also be the root cause of https://github.com/square/okhttp/issues/4470.

It seems to be related to the fact that the HTTP2Codec has no stream assigned at that point, so HTTP2Codec.cancel() is a no-op.


amirlivneh:
Based on @swankjesse's feedback, I think this can be merged as-is.

I was able to verify that this is the root cause for HttpOverHttp2Test.recoverFromMultipleCancelReusesConnection, and possibly other tests that cancel calls immediately, leaving a dirty connection pool to the next test. Once this lands, I'll submit a PR with a suggested fix.

Issue #4427, opened 2018-12-02
IndexOutOfBoundsException in RealConnection


andylizi:
#### Stacktrace
```
java.lang.IndexOutOfBoundsException: Index: 0
	at java.base/java.util.Collections$EmptyList.get(Collections.java:4481)
	at okhttp3.internal.connection.RealConnection.connectTls(RealConnection.java:326)
	at okhttp3.internal.connection.RealConnection.establishProtocol(RealConnection.java:283)
	at okhttp3.internal.connection.RealConnection.connect(RealConnection.java:168)
	at okhttp3.internal.connection.StreamAllocation.findConnection(StreamAllocation.java:257)
	at okhttp3.internal.connection.StreamAllocation.findHealthyConnection(StreamAllocation.java:135)
	at okhttp3.internal.connection.StreamAllocation.newStream(StreamAllocation.java:114)
	at okhttp3.internal.connection.ConnectInterceptor.intercept(ConnectInterceptor.java:42)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:147)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:121)
	at okhttp3.internal.cache.CacheInterceptor.intercept(CacheInterceptor.java:93)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:147)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:121)
	at okhttp3.internal.http.BridgeInterceptor.intercept(BridgeInterceptor.java:93)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:147)
	at okhttp3.internal.http.RetryAndFollowUpInterceptor.intercept(RetryAndFollowUpInterceptor.java:126)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:147)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:121)
	at okhttp3.RealCall.getResponseWithInterceptorChain(RealCall.java:254)
	at okhttp3.RealCall.execute(RealCall.java:92)
```

#### Environment
* OkHttp 3.12.0
* Java 11.0.1 LTS
* Windows 10 64-bit

This bug seems cannot be stably reproduce, so I'm not sure how to write a failing test for this. After I downgraded okhttp to 3.11.0 it works fine.

[...]

PR #4489, opened 2018-12-29
Exercise TLS with no server certificates

swankjesse:
Closes: https://github.com/square/okhttp/issues/4427

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/connection/RealConnection.java
+++ b/okhttp/src/main/java/okhttp3/internal/connection/RealConnection.java
@@ -26,6 +26,7 @@ import java.net.Socket;
 import java.net.SocketException;
 import java.net.SocketTimeoutException;
 import java.net.UnknownServiceException;
+import java.security.cert.Certificate;
 import java.security.cert.X509Certificate;
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


okhttp/src/main/java/okhttp3/internal/connection/RealConnection.java
```diff
@@ -323,11 +324,16 @@ private void connectTls(ConnectionSpecSelector connectionSpecSelector) throws IO
...
-            + "\n    subjectAltNames: " + OkHostnameVerifier.allSubjectAltNames(cert));
+        String message = "Hostname " + address.url().host() + " not verified";
+        List<Certificate> peerCertificates = unverifiedHandshake.peerCertificates();
+        if (!peerCertificates.isEmpty()) {
+          X509Certificate cert = (X509Certificate) peerCertificates.get(0);
+          message += ":"
+              + "\n    certificate: " + CertificatePinner.pin(cert)
+              + "\n    DN: " + cert.getSubjectDN().getName()
+              + "\n    subjectAltNames: " + OkHostnameVerifier.allSubjectAltNames(cert);
+        }
```
yschimke: I'd suggest appending something in the else case also, since it's a strange case.

swankjesse: Great idea. New message: `Hostname localhost not verified (no certificates)`.

Issue #4282, opened 2018-09-21
An explicitly set Cookie header is overwritten when using a CookieJar


nicktrav:
If a `Request` contains a `Cookie` header, it is overridden by the values returned by the `CookieJar` (if one has been installed).

This issue is half "bug report" in the sense that the principle of least surprise would have me reason that a cookie header set _explicitly_ on a request would be honored when using a `CookieJar`, and half "question" in the sense that maybe this is actually intended behavior?

Here's a test case to reproduce what I'm noticing: https://github.com/nicktrav/okhttp/commit/1cc51f7c33e0d110e85e97cc51528685a6c1a565

The following is a naive patch I'd be happy to offer up, though this change in behavior might break existing clients, so I'm open to thoughts / ideas.

```diff
--- a/okhttp/src/main/java/okhttp3/internal/http/BridgeInterceptor.java
+++ b/okhttp/src/main/java/okhttp3/internal/http/BridgeInterceptor.java
@@ -82,7 +82,7 @@ public final class BridgeInterceptor implements Interceptor {
     }

     List<Cookie> cookies = cookieJar.loadForRequest(userRequest.url());
-    if (!cookies.isEmpty()) {
+    if (!cookies.isEmpty() && userRequest.header("Cookie") == null) {
       requestBuilder.header("Cookie", cookieHeader(cookies));
     }
```

Thanks!

swankjesse:
We could union ’em. You’re allowed to have multiple `Cookie` headers. Want to submit a pull request?!

nicktrav:
My understanding is that you can have multiple values in a single `Cookie` header (i.e. you can't have multiple `Cookie` headers, as per [RFC 6265][0]):

> When the user agent generates an HTTP request, the user agent MUST
> NOT attach more than one Cookie header

A union of the cookie values sounds reasonable! There's still the question of which cookie value is authoritative if the cookie value is in the `CookieJar` as well as set on the `Request`. Taking the value set on the `Request` would solve the issue in our case.

@yschimke - feel free to assign this one to me. I'll submit a patch in the next couple days.

[0]: https://tools.ietf.org/html/rfc6265#section-5.4

yschimke:
I think @swankjesse is against "licking the cookie", If you are worried, then I suggest putting up a PR even if it's still WIP.  Not sure if there is some other process for square employees :)

swankjesse:
[...]

PR #4490, opened 2018-12-29
Concatenate cookies that come from competing sources.

swankjesse:
If the HTTP request and the cookie jar both provide cookies for a request,
concatenate them into a single cookie header.

Closes: https://github.com/square/okhttp/issues/4282

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/http/BridgeInterceptor.java
+++ b/okhttp/src/main/java/okhttp3/internal/http/BridgeInterceptor.java
@@ -26,10 +26,12 @@ import okhttp3.MediaType;
 import okhttp3.Request;
 import okhttp3.RequestBody;
 import okhttp3.Response;
+import okhttp3.internal.Util;
 import okhttp3.internal.Version;
 import okio.GzipSource;
 import okio.Okio;
 
+import static okhttp3.internal.Util.concat;
 import static okhttp3.internal.Util.hostHeader;
 
 /**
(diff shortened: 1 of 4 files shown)


okhttp/src/main/java/okhttp3/internal/http/BridgeInterceptor.java
```diff
@@ -81,9 +83,12 @@ public BridgeInterceptor(CookieJar cookieJar) {
       requestBuilder.header("Accept-Encoding", "gzip");
     }
 
-    List<Cookie> cookies = cookieJar.loadForRequest(userRequest.url());
-    if (!cookies.isEmpty()) {
-      requestBuilder.header("Cookie", cookieHeader(cookies));
+    List<Cookie> cookieJarCookies = cookieJar.loadForRequest(userRequest.url());
+    if (!cookieJarCookies.isEmpty()) {
+      // Concatenate cookies from the cookie jar and the user's request. If a cookie is in both the
+      // request and the cookie jar we send both! Well-behaved calling code shouldn't do this.
```
yschimke: What specifically means that well-behaved code shouldn't do this?  Curious, because it wasn't immediately clear to me that concatenating instead of taking the request cookies *over* cookie-jar cookies was the right approach.

swankjesse: I was thinking that if you’re manually providing cookies, it’s your responsibility to make sure those cookies don’t conflict with the cookie store.

But you’re right, I think it’s a bit of a bad use-case, and gets us into solving problems we don’t have enough information to solve. Might be simpler to keep our current policy, where if you have a `CookieJar` that’s where the cookies all come from.

@nicktrav can you help us here? Some context on how you ended up having two different things trying to supply cookies?

yschimke: I would have assumed the opposite, that putting cookies on the request was intentional to override the default behaviour.  But deferring to you.

swankjesse:
New decision: keep the current behavior. Recommend users either use a cookie jar or do their own thing. Living in both worlds means we have to resolve collisions and we’re not prepared to do that well.

yschimke:
That seems fair, but it's essentially

a) manual cookies, e.g. your client code uses Set-Cookie and Cookie headers manually
b) You behave like a user agent, the server sets cookies and you respect them

For other behaviours, you can use methods of CookieJar, or concrete implementations.  n.b. even that is awkward because CookieJar is based around URL whereas within a server cookies may be per user, and that might be based on other fields of the Request.  So fair enough.

Issue #4451, opened 2018-12-18
HttpUrl host decode IPv6 address issue


Yphyllis:
@Test
    public void test() {

        String host_ipv4 = "192.168.1.254";
        String host_ipv6 = "::ffff:c0a8:1fe";

        HttpUrl.Builder builder = new HttpUrl.Builder();
        builder.scheme("http");
        builder.host(host_ipv6);

        System.out.println(builder.build().toString());
    }

java.lang.AssertionError: Invalid IPv6 address: '::ffff:c0a8:1fe'

	at okhttp3.internal.Util.canonicalizeHost(Util.java:408)
	at okhttp3.HttpUrl$Builder.canonicalizeHost(HttpUrl.java:1592)
	at okhttp3.HttpUrl$Builder.host(HttpUrl.java:1026)
	at us.nonda.zus.TimestampTest.test(TimestampTest.java:22)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:483)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
[...]

PR #4491, opened 2018-12-29
Don't crash on IPv4-mapped IPv6 addresses

swankjesse:
https://github.com/square/okhttp/issues/4451

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/HttpUrlTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/HttpUrlTest.java
@@ -42,16 +42,16 @@ public final class HttpUrlTest {
         new Object[] { false }
     );
   }
-  
+
   @Parameterized.Parameter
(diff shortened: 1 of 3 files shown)


okhttp-tests/src/test/java/okhttp3/HttpUrlTest.java
```diff
@@ -554,10 +554,11 @@ HttpUrl parse(String url) {
     assertEquals("1::", parse("http://[1:0:0:0:0:0:0:0]/").host());
     assertEquals("::1", parse("http://[0:0:0:0:0:0:0:1]/").host());
     assertEquals("::", parse("http://[0:0:0:0:0:0:0:0]/").host());
+    assertEquals("192.168.1.254", parse("http://[::ffff:c0a8:1fe]/").host());
```
swankjesse: # 🤯

Issue #4457, opened 2018-12-20
Add default timeouts value to javadoc


s3r6:
Add actual values from `okhttp3.OkHttpClient.Builder#Builder()`
```
connectTimeout = 10_000;
      readTimeout = 10_000;
      writeTimeout = 10_000;
```
to `okhttp3.OkHttpClient.Builder#readTimeout` and other builders' timeout setter methods javadocs for not to search default values in source or internet.

PR #4502, opened 2018-12-30
Make sure default timeouts are documented explicitly.

swankjesse:
Closes: https://github.com/square/okhttp/issues/4457

### DIFF
--- a/okhttp/src/main/java/okhttp3/OkHttpClient.java
+++ b/okhttp/src/main/java/okhttp3/OkHttpClient.java
@@ -316,27 +316,30 @@ public class OkHttpClient implements Cloneable, Call.Factory, WebSocket.Factory 
     }
   }
 
-  /** Default call timeout (in milliseconds). */
+  /**
+   * Default call timeout (in milliseconds). By default there is no timeout for complete calls, but
+   * there is for the connect, write, and read actions within a call.
+   */
   public int callTimeoutMillis() {
     return callTimeout;
   }
 
-  /** Default connect timeout (in milliseconds). */
+  /** Default connect timeout (in milliseconds). The default is 10 seconds. */
   public int connectTimeoutMillis() {
     return connectTimeout;
   }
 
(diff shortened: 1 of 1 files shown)


okhttp/src/main/java/okhttp3/OkHttpClient.java
```diff
@@ -316,27 +316,27 @@ private static SSLSocketFactory newSslSocketFactory(X509TrustManager trustManage
...
   }
 
-  /** Default write timeout (in milliseconds). */
+  /** Default write timeout (in milliseconds). The default is 10 seconds. */
   public int writeTimeoutMillis() {
     return writeTimeout;
   }
 
-  /** Web socket ping interval (in milliseconds). */
+  /** Web socket and HTTP/2 ping interval (in milliseconds). By default no pings are sent. */
```
oldergod: super nit but 'no ping is sent' ?

okhttp/src/main/java/okhttp3/OkHttpClient.java
```diff
@@ -316,27 +316,27 @@ private static SSLSocketFactory newSslSocketFactory(X509TrustManager trustManage
     }
   }
 
-  /** Default call timeout (in milliseconds). */
+  /** Default call timeout (in milliseconds). By default calls have no timeout. */
```
JakeWharton: This should probably clarify that individual timeouts may still apply

okhttp/src/main/java/okhttp3/OkHttpClient.java
swankjesse: _By default pings are not sent_

okhttp/src/main/java/okhttp3/OkHttpClient.java
swankjesse: _Default call timeout (in milliseconds). By default there is no timeout for complete calls, but there is for the connect, write, and read actions within a call._

PR #4507, opened 2019-01-01
Get rid of some Java 7 stuff.

swankjesse:
Still more to do, but this is a start.

### DIFF
--- a/okhttp-urlconnection/src/main/java/okhttp3/internal/huc/DelegatingHttpsURLConnection.java
+++ b/okhttp-urlconnection/src/main/java/okhttp3/internal/huc/DelegatingHttpsURLConnection.java
@@ -147,7 +147,7 @@ abstract class DelegatingHttpsURLConnection extends HttpsURLConnection {
     return delegate.getContentLength();
   }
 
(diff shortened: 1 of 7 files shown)


okhttp-urlconnection/src/main/java/okhttp3/internal/huc/DelegatingHttpsURLConnection.java
```diff
@@ -147,7 +147,7 @@
     return delegate.getContentLength();
   }
 
-  @IgnoreJRERequirement // Should only be invoked on Java 7+.
+  @IgnoreJRERequirement // Should only be invoked on Java 7+ or Android API 24+.
```
JakeWharton: Remove Java 7 mention from these?

swankjesse: Changed to _Should only be invoked on Java 8+ or Android API 24+_

JakeWharton: I wouldn't even mention 8 because it's the baseline. You're only ignoring for Android's sake.

PR #4508, opened 2019-01-01
Drop code for Android 2.3 through 4.4.

swankjesse:
No more Gingerbread, Honeycomb, Ice Cream Sandwich, Jelly Bean or KitKat.


JakeWharton:
Nice to see the new years PRs are back!

okhttp/src/main/java/okhttp3/internal/tls/BasicCertificateChainCleaner.java
```diff
@@ -38,10 +44,20 @@
...
   private static final int MAX_SIGNERS = 9;
 
-  private final TrustRootIndex trustRootIndex;
+  /** A simple index that of trusted root certificates that have been loaded into memory. */
+  private final Map<X500Principal, Set<X509Certificate>> subjectToCaCerts;
 
-  public BasicCertificateChainCleaner(TrustRootIndex trustRootIndex) {
-    this.trustRootIndex = trustRootIndex;
+  public BasicCertificateChainCleaner(X509Certificate... caCerts) {
+    subjectToCaCerts = new LinkedHashMap<>();
```
swankjesse: Promoted from BasicTrustRootIndex.

okhttp/src/main/java/okhttp3/internal/tls/BasicCertificateChainCleaner.java
```diff
@@ -112,13 +128,31 @@ private boolean verifySignature(X509Certificate toVerify, X509Certificate signin
     }
   }
 
+  /** Returns the trusted CA certificate that signed {@code cert}. */
+  private X509Certificate findByIssuerAndSignature(X509Certificate cert) {
```
swankjesse: Same!

swankjesse:
screen shot 2019-01-01 at 12 02 39 am

yschimke:
When you update http://square.github.io/okhttp/ with requirements, can you call out the ongoing support for critical security updates for 3.12.X?  But that it won't be generally active.  It would be good to be able to point people to our documented policy (whatever specifically that is).

okhttp/src/main/java/okhttp3/internal/platform/AndroidPlatform.java
```diff
@@ -50,8 +46,6 @@
   private final Class<?> sslParametersClass;
```
yschimke: "Android 2.3 or better" above

yschimke:
Maybe worth logging a warning when this code runs on older platforms as guidance when a developer runs on a legacy device.  Is that easy to do?

okhttp/src/main/java/okhttp3/internal/platform/AndroidPlatform.java
swankjesse: Snuck in a fix here: https://github.com/square/okhttp/pull/4511

PR #4511, opened 2019-01-01
Run IntelliJ inspections on the codebase

swankjesse:
Some nullability warnings, some diamond operators, adopting Objects.equals(),
and some dead code elimination.


okhttp/src/main/java/okhttp3/internal/connection/ConnectionSpecSelector.java
```diff
@@ -124,9 +123,7 @@ public boolean connectionFailed(IOException e) {
 
     // On Android, SSLProtocolExceptions can be caused by TLS_FALLBACK_SCSV failures, which means we
     // retry those when we probably should not.
```
JakeWharton: Update or remove comment?

swankjesse: Updated.

PR #4513, opened 2019-01-01
Fix connection leak in DuplexTest

amirlivneh:


okhttp-tests/src/test/java/okhttp3/DuplexTest.java
```diff
@@ -89,6 +89,8 @@
     latchParty.step(6);
     assertTrue(source.exhausted());
     assertEquals(Headers.of("trailers", "boom"), response.trailers());
+
+    response.close();
```
yschimke: Is there a reason we don't use try-with-resources here?

amirlivneh: I haven't seen try-with-resources used in any of the tests. Perhaps it's because we expect a certain flow in a successful run, and don't want to clutter the test with code that handles anything other than that flow, which would be a test failure anyway.

I'm open to change it, if I'm misunderstanding.

yschimke: I suspect it might be fixed with the bump to Android 5+ 

https://stackoverflow.com/questions/24290830/try-with-resources-are-not-supported-at-this-language-level-android

amirlivneh: Updated to use try-with-resources.

PR #4521, opened 2019-01-02
Fix WebSocketRecorder logging

amirlivneh:
Both WebSocketRecorder.onClosing() and WebSocketRecorder.onClose() currently use the string "onClose":
https://github.com/square/okhttp/blob/775d5d9d5c6a210a49fb6ef5ebfcc47dcbb0865a/okhttp-tests/src/test/java/okhttp3/internal/ws/WebSocketRecorder.java#L88-L89

https://github.com/square/okhttp/blob/775d5d9d5c6a210a49fb6ef5ebfcc47dcbb0865a/okhttp-tests/src/test/java/okhttp3/internal/ws/WebSocketRecorder.java#L100-L101


oldergod:
maybe we can also rename the other one to `onClosed` to match method name.

amirlivneh:
Fixed

PR #4523, opened 2019-01-02
Fix retry infinite loop

feathj:
Infinite loop is created if a file IO exception is thrown on upload. This fixes that issue.


okhttp/src/main/java/okhttp3/internal/http/CallServerInterceptor.java
```diff
@@ -78,8 +78,18 @@ public CallServerInterceptor(boolean forWebSocket) {
...
 
-          request.body().writeTo(bufferedRequestBody);
-          bufferedRequestBody.close();
+          // if the requestBody is a file, this can throw an IOException
+          // which further up by the retry interceptor gets treated as a server
+          // IOException and assumes the socket is closed.
+          // We mitigate this by catching here, ensuring a close,
+          // and throwing
+          try {
+            request.body().writeTo(bufferedRequestBody);
```
swankjesse: Unfortunately this will catch problems where the bufferedRequestBody stream throws a retryable IOException. But you’re on the right track. I think we need to detec _those_ inside requestBodyOut or something

swankjesse:
Obsoleted by https://github.com/square/okhttp/pull/4635

Issue #4515, opened 2019-01-01
Most tests in WebSocketHttpTest are leaking connections


amirlivneh:
This can be reproduced by adding these lines to `WebSocketHttpTest.tearDown()`:
```
// Ensure a fresh connection pool for the next test.
client.connectionPool().evictAll();
assertEquals(0, client.connectionPool().connectionCount());
```

amirlivneh:
In some cases there is an actual issue in the code under test. For example, in `throwingOnOpenFailsImmediately`, an exception is thrown from `WebSocketListener.onOpen()`: https://github.com/square/okhttp/blob/32bf2f0fc10cc35e92d70a50436d84ef591a399a/okhttp-tests/src/test/java/okhttp3/internal/ws/WebSocketHttpTest.java#L150-L152

But `onOpen()` is called and throws before `initReaderAndWriter()`, which updates `this.streams` with the allocated streams, is called:
https://github.com/square/okhttp/blob/550afce3f855115bbb7de9b2080684dd8a6e1fe0/okhttp/src/main/java/okhttp3/internal/ws/RealWebSocket.java#L206-L212

As a result, the allocated streams are never released in `failWebSocket()`:
https://github.com/square/okhttp/blob/550afce3f855115bbb7de9b2080684dd8a6e1fe0/okhttp/src/main/java/okhttp3/internal/ws/RealWebSocket.java#L564-L564

yschimke:
Fixed

PR #4525, opened 2019-01-03
Fix connection leak if WebSocketListener.onOpen() throws

amirlivneh:
If WebSocketListener.onOpen() throws, failWebSocket() is responsible for releasing the allocated streams. Specifically it releases this.streams. Previously WebSocketListener.onOpen() was called after the streams were allocated but before they were assigned to this.streams. As a result the streams were not allocated if WebSocketListener.onOpen() threw.

Fixes part of https://github.com/square/okhttp/issues/4515.


swankjesse:
Nice find.

PR #4529, opened 2019-01-04
Add null-check on fail-websocket

alvince:


JakeWharton:
When would this happen? The listener is never null. Can you write a failing test case?

alvince:
> When would this happen? The listener is never null. Can you write a failing test case?

Yes, when I change listener null by reflection.
Crash happened sometimes.

swankjesse:
Change the listener to one that no-ops!

PR #4534, opened 2019-01-06
Test for and fix connection leaks in CallTest

amirlivneh:
Also removed calls to ConnectionPool.evictAll() that are now redundant.

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/CallTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/CallTest.java
@@ -117,6 +117,10 @@ public final class CallTest {
   @After public void tearDown() throws Exception {
     cache.delete();
     logger.removeHandler(logHandler);
+
+    // Ensure the test has released all connections.
+    client.connectionPool().evictAll();
+    assertEquals(0, client.connectionPool().connectionCount());
   }
 
   @Test public void get() throws Exception {
@@ -670,6 +674,8 @@ public final class CallTest {
     }
 
     assertEquals("SyncApiTest", server.takeRequest().getHeader("User-Agent"));
+
+    callback.await(request.url()).assertSuccessful();
   }
 
   @Test public void legalToExecuteTwiceCloning() throws Exception {
@@ -1671,7 +1677,6 @@ public final class CallTest {
         .assertBody("A");
 
     // Attempt conditional cache validation and a DNS miss.
-    client.connectionPool().evictAll();
     client = client.newBuilder()
         .dns(new FakeDns())
         .build();
@@ -3399,7 +3404,6 @@ public final class CallTest {
     server.enqueue(new MockResponse()
         .setResponseCode(401));
 
-    client.connectionPool().evictAll();
     client = client.newBuilder()
         .authenticator(new Authenticator() {
           @Override public Request authenticate(Route route, Response response) throws IOException {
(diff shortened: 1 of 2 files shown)


amirlivneh:
I still need to fix the failure in asyncCallEngineInitialized. It only reproduces in CI.

amirlivneh:
This is now ready for review.

okhttp-tests/src/test/java/okhttp3/CallTest.java
```diff
@@ -3560,6 +3563,8 @@ private RequestBody requestBody(final boolean chunked, final long size, final in
 
     assertTrue(source.exhausted());
     assertEquals(Headers.of("trailers", "boom"), response.trailers());
+
+    response.close();
```
yschimke: Should this be try with resources? Do you want everything closing cleanly even when tests fail?

okhttp-tests/src/test/java/okhttp3/ConnectionReuseTest.java
```diff
@@ -301,14 +301,22 @@
...
-        Response response = chain.proceed(chain.request());
-        return response.newBuilder()
-            .body(ResponseBody.create(null, "unrelated response body!"))
-            .build();
-      }
-    }).build();
+    client = client.newBuilder()
+        // Since this test knowingly leaks a connection, avoid using the default shared connection
+        // pool, which should remain clean for subsequent tests.
+        .connectionPool(new ConnectionPool())
```
yschimke: Is it worth cleaning up this pool also?

okhttp-tests/src/test/java/okhttp3/CallTest.java
amirlivneh: Fixed.

okhttp-tests/src/test/java/okhttp3/ConnectionReuseTest.java
amirlivneh: It does makes sense, but I'm not aware of a way to do this using ConnectionPool's public methods. Let me know if you can think of a way.

amirlivneh: @yschimke, any idea on how to do this?

amirlivneh: I'm not aware of a real-world problem that calls for adding such a public API, but I'm going to leave this outside the scope of this PR.

yschimke: Mainly being a good Android citizen.  It's a legitimate thing to want to be able to shutdown an App cleanly, close connections etc.  You may even want to do something like this on backgrounding an app to save battery etc.

amirlivneh: Should we open an issue for this?

yschimke: @swankjesse has always pushed back on these types of close methods etc, I think generally because the good usage patterns of client instances often reuse connection pools and dispatchers, and the close APIs can corrupt that pattern.

swankjesse: There is a way to shut down an idle HTTP client. But it's cumbersome.

I'm reluctant to make it easier because I don't want people to think of their OkHttp clients as things they need to stress about managing. And I really don't want people doing the obvious-looking thing and creating a new client for every call.

PR #4549, opened 2019-01-12
Use lambdas where appropriate

swankjesse:


swankjesse:
This causes Android studio to require the following configuration:

```
    compileOptions {
        sourceCompatibility JavaVersion.VERSION_1_8
        targetCompatibility JavaVersion.VERSION_1_8
    }
```

There is no problem running the compiled code on older versions of Android.

PR #4550, opened 2019-01-12
Core TLS features are now non-optional on Android

swankjesse:
With a baseline of 5.0+ we will have session tickets, SNI, and ALPN.


swankjesse:
```

    Caused by: java.lang.IllegalStateException: Expected Android API level 21+ but was 19
        at okhttp3.internal.platform.AndroidPlatform.buildIfSupported(AndroidPlatform.java:227)
        at okhttp3.internal.platform.Platform.findPlatform(Platform.java:200)
        at okhttp3.internal.platform.Platform.<clinit>(Platform.java:77)
```

swankjesse:
a) not as far as I know. I think Robolectric will always use a JDK platform.
b) I don’t think so. I tested on Android 4.4, 5.0, and 9.

Unfortunately to compile calls to new public APIs we need an update to [this thing](https://search.maven.org/search?q=a:android%20g:com.google.android). Not sure why that stops at Android 4.1.

JakeWharton:
It stops because it's not official. We deployed that artifact manually by
compiling Android ourselves and uploading it.

PR #4554, opened 2019-01-14
Added constants for WebSocket close code

PhilLab:
Things to check:

1. I could not determine where (if anywhere) constants are usually placed so I added a new class.

2. I copied the documentation from Java Websockets: 
https://github.com/TooTallNate/Java-WebSocket/blob/master/src/main/java/org/java_websocket/framing/CloseFrame.java . This is why I did (not yet) add the copyright and license to the beginning of WebSocketStatusCode.java

@TooTallNate As author of the awesome Java Websockets library, is it okay for you to place the file ``WebSocketStatusCode.java`` from this pull request under [okhttp's license](https://github.com/PhilLab/okhttp/blob/e82b0561c46eafbf99fc2d7bb5edc87232d7cd27/okhttp/src/main/java/okhttp3/WebSocket.java#L1-L15)? I figured that this should be okay as it is just a copy of the RFC plus some additions but I wanted to ask you first :-)

### DIFF
--- a/okhttp/src/main/java/okhttp3/WebSocketStatusCode.java
+++ b/okhttp/src/main/java/okhttp3/WebSocketStatusCode.java
@@ -1,0 +1,114 @@
+package okhttp3;
+
+/**
+ * Status code as defined by
+ * <a href="http://tools.ietf.org/html/rfc6455#section-7.4">
+ * Section 7.4 of RFC 6455
+ * </a>.
+ */
+public class WebSocketStatusCode {
+    /**
+     * indicates a normal closure, meaning whatever purpose the
+     * connection was established for has been fulfilled.
+     */
+    public static final int NORMAL = 1000;
+    /**
+     * 1001 indicates that an endpoint is "going away", such as a server
+     * going down, or a browser having navigated away from a page.
+     */
+    public static final int GOING_AWAY = 1001;
+    /**
+     * 1002 indicates that an endpoint is terminating the connection due
+     * to a protocol error.
+     */
+    public static final int PROTOCOL_ERROR = 1002;
+    /**
+     * 1003 indicates that an endpoint is terminating the connection
+     * because it has received a type of data it cannot accept (e.g. an
+     * endpoint that understands only text data MAY send this if it
+     * receives a binary message).
+     */
+    public static final int REFUSE = 1003;
+    /*1004: Reserved. The specific meaning might be defined in the future.*/
+    /**
+     * 1005 is a reserved value and MUST NOT be set as a status code in a
+     * Close control frame by an endpoint. It is designated for use in
+     * applications expecting a status code to indicate that no status
+     * code was actually present.
+     */
+    public static final int NOCODE = 1005;
+    /**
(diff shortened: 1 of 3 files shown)


swankjesse:
Please strip the documentation that was copied in. We don’t want to do open source that way.

okhttp/src/main/java/okhttp3/WebSocketStatusCode.java
```diff
@@ -0,0 +1,114 @@
+package okhttp3;
+
+/**
+ * Status code as defined by
+ * <a href="http://tools.ietf.org/html/rfc6455#section-7.4">
+ * Section 7.4 of RFC 6455
+ * </a>.
+ */
+public class WebSocketStatusCode {
```
PromanSEW: `final`?

JakeWharton: and private constructor, and two-space indent

swankjesse:
The right place for these constants is WebSocket.java.

Are all of these actually useful? Like, are the only interesting values 1000 and _other_ ? Which status codes are you using?

PhilLab:
@swankjesse We are using 1000 (Normal), 1002(PROTOCOL_ERROR) and 1003(REFUSE) right now.

swankjesse:
@PhilLab that helps, thanks! What do you do differently for PROTOCOL_ERROR vs. any other code?

PhilLab:
@swankjesse I my use case I handle 1003 like I would handle 404: the server (which is also under our control) does not offer me this socket. 1002 throws an assertion as a wrongly formatted message in my custom "protocol" should not happen and is a programming error.

After all, it boils down to which codes your server will send and if you can program the server yourself. To be honest, I could work with only 1000 and not-1000 in our use case as well. I don't know if others need more codes but after all the okhttp interface does return the code, so why not providing the constants for interpretation.

swankjesse:
I worry that if we give our users a dozen codes they’ll feel compelled to write code to treat each uniquely. Offering just 1000 / other might help guide users to keep their code simple.

PhilLab:
@swankjesse Yes, that is an important point. But on the other hand, the current interface does return the code as integer so why not giving the users the means to interpret it?

Proposal 1: Don't add any codes and dismiss this PR
Proposal 2: Add only ``NORMAL`` as constant to WebSocket.java
Proposal 3: Add a reduced set, I would suggest ``NORMAL=1000``, ``REFUSE=1003`` (as equivalent to the often used 404) and ``GOING_AWAY=1001`` which is sent when the server shuts down. Users like me who also use additional codes will then define them "manually"

Which proposal do you prefer I implement?

swankjesse:
I like proposal #2 best. If you can show me code that behaves differently for 1003 and other, or for 1004 and other, I can be convinced of more!
https://github.com/Tinder/Scarlet/blob/5c74eff525d4c42d7d072f621353ee1d38c184d0/scarlet-core/src/main/java/com/tinder/scarlet/ShutdownReason.kt

swankjesse:
Please reopen if you’re still into this!

PR #4565, opened 2019-01-17
Add tests that exercise real duplex communication

swankjesse:
We're currently broken because the Pipe doesn't flush. Will fix in Okio.


okhttp/src/main/java/okhttp3/internal/http2/Http2Stream.java
```diff
@@ -181,22 +181,21 @@ public synchronized ErrorCode getErrorCode() {
...
+      throws IOException {
     assert (!Thread.holdsLock(Http2Stream.this));
     if (responseHeaders == null) {
       throw new NullPointerException("headers == null");
     }
-    boolean outFinished = false;
-    boolean flushHeaders = false;
     synchronized (this) {
       this.hasResponseHeaders = true;
-      if (!out) {
```
swankjesse: This was too simple for duplex. We want to flush headers for duplex even though a body is forthcoming.

mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.java
```diff
@@ -1052,11 +1054,11 @@ private void writeResponse(final Http2Stream stream,
         throttledTransfer(response, socket, body, sink, body.size(), false);
         sink.close();
       } else if (response.isDuplex()) {
-        final BufferedSink sink = Okio.buffer(stream.getSink());
-        final BufferedSource source = Okio.buffer(stream.getSource());
-        final DuplexResponseBody duplexResponseBody = response.getDuplexResponseBody();
+        BufferedSink sink = Okio.buffer(stream.getSink());
```
oldergod: what there something wrong for them being `final`?

okhttp-tests/src/test/java/okhttp3/DuplexTest.java
```diff
@@ -45,6 +47,115 @@
...
+
+        responseBody.writeUtf8("response C\n");
+        responseBody.flush();
+        assertEquals("request D", requestBody.readUtf8Line());
+
+        responseBody.writeUtf8("response E\n");
+        responseBody.flush();
+        assertEquals("request F", requestBody.readUtf8Line());
+
+        responseBody.close();
```
oldergod: I thought asserting on non-test thread would be dangerous but you're safe because test don't pass if the response body doesn't close.

mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.java
swankjesse: `final` here won’t benefit the compiler and I claim that it harms readability. It’s using a lot of characters to say something that’s obvious.

okhttp-tests/src/test/java/okhttp3/DuplexTest.java
swankjesse: In general yes! But in this case we’re protected by the fact that the other thread won’t complete normally unless this one does.

PR #4566, opened 2019-01-18
Add Jetty-APLN profiles for Java 1.8.0_201 and 1.8.0_202

swankjesse:
https://github.com/jetty-project/jetty-alpn/blob/master/docs/version_mapping.properties


swankjesse:
Rebasing over this should fix the build break: https://github.com/square/okhttp/pull/4566

PR #4576, opened 2019-01-27
Allow users to set UTF-8 characters in Multipart Parts when they use addFormDataPart(...)

### LINKED ISSUE #4564: okhttp3 3.12.x parse Part BUG
Good bug reports include a failing test! Writing a test helps you to isolate and describe the problem, and it helps us to fix it fast. Bug reports without a failing test or reproduction steps are likely to be closed.

Here’s an example test to get you started.
https://gist.github.com/swankjesse/981fcae102f513eb13ed

# Version :
3.12.x

# Error description :
When the `MultipartBody.Part.createFormData(String name, @Nullable String filename, RequestBody body) ` method is called, if the file name contains Chinese, an IllegalArgumentException will be thrown to locate the calling chain:
```
MultipartBody.Part.createFormData(String name, @Nullable String filename, RequestBody body) 
Headers.of(String... namesAndValues)
Headers.checkValue(String value, String name)
```
The final code is as follows:
```java
  static void checkValue(String value, String name) {
    if (value == null) throw new NullPointerException("value for name " + name + " == null");
    for (int i = 0, length = value.length(); i < length; i++) {
      char c = value.charAt(i);
      if ((c <= '\u001f' && c != '\t') || c >= '\u007f') {
        throw new IllegalArgumentException(Util.format(
            "Unexpected char %#04x at %d in %s value: %s", (int) c, i, name, value));
      }
    }
  }
```
As shown above, this code does not handle Chinese correctly.

monkey-mas:
Closes #4564 

Currently the user is not allowed to set Non-ASCII characters to `filename` when calling [addFormDataPart(...)](https://github.com/square/okhttp/blob/parent-3.12.1/okhttp/src/main/java/okhttp3/MultipartBody.java#L323-L325). When `addFormDataPart(...)` calls [Part.createFormData(...)](https://github.com/square/okhttp/blob/parent-3.12.1/okhttp/src/main/java/okhttp3/MultipartBody.java#L247-L260), [Headers.of(...)](https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/Headers.java#L236), which rejects Non-ASCII characters, is executed. By employing `Headers.Builder` with [addUnsafeNonAscii(...)](https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/Headers.java#L332-L335) when calling `Part.createFormData(...)`, users can set Non-ASCII characters.

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/MultipartBodyTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/MultipartBodyTest.java
@@ -188,6 +188,10 @@ public final class MultipartBodyTest {
     Buffer buffer = new Buffer();
     body.writeTo(buffer);
     assertEquals(expected, buffer.readUtf8());
+    assertEquals(489, body.contentLength());
+    assertEquals("form-data; name=\"name1\"; filename=\"value.txt\"", body.part(0).headers.value(0));
+    assertEquals("form-data; name=\"name2\"; filename=\"零壱弐参.txt\"", body.part(1).headers.value(0));
+    assertEquals("form-data; name=\"name3\"; filename=\"Star Fox 零.txt\"", body.part(2).headers.value(0));
   }
 
   @Test public void stringEscapingIsWeird() throws Exception {
--- a/okhttp-tests/src/test/java/okhttp3/MultipartBodyTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/MultipartBodyTest.java
@@ -102,7 +102,7 @@ public final class MultipartBodyTest {
         + "--AaB03x\r\n"
         + "Content-Disposition: form-data; name=\"files\"\r\n"
         + "Content-Type: multipart/mixed; boundary=BbC04y\r\n"
-        + "Content-Length: 343\r\n"
+        + "Content-Length: 337\r\n"
         + "\r\n"
         + "--BbC04y\r\n"
         + "Content-Disposition: file; filename=\"file1.txt\"\r\n"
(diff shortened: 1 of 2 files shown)


okhttp-tests/src/test/java/okhttp3/MultipartBodyTest.java
```diff
@@ -131,27 +131,66 @@
                     RequestBody.create(
                         MediaType.get("text/plain"), "... contents of file1.txt ..."))
                 .addPart(
-                    Headers.of(
-                        "Content-Disposition", "file; filename=\"file2.gif\"",
-                        "Content-Transfer-Encoding", "binary"),
+                    new Headers.Builder()
+                        .addUnsafeNonAscii("Content-Disposition", "file; filename=\"file2€.gif\"")
```
yschimke: Can you move this out into a separate targeted test and leave the existing test unchanged?  The new test can be shorter and just focus on length and filename encoding.

monkey-mas: Yeah, I undid the change. Do you think I can even remove `expected` to do [assertEquals(expected, buffer.readUtf8())](https://github.com/square/okhttp/pull/4576/commits/a1182fc95d61fdefc9418afcb11729e6671eb439#diff-c0fbf53145399b63adc71321a0d0b864R190)?

swankjesse:
Thanks! Obsoleted by https://github.com/square/okhttp/pull/4584

PR #4577, opened 2019-01-27
Allow users to set the encoded value to `filename*` in MultiPart Parts

monkey-mas:
Cont. #4576 

Currently the user can only set `filename` in MultiPart. [RFC6266](https://tools.ietf.org/html/rfc6266#section-6) says that `filename*` field is used to set the encoding of `filename`. Moreover, as defined in [RFC5987](https://tools.ietf.org/html/rfc5987#section-3.2.2), we don't provide quotations to wrap the `filename*` value.

To improve user exprience, it'd be better to give this option to users.

Useful resources:
- [RFC6266 Appendix D.  Advice on Generating Content-Disposition Header Fields](https://tools.ietf.org/html/rfc6266#appendix-D)
- [Stackoverflow: How to encode the filename parameter of Content-Disposition header in HTTP?](https://stackoverflow.com/questions/93551/how-to-encode-the-filename-parameter-of-content-disposition-header-in-http)

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/MultipartBodyTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/MultipartBodyTest.java
@@ -102,7 +102,7 @@ public final class MultipartBodyTest {
         + "--AaB03x\r\n"
         + "Content-Disposition: form-data; name=\"files\"\r\n"
         + "Content-Type: multipart/mixed; boundary=BbC04y\r\n"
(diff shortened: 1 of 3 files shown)


swankjesse:
I believe this is addressed by https://github.com/square/okhttp/pull/4584

I haven’t been able to get any browsers (Chrome, Firefox, Safari) to upload files using the `filename*` header; it looks to me like that’s mostly used when files are downloaded.

PR #4584, opened 2019-02-02
Permit multipart filenames to contain non-ascii characters

swankjesse:
Closes: https://github.com/square/okhttp/issues/4564

### DIFF
--- a/mockwebserver/src/test/java/okhttp3/mockwebserver/RecordedRequestTest.java
+++ b/mockwebserver/src/test/java/okhttp3/mockwebserver/RecordedRequestTest.java
@@ -22,13 +22,14 @@ import java.net.Socket;
 import java.net.UnknownHostException;
 import java.util.Collections;
 import okhttp3.Headers;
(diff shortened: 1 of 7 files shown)


swankjesse:
And upon further investigation I’ve convinced myself that UTF-8 is safe. I just needed to fix my tests so that the form I was submitting from was declared with a proper UTF-8 HTML charset. With that in place I get UTF-8 filenames in Chrome, Firefox, and Safari.

Whew.

screen shot 2019-02-02 at 2 02 17 pm
[test.html.txt](https://github.com/square/okhttp/files/2824747/test.html.txt)
[uploaded_chrome.txt](https://github.com/square/okhttp/files/2824742/uploaded_chrome.txt)
[uploaded_firefox.txt](https://github.com/square/okhttp/files/2824743/uploaded_firefox.txt)
[uploaded_safari.txt](https://github.com/square/okhttp/files/2824744/uploaded_safari.txt)

PR #4585, opened 2019-02-03
Improve testing in MockDuplexResponseBody

swankjesse:


okhttp-tests/src/test/java/okhttp3/DuplexTest.java
```diff
@@ -46,45 +43,30 @@
...
+        new MockResponse()
+            .clearHeaders(),
+        new MockDuplexResponseBody()
+            .receiveRequest("request A\n")
+            .sendResponse("response B\n")
+            .receiveRequest("request C\n")
+            .sendResponse("response D\n")
+            .receiveRequest("request E\n")
+            .sendResponse("response F\n")
+            .exhaustRequest());
```
oldergod: neat

mockwebserver/src/main/java/okhttp3/mockwebserver/internal/duplex/MockDuplexResponseBody.java
```diff
@@ -0,0 +1,92 @@
...
+  @Override public void onRequest(RecordedRequest request, BufferedSource requestBody,
+      BufferedSink responseBody) {
+    FutureTask<Void> futureTask = new FutureTask<>(() -> {
+      for (Action action; (action = actions.poll()) != null; ) {
+        action.execute(request, requestBody, responseBody);
+      }
+      return null; // Success!
+    });
+    results.add(futureTask);
+    futureTask.run();
```
oldergod: How is it different from just using `Runnable`s?

oldergod: Used as a way to store exceptions they may throw

PR #4589, opened 2019-02-05
Fix some maven problems that are blocking the 3.13 release

swankjesse:


bom/pom.xml
```diff
@@ -74,4 +74,11 @@
     </dependencies>
   </dependencyManagement>
 
+  <distributionManagement>
+    <repository>
+      <id>sonatype-nexus-staging</id>
+      <name>Nexus Release Repository</name>
+      <url>https://oss.sonatype.org/service/local/staging/deploy/maven2/</url>
+    </repository>
+  </distributionManagement>
```
swankjesse: Stripping the parent means that Maven doesn’t know where to upload releases to. Ick.

pom.xml
```diff
@@ -213,13 +213,6 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-release-plugin</artifactId>
         <version>2.5.3</version>
-        <dependencies>
-          <dependency>
-            <groupId>org.apache.maven.scm</groupId>
-            <artifactId>maven-scm-provider-gitexe</artifactId>
-            <version>1.11.1</version>
-          </dependency>
-        </dependencies>
```
swankjesse: Not sure why we needed this, but by removing it we avoid this crash when tagging the release:

```
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-release-plugin:2.5.3:prepare (default-cli) on project parent: Execution default-cli of goal org.apache.maven.plugins:maven-release-plugin:2.5.3:prepare failed: An API incompatibility was encountered while executing org.apache.maven.plugins:maven-release-plugin:2.5.3:prepare: java.lang.NoSuchMethodError: org.apache.maven.scm.ScmTagParameters.isSign()Z
[ERROR] -----------------------------------------------------
[ERROR] realm =    plugin>org.apache.maven.plugins:maven-release-plugin:2.5.3
```

PR #4590, opened 2019-02-05
Drop the okhttp-bom module

swankjesse:
I think the benefit is unclear and it's already cost me 90 minutes of fighting
with Maven central.


swankjesse:
screen shot 2019-02-04 at 11 16 40 pm

swankjesse:
It might be possible to make the BOM work, but it’s not working now and I don’t have the patience to fight it any longer.

yschimke:
@swankjesse damn, just as gradle 5.2 adds better support.  Agree this is appropriate.  FWIW I've found the benefit of bom to be when using common libraries like jackson or netty and avaoiding mixing versions transitively.

swankjesse:
I think we gotta follow their structure, possibly making the parent project extend the bom vs. being independent?

I assume the parent pom can't be the the boom?

Issue #4592, opened 2019-02-05
OkHttp 3.13 on Android with custom trust manager fails with AssertionError


laalto:
TL;DR: How to use custom trust managers with OkHttp 3.13?

We have an app that uses a custom `X509TrustManager` (`X509ExtendedTrustManager` on API >= 24) to implement additional certificate whitelisting requirements. Basically the custom trust manager takes system trust manager as a delegate and adds custom whitelist checks on top.

This used to work fine up to OkHttp 3.12.1.

On OkHttp 3.13.0 this fails when initialising `OkHttpClient.Builder#sslSocketFactory` like:

```
        X509TrustManager trustManager;
        SSLSocketFactory sslSocketFactory;
        try {
            trustManager = buildTrustManager(...);
            SSLContext sslContext = SSLContext.getInstance("TLS");
            sslContext.init(null, new TrustManager[] { trustManager }, null);
            sslSocketFactory = sslContext.getSocketFactory();
        } catch (NoSuchAlgorithmException | KeyManagementException e) {
            throw new RuntimeException(e);
        }
        okClientBuilder.sslSocketFactory(sslSocketFactory, trustManager);
```

with stacktrace like:

```
java.lang.AssertionError: java.lang.reflect.InvocationTargetException
        at okhttp3.internal.platform.AndroidPlatform.buildCertificateChainCleaner(AndroidPlatform.java:201)
        at okhttp3.internal.tls.CertificateChainCleaner.get(CertificateChainCleaner.java:41)
        at okhttp3.OkHttpClient$Builder.sslSocketFactory(OkHttpClient.java:821)

...

     Caused by: java.lang.reflect.InvocationTargetException
        at java.lang.reflect.Constructor.newInstance0(Native Method)
        at java.lang.reflect.Constructor.newInstance(Constructor.java:343)
        at okhttp3.internal.platform.AndroidPlatform.buildCertificateChainCleaner(AndroidPlatform.java:196)
        at okhttp3.internal.tls.CertificateChainCleaner.get(CertificateChainCleaner.java:41) 
        at okhttp3.OkHttpClient$Builder.sslSocketFactory(OkHttpClient.java:821)  

...

Caused by: java.lang.IllegalArgumentException: Required method checkServerTrusted(X509Certificate[], String, String, String) missing
        at android.net.http.X509TrustManagerExtensions.<init>(X509TrustManagerExtensions.java:72)
        at java.lang.reflect.Constructor.newInstance0(Native Method) 
[...]

PR #4593, opened 2019-02-05
Restore support for custom trust managers on Android

swankjesse:
I took this away because I thought it was only useful for Android versions
we've since dropped. But it turns out it's also necessary when there's a
custom TrustManager in use.

Closes: https://github.com/square/okhttp/issues/4592

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/tls/BasicTrustRootIndex.java
+++ b/okhttp/src/main/java/okhttp3/internal/tls/BasicTrustRootIndex.java
@@ -1,0 +1,70 @@
+/*
+ * Copyright (C) 2016 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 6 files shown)


okhttp/src/main/java/okhttp3/internal/tls/BasicTrustRootIndex.java
```diff
@@ -0,0 +1,70 @@
+/*
+ * Copyright (C) 2016 Square, Inc.
```
oldergod: should this now be 2019?

okhttp/src/main/java/okhttp3/internal/tls/TrustRootIndex.java
```diff
@@ -0,0 +1,23 @@
+/*
+ * Copyright (C) 2016 Square, Inc.
```
oldergod: same, I just don't know

okhttp/src/main/java/okhttp3/internal/tls/BasicTrustRootIndex.java
swankjesse: nah, it’s the original date of the file.

PR #4605, opened 2019-02-11
Added StatusFamily enum to Response class.

aresok:
Hello,
I added StatusFamily enum to Response class.
Intension is to provide convenient method to use switch-case on status code family (class), instead of status arithmetic in client code.
StatusFamily#family() method has codes ordered with probability of particular code occurrence.

### DIFF
--- a/okhttp/src/main/java/okhttp3/Response.java
+++ b/okhttp/src/main/java/okhttp3/Response.java
@@ -289,6 +289,14 @@ public final class Response implements Closeable {
     body.close();
   }
 
+  /**
+   * Gets the family of status code for this response.
+   * @return family of status code
+   */
+  public StatusFamily statusFamily() {
+    return StatusFamily.family(code);
+  }
+
   @Override public String toString() {
     return "Response{protocol="
         + protocol
@@ -300,6 +308,53 @@ public final class Response implements Closeable {
         + request.url()
         + '}';
   }
+
+
+  /**
+   * Enum for status code class (family).
+   * It is intended to provide convenient method for switch-case branching on status class (family),
+   * instead of HTTP code calculations within client code.
+   */
+  public enum StatusFamily {
+    /** 1xx HTTP status codes. */
+    INFORMATIONAL,
+    /** 2xx HTTP status codes. */
+    SUCCESSFUL,
+    /** 3xx HTTP status codes. */
+    REDIRECTION,
+    /** 4xx HTTP status codes. */
+    CLIENT_ERROR,
+    /** 5xx HTTP status codes. */
+    SERVER_ERROR,
+    /** All other unrecognized HTTP status codes. */
+    OTHER;
+
+    /**
+     * Gets the family of particular http status code.
+     * @param httpCode http status code to check
+     * @return family of HTTP status
+     */
+    public static StatusFamily family(int httpCode) {
+      if (httpCode >= 200 && httpCode < 300) return SUCCESSFUL;
+      if (httpCode >= 400 && httpCode < 500) return CLIENT_ERROR;
+      if (httpCode >= 500 && httpCode < 600) return SERVER_ERROR;
+      if (httpCode >= 300 && httpCode < 400) return SERVER_ERROR;
+      if (httpCode >= 100 && httpCode < 200) return INFORMATIONAL;
+      return OTHER;
+    }
+
+    /**
+     * Checks if specified {@code httpCode} belongs to specified {@code family}.
+     * @param httpCode HTTP status code to check
+     * @param family family to check
+     * @return {@code true} if {@code httpCode} belongs to specified {@code family},
+     *         {@code false} otherwise.
+     */
+    public static boolean isFamily(int httpCode, StatusFamily family) {
+      return family(httpCode) == family;
+    }
+  }
+
 
   public static class Builder {
     @Nullable Request request;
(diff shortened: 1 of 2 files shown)


okhttp/src/main/java/okhttp3/Response.java
```diff
@@ -301,6 +309,53 @@ public long receivedResponseAtMillis() {
         + '}';
   }
 
+
+  /**
+   * Enum for status code class (family).
+   * It is intended to provide convenient method for switch-case branching on status class (family),
+   * instead of HTTP code calculations within client code.
+   */
+  public enum StatusFamily {
```
yschimke: Spec uses "class" https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html.

yschimke:
No strong opinion on this from me.  How do other libraries handle this?  I always assumed that OkHttp not providing constants for Method (PUT/POST etc) and Status code was deliberate, so interested in the thinking.

aresok:
Motivation behind is to be able to do sth like this:
```
switch(response.family()) {
  case StatusFamily.SUCCESS: return onSuccess(response);
  case StatusFamily.CLIENT_ERROR: return onClientError(response);
  default: return someOtherAction(response);
}
```
instead of, sth like this:
```
switch(response.code() / 100) {
  case 2: return onSuccess(response);
  case 4: return onClientError(response);
  default: return someOtherAction(response);
}
```

Spring for that uses [HttpStatus.Series](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/http/HttpStatus.Series.html).
javax.ws.rs.core has [Response.Status.Family](https://docs.oracle.com/javaee/7/api/javax/ws/rs/core/Response.Status.Family.html) - and I take name from here.

okhttp/src/main/java/okhttp3/Response.java
aresok: Yes, but "class" seems already be meaningful in java :)

swankjesse:
I assume the only interesting cases are 200s vs. non-200s. Are you doing something different for the other cases?

aresok:
If this were only about 200s vs non-200s, then it wouldn't be needed. In my case, I need to distinguish between: 2xx, 3xx, 4xx, 5xx & others.
It seems not to be something uncommon, as libraries like Spring are providing such kind of enums.

swankjesse:
Could you provide examples of different behavior based on status code class?

aresok:
I can't share commercial code, but basically it goes like this:
```
switch(response.code() / 100) {
  case 2: return onSuccess(response);
  case 3: return followUpOnRedirectAction(response);
  case 4: return onClientError(response.code());
  case 5: throw new ParticularServerError(response.code());
  default: throw new GenericHttpException(response.code());
}
```
I'm not sure what level of details you are asking for - is this enough, or you want to see actual business/technical/infrastructural case?

swankjesse:
What do these cases do differently? I'm trying to find out if you've got 5 cases when 2 is sufficient.

aresok:
Please, don't claim that 2 is sufficient, if you don't know the problem. I'll try to explain.

- On "2xx" - just follow the happy path, everything is great.
- On "3xx" logic needs to go to new location with the same request, and do some cashing with new location address, to avoid unnecessary redirections next time(s).
- On "4xx": the system is kind-of middleware, and is using some parts of query and credentials from other system (requestor) to craft a request to next system. If 'next system' answers with "4xx", requestor needs to get detailed information on what happened, to be able to fix its request, and eventually try again. Usually it will be token expiration, but not only that.
- On "5xx" requestor just need to learn on 'next system' error, and different logging for monitoring purposes is launched.
- "default" should never happen.

I hope this explains the use case.

swankjesse:
Does your system respond to 3xx codes? Which ones? OkHttp handles some of these automatically.

aresok:
Yes, it responds - mainly 302, but also 301.
I know, that by default OkHttp handles them, in order to add some custom logic it's turned off (with HttpURLConnection#setInstanceFollowRedirects).

Please let me know what's your opinion on the change I'm proposing.
If you agree, that it provides some value (for me it does), please accept. If not, please reject with some reasoning.

swankjesse:
> If you agree, that it provides some value (for me it does), please accept. If not, please reject with some reasoning.

As you wish.

APIs like this create checklists that developers may feel compelled to complete.

In my experience most HTTP calls have a small list of **codes** they care about, and everything else gets a log message and and exception: "unexpected code". You might handle 301 and 302 but not 304 or 399, and it's not particularly useful to handle 399 separately from 599; both codes are unexpected and suggest a server bug. Similarly you'll want to know about 401 but not 404 and its not generally useful to group them together.

The unexpected status class case for HTTP 100s (which are specified) and 600s (which don't) is yet another case where putting codes into buckets feels organized but doesn't do anything useful.

Most importantly I think it's actually not completely insane for you to just check the ranges 200..299, 300..399, etc in your code. Anyone who has done HTTP before will recognize it, and these ranges are not likely to change. Creating our own names and symbols for the ranges feels like it's just more to learn.

Lastly I think a very good use case of OkHttp is behind another abstraction like Retrofit or Picasso or a hypothetical MyCompanyApiClient. These layers are where the status classes go because they know which ones are relevant to them. You're special for disabling automatic redirects in yours, for example.

PR #4613, opened 2019-02-14
Prepare to fold StreamAllocation and Transmitter

swankjesse:
This is a mechanical refactoring that changes all code that directly
accesses a StreamAllocation to instead access a Transmitter. The
next step is to combine StreamAllocation and Transmitter into
a single class.


swankjesse:
Thanks for the quick review!

PR #4616, opened 2019-02-15
Fix Jdk8WithJettyBootPlatform loading ALPN class through normal classpath

### LINKED ISSUE #4615: Okhttp breaks jetty on Java 8 with ALPN enabled *dynamically*
Specifically, if this line is executed:

https://github.com/square/okhttp/blob/5ecd590c8c0cd7833ed378c8702d3a228817b4e8/okhttp/src/main/java/okhttp3/internal/platform/Jdk8WithJettyBootPlatform.java#L85

and that ALPN class is on the normal classpath, in addition to the boot classpath, then the class will be initialized from the class' ClassLoader, _not_ from the 'null' (system / bootstrap) classloader.

Then, Jetty will fail to initialize its `OpenJDK8ServerALPNProcessor` because the following check will fail:

https://github.com/eclipse/jetty.project/blob/f88f856673f2e8768042c973481ad78d41910e22/jetty-alpn/jetty-alpn-openjdk8-server/src/main/java/org/eclipse/jetty/alpn/openjdk8/server/OpenJDK8ServerALPNProcessor.java#L44-L45

We've usually hit this failure case when initializing okhttp clients in tests first (simply due to an unfortunate class load ordering), and only later loading the [jetty ALPN agent](https://github.com/jetty-project/jetty-alpn-agent) dynamically by calling a method [Http2Agent.install](https://github.com/palantir/conjure-java-runtime/blob/4f7e9c6428b356394771ef136f22bb4ad825877e/jetty-http2-agent/src/main/java/com/palantir/conjure/java/http2/Http2Agent.java#L33-L50) in a static initializer.

TODO:

- [x] test: https://github.com/square/okhttp/pull/4617

dansanduleac:
Fixes #4615 by ensuring that when loading the ALPN class, is can only be loaded through the bootstrap classloader.

I got this idea from what grpc-java was doing: 
https://github.com/grpc/grpc-java/blob/9a38dea91c6194f7a882e2be1c2846567e9cdcda/netty/src/main/java/io/grpc/netty/JettyTlsUtil.java#L64

I will follow-up with a currently failing test and link that onto the issue.

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/platform/Jdk8WithJettyBootPlatform.java
+++ b/okhttp/src/main/java/okhttp3/internal/platform/Jdk8WithJettyBootPlatform.java
@@ -82,10 +82,10 @@ class Jdk8WithJettyBootPlatform extends Platform {
     // Find Jetty's ALPN extension for OpenJDK.
     try {
       String alpnClassName = "org.eclipse.jetty.alpn.ALPN";
-      Class<?> alpnClass = Class.forName(alpnClassName);
-      Class<?> providerClass = Class.forName(alpnClassName + "$Provider");
-      Class<?> clientProviderClass = Class.forName(alpnClassName + "$ClientProvider");
-      Class<?> serverProviderClass = Class.forName(alpnClassName + "$ServerProvider");
+      Class<?> alpnClass = Class.forName(alpnClassName, true, null);
+      Class<?> providerClass = Class.forName(alpnClassName + "$Provider", true, null);
+      Class<?> clientProviderClass = Class.forName(alpnClassName + "$ClientProvider", true, null);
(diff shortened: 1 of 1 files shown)


dansanduleac:
Sadly I've tried merging this into #4617 but it still fails nonetheless..
It might be because in the test case, the ALPN class is not on the boot classpath _at all_ at the point where `Class.forName` is called, and it somehow can't avoid loading it through the wrong classloader.

In that case, I'm not exactly sure how to fix it at the moment.

yschimke:
This change itself makes sense, it seems worth landing and confirm with local testing that it doesn't break anything.  If ALPN is coming from a secondary classloader then we should probably drop back to the normal Platform.

yschimke:
@swankjesse should we land this?

swankjesse:
Yep! Thanks for the nudge.

swankjesse:
Thanks!

Issue #4603, opened 2019-02-11
Rewrite StreamAllocation + Events


swankjesse:
I’m unhappy with the way events are sent. It feels pretty haphazard, and we’re not reliably publishing all of the events we should be. For example, with 100-continue we don’t publish an endHeaders event.

I’m also unhappy with the way StreamAllocation interacts with duplex requests. It seems we’re likely to inadvertently leak when the request stream outlives the response stream.

I’d like to rewrite the StreamAllocation code as a new `Transmitter` class. It’s the bridge between OkHttp’s application layer stuff (built-in interceptors) and its network layer stuff (route planning, connection pooling, etc.). The transmitter should look kind of like a state machine and it should publish events on state transitions.

yschimke:
The test coverage pre-events and events specific is pretty good, so if you can run that gauntlet, you should be pretty safe.  :)

swankjesse:
Happy with where this landed. Thanks @yschimke, @oldergod, and @amirlivneh for quick code reviews!

PR #4620, opened 2019-02-15
Don't call callEnd() twice when application interceptors proceed twice.

swankjesse:
I'm working towards reducing the role StreamAllocation plays.

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
@@ -1139,4 +1139,33 @@ public final class EventListenerTest {
         "CallEnd");
     assertEquals(expectedEvents, listener.recordedEventTypes());
   }
+
+  @Test public void applicationInterceptorProceedsMultipleTimes() throws Exception {
+    server.enqueue(new MockResponse().setBody("a"));
+    server.enqueue(new MockResponse().setBody("b"));
+
+    client = client.newBuilder()
+        .addInterceptor(chain -> {
+          try (Response a = chain.proceed(chain.request())) {
+            assertEquals("a", a.body().string());
+          }
+          return chain.proceed(chain.request());
+        })
+        .build();
+
+    Call call = client.newCall(new Request.Builder().url(server.url("/")).build());
+    Response response = call.execute();
+    assertEquals("b", response.body().string());
+
+    List<String> expectedEvents = Arrays.asList("CallStart", "DnsStart", "DnsEnd",
+        "ConnectStart", "ConnectEnd", "ConnectionAcquired", "RequestHeadersStart",
+        "RequestHeadersEnd", "ResponseHeadersStart", "ResponseHeadersEnd", "ResponseBodyStart",
+        "ResponseBodyEnd", "RequestHeadersStart", "RequestHeadersEnd", "ResponseHeadersStart",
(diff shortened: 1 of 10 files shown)


okhttp/src/main/java/okhttp3/internal/http/RetryAndFollowUpInterceptor.java
```diff
@@ -70,13 +71,12 @@ public RetryAndFollowUpInterceptor(OkHttpClient client) {
     RealInterceptorChain realChain = (RealInterceptorChain) chain;
     Transmitter transmitter = realChain.transmitter();
 
-    transmitter.newStreamAllocation(request);
```
swankjesse: This PR has 2 interesting parts.

1. Replacing newStreamAllocation() with prepareToConnect(). The difference is prepareToConnect() works if there’s another stream already ready.
2. Moving the releaseStreamAllocation() calls out of here and into RealCall.

okhttp/src/main/java/okhttp3/internal/Transmitter.java
```diff
@@ -214,8 +230,8 @@ public boolean supportsUrl(HttpUrl url) {
     return streamAllocation.connection().supportsUrl(url);
   }
 
-  public void acquire(RealConnection connection, boolean reportedAcquired) {
-    streamAllocation.acquire(connection, reportedAcquired);
+  public void transmitterAcquireConnection(RealConnection connection, boolean reportedAcquired) {
```
oldergod: you prefixed it with `transmitter` for not to be confused with `streamAllocation`? Otherwise, since the class is `Transmitter` already, let's remove the prefix?

okhttp/src/main/java/okhttp3/ConnectionPool.java
```diff
@@ -110,17 +110,19 @@ public synchronized int connectionCount() {
   }
 
   /**
-   * Acquires a recycled connection to {@code address} for {@code streamAllocation}. If non-null
-   * {@code route} is the resolved route for a connection.
+   * Acquires a recycled connection to {@code address} for {@code transmitter}. If non-null {@code
+   * route} is the resolved route for a connection.
```
oldergod: Do you usually specify the returned parameter in comments as well? Like _returns true if was able to acquire a connection_ kind of thing

swankjesse: Yes, definitely. Lemme fix!

okhttp/src/main/java/okhttp3/internal/Transmitter.java
swankjesse: Fixed!

okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
```diff
@@ -1139,4 +1139,33 @@ public void redirectUsingNewConnectionEventSequence() throws IOException {
         "CallEnd");
     assertEquals(expectedEvents, listener.recordedEventTypes());
   }
+
+  @Test public void applicationInterceptorProceedsMultipleTimes() throws Exception {
```
yschimke: Any change to the currently ignored tests?

swankjesse: Will do a pass when things settle down.

PR #4621, opened 2019-02-15
Split ConnectionPool API from implementation

swankjesse:
Our 'Internal' trampoline really gets in the way with this one. Just
move the actual implementation code into its own type so there isn't
as much indirection.

Unfortunately we can't easily make RealConnectionPool extend from
ConnectionPool because otherwise we'd be open to any user-provided
connection pools and our API isn't that complete yet.

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/connection/RealConnectionPool.java
+++ b/okhttp/src/main/java/okhttp3/internal/connection/RealConnectionPool.java
@@ -1,0 +1,263 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one or more
+ *  contributor license agreements.  See the NOTICE file distributed with
+ *  this work for additional information regarding copyright ownership.
+ *  The ASF licenses this file to You under the Apache License, Version 2.0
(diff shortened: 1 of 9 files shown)


okhttp/src/main/java/okhttp3/internal/connection/RealConnectionPool.java
```diff
@@ -0,0 +1,263 @@
+/*
```
yschimke: What happened to the 

Copyright (C) 2019 Square, Inc.?

swankjesse: Oooh, I don’t know if you want to know just how old this class is. 2007!

http://svn.apache.org/viewvc/harmony/enhanced/java/trunk/classlib/modules/luni/src/main/java/org/apache/harmony/luni/internal/net/www/protocol/http/HttpConnectionManager.java?revision=552679&view=markup#l99

yschimke: "Son, I've got Util.java classes older than you."

PR #4623, opened 2019-02-15
Move connection.noNewStreams into the individual streams

swankjesse:


okhttp/src/main/java/okhttp3/internal/connection/RealConnection.java
```diff
@@ -99,7 +99,10 @@
 
   // The fields below track connection state and are guarded by connectionPool.
 
-  /** If true, no new streams can be created on this connection. Once true this is always true. */
+  /**
+   * If true, no new streams can be created on this connection. Once true this is always true.
+   * Guarded by {@link #connectionPool}.
```
oldergod: side-talk: It's kind of surprising that those unidirectional mutable variable don't have a type yet.

PR #4624, opened 2019-02-16
Push more StreamAllocation state into Connection and HttpCodec

swankjesse:


okhttp-tests/src/test/java/okhttp3/EventListenerTest.java
```diff
@@ -152,6 +152,8 @@
     assertEquals(expectedEvents, listener.recordedEventTypes());
   }
 
+  /** We incorrectly report a CallEnd event with the CallFailed event. */
+  @Ignore
```
swankjesse: This is a behavior regression! I’m attempting to split up StreamAllocation so that the state local to the codec lives there, and until that’s done I don’t mind breaking very subtle things like this!

okhttp/src/main/java/okhttp3/internal/connection/StreamAllocation.java
```diff
@@ -402,16 +371,15 @@ public void cancel() {
   }
 
   public void streamFailed(@Nullable IOException e) {
-    Socket socket;
-    Connection releasedConnection;
-
     synchronized (connectionPool) {
+      if (released) throw new IllegalStateException();
```
yschimke: Won't this likely hide the triggering exception?

swankjesse: I should have used AssertionError. It's just a sanity check to confirm this is called before the connection is released.

PR #4626, opened 2019-02-16
Manage releasing connection separately from HTTP framing

swankjesse:
This creates a new class ResponseBodySource that manages events and
connections. It's symmetric with our existing RequestBodySink class.

Continuing to work towards these things having their own lifecycles
that are separate from HTTP framing mechanics.


swankjesse:
Thanks! Definitely having a huge test suite is a big win here, especially tests that confirm connections are properly pooled.

PR #4627, opened 2019-02-16
Handle stream exceptions in Transmitter

swankjesse:
This changes failure recovery to only attempt when the failure
occurs when OkHttp is doing I/O. If a failure is triggered by
something else (such as a bad RequestBody or Interceptor) then
that failure will not be retried.

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/connection/RealConnection.java
+++ b/okhttp/src/main/java/okhttp3/internal/connection/RealConnection.java
@@ -30,7 +30,6 @@ import java.security.cert.Certificate;
 import java.security.cert.X509Certificate;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.concurrent.TimeUnit;
 import javax.annotation.Nullable;
 import javax.net.ssl.SSLPeerUnverifiedException;
 import javax.net.ssl.SSLSession;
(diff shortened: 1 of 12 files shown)


okhttp/src/main/java/okhttp3/internal/http/RetryAndFollowUpInterceptor.java
```diff
@@ -153,8 +151,6 @@ public RetryAndFollowUpInterceptor(OkHttpClient client) {
    */
   private boolean recover(IOException e, Transmitter transmitter,
       boolean requestSendStarted, Request userRequest) {
-    transmitter.streamFailed(e);
```
swankjesse: This is the most important part of the PR. We’re now tracking failures where they’re thrown, not far away in interceptors.

swankjesse:
This has been a pretty huge refactoring. I’ve been rearranging responsibilities between connections, stream allocations, codecs, and interceptors.

I think I’m near the turning point, where the types that are most semantic (RealConnectionPool, RealConnection, HttpCodec) can carry most of the state.

Next steps:
 * rename HttpCodec to HttpExchange. (Better names?)
 * rename StreamAllocation to HttpExchangeFactory
 * move state not related to creating HttpExchange values into Transmitter
 * make Transmitter work like a state machine

okhttp/src/main/java/okhttp3/internal/connection/RealConnection.java
```diff
@@ -105,6 +104,12 @@
    */
   public boolean noNewStreams;
 
+  /**
+   * If true there was a problem establishing this connection that may be due to route chosen. Once
```
yschimke: Do you mean "If non-zero"?

yschimke:
It's almost like there was some planning behind internal packages and great test coverage here

okhttp/src/main/java/okhttp3/internal/connection/RealConnection.java
swankjesse: I sure do! Fixed!

PR #4628, opened 2019-02-17
Make connection coalescing less special.

swankjesse:
Today we have to suppress 'connectionAcquired' events because we attach the
constructed connection and then replace it with the pooled one. This changes
the logic to use separate fields for the connecting connection vs. the
pooled one.

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/connection/RealConnection.java
+++ b/okhttp/src/main/java/okhttp3/internal/connection/RealConnection.java
@@ -471,7 +471,7 @@ public final class RealConnection extends Http2Connection.Listener implements Co
    * Returns true if this connection can carry a stream allocation to {@code address}. If non-null
    * {@code route} is the resolved route for a connection.
    */
-  public boolean isEligible(Address address, @Nullable Route route) {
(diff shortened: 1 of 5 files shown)


okhttp/src/main/java/okhttp3/internal/connection/RealConnection.java
```diff
@@ -494,13 +494,10 @@ public boolean isEligible(Address address, @Nullable Route route) {
     // 2. The routes must share an IP address. This requires us to have a DNS address for both
     // hosts, which only happens after route planning. We can't coalesce connections that use a
     // proxy, since proxies don't tell us the origin server's IP address.
-    if (route == null) return false;
-    if (route.proxy().type() != Proxy.Type.DIRECT) return false;
-    if (this.route.proxy().type() != Proxy.Type.DIRECT) return false;
-    if (!this.route.socketAddress().equals(route.socketAddress())) return false;
+    if (routes == null || !routeMatchesAny(routes)) return false;
```
yschimke: The comment above can probably move to the method, below.

swankjesse: Can fix in follow-up!

PR #4629, opened 2019-02-17
Introduce Exchange.

swankjesse:
This is the payoff that preceding changes have been working towards.
This splits StreamAllocation into three parts:

 - ExchangeFinder: strategy to allocate connections for exchanges
 - Exchange: a single request/response pair
 - Transmitter: a sequence of exchanges for a Call

The refactoring isn't totally complete, but this should be the last
big one.


okhttp/src/main/java/okhttp3/internal/Transmitter.java
```diff
@@ -16,55 +16,56 @@
...
 /**
  * Bridge between OkHttp's application and network layers. This class exposes high-level application
  * layer primitives: connections, requests, responses, and streams.
+ *
+ * <p>This class supports {@linkplain #cancel asynchronous canceling}. This is intended to have the
+ * smallest blast radius possible. If an HTTP/2 stream is active, canceling will cancel that stream
+ * but not the other streams sharing its connection. But if the TLS handshake is still in progress
+ * then canceling may break the entire connection.
  */
 public final class Transmitter {
```
swankjesse: Make sure to load this diff!

okhttp/src/main/java/okhttp3/internal/Transmitter.java
```diff
@@ -113,283 +114,161 @@ private Address createAddress(HttpUrl url) {
...
+    }
   }
 
-  public void responseBodyComplete(long bytesRead, IOException e) {
-    if (e != null) {
-      streamAllocation.streamFailed(e);
-    }
-    if (streamAllocation != null) {
-      streamAllocation.responseBodyComplete(bytesRead, e);
+  public void exchangeDone(@Nullable IOException e) {
```
swankjesse: FYI @oldergod this is the method we didn’t have before that we really needed. With duplex we’ll be able to have the request stream end after the response stream ends.

okhttp/src/main/java/okhttp3/internal/http/RealInterceptorChain.java
```diff
@@ -92,14 +92,19 @@ public RealInterceptorChain(List<Interceptor> interceptors, Transmitter transmit
...
+    return new RealInterceptorChain(interceptors, transmitter, exchange, index, request, call,
         connectTimeout, readTimeout, millis);
   }
 
   public Transmitter transmitter() {
     return transmitter;
   }
 
+  public Exchange exchange() {
+    if (exchange == null) throw new IllegalStateException();
```
oldergod: that's weird, isn't it?
Field is final and nullable but you don't allow access to it in some cases? How can the caller know it's good to call it?

swankjesse:
@oldergod in both cases the exceptions are just defensive programming for use of internal APIs. As far as I know these can’t happen in practice!

These could have just as easily been AssertionErrors, though the difference between AssertionError and IllegalStateException is negligible when it’s an internal caller.

PR #4630, opened 2019-02-17
Restore tests that now report the correct events.

swankjesse:
The change to introduce Transmitter and Exchange has caused us to report
the correct events. Yay!


amirlivneh:
Great job on rearchitecting this!

PR #4641, opened 2019-02-22
EventListener supports response overlasting request

oldergod:


okhttp/src/main/java/okhttp3/internal/connection/Transmitter.java
```diff
@@ -123,6 +125,8 @@ Exchange newExchange(Interceptor.Chain chain, boolean doExtensiveHealthChecks) {
 
     synchronized (connectionPool) {
       this.exchange = result;
+      this.exchangeRequestDone = false;
+      this.exchangeResponseDone = false;
```
oldergod: those two are needed when redirection happens, because they have to be bound to the exchange lifecycle, not the transmitter.

okhttp/src/main/java/okhttp3/internal/connection/Exchange.java
```diff
@@ -152,16 +152,27 @@ void trackFailure(IOException e) {
     codec.connection().trackFailure(e);
   }
 
-  void responseBodyComplete(long bytesRead, @Nullable IOException e) {
+  void bodyComplete(
+      long bytesRead, boolean responseDone, boolean requestDone, @Nullable IOException e) {
```
oldergod: changed the params order to match `exchangeMessageDone`'s

PR #4642, opened 2019-02-22
Crashes when duplex meets http1

oldergod:


okhttp/src/main/java/okhttp3/internal/http1/Http1ExchangeCodec.java
```diff
@@ -102,7 +103,11 @@ public Http1ExchangeCodec(OkHttpClient client, RealConnection realConnection, Bu
     return realConnection;
   }
 
-  @Override public Sink createRequestBody(Request request, long contentLength) {
+  @Override public Sink createRequestBody(Request request, long contentLength) throws IOException {
+    if (request.body() instanceof DuplexRequestBody) {
+      throw new ProtocolException("Duplex connections are not supported for HTTP1");
```
swankjesse: Supernit: HTTP1 » HTTP/1

oldergod: why don't you let me shake that off?

Issue #1605, opened 2015-05-05
Flexible Cache / Caching Policy


snodnipper:
It would be great if devs had a cache interface so they could define a custom cache implementation. @swankjesse suggested this requirement may be better served (or certainly complemented) by a caching policy and asked for a separate issue to be raised.

Some use cases where a custom cache would be useful:
- offline mode: user selects certain content (URLs) to be available offline - which overrides any default cache eviction behaviour.  This could be useful for media content etc. _think Spotify’s offline mode for specific content_.
  - I have had some success by abstracting two caches (offline/steaming) with deferent properties.  The offline bucket could only be evicted by the user whereas streaming behaved normally.
- abstracting multiple storage locations: I have been working on a project where users have cached content directories of several GiB, sometimes on different volumes (which can be unmounted etc.).  I have written a cache wrapper to ensure the desired behaviour.
- sensitive data: some content requires encryption, others do not.

My immediate thoughts are that a contract similar to the following could be useful.

```
DataProvider
void PUT(String url, byte[] bytes);
byte[] GET(String url);
```

Without too much thought I can imagine a chain of responsibility pattern for caching: MemoryDataProvider -> DiskDataProvider -> NetworkDataProvider

Related:
https://github.com/square/okhttp/issues/1021
https://github.com/square/okhttp/issues/1459

swankjesse:
If you'd like different storage locations, you can already do that by using two OkHttpClient instances, each configured with a different cache.

For offline mode, it's a little bit weird to use the HTTP cache to power that. In particular, because you need to be extremely careful around how you seed the cache, and what URLs your application attempts to fetch. Doing this in the application layer might be a better fit.

snodnipper:
agreed.  Different OkHttpClient instances certainly could get the job done - and potentially, the offline instance could simply have an extremely large cache size with headers to not expire anytime soon / accept anything in the bucket.  Cheers.

mandrachek:
The sensitive data bit is what I need. It would be so lovely to be able to transparently encrypt/decrypt cached responses on the fly (HIPAA stuff - has to be encrypted on the wire and "at rest").  

So it would be good if we could intercept the read and writes of cache entries (and keys!). At this point I'm stuck manually caching stuff. :(

swankjesse:
If you're feeling adventurous and want to depend on an unstable internal API, take a look at these two APIs:
[...]

PR #4643, opened 2019-02-23
Allow different types of response caches to be used

rrva:
Refactor so that you can provide your own custom response cache
instead of the the default filesystem-based cache.


rrva:
@swankjesse I'd like to implement a cache in redis. Will this be easy/stable across future okhttp upgrades when I only have the Filesystem interface?

rrva:
Also, the `File` interface is quite large and toilsome to implement

rrva:
Apache HttpClient has https://hc.apache.org/httpcomponents-client-4.5.x/httpclient-cache/apidocs/org/apache/http/client/cache/HttpCacheStorage.html, please consider creating something similar

swankjesse:
Do your want to support multiple concurrent OkHttp clients?

My long-stalled plan is to use [Relay](https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/internal/cache2/Relay.java) to allow a single client to make multiple concurrent requests for a single resource, and for it to fold those together. It’s intended to benefit HTTP/2 server push, where a cache write might be initiated by a server.

swankjesse:
Closing as a duplicate of #1605. Apache HTTP’s cache API is big! And we’ll have to do a lot of work to offer similar functionality.

PR #4645, opened 2019-02-23
Support follow-ups for duplex requests.

swankjesse:
https://github.com/square/okhttp/issues/4267

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/DuplexTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/DuplexTest.java
@@ -16,10 +16,12 @@
 package okhttp3;
 
 import java.io.IOException;
+import java.net.HttpURLConnection;
 import java.net.ProtocolException;
 import java.util.Arrays;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
+import okhttp3.internal.RecordingOkAuthenticator;
 import okhttp3.internal.duplex.AsyncRequestBody;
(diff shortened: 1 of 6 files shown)


okhttp-tests/src/test/java/okhttp3/DuplexTest.java
```diff
@@ -271,6 +273,116 @@
...
+    }
+
+    mockDuplexResponseBody.awaitSuccess();
+
+    List<String> expectedEvents = Arrays.asList("CallStart", "DnsStart", "DnsEnd", "ConnectStart",
+        "SecureConnectStart", "SecureConnectEnd", "ConnectEnd", "ConnectionAcquired",
+        "RequestHeadersStart", "RequestHeadersEnd", "RequestBodyStart", "ResponseHeadersStart",
+        "ResponseHeadersEnd", "ResponseBodyStart", "ResponseBodyEnd", "RequestHeadersStart",
+        "RequestHeadersEnd", "ResponseHeadersStart", "ResponseHeadersEnd", "ResponseBodyStart",
+        "ResponseBodyEnd", "ConnectionReleased", "CallEnd", "RequestBodyEnd");
```
yschimke: Are you happy with this ordering of CallEnd not being last?

swankjesse: No, definitely not. I’m gonna see if I can change things up so the event gets published when the stream gets detached. Follow up!

PR #4649, opened 2019-02-23
Introduce EventListener.requestFailed, responseFailed events

swankjesse:
These replace requestBodyEnd() / responseBodyEnd() in some failure scenarios.
They may also be issued in cases where no event was published previously.


okhttp-tests/src/test/java/okhttp3/DuplexTest.java
```diff
@@ -319,7 +319,7 @@
         "RequestHeadersStart", "RequestHeadersEnd", "RequestBodyStart", "ResponseHeadersStart",
         "ResponseHeadersEnd", "ResponseBodyStart", "ResponseBodyEnd", "RequestHeadersStart",
         "RequestHeadersEnd", "ResponseHeadersStart", "ResponseHeadersEnd", "ResponseBodyStart",
-        "ResponseBodyEnd", "ConnectionReleased", "CallEnd", "RequestBodyEnd");
+        "ResponseBodyEnd", "ConnectionReleased", "CallEnd", "RequestFailed");
```
swankjesse: FYI @yschimke ... slightly better already

Issue #4583, opened 2019-02-01
RealCall possible response leak?


Tolriq:
I'm trying to understand why I have a what seems impossible very rare  LeakedClosableViolation error when doing lots of call.enqueue cancellations.

I'm sure that I close everything everywhere possible.

While looking at OkHTTP source I saw this: 
https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/RealCall.java#L209

I don't know all the internals but it seems there can be a racy condition here. Where the response is correctly get but call is cancelled just before the isCanceled() test. Leading to the leak as the response is never closed since never passed back to the caller.

cc: @swankjesse 

StrictMode log:
```
2019-02-01 12:38:45.968 4693-4777/org.leetzone.android.yatsewidgetfree.debug D/StrictMode: StrictMode policy violation: android.os.strictmode.LeakedClosableViolation: A resource was acquired at attached stack trace but never released. See java.io.Closeable for information on avoiding resource leaks.
        at android.os.StrictMode$AndroidCloseGuardReporter.report(StrictMode.java:1786)
        at dalvik.system.CloseGuard.warnIfOpen(CloseGuard.java:264)
        at java.lang.reflect.Method.invoke(Native Method)
        at okhttp3.internal.platform.AndroidPlatform$CloseGuard.warnIfOpen(AndroidPlatform.java:355)
        at okhttp3.internal.platform.AndroidPlatform.logCloseableLeak(AndroidPlatform.java:162)
        at okhttp3.ConnectionPool.pruneAndGetAllocationCount(ConnectionPool.java:271)
        at okhttp3.ConnectionPool.cleanup(ConnectionPool.java:211)
        at okhttp3.ConnectionPool$1.run(ConnectionPool.java:60)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
        at java.lang.Thread.run(Thread.java:764)
     Caused by: java.lang.Throwable: Explicit termination method 'response.body().close()' not called
        at dalvik.system.CloseGuard.open(CloseGuard.java:221)
        at java.lang.reflect.Method.invoke(Native Method)
        at okhttp3.internal.platform.AndroidPlatform$CloseGuard.createAndOpen(AndroidPlatform.java:343)
[...]

PR #4651, opened 2019-02-24
Fix a bug where responses weren't closed on a late cancel (3.12.x branch)

swankjesse:
Closes: https://github.com/square/okhttp/issues/4583

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/CallTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/CallTest.java
@@ -73,6 +73,7 @@ import okhttp3.tls.HeldCertificate;
 import okio.Buffer;
 import okio.BufferedSink;
 import okio.BufferedSource;
+import okio.ForwardingSource;
 import okio.GzipSink;
 import okio.Okio;
 import org.junit.After;
@@ -3379,6 +3380,35 @@ public final class CallTest {
         .assertFailure(FileNotFoundException.class);
 
     assertEquals(1L, called.get());
+  }
+
+  /** https://github.com/square/okhttp/issues/4583 */
+  @Test public void lateCancelCallsOnFailure() throws Exception {
+    server.enqueue(new MockResponse()
+        .setBody("abc"));
+
+    final AtomicBoolean closed = new AtomicBoolean();
+
+    client = client.newBuilder()
+        .addInterceptor(new Interceptor() {
+          @Override public Response intercept(Chain chain) throws IOException {
+            Response response = chain.proceed(chain.request());
+            chain.call().cancel(); // Cancel after we have the response.
+            ForwardingSource closeTrackingSource = new ForwardingSource(response.body().source()) {
+              @Override public void close() throws IOException {
+                closed.set(true);
+                super.close();
+              }
+            };
+            return response.newBuilder()
+                .body(ResponseBody.create(null, -1L, Okio.buffer(closeTrackingSource)))
+                .build();
+          }
+        })
+        .build();
+
+    executeSynchronously("/").assertFailure("Canceled");
+    assertTrue(closed.get());
   }
 
   private void makeFailingCall() {
--- a/okhttp/src/main/java/okhttp3/RealCall.java
+++ b/okhttp/src/main/java/okhttp3/RealCall.java
@@ -35,6 +35,7 @@ import okio.AsyncTimeout;
 import okio.Timeout;
 
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
+import static okhttp3.internal.Util.closeQuietly;
 import static okhttp3.internal.platform.Platform.INFO;
 
 final class RealCall implements Call {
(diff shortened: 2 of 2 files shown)


Tolriq:
@swankjesse @yschimke  Beware that this way of fixing the issue is a breaking change. (Including the merged one for 3.14)

It's still totally possible to have another thread that cancels at https://github.com/square/okhttp/blob/bf394560e1c83d05a552079634359fad9e0a640b/okhttp/src/main/java/okhttp3/RealCall.java#L202 point.

Before it would have thrown, call failure and leaked the response, now it will call onResponse (so success) and ignore the cancel.

Personally I'm fine with the change as I find better to have a success call than a leak but I'm not sure this was wanted with that version of the fix.

swankjesse:
Needs to be rebased on this:
https://github.com/square/okhttp/pull/4652

swankjesse:
@Tolriq yep! You can also cancel a call while it’s being processed in `onResponse()`. When this happens the response body stream will throw an IOException, but the `onResponse()` implementation is still responsible for closing it.

Tolriq:
@swankjesse Yes I know and handle all the cases and found this issue, just pointing that the way of fixing it is a breaking change in some user cases. 
A late cancel would have called onFailure now it can call onReponse.

If a user expect that onResponse is not called for cancelled calls this is breaking change and may require proper warning in the changelog that now a check for call.isCanceled() might be wanted in the onResponse callback due to this change since it can now happens.

Tolriq:
Just a side note that changelog does not contains a comment about the fact that a cancelled call can reach onResponse now after those fixes. (3.14 and 3.12.x)

swankjesse:
@Tolriq it was already possible previously. The call is often canceled while reading the response body.

Tolriq:
@swankjesse I guess I badly explained in all previous comments so :)

The breaking change is that onResponse can be called with a call that is already cancelled. This was not possible before.
This change behaviors for API consumers, not all consumers want to actually use the body and will see an error if there's an error when using a body from a cancelled but successful call (I don't think so). And can start job / stuff that they are not supposed to do if the call was cancelled and never reached onResponse.

Furthermore the documentation says:
/** Cancels the request, if possible. Requests that are already complete cannot be canceled. */

So if onResponse is called a call.cancel() is said to do nothing. 

Now onResponse can be called with a call that is already cancelled and the consumer should not use the response, yet 99% of users certainly don't do call.isCancelled() check at the start of this function since it was not possible that is happened before.

Hope it's more clear why it's a breaking change and people should be aware, the race condition is now move to end user code.

swankjesse:
There’s no ordering guarantees between cancel() and onResponse(), and a call can be canceled immediately after an isCanceled() check returns false.

swankjesse:
Does this clarify? A call isn’t complete until the response body is exhausted or closed, which is also independent of when onResponse() is invoked.

Tolriq:
Then docs are not clear enough about what complete is.

And this also means that it's virtually impossible to prevent the leaks in coroutines world so :(

```
return suspendCancellableCoroutine { continuation ->
            continuation.invokeOnCancellation {
                cancel()
            }
            enqueue(object : Callback {
                    continuation.resume(WorkerResult.Success(response))
                }

                override fun onFailure(call: Call, e: IOException) {
                    if (call.isCanceled || !isActive) {
                        continuation.resume(WorkerResult.Error(CancellationException("Cancelled")))
                    } else {
                        continuation.resume(WorkerResult.Error(e))
                    }
                }
            })
        }
```

Can leak for example  coroutines 1.1.x ignores the resume when the call is cancelled and now with the race on client side the response is not closed :(
Maybe @JakeWharton have more idea about how to manage those races as retrofit coroutines will face that too.
Adding the isActive isCancelled check can face the race condition in onResponse,

JakeWharton:
Retrofit doesn't have this problem because it doesn't propagate an open
resource. Seems like any coroutine that resumes with a resource will have
this problem. Is there anything specific to OkHttp?

swankjesse:
Can you register a different cancelation handler in `onResponse()` ?

Tolriq:
@JakeWharton you do the exact same thing:
https://github.com/square/retrofit/blob/master/retrofit/src/main/java/retrofit2/KotlinExtensions.kt#L69
and
https://github.com/square/retrofit/blob/master/retrofit/src/main/java/retrofit2/KotlinExtensions.kt#L89

Don't know how those extensions method are used currently but they present the same issue :(

@swankjesse not sure to understand your question :(

JakeWharton:
Those aren't resources. They don't need closed. They're values.

Tolriq:
Oh sorry, long time not used Retrofit, did not notice the Call is the same interface but not same API :(.

Will move this to coroutines support so.

Tolriq:
See https://github.com/Kotlin/kotlinx.coroutines/issues/1044 for the coroutine part and design thinking by Roman Elizarov, maybe you have ideas to share :)

PR #4653, opened 2019-02-24
Fix the name of a TLSv1.3 cipher suite, TLS_AES_128_CCM_8_SHA256

swankjesse:
I think we made either a transcription error or grabbed a suite
from an earlier version of the spec.

### DIFF
--- a/okhttp/src/main/java/okhttp3/CipherSuite.java
+++ b/okhttp/src/main/java/okhttp3/CipherSuite.java
@@ -398,7 +398,7 @@ public final class CipherSuite {
   public static final CipherSuite TLS_AES_256_GCM_SHA384 = init("TLS_AES_256_GCM_SHA384", 0x1302);
   public static final CipherSuite TLS_CHACHA20_POLY1305_SHA256 = init("TLS_CHACHA20_POLY1305_SHA256", 0x1303);
   public static final CipherSuite TLS_AES_128_CCM_SHA256 = init("TLS_AES_128_CCM_SHA256", 0x1304);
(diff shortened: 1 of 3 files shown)


okhttp/src/main/java/okhttp3/CipherSuite.java
```diff
@@ -398,7 +398,7 @@
   public static final CipherSuite TLS_AES_256_GCM_SHA384 = init("TLS_AES_256_GCM_SHA384", 0x1302);
   public static final CipherSuite TLS_CHACHA20_POLY1305_SHA256 = init("TLS_CHACHA20_POLY1305_SHA256", 0x1303);
   public static final CipherSuite TLS_AES_128_CCM_SHA256 = init("TLS_AES_128_CCM_SHA256", 0x1304);
-  public static final CipherSuite TLS_AES_256_CCM_8_SHA256 = init("TLS_AES_256_CCM_8_SHA256", 0x1305);
+  public static final CipherSuite TLS_AES_128_CCM_8_SHA256 = init("TLS_AES_128_CCM_8_SHA256", 0x1305);
```
swankjesse: This is a backwards-incompatible API change. But it would never have worked previously.

swankjesse: https://tools.ietf.org/html/draft-ietf-tls-tls13-28#appendix-B.4

swankjesse: https://tools.ietf.org/html/rfc8446#appendix-B.4

yschimke:
Assuming my mistake

swankjesse:
We would have found it earlier if it was getting any use!

Issue #3865, opened 2018-02-19
Update cipher suites for February 2019


swankjesse:
We most recently updated the cipher suites in February 2018. We should update 'em again in February 2019.

When that's done, please open another tracking bug to update cipher suites in February 2020!

swankjesse:
One recommendation from an infosec person at Square:

> “If I were picking the ciphers, I would drop the SHA1 ciphers (#6, #7, #10, #11) and the 3DES one too (#12).”

swankjesse:
Next: https://github.com/square/okhttp/issues/4655

PR #4654, opened 2019-02-24
Remove default cipher suites: TLS_AES_128_CCM_SHA256, TLS_AES_128_CCM_8_SHA256

swankjesse:
As far as I know we never would have used these anyway, as they're
not available on Android, Java, or Conscrypt.

They aren't enabled by default in either Firefox or Chrome.

Closes: https://github.com/square/okhttp/issues/3865


swankjesse:
https://tinyurl.com/okhttp-cipher-suites

PR #4656, opened 2019-02-25
Update the public suffixes list

swankjesse:


swankjesse:
FYI: diff looks good: https://github.com/square/okhttp/commit/5fc67f78c6f24727e45edf334be67fb7d7f41f06

Generate before & after like so:
```
gunzip okhttp/src/main/resources/okhttp3/internal/publicsuffix/publicsuffixes.gz
cat okhttp/src/main/resources/okhttp3/internal/publicsuffix/publicsuffixes | strings > okhttp/src/main/resources/okhttp3/internal/publicsuffix/publicsuffixes.txt
```

PR #4660, opened 2019-02-25
Include 'timeout' in exceptions when that's the root cause

swankjesse:
I'm not particularly happy with the way this code looks, but I like
the result. Otherwise exceptions are like these:

  * stream was reset: CANCEL
  * Socket closed
  * Canceled

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/connection/Exchange.java
+++ b/okhttp/src/main/java/okhttp3/internal/connection/Exchange.java
@@ -166,7 +166,7 @@ public final class Exchange {
     codec.connection().trackFailure(e);
   }
 
-  void bodyComplete(
+  @Nullable IOException bodyComplete(
       long bytesRead, boolean responseDone, boolean requestDone, @Nullable IOException e) {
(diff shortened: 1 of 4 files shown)


okhttp/src/main/java/okhttp3/internal/connection/Exchange.java
```diff
@@ -215,17 +215,15 @@ public void noRequestBody() {
         super.write(source, byteCount);
         this.bytesReceived += byteCount;
       } catch (IOException e) {
-        complete(e);
-        throw e;
+        throw complete(e);
```
yschimke: This doesn't read quite right.  I can see what it's doing, and it looks correct, just feels a bit off.

Even reading into exchangeMessageDone there is a lot of assumptions about the nullability of e

swankjesse: One invariant we do hold: if the parameter in is non-null, the return value is non-null.

Still, if you’d like to take a pass at tidying this up please do!

Issue #95, opened 2013-01-31
Domain name lookup doesn't support cancel or timeout


swankjesse:
We might want to work-around this with helper threads or interrupts, or something.
See also https://code.google.com/p/android/issues/detail?id=40874

DanielNovak:
Any news on this? It would be great if we could set a timeout for the DNS lookup. Typical use case is when a user connects to a WiFi that does not have an internet connection. It takes 20-30 seconds until the request fails and there is no way how to reduce this time.

rtzoeller:
We are also running into issues with this. Do you know if any progress has been made on a workaround?

swankjesse:
Nothing at the moment. It's a difficult bug to fix because DNS is completely opaque in Java.

czfyzy:
see:http://stackoverflow.com/questions/693997/how-to-set-httpresponse-timeout-for-android-in-java/31643186#31643186

swankjesse:
Maybe we do DNS on a background thread? That’d probably be worth the cost, although potentially surprising.

swankjesse:
One more thing: now that our `Dns` interface is public, end-users can do this if they like.

yschimke:
n.b. This is relevant for dnsoverhttps and there are also other viable dns clients now e.g. Netty impl is straightforward and can support timeouts.  It would be nice if this was part of an overall request timeout even for a synchronous call.execute

https://github.com/yschimke/oksocial/blob/master/src/main/kotlin/com/baulsupp/oksocial/network/NettyDns.kt

swankjesse:
I was pondering the right API for this. Is it DnsCall + DnsCallback interfaces? That seems like a lot of code.

Another option is to make the DNS method async with coroutines, but that means we need a Kotlin dependency.

Other ideas?

yschimke:
Punt until there is a story for a (future?) kotlin based OkHttp, hopefully with native support for coroutines?

galderz:
Hi, I've been working on this. I've created a PR (#4674) to get some feedback. Thx

yschimke:
Dupe of https://github.com/square/okhttp/issues/5042?

swankjesse:
Yep, these are mutually redundant.

distinctdan:
[...]

PR #4674, opened 2019-03-01
[WIP] Async DNS

galderz:
I've been working on #95 and wanted to share progress with it and get some guidance on next steps.

As per #95 I started by looking into Call/Callback to see if DNS requests could be represented using them but these classes are specific to HTTP calls, so I discarded that option. I was also not convinced about adding DnsCall/DnsCallback.

So, instead I've worked backwards by starting first with an async DNS implementation (`AsyncDns`) and a unit test for it (`AsyncDnsTest` and `MockDnsCache`). I've made `AsyncDns` a subclass of `Dns` to make it easier to plug it in. It currently delegates to a given DNS implementation wrapping calls around `CompletableFuture.supplyAsync()` and configurable executor (as suggested in #95, it should be the IO pool). 

Then to try integrate the async DNS implementation I've decided to separate route state in `RouteSelector` into two implementations `SyncRouteState` and `AsyncRouteState` of `RouteState` interface. The former for sync DNS implementations and the latter for async DNS implementations. This interface is a bit clunky at the moment, so should be considered WIP as there are bigger problems to tackle first. 

`AsyncRouteState` just keeps track of the async lookup future and replies accordingly. The key aspect comes when `Selection` needs to be built. If the future is not complete when the `Selection` needs to be built, it throws an `IOException` indicating the async DNS lookup is not done. My rationale here is that this would eventually bubble up to `RetryAndFollowUpInterceptor` and it'd retry. Of course the retry would need to be slightly different since you'd want to just use the same connection and retry the logic all the way to selecting the route.

I wanted to find out from OkHttp developers if this is the right approach? This is not yet implemented and there are no end-to-end tests for async DNS, but before working on those I wanted to verify that this is a valid approach that can work with the current APIs.

On top of `AsyncDnsTest` I've developed a `AsyncRouteSelectorTest` test that extends `RouteSelectorTest` and uses `AsyncDns`. This test randomly fails because there are times when async DNS has not completed (to be expected), but I'm hoping to use the test to verify that if `AsyncRouteState` works as expected. For that, I'm planning to modify the test so that async DNS completes in time.

Once `AsyncRouteSelectorTest` works (with the assumption that async DNS always completes in time for the selection router `next()` to be called), I was planning to work on at least two end-to-end tests: one that verifies that an async DNS implementation that replies in time of router selection being queried works, and another where the DNS implementation is slow and see if retry logic

Summarising, still TODOs:
- Hook IO executor thread pool into AsyncDns (needs investigating).
- End-to-end async DNS test where async DNS completes in time first time.
- End-to-end async DNS test with slow async DNS and retries.
- Cancel an async DNS call if the HTTP `Call` instance is cancelled.
- End-to-end async DNS test with slow async DNS and cancel.
- Calls that timeout should cancel async DNS calls.

### DIFF
--- a/okhttp/src/main/java/okhttp3/AsyncDns.java
+++ b/okhttp/src/main/java/okhttp3/AsyncDns.java
@@ -12,15 +12,17 @@ import javax.annotation.Nullable;
 final class AsyncDns implements Dns {
 
   private final ExecutorService executor;
+  private final Dns dns;
 
   AsyncDns(Builder builder) {
     this.executor = builder.executor;
+    this.dns = builder.dns;
   }
 
   public CompletableFuture<List<InetAddress>> lookupAsync(String hostName) {
     return CompletableFuture.supplyAsync(() -> {
       try {
-        return Dns.SYSTEM.lookup(hostName);
+        return dns.lookup(hostName);
       } catch (UnknownHostException e) {
         throw new CompletionException(e);
       }
@@ -44,6 +46,7 @@ final class AsyncDns implements Dns {
 
   public static final class Builder {
     @Nullable ExecutorService executor = null;
+    Dns dns = Dns.SYSTEM;
 
     public AsyncDns build() {
       return new AsyncDns(this);
@@ -51,6 +54,11 @@ final class AsyncDns implements Dns {
 
     public Builder executor(ExecutorService executor) {
       this.executor = executor;
+      return this;
+    }
+
+    public Builder dns(Dns dns) {
+      this.dns = dns;
       return this;
     }
   }
--- a/okhttp/src/main/java/okhttp3/AsyncDns.java
+++ b/okhttp/src/main/java/okhttp3/AsyncDns.java
@@ -9,7 +9,7 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import javax.annotation.Nullable;
 
-final class AsyncDns implements Dns {
+public final class AsyncDns implements Dns {
 
   private final ExecutorService executor;
   private final Dns dns;
--- a/okhttp/src/main/java/okhttp3/AsyncDns.java
+++ b/okhttp/src/main/java/okhttp3/AsyncDns.java
@@ -1,0 +1,58 @@
+package okhttp3;
+
+import java.net.InetAddress;
+import java.net.UnknownHostException;
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import javax.annotation.Nullable;
+
+final class AsyncDns implements Dns {
+
+  private final ExecutorService executor;
+
+  AsyncDns(Builder builder) {
+    this.executor = builder.executor;
+  }
+
+  public CompletableFuture<List<InetAddress>> lookupAsync(String hostName) {
+    return CompletableFuture.supplyAsync(() -> {
+      try {
+        return Dns.SYSTEM.lookup(hostName);
+      } catch (UnknownHostException e) {
+        throw new CompletionException(e);
+      }
+    }, executor);
+  }
+
+  public @Override List<InetAddress> lookup(String hostname) throws UnknownHostException {
+    CompletableFuture<List<InetAddress>> future = lookupAsync(hostname);
+    try {
+      return future.get();
+    } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
+      throw new RuntimeException("Interrupted waiting for DNS resolution");
+    } catch (ExecutionException e) {
+      if (e.getCause() instanceof UnknownHostException)
+        throw (UnknownHostException) e.getCause();
+
+      throw new RuntimeException(e);
+    }
+  }
+
+  public static final class Builder {
+    @Nullable ExecutorService executor = null;
+
+    public AsyncDns build() {
+      return new AsyncDns(this);
+    }
+
+    public Builder executor(ExecutorService executor) {
+      this.executor = executor;
+      return this;
+    }
+  }
+
+}
--- a/okhttp/src/main/java/okhttp3/internal/connection/AsyncRouteState.java
+++ b/okhttp/src/main/java/okhttp3/internal/connection/AsyncRouteState.java
@@ -1,0 +1,167 @@
+package okhttp3.internal.connection;
+
+import java.io.IOException;
+import java.net.InetAddress;
+import java.net.InetSocketAddress;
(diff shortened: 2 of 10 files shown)


okhttp/src/main/java/okhttp3/AsyncDns.java
```diff
@@ -0,0 +1,66 @@
...
+import java.net.InetAddress;
+import java.net.UnknownHostException;
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import javax.annotation.Nullable;
+
+public final class AsyncDns implements Dns {
```
yschimke: Not sure how well this will fly given it introduces new public API.  I think this is high impact so would like some solution like this that doesn't involve blocking sync calls for Dns.

I wonder whether the public API part should be limited to a sub-interface with an additional method.

okhttp/src/main/java/okhttp3/AsyncDns.java
```diff
@@ -0,0 +1,66 @@
...
+
+  private final ExecutorService executor;
+  private final Dns dns;
+
+  AsyncDns(Builder builder) {
+    this.executor = builder.executor;
```
yschimke: presumably should fail here if builder.executor is null here.

okhttp/src/main/java/okhttp3/AsyncDns.java
```diff
@@ -0,0 +1,66 @@
...
+    this.dns = builder.dns;
+  }
+
+  public CompletableFuture<List<InetAddress>> lookupAsync(String hostName) {
```
yschimke: My major problem here is that I don't think this is the correct return type of a OkHttp wide single instance Dns implementation.  Just considering the dual stack world (https://en.wikipedia.org/wiki/Happy_Eyeballs) if we want to do this right, then it problem means a stream of InetAddresses e.g. IPv4 might return more quickly than IPv6.

Dns over Http already does this but blocks until both are ready.

yschimke:
As a thought experiment, maybe worth seeing how it would fit with Dns over Http which could be made async

okhttp/src/main/java/okhttp3/internal/connection/AsyncRouteState.java
```diff
@@ -0,0 +1,167 @@
...
+        throw new UnknownHostRuntimeException(e);
+
+      throw new RuntimeException(e);
+    }
+  }
+
+  private AllRoutes getComputedRoutes() throws IOException {
+    try {
+      if (computedRoutes == null)
+        computedRoutes = futureComputedRoutes.get();
```
yschimke: This is my main concern, a blocking call here is making this look simple while it's actually problematic.  e.g. if we use a single executor for both the OkHttp client and the Dns implementation, its simpler for users.  But now we have something capable of deadlocks IIUC.

yschimke:
This is high value, but I expect a lot of headwind to get this into a state ready to ship without causing awkward hard to reproduce bugs on different platforms etc.  See what others think.  I'm guessing it might be considered too big a change until OkHttp 4.  And at that point there might be nicer ways to crack that walnut *cough* kotlin coroutines :)

swankjesse:
First off, I’m frustrated that you spent a lot of time on this without first telling us you were going to do so. It’s unlikely that *any* large pull request from any new contributor will land. If you want background on this stance I’ve got a [presentation here](https://speakerdeck.com/swankjesse/an-open-source-advantage-droidcon-mtl-2015).

As for cancelable DNS lookups, it’s definitely something we want, and probably something we want in 2019. We’ve finally got a cancelable implementation with DNS-over-HTTPS. And I hope our host JVM and Android platforms will add similiar things.

But I think the API we want for this is Kotlin coroutines. They’re a more capable concurrency primitive than futures and work nicely both to implement and to call.

Next steps: I’m going to close this PR for now. I’m going to encourage us to more aggressively upgrade to Kotlin, and once that’s somewhat stable we’ll revisit this.

galderz:
> First off, I’m frustrated that you spent a lot of time on this without first telling us you were going to do so. 

I didn't meant to make you frustrated :). I was just giving a go at an issue that had been lying around for a while. I was fully aware that it was not a trivial change, but it felt like something I could have a go.

> It’s unlikely that any large pull request from any new contributor will land. If you want background on this stance I’ve got a presentation here.

I fully understand your stance. Do you have any labels for issues that are suitable for new contributors? I don't currently see any in [the project](https://github.com/square/okhttp/labels). I feel having such labels might help better steer new contributors towards issues that are the right size for them.

> Next steps: I’m going to close this PR for now. I’m going to encourage us to more aggressively upgrade to Kotlin, and once that’s somewhat stable we’ll revisit this.

Fair enough. 

What do you mean exactly by upgrading to Kotlin? Providing a Kotlin implementation of OkHttp that is then consumed by Java apps? Hence ditching the Java impl? Or am I misunderstanding it?

I will say that working on this PR, even though it's not integrated or discussed further , gave me a good insight into how things work :)

galderz:
> Do you have any labels for issues that are suitable for new contributors? I don't currently see any in the project. I feel having such labels might help better steer new contributors towards issues that are the right size for them.

@yschimke @swankjesse Btw, related to the above, if you know of any such issues I'd love to help.

swankjesse:
OkHttp + Kotlin will be something like this: https://youtu.be/Q8B4eDirgk0

I need to create a tracking bug for that work; please save your feedback for that.

swankjesse:
For new contributors: one of my core goals of this project is to keep it small. The best contributions improve the project’s quality without adding much code.

amirlivneh:
How about fixing flaky tests?

swankjesse:
Yes please!

LouisCAD:
Is there a way to pass a custom DNS resolution implementation to OkHttp 3.12.2 so it becomes possible to override the timeout? I'd like to be able to do something like this: https://stackoverflow.com/questions/18217335/can-i-set-the-getaddrinfo-timeout-in-android-for-defaulthttpclient

yschimke:
Some example custom DNS resolvers

https://github.com/yschimke/okurl/blob/master/src/main/kotlin/com/baulsupp/okurl/network/NettyDns.kt
https://github.com/yschimke/okurl/blob/master/src/main/kotlin/com/baulsupp/okurl/network/DnsOverride.kt
https://github.com/yschimke/okurl/blob/master/src/main/kotlin/com/baulsupp/okurl/network/GoogleDns.kt

Also Dns over HTTPS

https://github.com/square/okhttp/blob/master/okhttp-dnsoverhttps/src/main/java/okhttp3/dnsoverhttps/DnsOverHttps.kt

PR #4675, opened 2019-03-02
Delete long-deprecated OkUrlFactory.

swankjesse:
Some OkHttp behavior was only tested via this API, so this PR includes a bunch
of test migration. The test migration is as mechanical as possible, so we aren't
really using idiomatic APIs yet in these tests.

The `okhttp-urlconnection` survives for the JavaNetCookieJar and the
JavaNetAuthenticator. We can't remove these any time soon unfortunately.


okhttp-tests/src/test/java/okhttp3/URLConnectionTest.java
```diff
@@ -108,34 +103,30 @@
...
 import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertThat;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 import static org.junit.Assume.assumeTrue;
 
-/** Android's URLConnectionTest. */
+/** Android's URLConnectionTest, ported to exercise OkHttp's Call API. */
```
swankjesse: This file is where most of the deltas are.

swankjesse:
Very Large?

swankjesse:
Vrai Lagume?

oldergod:
Looks Very Looking Good To Me

swankjesse:
I’ve created a Gist that people impacted by this can use in shame until they upgrade to our public APIs:
https://gist.github.com/swankjesse/dd91c0a8854e1559b00f5fc9c7bfae70

Issue #4134, opened 2018-07-09
Public API for UnrepeatableRequestBody


swankjesse:
Would be handy to have a mechanism to signal one-shot request bodies.

swankjesse:
Perhaps `isRepeatable()` as a method that users can override?

PR #4676, opened 2019-03-02
New API: RequestBody.isOneShot()

swankjesse:
Closes: https://github.com/square/okhttp/issues/4134


okhttp/src/main/java/okhttp3/RequestBody.java
```diff
@@ -76,6 +77,21 @@ public boolean isDuplex() {
...
+  /**
+   * Returns true if this body expects at most one call to {@link #writeTo} and can be transmitted
+   * at most once. This is typically used when writing the request body is destructive and it is not
+   * possible to recreate the request body after it has been sent.
+   *
+   * <p>This method returns false unless it is overridden by a subclass.
+   *
+   * <p>OkHttp may attempt to retransmit request bodies when the original request fails due to a
+   * stale connection, a client timeout (HTTP 408), a satisfied authorization challenge (HTTP 401
+   * and 407), or a retryable server failure (HTTP 503 with a {@code Retry-After: 0} header).
```
oldergod: OkHttp may attempt to retransmit things but the `cannotRetryOneShotRequestBody` doesn't seem to do so?

swankjesse: Maybe rewrite this paragraph to be “by default...”

PR #4677, opened 2019-03-02
Test for connection leaks in more tests

amirlivneh:
This will help prevent regressions and point to the leaking test, rather than to a later test that happens to checks for leaks. The change is applied to all tests that have an OkHttpClient member variable.

### DIFF
--- a/okhttp-testing-support/src/main/java/okhttp3/TestUtil.java
+++ b/okhttp-testing-support/src/main/java/okhttp3/TestUtil.java
@@ -23,6 +23,8 @@ import java.util.Collections;
 import java.util.List;
 import okhttp3.internal.http2.Header;
 
+import static org.junit.Assert.assertEquals;
+
 public final class TestUtil {
   public static final InetSocketAddress UNREACHABLE_ADDRESS
       = new InetSocketAddress("198.51.100.1", 8080);
@@ -82,5 +84,10 @@ public final class TestUtil {
     Runtime.getRuntime().gc();
     Thread.sleep(100);
     System.runFinalization();
+  }
+
+  public static void ensureAllConnectionsReleased(OkHttpClient client) {
+    client.connectionPool().evictAll();
+    assertEquals(0, client.connectionPool().idleConnectionCount());
   }
 }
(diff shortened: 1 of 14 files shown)


okhttp-testing-support/src/main/java/okhttp3/TestUtil.java
```diff
@@ -83,4 +85,9 @@ public static void awaitGarbageCollection() throws Exception {
     Thread.sleep(100);
     System.runFinalization();
   }
+
+  public static void ensureAllConnectionsReleased(OkHttpClient client) {
+    client.connectionPool().evictAll();
+    assertEquals(0, client.connectionPool().idleConnectionCount());
```
yschimke: How does it fail when there is a bug causing the original test to fail, and also retaining connections?  Does this hide the root cause?

yschimke:
Looks good, provided it doesn't obscure the root cause.  IT does make me wonder whether the client creation and this check should be part of a @Rule instead?

swankjesse:
Agreed. If you wanna do a follow-up to make a rule for the default OkHttpClient, that’d be rad.

okhttp-testing-support/src/main/java/okhttp3/TestUtil.java
amirlivneh: > How does it fail when there is a bug causing the original test to fail, and also retaining connections?

If there is a failed assertion inside a test and the test also retains connections, the test output will show both the assertion inside the test and the leak assertion.
Without the change, the output shows the failed assertion inside the test and, sometimes a few minutes later, a leak assertion, when the first test that checks for leaks is executed.

> Does this hide the root cause?

I don't think it does.

amirlivneh:
Are you suggesting to add a new `OkHttpClientRule` class that extends `OkHttpClient` and implements `TestRule`.

swankjesse:
Or maybe `OkHttpClientTestingRule` that _has_ an `OkHttpClient` instance that you can use.

swankjesse:
```
@Rule public OkHttpClientTestingRule clientTestingRule = ...
private OkHttpClient client = clientTestingRule.client;
```

PR #4680, opened 2019-03-03
Use a TestRule to check for connection leaks

amirlivneh:
This implements the suggestion in https://github.com/square/okhttp/pull/4677.

I'm not very happy with the result and I personally prefer it not to be merged, at least not as-is: The problem is that many tests use this pattern to overwrite the `client` field:
```
client = client.newBuilder()
      ...
      .build();
```
Since updating `OkHttpClientTestingRule.client` after every instance of this would be tedious, `OkHttpClientTestingRule` will apply its checks to the original `OkHttpClient`. This is usually OK since the original and new `OkHttpClient` instances share most of the fields, but since some fields do differ, the result is inconsistent.

### DIFF
--- a/okhttp-testing-support/src/main/java/okhttp3/OkHttpClientTestingRule.java
+++ b/okhttp-testing-support/src/main/java/okhttp3/OkHttpClientTestingRule.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright (C) 2019 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okhttp3;
+
+import org.junit.rules.TestRule;
+import org.junit.runner.Description;
+import org.junit.runners.model.Statement;
+
+import static okhttp3.TestUtil.defaultClient;
+
+public class OkHttpClientTestingRule implements TestRule {
+  public OkHttpClient client = defaultClient();
+
+  @Override
+  public Statement apply(Statement base, Description description) {
+    return statement(base);
+  }
+
+  private Statement statement(final Statement base) {
+    return new Statement() {
+      public void evaluate() throws Throwable {
+        try {
+          base.evaluate();
+        } finally {
+          TestUtil.ensureAllConnectionsReleased(client);
+        }
+      }
+    };
+  }
+}
--- a/okhttp-sse/src/test/java/okhttp3/internal/sse/EventSourceHttpTest.java
+++ b/okhttp-sse/src/test/java/okhttp3/internal/sse/EventSourceHttpTest.java
@@ -17,6 +17,7 @@ package okhttp3.internal.sse;
 
 import java.util.concurrent.TimeUnit;
 import okhttp3.OkHttpClient;
+import okhttp3.OkHttpClientTestingRule;
 import okhttp3.Request;
 import okhttp3.mockwebserver.MockResponse;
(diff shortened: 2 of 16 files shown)


yschimke:
I think it's a net improvement, we can add functionality to the Rule, but happy to wait for other feedback.

amirlivneh:
> I think it's a net improvement

@yschimke, not necessarily. For example, consider this test:
https://github.com/square/okhttp/blob/d261896d03c3afc9a39e64cf0a7f31fb8d971d0d/okhttp-tests/src/test/java/okhttp3/ConnectionReuseTest.java#L114-L117
The existing code will detect a connection leak in this test because it looks at the new `client` instance and new `ConnectionPool`. The new code will not detect a leak because it looks at the overwritten `client` and its `ConnectionPool`.

yschimke:
@amirlivneh maybe this should be part of the functionality of the rule, e.g. registering the client with the TestRule for cleanup?

amirlivneh:
I considered it but was concerned about the boilerplate it would add to tests and the fact that it's easy to forget to register the client. This would make it inferior to the `tearDown()` solution, where these issues don't exist.

swankjesse:
I like it. I don’t think it’s harder to remember a rule than to remember a tearDown() method.

We could get rid of `TestUtil.defaultClient()` so the only way to get an OkHttpClient for a new test is to either instantiate one directly or to use this rule? Most tests that I write copy-paste an existing one for setup so as long as we start consistent it’s easy to stay consistent.

amirlivneh:
@swankjesse, my concern is not about forgetting to add a rule per file - that’s easy. It’s about the added boilerplate and the chance of forgetting to reassign the new client instance to the rule in each test method that uses this pattern:
https://github.com/square/okhttp/blob/d261896d03c3afc9a39e64cf0a7f31fb8d971d0d/okhttp-tests/src/test/java/okhttp3/ConnectionReuseTest.java#L114-L117

JakeWharton:
How about the rule becomes the factory:
```java
OkHttpClient client = rule.defaultClient();
```
```java
OkHttpClient client = rule.build(builder -> {
  builder.connectionPool(...)
});
```

amirlivneh:
Yes, that should work.

amirlivneh:
I tried the approach suggested by @JakeWharton, and it works. I personally still prefer the `tearDown()` solution because it requires adding only one call site per test class, leaving test methods unconcerned with the whole thing, but it looks like I'm in minority here :)

If we move forward with the TestRule solution, I recommend merging this PR as-is and sending another PR that only applies the pattern suggested by Jake. I expect the change to be in dozens of call sites and will be easier to review if it consists only of that single mechanical change.

okhttp-testing-support/src/main/java/okhttp3/OkHttpClientTestingRule.java
```diff
@@ -0,0 +1,28 @@
+package okhttp3;
```
swankjesse: nit: we put copyright statements on every file!

amirlivneh: Fixed

okhttp-testing-support/src/main/java/okhttp3/OkHttpClientTestingRule.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+package okhttp3;
+
+import org.junit.rules.TestRule;
+import org.junit.runner.Description;
+import org.junit.runners.model.Statement;
+
+import static okhttp3.TestUtil.defaultClient;
+
+public class OkHttpClientTestingRule implements TestRule {
```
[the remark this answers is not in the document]
swankjesse: Sure!

amirlivneh:
> How about the rule becomes the factory:
> 
> ```java
> OkHttpClient client = rule.defaultClient();
> ```
> 
> ```java
> OkHttpClient client = rule.build(builder -> {
>   builder.connectionPool(...)
> });
> ```

How about making `OkHtttpClientTestRule.client.newBuilder()` return a class derived from `OkHttpClient.Builder()` whose `build()` method updates its enclosing TestRule with the built `OkHttpClient` instance? This will be transparent to call sites, but will require making `OkHttpClient.Builder` non-final.

JakeWharton:
I'm a bit wary of changing public API to accommodate tests. Especially
since I'm a 'final' warrior.

Issue #4658, opened 2019-02-25
Connection leaks if WebSocket handshake fails


amirlivneh:
This is a recent regression bisected to dd70407455f09f3eda7288f0bb9610f8943af03e.

It may be the cause for the recent increase in test flakiness.

To reproduce, run this modified test:
```
  @Test public void missingConnectionHeader() throws IOException {
    webServer.enqueue(new MockResponse()
        .setResponseCode(101)
        .setHeader("Upgrade", "websocket")
        .setHeader("Sec-WebSocket-Accept", "ujmZX4KXZqjwy6vi1aQFH5p4Ygk="));
    newWebSocket();

    clientListener.assertFailure(101, null, ProtocolException.class,
        "Expected 'Connection' header value 'Upgrade' but was 'null'");

    client.connectionPool().evictAll();
    assertEquals(0, client.connectionPool().connectionCount());
  }
```

swankjesse:
Nice find. Can fix.

swankjesse:
I was playing with this and experimented running websockets calls over our new duplex mechanisms. Seems doable though there are some drawbacks. We’d need `GET` upgrade requests to have a duplex body! But the upside is you’d be able to optimistically send request frames before the response was received.

galderz:
Not sure if strictly related, but I've just seen `WebSocketHttpTest .missingConnectionHeader` fail (see #4695)

PR #4684, opened 2019-03-06
Don't leak a connection if a WebSocket handshake fails

swankjesse:
I'm not particularly happy with all of the moving parts here. I think
perhaps doing web sockets over duplex is a possible fix here.

Closes: https://github.com/square/okhttp/issues/4658

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/internal/ws/WebSocketHttpTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/internal/ws/WebSocketHttpTest.java
@@ -51,6 +51,7 @@ import org.junit.Rule;
 import org.junit.Test;
 
 import static okhttp3.TestUtil.defaultClient;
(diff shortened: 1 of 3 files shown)


okhttp-tests/src/test/java/okhttp3/internal/ws/WebSocketHttpTest.java
```diff
@@ -304,6 +304,9 @@
 
     clientListener.assertFailure(101, null, ProtocolException.class,
         "Expected 'Connection' header value 'Upgrade' but was 'null'");
+
+    client.connectionPool().evictAll();
```
amirlivneh: nit: It's better to use `TestUtil.ensureAllConnectionsReleased()`.

swankjesse: Yes! Fixed.

Issue #4685, opened 2019-03-06
LoggingEventListenerTest.secureGet failing with incorrect cipher suite


galderz:
`LoggingEventListenerTest.secureGet` failing on `jdk1.8.0_144` with:

```
java.lang.AssertionError: 
<[382 ms] secureConnectEnd: Handshake{tlsVersion=TLS_1_2 cipherSuite=TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256 peerCertificates=[CN=localhost] localCertificates=[]}>
did not match pattern
<\[\d+ ms] secureConnectEnd: Handshake\{tlsVersion=TLS_1_[23] cipherSuite=(?:TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384|TLS_AES_128_GCM_SHA256) peerCertificates=\[CN=localhost\] localCertificates=\[\]}>

	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at okhttp3.logging.HttpLoggingInterceptorTest$LogRecorder.assertLogMatch(HttpLoggingInterceptorTest.java:861)
	at okhttp3.logging.LoggingEventListenerTest$LogRecorder.assertLogMatch(LoggingEventListenerTest.java:224)
	at okhttp3.logging.LoggingEventListenerTest.secureGet(LoggingEventListenerTest.java:148)
```

galderz:
PR: https://github.com/square/okhttp/pull/4686

PR #4686, opened 2019-03-06
Expand possible cipher suite options #4685

galderz:


okhttp-logging-interceptor/src/test/java/okhttp3/logging/LoggingEventListenerTest.java
```diff
@@ -147,7 +147,7 @@ public void secureGet() throws Exception {
         .assertLogMatch("secureConnectStart")
         .assertLogMatch("secureConnectEnd: Handshake\\{"
             + "tlsVersion=TLS_1_[23] "
-            + "cipherSuite=(?:TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384|TLS_AES_128_GCM_SHA256) "
+            + "cipherSuite=(?:TLS_(?:ECDHE_ECDSA_WITH_|)AES_(?:256_GCM_SHA384|128_GCM_SHA256)) "
```
yschimke: This reads very intentional on the cipher, we could instead just check it starts with TLS_?

galderz: Makes sense, less likely to break in the future. I'll amend.

galderz:
@yschimke Updated PR

Issue #4682, opened 2019-03-04
InterceptorTest.connectTimeout fails when wifi disabled


galderz:
`InterceptorTest.connectTimeout` fails when WIFI is disabled on my laptop.

```
INFO: MockWebServer[64062] starting to accept connections
Mar 04, 2019 8:18:54 PM okhttp3.mockwebserver.MockWebServer$3 acceptConnections
INFO: MockWebServer[64062] done accepting connections: Socket closed

java.net.ConnectException: Operation timed out (Connection timed out)
	at java.net.PlainSocketImpl.socketConnect(Native Method)
	at java.net.AbstractPlainSocketImpl.doConnect(AbstractPlainSocketImpl.java:350)
	at java.net.AbstractPlainSocketImpl.connectToAddress(AbstractPlainSocketImpl.java:204)
	at java.net.AbstractPlainSocketImpl.connect(AbstractPlainSocketImpl.java:188)
	at java.net.SocksSocketImpl.connect(SocksSocketImpl.java:392)
	at java.net.Socket.connect(Socket.java:589)
	at java.net.Socket.connect(Socket.java:538)
	at okhttp3.InterceptorTest.connectTimeout(InterceptorTest.java:693)
```

I can look into it

galderz:
PR(#4693) sent

yschimke:
Seems fixed

PR #4693, opened 2019-03-08
Bind temporary socket to loopback address #4682

galderz:


yschimke:
Failing on the test lockup we have seen elsewhere

okhttp/src/test/java/okhttp3/InterceptorTest.java
```diff
@@ -679,7 +680,8 @@ private void interceptorThrowsRuntimeExceptionAsynchronous(boolean network) thro
       return chain.proceed(chain.request());
     };
 
-    ServerSocket serverSocket = new ServerSocket(0, 1);
+    InetAddress localhost = InetAddress.getLoopbackAddress();
```
yschimke: We should do this for MockWebServer also, it's used in the start method

galderz: MockWebServer calls `InetAddress.getByName("localhost")`, are you suggesting to change that for `InetAddress. getLoopbackAddress()`? Or do you mean something else?

yschimke: Yes, what do you think?  I think in practice it can be different, but shouldn't.

https://tools.ietf.org/html/rfc6761

> 6.3.  Domain Name Reservation Considerations for "localhost."
> 
>    The domain "localhost." and any names falling within ".localhost."
>    are special in the following ways:
> 
>    1.  Users are free to use localhost names as they would any other
>        domain names.  Users may assume that IPv4 and IPv6 address
>        queries for localhost names will always resolve to the respective
>        IP loopback address.
> 
>    2.  Application software MAY recognize localhost names as special, or
>        MAY pass them to name resolution APIs as they would for other
>        domain names.
> 
>    3.  Name resolution APIs and libraries SHOULD recognize localhost
>        names as special and SHOULD always return the IP loopback address
>        for address queries and negative responses for all other query
>        types.  Name resolution APIs SHOULD NOT send queries for
>        localhost names to their configured caching DNS server(s).

swankjesse: I want to keep it as is because we've had problems with other addresses.

galderz: I've not personally encountered issues with what `MockWebServer` was doing, so I'd leave it as is too.

PR #4701, opened 2019-03-09
Use OkHttpClientTestRule as a factory

amirlivneh:
This implements @JakeWharton's suggestion in https://github.com/square/okhttp/pull/4680#issuecomment-469672041.


swankjesse:
I don't prefer it. I like my tests to be as close to the regular API as possible, and this isn't it.

amirlivneh:
> I don't prefer it. I like my tests to be as close to the regular API as possible, and this isn't it.

I agree. I suggested an alternative in https://github.com/square/okhttp/pull/4680#issuecomment-470575491 and am open to other suggestions.

PR #4705, opened 2019-03-11
Provide better exceptions when the shared HTTP/2 stream fails

swankjesse:
Previously we always provided a generic StreamResetException. Now we provide
the exception that triggered the stream to be closed, if any.

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/http2/Http2Stream.java
+++ b/okhttp/src/main/java/okhttp3/internal/http2/Http2Stream.java
@@ -74,7 +74,10 @@ public final class Http2Stream {
    * close this stream (such as both peers closing it near-simultaneously) then this is the first
    * reason known to this peer.
    */
-  ErrorCode errorCode = null;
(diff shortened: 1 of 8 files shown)


okhttp/src/main/java/okhttp3/internal/http2/Http2Stream.java
```diff
@@ -153,7 +156,8 @@ public synchronized Headers takeHeaders() throws IOException {
     if (!headersQueue.isEmpty()) {
       return headersQueue.removeFirst();
     }
-    throw new StreamResetException(errorCode);
+    if (errorException == null) throw new StreamResetException(errorCode);
+    throw errorException;
```
amirlivneh: nit: consider making this and the other occurrences more concise:
```
throw errorException != null ? errorException : new StreamResetException(errorCode);
```

swankjesse: Fixed.

PR #4707, opened 2019-03-11
Recipe for Duplex Calls

oldergod:
First, not working as is: I still have a `java.net.ProtocolException: Duplex connections are not supported for HTTP/1` that I need to fix.

But before that, is it something one would be happy to find in recipes? I think so (even though I don't like we have to mock the server as well, and overall it's really just like one of your duplex tests).


swankjesse:
I don’t think this is that useful. Unlike most of our features duplex doesn’t stand on its own; you need something like gRPC to leverage it.

PR #4725, opened 2019-03-15
Switch from Maven to Gradle

swankjesse:


yschimke:
I've found the kotlin DSL for gradle build files to be a huge win.  Slightly glitchy when you edit the build files, but getting better fast.  But more importantly I don't find myself copying and pasting gradle snippets from other projects.  The type safety of the build files and IDE completion is a massive win.

build.gradle
```diff
@@ -0,0 +1,114 @@
+buildscript {
+  ext.versions = [
```
yschimke: For comparison with the Kotlin DSL

https://github.com/yschimke/okurl/blob/master/buildSrc/src/main/kotlin/Deps.kt

Egorand:
AFAIK the biggest issue with Kotlin DSL is performance, I heard people reporting it being extremely slow. The sole fact that it spawns multiple Gradle daemons to compile different parts of the config is gross. [KotlinPoet uses KotlinDSL](https://github.com/square/kotlinpoet/blob/master/build.gradle.kts), but it's a tiny single-module project. As for OkHttp, it might be worth the experiment, but we might have to revert to Groovy if we notice IDEs becoming very slow when working with OkHttp.

JakeWharton:
I also find its design to be bad. They tried to make it look kinda like the old Groovy instead of just designing nice, new APIs.

yschimke:
I'm happy just to get gradle to be fair, has the open source developer mindshare right now and seems to be really pushing the envelope to stay competitive with the blaze-clones.

swankjesse:
A couple things left for follow-up:

 * [ ] Error Prone
 * [ ] Maven publishing, including snapshot publishing from Travis

swankjesse:
# 🐘

yschimke:
For caching wins https://docs.travis-ci.com/user/languages/java/#caching

PR #4732, opened 2019-03-17
Configure travis caching for gradle.

swankjesse:


yschimke:
Looks like it's trying to work

image

yschimke:
Looks like win is measurable but basically noise compared to test times.  I might experiment with the parallel module or test running (IIRC) and if I can get it as stable, look at turning that on.

PR #4735, opened 2019-03-17
Fix MockWebServer leak in CookiesTest

amirlivneh:
MockWebServer instances were never shut down.

Fixes https://github.com/square/okhttp/issues/4734.


yschimke:
Nice cleanup

PR #4737, opened 2019-03-17
Aggressive parallel test execution [testing]

yschimke:
Testing for now


yschimke:
I suspect travis is only giving us so no clear wins here.  But locally runs in 90 seconds e.g.

https://scans.gradle.com/s/zhlsqk5azvs3q

src/test/resources/junit-platform.properties
```diff
@@ -0,0 +1,2 @@
+junit.jupiter.execution.parallel.enabled=true
+junit.jupiter.execution.parallel.mode.default=concurrent
```
swankjesse: Is JUnit 5 involved?

yschimke: There are 3 potential things here

1) Parallel Modules
2) Forking Multiple JVMs per module test run and dividing Test Cases between them
3) Concurrency within a single VM - https://junit.org/junit5/docs/snapshot/user-guide/#writing-tests-parallel-execution

Junit 5 runner is relevant for the 3rd, but shouldn't affect the test classes, just a different test runner.  I may not have unlocked this yet...

yschimke: I'm not in a rush until I confirm all 3 are active and see if results are ok.

yschimke:
Flaky tests when run in parallel mode

https://circleci.com/build-insights/gh/square/okhttp/pull%252F4737

PR #4738, opened 2019-03-17
Hook up japicmp for all stable artifacts

swankjesse:
This excludes okhttp-sse and okhttp-dnsoverhttps

### DIFF
--- a/okhttp-urlconnection/build.gradle
+++ b/okhttp-urlconnection/build.gradle
@@ -17,3 +17,15 @@ dependencies {
   testImplementation deps.junit
   testImplementation deps.assertj
 }
+
+apply plugin: 'me.champeau.gradle.japicmp'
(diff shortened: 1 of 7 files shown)


okhttp-urlconnection/build.gradle
```diff
@@ -17,3 +17,15 @@ dependencies {
   testImplementation deps.junit
   testImplementation deps.assertj
 }
+
+apply plugin: 'me.champeau.gradle.japicmp'
+task japicmp(type: me.champeau.gradle.japicmp.JapicmpTask, dependsOn: 'jar') {
+  oldClasspath = files(baselineJar(project, "3.14.0"))
```
JakeWharton: Move string to ext in the root?

swankjesse: Good idea. Done.

PR #4741, opened 2019-03-18
Convert some basic types to Kotlin

swankjesse:


swankjesse:
This is 2 commits.

Commit 1 is the actual Kotlinization:
https://github.com/square/okhttp/pull/4741/commits/3a29e5525591c4162405c911a1a4d47c646501fa

Commit 2 just renames .kt files to .java files.

okhttp/src/main/java/okhttp3/Dns.kt
```diff
@@ -0,0 +1,59 @@
...
+  @Throws(UnknownHostException::class)
+  fun lookup(hostname: String): List<InetAddress>
+
+  companion object {
+    /**
+     * A DNS that uses [InetAddress.getAllByName] to ask the underlying operating system to
+     * lookup IP addresses. Most custom [Dns] implementations should delegate to this instance.
+     */
+    @JvmField
+    val SYSTEM = object : Dns {
```
swankjesse: The old version of this used to check its argument for null, throwing an UnknownHostException if it is.

We can’t do that in Kotlin without making the argument nullable, which we don’t really want.

okhttp/src/main/java/okhttp3/Authenticator.kt
```diff
@@ -0,0 +1,119 @@
...
+   * `response`. Returns null if the challenge cannot be satisfied.
+   *
+   * The route is best effort, it currently may not always be provided even when logically
+   * available. It may also not be provided when an authenticator is re-used manually in an
+   * application interceptor, such as when implementing client-specific retries.
+   */
+  @Throws(IOException::class)
+  fun authenticate(route: Route?, response: Response): Request?
+
+  companion object {
```
PromanSEW: Can be
```kotlin
companion object NONE : Authenticator {
...
}
```

okhttp/src/main/java/okhttp3/CookieJar.kt
```diff
@@ -0,0 +1,61 @@
...
+  /**
+   * Load cookies from the jar for an HTTP request to `url`. This method returns a possibly
+   * empty list of cookies for the network request.
+   *
+   * Simple implementations will return the accepted cookies that have not yet expired and that
+   * [match][Cookie.matches] `url`.
+   */
+  fun loadForRequest(url: HttpUrl): List<Cookie>
+
+  companion object {
```
PromanSEW: Can be
```kotlin
companion object NO_COOKIES : CookieJar {
...
}
```

okhttp/src/main/java/okhttp3/Dns.kt
```diff
@@ -0,0 +1,59 @@
...
+interface Dns {
+  /**
+   * Returns the IP addresses of `hostname`, in the order they will be attempted by OkHttp. If a
+   * connection to an address fails, OkHttp will retry the connection with the next address until
+   * either a connection is made, the set of IP addresses is exhausted, or a limit is exceeded.
+   */
+  @Throws(UnknownHostException::class)
+  fun lookup(hostname: String): List<InetAddress>
+
+  companion object {
```
PromanSEW: The same

okhttp/src/main/java/okhttp3/Dns.kt
```diff
@@ -0,0 +1,59 @@
...
+    /**
+     * A DNS that uses [InetAddress.getAllByName] to ask the underlying operating system to
+     * lookup IP addresses. Most custom [Dns] implementations should delegate to this instance.
+     */
+    @JvmField
+    val SYSTEM = object : Dns {
+      override fun lookup(hostname: String): List<InetAddress> {
+        try {
+          return asList(*InetAddress.getAllByName(hostname))
```
JakeWharton:
```suggestion
          return InetAddress.getAllByName(hostname).toList()
```

okhttp/src/main/java/okhttp3/Call.kt
```diff
@@ -0,0 +1,99 @@
...
+   *
+   * Configure the client's default timeout with [OkHttpClient.Builder.callTimeout].
+   */
+  fun timeout(): Timeout
+
+  /**
+   * Create a new, identical call to this one which can be enqueued or executed even if this call
+   * has already been.
+   */
+  public override fun clone(): Call
```
JakeWharton: public is needed?

okhttp/src/main/java/okhttp3/EventListener.kt
```diff
@@ -0,0 +1,356 @@
...
+
+  /**
+   * Invoked immediately after a DNS lookup.
+   *
+   * This method is invoked after [dnsStart].
+   */
+  open fun dnsEnd(
+    call: Call,
+    domainName: String,
+    inetAddressList: List<@JvmSuppressWildcards InetAddress>
```
yschimke: TIL

okhttp/src/main/java/okhttp3/Call.kt
swankjesse: Fixed!

PR #4751, opened 2019-03-21
Refactor exceptions and static classes to Kotlin

yschimke:


okhttp/src/main/java/okhttp3/internal/connection/RouteException.kt
```diff
@@ -13,36 +13,26 @@
...
-    return lastException;
+  init {
+    lastConnectException = firstConnectException
   }
 
-  void addConnectException(IOException e) {
-    addSuppressedIfPossible(firstException, e);
-    lastException = e;
+  fun addConnectException(e: IOException) {
+    addSuppressedIfPossible(firstConnectException, e)
```
JakeWharton: Sorry I conflicted you here by removing this method. We can use `addSuppressed` directly now.

okhttp/src/main/java/okhttp3/internal/connection/RouteException.kt
```diff
@@ -13,36 +13,26 @@
...
-    firstException = cause;
-    lastException = cause;
-  }
-
-  public IOException getFirstConnectException() {
-    return firstException;
-  }
+class RouteException internal constructor(val firstConnectException: IOException) :
+    RuntimeException(firstConnectException) {
+  var lastConnectException: IOException? = null
```
JakeWharton: This doesn't seem to be actually nullable. You can change the initializer to be `firstConnectException` so it always has a non-null value and remove the `init` block.

PR #4752, opened 2019-03-21
Cleanup kotlin warnings

yschimke:


okhttp/src/main/java/okhttp3/HttpUrl.kt
```diff
@@ -310,6 +310,7 @@ class HttpUrl internal constructor(builder: Builder) {
    * the empty string. Each segment is formatted with a leading '/', so if path segments were ["a",
    * "b", ""], then the encoded path would be "/a/b/".
    */
+  @Suppress("UNCHECKED_CAST")
```
yschimke: List<String?> -> List<String>

okhttp/src/main/java/okhttp3/internal/platform/ConscryptPlatform.kt
```diff
@@ -31,6 +31,7 @@ import javax.net.ssl.X509TrustManager
  */
 class ConscryptPlatform private constructor() : Platform() {
 
+  @Suppress("DEPRECATION")
```
yschimke: provideTrustManager is deprecated

PR #4758, opened 2019-03-22
Add API to close websocket without sending status code.

inwotep:
The WebSocket specification allows a close frame to be sent
without a body.  In this case, neither a code or reason are
sent in the frame.  Add a new API, `close()`, which supports
this.

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/internal/ws/RealWebSocketTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/internal/ws/RealWebSocketTest.java
@@ -90,6 +90,18 @@ public final class RealWebSocketTest {
     }
   }
 
+  @Test public void clientCloseWithoutStatus() throws IOException {
+    client.webSocket.close();
+    // This will trigger a close response.
+    assertThat(server.processNextFrame()).isFalse();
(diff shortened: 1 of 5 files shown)


inwotep:
That's fair.   I didn't think changing the behavior of the existing close method (allowing 0 as "no status") was desirable either.  Do you have any other suggestions?

swankjesse:
What’s the motivation for not providing a code?

inwotep:
I'm using Okhttp as the underlying implementation to a higher level abstraction, which already exposes close without a status code (expecting the code not be sent at all).  This is simply to provide, at an API level, the feature included in the specification.

I would think the receiving end would translate this as 1000 (closed normally) anyway, and not expose whether it was sent or not, but I can't really be sure that's the case.

swankjesse:
Use 1000?

inwotep:
Sure, that's what I'm currently doing.  However, I can't be sure that the receiving end interprets no status code as 1000.  From the specification (7.1.5), it looks like it should actually be interpreted as 1005 (does that mean APIs expose it as 1005?).

swankjesse:
I think I’d like to close this & claim that sending a definitive status code is more intentional than omitting it.

PR #4762, opened 2019-03-22
Convert CacheControl to Kotlin

swankjesse:


okhttp/src/main/java/okhttp3/CacheControl.kt
```diff
@@ -0,0 +1,346 @@
...
+ */
+class CacheControl private constructor(
+  /**
+   * In a response, this field's name "no-cache" is misleading. It doesn't prevent us from caching
+   * the response; it only means we have to validate the response with the origin server before
+   * returning it. We can do this with a conditional GET.
+   *
+   * In a request, it means do not use a cache to satisfy the request.
+   */
+  @get:JvmName("noCache") val noCache: Boolean,
```
JakeWharton: Isn't this breaking your rule of compatibility? This is a source breaking change for Kotlin users. Or did we relax that?

okhttp/src/main/java/okhttp3/CacheControl.kt
```diff
@@ -0,0 +1,346 @@
...
+      var onlyIfCached = false
+      var noTransform = false
+      var immutable = false
+
+      var canUseHeaderValue = true
+      var headerValue: String? = null
+
+      var i = 0
+      val size = headers.size()
+      loop@ while (i < size) {
```
JakeWharton: `for (i in 0 until headers.size())`

okhttp/src/main/java/okhttp3/CacheControl.kt
swankjesse: Oooh, it does. Lemme fix.

yschimke: I've tried to revert almost all test changes by the end to ensure the code is at least source compatible.  But I understood we really want binary compatible via the build checks.

okhttp/src/main/java/okhttp3/CacheControl.kt
swankjesse: Fixed.

Issue #1733, opened 2015-06-27
Consider permessage-deflate extension for WebSocket


JakeWharton:
The HyBi IETF WG has picked up recently and it looks to be moving this to RFC. Chrome, Firefox, and many libraries support this extension. Given the widespread use of text-based messages for content, compression has an opportunity for giving [up-to 75% reduction in payload size](https://www.igvita.com/2013/11/27/configuring-and-optimizing-websocket-compression/#parameters). The [latest draft](https://tools.ietf.org/html/draft-ietf-hybi-permessage-compression-24) was just published two days ago.

JakeWharton:
I think this just got a thumbs up on its way to publication (if i'm reading the email correctly): https://datatracker.ietf.org/doc/draft-ietf-hybi-permessage-compression/

ashoemaker:
:+1:

OFmarting:
It is not possible to fully implement the specification on Android/with Java. This is the case because the Android/Java API doesn't offer the possibility to set the window bits of the Inflater/Deflater. Further the required SYNC_FLUSH option is not supported in Java until Java 7 and on Android until API Level 19 (as @swankjesse  pointed out in the [Okio source](https://github.com/square/okio/blob/4023ed0d4a6a09a485963a1d608030bdeeea8135/okio/src/main/java/okio/DeflaterSink.java#L89)). Although the option can be used since Android 2.3.

swankjesse:
Not a big fan of that spec. Seems like a lot more machinery than should be necessary. I think we'll be able to implement enough of the spec to be useful!

gdaniels:
Is this still on the radar?  The spec is an RFC now... https://tools.ietf.org/html/rfc7692.  Would love to see this in okhttp.

b95505017:
Any news? BTW nv-websocket-client support this spec.
https://github.com/TakahikoKawasaki/nv-websocket-client

JakeWharton:
At this time we don't have plans to implement.

yosiat:
Hi @JakeWharton ,
Will you accept any PR for implementing per message deflate in okhttp?

monil1334:
@JakeWharton  any updates on implementing the extension?

JakeWharton:
No plans to. Feel free to send a PR

swankjesse:
[...]

PR #4763, opened 2019-03-22
Allow specifying pongTimeout smaller than pingInterval

AntonRutkevich:
Current ping-pong implementation introduced in #3870 for websocket awaits for the next ping to detect the fact that the previous ping got lost. So the worst case for detecting a broken connection is `pingInterval` ms.

This works well for small pingInterval numbers, however if the number is high enough (let's say 15 or more seconds), we'll for 15 seconds instead of ~3-5 which looks more than enough even for mobile networks. This is a real use case for us, and we plan to try raising the pingInterval even more. 

This PR adds the ability to specify a pongTimeout smaller than pingInterval. Example:
pingInterval = 15_000
pongTimeout = 3_000

If pongTimeout is not used, the previous strategy is used. 

This is a draft PR, if I get an approve of the general idea, I'll contribute the same for HTTP/2 and update the javadocs. 
Also, with the whole Kotlin migration, where should the PR be targeted to have a chance to be released before okhttp4?

### DIFF
--- a/mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.java
+++ b/mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.java
@@ -738,7 +738,7 @@ public final class MockWebServer extends ExternalResource implements Closeable {
       }
     };
     RealWebSocket webSocket = new RealWebSocket(fancyRequest,
-        response.getWebSocketListener(), new SecureRandom(), 0);
+        response.getWebSocketListener(), new SecureRandom(), 0, 0);
     response.getWebSocketListener().onOpen(webSocket, fancyResponse);
     String name = "MockWebServer WebSocket " + request.getPath();
     webSocket.initReaderAndWriter(name, streams);
--- a/okhttp/src/main/java/okhttp3/OkHttpClient.java
+++ b/okhttp/src/main/java/okhttp3/OkHttpClient.java
@@ -197,6 +197,7 @@ public class OkHttpClient implements Cloneable, Call.Factory, WebSocket.Factory 
   final int readTimeout;
   final int writeTimeout;
   final int pingInterval;
+  final int pongTimeout;
 
   public OkHttpClient() {
     this(new Builder());
@@ -249,6 +250,7 @@ public class OkHttpClient implements Cloneable, Call.Factory, WebSocket.Factory 
     this.readTimeout = builder.readTimeout;
     this.writeTimeout = builder.writeTimeout;
     this.pingInterval = builder.pingInterval;
+    this.pongTimeout = builder.pongTimeout;
 
     if (interceptors.contains(null)) {
       throw new IllegalStateException("Null interceptor: " + interceptors);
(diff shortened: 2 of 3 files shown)


swankjesse:
Doesn't 15_000s/3_000s retain a worst case of 15s? What happens if the connection is severed immediately after the pong is transmitted?

AntonRutkevich:
Thanks for the input, checked the math again, hopefully I didn't miss anything this time. 

With the initial environment of high ping interval (lets take 20s), fast pongs (~1s from ping), and a reasonable timeout (~3s from ping), the current and new approaches behave like the following.

**X** axis is a timeline
**Y** is how long it takes to detect a disconnect that happened at second x
Both cases have pings at 0s, 20s and corresponding pongs at 1s, 21s. 
New plot has timeouts triggered at 3s, 23s.

**Current** 
https://www.desmos.com/calculator/f9ka8jvrrb
current
If connection is severed right after the pong, the next writePing in 19 seconds will not detect it, as `awaitingPong` is `false`. It will only be detected 20 seconds later. _(Is this indeed the case? Because if not then all the below ideas are not relevant.)_
So, current approach has best case of  **(pingInterval - rtt)** and the worst case of **(2*pingInterval - rtt)**. Here it's 19..39s.

**New**
https://www.desmos.com/calculator/tjfnbyzkzy
new
In the new approach, if connection is severed right after pong, it will be detected at **next** timeout event, as current timeout has already been cancelled due to arrived pong.
So, the new approach gives the best case of **(pongTimeout - rtt)** and the worst case of **(pingInterval + pongTimeout - rtt)**. Here it's 2-22s.

So, looks like new approach improves the average detection time by **(pingInterval - pongTimeout)**, leaving the network traffic the same. 

Does it make sense to you?

swankjesse:
Yeah, that helps! 

I’m trying to understand how this works in practice. What values will you be using? If you have a maximum tolerance of 21 seconds, just set the ping interval to 10.5 seconds?

AntonRutkevich:
In practice we expect to use values like 20/5 or 30/5. 
`pingInterval = 10.5s` 
indeed gives close results with 
`pingInterval = 21s`
`pongTimeout = 4s`
but consumes x2 traffic.

I understand that it does not make a difference in many cases, however in our case we pay for the traffic, and connections are being kept alive for quite a long time. "permessage-deflate" (#1733) is another feature I hope to manage to contribute with the same goal of reducing websocket traffic consumption.

AntonRutkevich:
@swankjesse , is there anything else I can do to get the green light? :)

swankjesse:
Gimme some time to consider this. I’m on board with the behavior but want to think about the API.

AntonRutkevich:
Sure, thanks!

yschimke:
@swankjesse what was the conclusion here?

swankjesse:
All of the ping interval stuff feels clumsy. It's hard for users to know what the internal should be and we likely waste bandwidth and battery pinging unused connections.

I'd prefer to do it dynamically. We have lots of information:
 - With the initial TLS handshake we can capture the round trip time
 - When the connection is busy we don't need to send pings, regular traffic is enough
 - When all calls are blocked on read a ping is potentially useful to fail fast
 - When there are no calls in flight we can do exponential back off on assumption that the connection probably won't be used anymore

yschimke:
Closing out because it's not inline with the next target, is public API and currently Java based.

@AntonRutkevich Thanks for your patience here and no (intended) slight on the high quality of the PR.

AntonRutkevich:
@swankjesse , @yschimke no problem, thanks for the input! 
It should be a good starting point for someone who would implement it properly.

PR #4770, opened 2019-03-23
Skip if tests are missing

yschimke:


okhttp-hpacktests/src/test/java/okhttp3/internal/http2/HpackDecodeTestBase.java
```diff
@@ -23,6 +23,8 @@
 import okhttp3.internal.http2.hpackjson.HpackJsonUtil;
 import okhttp3.internal.http2.hpackjson.Story;
 import okio.Buffer;
+import org.junit.Assume;
```
JakeWharton: Unused. Any reason you didn't use `assumeTrue`/`assumeFalse`?

okhttp-hpacktests/src/test/java/okhttp3/internal/http2/HpackDecodeTestBase.java
```diff
@@ -41,7 +43,7 @@
     for (String interopTestName : interopTests) {
       List<Story> stories = HpackJsonUtil.readStories(interopTestName);
       if (stories.isEmpty()) {
-        fail("No stories for: " + interopTestName);
+        throw new AssumptionViolatedException("No stories for: " + interopTestName);
```
JakeWharton: The past-tense of this class bothers me. Shouldn't it be `AssumptionViolationException`?

yschimke: It seems like the dynamic tests cases don't work with assumption failures, so I've replaced with just not running any tests if that is the case.

okhttp-hpacktests/src/test/java/okhttp3/internal/http2/HpackDecodeTestBase.java
yschimke: Absolutely no good reason :)

More that I was in the middle of an if statement, so I wrote some bad code without thought.  I was testing out the fix for missing git submodules.

okhttp-hpacktests/src/test/java/okhttp3/internal/http2/HpackDecodeTestBase.java
```diff
@@ -41,7 +41,7 @@
     for (String interopTestName : interopTests) {
       List<Story> stories = HpackJsonUtil.readStories(interopTestName);
       if (stories.isEmpty()) {
-        fail("No stories for: " + interopTestName);
+        Collections.emptyList();
```
swankjesse: what does this do?

yschimke: So options we have

1) Like maven - don't build this module
2) Build and fail if git submodule is missing.  This is master currently.
3) Use test assumptions to (visibly) skip tests if not available.  This failed when the test suite generation throws an Assume failure :(
4) Return 0 test cases, (silently) skips tests but compiles the code etc.  Still runs locally etc.

Above is an implementation of option 4.  Option 3 was my first attempt on the PR.

JakeWharton: I think Jesse is referring to the fact that you don't assign this to a variable. It's a bare statement with no effect.

okhttp-hpacktests/src/test/java/okhttp3/internal/http2/HpackDecodeTestBase.java
yschimke: Too much Kotlin.  apologies for the shoddy diffs recently.

okhttp-hpacktests/src/test/java/okhttp3/internal/http2/HpackDecodeTestBase.java
yschimke: Too much Kotlin.  apologies for the shoddy diffs recently.

yschimke: Same result luckily. since it loops through an empty collection next.

swankjesse:
I’d prefer an assumption failure ’cause you see an ignored test in the output.

yschimke:
@swankjesse ok, will change this around to an assumption failure, I'll try with some sentinel value for no tests.

yschimke:
image

PR #4771, opened 2019-03-23
Convert MediaType to Kotlin

swankjesse:


okhttp/src/main/java/okhttp3/CacheControl.kt
```diff
@@ -25,57 +25,39 @@ import java.util.concurrent.TimeUnit
  * See [RFC 7234, 5.2](https://tools.ietf.org/html/rfc7234#section-5.2).
  */
 class CacheControl private constructor(
-  /**
```
swankjesse: I put these in the wrong place on the last PR

okhttp/src/main/java/okhttp3/MediaType.kt
```diff
@@ -0,0 +1,132 @@
...
+            token.substring(1, token.length - 1)
+          }
+          else -> token
+        }
+        require(charset == null || charsetParameter.equals(charset, ignoreCase = true)) {
+          "Multiple charsets defined: \"$charset\" and: \"$charsetParameter\" for: \"$string\""
+        }
+        charset = charsetParameter
+        s = parameter.end()
+        continue
```
JakeWharton: Last statement in loop body. Not needed.

PR #4777, opened 2019-03-24
Use intrinsic version of toArray

yschimke:
Intellij advises to use this form post JVM 1.6 as it's an intrinsic that performs as well or better

### DIFF
--- a/okhttp/src/main/java/okhttp3/Headers.java
+++ b/okhttp/src/main/java/okhttp3/Headers.java
@@ -53,7 +53,7 @@ public final class Headers {
   private final String[] namesAndValues;
 
   Headers(Builder builder) {
-    this.namesAndValues = builder.namesAndValues.toArray(new String[builder.namesAndValues.size()]);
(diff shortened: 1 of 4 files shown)


swankjesse:
Does Kotlin do it better?!

yschimke:
Not sure, this is java -> java > 1.6

image

JakeWharton:
FYI this comes from https://shipilev.net/blog/2016/arrays-wisdom-ancients/. I'll check ART later. It lacks an intrinsic for Arrays.copyOf and copyOfRange like the JVM so I was going to try to add it. Maybe we can get this added as well if it's missing.

yschimke:
@JakeWharton do you want me to revert this change? or acceptable even if slower on ART for now?

JakeWharton:
No. I don't think these are in hot paths where it should matter. Just want
to double check ART.

JakeWharton:
Looks like Android doesn't do anything for either `toArray` variant so we'd probably need benchmarks to figure out which is better. But... who cares these aren't hotspots! Left as an exercise for someone else.

PR #4748, opened 2019-03-19
Convert OkHttpClient to Kotlin

ZacSweers:
Resolves #4747 

Renamed the file first with no changes to try to protect git history (unless you want to squash)

PR #4785, opened 2019-03-24
 Convert ConnectionPool to Kotlin

Egorand:


okhttp/src/main/java/okhttp3/ConnectionPool.kt
```diff
@@ -0,0 +1,49 @@
...
+ * application. The tuning parameters in this pool are subject to change in future OkHttp releases.
+ * Currently this pool holds up to 5 idle connections which will be evicted after 5 minutes of
+ * inactivity.
+ */
+class ConnectionPool @JvmOverloads constructor(
+  maxIdleConnections: Int = 5,
+  keepAliveDuration: Long = 5,
+  timeUnit: TimeUnit = TimeUnit.MINUTES
+) {
+  internal val delegate = RealConnectionPool(maxIdleConnections, keepAliveDuration, timeUnit)
```
Egorand: This is currently broken since `OkHttpClient.java` can't access `internal` members. Should be resolved once #4748 is merged.

okhttp/src/main/java/okhttp3/ConnectionPool.kt
```diff
@@ -0,0 +1,49 @@
...
+ * Manages reuse of HTTP and HTTP/2 connections for reduced network latency. HTTP requests that
+ * share the same [Address] may share a [Connection]. This class implements the policy
+ * of which connections to keep open for future use.
+ *
+ * @constructor Create a new connection pool with tuning parameters appropriate for a single-user
+ * application. The tuning parameters in this pool are subject to change in future OkHttp releases.
+ * Currently this pool holds up to 5 idle connections which will be evicted after 5 minutes of
+ * inactivity.
+ */
+class ConnectionPool @JvmOverloads constructor(
```
Egorand: Java version only had two constructors, the no-arg and the all-arg. `@JvmOverloads` will instead produce 4 constructors. Are we fine with it, or should I change it to two constructors declared explicitly?

JakeWharton: Two explicit constructors. The variations produced by `JvmOverloads` don't make sense. You always must pass a `TimeUnit` when supplying a time value

yschimke: my 2c: the @JvmOverloads approach is better as its source and binary compatible, but sets a nicer precedent.

yschimke: hmmm....., weird to specify keepAliveDuration but not timeUnit.  So worth discussion.

Egorand: @JakeWharton right, I didn't notice that. Will revert to two constructors and can follow up with a more idiomatic version if needed.

okhttp/src/main/java/okhttp3/ConnectionPool.kt
swankjesse: Wanna make it public and comment it in a way we won't miss?

```
TODO(kotlin): should be internal
```

Egorand: Yeah, can do as well

Egorand:
Actually managed to make it compile by adding `@JvmField` to delegate, will remove it once `OkHttpClient` is in Kotlin.

PR #4795, opened 2019-03-25
Add local.properties to .gitignore

elevenfive:


JakeWharton:
Do we need this? This should only be generated for Android projects opened in Android Studio.

elevenfive:
> Do we need this? This should only be generated for Android projects opened in Android Studio.

Great question -- I guess it depends on how many folks will open this in Android Studio -- I at least did :).

swankjesse:
Doesn’t hurt.

PR #4796, opened 2019-03-26
 Convert RealCall.java to Kotlin

Egorand:


okhttp/build.gradle
```diff
@@ -180,6 +180,15 @@ task japicmp(type: me.champeau.gradle.japicmp.JapicmpTask, dependsOn: 'jar') {
...
       'okhttp3.MediaType#type()',
+      'okhttp3.RealCall#getResponseWithInterceptorChain()',
+      'okhttp3.RealCall#redactedUrl()',
+      'okhttp3.RealCall#toLoggableString()',
+      'okhttp3.RealCall$AsyncCall#callsPerHost()',
+      'okhttp3.RealCall$AsyncCall#executeOn(java.util.concurrent.ExecutorService)',
+      'okhttp3.RealCall$AsyncCall#get()',
+      'okhttp3.RealCall$AsyncCall#host()',
+      'okhttp3.RealCall$AsyncCall#request()',
+      'okhttp3.RealCall$AsyncCall#reuseCallsPerHostFrom(okhttp3.RealCall$AsyncCall)',
```
JakeWharton: This is getting ridiculous. We need to ensure we pull in an updated japicmp before we ship so we can nuke 99% of these!

okhttp/src/main/java/okhttp3/RealCall.kt
```diff
@@ -0,0 +1,214 @@
...
+
+  // Guarded by this.
+  @get:Synchronized override var isExecuted: Boolean = false
+
+  override val isCanceled: Boolean
+    get() = transmitter.isCanceled
+
+  override fun request(): Request = originalRequest
+
+  @Throws(IOException::class)
```
JakeWharton: Might not be needed since non-public type and an override

okhttp/src/main/java/okhttp3/RealCall.kt
```diff
@@ -0,0 +1,214 @@
...
+     */
+    fun executeOn(executorService: ExecutorService) {
+      assert(!Thread.holdsLock(client.dispatcher()))
+      var success = false
+      try {
+        executorService.execute(this)
+        success = true
+      } catch (e: RejectedExecutionException) {
+        val ioException = InterruptedIOException("executor rejected")
+        ioException.initCause(e)
```
JakeWharton: We should file a JDK bug to get this added to constructor.

okhttp/src/main/java/okhttp3/RealCall.kt
Egorand: Removed

okhttp/src/main/java/okhttp3/RealCall.kt
Egorand: Good call, wanna do that? Guess you've got more context.

JakeWharton: Uh I can try! We'll basically be dead before we can use it, but still worth trying to make the world a better place.

PR #4797, opened 2019-03-26
 Convert Route.java to Kotlin

Egorand:
Was tempted to use data class for this, but it breaks Kotlin compatibility. Is this fine?


JakeWharton:
> Is this fine?

We should start with direct conversion and then expose properties and create deprecated(hidden) functions afterward to make the API idiomatic.

Egorand:
Makes sense. I'll decline and make another pass.

PR #4798, opened 2019-03-26
 Convert Route.java to Kotlin

Egorand:


okhttp/src/main/java/okhttp3/Route.kt
```diff
@@ -0,0 +1,76 @@
...
+  /**
+   * Returns true if this route tunnels HTTPS through an HTTP proxy.
+   * See [RFC 2817, Section 5.2](http://www.ietf.org/rfc/rfc2817.txt).
+   */
+  fun requiresTunnel(): Boolean {
+    return address.sslSocketFactory != null && proxy.type() == Proxy.Type.HTTP
+  }
+
+  override fun equals(other: Any?): Boolean {
+    return (other is Route
```
JakeWharton: nit: parens not needed

Egorand: Fixed

Egorand:
Circle CI failure looks unrelated, merging

PR #4799, opened 2019-03-26
Move Handshake to Kotlin

shaishavgandhi05:


okhttp/src/main/java/okhttp3/Handshake.kt
```diff
@@ -0,0 +1,156 @@
...
+ * is its *peer*.
+ *
+ * This value object describes a completed handshake. Use [ConnectionSpec] to set policy
+ * for new handshakes.
+ */
+class Handshake private constructor(
+    private val tlsVersion: TlsVersion,
+    private val cipherSuite: CipherSuite,
+    private val peerCertificates: List<Certificate>,
+    private val localCertificates: List<Certificate>
```
JakeWharton: nit: these should be +2 not +4

okhttp/src/main/java/okhttp3/Handshake.kt
```diff
@@ -0,0 +1,156 @@
...
+  fun cipherSuite() = cipherSuite
+
+  /** Returns a possibly-empty list of certificates that identify the remote peer. */
+  fun peerCertificates() = peerCertificates
+
+  /** Returns the remote peer's principle, or null if that peer is anonymous. */
+  fun peerPrincipal(): Principal? {
+    return if (peerCertificates.isNotEmpty())
+      (peerCertificates[0] as X509Certificate).subjectX500Principal
+    else
```
JakeWharton: curlies on both branches

okhttp/src/main/java/okhttp3/Handshake.kt
```diff
@@ -0,0 +1,156 @@
...
+  }
+
+  /** Returns a possibly-empty list of certificates that identify this peer. */
+  fun localCertificates() = localCertificates
+
+  /** Returns the local principle, or null if this peer is anonymous. */
+  fun localPrincipal(): Principal? {
+    return if (localCertificates.isNotEmpty())
+      (localCertificates[0] as X509Certificate).subjectX500Principal
+    else
```
JakeWharton: curlies

okhttp/src/main/java/okhttp3/Handshake.kt
```diff
@@ -0,0 +1,156 @@
...
+      null
+  }
+
+  override fun equals(other: Any?): Boolean {
+    if (other !is Handshake) return false
+    return (tlsVersion == other.tlsVersion
```
JakeWharton: redundant parens. you can also fold `other is Handshake` into `&&` chain

okhttp/src/main/java/okhttp3/Handshake.kt
```diff
@@ -0,0 +1,156 @@
...
+    return ("Handshake{"
+        + "tlsVersion="
+        + tlsVersion
+        + " cipherSuite="
+        + cipherSuite
+        + " peerCertificates="
+        + names(peerCertificates)
+        + " localCertificates="
+        + names(localCertificates)
+        + '}'.toString())
```
JakeWharton: redundant parens, you shouldn't need `.toString()` call either

okhttp/src/main/java/okhttp3/Handshake.kt
```diff
@@ -0,0 +1,156 @@
...
+
+      val peerCertificates: Array<Certificate>? = try {
+        session.peerCertificates
+      } catch (ignored: SSLPeerUnverifiedException) {
+        null
+      }
+
+      val peerCertificatesList = if (peerCertificates != null)
+        Util.immutableList(*peerCertificates)
+      else
```
JakeWharton: curlies

okhttp/src/main/java/okhttp3/Handshake.kt
```diff
@@ -0,0 +1,156 @@
...
+        emptyList()
+
+      val localCertificates = session.localCertificates
+      val localCertificatesList = if (localCertificates != null)
+        Util.immutableList(*localCertificates)
+      else
```
JakeWharton: curlies

okhttp/src/main/java/okhttp3/Handshake.kt
```diff
@@ -0,0 +1,156 @@
...
+    val strings = ArrayList<String>()
+
+    for (cert in certificates) {
+      if (cert is X509Certificate) {
+        strings.add(cert.subjectDN.toString())
+      } else {
+        strings.add(cert.type)
+      }
+    }
+    return strings
```
JakeWharton: Implementation could be
```kotlin
return certificates
    .map {
      if (it is ..) {
        it.subjectDN.toString()
      } else {
        it.type
      }
    }
```
which will also pre-size the new list

okhttp/src/main/java/okhttp3/Handshake.kt
shaishavgandhi05: Done 26a75752e3edd3600fbc2f4c0e796134e3b08277

okhttp/src/main/java/okhttp3/Handshake.kt
shaishavgandhi05: Done 26a75752e3edd3600fbc2f4c0e796134e3b08277

okhttp/src/main/java/okhttp3/Handshake.kt
shaishavgandhi05: Done 26a75752e3edd3600fbc2f4c0e796134e3b08277

okhttp/src/main/java/okhttp3/Handshake.kt
shaishavgandhi05: Done 26a75752e3edd3600fbc2f4c0e796134e3b08277

okhttp/src/main/java/okhttp3/Handshake.kt
shaishavgandhi05: 26a75752e3edd3600fbc2f4c0e796134e3b08277

okhttp/src/main/java/okhttp3/Handshake.kt
shaishavgandhi05: 26a75752e3edd3600fbc2f4c0e796134e3b08277

okhttp/src/main/java/okhttp3/Handshake.kt
shaishavgandhi05: Nice! 26a75752e3edd3600fbc2f4c0e796134e3b08277

PR #4800, opened 2019-03-26
Move tests from okhttp-tests/src/test/java to okhttp/src/test/java

swankjesse:
The previous module structure was necessary because Maven wouldn't allow
mockwebserver to depend on okhttp/src and for okhttp/test to depend on
mockwebserver. With Gradle this constraint is lifted and we can fold
everything into a single module.

I'm in a rush to do this now because it's necessary for Kotlin 'internal'
tests to be in the same module.


okhttp/build.gradle
```diff
@@ -29,8 +29,16 @@ dependencies {
   compileOnly deps.jsr305
   compileOnly deps.animalSniffer
 
+  testImplementation deps.okio
+  testImplementation project(':okhttp-testing-support')
+  testImplementation project(':okhttp-tls')
+  testImplementation project(':okhttp-urlconnection')
```
JakeWharton: This one feels strange...

okhttp/build.gradle
```diff
@@ -29,8 +29,16 @@ dependencies {
...
+  testImplementation project(':mockwebserver')
+  testImplementation project(':okhttp-logging-interceptor')
```
JakeWharton: And this one...

swankjesse: EventListenerTest.java uses it. Can fix, but not this PR.

okhttp/build.gradle
swankjesse: CookiesTest.java uses JavaNetCookieJar. Can fix, but not this PR.

PR #4802, opened 2019-03-26
Convert Challenge to Kotlin

shaishavgandhi05:
I have a few concerns here that I'd like to get some opinions on. Pointing them out inline.


okhttp/src/main/java/okhttp3/Challenge.kt
```diff
@@ -0,0 +1,90 @@
...
+  fun scheme() = scheme
+
+  /**
+   * Returns the auth params, including [realm] and [charset] if present, but as
+   * strings. The map's keys are lowercase and should be treated case-insensitively.
+   */
+  fun authParams() = authParams
+
+  /** Returns the protection space. */
+  fun realm(): String? = authParams["realm"]
```
shaishavgandhi05: This seems like a breaking change to me and i'm unsure how to proceed. The java code returned a `String` with no `@Nullable` annotation. However Kotlin won't let me do that since it's technically nullable.

okhttp/src/main/java/okhttp3/Challenge.kt
```diff
@@ -0,0 +1,90 @@
...
+class Challenge(
+  private val scheme: String,
+  authParams: Map<String?, String>
+) {
+  private val authParams: Map<String?, String>
+
+  init {
+    val newAuthParams: MutableMap<String?, String> = LinkedHashMap()
+    for ((key, value) in authParams) {
+      val newKey = key?.toLowerCase(US)
```
shaishavgandhi05: In the Java code, we check for whether the `key` is `null` and so I'm carrying this over. This does mean that the `authParams` are `Map<String?, String>`.

okhttp/src/main/java/okhttp3/Challenge.kt
JakeWharton: It's binary compatible but source-incompatible for Kotlin users and NullAway+Java users. I don't think it's a big deal since it'll be attached with a major version. Correctness for nullability is paramount.

okhttp/src/main/java/okhttp3/Challenge.kt
JakeWharton: All maps coming from Java may have a null key. Whether we actually specifically allow that here I'm not sure.

okhttp/src/main/java/okhttp3/Challenge.kt
swankjesse: Should be nullable. This code is right.

okhttp/src/main/java/okhttp3/Challenge.kt
swankjesse: We describe it as the key used for token68.

PR #4807, opened 2019-03-27
Convert CipherSuite to Kotlin

swankjesse:


okhttp/src/test/java/okhttp3/CipherSuiteTest.java
```diff
@@ -88,9 +88,9 @@
 
   @Test public void javaName_equalsToString() {
     assertThat(TLS_RSA_EXPORT_WITH_RC4_40_MD5.toString()).isEqualTo(
-        TLS_RSA_EXPORT_WITH_RC4_40_MD5.javaName);
+        TLS_RSA_EXPORT_WITH_RC4_40_MD5.javaName());
```
yschimke: Nice, more use of public API in tests.

PR #4812, opened 2019-03-27
Convert ConnectionSpec to Kotlin

swankjesse:


okhttp/src/main/java/okhttp3/ConnectionSpec.kt
```diff
@@ -0,0 +1,324 @@
...
+  fun cipherSuites(): List<CipherSuite>? {
+    return if (cipherSuites != null) CipherSuite.forJavaNames(*cipherSuites) else null
+  }
+
+  /**
+   * Returns the TLS versions to use when negotiating a connection. Returns null if all of the SSL
+   * socket's enabled TLS versions should be used.
+   */
+  fun tlsVersions(): List<TlsVersion>? {
+    return if (tlsVersions != null) tlsVersionsForJavaNames(*tlsVersions) else null
```
oldergod: I think you don't like it but `return tlsVersions?.let { tlsVersionsForJavaNames(*it) }`

JakeWharton: We shouldn't need any helpers that convert an array/list to a list with a transformation anymore. This could be `tlsVersions?.map(::tlsVersionForJavaName)`, right?

okhttp/src/main/java/okhttp3/ConnectionSpec.kt
```diff
@@ -0,0 +1,324 @@
...
+
+  override fun equals(other: Any?): Boolean {
+    if (other !is ConnectionSpec) return false
+    if (other === this) return true
+
+    val that = other as ConnectionSpec?
+    if (this.isTls != that!!.isTls) return false
+
+    if (isTls) {
+      if (!Arrays.equals(this.cipherSuites, that.cipherSuites)) return false
```
JakeWharton: Kotlin uses `contentEquals` extension on array. It just ends up calling this though so not a big deal

okhttp/src/main/java/okhttp3/ConnectionSpec.kt
```diff
@@ -0,0 +1,324 @@
...
+      if (this.supportsTlsExtensions != that.supportsTlsExtensions) return false
+    }
+
+    return true
+  }
+
+  override fun hashCode(): Int {
+    var result = 17
+    if (isTls) {
+      result = 31 * result + Arrays.hashCode(cipherSuites)
```
JakeWharton: Similarly, `contentHashCode`

okhttp/src/main/java/okhttp3/ConnectionSpec.kt
swankjesse: Oooh, nice. Fixed.

okhttp/src/main/java/okhttp3/ConnectionSpec.kt
swankjesse: Tried it, and ran into a type inference issue. Will come back to it later.

okhttp/src/main/java/okhttp3/ConnectionSpec.kt
swankjesse: This one worked just fine.

okhttp/src/main/java/okhttp3/ConnectionSpec.kt
swankjesse: (I think I’ll be happier migrating this class to use Lists instead of Arrays.)

PR #4813, opened 2019-03-27
Convert http2 Header to Kotlin

shaishavgandhi05:


okhttp/src/main/java/okhttp3/internal/http2/Header.kt
```diff
@@ -0,0 +1,68 @@
...
+
+  override fun hashCode(): Int {
+    var result = 17
+    result = 31 * result + name.hashCode()
+    result = 31 * result + value.hashCode()
+    return result
+  }
+
+  override fun toString(): String {
+    return Util.format("%s: %s", name.utf8(), value.utf8())
```
JakeWharton: Use interpolation directly?

shaishavgandhi05: `Util.format` formats the string according to US locale. Just wanted to confirm that moving to direct interpolation won't affect that? (unsure about locale implications with direct interpolation).

swankjesse: It's relevant for `%d` but not `%s`.
https://github.com/square/okhttp/issues/2522

PR #4814, opened 2019-03-28
Okio upgrade post kotlin

yschimke:


okhttp/src/main/java/okhttp3/internal/Util.java
```diff
@@ -678,4 +679,20 @@ public static boolean sameConnection(HttpUrl a, HttpUrl b) {
...
+   *
+   * @param buffer the in-memory buffer
+   * @return byte returned by buffer.readByte()
+   * @throws AssertionError in place of EOFException
+   */
+  public static byte readIOSafeByte(Buffer buffer) {
+    try {
+      return buffer.readByte();
+    } catch (EOFException eofe) {
+      throw new AssertionError("unexpected eod of stream", eofe);
```
swankjesse: eod -> eof

okhttp/src/main/java/okhttp3/internal/Util.java
```diff
@@ -678,4 +679,20 @@ public static boolean sameConnection(HttpUrl a, HttpUrl b) {
...
+
+  /**
+   * Read a byte from a buffer where the operation is known to be safe, e.g. reading
+   * a stream in-memory with remaining bytes.
+   *
+   * @param buffer the in-memory buffer
+   * @return byte returned by buffer.readByte()
+   * @throws AssertionError in place of EOFException
+   */
+  public static byte readIOSafeByte(Buffer buffer) {
```
swankjesse: I don't get the motivation for this method. Is it to avoid a checked exception?

yschimke: Yep, an incompatibility between okio 1 and 2.  public method that now throws an IOException.

yschimke: Do you want me to remove and add throws to each of the callers?

JakeWharton: Huh, why wasn't this caught?

JakeWharton: https://github.com/square/okio/commit/487a1f2392d8a9f017dceaca15656a82ab4ed46d

swankjesse: Yeah, please add throws to the callers. I'm expecting no new catch clauses.

yschimke: I had a branch&PR previously with the changes to switch the java code to okio 2.  

https://github.com/square/okhttp/pull/4256/files

PR #4816, opened 2019-03-28
Move more interfaces to Kotlin

shaishavgandhi05:
Moves `CacheRequest`, `ExchangeCodec`, `CacheRequest` and `TrustRootIndex` to Kotlin.


okhttp/src/main/java/okhttp3/internal/tls/TrustRootIndex.kt
```diff
@@ -13,11 +13,11 @@
...
 
-import java.security.cert.X509Certificate;
+import java.security.cert.X509Certificate
 
-public interface TrustRootIndex {
-  /** Returns the trusted CA certificate that signed {@code cert}. */
-  X509Certificate findByIssuerAndSignature(X509Certificate cert);
+interface TrustRootIndex {
+  /** Returns the trusted CA certificate that signed [cert].  */
+  fun findByIssuerAndSignature(cert: X509Certificate): X509Certificate
```
shaishavgandhi05: Build will currently fail. This returns a non-null `X509Certificate` but `AndroidPlatform` returns null when overriding this method.

shaishavgandhi05: Any ideas on how to proceed here? Should this be nullable because `AndroidPlatform` overrides this and returns null?

swankjesse: Should return `X509Certificate?`.

Classes in `internal` are not annotated `@Nullable` precisely!

shaishavgandhi05: Thanks Jesse, good to know! 
92433df501d10342a66c242a114fdee52aa33218

okhttp/src/main/java/okhttp3/internal/http/ExchangeCode.kt
```diff
@@ -0,0 +1,81 @@
...
+
+  /** Flush the request to the underlying socket and signal no more bytes will be transmitted.  */
+  @Throws(IOException::class)
+  fun finishRequest()
+
+  /**
+   * Parses bytes of a response header from an HTTP transport.
+   *
+   * @param expectContinue true to return null if this is an intermediate response with a "100"
+   * response code. Otherwise this method never returns null.
```
swankjesse: nit: we like to indent wrapped KDoc tags +4 spaces

swankjesse:
Thanks!

PR #4819, opened 2019-03-30
Convert RouteDatabase to Kotlin

ychescale9:


okhttp/src/main/java/okhttp3/internal/connection/RouteDatabase.kt
```diff
@@ -13,33 +13,32 @@
...
 /**
  * A blacklist of failed routes to avoid when creating a new connection to a target address. This is
  * used so that OkHttp can learn from its mistakes: if there was a failure attempting to connect to
  * a specific IP address or proxy server, that failure is remembered and alternate routes are
  * preferred.
  */
-final class RouteDatabase {
-  private final Set<Route> failedRoutes = new LinkedHashSet<>();
+class RouteDatabase {
+  private val failedRoutes = LinkedHashSet<Route>()
```
yschimke: linkedSetOf<Route>()

okhttp/src/main/java/okhttp3/internal/connection/RouteDatabase.kt
```diff
@@ -13,33 +13,32 @@
...
+  @Synchronized fun connected(route: Route) {
+    failedRoutes.remove(route)
   }
 
-  /** Returns true if {@code route} has failed recently and should be avoided. */
-  public synchronized boolean shouldPostpone(Route route) {
-    return failedRoutes.contains(route);
+  /** Returns true if [route] has failed recently and should be avoided. */
+  @Synchronized fun shouldPostpone(route: Route): Boolean {
+    return failedRoutes.contains(route)
```
yschimke: change the block to inline expression

okhttp/src/main/java/okhttp3/internal/connection/RouteDatabase.kt
BoxResin: Specify the type for `failedRoutes` explicitly to prevent its type determined as `LinkedHashSet<Route>`.
```kotlin
private val failedRoutes: Set<Route> = ...
```

ychescale9: `add` and `remove` aren't available in `Set`. I've used `HashSet` instead:
`private val failedRoutes: HashSet<Route> = linkedSetOf()`

ychescale9: Changed to `private val failedRoutes: HashSet<Route> = linkedSetOf()`

okhttp/src/main/java/okhttp3/internal/connection/RouteDatabase.kt
ychescale9: Done and rebased

okhttp/src/main/java/okhttp3/internal/connection/RouteDatabase.kt
BoxResin: You can use `mutableSetOf()` instead to use `add` or `remove`.

yschimke: use MutableSet

BoxResin: This method is more the kotlin-way.
```kotlin
private val failedRoutes: MutableSet<Route> = mutableSetOf()
```

yschimke: Yep, i missed that it was linkedHashSet by default.  TIL

BoxResin: In short, it can be like:
```kotlin
private val failedRoutes = mutableSetOf<Route>()
```
The type of failedRoutes is still `MutableSet<Route>`.

ychescale9: Thanks. Updated and rebased.

oldergod:
thank you

PR #4822, opened 2019-03-30
Timeout on slow tests and fix CircleCI slow test

yschimke:
We should aim to have timeout on any tests we identify as slow, can't land this until we fix the terrible one

### DIFF
--- a/okhttp/src/test/java/okhttp3/EventListenerTest.java
+++ b/okhttp/src/test/java/okhttp3/EventListenerTest.java
@@ -55,11 +55,13 @@ import org.hamcrest.BaseMatcher;
 import org.hamcrest.CoreMatchers;
 import org.hamcrest.Description;
 import org.hamcrest.Matcher;
+import org.jetbrains.annotations.NotNull;
 import org.junit.After;
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
+import org.junit.rules.Timeout;
 
 import static java.util.Arrays.asList;
 import static okhttp3.tls.internal.TlsUtil.localhost;
@@ -73,6 +75,7 @@ public final class EventListenerTest {
   public static final Matcher<Response> anyResponse = CoreMatchers.any(Response.class);
   @Rule public final MockWebServer server = new MockWebServer();
   @Rule public final OkHttpClientTestRule clientTestRule = new OkHttpClientTestRule();
+  @Rule public final Timeout timeoutRule = new Timeout(20, TimeUnit.SECONDS);
 
   private final RecordingEventListener listener = new RecordingEventListener();
   private final HandshakeCertificates handshakeCertificates = localhost();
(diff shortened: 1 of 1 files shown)


yschimke:
Test output https://158-5152285-gh.circle-artifacts.com/0/home/circleci/reports/okhttp/build/reports/tests/test/classes/okhttp3.EventListenerTest.html#tab2

yschimke:
With http/2 I don't think the timeouts kick in, and we have a 1 second pause between each IO operation  because of the throttle behaviour.

yschimke:
With this fix, the slowest test in CI drops from 100+ seconds to 11 seconds.

okhttp/src/test/java/okhttp3/EventListenerTest.java
```diff
@@ -943,43 +948,32 @@ private void responseBodyFail(Protocol expectedProtocol) throws IOException {
...
+  @Test public void requestBodyFailHttp2OverHttps() {
     enableTlsWithTunnel(false);
     server.setProtocols(asList(Protocol.HTTP_2, Protocol.HTTP_1_1));
-    requestBodyFail();
-  }
 
-  @Test public void requestBodyFailHttp() throws IOException {
-    requestBodyFail();
+    requestBodyFail(new MockResponse()
+        .setSocketPolicy(SocketPolicy.DISCONNECT_DURING_REQUEST_BODY), sizedRequest(2), Protocol.HTTP_2);
```
yschimke: I'm worried about this, it might be better to have a custom request body that throttles the writes as well...

yschimke:
I suspect the reason for this showing up in CircleCI is environmental, within the containers we read a different size of buffer from the network, or "localhost" is not being selected as true loopback device and we see a sleep(1 second) in a tight loop in MockWebServer.

amirlivneh:
I think the test is also very slow in Travis CI with JDK8. It runs must faster with JDK9+.

yschimke:
@amirlivneh Any comments on the new test format?

okhttp/src/test/java/okhttp3/EventListenerTest.java
```diff
@@ -943,56 +948,77 @@ private void responseBodyFail(Protocol expectedProtocol) throws IOException {
...
 
       @Override public void writeTo(BufferedSink sink) throws IOException {
-        for (int i = 0; i < 1024; i++) {
-          sink.write(new byte[1024 * 256]);
-          sink.flush();
+        sink.write(new byte[1024 * 1024]);
+        sink.flush();
+        sink.write(new byte[1024 * 1024]);
+        sink.flush();
+        sink.write(new byte[1024 * 1024]);
```
amirlivneh: To cause the client to fail writing the request body, don't you have to ensure that the client tries to write at least once *after* the server disconnects? I don't see what guarantees that this is always the case here.

yschimke:
@amirlivneh take another look

okhttp/src/test/java/okhttp3/EventListenerTest.java
```diff
@@ -943,56 +947,80 @@ private void responseBodyFail(Protocol expectedProtocol) throws IOException {
...
-    };
+  @Test public void requestBodyFailHttp() {
+    requestBodyFail(null);
+  }
 
+  private void requestBodyFail(@Nullable Protocol requiredProtocol) {
     server.enqueue(new MockResponse()
         .setSocketPolicy(SocketPolicy.DISCONNECT_DURING_REQUEST_BODY));
 
+    NonCompletingRequest request = new NonCompletingRequest();
```
amirlivneh: nit: It's not really a request.

```suggestion
    NonCompletingRequestBody requestBody = new NonCompletingRequestBody();
```

PR #4826, opened 2019-03-30
Convert FaultHidingSink to Kotlin

shaishavgandhi05:


okhttp/src/main/java/okhttp3/internal/cache/FaultHidingSink.kt
```diff
@@ -0,0 +1,68 @@
...
+    }
+    try {
+      super.flush()
+    } catch (e: IOException) {
+      hasErrors = true
+      onException(e)
+    }
+  }
+
+  @Throws(IOException::class)
```
yschimke: should we remove these throws?

swankjesse: Yes.

shaishavgandhi05: 8fb70d3b7bfd0669a5bca5d52def811c15ab54a3

PR #4827, opened 2019-03-30
Convert OkHttpClient to Kotlin

swankjesse:


okhttp/src/main/java/okhttp3/OkHttpClient.kt
```diff
@@ -0,0 +1,827 @@
...
+  private val writeTimeout: Int = builder.writeTimeout
+  private val pingInterval: Int = builder.pingInterval
+  private val internalCache: InternalCache? = builder.internalCache
+  private val certificateChainCleaner: CertificateChainCleaner?
+
+  constructor() : this(Builder())
+
+  init {
+    var isTLS = false
+    for (spec in connectionSpecs) {
```
yschimke: Is this more idiomatic?

```
isTLS = connectionSpecs.any { it.isTLS }
```

okhttp/src/main/java/okhttp3/OkHttpClient.kt
```diff
@@ -0,0 +1,827 @@
...
+    }
+
+    if (sslSocketFactory != null) {
+      Platform.get().configureSslSocketFactory(sslSocketFactory)
+    }
+
+    this.certificatePinner = builder.certificatePinner
+        .withCertificateChainCleaner(certificateChainCleaner)
+
+    check(null !in (interceptors as List<Interceptor?>)) {
```
yschimke: Do we really need this check?  Should we trust the non null types.

okhttp/src/main/java/okhttp3/OkHttpClient.kt
```diff
@@ -0,0 +1,827 @@
...
+     * be used.
+     *
+     * @deprecated [SSLSocketFactory] does not expose its [X509TrustManager], which is a field that
+     *     OkHttp needs to build a clean certificate chain. This method instead must use reflection
+     *     to extract the trust manager. Applications should prefer to call
+     *     `sslSocketFactory(SSLSocketFactory, X509TrustManager)`, which avoids such reflection.
+     */
+    @Deprecated(
+        message = "Use the sslSocketFactory overload that accepts a X509TrustManager.",
+        level = ERROR
```
yschimke: Nice!

okhttp/src/main/java/okhttp3/OkHttpClient.kt
swankjesse: Unfortunately Java doesn't honor them. We've had bugs where users get nulls into these collections and the crashes are difficult to diagnose!

okhttp/src/main/java/okhttp3/OkHttpClient.kt
swankjesse: Yes! I wanna save for follow up. Would be excellent for IntelliJ to quick fix this for me.

PR #4829, opened 2019-03-30
Configure Dokka

swankjesse:


okhttp/src/main/java/okhttp3/Authenticator.kt
```diff
@@ -50,8 +49,7 @@ import java.io.IOException
  * return null; // Didn't find a preemptive auth scheme.
  * ```
  *
- * Reactive Authentication
- * -----------------------
```
swankjesse: Dokka doesn’t render this properly.

okhttp/src/main/java/okhttp3/Authenticator.kt
```diff
@@ -22,15 +22,14 @@ import java.io.IOException
  * **reactive** authentication after receiving a challenge from either an origin web server or proxy
  * server.
  *
- * Preemptive Authentication
- * -------------------------
+ * ## Preemptive Authentication
  *
  * To make HTTPS calls using an HTTP proxy server OkHttp must first negotiate a connection with
- * the proxy. This proxy connection is called a "TLS Tunnel" and is specified by [RFC
- * 2817](https://tools.ietf.org/html/rfc2817). The HTTP CONNECT request that creates this tunnel
```
swankjesse: Dokka can’t handle line wraps within links.

okhttp/src/main/java/okhttp3/ConnectionSpec.kt
```diff
@@ -196,7 +196,7 @@ class ConnectionSpec internal constructor(builder: Builder) {
       this.cipherSuites = null
     }
 
-    fun cipherSuites(vararg cipherSuites: CipherSuite) = apply {
+    fun cipherSuites(vararg cipherSuites: CipherSuite): Builder = apply {
```
swankjesse: https://github.com/Kotlin/dokka/issues/457

build.gradle
```diff
@@ -119,6 +123,16 @@ subprojects { project ->
     errorproneJavac 'com.google.errorprone:javac:9+181-r4173-1'
     errorprone 'com.google.errorprone:error_prone_core:2.3.3'
   }
+
+  apply plugin: 'org.jetbrains.dokka'
+  dokka {
```
ZacSweers: FYI, this may not work with the `gradle-maven-publish-plugin` in the sense that it may just redeclare its own configuration that overwrites it. Could be wrong though, but I was having issues around that in a separate project

ZacSweers: Some other useful ones 
- `Module.md` files
- `externalDocumentationLink` for dependencies

```gradle
externalDocumentationLink {
    url = new URL("https://docs.oracle.com/javase/7/docs/api/")
}
```
- `includeNonPublic = false`
- `outputFormat = 'html'`

ZacSweers: and for docs referencing other sibling projects as deps: https://github.com/Kotlin/dokka/issues/269#issuecomment-468733943

```gradle
evaluationDependsOn(":okhttp")
afterEvaluate {
  dokka {
    classpath = configurations.compile
    externalDocumentationLink {
      packageListUrl = new URL("file://${project(":okhttp").buildDir}/dokka/okhttp/package-list")
      url = new URL("file://${project(":okhttp").buildDir}/dokka/okhttp/")
    }
  }
}
```

swankjesse: thanks for the heads up

PR #4831, opened 2019-03-31
Migrate Request/Response + their body types to Kotlin

### LINKED ISSUE #4830: Migrate (Request/Response)(Body) to Kotlin
Request
RequestBody
Response
ResponseBody

I'll take these on!

ZacSweers:
Resolves #4830

Notes:
* Renamed first with no change to preserve git history if desired
* ~~Apparently Kotlin allows swizzling a java `isFoo()` method to a `isFoo` property accessor in Kotlin, meaning that changing `boolean isFoo()` in Java to `fun isFoo(): Boolean` in Kotlin is actually a potential breaking change for Kotlin consumers~~
  * EDIT: Keeping them as `val` in kotlin actually preserves APIs for both! `val isFoo: Boolean` is `isFoo()` in java and `.isFoo` in kotlin, so both are maintained. 
* One `url()` builder method said it would check and throw on http/https scheme, but doesn't actually in the implementation (noted inline). Can choose to add or change the doc here
* Opted for idiomatic implementations to avoid keeping around mutable/nullable intermediate states in some body functions
* Opted to move required properties of request/response to the constructor since the builder `build()` functions check them. I figured this was better than doing the check twice (once in the builder and again as a `!!` in the constructor), but we can also opt to move the checks to the constructor rather than the `build()` functions.


swankjesse:
```


> Task :okhttp-dnsoverhttps:compileKotlin
e: /home/travis/build/square/okhttp/okhttp-dnsoverhttps/src/main/java/okhttp3/dnsoverhttps/DnsOverHttps.kt: (218, 21): Function invocation 'isSuccessful()' expected
```

ZacSweers:
derp, will fix! Also - Builders aren't currently final, which fails the japi check. Assuming they should keep `open` to match, but want to clarify before since the actual request/response classes are final

ZacSweers:
> Builders aren't currently final, which fails the japi check. Assuming they should keep open to match, but want to clarify before since the actual request/response classes are final

Actually just demo'd it in f457f24. Can revert + adjust the japi settings if we don't want that here

ZacSweers:
Another consideration - for the `tag()` API, should it use `KClass` under the hood + extension function for reified overload like we did in KotlinPoet? Or punt that for another time?

swankjesse:
Definitely avoid behavior changes in the language conversion PRs. That means leaving it as `Class`.

ZacSweers:
I can't repro that test failure locally, though I haven't tried running jdk 11 locally before this either so maybe there's something I'm missing. I've tried with both setting 11 in the IDE's runner and via gradle with `./gradlew build -Dorg.gradle.java.home=/Library/Java/JavaVirtualMachines/openjdk-11.0.1.jdk/Contents/Home` (my jdk home for 11). Any chance CI could rerun in case it's flaky? Or re-run with verbose output?

ZacSweers:
Ok I was able to repro! But it looks unrelated to my changes here

```
java.lang.AssertionError: Expected exception type among [class javax.net.ssl.SSLHandshakeException], got javax.net.ssl.SSLPeerUnverifiedException: Hostname localhost not verified (no certificates)
	at okhttp3.RecordedResponse.assertFailure(RecordedResponse.java:154)
	at okhttp3.CallTest.anonCipherSuiteUnsupported(CallTest.java:1281)
	at okhttp3.OkHttpClientTestRule$1.evaluate(OkHttpClientTestRule.java:37)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.lang.Thread.run(Thread.java:834)
```

swankjesse:
@ZacSweers you can ignore that. There’s a problem with more recent JDKs.

ZacSweers:
Cool. Pushing a rebase now

swankjesse:
```


okhttp3.CookiesTest > cookieHandlerLikeAndroid FAILED
    java.lang.IllegalArgumentException: networkResponse.body != null
        at okhttp3.Response$Builder.checkSupportResponse(Response.kt:321)
        at okhttp3.Response$Builder.networkResponse(Response.kt:310)
        at okhttp3.internal.cache.CacheInterceptor.intercept(CacheInterceptor.java:126)
        at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:115)
        at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:90)
        at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:115)
        at okhttp3.internal.http.RetryAndFollowUpInterceptor.intercept(RetryAndFollowUpInterceptor.java:88)
        at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:115)
        at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:90)
        at okhttp3.RealCall.getResponseWithInterceptorChain(RealCall.kt:184)
        at okhttp3.RealCall.execute(RealCall.kt:67)
        at okhttp3.CookiesTest.get(CookiesTest.java:328)
        at okhttp3.CookiesTest.cookieHandlerLikeAndroid(CookiesTest.java:180)
```

okhttp/src/main/java/okhttp3/Request.kt
```diff
@@ -0,0 +1,236 @@
...
+class Request internal constructor(
+  internal val url: HttpUrl,
+  builder: Builder
+) {
+  internal val method: String = builder.method
+  internal val headers: Headers = builder.headers.build()
+  internal val body: RequestBody? = builder.body
+  internal val tags: Map<Class<*>, Any> = Util.immutableMap(builder.tags)
+
+  private val cacheControl: CacheControl by lazy {
```
swankjesse: does this allocate?

swankjesse: I’m anxious about performance regressions going from Java to Kotlin, and features like `by lazy` might be a bad fit

okhttp/src/main/java/okhttp3/Request.kt
```diff
@@ -0,0 +1,236 @@
...
+  fun headers(): Headers = headers
+
+  fun header(name: String): String? = headers[name]
+
+  fun headers(name: String): List<String> = headers.values(name)
+
+  fun body(): RequestBody? = body
+
+  /**
+   * Returns the tag attached with `Object.class` as a key, or `null` if no tag is attached with
```
swankjesse: Why are you putting null in backticks? Don’t do that.

We want to avoid backticks. Null is a normal word like tomato or taco or true, and it’s completely unambiguous without backticks.

okhttp/src/main/java/okhttp3/Request.kt
swankjesse: Keep it as-is for now, but it’s on my list for things to reconsider

okhttp/src/main/java/okhttp3/Request.kt
```diff
@@ -0,0 +1,236 @@
...
+    constructor() {
+      this.method = "GET"
+      this.headers = Headers.Builder()
+    }
+
+    internal constructor(request: Request) {
+      this.url = request.url
+      this.method = request.method
+      this.body = request.body
+      this.tags = if (request.tags.isEmpty())
```
swankjesse: curlies on ternary if and else please

okhttp/src/main/java/okhttp3/Request.kt
```diff
@@ -0,0 +1,236 @@
...
+
+    open fun url(url: HttpUrl): Builder = apply {
+      this.url = url
+    }
+
+    /**
+     * Sets the URL target of this request.
+     *
+     * @throws IllegalArgumentException if [url] is not a valid HTTP or HTTPS URL. Avoid this
+     * exception by calling [HttpUrl.parse]; it returns `null` for invalid URLs.
```
swankjesse: +4 on wrapped KDoc params

okhttp/src/main/java/okhttp3/Request.kt
```diff
@@ -0,0 +1,236 @@
...
+
+      return url(HttpUrl[finalUrl])
+    }
+
+    /**
+     * Sets the URL target of this request.
+     *
+     * @throws IllegalArgumentException if the scheme of [url] is not `http` or `https`.
+     */
+    // Code review TODO - this never checked the scheme before like the doc says. Should it?
```
swankjesse: I think it’ll explode on HttpUrl.get()

okhttp/src/main/java/okhttp3/Request.kt
```diff
@@ -0,0 +1,236 @@
...
+    open fun url(url: URL) = url(HttpUrl[url.toString()])
```
swankjesse: can you please remove the `operator` on HttpUrl.get()? It’s not a collection and this syntax is not good

okhttp/src/main/java/okhttp3/Request.kt
```diff
@@ -0,0 +1,236 @@
...
+    }
+
+    open fun get() = method("GET", null)
+
+    open fun head() = method("HEAD", null)
+
+    open fun post(body: RequestBody) = method("POST", body)
+
+    @JvmOverloads
+    open fun delete(body: RequestBody? = Util.EMPTY_REQUEST) = method("DELETE", body)
```
swankjesse: nice

okhttp/src/main/java/okhttp3/Request.kt
```diff
@@ -0,0 +1,236 @@
...
+
+    open fun put(body: RequestBody) = method("PUT", body)
+
+    open fun patch(body: RequestBody) = method("PATCH", body)
+
+    open fun method(method: String, body: RequestBody?): Builder = apply {
+      require(method.isNotEmpty()) {
+        "method.isEmpty() == true"
```
swankjesse: weird that the exception message is changing here. The `== true` part weirds me out also

okhttp/src/main/java/okhttp3/Request.kt
ZacSweers: It does allocate a `Lazy` instance. Can revert

okhttp/src/main/java/okhttp3/Request.kt
ZacSweers: We always put `{@code null}` in our codebase since it's kind of like a constant value ref, so just was doing it out of habit. Will fix

okhttp/src/main/java/okhttp3/Request.kt
```diff
@@ -0,0 +1,236 @@
...
+        tags.remove(type)
+      } else {
+        if (tags.isEmpty()) {
+          tags = mutableMapOf()
+        }
+        tags[type] = type.cast(tag)!! // Force-unwrap due to lack of contracts on Class#cast()
+      }
+    }
+
+    open fun build(): Request =  Request(
```
swankjesse: nit: weird spacing

okhttp/src/main/java/okhttp3/Request.kt
```diff
@@ -0,0 +1,236 @@
...
+import okhttp3.internal.Util
+import okhttp3.internal.http.HttpMethod
+import java.net.URL
+
+/**
+ * An HTTP request. Instances of this class are immutable if their [body] is `null` or itself
+ * immutable.
+ */
+class Request internal constructor(
+  internal val url: HttpUrl,
```
swankjesse: just curious: why does URL get constructor treatment but the other vals do not?

okhttp/src/main/java/okhttp3/RequestBody.kt
```diff
@@ -0,0 +1,153 @@
...
+import kotlin.text.Charsets.UTF_8
+
+abstract class RequestBody {
+
+  /** Returns the Content-Type header for this body. */
+  abstract fun contentType(): MediaType?
+
+  /**
+   * Returns the number of bytes that will be written to `sink` in a call to [writeTo],
+   * or `-1` if that count is unknown.
```
swankjesse: please stop backticking things

okhttp/src/main/java/okhttp3/Request.kt
ZacSweers: d853c19

okhttp/src/main/java/okhttp3/Request.kt
ZacSweers: 24a7ae6

okhttp/src/main/java/okhttp3/RequestBody.kt
```diff
@@ -0,0 +1,153 @@
...
+  abstract fun writeTo(sink: BufferedSink)
+
+  /**
+   * A duplex request body is special in how it is **transmitted** on the network and
+   * in the **API contract** between OkHttp and the application.
+   *
+   * This method returns false unless it is overridden by a subclass.
+   *
+   * Duplex Transmission
+   * -------------------
```
swankjesse: Sadly Dokka can’t handle this.
https://github.com/square/okhttp/pull/4829

okhttp/src/main/java/okhttp3/RequestBody.kt
```diff
@@ -0,0 +1,153 @@
...
+   *
+   * With regular HTTP calls the request always completes sending before the response may begin
+   * receiving. With duplex the request and response may be interleaved! That is, request body bytes
+   * may be sent after response headers or body bytes have been received.
+   *
+   * Though any call may be initiated as a duplex call, only web servers that are specially
+   * designed for this nonstandard interaction will use it. As of 2019-01, the only widely-used
+   * implementation of this pattern is
+   * [gRPC](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md).
```
swankjesse: wanna do a labeled link at the bottom? I switched most of the codebase over to that style with the Dokka PR

okhttp/src/main/java/okhttp3/RequestBody.kt
```diff
@@ -0,0 +1,153 @@
...
+   *
+   * Because the encoding of interleaved data is not well-defined for HTTP/1, duplex request
+   * bodies may only be used with HTTP/2. Calls to HTTP/1 servers will fail before the HTTP request
+   * is transmitted. If you cannot ensure that your client and server both support HTTP/2, do not
+   * use this feature.
+   *
+   * Duplex APIs
```
swankjesse: Maybe add `###` ? Weird otherwise

okhttp/src/main/java/okhttp3/RequestBody.kt
swankjesse: Use `###` instead

okhttp/src/main/java/okhttp3/ResponseBody.kt
```diff
@@ -0,0 +1,250 @@
...
+      closed = true
+      delegate?.close() ?: run { source.close() }
+    }
+  }
+
+  companion object {
+
+    /**
+     * Returns a new response body that transmits [content]. If `contentType` is non-null
+     * and lacks a charset, this will use `UTF-8`.
```
swankjesse: `please`, don’t `backtick` things unless `they` had `backticks` previously!

okhttp/src/main/java/okhttp3/ResponseBody.kt
```diff
@@ -0,0 +1,250 @@
...
+     */
+    @JvmStatic
+    fun create(contentType: MediaType?, content: String): ResponseBody {
+      val (charset, finalContentType) = contentType?.charset()
```
swankjesse: This is too fancy for me. Could you just use regular control flow?

swankjesse: Also – regular control flow doesn’t allocate!

okhttp/src/main/java/okhttp3/Request.kt
ZacSweers: kk, will remove this line and leave as is

okhttp/src/main/java/okhttp3/Request.kt
ZacSweers: Wow, that was an autoconversion thing I missed and I'm not sure I even know what it's doing 🤔.

ZacSweers: Should they be removed from HttpUrl then? 

image

okhttp/src/main/java/okhttp3/Request.kt
ZacSweers: I was matching the referenced check, would you rather keep the `.length == 0`?

okhttp/src/main/java/okhttp3/Request.kt
ZacSweers: Because this is the only one that is technically nullable in the builder property but required to not be by the time it gets here. It's checked (and thus smart casted safely) in the builder, but we could move that check to the constructor here instead if you'd rather keep it single-param

okhttp/src/main/java/okhttp3/RequestBody.kt
ZacSweers: TIL you can do that in kdocs!

okhttp/src/main/java/okhttp3/Request.kt
ZacSweers: 070f7a2

okhttp/src/main/java/okhttp3/Request.kt
ZacSweers: 988aee8

okhttp/src/main/java/okhttp3/Request.kt
ZacSweers: cf36f06

okhttp/src/main/java/okhttp3/Request.kt
ZacSweers: d63759b

okhttp/src/main/java/okhttp3/Request.kt
ZacSweers: 4d7bc7a

okhttp/src/main/java/okhttp3/RequestBody.kt
ZacSweers: fb3b296

okhttp/src/main/java/okhttp3/RequestBody.kt
ZacSweers: 619032c

okhttp/src/main/java/okhttp3/RequestBody.kt
ZacSweers: e239395

okhttp/src/main/java/okhttp3/RequestBody.kt
ZacSweers: I spend way too much time figuring out a regex to find all instances of these only to find that this was the only one

7ead57b

okhttp/src/main/java/okhttp3/ResponseBody.kt
ZacSweers: Part of fb3b296

okhttp/src/main/java/okhttp3/ResponseBody.kt
ZacSweers: 79622d0, though I kind of think this is fine aside from the extra allocation. The updated version feels less readable than the kotlin code since I'm now having to think about 2 levels of control flow, and default values on mutable possibly-rewritten variables

PR #4832, opened 2019-03-31
 Replace 'for loop' with 'foreach'

TangHuaiZhe:
- Replace 'for loop' with 'foreach'
- Two Anonymous classes replaced with lambda


swankjesse:
We’re aware of foreach. It allocates. We’re avoiding allocations.

TangHuaiZhe:
ok, you are right

PR #4833, opened 2019-03-31
New KotlinSourceCompatibilityTest

swankjesse:
As we start to take advantage of Kotlin language features this test will keep
us from inadvertently breaking existing callers.


swankjesse:
Still lots of stuff to cover here.

swankjesse:
We've got lots of code to exercise behavior. This is intended to avoid is accidentally dropping an accessor or something.

PR #4835, opened 2019-03-31
Check different JDKs in CircleCI

yschimke:


yschimke:
@swankjesse Thanks, going to leave this stream of work at this point.  I suggest giving it a few weeks, and then making a call to stick with travis or switch publishing also to circleci. 

There is also a lot of cleanup that is possible, reuse of jobs and workflow.  But it's faster, easier to navigate to test failures, so I'm guessing changing is worth it.

swankjesse:
Great! I like the idea of switching.

PR #4839, opened 2019-03-31
Convert RealInterceptorChain to Kotlin

shaishavgandhi05:


okhttp/src/main/java/okhttp3/internal/http/RealInterceptorChain.kt
```diff
@@ -0,0 +1,130 @@
...
+  override fun withWriteTimeout(timeout: Int, unit: TimeUnit): Interceptor.Chain {
+    val millis = checkDuration("timeout", timeout.toLong(), unit)
+    return RealInterceptorChain(interceptors, transmitter, exchange, index, request, call,
+        connectTimeout, readTimeout, millis)
+  }
+
+  fun transmitter(): Transmitter = transmitter
+
+  fun exchange(): Exchange {
+    if (exchange == null) throw IllegalStateException()
```
oldergod: you probably could `return checkNotNull(exchange)`

PR #4844, opened 2019-04-01
Add Module.md files for different artifacts with module and package info

ZacSweers:
This is what dokka uses to document packages and modules. Deleted the `package-info.kt` file as it doesn't do anything on its own, but can restore if it's still wanted

<kbd>
  
</kbd>

<kbd>
  
</kbd>

### DIFF
--- a/okhttp/Module.md
+++ b/okhttp/Module.md
@@ -1,0 +1,7 @@
+# Module okhttp
+
+An HTTP+HTTP/2 client for Android and Java applications.
+
+# Package okhttp3
+
+An HTTP+HTTP/2 client for Android and Java applications.
--- a/okhttp/Module.md
+++ b/okhttp/Module.md
@@ -1,7 +1,3 @@
 # Module okhttp
 
 An HTTP+HTTP/2 client for Android and Java applications.
-
-# Package okhttp3
-
-An HTTP+HTTP/2 client for Android and Java applications.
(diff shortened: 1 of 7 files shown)


okhttp/Module.md
```diff
@@ -0,0 +1,7 @@
+# Module okhttp
+
+An HTTP+HTTP/2 client for Android and Java applications.
+
+# Package okhttp3
+
+An HTTP+HTTP/2 client for Android and Java applications.
```
swankjesse: What's the utility in repeating these sentences? We should pick one.

ZacSweers: One's for the module, and you can have any number of packages with their own. Basically the package lines are the replacement for package-info documentation, and the module line is per-module documentation (either a blurb or you could make it fully filled like an embedded README)

Example of multiple packages

```markdown
# Module okhttp

An HTTP+HTTP/2 client for Android and Java applications.

# Package okhttp3

An HTTP+HTTP/2 client for Android and Java applications.

# Package okhttp3.hypothetical

A hypothetical other package doc.

```

<kbd>
  
</kbd>

swankjesse: Can we defer that until we actually have a module with multiple packages?

ZacSweers: That works. Then it looks like this

<kbd>
  
</kbd>

ZacSweers: Or alternatively you could have the package only and not the module. I did modules only in 3dc147c but should have asked first. Lmk which you prefer

<kbd>
  
</kbd>

PR #4845, opened 2019-04-01
Specify jdk 8 for dokka

ZacSweers:
Since the project targets java 8


ZacSweers:
Will dig in the issue, works locally for me when I just run `./gradlew dokka` 🤔

PR #4848, opened 2019-04-01
Convert CertificatePinner to a data class

amirlivneh:


swankjesse:
I'm curious what happens to the copy method since Pin is not a public type.

amirlivneh:
Screen Shot 2019-04-01 at 5 35 46 AM

PR #4852, opened 2019-04-01
 Convert Dispatcher.java to Kotlin

Egorand:


Egorand:
How are we dealing with cases when the code in `okhttp-tests` tries to access types/methods that were package-private before? `internal` doesn't work here, are we just making them `public`?

amirlivneh:
> How are we dealing with cases when the code in okhttp-tests tries to access types/methods that were package-private before? internal doesn't work here, are we just making them public?

From what I've seen, although IntelliJ inspection reports an error when trying to access an internal class from a test file, these instances do compile fine.

swankjesse:
InternalKt.kt is what we've used this far. My hope is that when the full conversion is done that file will be empty or almost empty.

Egorand:
@amirlivneh `okhttp-tests` isn't supposed to access `internal` code from `okhttp`, since these are separate modules. I guess you mean `okhttp/src/test/java` accessing code inside `okhttp/src/main/java` - that should indeed work and I vaguely recall there being an IDE issue related to it.

@swankjesse cool, I'll look into it and update the PR. Currently it fails cause Japicmp reports visibility changes.

amirlivneh:
@Egorand, yes, it was a test that lives in _okhttp/src/test/java/okhttp3_, which is also where _DispatcherTest.java_ lives. I was under the impression that the okhttp-tests module was removed in https://github.com/square/okhttp/commit/9837fea7684eb5b964e736c45463bfca2de361c7, but I may be missing something.

okhttp/src/main/java/okhttp3/Dispatcher.kt
```diff
@@ -0,0 +1,234 @@
...
+  }
+
+  /** Returns a snapshot of the calls currently awaiting execution.  */
+  @Synchronized fun queuedCalls(): List<Call> {
+    return Collections.unmodifiableList(readyAsyncCalls.map { it.get() })
+  }
+
+  /** Returns a snapshot of the calls currently being executed.  */
+  @Synchronized fun runningCalls(): List<Call> {
+    return Collections.unmodifiableList(runningSyncCalls + runningAsyncCalls.map { it.get() })
```
oldergod: method reference? might be longer but I kind of remember the method counts to be lower?

Egorand:
@amirlivneh oh, you're right, I've missed the fact that the tests moved. It is weird then, I'll investigate. The test failed on CI so I guess it's a real failure and not a buggy IDE check.

Egorand:
Oh well, this is actually easy, the `DisptacherTest` is still in Java, hence it can't access `internal` members. I'll move those tests into `DispatcherKotlinTest`.

amirlivneh:
@Egorand, I was able to access the class `CertificatePinner.Pin`, which is internal, from _CertificatePinnerTest.java_.

Egorand:
@amirlivneh the failure is for an `internal` method. Looks like `internal` types can be accessed by Java code, but not methods, since Kotlin jumbles their names.

okhttp/src/main/java/okhttp3/Dispatcher.kt
Egorand: The code gets inlined so there should be no effect on method count. Also, bytecode for both versions looks pretty similar.

okhttp/src/test/java/okhttp3/RecordingExecutor.kt
```diff
@@ -0,0 +1,57 @@
+package okhttp3
```
swankjesse: nit: copyright

Egorand: Fixed!

PR #4854, opened 2019-04-01
Use the build cache in CircleCI

yschimke:
TL;DR it probably saves a minute per build on average, and we could debug to improve.

Aggressive build caching within gradle and circleci.  Not sure we want this, it makes re-running a flaky test potentially awkward. It should be safe, but ultimately we want any snapshot or tagged release publishing step to do a fresh build I assume.

Normal: 5:40
https://circleci.com/workflow-run/7c6cd465-9609-41bd-b63c-d98334c9a027
image

Cache First Run: 5:41
https://circleci.com/workflow-run/a69206f5-ef50-415b-ba90-b0f97fefa33b
image

Second Run: 4:53
https://circleci.com/workflow-run/0a782b8a-5db8-4f4b-b76d-9792d9fef0fd
image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,14 +1,14 @@
 # Check https://circleci.com/docs/2.0/language-java/ for more details
 version: 2.1
 jobs:
-  dependencies:
+  compile:
     docker:
       - image: circleci/openjdk:8u171-jdk
 
     working_directory: ~/repo
 
     environment:
-      JVM_OPTS: -Xmx1g
+      JVM_OPTS: -Xmx768m
       TERM: dumb
 
     steps:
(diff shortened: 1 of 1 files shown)


yschimke:
Sorry, should be a draft PR

yschimke:
n.b. times should get better as a master build should populate the cache for branch builds.  We can have a cache key command that generates a cache key from all build.gradle files, maybe combined with the current date.

Otherwise the cache becomes less relevant over time

image

JakeWharton:
There shouldn't really be a need to use the date. I mean we shouldn't even
need to use the build file hash either except the cache would just grow
indefinitely. Maybe just the year and month as a key?

yschimke:
https://guides.gradle.org/using-build-cache/

Lot of things to investigate e.g. debug the internal cache keys, since the environment variables probably kill the test result caching

image

yschimke:
@JakeWharton what would you be comfortable caching across PRs?  e.g. would you like it if gradle only ran the tests for modules we had changed?  Should the target be the 10-20 second build & test times you can often experience locally?

yschimke:
Also, once you have written the cache with a key it is immutable.  And restores are prefix based, so it's nuanced in an awkward way.

JakeWharton:
Compilation for changed modules only. Tests for changed or downstream
(i.e., depends on a re-compiled module, potentially transitively).

yschimke:
OK, leave open and I'll dig deeper tomorrow.

yschimke:
With a change in a smaller test module, it's down to 2 minutes for a followup build

https://circleci.com/workflow-run/2f31ac40-681f-4d5f-bd32-d688cf512232

image

image

PR #4860, opened 2019-04-02
Convert BasicCertificateChainCleaner to Kotlin

shaishavgandhi05:


okhttp/src/main/java/okhttp3/internal/tls/BasicCertificateChainCleaner.kt
```diff
@@ -0,0 +1,124 @@
...
+
+  /**
+   * Returns a cleaned chain for [chain].
+   *
+   * This method throws if the complete chain to a trusted CA certificate cannot be constructed.
+   * This is unexpected unless the trust root index in this class has a different trust manager than
+   * what was used to establish [chain].
+   */
+  @Throws(SSLPeerUnverifiedException::class)
+  override fun clean(chain: MutableList<Certificate>, hostname: String): MutableList<Certificate> {
```
JakeWharton: Parameter and return should be `List`.

shaishavgandhi05: db2e652e414d93a134691ab80ffd4ccaadd06d7e

PR #4861, opened 2019-04-03
More coverage in KotlinSourceCompatibilityTest

swankjesse:


okhttp/src/test/java/okhttp3/KotlinSourceCompatibilityTest.kt
```diff
@@ -67,6 +85,13 @@ class KotlinSourceCompatibilityTest {
     val certificatePinner: CertificatePinner? = address.certificatePinner()
   }
 
+  @Test @Ignore
```
Egorand: Are these just proving that the code compiles and is idiomatic?

yschimke:
This fails to compile against 3.14

```

> Task :compileTestKotlin
e: /Users/yuri/workspace/okurl/src/test/kotlin/okhttp/KotlinSourceCompatibilityTest.kt: (148, 16): Object is not abstract and does not implement abstract member public abstract fun isExecuted(): Boolean defined in okhttp3.Call
e: /Users/yuri/workspace/okurl/src/test/kotlin/okhttp/KotlinSourceCompatibilityTest.kt: (153, 7): 'isExecuted' overrides nothing
e: /Users/yuri/workspace/okurl/src/test/kotlin/okhttp/KotlinSourceCompatibilityTest.kt: (154, 7): 'isCanceled' overrides nothing
e: /Users/yuri/workspace/okurl/src/test/kotlin/okhttp/KotlinSourceCompatibilityTest.kt: (360, 7): 'requestFailed' overrides nothing
e: /Users/yuri/workspace/okurl/src/test/kotlin/okhttp/KotlinSourceCompatibilityTest.kt: (365, 7): 'responseFailed' overrides nothing
e: /Users/yuri/workspace/okurl/src/test/kotlin/okhttp/KotlinSourceCompatibilityTest.kt: (387, 34): Type mismatch: inferred type is MediaType? but MediaType was expected
```

okhttp/src/test/java/okhttp3/KotlinSourceCompatibilityTest.kt
swankjesse: Exactly. I wanna run these against 3.14 and make sure they pass there and on 4.0.

swankjesse:
@yschimke the motivation of this PR is to find those kinds of regressions! Once I've got the entire API I'd like to do exactly what you're doing, and fix the test to match 3.14. Then fix the code to pass the test.

I already know of a few places where I think we should deliberately break source compatibility:
 - List/MutableList
 - Ambiguous nullability to unambiguous nullability

I want all of these differences to be known and to not surprise us in the release.

Next steps:
1. Finish coverage. Includes MockWebServer, LoggingInterceptor, TLS, etc.
2. Get it passing on a branch of 3.14, probably requiring some changes.
3. Anywhere that 4.0 and 3.14 are different is a source compatibility regression. On a case-by-case basis fix these in code or document them.

yschimke:
@swankjesse apologies for jumping the gun :)

yschimke:
A big +1 to breaking Kotlin source slightly and getting it right.  It will only get more painful to do later if you don't now.

Will kotlin still be binary compatible?

swankjesse:
Yep, still binary compatible. That’s the most important one ’cause it impacts transitive dependencies quite dramatically.

Issue #1706, opened 2015-06-16
Support Link-Local IPv6 with Scope in URLs


yydl:
[IPv6 link local addresses](https://en.wikipedia.org/wiki/IPv6_address#Link-local_addresses_and_zone_indices) need to specify a scope ID in order to be routed correctly. I tried making requests with URLs that contain this scope ID but OkHttp doesn't accept the url as it doesn't know how to parse it correctly.

First I tried by escaping the percent sign before the scope:

```
http://[fe80::45c:d71:aec1:9ec3%25eth0]:7001/path.file
```

This resulted in an "IllegalArgumentException: unexpected url" error

Then I tried just putting in the percent sign as is:

```
http://[fe80::45c:d71:aec1:9ec3%eth0]:7001/path.file
```

Once again, I got the same error.

JakeWharton:
Interesting feature. Never knew this was a thing.

swankjesse:
Explanation: https://msdn.microsoft.com/en-us/library/aa917150.aspx
Proposed Standard: https://tools.ietf.org/html/rfc6874
Chrome bug report: https://code.google.com/p/chromium/issues/detail?id=70762
Firefox bug report: https://bugzilla.mozilla.org/show_bug.cgi?id=700999

Interesting constraint on these URLs:
> To limit this risk, implementations MUST NOT allow use of this format except for well-defined usages, such as sending to link-local addresses under prefix fe80::/10.  At the time of writing, this is the only well-defined usage known.

Next steps:
1. We'll accept a pull request to add link-local support to HttpUrl, as long as it's accompanied by appropriate tests. Preferably link local address tests are also submitted upstream to [web-platform-tests](https://github.com/w3c/web-platform-tests/tree/master/url).
2. We'll accept a pull request to honor those `HttpUrl` instances when selecting an interface for a HTTP connection. There's some nuance around link-local addresses and proxies - we connect through the proxy so that's where the scope goes.

Those interested in seeing this feature should volunteer to write those pull requests. Until then I'm putting this feature request on ice.

marsounjan:
Hi @swankjesse, I have made a PR to support IPv6 link local addresses with interface scopes. Could you please check it out? :)

swankjesse:
Guava implemented support by always discarding the scope ID:
https://github.com/google/guava/commit/1d3780724431da91b28bbf50fcb13f27459f4710

yschimke:
Closing as we rejected this PR https://github.com/square/okhttp/pull/4865

PR #4865, opened 2019-04-03
Support for IPv6 link local

marsounjan:
https://github.com/square/okhttp/issues/1706

### DIFF
--- a/okhttp/src/main/java/okhttp3/HttpUrl.kt
+++ b/okhttp/src/main/java/okhttp3/HttpUrl.kt
@@ -1246,6 +1246,36 @@ class HttpUrl internal constructor(builder: Builder) {
         }
       }
 
+      /*
+      * add IPv6 interface scope to host when base host and new host are IPv6 link local addresses
+      * used mainly for correct redirection while using link local addresses
+       */
+      this.host?.also { host ->
+        if(base != null
+                && base.host.contains(":")
+                && (base.host.startsWith("fe80::") || base.host.startsWith("fe80::", 1))
+                && base.host.contains("%")
+                && host.contains(":")
+                && !host.contains("%")
+                && (host.startsWith("fe80::") || host.startsWith("fe80::", 1))
+        ){
+          val interfaceScope = base.host.substring(
+                  base.host.indexOf("%"),
+                  if(base.host.endsWith("]")){
+                    base.host.length - 1
+                  } else {
+                    base.host.length
+                  }
+          )
+
+          this.host = if(host.endsWith("]")){
+            host.replace("]", "$interfaceScope]")
+          } else {
+            "$host$interfaceScope"
+          }
+        }
+      }
+
       // Resolve the relative path.
       val pathDelimiterOffset = delimiterOffset(input, pos, limit, "?#")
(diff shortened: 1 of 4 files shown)


swankjesse:
This is cool!

We aren't in a position to merge new features now, and for about 8 weeks. We're doing a Kotlin conversion and that is complex enough that I don't want to make any other changes alongside it.

Can you explain how you found yourself using this? I read it in the spec but really want context around how this gets used in practice.

marsounjan:
I work for a company which makes network appliances. Using IPv6 link local address solves for example the connection issues when you take the device configured with static IPv4 address from some network and plug it into you LAN with different subnet configured.
With IPv6 Link Local you can always communicate with the device when it's plugged in LAN :)

swankjesse:
Now I'm confused. I was expecting linklocal to be about the local interface. I need to do some studying.

In general I'm somewhat reluctant to do this because it opens up attack vectors.

marsounjan:
it's definitely about the local interface. To use link local address you must specify the interface.

I don't see how it could open up attack vendors.

swankjesse:
An attacker server could redirect a client to DoS a server that isn’t on the public internet.

yschimke:
Util.java change will need to target kotlin.  Seems like something we could support via caller urls (with some risk), but disallow for automatic redirects.

swankjesse:
I'm gonna close this because I'm not ready to impose the security consequences of link local on all of our users.

marsounjan:
@swankjesse @yschimke we'd really like okhttp to support link local, since it's really pain for us updating our local fork every time when we need to update okhttp version..

So what's exactly the problem you see here? If it's just you worring about being redirected to different network interface, it can be solved quite easily. 

Also I can see several other users would like to have this feature in okhttp :)

marsounjan:
well, I wanted to write that it'll be easy to implement it the way that doesn't allow to switch interfaces by redirections, because that doesn't make much sense anyway. 

But then I look at the code and found out that it's already implemented that way :)

Is not allowing redirects to switch network interfaces for traffic enough for you to stop worrying about security?
Other then that I really can't imagine any possible security risk or anything like that since it's behaving the same as IPv6

swankjesse:
I don't want new calls to switch interfaces either. Imagine crawling the web with OkHttp and a malicious page attempts to get the crawler to index URLs on it's internal interface.

marsounjan:
is this really a security issue which should be okhttp client library worried about? Isn't it more about good/bad crawler implementation?

I am not proposing okhttp to support some custom hacky protocol, just RFC documented addressing :)

swankjesse:
Do browsers support this? If they do we should; if they don't we shouldn't.

swankjesse:
Looks like the answer is no.
https://stackoverflow.com/a/46881540/40013

swankjesse:
Also! Not supporting link local addresses URLs does not mean OkHttp cannot access the resources at those URLs! You just need to specify a SocketFactory that chooses the link you want. That's not too much code and it keeps this feature safe.

yschimke:
@marsounjan What's the typical use case for this?  Any examples of where you need it for typical development or app uses?

PR #4866, opened 2019-04-03
Adjust CircleCI caching

yschimke:
Master compile cache is built once a day by master compile job or each commit on master.
Each Test build uses a previous run on the same branch, or the Master compile cache.

Experience should generally be

PR: check and test jobs start immediately using the master build cache (or previous good run).
Master commit: Runs 3 test jobs which use master build cache.  Compile updates the cache without growing.
Daily: All jobs run and Master compile cache is updated.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,7 +23,7 @@ jobs:
             - ~/.gradle/caches
             - ~/.gradle/wrapper
           # Under normal usage, saves compiled results from master once a day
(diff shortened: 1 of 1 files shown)


yschimke:
My one concern is whether we will skip too many tests to get good signal on flaky tests.  We may want to ensure that daily jobs, maybe master commit jobs always run all tests.

yschimke:
Landing to continue testing in CI, cc @swankjesse

PR #4868, opened 2019-04-03
Conscrypt 2.1.0 upgrade

yschimke:


yschimke:
Questions that the upgrade notes raise

https://groups.google.com/forum/#!topic/conscrypt/BKFfRkMvlBo

- Should we use more of Conscrypt support e.g. hostname verification?
- Overall is there functionality hidden behind optionalmethod in various places that can now be called directly in JDK 8?
- Are some of the excluded tests now passable? e.g. fallback.

.circleci/config.yml
```diff
@@ -234,7 +234,7 @@ workflows:
       - testconscrypt:
           filters:
             branches:
-              only: master
+              ignore: master
```
yschimke: To be reverted

swankjesse:
We shouldn't use Conscrypt hostname verification. We might also want to confirm it isn't enabled. There's little benefit to doing it twice, and we need to do it in our later anyway because other TLS backends won't do it for us.

swankjesse:
Replacing OptionalMethod stuff is good. I'm reluctant to make such behavior changes until 4.0 lands. I don't want that release to include anything risky other than Kotlin! And for that reason I'd also like to sit on this until that is out. Should be a few weeks.

swankjesse:
And hooray for fixed tests!

PR #4874, opened 2019-04-04
Convert MultipartBody to Kotlin

swankjesse:


okhttp/src/main/java/okhttp3/MultipartBody.kt
```diff
@@ -0,0 +1,298 @@
...
+    }
+
+    /** Add a part to the body.  */
+    fun addPart(part: Part) = apply {
+      parts.add(part)
+    }
+
+    /** Assemble the specified parts into a request body.  */
+    fun build(): MultipartBody {
+      require(!parts.isEmpty()) { "Multipart body must have at least one part." }
```
JakeWharton:
```suggestion
      require(parts.isNotEmpty()) { "Multipart body must have at least one part." }
```

okhttp/src/main/java/okhttp3/MultipartBody.kt
```diff
@@ -0,0 +1,298 @@
...
+     * Appends a quoted-string to a StringBuilder.
+     *
+     * RFC 2388 is rather vague about how one should escape special characters in form-data
+     * parameters, and as it turns out Firefox and Chrome actually do rather different things, and
+     * both say in their comments that they're not really sure what the right approach is. We go
+     * with Chrome's behavior (which also experimentally seems to match what IE does), but if you
+     * actually want to have a good chance of things working, please avoid double-quotes, newlines,
+     * percent signs, and the like in your field names.
+     */
+    internal fun appendQuotedString(target: StringBuilder, key: String) {
```
JakeWharton: Could be extension if you want

okhttp/src/main/java/okhttp3/MultipartBody.kt
swankjesse: Fixed.

okhttp/src/main/java/okhttp3/MultipartBody.kt
swankjesse: Good call. Fixed.

JakeWharton:
Uh. Something weird happened here. Rebase?

JakeWharton:
Oh you just sorted the list. If you want to be cute we can add a check to make sure it's sorted.

okhttp/src/main/java/okhttp3/MultipartBody.kt
```diff
@@ -0,0 +1,302 @@
...
+        return Part(headers, body)
+      }
+
+      @JvmStatic
+      fun createFormData(name: String, value: String): Part =
+          createFormData(name, null, RequestBody.create(null, value))
+
+      @JvmStatic
+      fun createFormData(name: String, filename: String?, body: RequestBody): Part {
+        val disposition = StringBuilder("form-data; name=")
```
JakeWharton: Could be `buildString`, or we can clean them all up later

swankjesse: Later!

PR #4881, opened 2019-04-05
Update japicmp plugin

JakeWharton:
This bumps japicmp to 0.13.1 which still doesn't contain the effectively-final fix, but at least provides compatibility with newer japicmp versions so that we can upgrade once that change is released.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -54,7 +54,7 @@ buildscript {
 plugins {
   id 'ru.vyarus.animalsniffer' version '1.5.0'
   id 'com.github.johnrengelman.shadow' version '4.0.1'
-  id 'me.champeau.gradle.japicmp' version '0.2.6'
+  id 'me.champeau.gradle.japicmp' version '0.2.8'
 }
 
 allprojects {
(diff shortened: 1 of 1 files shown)


JakeWharton:
So updating this checks the classfile version which is changing from 52 (Java 8) to 50 (Java 6). Changing Kotlin to target 1.8 bytecode produce references to missing methods like `Long.hashCode` in the bytecode which fails animal sniffer. These methods as generated by `kotlinc` will be desugared by R8 to all API levels, but unfortunately they forgot `Long.hashCode`. I added support for it this week into R8 but it won't be available in the canary for two weeks and not in stable until likely June-ish. Even then it will require that people be using AGP 3.5 and will otherwise silently fail at runtime on pre-API 24.

So... since there's no japicmp version with effectively-final yet (the primary motivation for updating the plugin), I'm going to punt on this for now and just leave things as is.

swankjesse:
Oooh, that’s super annoying. Thanks for fixing R8.

swankjesse:
Closing for now.

PR #4884, opened 2019-04-05
Proposal: use gradle-consistent-versions to enable easy CR of dependency changes

iamdanfox:
As a pretty widely-used library, any changes to OkHttp's transitive dependencies can have quite a far-reaching impact.

We've started using a plugin called [gradle-consistent-versions](https://github.com/palantir/gradle-consistent-versions) to make it easier to code-review changes like this. The gist is you:

1. define what versions you want in `versions.props` 
1. then run `./gradlew --write-locks` to write them to `versions.lock`.

This `versions.lock` file gives you a bird's eye view of all version numbers in your repo.  The plugin also means you can use a slightly more vanilla style of declaring dependencies and don't need any intermediate variable names e.g.

```gradle
dependencies {
   implementation 'com.google.guava:guava'
}
```

I realise you've just moved to Gradle, so probably don't want to rock the boat with new plugins too quickly - mainly just put this up to see if you think it might be useful to OkHttp!

cc @dansanduleac


versions.lock
```diff
@@ -0,0 +1,44 @@
...
+com.github.jnr:jnr-enxio:0.20 (1 constraints: d20cde08)
+com.github.jnr:jnr-ffi:2.1.9 (3 constraints: d722dd54)
+com.github.jnr:jnr-posix:3.0.47 (1 constraints: 3c0d2523)
+com.github.jnr:jnr-unixsocket:0.22 (1 constraints: d804f130)
+com.github.jnr:jnr-x86asm:1.0.2 (1 constraints: e909bfb2)
+com.google.android:android:4.1.1.4 (1 constraints: 6a05a340)
+com.google.code.findbugs:jsr305:3.0.2 (3 constraints: c81702a8)
+com.google.errorprone:error_prone_annotations:2.2.0 (1 constraints: 160aebb4)
+com.google.guava:failureaccess:1.0.1 (1 constraints: 140ae1b4)
+com.google.guava:guava:27.0.1-jre (2 constraints: 250f4b81)
```
iamdanfox: Running `./gradlew why --hash 250f4b81` explains what's going on here

```
> Task :why
com.google.guava:guava:27.0.1-jre
        projects -> 27.0.1-jre
        io.airlift:airline -> 21.0
```

If a new airline version happened to pull in a newer version of guava, Gradle would pick the newer version for you and you'd see the change reflected here.

JakeWharton:
So if we assume that there's a problem with our transitive dependencies (which I'm not necessarily convinced there is), how is this different than Gradle's built-in dependency locking?

Also, as far as I can tell, your plugin requires that the project resolve version conflicts globally across all configurations whereas Gradle normally resolves these on a per-configuration basis. In the airlift example, that is only used by the okcurl project and Guava is only used by tests. So if airlift updates Guava, every project sees the dependency update on their test classpath rather than only the okcurl project on its compilation and test classpaths.

iamdanfox:
I just floated the PR as I saw you'd just moved to Gradle and we'd found the plugin beneficial for some of our libraries - not that we'd encountered problems with OkHttp's transitive dependencies!

I think you've hit the nail on the head - gradle-consistent-versions tries to align the same jar across projects, and with that premise it means you can have one single lock file instead of 10. Having the constraints included in the file are just for convenience in CRs.

okhttp/build.gradle
```diff
@@ -22,23 +22,23 @@ task copyJavaTemplates(type: Copy) {
...
 
 dependencies {
-  api deps.okio
-  api deps.kotlinStdlib
-  compileOnly deps.conscrypt
-  compileOnly deps.android
-  compileOnly deps.jsr305
-  compileOnly deps.animalSniffer
+  api 'com.squareup.okio:okio'
+  api "org.jetbrains.kotlin:kotlin-stdlib:${versions.kotlin}"
```
yschimke: why versions here?

yschimke:
@iamdanfox Can you clarify one thing, does the eventual gradle build have the same dependency structure?  This looks like a win, particularly when combined with plugins to upgrade dependencies so we could track okio, conscrypt etc.

I think I understand Jake's concern, but I'm swayed by how clean this makes things.  FWIW I'm OCD like on updating deps across OS projects I tinker, as a user of com.github.ben-manes.versions, this looks great and a good step towards having automatic updates/resolutions just accepted by running a script and landing a PR.

yschimke:
I'm sold on this an am applying to my other projects.  Thanks @iamdanfox.  But no implications for okhttp itself landing this PR.

okhttp/build.gradle
iamdanfox: I kept this one specific version here because I figured you might want to keep the stdlib in sync with the gradle plugin so you only have one place to bump versions:

```
    classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:${versions.kotlin}"
```

If this isn't an important constraint then we could absolutely just treat the kotling stdlib like all other dependencies.

iamdanfox:
Glad to hear you're interested @yschimke, I'm always keen to hear any first-impressions when you get the chance to try it out!

As for published artifacts, we use it on our [tracing-java](https://mvnrepository.com/artifact/com.palantir.tracing/tracing/2.10.1) library, and the eventual [POM](https://dl.bintray.com/palantir/releases/com/palantir/tracing/tracing/2.10.1/tracing-2.10.1.pom) has the usual dependencies block. (Although we do use gradle's baked in `maven-publish` instead of `com.vanniktech.maven.publish`)

```xml
<dependencies>
   ...
   <dependency>
      <groupId>org.slf4j</groupId>
      <artifactId>slf4j-api</artifactId>
      <version>1.7.12</version>
      <scope>compile</scope>
    </dependency>
    <dependency>
      <groupId>com.fasterxml.jackson.core</groupId>
      <artifactId>jackson-annotations</artifactId>
      <version>2.9.8</version>
      <scope>runtime</scope>
    </dependency>
    ...
</dependencies>
```

I'm definitely a big fan of keeping things as clean and vanilla gradle as possible! Having the nicely structured versions.props file means automation can come round and open PRs to update our deps for us ([example](https://github.com/palantir/tracing-java/pull/123/files)).

Totally fine by me if you'd prefer for okhttp itself to keep rolling it's own dependency management story though. Happy to fix up merge conflicts, or close out if you'd prefer.

swankjesse:
After seeking further opinions, I’m going to decline this because it doesn’t solve problems we currently have.

PR #4886, opened 2019-04-06
Optimize code and eliminate Lint warning

TangHuaiZhe:
- Statement lambda can be replaced with expression lambda
- Delete unused imports


swankjesse:
Can you send the `Delete unused imports` PR separately? 

I don’t mind the lift return lint warning; that makes lint happy but it makes *me* unhappy.

TangHuaiZhe:
> Can you send the `Delete unused imports` PR separately?
> 
> I don’t mind the lift return lint warning; that makes lint happy but it makes _me_ unhappy.

Done
Refer to `Delete unused imports` #4888

PR #4887, opened 2019-04-06
Fix source compatibility for some Kotlin callsites.

swankjesse:
We had a problem where boolean vals needed to be reverted back to
boolean funs. I'd like to go back to vals later, but supporting existing
source patterns is more important.

We also had a problem where single abstract method types (SAM types)
could be supplied as lambas when calling into Java but not when calling
into Kotlin.

I found these by pointing KotlinSourceCompatibilityTest at the OkHttp

Get there like this:
```
  git co 7eb897ab2e223632b3316bf46a15d37307a3d3b6^
  git co 7eb897ab2e223632b3316bf46a15d37307a3d3b6 okhttp/build.gradle
  git co 7eb897ab2e223632b3316bf46a15d37307a3d3b6 build.gradle
```

### DIFF
--- a/okhttp/src/main/java/okhttp3/Dns.kt
+++ b/okhttp/src/main/java/okhttp3/Dns.kt
@@ -54,5 +54,13 @@ interface Dns {
         }
       }
     }
+
+    // This lambda conversion is for Kotlin callers expecting a Java SAM (single-abstract-method).
+    @JvmName("-deprecated_Dns")
+    inline operator fun invoke(
+      crossinline block: (String) -> List<InetAddress>
+    ): Dns = object : Dns {
+      override fun lookup(hostname: String): List<InetAddress> = block(hostname)
+    }
   }
 }
--- a/okhttp/src/main/java/okhttp3/OkHttpClient.kt
+++ b/okhttp/src/main/java/okhttp3/OkHttpClient.kt
@@ -755,6 +755,16 @@ open class OkHttpClient internal constructor(
       interceptors += interceptor
     }
 
+    // This lambda conversion is for Kotlin callers expecting a Java SAM (single-abstract-method).
+    @JvmName("-deprecated_addInterceptor")
+    inline fun addInterceptor(
+      crossinline interceptor: (chain: Interceptor.Chain) -> Response
+    ) = apply {
+      addInterceptor(object : Interceptor {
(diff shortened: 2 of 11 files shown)


okhttp/src/main/java/okhttp3/Dns.kt
```diff
@@ -54,5 +54,10 @@ interface Dns {
         }
       }
     }
+
+    // This lambda conversion is for Kotlin callers expecting a Java SAM (single-abstract-method).
+    @JvmName("-deprecated_Dns")
+    inline operator fun invoke(crossinline block: (String) -> List<InetAddress>): Dns =
+        Dns { block(it) }
```
swankjesse: An alternative we should consider – top-level function named with the type name:

```
inline fun Dns(crossinline block: (String) -> List<InetAddress>): Dns = object : Dns {
  override fun lookup(hostname: String) = block(hostname)
}
```

swankjesse: As also found here:
https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-core/common/src/Job.kt#L357

okhttp/src/main/java/okhttp3/OkHttpClient.kt
```diff
@@ -786,6 +806,14 @@ open class OkHttpClient internal constructor(
       this.eventListenerFactory = eventListenerFactory
     }
 
+    // This lambda conversion is for Kotlin callers expecting a Java SAM (single-abstract-method).
+    @JvmName("-deprecated_eventListenerFactory")
```
swankjesse: Also considered adding this:
```
    @Deprecated(
        message = "moved to object",
        replaceWith = ReplaceWith(
            expression = "" +
                "eventListenerFactory(object : EventListener.Factory {\n" +
                "  override fun create(call: Call) = block(call)\n" +
                "})",
            imports = [
              "okhttp3.Call", 
              "okhttp3.EventListener.Factory"
            ]),
    inline fun eventListenerFactory(
      crossinline block: (call: Call) -> EventListener
    )
```

swankjesse: It works, but when you apply the advice it looks a bit weird ’cause it actually creates the lambda and then calls it.

PR #4890, opened 2019-04-06
RFC: Platform Thoughts

yschimke:
I'm still assuming Platform is a JVM wide instance, but it seems in almost all cases OkHttpClient is the best way to reference the current platform.  Would make it easier to test different platforms in our tests.

I do wonder whether some control over the choice of Platform should be eventually made into public API, e.g. OkHttp users can't implement a platform but can force downgrade to Platform instead of ConscryptPlatform etc.  Or conversely can force Jdk9Platform to be used.

### DIFF
--- a/okhttp/src/main/java/okhttp3/OkHttpClient.kt
+++ b/okhttp/src/main/java/okhttp3/OkHttpClient.kt
@@ -121,6 +121,7 @@ import kotlin.DeprecationLevel.ERROR
 open class OkHttpClient internal constructor(
   builder: Builder
 ) : Cloneable, Call.Factory, WebSocket.Factory {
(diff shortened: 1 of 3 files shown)


swankjesse:
I'd like to consider another direction. Prior to Conscrypt the Platform was 1:1 with the host VM. With Conscrypt the Platform has two competing responsibilities: TLS and everything else.

We could split TLS out, and mix and match to make bundled Conscrypt work on Android.

yschimke:
There is a nice future API for managing common things in mobile apps or corporate apps related to custom certificates etc, that ideally may even exist outside of OkHttp, extending the direction of okhttp-tls.

PR #4892, opened 2019-04-06
Conscrypt Test Fix

yschimke:
Another attempt at fixing Conscrypt tests

PlatformRule:
- identifies tests that are sensitive to the platform, all such tests should use it.
- flags methods by identifying the intent of specific tests e.g. assumeNotConscrypt
- initializes the JVM including Security before any such tests run.  This is one way setup.

Adds a Platform.reset to allow setting and clearing a specific Platform, but this is generally not useful as you can rely on other test runs in CI.


okhttp/src/test/java/okhttp3/ConscryptTest.java
```diff
@@ -31,23 +32,22 @@
...
-  private OkHttpClient client = new OkHttpClient();
-
   @BeforeClass
-  public static void setup() {
-    assumeConscrypt();
-
-    OpenSSLProvider provider = new OpenSSLProvider();
-    Security.insertProviderAt(provider, 1);
+  public static void installProvider() {
+    if ("conscrypt".equals(System.getProperty("okhttp.platform"))) {
```
yschimke: This is the main fix.  We need to insert the provider before the JDK and OkHttp cache anything SSL related.

But it also means that assumptions in ConnectionSpecTest break.  And some behaviour differs in Intellij running one test vs running in gradle with daemon or locally.  Hence the bad commit.

.circleci/config.yml
```diff
@@ -238,10 +238,11 @@ workflows:
           filters:
             branches:
               only: master
-      - testconscrypt:
-          filters:
-            branches:
-              only: master
+      - testconscrypt
+#      - testconscrypt:
```
yschimke: TODO revert

okhttp/src/test/java/okhttp3/ConscryptTest.java
yschimke: Tests just passed.... but....


```
okhttp3.ConscryptTest > testMozilla SKIPPED

okhttp3.ConscryptTest > testGoogle SKIPPED
```

okhttp/src/test/java/okhttp3/ConscryptTest.java
```diff
@@ -31,23 +32,22 @@
...
+  private OkHttpClient client = new OkHttpClient();
+
   @After
   public void tearDown() {
     TestUtil.ensureAllConnectionsReleased(client);
   }
 
   private static void assumeConscrypt() {
-    Assume.assumeTrue("conscrypt".equals(System.getProperty("okhttp.platform")));
+    Assume.assumeTrue(Conscrypt.isConscrypt(Platform.get().platformTrustManager()));
```
yschimke: This is the honest check of whether we are running the test with Conscrypt enabled.

build.gradle
```diff
@@ -110,6 +110,7 @@ subprojects { project ->
     jvmArgs += "-Dokhttp.platform=$platform"
 
     testLogging {
+//      events "skipped", "failed"
```
swankjesse: ?

okhttp/src/test/java/okhttp3/ConscryptTest.java
swankjesse: Want a method like `Platform.reset()` ?

yschimke: I want internal methods to reset to null or a specific Platform instance :)

But generally if we have only a few places in the code that call Platform.get() e.g. an internal var of OkHttp client it makes this testing soooooo much easier.

build.gradle
yschimke: It's useful feedback when debugging in CI, so keeping here to remember the code, I think instead I need to work out how to control the logging levels of the gradlew.

okhttp/src/test/java/okhttp3/ConscryptTest.java
yschimke: Specifically Platform.reset() is still a JVM singleton, which makes strong assumptions about how the tests run in parallel.  It's not clear to me if gradle parallel is within same test runner instance or not.

yschimke:
https://circleci.com/workflow-run/138023bb-e3f6-4bd0-9f30-388a4be708d7

image

okhttp/src/main/java/okhttp3/internal/platform/Platform.kt
```diff
@@ -24,6 +24,7 @@ import okhttp3.internal.tls.BasicTrustRootIndex
 import okhttp3.internal.tls.CertificateChainCleaner
 import okhttp3.internal.tls.TrustRootIndex
 import okio.Buffer
+import org.jetbrains.annotations.TestOnly
```
yschimke: oops

okhttp/src/test/java/okhttp3/PlatformRule.kt
```diff
@@ -0,0 +1,113 @@
...
+import org.junit.rules.ExternalResource
+import java.security.Security
+
+/**
+ * Marks a test as Platform aware, before the test runs a consistent Platform will be
+ * established e.g. SecurityManager for Conscrypt installed.
+ *
+ * Also allows a test file to state general platform assumptions, or for individual test.
+ */
+open class PlatformRule @JvmOverloads constructor(
```
yschimke: We may want to extend this to allow testing the failure modes on specific platforms instead.  Instead of failing an assumption, we could verify that it fails with a specific error in Conscrypt.  Then if Conscrypt starts working, we would detect the regression.

okhttp/src/main/java/okhttp3/internal/platform/Jdk8WithJettyBootPlatform.kt
```diff
@@ -121,6 +121,16 @@ class Jdk8WithJettyBootPlatform(
   companion object {
     @JvmStatic
     fun buildIfSupported(): Platform? {
+      val jvmVersion = System.getProperty("java.specification.version", "unknown")
+      try {
+        // 1.8, 9, 10, 11, 12 etc
+        val version = jvmVersion.toInt()
+        if (version >= 9)
+          return null
+      } catch (re: NumberFormatException) {
```
swankjesse: re?

okhttp/src/main/java/okhttp3/internal/platform/Jdk8WithJettyBootPlatform.kt
```diff
@@ -121,6 +121,16 @@ class Jdk8WithJettyBootPlatform(
   companion object {
     @JvmStatic
     fun buildIfSupported(): Platform? {
+      val jvmVersion = System.getProperty("java.specification.version", "unknown")
+      try {
+        // 1.8, 9, 10, 11, 12 etc
+        val version = jvmVersion.toInt()
+        if (version >= 9)
```
swankjesse: parens or don’t wrap

okhttp/src/main/java/okhttp3/internal/platform/Platform.kt
```diff
@@ -188,27 +189,27 @@ open class Platform {
...
     const val INFO = 4
     const val WARN = 5
 
     private val logger = Logger.getLogger(OkHttpClient::class.java.name)
 
     @JvmStatic
-    fun get(): Platform = PLATFORM
+    fun get(): Platform = platform
+
+    @TestOnly
```
swankjesse: FWIW, I’d prefer the method name be like `resetForTests` so the test-only nature of it is visible at the callsite. I find `@TestOnly` and `@VisibleForTesting` to be weird because they exist to discourage future changes that might otherwise make sense.

okhttp/src/test/java/okhttp3/ConscryptTest.kt
```diff
@@ -0,0 +1,99 @@
...
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+import java.net.InetAddress
+import java.net.UnknownHostException
+
+class ConscryptTest {
+  @Suppress("RedundantVisibilityModifier")
+  @JvmField
+  @Rule public val platform = PlatformRule.conscrypt()
```
swankjesse: neat!

okhttp/src/test/java/okhttp3/ConscryptTest.kt
```diff
@@ -0,0 +1,99 @@
...
+import org.junit.Assert.assertTrue
+import org.junit.Assume
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+import java.net.InetAddress
+import java.net.UnknownHostException
+
+class ConscryptTest {
+  @Suppress("RedundantVisibilityModifier")
```
swankjesse: why ?

okhttp/src/test/java/okhttp3/ConscryptTest.kt
```diff
@@ -0,0 +1,99 @@
...
+  @JvmField
+  @Rule public val platform = PlatformRule.conscrypt()
+
+  private lateinit var client: OkHttpClient
+
+  @Before
+  fun createClient() {
+    assertTrue(Conscrypt.isConscrypt(Platform.get().platformTrustManager()))
```
swankjesse: `assertThat(...).isTrue()`

I don’t wanna mix & match two different assert patterns

okhttp/src/test/java/okhttp3/ConscryptTest.kt
```diff
@@ -0,0 +1,99 @@
...
+  @Test
+  fun testGoogle() {
+    assumeNetwork()
+
+    val request = Request.Builder().url("https://google.com/robots.txt").build()
+
+    val response = client.newCall(request).execute()
+
+    assertThat(response.protocol()).isEqualTo(Protocol.HTTP_2)
+//    assertThat(response.handshake()!!.tlsVersion()).isEqualTo(TlsVersion.TLS_1_3)
```
swankjesse: ?

okhttp/src/test/java/okhttp3/PlatformRule.kt
```diff
@@ -0,0 +1,113 @@
...
+import org.hamcrest.CoreMatchers.equalTo
+import org.hamcrest.CoreMatchers.not
+import org.junit.Assume
+import org.junit.Assume.assumeThat
+import org.junit.rules.ExternalResource
+import java.security.Security
+
+/**
+ * Marks a test as Platform aware, before the test runs a consistent Platform will be
+ * established e.g. SecurityManager for Conscrypt installed.
```
swankjesse: SecurityManager?!

swankjesse: I think you mean SecurityProvider. SecurityManager is an applet thing

okhttp/src/test/java/okhttp3/PlatformRule.kt
swankjesse: Cool!

okhttp/src/test/java/okhttp3/PlatformRule.kt
```diff
@@ -0,0 +1,113 @@
...
+    assumeThat(getPlatform(), equalTo("jdk9"))
+  }
+  fun assumeLegacy() {
+    assumeThat(getPlatform(), equalTo("platform"))
+  }
+  fun assumeNotConscrypt() {
+    assumeThat(getPlatform(), not("conscrypt"))
+  }
+  fun assumeNotJdk9() {
+    assumeThat(getPlatform(), not("jdk9"))
```
swankjesse: does AssertJ do assumes? Sigh

okhttp/src/test/java/okhttp3/PlatformRule.kt
```diff
@@ -0,0 +1,113 @@
...
+    }
+  }
+
+  fun isConscrypt() = getPlatform() == "conscrypt"
+  fun isJdk9() = getPlatform() == "jdk9"
+  fun isLegacy() = getPlatform() == "platform"
+  fun assumeConscrypt() {
+    assumeThat(getPlatform(), equalTo("conscrypt"))
+  }
+  fun assumeJdk9() {
```
swankjesse: nit: blank between funs?

okhttp/src/test/java/okhttp3/PlatformRule.kt
```diff
@@ -0,0 +1,113 @@
...
+  }
+  fun assumeNotJdk9() {
+    assumeThat(getPlatform(), not("jdk9"))
+  }
+  fun assumeNotLegacy() {
+    assumeThat(getPlatform(), not("platform"))
+  }
+
+  fun assumeJettyBootEnabled() {
+    Assume.assumeTrue("ALPN Boot not enabled", isAlpnBootEnabled())
```
swankjesse: nit: Omit `Assume.`  ?

okhttp/src/test/java/okhttp3/PlatformRule.kt
```diff
@@ -0,0 +1,113 @@
...
+        Security.insertProviderAt(provider, 1)
+      } else if (getPlatform() == "platform") {
+        if (!isAlpnBootEnabled()) {
+          System.err.println("Warning: ALPN Boot not enabled")
+        }
+      }
+    }
+
+    @JvmStatic
+    fun getPlatform(): String = System.getProperty("okhttp.platform", "platform")
```
swankjesse: Rename to getPlatformSystemProperty() ?

swankjesse: nit: null would be a better no-platform-defined constant than the string `"platform"`

okhttp/src/test/java/okhttp3/PlatformRule.kt
```diff
@@ -0,0 +1,113 @@
...
+
+    @JvmStatic
+    fun conscrypt() = PlatformRule("conscrypt")
+
+    @JvmStatic
+    fun jdk9() = PlatformRule("jdk9")
+
+    @JvmStatic
+    fun legacy() = PlatformRule("platform")
```
swankjesse: what does legacy mean? Does it mean Android or JVM? Want a more precise word?

build.gradle
swankjesse: if (false) ? That way find-usages etc. work

swankjesse: Never comment out code, but `if (false)` is completely fine ’cause it can be managed through refactoring and other static tools

okhttp/src/test/java/okhttp3/ConscryptTest.kt
yschimke: Unused code in a new Kotlin test, that needs to be public for junit.

```
class ConscryptTest {
  @Suppress("RedundantVisibilityModifier")
  @JvmField
  @Rule public val platform = PlatformRule.conscrypt()
```

okhttp/src/test/java/okhttp3/ConscryptTest.kt
yschimke: It appears flaky.  And technically is out of our control.  I suspect some weirdness between conscrypt and google.  e.g. maybe they have a 25/75 backtest for Conscrypt users to flush out problems?

The mozilla one isn't flaky.

cc @flooey

okhttp/src/test/java/okhttp3/PlatformRule.kt
yschimke: Changed to jdk8.  As long as it's test only we can restrict to jdk8 == jdk8 with alpn-boot.  But we might want to test without alpn-boot also.

yschimke: https://github.com/square/okhttp/issues/4904

okhttp/src/test/java/okhttp3/PlatformRule.kt
yschimke: See what you think about jdk8?  Maybe we should have "default"

okhttp/src/test/java/okhttp3/ConscryptTest.kt
JakeWharton:
```kotlin
@get:Rule
val platform = PlatformRule.conscrypt()
```
should work

PR #4895, opened 2019-04-07
Convert okhttp-tls classes to Kotlin

swankjesse:


okhttp/src/main/java/okhttp3/WebSocketListener.kt
```diff
@@ -22,29 +22,35 @@ abstract class WebSocketListener {
    * Invoked when a web socket has been accepted by the remote peer and may begin transmitting
    * messages.
    */
-  open fun onOpen(webSocket: WebSocket, response: Response) {}
+  open fun onOpen(webSocket: WebSocket, response: Response) {
```
swankjesse: we avoided `{}` in Java. Is it idiomatic in Kotlin?

okhttp/src/main/java/okhttp3/internal/platform/Platform.kt
```diff
@@ -80,27 +77,17 @@ open class Platform {
...
-          TrustManagerFactory.getDefaultAlgorithm())
-      trustManagerFactory.init(null as KeyStore?)
-      val trustManagers = trustManagerFactory.trustManagers
-      if (trustManagers.size != 1 || trustManagers[0] !is X509TrustManager) {
-        throw IllegalStateException(
-            "Unexpected default trust managers:" + Arrays.toString(trustManagers))
-      }
-      return trustManagers[0] as X509TrustManager
-    } catch (e: GeneralSecurityException) {
-      throw AssertionError("No System TLS", e) // The system has no TLS. Just give up.
```
swankjesse: removing places where we catch only to rethrow AssertionErrors

okhttp/src/main/java/okhttp3/internal/platform/Platform.kt
```diff
@@ -150,17 +139,18 @@ open class Platform {
    * should be used specifically for [java.io.Closeable] objects and in conjunction with
    * [logCloseableLeak].
    */
-  open fun getStackTraceForCloseable(closer: String): Any? =
-      when {
```
swankjesse: I’m not a fan of big expression functions

okhttp/src/main/java/okhttp3/internal/platform/Platform.kt
```diff
@@ -80,27 +77,17 @@ open class Platform {
   /** Prefix used on custom headers.  */
   fun getPrefix() = "OkHttp"
 
-  open fun getSSLContext(): SSLContext = try {
-    SSLContext.getInstance("TLS")
-  } catch (e: NoSuchAlgorithmException) {
-    throw IllegalStateException("No TLS provider", e)
-  }
+  open fun newSSLContext(): SSLContext = SSLContext.getInstance("TLS")
```
swankjesse: this seemed misnamed, especially since we mutate it immediately after access

okhttp/src/main/java/okhttp3/WebSocketListener.kt
yschimke: But don't make this your decision, pick a code formatter and apply e.g. ktlint :)

okhttp/src/main/java/okhttp3/internal/platform/Platform.kt
yschimke: I'd argue against this one, when is effectively the block and the new form just replaces ` = ` for `{ return`

okhttp/src/main/java/okhttp3/internal/platform/Platform.kt
yschimke: Good call

okhttp/src/main/java/okhttp3/internal/platform/Platform.kt
```diff
@@ -80,27 +77,17 @@ open class Platform {
...
-            "Unexpected default trust managers:" + Arrays.toString(trustManagers))
-      }
-      return trustManagers[0] as X509TrustManager
-    } catch (e: GeneralSecurityException) {
-      throw AssertionError("No System TLS", e) // The system has no TLS. Just give up.
+    val factory = TrustManagerFactory.getInstance(
+        TrustManagerFactory.getDefaultAlgorithm())
+    factory.init(null as KeyStore?)
+    val trustManagers = factory.trustManagers!!
+    check(trustManagers.size == 1 && trustManagers[0] is X509TrustManager) {
```
yschimke: TIOLI: If writing in kotlin from scratch I might go for

```
      val tmf = TrustManagerFactory.getInstance(TrustManagerFactory.getDefaultAlgorithm()).apply { 
        init(null as KeyStore?)
      }
      
      return tmf.trustManagers?.firstOrNull() as? X509TrustManager ?: throw IllegalStateException("XXX")
```

Mainly to focus on the good path, not the bad

okhttp-tls/src/main/java/okhttp3/tls/HandshakeCertificates.kt
```diff
@@ -0,0 +1,139 @@
...
+     * The chain should include all intermediate certificates but does not need the root certificate
+     * that we expect to be known by the remote peer. The peer already has that certificate so
+     * transmitting it is unnecessary.
+     */
+    fun heldCertificate(
+      heldCertificate: HeldCertificate,
+      vararg intermediates: X509Certificate
+    ) = apply {
+      this.heldCertificate = heldCertificate
+      this.intermediates = arrayOf(*intermediates) // Defensive copy.
```
JakeWharton: `intermediates.copyOf()`

okhttp-tls/src/main/java/okhttp3/tls/HandshakeCertificates.kt
```diff
@@ -0,0 +1,139 @@
...
+     * Most TLS clients that connect to hosts on the public Internet should call this method.
+     * Otherwise it is necessary to manually prepare a comprehensive set of trusted roots.
+     *
+     * If the host platform is compromised or misconfigured this may contain untrustworthy root
+     * certificates. Applications that connect to a known set of servers may be able to mitigate
+     * this problem with [certificate pinning][CertificatePinner].
+     */
+    fun addPlatformTrustedCertificates() = apply {
+      val platformTrustManager = Util.platformTrustManager()
+      Collections.addAll(trustedCertificates, *platformTrustManager.acceptedIssuers)
```
JakeWharton: Did `+=` not work here? Otherwise there's an `addAll` that accepts arrays.

PR #4899, opened 2019-04-07
Use more idiomatic Kotlin in EventSources and elsewhere

swankjesse:


okhttp/src/main/java/okhttp3/CertificatePinner.kt
```diff
@@ -264,21 +242,6 @@ data class CertificatePinner internal constructor(
...
     }
 
-    override fun equals(other: Any?): Boolean {
-      return other is Pin
-          && pattern == other.pattern
-          && hashAlgorithm == other.hashAlgorithm
-          && hash == other.hash
-    }
-
-    override fun hashCode(): Int {
```
JakeWharton: Did you mean to make this a data class? Or is it just not compared or used as a key?

swankjesse: I meant to make it a data class.

swankjesse: Good catch!

okhttp/src/main/java/okhttp3/Handshake.kt
```diff
@@ -72,33 +72,26 @@ data class Handshake private constructor(
...
+      }
+    }
 
   companion object {
     @Throws(IOException::class)
     @JvmStatic
     fun get(session: SSLSession): Handshake {
-      val cipherSuiteString = session.cipherSuite ?: throw IllegalStateException("cipherSuite == null")
+      val cipherSuiteString =
+          session.cipherSuite ?: throw IllegalStateException("cipherSuite == null")
```
Egorand: nit: `checkNotNull(session.cipherSuite) { "cipherSuite == null" }`

okhttp/src/main/java/okhttp3/Handshake.kt
```diff
@@ -72,33 +72,26 @@ data class Handshake private constructor(
...
-          if (it is X509Certificate) {
-            it.subjectDN.toString()
-          } else {
-            it.type
-          }
-        }
-  }
+  private val Certificate.name: String
+    get() {
+      return when (this) {
```
Egorand: nit: 
```kotlin
get() = when (this) {
  is X509Certificate -> subjectDN.toString()
  else -> type
}
```

okhttp/src/main/java/okhttp3/Handshake.kt
swankjesse: Fixed.

okhttp/src/main/java/okhttp3/Handshake.kt
swankjesse: Fixed.

okhttp/src/main/java/okhttp3/Handshake.kt
yschimke: That checkNotNull is nice. TIL

PR #4900, opened 2019-04-07
Add MockWebServer to KotlinSourceCompatibilityTest

swankjesse:
Unlike most of OkHttp, the API for MockWebServer is pretty rough. We don't
have Nullable annotations. Everything is mutable. There are even weird bad
things where public methods return non-public types.


swankjesse:
I want to get very close to 100% source and binary compatibility, leaving no reason to stay on 3.x.

For a better API I think we create a new package name?

Egorand:
Yep, and deprecate okhttp3.* code.

PR #4901, opened 2019-04-08
Don't crash on retry when there are no more routes (3.x branch)

swankjesse:
https://github.com/square/okhttp/issues/4875

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/connection/ExchangeFinder.java
+++ b/okhttp/src/main/java/okhttp3/internal/connection/ExchangeFinder.java
@@ -258,11 +258,16 @@ final class ExchangeFinder {
     }
   }
 
-  boolean canRetry() {
+  /** Returns true if there is a failure that retrying might fix. */
+  boolean hasStreamFailure() {
     synchronized (connectionPool) {
-      // Don't try if the failure wasn't our fault!
-      if (!hasStreamFailure) return false;
+      return hasStreamFailure;
+    }
+  }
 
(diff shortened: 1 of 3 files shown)


okhttp/src/main/java/okhttp3/internal/connection/ExchangeFinder.java
```diff
@@ -258,11 +258,16 @@ void trackFailure() {
     }
   }
 
-  boolean canRetry() {
```
swankjesse: this method got split into 2 halves: hasStreamFailure() and hasRouteToTry().

okhttp/src/main/java/okhttp3/internal/connection/Transmitter.java
```diff
@@ -303,7 +305,7 @@ public void exchangeDoneDueToException() {
   }
 
   public boolean canRetry() {
-    return exchangeFinder.canRetry();
```
swankjesse: this now calls the two halves in sequence

okhttp/src/main/java/okhttp3/internal/connection/Transmitter.java
```diff
@@ -123,7 +123,9 @@ public void callStart() {
    */
   public void prepareToConnect(Request request) {
     if (this.request != null) {
-      if (sameConnection(this.request.url(), request.url())) return; // Already ready.
```
swankjesse: and this calls the 2nd half only!

PR #4902, opened 2019-04-08
Parallelize tests based on available processors

ZacSweers:
This brings down `:okhttp:test` (the biggest bottle neck) time on my local machine from ~2m11s to ~39s.

Build scans:
* Without parallelization: https://scans.gradle.com/s/nmjkdjflng4o4
* With parallelization: https://scans.gradle.com/s/3oebsmkfwetiy

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ jdk:
   - openjdk11
 
 script:
- - ./gradlew check
+ - ./gradlew check --parallel
 
 after_success:
   - .buildscript/deploy_snapshot.sh
--- a/build.gradle
+++ b/build.gradle
@@ -121,6 +121,7 @@ subprojects { project ->
     jvmArgs += "-Dlistener=okhttp3.testing.InstallUncaughtExceptionHandlerListener"
     jvmArgs += "-Dokhttp.platform=$platform"
 
+    maxParallelForks Runtime.runtime.availableProcessors().intdiv(2) ?: 1
     testLogging {
       exceptionFormat = 'full'
     }
--- a/build.gradle
+++ b/build.gradle
@@ -121,7 +121,7 @@ subprojects { project ->
     jvmArgs += "-Dlistener=okhttp3.testing.InstallUncaughtExceptionHandlerListener"
     jvmArgs += "-Dokhttp.platform=$platform"
 
-    maxParallelForks Runtime.runtime.availableProcessors().intdiv(2) ?: 1
+    maxParallelForks Runtime.runtime.availableProcessors() * 2
     testLogging {
       exceptionFormat = 'full'
     }


yschimke:
I noticed a lot more flakiness and timeouts when running with high concurrency. Although to be honest there are so many knobs in Gradle + ci concurrency I may have been doing it wrong. Anyway if you can get it consistently passing locally and in CI with build cache disabled (for PR testing) then this would be awesome. 

yschimke:
BTW I really like this and fixing the tests should be an aim on its own. So if it’s delayed we should still track the work e.g. a high concurrency workflow in circle ci that is collecting flaky concurrent tests. 

ZacSweers:
I definitely had everything passing locally (hence being able to get the scans). I didn't explicitly disable the build cache though so I'll try that to be doubly sure. Will see if I can induce it from the CI reports here, maybe it's something unique to Circle's env. Will keep tinkering!

yschimke:
It might be useful to get a scan from CI with build cache off etc.

Also I notice parallel is off - https://scans.gradle.com/s/3oebsmkfwetiy/switches, maybe try with that on also.  Although your 12 core machine will probably soak up a lot of time sensitive delays anyway.

build.gradle
```diff
@@ -109,6 +109,7 @@ subprojects { project ->
     jvmArgs += "-Dlistener=okhttp3.testing.InstallUncaughtExceptionHandlerListener"
     jvmArgs += "-Dokhttp.platform=$platform"
 
+    maxParallelForks Runtime.runtime.availableProcessors().intdiv(2) ?: 1
```
yschimke: Would love to see this land.  CI env in circle is 2 CPU only, so I think this is no-op.

image

I'd suggest we try to flush out the flaky tests by running with Runtime.runtime.availableProcessors() * 2, and seeing what breaks.  A lot of tests are heavily CPU bound anyway.  If you enable this before any capacity boost, then it might block running on a bigger paid container.

yschimke:
Rebuild gave you a clean build.

ZacSweers:
Sorry for the delay, I've been swamped the past few days. I'm fine with either case, and can do the no-build-cache test still to check

ZacSweers:
Alright finally circling back, this is the only test I'm able to somewhat consistently repro after rebasing master. 

```
okhttp3.InterceptorTest > connectTimeout FAILED
    java.net.ConnectException: Operation timed out (Connection timed out)
        at java.net.PlainSocketImpl.socketConnect(Native Method)
        at java.net.AbstractPlainSocketImpl.doConnect(AbstractPlainSocketImpl.java:350)
        at java.net.AbstractPlainSocketImpl.connectToAddress(AbstractPlainSocketImpl.java:204)
        at java.net.AbstractPlainSocketImpl.connect(AbstractPlainSocketImpl.java:188)
        at java.net.SocksSocketImpl.connect(SocksSocketImpl.java:392)
        at java.net.Socket.connect(Socket.java:589)
        at java.net.Socket.connect(Socket.java:538)
        at okhttp3.InterceptorTest.connectTimeout(InterceptorTest.java:684)
```

build.gradle
ZacSweers: 3236112

yschimke:
It seems certainly the case that in a constrained environment like CI, this noticeably triggers existing flaky tests

https://circleci.com/gh/square/okhttp/1503?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

.travis.yml
```diff
@@ -6,7 +6,7 @@ jdk:
   - openjdk11
 
 script:
- - ./gradlew check
+ - ./gradlew check --parallel
```
yschimke: We can probably remove this as CircleCI should be the main builds for PRs

yschimke: And maybe activate for .circleci/config.yml

yschimke:
Going to land and observe whether there is an increase in flakiness in CI tests.  That is an issue with those tests, not this PR.

yschimke:
Already seen some possibly related failures, but in known flaky tests.  e.g. ConscryptTest which are disabled anyway in

https://github.com/square/okhttp/pull/5212/files

yschimke:
Nice win in CI

image

PR #4905, opened 2019-04-08
Don't crash on retry when there are no more routes (4.x branch)

swankjesse:
https://github.com/square/okhttp/issues/4875

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/connection/Transmitter.java
+++ b/okhttp/src/main/java/okhttp3/internal/connection/Transmitter.java
@@ -123,7 +123,9 @@ public final class Transmitter {
    */
   public void prepareToConnect(Request request) {
     if (this.request != null) {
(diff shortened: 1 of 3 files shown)


okhttp/src/main/java/okhttp3/internal/connection/Transmitter.java
```diff
@@ -303,7 +305,7 @@ public void exchangeDoneDueToException() {
   }
 
   public boolean canRetry() {
-    return exchangeFinder.canRetry();
+    return exchangeFinder.hasStreamFailure() && exchangeFinder.hasRouteToTry();
```
JakeWharton: Is it problematic that this locks twice allowing code to run between the two expressions?

swankjesse: Nope. Whew!

PR #4906, opened 2019-04-08
Use gradle --continue to flush out all test results

yschimke:
For the PR builds, getting all failures is more useful


yschimke:
@JakeWharton I'm going to park this stream for a bit, it's a bit like playing Windows Solitaire or some LevelQuest online game :). I need to focus on other personal projects that need more attention right now.

PR #4908, opened 2019-04-08
Convert Version to Kotlin

swankjesse:


okhttp/src/main/java-templates/okhttp3/internal/Version.kt
```diff
@@ -13,13 +13,9 @@
...
-public final class Version {
-  public static String userAgent() {
-    return "okhttp/${projectVersion}";
-  }
-
-  private Version() {
-  }
+object Version {
+  @JvmStatic
+  val userAgent = "okhttp/${projectVersion}"
```
JakeWharton: Make top level, remove `object`, and add

```kotlin
@file:JvmName("Version")
```

To avoid the needless singleton instance.

swankjesse: Oh Kotlin, some day I will understand you. Fixed.

PR #4909, opened 2019-04-08
Convert JavaNetAuthenticator to Kotlin

swankjesse:


okhttp-urlconnection/src/main/java/okhttp3/JavaNetAuthenticator.kt
```diff
@@ -0,0 +1,92 @@
...
+import java.net.Authenticator
+import java.net.InetAddress
+import java.net.InetSocketAddress
+import java.net.Proxy
+
+/**
+ * Adapts [Authenticator] to [okhttp3.Authenticator]. Configure OkHttp to use [Authenticator] with
+ * [OkHttpClient.Builder.authenticator] or [OkHttpClient.Builder.proxyAuthenticator].
+ */
+@EverythingIsNonNull
```
JakeWharton: Delete?

okhttp-urlconnection/src/main/java/okhttp3/JavaNetAuthenticator.kt
```diff
@@ -0,0 +1,92 @@
...
+    val url = request.url()
+    val proxyAuthorization = response.code() == 407
+    val proxy = route?.proxy() ?: Proxy.NO_PROXY
+
+    for (challenge in challenges) {
+      if (!"Basic".equals(challenge.scheme(), ignoreCase = true)) {
+        continue
+      }
+
+      val auth = when {
```
JakeWharton: Switch to
```kotlin
val auth = if (proxyAuthorization) {
  ..
} else {
  ..
}
```

okhttp-urlconnection/src/main/java/okhttp3/JavaNetAuthenticator.kt
swankjesse: Fixed!

okhttp-urlconnection/src/main/java/okhttp3/JavaNetAuthenticator.kt
swankjesse: Fixed.

PR #4910, opened 2019-04-08
Convert RealEventSource.java to .kt

swankjesse:
I broke my rule and did some extra refactoring. In particular I adopted Okio’s Options.


okhttp-sse/src/main/java/okhttp3/internal/sse/ServerSentEventReader.kt
```diff
@@ -0,0 +1,152 @@
...
+  @Throws(IOException::class)
+  fun processNextEvent(): Boolean {
+    var id = lastId
+    var type: String? = null
+    val data = Buffer()
+
+    while (true) {
+      val option = source.select(options)
+      when (option) {
+        0, 1, 2 -> {
```
swankjesse: gross . . . ideas?

okhttp-sse/src/main/java/okhttp3/internal/sse/RealEventSource.kt
```diff
@@ -0,0 +1,106 @@
...
+        }
+      } catch (e: Exception) {
+        listener.onFailure(this, e, response)
+        return
+      }
+      listener.onClosed(this)
+    }
+  }
+
+  private fun MediaType?.isEventStream(): Boolean =
```
JakeWharton: Hang this on `ResponseBody` instead? I'm not sure about the nullable receiver. I know it's used once or twice in the stdlib for things like `toString()`, `hashCode`, and `isNullOrEmpty()`, but it seems like the call site could just as easily be `contentType?.isEventStream() != true`.

okhttp-sse/src/main/java/okhttp3/internal/sse/ServerSentEventReader.kt
```diff
@@ -0,0 +1,152 @@
...
+      return try {
+        retryString.toLong()
+      } catch (_: NumberFormatException) {
+        -1L
+      }
+    }
+
+    private fun String.toNullIfEmpty(): String? {
+      if (isEmpty()) return null
+      return this
```
JakeWharton: You could use `takeIf { it.isNotEmpty() }` for the implementation here (or inlined at the call sites).

okhttp-sse/src/main/java/okhttp3/internal/sse/ServerSentEventReader.kt
```diff
@@ -0,0 +1,152 @@
...
+   * @return false when EOF is reached
+   */
+  @Throws(IOException::class)
+  fun processNextEvent(): Boolean {
+    var id = lastId
+    var type: String? = null
+    val data = Buffer()
+
+    while (true) {
+      val option = source.select(options)
```
JakeWharton: You can move this inside the `when`:
```kotlin
when (val option = source.select(options)) {
```

okhttp-sse/src/main/java/okhttp3/internal/sse/ServerSentEventReader.kt
JakeWharton: `0..2`?

I thought there was a reason that I avoided using select... I don't recall it.

swankjesse:
The big drawback of select is you need 3 choices for each!

okhttp-sse/src/main/java/okhttp3/internal/sse/ServerSentEventReader.kt
```diff
@@ -0,0 +1,152 @@
...
+    val options = Options.of(
+        // 0, 1, 2
+        "\r\n".encodeUtf8(),
+        "\r".encodeUtf8(),
+        "\n".encodeUtf8(),
+
+        // 3, 4, 5
+        "data: ".encodeUtf8(),
+        "data:".encodeUtf8(),
+        "data".encodeUtf8(),
```
JakeWharton: This is a behavior change. Lines starting with `database: foo` will be erroneously parsed as being data. The only time you can match on the non-colon version is when the string occupies the entire line. So you would need to make two entires for this (`"data\r"` and `"data\n"`) and the other keys.

I believe this is why I elected to not use select.

okhttp-sse/src/main/java/okhttp3/internal/sse/ServerSentEventReader.kt
swankjesse: Fixed.

okhttp-sse/src/main/java/okhttp3/internal/sse/RealEventSource.kt
swankjesse: Moved it to ResponseBody.

okhttp-sse/src/main/java/okhttp3/internal/sse/ServerSentEventReader.kt
swankjesse: Done!

okhttp-sse/src/main/java/okhttp3/internal/sse/ServerSentEventReader.kt
swankjesse: Done!

Issue #4761, opened 2019-03-22
Regression in Exchange in 3.14


tmtrademarked:
Hey, folks - I think I'm running into something that appears to be a regression in 3.14 around the new exchange mechanism. I can reproduce this 100% of the time in my app by creating a specific scenario.

Here's the basic setup. We have an interceptor that does a lot of things for us, including setting headers and handling retries for auth errors. Here's a simplified version of the code, eliding a lot of details around how the retries are triggered:

```
@Override
public Response intercept(Chain chain) throws IOException {
  Request originalRequest = chain.request();
  Request.Builder builder = originalRequest.newBuilder();
  // Add headers/etc here.
  Request request = builder.build();
  Response response = chain.proceed(request);

  if (isError(response)) {
    // Grab a new auth token, etc.
    response = chain.proceed(retry);
    return response;
  }
}
```

In 3.13.1, this seems to work just fine. In 3.14.0, this crashes with the following error:

```
03-22 11:52:21.889 E/AndroidRuntime(10135): java.lang.IllegalStateException: exchange != null
03-22 11:52:21.889 E/AndroidRuntime(10135): 	at okhttp3.internal.connection.Transmitter.newExchange(Transmitter.java:159)
03-22 11:52:21.889 E/AndroidRuntime(10135): 	at okhttp3.internal.connection.ConnectInterceptor.intercept(ConnectInterceptor.java:41)
...
```

It doesn't seem to matter if you mutate the request or not. Even just using (as above) a newly built version of the request produces the same behavior.

Any ideas what's happening here? I'm going to revert our library back to 3.13 for now to work around this, but I'm curious what could be causing this.

swankjesse:
Good find. My guess is that we're crashing because the first response body is still active. Can you see if closing it fixes things?

tmtrademarked:
Ah, good callout! It looks like closing the previous response before retrying *does* fix the crash on 3.14. I was able to change my code to the follow and re-run the same scenarios with no crashes:

```
@Override
public Response intercept(Chain chain) throws IOException {
  Request originalRequest = chain.request();
  Request.Builder builder = originalRequest.newBuilder();
  // Add headers/etc here.
  Request request = builder.build();
  Response response = chain.proceed(request);

  if (isError(response)) {
    // Grab a new auth token, etc.
    response.close();
    Request retry = request.newBuilder().build();
    response = chain.proceed(retry);
    return response;
  }
}
```

[...]

PR #4911, opened 2019-04-08
Provide actionable advice when the exchange is non-null (3.x branch)

swankjesse:
https://github.com/square/okhttp/issues/4761


swankjesse:
When this lands I’ll cut 3.14.1.

swankjesse:
Cherry-picked to 4.x here:
https://github.com/square/okhttp/pull/4922/files

PR #4914, opened 2019-04-09
 Convert AddressTest to Kotlin

TangHuaiZhe:
Convert AddressTest to Kotlin.
All four tests passed.
Code style follow [square/java-code-styles](https://github.com/square/java-code-styles)


yschimke:
Moving tests like these is premature. These currently function as Java source compatibility regression tests. 

TangHuaiZhe:
> Moving tests like these is premature. These currently function as Java source compatibility regression tests.
> 
> Sent with [GitHawk](http://githawk.com)

I find `Address.java`  was already  converted to Kotlin , so I changed the AddressTest.
Should I close the pull request?

yschimke:
We need specific Public API coverage for both Kotlin and Java before we refactor these.  Also there is less priority for moving existing tests and it introduces some risk when trying to ensure we don't have regressions between 3.14 and 4.0.

swankjesse:
Yes. No plans to migrate tests.

PR #4916, opened 2019-04-09
Use @JvmOverloads to generate overloads where possible

amirlivneh:


okhttp/src/main/java/okhttp3/ConnectionPool.kt
```diff
@@ -29,17 +29,15 @@ import java.util.concurrent.TimeUnit
  * Currently this pool holds up to 5 idle connections which will be evicted after 5 minutes of
  * inactivity.
  */
-class ConnectionPool(
-  maxIdleConnections: Int,
-  keepAliveDuration: Long,
-  timeUnit: TimeUnit
+class ConnectionPool @JvmOverloads constructor (
```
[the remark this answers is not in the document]
amirlivneh: Fixed

okhttp/src/main/java/okhttp3/OkHttpClient.kt
```diff
@@ -118,8 +118,8 @@ import kotlin.DeprecationLevel.ERROR
  * OkHttp also uses daemon threads for HTTP/2 connections. These will exit automatically if they
  * remain idle.
  */
-open class OkHttpClient internal constructor(
-  builder: Builder
+open class OkHttpClient internal @JvmOverloads constructor(
```
yschimke: Doesn't this make `new OkHttpClient()` an internal call?  That is a really common pattern for java clients.

amirlivneh: You're right. Interestingly, it looks like this doesn't change the public API for Java consumers, which can still use that constructor but does change the public API for Kotlin consumers. Perhaps the name mangling used to prevent access to internal members isn't applied to constructors.

If this is true, it means that all constructors that were originally private or package-private and become `internal` as part of the migration, became effectively public to Java consumers.

PR #4920, opened 2019-04-09
Improve PlatformRule autodetection

yschimke:
- Adds jdk8 without alpn
- Auto detect based on JVM & features


yschimke:
Running here https://circleci.com/workflow-run/5c94c6e5-404e-4664-b6bb-a62444f390fd

Will reduce the checks on PRs after testing

okhttp-testing-support/src/main/java/okhttp3/PlatformRule.kt
```diff
@@ -55,55 +59,107 @@ open class PlatformRule @JvmOverloads constructor(
...
+      var property: String? = System.getProperty(PROPERTY_NAME)
+
+      if (property == null) {
+        property = when (Platform.get()) {
+          is ConscryptPlatform -> CONSCRYPT_PROPERTY
+          is Jdk8WithJettyBootPlatform -> CONSCRYPT_PROPERTY
+          is Jdk9Platform -> JDK9_PROPERTY
+          else -> JDK8_PROPERTY
+        }
+      }
```
yschimke: TODO: I should have an else check that warns in the reverse case, e.g. checks a specified system property matches Platform.get().  For a later PR.

okhttp-testing-support/src/main/java/okhttp3/PlatformRule.kt
```diff
@@ -55,55 +59,107 @@ open class PlatformRule @JvmOverloads constructor(
...
+      } else if (getPlatformSystemProperty() == JDK8_PROPERTY) {
+        if (isAlpnBootEnabled()) {
+          System.err.println("Warning: ALPN Boot enabled unintentionally")
+        }
       }
     }
 
     @JvmStatic
-    fun getPlatformSystemProperty(): String = System.getProperty(PROPERTY_NAME, DEFAULT_PROPERTY)
+    fun getPlatformSystemProperty(): String {
```
yschimke: TODO: How to test AndroidPlatform?

okhttp-testing-support/src/main/java/okhttp3/PlatformRule.kt
```diff
@@ -55,55 +59,107 @@ open class PlatformRule @JvmOverloads constructor(
...
         val provider = Conscrypt.newProviderBuilder().provideTrustManager(true).build()
         Security.insertProviderAt(provider, 1)
-      } else if (getPlatformSystemProperty() == DEFAULT_PROPERTY) {
+      } else if (getPlatformSystemProperty() == JDK8_ALPN_PROPERTY) {
         if (!isAlpnBootEnabled()) {
           System.err.println("Warning: ALPN Boot not enabled")
         }
+      } else if (getPlatformSystemProperty() == JDK8_PROPERTY) {
+        if (isAlpnBootEnabled()) {
+          System.err.println("Warning: ALPN Boot enabled unintentionally")
```
swankjesse: nice

okhttp/src/test/java/okhttp3/PlatformRuleTest.kt
```diff
@@ -0,0 +1,20 @@
+package okhttp3
```
swankjesse: supernit: copyright

okhttp-testing-support/src/main/java/okhttp3/PlatformRule.kt
```diff
@@ -55,55 +59,107 @@ open class PlatformRule @JvmOverloads constructor(
...
+    const val PROPERTY_NAME = "okhttp.platform"
+    const val CONSCRYPT_PROPERTY = "conscrypt"
+    const val JDK9_PROPERTY = "jdk9"
+    const val JDK8_ALPN_PROPERTY = "jdk8alpn"
+    const val JDK8_PROPERTY = "jdk8"
 
     init {
       if (getPlatformSystemProperty() == CONSCRYPT_PROPERTY && Security.getProviders()[0].name != "Conscrypt") {
+        if (!Conscrypt.isAvailable()) {
+          System.err.println("Warning: Conscrypt not available")
```
JakeWharton: Does this need to throw? Or we expect it to just fail somewhere else down the line maybe.

yschimke: I think you are right.

But to start with I didn't really want to change much behaviour, just make things more predictable.  I don't want it throwing new exceptions that we hadn't see before.  I think System.err will be a good pointer in an IDE console that something isn't right.

Let's land as is and see how the experience is in CI and IDEs, at least now it's behind this API and the tests are quite intentional, so you can search for each assumption method.

PR #4924, opened 2019-04-10
Ignore dependency:resolve failure on 3.14 branch

yschimke:
resolve seems to break on internal deps

https://travis-ci.org/square/okhttp/jobs/518372391


yschimke:
@swankjesse Damn, think I broke the first build post tag :(

If it works on 3.14, I'll make same change on master for travis.

PR #4927, opened 2019-04-11
Fix kotlin source compatibility with logger

yschimke:
If I understand correctly the Kotlin SAM rules mean this case will no longer work, so just documenting.


okhttp/src/test/java/okhttp3/KotlinSourceCompatibilityTest.kt
```diff
@@ -891,6 +891,7 @@ class KotlinSourceCompatibilityTest {
     builder = builder.eventListenerFactory(object : EventListener.Factory {
       override fun create(call: Call): EventListener = TODO()
     })
+    builder = builder.eventListenerFactory(LoggingEventListener.Factory { s -> TODO() })
```
yschimke: This is a source test of LoggingEventListener.Factory, but included here to show working end to end.

yschimke:
@swankjesse It broke downstream code, I didn't sense a disturbance in the source...

PR #4935, opened 2019-04-13
Move http classes to Kotlin

shaishavgandhi05:


okhttp/src/main/java/okhttp3/internal/http/StatusLine.kt
```diff
@@ -0,0 +1,107 @@
...
+      } else {
+        throw ProtocolException("Unexpected status line: $statusLine")
+      }
+
+      // Parse response code like "200". Always 3 digits.
+      if (statusLine.length < codeStart + 3) {
+        throw ProtocolException("Unexpected status line: $statusLine")
+      }
+      val code: Int
+      try {
```
JakeWharton:
```kotlin
val code = try { ... }
```

shaishavgandhi05: 6c606b204efd34392d03fbe5b2e482539d612b4a

swankjesse:
Thanks!

PR #4936, opened 2019-04-13
Fix #4915 nullPointerException caused by realConnection is null

### LINKED ISSUE #4915: NullPointerException in Http1ExchangeCodec.readResponseHeaders() in OkHttp 3.14.0
Since upgrading OkHttp from 3.13.1 to 3.14.0 in our Android app, some of our users experience crashes because of a NullPointerException in [`Http1ExchangeCodec` line 233](https://github.com/square/okhttp/blob/df95c77adfbf4f5ddd30bfc6a2bb684295d7d313/okhttp/src/main/java/okhttp3/internal/http1/Http1ExchangeCodec.java#L233).

Relevant stacktrace:
```
Fatal Exception: java.lang.NullPointerException: Attempt to invoke virtual method 'okhttp3.Route okhttp3.internal.connection.RealConnection.route()' on a null object reference
       at okhttp3.internal.http1.Http1ExchangeCodec.readResponseHeaders(Http1ExchangeCodec.java:233)
       at okhttp3.internal.connection.RealConnection.createTunnel(RealConnection.java:400)
       at okhttp3.internal.connection.RealConnection.connectTunnel(RealConnection.java:236)
       at okhttp3.internal.connection.RealConnection.connect(RealConnection.java:177)
       at okhttp3.internal.connection.ExchangeFinder.findConnection(ExchangeFinder.java:224)
       at okhttp3.internal.connection.ExchangeFinder.findHealthyConnection(ExchangeFinder.java:107)
       at okhttp3.internal.connection.ExchangeFinder.find(ExchangeFinder.java:87)
       at okhttp3.internal.connection.Transmitter.newExchange(Transmitter.java:162)
       at okhttp3.internal.connection.ConnectInterceptor.intercept(ConnectInterceptor.java:41)
       at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:142)
...
```

The majority of the crashes occur on Motorola devices. We have not been able to reproduce the issue on a test device, but it seems related to the app forcing a request over WiFi, and users having a proxy set on the device.

The cause of the NullPointerException is easy to find from the stacktrace, though. From [RealConnection](https://github.com/square/okhttp/blob/df95c77adfbf4f5ddd30bfc6a2bb684295d7d313/okhttp/src/main/java/okhttp3/internal/connection/RealConnection.java#L395):
```
      Http1ExchangeCodec tunnelCodec = new Http1ExchangeCodec(null, null, source, sink);
      source.timeout().timeout(readTimeout, MILLISECONDS);
      sink.timeout().timeout(writeTimeout, MILLISECONDS);
      tunnelCodec.writeRequest(tunnelRequest.headers(), requestLine);
      tunnelCodec.finishRequest();
      Response response = tunnelCodec.readResponseHeaders(false)
          .request(tunnelRequest)
          .build();
```
The second argument to `Http1ExchangeCodec()` (realConnection) is passed as `null`, which causes `readResponseHeaders()` to throw a NullPointerException when an EOFException is encountered.

TangHuaiZhe:
When make an HTTPS connection over an HTTP proxy, the `realConnection` is always set to `null`.
In `RealConnection.java` line 395:
`Http1ExchangeCodec tunnelCodec = new Http1ExchangeCodec(client:null, realConnection:null, source, sink);`
`...`
`Response response = tunnelCodec.readResponseHeaders(false)`

Refer to #4915 NullPointerException in Http1ExchangeCodec.readResponseHeaders() in OkHttp 3.14.0

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/http1/Http1ExchangeCodec.java
+++ b/okhttp/src/main/java/okhttp3/internal/http1/Http1ExchangeCodec.java
@@ -229,8 +229,12 @@ public final class Http1ExchangeCodec implements ExchangeCodec {
       return responseBuilder;
     } catch (EOFException e) {
       // Provide more context if the server ends the stream before sending a response.
+      String address = "unknown";
+      if (realConnection != null) {
+        address = realConnection.route().address().url().redact();
+      }
       throw new IOException("unexpected end of stream on "
-          + realConnection.route().address().url().redact(), e);
+          + address, e);
     }
   }
 


yschimke:
Ideally we would have a test for this in CallTest (?) but not sure how easy it is to reproduce this case.

FWIW you may want to target the 3.14 branch depending how serious the problem is. Http1ExchangeCodec could be kotlinized before this lands also.

TangHuaiZhe:
> Ideally we would have a test for this in CallTest (?) but not sure how easy it is to reproduce this case.
> 
> FWIW you may want to target the 3.14 branch depending how serious the problem is. Http1ExchangeCodec could be kotlinized before this lands also.

The exception looks like related with a commit 20c46127a65d58924910be351f896b31ce323232 of Jesse

>Manage releasing connection separately from HTTP framing


```
      - Http1Codec tunnelConnection = new Http1Codec(null, this, null, source, sink);
      + Http1Codec tunnelConnection = new Http1Codec(null, null, source, sink);

```
The majority of the crashes occurs on Motorola devices means that it is also related to the phone model and hard to reproduce, perhaps just add a null check in exception statement is a cheap solution.

yschimke:
@TangHuaiZhe I'll fix on master as part of Kotlin conversion.  https://github.com/square/okhttp/pull/4938

TangHuaiZhe:
@yschimke  OK, my request will  target the 3.14 branch

PR #4940, opened 2019-04-13
Convert OkHostnameVerifier and BasicTrustRootIndex to Kotlin

swankjesse:


okhttp/src/main/java/okhttp3/internal/tls/BasicTrustRootIndex.kt
```diff
@@ -13,58 +13,41 @@
...
+    val subjectCaCerts = subjectToCaCerts[issuer] ?: return null
 
-    for (X509Certificate caCert : subjectCaCerts) {
-      PublicKey publicKey = caCert.getPublicKey();
+    return subjectCaCerts.firstOrNull {
       try {
-        cert.verify(publicKey);
-        return caCert;
-      } catch (Exception ignored) {
+        cert.verify(it.publicKey)
```
yschimke: it.publicKey isn't guaranteed to be cheap

```
    @Override
    public PublicKey getPublicKey() {
        /* First try to generate the key from supported OpenSSL key types. */
        try {
            OpenSSLKey pkey = new OpenSSLKey(NativeCrypto.X509_get_pubkey(mContext, this));
            return pkey.getPublicKey();
        } catch (NoSuchAlgorithmException ignored) {
        } catch (InvalidKeyException ignored) {
        }

        /* Try generating the key using other Java providers. */
        String oid = NativeCrypto.get_X509_pubkey_oid(mContext, this);
        byte[] encoded = NativeCrypto.i2d_X509_PUBKEY(mContext, this);
        try {
            KeyFactory kf = KeyFactory.getInstance(oid);
            return kf.generatePublic(new X509EncodedKeySpec(encoded));
        } catch (NoSuchAlgorithmException ignored) {
        } catch (InvalidKeySpecException ignored) {
        }

        /*
         * We couldn't find anything else, so just return a nearly-unusable
         * X.509-encoded key.
         */
        return new X509PublicKey(oid, encoded);
    }
```

okhttp/src/main/java/okhttp3/internal/tls/OkHostnameVerifier.kt
```diff
@@ -14,130 +14,79 @@
...
-      return verify(host, (X509Certificate) certificates[0]);
-    } catch (SSLException e) {
-      return false;
+object OkHostnameVerifier : HostnameVerifier {
+  private const val ALT_DNS_NAME = 2
+  private const val ALT_IPA_NAME = 7
+
+  override fun verify(host: String, session: SSLSession): Boolean {
+    return try {
+      verify(host, session.peerCertificates[0] as X509Certificate)
```
yschimke: tioli: first looks nicer even for arrays

okhttp/src/main/java/okhttp3/internal/tls/OkHostnameVerifier.kt
```diff
@@ -14,130 +14,79 @@
...
    */
-  public boolean verifyHostname(String hostname, String pattern) {
+  fun verifyHostname(hostname: String?, pattern: String?): Boolean {
+    var hostname = hostname
+    var pattern = pattern
     // Basic sanity checks
-    // Check length == 0 instead of .isEmpty() to support Java 5.
-    if ((hostname == null) || (hostname.length() == 0) || (hostname.startsWith("."))
-        || (hostname.endsWith(".."))) {
+    if (hostname == null ||
```
yschimke: isNullOrEmpty()

okhttp/src/main/java/okhttp3/internal/tls/BasicTrustRootIndex.kt
swankjesse: Recommendations?

okhttp/src/main/java/okhttp3/internal/tls/OkHostnameVerifier.kt
swankjesse: I think leave it, ’cause it adds an unnecessary range check and replaces an instruction with a method call

okhttp/src/main/java/okhttp3/internal/tls/OkHostnameVerifier.kt
swankjesse: Fixed. This is better if we have contracts (we do) and worse if we didn’t.

okhttp/src/main/java/okhttp3/internal/tls/BasicTrustRootIndex.kt
yschimke: Nope, I misread, thought you had inlined inside a loop.

yschimke: Actually technically you have since it's a predicate.

okhttp/src/main/java/okhttp3/internal/tls/BasicTrustRootIndex.kt
```diff
@@ -0,0 +1,53 @@
...
+ * limitations under the License.
+ */
+package okhttp3.internal.tls
+
+import java.security.cert.X509Certificate
+import javax.security.auth.x500.X500Principal
+
+/** A simple index that of trusted root certificates that have been loaded into memory.  */
+class BasicTrustRootIndex(vararg caCerts: X509Certificate) : TrustRootIndex {
+  private val subjectToCaCerts = mutableMapOf<X500Principal, MutableSet<X509Certificate>>()
```
yschimke: I think it's worth removing the mutable nature here.  Either be converting to just map, or inlining the whole calculation in functional zealot style :)

okhttp/src/main/java/okhttp3/internal/tls/BasicTrustRootIndex.kt
swankjesse: Unfortunately I think we must call `.getPublicKey()` on each object iterated over.

okhttp/src/main/java/okhttp3/internal/tls/BasicTrustRootIndex.kt
swankjesse: Made non-mutable without going full functional zealot. The functional zealot part is difficult ’cause we don’t have a multimap type.

okhttp/src/main/java/okhttp3/internal/tls/BasicTrustRootIndex.kt
yschimke: Maybe one to benchmark later whether it’s worth doing this at init time and storing in a field?

swankjesse: Ahhhh, like storing a `Map<X500Principal, Set<PublicKey>>` instead. Maybe. Out of scope for this PR!

okhttp/src/main/java/okhttp3/internal/tls/OkHostnameVerifier.kt
```diff
@@ -0,0 +1,182 @@
...
+    // yet absolute. This is needed because server certificates do not normally contain absolute
+    // names or patterns, but they should be treated as absolute. At the same time, any hostname
+    // presented to this method should also be treated as absolute for the purposes of matching
+    // to the server certificate.
+    //   www.android.com  matches www.android.com
+    //   www.android.com  matches www.android.com.
+    //   www.android.com. matches www.android.com.
+    //   www.android.com. matches www.android.com
+    if (!hostname.endsWith(".")) {
+      hostname += '.'.toString()
```
JakeWharton:
```suggestion
      hostname += "."
```

okhttp/src/main/java/okhttp3/internal/tls/OkHostnameVerifier.kt
```diff
@@ -0,0 +1,182 @@
...
+    }
+    if (!pattern.endsWith(".")) {
+      pattern += '.'.toString()
```
JakeWharton:
```suggestion
      pattern += "."
```

okhttp/src/main/java/okhttp3/internal/tls/OkHostnameVerifier.kt
```diff
@@ -0,0 +1,182 @@
...
+    }
+    // Hostname and pattern are now absolute domain names.
+
+    pattern = pattern.toLowerCase(Locale.US)
+    // Hostname and pattern are now in lower case -- domain names are case-insensitive.
+
+    if (!pattern.contains("*")) {
```
JakeWharton:
```suggestion
    if ("*" !in pattern) {
```

PR #4938, opened 2019-04-13
HttpExchange Kotlin conversion

yschimke:

PR #4942, opened 2019-04-14
Slightly More Idiomatic StatusLine

yschimke:
Accidentally hit this on another PR, when flushing out build issues.  merged the existing one with my attempt.


yschimke:
Spec says message should always be there via presence of Space, but I think we used to be tolerant.

yschimke:
This PR is included in https://github.com/square/okhttp/pull/4938/files, so no intention to land separately.

yschimke:
cc @shaishavgandhi05 @swankjesse

PR #4943, opened 2019-04-14
ExchangeCodec Kotlin + StatusLine changes

yschimke:
Another attempt - post merge conflicts and one force push too many


okhttp/src/main/java/okhttp3/Response.kt
```diff
@@ -269,7 +269,7 @@ class Response internal constructor(
       this.code = code
     }
 
-    open fun message(message: String) = apply {
+    open fun message(message: String?) = apply {
```
yschimke: The field in Response is Nullable.  StatusLine toString assumes nullable.  So happy to discuss in the review.  But we would previously allow a java client to provide null here, and now we are stricter because of Kotlin checks...

yschimke:
This all went a bit pear shaped in the last PR, once I hit a trivial issue with StatusLine constants in Travis (not CircleCI).  But the nullability issue with Response message seems real.

yschimke:
FWIW I think what we are missing from CircleCI compared to Travis is the two builds - PR & Push

image

https://docs.travis-ci.com/user/pull-requests/#double-builds-on-pull-requests

yschimke:
Going to land *without* StatusLine/Response changes based on approval in https://github.com/square/okhttp/pull/4938

swankjesse:
LGTM

PR #4945, opened 2019-04-14
Testing with unsigned ints

yschimke:
Can't ship with Kotlin 1.3 because experimental, so just exploring.  FYI


okhttp-dnsoverhttps/src/main/java/okhttp3/dnsoverhttps/DnsRecordCodec.kt
```diff
@@ -117,17 +115,17 @@ object DnsRecordCodec {
...
+    var length = source.readByte().toUInt()
 
-    if (length < 0) {
+    if (length and 0xC000u != 0u) {
       // compressed name pointer, first two bits are 1
       // drop second byte of compression offset
       source.skip(1)
     } else {
-      while (length > 0) {
+      while (length and 0xC000u != 0u) {
```
yschimke: could check == 0xC0000u

yschimke:
Will close out this PR, once tests are passing. Then later reopen for Kotlin 1.4?

swankjesse:
Yep!

PR #4947, opened 2019-04-14
Convert DiskLruCache to Kotlin

swankjesse:


okhttp/src/main/java/okhttp3/internal/cache/DiskLruCache.kt
```diff
@@ -13,371 +13,350 @@
...
-      } catch (IOException journalIsCorrupt) {
-        Platform.get().log(WARN, "DiskLruCache " + directory + " is corrupt: "
-            + journalIsCorrupt.getMessage() + ", removing", journalIsCorrupt);
+        readJournal()
+        processJournal()
+        initialized = true
+        return
+      } catch (journalIsCorrupt: IOException) {
+        Platform.get().log(WARN, "DiskLruCache $directory is corrupt: ${journalIsCorrupt.message}" +
+            ", removing", journalIsCorrupt)
```
yschimke: might be worth putting log statement on the next line to have it all together

okhttp/src/main/java/okhttp3/internal/cache/DiskLruCache.kt
```diff
@@ -13,371 +13,350 @@
...
+   * compacted by dropping redundant lines. A temporary file named "journal.tmp" will be used during
+   * compaction; that file should be deleted if it exists when the cache is opened.
+   */
+
+  private val journalFile: File
+  private val journalFileTmp: File
+  private val journalFileBackup: File
+  private var size: Long = 0L
+  private var journalWriter: BufferedSink? = null
+  internal val lruEntries = LinkedHashMap<String, Entry>(0, 0.75f, true)
```
yschimke: mutableMapOf?

okhttp/src/main/java/okhttp3/internal/cache/DiskLruCache.kt
```diff
@@ -386,664 +365,642 @@ public final class DiskLruCache implements Closeable, Flushable {
...
 
-    redundantOpCount++;
-    entry.currentEditor = null;
-    if (entry.readable | success) {
-      entry.readable = true;
-      journalWriter.writeUtf8(CLEAN).writeByte(' ');
-      journalWriter.writeUtf8(entry.key);
-      entry.writeLengths(journalWriter);
-      journalWriter.writeByte('\n');
+    val journalWriter = this.journalWriter!!
```
yschimke: This block is a good candidate for `this.journalWriter.apply { ... }`

okhttp/src/main/java/okhttp3/internal/cache/DiskLruCache.kt
```diff
@@ -386,664 +365,642 @@ public final class DiskLruCache implements Closeable, Flushable {
...
+    val removed = removeEntry(entry)
+    if (removed && size <= maxSize) mostRecentTrimFailed = false
+    return removed
   }
 
-  boolean removeEntry(Entry entry) throws IOException {
-    if (entry.currentEditor != null) {
-      entry.currentEditor.detach(); // Prevent the edit from completing normally.
-    }
+  @Throws(IOException::class)
```
yschimke: For internal classes, a lot of the throws can be presumably dropped.  Left just on public APIs.

okhttp/src/main/java/okhttp3/internal/cache/DiskLruCache.kt
```diff
@@ -13,371 +13,350 @@
...
-  private BufferedSink newJournalWriter() throws FileNotFoundException {
-    Sink fileSink = fileSystem.appendingSink(journalFile);
-    Sink faultHidingSink = new FaultHidingSink(fileSink) {
-      @Override protected void onException(IOException e) {
-        assert (Thread.holdsLock(DiskLruCache.this));
-        hasJournalErrors = true;
+  @Throws(FileNotFoundException::class)
+  private fun newJournalWriter(): BufferedSink {
+    val fileSink = fileSystem.appendingSink(journalFile)
+    val faultHidingSink = object : FaultHidingSink(fileSink) {
```
yschimke: TIOLI: Might be a good time to make FaultHidingSink a final class and take an exception handler :)

```
val faultHidingSink = FaultHidingSink(fileSink) { ex ->
   assert(Thread.holdsLock(this@DiskLruCache))
   hasJournalErrors = true
}
```

okhttp/src/main/java/okhttp3/internal/cache/DiskLruCache.kt
swankjesse: Good call, fixed

okhttp/src/main/java/okhttp3/internal/cache/DiskLruCache.kt
swankjesse: The `,true` at the end is key

okhttp/src/main/java/okhttp3/internal/cache/DiskLruCache.kt
swankjesse: Dig.

okhttp/src/main/java/okhttp3/internal/cache/DiskLruCache.kt
swankjesse: I want to keep them as long as we’re supporting Java callers (ie. forever). Otherwise it’s very difficult to know when an `@Throws` is necessary.

okhttp/src/main/java/okhttp3/internal/cache/DiskLruCache.kt
swankjesse: Yeah, I like that. Done.

PR #4952, opened 2019-04-15
Example Kotlin Cleanups

yschimke:
These are borderline terrible, but they flag in error prone and highlight in the IDE.  Putting this here for discussion only, what's the approach to getting a clean warning free Kotlin experience?

### DIFF
--- a/okhttp/src/main/java/okhttp3/Cookie.kt
+++ b/okhttp/src/main/java/okhttp3/Cookie.kt
@@ -39,6 +39,7 @@ import java.util.regex.Pattern
  *
  * [chromium_extension]: https://code.google.com/p/chromium/issues/detail?id=232693
  */
+@Suppress("DataClassPrivateConstructor")
 data class Cookie private constructor(
   private val name: String,
   private val value: String,
--- a/okhttp/src/main/java/okhttp3/Cookie.kt
+++ b/okhttp/src/main/java/okhttp3/Cookie.kt
@@ -39,7 +39,6 @@ import java.util.regex.Pattern
  *
  * [chromium_extension]: https://code.google.com/p/chromium/issues/detail?id=232693
  */
-@Suppress("DataClassPrivateConstructor")
 data class Cookie private constructor(
   private val name: String,
   private val value: String,
--- a/okhttp/src/main/java/okhttp3/internal/connection/RealConnectionPool.kt
+++ b/okhttp/src/main/java/okhttp3/internal/connection/RealConnectionPool.kt
@@ -21,7 +21,10 @@ import okhttp3.Route
 import okhttp3.internal.Util
 import okhttp3.internal.Util.closeQuietly
 import okhttp3.internal.connection.Transmitter.TransmitterReference
+import okhttp3.internal.notifyAll
 import okhttp3.internal.platform.Platform
+import okhttp3.internal.wait
+import okhttp3.internal.waitNanos
 import java.io.IOException
 import java.net.Proxy
 import java.util.ArrayDeque
@@ -111,8 +114,7 @@ class RealConnectionPool(
       return true
     } else {
       // Awake the cleanup thread: we may have exceeded the idle connection limit.
-      @Suppress("PLATFORM_CLASS_MAPPED_TO_KOTLIN")
-      (this as java.lang.Object).notifyAll()
+      this.notifyAll()
       return false
     }
   }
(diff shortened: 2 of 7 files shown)


okhttp/src/main/java/okhttp3/internal/http/RealInterceptorChain.kt
```diff
@@ -120,11 +120,6 @@ class RealInterceptorChain(
           " must call proceed() exactly once")
     }
 
-    // Confirm that the intercepted response isn't null.
-    if (response == null) {
```
JayNewstrom: I thought this was here to ensure nullability of what's returned from java callers.

yschimke: My understanding is that by this point in Kotlin, it can't be null.

yschimke:
@JakeWharton are we happy with sprinkling \@Suppress around our code?

okhttp/src/main/java/okhttp3/Cookie.kt
```diff
@@ -39,6 +39,7 @@ import java.util.regex.Pattern
  *
  * [chromium_extension]: https://code.google.com/p/chromium/issues/detail?id=232693
  */
+@Suppress("DataClassPrivateConstructor")
```
swankjesse: Don't suppress this one? It's sweeping a real problem under the rug and we haven't yet addressed that problem.

okhttp/src/main/java/okhttp3/internal/connection/RealConnectionPool.kt
```diff
@@ -111,6 +111,7 @@ class RealConnectionPool(
       return true
     } else {
       // Awake the cleanup thread: we may have exceeded the idle connection limit.
+      @Suppress("PLATFORM_CLASS_MAPPED_TO_KOTLIN")
```
swankjesse: Maybe create extensions for these? And suppress all in one place rather than littering callsites.

okhttp/src/main/java/okhttp3/internal/http/RealInterceptorChain.kt
yschimke: TIL - Kotlin null checks are a sham

yschimke: OK, so situation is

1) Without explicit null check in our code, the return value is enforced

```
    val response = interceptor.intercept(next)

    // Confirm that the next interceptor made its required call to chain.proceed().
    if (exchange != null && index + 1 < interceptors.size && next.calls != 1) {
      throw IllegalStateException("network interceptor " + interceptor +
          " must call proceed() exactly once")
    }

//    // Needed to enforce Kotlin null checks
//    @Suppress("SENSELESS_COMPARISON")
//    if (response == null) {
//      throw NullPointerException("interceptor $interceptor returned null")
//    }

//    if (response.body() == null) {
//      throw IllegalStateException(
//          "interceptor $interceptor returned a response with no body")
//    }

    return response
```

NPE

```
	at okhttp3.internal.cache.CacheInterceptor.intercept(CacheInterceptor.java:124)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:115)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:90)
```

yschimke: With *later* explicit check

```
java.lang.NullPointerException: interceptor okhttp3.InterceptorTest$$Lambda$62/0x0000000800157c40@45752059 returned null
	at okhttp3.internal.connection.ConnectInterceptor.intercept(ConnectInterceptor.java:43)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:115)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:90)
```

yschimke: So Kotlin doesn't differentiate between a NPE without a message and with one.  But our test does.  So we are both right.

JayNewstrom: Cool! 👍Good to learn regardless!

PR #4954, opened 2019-04-15
 Convert CallServerInterceptor to Kotlin

TangHuaiZhe:
Convert CallServerInterceptor to Kotlin


yschimke:
It would be nice to see if we can simplify the logic to remove !! cases. Also the “close” checks might be worth pulling out into internal Request and Response receiver methods to make the code logic simpler. 

TangHuaiZhe:
@yschimke There are two places use `!!` statement which is safe because `exchange.readResponseHeaders(false)` never be null.
Wait until  `Exchange.java` be converted to Kotlin we can remove these `!!` case.

okhttp/src/main/java/okhttp3/internal/http/CallServerInterceptor.kt
```diff
@@ -0,0 +1,129 @@
...
+    }
+    if (!responseHeadersStarted) {
+      exchange.responseHeadersStart()
+    }
+    if (responseBuilder == null) {
+      // exchange.readResponseHeaders(false) never returns null
+      // while exchange.readResponseHeaders(true) always returns null
+      responseBuilder = exchange.readResponseHeaders(false)
+    }
+      // safe to use!!
```
swankjesse: I very much dislike this comment because if it weren't safe you wouldn't have used it. Remove the comment please.

okhttp/src/main/java/okhttp3/internal/http/CallServerInterceptor.kt
```diff
@@ -0,0 +1,129 @@
...
+    if (requestBody == null || !requestBody.isDuplex()) {
+      exchange.finishRequest()
+    }
+    if (!responseHeadersStarted) {
+      exchange.responseHeadersStart()
+    }
+    if (responseBuilder == null) {
+      // exchange.readResponseHeaders(false) never returns null
+      // while exchange.readResponseHeaders(true) always returns null
+      responseBuilder = exchange.readResponseHeaders(false)
```
swankjesse: You want !! at the end of this line, not on like 87. And remove the 2 comment lines above.

okhttp/src/main/java/okhttp3/internal/http/CallServerInterceptor.kt
```diff
@@ -0,0 +1,129 @@
...
+        .request(request)
+        .handshake(exchange.connection().handshake())
+        .sentRequestAtMillis(sentRequestMillis)
+        .receivedResponseAtMillis(System.currentTimeMillis())
+        .build()
+    var code = response.code()
+    if (code == 100) {
+      // server sent a 100-continue even though we did not request one.
+      // try again to read the actual response
+      // exchange.readResponseHeaders(false) never be null
```
swankjesse: Please remove this comment.

swankjesse: I don't like the comment because it acts like a Jedi mind trick. “this value is not null” in a comment doesn't make it so.

okhttp/src/main/java/okhttp3/internal/http/CallServerInterceptor.kt
```diff
@@ -0,0 +1,129 @@
...
+          .body(exchange.openResponseBody(response))
+          .build()
+    }
+    if ("close".equals(response.request().header("Connection"), ignoreCase = true) ||
+        "close".equals(response.header("Connection"), ignoreCase = true)) {
+      exchange.noNewExchangesOnConnection()
+    }
+    if ((code == 204 || code == 205) && response.body()?.contentLength() ?: -1 > 0) {
+      throw ProtocolException(
+          "HTTP " + code + " had non-zero Content-Length: " + response.body()?.contentLength())
```
swankjesse: Template please

okhttp/src/main/java/okhttp3/internal/http/CallServerInterceptor.kt
TangHuaiZhe: Fixed

okhttp/src/main/java/okhttp3/internal/http/CallServerInterceptor.kt
TangHuaiZhe: Yes, Add `!!` at the end of this line is much better.
Fixed.

okhttp/src/main/java/okhttp3/internal/http/CallServerInterceptor.kt
TangHuaiZhe: Fixed

okhttp/src/main/java/okhttp3/internal/http/CallServerInterceptor.kt
TangHuaiZhe: Fixed

PR #4961, opened 2019-04-17
Convert WebSocketProtocol to Kotlin

shaishavgandhi05:


okhttp/src/main/java/okhttp3/internal/InternalKt.kt
```diff
@@ -70,3 +72,9 @@ fun Any.waitMillis(timeout: Long, nanos: Int = 0) {
 fun Any.notify() = (this as Object).notify()
 
 fun Any.notifyAll() = (this as Object).notifyAll()
+
+fun toggleMask(cursor: Buffer.UnsafeCursor, key: ByteArray) = WebSocketProtocol.toggleMask(cursor, key)
```
yschimke: What is the value in having these here as opposed to just in WebSocketProtocol? and used by reader/writer?

shaishavgandhi05: They're internal in `WebSocketProtocol` and can't be consumed by the Java files. Once the reader/writer are Kotlin, we can purge the functions in `InternalKt` and things will work as intended. 

The above strategy seemed to be what we did for public types. If we're not as strict for the internal classes, then I don't mind making those functions public in `WebSocketProtocol` itself.

yschimke: Yep, I'd suggest making them public for now.  Internal is off limits anyway.

okhttp/src/main/java/okhttp3/internal/ws/WebSocketProtocol.kt
```diff
@@ -0,0 +1,141 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okhttp3.internal.ws
+
+import okio.Buffer
+import okio.ByteString.Companion.encodeUtf8
+import kotlin.experimental.xor
```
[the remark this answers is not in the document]
shaishavgandhi05: Updated 7a3627f34fdaa27e893a1aad10bc110e356a7d15

okhttp/src/main/java/okhttp3/internal/InternalKt.kt
shaishavgandhi05: 543ddbe109061a431070a197e4c4acc290187069

okhttp/src/main/java/okhttp3/internal/ws/WebSocketProtocol.kt
```diff
@@ -0,0 +1,146 @@
...
+    val keyLength = key.size
+    do {
+      val buffer = cursor.data
+      var i = cursor.start
+      val end = cursor.end
+      if (buffer != null) {
+        while (i < end) {
+          keyIndex %= keyLength // Reassign to prevent overflow breaking counter.
+
+          // Byte xor is experimental in Kotlin so we coerce bytes to int, xor them
```
swankjesse: This is what Java does anyway

PR #4970, opened 2019-04-21
Fix race condition in RecordingHostnameVerifier

amirlivneh:
When connections are established concurrently, verify() sometimes throws a `java.lang.ArrayIndexOutOfBoundsException: Index 1 out of bounds for length 0` because of unsynchronized access to the `calls` list.


swankjesse:
Wanna make `verify()` synchronized instead? `synchronizedList()` is bad because it uses a weird lock, and won’t work if we want another method like `take()` later.

amirlivneh:
I made `verify()` synchronized instead.

PR #4983, opened 2019-04-27
Increase documentation of HttpUrl.Builder

romeara:
Add documentation for the `HttpUrl.newBuilder()` method indicating that initial values are read from the existing URL, and for `HttpUrl.Builder.scheme` indicating the legal values for the `scheme` parameter

### DIFF
--- a/okhttp/src/main/java/okhttp3/HttpUrl.kt
+++ b/okhttp/src/main/java/okhttp3/HttpUrl.kt
@@ -832,7 +832,7 @@ class HttpUrl internal constructor(builder: Builder) {
     }
 
     /**
-     * @param scheme either http or https
+     * @param scheme either "http" or "https"
      */
     fun scheme(scheme: String): Builder {
       when {
--- a/okhttp/src/main/java/okhttp3/HttpUrl.kt
+++ b/okhttp/src/main/java/okhttp3/HttpUrl.kt
@@ -754,7 +754,7 @@ class HttpUrl internal constructor(builder: Builder) {
   fun resolve(link: String): HttpUrl? = newBuilder(link)?.build()
 
   /**
-   * Returns a builder based on this URL
+   * Returns a builder based on this URL.
    */
   fun newBuilder(): Builder {
     val result = Builder()
@@ -832,7 +832,7 @@ class HttpUrl internal constructor(builder: Builder) {
     }
 
     /**
-     * @param scheme either "http" or "https"
+     * @param scheme either "http" or "https".
      */
     fun scheme(scheme: String): Builder {
       when {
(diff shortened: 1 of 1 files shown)


okhttp/src/main/java/okhttp3/HttpUrl.kt
```diff
@@ -753,6 +753,9 @@ class HttpUrl internal constructor(builder: Builder) {
    */
   fun resolve(link: String): HttpUrl? = newBuilder(link)?.build()
 
+  /**
+   * Returns a builder based on this URL
```
swankjesse: https://publicobject.com/2007/08/05/stop_writing_redundant_javadoc/

okhttp/src/main/java/okhttp3/HttpUrl.kt
```diff
@@ -828,6 +831,9 @@ class HttpUrl internal constructor(builder: Builder) {
       encodedPathSegments.add("") // The default path is '/' which needs a trailing space.
     }
 
+    /**
+     * @param scheme either http or https
```
swankjesse: Can you quote these, "http" or "https" ? Thanks!

okhttp/src/main/java/okhttp3/HttpUrl.kt
romeara: I felt this was necessary because while it seemed reasonable that this started with HttpUrl's values, it also seemed a possibility that it started "blank" - hence why it seemed like it might be worthwhile to add a small bit of doc clarifying this.
If you'd rather see it removed, or additional information added to make it a little less basic, I can alter the change

okhttp/src/main/java/okhttp3/HttpUrl.kt
romeara: Sure! Quotes have been added

okhttp/src/main/java/okhttp3/HttpUrl.kt
swankjesse: Sounds good. Add a period on the end of the sentence please.

okhttp/src/main/java/okhttp3/HttpUrl.kt
```diff
@@ -828,6 +831,9 @@ class HttpUrl internal constructor(builder: Builder) {
       encodedPathSegments.add("") // The default path is '/' which needs a trailing space.
     }
 
+    /**
+     * @param scheme either "http" or "https"
```
swankjesse: period at the end of this too. https://nedbatchelder.com/blog/201401/comments_should_be_sentences.html

okhttp/src/main/java/okhttp3/HttpUrl.kt
romeara: Period added, sorry for missing that convention from the other doc :)

okhttp/src/main/java/okhttp3/HttpUrl.kt
romeara: Period added

PR #4985, opened 2019-04-27
Travis CI builds in JDK 8 only

yschimke:
No need to build all here given progress towards CircleCI.


yschimke:
Prompted by https://travis-ci.org/square/okhttp/jobs/525225475

PR #4987, opened 2019-04-28
Convert Hpack to Kotlin

swankjesse:


okhttp/src/test/java/okhttp3/internal/http2/HpackTest.java
```diff
@@ -183,7 +183,7 @@
   /** Header table backing array is initially 8 long, let's ensure it grows. */
   @Test public void dynamicallyGrowsBeyond64Entries() throws IOException {
     // Lots of headers need more room!
-    hpackReader = new Hpack.Reader(16384, 4096, bytesIn);
+    hpackReader = new Hpack.Reader(bytesIn, 16384, 4096);
```
yschimke: We should start naming params of ints and booleans to avoid confusion in the codebase, I suggest starting with changes that come through these PRs.

okhttp/src/main/java/okhttp3/internal/http2/Hpack.kt
```diff
@@ -13,576 +13,584 @@
...
-    private final List<Header> headerList = new ArrayList<>();
-    private final BufferedSource source;
+  /**
+   * The decoder has ultimate control of the maximum size of the dynamic table but we can choose
+   * to use less. We'll put a cap at 16K. This is arbitrary but should be enough for most purposes.
+   */
+  private const val SETTINGS_HEADER_TABLE_SIZE_LIMIT = 16_384
+
+  val STATIC_HEADER_TABLE = arrayOf(
+      Header(Header.TARGET_AUTHORITY, ""),
```
yschimke: If you import the "Header." part of the key, it would be a nice cleanup.

`Header(Header.TARGET_AUTHORITY, ""),`

okhttp/src/main/java/okhttp3/internal/http2/Header.kt
```diff
@@ -26,16 +25,14 @@ data class Header(
   /** Value in UTF-8 encoding. */
   @JvmField val value: ByteString
```
yschimke: Can you default value to "" to cleanup definitions below?

okhttp/src/main/java/okhttp3/internal/http2/Hpack.kt
```diff
@@ -13,576 +13,584 @@
...
+    @Throws(IOException::class)
+    fun readByteString(): ByteString {
+      val firstByte = readByte()
+      val huffmanDecode = firstByte and 0x80 == 0x80 // 1NNNNNNN
+      val length = readInt(firstByte, PREFIX_7_BITS)
 
       if (huffmanDecode) {
-        return ByteString.of(Huffman.get().decode(source.readByteArray(length)));
+        val bytes = source.readByteArray(length.toLong())
+        return ByteString.of(*Huffman.get().decode(bytes))
```
yschimke: Does the splat (*) to a vararg method imply an extra copy?

okhttp/src/main/java/okhttp3/internal/http2/Hpack.kt
```diff
@@ -13,576 +13,584 @@
...
+  private fun nameToFirstIndex(): Map<ByteString, Int> {
+    val result = LinkedHashMap<ByteString, Int>(STATIC_HEADER_TABLE.size)
+    for (i in STATIC_HEADER_TABLE.indices) {
       if (!result.containsKey(STATIC_HEADER_TABLE[i].name)) {
-        result.put(STATIC_HEADER_TABLE[i].name, i);
+        result[STATIC_HEADER_TABLE[i].name] = i
       }
     }
-    return Collections.unmodifiableMap(result);
+    return Collections.unmodifiableMap(result)
```
yschimke: Is this still needed here, or is immutable Map vs MutableMap strong enough here?

yschimke:
Failure do to travis - JDK11 issues not a build failure

okhttp/src/test/java/okhttp3/internal/http2/HpackTest.java
swankjesse: Can do for Kotlin callsites!

okhttp/src/main/java/okhttp3/internal/http2/Hpack.kt
swankjesse: Done!

okhttp/src/main/java/okhttp3/internal/http2/Header.kt
swankjesse: I don’t think it’s a good default for `Header` because only Hpack wants empty strings here.

okhttp/src/main/java/okhttp3/internal/http2/Hpack.kt
swankjesse: It does and that’s sad. Here’s the decompiled bytecode:

```
      @NotNull
      public final ByteString readByteString() throws IOException {
         int firstByte = this.readByte();
         boolean huffmanDecode = (firstByte & 128) == 128;
         int length = this.readInt(firstByte, 127);
         if (huffmanDecode) {
            byte[] bytes = this.source.readByteArray((long)length);
            Companion var10000 = ByteString.Companion;
            byte[] var10001 = Huffman.get().decode(bytes);
            Intrinsics.checkExpressionValueIsNotNull(var10001, "Huffman.get().decode(bytes)");
            return var10000.of(Arrays.copyOf(var10001, var10001.length));
         } else {
            return this.source.readByteString((long)length);
         }
      }
```

swankjesse: ... I changed to use a better API, `ByteArray.toByteString()` which saves the copy. Good catch!

```
      @NotNull
      public final ByteString readByteString() throws IOException {
         int firstByte = this.readByte();
         boolean huffmanDecode = (firstByte & 128) == 128;
         int length = this.readInt(firstByte, 127);
         if (huffmanDecode) {
            byte[] bytes = this.source.readByteArray((long)length);
            Companion var10000 = ByteString.Companion;
            byte[] var10001 = Huffman.get().decode(bytes);
            Intrinsics.checkExpressionValueIsNotNull(var10001, "Huffman.get().decode(bytes)");
            return Companion.of$default(var10000, var10001, 0, 0, 3, (Object)null);
         } else {
            return this.source.readByteString((long)length);
         }
      }
```

okhttp/src/main/java/okhttp3/internal/http2/Hpack.kt
swankjesse: I don’t like Kotlin’s immutable types. They aren’t immutable to Java callers, and they aren’t immutable if you cast.

okhttp/src/main/java/okhttp3/internal/http2/Hpack.kt
swankjesse: ... also fixed 3 similar splats in Cache, CertificatePinner, and HeldCertificate.

okhttp/src/main/java/okhttp3/internal/http2/Hpack.kt
JakeWharton: Not sure we should care about casts. Once you go down that road you can always get access to mutable internals.

swankjesse: I suppose it’s okay for internal APIs.

For published APIs returning a mutable result discourages us from returning an immutable result later.

It'd be much less problematic if Java’s type system disambiguated the two.

PR #4989, opened 2019-04-28
Convert RetryAndFollowUpInterceptor to Kotlin

TangHuaiZhe:
Update RetryAndFollowUpInterceptor.kt


okhttp/src/main/java/okhttp3/internal/http/RetryAndFollowUpInterceptor.kt
```diff
@@ -0,0 +1,330 @@
...
+        success = true
+      } catch (e: RouteException) {
+        // The attempt to connect via a route failed. The request will not have been sent.
+        if (!recover(e.lastConnectException, transmitter, false, request)) {
+          throw e.firstConnectException
+        }
+        continue
+      } catch (e: IOException) {
+        // An attempt to communicate with a server failed. The request may have been sent.
+        val requestSendStarted = e !is ConnectionShutdownException
```
swankjesse: nice. `x !is y` is so much nicer than `!(x instanceof y)`

okhttp/src/main/java/okhttp3/internal/http/RetryAndFollowUpInterceptor.kt
```diff
@@ -0,0 +1,330 @@
...
+
+    return requestBuilder.url(url).build()
+  }
+
+  private fun retryAfter(userResponse: Response, defaultDelay: Int): Int {
+    val header = userResponse.header("Retry-After") ?: return defaultDelay
+
+    // https://tools.ietf.org/html/rfc7231#section-7.1.3
+    // currently ignores a HTTP-date, and assumes any non int 0 is a delay
+    if (header.matches("\\d+".toRegex())) {
```
yschimke: Can be simplified 

`header.toIntOrNull() ?: Int.MAX_VALUE`

okhttp/src/main/java/okhttp3/internal/http/RetryAndFollowUpInterceptor.kt
```diff
@@ -0,0 +1,330 @@
...
+      }
+
+      HTTP_UNAVAILABLE -> {
+        val priorResponse = userResponse.priorResponse()
+        if (priorResponse != null && priorResponse.code() == HTTP_UNAVAILABLE) {
+          // We attempted to retry and got another timeout. Give up.
+          return null
+        }
+
+        if (retryAfter(userResponse, Integer.MAX_VALUE) == 0) {
```
yschimke: Should we use Int references here?

swankjesse: Once we're all migrated over we'll need a couple passes to use Kotlin things wherever they're preferred.

PR #4990, opened 2019-04-28
Delete the unused DistinguishedNameParser.

swankjesse:
This hasn't been used since we stopped supporting CN verification in
#3764.


swankjesse:
Tough to do mechanically ’cause it's covered by tests. Only hint is it's an internal class not reachable by non-internal classes.

JakeWharton:
I'll play with it.

PR #4994, opened 2019-04-30
Update README.md

bigman212:
Added link to Maven repo to simplify search for latest version number.

### DIFF
--- a/okhttp-logging-interceptor/README.md
+++ b/okhttp-logging-interceptor/README.md
@@ -45,11 +45,12 @@ Get via Maven:
 </dependency>
 ```
 
-or via Gradle 
+or via Gradle
 ```groovy
 implementation 'com.squareup.okhttp3:logging-interceptor:(insert latest version)'
 ```
-
+Latest version can be found in [Maven][2] repository.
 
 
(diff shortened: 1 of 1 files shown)


JakeWharton:
Even if we wanted to link, search.maven.org is the correct place. mvnrepository is a third-party index that's frequently out-of-date and shouldn't be used.

bigman212:
> Even if we wanted to link, search.maven.org is the correct place. mvnrepository is a third-party index that's frequently out-of-date and shouldn't be used.

Oops, sorry for that, my mistake.
Should "same version as OkHttp dependency" recommendation go to readme.md then? How people, for example, that, are using retrofit should check OkHttp version?

swankjesse:
If you’re using the logging interceptor, you should explicitly specify it’s version and the OkHttp version.

bigman212:
Okay. What should i do next then with this PR?

swankjesse:
Change the doc to recommend using the same version as your OkHttp artifact? Mind changing our other optional modules also?

swankjesse:
Fixed our tools so the latest version is easy.
https://github.com/square/okhttp/pull/5129

PR #4997, opened 2019-05-02
Convert CacheInterceptor to kotlin

TangHuaiZhe:


okhttp/src/main/java/okhttp3/internal/cache/CacheInterceptor.kt
```diff
@@ -0,0 +1,267 @@
...
+      if (networkResponse?.code() == HTTP_NOT_MODIFIED) {
+        val response = cacheResponse.newBuilder()
+            .headers(combine(cacheResponse.headers(), networkResponse.headers()))
+            .sentRequestAtMillis(networkResponse.sentRequestAtMillis())
+            .receivedResponseAtMillis(networkResponse.receivedResponseAtMillis())
+            .cacheResponse(stripBody(cacheResponse))
+            .networkResponse(stripBody(networkResponse))
+            .build()
+
+        networkResponse.body()?.close()
```
swankjesse: strictly speaking this should be `!!.close()`, not `?.close()`

okhttp/src/main/java/okhttp3/internal/cache/CacheInterceptor.kt
```diff
@@ -0,0 +1,267 @@
...
+
+        // Update the cache after combining headers but before stripping the
+        // Content-Encoding header (as performed by initContentStream()).
+        if (cache != null) {
```
swankjesse: here we know the cache is non-null, the null check should be replaced with the `!!` operator

okhttp/src/main/java/okhttp3/internal/cache/CacheInterceptor.kt
```diff
@@ -0,0 +1,267 @@
...
+    }
+
+    // If we don't need the network, we're done.
+    if (networkRequest == null) {
+      return cacheResponse!!.newBuilder()
+          .cacheResponse(stripBody(cacheResponse))
+          .build()
+    }
+
+    var networkResponse: Response? = null
```
swankjesse: Can you declare this as

```
var networkResponse: Response
```

don’t declare as nullable, don’t assign to null

okhttp/src/main/java/okhttp3/internal/cache/CacheInterceptor.kt
```diff
@@ -0,0 +1,267 @@
...
+                MILLISECONDS)) {
+          cacheRequestClosed = true
+          cacheRequest.abort()
+        }
+        source.close()
+      }
+    }
+
+    val contentType = response.header("Content-Type")
+    val contentLength = response.body()?.contentLength() ?: -1
```
swankjesse: replace the `?.` operator with the `!!.` operator. There is no `-1` here.

okhttp/src/main/java/okhttp3/internal/cache/CacheInterceptor.kt
```diff
@@ -0,0 +1,267 @@
...
+    return response.newBuilder()
+        .body(RealResponseBody(contentType, contentLength, cacheWritingSource.buffer()))
+        .build()
+  }
+
+  companion object {
+
+    private fun stripBody(response: Response?): Response? {
+      if (response?.body() != null)
```
swankjesse: braces on all if clauses and else clauses that wrap

okhttp/src/main/java/okhttp3/internal/cache/CacheInterceptor.kt
```diff
@@ -0,0 +1,267 @@
...
+
+    /** Combines cached headers with a network headers as defined by RFC 7234, 4.3.4.  */
+    private fun combine(cachedHeaders: Headers, networkHeaders: Headers): Headers {
+      val result = Headers.Builder()
+
+      for (index in cachedHeaders.names().indices) {
+        val fieldName = cachedHeaders.name(index)
+        val value = cachedHeaders.value(index)
+        if ("Warning".equals(fieldName, ignoreCase = true) && value.startsWith("1")) {
+          continue
```
swankjesse: needs a comment, ` // Drop 100-level freshness warnings.`

okhttp/src/main/java/okhttp3/internal/cache/CacheInterceptor.kt
```diff
@@ -0,0 +1,267 @@
...
+        }
+        if (isContentSpecificHeader(fieldName) ||
+            !isEndToEnd(fieldName) ||
+            networkHeaders[fieldName] == null) {
+          addHeaderLenient(result, fieldName, value)
+        }
+      }
+
+      for (index in networkHeaders.names().indices) {
```
swankjesse: does the `.indices` call allocate?

okhttp/src/main/java/okhttp3/internal/cache/CacheInterceptor.kt
```diff
@@ -0,0 +1,267 @@
...
+
+      return result.build()
+    }
+
+    /**
+     * Returns true if `fieldName` is an end-to-end HTTP header, as defined by RFC 2616,
+     * 13.5.1.
+     */
+    internal fun isEndToEnd(fieldName: String): Boolean {
+      return (!"Connection".equals(fieldName, ignoreCase = true) &&
```
swankjesse: are the parens necessary?

okhttp/src/main/java/okhttp3/internal/cache/CacheInterceptor.kt
TangHuaiZhe: > Can you declare this as
> 
> ```
> var networkResponse: Response
> ```
> 
> don’t declare as nullable, don’t assign to null

I tried to declare `networkResponse:Response `means
`networkResponse` will never be null,
But actually we have a `networkResponse == null `scene:
In the finally statement:
```
// If we're crashing on I/O or otherwise, don't leak the cache body.
If (networkResponse == null && cacheCandidate != null) {
  closeQuietly(cacheCandidate.body())
}
```
After modification, it becomes:
```
If (cacheCandidate != null){
  closeQuietly(cacheCandidate.body())
}
```

When `networkResponse != null && cacheCandidate != null`,
The original logic does not `closeQuietly`,
After modification, it will `closeQuietly`, causing `networkResponse.body.bodySource.closed = true`

This causes `assertThat(response2.body().string()).isEqualTo("A")` be failed in the `conditionalHitUpdatesCache()` method `in CacheTest`.

```
java.lang.IllegalStateException: closed

At okio.RealBufferedSource.select(RealBufferedSource.kt:83)
At okhttp3.internal.Util.bomAwareCharset(Util.java:443)
At okhttp3.ResponseBody.string(ResponseBody.kt:170)
At okhttp3.CacheTest.conditionalHitUpdatesCache(CacheTest.java:1925)
```

okhttp/src/main/java/okhttp3/internal/cache/CacheInterceptor.kt
TangHuaiZhe: > does the `.indices` call allocate?

I tried to decompile `for (index in cachedHeaders.names().indices)` into a class file and got:
```
Int i = 0;
For (int j = ((Collection)cachedHeaders.names()).size(); i < j; index++)

```
It does not seem to allocate extra memory to the `IntRange` object.

There is a blog about this issue, it seems that Kotlin is doing very well:
[Exploring Kotlin’s hidden costs](https://medium.com/@BladeCoder/exploring-kotlins-hidden-costs-part-3-3bf6e0dbf0a4)
In this article, you can directly search for the keyword **collection indices** and you can see related information.

okhttp/src/main/java/okhttp3/internal/cache/CacheInterceptor.kt
TangHuaiZhe: Fixed

okhttp/src/main/java/okhttp3/internal/cache/CacheInterceptor.kt
TangHuaiZhe: Fixed

okhttp/src/main/java/okhttp3/internal/cache/CacheInterceptor.kt
TangHuaiZhe: Fixed

okhttp/src/main/java/okhttp3/internal/cache/CacheInterceptor.kt
TangHuaiZhe: Fixed

okhttp/src/main/java/okhttp3/internal/cache/CacheInterceptor.kt
TangHuaiZhe: Fixed

okhttp/src/main/java/okhttp3/internal/cache/CacheInterceptor.kt
TangHuaiZhe: Fixed

okhttp/src/main/java/okhttp3/internal/cache/CacheInterceptor.kt
swankjesse: Cool!

swankjesse:
Thanks!

PR #4999, opened 2019-05-03
Convert Huffman to Kotlin

swankjesse:


okhttp/src/main/java/okhttp3/internal/http2/Huffman.kt
```diff
@@ -0,0 +1,208 @@
...
+  }
+
+  private fun addCode(symbol: Int, code: Int, codeBitCount: Int) {
+    val terminal = Node(symbol, codeBitCount)
+
+    var accumulatorBitCount = codeBitCount
+    var node = root
+    while (accumulatorBitCount > 8) {
+      accumulatorBitCount -= 8
+      val childIndex = code.ushr(accumulatorBitCount) and 0xff
```
JakeWharton:
```suggestion
      val childIndex = code ushr accumulatorBitCount and 0xff
```
To match line 176

okhttp/src/main/java/okhttp3/internal/http2/Huffman.kt
```diff
@@ -0,0 +1,208 @@
...
+      }
+      node = child
+    }
+
+    val shift = 8 - accumulatorBitCount
+    val start = code shl shift and 0xff
+    val end = 1 shl shift
+    for (i in start until start + end) {
+      node.children!![i] = terminal
+    }
```
JakeWharton:
```kotlin
node.children!!.fill(terminal, start, start + end)
```
Avoids a null check on each loop iteration

okhttp/src/main/java/okhttp3/internal/http2/Huffman.kt
swankjesse: Fixed!

okhttp/src/main/java/okhttp3/internal/http2/Huffman.kt
swankjesse: Fixed!

PR #5000, opened 2019-05-03
Convert HttpDate to Kotlin

shaishavgandhi05:


okhttp/src/main/java/okhttp3/internal/http/HttpDate.kt
```diff
@@ -13,40 +13,38 @@
...
-          DateFormat rfc1123 = new SimpleDateFormat("EEE, dd MMM yyyy HH:mm:ss 'GMT'", Locale.US);
-          rfc1123.setLenient(false);
-          rfc1123.setTimeZone(UTC);
-          return rfc1123;
-        }
-      };
+  private val STANDARD_DATE_FORMAT = object : ThreadLocal<DateFormat>() {
+    override fun initialValue(): DateFormat {
+      // Date format specified by RFC 7231 section 7.1.1.1.
+      val rfc1123 = SimpleDateFormat("EEE, dd MMM yyyy HH:mm:ss 'GMT'", Locale.US)
```
jemaystermind: Just throwing this out here. But I'm not sure what's more readable.
```suggestion
      val rfc1123 = SimpleDateFormat("EEE, dd MMM yyyy HH:mm:ss 'GMT'", Locale.US).apply {
          isLenient = false
          timeZone = UTC
      }
```

Egorand: Agree, it's a good use-case for `apply`

okhttp/src/main/java/okhttp3/internal/http/HttpDate.kt
```diff
@@ -65,55 +63,55 @@
...
-      return result;
+      return result
     }
-    synchronized (BROWSER_COMPATIBLE_DATE_FORMAT_STRINGS) {
-      for (int i = 0, count = BROWSER_COMPATIBLE_DATE_FORMAT_STRINGS.length; i < count; i++) {
-        DateFormat format = BROWSER_COMPATIBLE_DATE_FORMATS[i];
+    synchronized(BROWSER_COMPATIBLE_DATE_FORMAT_STRINGS) {
+      var i = 0
+      val count = BROWSER_COMPATIBLE_DATE_FORMAT_STRINGS.size
+      while (i < count) {
```
Egorand: Use `for (i in 0 until BROWSER_COMPATIBLE_DATE_FORMAT_STRINGS.size)`: the converter changed it to `while` as it got confused by the `count` variable declared inside `for`

okhttp/src/main/java/okhttp3/internal/http/HttpDate.kt
```diff
@@ -65,55 +63,55 @@
...
+        var format: DateFormat? = BROWSER_COMPATIBLE_DATE_FORMATS[i]
```
Egorand:
```kotlin
val format = BROWSER_COMPATIBLE_DATE_FORMATS[i] ?:
    SimpleDateFormat(BROWSER_COMPATIBLE_DATE_FORMAT_STRINGS[i], Locale.US).apply {
        timeZone = UTC
    }
```

okhttp/src/main/java/okhttp3/internal/http/HttpDate.kt
shaishavgandhi05: 6b9b76143563db1b3de49dd87348682afd891542

okhttp/src/main/java/okhttp3/internal/http/HttpDate.kt
shaishavgandhi05: 6b9b76143563db1b3de49dd87348682afd891542

okhttp/src/main/java/okhttp3/internal/http/HttpDate.kt
shaishavgandhi05: 6b9b76143563db1b3de49dd87348682afd891542

okhttp/src/main/java/okhttp3/internal/http/HttpDate.kt
```diff
@@ -65,55 +63,51 @@
       "EEE, dd-MM-yyyy HH:mm:ss z",
 
       /* RI bug 6641315 claims a cookie of this format was once served by www.yahoo.com */
-      "EEE MMM d yyyy HH:mm:ss z",
-  };
+      "EEE MMM d yyyy HH:mm:ss z")
```
BoxResin: According to [kotlin official guide](https://kotlinlang.org/docs/reference/coding-conventions.html#method-call-formatting), you should write like:

```kotlin
      "EEE MMM d yyyy HH:mm:ss z"
    )
```

okhttp/src/main/java/okhttp3/internal/http/HttpDate.kt
```diff
@@ -65,55 +63,51 @@
...
   }
 
-  private HttpDate() {
+  /** Returns the string for [value].  */
+  @JvmStatic
+  fun format(value: Date): String {
+    return STANDARD_DATE_FORMAT.get().format(value)
   }
-}
+}
```
BoxResin: No newline at end of file.

okhttp/src/main/java/okhttp3/internal/http/HttpDate.kt
```diff
@@ -65,55 +63,51 @@
...
-        if (position.getIndex() != 0) {
+    synchronized(BROWSER_COMPATIBLE_DATE_FORMAT_STRINGS) {
+      for (i in 0 until BROWSER_COMPATIBLE_DATE_FORMAT_STRINGS.size) {
+        val format: DateFormat = BROWSER_COMPATIBLE_DATE_FORMATS[i]
+            ?: SimpleDateFormat(BROWSER_COMPATIBLE_DATE_FORMAT_STRINGS[i], Locale.US).apply {
+              // Set the timezone to use when interpreting formats that don't have a timezone. GMT is
+              // specified by RFC 7231.
+              timeZone = UTC
+            }
+        BROWSER_COMPATIBLE_DATE_FORMATS[i] = format
```
swankjesse: I don’t like how this adds an assignment when the value hasn’t changed. Could you make it an if instead?

okhttp/src/main/java/okhttp3/internal/http/HttpDate.kt
swankjesse: You can disregard this; our style guide doesn’t care.

okhttp/src/main/java/okhttp3/internal/http/HttpDate.kt
shaishavgandhi05: Fixed in 2664c76ddba226c8a6fbf510bc9af4ee2606eada

Issue #4726, opened 2019-03-16
NPE Crash At java.net.Proxy.type() on 3.12.x


swankjesse:
Make sure this test passes:
https://github.com/square/okhttp/commit/77f3cec493e09a704301b18d577ca7220e46231e

Possibly borrowing these code changes:
https://github.com/square/okhttp/pull/4611/files

From this mainline issue:
https://github.com/square/okhttp/issues/3995

PR #5004, opened 2019-05-04
fix: StreamAllocation should keep route from reused connection (3.12.x branch)

swankjesse:


swankjesse:
For https://github.com/square/okhttp/issues/4726

swankjesse:
(Circle CI doesn’t exist on the 3.12.x branch.)

PR #5010, opened 2019-05-04
Convert RouteSelector to Kotlin

swankjesse:


okhttp/src/main/java/okhttp3/internal/connection/RouteSelector.kt
```diff
@@ -0,0 +1,196 @@
...
+  private val routeDatabase: RouteDatabase,
+  private val call: Call,
+  private val eventListener: EventListener
+) {
+  /* State for negotiating the next proxy to use. */
+  private var proxies = emptyList<Proxy>()
+  private var nextProxyIndex: Int = 0
+
+  /* State for negotiating the next socket address to use. */
+  private var inetSocketAddresses: List<InetSocketAddress> = emptyList()
```
JakeWharton:
```suggestion
  private var inetSocketAddresses = emptyList<InetSocketAddress>()
```
You're mixing with type / without type

PR #5011, opened 2019-05-04
Update RealConnection.java

DevResRod:
Re-create the connection when Proxy server closes the connection after an unauthorized request, to avoid IO issue with closed socket.

### DIFF
--- a/okhttp/src/main/java/com/squareup/okhttp/internal/io/RealConnection.java
+++ b/okhttp/src/main/java/com/squareup/okhttp/internal/io/RealConnection.java
@@ -16,29 +16,6 @@
  */
 package com.squareup.okhttp.internal.io;
 
-import static com.squareup.okhttp.internal.Util.closeQuietly;
-import static java.net.HttpURLConnection.HTTP_OK;
-import static java.net.HttpURLConnection.HTTP_PROXY_AUTH;
-import static java.util.concurrent.TimeUnit.MILLISECONDS;
-
-import java.io.IOException;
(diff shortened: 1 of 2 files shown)


swankjesse:
Please don’t combine behavior changes with refactorings. Ie. don’t extract methods.

Please include a test case that fails without the behavior change.

DevResRod:
I have reverted the method extraction (refactor) to ease the change tracking. Is it enought?
Also, added a new test method in CallTest.java that fails with the original code and succeeds with this commit.

okhttp/src/main/java/com/squareup/okhttp/internal/io/RealConnection.java
```diff
@@ -292,7 +293,16 @@ private void createTunnel(int readTimeout, int writeTimeout) throws IOException
         case HTTP_PROXY_AUTH:
           tunnelRequest = OkHeaders.processAuthHeader(
               route.getAddress().getAuthenticator(), response, route.getProxy());
-          if (tunnelRequest != null) continue;
+          if (tunnelRequest != null) {
+              String header = response.header("Connection");
+              if (header != null && header.toLowerCase().contains("close")) {
```
yschimke: .contains() seems a bit weak here, value is generally keep-alive or close.

yschimke:
Any plans to migrate this to kotlin?  Seems like a useful fix, but not essential (it should recover anyway) unless you suffer this daily with additional latency and timeouts.

PR #5013, opened 2019-05-04
Reproduce NoSuchElementException in RouteSelector.next() in a test

amirlivneh:
I was unable to write a test that reproduces https://github.com/square/okhttp/issues/4875 consistently, but this is the closest I got. When I run it repeatedly in IntelliJ, it fails about 1 out of 5 times.


swankjesse:
Great find.

PR #5019, opened 2019-05-07
Convert CacheStrategy to Kotlin

swankjesse:


okhttp/src/main/java/okhttp3/internal/cache/CacheStrategy.kt
```diff
@@ -13,314 +13,320 @@
...
+        return CacheStrategy(null, null)
       }
 
-      return candidate;
+      return candidate
     }
 
     /** Returns a strategy to use assuming the request can use the network. */
-    private CacheStrategy getCandidate() {
+    private fun computeCandidate(): CacheStrategy {
```
swankjesse: renamed from getCandidate() to computeCandidate() ’cause Kotlin really likes get to be for accessors.

PR #5023, opened 2019-05-07
Convert Exchange to Kotlin

swankjesse:


okhttp/src/main/java/okhttp3/internal/connection/ExchangeFinder.kt
```diff
@@ -0,0 +1,310 @@
...
+    var result: RealConnection? = null
+    var selectedRoute: Route? = null
+    var releasedConnection: RealConnection?
+    val toClose: Socket?
+    synchronized(connectionPool) {
+      if (transmitter.isCanceled) throw IOException("Canceled")
+      hasStreamFailure = false // This is a fresh attempt.
+
+      val previousRoute = if (retryCurrentRoute()) {
+        // Attempt to use an already-allocated connection. We need to be careful here because our
```
jemaystermind: Rest of the comment is missing and should be in line 163. 

See: https://github.com/square/okhttp/pull/5023/files#diff-a2fc2ec01d45df9d9a8a6621cbfb4032L147

okhttp/src/main/java/okhttp3/internal/connection/Transmitter.kt
```diff
@@ -0,0 +1,341 @@
...
+    return e as E
+  }
+
+  fun callStart() {
+    this.callStackTrace = Platform.get().getStackTraceForCloseable("response.body().close()")
+    eventListener.callStart(call)
+  }
+
+  /**
+   * Prepare to create a stream to carry `request`. This prefers to use the existing connection if
```
jemaystermind: Should be `[request]` instead of \`request\`?

okhttp/src/main/java/okhttp3/internal/connection/ExchangeFinder.kt
oldergod: Yeah Jesse you probably want
```
val previousRoute = if (retryCurrentRoute())  transmitter.connection!!.route() else null

// Attempt to use an already-allocated connection. We need to be careful here because our
// already-allocated connection may have been restricted from creating new exchanges.
releasedConnection = transmitter.connection
toClose = if (transmitter.connection?.noNewExchanges ?: false) {
      transmitter.releaseConnectionNoEvents()
} else {
      null
}
```

okhttp/src/main/java/okhttp3/internal/connection/ExchangeFinder.kt
```diff
@@ -13,264 +13,287 @@
...
+      eventListener.connectionReleased(call, releasedConnection!!)
     }
     if (foundPooledConnection) {
-      eventListener.connectionAcquired(call, result);
+      eventListener.connectionAcquired(call, result!!)
     }
     if (result != null) {
       // If we found an already-allocated or pooled connection, we're done.
-      return result;
+      return result!!
```
oldergod: probably don't need those bangers; might apply elsewhere as well

okhttp/src/main/java/okhttp3/internal/connection/Transmitter.kt
oldergod: nope

okhttp/src/main/java/okhttp3/internal/connection/Transmitter.kt
```diff
@@ -13,309 +13,300 @@
...
+      check(!noMoreExchanges)
+      exchange = null
     }
   }
 
   /**
-   * Releases resources held with the request or response of {@code exchange}. This should be called
-   * when the request completes normally or when it fails due to an exception, in which case {@code
-   * e} should be non-null.
+   * Releases resources held with the request or response of [exchange]. This should be called when
```
oldergod: you replaced all the `@code`s with `[]`, is it intentional?

okhttp/src/main/java/okhttp3/internal/connection/Transmitter.kt
```diff
@@ -13,309 +13,300 @@
...
+    return maybeReleaseConnection(e, false)
   }
 
   /**
    * Release the connection if it is no longer needed. This is called after each exchange completes
    * and after the call signals that no more exchanges are expected.
    *
-   * <p>If the transmitter was canceled or timed out, this will wrap {@code e} in an exception that
-   * provides that additional context. Otherwise {@code e} is returned as-is.
+   * If the transmitter was canceled or timed out, this will wrap [e] in an exception that provides
```
oldergod: ``` `e` ``` instead of `[e]`? same next line

okhttp/src/main/java/okhttp3/internal/connection/ExchangeFinder.kt
swankjesse: Nice catch. Thanks!

okhttp/src/main/java/okhttp3/internal/connection/Transmitter.kt
swankjesse: Fixed to use `[square braces]`. My rule is to use them whenever they’re supported.

okhttp/src/main/java/okhttp3/internal/connection/ExchangeFinder.kt
swankjesse: Sadly anything initialized in `synchronized` needs bangers.

okhttp/src/main/java/okhttp3/internal/connection/Transmitter.kt
swankjesse: Yep.

okhttp/src/main/java/okhttp3/internal/connection/Transmitter.kt
swankjesse: The benefit of square braces is it participates in refactoring if I rename the parameter.

swankjesse:
> I don't understand why you use generics for exception returning methods; is this required?

When the caller passes in a non-null IOException the return type is non-null. Otherwise it’s nullable. Generics seemed like the right solution.

PR #5027, opened 2019-05-10
Convert RealConnection to Kotlin

swankjesse:


okhttp/src/main/java/okhttp3/internal/connection/RealConnection.kt
```diff
@@ -0,0 +1,714 @@
...
+
+        // If this route hasn't completed a call, avoid it for new connections.
+        if (successCount == 0) {
+          if (e != null) {
+            connectionPool.connectFailed(route, e)
+          }
+          routeFailureCount++
+        }
+      }
+      return@synchronized // Keep synchronized {} happy.
```
swankjesse: Kotlin Kompiler Korner Kase

PR #5036, opened 2019-05-11
Convert FileOperator to Kotlin

shaishavgandhi05:


okhttp/src/main/java/okhttp3/internal/cache2/FileOperator.kt
```diff
@@ -0,0 +1,70 @@
...
+      mutableByteCount -= bytesWritten
+    }
+  }
+
+  /**
+   * Copy [byteCount] bytes from the file at [pos] into `source`. It is the
+   * caller's responsibility to make sure there are sufficient bytes to read: if there aren't this
+   * method throws an `EOFException`.
+   */
+  fun read(pos: Long, sink: Buffer, byteCount: Long) {
```
yschimke: Is it worth rewriting any of these in a functional kotlin style?

```
  tailrec fun read(pos: Long, sink: Buffer, byteCount: Long) {
    check(byteCount >= 0)

    if (byteCount > 0) {
      val bytesRead = fileChannel.transferTo(pos, byteCount, sink)
      read(pos + bytesRead, sink, byteCount - bytesRead)
    }
  }
```

cc @swankjesse

swankjesse: No, I don’t think so.

PR #5037, opened 2019-05-12
Convert Relay to Kotlin

swankjesse:


okhttp/src/main/java/okhttp3/internal/cache2/Relay.kt
```diff
@@ -0,0 +1,357 @@
...
+  val isClosed: Boolean
+    get() = file == null
+
+  @Throws(IOException::class)
+  private fun writeHeader(
+    prefix: ByteString,
+    upstreamSize: Long,
+    metadataSize: Long
+  ) {
+    val header = Buffer()
```
jemaystermind:
```suggestion
    val header = Buffer().apply{
      write(prefix)
      writeLong(upstreamSize)
      writeLong(metadataSize)
    }
```

okhttp/src/main/java/okhttp3/internal/cache2/Relay.kt
```diff
@@ -13,349 +13,345 @@
...
-          return -1L;
+          commit(upstreamPos)
+          return -1L
         }
 
         // Update this source and prepare this call's result.
-        long bytesRead = Math.min(upstreamBytesRead, byteCount);
-        upstreamBuffer.copyTo(sink, 0, bytesRead);
-        sourcePos += bytesRead;
+        val bytesRead = Math.min(upstreamBytesRead, byteCount)
```
yschimke: minOf

okhttp/src/main/java/okhttp3/internal/cache2/Relay.kt
swankjesse: Done.

okhttp/src/main/java/okhttp3/internal/cache2/Relay.kt
swankjesse: Fixed all.

PR #5038, opened 2019-05-12
Replace NamedRunnable.java with a threadName function

swankjesse:


mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.kt
```diff
@@ -331,55 +331,53 @@ class MockWebServer : ExternalResource(), Closeable {
...
     port = serverSocket!!.localPort
-    executor!!.execute(object : NamedRunnable("MockWebServer $port") {
-      override fun execute() {
-        try {
-          logger.info("${this@MockWebServer} starting to accept connections")
-          acceptConnections()
-        } catch (e: Throwable) {
-          logger.log(Level.WARNING, "${this@MockWebServer}  failed unexpectedly", e)
-        }
+    executor!!.execute("MockWebServer $port") {
```
swankjesse: GitHub doesn’t do this diff well. It’s a 17-line diff when you ignore whitespace.

JakeWharton: https://github.com/square/okhttp/pull/5038/files?w=1

okhttp/src/main/java/okhttp3/internal/UtilKt.kt
```diff
@@ -57,13 +57,23 @@ inline fun ignoreIoExceptions(block: () -> Unit) {
   }
 }
 
+inline fun threadName(name: String, block: () -> Unit) {
+  val oldName = Thread.currentThread().name
```
JakeWharton: Is `Thread.currentThread()` expensive? Could make a local

okhttp/src/main/java/okhttp3/RealCall.kt
```diff
@@ -115,7 +115,7 @@ internal class RealCall private constructor(
       assert(!Thread.holdsLock(client.dispatcher()))
       var success = false
       try {
-        executorService.execute(this)
+        executorService.execute("OkHttp ${redactedUrl()}", this@AsyncCall::run)
```
yschimke: This is breaking the test expectations of RecordingExecutor :(

yschimke: I was trying to be helpful but maybe overstepped a line on the PR.  No issue if you revert :)

okhttp/src/main/java/okhttp3/internal/UtilKt.kt
swankjesse: Fixed.

okhttp/src/main/java/okhttp3/RealCall.kt
swankjesse: yeah, I fixed

PR #5041, opened 2019-05-12
Convert Internal to Kotlin

yschimke:


okhttp/src/main/java/okhttp3/internal/Internal.kt
```diff
@@ -0,0 +1,34 @@
...
+
+import okhttp3.ConnectionPool
+import okhttp3.OkHttpClient
+import okhttp3.Response
+import okhttp3.internal.connection.Exchange
+import okhttp3.internal.connection.RealConnectionPool
+
+/**
+ * Escalate internal APIs in `okhttp3` so they can be used from OkHttp's implementation
+ * packages. The only implementation of this interface is in [OkHttpClient].
```
swankjesse: Maybe change these docs to `Escalate internal APIs in okhttp3 so they can be used by friend modules.

swankjesse: You beat me to it!

PR #5044, opened 2019-05-12
Adopt idiomatic Kotlin in Handshake

swankjesse:
I'm using DeprecationLevel.WARNING because I can't make the test
work otherwise:

https://stackoverflow.com/questions/56102107/suppress-deprecationlevel-error-in-kotlin

### DIFF
--- a/okhttp/src/main/java/okhttp3/Handshake.kt
+++ b/okhttp/src/main/java/okhttp3/Handshake.kt
@@ -24,50 +24,96 @@ import javax.net.ssl.SSLPeerUnverifiedException
 import javax.net.ssl.SSLSession
 
 /**
- * A record of a TLS handshake. For HTTPS clients, the client is *local* and the remote server
- * is its *peer*.
+ * A record of a TLS handshake. For HTTPS clients, the client is *local* and the remote server is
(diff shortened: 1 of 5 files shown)


okhttp/src/main/java/okhttp3/Handshake.kt
```diff
@@ -24,50 +24,96 @@ import javax.net.ssl.SSLPeerUnverifiedException
...
+ * handshakes.
  */
-data class Handshake private constructor(
-  private val tlsVersion: TlsVersion,
-  private val cipherSuite: CipherSuite,
-  private val peerCertificates: List<Certificate>,
-  private val localCertificates: List<Certificate>
-) {
-
+class Handshake private constructor(
```
oldergod: Could still have ben a `data class` no?

swankjesse: I think so. Biggest drawback of data class is that it’ll accept mutable lists, particularly from Java.

swankjesse: (I’m gonna punt on data class or not in public APIs for now.)

PR #5046, opened 2019-05-14
Adopt idiomatic Kotlin in CipherSuite, Cookie, and MediaType

swankjesse:


okhttp/src/main/java/okhttp3/ConnectionSpec.kt
```diff
@@ -198,7 +198,7 @@ class ConnectionSpec internal constructor(builder: Builder) {
 
     fun cipherSuites(vararg cipherSuites: CipherSuite): Builder = apply {
       require(tls) { "no cipher suites for cleartext connections" }
-      val strings = cipherSuites.map(CipherSuite::javaName).toTypedArray()
+      val strings = cipherSuites.map { it.javaName }.toTypedArray()
```
JakeWharton: Somewhat ironic given that you're trying to maintain source compatibility

swankjesse: yeah. Sigh

PR #5053, opened 2019-05-16
Adopt idiomatic Kotlin in Headers

swankjesse:


mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.kt
```diff
@@ -751,10 +751,10 @@ class MockWebServer : ExternalResource(), Closeable {
 
   @Throws(IOException::class)
   private fun writeHeaders(sink: BufferedSink, headers: Headers) {
-    for (i in 0 until headers.size()) {
-      sink.writeUtf8(headers.name(i))
+    for ((name, value) in headers) {
```
Egorand: Neat

okhttp/src/main/java/okhttp3/Headers.kt
```diff
@@ -122,6 +130,10 @@ class Headers private constructor(
     return result
   }
 
+  override operator fun iterator(): Iterator<Pair<String, String>> {
+    return Array(size) { name(it) to value(it) }.iterator()
```
JakeWharton: This is unfortunately heavy on allocation

swankjesse: yeah, and as a consequence I’m not using it internally as much as I’d prefer otherwise.

I think the alternative is an inline function, `forEachHeader()`, but I don’t think that’s as flexible.

swankjesse: One option this pushes us towards is changing the internal store from alternating names and values to an array of pairs. That could be fewer allocations if you iterate repeatedly.

PR #5058, opened 2019-05-17
Fix #5057: Remove unnecessary `isAndroidGetsocknameError` function

### LINKED ISSUE #5057: Remove isAndroidGetsocknameError
It isn't necessary for Android 5+.

MGaetan89:
I also had a quick look at `android.os.Build.VERSION.SDK_INT` usages, and I found the following two usages, but I don't know if they could be simplified or if they were left on purpose:

https://github.com/square/okhttp/blob/fbfb5f84612d416c5977f4dd312991f36d3bb0f8/okhttp/src/main/java/okhttp3/internal/platform/AndroidPlatform.kt#L56-L60

https://github.com/square/okhttp/blob/fbfb5f84612d416c5977f4dd312991f36d3bb0f8/okhttp/src/main/java/okhttp3/internal/platform/AndroidPlatform.kt#L377-L403

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/connection/RealConnection.kt
+++ b/okhttp/src/main/java/okhttp3/internal/connection/RealConnection.kt
@@ -385,8 +385,6 @@ class RealConnection(
       handshake = unverifiedHandshake
       protocol = if (maybeProtocol != null) Protocol.get(maybeProtocol) else Protocol.HTTP_1_1
       success = true
-    } catch (e: AssertionError) {
-      throw e
     } finally {
       if (sslSocket != null) {
         Platform.get().afterHandshake(sslSocket)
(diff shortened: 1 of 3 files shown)


okhttp/src/main/java/okhttp3/internal/connection/RealConnection.kt
```diff
@@ -387,7 +386,6 @@ class RealConnection(
       protocol = if (maybeProtocol != null) Protocol.get(maybeProtocol) else Protocol.HTTP_1_1
       success = true
     } catch (e: AssertionError) {
```
swankjesse: This entire clause is now unnecessary.

okhttp/src/main/java/okhttp3/internal/platform/AndroidPlatform.kt
```diff
@@ -82,7 +81,6 @@ class AndroidPlatform(
     try {
       socket.connect(address, connectTimeout)
     } catch (e: AssertionError) {
-      if (isAndroidGetsocknameError(e)) throw IOException(e)
```
swankjesse: Unnecessary catch clause?

okhttp/src/main/java/okhttp3/internal/connection/RealConnection.kt
MGaetan89: Indeed. I updated the PR

PR #5059, opened 2019-05-18
Fix/Simplify regex usage

yschimke:


yschimke:
Two cases not touched

okhttp/src/main/java/okhttp3/Cookie.kt - reuses matcher
okhttp/src/main/java/okhttp3/MediaType.kt - uses lookingAt instead of find or matchesEntire

swankjesse:
Nice!

okcurl/src/main/java/okhttp3/curl/Main.kt
```diff
@@ -205,7 +205,7 @@ class Main : Runnable {
   private fun mediaType(): MediaType? {
     val mimeType = headers?.let {
       for (header in it) {
-        val parts = header.split(":".toRegex()).toTypedArray()
+        val parts = header.split(':')
```
JakeWharton: This seems like it should have a limit as well

swankjesse: https://github.com/square/okhttp/pull/5073

PR #5060, opened 2019-05-18
Intellij cleanup

yschimke:
Doing a pass over with Intellij to cleanup and confirm our approach to quickfixes.

I expect most questionable is lifting return out of try etc, which I love.  Everything is an expression is a big win for me.


yschimke:
Can you explain the last sentence?  If you meant to say internals, then yep, I can revert.  otherwise I don't follow.

swankjesse:
In Java we use package-private to avoid synthetics. It's a small increase in visibility so it's worth the very small runtime benefit.

Kotlin’s next narrowest visibility is internal, which is a large increase in visibility for a very small runtime benefit.

Therefore I don't mind private for Kotlin even though we'll get synthetics in the bytecode.

PR #5061, opened 2019-05-18
Use startsWith instead of regionMatches

yschimke:


okhttp/src/main/java/okhttp3/CertificatePinner.kt
```diff
@@ -237,8 +237,7 @@ data class CertificatePinner internal constructor(
       if (pattern.startsWith(WILDCARD)) {
         val firstDot = hostname.indexOf('.')
         return hostname.length - firstDot - 1 == canonicalHostname.length &&
-            hostname.regionMatches(firstDot + 1, canonicalHostname, 0, canonicalHostname.length,
-            ignoreCase = false)
+            hostname.startsWith(canonicalHostname, ignoreCase = false, startIndex = firstDot + 1)
```
JakeWharton: This is the default value for `ignoreCase`. No need to supply.

yschimke: Yep. Good catch

swankjesse: https://github.com/square/okhttp/pull/5073

PR #5062, opened 2019-05-18
Build in Travis CI only on master

yschimke:
Build only when we actually want a snapshot built as part of .buildscript/deploy_snapshot.sh


yschimke:
This may not have worked...

image

yschimke:
Also disabled here https://travis-ci.com/square/okhttp/settings, can reenable if we don't want to land this.

image

cc @swankjesse

PR #5067, opened 2019-05-18
RFC: PlatformRegistry and more predictable running mode

yschimke:
By observation we are generally only running with Platform (no alpn) in IDE mode.  Gradle config doesn't get the right things onto bootclasspath.

Changing the default to Conscrypt will give us better coverage in IDE mode with JDK 8.  With JDK 9 we should be using the Jdk9Platform.

TODO confirm run modes:  

- `./gradlew -Pokhttp.platform=jdk9 test` -> Jdk9Platform
- `./gradlew -Pokhttp.platform=conscrypt test` -> ConscryptPlatform
- `./gradlew -Pokhttp.platform=jdk8 test` -> Platform
- `./gradlew -Pokhttp.platform=jdk8alpn test` -> Jdk8WithJettyBootPlatform
- `./gradlew test`with JDK 9+ -> Jdk9Platform
- `./gradlew test`with JDK 8 -> ConscryptPlatform
- Intellij with JDK 9+ -> Jdk9Platform
- Intellij with JDK 8 -> ConscryptPlatform

### DIFF
--- a/okhttp-testing-support/src/main/java/okhttp3/PlatformRule.kt
+++ b/okhttp-testing-support/src/main/java/okhttp3/PlatformRule.kt
@@ -15,7 +15,10 @@
  */
 package okhttp3
 
+import okhttp3.internal.platform.Jdk8WithJettyBootPlatform.Companion.isAlpnBootEnabled
+import okhttp3.internal.platform.Jdk9Platform
 import okhttp3.internal.platform.Platform
+import okhttp3.internal.platform.PlatformRegistry
 import org.conscrypt.Conscrypt
 import org.hamcrest.CoreMatchers.equalTo
 import org.hamcrest.CoreMatchers.not
@@ -39,7 +42,7 @@ open class PlatformRule @JvmOverloads constructor(
       assumeThat(getPlatformSystemProperty(), equalTo(requiredPlatformName))
     }
 
-    Platform.resetForTests()
+    PlatformRegistry.resetForTests()
   }
 
   fun isConscrypt() = getPlatformSystemProperty() == CONSCRYPT_PROPERTY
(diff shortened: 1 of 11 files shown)


yschimke:
I'm expecting this to be too big for landing in 4.0, but was getting frustrated at weird class loading cycles between Platforms while debugging etc.  Trying to simplify based on how we would write this in a kotlin world.

okhttp-testing-support/src/main/java/okhttp3/PlatformRule.kt
```diff
@@ -144,22 +133,16 @@ open class PlatformRule @JvmOverloads constructor(
...
     }
 
     @JvmStatic
     fun getPlatformSystemProperty(): String {
-      var property: String? = System.getProperty(PROPERTY_NAME)
-
-      if (property == null) {
-        property = when (Platform.get()) {
-          is ConscryptPlatform -> CONSCRYPT_PROPERTY
-          is Jdk8WithJettyBootPlatform -> CONSCRYPT_PROPERTY
```
yschimke: This was a bug

swankjesse: Yikes

okhttp/src/main/java/okhttp3/internal/Util.kt
```diff
@@ -505,4 +506,43 @@ object Util {
...
+  @JvmStatic
+  fun <T> readFieldOrNull(instance: Any, fieldType: Class<T>, fieldName: String): T? {
+    var c: Class<*> = instance.javaClass
+    while (c != Any::class.java) {
+      try {
+        val field = c.getDeclaredField(fieldName)
+        field.isAccessible = true
+        val value = field.get(instance)
+        return if (!fieldType.isInstance(value)) null else fieldType.cast(value)
+      } catch (ignored: NoSuchFieldException) {
```
swankjesse: nit: `_`

okhttp/src/main/java/okhttp3/internal/platform/Jdk8WithJettyBootPlatform.kt
```diff
@@ -123,15 +122,17 @@ class Jdk8WithJettyBootPlatform(
   }
 
   companion object {
+    fun isAlpnBootEnabled(): Boolean = try {
+      Class.forName("org.eclipse.jetty.alpn.ALPN", true, null)
+      true
+    } catch (cnfe: ClassNotFoundException) {
```
swankjesse: `_`

okhttp/src/main/java/okhttp3/internal/platform/Jdk9Platform.kt
```diff
@@ -75,16 +74,27 @@ class Jdk9Platform(
...
+          else -> null
         }
+
+    fun isAvailable(): Boolean {
+      val jvmVersion = System.getProperty("java.specification.version", "unknown")
+      return try {
+        // 1.8, 9, 10, 11, 12 etc
+        val version = jvmVersion.toInt()
+        version >= 9
+      } catch (nfe: NumberFormatException) {
```
swankjesse: `_`

okhttp-testing-support/src/main/java/okhttp3/PlatformRule.kt
yschimke: In tests and likely never hit in practice

PR #5068, opened 2019-05-19
Lock in a route in hasRouteToTry() (3.14.x branch)

swankjesse:


okhttp-tests/src/test/java/okhttp3/internal/http2/HttpOverHttp2Test.java
```diff
@@ -1500,4 +1502,39 @@ private Buffer gzip(String bytes) throws IOException {
       }
     }
   }
+
+  /** https://github.com/square/okhttp/issues/4875 */
```
swankjesse: @amirlivneh FYI, this is your commit on my PR

swankjesse: Also – I got the test to pass 100 times in a row, giving me confidence I got the fix right.

amirlivneh: Yes, it does look promising. When I run the test locally with the fix, the test now times out one out of ~100 runs because `Callback.onFailure()` is called once and not twice as expected. When I remove the test timeout, the test sometimes hangs forever, which means one of the calls never fails and never times out. This looks like an unrelated yet real issue.

PR #5069, opened 2019-05-19
Adopt idiomatic Kotlin in CacheControl

TangHuaiZhe:


okhttp/src/main/java/okhttp3/CacheControl.kt
```diff
@@ -344,7 +405,7 @@ class CacheControl private constructor(
      */
     private fun String.indexOfElement(characters: String, startIndex: Int = 0): Int {
```
yschimke: TIOLI: We could drop this (at cost of storing immutable list as a const instead of strings) and replace with

    "asb".indexOfAny("abc".toList(), startIndex = 3)

okhttp/src/main/java/okhttp3/internal/cache/CacheStrategy.kt
```diff
@@ -167,21 +167,21 @@ class CacheStrategy internal constructor(
       val ageMillis = cacheResponseAge()
       var freshMillis = computeFreshnessLifetime()
 
-      if (requestCaching.maxAgeSeconds() != -1) {
-        freshMillis = minOf(freshMillis, SECONDS.toMillis(requestCaching.maxAgeSeconds().toLong()))
+      if (requestCaching.maxAgeSeconds != -1) {
+        freshMillis = minOf(freshMillis, SECONDS.toMillis(requestCaching.maxAgeSeconds.toLong()))
```
yschimke: I suggest an extension function `TimeDuration.toMillis(Int)` in UtilKt to avoid the toLong operations here.

okhttp/src/main/java/okhttp3/CacheControl.kt
swankjesse: I think I prefer what we have here. Fewer allocations!

PR #5071, opened 2019-05-20
Replace Math.min() with minOf

TangHuaiZhe:
- Replace `Math.min()` with `minOf`
- Use `0L` replace `0` when compared with a long var

### DIFF
--- a/mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.kt
+++ b/mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.kt
@@ -92,7 +92,6 @@ import javax.net.ssl.SSLSocket
 import javax.net.ssl.SSLSocketFactory
 import javax.net.ssl.TrustManager
 import javax.net.ssl.X509TrustManager
-import kotlin.math.min
 
 /**
  * A scriptable web server. Callers supply canned responses and the server replays them upon request
@@ -659,7 +658,7 @@ class MockWebServer : ExternalResource(), Closeable {
     val chunkSizes = ArrayList<Int>()
     val policy = dispatcher.peek()
     if (contentLength != -1L) {
-      hasBody = contentLength > 0
+      hasBody = contentLength > 0L
       throttledTransfer(policy, socket, source, requestBody.buffer(), contentLength, true)
     } else if (chunked) {
       hasBody = true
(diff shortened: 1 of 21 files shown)


okhttp/src/main/java/okhttp3/internal/http/CallServerInterceptor.kt
```diff
@@ -116,7 +116,7 @@ class CallServerInterceptor(private val forWebSocket: Boolean) : Interceptor {
         "close".equals(response.header("Connection"), ignoreCase = true)) {
       exchange.noNewExchangesOnConnection()
     }
-    if ((code == 204 || code == 205) && response.body()?.contentLength() ?: -1 > 0) {
+    if ((code == 204 || code == 205) && response.body()?.contentLength() ?: -1 > 0L) {
```
yschimke: -1L?

okhttp/src/main/java/okhttp3/internal/http2/Http2Writer.kt
```diff
@@ -100,7 +100,7 @@ class Http2Writer(
     hpackWriter.writeHeaders(requestHeaders)
 
     val byteCount = hpackBuffer.size
-    val length = Math.min((maxFrameSize - 4).toLong(), byteCount).toInt()
+    val length = minOf((maxFrameSize - 4).toLong(), byteCount).toInt()
```
yschimke: the toLong seems weird here given there is toInt, is there better way to express this?

TangHuaiZhe: How about` maxFrameSize - 4L`?

okhttp/src/main/java/okhttp3/internal/http/CallServerInterceptor.kt
TangHuaiZhe: Fixed.

mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.kt
```diff
@@ -659,7 +658,7 @@ class MockWebServer : ExternalResource(), Closeable {
     val chunkSizes = ArrayList<Int>()
     val policy = dispatcher.peek()
     if (contentLength != -1L) {
-      hasBody = contentLength > 0
+      hasBody = contentLength > 0L
```
swankjesse: thanks!

okhttp/src/main/java/okhttp3/internal/http2/Http2Writer.kt
```diff
@@ -100,7 +100,7 @@ class Http2Writer(
     hpackWriter.writeHeaders(requestHeaders)
 
     val byteCount = hpackBuffer.size
-    val length = Math.min((maxFrameSize - 4).toLong(), byteCount).toInt()
+    val length = minOf(maxFrameSize - 4L, byteCount).toInt()
```
swankjesse: nice!

PR #5077, opened 2019-05-20
Use more Kotlin in HttpUrl

swankjesse:


okhttp/src/main/java/okhttp3/HttpUrl.kt
```diff
@@ -965,53 +962,49 @@ class HttpUrl internal constructor(builder: Builder) {
...
-      this.encodedQueryNamesAndValues = if (query != null) {
-        queryStringToNamesAndValues(canonicalize(query, QUERY_ENCODE_SET, alreadyEncoded = false,
-                strict = false, plusIsSpace = true, asciiOnly = true))
-      } else {
-        null
-      }
+      this.encodedQueryNamesAndValues = query?.canonicalize(
+          encodeSet = QUERY_ENCODE_SET,
+          plusIsSpace = true
+      )?.toQueryNamesAndValues()
```
swankjesse: This is a nice example of how extension functions and the safe call operator work together.

JakeWharton: Yeah we traditionally have a lot of utils that accept null and early return that we can start removing. Close quietly comes to mind...

okhttp/src/main/java/okhttp3/HttpUrl.kt
```diff
@@ -1528,25 +1541,24 @@ class HttpUrl internal constructor(builder: Builder) {
...
-     * Cuts `encodedQuery` up into alternating parameter names and values. This divides a query
-     * string like `subject=math&easy&problem=5-2=3` into the list `["subject", "math", "easy",
-     * null, "problem", "5-2=3"]`. Note that values may be null and may contain '=' characters.
+     * Cuts [this] up into alternating parameter names and values. This divides a query string like
+     * `subject=math&easy&problem=5-2=3` into the list `["subject", "math", "easy", null, "problem",
+     * "5-2=3"]`. Note that values may be null and may contain '=' characters.
      */
-    @JvmStatic
-    internal fun queryStringToNamesAndValues(encodedQuery: String): MutableList<String?> {
+    internal fun String.toQueryNamesAndValues(): MutableList<String?> {
```
JakeWharton: Does return need to stay mutable?

swankjesse: It does. This is for a weird API where you provide the entire unparsed query, and then potentially build upon it.

PR #5078, opened 2019-05-20
Don't use @JvmStatic for internal calls

swankjesse:


okhttp/src/test/java/okhttp3/internal/http/HttpDateTest.kt
```diff
@@ -13,80 +13,84 @@
...
 
-public class HttpDateTest {
-
-  private TimeZone originalDefault;
+  private lateinit var originalDefault: TimeZone
 
   @Before
-  public void setUp() throws Exception {
-    originalDefault = TimeZone.getDefault();
+  @Throws(Exception::class)
```
JakeWharton: nit: could be removed from all functions in this file

PR #5079, opened 2019-05-20
Kotlin improvements

JakeWharton:
Around string index/substring/contains and collection contains mostly.


okhttp/src/main/java/okhttp3/Cache.kt
```diff
@@ -752,11 +752,7 @@ class Cache internal constructor(
     }
 
     /** Returns true if a Vary header contains an asterisk. Such responses cannot be cached. */
-    fun Response.hasVaryAll(): Boolean {
-      val responseHeaders = headers()
-      val varyFields = responseHeaders.varyFields()
-      return varyFields.contains("*")
-    }
+    fun Response.hasVaryAll() = "*" in headers().varyFields()
```
swankjesse: golf hole-in-one

PR #5080, opened 2019-05-20
Eliminate Util object in favor of top-level functions

JakeWharton:


JakeWharton:
Have a bunch of commits migrating these to extension functions next

PR #5082, opened 2019-05-20
Use apply {} in builders

swankjesse:


yschimke:
Nice change

PR #5085, opened 2019-05-21
Adopt HttpUrl's new Kotlin API

swankjesse:


swankjesse:
All via fancy deprecation.

PR #5087, opened 2019-05-21
Smattering of Kotlin part 2

JakeWharton:
Again, squash or not. Shouldn't matter.

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/Util.kt
+++ b/okhttp/src/main/java/okhttp3/internal/Util.kt
@@ -289,5 +289,4 @@ fun HttpUrl.canReuseConnectionFor(other: HttpUrl): Boolean = host == other.host 
     port == other.port &&
     scheme == other.scheme
 
-fun eventListenerFactory(listener: EventListener): EventListener.Factory =
-    EventListener.Factory { listener }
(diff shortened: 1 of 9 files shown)


okhttp/src/main/java/okhttp3/internal/Util.kt
```diff
@@ -138,14 +136,14 @@ fun nonEmptyIntersection(
   return false
 }
 
-fun hostHeader(url: HttpUrl, includeDefaultPort: Boolean): String {
-  val host = if (":" in url.host) {
-    "[${url.host}]"
+fun HttpUrl.toHostHeader(includeDefaultPort: Boolean = false): String {
```
swankjesse: Tempted to drop the `to` prefix and just call this` hostHeader()`. I like `to` to be more or less reversible. (Or did I pattern match to far?)

JakeWharton: This is reversible, I just didn't do the other one yet. It's directly above this function. Let me convert it too in the same commit.

JakeWharton: Done

PR #5088, opened 2019-05-21
Brotli Interceptor

yschimke:


okhttp-brotli/README.md
```diff
@@ -0,0 +1,5 @@
+OkHttp DNS over HTTPS Implementation
```
yschimke: TODO

okhttp-brotli/src/test/java/okhttp3/brotli/BrotliInterceptorTest.kt
```diff
@@ -0,0 +1,92 @@
...
+    val uncompressed = BrotliInterceptor.uncompress(response)
+
+    val responseString = uncompressed.body()?.string()
+    assertThat(responseString).contains("\"brotli\": true,")
+    assertThat(responseString).contains("\"Accept-Encoding\": \"br\"")
+  }
+
+  @Test
+  fun testNoUncompress() {
+    val response = response("https://httpbin.org/brotli", ByteString.of(*"XXXX".toByteArray()))
```
yschimke: whats the solution for this awkwardness?

yschimke:
95kb Jar

http://central.maven.org/maven2/org/brotli/dec/0.1.2/

okhttp-brotli/src/test/java/okhttp3/brotli/BrotliInterceptorTest.kt
NightlyNexus: `"XXXX"..encodeUtf8()`?

okhttp-brotli/src/test/java/okhttp3/brotli/BrotliInterceptorTest.kt
```diff
@@ -0,0 +1,92 @@
...
+            "5cb384cd0912aeb5b5dd9555f2dd1a9b20688201"
+
+    val response = response("https://httpbin.org/brotli", s.decodeHex()) {
+      header("Content-Encoding", "br")
+    }
+
+    val uncompressed = BrotliInterceptor.uncompress(response)
+
+    val responseString = uncompressed.body()?.string()
+    assertThat(responseString).contains("\"brotli\": true,")
```
swankjesse: Nit: kind of weird for the body contents to look like headers

okhttp-brotli/src/test/java/okhttp3/brotli/BrotliInterceptorTest.kt
```diff
@@ -0,0 +1,92 @@
...
+    bodyHex: ByteString,
+    fn: Response.Builder.() -> Unit = {}
+  ): Response {
+    return Response.Builder()
+        .body(ResponseBody.create(MediaType.get("text/plain"), bodyHex))
+        .code(200)
+        .message("OK")
+        .request(Request.Builder().url(url).build())
+        .protocol(Protocol.HTTP_2)
+        .apply(fn)
```
swankjesse: Cute

okhttp-brotli/src/test/java/okhttp3/brotli/BrotliInterceptorTest.kt
yschimke: That's httpbin for you

okhttp-brotli/src/test/java/okhttp3/brotli/BrotliInterceptorTest.kt
yschimke: appears broken, but works manually.

okhttp-brotli/src/main/java/okhttp3/brotli/package-info.kt
```diff
@@ -0,0 +1,2 @@
+/** A Brotli compression interceptor. */
+package okhttp3.brotli
```
yschimke: Should I delete?

okhttp-brotli/src/main/java/okhttp3/brotli/BrotliInterceptor.kt
```diff
@@ -0,0 +1,52 @@
...
+import org.brotli.dec.BrotliInputStream
+
+/**
+ * Transparent Brotli response support.
+ *
+ * Adds Accept-Encoding: br to existing encodings and checks (and strips) for Content-Encoding: br in responses
+ */
+object BrotliInterceptor : Interceptor {
+  override fun intercept(chain: Interceptor.Chain): Response {
+    val request = chain.request().newBuilder().header("Accept-Encoding", "br").build()
```
krombel: I would like okhttp to let me extend the "Accept-Encoding" header by "br" and not replace it completely to keep e.g. gzip compression in case the webserver does not support brotli yet.
Would that be possible?

yschimke: Let me confirm with a test. The intention was to add a new header value, not remove all others.

MatushkaRossiya:
Why was this closed? Is it abandoned?

yschimke:
My mistake.

MatushkaRossiya:
Ah, sorry then, thanks for clarifying.

okhttp-brotli/src/main/java/okhttp3/brotli/BrotliInterceptor.kt
yschimke: @krombel thanks, good catch

yschimke: @swankjesse whats the right handling here?  add gzip also which BridgeInterceptor would otherwise?  Or check and only add if the client didn't define?  But still process if response is brotli?

swankjesse:
Probably the simplest thing to do is for this to support all compression mechanisms. Otherwise there’s an implicit precedence thing and I don’t love that.

yschimke:
Is okhttp-brotli the right module name still?

okhttp-brotli/src/main/java/okhttp3/brotli/BrotliInterceptor.kt
```diff
@@ -0,0 +1,64 @@
...
+
+  internal fun uncompress(response: Response): Response {
+    val body = response.body ?: return response
+    val encoding = response.header("Content-Encoding") ?: return response
+
+    val decompressedSource = when {
+      encoding.equals("br", ignoreCase = true) ->
+        BrotliInputStream(body.source().inputStream()).source().buffer()
+      encoding.equals("gzip", ignoreCase = true) ->
+        GzipSource(body.source()).buffer()
```
krombel: I am not sure as I don't know the implementation in detail but:
Shouldn't gzip be handled in another interceptor in the chain?
It is currently supported by default so should be enchained already
(Or do I understand sth. wrong?)

yschimke: It’s handled transparently IFF you don’t set your own headers.

yschimke:
@swankjesse Landing based on "swankjesse approved these changes on 22 May"

okhttp-brotli/src/main/java/okhttp3/brotli/BrotliInterceptor.kt
```diff
@@ -0,0 +1,64 @@
...
+import okio.GzipSource
+import okio.buffer
+import okio.source
+import org.brotli.dec.BrotliInputStream
+
+/**
+ * Transparent Brotli response support.
+ *
+ * Adds Accept-Encoding: br to request and checks (and strips) for Content-Encoding: br in
+ * responses.  n.b. this replaces the transparent gzip compression InBridgeInterceptor.
```
swankjesse: InBridgeInterceptor –> in BridgeInterceptor

okhttp-brotli/src/main/java/okhttp3/brotli/BrotliInterceptor.kt
```diff
@@ -0,0 +1,64 @@
...
+        chain.proceed(chain.request())
+      }
+
+  internal fun uncompress(response: Response): Response {
+    val body = response.body ?: return response
+    val encoding = response.header("Content-Encoding") ?: return response
+
+    val decompressedSource = when {
+      encoding.equals("br", ignoreCase = true) ->
+        BrotliInputStream(body.source().inputStream()).source().buffer()
```
swankjesse: It’s too bad this stuff is package-private:
https://github.com/google/brotli/blob/master/java/org/brotli/dec/Decode.java
https://github.com/google/brotli/blob/master/java/org/brotli/dec/State.java

If it weren’t we might be able to save some copies, especially with Buffer’s UnsafeCursor.

okhttp-brotli/src/test/java/okhttp3/brotli/BrotliInterceptorTest.kt
```diff
@@ -0,0 +1,92 @@
...
+import okhttp3.ResponseBody.Companion.toResponseBody
+import okio.ByteString
+import okio.ByteString.Companion.decodeHex
+import okio.ByteString.Companion.encodeUtf8
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.Assert.fail
+import org.junit.Test
+import java.io.IOException
+
+class BrotliInterceptorTest {
```
swankjesse: Sweet.

yschimke:
I’ll address your last comments and also test a snapshot build on device against google.

okhttp-brotli/src/main/java/okhttp3/brotli/BrotliInterceptor.kt
yschimke: Maybe worth a discussion, for a compression library the efficiency seems like a core feature.

blundell:
Just to confirm, to use Brotli I need to add it as an interceptor? OkHttp doesn't use it be default (like it does with GZIP)?

blundell:
Found it here "Install the interceptor to introduce Botli compression" https://square.github.io/okhttp/changelog/#version-410

PR #5089, opened 2019-05-22
Replace statics with extension functions to create HttpUrls

swankjesse:


okhttp/src/main/java/okhttp3/HttpUrl.kt
```diff
@@ -1130,8 +1130,7 @@ class HttpUrl internal constructor(
     }
 
     private fun removeAllCanonicalQueryParameters(canonicalName: String) {
-      var i = encodedQueryNamesAndValues!!.size - 2
-      while (i >= 0) {
+      for (i in encodedQueryNamesAndValues!!.size - 2 downTo 0 step 2) {
```
Egorand: Nice

PR #5091, opened 2019-05-22
Refactor Array utilities to extensions

JakeWharton:


JakeWharton:
GitHub notifications are delayed so CI won't kick in for a while...

okhttp/src/main/java/okhttp3/internal/Util.kt
```diff
@@ -92,17 +92,16 @@ fun threadFactory(
...
+ * Returns an array containing only elements found in [this] and also in [other].
+ * The returned elements are in the same order as in [this].
  */
-fun intersect(
-  comparator: Comparator<in String>,
-  first: Array<String>,
-  second: Array<String>
+fun Array<String>.intersect(
+  other: Array<String>,
+  comparator: Comparator<in String>
```
yschimke: naturalOrder is a compelling default for these.

okhttp/src/main/java/okhttp3/internal/Util.kt
```diff
@@ -113,21 +112,20 @@ fun intersect(
...
  * so small, and the likelihood of an intersection so great, that it is not worth the CPU cost of
  * sorting or the memory cost of hashing.
  */
-fun nonEmptyIntersection(
-  comparator: Comparator<String>,
-  first: Array<String>?,
-  second: Array<String>?
+fun Array<String>.hasIntersection(
+  other: Array<String>?,
+  comparator: Comparator<in String>
```
yschimke: naturalOrder is a compelling default for these.

JakeWharton: Yeah I was debating whether to do that or not. They're actually only used in one place where a custom comparator is always specified so I elected not to.

PR #5092, opened 2019-05-22
Use idiomatic Kotlin in Request.kt

monkey-mas:
- define `@get:JvmName(...)` for the following vals in constructor instead of passing `builder: Builder`.
  - `url: HttpUrl`
  - `method: String`
  - `headers: Headers`
  - `body: RequestBody?`

  - `fun url(): HttpUrl`
  - `fun method(): String`
  - `fun headers(): Headers`
  - `fun body(): RequestBody?`
  - `fun cacheControl(): CacheControl`

### DIFF
--- a/okhttp/src/main/java/okhttp3/Request.kt
+++ b/okhttp/src/main/java/okhttp3/Request.kt
@@ -26,31 +26,21 @@ import java.net.URL
  * immutable.
  */
 class Request internal constructor(
-  internal val url: HttpUrl,
-  builder: Builder
+  @get:JvmName("url") val url: HttpUrl,
+  @get:JvmName("method") val method: String,
+  @get:JvmName("headers") val headers: Headers,
+  @get:JvmName("body") val body: RequestBody?,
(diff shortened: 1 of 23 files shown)


okhttp/src/main/java/okhttp3/Request.kt
```diff
@@ -74,11 +64,49 @@ class Request internal constructor(
...
+      level = DeprecationLevel.WARNING)
+  fun method(): String = method
+
+  @JvmName("-deprecated_headers")
+  @Deprecated(
+      message = "moved to val",
+      replaceWith = ReplaceWith(expression = "headers"),
+      level = DeprecationLevel.WARNING)
+  fun headers(): Headers = headers
+  @JvmName("-deprecated_body")
```
JakeWharton: super nit: should have empty line above this

monkey-mas: Thanks! fixed it! :)

PR #5096, opened 2019-05-23
Adopt idiomatic Kotlin in Cache

swankjesse:
Also remove the obsolete InternalCache abstraction. That was used by AOSP.


okhttp/src/main/java/okhttp3/OkHttpClient.kt
```diff
@@ -254,13 +252,9 @@ open class OkHttpClient internal constructor(
...
   fun pingIntervalMillis(): Int = pingInterval
 
-  internal fun internalCache(): InternalCache? {
-    return cache?.internalCache ?: internalCache
-  }
-
   /** Prepares the [request] to be executed at some point in the future. */
   override fun newCall(request: Request): Call {
-    return RealCall.newRealCall(this, request, false /* for web socket */)
+    return RealCall.newRealCall(this, request, forWebSocket = false)
```
yschimke: Nice!  We should do this for all ints and bools.

jemaystermind: @yschimke Is someone working on it? I'd like to do it.

swankjesse: @jemaystermind it's easy to screw this up because the names are noisy if the context makes their purposes clear. Starting with parameters we're already commenting would be a good start though.

jemaystermind: @swankjesse Good point. I'll take your advice and see what I can do. Thanks!

PR #5101, opened 2019-05-23
Use kdoc parameter link for CacheInterceptor

jemaystermind:


okhttp/src/main/java/okhttp3/internal/cache/CacheInterceptor.kt
```diff
@@ -239,7 +239,7 @@ class CacheInterceptor(internal val cache: Cache?) : Interceptor {
     }
 
     /**
-     * Returns true if `fieldName` is an end-to-end HTTP header, as defined by RFC 2616,
+     * Returns true if [fieldName] is an end-to-end HTTP header, as defined by RFC 2616,
```
jemaystermind: Would adding the RFC link here be an unnecessary noise?

swankjesse: Yep.

swankjesse: It's not a public API so we don't need to be as formal with the docs.

PR #5104, opened 2019-05-23
Remove obsolete API version check

JakeWharton:


okhttp/src/main/java/okhttp3/internal/platform/AndroidPlatform.kt
```diff
@@ -48,23 +48,7 @@ class AndroidPlatform(
 ) : Platform() {
   private val closeGuard = CloseGuard.get()
 
-  // Not a real Android runtime; probably RoboVM or MoE
-  // Try to load TLS 1.2 explicitly.
-  // fallback to TLS
   override fun newSSLContext(): SSLContext {
```
yschimke: Do we need this override anymore?

JakeWharton: Nope! Removed.

PR #5105, opened 2019-05-23
Convert Route to expose properties

JakeWharton:


okhttp/src/test/java/okhttp3/KotlinSourceCompatibilityTest.kt
```diff
@@ -1122,9 +1122,9 @@ class KotlinSourceCompatibilityTest {
   @Test @Ignore
   fun route() {
     val route: Route = newRoute()
-    val address: Address = route.address()
-    val proxy: Proxy = route.proxy()
-    val inetSocketAddress: InetSocketAddress = route.socketAddress()
+    val address: Address = route.address
```
swankjesse: Nit: we've been keeping these original so our deprecated methods get exercise.

JakeWharton: Reverted

PR #5110, opened 2019-05-24
Idiomatic Kotlin for Response.kt

monkey-mas:
- define `@get:JvmName(...)` for the following vals in constructor instead of passing `builder: Builder`.
  - `request: Request`
  - `protocol: Protocol`
  - `message: String`
  - `code: Int`
  - `handshake: Handshake?`
  - `headers: Headers`
  - `body: ResponseBody?`
  - `networkResponse: Response?`
  - `cacheResponse: Response?`
  - `priorResponse: Response?`
  - `sentRequestAtMillis: Long`
  - `receivedResponseAtMillis: Long`
  - `exchange: Exchange?`

  - `fun request(): Request`
  - `fun protocol(): Protocol`
  - `fun message(): String`
  - `fun code(): Int`
  - `fun handshake(): Handshake?`
  - `fun headers(): Headers`
  - `fun body(): ResponseBody?`
  - `fun networkResponse(): Response?`
  - `fun cacheResponse(): Response?`
  - `fun priorResponse(): Response?`
  - `fun sentRequestAtMillis(): Long`
  - `fun receivedResponseAtMillis(): Long`
  - `fun cacheControl(): CacheControl`

### DIFF
--- a/okhttp/src/main/java/okhttp3/Response.kt
+++ b/okhttp/src/main/java/okhttp3/Response.kt
@@ -37,25 +37,6 @@ import java.net.HttpURLConnection.HTTP_UNAUTHORIZED
  * [ResponseBody] for an explanation and examples.
  */
 class Response internal constructor(
-  internal val request: Request,
-  internal val protocol: Protocol,
-  internal val message: String,
-  builder: Builder
-) : Closeable {
-  internal val code: Int = builder.code
-  internal val handshake: Handshake? = builder.handshake
-  internal val headers: Headers = builder.headers.build()
-  internal val body: ResponseBody? = builder.body
-  internal val networkResponse: Response? = builder.networkResponse
-  internal val cacheResponse: Response? = builder.cacheResponse
-  internal val priorResponse: Response? = builder.priorResponse
-  internal val sentRequestAtMillis: Long = builder.sentRequestAtMillis
-  internal val receivedResponseAtMillis: Long = builder.receivedResponseAtMillis
-  internal val exchange: Exchange? = builder.exchange
-
-  @Volatile
-  private var cacheControl: CacheControl? = null // Lazily initialized.
-
   /**
    * The wire-level request that initiated this HTTP response. This is not necessarily the same
    * request issued by the application:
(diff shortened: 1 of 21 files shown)


okhttp/src/main/java/okhttp3/Response.kt
```diff
@@ -353,7 +424,17 @@ class Response internal constructor(
...
-          this)
+          code,
+          handshake,
+          headers.build(),
+          body,
+          networkResponse,
+          cacheResponse,
+          priorResponse,
+          sentRequestAtMillis,
+          receivedResponseAtMillis,
```
monkey-mas: Should we check `sentRequestAtMillis >= 0` and `receivedResponseAtMillis >= 0` like we do `check(code > 0) { "code < 0: $code" }`?

okhttp/src/main/java/okhttp3/Response.kt
```diff
@@ -65,12 +63,27 @@ class Response internal constructor(
    * * It may be the request generated in response to an HTTP redirect or authentication
    *   challenge. In this case the request URL may be different than the initial request URL.
    */
```
swankjesse: Move the docs from the deprecated functions to the corresponding vals?

okhttp/src/main/java/okhttp3/Response.kt
swankjesse: No, 0 isn’t special for timestamps

okhttp/src/main/java/okhttp3/Response.kt
monkey-mas: For sure! Thanks for pointing it out!

okhttp/src/main/java/okhttp3/Response.kt
```diff
@@ -65,12 +46,96 @@ class Response internal constructor(
...
   /** Returns the HTTP status code. */
+  @get:JvmName("code") val code: Int,
+
+  /**
+   * Returns the TLS handshake of the connection that carried this response, or null if the
+   * response was received without TLS.
+   */
+  @get:JvmName("handshake") val handshake: Handshake?,
+
+  /** Returns the HTTP headers. */
```
monkey-mas: added this comment as well for "headers"

PR #5112, opened 2019-05-24
Use apply to simplify MockResponse

TangHuaiZhe:
- Use `apply` to simplify `MockResponse`
- And a `with` in `Address`

### DIFF
--- a/okhttp/src/main/java/okhttp3/Address.kt
+++ b/okhttp/src/main/java/okhttp3/Address.kt
@@ -184,17 +184,17 @@ class Address(
     return result
   }
 
-  internal fun equalsNonHost(that: Address): Boolean = with(this) {
-    dns == that.dns &&
-        proxyAuthenticator == that.proxyAuthenticator &&
-        protocols == that.protocols &&
-        connectionSpecs == that.connectionSpecs &&
-        proxySelector == that.proxySelector &&
-        proxy == that.proxy &&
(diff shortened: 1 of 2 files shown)


okhttp/src/main/java/okhttp3/Address.kt
```diff
@@ -184,17 +184,17 @@ class Address(
...
-        this.proxyAuthenticator == that.proxyAuthenticator &&
-        this.protocols == that.protocols &&
-        this.connectionSpecs == that.connectionSpecs &&
-        this.proxySelector == that.proxySelector &&
-        this.proxy == that.proxy &&
-        this.sslSocketFactory == that.sslSocketFactory &&
-        this.hostnameVerifier == that.hostnameVerifier &&
-        this.certificatePinner == that.certificatePinner &&
-        this.url.port == that.url.port
+  internal fun equalsNonHost(that: Address): Boolean = with(this) {
```
Egorand: Do we need `with` here? Can simply get rid of `this`, although I'd keep it, it reads better

JakeWharton: I have a hard time justifying the use of `with` anywhere, but as you state `this` is already in scope so it's pointless.

TangHuaiZhe: Well..Reverted.

PR #5113, opened 2019-05-25
Beef up Request's toString

JakeWharton:
Previously it unconditionally showed tags even if empty and would not show headers.

### DIFF
--- a/okhttp/src/main/java/okhttp3/Request.kt
+++ b/okhttp/src/main/java/okhttp3/Request.kt
@@ -109,7 +109,29 @@ class Request internal constructor(
       level = DeprecationLevel.WARNING)
   fun cacheControl(): CacheControl = cacheControl
 
-  override fun toString(): String = "Request{method=$method, url=$url, tags=$tags}"
+  override fun toString() = buildString {
+    append("Request{method=")
+    append(method)
+    append(", url=")
+    append(url)
+    if (headers.size != 0) {
+      append(", headers=[")
+      headers.forEachIndexed { index, (name, value) ->
+        if (index > 0) {
+          append(", ")
+        }
+        append(name)
+        append(':')
+        append(value)
+      }
+      append(']')
+    }
+    if (tags.isNotEmpty()) {
+      append(", tags=")
+      append(tags)
+    }
+    append('}')
+  }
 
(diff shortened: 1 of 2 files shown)


okhttp/src/main/java/okhttp3/Request.kt
```diff
@@ -109,7 +109,29 @@ class Request internal constructor(
       level = DeprecationLevel.WARNING)
   fun cacheControl(): CacheControl = cacheControl
 
-  override fun toString(): String = "Request{method=$method, url=$url, tags=$tags}"
+  override fun toString() = buildString {
+    append("Request{method=")
+    append(method)
+    append(", url=")
+    append(url)
+    if (headers.size != 0) {
```
JakeWharton: Annoyed this doesn't have `isEmpty` / `isNotEmpty`. Probably coming soon from me. This should behave more like a collection!

JakeWharton:
Will fix tomorrow

okhttp/src/main/java/okhttp3/Request.kt
```diff
@@ -109,7 +109,29 @@ class Request internal constructor(
...
+      append(", headers=[")
+      headers.forEachIndexed { index, (name, value) ->
```
swankjesse: should this be Headers’ toString() ?

JakeWharton: Probably, with it's current toString changing to some other function

swankjesse:
Some predictable failures:
```
java.lang.AssertionError: 
Expecting:
 "[0 ms] callStart: Request{method=GET, url=http://localhost:54263/}"
to match pattern:
 "\[\d+ ms] callStart: Request\{method=GET, url=http://localhost:54263/, tags=\{\}\}"
	at okhttp3.logging.HttpLoggingInterceptorTest$LogRecorder.assertLogMatch(HttpLoggingInterceptorTest.java:851)
	at okhttp3.logging.LoggingEventListenerTest$LogRecorder.assertLogMatch(LoggingEventListenerTest.java:228)
	at okhttp3.logging.LoggingEventListenerTest.get(LoggingEventListenerTest.java:76)
```

PR #5117, opened 2019-05-25
Add githubs security policy

yschimke:
Copied from  okhttp/BUG-BOUNTY.md

Seems to be a standard location e.g.

https://github.com/electron/electron/blob/master/SECURITY.md
https://github.com/standard/standard/blob/master/SECURITY.md

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,18 @@
+# Security Policy
+
+## Supported Versions
+
+| Version | Supported          |
+| ------- | ------------------ |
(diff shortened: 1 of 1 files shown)


yschimke:
Happy to drop if we are happy with current state

yschimke:
@swankjesse just noticed that hackerone relies on BUG-BOUNTY.md file :(  I presume I can't delete.

yschimke:
@swankjesse Sorry about the copy and paste :(

https://github.com/square/okhttp/security/policy

image

PR #5118, opened 2019-05-25
Idiomatic Kotlin in ConnectionSpec.kt

monkey-mas:
Cont. #5115 

- define the following vals in constructor instead of passing `builder: Builder`.
  - `@get:JvmName("isTls") val isTls: Boolean`
  - `@get:JvmName("supportsTlsExtensions") val supportsTlsExtensions: Boolean`
  - `private val cipherSuitesAsString: Array<String>?`
  - `private val tlsVersionsAsString: Array<String>?`

  - `fun cipherSuites(): List<CipherSuite>?`
  - `fun tlsVersions(): List<TlsVersion>?`

### DIFF
--- a/okhttp/src/main/java/okhttp3/ConnectionSpec.kt
+++ b/okhttp/src/main/java/okhttp3/ConnectionSpec.kt
@@ -43,24 +43,50 @@ import javax.net.ssl.SSLSocket
  *
  * [tls_history]: https://github.com/square/okhttp/wiki/TLS-Configuration-History
  */
-class ConnectionSpec internal constructor(builder: Builder) {
-  val isTls: Boolean = builder.tls
-  internal val supportsTlsExtensions: Boolean = builder.supportsTlsExtensions
-  internal val cipherSuites: Array<String>? = builder.cipherSuites
-  internal val tlsVersions: Array<String>? = builder.tlsVersions
+class ConnectionSpec internal constructor(
+  @get:JvmName("isTls") val isTls: Boolean,
+  @get:JvmName("supportsTlsExtensions") val supportsTlsExtensions: Boolean,
+  private val cipherSuitesAsString: Array<String>?,
+  private val tlsVersionsAsString: Array<String>?
+) {
 
   /**
    * Returns the cipher suites to use for a connection. Returns null if all of the SSL socket's
    * enabled cipher suites should be used.
    */
-  fun cipherSuites(): List<CipherSuite>? = cipherSuites?.map(CipherSuite.Companion::forJavaName)
+  @get:JvmName("cipherSuites") val cipherSuites: List<CipherSuite>?
+    get() {
+      return cipherSuitesAsString?.map { CipherSuite.forJavaName(it) }?.toList()
+    }
+
+  @JvmName("-deprecated_cipherSuites")
+  @Deprecated(
+      message = "moved to val",
+      replaceWith = ReplaceWith(expression = "cipherSuites"),
+      level = DeprecationLevel.WARNING)
+  fun cipherSuites(): List<CipherSuite>? = cipherSuites
 
   /**
    * Returns the TLS versions to use when negotiating a connection. Returns null if all of the SSL
    * socket's enabled TLS versions should be used.
    */
-  fun tlsVersions(): List<TlsVersion>? = tlsVersions?.map(TlsVersion.Companion::forJavaName)
+  @get:JvmName("tlsVersions") val tlsVersions: List<TlsVersion>?
+    get() {
+      return tlsVersionsAsString?.map { TlsVersion.forJavaName(it) }?.toList()
+    }
 
+  @JvmName("-deprecated_tlsVersions")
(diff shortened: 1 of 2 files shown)


okhttp/src/main/java/okhttp3/ConnectionSpec.kt
```diff
@@ -43,51 +43,90 @@ import javax.net.ssl.SSLSocket
...
-  internal val cipherSuites: Array<String>? = builder.cipherSuites
-  internal val tlsVersions: Array<String>? = builder.tlsVersions
+class ConnectionSpec internal constructor(
+  @get:JvmName("isTls") val isTls: Boolean,
+  @get:JvmName("supportsTlsExtensions") val supportsTlsExtensions: Boolean,
+  private val cipherSuitesAsString: Array<String>?,
+  private val tlsVersionsAsString: Array<String>?
+) {
+
+  private var lazyCipherSuites: List<CipherSuite>? = null
```
yschimke: This seems problematic, e.g. makes us consider concurrency.  Why bundle this as part of this PR?

swankjesse: It’s the same pattern we do in Request.cacheControl, which works. Regardless for consistent behavior with what we did previously this should probably just compute on each call.

okhttp/src/main/java/okhttp3/ConnectionSpec.kt
```diff
@@ -43,51 +43,90 @@ import javax.net.ssl.SSLSocket
...
   /**
    * Returns the TLS versions to use when negotiating a connection. Returns null if all of the SSL
    * socket's enabled TLS versions should be used.
    */
-  fun tlsVersions(): List<TlsVersion>? = tlsVersions?.map(TlsVersion.Companion::forJavaName)
+  @get:JvmName("tlsVersions") val tlsVersions: List<TlsVersion>?
+    get() {
+      var result = lazyTlsVersions
+      if (result == null) {
+        result = tlsVersionsAsString?.map { TlsVersion.forJavaName(it) }?.toList()
```
swankjesse: similarly - change this to not cache?

okhttp/src/main/java/okhttp3/ConnectionSpec.kt
```diff
@@ -185,8 +224,8 @@ class ConnectionSpec internal constructor(builder: Builder) {
 
     constructor(connectionSpec: ConnectionSpec) {
       this.tls = connectionSpec.isTls
-      this.cipherSuites = connectionSpec.cipherSuites
-      this.tlsVersions = connectionSpec.tlsVersions
+      this.cipherSuites = connectionSpec.cipherSuites?.map { it.javaName }?.toTypedArray()
```
swankjesse: connectionSpec.cipherSuitesAsString

okhttp/src/main/java/okhttp3/ConnectionSpec.kt
```diff
@@ -216,7 +255,7 @@ class ConnectionSpec internal constructor(builder: Builder) {
     fun tlsVersions(vararg tlsVersions: TlsVersion): Builder = apply {
       require(tls) { "no TLS versions for cleartext connections" }
 
-      val strings = tlsVersions.map(TlsVersion::javaName).toTypedArray()
+      val strings = tlsVersions.map { it.javaName }.toTypedArray()
```
swankjesse: tlsVersionsAsString

okhttp/src/main/java/okhttp3/TlsVersion.kt
```diff
@@ -19,13 +19,20 @@ package okhttp3
  * Versions of TLS that can be offered when negotiating a secure socket. See
  * [javax.net.ssl.SSLSocket.setEnabledProtocols].
  */
-enum class TlsVersion(private val javaName: String) {
+enum class TlsVersion(
+  @get:JvmName("javaName") val javaName: String
```
swankjesse: 👍🏻

okhttp/src/main/java/okhttp3/ConnectionSpec.kt
monkey-mas: Removed the cache!

okhttp/src/test/java/okhttp3/ConnectionSpecTest.java
```diff
@@ -181,6 +181,21 @@
         .build();
   }
 
+  @Test public void tls_recomputesCiphersAndVersions() throws Exception {
```
monkey-mas: Sounds like this is what we expect instead of using a cache?

swankjesse: I don't want this test. It's not a behavior we guarantee either way. Just don't want PRs that intend to add Kotlin idioms also adding caches.

monkey-mas: sure! removed the test :)

PR #5120, opened 2019-05-25
Use vars in Dispatcher.

swankjesse:
This breaks Kotlin source compatibility in a small way that shouldn't be too
difficult for developers to figure out and fix.

### DIFF
--- a/okhttp/src/main/java/okhttp3/Dispatcher.kt
+++ b/okhttp/src/main/java/okhttp3/Dispatcher.kt
@@ -37,7 +37,7 @@ class Dispatcher constructor() {
    * The maximum number of requests to execute concurrently. Above this requests queue in memory,
    * waiting for the running calls to complete.
    *
-   * If more than `maxRequests` requests are in flight when this is invoked, those requests will
+   * If more than [maxRequests] requests are in flight when this is invoked, those requests will
    * remain in flight.
    */
   @get:Synchronized var maxRequests = 64
--- a/okhttp/src/main/java/okhttp3/Dispatcher.kt
+++ b/okhttp/src/main/java/okhttp3/Dispatcher.kt
@@ -28,18 +28,17 @@ import java.util.concurrent.TimeUnit
 /**
  * Policy on when async requests are executed.
  *
- *
  * Each dispatcher uses an [ExecutorService] to run calls internally. If you supply your own
  * executor, it should be able to run [the configured maximum][maxRequests] number of calls
  * concurrently.
  */
 class Dispatcher constructor() {
   /**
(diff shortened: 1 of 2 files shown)


okhttp/src/test/java/okhttp3/KotlinSourceCompatibilityTest.kt
```diff
@@ -308,9 +308,6 @@ class KotlinSourceCompatibilityTest {
     val maxRequestsPerHost: Int = dispatcher.maxRequestsPerHost
     dispatcher.maxRequestsPerHost = 0
     val executorService: ExecutorService = dispatcher.executorService()
-    dispatcher.setIdleCallback(object : Runnable {
```
swankjesse: turns out to be quite difficult to provide a `ReplaceWith` for this; I think it might be an IntelliJ bug.

okhttp/src/main/java/okhttp3/Dispatcher.kt
```diff
@@ -28,18 +28,17 @@ import java.util.concurrent.TimeUnit
...
 class Dispatcher constructor() {
   /**
-   * Set the maximum number of requests to execute concurrently. Above this requests queue in
-   * memory, waiting for the running calls to complete.
+   * The maximum number of requests to execute concurrently. Above this requests queue in memory,
+   * waiting for the running calls to complete.
    *
-   * If more than `maxRequests` requests are in flight when this is invoked, those requests
-   * will remain in flight.
+   * If more than `maxRequests` requests are in flight when this is invoked, those requests will
```
yschimke:
```suggestion
   * If more than [maxRequests] requests are in flight when this is invoked, those requests will
```

okhttp/src/main/java/okhttp3/Dispatcher.kt
```diff
@@ -51,13 +50,12 @@ class Dispatcher constructor() {
...
-   * by the URL's host name. Note that concurrent requests to a single IP address may still exceed
-   * this limit: multiple hostnames may share an IP address or be routed through the same HTTP
-   * proxy.
+   * The maximum number of requests for each host to execute concurrently. This limits requests by
+   * the URL's host name. Note that concurrent requests to a single IP address may still exceed this
+   * limit: multiple hostnames may share an IP address or be routed through the same HTTP proxy.
    *
-   * If more than `maxRequestsPerHost` requests are in flight when this is invoked, those
-   * requests will remain in flight.
+   * If more than `maxRequestsPerHost` requests are in flight when this is invoked, those requests
```
yschimke:
```suggestion
   * If more than [maxRequestsPerHost] requests are in flight when this is invoked, those requests
```

PR #5121, opened 2019-05-25
Extension Functions toRequestBody(), toResponseBody()

swankjesse:


okhttp/src/main/java/okhttp3/RequestBody.kt
```diff
@@ -91,11 +91,12 @@ abstract class RequestBody {
...
+     * Returns a new request body that transmits this string. If [contentType] is non-null and lacks
+     * a charset, this will use UTF-8.
      */
     @JvmStatic
-    fun create(contentType: MediaType?, content: String): RequestBody {
+    @JvmName("create")
+    fun String.toRequestBody(contentType: MediaType? = null): RequestBody {
       var charset: Charset = UTF_8
       var finalContentType: MediaType? = contentType
       if (contentType != null) {
```
yschimke: 4.1: this logic of resolving a final MediaType with charset seems like it should be part of MediaType.  Especially if it's something we deem important enough to do automatically for users of the RequestBody API.

okhttp/src/main/java/okhttp3/RequestBody.kt
```diff
@@ -107,56 +108,108 @@ abstract class RequestBody {
           charset = resolvedCharset
         }
       }
-      val bytes = content.toByteArray(charset)
-      return create(finalContentType, bytes)
+      val bytes = toByteArray(charset)
```
yschimke: We should arguably prefer forms using ByteString here over ByteArray

okhttp/src/main/java/okhttp3/ResponseBody.kt
```diff
@@ -221,36 +221,84 @@ abstract class ResponseBody : Closeable {
...
 
-    /** Returns a new response body that transmits [content]. */
+    /** Returns a new response body that transmits this byte string. */
     @JvmStatic
-    fun create(contentType: MediaType?, content: ByteString): ResponseBody {
-      val buffer = Buffer().write(content)
-      return create(contentType, content.size.toLong(), buffer)
+    @JvmName("create")
+    fun ByteString.toResponseBody(contentType: MediaType? = null): ResponseBody {
+      return Buffer()
```
yschimke: Seems like we should be able to avoid the copy into new Buffer, since the source is an immutable ByteArray wrapper.

okhttp-logging-interceptor/src/test/java/okhttp3/logging/HttpLoggingInterceptorTest.java
```diff
@@ -140,7 +140,7 @@ private void setLevel(Level level) {
     setLevel(Level.BASIC);
 
     server.enqueue(new MockResponse());
-    client.newCall(request().post(RequestBody.create(PLAIN, "Hi?")).build()).execute();
+    client.newCall(request().post(RequestBody.create("Hi?", PLAIN)).build()).execute();
```
yschimke: This seems like a major breaking public API change.

swankjesse:
The Java API is still there, just deprecated.

okhttp-logging-interceptor/src/test/java/okhttp3/logging/HttpLoggingInterceptorTest.java
swankjesse: For most of our API changes Java is not impacted at all. This one is necessary because the extension function’s receiver needs to be the first parameter, but our API had content-type there. 

The API change Java users will see is a change in argument order only. If Java users don't make the change they'll still be fine.

okhttp/src/main/java/okhttp3/ResponseBody.kt
swankjesse: The catch is this ultimately needs to produce a BufferedSource, so copying into a Buffer is inescapable.

The other catch is the source has to be one-shot, so if we defer that work we have to do some cursor management.

okhttp/src/main/java/okhttp3/RequestBody.kt
swankjesse: This is String’s built in API. Boxing it is just an extra allocation.

okhttp/src/main/java/okhttp3/RequestBody.kt
swankjesse: `mediaTpe.copy(charset=UTF_8)` ? Dig.

(For follow up!)

okhttp/src/main/java/okhttp3/RequestBody.kt
yschimke: Yep, following through it appears optimal for allocations.

JakeWharton:
These feel like they should be `asResponseBody` / `asRequestBody` since they encapsulate rather than convert.

PR #5124, opened 2019-05-25
Idiomatic Kotlin in HandshakeCertificates

monkey-mas:
- define `@get:JvmName(...)` for the following vals in constructor.
  - `keyManager: X509KeyManager`
  - `trustManager: X509TrustManager`

  - `fun keyManager(): X509KeyManager`
  - `fun trustManager(): X509TrustManager`
  - `fun sslSocketFactory(): SSLSocketFactory`
  - `fun sslContext(): SSLContext`

### DIFF
--- a/okhttp-tls/src/main/java/okhttp3/tls/HandshakeCertificates.kt
+++ b/okhttp-tls/src/main/java/okhttp3/tls/HandshakeCertificates.kt
@@ -67,11 +67,22 @@ import javax.net.ssl.X509TrustManager
  *    roots private to an organization or service.
  */
 class HandshakeCertificates private constructor(
-  private val keyManager: X509KeyManager,
-  private val trustManager: X509TrustManager
+  @get:JvmName("keyManager") val keyManager: X509KeyManager,
+  @get:JvmName("trustManager") val trustManager: X509TrustManager
 ) {
+
+  @JvmName("-deprecated_keyManager")
+  @Deprecated(
+      message = "moved to val",
+      replaceWith = ReplaceWith(expression = "keyManager"),
+      level = DeprecationLevel.WARNING)
   fun keyManager(): X509KeyManager = keyManager
 
+  @JvmName("-deprecated_trustManager")
+  @Deprecated(
+      message = "moved to val",
+      replaceWith = ReplaceWith(expression = "trustManager"),
+      level = DeprecationLevel.WARNING)
   fun trustManager(): X509TrustManager = trustManager
 
   fun sslSocketFactory(): SSLSocketFactory = sslContext().socketFactory


okhttp-tls/src/main/java/okhttp3/tls/HandshakeCertificates.kt
```diff
@@ -124,10 +153,20 @@ class HandshakeCertificates private constructor(
...
+          Collections.addAll(trustedCertificates, *platformTrustManager.acceptedIssuers)
+        }
+      }
+
+    @JvmName("-deprecated_sslSocketFactory")
+    @Deprecated(
+        message = "moved to val",
+        replaceWith = ReplaceWith(expression = "sslSocketFactory"),
+        level = DeprecationLevel.WARNING)
+    fun addPlatformTrustedCertificates(): Builder = addPlatformTrustedCertificates
```
monkey-mas: Should we apply idiomatic Kotlin to Builder methods as well? 👀 My question could be: can we do an idiomatic Kotlin way like the following without `()` for Builder methods? 🤔 
```
Builder
  .computeA
  .build
```

okhttp-tls/src/main/java/okhttp3/tls/HandshakeCertificates.kt
```diff
@@ -67,20 +67,49 @@ import javax.net.ssl.X509TrustManager
  *    roots private to an organization or service.
  */
 class HandshakeCertificates private constructor(
-  private val keyManager: X509KeyManager,
-  private val trustManager: X509TrustManager
+  @get:JvmName("keyManager") val keyManager: X509KeyManager,
+  @get:JvmName("trustManager") val trustManager: X509TrustManager
```
monkey-mas: I think they don't have to be `private` given that we have `fun keyManager(): X509KeyManager`.

okhttp-tls/src/main/java/okhttp3/tls/HandshakeCertificates.kt
```diff
@@ -67,20 +67,49 @@ import javax.net.ssl.X509TrustManager
...
+  @Deprecated(
+      message = "moved to val",
+      replaceWith = ReplaceWith(expression = "trustManager"),
+      level = DeprecationLevel.WARNING)
   fun trustManager(): X509TrustManager = trustManager
 
-  fun sslSocketFactory(): SSLSocketFactory = sslContext().socketFactory
+  @get:JvmName("sslSocketFactory") val sslSocketFactory: SSLSocketFactory
+    get() {
+      return sslContext.socketFactory
```
yschimke: I'd argue against this being a val, it calls functions and creates a new heavyweight object, in fact two in a chain.

okhttp-tls/src/main/java/okhttp3/tls/HandshakeCertificates.kt
```diff
@@ -67,20 +67,49 @@ import javax.net.ssl.X509TrustManager
...
+      replaceWith = ReplaceWith(expression = "sslSocketFactory"),
+      level = DeprecationLevel.WARNING)
+  fun sslSocketFactory(): SSLSocketFactory = sslSocketFactory
 
-  fun sslContext(): SSLContext {
-    return Platform.get().newSSLContext().apply {
-      init(arrayOf<KeyManager>(keyManager), arrayOf<TrustManager>(trustManager), SecureRandom())
+  @get:JvmName("sslContext") val sslContext: SSLContext
+    get() {
+      return Platform.get().newSSLContext().apply {
```
yschimke: I'd argue against this being a val

swankjesse: Hmmm . . . I think SSLContext is immutable, which feels like a `val` to me.

okhttp-tls/src/main/java/okhttp3/tls/HandshakeCertificates.kt
```diff
@@ -124,10 +153,20 @@ class HandshakeCertificates private constructor(
      * certificates. Applications that connect to a known set of servers may be able to mitigate
      * this problem with [certificate pinning][CertificatePinner].
      */
-    fun addPlatformTrustedCertificates() = apply {
-      val platformTrustManager = Platform.get().platformTrustManager()
-      Collections.addAll(trustedCertificates, *platformTrustManager.acceptedIssuers)
-    }
+    @get:JvmName("addPlatformTrustedCertificates") val addPlatformTrustedCertificates: Builder
```
swankjesse: No, this is not a val.

okhttp-tls/src/main/java/okhttp3/tls/HandshakeCertificates.kt
swankjesse: No, definitely not

swankjesse: The goal isn’t to remove the parens, it’s to use `vals` where the semantics of them work.
https://kotlinlang.org/docs/reference/properties.html

okhttp-tls/src/main/java/okhttp3/tls/HandshakeCertificates.kt
yschimke: But it makes it look externally cheap, while internally it's creating this expensive (potentially) native implement SSLContext and calling init on it.  It's the kind of thing that Intellij would suggest extracting a common variable and reusing, but calling once vs calling 10 times is very different.  The equality of two SSLContext isn't known IIUC.  And they may be immutable, but there is state in SSLContext potentially like session ids etc.

swankjesse: Works for me.

monkey-mas: Thanks for your feedback! I removed the wrong `@get:JvmName(...) val` for `sslContext` and `sslSocketFactory`.

swankjesse:
Nice!

PR #5128, opened 2019-05-26
Adopt idiomatic Kotlin in HttpLoggingInterceptor

swankjesse:


okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.kt
```diff
@@ -283,37 +289,34 @@ class HttpLoggingInterceptor @JvmOverloads constructor(
...
+   * Returns true if the body in question probably contains human readable text. Uses a small
+   * sample of code points to detect unicode control characters commonly used in binary file
+   * signatures.
+   */
+  private fun Buffer.isUtf8(): Boolean {
+    try {
+      val prefix = Buffer()
+      val byteCount = size.coerceAtMost(64)
+      copyTo(prefix, 0, byteCount)
+      for (i in 0 until 16) {
```
egor-n: Can be `repeat(16) {`.

okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.kt
```diff
@@ -44,7 +43,8 @@ class HttpLoggingInterceptor @JvmOverloads constructor(
 
   @Volatile private var headersToRedact = emptySet<String>()
 
-  @Volatile @set:JvmName("-deprecated_setLevel") var level = Level.NONE
+  @set:JvmName("level")
```
monkey-mas: Is it good to define `@get:JvmName("level")` to match the setter name? 🤔WDYT? :)

okhttp-logging-interceptor/src/test/java/okhttp3/logging/HttpLoggingInterceptorTest.java
```diff
@@ -70,8 +70,8 @@
       new HttpLoggingInterceptor(applicationLogs);
 
   private void setLevel(Level level) {
-    networkInterceptor.setLevel(level);
-    applicationInterceptor.setLevel(level);
+    networkInterceptor.level(level);
+    applicationInterceptor.level(level);
```
monkey-mas: If it's not good to use `level(...)` this time, let's remove my commit. Or please rebase it into your commit it if the changes are fine :)

yschimke: We shouldn't be updating the java tests as part of these PRs, we lose an important check of public API checks.

Unfortunately these methods are excluded from the normal controls for other reasons

```
    // Became 'final' in 4.0.0:
    'okhttp3.logging.HttpLoggingInterceptor#redactHeader(java.lang.String)',
    'okhttp3.logging.HttpLoggingInterceptor#getLevel()',
    'okhttp3.logging.HttpLoggingInterceptor#setLevel(okhttp3.logging.HttpLoggingInterceptor$Level)'
```

monkey-mas: Sure! Deleted the test change commit :)

swankjesse:
@monkey-mas Please don’t add commits to other people’s pull requests, it’s not something we do on this project.

okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.kt
swankjesse: this is necessary to avoid a conflict with the `setLevel()` method that returns a `HttpLoggingInterceptor`.

monkey-mas:
> @monkey-mas Please don’t add commits to other people’s pull requests, it’s not something we do on this project.

Sure, sorry about it 🙇

okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.kt
```diff
@@ -129,16 +129,24 @@ class HttpLoggingInterceptor @JvmOverloads constructor(
...
-    newHeadersToRedact.addAll(headersToRedact)
-    newHeadersToRedact.add(name)
+    newHeadersToRedact += headersToRedact
+    newHeadersToRedact += name
     headersToRedact = newHeadersToRedact
   }
 
+  @Deprecated(
+      message = "Moved to var. Replace setLevel(...) with level(...) to fix Java",
+      replaceWith = ReplaceWith(expression = "run { this.level = level }"),
```
yschimke: This looks wrong, I would have thought run should be apply here.

https://proandroiddev.com/the-difference-between-kotlins-functions-let-apply-with-run-and-else-ca51a4c696b8

run should be returning the level as the result instead of the interceptor.

yschimke:
> > @monkey-mas Please don’t add commits to other people’s pull requests, it’s not something we do on this project.
> 
> Sure, sorry about it 🙇

Yeah, I did this on one PR and it felt immediately awkward.

https://github.com/square/okhttp/pull/5038#discussion_r283121957

okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.kt
monkey-mas: Yeah, I see your point. `apply { this.level = level; this }` gives a way for the user to do 100% compatible replacement for a while. Once we remove the deprecated `setLevel()`, they'll end up with changing the code into `this.level = level` later. :eyes:

yschimke: you shouldn't even need the "; this".  apply should execute the block but still return the interceptor.

monkey-mas: ah yeah, you're right! `; this` shouldn't be there for `apply(...)`. Thanks!
`public inline fun <T> T.apply(block: T.() -> Unit): T`
`public inline fun <T, R> T.run(block: T.() -> R): R`

After doing this kind of stuff in #5144 , it feels to me that `apply { this.level = level }` would be nice here? 🤔

JakeWharton: +1 to apply

okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.kt
swankjesse: Unfortunately that doesn’t support `break`.

okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.kt
swankjesse: Fixed.

PR #5129, opened 2019-05-26
New releasing guide.

swankjesse:
Also make readmes offer artifacts with Gradle syntax

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -1,0 +1,47 @@
+Releasing
+=========
+
+1. Update `CHANGELOG.md`.
+
+2. Set versions:
+
+    ```
+    export RELEASE_VERSION=X.Y.Z
+    export NEXT_VERSION=X.Y.Z-SNAPSHOT
+    ```
+
+3. Update, build, and upload:
+
+    ```
+    sed -i "" \
+      "s/VERSION_NAME=.*/VERSION_NAME=$RELEASE_VERSION/g" \
+      gradle.properties
+    sed -i "" \
+      "s/\"com.squareup.okhttp3:\([^\:]*\):[^\"]*\"/\"com.squareup.okhttp3:\1:$RELEASE_VERSION\"/g" \
+      `find . -name "README.md"`
+    ./gradlew clean uploadArchives
+    ```
+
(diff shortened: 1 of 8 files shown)


RELEASING.md
```diff
@@ -0,0 +1,47 @@
+Releasing
```
yschimke: Do you intend to automate this? e.g. a normal flow for some projects would be that the act of creating a tag matching the version pattern is enough to create 

https://help.github.com/en/articles/creating-releases
https://developer.github.com/v3/repos/releases/#create-a-release
https://circleci.com/blog/how-my-distributed-team-communicates-so-no-context-is-left-behind/

Or just move these steps into scripts checked in instead of docs.

swankjesse: It’s _almost_ reliable enough to be a script. For now this is for my benefit as the person cutting the releases. (This was more automated with Maven, which would prompt for old and new versions.)

yschimke: I find the gradle versions plugins great as they derive sensible from the git tag.  So it boils down to whether the person or CI system running the release process has rights and keys to github, bintray etc.

swankjesse: Not sure I follow. Most of the work in this process is writing the changelog and using the awkward Maven Central UI.

yschimke: Process can be driven using the Github New Release screen.

`export RELEASE_VERSION=X.Y.Z` can come from the git tag
`./gradlew clean uploadArchives` could be in the CI workflow definition for release tags.
`git tag` done by the Github Flow
Changelog text could be the text entered into the release to start with.

n.b. feeling like this at the moment though - https://en.wikipedia.org/wiki/Teaching_grandmother_to_suck_eggs

swankjesse: I finally understand your recommendation: build the release in CI rather than on my Mac. That's something worth considering.

I’m reluctant to use GitHub releases ’cause they're too fancy. I prefer the naked simplicity of a changelog file. I'd also like to move our wiki out of GitHub’s wiki and into our repo!

JakeWharton: RxAndroid releases by just creating and pushing a tag to CI. I hate it. I'm all for automating the readme updating, but as for the actual release, I don't mind doing a few manual steps where I can control what's happening.

PR #5132, opened 2019-05-26
Replace MediaType.get(string) with string.toMediaType()

swankjesse:


okhttp/src/main/java/okhttp3/internal/platform/Jdk8WithJettyBootPlatform.kt
```diff
@@ -129,7 +129,7 @@ class Jdk8WithJettyBootPlatform(
         // 1.8, 9, 10, 11, 12 etc
         val version = jvmVersion.toInt()
         if (version >= 9) return null
-      } catch (nfe: NumberFormatException) {
+      } catch (_: NumberFormatException) {
         // expected on >= JDK 9
```
monkey-mas: `NumberFormatException` is thrown and caught when JVM version is <= 1.8? but not for JDK >= 9? 🤔

yschimke:
```
// 1.8, 9, 10, 11, 12 etc
val version = jvmVersion.toInt()
```

toInt() fails on 1.8.

PR #5135, opened 2019-05-28
Silent Code Cleanup

yschimke:
next batch of automated fixes


yschimke:
I’m just clicking the menu item in IntelliJ and it applies these changes. No need to do one by one. 

Issue #4950, opened 2019-04-15
HostName becomes "[localhost]" inside MockWebServer RecordedRequest Constructor for ipv6 host


CoderSpinoza:
### Symptom

**RecordedRequest#getRequestUrl() is returning null because hostname is set to "[localhost]", which causes IllegalArgumentException inside HttpUrl#parse().**

### Reproducible Env

Simple test using MockWebServer can reproduce this issue.

Affected versions: Above 3.12.0 (Probably related to https://github.com/square/okhttp/pull/4235)
OS: macOS 10.14.4, Android API Level 28
Network env: Secure VPN (ipv6)

#### Test

```kotlin
package auth

import auth.model.AccessTokenResponse
import io.reactivex.observers.TestObserver
import okhttp3.mockwebserver.MockResponse
import okhttp3.mockwebserver.MockWebServer
import org.junit.jupiter.api.Assertions
import org.junit.jupiter.api.Test
import retrofit2.Retrofit
import retrofit2.adapter.rxjava2.RxJava2CallAdapterFactory
import retrofit2.converter.gson.GsonConverterFactory

/**
 * @author kevin.kang. Created on 15/04/2019..
 */
class MockWebServerTest {
    @Test fun requestUrl() {
        val server = MockWebServer()
        server.enqueue(MockResponse().setResponseCode(200))
        val api = Retrofit.Builder().baseUrl(server.url("/"))
            .addCallAdapterFactory(RxJava2CallAdapterFactory.create())
            .addConverterFactory(GsonConverterFactory.create()).build().create(AuthApi::class.java)
        api.issueAccessToken(
            authCode = "auth_code",
            clientId = "some_client_id",
            redirectUri = "https://example.com/oauth",
            androidKeyHash = "sample_key_hash"
        ).subscribe(TestObserver<AccessTokenResponse>())
        val request = server.takeRequest()
        Assertions.assertNotNull(request.requestUrl)
        server.shutdown()
    }
}
```

Since the reproducibility might depend on the network environment, I attach extra debug infos.

Hostname becomes "[localhost]" inside the constructor of RecordedRequest.

```java
InetAddress inetAddress = socket.getLocalAddress(); // inetAddress: "/0:0:0:0:0:0:0:1"
String hostname = inetAddress.getHostName(); // hostname: "localhost"
if (inetAddress instanceof Inet6Address) {
hostname = "[" + hostname + "]"; // hostname: "[localhost]" 
```

[...]

PR #5137, opened 2019-05-28
Do not enclose localhost hostname with square brackets #4950

CoderSpinoza:
Inet6Address is returning localhost as hostname under specific network env (ex. Vpn).
“Localhost” should not be enclosed with square brackets since they will produce IllegalArugmentException in HttpUrl if so.


CoderSpinoza:
This is the most simple fix, which asserts that hostname never becomes "[localhost]".

yschimke:
Can you add a test to RecordedRequestTest

swankjesse:
Please complete our contributor’s agreement:

yschimke:
@CoderSpinoza Any plans to sign the CLA and add a test, or should we take over and reimplement with tests?

PR #5140, opened 2019-05-28
Testing with Quiche

yschimke:


yschimke:
Expecting this to fail, but curious what a HTTP/2 client connecting to HTTP/3 services looks like.  Don't have ipv6 at home.

yschimke:
https://github.com/cloudflare/quiche

PR #5141, opened 2019-05-30
Add ResponseBody.byteString convenience function

JakeWharton:


okhttp/src/main/java/okhttp3/ResponseBody.kt
```diff
@@ -121,16 +121,31 @@ abstract class ResponseBody : Closeable {
    * possibility for your response.
    */
   @Throws(IOException::class)
-  fun bytes(): ByteArray {
+  fun bytes() = consumeSource(BufferedSource::readByteArray) { it.size }
```
JakeWharton: Can't be function reference because of https://youtrack.jetbrains.com/issue/KT-21014

okhttp/src/main/java/okhttp3/ResponseBody.kt
```diff
@@ -121,16 +121,31 @@ abstract class ResponseBody : Closeable {
...
+
+  /**
+   * Returns the response as a [ByteString].
+   *
+   * This method loads entire response body into memory. If the response body is very large this
+   * may trigger an [OutOfMemoryError]. Prefer to stream the response body if this is a
+   * possibility for your response.
+   */
+  @Throws(IOException::class)
+  fun byteString() = consumeSource(BufferedSource::readByteString) { it.size }
```
JakeWharton: Can't be function reference because of property / function name overload

okhttp/src/main/java/okhttp3/ResponseBody.kt
```diff
@@ -121,16 +121,31 @@ abstract class ResponseBody : Closeable {
...
+    if (contentLength > Int.MAX_VALUE) {
       throw IOException("Cannot buffer entire body for content length: $contentLength")
     }
 
-    val bytes: ByteArray = source().use(BufferedSource::readByteArray)
-    if (contentLength != -1L && contentLength != bytes.size.toLong()) {
-      throw IOException(
-          "Content-Length ($contentLength) and stream length (${bytes.size}) disagree")
+    val bytes = source().use(consumer)
+    val size = sizeMapper(bytes)
```
yschimke: TIOLI: maybe for 4.1?

Could sizeMapper be replaced by just checking the source() is exhausted?  This would also check the contentLength case, when we presumably would want to consume the whole stream anyway.

swankjesse:
I’d like to sit on all API + behavior changes until after 4.0 is cut. Only because I’d like to really minimize the perceived upgrade risk to our users.

yschimke:
@swankjesse @JakeWharton merged assuming any quick fixes would be a 4.0.1 maintenance release, and master is now 4.1.x

PR #5003, opened 2019-05-03
Convert mockweserver/PushPromise SocketPolicy RecordedRequest to Kotlin

TangHuaiZhe:

PR #5147, opened 2019-05-31
Idiomatic Kotlin in RecordedRequest.kt

monkey-mas:
- define `@get:JvmName(...)` for the following vals.
  - `requestLine: String`
  - `headers: Headers`
  - `chunkSizes: List<Int>`
  - `bodySize: Long`
  - `body: Buffer`
  - `sequenceNumber: Int`
  - `method: String?`
  - `path: String?`
  - `handshake: Handshake?`
  - `requestUrl: HttpUrl?`
  - `tlsVersion: TlsVersion?`

  - `fun getRequestLine(): String`
  - `fun getHeaders(): Headers`
  - `fun getChunkSizes(): List<Int>`
  - `fun getBodySize(): Long`
  - `fun getBody(): Buffer`
  - `fun getSequenceNumber(): Int`
  - `fun getMethod(): String?`
  - `fun getPath(): String?`
  - `fun getHandshake(): Handshake?`
  - `fun getRequestUrl(): HttpUrl?`
  - `fun getTlsVersion(): TlsVersion?`

- update `KotlinSourceCompatibilityTest` and `KotlinSourceModernTest`

### DIFF
--- a/mockwebserver/src/main/java/okhttp3/mockwebserver/RecordedRequest.kt
+++ b/mockwebserver/src/main/java/okhttp3/mockwebserver/RecordedRequest.kt
@@ -30,40 +30,50 @@ import javax.net.ssl.SSLSocket
 
 /** An HTTP request that came into the mock web server. */
 class RecordedRequest(
-  val requestLine: String,
+  @get:JvmName("requestLine") val requestLine: String,
+
   /** All headers. */
-  val headers: Headers,
+  @get:JvmName("headers") val headers: Headers,
+
   /**
    * The sizes of the chunks of this request's body, or an empty list if the request's body
    * was empty or unchunked.
    */
-  val chunkSizes: List<Int>,
+  @get:JvmName("chunkSizes") val chunkSizes: List<Int>,
+
   /** The total size of the body of this POST request (before truncation).*/
-  val bodySize: Long,
+  @get:JvmName("bodySize") val bodySize: Long,
+
   /** The body of this POST request. This may be truncated. */
-  val body: Buffer,
+  @get:JvmName("body") val body: Buffer,
+
   /**
    * The index of this request on its HTTP connection. Since a single HTTP connection may
    * serve multiple requests, each request is assigned its own sequence number.
    */
-  val sequenceNumber: Int,
-  socket: Socket
+  @get:JvmName("sequenceNumber") val sequenceNumber: Int,
+
+  internal val socket: Socket
 ) {
-  val method: String?
-  val path: String?
+  @get:JvmName("method") val method: String?
+  @get:JvmName("path") val path: String?
+
   /**
    * The TLS handshake of the connection that carried this request, or null if the request
    * was received without TLS.
    */
-  val handshake: Handshake?
-  val requestUrl: HttpUrl?
+  @get:JvmName("handshake") val handshake: Handshake?
+  @get:JvmName("requestUrl") val requestUrl: HttpUrl?
 
   val utf8Body: String
-    @Deprecated("Use {@link #getBody() getBody().readUtf8()}. ")
+    @Deprecated(
+        message = "Use body().readUtf8() instead.",
+        replaceWith = ReplaceWith(expression = "body.readUtf8"),
+        level = DeprecationLevel.WARNING)
     get() = body.readUtf8()
 
   /** Returns the connection's TLS version or null if the connection doesn't use SSL. */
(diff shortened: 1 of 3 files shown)


okhttp/src/test/java/okhttp3/KotlinSourceCompatibilityTest.kt
```diff
@@ -943,31 +943,31 @@ class KotlinSourceCompatibilityTest {
     var recordedRequest: RecordedRequest = RecordedRequest(
         "", Headers.of(), listOf(), 0L, Buffer(), 0, Socket())
     recordedRequest = RecordedRequest("", Headers.of(), listOf(), 0L, Buffer(), 0, Socket())
-    var requestUrl: HttpUrl? = recordedRequest.requestUrl
-    var requestLine: String = recordedRequest.requestLine
-    var method: String? = recordedRequest.method
-    var path: String? = recordedRequest.path
-    var headers: Headers = recordedRequest.headers
+    var requestUrl: HttpUrl? = recordedRequest.getRequestUrl()
```
JakeWharton: This file shouldn't be changed. Only the "modern" one should be updated.

monkey-mas: ah, I thought `KotlinSourceCompatibilityTest` tests _deprecated_ functions and `KotlinSourceModernTest ` tests _idiomatic_ functions. But it sounds like I'm probably wrong?

JakeWharton: It tests how Kotlin compiles against OkHttp 3.x.

yschimke: You are both right, the file got cleaned up in https://github.com/square/okhttp/pull/5003#discussion_r280961308

@swankjesse ok'd it

swankjesse: Yeah, the goal of this file is to have a record of all of the source-incompatible changes we made so we can either avoid them or flag them to users.

mockwebserver/src/main/java/okhttp3/mockwebserver/RecordedRequest.kt
```diff
@@ -30,40 +30,50 @@ import javax.net.ssl.SSLSocket
...
-  val handshake: Handshake?
-  val requestUrl: HttpUrl?
+  @get:JvmName("handshake") val handshake: Handshake?
+  @get:JvmName("requestUrl") val requestUrl: HttpUrl?
 
   val utf8Body: String
-    @Deprecated("Use {@link #getBody() getBody().readUtf8()}. ")
+    @Deprecated(
+        message = "Use body().readUtf8() instead.",
+        replaceWith = ReplaceWith(expression = "body.readUtf8"),
```
swankjesse: the replace with expression is missing parens, should be `body.readUtf8()`

mockwebserver/src/main/java/okhttp3/mockwebserver/RecordedRequest.kt
```diff
@@ -30,40 +30,50 @@ import javax.net.ssl.SSLSocket
 
 /** An HTTP request that came into the mock web server. */
 class RecordedRequest(
-  val requestLine: String,
+  @get:JvmName("requestLine") val requestLine: String,
```
swankjesse: these were already idiomatic Kotlin; nothing to change

mockwebserver/src/main/java/okhttp3/mockwebserver/RecordedRequest.kt
```diff
@@ -30,40 +30,50 @@ import javax.net.ssl.SSLSocket
...
   /**
    * The TLS handshake of the connection that carried this request, or null if the request
    * was received without TLS.
    */
-  val handshake: Handshake?
-  val requestUrl: HttpUrl?
+  @get:JvmName("handshake") val handshake: Handshake?
+  @get:JvmName("requestUrl") val requestUrl: HttpUrl?
 
   val utf8Body: String
```
swankjesse: We should replace this `val` with a `fun`; it doesn’t behave like a `val` in that reading it has a side-effect

swankjesse:
Another shot at the same problem: https://github.com/square/okhttp/pull/5152

PR #5149, opened 2019-05-31
Code clean up

TangHuaiZhe:


mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.kt
```diff
@@ -270,7 +270,7 @@ class MockWebServer : ExternalResource(), Closeable {
    * @return the head of the request queue
    */
   @Throws(InterruptedException::class)
-  fun takeRequest(timeout: Long, unit: TimeUnit): RecordedRequest? =
+  fun takeRequest(timeout: Long, unit: TimeUnit): RecordedRequest =
```
JakeWharton: Don't think we want to make this change. It changes the behavior from 3.x which returned null (despite saying it never did) to throwing NPE. I'm comfortable with this source incompatibility.

TangHuaiZhe: Thanks, reverted.

PR #5152, opened 2019-06-01
Adopt idiomatic Kotlin in RecordedRequest

swankjesse:
Not much to change.

### DIFF
--- a/mockwebserver/src/main/java/okhttp3/mockwebserver/RecordedRequest.kt
+++ b/mockwebserver/src/main/java/okhttp3/mockwebserver/RecordedRequest.kt
@@ -31,35 +31,45 @@ import javax.net.ssl.SSLSocket
 /** An HTTP request that came into the mock web server. */
 class RecordedRequest(
   val requestLine: String,
+
   /** All headers. */
   val headers: Headers,
+
   /**
    * The sizes of the chunks of this request's body, or an empty list if the request's body
    * was empty or unchunked.
    */
   val chunkSizes: List<Int>,
+
   /** The total size of the body of this POST request (before truncation).*/
   val bodySize: Long,
+
   /** The body of this POST request. This may be truncated. */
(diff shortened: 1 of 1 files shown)


mockwebserver/src/main/java/okhttp3/mockwebserver/RecordedRequest.kt
```diff
@@ -105,12 +115,16 @@ class RecordedRequest(
...
 
+  @Deprecated(
+      message = "Use body.readUtf8()",
+      replaceWith = ReplaceWith("body.readUtf8()"),
+      level = DeprecationLevel.WARNING)
+  fun getUtf8Body(): String = body.readUtf8()
+
   /** Returns the first header named [name], or null if no such header exists. */
   fun getHeader(name: String): String? {
     return headers.values(name).firstOrNull()
```
yschimke: nit: could return expression like below also

mockwebserver/src/main/java/okhttp3/mockwebserver/RecordedRequest.kt
```diff
@@ -105,12 +115,16 @@ class RecordedRequest(
     }
   }
 
+  @Deprecated(
+      message = "Use body.readUtf8()",
+      replaceWith = ReplaceWith("body.readUtf8()"),
```
yschimke: curious: Does Intellij also convert the kotlin suggestions to Java?

swankjesse: No, Java users have to fix manually.

mockwebserver/src/main/java/okhttp3/mockwebserver/RecordedRequest.kt
swankjesse: Fixed!

PR #5153, opened 2019-06-01
Adopt idiomatic Kotlin in MockWebServer

swankjesse:


mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.kt
```diff
@@ -104,36 +104,84 @@ class MockWebServer : ExternalResource(), Closeable {
...
   var dispatcher: Dispatcher = QueueDispatcher()
-  private var port = -1
-  private var inetSocketAddress: InetSocketAddress? = null
-  private var protocolNegotiationEnabled = true
-  private var protocols = immutableListOf(Protocol.HTTP_2, Protocol.HTTP_1_1)
 
-  private var started: Boolean = false
+  var port: Int = -1
+    get() {
+      before()
```
swankjesse: this is my least favorite change. I don’t like vals that have side effects when you read them.

This is consistent with hostName below. I’m tempted to change both methods to only be available through a function rather than a val.

mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.kt
```diff
@@ -174,40 +229,38 @@ class MockWebServer : ExternalResource(), Closeable {
...
-   */
-  fun setProtocolNegotiationEnabled(protocolNegotiationEnabled: Boolean) {
+  @JvmName("-deprecated_bodyLimit")
+  @Deprecated(
+      message = "moved to var",
+      replaceWith = ReplaceWith(
+          expression = "run { this.bodyLimit = bodyLimit }"
+      ),
+      level = DeprecationLevel.WARNING)
+  fun setBodyLimit(bodyLimit: Long) = run { this.bodyLimit = bodyLimit }
```
yschimke: apply instead of run, otherwise you return bodyLimit as the result.  OR should these be returning Unit/Void?

mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.kt
yschimke: It's tough since this was the existing 3.14 behaviour also.  Essentially for any awkward kotlin source changes, it only get's harder to make these changes in 4.1.  So make them now since you already have made a few.

mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.kt
swankjesse: These return Unit.

mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.kt
swankjesse: I’m gonna keep it as-is for now.

PR #5144, opened 2019-05-30
Idiomatic Kotlin in MockResponse.kt

monkey-mas:

PR #5154, opened 2019-06-01
Use apply {} to adopt idiomatic Kotlin in Dispatcher and QueueDispatcher

monkey-mas:
Follow up of https://github.com/square/okhttp/pull/5144#issuecomment-497954889


swankjesse:
Thanks!

PR #5156, opened 2019-06-02
Fix Kotlin idioms in MockResponse. Headers and Trailers should be vals.

swankjesse:
Also track OkHttp 3.14.x in KotlinSourceCompatibilityTest. This shouldn't
have been updated to use new APIs.

### DIFF
--- a/mockwebserver/src/main/java/okhttp3/mockwebserver/MockResponse.kt
+++ b/mockwebserver/src/main/java/okhttp3/mockwebserver/MockResponse.kt
@@ -29,8 +29,23 @@ class MockResponse : Cloneable {
   @set:JvmName("status")
   var status: String = ""
 
-  private var headers = Headers.Builder()
-  private var trailers = Headers.Builder()
+  private var headersBuilder = Headers.Builder()
+  private var trailersBuilder = Headers.Builder()
+
+  /** The HTTP headers, such as "Content-Length: 0". */
+  @set:JvmName("headers")
+  var headers: Headers
+    get() = headersBuilder.build()
+    set(value) {
+      this.headersBuilder = value.newBuilder()
+    }
+
+  @set:JvmName("trailers")
+  var trailers: Headers
+    get() = trailersBuilder.build()
+    set(value) {
+      this.trailersBuilder = value.newBuilder()
+    }
 
   private var body: Buffer? = null
 
@@ -78,7 +93,7 @@ class MockResponse : Cloneable {
 
   public override fun clone(): MockResponse {
     val result = super.clone() as MockResponse
-    result.headers = headers.build().newBuilder()
+    result.headersBuilder = headersBuilder.build().newBuilder()
     result.promises = promises.toMutableList()
     return result
   }
@@ -110,17 +125,12 @@ class MockResponse : Cloneable {
     return apply { status = "HTTP/1.1 $code $reason" }
   }
 
-  /** Returns the HTTP headers, such as "Content-Length: 0". */
-  fun getHeaders(): Headers = headers.build()
-
-  fun getTrailers(): Headers = trailers.build()
-
   /**
    * Removes all HTTP headers including any "Content-Length" and "Transfer-encoding" headers that
    * were added by default.
    */
   fun clearHeaders() = apply {
-    headers = Headers.Builder()
+    headersBuilder = Headers.Builder()
   }
 
   /**
@@ -128,7 +138,7 @@ class MockResponse : Cloneable {
    * name followed by a colon and a value.
    */
   fun addHeader(header: String) = apply {
-    headers.add(header)
+    headersBuilder.add(header)
   }
 
   /**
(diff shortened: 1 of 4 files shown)


okhttp/src/test/java/okhttp3/KotlinSourceCompatibilityTest.kt
```diff
@@ -676,35 +682,35 @@ class KotlinSourceCompatibilityTest {
   fun mockResponse() {
     var mockResponse: MockResponse = MockResponse()
     var status: String = mockResponse.getStatus()
-    status = mockResponse.getStatus()
+    status = mockResponse.status
```
swankjesse: many of these accesses were migrated and shouldn’t have been

okhttp/src/test/java/okhttp3/KotlinSourceCompatibilityTest.kt
```diff
@@ -957,18 +961,6 @@ class KotlinSourceCompatibilityTest {
     var sequenceNumber: Int = recordedRequest.sequenceNumber
     var tlsVersion: TlsVersion? = recordedRequest.tlsVersion
     var handshake: Handshake? = recordedRequest.handshake
-    requestUrl = recordedRequest.requestUrl
```
swankjesse: these were all just repeated

okhttp/src/test/java/okhttp3/KotlinSourceCompatibilityTest.kt
```diff
@@ -1039,27 +1031,27 @@ class KotlinSourceCompatibilityTest {
   @Test @Ignore
   fun response() {
     val response: Response = Response.Builder().build()
-    val request: Request = response.request
-    val protocol: Protocol = response.protocol
-    val code: Int = response.code
+    val request: Request = response.request()
```
swankjesse: migrated and shouldn’t have been

mockwebserver/src/main/java/okhttp3/mockwebserver/MockResponse.kt
```diff
@@ -29,8 +29,23 @@ class MockResponse : Cloneable {
...
+  /** The HTTP headers, such as "Content-Length: 0". */
+  @set:JvmName("headers")
+  var headers: Headers
+    get() = headersBuilder.build()
+    set(value) {
+      this.headersBuilder = value.newBuilder()
+    }
+
+  @set:JvmName("trailers")
+  var trailers: Headers
```
monkey-mas: Probably the PR title wanted to say "Headers and Trailers should be `vars`."? 👀

mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.kt
```diff
@@ -329,8 +329,7 @@ class MockWebServer : ExternalResource(), Closeable {
    * @return the head of the request queue
    */
   @Throws(InterruptedException::class)
-  fun takeRequest(timeout: Long, unit: TimeUnit): RecordedRequest? =
-      requestQueue.poll(timeout, unit)
+  fun takeRequest(timeout: Long, unit: TimeUnit): RecordedRequest = requestQueue.poll(timeout, unit)
```
monkey-mas: Can we change the return type to non-nul here? 🤔

okhttp/src/test/java/okhttp3/KotlinSourceCompatibilityTest.kt
monkey-mas: My bad 🤦‍♂ Thanks for fixing!

monkey-mas:
Is this PR for `3.14.x` branch or `master`? 👀

mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.kt
swankjesse: Yes we can!

swankjesse:
PR is for master, KotlinSourceCompatibilityTest works on 3.14.x APIs and today’s.

okhttp/src/test/java/okhttp3/KotlinSourceCompatibilityTest.kt
swankjesse: It's a weird test.

mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.kt
JakeWharton: This changes it to a platform type, not non-null. We should switch it back.

swankjesse: I changed it to explicitly return `RecordedResponse`, which is non-null. Definitely need to avoid mystery types.

JakeWharton: There's probably a lint or something we can enable. Android's lint has a check and works on Java projects. I'll look into it later.

JakeWharton: Doesn't seem to run on Kotlin sources. Filed https://issuetracker.google.com/issues/134237547

PR #5157, opened 2019-06-02
Remove the Kotlin migration bridges from DnsOverHttps

swankjesse:
We promise no API stability in this package.

### DIFF
--- a/okhttp-dnsoverhttps/src/main/java/okhttp3/dnsoverhttps/DnsOverHttps.kt
+++ b/okhttp-dnsoverhttps/src/main/java/okhttp3/dnsoverhttps/DnsOverHttps.kt
@@ -36,18 +36,18 @@ import java.util.ArrayList
 import java.util.concurrent.CountDownLatch
 
 /**
- * [DNS over HTTPS implementation][[doh_spec]].
+ * [DNS over HTTPS implementation][doh_spec].
  *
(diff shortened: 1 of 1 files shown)


okhttp-dnsoverhttps/src/main/java/okhttp3/dnsoverhttps/DnsOverHttps.kt
```diff
@@ -57,55 +57,8 @@ class DnsOverHttps internal constructor(
...
-      replaceWith = ReplaceWith(expression = "includeIPv6"),
-      level = DeprecationLevel.WARNING)
-  fun includeIPv6(): Boolean = includeIPv6
-
-  @JvmName("-deprecated_client")
-  @Deprecated(
-      message = "moved to val",
-      replaceWith = ReplaceWith(expression = "client"),
-      level = DeprecationLevel.WARNING)
-  fun client(): OkHttpClient = client.newBuilder()
```
swankjesse: this was broken; the actual client val is what gets used, but only this one that is configured properly.

monkey-mas: Are we going to write API compatibility doc for 4.0.0-ALPHA02 or 4.0.0-RC1 in CHANGELOG.md? 👀

PR #5158, opened 2019-06-02
Remove most of the SAM bridges

swankjesse:
The ReplaceWith() on these doesn't work well.

For interceptors I expect the need to be common enough that I'm making these
new APIs that we'll keep going forward. This is for both creating interceptors
directly and as arguments to addInterceptor() and addNetworkInterceptor().

### DIFF
--- a/okhttp/src/main/java/okhttp3/Interceptor.kt
+++ b/okhttp/src/main/java/okhttp3/Interceptor.kt
@@ -28,13 +28,20 @@ interface Interceptor {
   fun intercept(chain: Chain): Response
 
   companion object {
-    @JvmName("-deprecated_Interceptor")
-    @Deprecated(
-        message = "No SAM (single-abstract-method) conversions for Kotlin declarations",
-        level = DeprecationLevel.WARNING)
-    operator fun invoke(block: (chain: Chain) -> Response): Interceptor = object : Interceptor {
-      override fun intercept(chain: Chain) = block(chain)
-    }
+    /**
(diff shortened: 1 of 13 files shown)


okhttp/src/test/java/okhttp3/KotlinSourceCompatibilityTest.kt
```diff
@@ -581,6 +576,7 @@ class KotlinSourceCompatibilityTest {
       override fun intercept(chain: Interceptor.Chain): Response = TODO()
     }
     interceptor = Interceptor { it: Interceptor.Chain -> TODO() }
```
yschimke: Is this example still useful?

swankjesse: Nope! Fixed!

okhttp/src/main/java/okhttp3/Interceptor.kt
```diff
@@ -28,13 +28,20 @@ interface Interceptor {
...
+     *
+     * ```
+     * val interceptor = Interceptor { chain: Interceptor.Chain ->
+     *     chain.proceed(chain.request())
+     * }
+     * ```
+     */
+    inline operator fun invoke(crossinline block: (chain: Chain) -> Response): Interceptor =
+        object : Interceptor {
+          override fun intercept(chain: Chain) = block(chain)
```
swankjesse: blockchain!

PR #5173, opened 2019-06-07
Add HttpLoggingInterceptor.Level.JUSTBODY for logging request and response bodies without headers

eldad-sk:
(if required, I could also copy create copies of the HttpLoggingInterceptorTest.body*() tests)

### DIFF
--- a/okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.kt
+++ b/okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.kt
@@ -82,6 +82,22 @@ class HttpLoggingInterceptor @JvmOverloads constructor(
     HEADERS,
 
     /**
+     * Logs request and response lines and their respective bodies (if present) without headers.
+     *
+     * Example:
+     * ```
+     * --> POST /greeting http/1.1
+     * Hi?
+     * --> END POST
+     *
+     * <-- 200 OK (22ms)
+     * Hello!
+     * <-- END HTTP
+     * ```
+     */
+    JUSTBODY,
+
+    /**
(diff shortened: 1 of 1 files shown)


swankjesse:
I think our logging interceptor is simple enough that if you wanna customize it, it's better to copy it into your project instead of growing the API for everyone.

eldad-sk:
The reason my project needs this feature is that in order to deal with legacy servers my app has to send a dozen headers for each request (most never change, a couple are related to session/auth). Logging headers makes it very difficult to focus on request / response bodies.

The question is whether this situation is all that rare.

swankjesse:
Wanna use `redactHeader()` to redact the headers you don’t want to see? That’s an easy fix that requires no new APIs.

yschimke:
I don't think we should build out in this direction.  Consider the HttpLoggingInterceptor sample code you can take a do whatever you need.  But making it too easy or defaulting to log request bodies is probably a net drag as client in real user devices spend time logging etc.  If I could go back in time, I'd probably suggest this interceptor in OkHttp should only show headers and have a hook for developers to opt in to logging bodies.

yschimke:
Closing as a more full featured HttpLoggingInterceptor should be a utility library or project specific.

eldad-sk:
@swankjesse Unfortunately, the redactHeader approach is not only messy and cumbersome (for one thing, you need to know the possible headers in advance), it still results in one line in the log file for every header sent or received.

I'm not sure I can copy HttpLoggingInterceptor into my own project since it seems to require internal classes.

yschimke:
@eldad-sk FWIW the only internal usage is the hook to Android logging, which isn't required in your code since you will call Log.i directly.  And also Response.promisesBody which is one that we will need to continue to support but you could copy anyway, or even simplify based on your needs.

PR #5176, opened 2019-06-09
More notes for the upgrade guide

swankjesse:
SAM improvements to Kotlin
Companion imports
R8 and ProGuard

### DIFF
--- a/UPGRADING_TO_OKHTTP_4.md
+++ b/UPGRADING_TO_OKHTTP_4.md
@@ -105,6 +105,9 @@ SAM conversion impacts these APIs:
  * LoggingEventListener.Factory
  * OkHttpClient.Builder.hostnameVerifier(HostnameVerifier)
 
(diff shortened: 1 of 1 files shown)


UPGRADING_TO_OKHTTP_4.md
```diff
@@ -198,3 +201,48 @@ null are uncommon and Kotlin callers may have incorrectly assigned the result to
 [japicmp_gradle]: https://github.com/melix/japicmp-gradle-plugin
 [mockito]: https://site.mockito.org/
 [kotlin_sam]: https://kotlinlang.org/docs/reference/java-interop.html#sam-conversions
+
+
+Companion Imports
```
ZacSweers: Would it be worth adding a line explaining why `@JvmStatic` isn't being used for better java compatibility here? I could guess (bytecode duplication), but I'm not actually sure from reading this and would wonder why

ZacSweers: d'oh ignore me, this is for kotlin imports

Issue #4894, opened 2019-04-06
Junit "listener" appears to be a maven surefire configuration


yschimke:
https://github.com/apache/maven-surefire/blob/5f6e1bd3af79acde6eeeb71477618b5644622a9b/surefire-providers/surefire-junit4/src/main/java/org/apache/maven/surefire/junit4/JUnit4Provider.java#L107

n.b. I haven't seen this for gradle, and the current one doesn't seem to fire in gradle

yschimke:
Configured here 


    jvmArgs += "-Dlistener=okhttp3.testing.InstallUncaughtExceptionHandlerListener"

yschimke:
One fix is adopting something like this 

https://github.com/Nordstrom/JUnit-Foundation#support-for-standard-junit-runlistener-providers

ZacSweers:
I've been trying that option out and have not had any luck :/

Another solution is to convert the runlistener to a `TestWatcher` rule and have a base test class for all tests to inherit from. Not great, but better than nothing.

ZacSweers:
Branch of what I've tried here: https://github.com/ZacSweers/okhttp/tree/z/testingJunitFoundation

TestWatcher impl

```java
package okhttp3.testing;

import java.io.PrintWriter;
import java.io.StringWriter;
import java.util.LinkedHashMap;
import java.util.Map;
import org.junit.rules.TestWatcher;
import org.junit.runner.Description;
import org.junit.runners.model.Statement;

/**
 * A {@link org.junit.rules.TestWatcher} used to install an aggressive default {@link
 * java.lang.Thread.UncaughtExceptionHandler} similar to the one found on Android. No exceptions
 * should escape from OkHttp that might cause apps to be killed or tests to fail on Android.
 */
public class InstallUncaughtExceptionHandlerListener extends TestWatcher {

  private Thread.UncaughtExceptionHandler oldDefaultUncaughtExceptionHandler;
  private Description lastTestStarted;
  private final Map<Throwable, String> exceptions = new LinkedHashMap<>();

  @Override protected void starting(Description description) {
    System.err.println("Installing aggressive uncaught exception handler");
    oldDefaultUncaughtExceptionHandler = Thread.getDefaultUncaughtExceptionHandler();
    Thread.setDefaultUncaughtExceptionHandler((thread, throwable) -> {
      StringWriter errorText = new StringWriter(256);
      errorText.append("Uncaught exception in OkHttp thread \"");
      errorText.append(thread.getName());
[...]

PR #5189, opened 2019-06-15
Make OkHttpClientTestRule more powerful.

swankjesse:
I'm hoping to follow this up with a change so that this rule can also do the
uncaught exception handling that we're currently doing elsewhere.

See https://github.com/square/okhttp/issues/4894


okhttp-testing-support/src/main/java/okhttp3/OkHttpClientTestRule.kt
```diff
@@ -0,0 +1,77 @@
...
+    assertThat(connectionPool.idleConnectionCount()).isEqualTo(0)
+  }
+
+  override fun apply(base: Statement, description: Description): Statement {
+    return object : Statement() {
+      override fun evaluate() {
+        try {
+          base.evaluate()
+        } finally {
+          ensureAllConnectionsReleased()
```
yschimke: This pretty much kills off concurrency in our test running, running this against a JVM singleton connectionPool.  Perhaps there is a way we can use internal access to the connection pool to count/cleanup only the relevant connections. Since connections are tied to Address, even making Dns a non singleton and tracking based on object identity would allow you to filter.

swankjesse: Hmmmm. Yep, lemme to that.

swankjesse: I changed it up to pool full OkHttpClient instances, which allows us concurrency and safety.

okhttp/src/test/java/okhttp3/ConnectionReuseTest.java
```diff
@@ -41,7 +42,11 @@
   @Rule public final OkHttpClientTestRule clientTestRule = new OkHttpClientTestRule();
 
   private HandshakeCertificates handshakeCertificates = localhost();
-  private OkHttpClient client = clientTestRule.client;
+  private OkHttpClient client;
+
+  @Before public void setUp() {
```
yschimke: This is really unfortunate, adding the setup methods.  I wonder whether we can avoid this boilerplate.  What is the effect of only replacing `.client` with `.newClient()`?  Does it negate pooling by eagerly creating a client for every test before running them?

Could we go the opposite way?  remove the field as a cache of the client and use the rule as the cache?  The cache knows both a) the test it is installed for, b) the lifecycle of the test.  And could have an optional param for second/third client instances?

yschimke: I'm more concerned if we start doing more of this and having setup in every test.  Currently the change looks like a big net positive

okhttp/src/test/java/okhttp3/CacheTest.java
```diff
@@ -1109,7 +1109,7 @@ private void assertNonIdentityEncodingCached(MockResponse response) throws Excep
   }
 
   @Test public void conditionalCacheHitIsNotDoublePooled() throws Exception {
-    TestUtil.ensureAllConnectionsReleased(client);
+    clientTestRule.ensureAllConnectionsReleased();
```
yschimke: We shouldn't need this line, it should be enforced internally already?

okhttp/src/test/java/okhttp3/ConnectionReuseTest.java
swankjesse: Let's use the rule as a cache. That's a nice improvement.

yschimke:
Are you planning to land this? I'd like to build on top of it.

swankjesse:
Lemme land as-is and circle back to avoid the setUp methods.

swankjesse:
Not creating clients in setUp() should be easier soon. I need to change the mechanism we use to create threads so each client is less expensive.
https://github.com/square/okhttp/pull/5485/files

PR #5190, opened 2019-06-16
Hide Kotlin-internal APIs from Java

monkey-mas:
Define `@file:JvmName(Xxx)` to hide Kotlin internal APIs instead of using a translated class name/path, e.g. `XxxKt`.

- Kotlint-internal API lowercase files.
`$ find . -name "[a-z]*\.kt" | xargs -I {}`
```
./okhttp-dnsoverhttps/src/main/java/okhttp3/dnsoverhttps/package-info.kt
./okhttp-logging-interceptor/src/main/java/okhttp3/logging/utf8.kt
./okhttp/src/main/java/okhttp3/internal/http/dates.kt
./okhttp/src/main/java/okhttp3/internal/internal.kt
./okhttp/src/main/java/okhttp3/internal/hostnames.kt
```

- Kotlin classes that doesn't have keywords starting from `class`, `interface`, `object`, `abstract`, `open`, `internal`, `enum`, and `data`.
`$ find . -name "*\.kt" | xargs -I {} grep -iL -e '^CLASS' -e '^INTERFACE' -e '^OBJECT' -e '^ABSTRACT' -e '^OPEN' -e '^INTERNAL' -e '^ENUM' -e '^DATA'  {}`
```
./okhttp-dnsoverhttps/src/main/java/okhttp3/dnsoverhttps/package-info.kt
./okhttp/build/generated/sources/java-templates/java/main/okhttp3/internal/Version.kt
./okhttp/src/main/java/okhttp3/internal/http/HttpHeaders.kt
./okhttp/src/main/java/okhttp3/internal/internal.kt
./okhttp/src/main/java/okhttp3/internal/hostnames.kt
./okhttp/src/main/java/okhttp3/internal/Util.kt
./okhttp/src/main/java-templates/okhttp3/internal/Version.kt
```


okhttp-dnsoverhttps/src/main/java/okhttp3/dnsoverhttps/package-info.kt
```diff
@@ -1,2 +1,20 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 /** A DNS over HTTPS implementation for OkHttp. */
+@file:JvmName("PackageInfo")
+
 package okhttp3.dnsoverhttps
```
monkey-mas: Do we need this kt class?

okhttp/src/main/java-templates/okhttp3/internal/Version.kt
```diff
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 @file:JvmName("Version")
```
monkey-mas: Should we use lowercase file name, e.g., `version.kt`, instead of `Version.kt`?

okhttp-dnsoverhttps/src/main/java/okhttp3/dnsoverhttps/package-info.kt
yschimke: I'd delete.  it technically provides a package comment, or it did.  But it was mainly used for carrying the non null by default annotation we used pre-kotlin.

okhttp-logging-interceptor/src/main/java/okhttp3/logging/utf8.kt
```diff
@@ -13,6 +13,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+@file:JvmName("UTF8")
```
yschimke: Nice catch, we should have some test that covers this public API.

okhttp/src/main/java-templates/okhttp3/internal/Version.kt
yschimke: We should have some test that covers this public API.  Please add a test.

yschimke:
if you look at generated class files, do we have any other *Kt.class classes?

okhttp-logging-interceptor/src/main/java/okhttp3/logging/utf8.kt
swankjesse: This name isn't hidden from Java.

swankjesse: My preference is to do this as we do in Okio.
https://github.com/square/okio/blob/master/okio/src/jvmMain/kotlin/okio/-Platform.kt#L17

okhttp/src/main/java-templates/okhttp3/internal/Version.kt
swankjesse: I don't want to change any `internal.*` packages, only internal-visibility in public packages.

monkey-mas: Ah, just to confirm, we don't change anything for `internal.*` packages(kotlin files), e.g., hostnames.kt and dates.kt as well? Will undo the changes if you meant so :)

okhttp-logging-interceptor/src/main/java/okhttp3/logging/utf8.kt
monkey-mas: - Got it. Looks like the file name could be `-UTF8.kt`?
- Seems like test is nicely written in `HttpLoggingInterceptor` and `IsProbablyUtf8Test` :)

okhttp-dnsoverhttps/src/main/java/okhttp3/dnsoverhttps/package-info.kt
monkey-mas: deleted!

monkey-mas:
>  if you look at generated class files, do we have any other *Kt.class classes?

@yschimke Seems like no. There're currently only two candidates in test.
```
$ find . -name "*\.java" | xargs -I {} grep -e "^import.*Kt" {}
import static okhttp3.internal.HostnamesKt.toCanonicalHost;
import static okhttp3.internal.http.DatesKt.MAX_DATE;
```

yschimke:
I was more interested in classfiles we include in the Jar that are exposed than just what is covered by tests

swankjesse:
I don't think this is the approach we want.

PR #5195, opened 2019-06-18
Correctly calculate HTTP2 response body size

### LINKED ISSUE #4948: HEAD requests break if the response contains the content-length header over HTTP2
This is similar to #2998 except it only happens on HTTP2 requests. The following example with okhttp `3.14.1` shows the issue:

````java
@Test
public void headOverHttp2() throws IOException {
    try (Response resp = client.newCall(
        new Request.Builder()
            .head()
            .url("https://http2.akamai.com/demo")
            .build()
    ).execute()) {
        System.out.println("resp.protocol() = " + resp.protocol().toString());
        System.out.println("resp.header(\"content-length\") = " + resp.header("content-length"));
        System.out.println("resp.body().contentLength() = " + resp.body().contentLength());
        System.out.println("resp.body().bytes().length = " + resp.body().bytes().length);
    }
}
````

I get this output:

	resp.protocol() = h2
	resp.header("content-length") = 2421
	resp.body().contentLength() = 2421

	java.io.IOException: Content-Length (2421) and stream length (0) disagree
		at okhttp3.ResponseBody.bytes(ResponseBody.java:139)

It is `resp.body().bytes().length` that throws the exception. When not using HTTP2, the `resp.body().contentLength()` correctly returns 0, as does `resp.body().bytes().length`.

### PR DESCRIPTION
**Pull request by monkey-mas**
Fixes #4948

Currently response body size is determined by `Content-Length` header value, i.e. we use the header value if it exists, otherwise set -1.

This doesn't work for some HTTP responses. For example, `response.body().contentLength()` should return 0 for a response of HEAD request instead of using the `Content-Length` value. Also it should return 0 for a 204 responses instead of providing -1 when Content-Length header is not present.

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/http2/Http2ExchangeCodec.kt
+++ b/okhttp/src/main/java/okhttp3/internal/http2/Http2ExchangeCodec.kt
@@ -27,6 +27,7 @@ import okhttp3.internal.http.ExchangeCodec
 import okhttp3.internal.http.RequestLine
 import okhttp3.internal.http.StatusLine
 import okhttp3.internal.http.StatusLine.Companion.HTTP_CONTINUE
+import okhttp3.internal.http.promisesBody
 import okhttp3.internal.http2.Header.Companion.RESPONSE_STATUS_UTF8
 import okhttp3.internal.http2.Header.Companion.TARGET_AUTHORITY
 import okhttp3.internal.http2.Header.Companion.TARGET_AUTHORITY_UTF8
@@ -106,7 +107,10 @@ class Http2ExchangeCodec(
   }
 
   override fun reportedContentLength(response: Response): Long {
-    return response.headersContentLength()
+    return when {
+      !response.promisesBody() -> 0L
+      else -> response.headersContentLength()
+    }
   }
 
   override fun openResponseBodySource(response: Response): Source {
--- a/okhttp/src/test/java/okhttp3/internal/http2/HttpOverHttp2Test.java
+++ b/okhttp/src/test/java/okhttp3/internal/http2/HttpOverHttp2Test.java
@@ -187,6 +187,53 @@ public final class HttpOverHttp2Test {
         (server.getHostName() + ":" + server.getPort()));
   }
 
+  @Test public void get204Response() throws Exception {
+    MockResponse responseWithoutBody = new MockResponse();
+    responseWithoutBody.status("HTTP/1.1 204");
+    responseWithoutBody.removeHeader("Content-Length");
+    server.enqueue(responseWithoutBody);
+
+    Call call = client.newCall(new Request.Builder()
+        .url(server.url("/foo"))
+        .build());
+    Response response = call.execute();
+
+    // Body contains nothing.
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: okhttp/src/main/java/okhttp3/internal/http2/Http2ExchangeCodec.kt [hunk 1]
```diff
@@ -106,7 +107,10 @@ class Http2ExchangeCodec(
   }
 
   override fun reportedContentLength(response: Response): Long {
-    return response.headersContentLength()
+    return when {
+      !response.promisesBody() -> 0L
+      else -> response.headersContentLength()
+    }
```
monkey-mas: I think we don't have to handle chunked (Transfer-Encoding) response unlike HTTP1.

https://tools.ietf.org/html/rfc7540#section-8.1 says `HTTP/2 uses DATA frames to carry message payloads. The "chunked" transfer encoding defined in Section 4.1 of [RFC7230] MUST NOT be used in HTTP/2.`

okhttp/src/test/java/okhttp3/internal/http2/HttpOverHttp2Test.java
```diff
@@ -187,6 +187,53 @@ private OkHttpClient buildHttp2Client() {
...
+        .head()
+        .url(server.url("/foo"))
+        .build());
+
+    Response response = call.execute();
+
+    // Body contains nothing.
+    assertThat(response.body().bytes().length).isEqualTo(0);
+    assertThat(response.body().contentLength()).isEqualTo(0);
+
```
monkey-mas: `assertThat(response.body().string()).isEqualTo("");` triggered an exception. Is this expected?
```
java.lang.IllegalStateException: closed

	at okio.RealBufferedSource.select(RealBufferedSource.kt:83)
	at okhttp3.internal.Util.readBomAsCharset(Util.kt:256)
	at okhttp3.ResponseBody.string(ResponseBody.kt:172)
	at okhttp3.internal.http2.HttpOverHttp2Test.head(HttpOverHttp2Test.java:229)
	at okhttp3.OkHttpClientTestRule$1.evaluate(OkHttpClientTestRule.java:37)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.lang.Thread.run(Thread.java:834)
```

okhttp/src/test/java/okhttp3/internal/http2/HttpOverHttp2Test.java
```diff
@@ -187,6 +187,53 @@ private OkHttpClient buildHttp2Client() {
         (server.getHostName() + ":" + server.getPort()));
   }
```
monkey-mas: Can we write other integration/unit tests?

okhttp/src/test/java/okhttp3/internal/http2/HttpOverHttp2Test.java
swankjesse: Yep, you can’t request the body multiple times.

swankjesse:
Thanks! Sorry for dropping the ball on this one, it’s a nice improvement!

PR #5202, opened 2019-06-20
Use MkDocs to generate website

ajalt:
As discussed on Slack, this PR sets up [MkDocs](https://squidfunk.github.io/mkdocs-material/) static site generation for both user documentation and API docs. 

The generated site is configured in the new `mkdocs.yml` file; I set it up with reasonable settings and the extensions that I think are worth using, but you will probably want to tweak it.

All files for the site live in the `docs/` directory. You can change this by setting [`docs_dir`](https://www.mkdocs.org/user-guide/configuration/#docs_dir) in the yaml file.

The API docs are generated by Dokka with the command `./gradlew :okhttp:dokka`. I chose not to commit them, but you could do so if you want people to be able to browse them through github.

I moved most of the MD files that came from the wiki into the website; you'll probably want to reorganize or edit them outside of this PR. You could also move them out of the website entirely and just use it for API docs if you wanted to. The navbar is also configured in `mkdocs.yml`. Right now I have one entry for each MD file, one for stack overflow, and one for the API docs. MkDocs also support a level of nesting in the navbar, which is useful if you want to link to sections inside of a page. Check out [Clikt's nav setup](https://github.com/ajalt/clikt/blob/4db82a772ed61e12681017f1dcb4c2d0b1235a6e/mkdocs.yml#L38) to see what that looks like. 

I also moved the existing site index page into markdown in a format that MkDocs can work with, but I'm sure it needs more work.

Right now I only have the main `okhttp` module set up to generate API docs for the site, but of course you can add other modules if you want to.

If you're working on the site, you can serve it locally with `mkdocs serve`. That will start a server at `http://127.0.0.1:8000/` that has auto-reload whenever you edit a file. You can generate the API docs for the site with `./gradlew :okhttp:dokka`, but that will slow down site generation quite a bit since there's so many files.

Finally, I updated `deploy_website.sh` so that it will work with the new setup.


swankjesse:
This is a very generous contribution. Thank you!

build.gradle
```diff
@@ -156,20 +156,24 @@ subprojects { project ->
...
-      suppress = true
-    }
-    if (project.file('Module.md').exists()) {
-      includes = ['Module.md']
-    }
-    externalDocumentationLink {
-      url = new URL("https://square.github.io/okio/2.x/okio/jvm/index.html")
-      packageListUrl = new URL("https://square.github.io/okio/2.x/okio/jvm/package-list")
+  // We have to set the dokka configuration after evaluation since the com.vanniktech.maven.publish
+  // plugin overwrites our dokka configuration on projects where it's applied.
```
swankjesse: handy comment

docs/calls.md
```diff
@@ -46,17 +45,8 @@ Calls are executed in one of two ways:
...
+## Dispatch
 
 For synchronous calls, you bring your own thread and are responsible for managing how many simultaneous requests you make. Too many simultaneous connections wastes resources; too few harms latency.
 
 For asynchronous calls, [`Dispatcher`](http://square.github.io/okhttp/3.x/okhttp/okhttp3/Dispatcher.html) implements policy for maximum simultaneous requests. You can set maximums per-webserver (default is 5), and overall (default is 64).
-
-
------
-
-<table><tr>
```
swankjesse: Yay!

docs/security.md
```diff
@@ -3,12 +3,13 @@ Security Policy
...
-| 4.x     | :x:                | Not released.  |
-| 3.14.x  | :white_check_mark: |                |
-| 3.12.x  | :white_check_mark: | Android 2.3+ (API level 9+) and Java 7+.  Platforms may not support TLSv1.2. Until December 31, 2020  |
-| 2.7.x   | :x:                | Not supported. |
+| Version | Supported  | Notes          |
+| ------- | ---------- | -------------- |
+| 4.x     | ❌         | Not released.  |
+| 3.14.x  | ✅         |                |
+| 3.12.x  | ✅         | Android 2.3+ (API level 9+) and Java 7+.  Platforms may not support TLSv1.2. Until December 31, 2020  |
+| 2.7.x   | ❌         | Not supported. |
```
swankjesse: sweet

PR #5207, opened 2019-06-20
Picocli

yschimke:
Better mix of features and better for testing graal nativeImage


yschimke:
image

yschimke:
`./okcurl --completionScript "a"` will generate a bash tab completion script

yschimke:
@swankjesse Thoughts on this?  It mainly helps as a testbed for confirming graal compatibility for OkHttp.  That is useful for me, I ended up porting a CLI client to python requests and I'd prefer to stay within a single codebase.

okcurl/okcurl
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh -e
+
+../gradlew -q --console plain installDist
```
swankjesse: What does this do?

build.gradle
```diff
@@ -18,11 +17,12 @@ buildscript {
       'kotlin': '1.3.31',
       'moshi': '1.8.0',
       'okio': '2.2.2',
-      'ktlint': '0.31.0'
+      'ktlint': '0.31.0',
+      'picocli': '4.0.0-beta-1b'
```
swankjesse: Should we wait for a proper release?

yschimke: Sounds good, I'll put this on a branch or similar until then.  Test with that.

okcurl/okcurl
yschimke: silently builds (with gradle caching etc), so that you can run "./okcurl https://..." from a git checkout and it *just* works

PR #5210, opened 2019-06-20
Kotlin 1.3.40 upgrade

yschimke:


yschimke:
Was mainly interested for Uint but I believe that is still experimental.

Issue #4703, opened 2019-03-10
When TLS uses pre-shared keys the Handshake has no peer certs


swankjesse:
Java 11’s TLS stack [automatically supports](https://bugs.openjdk.java.net/browse/JDK-8049402) pre-shared keys when attempting a repeat TLS connection to a server.

Unfortunately these TLS handshakes don’t have peer certificates and `SSLSession.getPeerCertificates()` throws.

Without the peer certificates other things break including certificate pinning.

```
java.util.NoSuchElementException
	at java.base/java.util.ArrayDeque.removeFirst(ArrayDeque.java:363)
	at okhttp3.internal.tls.BasicCertificateChainCleaner.clean(BasicCertificateChainCleaner.java:58)
	at okhttp3.CertificatePinner.check(CertificatePinner.java:166)
	at okhttp3.internal.connection.RealConnection.connectTls(RealConnection.java:358)
	at okhttp3.internal.connection.RealConnection.establishProtocol(RealConnection.java:300)
```

This is reproduced by `CallTest.matchingPinnedCertificate()` on Java 11.

swankjesse:
I submitted a bug report to OpenJDK using this isolated test:
https://gist.github.com/swankjesse/589e3e1c4a566e4a9aaec66f0f3a6223

pol12va:
@swankjesse 
We've faced strange exception after updating to 3.14.2 version and migrating to Java 11
from time to time requests fail with the exception:
```
Caused by: javax.net.ssl.SSLPeerUnverifiedException: Hostname graph.facebook.com not verified (no certificates)
	at okhttp3.internal.connection.RealConnection.connectTls(RealConnection.java:353) ~[okhttp-3.14.2.jar:?]
	at okhttp3.internal.connection.RealConnection.establishProtocol(RealConnection.java:300) ~[okhttp-3.14.2.jar:?]
	at okhttp3.internal.connection.RealConnection.connect(RealConnection.java:185) ~[okhttp-3.14.2.jar:?]
	at okhttp3.internal.connection.ExchangeFinder.findConnection(ExchangeFinder.java:224) ~[okhttp-3.14.2.jar:?]
	at okhttp3.internal.connection.ExchangeFinder.findHealthyConnection(ExchangeFinder.java:108) ~[okhttp-3.14.2.jar:?]
	at okhttp3.internal.connection.ExchangeFinder.find(ExchangeFinder.java:88) ~[okhttp-3.14.2.jar:?]
[...]

PR #5212, opened 2019-06-20
Testing with JDK 12 and latest 11.0.3

yschimke:
Intention of the change is to support JDK 12, and rather than skipping the test, confirm it fails as expected.  But 11.0.3 appears to have fixed a number of issues.


yschimke:
https://jdk.java.net/13/release-notes

Reimplement the Legacy Socket API (JDK-8221481)
Update the default enabled cipher suites preference (JDK-8163326)
Use server cipher suites preference by default (JDK-8168261)
Deprecated javax.security.cert APIs with forRemoval=true (JDK-8160247)

okhttp-testing-support/src/main/java/okhttp3/FakeSSLSession.java:import javax.security.cert.X509Certificate;

okhttp/src/test/java/okhttp3/CallTest.java
```diff
@@ -1396,9 +1397,6 @@ private void postBodyRetransmittedAfterAuthorizationFail(String body) throws Exc
   }
 
   @Test public void matchingPinnedCertificate() throws Exception {
-    // TODO https://github.com/square/okhttp/issues/4703
-    assumeFalse(getJvmSpecVersion().equals("11"));
```
yschimke: Fixed in 11.0.3, will close out #4703 if confirmed

okhttp/src/test/java/okhttp3/internal/tls/CertificatePinnerChainValidationTest.java
```diff
@@ -183,7 +185,7 @@
 
   @Test public void unrelatedPinnedLeafCertificateInChain() throws Exception {
     // https://github.com/square/okhttp/issues/4729
-    assumeFalse(getJvmSpecVersion().matches("1[123]"));
+    jdkMatchRule.expectFailure(fromMajor(11), isA(KeyStoreException.class));
```
yschimke: Draft API for setting test expectations

swankjesse:
Wow, that’s a lot of infrastructure changes.

yschimke:
> Wow, that’s a lot of infrastructure changes.

Hope your talking about JDK 13.

Within the scope of this PR, I just wanted to test against the latest JDK 12 (maybe 13) in CircleCI, and I hit the bugs fixed in 11.0.2 and 11.0.3, plus whether they still exist in 12.0.1 etc.  So I wanted to stop skipping tests and instead confirm that bugs exist (regressions in regressions).

I don't think we should concern ourselves with tests running in 11.0.1, because developers can upgrade, so once fixed we should just remove all trace of test code handling weird cases like the PSK or session resumption failures.

.circleci/config.yml
```diff
@@ -191,10 +205,8 @@ workflows:
             branches:
               ignore:
                 - gh-pages
-      - testjdk11:
-          filters:
-            branches:
-              only: master
+      - testjdk11
```
yschimke: Will revert and filter only on master before landing

okhttp/src/test/java/okhttp3/CallTest.java
```diff
@@ -1315,7 +1321,7 @@ private void postBodyRetransmittedAfterAuthorizationFail(String body) throws Exc
 
     // The _anon_ suites became unsupported in "1.8.0_201" and "11.0.2".
     assumeFalse(System.getProperty("java.version", "unknown").matches("1\\.8\\.0_1\\d\\d"));
-    assumeFalse(System.getProperty("java.version", "unknown").matches("11"));
+    jdkMatchRule.expectFailure(fromMajor(11), anything());
```
yschimke: This seems to pass for me... I would like to understand why.

okhttp/src/test/java/okhttp3/CallTest.java
```diff
@@ -1315,7 +1321,7 @@ private void postBodyRetransmittedAfterAuthorizationFail(String body) throws Exc
 
     // The _anon_ suites became unsupported in "1.8.0_201" and "11.0.2".
     assumeFalse(System.getProperty("java.version", "unknown").matches("1\\.8\\.0_1\\d\\d"));
-    assumeFalse(System.getProperty("java.version", "unknown").matches("11"));
```
yschimke: This matches was probably overly specific
```

jshell> System.getProperty("java.version", "unknown")
$1 ==> "12.0.1"

jshell> System.getProperty("java.version", "unknown").matches("12")
$2 ==> false
```

yschimke:
See test failures in JDK 11
https://circleci.com/gh/square/okhttp/3516?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

JDK 12
https://circleci.com/gh/square/okhttp/3519?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

New in JDK12 

missingPongsFailsConnection[h2] - okhttp3.internal.http2.HttpOverHttp2Test
org.junit.runners.model.TestTimedOutException: test timed out after 5 seconds

yschimke:
I'd like to add an easy way to automatically log http client listener events for failing tests.  Ideally by default OkHttpClientTestRule would cache all listener events, and if a test is failing would print for informational purposes to System.err.  So we get some useful information out of failing CI tests.

yschimke:
image

.circleci/config.yml
```diff
@@ -154,6 +154,21 @@ jobs:
       - runtests:
           platform: jdk9
 
+  testjdk12:
+    docker:
+      # best source of JDK 12 for now
```
swankjesse: 🙀

okhttp/src/test/java/okhttp3/URLConnectionTest.java
```diff
@@ -685,6 +688,10 @@ private void connectViaHttpsReusingConnections(boolean rebuildClient) throws Exc
    * http://code.google.com/p/android/issues/detail?id=13178
    */
   @Test public void connectViaHttpsToUntrustedServer() throws Exception {
+    // https://github.com/square/okhttp/issues/5222
+    jdkMatchRule.expectFailure(platform.platformMatcher(PlatformRule.CONSCRYPT_PROPERTY),
```
swankjesse: I find this syntax perplexing

okhttp/src/test/java/okhttp3/internal/http2/HttpOverHttp2Test.java
```diff
@@ -104,11 +110,13 @@
     return asList(Protocol.H2_PRIOR_KNOWLEDGE, Protocol.HTTP_2);
   }
 
-  @Rule public final PlatformRule platform = new PlatformRule();
+  private JdkMatchRule jdkMatchRule = new JdkMatchRule();
+  private PlatformRule platform = new PlatformRule();
+  @Rule public final TestRule chain =
```
swankjesse: Don’t they chain on their own?

okhttp/src/test/java/okhttp3/internal/tls/CertificatePinnerChainValidationTest.java
```diff
@@ -120,7 +128,7 @@
   /** The pinner should accept an intermediate from the server's chain. */
   @Test public void pinIntermediatePresentInChain() throws Exception {
     // TODO https://github.com/square/okhttp/issues/4703
-    assumeFalse(getJvmSpecVersion().equals("11"));
+    //jdkMatchRule.expectFailure(fromMajor(11), anything());
```
swankjesse: why commented out code?

okhttp/src/test/java/okhttp3/internal/tls/ClientAuthTest.java
```diff
@@ -234,7 +238,7 @@ public void setUp() {
   @Test public void invalidClientAuthFails() throws Throwable {
     // TODO https://github.com/square/okhttp/issues/4598
     // StreamReset stream was reset: PROT...
-    assumeFalse(getJvmSpecVersion().matches("1[123]"));
+    //jdkMatchRule.expectFailure(fromMajor(11), anything());
```
swankjesse: I think I want different syntax in tests, even if it makes the JdkMatchRule more complex. Maybe this?

```
  @Rule public final JdkMatchRule jdkMatchRule = new JdkMatchRule();

  @Test public void invalidClientAuthFails() throws Throwable {
    jdkMatchRule.assumeJdkLessThan(12)
  }
```

swankjesse: The syntax of Hamcrest matchers seems so awkward

okhttp/src/test/java/okhttp3/internal/tls/CertificatePinnerChainValidationTest.java
monkey-mas: Yeah, TLS issue should be resolved after OpenJDK 11.0.3. I guess we can safely remove the TODO comment and `expectFailure` related to 4703? :)

okhttp-testing-support/src/main/java/okhttp3/testing/JdkMatchRule.kt
```diff
@@ -0,0 +1,105 @@
...
+        val description = StringDescription()
+        versionMatcher.describeTo(description)
+        description.appendText(" expected to fail with exception that ")
+        failureMatcher.describeTo(description)
+
+        fail(description.toString())
+      }
+    }
+  }
+}
```
monkey-mas: nit: new line :)

okhttp-testing-support/src/test/kotlin/okhttp3/testing/JdkMatchRuleTest.kt
```diff
@@ -0,0 +1,41 @@
...
+    jdkMatchRule.expectFailure(fromMajor(VersionInfo.majorVersion + 1))
+  }
+
+  @Test
+  fun failureCase() {
+    jdkMatchRule.expectFailure(fromMajor(VersionInfo.majorVersion))
+
+    check(false)
+  }
+}
```
monkey-mas: nit: new line :)

okhttp/src/test/java/okhttp3/internal/http2/HttpOverHttp2Test.java
yschimke: Predictably and stable, yes!? But just not intuitively e.g. field ordering.  I suspect by design if each Rule only affects the test execution you are grand.  But here the Rules can fail e.g. timeout, and we want those failures to be seen or not seen by other rules.

okhttp/src/test/java/okhttp3/internal/tls/CertificatePinnerChainValidationTest.java
yschimke: Happy to remove.  I might leave the comment without "TODO" and remove the dead code.  Essentially someone running with 11.0.1 will see test failures and I don't want those reported.

okhttp/src/test/java/okhttp3/internal/tls/ClientAuthTest.java
yschimke: I specifically designed to not use assumptions.  I wanted failures and to verify the failures on a secondary path, such that tests are not skipped or flaky, but understood including when they start passing because of fixes upstream.

okhttp/src/test/java/okhttp3/URLConnectionTest.java
yschimke: I like this simpler syntax you suggested.  Also I'll merge with PlatformTestRule

jdkMatchRule.assumeJdkLessThan(12)

yschimke: Have left a few helper methods, but cleaned it up to be externally understable I hope.

yschimke:
@swankjesse I assumed I was waiting on a shipit on this, but I'm unsure about any outstanding concerns. It's tests only, ok to ship?

okhttp-testing-support/src/main/java/okhttp3/testing/VersionInfo.kt
```diff
@@ -0,0 +1,14 @@
+package okhttp3.testing
+
+object VersionInfo {
```
swankjesse: PlatformVersion ?

okhttp-testing-support/src/main/java/okhttp3/testing/VersionInfo.kt
```diff
@@ -0,0 +1,14 @@
+package okhttp3.testing
```
swankjesse: Nit: copyright

okhttp/src/test/java/okhttp3/internal/tls/CertificatePinnerChainValidationTest.java
```diff
@@ -262,7 +261,7 @@
 
   @Test public void unrelatedPinnedIntermediateCertificateInChain() throws Exception {
     // https://github.com/square/okhttp/issues/4729
-    assumeFalse(getJvmSpecVersion().matches("1[123]"));
+    platform.expectFailure(platformMatches(CONSCRYPT_PROPERTY), KeyStoreException.class);
```
swankjesse: I still think this reads awkwardly. I'd prefer a try/catch and then an assertion rather than framework.

okhttp-testing-support/src/main/java/okhttp3/testing/PlatformRule.kt
```diff
@@ -50,12 +86,54 @@ open class PlatformRule @JvmOverloads constructor(
...
 
-  override fun after() {
+  fun resetPlatform() {
     if (platform != null) {
       Platform.resetForTests()
     }
   }
 
+  @JvmOverloads
+  fun expectFailure(
```
swankjesse: This feels like too much framework just like JUnits bad expected exceptions feature. 

Just use try/catch? Then you can isolate the expectation properly, and assert the state of things after the exception if you need to.

yschimke: Truth be told it’s less about how it’s failing. More about understanding what is failing in all the weird combinations. So I think I’ll drop the how aspect and just focus on when. So we detect that a conscrypt release or JDK 11.0.4 fixed it.

yschimke: But I also want to keep the test focused on the correct case. So ideally without an embedded try catch check env. And a time-out is thrown externally so doesn’t work for that.

okhttp-testing-support/src/main/java/okhttp3/testing/PlatformRule.kt
```diff
@@ -50,12 +86,54 @@ open class PlatformRule @JvmOverloads constructor(
...
+      }
+    }
+
+    throw e
+  }
+
+  fun failIfExpected() {
+    versionChecks.forEach { (versionMatcher, failureMatcher) ->
+      if (versionMatcher.matches(VersionInfo)) {
+        val description = StringDescription()
```
swankjesse: For follow up I’d love to find something other than Hamcrest. It’s so awkward.

okhttp-testing-support/src/main/java/okhttp3/testing/PlatformRule.kt
```diff
@@ -50,12 +86,54 @@ open class PlatformRule @JvmOverloads constructor(
...
+    versionMatcher: Matcher<out Any>,
+    expectedType: Class<*>
+  ) {
+    versionChecks.add(Pair(versionMatcher, isA(expectedType)))
+  }
+
+  fun rethrowIfNotExpected(e: Throwable) {
+    versionChecks.forEach { (versionMatcher, failureMatcher) ->
+      if (versionMatcher.matches(VersionInfo)) {
+        if (!failureMatcher.matches(e)) {
```
swankjesse: invert this if? Then you only need one `throw e`

okhttp-testing-support/src/main/java/okhttp3/testing/PlatformRule.kt
```diff
@@ -50,12 +86,54 @@ open class PlatformRule @JvmOverloads constructor(
...
+  fun expectFailure(
+    versionMatcher: Matcher<out Any>,
+    failureMatcher: Matcher<out Any> = CoreMatchers.anything()
+  ) {
+    versionChecks.add(Pair(versionMatcher, failureMatcher))
+  }
+
+  fun expectFailure(
+    versionMatcher: Matcher<out Any>,
+    expectedType: Class<*>
```
swankjesse: constrain this to Throwable subtypes?

okhttp-testing-support/src/main/java/okhttp3/testing/PlatformRule.kt
swankjesse: Okay.

okhttp-testing-support/src/main/java/okhttp3/testing/PlatformRule.kt
yschimke: Could be simple kotlin functions if we move some (non API enforcing) tests to kotlin.

yschimke:
unrelated test failure

Issue #5209, opened 2019-06-20
Response.Builder#build() not accepting zero code; misleading message


laalto:
OkHttp 4.0.0-RC1

`Response.Builder#build()` checks the response code:

```
      check(code > 0) { "code < 0: $code" }
```

given `code == 0` this would fail with `code < 0: 0` which is misleading

Earlier versions of okhttp allowed 0 as response code here:

```
if (code < 0) throw new IllegalStateException("code < 0: " + code);
```

I believe changing the check to `code >= 0` would preserve compatibility and fix message correctness.

Discovered this when trying out 4.0.0-RC1 on a codebase with unit tests that pass around fake responses without explicitly set response code.

PR #5215, opened 2019-06-21
Allow users to set status code 0 in Response to maintain backward compatibility in OkHttp4.0

monkey-mas:
Fixes https://github.com/square/okhttp/issues/5209

Previously in OkHttp3, status code == 0 was acceptable.
ref. [status code validity check in 3.14.2](https://github.com/square/okhttp/blob/parent-3.14.2/okhttp/src/main/java/okhttp3/Response.java#L453)

### DIFF
--- a/okhttp/src/main/java/okhttp3/Response.kt
+++ b/okhttp/src/main/java/okhttp3/Response.kt
@@ -433,7 +433,7 @@ class Response internal constructor(
     }
 
     open fun build(): Response {
-      check(code > 0) { "code < 0: $code" }
+      check(code >= 0) { "code < 0: $code" }
       return Response(
           checkNotNull(request) { "request == null" },
           checkNotNull(protocol) { "protocol == null" },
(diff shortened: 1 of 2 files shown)


JakeWharton:
Test to prevent regression?

monkey-mas:
> tests
Okay will do :)

okhttp/src/main/java/okhttp3/Response.kt
```diff
@@ -433,7 +433,7 @@ class Response internal constructor(
     }
 
     open fun build(): Response {
-      check(code > 0) { "code < 0: $code" }
+      check(code >= 0) { "code < 0: $code" }
```
monkey-mas: nit: I kinda feel this should probably throw `IllegalArgumentException`? Anyways, this is unrelated to backward compatibility. Let's discuss it in another issue/pr :)

JakeWharton: The function accepts no arguments though

monkey-mas: ah, correct..  Okay, so if `Builder#code(...)` threw an exception, the exception could be `IllegalArgumentException `, which Builder doesn't do. In our case, `Builder#build()` uses _state_ that was set via setters and throws `IllegalStateException`.

Hope my understanding is correct, thanks anyway :)

PR #5216, opened 2019-06-21
Change return type of MockWebServer#takeRequest(...) to nullable to maintain backward compatibility in OkHttp4.0

### LINKED ISSUE #5208: MockWebServer takeRequest with timeout should be nullable
OkHttp 4.0.0-RC1

`MockWebServer` `takeRequest()` with timeout function signature is NotNull:

```
fun takeRequest(timeout: Long, unit: TimeUnit): RecordedRequest = requestQueue.poll(timeout, unit)
```

However the underlying `LinkedBlockingQueue` `poll()` returns nulls on timeout, causing an `IllegalStateException` when a NotNull kotlin function tries to return null.

I believe the return type should be Nullable instead, e.g. 

```
fun takeRequest(timeout: Long, unit: TimeUnit): RecordedRequest? = requestQueue.poll(timeout, unit)
```

I discovered this when trying out 4.0.0-RC1 migration on a codebase that has some unit test cases testing caching functionality using `takeRequest()` with (short) timeout to make sure the feature being tested does not hit the network.

monkey-mas:
Fixes #5208 

Previously in OkHttp3, takeRequest(...) returns RecordedRequest?.
ref. [takeRequest(timeout: Long, unit: TimeUnit) in 3.14.2](https://github.com/square/okhttp/blob/parent-3.14.2/mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.java#L308)

### DIFF
--- a/mockwebserver/src/test/java/okhttp3/mockwebserver/MockWebServerTest.java
+++ b/mockwebserver/src/test/java/okhttp3/mockwebserver/MockWebServerTest.java
@@ -48,6 +48,7 @@ import org.junit.runners.model.Statement;
 
 import static java.nio.charset.StandardCharsets.UTF_8;
 import static java.util.Arrays.asList;
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
 import static java.util.concurrent.TimeUnit.NANOSECONDS;
 import static java.util.concurrent.TimeUnit.SECONDS;
 import static okhttp3.tls.internal.TlsUtil.localhost;
@@ -140,6 +141,9 @@ public final class MockWebServerTest {
     RecordedRequest request = server.takeRequest();
     assertThat(request.getRequestLine()).isEqualTo("GET / HTTP/1.1");
     assertThat(request.getHeader("Accept-Language")).isEqualTo("en-US");
+
+    // Server has no more requests.
+    assertThat(server.takeRequest(100, MILLISECONDS)).isNull();
   }
 
   @Test public void redirect() throws Exception {
--- a/okhttp/src/test/java/okhttp3/KotlinSourceModernTest.kt
+++ b/okhttp/src/test/java/okhttp3/KotlinSourceModernTest.kt
@@ -735,8 +735,8 @@ class KotlinSourceModernTest {
     mockWebServer.noClientAuth()
     mockWebServer.requestClientAuth()
     mockWebServer.requireClientAuth()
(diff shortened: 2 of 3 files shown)


mockwebserver/src/test/java/okhttp3/mockwebserver/MockWebServerTest.java
```diff
@@ -140,6 +141,9 @@
     RecordedRequest request = server.takeRequest();
     assertThat(request.getRequestLine()).isEqualTo("GET / HTTP/1.1");
     assertThat(request.getHeader("Accept-Language")).isEqualTo("en-US");
+
+    // Server has no more requests.
+    assertThat(server.takeRequest(100, MILLISECONDS)).isNull();
```
monkey-mas: Timeout could be shorter? I believe it always returns `null` when we set timeout 0.

ref. OpenJDK 11.0.3 LinkedBlockingQueue
```java
    public E poll(long timeout, TimeUnit unit) throws InterruptedException {
        final E x;
        final int c;
        long nanos = unit.toNanos(timeout);
        final AtomicInteger count = this.count;
        final ReentrantLock takeLock = this.takeLock;
        takeLock.lockInterruptibly();
        try {
            while (count.get() == 0) {
                if (nanos <= 0L)
                    return null;
                nanos = notEmpty.awaitNanos(nanos);
            }
            x = dequeue();
            c = count.getAndDecrement();
            if (c > 1)
                notEmpty.signal();
        } finally {
            takeLock.unlock();
        }
        if (c == capacity)
            signalNotFull();
        return x;
    }
```

okhttp/src/test/java/okhttp3/KotlinSourceModernTest.kt
```diff
@@ -735,8 +735,8 @@ class KotlinSourceModernTest {
     mockWebServer.noClientAuth()
     mockWebServer.requestClientAuth()
     mockWebServer.requireClientAuth()
-    var request: RecordedRequest = mockWebServer.takeRequest()
-    request = mockWebServer.takeRequest(0L, TimeUnit.SECONDS)
+    val request: RecordedRequest = mockWebServer.takeRequest()
+    val nullableRequest: RecordedRequest? = mockWebServer.takeRequest(0L, TimeUnit.SECONDS)
```
monkey-mas: timeout could be greater than 0?

PR #5217, opened 2019-06-21
Output client events after tagged @Flaky tests or failures

yschimke:
Print client events on test failure, or for known flaky tests (successful or otherwise).

Also dogfoods our listeners as our own debug tools for CI after a flaky test.


okhttp-testing-support/src/main/java/okhttp3/OkHttpClientTestRule.kt
```diff
@@ -92,5 +120,18 @@ class OkHttpClientTestRule : TestRule {
...
+    private fun freshClient(): OkHttpClient {
+      val listener = ClientRuleEventListener { client, line ->
+        clientEventsMap[client]?.add(line)
+      }
+
+      return OkHttpClient.Builder()
+          .dns(SINGLE_INET_ADDRESS_DNS) // Prevent unexpected fallback addresses.
+          .eventListener(listener)
+          .build().apply {
+            listener.client = this
```
yschimke: bit ugly, listener needs to use client as the key

okhttp/src/test/java/okhttp3/ConnectionCoalescingTest.java
```diff
@@ -109,6 +109,8 @@
     assert200Http2Response(execute(sanUrl), "san.com");
 
     assertThat(client.connectionPool().connectionCount()).isEqualTo(1);
+    
+    fail("Revert before landing");
```
yschimke: Testing the test code

okhttp-testing-support/src/main/java/okhttp3/OkHttpClientTestRule.kt
```diff
@@ -55,25 +55,51 @@ class OkHttpClientTestRule : TestRule {
...
         prototypes.push(prototype)
+        clientEventsMap.remove(prototype)
         prototype = null
       }
     }
   }
 
+  private fun logEventsIfFlaky() {
+    // TODO check if test is a known flaky test that should always log
+//    logEvents()
```
yschimke: src/test/resources/flakytests.txt?

okhttp-testing-support/src/main/java/okhttp3/OkHttpClientTestRule.kt
```diff
@@ -55,25 +55,51 @@ class OkHttpClientTestRule : TestRule {
...
+  }
+
+  private fun logEventsOnFailure(t: Throwable) {
+    logEvents()
+  }
+
+  private fun logEvents() {
+    val events = clientEventsMap[prototype]
+
+    // Will fail if test overrides the listener
```
yschimke: This is pretty common, but once a test is recording events, we should avoid causing additional noise.

yschimke:
Output https://3552-5152285-gh.circle-artifacts.com/0/home/circleci/reports/okhttp/build/reports/tests/test/classes/okhttp3.ConnectionCoalescingTest.html

image

okhttp-testing-support/src/main/java/okhttp3/ClientRuleEventListener.kt
```diff
@@ -0,0 +1,138 @@
...
+ */
+package okhttp3
+
+import java.io.IOException
+import java.net.InetAddress
+import java.net.InetSocketAddress
+import java.net.Proxy
+import java.util.concurrent.TimeUnit
+
+class ClientRuleEventListener(var logger: (OkHttpClient, String) -> Unit) : EventListener(),
```
swankjesse: Can we combine this with our other event listener that works similarly?

okhttp-testing-support/src/main/java/okhttp3/OkHttpClientTestRule.kt
```diff
@@ -55,25 +55,52 @@ class OkHttpClientTestRule : TestRule {
...
     }
   }
 
+  private fun logEventsIfFlaky(testClass: String, testMethod: String) {
+    if (flakyTests.contains(testClass) || flakyTests.contains("$testClass#$testMethod")) {
+      logEvents()
+    }
+  }
+
+  private fun logEventsOnFailure(t: Throwable) {
```
swankjesse: why not just call logEvents ?

okhttp-testing-support/src/main/java/okhttp3/OkHttpClientTestRule.kt
```diff
@@ -55,25 +55,52 @@ class OkHttpClientTestRule : TestRule {
...
+    logEvents()
+  }
+
+  private fun logEvents() {
+    val events = clientEventsMap[prototype]
+
+    // Will be ineffective if test overrides the listener
```
swankjesse: Good comment

okhttp-testing-support/src/main/java/okhttp3/OkHttpClientTestRule.kt
```diff
@@ -55,25 +55,52 @@ class OkHttpClientTestRule : TestRule {
...
+    if (events != null) {
+      println("Events (${events.size})")
+
+      events.forEach {
```
swankjesse: I prefer `for` so much. You can break!

swankjesse: (in this case you don‘t need to, but why ever use the worse one)

okhttp-testing-support/src/main/java/okhttp3/OkHttpClientTestRule.kt
```diff
@@ -82,6 +109,13 @@ class OkHttpClientTestRule : TestRule {
      */
     internal val prototypes = ConcurrentLinkedDeque<OkHttpClient>()
 
+    val clientEventsMap = mutableMapOf<OkHttpClient, MutableList<String>>()
+
+    val flakyTests: Set<String> by lazy {
+      (this::class.java.getResource("/flakytests.txt")?.readText() ?: "")
```
swankjesse: should we mark with an annotation instead?

okhttp-testing-support/src/main/java/okhttp3/OkHttpClientTestRule.kt
```diff
@@ -82,6 +109,13 @@ class OkHttpClientTestRule : TestRule {
      */
     internal val prototypes = ConcurrentLinkedDeque<OkHttpClient>()
 
+    val clientEventsMap = mutableMapOf<OkHttpClient, MutableList<String>>()
```
swankjesse: Why isn’t this just a property of the OkHttpClientTestRule?

okhttp-testing-support/src/main/java/okhttp3/ClientRuleEventListener.kt
yschimke: RecordingEventListener?  Good shout, I could just make that the default and pinch the events out.

okhttp-testing-support/src/main/java/okhttp3/OkHttpClientTestRule.kt
yschimke: I think I wanted this to also account for flaky tests... I forget :)

okhttp-testing-support/src/main/java/okhttp3/OkHttpClientTestRule.kt
yschimke: Because of static caching of the client instances, doesn't match the Rule lifecycle.

okhttp-testing-support/src/main/java/okhttp3/ClientRuleEventListener.kt
yschimke: This is tough, RecordingEventListener is mutable in body of tests, doesn't have great toString etc and LoggingEventListener is in inaccessible location.  Would prefer to defer this for now.

okhttp-testing-support/src/main/java/okhttp3/OkHttpClientTestRule.kt
swankjesse: Got it. Lets put the current `Rule` in the `EventListener.Factory`, so each `EventListener` knows which `Rule` created it.

swankjesse: Maybe something like this:

```
  class OkHttpClientTestRule {
    val lines = mutableListOf<String>()

    ...

    private fun freshClient(): OkHttpClient {
      return OkHttpClient.Builder()
          .dns(SINGLE_INET_ADDRESS_DNS) // Prevent unexpected fallback addresses.
          .eventListenerFactory(object: EventListener.Factory {
             fun create(call: Call) = ClientRuleEventListener { client, line ->
               this@OkHttpClientTestRule.events.add(line)
             }
         })
         .build()
```

yschimke: Nice, think I finally grokked it.

yschimke:
@swankjesse same, tests only.  Any concerns?

okhttp-testing-support/src/main/java/okhttp3/ClientRuleEventListener.kt
```diff
@@ -0,0 +1,135 @@
...
+ */
+package okhttp3
+
+import java.io.IOException
+import java.net.InetAddress
+import java.net.InetSocketAddress
+import java.net.Proxy
+import java.util.concurrent.TimeUnit
+
+class ClientRuleEventListener(var logger: (String) -> Unit) : EventListener(),
```
swankjesse: this is great

okhttp-testing-support/src/main/java/okhttp3/OkHttpClientTestRule.kt
```diff
@@ -107,5 +137,11 @@ class OkHttpClientTestRule : TestRule {
         return listOf(addresses[0])
       }
     }
+
+    fun freshClient(): OkHttpClient {
```
swankjesse: does this need to be public?

okhttp-testing-support/src/main/java/okhttp3/testing/Flaky.kt
```diff
@@ -0,0 +1,9 @@
+package okhttp3.testing
```
swankjesse: nit: copyright

okhttp/src/test/java/okhttp3/internal/http2/HttpOverHttp2Test.java
```diff
@@ -97,6 +98,8 @@
 
 /** Test how HTTP/2 interacts with HTTP features. */
 @RunWith(Parameterized.class)
+@Flaky(issues = {"https://github.com/square/okhttp/issues/4632",
```
swankjesse: FWIW, having two issues to track the same flaky test feels bad

swankjesse: I think I’d rather track these as normal comments because when the test ceases to be flaky we still probably want the comment link?

okhttp/src/test/java/okhttp3/internal/ws/WebSocketHttpTest.java
```diff
@@ -644,8 +648,9 @@
     latch.countDown();
 
     long elapsedUntilFailure = System.nanoTime() - openAtNanos;
-    assertThat((double) TimeUnit.NANOSECONDS.toMillis(elapsedUntilFailure)).isCloseTo((double) 1000, offset(
-        250d));
+    assertThat((double) TimeUnit.NANOSECONDS.toMillis(elapsedUntilFailure)).isCloseTo((double) 1000,
+        offset(
+            250d));
```
swankjesse: nit: this wrapping is awkward

okhttp/src/test/java/okhttp3/internal/http2/HttpOverHttp2Test.java
yschimke: I don't think it's always two test tracking the bug.  It's often two valid tests failing due to the same bug.  I can move out the issues to comments.

Reification is a big theme for my thinking at the moment, so having it introspectible via code was interesting.  

image

okhttp-testing-support/src/main/java/okhttp3/ClientRuleEventListener.kt
yschimke: I didn't think your review comments were achievable, but it ended up really nice and clean.

okhttp/src/test/java/okhttp3/internal/ws/WebSocketHttpTest.java
```diff
@@ -644,8 +647,8 @@
     latch.countDown();
 
     long elapsedUntilFailure = System.nanoTime() - openAtNanos;
-    assertThat((double) TimeUnit.NANOSECONDS.toMillis(elapsedUntilFailure)).isCloseTo((double) 1000, offset(
-        250d));
+    assertThat(TimeUnit.NANOSECONDS.toMillis(elapsedUntilFailure))
+        .isCloseTo(1000L, offset(250L));
```
swankjesse: thanks

PR #5220, opened 2019-06-22
Move the API docs from /okhttp/api to /okhttp/4.x

swankjesse:


swankjesse:
Merging doc change optimistically. Review welcome!

Issue #5223, opened 2019-06-22
this can be null (Util.closeQuietly)


yschimke:
When called from java, this can be null

```
okhttp3.internal.http2.Http2ConnectionTest > socketExceptionWhileWritingHeaders FAILED
    java.lang.IllegalArgumentException: Parameter specified as non-null is null: method okhttp3.internal.Util.closeQuietly, parameter $this$closeQuietly
        at okhttp3.internal.Util.closeQuietly(Util.kt)
        at okhttp3.internal.http2.MockHttp2Peer.close(MockHttp2Peer.java:186)
        at okhttp3.internal.http2.Http2ConnectionTest.tearDown(Http2ConnectionTest.java:70)
2295 tests completed, 1 failed, 188 skipped
```

monkey-mas:
#5224 looks good to me but I'm not 100% sure why the socket can be null after the test run. 🤔

vickyleu:
4 Caused by:
--
5 java.lang.IllegalArgumentException:Parameter specified as non-null is null: method kotlin.jvm.internal.Intrinsics.checkParameterIsNotNull, parameter $this$closeQuietly
6 okhttp3.internal.Util.closeQuietly(Unknown Source:2)
7 com.baijiayun.playback.dataloader.PlayerDataLoader.a(SourceFile:178)
8 com.baijiayun.playback.dataloader.PlayerDataLoader.lambda$QwSiiiasslIm5mZ0BFT1stIH9m0(Unknown Source:0)

########## aar file compile used okhttp3.8.1 java version,app project use okhttp4.0.1 kotlin version

######### aar code 
public Observable<File> a(String var1, File var2) {
        return Observable.create((var3) -> {
            BJNetCall var4 = this.C.newDownloadCall(var1, var2, (Map)null);
            BufferedSink var5 = null;
            File var6 = var2;
            if (var2.isDirectory()) {
                var6 = new File(var2, PBUtils.md5Hex(var1));
            }

            try {
                BJResponse var7 = var4.executeSync(this);
                if (var7 != null) {
                    if (var7.isSuccessful()) {
                        var5 = Okio.buffer(Okio.sink(var6));
                        var5.writeAll(var7.getResponse().body().source());
                        var3.onNext(var6);
                    } else {
                        LPRxUtils.onError(var3, new LPError(var7.code(), var7.message()));
                    }

                    return;
                }

                LPRxUtils.onError(var3, this.E);
            } catch (FileNotFoundException var12) {
[...]

PR #5224, opened 2019-06-22
Allow nullable on Util.close

yschimke:
From Java can be null

https://github.com/square/okhttp/issues/5223


okhttp/src/main/java/okhttp3/internal/Util.kt
```diff
@@ -485,9 +485,9 @@ fun Closeable.closeQuietly() {
 }
 
 /** Closes this, ignoring any checked exceptions. Does nothing if this is null. */
```
yschimke: FWIW also now matches the javadoc

Issue #5194, opened 2019-06-17
Public API to monitor and manage connections 


yschimke:
https://github.com/yschimke/OkHttpAndroidApp/blob/master/android/app/src/main/java/ee/schimke/okhttp/android/factory/Accessors.kt

Shows the internal APIs and reflection I've needed to monitor and actively manage connections e.g. soft/hard close them on network events.

Raising this to track a 4.1 API for this sort of functionality.

Want to be able to:
- enumerate the connections with some form of identity (id? or reference?), 
- Actively close or soft close a connection
- Understand network (socket? local ip?)

yschimke:
Relates to https://github.com/square/okhttp/issues/4530

yschimke:
Also relates to https://github.com/square/okhttp/pull/3855

yschimke:
@swankjesse how against this are you? I could really use this on Wear, where controlling which connections between (low bandwidth) BT, Wifi and expensive LTE is critical for apps.

I'm looking to accurately report the number of each type of connections, since there can me multiple at the same time.  We might have started a bluetooth connection for fetching artwork images, then temporarily open up LTE to download an 80mb podcast.

swankjesse:
I'm not opposed! Doing a multiplatform release with Android-specific code also gives us the opportunity to do this in a way that gives users good defaults.

My top line goal here is that if we offer a tuning API, we should also do the work to make use of it ourselves.

yschimke:
A draft for discussion here https://github.com/square/okhttp/pull/7456

PR #5225, opened 2019-06-22
[RFC] Thoughts on ConnectionPool API

yschimke:
**This is not an indication of a planned public API, just for discussion.**

https://github.com/square/okhttp/issues/5194

- Want to be able to enumerate the connections with some form of identity (id? or reference?), 
- Actively close or soft close a connection
- Understand network (socket? local ip?)

I'd argue for a midsized app, this is behaviour that should be possible e.g. evaluating whether some ConnectionPool tuning change makes a measurable improvement to request latency via higher connection reuse etc.  n.b. This is implicitly available via events on an EventListener.

### DIFF
--- a/okhttp/src/main/java/okhttp3/Call.kt
+++ b/okhttp/src/main/java/okhttp3/Call.kt
@@ -87,6 +87,8 @@ interface Call : Cloneable {
    */
   fun timeout(): Timeout
 
+  fun eventListener(): EventListener
+
   /**
    * Create a new, identical call to this one which can be enqueued or executed even if this call
    * has already been.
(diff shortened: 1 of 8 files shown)


yschimke:
Other relevant events - ping latencies would be useful to track/expose for performance reasons.

swankjesse:
I’d like to start with a design of what the inputs and outputs of the connection pool are, probably starting in a GitHub issue or markdown doc before code.

I’ve been mulling over this for a while and I’ve got a bunch of different complementary problems that I think we solve all in one go.

 * Seeding the connection pool.
 * Client-side load balancing.
 * Active DNS / quota management.

My ideal is that we can separate the complex concerns using connections from the also complex concerns of defining policy.

yschimke:
I would like to add connection pool listener, state queries (am I connected to my API server) and some simple core API that background sensitive apps could use. But I like the plan.

PR #5226, opened 2019-06-22
OkHttpClientTestRule check connectionCount instead of idle

yschimke:
Clients should be clean after use, not just from idle connections.

### DIFF
--- a/okhttp-testing-support/src/main/java/okhttp3/OkHttpClientTestRule.kt
+++ b/okhttp-testing-support/src/main/java/okhttp3/OkHttpClientTestRule.kt
@@ -44,9 +44,11 @@ class OkHttpClientTestRule : TestRule {
   }
 
   fun ensureAllConnectionsReleased() {
-    val connectionPool = prototype!!.connectionPool
-    connectionPool.evictAll()
-    assertThat(connectionPool.connectionCount()).isEqualTo(0)
+    prototype?.let {
+      val connectionPool = it.connectionPool
+      connectionPool.evictAll()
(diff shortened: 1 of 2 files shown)


yschimke:
hmmm... maybe can't land without some fixes, probably a good sign?

swankjesse:
Interesting!

okhttp/src/test/java/okhttp3/internal/ws/WebSocketHttpTest.java
```diff
@@ -84,6 +84,7 @@
     clientListener.assertExhausted();
 
     // TODO: assert all connections are released once leaks are fixed
+    clientTestRule.abandonClient(client);
```
yschimke: The comment above makes it clear why this test was failing...

The rule is still worth using here to standardise the test client creation, but we just don't end up with clean client instances.

okhttp-testing-support/src/main/java/okhttp3/OkHttpClientTestRule.kt
```diff
@@ -44,9 +44,11 @@ class OkHttpClientTestRule : TestRule {
   }
 
   fun ensureAllConnectionsReleased() {
-    val connectionPool = prototype!!.connectionPool
-    connectionPool.evictAll()
-    assertThat(connectionPool.idleConnectionCount()).isEqualTo(0)
+    prototype?.let {
```
yschimke: using let over simple if check because of a mutable field

PR #5227, opened 2019-06-22
Enable additional passing tests for conscrypt

yschimke:


okhttp/src/test/java/okhttp3/ConnectionCoalescingTest.java
```diff
@@ -302,20 +301,6 @@
     assertThat(client.connectionPool().connectionCount()).isEqualTo(1);
   }
 
-  /** Run against public external sites, doesn't run by default. */
-  @Ignore
```
yschimke: Don't need external tests with network requirements.

PR #5228, opened 2019-06-22
Fix samples and add test

yschimke:
some minor fixes and adding a test to confirm they run


samples/guide/src/main/java/okhttp3/recipes/PostMultipart.java
```diff
@@ -40,7 +41,9 @@ public void run() throws Exception {
         .setType(MultipartBody.FORM)
         .addFormDataPart("title", "Square Logo")
         .addFormDataPart("image", "logo-square.png",
-            RequestBody.create(new File("website/static/logo-square.png"), MEDIA_TYPE_PNG))
+            RequestBody.create(
+                new File(SamplesKt.getPrefix() + "/docs/images/logo-square.png"),
```
yschimke: the path change is the only required fix, we can ditch the test and getPrefix() logic

samples/guide/src/main/java/okhttp3/recipes/PostFile.java
```diff
@@ -30,7 +31,7 @@
   private final OkHttpClient client = new OkHttpClient();
 
   public void run() throws Exception {
-    File file = new File("README.md");
+    File file = new File(SamplesKt.getPrefix() + "README.md");
```
swankjesse: I don't really like this. The goal of these samples is to stand alone.

Any alternatives?

samples/guide/src/main/java/okhttp3/samples.kt
```diff
@@ -0,0 +1,5 @@
+package okhttp3
+
+import java.io.File
+
+val prefix = if (File("README.md").exists()) "" else "../../"
```
swankjesse: Should we just create two README.md files so users always get lucky?

samples/guide/src/test/kotlin/okhttp3/AllMainsTest.kt
```diff
@@ -0,0 +1,37 @@
+package okhttp3
```
swankjesse: Nit: copyright

samples/guide/src/test/kotlin/okhttp3/AllMainsTest.kt
```diff
@@ -0,0 +1,37 @@
...
+
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.junit.runners.Parameterized
+import java.io.File
+import java.lang.reflect.InvocationTargetException
+
+@RunWith(Parameterized::class)
+class AllMainsTest(val className: String) {
```
swankjesse: This is really clever. I'd be tempted to just write 15 regular test methods, then you could assert exceptions here rather than in the samples themselves? These don't change too frequently!

samples/guide/src/test/kotlin/okhttp3/AllMainsTest.kt
```diff
@@ -0,0 +1,37 @@
...
+  @Test
+  fun runMain() {
+    val mainMethod = Class.forName(className).methods.find { it.name == "main" }
+    try {
+      mainMethod?.invoke(null, arrayOf<String>())
+    } catch (ite: InvocationTargetException) {
+      throw ite.targetException
```
monkey-mas: nit: maybe we use `getCause()` instead? The doc says `The {@link Throwable#getCause()} method is now the preferred means of obtaining this information.`

```
    /**
     * Get the thrown target exception.
     *
     * <p>This method predates the general-purpose exception chaining facility.
     * The {@link Throwable#getCause()} method is now the preferred means of
     * obtaining this information.
     *
     * @return the thrown target exception (cause of this exception).
     */
    public Throwable getTargetException() {
        return target;
    }

    /**
     * Returns the cause of this exception (the thrown target exception,
     * which may be {@code null}).
     *
     * @return  the cause of this exception.
     * @since   1.4
     */
    public Throwable getCause() {
        return target;
    }
```

samples/guide/src/test/kotlin/okhttp3/AllMainsTest.kt
```diff
@@ -0,0 +1,37 @@
...
+      }.sorted()
+    }
+
+    private fun mainFiles(): List<File> {
+      return File("$prefix/samples/guide/src/main/java/okhttp3").listFiles()?.flatMap {
+        it?.listFiles()?.toList().orEmpty()
+      }.orEmpty()
+    }
+  }
+}
```
monkey-mas: nit: new line?

samples/guide/src/main/java/okhttp3/recipes/CheckHandshake.java
```diff
@@ -59,6 +59,10 @@ public void run() throws Exception {
   }
 
   public static void main(String... args) throws Exception {
-    new CheckHandshake().run();
+    try {
+      new CheckHandshake().run();
+    } catch (IOException ioe) {
+      // expected
```
monkey-mas: nit: maybe change the variable name from `ioe` to `ignored` or `expected` and remove the comment? :)

samples/guide/src/main/java/okhttp3/recipes/RequestBodyCompression.java
```diff
@@ -66,7 +66,11 @@ public void run() throws Exception {
   }
 
   public static void main(String... args) throws Exception {
-    new RequestBodyCompression().run();
+    try {
+      new RequestBodyCompression().run();
+    } catch (IOException ioe) {
```
monkey-mas: nit: maybe `ignored`? :)

samples/guide/src/main/java/okhttp3/recipes/PostFile.java
yschimke: removed with body doubles

samples/guide/src/test/kotlin/okhttp3/AllMainsTest.kt
yschimke: Left this mostly as is, my goal is keeping it stupidly simple to add a normal passing sample, but detect when that breaks or is flaky

samples/guide/src/main/java/okhttp3/recipes/PostMultipart.java
```diff
@@ -40,7 +40,9 @@ public void run() throws Exception {
         .setType(MultipartBody.FORM)
         .addFormDataPart("title", "Square Logo")
         .addFormDataPart("image", "logo-square.png",
-            RequestBody.create(new File("website/static/logo-square.png"), MEDIA_TYPE_PNG))
+            RequestBody.create(
+                new File("docs/images/logo-square.png"),
```
yschimke: Still the only essential change.

swankjesse:
We need to disable these by default ’cause they don’t work in CI.

```
okhttp3.AllMainsTest > runMain[okhttp3.recipes.ParseResponseWithMoshi] FAILED
    java.io.IOException: Unexpected code Response{protocol=http/1.1, code=403, message=Forbidden, url=https://api.github.com/gists/c2a7c39532239ff261be}
        at okhttp3.recipes.ParseResponseWithMoshi.run(ParseResponseWithMoshi.java:36)
        at okhttp3.recipes.ParseResponseWithMoshi.main(ParseResponseWithMoshi.java:56)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at okhttp3.AllMainsTest.runMain(AllMainsTest.kt:30)
```

monkey-mas:
Hmm, so the CI is not allowed to  access `https://api.github.com/gists/c2a7c39532239ff261be`...? 😕

PR #5239, opened 2019-06-26
Run error-prone on JDK 11

JakeWharton:
JDK 12+ is where it currently is unsupported.


swankjesse:
I'm not super enthusiastic about non-LTS releases anyway.

PR #5243, opened 2019-06-26
More supported socket types on Android

yschimke:
Ensures we are using H2 on Android with all of

- Android base
- Google Play Provider
- Conscrypt (lib)

- [x]  Test Android 22 (Lollipop)
- [x]  Test Android 28 (Pie)
- [x]  Test GMS
- [x]  Test Conscrypt

Includes a unit test of the core socket handling via Conscrypt on JDK


yschimke:
```
h2 class com.android.org.conscrypt.Java8FileDescriptorSocket
h2 class org.conscrypt.Java8FileDescriptorSocket

```

yschimke:
```
h2 class com.android.org.conscrypt.Java8FileDescriptorSocket
h2 class org.conscrypt.Java8FileDescriptorSocket
h2 class com.google.android.gms.org.conscrypt.Java8FileDescriptorSocket
```

okhttp/src/main/java/okhttp3/internal/platform/AndroidSocketAdapter.kt
```diff
@@ -0,0 +1,149 @@
...
+        context!!, X509TrustManager::class.java, "x509TrustManager")
+    return x509TrustManager ?: Platform.readFieldOrNull(context, X509TrustManager::class.java,
+        "trustManager")
+  }
+
+  fun matchesSocket(sslSocket: SSLSocket): Boolean {
+    val matches = matchesSocketInternal(sslSocket)
+
+    if (matches) {
+      initSocketClasses(sslSocket.javaClass)
```
yschimke: This is the core of the fix, finding the Gms classes isn't obvious (classpath tricks?), but the actual class can provide it.

okhttp/src/test/java/okhttp3/internal/platform/AndroidSocketAdapterTest.kt
```diff
@@ -0,0 +1,65 @@
...
+  val provider = Conscrypt.newProviderBuilder().provideTrustManager(true).build()
+  val context = SSLContext.getInstance("TLS", provider)
+  init {
+    context.init(null, null, null)
+  }
+
+  @Test
+  fun testInit() {
+    val socketFactory = context.socketFactory
+    adapter.matchesSocketFactory(socketFactory)
```
yschimke: I will add asserts here and next match check

okhttp/src/main/java/okhttp3/internal/platform/AndroidPlatform.kt
```diff
@@ -65,87 +61,26 @@ class AndroidPlatform(
...
-      }
-
-      // Enable ALPN.
-      setAlpnProtocols.invoke(sslSocket, concatLengthPrefixed(protocols))
-    } catch (e: IllegalAccessException) {
-      throw AssertionError(e)
-    } catch (e: InvocationTargetException) {
-      throw AssertionError(e)
-    }
+    // No TLS extensions if the socket class is custom.
```
yschimke: Should we replace the comment with a once off Log.warning?

yschimke:
@swankjesse any steer on this PR? It fixes a real bug, but also hopefully makes the code cleaner and tests the Android code via Conscrypt.

4.1? 4.2 with changes? never?

okhttp/src/main/java/okhttp3/internal/platform/AndroidSocketAdapter.kt
```diff
@@ -0,0 +1,157 @@
...
+
+  @Synchronized private fun initSocketClasses(actualSSLSocketClass: Class<in SSLSocket>) {
+    if (!this::sslSocketClass.isInitialized) {
+      var possibleClass: Class<in SSLSocket> = actualSSLSocketClass
+      while (possibleClass.name != "$packageName.OpenSSLSocketImpl") {
+        possibleClass = possibleClass.superclass
+
+        if (possibleClass == null) {
+          throw AssertionError(
+              "No OpenSSLSocketImpl superclass of socket of type $actualSSLSocketClass")
```
swankjesse: Seems fragile, especially to OS upgrades

okhttp/src/main/java/okhttp3/internal/platform/AndroidSocketAdapter.kt
```diff
@@ -0,0 +1,157 @@
...
+import javax.net.ssl.SSLSocketFactory
+import javax.net.ssl.X509TrustManager
+
+open class AndroidSocketAdapter internal constructor(private val packageName: String) {
+  internal lateinit var sslSocketClass: Class<in SSLSocket>
+  internal lateinit var paramClass: Class<*>
+  internal val setUseSessionTickets: Method by lazy {
+    sslSocketClass.getDeclaredMethod("setUseSessionTickets", Boolean::class.javaPrimitiveType)
+  }
+  internal val setHostname by lazy {
```
swankjesse: Why are these lazy? I don’t really like lazy because you always pay for the lazy indirection, which will often make it more work than not lazy

okhttp/src/main/java/okhttp3/internal/platform/ConscryptPlatform.kt
```diff
@@ -106,18 +106,22 @@ class ConscryptPlatform private constructor() : Platform() {
...
 
   companion object {
-    fun buildIfSupported(): ConscryptPlatform? = try {
-      // Trigger an early exception over a fatal error, prefer a RuntimeException over Error.
-      Class.forName("org.conscrypt.Conscrypt\$Version")
-
-      when {
-        Conscrypt.isAvailable() && atLeastVersion(2, 1, 0) -> ConscryptPlatform()
-        else -> null
+    val isSupported: Boolean by lazy {
```
swankjesse: why lazy?

okhttp/src/main/java/okhttp3/internal/platform/AndroidSocketAdapter.kt
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    return matches
+  }
+
+  internal open fun matchesSocketInternal(sslSocket: SSLSocket) =
+      sslSocket.javaClass.name.startsWith("${this.packageName}.")
+
+  @Synchronized private fun initSocketClasses(actualSSLSocketClass: Class<in SSLSocket>) {
+    if (!this::sslSocketClass.isInitialized) {
```
swankjesse: ew, can you find a way to do this without isInitialized?

swankjesse: maybe do it eagerly? Like when this class is loaded or something?

yschimke:
> I don’t love it.
> 
> I think the SocketAdapter class is a good idea, but this code feels both frameworkey and loose.
> 
> The other platform code is intended to be really procedural: interrogate the incoming socket and make a decision about it. The failure mode was mostly that we’d end up with a basic Platform instance.
> 
> With this we can end up selecting a SocketAdapter without initializing it due to all the use of `lazy` and `lateinit`, and the whole thing just creates a bunch of different possibly-initialized states that I find hard to reason about.

But both the GMS provider and Conscrypt may be activated after we've already loaded the Platform, it's too racey.  And specifically for GMS, I don't think we can reliably look for the classes.  You have to wait until they come by!

I'll drop the loop over socket implementations and make it more procedural, consider the GMS and Conscrypt cases two special cases triggered as needed.

A bit value add from this change is that we can unit test the Android socket behaviour on JVM builds via Conscrypt!  So to me it's a big win overall!

I'll put up another cut when I have time this wek for re-review!

Thanks for the feedback!

okhttp/src/main/java/okhttp3/internal/platform/ConscryptPlatform.kt
yschimke: IIRC mainly why run this sort of code unintentionally on other platforms e.g. Android if we aren't using it.  I can simplify.

okhttp/src/main/java/okhttp3/internal/platform/AndroidSocketAdapter.kt
yschimke: Good call.  I can make it fail into a dumb mode.  But overall the major functional change of this PR is to not look for classes via reflection by name.  But to wait for those sockets to come by and use their runtime implementations to discover the methods.

okhttp/src/main/java/okhttp3/internal/platform/android/util.kt
```diff
@@ -0,0 +1,41 @@
...
+ * limitations under the License.
+ */
+package okhttp3.internal.platform.android
+
+import android.util.Log
+import okhttp3.internal.platform.Platform
+
+private const val MAX_LOG_LENGTH = 4000
+
+internal fun androidLog(level: Int, message: String, t: Throwable?) {
```
yschimke: Split out to keep Android Log logic out of the main platform class.

okhttp/src/main/java/okhttp3/internal/platform/android/CloseGuard.kt
```diff
@@ -0,0 +1,75 @@
...
+package okhttp3.internal.platform.android
+
+import java.lang.reflect.Method
+
+/**
+ * Provides access to the internal dalvik.system.CloseGuard class. Android uses this in
+ * combination with android.os.StrictMode to report on leaked java.io.Closeable's. Available since
+ * Android API 11.
+ */
+internal class CloseGuard(
```
yschimke: With a specific Android package, moving CloseGuard out keeps AndroidPlatform smaller and focused.

yschimke:
> I don’t love it.
> 
> I think the SocketAdapter class is a good idea, but this code feels both frameworkey and loose.
> 
> The other platform code is intended to be really procedural: interrogate the incoming socket and make a decision about it. The failure mode was mostly that we’d end up with a basic Platform instance.
> 
> With this we can end up selecting a SocketAdapter without initializing it due to all the use of `lazy` and `lateinit`, and the whole thing just creates a bunch of different possibly-initialized states that I find hard to reason about.

Hopefully it's ready to land now.  FWIW the AndroidPlatform logic is now, are we on a phone, yes use AndroidPlatform.  Then the vagueries of different socket types (android, conscrypt, gms, custom) are pull out to the Adapters.  So the design is cleaner and allows for changes over time e.g. when Gms is activated after startup, or some custom sockets are used.

yschimke:
Bumping to 4.2, because this change will need some bake time and testing.

okhttp/src/main/java/okhttp3/internal/platform/AndroidPlatform.kt
```diff
@@ -342,32 +224,21 @@ class AndroidPlatform(
...
-      // Attempt to find Android 5+ APIs.
-      val sslParametersClass: Class<*>
-      val sslSocketClass: Class<*>
-      try {
-        sslParametersClass = Class.forName("com.android.org.conscrypt.SSLParametersImpl")
-        sslSocketClass = Class.forName("com.android.org.conscrypt.OpenSSLSocketImpl")
-      } catch (_: ClassNotFoundException) {
-        return null // Not an Android runtime.
+    val isSupported: Boolean = try {
+      // Trigger an early exception over a fatal error, prefer a RuntimeException over Error.
```
swankjesse: Nice comment.

okhttp/src/main/java/okhttp3/internal/platform/AndroidPlatform.kt
```diff
@@ -342,32 +224,21 @@ class AndroidPlatform(
...
-          val getAlpnSelectedProtocol = sslSocketClass.getMethod("getAlpnSelectedProtocol")
-          val setAlpnProtocols = sslSocketClass.getMethod("setAlpnProtocols", ByteArray::class.java)
-          return AndroidPlatform(sslParametersClass, sslSocketClass, setUseSessionTickets,
-              setHostname, getAlpnSelectedProtocol, setAlpnProtocols)
-        } catch (_: NoSuchMethodException) {
-        }
-      }
-      throw IllegalStateException("Expected Android API level 21+ but was ${Build.VERSION.SDK_INT}")
+      true
+    } catch (e: ClassNotFoundException) {
```
swankjesse: supernit: `e` -> `_`

okhttp/src/main/java/okhttp3/internal/platform/android/SocketAdapter.kt
```diff
@@ -0,0 +1,36 @@
...
+ * limitations under the License.
+ */
+package okhttp3.internal.platform.android
+
+import okhttp3.Protocol
+import javax.net.ssl.SSLSocket
+import javax.net.ssl.SSLSocketFactory
+import javax.net.ssl.X509TrustManager
+
+interface SocketAdapter {
```
swankjesse: dig

yschimke:
Merging the day after a release is the right call. Needs some bake time and snapshot testing.

Issue #5246, opened 2019-06-27
HttpHeaders.hasBody removed


PaulWoitaschek:
There is no replacement for the old function `HttpHeaders.hasBody`

Therefore it's not binary compatible and crashes using chuck:

https://github.com/jgilfelt/chuck/blob/152f9a79f94ea23e9f0542137765961918909a76/library/src/main/java/com/readystatesoftware/chuck/ChuckInterceptor.java#L196

Feel free to close as it's in the internal package.

yschimke:
It's generally inline with https://github.com/square/okhttp/commit/ab72e81808533957e4d669d8ae1a817056e4bce2

koral--:
There is a `Response.promisesBody()` extension function which looks like possible replacement.

https://github.com/square/okhttp/blob/eee838aebe8d3524d7e0e2dbf8f9bf357512f038/okhttp/src/main/java/okhttp3/internal/http/HttpHeaders.kt#L213

swankjesse:
Do not use APIs in `internal`. We will move them in a future release without warning.

If you need this functionality, copy-paste it into your app.

PR #5251, opened 2019-06-29
Java internal compatibility e.g. HttpHeaders.hasBody

yschimke:
For discussion, may not land


okhttp/src/main/java/okhttp3/internal/http/HttpHeaders.kt
```diff
@@ -232,3 +233,9 @@ fun Response.promisesBody(): Boolean {
 
   return false
 }
+
+@Deprecated(message = "No longer supported", level = DeprecationLevel.ERROR,
+    replaceWith = ReplaceWith(expression = "response.promisesBody()"))
```
monkey-mas: 1) Since we allow Java users to use this deprecated method, it'd be nice for Kotlin developers to use this function for a while as well? i.e. We might downgrade the error level from `ERROR` to `WARNING`?

2) nit: seems like we usually indent for each param? e.g.
```
@Deprecated(
    message = "No longer supported",
    replaceWith = ReplaceWith(expression = "response.promisesBody()"),
```

okhttp/src/test/java/okhttp3/HeadersTest.java
```diff
@@ -36,6 +36,7 @@
 import static java.util.Collections.singletonMap;
 import static okhttp3.TestUtil.headerEntries;
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.Assert.assertTrue;
```
monkey-mas: nit: we don't need this?

okhttp/src/test/java/okhttp3/PublicInternalApiTest.java
```diff
@@ -0,0 +1,45 @@
...
+package okhttp3;
+
+import okhttp3.internal.http.HttpHeaders;
+import okhttp3.internal.http.HttpMethod;
+import org.junit.Test;
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+@SuppressWarnings("ALL") public class PublicInternalApiTest {
```
monkey-mas: Do we need `@SuppressWarnings("ALL")`? 👀 Seems like tests are green without it.

okhttp/src/test/java/okhttp3/PublicInternalApiTest.java
```diff
@@ -0,0 +1,45 @@
...
+  @Test public void permitsRequestBody() {
+    assertTrue(HttpMethod.permitsRequestBody("POST"));
+    assertFalse(HttpMethod.permitsRequestBody("GET"));
```
monkey-mas: nit: I'd probably test `assertFalse(HttpMethod.permitsRequestBody("HEAD"));` as well :)

okhttp/src/test/java/okhttp3/PublicInternalApiTest.java
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  @Test public void requiresRequestBody() {
+    assertTrue(HttpMethod.requiresRequestBody("PUT"));
```
monkey-mas: nit: Given that `requiresRequestBody` is a public method and there aren't many test cases, it'd be okay to test all the assertTrue cases. i.e.
```
    assertTrue(HttpMethod.requiresRequestBody("POST"));
    assertTrue(HttpMethod.requiresRequestBody("PUT"));
    assertTrue(HttpMethod.requiresRequestBody("PATCH"));
    assertTrue(HttpMethod.requiresRequestBody("PROPPATCH"));
    assertTrue(HttpMethod.requiresRequestBody("REPORT"));
```

okhttp/src/test/java/okhttp3/PublicInternalApiTest.java
```diff
@@ -0,0 +1,45 @@
...
+  @Test public void hasBody() {
+    Request request = new Request.Builder().url("http://example.com").build();
+    Response response = new Response.Builder().code(200)
+        .message("OK")
+        .request(request)
+        .protocol(Protocol.HTTP_2)
+        .build();
+    assertTrue(HttpHeaders.hasBody(response));
+  }
+}
```
monkey-mas: nit: newline :)

okhttp/src/test/java/okhttp3/PublicInternalApiTest.java
yschimke: I deliberately didn't do that because the test is a pass through and so I'd be copying other tests verbatim, seems like maintenance for not much gain.

yschimke:
Closing - as per https://github.com/square/okhttp/issues/5246

swankjesse:
Reopening! I'm going to both make strong statements to users to not access our internal APIs, and also in rare cases add workarounds when they do.  I really do not want users calling our _new_ internal APIs; that’s just obnoxious.

yschimke:
Ok. I’ll fix up and land tonight.

okhttp/src/main/java/okhttp3/internal/http/HttpHeaders.kt
yschimke: We have chosen to break source compatibility for kotlin users when upgrading

> OkHttp is not source-compatible for Kotlin callers, but upgrading should be automatic thanks to Kotlin’s powerful deprecation features. Most developers should be able to use IntelliJ’s Code Cleanup for a safe and fast upgrade.

okhttp/src/test/java/okhttp3/PublicInternalApiTest.java
yschimke: This class doesn't need to worry about deprecations, or non sensical code. e.g. the other similar classes don't even run some code, so it's ok to do weird things to confirm API compat.

yschimke:
@swankjesse when you get a chance, could you cut a 4.0.x branch?  Is it literally just pushing a branch to origin? or is there some setup?

swankjesse:
LGTM

`master` is our 4.x branch. What else?

yschimke:
Wasn’t sure if you wanted fixes to target a 4.0.1?

swankjesse:
I think it can probably wait til 4.1!

PR #5253, opened 2019-06-29
Basic Android Instrumentation Test

yschimke:
Run separately against release version currently.  i.e. not to run inside CI.

Mainly useful to define a set of tests we expect to run against real emulators or phones to check for regressions or confirm reported bugs.

Specifically:
 - Tests don't run against a mocked environment like robolectric.  Only emulator or phone.
 - Tests are not part of regular CI.
 - Tests are currently against a released version, but could change to master if that is easy.
 - Tests are a separate project in same repo.

Run with

```
$ ./gradlew :android-test:connectedCheck
```

### DIFF
--- a/android-test/build.gradle
+++ b/android-test/build.gradle
@@ -1,0 +1,28 @@
+apply plugin: 'com.android.library'
+apply plugin: 'org.jetbrains.kotlin.android'
+
+android {
+  compileOptions {
+    sourceCompatibility JavaVersion.VERSION_1_8
+    targetCompatibility JavaVersion.VERSION_1_8
+  }
+
+  compileSdkVersion 29
+
+  defaultConfig {
+    minSdkVersion 21
+    targetSdkVersion 29
+    versionCode 1
+    versionName "1.0"
+
+    testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
+  }
+}
+
+dependencies {
+  implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8:${versions.kotlin}"
+  implementation project(':okhttp')
+  testImplementation 'junit:junit:4.12'
+  androidTestImplementation 'com.android.support.test:runner:1.0.2'
+  androidTestImplementation 'com.android.support.test.espresso:espresso-core:3.0.2'
+}
--- a/android-test/src/androidTest/java/okhttp/android/test/OkHttpTest.kt
+++ b/android-test/src/androidTest/java/okhttp/android/test/OkHttpTest.kt
@@ -1,3 +1,18 @@
+/*
+ * Copyright (C) 2019 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package okhttp.android.test
 
 import android.os.Build
@@ -12,9 +27,12 @@ import okhttp3.TlsVersion
 import org.junit.After
 import org.junit.Assert.assertEquals
 import org.junit.Assert.fail
+import org.junit.Assume
 import org.junit.Before
 import org.junit.Test
 import org.junit.runner.RunWith
+import java.net.InetAddress
+import java.net.UnknownHostException
 
 /**
  * Run with "./gradlew :android-test:connectedCheck" and make sure ANDROID_SDK_ROOT is set.
@@ -36,6 +54,8 @@ class OkHttpTest {
 
   @Test
   fun testRequest() {
+    assumeNetwork()
+
     val request = Request.Builder().url("https://api.twitter.com/robots.txt").build()
 
     val response = client.newCall(request).execute()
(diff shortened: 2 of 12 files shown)


android-test/app/src/main/res/values-w820dp/dimens.xml
```diff
@@ -0,0 +1,22 @@
...
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+<resources>
+    <!-- Example customization of dimensions originally defined in res/values/dimens.xml
+         (such as screen margins) for screens with more than 820dp of available width. This
+         would include 7" and 10" devices in landscape (~960dp and ~1280dp respectively). -->
+    <dimen name="activity_horizontal_margin">64dp</dimen>
```
JakeWharton: Almost all of these resource files are not necessary.

android-test/build.gradle
```diff
@@ -0,0 +1,31 @@
...
+
+allprojects {
+    repositories {
+        jcenter()
+        google()
+    }
+}
+
+ext {
+    kotlinVersion = "1.3.40"
```
JakeWharton: Don't like that all of these are separate from the other variables in the enclosing project

yschimke:
@JakeWharton Thanks for the review.  I took one of the google example instrumentation tests, I didn't want to spend too much time cleaning it up before getting feedback from others on whether it was desirable.  Based on feedback I'll clean it up and land when I have some time.  Thanks!

settings.gradle
```diff
@@ -1,4 +1,11 @@
+rootProject.name = 'okhttp-parent'
```
yschimke:
```
> Failed to apply plugin [id 'com.android.library']
   > Your project contains 2 or more modules with the same identification com.squareup.okhttp3:okhttp
     at ":" and ":okhttp".
     You must use different identification (either name or group) for each modules.
```

settings.gradle
```diff
@@ -1,4 +1,11 @@
...
+
 include ':mockwebserver'
+if (!properties.containsKey('android.injected.invoked.from.ide') && System.getenv(
```
yschimke: @swankjesse @JakeWharton  This should make the code pretty safe for both normal development and also CI.  But allow us to do real local tests against phones etc to flush out bugs.

Thoughts?

android-test/src/androidTest/java/okhttp/android/test/OkHttpTest.kt
```diff
@@ -0,0 +1,86 @@
+package okhttp.android.test
```
swankjesse: nit: copyright

android-test/src/androidTest/java/okhttp/android/test/OkHttpTest.kt
```diff
@@ -0,0 +1,86 @@
...
+
+    response.use {
+      assertEquals(Protocol.HTTP_2, response.protocol)
+      if (Build.VERSION.SDK_INT >= 29) {
+        assertEquals(TlsVersion.TLS_1_3, response.handshake?.tlsVersion)
+      } else {
+        assertEquals(TlsVersion.TLS_1_2, response.handshake?.tlsVersion)
+      }
+      assertEquals(200, response.code)
+      assertEquals("com.android.org.conscrypt.Java8FileDescriptorSocket", socketClass)
```
swankjesse: Not in love with this. Testing behavior above is sufficient?

android-test/src/androidTest/java/okhttp/android/test/OkHttpTest.kt
```diff
@@ -0,0 +1,86 @@
...
+    }
+  }
+
+  @Test
+  fun testHttpRequestBlocked() {
```
swankjesse: This won’t work on all Androids

settings.gradle
swankjesse: Works for me. Got an issue to link to so we know when we can delete this?

swankjesse:
Also! Tests that call real hosts doesn’t seem like a very reliable strategy.

android-test/src/androidTest/java/okhttp/android/test/OkHttpTest.kt
yschimke: I specifically want to confirm that we are using an Android socket here.  So I'd like one test that confirms this.

android-test/src/androidTest/java/okhttp/android/test/OkHttpTest.kt
yschimke: I'll add junit assumptions

yschimke:
> Mixed feels on this.
> 
> When I was playing with this I wrote an Android app that ran the standard OkHttp test suite, with a custom suppression list. It was hackier and uglier but it meant I could pretty easily figure out where the problems were.

I'll try to make sure it stays out of the way, but TBH it was a godsend for testing out the socket changes etc.  So much quicker.  Benefits overall

1) It's the quickest target for someone to add a bug reproduction to, without generating a new project and uploading to github.
2) It's good for us developing anything internally that needs to run on Android.

If I keep this out of the way, give me some lee-way and I'll land close to as-is, then add more MockWebserver tests over time to beef up the capabilities.  It should give us better coverage of not breaking android with changes sooner.  It becomes trivial to run these tests locally and frequently against a real phone etc.

yschimke:
> Also! Tests that call real hosts doesn’t seem like a very reliable strategy.

I think we need a couple but I can put them behind an assumeNetwork() check.

android-test/src/androidTest/java/okhttp/android/test/OkHttpTest.kt
swankjesse: Sounds good!

PR #5268, opened 2019-07-04
[WIP] Cleanup of error prone and warnings

yschimke:


yschimke:
FWIW I dislike this change, but if we want value out of errorprone and compiler warnings, we need to reduce the noise on every build.

okhttp/src/main/java/okhttp3/CertificatePinner.kt
```diff
@@ -137,15 +137,15 @@ class CertificatePinner internal constructor(
    */
   @Throws(SSLPeerUnverifiedException::class)
   fun check(hostname: String, peerCertificates: List<Certificate>) {
-    var peerCertificates = peerCertificates
+    var cleanPeerCertificates = peerCertificates
```
swankjesse: I really dislike this warning

okhttp/src/main/java/okhttp3/internal/tls/OkHostnameVerifier.kt
```diff
@@ -70,6 +70,7 @@ object OkHostnameVerifier : HostnameVerifier {
    * @param pattern domain name pattern from certificate. May be a wildcard pattern such as
    *     `*.android.com`.
    */
+  @Suppress("NAME_SHADOWING")
```
swankjesse: Can we suppress this warning in our compiler config?

okhttp/src/test/java/okhttp3/CertificatePinnerKotlinTest.kt
```diff
@@ -23,6 +23,7 @@ import org.junit.Test
 
 class CertificatePinnerKotlinTest {
 
+  @Suppress("DEPRECATION")
```
swankjesse: I don’t get this. What‘s warning here?

swankjesse: Should we just fix it to call the non-deprecated method?

samples/crawler/src/main/java/okhttp3/sample/Crawler.java
```diff
@@ -48,6 +48,7 @@ public Crawler(OkHttpClient client) {
     this.client = client;
   }
 
+  @SuppressWarnings("CatchAndPrintStackTrace")
```
swankjesse: yuck, this is a huge distraction in sample code

samples/guide/src/main/java/okhttp3/recipes/CancelCall.java
```diff
@@ -24,6 +24,7 @@
 import okhttp3.Request;
 import okhttp3.Response;
 
+@SuppressWarnings("FutureReturnValueIgnored")
```
swankjesse: remove these from sample code? Maybe we can suppress the warnings in the build file instead of the source?

swankjesse: It grosses me out that our users will just copy these suppressions into their code, not knowing why they exist or what they do

samples/static-server/src/main/java/okhttp3/sample/SampleServer.java
```diff
@@ -18,7 +18,7 @@
 import okio.Buffer;
 import okio.Okio;
 
-public class SampleServer extends Dispatcher {
+@SuppressWarnings("deprecation") public class SampleServer extends Dispatcher {
```
swankjesse: no suppressions in sample code please

if we’re using deprecated things in sample code, that’s what we should fix.

okhttp/src/test/java/okhttp3/KotlinSourceModernTest.kt
```diff
@@ -304,6 +304,7 @@ class KotlinSourceModernTest {
     val basic: String = Credentials.basic("", "")
   }
 
+  @Suppress("IMPLICIT_NOTHING_AS_TYPE_PARAMETER")
```
swankjesse: move this to the top of the file?

okhttp/src/test/java/okhttp3/KotlinSourceModernTest.kt
```diff
@@ -474,6 +475,7 @@ class KotlinSourceModernTest {
     val headers: Headers = builder.build()
   }
 
+  @Suppress("DEPRECATION")
```
swankjesse: move this to the top of the file

JakeWharton:
The error prone checks should be made to error if you're fixing them. Warnings are generally useless. It should either always error and require suppression with a comment or just be disabled.

okhttp/src/main/java/okhttp3/internal/tls/OkHostnameVerifier.kt
yschimke: Haven't worked that out yet.  Moving away from the code to the top of the file for now.

yschimke:
@swankjesse @JakeWharton Thanks for the discussion.  Agreed on the errorprone becoming errors, I'll work towards this module by module.

PR #5272, opened 2019-07-06
Kotlin upgrade to 1.3.41

yschimke:
Automatically prompted by IDE


yschimke:
If we have a defined policy, then happy to follow it.  Otherwise I'll follow Intellij prompts.

yschimke:
I will next time.  It looks like a lot of type inference fixes, which are unlikely to help once we already have compiling code and explicit return types on public api.  

https://github.com/JetBrains/kotlin/releases/tag/v1.3.41

PR #5273, opened 2019-07-07
Document kotlin library dependency

yschimke:


yschimke:
I suspect this say too much about Kotlin for this doc, but I thought I would start there and work back to what we want to say.  This is meant to talk openly about the most contentious part of Kotlin adoption at least observed in twitter etc.

PR #5279, opened 2019-07-10
Update the changelog for OkHttp 4.0.1

swankjesse:


swankjesse:
Self-merging docs change.

PR #5283, opened 2019-07-12
Avoid using non-null assertion in ConnectionSpec

### LINKED ISSUE #5280: KotlinNullPointerException on any network request on old Android 5 phone with new okhttp-logging-interceptor 4.0.0+
After upgrading from com.squareup.okhttp3:logging-interceptor 3.14.1 to 4.0.0 (or 4.0.1), with no other code changes, I am getting a KotlinNullPointerException on any network request, when our app installed on HTC One with Android 5.0.2. I am unsure of which other devices also have this issue. It also crashes even if I remove all calls to addInterceptor.

It crashes because cipherSuites is null in the ConnectionSpec. cipherSuites is null in the connection spec because the code I was using to enable TLS 1.2 on API 21 was calling .allEnabledCipherSuites() on the ConnectionSpec.Builder. As far as I can tell, all that allEnabledCipherSuites does is set the cipherSuites to null.

Is this the expected behavior? Do should I be calling allEnabledCipherSuites() or not?

```
E/AndroidRuntime_2_crash: crash in the same process: OkHttp Dispatcher
    kotlin.KotlinNullPointerException
        at okhttp3.ConnectionSpec.hashCode(ConnectionSpec.kt:186)
        at java.util.ArrayList.hashCode(ArrayList.java:603)
        at java.util.Collections$UnmodifiableList.hashCode(Collections.java:1054)
        at okhttp3.Address.hashCode(Address.kt:178)
        at okhttp3.Route.hashCode(Route.kt:84)
        at java.util.Collections.secondaryHash(Collections.java:3405)
        at java.util.HashMap.containsKey(HashMap.java:325)
        at java.util.HashSet.contains(HashSet.java:138)
        at okhttp3.internal.connection.RouteDatabase.shouldPostpone(RouteDatabase.kt:40)
        at okhttp3.internal.connection.RouteSelector.next(RouteSelector.kt:74)
        at okhttp3.internal.connection.ExchangeFinder.findConnection(ExchangeFinder.kt:199)
        at okhttp3.internal.connection.ExchangeFinder.findHealthyConnection(ExchangeFinder.kt:109)
        at okhttp3.internal.connection.ExchangeFinder.find(ExchangeFinder.kt:77)
        at okhttp3.internal.connection.Transmitter.newExchange$okhttp(Transmitter.kt:162)
        at okhttp3.internal.connection.ConnectInterceptor.intercept(ConnectInterceptor.kt:35)
        at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:112)
        at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:87)
        at okhttp3.internal.cache.CacheInterceptor.intercept(CacheInterceptor.kt:82)
        at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:112)
        at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:87)
        at okhttp3.internal.http.BridgeInterceptor.intercept(BridgeInterceptor.kt:84)
        at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:112)
        at okhttp3.internal.http.RetryAndFollowUpInterceptor.intercept(RetryAndFollowUpInterceptor.kt:71)
        at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:112)
        at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:87)
        at okhttp3.RealCall.getResponseWithInterceptorChain(RealCall.kt:184)
        at okhttp3.RealCall$AsyncCall.run(RealCall.kt:136)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1112)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:587)
        at java.lang.Thread.run(Thread.java:818)
```

monkey-mas:
Fix #5280 

To be compatible with OkHttp3, OkHttp4.0 doesn't use non-null assertion.

### DIFF
--- a/okhttp/src/main/java/okhttp3/ConnectionSpec.kt
+++ b/okhttp/src/main/java/okhttp3/ConnectionSpec.kt
@@ -168,13 +168,12 @@ class ConnectionSpec internal constructor(
     if (other !is ConnectionSpec) return false
     if (other === this) return true
 
-    val that = other as ConnectionSpec?
-    if (this.isTls != that!!.isTls) return false
+    if (this.isTls != other.isTls) return false
 
     if (isTls) {
-      if (!Arrays.equals(this.cipherSuitesAsString, that.cipherSuitesAsString)) return false
-      if (!Arrays.equals(this.tlsVersionsAsString, that.tlsVersionsAsString)) return false
-      if (this.supportsTlsExtensions != that.supportsTlsExtensions) return false
+      if (!Arrays.equals(this.cipherSuitesAsString, other.cipherSuitesAsString)) return false
+      if (!Arrays.equals(this.tlsVersionsAsString, other.tlsVersionsAsString)) return false
+      if (this.supportsTlsExtensions != other.supportsTlsExtensions) return false
     }
 
     return true
@@ -183,8 +182,8 @@ class ConnectionSpec internal constructor(
   override fun hashCode(): Int {
     var result = 17
     if (isTls) {
-      result = 31 * result + cipherSuitesAsString!!.contentHashCode()
-      result = 31 * result + tlsVersionsAsString!!.contentHashCode()
+      result = 31 * result + (cipherSuitesAsString?.contentHashCode() ?: 0)
+      result = 31 * result + (tlsVersionsAsString?.contentHashCode() ?: 0)
       result = 31 * result + if (supportsTlsExtensions) 0 else 1
     }
     return result
--- a/okhttp/src/test/java/okhttp3/ConnectionSpecTest.java
+++ b/okhttp/src/test/java/okhttp3/ConnectionSpecTest.java
@@ -284,6 +284,8 @@ public final class ConnectionSpecTest {
     assertThat(set.add(ConnectionSpec.CLEARTEXT)).isTrue();
     assertThat(set.add(allTlsVersions)).isTrue();
     assertThat(set.add(allCipherSuites)).isTrue();
+    allCipherSuites.hashCode();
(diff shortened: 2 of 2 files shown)


okhttp/src/main/java/okhttp3/ConnectionSpec.kt
```diff
@@ -183,8 +183,8 @@ class ConnectionSpec internal constructor(
   override fun hashCode(): Int {
```
monkey-mas: In `equals(other: Any?)`, it's good to remove `!!` from `if (this.isTls != that!!.isTls) return false`? 🤔 I don't think there's a chance that NEP can happen though (since we have a sanity check `if (other !is ConnectionSpec) return false`) at the top of the function.
```
override fun equals(other: Any?): Boolean {
  if (other !is ConnectionSpec) return false	
  if (other === this) return true

  val that = other as ConnectionSpec?
  if (this.isTls != that!!.isTls) return false
```

JakeWharton: The cast for that is wrong and shouldn't be nullable

monkey-mas: Ah, right. Removed the wrong cast in `equals(...)`. Thank you!

PR #5288, opened 2019-07-16
Add docs build to CircleCI

adrw:


.circleci/config.yml
```diff
@@ -183,6 +183,47 @@ jobs:
...
+          cat README.md | grep -v 'project website' > docs/index.md
+          cp CHANGELOG.md docs/changelog.md
+          cp CONTRIBUTING.md docs/contributing.md
+      - run: |
+          git remote set-url origin https://github.com/square/okhttp.git
+          git config --global user.email "circleci@squareup.com"
+          git config --global user.name "CircleCI"
+          git cherry-pick bb229b9dcc9a21a73edbf8d936bea88f52e0a3ff
+          git cherry-pick c695732f1d4aea103b826876c077fbfea630e244
+      - run: mkdocs build
```
swankjesse: This will attempt to write to GitHub. I don’t think that makes sense?

.circleci/config.yml
```diff
@@ -218,6 +259,7 @@ workflows:
           filters:
             branches:
               only: master
+      - docs
```
swankjesse: presumably this should be only on master?

.circleci/config.yml
adrw: `mkdocs build` does a build. `mkdocs gh-pages` does the Github write, which is in the deploy script. Running `mkdocs build` in CI ensures that the docs site is deployable.

yschimke: Then we presumably don't need the build on master.  Just PR build?

adrw: Fair point, will update the branch filter 👍

.circleci/config.yml
adrw: Updating to only on non-master, re: conversation above.

.circleci/config.yml
JakeWharton: Still good to build on master since concurrent PRs which are each valid in isolation might produce something which fails when both merged.

.circleci/config.yml
adrw: Update: should run on all branches to ensure docs site is deployable.

yschimke:
Build failure is tracked here - https://github.com/square/okhttp/issues/5316

https://circleci.com/gh/square/okhttp/4533?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

yschimke:
@adrw do you want to rebase?  Should fix the build failure.

adrw:
@yschimke rebased, and @swankjesse 's requested changed have been addressed. Feel free to merge!

.circleci/config.yml
```diff
@@ -197,6 +197,47 @@ jobs:
...
+          ./gradlew \
+            :mockwebserver:dokka \
+            :okhttp-dnsoverhttps:dokka \
+            :okhttp-logging-interceptor:dokka \
+            :okhttp-sse:dokka \
+            :okhttp-tls:dokka \
+            :okhttp-urlconnection:dokka \
+            :okhttp:dokka
+      - run: |
+          cat README.md | grep -v 'project website' > docs/index.md
```
swankjesse: I’m unhappy that this code is now duplicate here and in deploy_website.sh. Can you just invoke that script?

yschimke:
Needs resolve and one remaining comment, otherwise I'll close out in a month or so.

adrw:
@yschimke Thanks, I'll try to address feedback this week so it can be merged.

.circleci/config.yml
```diff
@@ -277,9 +307,11 @@ workflows:
             branches:
               ignore:
                 - gh-pages
-      #- testopenjsse
-      #- testconscrypt
-      #- testcorretto
+              #- testopenjsse
+              #- testconscrypt
+              #- testcorretto
+              only: master
```
yschimke: This change seems unintended.

yschimke:
Appears to be failing - needs to buidl against JDK11 now

```
Starting a Gradle Daemon (subsequent builds will be faster)
> Task :okhttp:copyJavaTemplates
> Task :okhttp:compileKotlin
e: /home/circleci/project/okhttp/src/main/java/okhttp3/internal/platform/Jdk9Platform.kt: (34, 19): Unresolved reference: applicationProtocols
e: /home/circleci/project/okhttp/src/main/java/okhttp3/internal/platform/Jdk9Platform.kt: (42, 43): Unresolved reference: applicationProtocol
e: /home/circleci/project/okhttp/src/main/java/okhttp3/internal/platform/android/Android10SocketAdapter.kt: (44, 38): Unresolved reference: applicationProtocol
e: /home/circleci/project/okhttp/src/main/java/okhttp3/internal/platform/android/Android10SocketAdapter.kt: (59, 19): Unresolved reference: applicationProtocols
```

yschimke:
No movement since feb, and since then I think we'd add to github actions instead of circleci.

PR #5290, opened 2019-07-17
Test migration of CallTest

yschimke:
Testing the migration process, and a possible candidate for remaining in Kotlin since we heavily maintain this test.

Initial Feedback

1)
        assertThat(sink.timeout().hasDeadline()).isFalse

```
Error:(1103, 50) Kotlin: Cannot infer type parameter SELF in val <SELF : AbstractBooleanAssert<SELF!>!> AbstractBooleanAssert<SELF>.isFalse: SELF!
None of the following substitutions
receiver: AbstractBooleanAssert<CapturedTypeConstructor(*)>  arguments: ()
receiver: AbstractBooleanAssert<AbstractBooleanAssert<*>!>  arguments: ()
receiver: AbstractBooleanAssert<Nothing!>  arguments: ()
receiver: AbstractBooleanAssert<AbstractBooleanAssert<Nothing!>!>  arguments: ()
receiver: AbstractBooleanAssert<AbstractBooleanAssert<in AbstractBooleanAssert<*>!>!>  arguments: ()
receiver: AbstractBooleanAssert<AbstractBooleanAssert<CapturedTypeConstructor(*)!>!>  arguments: ()
can be applied to
receiver: AbstractBooleanAssert<*>!  arguments: ()
```

2) It's at the limits of editable Kotlin inside Intellij, a useful stress test for Jetbrains.  Noticeablly worse when there were warnings after initial migration.

3) Migration of RequestBody.create(c, mt) to c.toRequestBody(mt) failed badly every time.
It left it as the incorrect  RequestBody.create(mt, c), which works as Java, but breaks Kotlin.


okhttp/src/main/java/okhttp3/Authenticator.kt
```diff
@@ -113,5 +113,22 @@ interface Authenticator {
...
       override fun authenticate(route: Route?, response: Response): Request? = null
     }
+
+
+    /**
+     * Constructs an interceptor for a lambda. This compact syntax is most useful for inline
+     * interceptors.
+     *
+     * ```
+     * val interceptor = Interceptor { chain: Interceptor.Chain ->
```
yschimke: todo

okhttp/src/test/java/okhttp3/CallTest.kt
```diff
@@ -0,0 +1,4168 @@
...
+    }
+
+    val credential = Credentials.basic("jesse", "secret")
+    client = client.newBuilder()
+        .authenticator(RecordingOkAuthenticator(credential, null))
+        .build()
+
+    try {
+      client.newCall(Request.Builder().url(server.url("/0")).build()).execute()
+      fail()
```
yschimke: I'm tempted to introduce an executeFailing() extension method to avoid this repeated pattern, but the current extra line is pretty explanatory, so avoided for now.

yschimke:
Follow up questions

1) What would a useful CallTest DSL look like?  Does that defeat the purpose.
2) Is it concerning that lines increased?

okhttp/src/test/java/okhttp3/CallTest.kt
swankjesse: There's something in the Kotlin test artifact that does this.

swankjesse:
> assertThat(sink.timeout().hasDeadline()).isFalse

I really don't like how `isFalse` reads like a val access. This should always have parens because it's the side effect of an AssertionError that we care about.

swankjesse:
If this file isn't comfortable in IntelliJ we should cut it up. I dont love that because organizing tests is difficult and not useful.

yschimke:
@swankjesse Should I ship this?  I don't want to start some goldrush to migrate tests, it was part testing the migration, and because we should care about the efficiency of changing this test file?

swankjesse:
I’d prefer to keep this on ice. I’d prefer to do new tests in Kotlin, but the benefits of converting existing tests is rather unclear.

Issue #5151, opened 2019-05-31
Non-IOException subtypes thrown from interceptor never notify Callback


JakeWharton:
A programming error in an interceptor or in OkHttp or an error of any kind will bypass the `Callback` and crash the `Dispatcher` thread leaving your consumer hanging forever. This was reported to Retrofit's Kotlin coroutine support, but is not specific to it.

My proposal is for adding something like
```java
default void onFailure(Call call, Throwable t) {
  if (t instanceof IOException) {
    onFailure(call, (IOException) t);
  }
}
```
and deprecating the existing `onFailure` callback in a 3.14.3 or 3.15 and the equivalent added to 4.0. This will force Retrofit onto Java 8, although given the added safety I think that's acceptable.

swankjesse:
It works for continuations where unexpected exceptions can be caught, but I dislike it otherwise. There’s nothing reasonable to do in onFailure().

JakeWharton:
That is a dangerous assumption. Continuations are just sugar over callbacks. There's nothing specific to them that isn't also true of Rx or futures or promises or this callback. There's plenty you can do when a RuntimeException is the result of an async invocation. Imagine if the synchronous API just blocked your thread forever on anything but an IOException.

swankjesse:
### Counterproposal!

I agree that we should offer a mechanism for callers to release resources after an unexpected exception. But I don't like sending expected IOExceptions and unexpected NullPointerException/OutOfMemoryError/AssertionError failures to the same place. It risks sweeping serious bugs under the rug.

When a Call suffers an unexpected exception, let's do this:
1. Cancel the call by delivering `IOException("canceled")` or similar to the callback.
2. Deliver the actual failure to the UncaughtExceptionHandler.

This keeps the contract between caller and OkHttp simple: you get a response or an exception you can mostly ignore. We say, “OkHttp will cancel calls for you if there is an unexpected exception”

And it allows the real problem to be dealt with, by crashing the app or whatever has been configured.

JakeWharton:
Canceling the call is a very elegant approach. I like it a lot.

Would you put the underlying exception as the cause?

swankjesse:
I think we don’t add it as a cause because it could get double-reported in tracking tools.

yschimke:
You can add suppresses then?

Tolriq:
+1000 :) Tried a few times to ask for something allowing handling of non IOException errors and always rejected.
Currently using added interceptor for that ...

[...]

PR #5296, opened 2019-07-21
Install UncaughtExceptionHandler

yschimke:


yschimke:
Example when disabling the ExceptionCatchingExecutor, causes a hang.

image

yschimke:
Better example

Before PR

image

With PR

image

okhttp-testing-support/src/main/java/okhttp3/testing/UncaughtExceptionHandlerListener.kt
```diff
@@ -0,0 +1,104 @@
...
+ */
+package okhttp3.testing
+
+import org.junit.runner.Description
+import java.io.PrintWriter
+import java.io.StringWriter
+import java.lang.Thread.UncaughtExceptionHandler
+
+/**
+ * installs an aggressive default [UncaughtExceptionHandler] similar to the one found on Android.
```
swankjesse: Nit: https://nedbatchelder.com/blog/201401/comments_should_be_sentences.html

yschimke:
> Let's scope the uncaught exception handler to the test that is currently running? And maybe make our rule a factory for threads that have this thing configured?

My first thought was to configure the thread pool used by Dispatcher, implement afterExecute.  But that seems like tests won't run with the normal code.  And any tests overriding the dispatcher would break it.

So this approach was deliberate to mimic the existing listener. But with parallel runs, it's hard to be sure we catch the right tests.

Can you expand on what you mean by scope to the test? Scoping to the JVM (this approach) or a client/dispatcher are the obvious options.

yschimke:
We also don't want or expect this to fire. 

We will tighten up this logic if we implement https://github.com/square/okhttp/issues/5151

yschimke:
ping @swankjesse 

What are your thoughts on upgrading to junit 5, I think there are cleaner ways to do this if so.

swankjesse:
I think we can do better! Your recent PRs surface a bunch of problems related to threads, unexpected exceptions, and unexpected interruptions.

Why don’t we design an integrated solution? Maybe something that works by wrapping `Runnables`? In tests we know which test caused which `Runnable` to be enqueued, and we could track this.

yschimke:
Will rework.

swankjesse:
I was thinking we could create an ExecutiveService factory or wrapper to manage all of our threads with the same care that we manage sockets.

PR #5297, opened 2019-07-23
Handle localhost as ipv6 without bracket quoting

yschimke:


mockwebserver/src/test/java/okhttp3/mockwebserver/RecordedRequestTest.java
```diff
@@ -102,4 +102,14 @@ private FakeSocket(InetAddress localAddress, int localPort, InetAddress remoteAd
 
     assertThat(request.getRequestUrl().toString()).isEqualTo("http://127.0.0.1/");
   }
+
+  @Test public void testUsesLocalIpv6WithoutBrackets() throws UnknownHostException {
```
monkey-mas: 👍 

nit: we might add a test for IPv4 as well? ;)
e.g. 
```
  @Test public void testLocalhostIpv4() throws UnknownHostException {
    Socket socket =
        new FakeSocket(InetAddress.getByAddress("localhost", new byte[] {127, 0, 0, 1}), 80);

    RecordedRequest request = new RecordedRequest("GET / HTTP/1.1", headers,
        Collections.emptyList(), 0, new Buffer(), 0, socket);

    assertThat(request.getRequestUrl().toString()).isEqualTo("http://localhost/");
  }
```

mockwebserver/src/main/java/okhttp3/mockwebserver/RecordedRequest.kt
```diff
@@ -101,7 +101,11 @@ class RecordedRequest(
       val inetAddress = socket.localAddress
 
       var hostname = inetAddress.hostName
-      if (inetAddress is Inet6Address) {
+      if (inetAddress is Inet6Address && hostname.contains(':')) {
+        // hostname is likely some form representing the ipv6 bytes
```
monkey-mas: super-nit: it'd be nice to write down a formal name, e.g. `IPv6` ? :)

monkey-mas:
LGTM! :)

mockwebserver/src/test/java/okhttp3/mockwebserver/RecordedRequestTest.java
yschimke: Thanks!

Issue #5304, opened 2019-07-25
Flaky test: WholeOperationTimeoutTest


yschimke:
https://circleci.com/gh/square/okhttp/4441

https://4441-5152285-gh.circle-artifacts.com/0/home/circleci/reports/okhttp/build/reports/tests/test/classes/okhttp3.WholeOperationTimeoutTest.html#noTimeout

```
Events (16)
[0 ms] callStart: Request{method=POST, url=http://localhost:60795/}
[0 ms] dnsStart: localhost
[0 ms] dnsEnd: [localhost/127.0.0.1]
[0 ms] connectStart: localhost/127.0.0.1:60795 DIRECT
[0 ms] connectEnd: http/1.1
[0 ms] connectionAcquired: Connection{localhost:60795, proxy=DIRECT hostAddress=localhost/127.0.0.1:60795 cipherSuite=none protocol=http/1.1}
[0 ms] requestHeadersStart
[0 ms] requestHeadersEnd
[0 ms] requestBodyStart
[254 ms] requestBodyEnd: byteCount=6
[259 ms] responseHeadersStart
[936 ms] responseHeadersEnd: Response{protocol=http/1.1, code=200, message=OK, url=http://localhost:60795/}
[936 ms] responseBodyStart
[1187 ms] responseFailed: java.net.SocketException: Socket closed
[1187 ms] connectionReleased
[1187 ms] callFailed: java.io.InterruptedIOException: timeout
```

PR #5305, opened 2019-07-25
Increase timeout for flaky test

yschimke:
Increasing the timeout which only has 250ms tolerance after inbuilt delays etc. Issue info is here https://github.com/square/okhttp/issues/5304

### DIFF
--- a/okhttp/src/test/java/okhttp3/WholeOperationTimeoutTest.java
+++ b/okhttp/src/test/java/okhttp3/WholeOperationTimeoutTest.java
@@ -23,6 +23,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
 import okhttp3.mockwebserver.MockResponse;
 import okhttp3.mockwebserver.MockWebServer;
+import okhttp3.testing.Flaky;
 import okio.BufferedSink;
 import org.junit.Before;
 import org.junit.Rule;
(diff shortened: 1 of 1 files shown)


okhttp/src/test/java/okhttp3/WholeOperationTimeoutTest.java
```diff
@@ -286,7 +287,10 @@ public void timeoutFollowingRedirectOnNewConnection() throws Exception {
     }
   }
 
+  @Flaky
```
yschimke: To be clear it shouldn't be flaky after this point, but this will record the events for every test run, so we can observe typical response times.

yschimke:
Passing run https://4456-5152285-gh.circle-artifacts.com/0/home/circleci/reports/okhttp/build/reports/tests/test/classes/okhttp3.WholeOperationTimeoutTest.html

```
Events (18)
[0 ms] callStart: Request{method=POST, url=http://localhost:52107/}
[0 ms] proxySelectStart: http://localhost:52107/
[0 ms] proxySelectEnd: [DIRECT]
[0 ms] dnsStart: localhost
[0 ms] dnsEnd: [localhost/127.0.0.1]
[0 ms] connectStart: localhost/127.0.0.1:52107 DIRECT
[0 ms] connectEnd: http/1.1
[0 ms] connectionAcquired: Connection{localhost:52107, proxy=DIRECT hostAddress=localhost/127.0.0.1:52107 cipherSuite=none protocol=http/1.1}
[0 ms] requestHeadersStart
[0 ms] requestHeadersEnd
[0 ms] requestBodyStart
[250 ms] requestBodyEnd: byteCount=6
[250 ms] responseHeadersStart
[500 ms] responseHeadersEnd: Response{protocol=http/1.1, code=200, message=OK, url=http://localhost:52107/}
[500 ms] responseBodyStart
[751 ms] responseBodyEnd: byteCount=65536
[751 ms] connectionReleased
[751 ms] callEnd
```

okhttp/src/test/java/okhttp3/WholeOperationTimeoutTest.java
swankjesse: 👍🏻

Issue #5138, opened 2019-05-28
Handshake certificates are unclean


swankjesse:
We’re cleaning them for the certificate Pinner but not otherwise. We should put cleaned certificates into the handshake always.

Only need to be careful of the performance consequences of this.

https://github.com/Babylonpartners/certificate-transparency-android/issues/4#issuecomment-496485839

yschimke:
This is a good candidate for 4.1

yschimke:
Draft PR, will add tests, feel free to comment on approach.

https://github.com/square/okhttp/pull/5311/files

swankjesse:
Shipped in 4.2!

PR #5311, opened 2019-07-29
Handshake returns cleaned peer certificates

yschimke:
https://github.com/square/okhttp/issues/5138

Pass through clean certificates with only a single pass of cert cleaning.


okhttp/src/main/java/okhttp3/internal/connection/RealConnection.kt
```diff
@@ -370,9 +372,18 @@ class RealConnection(
         }
       }
 
+      handshake = if (certificateChainCleaner != null) {
+        val cleanedCertificates =
+            certificateChainCleaner.clean(unverifiedHandshake.peerCertificates, address.url.host)
+        Handshake.get(unverifiedHandshake.tlsVersion, unverifiedHandshake.cipherSuite,
```
yschimke: I know this is a bit yuck, but this is limiting

```
val unverifiedHandshake = sslSocketSession.handshake()
```

Plus the design of X509TrustManager, which is a JDK class outside our control.

yschimke:
@swankjesse Any advice for testing here?  We test RealConnection through CallTest, and the JVM code around KeyManager will likely send the clean certificates.

okhttp/src/main/java/okhttp3/internal/connection/RealConnection.kt
yschimke: Also this method probably is probably the most brittle in the codebase, so I'm hesitant to restructure more than this.

swankjesse:
This situation is frustrating.

Cleaning the certificates could be non-trivial, so this has the potential of breaking or slowing code that works otherwise.

I’d like to land 4.1 without this, and then put this in 4.2 early with lots of bake time.

swankjesse:
That said, the change itself looks great!

As for testing . . . wanna test with a fake cleaner?

yschimke:
> That said, the change itself looks great!
> 
> As for testing . . . wanna test with a fake cleaner?

That route isn't great.  It makes new internal APIs for setting the cleaner in OkHttpClient.Builder.  Cleaner currently comes from the Platform, or via the TrustManager if you customise that.  But they are on a public API, so they tend to show up for Java clients.

image

yschimke:
@swankjesse I'm tempted to move this into sslSocketSession.handshake()

But also wondering whether we should retain the uncleaned certificates?  Handshake could have both, potentially interesting when trying to understand why a decision was made.  e.g. if certificates are served that have multiple root CAs for different clients.  Not sure if that happens in the wild.

swankjesse:
We can hide things from Java clients with a `-` prefix on the `@JvmName`. Does that help? (We should probably do this universally!)

swankjesse:
Can we always use the cleaner from the pinner? What’s the benefit of wiring it through the connection? If _that_ one is null we shouldn’t bother.

I’m still unsatisfied with the performance cost here. It has a non-zero cost and a nearly-zero benefit.

yschimke:
> Can we always use the cleaner from the pinner? What’s the benefit of wiring it through the connection? If _that_ one is null we shouldn’t bother.
> 
> I’m still unsatisfied with the performance cost here. It has a non-zero cost and a nearly-zero benefit.

There is only one pass of cleaning, is it the additional copy of Handshake you are concerned about? It used to happen at CertificatePinner.kt, line 146.  

The benefit is that callers can rely on the handshake certificates being those that we trust and validate.  If we choose to expose the handshake certificates (public API), I'd argue that we should provide the correct ones.

yschimke:
Now deferring cleaning until we either 

1) need to check certificate pinning
2) caller looks at the peer certificates

But only done at most once.

yschimke:
n.b. I removed the test because the change otherwise becomes awkward with OkHttpClient, since that constructor generally always fetches a new cleaner from Platform :(

okhttp/src/main/java/okhttp3/Handshake.kt
```diff
@@ -41,12 +41,15 @@ class Handshake private constructor(
...
 
-  /** Returns a possibly-empty list of certificates that identify the remote peer. */
-  @get:JvmName("peerCertificates") val peerCertificates: List<Certificate>,
-
   /** Returns a possibly-empty list of certificates that identify this peer. */
-  @get:JvmName("localCertificates") val localCertificates: List<Certificate>
+  @get:JvmName(
+      "localCertificates") val localCertificates: List<Certificate>,
+
+  private val peerCertificatesFn: () -> List<Certificate>
```
swankjesse: Neato!

swankjesse: Does this need to be a val?

okhttp/src/main/java/okhttp3/Handshake.kt
```diff
@@ -152,20 +155,20 @@ class Handshake private constructor(
...
       val localCertificates = localCertificates
       val localCertificatesList = if (localCertificates != null) {
         immutableListOf(*localCertificates)
       } else {
         emptyList()
       }
 
-      return Handshake(tlsVersion, cipherSuite, peerCertificatesList, localCertificatesList)
+      return Handshake(tlsVersion, cipherSuite, localCertificatesList) {
+        if (peerCertificates != null) {
```
swankjesse: I think it’d be simpler to continue to compute the `peerCertificatesList` eagerly and just make this a boring function. Deferring the immutableListOf() call is more complexity than necessary!

okhttp/src/main/java/okhttp3/Handshake.kt
```diff
@@ -183,8 +186,8 @@ class Handshake private constructor(
       peerCertificates: List<Certificate>,
       localCertificates: List<Certificate>
     ): Handshake {
-      return Handshake(tlsVersion, cipherSuite, peerCertificates.toImmutableList(),
-          localCertificates.toImmutableList())
+      return Handshake(tlsVersion, cipherSuite, localCertificates.toImmutableList(),
+          { peerCertificates.toImmutableList() })
```
swankjesse: you gotta call toImmutableList() early, otherwise mutating the list after the call impacts the Handshake

okhttp/src/main/java/okhttp3/Handshake.kt
swankjesse: This wants a comment!

okhttp/src/main/java/okhttp3/CertificatePinner.kt
```diff
@@ -138,29 +138,31 @@ class CertificatePinner internal constructor(
...
   @Throws(SSLPeerUnverifiedException::class)
   fun check(hostname: String, peerCertificates: List<Certificate>) {
-    var peerCertificates = peerCertificates
+    return check(hostname) {
+      (certificateChainCleaner?.clean(peerCertificates, hostname) ?: peerCertificates)
+          .map { it as X509Certificate }
+    }
+  }
+
+  internal fun check(hostname: String, cleanedPeerCertificatesFn: () -> List<X509Certificate>) {
```
swankjesse: I think passing a lambda here is too surprising; it might be better to make this take `cleanedCertificates`, and then make the one callsite guard the call with like `hasPins()` check

okhttp/src/main/java/okhttp3/Handshake.kt
yschimke: Good catch

okhttp/src/main/java/okhttp3/CertificatePinner.kt
yschimke: It's not a simple check without cost if there is pinning in place.  e.g. a client configures pinning on their API host, but not on the CDN (unknown hosts).

In this case, `findMatchingPins(hostname)` is the check we need to do externally rather than hasPins.

okhttp/src/main/java/okhttp3/Handshake.kt
```diff
@@ -146,26 +150,22 @@ class Handshake private constructor(
       if ("NONE" == tlsVersionString) throw IOException("tlsVersion == NONE")
       val tlsVersion = TlsVersion.forJavaName(tlsVersionString)
 
-      val peerCertificates: Array<Certificate>? = try {
-        peerCertificates
+      val peerCertificatesCopy = try {
+        peerCertificates.toImmutableList()
       } catch (_: SSLPeerUnverifiedException) {
-        null
+        listOf<Certificate>()
```
yschimke: I should change this to emptyList()

okhttp/src/main/java/okhttp3/CertificatePinner.kt
yschimke: It's also internal API, for a class with public API.

swankjesse: SGTM

okhttp/src/main/java/okhttp3/Handshake.kt
```diff
@@ -41,12 +41,16 @@ class Handshake private constructor(
...
 
-  /** Returns a possibly-empty list of certificates that identify the remote peer. */
-  @get:JvmName("peerCertificates") val peerCertificates: List<Certificate>,
-
   /** Returns a possibly-empty list of certificates that identify this peer. */
-  @get:JvmName("localCertificates") val localCertificates: List<Certificate>
+  @get:JvmName(
+      "localCertificates") val localCertificates: List<Certificate>,
+
+  // Delayed provider of peerCertificates, to allow lazy cleaning.
```
swankjesse: Nice.

swankjesse:
I think this is a very nice improvement.

PR #5312, opened 2019-07-29
Avoid gradle assemble in Travis

yschimke:
Two changes

1) Externally - disabled here https://travis-ci.org/square/okhttp/settings
2) Travis config avoids assemble task


swankjesse:
Need me to flip a setting?

Issue #5266, opened 2019-07-03
407 proxy error is reported with proxy that doesn't require authentication/credentials


tincho5588:
This was working fine on 3.11.0.
After upgrading to 3.14.1, if using a proxy that does not require authentication, okHttp will throw an exception with "Failed to authenticate with proxy" message upon execution a Request. This is wrong, as the proxy does not require credentials or anything, so the request should be fullfilled without further issues.

swankjesse:
Executable test case?

tincho5588:
Not sure what you mean by "Executable test case?"

swankjesse:
I want this: http://www.sscce.org/

tincho5588:
There are no executable test cases, nor will I write them...
Set up a proxy that requires no authentication, configure an android device to go through it in the network settings, and make a simple HTTP query with your library... You'll see a 407...

swankjesse:
I'm unable to reproduce this issue. Are you talking about preemptive auth?

https://square.github.io/okhttp/3.x/okhttp/okhttp3/Authenticator.html

tincho5588:
Yes, but there shouldn't be any preemptive auth in this case, as the proxy does not require it...
Check this out:
https://github.com/square/okhttp/blob/okhttp_3.14.x/okhttp/src/main/java/okhttp3/internal/connection/RealConnection.java#L416
I'm hitting that code block, the response code I get should be HTTP_OK, not HTTP_PROXY_AUTH...
For the record, this is on 3.14.1...

swankjesse:
Could you share a stacktrace or something?

tincho5588:
I don't have one right now, because the exception is handled... But I could work on getting one over the week...
Only downside is that I have to limit the stack Trace to calls on your own lib only, so I don't break my DNA...
The code I'm working with is proprietary, so there's a lot I cannot share...

yschimke:
https://github.com/square/okhttp/pull/5323

```
I/TestRunner( 4816): started: testProxyRequest(okhttp.android.test.OkHttpTest)
I/MonitoringInstr( 4816): Activities that are still in CREATED to STOPPED: 0

D/OkHttp  ( 4816): [0 ms] callStart: Request{method=GET, url=https://www.proxy-list.download/api/v1/get?type=http&anon=elite&country=US}
D/OkHttp  ( 4816): [2 ms] proxySelectStart: https://www.proxy-list.download/
[...]

PR #5323, opened 2019-08-02
Android test using a proxy

yschimke:
Not sure whether to land this, using to confirm https://github.com/square/okhttp/issues/5266

It's not a CI worthy test, but it's high value for reproducing an external bug report.


yschimke:
Saving as PR for easy test reproductions

PR #5332, opened 2019-08-08
Avoid SecurityException on ProxySelector.getDefault when proxy set

yschimke:


yschimke:
We can. No change required then. But the counter argument is this is clearly part of the public API. 

Is seems like simple code to make it just work inside of containers with what is after all a nice sensible fallback default.

swankjesse:
I think we do need a change; we need to defer looking up the default proxy.

Is there a particular container that does this by default? I want to discourage developers from using the security manager to shoot themselves in the foot. I assume if you're using a container you're probably also using a proxy server to escape out to the Internet so defaulting to no proxy hides a problem.

yschimke:
OK, that seems like a good fix.  If you hardcode Proxy.NO_PROXY you won't fail because of effectively useless call.  This is non contentious and allows a workaround.

I'll make this change for now.

My remaining concern is that there are two distinct roles here, who are not connected.  But it's niche, so I don't want to block on this. 

1) library or app developer who uses OkHttp and relies on the proxies being configured external (system property etc).  They may or may not be aware of the security manager enforced by the container. 
2) User in a corporate environment who just wants to set a proxy before launching the command.

The most compatible code for "library or app developer" to write is to call ProxySelector.getDefault themselves, catch and swallow/log the error, and then set the results of getDefault

okhttp/src/test/java/okhttp3/OkHttpClientTest.java
```diff
@@ -283,4 +285,18 @@
...
   }
+
+  @Test public void testProxyDefaults() {
+    OkHttpClient client = new OkHttpClient.Builder().build();
+    assertThat(client.proxy()).isNull();
+    assertThat(client.proxySelector()).isNotInstanceOf(NullProxySelector.class);
+
+    client = new OkHttpClient.Builder().proxy(Proxy.NO_PROXY).build();
+    assertThat(client.proxy()).isSameAs(Proxy.NO_PROXY);
+    assertThat(client.proxySelector()).isInstanceOf(NullProxySelector.class);
```
swankjesse: dig.

swankjesse:
Library authors aught to to let application code provide the OkHttp instance. That makes everything easy.

Issue #3957, opened 2018-04-05
Provide a way to shutdown ConnectionPool.executor and stop WatchDog threads


robstoll:
What kind of issue is this?
 - [x] Feature Request. 

I am using OkHttpClient in a maven java:exec context and get the following error:
```maven
[WARNING] thread Thread[OkHttp ConnectionPool,5,ch.loewenfels.depgraph.runner.Main] was interrupted but is still alive after waiting at least 15000msecs
[WARNING] thread Thread[OkHttp ConnectionPool,5,ch.loewenfels.depgraph.runner.Main] will linger despite being asked to die via interruption
[WARNING] thread Thread[Okio Watchdog,5,ch.loewenfels.depgraph.runner.Main] will linger despite being asked to die via interruption
[WARNING] NOTE: 2 thread(s) did not finish despite being asked to  via interruption. This is not a problem with exec:java, it is a problem with the running code. Although not serious, it should be remedied.
[WARNING] Couldn't destroy threadgroup org.codehaus.mojo.exec.ExecJavaMojo$IsolatedThreadGroup[name=ch.loewenfels.depgraph.runner.Main,maxpri=10]
```
I am aware of that I can set `cleanupDaemonThreads` to false in the maven java:exec configuration but that seems wrong to me. It would be nicer if I can shutdown everything myself rather than relying on that the threads stop at some point. In the light of the many bugs concerning similar problematics (memory leak, thread does not stop etc.) I would appreciate if I can stop everything myself.

yschimke:
See the documentation for Client

https://github.com/square/okhttp/blob/5fe3cc2d089810032671d6135ad137af6f491d28/okhttp/src/main/java/okhttp3/OkHttpClient.java#L99

It tells you how to shutdown ASAP

robstoll:
nah... that does not work 👎
Please repoen the issue. This only shuts the executorService of the dispatcher down but not the executorService of the connection poll and neither does it stop the watchdog threads

yschimke:
I think the preview only shows a certain number of lines, but if you click through it also clears up the connection pool and even the cache.

robstoll:
You're right it does not show everything but also the rest does not help. `evictAll` doesn't shutdown the `ConnectionPool.executorService`. I could shut it down myself with the help of reflection but accessing a private field is ugly as hell. Also, I don't see how I can get hold oft he watchdog threads to kind of kill them. I hope you understand the problem now

yschimke:
[...]

PR #5334, opened 2019-08-08
Allow interrupt of singleton cleanup thread

yschimke:
If container aggressively interrupts the RealConnectionPool cleanup thread, then take notice to avoid waiting 60 seconds for daemon thread to cleanup normally.  However leave things in a sane state, so next connection established will restart it, and OkHttp will noisily log if this is done improperly without closing connections.

Related to https://github.com/square/okhttp/issues/3957 and Maven which violently interrupts daemon threads and noisily complains if they swallow interrupt exceptions.

TODO add tests if there is loose agreement this is correct logic.

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/connection/RealConnectionPool.kt
+++ b/okhttp/src/main/java/okhttp3/internal/connection/RealConnectionPool.kt
@@ -49,8 +49,7 @@ class RealConnectionPool(
           this@RealConnectionPool.lockAndWaitNanos(waitNanos)
         } catch (ie: InterruptedException) {
           cleanupRunning = false
-          val count = connectionCount()
-          if (count > 0) {
+          if (connectionCount() > 0) {
             Platform.get().log(
                 Platform.WARN,
                 "${Thread.currentThread().name} interrupted without cleanly closing connections",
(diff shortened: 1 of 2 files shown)


yschimke:
Not confident this will pass consensus, but Maven is a common enough container that we may want to allow for this usecase.  They are unlikely to change their thread handling for us.

swankjesse:
Alright, lets be a little less container-hostile.

Can you change it to call `evictAll()` once interrupted? Something like this might work?

```
      while (true) {
        val waitNanos = cleanup(System.nanoTime())
        if (waitNanos == -1L) return
        try {
          this@RealConnectionPool.lockAndWaitNanos(waitNanos)
        } catch (_: InterruptedException) {
          // Will cause the thread to exit unless other connections are created!
          evictAll() 
        }
      }
```

yschimke:
I'm a bit surprised you prefer this mode.  But it does keep the state in sync.  connections > 0 = cleanup running. It makes a weird API that interrupting a specific thread is a great way to cleanup all connections and threads forcefully.

yschimke:
Looking into test failures

image

swankjesse:
Only thing that’ll annoy container users is that our executor service will stick around for 60 seconds. Should we tighten this up to 1 second? 

(And perhaps we should use the same ExecutorService for this, HTTP/2’s readers, and HTTP/2’s callbacks.)

yschimke:
@swankjesse I like that, I think we should aim for using signals like evicting the connection pool to stop background tasks nicely.  So doing the right thing appears to just work within 1 second as you say, I'll raise a tracking issue.

PR #5339, opened 2019-08-09
Add TrustManager accessor to OkHttpClient.

AlexSzlavik:
We are able to access the SSLSocketFactory from the OkHttpClient but not the TrustManager used. Since there is no way to recover the used trust manager from the socket factory, we are adding explicit accessors.


swankjesse:
Thanks!

Issue #5342, opened 2019-08-10
Add test specifically for Session Resumption


yschimke:
We should have a test designed to prove that session resumption works in JDK11+

re: https://github.com/square/okhttp/pull/5325#issuecomment-517806509

yschimke:
Interesting page with nowhere else to post :) https://java.com/en/jre-jdk-cryptoroadmap.html

yschimke:
Android is now tested.  But we should have some tests for desktop VMs.

yschimke:
@swankjesse Any suggestions how to make a live status page? If it's an output of CI, I think we need it to combine results from multiple builds.

Going to close this as an open task, but I might try to make something that actually tells us what differences we have between platforms etc.  It seems really patchy on different JDKs.  Conscrypt just worked but only defaults to 10 sessions.  It also has a persistent cache API which we don't use or expose.

swankjesse:
I’d actually prefer a static page with what we have test coverage for and encourage our users to use.

PR #5345, opened 2019-08-10
Debug session reuse

yschimke:
tl;dr; Conscrypt works well and is clear about results.  JSSE is confusing, and depends on server.

Might need a table and record which combinations work.

Relates to https://github.com/square/okhttp/issues/5342

The defaults vary a lot and may affect overall performance significantly.


yschimke:
cc @swankjesse any advice?

PR #5347, opened 2019-08-11
Test Android session reuse

yschimke:
Test for https://github.com/square/okhttp/issues/5342

### DIFF
--- a/android-test/src/androidTest/java/okhttp/android/test/OkHttpTest.kt
+++ b/android-test/src/androidTest/java/okhttp/android/test/OkHttpTest.kt
@@ -23,11 +23,13 @@ import okhttp3.Connection
 import okhttp3.EventListener
 import okhttp3.OkHttpClient
 import okhttp3.Protocol
+import okhttp3.RecordingEventListener
 import okhttp3.Request
 import okhttp3.TlsVersion
 import okhttp3.mockwebserver.MockResponse
 import okhttp3.mockwebserver.MockWebServer
 import okhttp3.tls.internal.TlsUtil.localhost
+import okio.ByteString.Companion.toByteString
 import org.junit.After
 import org.junit.Assert.assertEquals
 import org.junit.Assert.assertTrue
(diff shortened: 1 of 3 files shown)


okhttp-testing-support/src/main/java/okhttp3/RecordingEventListener.java
```diff
@@ -75,7 +76,7 @@ private void logEvent(CallEvent e) {
     CallEvent startEvent = e.closes();
 
     if (startEvent != null) {
-      assertThat(eventSequence).contains(startEvent);
+      assertTrue(eventSequence.contains(startEvent));
```
yschimke: Not compatible with older android due to internal stream usage.

yschimke:
image

okhttp-testing-support/src/main/java/okhttp3/RecordingEventListener.java
yschimke: n.b. this is also a move to testing support

android-test/src/androidTest/java/okhttp/android/test/OkHttpTest.kt
```diff
@@ -190,6 +193,75 @@ class OkHttpTest {
...
+
+    client.connectionPool.evictAll()
+    assertEquals(0, client.connectionPool.connectionCount())
+
+    client.newCall(request).execute().use { response ->
+      assertEquals(200, response.code)
+    }
+
+    assertEquals(2, sessionIds.size)
+    assertEquals(sessionIds[0], sessionIds[1])
```
swankjesse: Sweet test.

ragner: Hi @yschimke, I am trying this session reuse test with Android 7.1.1 but it is not working for me. 
What version of Android is compatible with session reuse?

yschimke: I think that's a question for https://github.com/google/conscrypt

Issue #5346, opened 2019-08-11
Support OpenJSSE (TLSv1.3 on JDK 8)


yschimke:
FWIW it's on the Oracle roadmap

https://java.com/en/jre-jdk-cryptoroadmap.html

and supported here as well

https://github.com/openjsse/openjsse

Issue #4667, opened 2019-02-27
Provide a JUnit 5 extension for MockWebServer


fbiville:
I'm trying to get rid of JUnit 4 in my project and the last remaining bit is `MockWebServer` which depends on `org.junit.rules.ExternalResource`.

That is not a blocker as `org.junit.jupiter:junit-jupiter-migrationsupport` provides experimental support for this. 

However, it would be great to provide an implementation of `MockWebServer` following the Junit 5 [extension model](https://junit.org/junit5/docs/current/user-guide/#extensions).

yschimke:
This was discussed here https://github.com/square/okhttp/pull/3139

It's considered public api, so the change has to be very carefully handled.  Introducing a junit library is simpler but essentially wouldn't need to be a part of the core project either.

fbiville:
> but essentially wouldn't need to be a part of the core project either

Not sure I am following: by core project, do you mean OkHttp or MockWebServer? 

Given JUnit 4 and 5 can be both in the classpath afaik, what about adding JUnit 5 to `MockWebServer` and introducing a new class like `MockWebServer(JUnit5)Extension` without introducing visible changes to the existing rule?

JakeWharton:
You would still need 4 on the classpath because it's a super type so what's
the advantage of doing that?

swankjesse:
Also – we should assume there will be another incompatible library called JUnit 6 in a few years.

JakeWharton:
Yeah we need to just break this in OkHttp 4 and decouple the need for a
server from the idea that you're inside a test runner.

ati90ati:
As I know Junit 5 supports Java 11.
It will be a good idea to decouple and to rely only on the Junit 5, so we don't need to have the Junit 4 dependency with Junit 5 dependencies on the classpath because the MockWebServer dependency.

pyricau:
👋 just dumping a note that this is causing https://github.com/square/leakcanary/issues/1649

tokuhirom:
I'm migrating my project to junit5 and found this issue.  And, I guess `org.junit.jupiter:junit-jupiter-migrationsupport` may not help me. Since `org.junit.rules.ExternalResource` is not included in `junit-jupiter-migration-support`.
If a user wants to use a mockwebserver with junit5, the user needs to add `junit:junit` in the dependency tree.

yschimke:
[...]

PR #5348, opened 2019-08-11
Test for OpenJSSE

yschimke:
https://github.com/openjsse/openjsse from Azul

TLSv1.3 on JDK 8

At the moment, no code changes or custom support enabling HTTP/2, only a test that it works in JDK8.

### DIFF
--- a/okhttp/src/test/java/okhttp3/OpenJSSETest.kt
+++ b/okhttp/src/test/java/okhttp3/OpenJSSETest.kt
@@ -32,7 +32,7 @@ class OpenJSSETest {
   }
 
   @Test
-  fun testX() {
+  fun testTlsv13Works() {
     enableTls()
 
     server.enqueue(MockResponse().setBody("abc"))
@@ -49,6 +49,7 @@ class OpenJSSETest {
 
   private fun enableTls() {
     // Generate a self-signed cert for the server to serve and the client to trust.
+    // can't use localhost with a non OpenJSSE trust manager
     val heldCertificate = HeldCertificate.Builder()
         .commonName("localhost")
         .addSubjectAlternativeName(InetAddress.getByName("localhost").canonicalHostName)
--- a/okhttp/src/test/java/okhttp3/OpenJSSETest.kt
+++ b/okhttp/src/test/java/okhttp3/OpenJSSETest.kt
@@ -21,6 +21,7 @@ import okhttp3.mockwebserver.MockWebServer
 import okhttp3.testing.PlatformRule
 import okhttp3.tls.HandshakeCertificates
 import okhttp3.tls.HeldCertificate
+import org.assertj.core.api.Assertions.assertThat
 import org.junit.After
 import org.junit.Assert.assertEquals
 import org.junit.Before
@@ -65,6 +66,8 @@ class OpenJSSETest {
       assertEquals(200, response.code)
       assertEquals(TlsVersion.TLS_1_3, response.handshake?.tlsVersion)
       assertEquals(Protocol.HTTP_1_1, response.protocol)
+
+      assertThat(response.exchange?.connection()?.socket()).isInstanceOf(SSLSocketImpl::class.java)
     }
   }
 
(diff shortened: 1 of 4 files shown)


okhttp/src/test/java/okhttp3/OpenJSSETest.kt
```diff
@@ -0,0 +1,67 @@
...
+import java.security.Security
+
+class OpenJSSETest {
+  @JvmField @Rule val clientTestRule = OkHttpClientTestRule()
+  @JvmField @Rule val server = MockWebServer()
+  lateinit var client: OkHttpClient
+
+  @Before
+  fun setUp() {
+    Security.insertProviderAt(OpenJSSE(), 1)
```
yschimke: We may want to use junit 5 https://junit.org/junit5/docs/current/api/org/junit/jupiter/api/parallel/ResourceLock.html

To avoid weird interactions via providers?

yschimke:
Relates to https://github.com/square/okhttp/issues/5346

yschimke:
image

yschimke:
FWIW Error if you reuse the TlsUtil.localhost

```
Caused by: java.lang.IllegalArgumentException: TLSv1.3
	at sun.security.ssl.ProtocolVersion.valueOf(ProtocolVersion.java:187)
	at sun.security.ssl.X509TrustManagerImpl.checkTrusted(X509TrustManagerImpl.java:214)
	at sun.security.ssl.X509TrustManagerImpl.checkServerTrusted(X509TrustManagerImpl.java:132)
	at org.openjsse.sun.security.ssl.CertificateMessage$T13CertificateConsumer.checkServerCerts(CertificateMessage.java:1308)
	at org.openjsse.sun.security.ssl.CertificateMessage$T13CertificateConsumer.onConsumeCertificate(CertificateMessage.java:1199)
	at org.openjsse.sun.security.ssl.CertificateMessage$T13CertificateConsumer.consume(CertificateMessage.java:1146)
```

yschimke:
Current limitation, it won't negotiate HTTP/2 without one of

1) custom SSLSocketFactory 
2) An OpenJSSE Platform

yschimke:
image

okhttp/src/test/java/okhttp3/OpenJSSETest.kt
```diff
@@ -0,0 +1,102 @@
...
+    val s = factory.createSocket() as SSLSocketImpl
+
+    // A Public API available is available to use in a custom Platform
+    s.setHandshakeApplicationProtocolSelector { _, _ -> HTTP_2.toString() }
+    assertEquals(listOf("TLSv1.3", "TLSv1.2", "TLSv1.1", "TLSv1"), s.enabledProtocols.toList())
+  }
+
+  private fun enableTls() {
+    // Generate a self-signed cert for the server to serve and the client to trust.
+    // can't use localhost with a non OpenJSSE trust manager
```
swankjesse: is `localhost` here, `TlsUtil.localhost` ? I think that’s ambiguous

okhttp/src/test/java/okhttp3/OpenJSSETest.kt
swankjesse: JUnit 5 is a big can of worms because they broke API compatibility and MockWebServer needs JUnit 4.

swankjesse:
I think I like this better than Jetty-alpn?

okhttp/src/test/java/okhttp3/OpenJSSETest.kt
yschimke: Yep, I'll confirm.

yschimke:
> I think I like this better than Jetty-alpn?

Me too, 

Let me implement a PR for openjsse and as a provided jar like Conscrypt.  We can make a call to support as an optional dependency or abandon.

Two main risks

1) More options/surface area for tests.  But essentially if someone on the Desktop JVM uses OpenJSSE, then ideally we would support it with best possible features (TLSv1.3 + HTTP/2).
2) We add it and noone ever uses it...

okhttp/src/test/java/okhttp3/OpenJSSETest.kt
yschimke: Do you have a plan for how that will work? https://github.com/square/okhttp/issues/4667 ?

It would be good to break that dependency and have it working with both.  Over time more users will expect Junit 5.

PR #5355, opened 2019-08-12
Use the application plugin instead of shadow

swankjesse:


JakeWharton:
The downside is that you no longer get a single binary. Don't really have a preference. I tend to make my own fat jars and then make them self-executable with custom tasks. Then you have something you can run as if it were a "native" binary: https://github.com/JakeWharton/dex-member-list/blob/0cadcff27ac6238430e8ecf2e781a6824d0425d3/dex-members-list/build.gradle#L48-L83

swankjesse:
Ahh, I thought that the application plugin was doing that for me, but it’s far less satisfying.

```
$ jar tvf  okcurl/build/distributions/okcurl-4.1.0.zip
okcurl-4.1.0/
okcurl-4.1.0/lib/
okcurl-4.1.0/lib/okcurl-4.1.0.jar
okcurl-4.1.0/lib/okhttp-logging-interceptor-4.1.0.jar
okcurl-4.1.0/lib/okhttp-4.1.0.jar
okcurl-4.1.0/lib/picocli-4.0.1.jar
okcurl-4.1.0/lib/guava-27.0.1-jre.jar
okcurl-4.1.0/lib/okio-2.2.2.jar
okcurl-4.1.0/lib/kotlin-stdlib-1.3.41.jar
okcurl-4.1.0/lib/failureaccess-1.0.1.jar
okcurl-4.1.0/lib/listenablefuture-9999.0-empty-to-avoid-conflict-with-guava.jar
okcurl-4.1.0/lib/jsr305-3.0.2.jar
okcurl-4.1.0/lib/checker-qual-2.5.2.jar
okcurl-4.1.0/lib/error_prone_annotations-2.2.0.jar
okcurl-4.1.0/lib/j2objc-annotations-1.1.jar
okcurl-4.1.0/lib/animal-sniffer-annotations-1.17.jar
okcurl-4.1.0/lib/kotlin-stdlib-common-1.3.41.jar
okcurl-4.1.0/lib/annotations-13.0.jar
okcurl-4.1.0/bin/
okcurl-4.1.0/bin/okcurl.bat
okcurl-4.1.0/bin/okcurl
```

yschimke:
I hope in 6 months we can just use graal.  It's good for both CLI and fast start "serverless" apps.  But a bit twitchy with SSL (sunec, cacerts etc).

PR #5357, opened 2019-08-12
Change gradle resolution ordering to fix jdk8alpn

yschimke:


yschimke:
Testing a second fix for my own edification.  Resolving dependencies seems awkward as it crystallizes things before we are ready.

PR #5369, opened 2019-08-14
OpenJSSE Platform support

yschimke:
Support for HTTP/2 with https://github.com/openjsse/openjsse

Caveats

1) Tests based on MockWebserver (on JDK8) won't work without larger changes
2) Unclear what the overall status is of the project is, and edge cases like cipher selections on older VMs may need testing.

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/platform/OpenJSSEPlatform.kt
+++ b/okhttp/src/main/java/okhttp3/internal/platform/OpenJSSEPlatform.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2014 Square, Inc.
+ * Copyright (C) 2019 Square, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -30,7 +30,7 @@ import javax.net.ssl.X509TrustManager
  * Platform using OpenJSSE (https://github.com/openjsse/openjsse) if installed as the first
  * Security Provider.
  *
- * Requires org.conscrypt:conscrypt-openjdk-uber >= 2.1.0 on the classpath.
+ * Requires org.openjsse:openjsse >= 1.1.0 on the classpath.
  */
 class OpenJSSEPlatform private constructor() : Platform() {
   private val provider: Provider = org.openjsse.net.ssl.OpenJSSE()
--- a/okhttp/src/main/java/okhttp3/internal/platform/OpenJSSEPlatform.kt
+++ b/okhttp/src/main/java/okhttp3/internal/platform/OpenJSSEPlatform.kt
@@ -65,12 +65,6 @@ class OpenJSSEPlatform private constructor() : Platform() {
         sslParameters.applicationProtocols = names.toTypedArray()
       }
 
-      // Enable SNI.
-      // TODO: Maybe not need post JDK 1.7?
-//      if (hostname != null) {
-//        sslParameters.serverNames = listOf(SNIHostName(hostname))
-//      }
-
       sslSocket.sslParameters = sslParameters
     } else {
       super.configureTlsExtensions(sslSocket, hostname, protocols)
--- a/okhttp/src/main/java/okhttp3/internal/platform/OpenJSSEPlatform.kt
+++ b/okhttp/src/main/java/okhttp3/internal/platform/OpenJSSEPlatform.kt
@@ -15,7 +15,6 @@
  */
 package okhttp3.internal.platform
 
-import android.annotation.SuppressLint
 import okhttp3.Protocol
 import java.security.KeyStore
 import java.security.Provider
@@ -52,7 +51,6 @@ class OpenJSSEPlatform private constructor() : Platform() {
       throw UnsupportedOperationException(
           "clientBuilder.sslSocketFactory(SSLSocketFactory) not supported with OpenJSSE")
 
-  @SuppressLint("NewApi")
   override fun configureTlsExtensions(
     sslSocket: SSLSocket,
     hostname: String?,
--- a/okhttp/src/main/java/okhttp3/internal/platform/OpenJSSEPlatform.kt
+++ b/okhttp/src/main/java/okhttp3/internal/platform/OpenJSSEPlatform.kt
@@ -1,0 +1,105 @@
+/*
+ * Copyright (C) 2014 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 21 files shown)


yschimke:
This seems cool, but could be too early for us.

yschimke:
n.b. Failures relate to weird mix of jdk8 and TLSv1.3.  Not a bug in this PR per se, just something to sidestep more carefully than we are currently.

```
Caused by: java.lang.IllegalArgumentException: TLSv1.3
	at sun.security.ssl.ProtocolVersion.valueOf(ProtocolVersion.java:187)
	at sun.security.ssl.X509TrustManagerImpl.checkTrusted(X509TrustManagerImpl.java:206)
	at sun.security.ssl.X509TrustManagerImpl.checkServerTrusted(X509TrustManagerImpl.java:124)
	at org.openjsse.sun.security.ssl.CertificateMessage$T13CertificateConsumer.checkServerCerts(CertificateMessage.java:1308)
	at org.openjsse.sun.security.ssl.CertificateMessage$T13CertificateConsumer.onConsumeCertificate(CertificateMessage.java:1199)
	at org.openjsse.sun.security.ssl.CertificateMessage$T13CertificateConsumer.consume(CertificateMessage.java:1146)
```

yschimke:
image

yschimke:
@swankjesse Getting MockWebserver working with openjsse is a bigger effort.  Options

1) We could test twice.  Basic client only tests on JDK8, test client+server with JDK11? 
2) Fix MockWebserver completely to handle openjsse on JDK8.

swankjesse:
What’s difficult about getting MockWebServer working with OpenJSSE?

okhttp/src/main/java/okhttp3/internal/platform/OpenJSSEPlatform.kt
```diff
@@ -0,0 +1,96 @@
...
+ * Platform using OpenJSSE (https://github.com/openjsse/openjsse) if installed as the first
+ * Security Provider.
+ *
+ * Requires org.openjsse:openjsse >= 1.1.0 on the classpath.
+ */
+class OpenJSSEPlatform private constructor() : Platform() {
+  private val provider: Provider = org.openjsse.net.ssl.OpenJSSE()
+
+  override fun newSSLContext(): SSLContext =
+      SSLContext.getInstance("TLSv1.3", provider)
```
swankjesse: is this TLSv1.3+ ? We also wanna support TLSv1.2 right?

okhttp/src/main/java/okhttp3/internal/platform/OpenJSSEPlatform.kt
```diff
@@ -0,0 +1,96 @@
...
+    val trustManagers = factory.trustManagers!!
+    check(trustManagers.size == 1 && trustManagers[0] is X509TrustManager) {
+      "Unexpected default trust managers: ${trustManagers.contentToString()}"
+    }
+    return trustManagers[0] as X509TrustManager
+  }
+
+  public override fun trustManager(sslSocketFactory: SSLSocketFactory): X509TrustManager? =
+      throw UnsupportedOperationException(
+          "clientBuilder.sslSocketFactory(SSLSocketFactory) not supported with OpenJSSE")
```
swankjesse: nice

okhttp/src/main/java/okhttp3/internal/platform/OpenJSSEPlatform.kt
```diff
@@ -0,0 +1,96 @@
...
+    if (sslSocket is org.openjsse.javax.net.ssl.SSLSocket) {
+      val sslParameters = sslSocket.sslParameters
+
+      if (sslParameters is org.openjsse.javax.net.ssl.SSLParameters) {
+        // Enable ALPN.
+        val names = alpnProtocolNames(protocols)
+        sslParameters.applicationProtocols = names.toTypedArray()
+      }
+
+      sslSocket.sslParameters = sslParameters
```
swankjesse: what does this do?

okhttp/src/main/java/okhttp3/internal/platform/OpenJSSEPlatform.kt
```diff
@@ -0,0 +1,96 @@
...
+    } else {
+      super.configureTlsExtensions(sslSocket, hostname, protocols)
+    }
+  }
+
+  override fun getSelectedProtocol(sslSocket: SSLSocket): String? =
+      if (sslSocket is org.openjsse.javax.net.ssl.SSLSocket) {
+        when (val protocol = sslSocket.applicationProtocol) {
+          null, "" -> null
```
swankjesse: does it actually return both of these values? ick

okhttp/src/main/java/okhttp3/internal/platform/OpenJSSEPlatform.kt
yschimke: javax.net.ssl.SSLSocket creates a new params object when you use the getter, so to apply you specifically need to call the setter.

okhttp/src/main/java/okhttp3/internal/platform/OpenJSSEPlatform.kt
yschimke: I don't make the rules...

```
     * @return null if it has not yet been determined if application
     *         protocols might be used for this connection, an empty
     *         {@code String} if application protocols values will not
     *         be used, or a non-empty application protocol {@code String}
     *         if a value was successfully negotiated.
```

okhttp/src/main/java/okhttp3/internal/platform/OpenJSSEPlatform.kt
yschimke: For OpenJSSE it at least triggers changes to packet sizes https://github.com/openjsse/openjsse/blob/4b67b70d1ffe28f62f388e9a4cd7768be23b5f6c/src/main/java/org/openjsse/sun/security/ssl/SSLSocketImpl.java#L1048

okhttp/src/main/java/okhttp3/internal/platform/OpenJSSEPlatform.kt
yschimke: It does https://github.com/openjsse/openjsse/blob/82abe9803bbae50428be27189ab85b48d20f96bd/src/main/java/org/openjsse/sun/security/ssl/SSLContextImpl.java#L808

But in JDKs, supported vs defaulted often diverge so on other OpenJDK/Oracle/Android platforms we have had to select TLSv1.2 specifically or we might get held back.

yschimke:
> What’s difficult about getting MockWebServer working with OpenJSSE?

A mixture of JVM singletons (TlsUtils.localhost), provider setup timing, strict code in the JDK that throws exceptions if it sees the 1.3 version being selected, assumptions in MockWebServer e.g. it doesn't use the Platform SSLContext API.  So it's a separate larger task to fix it up.  Work we should do, but doesn't block using this, and we can already start testing this as a client only where it appears to mostly *just* work.

yschimke:
Raised https://github.com/square/okhttp/issues/5380 to fix up MockWebserver.  But not blocking this as we can test thoroughly with client only to flush out bugs before we bake the 4.2 cake.

okhttp/src/main/java/okhttp3/internal/platform/OpenJSSEPlatform.kt
swankjesse: Got it. APIs that mix mutable and accessors are hard to reason about.

swankjesse:
Thanks!

PR #5375, opened 2019-08-17
Ensure overridden content length/type headers are logged

### LINKED ISSUE #5373: HttpLoggingInterceptor does not print correct header if it's overridden
I've got a backend that I'm talking to which does not accept `Content-Type: application/json; charset=UTF-8`. Instead, I need to send `Content-Type: application/json;`.

Using a Network Interceptor I can do all of that but the problem is that `HttpLoggingInterceptor` logs the incorrect header.

Here's a failing test case using 4.1.0:

```kotlin
import okhttp3.Interceptor
import okhttp3.Interceptor.Chain
import okhttp3.MediaType.Companion.toMediaType
import okhttp3.OkHttpClient
import okhttp3.Request
import okhttp3.RequestBody.Companion.toRequestBody
import okhttp3.logging.HttpLoggingInterceptor
import okhttp3.logging.HttpLoggingInterceptor.Level.HEADERS
import okhttp3.logging.HttpLoggingInterceptor.Logger
import okhttp3.mockwebserver.MockResponse
import okhttp3.mockwebserver.MockWebServer
import okio.ByteString.Companion.encodeUtf8
import org.assertj.core.api.Java6Assertions.assertThat
import org.junit.Before
import org.junit.Rule
import org.junit.Test

class FooTest {
  @get:Rule val server = MockWebServer()

  private lateinit var client: OkHttpClient
  private lateinit var logs: MutableList<String>

  @Before fun setUp() {
    logs = mutableListOf()

    client = OkHttpClient.Builder()
        .addNetworkInterceptor(object : Interceptor {
          override fun intercept(chain: Chain) = chain.proceed(chain.request()
            .newBuilder()
            .header("Content-Type", "application/json")
            .build())
        })
        .addNetworkInterceptor(HttpLoggingInterceptor(object : Logger {
          override fun log(message: String) {
            logs.add(message)
          }
        }).apply { level = HEADERS })
        .build()
  }

  @Test fun contentTypeIsLogged() {
    server.enqueue(MockResponse())
    client.newCall(Request.Builder()
        .url(server.url("/"))
        .post("{}".encodeUtf8().toRequestBody("application/json; charset=UTF-8".toMediaType()))
        .build()).execute()

    val request = server.takeRequest()
    assertThat(request.headers.values("Content-Type")).containsExactly("application/json") // This passes correctly.

    // The HttpLoggingInterceptor should only see the overridden header. Instead it sees the Header from the Post request.
    assertThat(logs).contains("Content-Type: application/json;")
    assertThat(logs).doesNotContain("Content-Type: application/json; charset=UTF-8")
  }
}
```

JakeWharton:
Closes #5373

### DIFF
--- a/okhttp-logging-interceptor/src/test/java/okhttp3/logging/HttpLoggingInterceptorTest.java
+++ b/okhttp-logging-interceptor/src/test/java/okhttp3/logging/HttpLoggingInterceptorTest.java
@@ -23,6 +23,7 @@ import java.util.regex.Pattern;
 import javax.annotation.Nullable;
 import javax.net.ssl.HostnameVerifier;
 import okhttp3.HttpUrl;
+import okhttp3.Interceptor;
 import okhttp3.MediaType;
 import okhttp3.OkHttpClient;
 import okhttp3.testing.PlatformRule;
@@ -69,6 +70,8 @@ public final class HttpLoggingInterceptorTest {
   private final HttpLoggingInterceptor applicationInterceptor =
       new HttpLoggingInterceptor(applicationLogs);
 
+  private Interceptor extraNetworkInterceptor = null;
+
   private void setLevel(Level level) {
     networkInterceptor.setLevel(level);
     applicationInterceptor.setLevel(level);
@@ -76,6 +79,9 @@ public final class HttpLoggingInterceptorTest {
 
   @Before public void setUp() {
     client = new OkHttpClient.Builder()
+        .addNetworkInterceptor(chain -> extraNetworkInterceptor != null
+            ? extraNetworkInterceptor.intercept(chain)
+            : chain.proceed(chain.request()))
         .addNetworkInterceptor(networkInterceptor)
         .addInterceptor(applicationInterceptor)
         .sslSocketFactory(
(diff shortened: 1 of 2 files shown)


okhttp-logging-interceptor/src/test/java/okhttp3/logging/HttpLoggingInterceptorTest.java
```diff
@@ -69,13 +70,18 @@
   private final HttpLoggingInterceptor applicationInterceptor =
       new HttpLoggingInterceptor(applicationLogs);
 
+  private Interceptor extraNetworkInterceptor = null;
```
swankjesse: maybe a null object pattern would be nicer here?

```
private Interceptor extraNetworkInterceptor = object : NetworkInterceptor {
  fun intercept(chain: Chain) = chain.proceed(chain.request())
}
```

okhttp-logging-interceptor/src/test/java/okhttp3/logging/HttpLoggingInterceptorTest.java
```diff
@@ -325,6 +331,45 @@ private void setLevel(Level level) {
         .assertNoMoreLogs();
   }
 
+  @Test public void headersPostWithHeaderOverrides() throws IOException {
+    setLevel(Level.HEADERS);
+
+    extraNetworkInterceptor = chain -> chain.proceed(chain.request()
+        .newBuilder()
+        .header("Content-Length", "2")
```
swankjesse: 🙀

PR #5389, opened 2019-08-22
Support lambda construction for Logger

kamerok:
Support lambda construction for Logger the same way as [Interceptor](https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/Interceptor.kt) does, to write
```
val interceptor = HttpLoggingInterceptor(Logger {
    message -> Timber.tag("OkHttp").d(message)
})
```
instead of:
```
val interceptor = HttpLoggingInterceptor(object : Logger {
  override fun log(message: String) {
    Timber.tag("OkHttp").d(message)
  }
})
```


swankjesse:
I want to limit how many of these we add because a future update to Kotlins type inferencer will make SAM conversion automatic.

kamerok:
OK, I see now that [target version is 1.4](https://youtrack.jetbrains.com/issue/KT-11129)

PR #5394, opened 2019-08-23
Expose SSLContext in public API

yschimke:
Expose the SSLContext since it is created internally by Platform with non-trivial rules e.g. Conscrypt installed and available.

Open Question: 

1. Should it be internal, or exposed as public API?
2. Should we focus on a common API instead for performance aware users e.g. session resumption rates, connection pool stats etc.

n.b. I'd like to allow okhttp users to understand whether features like session resumption is working and tuned correctly with their servers.

### DIFF
--- a/okhttp-testing-support/src/main/java/okhttp3/OkHttpClientTestRule.kt
+++ b/okhttp-testing-support/src/main/java/okhttp3/OkHttpClientTestRule.kt
@@ -22,6 +22,7 @@ import org.junit.runner.Description
 import org.junit.runners.model.Statement
 import java.net.InetAddress
 import java.util.concurrent.ConcurrentLinkedDeque
+import javax.net.ssl.SSLSessionContext
 
 /** Apply this rule to tests that need an OkHttpClient instance. */
 class OkHttpClientTestRule : TestRule {
@@ -81,6 +82,11 @@ class OkHttpClientTestRule : TestRule {
 
       private fun releaseClient() {
         prototype?.let {
+          it.sslSessionContext?.let { sslSessionContext: SSLSessionContext ->
+            for (session in sslSessionContext.ids) {
+              sslSessionContext.getSession(session).invalidate()
+            }
+          }
           prototypes.push(it)
           prototype = null
         }
--- a/okhttp-testing-support/src/main/java/okhttp3/OkHttpClientTestRule.kt
+++ b/okhttp-testing-support/src/main/java/okhttp3/OkHttpClientTestRule.kt
@@ -82,14 +82,20 @@ class OkHttpClientTestRule : TestRule {
 
       private fun releaseClient() {
         prototype?.let {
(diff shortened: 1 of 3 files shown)


swankjesse:
Can users just keep a reference to their own and provide us with an SSLSocketFactory?

yschimke:
Yes. But I guess they should use Platform to create the SSLContext. So it ends up being code that you can get wrong easily. E.g. not using Platform and having differing logic. 

Maybe it should be an internal API that we expose metric and a simple session management API for. 

As I mentioned before with Conscrypt you can save these between restarts even.

yschimke:
Conscrypt uses SSLContext so I'll switch to that. which also exposes the client SSLSessionContext.

swankjesse:
I think the heart-wrenching part of this is that we’ve already deprecated `sslSocketFactory()` once, and to do this The Right Way we’d need to deprecate it again, this time to let the caller provide their own SSL context.

My preference on this one is that if we were starting from scratch, we’d build a fancy abstraction that’d make it easy to provide just what you know: SSLContext, SSLSocketFactory, TrustManager, and we’d figure it out.

But now it feels like that ship has sailed. Power users configure their own socket factories (and therefore have their own SSL contexts). Everyone else doesn’t care, and I don’t want to punish them by deprecating this again and making this API even more difficult to call.

I especially dislike having an SSLContext member that is sometimes null because of the way your OkHttpClient was created.

yschimke:
> I especially dislike having an SSLContext member that is sometimes null because of the way your OkHttpClient was created.

It's only null when you create a plaintext only client.

```
  @get:JvmName("sslContext") val sslContext: SSLContext
    get() = sslContextOrNull ?: throw IllegalStateException("CLEARTEXT-only client")
```

That said, this isn't a hill to die on.

I'm way less interested in configuring this or accessing it for reuse, my personal reason is more around having visibility into what OkHttp is doing in term of TLS events etc.  So I think next option is better visibility into the ConnectionPool stats and status.  Are clients tuned correctly etc.

swankjesse:
Isn't it also null if you configure an SSLSocketFactory?

yschimke:
Yep, I can see why it's so ugly.  I think this might be related to why the Platform singleton unsettles me, because a client can provide a SSLSocketFactory that doesn't match the installed Platform.

swankjesse:
Yeah, you’re right. When we went from (Android/Java 8+/Java 9+) to Android/Java 8+/Java 9+/Conscrypt/OpenJSSE) the Platform class was over-extended.

I assume most developers don’t use multiple TLS stacks concurrently. But they might use 2 if they asynchronously init Conscrypt or OpenJSSE.

Maybe finish your project to split Platform into actual immutable platform stuff (logging, strict mode) and TLS stuff?

yschimke:
Thank god it's internal.  We need to get that 100% right if parts ever become public.

Issue #5413, opened 2019-08-08
Cached header combination dropping headers when repeated name present


thierryd:
I have upgraded my project to Retrofit 2.6.1 and OkHttp 4.0.1
Everything seem to work, except when the server returns a 304. Retrofit can't seem to parse the response from the cache (it tries to parse an empty response).

Rolling back to Retrofit 2.5.0 and OkHttp 3.14.0 fix the issue.

swankjesse:
@thierryd fantastic investigation. We’ll fix this ASAP and cut a patch release.

PR #5415, opened 2019-09-05
Fix a regression in combining cache headers

swankjesse:
This was introduced with our migration to Kotlin. Iterating over the indices
of the header names doesn't work because header names are not distinct.

Closes: https://github.com/square/okhttp/issues/5413


swankjesse:
Introduced here:   https://github.com/square/okhttp/pull/4997

Issue #3772, opened 2018-01-11
java.lang.ClassNotFoundException: Didn't find class "android.security.NetworkSecurityPolicy"


wb357127021:
after the Version 6.0 did not appear.

swankjesse:
Got a full stack trace? Things are supposed to recover it this is absent.

yschimke:
re-open if you have a stack trace, thanks.

mueller-ma:
@yschimke I have the same problem. The app crashes on Android 4.2.2 since I added a network security policy to my app.
````
dalvik.system.BaseDexClassLoader.findClass (BaseDexClassLoader.java:53)
java.lang.ClassLoader.loadClass (ClassLoader.java:501)
java.lang.ClassLoader.loadClass (ClassLoader.java:461)
java.lang.Class.classForName (Class.java)
java.lang.Class.forName (Class.java:217)
java.lang.Class.forName (Class.java:172)
okhttp3.internal.platform.AndroidPlatform.isCleartextTrafficPermitted (AndroidPlatform.java:168)
okhttp3.internal.connection.RealConnection.connect (RealConnection.java:143)
okhttp3.internal.connection.StreamAllocation.findConnection (StreamAllocation.java:256)
okhttp3.internal.connection.StreamAllocation.findHealthyConnection (StreamAllocation.java:134)
okhttp3.internal.connection.StreamAllocation.newStream (StreamAllocation.java:113)
okhttp3.internal.connection.ConnectInterceptor.intercept (ConnectInterceptor.java:42)
okhttp3.internal.http.RealInterceptorChain.proceed (RealInterceptorChain.java:147)
okhttp3.internal.http.RealInterceptorChain.proceed (RealInterceptorChain.java:121)
okhttp3.internal.cache.CacheInterceptor.intercept (CacheInterceptor.java:93)
okhttp3.internal.http.RealInterceptorChain.proceed (RealInterceptorChain.java:147)
okhttp3.internal.http.RealInterceptorChain.proceed (RealInterceptorChain.java:121)
okhttp3.internal.http.BridgeInterceptor.intercept (BridgeInterceptor.java:93)
okhttp3.internal.http.RealInterceptorChain.proceed (RealInterceptorChain.java:147)
okhttp3.internal.http.RetryAndFollowUpInterceptor.intercept (RetryAndFollowUpInterceptor.java:125)
okhttp3.internal.http.RealInterceptorChain.proceed (RealInterceptorChain.java:147)
[...]

PR #5416, opened 2019-09-05
OkHttp 3.12.4 is released

swankjesse:


swankjesse:
For this: https://github.com/square/okhttp/issues/3772

swankjesse:
Self-merging documentation change.

Issue #5419, opened 2019-09-05
HttpLoggingInterceptor does not respect RequestBody.isOneShot()


DavidJRobertson:
HttpLoggingInterceptor should check the value of the `request.body.isOneShot()`, and if so, should not attempt to read/log the body contents and instead print a message stating that it was skipped.

PR #5420, opened 2019-09-05
HttpLoggingInterceptor: respect RequestBody.isOneShot()

DavidJRobertson:
Currently, if we have a one-shot request body and try to use the logging interceptor with `Level.BODY`, the `RequestBody.writeTo` method will be called twice. This is broken behaviour.

See #5419

### DIFF
--- a/okhttp-logging-interceptor/src/test/java/okhttp3/logging/HttpLoggingInterceptorTest.java
+++ b/okhttp-logging-interceptor/src/test/java/okhttp3/logging/HttpLoggingInterceptorTest.java
@@ -875,6 +875,53 @@ public final class HttpLoggingInterceptorTest {
         .assertNoMoreLogs();
   }
 
+
+
+  @Test public void oneShotRequestsAreNotLogged() throws Exception {
+    url = server.url("/");
+
+    setLevel(Level.BODY);
+
+    server.enqueue(new MockResponse()
+                           .setBody("Hello response!"));
+
+    RequestBody asyncRequestBody = new RequestBody() {
+      @Override public @Nullable MediaType contentType() {
+        return null;
+      }
+
+      int counter = 0;
+      @Override public void writeTo(BufferedSink sink) throws IOException {
+        counter++;
+        assertThat(counter).isLessThanOrEqualTo(1);
+
+        sink.writeUtf8("Hello request!");
+        sink.close();
+      }
+
+      @Override public boolean isOneShot() {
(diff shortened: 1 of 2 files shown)


JakeWharton:
We already buffer the entire request body. We can just `clone()` the `Buffer` and replace the original `RequestBody` (which also would make it no longer one-shot to downstream interceptors).

DavidJRobertson:
You could do, but if the `RequestBody` is trying to keep track of an upload's progress as well, that would be broken by cloning the buffer.

JakeWharton:
It's already broken in that case for non-one-shot bodies though, and I don't think one-shot is synonymous with indicating that a body that requires streaming.

DavidJRobertson:
True, but is it not reasonable to work around that by saying that if you're tracking progress in a RequestBody then it should/must be one-shot. If it wasn't one-shot, you would get garbage data anyway...

swankjesse:
I assume buffering a one-shot body is probably awkward – one shot is most useful for transient streams like live video from the camera or live audio from the microphone.

@DavidJRobertson mind adding a test?

okhttp-logging-interceptor/src/test/java/okhttp3/logging/HttpLoggingInterceptorTest.java
```diff
@@ -875,6 +875,53 @@ public void headersAreRedacted() throws Exception {
         .assertNoMoreLogs();
   }
 
+
```
yschimke: nit: remove 2 additional lines

yschimke:
```suggestion
```

okhttp-logging-interceptor/src/test/java/okhttp3/logging/HttpLoggingInterceptorTest.java
```diff
@@ -875,6 +875,53 @@ public void headersAreRedacted() throws Exception {
...
+
```
yschimke:
```suggestion
```

okhttp-logging-interceptor/src/test/java/okhttp3/logging/HttpLoggingInterceptorTest.java
yschimke: Not sure if this works :)

Issue #5412, opened 2019-09-04
HttpLoggingInterceptor: faulty binary request/response detection 


DavidJRobertson:
HttpLoggingInterceptor sometimes omits request/response bodies from logs if it thinks that it contains binary data. However, the `Buffer.isProbablyUtf8()` method doesn't reliably detect binary data, especially when the file format begins with a textual header.

This results in the logs being completely overwhelmed with useless binary data. For instance, this occurs frequently for me when uploading (10MB chunks of) mp4 video:
<img width="1969" alt="Screenshot 2019-09-04 at 15 01 28" src="https://user-images.githubusercontent.com/357772/64261780-f6ecd800-cf24-11e9-9ad8-05382e6e7f54.png">

Perhaps it would be better to examine the Content-Type and/or Content-Length headers first to see if we have a binary type or very large body, before falling back to `isProbablyUtf8()` if these headers are not present.

Alternatively, it might be worth making `isProbablyUtf8()` examine more than just the first 64 bytes / 16 UTF-8 code points before making a decision.

yschimke:
Won't fix, this should probably be a separate project.  It's interesting enough on it's own, but a distraction for OkHttp.

PR #5421, opened 2019-09-05
HttpLoggingInterceptor: improve detection of binary request/response bodies

DavidJRobertson:
See #5412 

Now checking the content type to see if it looks like a binary type when deciding whether or not to omit the body content from the logs.

### DIFF
--- a/okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.kt
+++ b/okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.kt
@@ -19,6 +19,7 @@ import okhttp3.Headers
 import okhttp3.Interceptor
 import okhttp3.OkHttpClient
 import okhttp3.Response
+import okhttp3.MediaType
 import okhttp3.internal.http.promisesBody
 import okhttp3.internal.platform.Platform
 import okhttp3.internal.platform.Platform.Companion.INFO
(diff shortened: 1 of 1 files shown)


okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.kt
```diff
@@ -285,4 +288,23 @@ class HttpLoggingInterceptor @JvmOverloads constructor(
     return !contentEncoding.equals("identity", ignoreCase = true) &&
         !contentEncoding.equals("gzip", ignoreCase = true)
   }
+    private fun mediaTypeIsProbablyBinary(mediaType: MediaType?): Boolean {
+        val type = mediaType?.type
+        val subType = mediaType?.subtype
+
+        val binaryTypes = listOf("audio", "binary", "font", "image", "video")
+        val binaryApplicationSubtypes = listOf("binary", "excel", "java-archive", "java-vm", "lzx",
```
swankjesse: I very much do not want to maintain this list of binary media types. Wanna make this the user’s problem by using the same strategy as `headersToRedact`?

I think this also wants us to finally finish this: https://github.com/square/okhttp/issues/782

yschimke:
@DavidJRobertson The other option here is for someone interested in this functionality to make a better HttpLoggingInterceptor as a library.  There are an infinite number of small features people want and having an active user/maintainer of this could help to point to a small focused library.  @JakeWharton has a ton of these small libraries and the model works well there. Maybe 2 small libraries, one with common media types, and the logger.

yschimke:
The reason I specifically suggest the library approach here, is this sort of feature only really works if it's implemented as sensible defaults.  By the point we ask users to define setup calls to headersToRedact, they are likely to consider other options like implementing their own or using a library.  At the same time, I agree that we don't want the core library to be making arbitrary decisions like this.  Maybe the right thing to do here is actually to show hex output in a nice format?

swankjesse:
Agreed. Closing for now.

Issue #5422, opened 2019-09-06
Acknowledge inbound SETTINGS before using new flow control window budget


swankjesse:
From [nghttp2](https://github.com/nghttp2/nghttp2/blob/master/lib/nghttp2_session.c), the HTTP/2 backend inside Envoy:

```
/*
 * Now we have SETTINGS synchronization, flow control error can be
 * detected strictly. If DATA frame is received with length > 0 and
 * current received window size + delta length is strictly larger than
 * local window size, it is subject to FLOW_CONTROL_ERROR, so return
 * -1. Note that local_window_size is calculated after SETTINGS ACK is
 * received from peer, so peer must honor this limit. If the resulting
 * recv_window_size is strictly larger than NGHTTP2_MAX_WINDOW_SIZE,
 * return -1 too.
 */
```

OkHttp transmits its outbound `SETTINGS ACK` frame asynchronously. It’s possible that it’ll be transmitted after outbound `DATA` frames that use the newer, larger window. And that’ll cause nghttp2 to fail our streams with `FLOW_CONTROL_ERROR`.

To interop best with nghttp2 we need to transmit the `SETTINGS ACK` frame before adjusting our flow control window.

PR #5423, opened 2019-09-06
Change the way flow control windows are tracked.

swankjesse:
Previously we tracked how many bytes we had left in a single value. With
this updatee we track the current value and the budget separately. I am
not worried about overflowing Long.MAX_VALUE.

I made this change in the process of diagnosing this:
https://github.com/square/okhttp/issues/5422

### DIFF
--- a/okhttp/src/test/java/okhttp3/internal/http2/Http2ConnectionTest.java
+++ b/okhttp/src/test/java/okhttp3/internal/http2/Http2ConnectionTest.java
@@ -37,7 +37,6 @@ import okio.Sink;
 import okio.Source;
 import okio.Utf8;
 import org.junit.After;
-import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.TestRule;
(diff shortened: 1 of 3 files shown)


okhttp/src/test/java/okhttp3/internal/http2/Http2ConnectionTest.java
```diff
@@ -1749,25 +1752,31 @@
...
+    assertThat(stream1.getWriteBytesMaximum()).isEqualTo(DEFAULT_INITIAL_WINDOW_SIZE);
 
     // receiving a window update on the connection will unblock new streams.
     connection.getReaderRunnable().windowUpdate(0, 3);
 
-    assertThat(connection.getBytesLeftInWriteWindow()).isEqualTo(3);
-    assertThat(connection.getStream(3).getBytesLeftInWriteWindow()).isEqualTo(0);
+    assertThat(connection.getWriteBytesMaximum()).isEqualTo(DEFAULT_INITIAL_WINDOW_SIZE + 3);
+    assertThat(connection.getWriteBytesTotal()).isEqualTo(DEFAULT_INITIAL_WINDOW_SIZE);
+    assertThat(stream1.getWriteBytesTotal()).isEqualTo(DEFAULT_INITIAL_WINDOW_SIZE);
```
yschimke: ubernit: switching the order of maximum and total, but not an issue.

swankjesse: Ooh yes, fixed.

PR #5427, opened 2019-09-07
Make it easier to decode PEM files

swankjesse:
I'm starting with the thing I want first; reusing a HeldCertificate
across VM invocations.

Later we might want to create entire HandshakeCertificates objects
by stringing together many certificates from a single file.

### DIFF
--- a/okhttp-tls/src/main/java/okhttp3/tls/HeldCertificate.kt
+++ b/okhttp-tls/src/main/java/okhttp3/tls/HeldCertificate.kt
@@ -16,7 +16,9 @@
 package okhttp3.tls
 
 import okhttp3.internal.canParseAsIpAddress
+import okio.Buffer
 import okio.ByteString
+import okio.ByteString.Companion.decodeBase64
 import okio.ByteString.Companion.toByteString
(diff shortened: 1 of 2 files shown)


okhttp-tls/src/main/java/okhttp3/tls/HeldCertificate.kt
```diff
@@ -27,17 +29,24 @@ import org.bouncycastle.asn1.x509.X509Extensions
...
+import java.security.cert.CertificateFactory
 import java.security.cert.X509Certificate
+import java.security.interfaces.ECPublicKey
 import java.security.interfaces.RSAPrivateKey
+import java.security.interfaces.RSAPublicKey
+import java.security.spec.PKCS8EncodedKeySpec
 import java.util.Date
 import java.util.UUID
 import java.util.concurrent.TimeUnit
+import java.util.regex.Pattern
```
yschimke: Why not use kotlin APIs here?

swankjesse: Just habit. Migrated to Kotlin APIs and they fit this problem even better.

PR #5431, opened 2019-09-08
Acknowledge and apply inbound settings atomically

swankjesse:
Closes: https://github.com/square/okhttp/issues/5422

Unfortunately testing this is awkward because it's racy. I did
run a stress test that used to reproduce the problem, and now it
doesn't, so I am satisfied.

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/http2/Http2Connection.kt
+++ b/okhttp/src/main/java/okhttp3/internal/http2/Http2Connection.kt
@@ -138,7 +138,6 @@ class Http2Connection internal constructor(builder: Builder) : Closeable {
   var writeBytesMaximum: Long = peerSettings.initialWindowSize.toLong()
     private set
 
-  internal var receivedInitialPeerSettings = false
   internal val socket: Socket = builder.socket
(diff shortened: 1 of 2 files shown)


okhttp/src/main/java/okhttp3/internal/http2/Http2Connection.kt
```diff
@@ -660,43 +659,53 @@ class Http2Connection internal constructor(builder: Builder) : Closeable {
...
+     * We need to apply the settings and ack them atomically. This is because some HTTP/2
+     * implementations (nghttp2) forbid peers from taking advantage of settings before they have
+     * acknowledged! In particular, we shouldn't send frames that assume a new `initialWindowSize`
+     * until we send the frame that acknowledges this new size.
+     *
+     * Since we can't ACK settings on the current reader thread (the reader thread can't write) we
+     * execute all peer settings logic on the writer thread. This relies on the fact that the
+     * writer executor won't reorder tasks; otherwise settings could be applied in the opposite
+     * order than received.
+     */
```
swankjesse: this function diffs better with GitHub’s _Hide whitespace changes_ setting.

Issue #5432, opened 2019-09-09
Certificate pinning: ability to apply pins to all subdomains


DavidJRobertson:
The wildcard support in CertificatePinner is very limited. In some cases, we may not know all of the subdomains which may be used in advance, but yet still want to apply pinning across them all.

This could be implemented by changing the behaviour of the wildcards, but this would potentially impact existing users. I think it's a better idea to add a boolean flag or separate builder method which causes all subdomains to be included.

PR #5433, opened 2019-09-09
Allow pinning all subdomains (including sub-subdomains) in CertificatePinner.Builder

DavidJRobertson:
See #5432

### DIFF
--- a/okhttp/src/main/java/okhttp3/CertificatePinner.kt
+++ b/okhttp/src/main/java/okhttp3/CertificatePinner.kt
@@ -240,36 +240,28 @@ class CertificatePinner internal constructor(
 
   internal data class Pin(
     /** A hostname like `example.com` or a pattern like `*.example.com`. */
-    val pattern: String,
+val pattern: String,
     /** The canonical hostname, i.e. `EXAMPLE.com` becomes `example.com`. */
-    private val canonicalHostname: String,
-    /** If true, all subdomains of the domain are included (including sub-subdomains etc). If true, wildcard matching is not applied. */
-    val includeSubdomains: Boolean,
+private val canonicalPattern: String,
     /** Either `sha1/` or `sha256/`. */
-    val hashAlgorithm: String,
+val hashAlgorithm: String,
     /** The hash of the pinned certificate using [hashAlgorithm]. */
-    val hash: ByteString
+val hash: ByteString
   ) {
     fun matches(hostname: String): Boolean {
-      if (includeSubdomains) {
-        val patternSegments = pattern.split('.').reversed()
-        val hostnameSegments = hostname.split('.').reversed()
-        if (patternSegments.size > hostnameSegments.size) {
+      val patternSegments = canonicalPattern.split('.').reversed()
+      val hostnameSegments = hostname.split('.').reversed()
+      if (patternSegments.size != hostnameSegments.size && patternSegments.last() != "**") {
+        return false
+      }
+      for (i in patternSegments.indices) {
+        val patternSegment = patternSegments[i]
+        val hostnameSegment = hostnameSegments.getOrNull(i) ?: return false
+        if (patternSegment != hostnameSegment && patternSegment != "*" && patternSegment != "**") {
           return false
         }
-        for (i in patternSegments.indices) {
-          if (patternSegments[i] != hostnameSegments[i]) {
-            return false
-          }
-        }
-        return true
-      } else if (pattern.startsWith(WILDCARD)) {
-        val firstDot = hostname.indexOf('.')
-        return hostname.length - firstDot - 1 == canonicalHostname.length &&
-                  hostname.startsWith(canonicalHostname, startIndex = firstDot + 1)
-      } else {
-        return hostname == canonicalHostname
       }
+      return true
     }
 
     override fun toString(): String = hashAlgorithm + hash.base64()
@@ -283,34 +275,19 @@ class CertificatePinner internal constructor(
      * Pins certificates for `pattern`.
      *
      * @param pattern lower-case host name or wildcard pattern such as `*.example.com`.
-     * @param includeSubdomains Set this to true to include all subdomains of the given domain. If
-     * true, wildcard matching is turned off.
-     * @param pins SHA-256 or SHA-1 hashes. Each pin is a hash of a certificate's Subject Public Key
-     * Info, base64-encoded and prefixed with either `sha256/` or `sha1/`.
-     */
-    fun add(pattern: String, includeSubdomains: Boolean, vararg pins: String) = apply {
-      for (pin in pins) {
-        this.pins.add(newPin(pattern, includeSubdomains, pin))
-      }
-    }
-
-    /**
-     * Pins certificates for `pattern`.
-     *
-     * @param pattern lower-case host name or wildcard pattern such as `*.example.com`.
      * @param pins SHA-256 or SHA-1 hashes. Each pin is a hash of a certificate's Subject Public Key
      * Info, base64-encoded and prefixed with either `sha256/` or `sha1/`.
      */
     fun add(pattern: String, vararg pins: String) = apply {
-      add(pattern, false, *pins)
+      for (pin in pins) {
+        this.pins.add(newPin(pattern, pin))
+      }
     }
 
     fun build(): CertificatePinner = CertificatePinner(pins.toSet(), null)
   }
 
   companion object {
-    internal const val WILDCARD = "*."
-
     @JvmField
     val DEFAULT = Builder().build()
 
@@ -333,26 +310,16 @@ class CertificatePinner internal constructor(
         publicKey.encoded.toByteString().sha256()
 
     internal fun newPin(pattern: String, pin: String): Pin {
-      return newPin(pattern, false, pin)
-    }
-    internal fun newPin(pattern: String, includeSubdomains: Boolean, pin: String): Pin {
-      val canonicalHostname = when {
-        pattern.startsWith(WILDCARD) -> {
-          "http://${pattern.substring(WILDCARD.length)}".toHttpUrl().host
-        }
-        else -> {
-          "http://$pattern".toHttpUrl().host
-        }
-      }
+      val canonicalPattern = "https://$pattern".toHttpUrl().host
 
       return when {
         pin.startsWith("sha1/") -> {
           val hash = pin.substring("sha1/".length).decodeBase64()!!
-          Pin(pattern, canonicalHostname, includeSubdomains, "sha1/", hash)
+          Pin(pattern, canonicalPattern, "sha1/", hash)
         }
         pin.startsWith("sha256/") -> {
           val hash = pin.substring("sha256/".length).decodeBase64()!!
-          Pin(pattern, canonicalHostname, includeSubdomains, "sha256/", hash)
+          Pin(pattern, canonicalPattern, "sha256/", hash)
         }
         else -> throw IllegalArgumentException("pins must start with 'sha256/' or 'sha1/': $pin")
       }
--- a/okhttp/src/main/java/okhttp3/CertificatePinner.kt
+++ b/okhttp/src/main/java/okhttp3/CertificatePinner.kt
@@ -260,8 +260,8 @@ val hash: ByteString
       for (i in patternSegments.indices) {
         val patternSegment = patternSegments[i]
         if (patternSegment != "**") {
-          val hostnameSegmentN = hostnameSegments.getOrNull(i)
-          if (patternSegment != hostnameSegmentN && (patternSegment != "*" || hostnameSegmentN == null)) {
+          val hostnameSegment = hostnameSegments.getOrNull(i)
+          if (patternSegment != hostnameSegment && (patternSegment != "*" || hostnameSegment == null)) {
             return false
           }
         }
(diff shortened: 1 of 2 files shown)


yschimke:
It seems preferable that pinning logic matches details from the certs. I.e. only a single level of sub domains. And wildcard seems already supported.

DavidJRobertson:
Android's network security config has a flag to allow using all subdomains including sub-subdomains (https://developer.android.com/training/articles/security-config#domain).

I understand that wildcard certs only match one level (like the existing pinning implementation) and maybe it is generally advisable to make the pins exactly match the certificates, but I don't think this should be forced upon developers.

In my use case, I am communicating with an API developed by another team, with multiple separate server instances. URLs for most actions come from server responses rather than being hard-coded on the client side. We have an agreement on pinning certain CA certificates, but still, the exact subdomains in use are an implementation detail of the API servers which may be subject to change without warning. For this reason, we would like to pin all subdomains of a given domain.

swankjesse:
I think this is pretty confusing, especially since it seems to compete with the wildcard rather than complement it.

How do you feel about say, `**.squareup.com` or `*.*.squareup.com` ?

DavidJRobertson:
That's a fair point. Double asterisk could work, for sure.

 It might also be worth supporting multiple asterisks like in your second example, though I presume this wouldn't be intended to match `a.b.c.squareup.com`, so it's probably not what I want to use in my app.

swankjesse:
Lets go with `**.squareup.com` for any number of subdomains.

swankjesse:
Please follow our contributor’s guideline, including completing our individual CLA.
https://square.github.io/okhttp/contributing/

okhttp/src/main/java/okhttp3/CertificatePinner.kt
```diff
@@ -240,21 +240,28 @@ class CertificatePinner internal constructor(
...
     /** The hash of the pinned certificate using [hashAlgorithm]. */
-    val hash: ByteString
+val hash: ByteString
   ) {
     fun matches(hostname: String): Boolean {
-      if (pattern.startsWith(WILDCARD)) {
-        val firstDot = hostname.indexOf('.')
-        return hostname.length - firstDot - 1 == canonicalHostname.length &&
-            hostname.startsWith(canonicalHostname, startIndex = firstDot + 1)
+      val patternSegments = canonicalPattern.split('.').reversed()
```
yschimke: I think avoiding this work when a simple hostname is used is still desirable.  Optimise for the == case.

Also Pin gives you a natural place to cache some of this work rather than computing each time.  Ultimately if a wildcard is in place, it may we worth checking endsWith before splitting anything.

okhttp/src/main/java/okhttp3/CertificatePinner.kt
```diff
@@ -305,23 +310,16 @@ class CertificatePinner internal constructor(
...
     internal fun newPin(pattern: String, pin: String): Pin {
-      val canonicalHostname = when {
-        pattern.startsWith(WILDCARD) -> {
-          "http://${pattern.substring(WILDCARD.length)}".toHttpUrl().host
-        }
-        else -> {
-          "http://$pattern".toHttpUrl().host
-        }
-      }
+      val canonicalPattern = "https://$pattern".toHttpUrl().host
```
yschimke: This seems questionable since * isn't a valid host character. Why simplify here?

okhttp/src/test/java/okhttp3/CertificatePinnerTest.java
```diff
@@ -189,4 +189,53 @@
...
+    } catch (SSLPeerUnverifiedException expected) {}
+    try {
+      certificatePinner.check("foo.bar.baz.example.co.uk", certB1.certificate());
+      fail();
+    } catch (SSLPeerUnverifiedException expected) {}
+
+    // Should not be pinned:
+    certificatePinner.check("uk", certB1.certificate());
+    certificatePinner.check("co.uk", certB1.certificate());
+    certificatePinner.check("example.co.uk", certB1.certificate());
```
yschimke: Is it worth discussing this case for **?

okhttp/src/test/java/okhttp3/CertificatePinnerTest.java
```diff
@@ -189,4 +189,53 @@
...
+    certificatePinner.check("anotherexample.co.uk", certB1.certificate());
+  }
+
+  @Test public void checkForHostnameWithMultipleWildcardSegments() throws Exception {
+    CertificatePinner certificatePinner = new CertificatePinner.Builder()
+            .add("*.*.example.co.uk", certA1Sha256Pin)
```
yschimke: Worth documenting this as it looks like a very generic feature.  Should *.example.* also work? e.g. companies that register with multiple countries?

okhttp/src/main/java/okhttp3/CertificatePinner.kt
DavidJRobertson: Yes, but it does seem to work fine. This is required for punycode expansion, otherwise, existing tests would fail.

DavidJRobertson:
> Please follow our contributor’s guideline, including completing our individual CLA.
> https://square.github.io/okhttp/contributing/

Done

okhttp/src/main/java/okhttp3/CertificatePinner.kt
yschimke: Yes, but maybe '(?:\\*{1,2}\\.)+' should be stripped before converting in case we make HttpUrl stricter here?

okhttp/src/test/java/okhttp3/CertificatePinnerTest.java
DavidJRobertson: I thought that if `*.example.com` did not include `example.com`, it would be confusing for `**.example.com` to include `example.com`.

yschimke: If people used this with Ant file patterns, which show up elsewhere.  It specifically means zero or more.

https://ant.apache.org/manual/dirtasks.html

> To make things a bit more flexible, we add one extra feature, which makes it possible to match multiple directory levels. This can be used to match a complete directory tree, or a file anywhere in the directory tree. To do this, "**" must be used as the name of a directory. When ** is used as the name of a directory in the pattern, it matches zero or more directories. For example: /test/** matches all files/directories under /test/, such as /test/x.java, or /test/foo/bar/xyz.html, but not /xyz.xml.

okhttp/src/main/java/okhttp3/CertificatePinner.kt
DavidJRobertson: I have added a shortcut path for exact matches.

okhttp/src/test/java/okhttp3/CertificatePinnerTest.java
DavidJRobertson: I am happy to implement it either way. I guess as long as it is documented it will be fine.

@swankjesse any thoughts on this?

okhttp/src/main/java/okhttp3/CertificatePinner.kt
DavidJRobertson: Now using String.toCanonicalHost() directly

okhttp/src/test/java/okhttp3/CertificatePinnerTest.java
DavidJRobertson: I don't see any good reason not to support it. I have added a test for wildcard segments that are not at the bottom level.

okhttp/src/test/java/okhttp3/CertificatePinnerTest.java
swankjesse: I think it’s easy enough if we say `*.example.com` matches exactly one level, and `**.example.com` matches any number.

okhttp/src/test/java/okhttp3/CertificatePinnerTest.java
swankjesse: I do not want `.*` at the end. There’s lots of weird new failure modes this introduces.

okhttp/src/test/java/okhttp3/CertificatePinnerTest.java
DavidJRobertson: I'm not sure whether you mean I should implement the change (to have `**.example.com` match `example.com`) or not.

swankjesse: ### Single star: match exactly one.

`*.example.com`
 - Matches `foo.example.com`
 - Does not match `foo.bar.example.com`, or `example.com`.

### Two stars: match zero or more.

`**.example.com`
Matches `example.com`, `foo.example.com`, and `foo.bar.example.com`.

DavidJRobertson: Thanks for the clarification.

okhttp/src/test/java/okhttp3/CertificatePinnerTest.java
DavidJRobertson: I have made attempts to use a wildcard TLD result in an IllegalArgumentException

okhttp/src/test/java/okhttp3/CertificatePinnerTest.java
DavidJRobertson: I have implemented this change.

okhttp/src/main/java/okhttp3/CertificatePinner.kt
```diff
@@ -90,13 +90,19 @@ import javax.net.ssl.SSLPeerUnverifiedException
...
  *
- *  1. Asterisk `*` is only permitted in the left-most domain name label and must be the only
- *     character in that label (i.e., must match the whole left-most label). For example,
- *     `*.example.com` is permitted, while `*a.example.com`, `a*.example.com`, `a*b.example.com`,
- *     `a.*.example.com` are not permitted.
- *  2. Asterisk `*` cannot match across domain name labels. For example, `*.example.com` matches
+ *  1. Where used, asterisk `*` (or double-asterisk `**`) must be the only text in the label. For
+ *     example, `*.example.com` is permitted, while `*a.example.com`, `a*.example.com`, and
+ *     `a*b.example.com` are not permitted.
+ *  2. Asterisk `*` can appear multiple times but must not appear in the top-level label. For
```
swankjesse: Is anyone asking for this? I’d really just prefer three modes

`example.com`
`*.example.com`
`**.example.com`

The multiple asterixes seems like an unnecessary departure from the SubjectAltName feature that we’re trying to emulate here.

okhttp/src/main/java/okhttp3/CertificatePinner.kt
```diff
@@ -240,21 +246,33 @@ class CertificatePinner internal constructor(
 
   internal data class Pin(
     /** A hostname like `example.com` or a pattern like `*.example.com`. */
-    val pattern: String,
+val pattern: String,
```
swankjesse: Kotlin style here: https://github.com/square/java-code-styles

okhttp/src/main/java/okhttp3/CertificatePinner.kt
DavidJRobertson: Personally I don't plan on using multiple asterisks, but I can conceive of uses for it. If you feel like it wouldn't be beneficial then I'm happy to get rid of it.

okhttp/src/main/java/okhttp3/CertificatePinner.kt
```diff
@@ -305,23 +305,18 @@ class CertificatePinner internal constructor(
...
     internal fun newPin(pattern: String, pin: String): Pin {
-      val canonicalHostname = when {
-        pattern.startsWith(WILDCARD) -> {
-          "http://${pattern.substring(WILDCARD.length)}".toHttpUrl().host
-        }
-        else -> {
-          "http://$pattern".toHttpUrl().host
-        }
-      }
+      require(!pattern.endsWith("*")) { "Invalid pattern: TLD cannot be a wildcard" }
```
yschimke: This line bugs me, because lots of other patterns are invalid.  e.g. * in the middle.

We should express this something closer to, "Patterns must start with wildcard when present".

okhttp/src/main/java/okhttp3/CertificatePinner.kt
```diff
@@ -239,22 +241,22 @@ class CertificatePinner internal constructor(
...
     /** The hash of the pinned certificate using [hashAlgorithm]. */
     val hash: ByteString
   ) {
+
     fun matches(hostname: String): Boolean {
-      if (pattern.startsWith(WILDCARD)) {
-        val firstDot = hostname.indexOf('.')
-        return hostname.length - firstDot - 1 == canonicalHostname.length &&
-            hostname.startsWith(canonicalHostname, startIndex = firstDot + 1)
+      val ending by lazy { pattern.substringAfter(".") }
```
yschimke: Having lazy inside the method seems like non idiomatic kotlin to me. I'd suggest another round of cleanup of the logic. I think it's right, and I'm not against landing, but I suspect there is a nicer succinct version of this hidden here.

One suggestion, TIOLI: maybe pulling out this wildcard matching into a internal sealed class that you test separately from Pin?

swankjesse:
Merged manually. I also changed the matching code to avoid allocations and added more tests.
https://github.com/square/okhttp/commit/4b0885860b9f1cf35f0a081b0916a7b835d224d6

Issue #5440, opened 2019-09-10
CipherSuite TLS_NULL_WITH_NULL_NULL


vittawang:
I met a problem.Here is the error message; This mistack happened 500 times.I have no idea to handle it.Please help me,Thank you!
```java
# OkHttp Dispatcher(3730)  java.lang.IllegalArgumentException  
No enum constant okhttp3.CipherSuite.TLS_NULL_WITH_NULL_NULL

1 java.lang.Enum.valueOf(Enum.java:258)
2 okhttp3.CipherSuite.okhttp3.CipherSuite valueOf(java.lang.String)(SourceFile:32)
3 okhttp3.CipherSuite.okhttp3.CipherSuite forJavaName(java.lang.String)(SourceFile:381)
4 okhttp3.Handshake.okhttp3.Handshake get(javax.net.ssl.SSLSession)(SourceFile:53)
5 okhttp3.internal.io.RealConnection.void connectTls(int,int,okhttp3.internal.ConnectionSpecSelector)(SourceFile:242)
6 okhttp3.internal.io.RealConnection.void establishProtocol(int,int,okhttp3.internal.ConnectionSpecSelector)(SourceFile:198)
7 okhttp3.internal.io.RealConnection.void buildConnection(int,int,int,okhttp3.internal.ConnectionSpecSelector)(SourceFile:173)
8 okhttp3.internal.io.RealConnection.void connect(int,int,int,java.util.List,boolean)(SourceFile:113)
9 okhttp3.internal.http.StreamAllocation.okhttp3.internal.io.RealConnection findConnection(int,int,int,boolean)(SourceFile:188)
10 okhttp3.internal.http.StreamAllocation.okhttp3.internal.io.RealConnection findHealthyConnection(int,int,int,boolean,boolean)(SourceFile:124)
11 okhttp3.internal.http.StreamAllocation.okhttp3.internal.http.HttpStream newStream(int,int,int,boolean,boolean)(SourceFile:94)
12 okhttp3.internal.http.HttpEngine.okhttp3.internal.http.HttpStream connect()(SourceFile:301)
13 okhttp3.internal.http.HttpEngine.void sendRequest()(SourceFile:253)
14 okhttp3.RealCall.okhttp3.Response getResponse(okhttp3.Request,boolean)(SourceFile:254)
15 okhttp3.RealCall$ApplicationInterceptorChain.okhttp3.Response proceed(okhttp3.Request)(SourceFile:212)
[...]

PR #5443, opened 2019-09-11
Handle TLS_NULL_WITH_NULL_NULL by throwing an IOException

swankjesse:
Closes: https://github.com/square/okhttp/issues/5440

### DIFF
--- a/okhttp/src/test/java/okhttp3/HandshakeTest.kt
+++ b/okhttp/src/test/java/okhttp3/HandshakeTest.kt
@@ -1,0 +1,126 @@
+/*
+ * Copyright (C) 2019 Square, Inc.
+ *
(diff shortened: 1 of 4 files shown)


okhttp/src/test/java/okhttp3/HandshakeTest.kt
```diff
@@ -0,0 +1,126 @@
...
+      assertThat(expected).hasMessage("cipherSuite == TLS_NULL_WITH_NULL_NULL")
+    }
+  }
+
+  class FakeSSLSession(
+    private val protocol: String,
+    private val cipherSuite: String,
+    private val peerCertificates: Array<Certificate>?,
+    private val localCertificates: Array<Certificate>?
+  ) : DelegatingSSLSession(null) {
```
yschimke: Nice!

PR #5457, opened 2019-09-14
Cancel calls on unexpected exceptions

swankjesse:
Closes: https://github.com/square/okhttp/issues/5151

### DIFF
--- a/okhttp/src/main/java/okhttp3/RealCall.kt
+++ b/okhttp/src/main/java/okhttp3/RealCall.kt
@@ -143,6 +143,14 @@ internal class RealCall private constructor(
           } else {
             responseCallback.onFailure(this@RealCall, e)
           }
+        } catch (t: Throwable) {
+          cancel()
+          if (!signalledCallback) {
+            val canceledException = IOException("canceled due to $t")
+            canceledException.addSuppressed(t)
+            responseCallback.onFailure(this@RealCall, canceledException)
+          }
+          throw t
         } finally {
           client.dispatcher.finished(this)
         }
--- a/okhttp/src/test/java/okhttp3/InterceptorTest.java
+++ b/okhttp/src/test/java/okhttp3/InterceptorTest.java
@@ -521,11 +521,12 @@ public final class InterceptorTest {
   }
 
   /**
-   * When an interceptor throws an unexpected exception, asynchronous callers are left hanging. The
+   * When an interceptor throws an unexpected exception, asynchronous calls are canceled. The
    * exception goes to the uncaught exception handler.
    */
   private void interceptorThrowsRuntimeExceptionAsynchronous(boolean network) throws Exception {
-    addInterceptor(network, chain -> { throw new RuntimeException("boom!"); });
+    RuntimeException boom = new RuntimeException("boom!");
+    addInterceptor(network, chain -> { throw boom; });
 
     ExceptionCatchingExecutor executor = new ExceptionCatchingExecutor();
     client = client.newBuilder()
@@ -535,9 +536,15 @@ public final class InterceptorTest {
     Request request = new Request.Builder()
         .url(server.url("/"))
         .build();
-    client.newCall(request).enqueue(callback);
+    Call call = client.newCall(request);
+    call.enqueue(callback);
+    RecordedResponse recordedResponse = callback.await(server.url("/"));
+    assertThat(recordedResponse.failure)
+        .hasMessage("canceled due to java.lang.RuntimeException: boom!");
+    assertThat(recordedResponse.failure).hasSuppressedException(boom);
+    assertThat(call.isCanceled()).isTrue();
 
-    assertThat(executor.takeException().getMessage()).isEqualTo("boom!");
+    assertThat(executor.takeException()).isEqualTo(boom);
   }
 
   @Test public void applicationInterceptorReturnsNull() throws Exception {


okhttp/src/main/java/okhttp3/RealCall.kt
```diff
@@ -143,6 +143,14 @@ internal class RealCall private constructor(
           } else {
             responseCallback.onFailure(this@RealCall, e)
           }
+        } catch (t: Throwable) {
+          cancel()
+          if (!signalledCallback) {
+            val canceledException = IOException("canceled due to $t")
+            canceledException.addSuppressed(t)
```
JakeWharton: I wonder if these are reported to crash tools

swankjesse: I'm hoping not. We rethrow `t` for the uncaught exception handler.

JakeWharton: It would also be nice if integrations could unwrap this. What about using a specific subtype of IOException so the Rx and Coroutine code in Retrofit can correctly unwrap this and propagate the real, crash-worthy cause?

swankjesse: We delivered that cause to the uncaught exception handler already. Rx and coroutines should treat this as a cancelation, not as an uncaught exception.

The call has an externality, but that should be handled elsewhere.

JakeWharton: The uncaught exception handler on the dispatcher executor is mostly useless. No one is listening to it and certainly not doing anything about exceptions that reach it. Moreover, you can't correlate it to the actual call which triggered the problem so even if you are listening, you have to guess as to the cause.

At least with this change we're no longer leaving calls hanging forever, but it still sweeps and entire class of programming errors under the rug under the guise of a transport problem.

swankjesse: Crashlytics and Bugsnag track uncaught exceptions. If you’d like, we could include a redacted URL in the stacktrace we re-throw.

JakeWharton: It seems pointless to propagate the exception when the cause mechanism was designed for this. I missed the rethrow, and wouldn't have approved with it.

This behavior is needlessly different from synchronous execution. I'm beginning to regret agreeing to use the cancelation mechanism for this.

I'm not sure who wants this behavior. It's weird. I want my Retrofit method to throw or propagate the original exception like it would if execute was used.

swankjesse: Hm. This is as proposed!
https://github.com/square/okhttp/issues/5151#issuecomment-497908703

Wanna submit a new PR?

JakeWharton: My goal is to allow Retrofit to unwrap the original cause in a way that's considered stable behavior and prevents duplicate reporting. Its users shouldn't need to know about the internal OkHttp mechanics of whether synchronous vs. asynchronous execution is used. That's why I suggested the exception subtype which maintains the cancelation and IOException but providing something people who are adapting OkHttp into different execution mechanisms can unwrap. Or the default method solution allows the current implementation in this PR to live in that new callback but with the ability to intercept it directly.

swankjesse: I think your proposal works great for Retrofit.

But wrapping unchecked exceptions in IOExceptions is broken for anyone not anticipating it. The new default method is only good if you know to look for it.

swankjesse: Why don't you like the uncaught exception handler? Is there a situation where an unchecked exception will be caught and handled application code? I read this as basically deciding which way this goes to the uncaught exception handler.

JakeWharton: > But wrapping unchecked exceptions in IOExceptions is broken for anyone not anticipating it.

Where does this claim come from? How would the presence of a cause of any type on an IOException cause problems? You're already given no guarantees of the IOException subtype nor the presence or absence of a cause of any type through this delivery channel.

> The new default method is only good if you know to look for it.

I agree, but I like each other proposed solution less. I would have preferred 5 years ago that onFailure was given a Throwable argument.

> Is there a situation where an unchecked exception will be caught and handled application code?

It's certainly not implausible. If you're rate limited or some other invariant is violated you're forced to use exceptions, and nothing tells you that only IOException subtypes will be delivered to async users. The desire for fixing this at the Retrofit layer comes from user reports of this not behaving as expected.

Retrofit's use of the callback-based API is an implementation detail. Ideally this exception-handling difference between synchronous vs. asynchronous shouldn't need to leak into callers if they're entirely abstracting OkHttp.

> I read this as basically deciding which way this goes to the uncaught exception handler.

I suspect in the majority of cases this is true. I would just rather it crashes up through the caller stack rather than being indistinguishable from a normal cancel, even if only from the Retrofit layer, but ideally by anyone adapting OkHttp callback into suspend-based APIs.

swankjesse: It's

swankjesse: IOExceptions are caused by problems in the environment, usually all kinds of network connectivity problems. These should be surfaced to app users who can fix em. (Can't reach the server). 

RuntimeExceptions are caused by bugs in the code. They should be sent to the developer who can fix em.

I would very much like to not deliver RuntimeExceptions via IOExceptions because I think they'll get lost.

It's too bad that for unchecked exceptions our callback API is asymmetric with our synchronous API. But I actually prefer the API as-is over it passing a Throwable because it punishes the user who has to now make sure their programming bugs aren't lost.

```
fun onFailure(call: Call, throwable: Throwable) {
  if (throwable is IOException) {
    // Tell the user theres a connectivity problem.
  } else {
    // The programmer’s new problem.
  }
}
```

JakeWharton: I think the problem is that Retrofit wants to use the async API to play nice but it doesn't want an abstraction over async in the same way that a direct consumer wants. I guess we'll have to wait and see if it becomes a problem in practice.

Did this land in 3.x branch?

swankjesse: Lemme do that next.

danielesegato: I'm not sure if I'm missing something but this caused issue to me when i updated retrofit and stared using suspend fun rather than Rx (not sure if one or the other is the issue)

I now have to wrap my interceptors and unwrap the exception on each single method of my interfaces defined with Retrofit, really really annoying thing.

I've created this issue: https://github.com/square/retrofit/issues/3505 please reconsider

Issue #3146, opened 2017-01-31
java.net.SocketTimeoutException from HTTP/2 connection leaves dead okhttp clients in pool


kenyee:
Tried writing a unit test w/ TestButler on Android w/ no luck, so I'll write up the steps to reproduce this and include some sample code.  This happens if you connect to an HTTP/2 server and your network goes down while the okhttp client is connected to it:
1) create an okhttp client
2) tell it to read from the HTTP/2 server
3) bring the network down
4) tell it to read from the HTTP/2 server (it'll get a SocketTimeoutException)
5) bring the network back up
6) tell it to read from the HTTP/2 server again (it'll be stuck w/ SocketTimeoutExceptions)
7) if you create new http clients at this point, it'll work, but the dead http client will eventually come back in the pool and fail.

okhttp client should attempt to reopen the HTTP/2 connection instead of being stuck in this state

Code sample for Android (create a trivial view w/ a button and a textview):

```
public class MainActivity extends AppCompatActivity {
    OkHttpClient okhttpClient = new OkHttpClient();

    @Override
    protected void onCreate(Bundle savedInstanceState) {
        super.onCreate(savedInstanceState);
        setContentView(R.layout.activity_main);

        Button loadButton = (Button) findViewById(R.id.loadButton);
        TextView outputView = (TextView) findViewById(R.id.outputView);

        loadButton.setOnClickListener(view -> Observable.fromCallable(() -> {
                    Request request = new Request.Builder()
                            .url(<INSERT URL TO YOUR HTTP/2 SERVER HERE>)
                            .build();

                    Response response = okhttpClient.newCall(request).execute();

                    return response.body().string();
                })
                .subscribeOn(Schedulers.io())
                .observeOn(AndroidSchedulers.mainThread())
                .subscribe(outputView::setText, t -> outputView.setText(t.toString()))
        );
    }
}

```

kenyee:
FYI, one of my coworkers found a workaround...set the connectionPool in the builder so it uses a new connection pool w/ a size of zero and also turn off HTTP/2 support by setting a new protocolList in the builder with only HTTP/1.1 support.

swankjesse:
You’re using 3.6.0?

kenyee:
yep...3.6.0 unfortunately.  Thought about rolling back to pre-http/2 support but that would mean 2.2 which is too far back because of all the okhttp3 dependencies :-(

swankjesse:
[...]

PR #5461, opened 2019-09-15
TaskRunner, an abstraction over ExecutorService

swankjesse:
I want to tighten up our executors for a few reasons
 - Fix daemon vs. non-daemon problems
 - Fix code unloading problems
 - Be able to wait for async work to complete in tearDown()
 - Fewer threads for pongs, acks, and window updates

This shouldn't be too much to layer on top of ExecutorService, but
it is not easy to do with those APIs alone.

The underlying motivation is to make scheduling work easier in
Http2Connection, which will help with differentiating between
dead streams and dead connections.

https://github.com/square/okhttp/issues/3146

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/concurrent/Task.kt
+++ b/okhttp/src/main/java/okhttp3/internal/concurrent/Task.kt
@@ -1,0 +1,53 @@
+/*
+ * Copyright (C) 2019 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okhttp3.internal.concurrent
+
+/**
+ * A unit of work that can be executed one or more times.
+ *
+ * Cancellation
+ * ------------
+ *
+ * Tasks control their cancellation. If the hosting queue is canceled, the [Task.tryCancel] function
+ * returns true if the task should skip the next-scheduled execution. Note that canceling a task is
+ * not permanent; it is okay to schedule a task after it has been canceled.
+ *
+ * Recurrence
+ * ----------
+ *
+ * Tasks control their recurrence schedule. The [runOnce] function returns -1L to signify that the
+ * task should not be executed again. Otherwise it returns a delay until the next execution.
+ *
+ * A task has at most one next execution. If the same task instance is scheduled multiple times, the
+ * earliest one wins. This applies to both executions scheduled with [TaskRunner.Queue.schedule] and
+ * those implied by the returned execution delay.
+ *
+ * Task Queues
+ * -----------
+ *
+ * Tasks are bound to the [TaskQueue] they are scheduled in. Each queue is sequential and the tasks
+ * within it never execute concurrently. It is an error to use a task in multiple queues.
+ */
+abstract class Task(
+  val name: String,
+  val daemon: Boolean = true
+) {
+  /** Returns the delay in nanoseconds until the next execution, or -1L to not reschedule. */
+  abstract fun runOnce(): Long
+
+  /** Return true to skip the scheduled execution. */
+  open fun tryCancel(): Boolean = false
+}
--- a/okhttp/src/main/java/okhttp3/internal/concurrent/TaskQueue.kt
+++ b/okhttp/src/main/java/okhttp3/internal/concurrent/TaskQueue.kt
@@ -1,0 +1,50 @@
+/*
+ * Copyright (C) 2019 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okhttp3.internal.concurrent
+
+/**
+ * A set of tasks that are executed in sequential order.
+ *
+ * Work within queues is not concurrent. This is equivalent to each queue having a dedicated thread
+ * for its work; in practice a set of queues may share a set of threads to save resources.
+ */
+interface TaskQueue {
+  /**
+   * An application-level object like a connection pool or HTTP call that this queue works on behalf
+   * of. This is intended to be useful for testing and debugging only.
+   */
+  val owner: Any
+
+  /** Returns a snapshot of tasks currently scheduled for execution. */
+  val scheduledTasks: List<Task>
+
+  /**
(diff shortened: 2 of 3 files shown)


yschimke:
I think my main question is around the synchronous nature of execution. Whether we should allow for a scheduled model and async execution?

swankjesse:
I’d like to migrate a few things to this. We’re using different executors for these today, but we end up keeping a lot of idle threads because we want both isolation of queues (can’t have two writers running in parallel) and concurrency (writers and listeners should run concurrently).

### RealConnectionPool.cleanupRunnable
This evicts old connections from the pool. It needs an option to run immediately in case the pool is larger than it’s supposed to be. And there’s no value in running multiple cleanups concurrently.

### Http2Connection.pushExecutor
Callbacks need to be sent in the order they’re received.

### Http2Connection.writerExecutor
Needs to send pings at a fixed rate. Also other writes happen on-demand.

### Http2Connection.listenerExecutor
We can’t notify multiple listeners concurrently.

### RealWebSocket.executor
Needs to send pings at a fixed rate. Needs to cancel after a fixed maximum time.

### DiskLruCache.executor
Needs to cleanup when the journal is too large.

abc52090241:
請問我是一個新手不知道該怎麼操作 可以幫幫我嗎.....

okhttp/src/main/java/okhttp3/internal/concurrent/Task.kt
```diff
@@ -0,0 +1,53 @@
...
+ * -----------
+ *
+ * Tasks are bound to the [TaskQueue] they are scheduled in. Each queue is sequential and the tasks
+ * within it never execute concurrently. It is an error to use a task in multiple queues.
+ */
+abstract class Task(
+  val name: String,
+  val daemon: Boolean = true
+) {
+  /** Returns the delay until the next execution, or -1L to not reschedule. */
```
yschimke: specify units here.  I assume nanos?

okhttp/src/main/java/okhttp3/internal/concurrent/TaskQueue.kt
```diff
@@ -0,0 +1,50 @@
...
+   *
+   * The target execution time is implemented on a best-effort basis. If another task in this queue
+   * is running when that time is reached, that task is allowed to complete before this task is
+   * started. Similarly the task will be delayed if the host lacks compute resources.
+   */
+  fun schedule(task: Task, delayNanos: Long = 0L)
+
+  /**
+   * Schedules immediate execution of [Task.tryCancel] on all currently-enqueued tasks. These calls
+   * will not be made until any currently-executing task has completed. Tasks that return true will
```
yschimke: Is the cancellation delay to ensure cancels are sequential also?  Not seen before when the task would be executed?

okhttp/src/main/java/okhttp3/internal/concurrent/TaskRunner.kt
```diff
@@ -0,0 +1,35 @@
...
+ * The task runner is also responsible for releasing held threads when the library is unloaded.
+ * This is for the benefit of container environments that implement code unloading.
+ *
+ * Most applications should share a process-wide [TaskRunner] and use queues for per-client work.
+ */
+interface TaskRunner {
+  fun newQueue(owner: Any): TaskQueue
+
+  /** Returns a snapshot of queues that currently have tasks scheduled. */
+  fun activeQueues(): List<TaskQueue>
```
yschimke: Why only active queues?  What about getting all queues to easily report on empty queues.  Or should this be implicit because a known queue (held externally by reference) is missing?

okhttp/src/main/java/okhttp3/internal/concurrent/TaskQueue.kt
```diff
@@ -0,0 +1,50 @@
...
+ */
+package okhttp3.internal.concurrent
+
+/**
+ * A set of tasks that are executed in sequential order.
+ *
+ * Work within queues is not concurrent. This is equivalent to each queue having a dedicated thread
+ * for its work; in practice a set of queues may share a set of threads to save resources.
+ */
+interface TaskQueue {
```
yschimke: Should a task be self describing?  e.g. have a name also?

okhttp/src/main/java/okhttp3/internal/concurrent/TaskQueue.kt
```diff
@@ -0,0 +1,50 @@
...
+  /**
+   * An application-level object like a connection pool or HTTP call that this queue works on behalf
+   * of. This is intended to be useful for testing and debugging only.
+   */
+  val owner: Any
```
yschimke: Is this also used for any synchronization locks?

okhttp/src/main/java/okhttp3/internal/concurrent/TaskQueue.kt
swankjesse: Yeah, everything on a queue is sequential.

swankjesse: ... and so cancels only prevent future runs, they don't impact executing ones.

okhttp/src/main/java/okhttp3/internal/concurrent/TaskRunner.kt
swankjesse: This is an implementation trick that I should document: the runner knows about the queues but not vice-versa.

okhttp/src/main/java/okhttp3/internal/concurrent/Task.kt
swankjesse: Fixed.

okhttp/src/main/java/okhttp3/internal/concurrent/TaskRunner.kt
swankjesse: Fixed.

okhttp/src/main/java/okhttp3/internal/concurrent/TaskQueue.kt
swankjesse: I think I’ll probably add a name in follow-up. For the thread name.

okhttp/src/main/java/okhttp3/internal/concurrent/TaskQueue.kt
swankjesse: No.

okhttp/src/main/java/okhttp3/internal/concurrent/Task.kt
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * Tasks are bound to the [TaskQueue] they are scheduled in. Each queue is sequential and the tasks
+ * within it never execute concurrently. It is an error to use a task in multiple queues.
+ */
+abstract class Task(
+  val name: String,
+  val daemon: Boolean = true
+) {
+  /** Returns the delay in nanoseconds until the next execution, or -1L to not reschedule. */
+  abstract fun runOnce(): Long
```
Egorand: nit: `Once` feels redundant here. The semantics of `run()` are well known and it usually implies that the task is executed exactly once.
nit: Would `execute()` be a better name? In the docs, the verb "run" relating to "task" is used 0 times vs "execute" which is used 3 times.

okhttp/src/main/java/okhttp3/internal/concurrent/TaskQueue.kt
```diff
@@ -0,0 +1,50 @@
...
+   * scheduled. If the task is already in the queue, the earliest execution time is used.
+   *
+   * The target execution time is implemented on a best-effort basis. If another task in this queue
+   * is running when that time is reached, that task is allowed to complete before this task is
+   * started. Similarly the task will be delayed if the host lacks compute resources.
+   */
+  fun schedule(task: Task, delayNanos: Long = 0L)
+
+  /**
+   * Schedules immediate execution of [Task.tryCancel] on all currently-enqueued tasks. These calls
```
Egorand: Should the results of `tryCancel()` be somehow propagated? Can't think of a good way to do it though.

PR #5477, opened 2019-09-21
Fix master build

yschimke:
The compile target needs to run against Jdk 11


yschimke:
Failing here https://app.circleci.com/jobs/github/square/okhttp/5653

PR #5482, opened 2019-09-22
Make Platform.log exception parameter optional

yschimke:


yschimke:
In running Android related tests and builds, I'm using this frequently and so scratching an itch. Not a critical change if there is concern.

okhttp/src/main/java/okhttp3/internal/platform/Platform.kt
```diff
@@ -126,7 +126,7 @@ open class Platform {
     socket.connect(address, connectTimeout)
   }
 
-  open fun log(level: Int, message: String, t: Throwable?) {
+  open fun log(level: Int = INFO, message: String, t: Throwable? = null) {
```
swankjesse: Put message first so you can safely omit level?

yschimke: Sounds good.

PR #5485, opened 2019-09-22
Implement TaskRunner.

swankjesse:
This is attempting to balance simplicity, efficiency, and testability.
It doesn't use ScheduledExecutorService because that class wants a
permanent scheduler thread. Instead this uses a coordinator thread
that does its own wait and notify, similar to the mechanism in the
ConnectionPool that this is intended to replace.

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/Util.kt
+++ b/okhttp/src/main/java/okhttp3/internal/Util.kt
@@ -516,10 +516,8 @@ fun Int.toHexString(): String = Integer.toHexString(this)
  */
 @Throws(InterruptedException::class)
 fun Any.lockAndWaitNanos(nanos: Long) {
-  val ms = nanos / 1_000_000L
-  val ns = nanos - (ms * 1_000_000L)
   synchronized(this) {
-    waitMillis(ms, ns.toInt())
+    objectWaitNanos(nanos)
   }
 }
 
@@ -527,13 +525,16 @@ fun Any.lockAndWaitNanos(nanos: Long) {
 inline fun Any.wait() = (this as Object).wait()
 
 /**
- * Lock and wait a duration in milliseconds and nanos.
- * Unlike [java.lang.Object.wait] this interprets 0 as "don't wait" instead of "wait forever".
+ * Wait a duration in nanoseconds. Unlike [java.lang.Object.wait] this interprets 0 as "don't wait"
+ * instead of "wait forever".
  */
+@Throws(InterruptedException::class)
 @Suppress("PLATFORM_CLASS_MAPPED_TO_KOTLIN")
-fun Any.waitMillis(timeout: Long, nanos: Int = 0) {
-  if (timeout > 0L || nanos > 0) {
-    (this as Object).wait(timeout, nanos)
+fun Any.objectWaitNanos(nanos: Long) {
+  val ms = nanos / 1_000_000L
+  val ns = nanos - (ms * 1_000_000L)
+  if (ms > 0L || nanos > 0) {
+    (this as Object).wait(ms, ns.toInt())
   }
 }
 
@@ -565,4 +566,8 @@ fun <T> readFieldOrNull(instance: Any, fieldType: Class<T>, fieldName: String): 
   }
 
   return null
+}
+
+internal fun <E> MutableList<E>.addIfAbsent(element: E) {
+  if (!contains(element)) add(element)
 }
--- a/okhttp/src/main/java/okhttp3/internal/concurrent/Task.kt
+++ b/okhttp/src/main/java/okhttp3/internal/concurrent/Task.kt
@@ -45,9 +45,43 @@ abstract class Task(
   val name: String,
   val daemon: Boolean = true
 ) {
+  // Guarded by the TaskRunner.
+  internal var queue: TaskQueue? = null
+
+  /** Undefined unless this is in [TaskQueue.futureTasks]. */
+  internal var nextExecuteNanoTime = -1L
+
+  internal var runRunnable: Runnable? = null
+  internal var cancelRunnable: Runnable? = null
+
   /** Returns the delay in nanoseconds until the next execution, or -1L to not reschedule. */
   abstract fun runOnce(): Long
 
   /** Return true to skip the scheduled execution. */
   open fun tryCancel(): Boolean = false
+
+  internal fun initQueue(queue: TaskQueue) {
+    if (this.queue === queue) return
+
+    check(this.queue === null) { "task is in multiple queues" }
+    this.queue = queue
+
+    this.runRunnable = Runnable {
+      var delayNanos = -1L
+      try {
+        delayNanos = runOnce()
+      } finally {
+        queue.runCompleted(this, delayNanos)
+      }
+    }
+
+    this.cancelRunnable = Runnable {
+      var skipExecution = false
+      try {
+        skipExecution = tryCancel()
+      } finally {
+        queue.tryCancelCompleted(this, skipExecution)
+      }
+    }
+  }
 }
--- a/okhttp/src/main/java/okhttp3/internal/concurrent/TaskQueue.kt
+++ b/okhttp/src/main/java/okhttp3/internal/concurrent/TaskQueue.kt
@@ -15,21 +15,44 @@
  */
 package okhttp3.internal.concurrent
 
+import okhttp3.internal.addIfAbsent
+
 /**
  * A set of tasks that are executed in sequential order.
  *
  * Work within queues is not concurrent. This is equivalent to each queue having a dedicated thread
  * for its work; in practice a set of queues may share a set of threads to save resources.
  */
-interface TaskQueue {
+class TaskQueue internal constructor(
+  private val taskRunner: TaskRunner,
+
   /**
    * An application-level object like a connection pool or HTTP call that this queue works on behalf
    * of. This is intended to be useful for testing and debugging only.
    */
   val owner: Any
+) {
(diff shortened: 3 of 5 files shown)


okhttp/src/main/java/okhttp3/internal/concurrent/Task.kt
```diff
@@ -45,9 +45,43 @@ abstract class Task(
...
+  internal var cancelRunnable: Runnable? = null
+
   /** Returns the delay in nanoseconds until the next execution, or -1L to not reschedule. */
   abstract fun runOnce(): Long
 
   /** Return true to skip the scheduled execution. */
   open fun tryCancel(): Boolean = false
+
+  internal fun initQueue(queue: TaskQueue) {
+    if (this.queue == queue) return
```
yschimke: Should this be ===?

okhttp/src/main/java/okhttp3/internal/Util.kt
```diff
@@ -516,24 +516,25 @@ fun Int.toHexString(): String = Integer.toHexString(this)
...
+ * instead of "wait forever".
  */
+@Throws(InterruptedException::class)
 @Suppress("PLATFORM_CLASS_MAPPED_TO_KOTLIN")
-fun Any.waitMillis(timeout: Long, nanos: Int = 0) {
-  if (timeout > 0L || nanos > 0) {
-    (this as Object).wait(timeout, nanos)
+fun Any.objectWaitNanos(nanos: Long) {
+  val ms = nanos / 1_000_000L
+  val ns = nanos - (ms * 1_000_000L)
```
yschimke:
```suggestion
  val ns = nanos % 1_000_000L
```

okhttp/src/main/java/okhttp3/internal/concurrent/TaskQueue.kt
```diff
@@ -39,12 +62,126 @@ interface TaskQueue {
...
+    // If the task is already scheduled, take the earlier of the two times.
+    val existingIndex = futureTasks.indexOf(task)
+    if (existingIndex != -1) {
+      if (task.nextExecuteNanoTime <= executeNanoTime) return false // Already scheduled earlier.
+      futureTasks.removeAt(existingIndex) // Already scheduled later: reschedule below!
+    }
+    task.nextExecuteNanoTime = executeNanoTime
+
+    // Insert in chronological order.
+    var insertAt = futureTasks.indexOfFirst { it.nextExecuteNanoTime - now > delayNanos }
```
yschimke: nit: simpler to compare to task.nextExecuteNanoTime? without backing out now?

okhttp/src/main/java/okhttp3/internal/concurrent/TaskQueue.kt
```diff
@@ -39,12 +62,126 @@ interface TaskQueue {
...
+    }
+  }
+
+  /** Adds [task] to run in [delayNanos]. Returns true if the coordinator should run. */
+  private fun scheduleAndDecide(task: Task, delayNanos: Long): Boolean {
+    val now = taskRunner.backend.nanoTime()
+    val executeNanoTime = now + delayNanos
+
+    // If the task is already scheduled, take the earlier of the two times.
+    val existingIndex = futureTasks.indexOf(task)
```
yschimke: If you trusted task.nextExecuteTime then presumably you could simplify this by using a java PriorityQueue?  

null to x = insert
x to <x = remove and insert
x to >x = don't change nextExecuteTime or modify queue

okhttp/src/main/java/okhttp3/internal/concurrent/TaskRunner.kt
```diff
@@ -27,12 +36,130 @@ package okhttp3.internal.concurrent
...
+    fun coordinatorWait(taskRunner: TaskRunner, nanos: Long)
+  }
+
+  class RealBackend : Backend {
+    private val coordinatorExecutor: Executor = ThreadPoolExecutor(
+        0, // corePoolSize.
+        1, // maximumPoolSize.
+        60L, TimeUnit.SECONDS, // keepAliveTime.
+        SynchronousQueue(),
+        threadFactory("OkHttp Task Coordinator", false)
```
yschimke: This seems like a good opportunity to throw the container users a bone and allow creating a backend with smaller timeouts or all daemon threads?  Can we make this simple e.g. a RealBackend factory method to create for different uses?

okhttp/src/main/java/okhttp3/internal/concurrent/TaskRunner.kt
```diff
@@ -27,12 +36,130 @@ package okhttp3.internal.concurrent
...
+    )
+
+    private val taskExecutor: Executor = ThreadPoolExecutor(
+        0, // corePoolSize.
+        Int.MAX_VALUE, // maximumPoolSize.
```
yschimke: Do you want to support contrained environments? e.g. at most X threads or a BlockingQueue etc? later?

Once we allow this to be customised in anyway like Dispatcher we presumably need to handle some weird edge cases.


``` 
val d = Dispatcher(
          ThreadPoolExecutor(1, 10, 1, TimeUnit.MINUTES, ArrayBlockingQueue<Runnable>(5),
              ThreadFactoryBuilder().apply {
                setDaemon(true)
                setNameFormat("XXX")
              }.build()))
```

okhttp/src/main/java/okhttp3/internal/concurrent/Task.kt
swankjesse: Yes! Good catch.

okhttp/src/main/java/okhttp3/internal/Util.kt
swankjesse: I'm assuming subtract and multiply is faster than mod. Good assumption to validate.

okhttp/src/main/java/okhttp3/internal/concurrent/TaskRunner.kt
swankjesse: My aspiration is to make container users happy by making the defaults work for them! Threads should promptly exit when there's no work to do.

okhttp/src/main/java/okhttp3/internal/concurrent/TaskRunner.kt
swankjesse: I'd like to keep this class and it's backend as implementation details. I think our resource limiting needs to be based on limiting the number of TaskQueues, which indirectly limits the number of threads.

okhttp/src/main/java/okhttp3/internal/concurrent/TaskQueue.kt
swankjesse: Unfortunately `System.nanoTime()` is documented to permit wrapping around. 

For example, `now` could be `Long.MAX_VALUE - 100`, the existing task could be `Long.MAX_VALUE` and the task-being scheduled could be `Long.MAX_VALUE + 100` which is negative. We want to subtract off now to compare `100 < 200` instead of comparing the values-as-is which yields a different result.

I’ll add a comment!

okhttp/src/main/java/okhttp3/internal/concurrent/TaskQueue.kt
swankjesse: Great idea. Will do.

swankjesse: ... upon actually trying it, it’s unsatisfying because of the above `now` thing. I’m going to leave it as lists for now, unsatisfying as that is.

swankjesse:
Thanks for the quick turnaround on this.

okhttp/src/main/java/okhttp3/internal/Util.kt
JakeWharton: https://www.agner.org/optimize/instruction_tables.pdf

For an i7, multiply + subtract takes 4 operations and causes a delay of 4. Division takes 60 with a delay of 40-100.

okhttp/src/main/java/okhttp3/internal/concurrent/TaskQueue.kt
yschimke: Thanks for humoring me.

okhttp/src/main/java/okhttp3/internal/concurrent/TaskQueue.kt
yschimke: Watch out for

"Differences in successive calls that span greater than approximately 292 years (263 nanoseconds) will not correctly compute elapsed time due to numerical overflow. "

:)

okhttp/src/main/java/okhttp3/internal/Util.kt
oldergod: To me this is the kind of things I would expect the compiler to do for me. Does it happen sometimes but not always? do we just wanna guarantee that?

swankjesse: Really tough for the compiler to anticipate this case.

okhttp/src/main/java/okhttp3/internal/concurrent/TaskQueue.kt
swankjesse: I’m not too worried about scheduling things 292 years into the future. I am worried about what the nanoTime initial value is. Is it 0?

yschimke: Yep, wasn't arguing against your valid justification.  Just going back to basics and re-reading nanoTime, where I found that gem of a warning.

PR #5486, opened 2019-09-23
DiskLruCache snapshot() method removing entries that are being cached (3.14.x branch)

swankjesse:
Fixed a bug I found.
When using DiskLruCache, hasNext() method of Iterator<Snapshot> returned by snapshot() method iterates through lruEntries, however if entry is not completed yet, it is removed (together with dirtyFiles) as cleanFiles don't exist yet - during saving response to disk. This caused sometimes file not being cached at all.


swankjesse:
Cherry-picking from https://github.com/square/okhttp/pull/3905

yschimke:
Sorry for not following up and cherry picking the fix.  Thanks for this!

PR #5491, opened 2019-09-23
Revise isAndroid test

yschimke:


yschimke:
I'm not sure of the right fix.  But putting up this strawman because it's awkward code that needs to handle 4 environments

1) Deployed JVM
2) Deployed Android
3) IDE
4) Gradle

And it's failing in option 3 above.

okhttp/src/main/java/okhttp3/internal/platform/AndroidPlatform.kt
```diff
@@ -194,22 +194,26 @@ class AndroidPlatform : Platform() {
     val isAndroid: Boolean = try {
       // Trigger an early exception over a fatal error, prefer a RuntimeException over Error.
       Class.forName("com.android.org.conscrypt.OpenSSLSocketImpl")
```
yschimke: Should catch deployed environments here.  I'll add comments.

okhttp/src/main/java/okhttp3/internal/platform/AndroidPlatform.kt
```diff
@@ -194,22 +194,26 @@ class AndroidPlatform : Platform() {
...
+
+      // account for android-all
+      check(Build.VERSION.SDK_INT > 0)
```
yschimke: Should fail before accidentally running in Android mode on JVM.  Defaults to 0 internally.

yschimke:
One option is to check for JVM class and short circuit e.g. sun.security.ssl.SSLContextImpl?

But overall I expect this issue to be annoying until we fix properly.

swankjesse:
Seems reasonable enough.

PR #5493, opened 2019-09-23
Set the thread name to the currently-executing task's name.

swankjesse:
Also implement tests for thread interruption.
Also add a test that uses the real backend.

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/concurrent/Task.kt
+++ b/okhttp/src/main/java/okhttp3/internal/concurrent/Task.kt
@@ -67,20 +67,30 @@ abstract class Task(
     this.queue = queue
 
     this.runRunnable = Runnable {
+      val currentThread = Thread.currentThread()
+      val oldName = currentThread.name
+      currentThread.name = name
+
       var delayNanos = -1L
       try {
         delayNanos = runOnce()
       } finally {
         queue.runCompleted(this, delayNanos)
+        currentThread.name = oldName
       }
     }
 
     this.cancelRunnable = Runnable {
+      val currentThread = Thread.currentThread()
+      val oldName = currentThread.name
+      currentThread.name = name
+
       var skipExecution = false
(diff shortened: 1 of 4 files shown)


okhttp/src/main/java/okhttp3/internal/concurrent/Task.kt
```diff
@@ -67,20 +67,30 @@ abstract class Task(
...
       try {
         delayNanos = runOnce()
       } finally {
         queue.runCompleted(this, delayNanos)
+        currentThread.name = oldName
       }
     }
 
     this.cancelRunnable = Runnable {
+      val currentThread = Thread.currentThread()
```
yschimke: Why not use the inline function Util.threadName?

okhttp/src/main/java/okhttp3/internal/concurrent/TaskRunner.kt
```diff
@@ -161,5 +160,10 @@ class TaskRunner(
     override fun coordinatorWait(taskRunner: TaskRunner, nanos: Long) {
       taskRunner.objectWaitNanos(nanos)
     }
+
+    fun shutDown() {
+      coordinatorExecutor.shutdown()
+      coordinatorExecutor.shutdown()
```
yschimke:
```suggestion
      taskExecutor.shutdown()
```

okhttp/src/main/java/okhttp3/internal/concurrent/Task.kt
swankjesse: I’m hoping to delete it!

okhttp/src/main/java/okhttp3/internal/concurrent/TaskRunner.kt
swankjesse: Whoa embarrassing mistake!

PR #5494, opened 2019-09-24
Adopt TaskRunner in RealConnectionPool

swankjesse:
This also configures tests to assert that the connection pool
isn't doing any work after the test completes.

### DIFF
--- a/okhttp-testing-support/src/main/java/okhttp3/OkHttpClientTestRule.kt
+++ b/okhttp-testing-support/src/main/java/okhttp3/OkHttpClientTestRule.kt
@@ -15,6 +15,9 @@
  */
 package okhttp3
 
+import okhttp3.internal.concurrent.Task
+import okhttp3.internal.concurrent.TaskQueue
+import okhttp3.internal.concurrent.TaskRunner
 import okhttp3.testing.Flaky
 import org.assertj.core.api.Assertions.assertThat
 import org.junit.rules.TestRule
@@ -22,6 +25,8 @@ import org.junit.runner.Description
 import org.junit.runners.model.Statement
 import java.net.InetAddress
 import java.util.concurrent.ConcurrentLinkedDeque
+import java.util.concurrent.CountDownLatch
+import java.util.concurrent.TimeUnit
 
 /** Apply this rule to tests that need an OkHttpClient instance. */
 class OkHttpClientTestRule : TestRule {
@@ -59,6 +64,14 @@ class OkHttpClientTestRule : TestRule {
     }
   }
 
+  private fun ensureAllTaskQueuesIdle() {
+    for (queue in TaskRunner.INSTANCE.activeQueues()) {
+      assertThat(queue.awaitIdle(500L, TimeUnit.MILLISECONDS))
+          .withFailMessage("Queue ${queue.owner} still active after 500ms")
+          .isTrue()
+    }
+  }
+
   override fun apply(base: Statement, description: Description): Statement {
     return object : Statement() {
       override fun evaluate() {
@@ -72,6 +85,7 @@ class OkHttpClientTestRule : TestRule {
         } finally {
           ensureAllConnectionsReleased()
           releaseClient()
+          ensureAllTaskQueuesIdle()
         }
       }
 
@@ -117,6 +131,19 @@ class OkHttpClientTestRule : TestRule {
       it.dispatcher.executorService.shutdownNow()
       it.connectionPool.evictAll()
     }
+  }
+
+  /** Returns true if this queue became idle before the timeout elapsed. */
+  private fun TaskQueue.awaitIdle(timeout: Long, timeUnit: TimeUnit): Boolean {
+    val latch = CountDownLatch(1)
+    schedule(object : Task("awaitIdle") {
+      override fun runOnce(): Long {
+        latch.countDown()
+        return -1L
+      }
+    })
+
+    return latch.await(timeout, timeUnit)
   }
 
   companion object {
--- a/okhttp/src/main/java/okhttp3/ConnectionPool.kt
+++ b/okhttp/src/main/java/okhttp3/ConnectionPool.kt
@@ -16,6 +16,7 @@
  */
 package okhttp3
 
+import okhttp3.internal.concurrent.TaskRunner
 import okhttp3.internal.connection.RealConnectionPool
 import java.util.concurrent.TimeUnit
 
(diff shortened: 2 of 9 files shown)


okhttp-testing-support/src/main/java/okhttp3/OkHttpClientTestRule.kt
```diff
@@ -119,6 +133,19 @@ class OkHttpClientTestRule : TestRule {
     }
   }
 
+  /** Returns true if this queue became idle before the timeout elapsed. */
+  private fun TaskQueue.awaitIdle(timeout: Long, timeUnit: TimeUnit): Boolean {
+    val latch = CountDownLatch(1)
+    schedule(object : Task("awaitIdle") {
```
yschimke: Making a Task companion factory would be nice here.

```
Task("awaitIdle") {
```

yschimke: Should we try to put some unique prefix on task names?  So an app with >1 okhttp client, threads can be disambiguated?

okhttp/src/main/java/okhttp3/ConnectionPool.kt
```diff
@@ -29,12 +30,19 @@ import java.util.concurrent.TimeUnit
...
+class ConnectionPool internal constructor(
+  internal val delegate: RealConnectionPool
 ) {
-  internal val delegate = RealConnectionPool(maxIdleConnections, keepAliveDuration, timeUnit)
+  constructor(
+    maxIdleConnections: Int,
+    keepAliveDuration: Long,
+    timeUnit: TimeUnit
+  ) : this(RealConnectionPool(
+      taskRunner = TaskRunner.INSTANCE,
```
yschimke: How will we detect contention of coordinatorExecutor since its now a process singleton thread?

okhttp/src/main/java/okhttp3/internal/connection/RealConnectionPool.kt
```diff
@@ -113,10 +99,10 @@ class RealConnectionPool(
     assert(Thread.holdsLock(this))
     return if (connection.noNewExchanges || maxIdleConnections == 0) {
       connections.remove(connection)
+      if (connections.isEmpty()) cleanupQueue.cancelAll()
       true
     } else {
-      // Awake the cleanup thread: we may have exceeded the idle connection limit.
-      this.notifyAll()
+      cleanupQueue.schedule(cleanupTask, 0L)
```
yschimke: worth making 0L a default?

yschimke:
I like where this is going.

okhttp-testing-support/src/main/java/okhttp3/OkHttpClientTestRule.kt
swankjesse: Good idea. But unfortunately from this layer they’re all pretty symmetric.

We’d need to either name the OkHttpClient, or name its component parts, or something. Any recommendations on what the names should be?

okhttp/src/main/java/okhttp3/ConnectionPool.kt
swankjesse: I don’t anticipate contention on it because it doesn’t do anything other than kickoff tasks and sleep.

swankjesse: ... and it should not do anything else unless we address that.

That said, it’d be interesting to shrink the total number of threads for a RealConnectionPool use case from 2 to 1 by sharing threads between task execution and the coordinator. The implementation gets a little more complex because threads would need to switch roles.

okhttp/src/main/java/okhttp3/internal/connection/RealConnectionPool.kt
swankjesse: Good idea. Done.

okhttp-testing-support/src/main/java/okhttp3/OkHttpClientTestRule.kt
swankjesse: I tried it and it doesn’t come out as satisfying as I was hoping because of the return value.

```
    schedule(Task("awaitIdle") {
      latch.countDown()
      return@Task -1L
    })
```

Lemme convert a few more of our thread pools over to use this new framework and then I’d like to take another attempt at tightening up the syntax.

okhttp/src/main/java/okhttp3/internal/concurrent/TaskRunner.kt
```diff
@@ -132,7 +132,7 @@ class TaskRunner(
         1, // maximumPoolSize.
         60L, TimeUnit.SECONDS, // keepAliveTime.
         SynchronousQueue(),
-        threadFactory("OkHttp Task Coordinator", false)
+        threadFactory("OkHttp Task Coordinator", true)
```
swankjesse: Dropping support for non-daemon threads for now.

swankjesse:
@yschimke please take a look! I’m going to start rolling the TaskRunner out to the HTTP/2 and web sockets thread pools next!

okhttp/src/main/java/okhttp3/internal/concurrent/TaskRunner.kt
```diff
@@ -166,4 +166,8 @@ class TaskRunner(
       taskExecutor.shutdown()
     }
   }
+
+  companion object {
+    val INSTANCE = TaskRunner(RealBackend())
```
yschimke: If this INSTANCE gets shutdown via INSTANCE.backend.shutDown() are you happy with RejectedExecutionException being raised?  Or do you want a clearer message to educate the developer?  OR should this instance not allow shutdown?

swankjesse: This one shouldn’t ever be shut down. It’s intended as an internal API but I’ve learned that sometimes developers can’t help themselves!

swankjesse:
Ahhh, yes I did spell shutdown wrong. Fixed!

Also tightened visibility to `internal` for the RealBackend APIs. Those APIs are not for application developers to use!

okhttp/src/main/java/okhttp3/internal/concurrent/TaskRunner.kt
```diff
@@ -126,13 +127,13 @@ class TaskRunner(
...
 
-  class RealBackend : Backend {
+  internal class RealBackend : Backend {
     private val coordinatorExecutor = ThreadPoolExecutor(
         0, // corePoolSize.
         1, // maximumPoolSize.
         60L, TimeUnit.SECONDS, // keepAliveTime.
-        SynchronousQueue(),
-        threadFactory("OkHttp Task Coordinator", false)
+        LinkedBlockingQueue<Runnable>(),
```
swankjesse: I had to switch this to prevent RejectedExecutionExceptions when a coordinator quickly exits and then starts again.

PR #5496, opened 2019-09-26
Make the FakeBackend a top-level type for testing task things.

swankjesse:
Adopt TaskRunner in DiskLruCache. This one is pretty straightforward.
No scheduling and no canceling means we only enqueue work for
immediate execution.


swankjesse:
Really good feedback on stuck queues.

One weird side effect of this is that it can be somewhat self-hosting. We can create a health checks queue, and when we enqueue say, the DiskLruCache task we can also enqueue a healthcheck task that confirms it completed normally. And we can cancel the healthcheck when it completes!

Nothing designed yet, and I’m not quite sure where that goes, but it’s fun.

Issue #5468, opened 2019-09-18
Ampersand "&" in the request in the filter body is not encoded as "%26"


Valentin0810:
When the field value contains the ampersand "&" for example, "Brothers & Co" and I use that value like filter in the query, after encoding string to query I have ampersand "&" in the filter body, like this:
 https://example.eu/exsource/User?ormat=json&$filter=(username%20eq%20%27examlpleusername%27%20and%20filtername%20eq%20%2Brothers%20&%20Co%27)&asOfDate=2019-09-13

So this request does not work and I become Http response with status code 500. For correct work I should to convert it to "%26" additionally.

swankjesse:
Seems like a bug. Which version are you using?

Valentin0810:
Now I am using the latest version - 4.2.0

archibishop:
@Valentin0810 Could you please add the snippet of the code that fails to encode the ampersand. I am trying to investigate the issue so i can take it up but it seems to be working fine. Thanks.

Valentin0810:
@archibishop The snippet of the code that fails to encode the ampersand: 
       
 String url = "https://example.eu/exsource/User?format=json&$filter=filtername eq 'Brothers & Co'";
 Request request = new Request.Builder().url(url).build();
 new OkHttpClient.Builder().build().newCall(request).execute();

Thank you very much for your efforts.

lutzhorn:
Do you expect OkHttp to parse the query part of the URL? How should it know that the first `&` is to b used as is while the second shall be encoded? Maybe it would be better to build the URL using a library or a helper method.

archibishop:
I am working on something to fix this issue but for now you can use what has been suggested by @lutzhorn.

swankjesse:
I’m sorry I misunderstood the original report here. I was under the impression that URLs constructed by OkHttp with our `HttpUrl.Builder` were not correctly encoding `&`. But in this case the URL in is the URL out, and that’s working as designed.

I believe the real problem is whatever creates this URL:
```
https://example.eu/exsource/User?ormat=json&$filter=(username%20eq%20%27examlpleusername%27%20and%20filtername%20eq%20%2Brothers%20&%20Co%27)&asOfDate=2019-09-13
```

If you use HttpUrl.Builder to make it, it’ll do the right thing.

```
    HttpUrl url = new HttpUrl.Builder()
        .scheme("https")
        .host("example.eu")
        .encodedPath("/exsource/User")
[...]

PR #5499, opened 2019-09-26
Encode ampersand that is part of a string in the query parameter.

archibishop:
Fixes [#5468](https://github.com/square/okhttp/issues/5468)

Encodes the ampersand that has been passed as part of a string in the query parameter to ```%26```. for example ```http://host/?foo=Foo&bar=Bar&Co&baz=Baz``` to ```http://host/?foo=Foo&bar=Bar%26Co&baz=Baz```.

Thanks.

### DIFF
--- a/okhttp/src/main/java/okhttp3/HttpUrl.kt
+++ b/okhttp/src/main/java/okhttp3/HttpUrl.kt
@@ -1785,6 +1785,13 @@ class HttpUrl internal constructor(
     ): String {
       var codePoint: Int
       var i = pos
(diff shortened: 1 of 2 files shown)


swankjesse:
This behavior is not to spec. The equals sign convention is not guaranteed, and there are perfectly normal situations where there is no equals sign.

We also have an aspiration where parse() and toString() retain the original encoding. This doesn’t honor that either.
https://publicobject.com/2017/08/01/url-encoding-is-material/

PR #5507, opened 2019-09-27
Create OkHttpClient instances eagerly in tests

swankjesse:
We used to do this lazily because OkHttpClient instances were
somewhat heavy: each standalone instance held its own
ExecutorService for the connection pool.

Now that we have TaskRunner each instance is much more
lightweight and the drawbacks of creating instances eagerly
is negligible.

### DIFF
--- a/okhttp-testing-support/src/main/java/okhttp3/OkHttpClientTestRule.kt
+++ b/okhttp-testing-support/src/main/java/okhttp3/OkHttpClientTestRule.kt
@@ -22,31 +22,36 @@ import org.junit.rules.TestRule
 import org.junit.runner.Description
 import org.junit.runners.model.Statement
 import java.net.InetAddress
-import java.util.concurrent.ConcurrentLinkedDeque
 import java.util.concurrent.TimeUnit
 
 /** Apply this rule to tests that need an OkHttpClient instance. */
 class OkHttpClientTestRule : TestRule {
   private val clientEventsList = mutableListOf<String>()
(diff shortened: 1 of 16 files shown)


yschimke:
Haven’t reviewed yet. But this should make things fun if new advice becomes always create new client instances. I assume you don’t go that far?

swankjesse:
Not that far. But clients are cheaper now than they’ve ever been.

okhttp-testing-support/src/main/java/okhttp3/OkHttpClientTestRule.kt
```diff
@@ -86,15 +91,8 @@ class OkHttpClientTestRule : TestRule {
...
-      private fun acquireClient() {
-        prototype = prototypes.poll() ?: freshClient()
-      }
-
       private fun releaseClient() {
-        prototype?.let {
-          prototypes.push(it)
-          prototype = null
-        }
+        (testClient?.dispatcher as? ExecutorService)?.shutdown()
```
yschimke: Are you sure the safe cast of dispatcher to ExecutorService actually fires?

swankjesse: It is definitely wrong. Needed to be this:

```
        testClient?.dispatcher?.executorService?.shutdown()
```

swankjesse: Great catch!

PR #5508, opened 2019-09-28
Rewrite Dokka-generated markdown to include titles.

swankjesse:
I tried rendering in jekyll mode but that doesn't look as nice with
Mkdocs.

### DIFF
--- a/deploy_website.sh
+++ b/deploy_website.sh
@@ -30,6 +30,26 @@ cd $DIR
   :okhttp-urlconnection:dokka \
   :okhttp:dokka
 
+# Dokka filenames like `-http-url/index.md` don't work well with MkDocs <title> tags.
+# Assign metadata to the file's first Markdown heading.
+# https://www.mkdocs.org/user-guide/writing-your-docs/#meta-data
+title_markdown_file() {
+  TITLE_PATTERN="s/^[#]+ *(.*)/title: \1 - OkHttp/"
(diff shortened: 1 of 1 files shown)


swankjesse:
Screen Shot 2019-09-28 at 11 51 47 AM

deploy_website.sh
```diff
@@ -30,6 +30,26 @@ cd $DIR
   :okhttp-urlconnection:dokka \
   :okhttp:dokka
 
+# Dokka filenames like `-http-url/index.md` don't work well with MkDocs <title> tags.
+# Assign metadata to the file's first Markdown heading.
+# https://www.mkdocs.org/user-guide/writing-your-docs/#meta-data
+title_markdown_file() {
+  TITLE_PATTERN="s/^[#][#]* *\(.*\)/title: \1 - OkHttp/g"
```
yschimke:
```suggestion
  TITLE_PATTERN="s/^[#]+ *\(.*\)/title: \1 - OkHttp/g"
```

Why g? all instances rather than first?

swankjesse: Just habit.

swankjesse: Oh and I got `+` to work but I needed the `-E` flag because sed is from the 1970s

swankjesse:
Worth it!
Screen Shot 2019-10-06 at 10 47 21 AM

PR #5509, opened 2019-09-28
Add Kotlin listings for OkHttp recipes doc

swankjesse:


mkdocs.yml
```diff
@@ -33,6 +33,7 @@ markdown_extensions:
   - pymdownx.magiclink
   - pymdownx.smartsymbols
   - pymdownx.superfences
+  - pymdownx.tilde
```
swankjesse: necessary for `~~foo~~` to render foo in strikeout, which we do on our TLS configuration history page

samples/guide/src/main/java/okhttp3/recipes/kt/Authenticate.kt
```diff
@@ -0,0 +1,60 @@
...
+          val credential = Credentials.basic("jesse", "password1")
+          return response.request.newBuilder()
+              .header("Authorization", credential)
+              .build()
+        }
+      })
+      .build()
+
+  fun run() {
+    val request = Request.Builder()
```
yschimke: Any idea what breaks github highlighting after this point?

image

samples/guide/src/main/java/okhttp3/recipes/kt/ParseResponseWithMoshi.kt
```diff
@@ -0,0 +1,54 @@
...
+      val gist = gistJsonAdapter.fromJson(response.body!!.source())
+
+      for ((key, value) in gist!!.files!!) {
+        println(key)
+        println(value.content)
+      }
+    }
+  }
+
+  @JsonClass(generateAdapter = true)
```
yschimke: nice example

samples/guide/src/main/java/okhttp3/recipes/kt/ParseResponseWithMoshi.kt
```diff
@@ -0,0 +1,54 @@
...
+ */
+package okhttp3.recipes.kt
+
+import com.squareup.moshi.JsonClass
+import com.squareup.moshi.Moshi
+import okhttp3.OkHttpClient
+import okhttp3.Request
+import java.io.IOException
+
+class ParseResponseWithMoshi {
```
yschimke: You wanted these examples completely self contained, so maybe add the kapt line from the gradle build file as a class comment?

swankjesse: I do, but I also don’t want to go deep into explaining Moshi stuff. I think it’ll become clear if they try and Moshi tells you you forgot to run the generator.

samples/guide/src/main/java/okhttp3/recipes/kt/Authenticate.kt
swankjesse: It usually renders comments in grey, so it’s got to be confused when it gets there.

PR #5510, opened 2019-09-29
Use TaskRunner in web sockets.

swankjesse:
The most consequential change here is that the background thread is
now a daemon thread. A program consisting of exactly one web socket
will exit when the reader reads an inbound close message. Previously
such a program would continue running until the writer acknowledged
this close.

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/http2/Http2Connection.kt
+++ b/okhttp/src/main/java/okhttp3/internal/http2/Http2Connection.kt
@@ -433,7 +433,7 @@ class Http2Connection internal constructor(builder: Builder) : Closeable {
 
   /**
    * Closes this connection. This cancels all open streams and unanswered pings. It closes the
-   * underlying input and output streams and shuts down internal executor services.
+   * underlying input and output streams and shuts down internal executor services and task queues.
    */
   override fun close() {
     close(ErrorCode.NO_ERROR, ErrorCode.CANCEL, null)
@@ -687,7 +687,7 @@ class Http2Connection internal constructor(builder: Builder) : Closeable {
      *
      * Since we can't ACK settings on the current reader thread (the reader thread can't write) we
      * execute all peer settings logic on the writer thread. This relies on the fact that the
-     * writer executor won't reorder tasks; otherwise settings could be applied in the opposite
+     * writer task queue won't reorder tasks; otherwise settings could be applied in the opposite
      * order than received.
      */
     fun applyAndAckSettings(clearPrevious: Boolean, settings: Settings) {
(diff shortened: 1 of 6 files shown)


swankjesse:
Possible solutions to the daemon thread problem:

1. Permit multiple TaskRunner instances, each daemon or non-daemon
2. Permit a TaskRunner instance to host both daemon and non-daemon queues
3. Ignore the bug; nobody cares that we don’t cleanly ack a close in a program running exactly one web socket and no other non-daemon threads.

yschimke:
I think overall most people will like and welcome the daemon thread behaviour.  But it will probably break some apps, so we should call it out in release notes.

We should handle cases where the app terminates during an operation e.g. writing a cache entry, so it shouldn't ever be terminal.

my 2c

okhttp/src/main/java/okhttp3/internal/http2/Http2Connection.kt
```diff
@@ -433,7 +433,7 @@ class Http2Connection internal constructor(builder: Builder) : Closeable {
 
   /**
    * Closes this connection. This cancels all open streams and unanswered pings. It closes the
-   * underlying input and output streams and shuts down internal executor services.
+   * underlying input and output streams and shuts down internal executor services and task queues.
```
yschimke: Didn't your last commit remove "internal executor services"?

okhttp/src/main/java/okhttp3/internal/ws/RealWebSocket.kt
```diff
@@ -129,13 +122,18 @@ class RealWebSocket(
     }
 
     this.key = ByteArray(16).apply { random.nextBytes(this) }.toByteString().base64()
-    this.writerRunnable = Runnable {
-      try {
-        while (writeOneFrame()) {
+    this.writerTask = object : Task("${taskQueue.owner} Writer") {
+      override fun runOnce(): Long {
+        try {
+          while (writeOneFrame()) {
```
yschimke: Why is the loop here rather than returning 0?

okhttp/src/main/java/okhttp3/internal/http2/Http2Connection.kt
swankjesse: Unfortunately there’s one more to sweep up that handles incoming pushes. That’s on the to-do list!

okhttp/src/main/java/okhttp3/internal/ws/RealWebSocket.kt
swankjesse: Good question! There’s a short answer and a long answer.

The short answer is that returning 0 is correct and I should do that.

The long answer is that I want to add a feature to TaskRunner so it requires fewer context switches. Right now returning 0 will context switch to the coordinator thread and then context switch back to the task thread. I think I can make it skip these context switches but it’s a code change I need to make.

I’m going to change this to return 0 now ’cause that’s correct locally, and I’ve got it in my plan to eliminate context switches in TaskRunner.

PR #5511, opened 2019-09-29
Fix Android Certificate Check with localhost certificates

yschimke:


android-test/src/androidTest/java/okhttp/android/test/OkHttpTest.kt
```diff
@@ -391,9 +392,20 @@ class OkHttpTest {
...
+    val trustManager = handshakeCertificates.trustManager
+    Platform.get().log("tm $trustManager", Platform.WARN)
+
+    val toList = trustManager.acceptedIssuers.toList().map {
+      it.subjectDN
+    }
+
+    Platform.get().log("tmIssuers $toList", Platform.WARN)
+
+    // TODO: Bug fixed with new client
```
yschimke: Working on the fix, be good to get in for this release.

Reusing a client via newBuilder and setting trustManager doesn't use the correct CertificateChainCleaner.

swankjesse: Ahhhh, that explains it. Good find.

okhttp/src/main/java/okhttp3/internal/platform/android/AndroidCertificateChainCleaner.kt
```diff
@@ -48,7 +50,8 @@ internal class AndroidCertificateChainCleaner(
   }
 
   override fun equals(other: Any?): Boolean =
-      other is AndroidCertificateChainCleaner // All instances are equivalent.
+      other is AndroidCertificateChainCleaner &&
+          other.trustManager === this.trustManager
```
yschimke: This could break connection pooling for some clients, but hopefully an edge case.

okhttp/src/main/java/okhttp3/internal/platform/android/AndroidCertificateChainCleaner.kt
```diff
@@ -48,7 +50,22 @@ internal class AndroidCertificateChainCleaner(
...
 
-  override fun hashCode(): Int = 0
+  override fun hashCode(): Int = System.identityHashCode(trustManager.hashCode())
+
+  companion object {
+    fun build(trustManager: X509TrustManager): AndroidCertificateChainCleaner? = try {
```
yschimke: Overall attempt to simplify AndroidPlatform, at some point we may want a gradual progression from Android21 to Android29 with different features enabled etc.

abc52090241: 我該怎麼做

okhttp/src/main/java/okhttp3/internal/platform/android/Android10CertificateChainCleaner.kt
```diff
@@ -48,7 +48,8 @@ internal class Android10CertificateChainCleaner(
   }
 
   override fun equals(other: Any?): Boolean =
-      other is Android10CertificateChainCleaner // All instances are equivalent.
+      other is Android10CertificateChainCleaner &&
+          other.trustManager === this.trustManager
```
swankjesse: 👍

okhttp/src/main/java/okhttp3/internal/platform/android/Android10CertificateChainCleaner.kt
```diff
@@ -48,7 +48,8 @@ internal class Android10CertificateChainCleaner(
...
 
-  override fun hashCode(): Int = 1
+  override fun hashCode(): Int = System.identityHashCode(trustManager.hashCode())
```
swankjesse: Why not `trustManager.hashCode()`

okhttp/src/main/java/okhttp3/internal/platform/android/AndroidCertificateChainCleaner.kt
```diff
@@ -48,7 +50,22 @@ internal class AndroidCertificateChainCleaner(
   }
 
   override fun equals(other: Any?): Boolean =
-      other is AndroidCertificateChainCleaner // All instances are equivalent.
+      other is AndroidCertificateChainCleaner &&
+          other.trustManager === this.trustManager
 
-  override fun hashCode(): Int = 0
+  override fun hashCode(): Int = System.identityHashCode(trustManager.hashCode())
```
swankjesse: `trustManager.hashCode()`

okhttp/src/main/java/okhttp3/internal/platform/android/Android10CertificateChainCleaner.kt
yschimke: To match the use of === above. I can change above to use == and therefore equals. 

They are effectively the same as far as I know. Implementations don’t seem to implement hashcode or equals.

swankjesse: Got it. You definitely don't want the identity hash code of the hash code, that's just a random value. I think probably regular equals and hashCode is best, that way you're not baking assumptions of how these classes are implemented.

yschimke: Thanks, typo.

swankjesse:
Thanks. Very nice to have this fixed.

PR #5513, opened 2019-09-29
Change HTTP/2 connections to track settings snapshots

swankjesse:
Previously we mutated one instance as directed by the peer. This
was difficult to reason about.

This change includes migration of settings listeners to TaskRunner.
TaskRunner makes it easier to notify the listener of settings
changes serially, which makes these snapshots more useful.

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/http2/Http2Connection.kt
+++ b/okhttp/src/main/java/okhttp3/internal/http2/Http2Connection.kt
@@ -21,14 +21,12 @@ import okhttp3.internal.closeQuietly
 import okhttp3.internal.concurrent.Task
 import okhttp3.internal.concurrent.TaskRunner
 import okhttp3.internal.connectionName
-import okhttp3.internal.execute
 import okhttp3.internal.http2.ErrorCode.REFUSED_STREAM
 import okhttp3.internal.http2.Settings.Companion.DEFAULT_INITIAL_WINDOW_SIZE
 import okhttp3.internal.ignoreIoExceptions
 import okhttp3.internal.notifyAll
 import okhttp3.internal.platform.Platform
 import okhttp3.internal.platform.Platform.Companion.INFO
-import okhttp3.internal.threadFactory
 import okhttp3.internal.toHeaders
 import okhttp3.internal.wait
 import okio.Buffer
@@ -42,8 +40,6 @@ import java.io.Closeable
 import java.io.IOException
 import java.io.InterruptedIOException
 import java.net.Socket
-import java.util.concurrent.SynchronousQueue
-import java.util.concurrent.ThreadPoolExecutor
 import java.util.concurrent.TimeUnit
 
 /**
@@ -71,10 +67,7 @@ class Http2Connection internal constructor(builder: Builder) : Closeable {
   /** True if this peer initiated the connection. */
   internal val client: Boolean = builder.client
 
-  /**
-   * User code to run in response to incoming streams or settings. Calls to this are always invoked
-   * on [listenerExecutor].
-   */
+  /** User code to run in response to incoming streams or settings. */
   internal val listener: Listener = builder.listener
   internal val streams = mutableMapOf<Int, Http2Stream>()
   internal val connectionName: String = builder.connectionName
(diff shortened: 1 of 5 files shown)


swankjesse:
Also makes sense to make Settings actually immutable. In follow-up!

okhttp/src/main/java/okhttp3/internal/http2/Http2Connection.kt
```diff
@@ -639,16 +635,22 @@ class Http2Connection internal constructor(builder: Builder) : Closeable {
...
-                newStream.close(ErrorCode.PROTOCOL_ERROR, e)
+
+          // Use a different task queue for each stream because they should be handled in parallel.
+          val taskName = "OkHttp $connectionName stream $streamId"
+          taskRunner.newQueue(taskName).schedule(object : Task(taskName) {
+            override fun runOnce(): Long {
+              try {
+                listener.onStream(newStream)
+              } catch (e: IOException) {
+                Platform.get().log("Http2Connection.Listener failure for $connectionName", INFO, e)
```
yschimke: Should this logging be a part of the task contract?  e.g. rethrow to log?

okhttp/src/main/java/okhttp3/internal/http2/Http2Connection.kt
```diff
@@ -639,16 +635,22 @@ class Http2Connection internal constructor(builder: Builder) : Closeable {
...
+                ignoreIoExceptions {
+                  newStream.close(ErrorCode.PROTOCOL_ERROR, e)
+                }
               }
+              return -1L
```
yschimke: I wonder if there is a better way to hide this magic value?  e.g. 

```
Task.once {
   ...
}
```

okhttp/src/main/java/okhttp3/internal/http2/Http2Connection.kt
yschimke: or task.logErrors() fluent API?

okhttp/src/main/java/okhttp3/internal/http2/Http2Connection.kt
```diff
@@ -691,21 +693,38 @@ class Http2Connection internal constructor(builder: Builder) : Closeable {
      * order than received.
      */
     fun applyAndAckSettings(clearPrevious: Boolean, settings: Settings) {
-      var delta = 0L
-      var streamsToNotify: Array<Http2Stream>? = null
+      var delta: Long
+      var streamsToNotify: Array<Http2Stream>?
```
yschimke: Should this change to List?

okhttp/src/main/java/okhttp3/internal/http2/Http2Connection.kt
```diff
@@ -691,21 +693,38 @@ class Http2Connection internal constructor(builder: Builder) : Closeable {
...
-          peerSettings.merge(settings)
-          val peerInitialWindowSize = peerSettings.initialWindowSize
-          if (peerInitialWindowSize != -1 && peerInitialWindowSize != priorWriteWindowSize) {
-            delta = (peerInitialWindowSize - priorWriteWindowSize).toLong()
-            streamsToNotify = if (streams.isNotEmpty()) streams.values.toTypedArray() else null
+          val previousPeerSettings = peerSettings
+          if (clearPrevious) {
+            newPeerSettings = settings
+          } else {
+            newPeerSettings = Settings()
```
yschimke: Maybe cleaner as an apply with merge inside the block?

okhttp/src/main/java/okhttp3/internal/http2/Http2Connection.kt
swankjesse: No I don’t think so. Uncaught exceptions should go to the VM’s UncaughtExceptionHandler because they indicate a bug.

This IOException is expected and recoverable and handled differently.

okhttp/src/main/java/okhttp3/internal/http2/Http2Connection.kt
swankjesse: Probably. I’d like to get the behavior all good in the first pass, and then think about improving the syntax. This is internal-facing only so my preference is to be pretty frugal with fancy things.

okhttp/src/main/java/okhttp3/internal/http2/Http2Connection.kt
swankjesse: Good idea. Can do.

okhttp/src/main/java/okhttp3/internal/http2/Http2Connection.kt
swankjesse: In this case I’m trading off efficiency vs. readability. Probably the extra allocation of the list is why I avoided it initially. I don’t feel too strongly but I like treating this as “inner loop” code by default.

PR #5517, opened 2019-10-02
Make TaskRunner cancellation less dynamic

swankjesse:
We don't need to be called back when a call is being canceled.
This was unnecessary complexity that I thought I'd need for OkHttp
but I don't.

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/concurrent/Task.kt
+++ b/okhttp/src/main/java/okhttp3/internal/concurrent/Task.kt
@@ -18,13 +18,6 @@ package okhttp3.internal.concurrent
 /**
  * A unit of work that can be executed one or more times.
  *
- * Cancellation
- * ------------
- *
- * Tasks control their cancellation. If the hosting queue is canceled, the [Task.tryCancel] function
- * returns true if the task should skip the next-scheduled execution. Note that canceling a task is
- * not permanent; it is okay to schedule a task after it has been canceled.
- *
  * Recurrence
  * ----------
  *
@@ -35,6 +28,18 @@ package okhttp3.internal.concurrent
  * earliest one wins. This applies to both executions scheduled with [TaskRunner.Queue.schedule] and
  * those implied by the returned execution delay.
  *
+ * Cancellation
+ * ------------
+ *
+ * Tasks may be canceled while they are waiting to be executed, or while they are executing.
+ *
+ * Canceling a task that is waiting to execute prevents that upcoming execution. Canceling a task
+ * that is currently executing does not impact the ongoing run, but it does prevent a recurrence
+ * from being scheduled.
+ *
+ * Tasks may opt-out of cancellation with `cancelable = false`. Such tasks will recur until they
+ * decide not to by returning -1L.
+ *
  * Task Queues
  * -----------
  *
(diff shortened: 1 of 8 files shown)


okhttp/src/main/java/okhttp3/internal/concurrent/TaskRunner.kt
```diff
@@ -113,8 +113,8 @@ class TaskRunner(
   }
 
   private fun cancelAll() {
-    for (queue in activeQueues) {
-      queue.cancelAll()
+    for (i in activeQueues.size - 1 downTo 0) {
```
yschimke:
```suggestion
    for (queue in activeQueues.asReversed()) {
```

okhttp/src/main/java/okhttp3/internal/concurrent/Task.kt
```diff
@@ -35,14 +28,26 @@ package okhttp3.internal.concurrent
...
+ * Cancellation
+ * ------------
+ *
+ * Tasks may be scheduled while they are waiting to be executed, or while they are executing.
+ *
+ * Canceling a task that is waiting to execute prevents that upcoming execution. Canceling a task
+ * that is currently executing does not impact the ongoing run, but it does prevent a recurrence
+ * from being scheduled.
+ *
+ * Tasks may opt-out of cancellation with `cancelable = false`.
```
yschimke: Can you expand on this, specifically does it mean that future executions will still be run even though task queue was cancelled?

yschimke:
Build Failure

```
> Task :okhttp:compileKotlin FAILED
e: /home/circleci/project/okhttp/src/main/java/okhttp3/internal/connection/RealConnectionPool.kt: (45, 5): 'tryCancel' overrides nothing
e: /home/circleci/project/okhttp/src/main/java/okhttp3/internal/http2/Http2Connection.kt: (150, 9): 'tryCancel' overrides nothing
```

okhttp/src/main/java/okhttp3/internal/concurrent/TaskRunner.kt
swankjesse: Tragically this allocates!

okhttp/src/test/java/okhttp3/internal/concurrent/TaskRunnerTest.kt
```diff
@@ -208,10 +193,24 @@ class TaskRunnerTest {
...
+    taskFaker.advanceUntil(0L)
+    assertThat(log).isEmpty()
+
+    taskFaker.advanceUntil(100L)
+    assertThat(log).containsExactly("run@100")
+
+    taskFaker.assertNoMoreTasks()
+  }
+
+  @Test fun cancelWhileExecutingDoesNotStopUncancelableTask() {
```
swankjesse: FYI @yschimke this is new since you approved! I added it because clarifying the docs reminded me of a case I’d missed.

okhttp/src/main/java/okhttp3/internal/concurrent/TaskRunner.kt
yschimke: It’s a view IIUC

swankjesse: Creative the view itself is the allocation. I'm really trying to limit allocations here!

JakeWharton: It is a view but also its iterator.

PR #5518, opened 2019-10-02
Default to jdk9 platform for tests

yschimke:
Since we build with jdk11 etc, default to jdk9 platform for tests unless overridden.


swankjesse:
For those following at home, read “jdk9” as “jdk9 and newer”

Issue #5519, opened 2019-10-02
Too many NullProxySelector instances break connection pooling


swankjesse:
We observed clients not pooling connections because although they shared a connection pool, they had different instances of NullProxySelector.

yschimke:
Which version did this start?  :(

swankjesse:
The SecurityManager managed to cause this! It's the worst.
https://github.com/square/okhttp/pull/5332

yschimke:
My bug.  ultimately NullProxySelector could have been a singleton, it is now.  Or had equality defined, it essentially does not via ==.  

Thanks!

yschimke:
The meta comment is we should be able to detect these regressions through stats, so hopefully with load balancing changes we would notice.

PR #5520, opened 2019-10-02
Fix connection pooling for different clients with the same pool.

swankjesse:
Closes: https://github.com/square/okhttp/issues/5519

### DIFF
--- a/okhttp/src/main/java/okhttp3/OkHttpClient.kt
+++ b/okhttp/src/main/java/okhttp3/OkHttpClient.kt
@@ -163,9 +163,9 @@ open class OkHttpClient internal constructor(
 
   @get:JvmName("proxySelector") val proxySelector: ProxySelector =
       when {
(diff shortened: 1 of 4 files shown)


okhttp/src/main/java/okhttp3/OkHttpClient.kt
```diff
@@ -163,9 +163,9 @@ open class OkHttpClient internal constructor(
 
   @get:JvmName("proxySelector") val proxySelector: ProxySelector =
       when {
-        // Avoid possible SecurityException from ProxySelector.getDefault
-        builder.proxy != null -> NullProxySelector()
-        else -> builder.proxySelector ?: ProxySelector.getDefault() ?: NullProxySelector()
```
swankjesse: these NullProxySelector instances weren’t equal to each other!

yschimke:
LGTM. On mobile.

swankjesse:
All good. SecurityManager is bad and we should discourage it.

PR #5526, opened 2019-10-03
Name tasks only in TaskRunner

swankjesse:
Naming queues seemed good initially, but the names are mostly
mutually-redundant with task names.

This PR reduces debug information when a queue is still busy
when the test completes. I have a fix for that in a follow-up
change.

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/concurrent/TaskQueue.kt
+++ b/okhttp/src/main/java/okhttp3/internal/concurrent/TaskQueue.kt
@@ -26,13 +26,7 @@ import java.util.concurrent.TimeUnit
  * for its work; in practice a set of queues may share a set of threads to save resources.
  */
 class TaskQueue internal constructor(
-  private val taskRunner: TaskRunner,
-
-  /**
-   * An application-level object like a connection pool or HTTP call that this queue works on behalf
-   * of. This is intended to be useful for testing and debugging only.
-   */
-  val owner: Any
+  private val taskRunner: TaskRunner
 ) {
   private var shutdown = false
 
--- a/okhttp/src/main/java/okhttp3/internal/ws/RealWebSocket.kt
+++ b/okhttp/src/main/java/okhttp3/internal/ws/RealWebSocket.kt
@@ -76,7 +76,7 @@ class RealWebSocket(
   /** Used for writes, pings, and close timeouts. */
   private var taskQueue = taskRunner.newQueue()
 
-  /** Names this this web socket for observability and debugging. */
+  /** Names this web socket for observability and debugging. */
   private var name: String? = null
 
   /**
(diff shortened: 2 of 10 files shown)


okhttp/src/main/java/okhttp3/internal/concurrent/TaskQueue.kt
```diff
@@ -26,13 +26,7 @@ import java.util.concurrent.TimeUnit
  * for its work; in practice a set of queues may share a set of threads to save resources.
  */
 class TaskQueue internal constructor(
-  private val taskRunner: TaskRunner,
-
-  /**
-   * An application-level object like a connection pool or HTTP call that this queue works on behalf
-   * of. This is intended to be useful for testing and debugging only.
-   */
-  val owner: Any
```
yschimke: I could see us adding this back for testing, debugging or monitoring purposes (as previously documented).  Maybe be useful to know which queue is active.  Happy to drop for now, it's easier to add than remove.

okhttp/src/main/java/okhttp3/internal/ws/RealWebSocket.kt
```diff
@@ -74,7 +74,10 @@ class RealWebSocket(
   private var writer: WebSocketWriter? = null
 
   /** Used for writes, pings, and close timeouts. */
-  private var taskQueue = taskRunner.newQueue("OkHttp WebSocket ${originalRequest.url.redact()}")
+  private var taskQueue = taskRunner.newQueue()
+
+  /** Names this this web socket for observability and debugging. */
```
yschimke:
```suggestion
  /** Names this web socket for observability and debugging. */
```

okhttp/src/main/java/okhttp3/internal/ws/RealWebSocket.kt
```diff
@@ -63,7 +63,7 @@ class RealWebSocket(
   private var call: Call? = null
 
   /** This task processes the outgoing queues. Call [runWriter] to after enqueueing. */
-  private val writerTask: Task
+  private var writerTask: Task? = null
```
yschimke: lateinit? or worried about allocation costs?

okhttp/src/main/java/okhttp3/internal/concurrent/TaskQueue.kt
swankjesse: My hypothesis is that the task info will imply the queue info

okhttp/src/main/java/okhttp3/internal/ws/RealWebSocket.kt
swankjesse: actually I believe lateinit is free! This is mostly for internal consistency with the code around it, that uses nullable where lateinit is probably an upgrade.

Lemme save for follow-up unless somebody else beats me to it

Issue #5388, opened 2019-08-21
Canceling a call may send the cancel frame too late


swankjesse:
I’m calling `Call.cancel()` followed by `requestBody.close()`. The outbound frames have the close() preceding the cancel(), which causes the receiving end to interpret it as a normal close and not an abnormal one.

```
Aug 21, 2019 2:37:12 PM okhttp3.internal.http2.Http2Reader nextFrame
FINE: << 0x00000003    14 DATA          
Aug 21, 2019 2:37:12 PM okhttp3.internal.http2.Http2Writer frameHeader
FINE: >> 0x00000003     0 DATA          END_STREAM
Aug 21, 2019 2:37:12 PM okhttp3.internal.http2.Http2Writer frameHeader
FINE: >> 0x00000003     4 RST_STREAM    
```

When canceling a stream the recipient should be notified that it’s an abnormal close by sending the RST_STREAM frame first.

PR #5528, opened 2019-10-04
Add a way for MockWebServer to track truncated requests

swankjesse:
It now tracks inbound requests that fail with an IOException.

Also add a fix for the bug where we'd send 'END OF STREAM' on a stream
that we'd previously canceled, which raced with the 'RST STREAM' and led
to flakiness.

Before we ship the new API in RecordedRequest we should go over the
other cases where inbound HTTP requests fail and make sure they
get reported through this channel.

Closes: https://github.com/square/okhttp/issues/5388

### DIFF
--- a/mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.kt
+++ b/mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.kt
@@ -331,7 +331,8 @@ class MockWebServer : ExternalResource(), Closeable {
    * @return the head of the request queue
    */
   @Throws(InterruptedException::class)
-  fun takeRequest(timeout: Long, unit: TimeUnit): RecordedRequest? = requestQueue.poll(timeout, unit)
+  fun takeRequest(timeout: Long, unit: TimeUnit): RecordedRequest? =
+      requestQueue.poll(timeout, unit)
 
   @JvmName("-deprecated_requestCount")
   @Deprecated(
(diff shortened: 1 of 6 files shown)


mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.kt
```diff
@@ -674,78 +682,80 @@ class MockWebServer : ExternalResource(), Closeable {
...
-        contentLength = header.substring(15).trim().toLong()
-      }
-      if (lowercaseHeader.startsWith("transfer-encoding:") && lowercaseHeader.substring(
-              18).trim() == "chunked") {
-        chunked = true
+    val requestBody = TruncatingBuffer(bodyLimit)
+    val chunkSizes = mutableListOf<Int>()
+    var failure: IOException? = null
+
+    try {
```
swankjesse: I moved more of this into try/catch so MockWebServer would track failed HTTP/1 requests.

swankjesse: (this is definitely an “ignore whitespace” PR

okhttp-testing-support/src/main/java/okhttp3/OkHttpClientTestRule.kt
```diff
@@ -68,8 +68,8 @@ class OkHttpClientTestRule : TestRule {
 
   private fun ensureAllTaskQueuesIdle() {
     for (queue in TaskRunner.INSTANCE.activeQueues()) {
-      assertThat(queue.awaitIdle(TimeUnit.MILLISECONDS.toNanos(500L)))
-          .withFailMessage("Queue still active after 500ms")
+      assertThat(queue.awaitIdle(TimeUnit.MILLISECONDS.toNanos(1000L)))
+          .withFailMessage("Queue still active after 1000ms")
```
swankjesse: Also bumped this from 500 ms to 1000 ms to reduce flakiness

PR #5530, opened 2019-10-04
Fix the broken link:

dmitry-timofeev:
There is no longer a markdown doc file on interceptors, 
but the article is available on the project website.

### DIFF
--- a/okhttp-logging-interceptor/README.md
+++ b/okhttp-logging-interceptor/README.md
@@ -42,4 +42,4 @@ implementation("com.squareup.okhttp3:logging-interceptor:4.2.1")
 
 
 
- [1]: ../INTERCEPTORS.md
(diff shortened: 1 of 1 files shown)


okhttp-logging-interceptor/README.md
```diff
@@ -42,4 +42,4 @@ implementation("com.squareup.okhttp3:logging-interceptor:4.2.1")
 
 
 
- [1]: ../INTERCEPTORS.md
+ [1]: https://square.github.io/okhttp/interceptors/
```
monkey-mas: Fixed link looks good to me! I'd change the reference from `[1]` to `[interceptors]` (and the corresponding part in `[OkHttp interceptor][interceptors]`).

monkey-mas:
I think the CI failed with a flakey test?

Issue #5512, opened 2019-09-29
TaskRunner tracking bug


swankjesse:
This issue is intended to help me organize what’s done and needs doing for TaskRunner.

 * [x] [API](https://github.com/square/okhttp/pull/5461)
 * [x] [Initial implementation](https://github.com/square/okhttp/pull/5485)
 * [x] [ConnectionPool](https://github.com/square/okhttp/pull/5494)
 * [x] [DiskLruCache](https://github.com/square/okhttp/pull/5496)
 * [x] [HTTP/2 writer and push](https://github.com/square/okhttp/pull/5502)
 * [x] [WebSockets](https://github.com/square/okhttp/pull/5510)
 * [x] MockWebServer
 * [x] HTTP/2 listeners
 * [x] Daemon threads
 * [x] Fewer context switches ([details](https://github.com/square/okhttp/pull/5510#discussion_r329355091))
 * [x] Consistent queue and task names
 * [x] Strict mode and/or other diagnostics for stuck queues ([details](https://github.com/square/okhttp/pull/5496))
 * [x] Tools for understanding real-world use ([details](https://github.com/square/okhttp/pull/5532#issuecomment-538617152))
 * [x] Re-evaluate how tryCancel() opt-in vs. opt-out
 * [x] [Better task names](https://github.com/square/okhttp/pull/5535)

See also
 * [Simplify test code](https://github.com/square/okhttp/pull/5507)
 * [Detect HTTP/2 liveness](https://github.com/square/okhttp/issues/3146)

swankjesse:
To improve visibility I think I'll borrow from what we do for HTTP/2 frames: use logger.fine() to print events and metrics. If you know it exists you can turn it on, but no public API.

swankjesse:
I wanna build debug logging for tasks. Here’s a sketch of a sample log:

```
Q1 ▌ scheduled after 100 ms: OkHttp example.com ping
Q1 ▌▌ scheduled after 200 ms: OkHttp example.com[3] writeSynReset
Q1 ▌▌ starting: OkHttp example.com ping
Q1 ▌▌ executed in 130 ms (next after 100 ms): OkHttp example.com ping
Q1 ▌▌ starting (31 ms late): OkHttp example.com[3] writeSynReset
Q1 ▌▌ executed in 2 ms: OkHttp example.com[3] writeSynReset
Q2 ▌ posted: OkHttp ConnectionPool
```

Notes:

[...]

PR #5532, opened 2019-10-05
Change TaskRunner to limit context switches.

swankjesse:
Now we don't have to alternate between the coordinator thread and the task
thread between task runs if the task returns 0. Instead the task thread can
stay resident.

This implementation works by having task runnables that can switch from
the coordinator role (sleeping until the next task starts) and the executor
role.

https://github.com/square/okhttp/issues/5512

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/concurrent/Task.kt
+++ b/okhttp/src/main/java/okhttp3/internal/concurrent/Task.kt
@@ -56,8 +56,6 @@ abstract class Task(
   /** Undefined unless this is in [TaskQueue.futureTasks]. */
   internal var nextExecuteNanoTime = -1L
 
-  internal var runRunnable: Runnable? = null
-
   /** Returns the delay in nanoseconds until the next execution, or -1L to not reschedule. */
   abstract fun runOnce(): Long
 
@@ -66,19 +64,5 @@ abstract class Task(
 
     check(this.queue === null) { "task is in multiple queues" }
     this.queue = queue
-
-    this.runRunnable = Runnable {
-      val currentThread = Thread.currentThread()
-      val oldName = currentThread.name
-      currentThread.name = name
-
-      var delayNanos = -1L
-      try {
-        delayNanos = runOnce()
-      } finally {
-        queue.runCompleted(this, delayNanos)
-        currentThread.name = oldName
-      }
-    }
   }
 }
--- a/okhttp/src/main/java/okhttp3/internal/concurrent/TaskQueue.kt
+++ b/okhttp/src/main/java/okhttp3/internal/concurrent/TaskQueue.kt
@@ -28,22 +28,16 @@ import java.util.concurrent.TimeUnit
 class TaskQueue internal constructor(
   private val taskRunner: TaskRunner
 ) {
-  private var shutdown = false
+  internal var shutdown = false
 
   /** This queue's currently-executing task, or null if none is currently executing. */
-  private var activeTask: Task? = null
-
-  /** True if the [activeTask] should not recur when it completes. */
-  private var cancelActiveTask = false
+  internal var activeTask: Task? = null
 
   /** Scheduled tasks ordered by [Task.nextExecuteNanoTime]. */
-  private val futureTasks = mutableListOf<Task>()
+  internal val futureTasks = mutableListOf<Task>()
 
-  internal fun isActive(): Boolean {
-    check(Thread.holdsLock(taskRunner))
-
-    return activeTask != null || futureTasks.isNotEmpty()
-  }
+  /** True if the [activeTask] should be canceled when it completes. */
+  internal var cancelActiveTask = false
 
   /**
    * Returns a snapshot of tasks currently scheduled for execution. Does not include the
@@ -87,7 +81,7 @@ class TaskQueue internal constructor(
   fun awaitIdle(delayNanos: Long): Boolean {
     val latch = CountDownLatch(1)
 
-    val task = object : Task("awaitIdle") {
+    val task = object : Task("awaitIdle", cancelable = false) {
       override fun runOnce(): Long {
         latch.countDown()
         return -1L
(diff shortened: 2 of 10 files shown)


okhttp/src/main/java/okhttp3/internal/concurrent/Task.kt
```diff
@@ -56,8 +56,6 @@ abstract class Task(
   /** Undefined unless this is in [TaskQueue.futureTasks]. */
   internal var nextExecuteNanoTime = -1L
 
-  internal var runRunnable: Runnable? = null
-
   /** Returns the delay in nanoseconds until the next execution, or -1L to not reschedule. */
```
yschimke: nit: Negative values < -1 seem to be effectively a priority, is that ok? Or should be an error?

Not a big concern given this is internal.

okhttp/src/main/java/okhttp3/internal/concurrent/TaskRunner.kt
```diff
@@ -38,117 +37,222 @@ import java.util.concurrent.TimeUnit
...
-   */
-  fun activeQueues(): List<TaskQueue> {
-    synchronized(this) {
-      return activeQueues.toList()
+  private val runnable: Runnable = object : Runnable {
+    override fun run() {
+      while (true) {
+        val task = synchronized(this@TaskRunner) {
+          awaitTaskToRun()
+        } ?: return
```
yschimke: I think my biggest concern is that the delay of the time to execute a solitary task becomes the limiting time to execute a later (but soon) task on another queue.

Or am I missing something?

Do you schedule a kick for next next task time somewhere?

okhttp/src/main/java/okhttp3/internal/concurrent/TaskRunner.kt
```diff
@@ -38,117 +37,222 @@ import java.util.concurrent.TimeUnit
...
+      if (readyQueues.isEmpty()) {
+        return null // Nothing to do.
+      }
+
+      val now = backend.nanoTime()
+      var minDelayNanos = Long.MAX_VALUE
+      var readyTask: Task? = null
+      var multipleReadyTasks = false
+
+      // Decide what to run. This loop's goal wants to:
```
yschimke: Instead of this loop goal, could you store the next two scheduled times, and loop to find the third at this point?

yschimke:
Overall, I'd love to see real world thread usage for this. Any thoughts of how it looks in a profiler? Or whether we could output something like the Google Chrome Trace Event format?

swankjesse:
Before & after effects of fewer context switches. These are measurements for the cost of 100 sequential empty tasks on the same queue.

### BEFORE
```
Benchmark                                          (executionCount)    Mode    Cnt     Score    Error  Units
SelectBenchmark.executeTasks                                    100  sample  10730   930.829 ± 10.726  us/op
SelectBenchmark.executeTasks:executeTasks·p0.00                 100  sample          655.360           us/op
SelectBenchmark.executeTasks:executeTasks·p0.50                 100  sample          761.856           us/op
SelectBenchmark.executeTasks:executeTasks·p0.90                 100  sample         1476.608           us/op
SelectBenchmark.executeTasks:executeTasks·p0.95                 100  sample         1673.216           us/op
SelectBenchmark.executeTasks:executeTasks·p0.99                 100  sample         2012.549           us/op
SelectBenchmark.executeTasks:executeTasks·p0.999                100  sample         3156.435           us/op
SelectBenchmark.executeTasks:executeTasks·p0.9999               100  sample         3675.513           us/op
SelectBenchmark.executeTasks:executeTasks·p1.00                 100  sample         3678.208           us/op
```

### AFTER
```
Benchmark                                          (executionCount)    Mode    Cnt     Score   Error  Units
SelectBenchmark.executeTasks                                    100  sample  24266   412.118 ± 1.106  us/op
SelectBenchmark.executeTasks:executeTasks·p0.00                 100  sample          375.808          us/op
SelectBenchmark.executeTasks:executeTasks·p0.50                 100  sample          394.752          us/op
SelectBenchmark.executeTasks:executeTasks·p0.90                 100  sample          445.952          us/op
SelectBenchmark.executeTasks:executeTasks·p0.95                 100  sample          488.960          us/op
SelectBenchmark.executeTasks:executeTasks·p0.99                 100  sample          635.218          us/op
SelectBenchmark.executeTasks:executeTasks·p0.999                100  sample          959.896          us/op
SelectBenchmark.executeTasks:executeTasks·p0.9999               100  sample         1423.714          us/op
SelectBenchmark.executeTasks:executeTasks·p1.00                 100  sample         1501.184          us/op
```

okhttp/src/main/java/okhttp3/internal/concurrent/Task.kt
swankjesse: Intended contract is for implementations to never return such values! If ever we wanted to make this more broadly usable we’d need to enforce that requirement.

okhttp/src/main/java/okhttp3/internal/concurrent/TaskRunner.kt
swankjesse: Yeah, awaitTaskToRun might be misnamed. It’ll always schedule a thread if there’s something to do now (run or sleep) that competes with something to do later (run or sleep).

okhttp/src/main/java/okhttp3/internal/concurrent/TaskRunner.kt
swankjesse: Maybe the question I should answer is, why two items?

The first one is what we’re going to execute on the current thread.
The second one justifies starting a new thread that’ll itself look for up to two items.

We stop after two because once we know we’re going to fork a thread, it becomes _that_ thread’s problem to fork a thread for the third thing.

swankjesse:
Figuring out how to mature this up is the next challenge. The best way to do that is some real world exercise. I’m going to run our crawler a bunch, and maybe write some torture tests.

okhttp/src/main/java/okhttp3/internal/concurrent/TaskQueue.kt
```diff
@@ -87,7 +81,7 @@ class TaskQueue internal constructor(
   fun awaitIdle(delayNanos: Long): Boolean {
     val latch = CountDownLatch(1)
 
-    val task = object : Task("awaitIdle") {
+    val task = object : Task("awaitIdle", cancelable = false) {
```
swankjesse: this was the problem with flaky tests!

okhttp/src/test/java/okhttp3/CallKotlinTest.kt
```diff
@@ -134,6 +134,9 @@ class CallKotlinTest {
     recordedRequest = server.takeRequest()
     assertEquals("HEAD", recordedRequest.method)
 
+    recordedRequest = server.takeRequest()
+    assertThat(recordedRequest.failure).isNotNull()
```
swankjesse: this fixes a green/green collision with two earlier pull requests

PR #5535, opened 2019-10-06
Be more consistent with task names

swankjesse:
Examples:
  - OkHttp TaskRunner
  - MockWebServer TaskRunner
  - OkHttp ConnectionPool
  - MockWebServer localhost applyAndAckSettings
  - OkHttp android.com applyAndAckSettings
  - OkHttp android.com onSettings
  - OkHttp awaitIdle
  - OkHttp localhost

I'm trying to use type names where appropriate, or method names otherwise.
Names include hostname and stream name if the task is working on behalf
of a specific stream or connection.

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/concurrent/Task.kt
+++ b/okhttp/src/main/java/okhttp3/internal/concurrent/Task.kt
@@ -48,7 +48,7 @@ package okhttp3.internal.concurrent
  */
 abstract class Task(
   val name: String,
-  val cancelable: Boolean = true
+  val cancelable: Boolean
 ) {
   // Guarded by the TaskRunner.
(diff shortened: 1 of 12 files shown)


okhttp/src/main/java/okhttp3/internal/concurrent/Task.kt
```diff
@@ -48,7 +48,7 @@ package okhttp3.internal.concurrent
  */
 abstract class Task(
   val name: String,
-  val cancelable: Boolean = true
```
swankjesse: I moved the default value to TaskQueue.schedule(). (I don’t like repeating the defaults; I find that hard to reason about.)

PR #5538, opened 2019-10-06
Use TaskRunner in MockWebServer

swankjesse:
The utility here is negligible. We're sharing threads between HTTP,
HTTP/2 and web sockets now.


yschimke:
LGTM

PR #5539, opened 2019-10-06
Change log for OkHttp 4.2.2.

swankjesse:


swankjesse:
This bug doesn‘t _really_ warrant a patch release, but we want it for Wire 3.0.

PR #5541, opened 2019-10-08
Add link to okio proguard rules

StefMa:
Always when I search for the proguard rules of my app I jump from dependency to dependency.
okhttp makes it a little bit easier by saying "You also need the Okio rules".
But just saying it means I have to open a new tab, go to github.com/square/okio (or google it), and copy the stuff to my proguard files.

With that link we simplify the process 😇

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,14 +106,6 @@ implementation("com.squareup.okhttp3:okhttp:4.2.1")
 Snapshot builds are [available][snap].
 
 
-R8 / ProGuard
--------------
-
-If you are using R8 or ProGuard add the options from [`okhttp3.pro`][okhttp3_pro].
-
-You might also need [rules for Okio][okio proguard] which is a dependency of this library.
-
-
 MockWebServer
 -------------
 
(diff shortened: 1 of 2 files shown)


JakeWharton:
The rules are embedded now so there's no action to take on the consumer side. We should probably just delete this section.

StefMa:
Is this true for proguard as well? 🤔

JakeWharton:
Yes, provided you're using the Android Gradle plugin to run it.

StefMa:
But maybe okhttp isn't used in an Android project but I'm a pure kotlin
project which uses proguard...
Then he needs to manually add these rules, right?

JakeWharton:
Yes. But in that case I'm okay with the rules being a little harder to
find. The overwhelmingly common case is consuming and shrinking through AGP
and in that case copying rules is actually the worst outcome since they
inevitably become stale.

JakeWharton:
A page on the website sounds okay. I don't think it's important enough to
put directly on the homepage though.

StefMa:
✅ I moved it into the docs 👍 
See https://github.com/square/okhttp/pull/5541/commits/2260512dcf47cc294b493336fec753df77086ea4

PR #5550, opened 2019-10-14
Add missing HTTP/2 Error Code (SETTINGS_TIMEOUT, STREAM_CLOSED, FRAME_SIZE_ERROR)

monkey-mas:
ref https://tools.ietf.org/html/rfc7540#section-11.4

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/http2/ErrorCode.kt
+++ b/okhttp/src/main/java/okhttp3/internal/http2/ErrorCode.kt
@@ -26,6 +26,12 @@ enum class ErrorCode constructor(val httpCode: Int) {
 
   FLOW_CONTROL_ERROR(3),
 
+  SETTINGS_TIMEOUT(4),
+
+  STREAM_CLOSED(5),
+
+  FRAME_SIZE_ERROR(6),
+
   REFUSED_STREAM(7),
 
   CANCEL(8),


okhttp/src/main/java/okhttp3/internal/http2/ErrorCode.kt
```diff
@@ -26,6 +26,12 @@ enum class ErrorCode constructor(val httpCode: Int) {
 
   FLOW_CONTROL_ERROR(3),
 
+  SETTINGS_TIMEOUT(4),
```
monkey-mas: We could use hexadecimal, eg `0x4`, to be consistent with other error code(`CONNECT_ERROR(0xa)`)?

okhttp/src/main/java/okhttp3/internal/http2/ErrorCode.kt
```diff
@@ -26,6 +26,12 @@ enum class ErrorCode constructor(val httpCode: Int) {
...
+
+  STREAM_CLOSED(5),
+
+  FRAME_SIZE_ERROR(6),
```
monkey-mas: Probably we might add a short description like RFC says?
```kotlin
/* Frame size incorrect */
FRAME_SIZE_ERROR(6)
```

PR #5551, opened 2019-10-14
Rename ErrorCode val from "httpCode" to "code"

monkey-mas:
Like [11.4.  Error Code Registry](https://tools.ietf.org/html/rfc7540#section-11.4) mentions, the error code value is `Code`.

```
New registrations are advised to provide the following information:

Name:  A name for the error code.  Specifying an error code name is
      optional.

Code:  The 32-bit error code value.
```

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/http2/ErrorCode.kt
+++ b/okhttp/src/main/java/okhttp3/internal/http2/ErrorCode.kt
@@ -16,7 +16,7 @@
 package okhttp3.internal.http2
 
 /** http://tools.ietf.org/html/draft-ietf-httpbis-http2-17#section-7 */
-enum class ErrorCode constructor(val httpCode: Int) {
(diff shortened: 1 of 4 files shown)


monkey-mas:
As long as the change is ok, we can safely merge it since all the code are from internal API, ie `okhttp/src/main/java/okhttp3/internal`. Thus OkHttp users shouldn't have touched them.

yschimke:
I'd be just as likely to accept a change from code to httpCode, so I don't really see the point of this PR. I'm tempted to close it, because it might just break a (silly) few users using internal API.

monkey-mas:
I'd say http code can imply HTTP status code, but if people can accept the current naming it's ok to close the PR.  I'd admit that migration steps are annoying. (I guess we need to write migration code/annotation like we did for OkHttp4.)

PR #5552, opened 2019-10-14
Fix test in CacheTest by using t2 timestamp as expected

monkey-mas:
We have a warning as follows, and we should've used `t2` timestamp.

```
okhttp/okhttp/src/test/java/okhttp3/CacheTest.java:1939: warning: [UnusedVariable] The local variable 't2' is never read.
    long t2 = System.currentTimeMillis();
         ^
    (see https://errorprone.info/bugpattern/UnusedVariable)
  Did you mean to remove this line or 'System.currentTimeMillis();'?
```

### DIFF
--- a/okhttp/src/test/java/okhttp3/CacheTest.java
+++ b/okhttp/src/test/java/okhttp3/CacheTest.java
@@ -1931,7 +1931,8 @@ public final class CacheTest {
     assertThat(response2.code()).isEqualTo(HttpURLConnection.HTTP_OK);
     assertThat(response2.body().string()).isEqualTo("A");
     assertThat(response2.header("Allow")).isEqualTo("GET, HEAD");
-    assertThat((double) (response2.receivedResponseAtMillis() - t1)).isCloseTo(
+    Long originalTimestamp = response2.receivedResponseAtMillis();
+    assertThat((double) (originalTimestamp - t1)).isCloseTo(
         (double) 0, offset(250.0));
 
     // A full cache hit reads the cache.
@@ -1940,8 +1941,7 @@ public final class CacheTest {
     Response response3 = get(server.url("/a"));
     assertThat(response3.body().string()).isEqualTo("A");
     assertThat(response3.header("Allow")).isEqualTo("GET, HEAD");
-    assertThat((double) (response3.receivedResponseAtMillis() - t1)).isCloseTo(
-        (double) 0, offset(250.0));
+    assertThat(response3.receivedResponseAtMillis()).isEqualTo(originalTimestamp);
 
     assertThat(server.getRequestCount()).isEqualTo(2);
   }


monkey-mas:
Ok, we can either a) remove the `t2` since we don't have to use the val definition or b) use `t2` and increase the _offset_ by "500ms" (due to a sleep before making a request for response3).

yschimke:
> Ok, we can either a) remove the `t2` since we don't have to use the val definition or b) use `t2` and increase the _"expected time"_ by 500ms (due to a sleep before making a request for response3).

Based on your comment, isn't dropping t2 the better solution?

monkey-mas:
I stand corrected. Probably this is what we should simply do...? IIUC, what we need to check are a) the last request is made after some sleep time(eg 500ms) and then b) the response has the original/cached timestamp. WDYT?

```diff
--- a/okhttp/src/test/java/okhttp3/CacheTest.java
+++ b/okhttp/src/test/java/okhttp3/CacheTest.java
@@ -1931,19 +1931,19 @@ public final class CacheTest {
     assertThat(response2.code()).isEqualTo(HttpURLConnection.HTTP_OK);
     assertThat(response2.body().string()).isEqualTo("A");
     assertThat(response2.header("Allow")).isEqualTo("GET, HEAD");
-    assertThat((double) (response2.receivedResponseAtMillis() - t1)).isCloseTo(
+    Long originalTimestamp = response2.receivedResponseAtMillis();
+    assertThat((double) (originalTimestamp - t1)).isCloseTo(
         (double) 0, offset(250.0));

     // A full cache hit reads the cache.
     Thread.sleep(500); // Make sure t1 and t2 are distinct.
     long t2 = System.currentTimeMillis();
+    assertThat((double) (t2 - t1)).isCloseTo((double) 500, offset(250.0));
+
     Response response3 = get(server.url("/a"));
     assertThat(response3.body().string()).isEqualTo("A");
     assertThat(response3.header("Allow")).isEqualTo("GET, HEAD");
-    // Since the received timestamp for response3 is the original timestamp from response2,
-    // we expect the difference from t2 to be greater than the sleep time(> 500ms).
-    assertThat((double) (t2 - response3.receivedResponseAtMillis())).isCloseTo(
-        (double) 500, offset(250.0));
+    assertThat(response3.receivedResponseAtMillis()).isEqualTo(originalTimestamp);

     assertThat(server.getRequestCount()).isEqualTo(2);
   }```

File: okhttp/src/test/java/okhttp3/CacheTest.java [hunk 1]
```diff
@@ -1940,8 +1940,10 @@ private RecordedRequest assertClientSuppliedCondition(MockResponse seed, String
     Response response3 = get(server.url("/a"));
     assertThat(response3.body().string()).isEqualTo("A");
     assertThat(response3.header("Allow")).isEqualTo("GET, HEAD");
-    assertThat((double) (response3.receivedResponseAtMillis() - t1)).isCloseTo(
-        (double) 0, offset(250.0));
+    // Since the received timestamp for response3 is the original timestamp from response2,
+    // we expect the difference from t2 to be greater than the sleep time(> 500ms).
+    assertThat((double) (t2 - response3.receivedResponseAtMillis())).isCloseTo(
```
yschimke: The comment doesn't match the logic.  Should the check be > 500ms?  This is 250 to 750.

monkey-mas: I'd like to remove the comment and change the test as I mentioned. (Let me add a new commit)
https://github.com/square/okhttp/pull/5552#issuecomment-542550095

Strictly specking, I should've done like this like you said?
```java
    assertThat(t2 - response3.receivedResponseAtMillis()).isGreaterThan(500);
    assertThat(t2 - response3.receivedResponseAtMillis()).isLessThan(750);
```

okhttp/src/test/java/okhttp3/CacheTest.java
```diff
@@ -1931,17 +1931,19 @@ private RecordedRequest assertClientSuppliedCondition(MockResponse seed, String
...
     assertThat(response2.header("Allow")).isEqualTo("GET, HEAD");
-    assertThat((double) (response2.receivedResponseAtMillis() - t1)).isCloseTo(
+    Long originalTimestamp = response2.receivedResponseAtMillis();
+    assertThat((double) (originalTimestamp - t1)).isCloseTo(
         (double) 0, offset(250.0));
 
     // A full cache hit reads the cache.
     Thread.sleep(500); // Make sure t1 and t2 are distinct.
     long t2 = System.currentTimeMillis();
+    assertThat((double) (t2 - t1)).isCloseTo((double) 500, offset(250.0));
```
yschimke: remove this, it's just testing variables from the test.

monkey-mas:
@yschimke Thanks! Removed the unnecessary assertion and rebased the commit!

PR #5559, opened 2019-10-18
Remove unnecessary variable in CacheTest

monkey-mas:
Like we discussed it in the previous PR, `t2` should be removed.
PR: https://github.com/square/okhttp/pull/5552

- Renamed `originalTimestamp` to `updatedTimestamp` since the name was not appropriate.
- Change the thread sleep time to 10ms.

### DIFF
--- a/okhttp/src/test/java/okhttp3/CacheTest.java
+++ b/okhttp/src/test/java/okhttp3/CacheTest.java
@@ -1931,17 +1931,16 @@ public final class CacheTest {
     assertThat(response2.code()).isEqualTo(HttpURLConnection.HTTP_OK);
     assertThat(response2.body().string()).isEqualTo("A");
     assertThat(response2.header("Allow")).isEqualTo("GET, HEAD");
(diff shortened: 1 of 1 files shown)


monkey-mas:
@yschimke My bad, sorry. PTAL.

okhttp/src/test/java/okhttp3/CacheTest.java
```diff
@@ -1936,8 +1936,7 @@ private RecordedRequest assertClientSuppliedCondition(MockResponse seed, String
         (double) 0, offset(250.0));
 
     // A full cache hit reads the cache.
-    Thread.sleep(500); // Make sure t1 and t2 are distinct.
-    long t2 = System.currentTimeMillis();
+    Thread.sleep(500);
```
yschimke: Can you drop this to 10?  And rename originalTimestamp to updateTimestamp.

monkey-mas: done! :)

PR #5565, opened 2019-10-22
Adds Maven coordinates to README

emilianbold:
The README still contains Java examples, but is missing what most Java projects would use: Maven coordinates. This PR adds that.

Note that something about the README (or the lack of an actual pom.xml file perhaps) confuses GitHub's Dependency graph page. Here https://github.com/nielsgron/quicksilver/network/dependencies it detects for com.squareup.okhttp3:okhttp the repository https://github.com/lizhangqu/PriorityOkHttp . It would be good to see with GitHub how to make this the canonical repository for OkHttp on Maven.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,17 +99,7 @@ Releases
 
 Our [change log][changelog] has release history.
 
-```kotlin
-implementation("com.squareup.okhttp3:okhttp:4.2.1")
-```
-
-```xml
-<dependency>
-  <groupId>com.squareup.okhttp3</groupId>
-  <artifactId>okhttp</artifactId>
-  <version>4.2.1</version>
-</dependency>
-```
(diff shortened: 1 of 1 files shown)


emilianbold:
PS: I contacted GitHub and they manually linked the com.squareup.okhttp3:okhttp dependency as shown with their Dependency graph page with this repository.

JakeWharton:
Java projects use Gradle. I think I would rather just show the coordinates and let the user figure out what to do for their build system. This is what we did for Retrofit: https://github.com/square/retrofit#download

emilianbold:
I've updates the PR. Just mentioning Maven and the coordinates instead of the Kotlin/Gradle line would also do.

emilianbold:
OK, I'm linking to Maven Central now and mention the 4.2.2 release in the README (there was 4.2.1 before).

Issue #5576, opened 2019-10-27
Failing CI test with JDK8: failingCallsDoNotInterfereWithConnection


yschimke:
https://app.circleci.com/jobs/github/square/okhttp/6685

```
okhttp3.CallTest > failingCallsDoNotInterfereWithConnection FAILED
    org.junit.ComparisonFailure: expected:<"Response [1]"> but was:<"Response [2]">
        at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
        at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
        at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
        at okhttp3.CallTest.failingCallsDoNotInterfereWithConnection(CallTest.java:2991)
```

yschimke:
Output from CI for successful test run

```
Events (29)
[0 ms] callStart: Request{method=POST, url=https://localhost:43321/}
[0 ms] proxySelectStart: https://localhost:43321/
[0 ms] proxySelectEnd: [DIRECT]
[0 ms] dnsStart: localhost
[0 ms] dnsEnd: [localhost/127.0.0.1]
[0 ms] connectStart: localhost/127.0.0.1:43321 DIRECT
[0 ms] secureConnectStart
[50 ms] secureConnectEnd: Handshake{tlsVersion=TLS_1_3 cipherSuite=TLS_AES_128_GCM_SHA256 peerCertificates=[CN=localhost] localCertificates=[]}
[51 ms] connectEnd: h2
[51 ms] connectionAcquired: Connection{localhost:43321, proxy=DIRECT hostAddress=localhost/127.0.0.1:43321 cipherSuite=TLS_AES_128_GCM_SHA256 protocol=h2}
[51 ms] requestHeadersStart
[51 ms] requestHeadersEnd
[51 ms] requestBodyStart
[0 ms] callStart: Request{method=POST, url=https://localhost:43321/}
[0 ms] proxySelectStart: https://localhost:43321/
[0 ms] proxySelectEnd: [DIRECT]
[0 ms] connectionAcquired: Connection{localhost:43321, proxy=DIRECT hostAddress=localhost/127.0.0.1:43321 cipherSuite=TLS_AES_128_GCM_SHA256 protocol=h2}
[0 ms] requestHeadersStart
[0 ms] requestHeadersEnd
[...]

PR #5577, opened 2019-10-27
Flaky test improvements

yschimke:
Keep the client event logging for flaky tests even when another listener is applied.

Used to generate a baseline for https://github.com/square/okhttp/issues/5576

### DIFF
--- a/okhttp/src/test/java/okhttp3/CallTest.java
+++ b/okhttp/src/test/java/okhttp3/CallTest.java
@@ -72,6 +72,7 @@ import okhttp3.mockwebserver.MockWebServer;
 import okhttp3.mockwebserver.QueueDispatcher;
 import okhttp3.mockwebserver.RecordedRequest;
 import okhttp3.mockwebserver.SocketPolicy;
+import okhttp3.testing.Flaky;
 import okhttp3.testing.PlatformRule;
 import okhttp3.tls.HandshakeCertificates;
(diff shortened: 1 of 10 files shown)


okhttp/src/test/java/okhttp3/CallTest.java
```diff
@@ -112,7 +113,7 @@
   private RecordingEventListener listener = new RecordingEventListener();
   private HandshakeCertificates handshakeCertificates = localhost();
   private OkHttpClient client = clientTestRule.newClientBuilder()
-      .eventListener(listener)
+      .eventListenerFactory(clientTestRule.wrap(listener))
```
yschimke: This could be simplified by putting the eventListener call on ClientTestRule, before getting the builder or client.

okhttp/src/test/java/okhttp3/EventListenerTest.java
```diff
@@ -71,6 +72,7 @@
 import static org.junit.Assert.fail;
 import static org.junit.Assume.assumeThat;
 
+@Flaky // STDOUT logging enabled for test
```
yschimke: This just makes sense while making any changes to the EventListenerTest.  Is there anywhere this creates too much noise?

swankjesse: Doesn’t bother me.

PR #5582, opened 2019-11-02
Run Jdk8 on CI and fix test expectations

yschimke:
Test doesn't need to care about which response, timing related issues are secondary.  Just that it succeeds and we don't somehow disconnect and reconnect.


yschimke:
@swankjesse I think testjdk8 and testjdk11 on commit is probably optimal for coverage (worst to best)

yschimke:
@swankjesse ping - simple one I hope

PR #5584, opened 2019-11-03
Testing fix/workaround for missingPongsFailsConnection with JDK12

yschimke:


okhttp/src/test/java/okhttp3/internal/http2/HttpOverHttp2Test.java
```diff
@@ -112,12 +112,12 @@
     return asList(Protocol.H2_PRIOR_KNOWLEDGE, Protocol.HTTP_2);
   }
 
-  private PlatformRule platform = new PlatformRule();
+  private final PlatformRule platform = new PlatformRule();
+  private final OkHttpClientTestRule clientTestRule = new OkHttpClientTestRule();
```
swankjesse: should we combine the two rule classes? Might make things simpler from callsites?

yschimke: Yep.  PlatformRule has a terrible name, so let's kill that and just make it a feature of OkHttpClientTestRule

Issue #5587, opened 2019-11-06
MockWebserver crashing with ssl negotiation error


nimeacuerdo:
Hi,

We are still not sure about the root cause but our acceptance tests execution phase is failing sometimes due to `MockWebserver` (**4.2.1**) crashing when the following error happens in `acceptConnections()`:

```
11-05 11:53:36.557  5607 23163 E AndroidRuntime: FATAL EXCEPTION: MockWebServer
11-05 11:53:36.557  5607 23163 E AndroidRuntime: Process: com.xxx.xxx.xxx, PID: 5607
11-05 11:53:36.557  5607 23163 E AndroidRuntime: java.lang.AssertionError: java.lang.reflect.InvocationTargetException
11-05 11:53:36.557  5607 23163 E AndroidRuntime: 	at okhttp3.internal.platform.android.AndroidSocketAdapter.a(SourceFile:86)
11-05 11:53:36.557  5607 23163 E AndroidRuntime: 	at okhttp3.internal.platform.AndroidPlatform.b(SourceFile:86)
11-05 11:53:36.557  5607 23163 E AndroidRuntime: 	at okhttp3.mockwebserver.MockWebServer$SocketHandler.handle(MockWebServer.kt:516)
11-05 11:53:36.557  5607 23163 E AndroidRuntime: 	at okhttp3.mockwebserver.MockWebServer$serveConnection$$inlined$execute$1.run(Util.kt:580)
11-05 11:53:36.557  5607 23163 E AndroidRuntime: 	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
11-05 11:53:36.557  5607 23163 E AndroidRuntime: 	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
11-05 11:53:36.557  5607 23163 E AndroidRuntime: 	at java.lang.Thread.run(Thread.java:764)
11-05 11:53:36.557  5607 23163 E AndroidRuntime: Caused by: java.lang.reflect.InvocationTargetException
11-05 11:53:36.557  5607 23163 E AndroidRuntime: 	at java.lang.reflect.Method.invoke(Native Method)
11-05 11:53:36.557  5607 23163 E AndroidRuntime: 	at okhttp3.internal.platform.android.AndroidSocketAdapter.a(SourceFile:81)
11-05 11:53:36.557  5607 23163 E AndroidRuntime: 	... 6 more
[...]

PR #5590, opened 2019-11-09
Workaround AOSP conscrypt NPE

yschimke:
https://github.com/square/okhttp/issues/5587
and
https://github.com/google/conscrypt/issues/781


okhttp/src/main/java/okhttp3/internal/platform/android/AndroidSocketAdapter.kt
```diff
@@ -74,6 +74,12 @@ open class AndroidSocketAdapter(private val sslSocketClass: Class<in SSLSocket>)
     return try {
       val alpnResult = getAlpnSelectedProtocol.invoke(sslSocket) as ByteArray?
       if (alpnResult != null) String(alpnResult, StandardCharsets.UTF_8) else null
+    } catch (e: NullPointerException) {
+        when {
+            // https://github.com/square/okhttp/issues/5587
+            e.message == "ssl == null" -> null
+            else -> throw AssertionError(e)
```
swankjesse: else -> throw e

Issue #5599, opened 2019-11-13
Allow limiting response size


KAbakumovAtl:
Currently there is no out of the box way to set hard limit on the response payload size. 

At the same time, large responses might cause out of memory error. Unless, we reject them before loading to memory. 

So it should be possible to drop huge response if one of the following happens:
a) Content-Length header with a value exceeding the threshold set in the resposne
b) While reading response from the network, we figured out that amount of bytes we received so far is higher than the threshold

swankjesse:
Really tough for us to do this well. Policy might be fail, throttle, truncate, or something else. Perhaps even backpressure.

I think it’s a good thing to do I just don’t see a simple way for us to express this as an API in OkHttp. Instead I think it fits better in a layer above like Retrofit.

KAbakumovAtl:
Problem of doing it in retrofit, is that we don't want huge response to be read into memory. Hence retrofit working on higher level, it might be able to reject response only after it was read (Maybe there is still a way to reject response before reading it though). Doing it on OkHTTP level - will help to reject response as early as possible.

However I got your concern. I am planning to go with suggested approach and create own repository with this interceptor.

MathianCode:
Hi, i have a clarification here... instead of having the check during the interceptor, would it also do fine if i have it after the call of httpClient.newCall(request.build()).execute() ??, if not i would kindly like to know the drawback of it.

**Here is the Code that i tried for limitting a response's body size to the provided threshold(responseBodyThreshold) in bytes.**
```
public Class HttpResponse {

    public HttpResponse(Response response, Long responseBodyThreshold) throws IOException {
        if (response.body() != null) {
            Source source = response.body().source();

            try (
                    response
                    Buffer buffer = new Buffer();
                    ByteArrayOutputStream str = new ByteArrayOutputStream()
            ) {
                for (long readCount; (readCount = source.read(buffer, 0x2000L)) != -1L; totalBytesRead += readCount) {
                    if (totalBytesRead > responseBodyThreshold) {
                        throw new IOException(
                                String.format(
                                        "Exceeded Response Body Threshold Size(%d Bytes)",
                                        responseBodyThreshold
                                )
                        );
                    }
                    buffer.writeTo(str);
                }

                this.responseCode = response.code();
                this.responseMessage = response.message();
                this.body = str.toString();
            }
        }
    }

}
```

PR #5600, opened 2019-11-14
Interceptor for huge responses rejection (#5599)

KAbakumov:
This pull request adds a module with size limiting interceptor and some instructions on how to use it


swankjesse:
Please post this as a standalone GitHub project! I don't want policy-heavy features like this in the core project because developers end up requesting features instead of just customizing to their own needs.

When it's posted I'd love to link it from the docs page of complementary projects.

yschimke:
+1 This could also be useful for users pinned on 3.12 and 3.14 as well.

PR #5601, opened 2019-11-16
Bump Travis to JDK 11

yschimke:
- Build using the minimum suitable LTS (JDK11)
- Avoid running tests because Travis has old versions of JDK11 that have SSL bugs

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ language: java
 dist: trusty
 
 jdk:
-  - oraclejdk8
+  - oraclejdk11
 
 # avoid ./gradlew assemble default which builds docs
 install:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -21,8 +21,10 @@ env:
     - secure: "WMkcWrsvzJNf48w7DJwipUNbhAoggCkC+NM31esq9/GDceGtVWj4hssQETynG4+ckxr0wGqUxsTRTz0uGhX6Fi58haG8yKp+g/HVClqI5EYjI44ptPcwlqlbYjuGbk65k1OGGZLctA6fQA3uT0zee05/yBjJx/jOqrN+PD1tW38="
 
 branches:
-  only:
-    - master
+  except:
```
yschimke: To revert

Issue #907, opened 2014-06-10
HTTP/3 support


mariotaku:
does okhttp have any plan on [QUIC](http://www.chromium.org/quic) support?

brunobowden:
Any update on QUIC support in okhttp? Google just announced that half of Chrome's traffic to Google servers now uses this. The benefits to mobile traffic is terms of reduced round trips is even more profound.

swankjesse:
Yup! We're still waiting on some sort of QUIC spec.

bgcr:
I'm planning to implement the QUIC support in okhttp source code. Does anyone could help me plotting the main stack responsible for transport connection? I would appreciate any kind of help. Thanks in advance.

swankjesse:
@bgcr please do your work in a fork! We don't have time or interest to have QUIC in trunk at this time.

swankjesse:
https://www.ietf.org/proceedings/96/slides/slides-96-quic-5.pdf

duanbo1983:
@bgcr how's your effort in QUIC support? BTW, I still haven't seen a stable QUIC spec.

realdeadfish:
I think Quic support would deserve a whole new library as it is a whole new transport protocol which has not much in common with HTTP (HTTP/2 even). `okquic` anyone? :)

JY482:
I can see "QUIC is not natively supported by OkHttp, but provided to allow a theoretical interceptor that provides support." in  OkHttp 3.11.0 API
how can I use interceptor?

yschimke:
@JY482 you would need to implement that interceptor yourself

derekargueta:
Hey everyone, just checking if there's been any movement here?

jkozlowski:
I would interested in the maintainers thoughts on this issue, I had a quick look around and there are a few things that could be blocking this:

* Quic runs on UDP. Currently, the notion of a Socket (which UDP does not use) is pretty pervasive across the codebase.
* Most importantly RealCall/RealConnection/StreamAllocation/ConnectionPool are pretty coupled to each other, and then to http1/http2 parts and then to Socket interface.
* All the important interceptors (CallServerInterceptor, ConnectInterceptor) all cast Chain->RealInterceptorChain. This might not be so important though.

From a quick look it feels like extracting a notion of a **Transport** and making it so that RealCall/RealConnection/StreamAllocation/CollectionPool don't think so much in terms of sockets but simply streams, or even better just HttpCodec, as that is ultimately what the higher interfaces require.

[...]

PR #5602, opened 2019-11-17
16 concurrent requests

yschimke:
Testing from discussion in https://github.com/square/okhttp/issues/907

### DIFF
--- a/okhttp/src/main/java/okhttp3/Dispatcher.kt
+++ b/okhttp/src/main/java/okhttp3/Dispatcher.kt
@@ -59,7 +59,7 @@ class Dispatcher constructor() {
    *
    * WebSocket connections to hosts **do not** count against this limit.
    */
-  @get:Synchronized var maxRequestsPerHost = 5
+  @get:Synchronized var maxRequestsPerHost = 10
     set(maxRequestsPerHost) {
       require(maxRequestsPerHost >= 1) { "max < 1: $maxRequestsPerHost" }
       synchronized(this) {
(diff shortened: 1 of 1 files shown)


swankjesse:
I think if we’re gonna bump this, we might as well bump it to 16? Or check to see what Chrome does? Maybe Netty also?

yschimke:
Does HTTP/2 ignore this and use the client/server negotiated stream limits?

swankjesse:
No, that'll often be unbounded. We're trading off latency vs. throughput, cost of client threads, and load on server.

yschimke:
What do you think about being somewhat adaptive?  Maybe having Platform indicate a rough device category or similar and retaining 5 for older devices? 16 for newer?

swankjesse:
I think 16 is probably the right number for 2019. I'd just like to confirm we're in the same ballpark as Chrome, Firefox, and perhaps also Netty and libcurl.

yschimke:
libcurl - unlimited https://curl.haxx.se/libcurl/c/CURLMOPT_MAX_HOST_CONNECTIONS.html by  default

yschimke:
Firefox network.http.max-persistent-connections-per-server 6

yschimke:
Chrome - 6 I believe, but couldn't easily find the exact config, most issues stop with a discussion we are sticking with 6 which Firefox also uses.

yschimke:
Closing this out.  HTTP/2 is probably where we people concerned about the perf should focus on.  This is a separate control until we add load balancing.

swankjesse:
Hang on. Those are limits on connections, not streams. I think maybe we need to make that distinction in our configuration?

swankjesse:
Okay, idea time. The thing that makes 256 or 8192 concurrent requests unreasonable is client-side threads. And we need one client-side thread per client-side request due to our interceptors model.

But if we do suspendable interceptors, the number of client side threads and concurrent requests become decoupled.

yschimke:
We should revisit, work out a more optimal number for mobile clients etc.  But likely it's more complicated than this PR as discussed above.

Issue #5586, opened 2019-11-06
Thread.holdsLock calls result in excessive overhead w/o ability to turn it off


astral303:
When profiling OkHttp3 4.2.1 under a high load (a microservice/RPC scenario), the profiler constantly shows `Thread.holdsLock(this)` as a non-trivial percentage of wall time spent servicing sending a request.

Here's an example hot spot in `Dispatcher.kt`:

```
  private fun promoteAndExecute(): Boolean {
    assert(!Thread.holdsLock(this))
```

The generated Java source makes the call to `Thread.holdsLock(this)` even if assertions are *off*:

```
  private final boolean promoteAndExecute() {
      boolean var1 = !Thread.holdsLock(this);
      boolean isRunning = false;
      boolean var3 = false;
      boolean var4;
      if (_Assertions.ENABLED && !var1) {
```

However, `Thread.holdsLock` is a native call:

```   
public static native boolean holdsLock(Object obj);
```

The JIT optimizer is unable to optimize across native call boundaries. This cause a hard unable-to-optimize situation, which, for example, inhibits inlining across virtual calls. In my profiling for my target use case, this placed a ceiling on performance that I was not able to work around (save for building a patched OkHttp version).

JakeWharton:
Relevant: https://youtrack.jetbrains.com/issue/KT-22292

JakeWharton:
We can mitigate by duplicating https://github.com/Kotlin/kotlinx.coroutines/commit/583d39dd4fe1d8f488df3e384901aa19f95166e2

yschimke:
@JakeWharton this looks worrying i.e. we can't reproduce exactly

```
// Need InlineOnly for efficient bytecode on Android
@file:Suppress("INVISIBLE_REFERENCE", "INVISIBLE_MEMBER")
```

PR #5604, opened 2019-11-17
Kotlin optimized lock asserts

yschimke:
Use a clear intentional method instead of kotlin assert() function vs JVM assert keyword. Guarded behind a cached assertion status check.

Addressing issue https://github.com/square/okhttp/issues/5586

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/Util.kt
+++ b/okhttp/src/main/java/okhttp3/internal/Util.kt
@@ -530,18 +530,18 @@ internal fun <E> MutableList<E>.addIfAbsent(element: E) {
 }
 
 @JvmField
-internal val AssertionsEnabled = OkHttpClient::class.java.desiredAssertionStatus()
+internal val assertionsEnabled = OkHttpClient::class.java.desiredAssertionStatus()
 
 @Suppress("NOTHING_TO_INLINE")
 internal inline fun Any.assertThreadHoldsLock() {
-  if (AssertionsEnabled && !Thread.holdsLock(this)) {
+  if (assertionsEnabled && !Thread.holdsLock(this)) {
     throw AssertionError("Thread ${Thread.currentThread().name} MUST hold lock on $this")
   }
 }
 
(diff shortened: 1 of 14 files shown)


yschimke:
FWIW we could optimize the byte code by avoiding a pretty exception.

```
@Suppress("NOTHING_TO_INLINE")
internal inline fun Any.assertThreadHoldsLock() {
  if (AssertionsEnabled) {
    assert(!Thread.holdsLock(this))
  }
}

@Suppress("NOTHING_TO_INLINE")
internal inline fun Any.assertThreadDoesntHoldLock() {
  if (AssertionsEnabled) {
    assert(Thread.holdsLock(this))
  }
}
```

```
      int isRunning = false;
      boolean var4;
      boolean var5;
      boolean var6;
      if (Util.AssertionsEnabled) {
         boolean var3 = Thread.holdsLock(this);
         var4 = false;
         var5 = false;
         if (_Assertions.ENABLED && !var3) {
            var6 = false;
            String var14 = "Assertion failed";
            throw (Throwable)(new AssertionError(var14));
         }
      }
```

okhttp/src/main/java/okhttp3/internal/Util.kt
```diff
@@ -527,3 +528,20 @@ fun <T> readFieldOrNull(instance: Any, fieldType: Class<T>, fieldName: String):
 internal fun <E> MutableList<E>.addIfAbsent(element: E) {
   if (!contains(element)) add(element)
 }
+
+@JvmField
+internal val AssertionsEnabled = OkHttpClient::class.java.desiredAssertionStatus()
```
swankjesse: Is the casing of this wrong? Should be `assertionsEnabled` ?

yschimke:
> So Kotlin’s assert function makes no sense?

Yep - see here https://youtrack.jetbrains.com/issue/KT-22292

Worse callers now rely on this eager assert, so it needs a migration after it gets deprecated and fixed.

PR #5628, opened 2019-12-03
Optimize judgment conditions

Lillls:
can reduce unnecessary loops


swankjesse:
Not all connections are multiplexed so this is not logically equivalent.

PR #5630, opened 2019-12-03
Actually remove SPDY_3 when configuring OkHttpClient protocols

dmgd:
Fixes https://github.com/square/okhttp/issues/5629


swankjesse:
Thanks!

PR #5631, opened 2019-12-05
Make okhttp native-image compatible

mageddo:
There is a [known issue][2] on Kotlin which make it impossible to compile kotlin softwares to binary using native-image when using [object expressions][3] inside interfaces. By this reason okhttp4+ can't be compiled to binary using [native-image][4]. Unfortunately there is no a road map for kotlin team to fix the issue. 

To make okhttp native-image friendly I'm proposing this pull request, this way it will compiles and run on native-image binaries successfully.

[here a working sample][5] after the fix

[1]: https://github.com/oracle/graal/issues/1521
[2]: https://youtrack.jetbrains.com/issue/KT-33097
[3]: https://kotlinlang.org/docs/reference/object-declarations.html#object-expressions
[4]: https://github.com/oracle/graal/tree/master/substratevm#introduction
[5]: https://github.com/mageddo/graalvm-examples/tree/okhttp-4-graalvm-19.2.1-working

### DIFF
--- a/okhttp/src/main/java/okhttp3/Authenticator.kt
+++ b/okhttp/src/main/java/okhttp3/Authenticator.kt
@@ -110,7 +110,8 @@ interface Authenticator {
   companion object {
     /** An authenticator that knows no credentials and makes no attempt to authenticate. */
     @JvmField
-    val NONE = object : Authenticator {
+    val NONE: Authenticator = AuthenticatorNone()
+    private class AuthenticatorNone : Authenticator {
       override fun authenticate(route: Route?, response: Response): Request? = null
     }
   }
--- a/okhttp/src/main/java/okhttp3/CookieJar.kt
+++ b/okhttp/src/main/java/okhttp3/CookieJar.kt
@@ -51,7 +51,8 @@ interface CookieJar {
   companion object {
     /** A cookie jar that never accepts any cookies. */
     @JvmField
-    val NO_COOKIES: CookieJar = object : CookieJar {
+    val NO_COOKIES: CookieJar = NoCookies()
+    private class NoCookies : CookieJar {
       override fun saveFromResponse(url: HttpUrl, cookies: List<Cookie>) {
       }
 
--- a/okhttp/src/main/java/okhttp3/Dns.kt
+++ b/okhttp/src/main/java/okhttp3/Dns.kt
@@ -42,7 +42,8 @@ interface Dns {
      * lookup IP addresses. Most custom [Dns] implementations should delegate to this instance.
      */
     @JvmField
-    val SYSTEM = object : Dns {
+    val SYSTEM: Dns = DnsSystem()
+    private class DnsSystem : Dns {
       override fun lookup(hostname: String): List<InetAddress> {
         try {
           return InetAddress.getAllByName(hostname).toList()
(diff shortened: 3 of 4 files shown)


JakeWharton:
This is randomly valid at best. Subtle changes to class definition could result in incompatibility with GraalVM.

mageddo:
See [this link][1] for a full working sample using the built snapshot . Thank you guys

[1]: https://github.com/mageddo/graalvm-examples/tree/e2e93db/okhttp4

yschimke:
@mageddo How could we make this simpler to use with existing apps building with gradle without affecting the bulk of non gradle clients?

A okhttp-module that includes ReflectionClasses from your project?

mageddo:
@yschimke We can make a simpler example for okhttp, my example is a little more complex because it is using jackson mapper and jackson is based on reflection, so the projects are required to map all vos which uses reflection at runtime.

Talking about the ReflectionClasses.java itself, it is a try to fix the ssl depedency which is not statically linked to graalvm binary [ref 1][1]. it's not fully working yet because you also have to [ship cacerts by you own][2].

I'm working on ways to automate that, actually it is required to all clients to do that configuration manually, looks like GraalVM team made a increase for that on [19.3.0][3] but I'm getting other issues with SSL on this version

[1]: https://github.com/oracle/graal/blob/master/substratevm/JCA-SECURITY-SERVICES.md
[2]: https://github.com/oracle/graal/issues/378
[3]: https://github.com/oracle/graal/issues/1336

yschimke:
@mageddo yep, this is a pain.  I've gone through this myself with okhttp, and got to a similar point but with a less automated solution.  So I guess when 19.3.0+ fixes remaining issues we can work towards a more push button solution.

mageddo:
@yschimke for graal **19.2.1**, I've made a [new example][1] which can generate a binary without the need of additional files or configurations.

1. you create an okhttp-module as mentioned with the following files
    *  `src/main/resources/META-INF/native-image/okhttp4/native-image.properties` 
    * `org.graalvm.config.SslBuildConfig` 
1. Create a static block on `OkHttpClient` class for example calling `org.graalvm.config.SslBuildConfig#installTrustStore()`

Then clients won't have to make additional configurations to get okhttp working.

You can ensure that by running the binary on a container which have not a jdk or jre

```bash
docker run --rm -it --name tmp -v $PWD/build/graal/graalvm-okhttp4:/graalvm-okhttp4 debian:10-slim ./graalvm-okhttp4
```

[1]: https://github.com/mageddo/graalvm-examples/tree/okhttp-4-graalvm-19.2.1-working-v2/okhttp4

yschimke:
What's stopping you using 19.3?

mageddo:
@yschimke

> What's blocking you using 19.3?

 [This one][1]

[1]: https://github.com/oracle/graal/issues/1934

mageddo:
Got okhttp4 working on **19.3.0**, looks like SunEC is really embedded on graalvm 19.3.0 but for some reason it doesn't have ECDH support.

The fix was to configure BouncyCastle on the project, TrustStore still required. [working sample](https://github.com/mageddo/graalvm-examples/tree/okhttp-4-graalvm-19.3.0-working/okhttp4)

PR #5637, opened 2019-12-08
Avoid running Corretto platform until fixed

yschimke:


yschimke:
hmmm..... should be running on this PR

yschimke:
Filtering seems broken based on regex, so easier to uncomment for specific PR builds

PR #5638, opened 2019-12-08
Degrade connections after a timeout

swankjesse:
This is based roughly on the 'Degraded Connections' proposal here
https://github.com/square/okhttp/issues/3146#issuecomment-471196032

I'm using 1000 ms instead of 500 ms. That's about the cost of a (slow)
TLS handshake, which is what callers would need to do if we degrade
unnecessarily.

### DIFF
--- a/mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.kt
+++ b/mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.kt
@@ -526,6 +526,7 @@ class MockWebServer : ExternalResource(), Closeable {
       }
 
       if (socketPolicy === STALL_SOCKET_AT_START) {
+        dispatchBookkeepingRequest(sequenceNumber, socket)
         return // Ignore the socket until the server is shut down!
       }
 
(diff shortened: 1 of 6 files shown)


mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.kt
```diff
@@ -526,6 +526,7 @@ class MockWebServer : ExternalResource(), Closeable {
       }
 
       if (socketPolicy === STALL_SOCKET_AT_START) {
+        dispatchBookkeepingRequest(sequenceNumber, socket)
```
yschimke: How does this relate to the change?

swankjesse:
For ping events it takes us towards an EventListener not attached to any individual Call instance. That's worth doing! And probably ties nicely into the client side load balancing stuff we're planning.

mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.kt
swankjesse: One of the new test cases uses this socket policy, and there was a bug where it would never proceed past it because it was peeked only.

PR #5642, opened 2019-12-09
Remove JDK 12 (non-LTS)

yschimke:
Suggested policy

JDK 8 - minimum with/out ALPN
JDK11 - Current LTS
JDK13 - Current JDK
JDK14ea - Next EA (when supported by toolchain)

As we have flaky integration tests, this is already ambitious.  So set this as the rough size of our integration tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -174,21 +174,6 @@ jobs:
       - runtests:
           platform: jdk9
 
-  testjdk12:
-    docker:
-      # best source of JDK 12 for now
-      - image: circleci/dynamodb:12.0.2-jdk
-
-    environment:
-      JVM_OPTS: -Xmx1g
-      TERM: dumb
-
-    steps:
-      - checkout
-
-      - runtests:
-          platform: jdk9
-
   testjdk13:
     docker:
       - image: circleci/openjdk:13.0.1-jdk-buster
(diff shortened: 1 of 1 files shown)


swankjesse:
Yep. I’d also potentially axe JDK 9. It is neither LTS nor current and so nobody should be using it.

yschimke:
@swankjesse Thanks for being patient through the noise of additional platforms, I think it's generally worth the pain and I think it's good we are reporting bugs up stream to OpenJDK.  Ideally I'd throw the minimum and most recent Android into the mix above also.

swankjesse:
Maybe we should add this table to our [Requirements documentation](https://square.github.io/okhttp/#requirements), possibly linking out to the corresponding CI tests?

| PLATFORM    | OkHttp | HTTP/2 | NOTES                                                      |
| :---------- | :----- | :----- | :--------------------------------------------------------- |
| JDK 7       | 3.12.x |        | OkHttp 3.12.x is the last release to support JDK 7         |
| JDK 8       | latest | ✻      | HTTP/2 available via Jetty-ALPN or Conscrypt               |
| JDK 9+      | latest | ✅      |                                                            |
| Android 4.x | 3.12.x |        | OkHttp 3.12.x is the latest release to support Android 4.x |
| Android 5+  | latest | ✅      |                                                            |

PR #5645, opened 2019-12-13
Debug logging for TaskRunner

swankjesse:
This is intended to be useful in debugging and otherwise operating
OkHttp in anger.

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/concurrent/TaskLogger.kt
+++ b/okhttp/src/main/java/okhttp3/internal/concurrent/TaskLogger.kt
@@ -1,0 +1,77 @@
+/*
+ * Copyright (C) 2019 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okhttp3.internal.concurrent
+
+import java.util.logging.Level
+
+internal inline fun taskLog(
+  task: Task,
+  queue: TaskQueue,
+  messageBlock: () -> String
+) {
+  if (TaskRunner.logger.isLoggable(Level.FINE)) {
+    log(task, queue, messageBlock())
+  }
+}
+
+internal inline fun <T> logElapsed(
+  task: Task,
+  queue: TaskQueue,
+  block: () -> T
+): T {
+  var startNs = -1L
+  val loggingEnabled = TaskRunner.logger.isLoggable(Level.FINE)
+  if (loggingEnabled) {
+    startNs = queue.taskRunner.backend.nanoTime()
+    log(task, queue, "starting")
+  }
+
+  var completedNormally = false
+  try {
+    val result = block()
+    completedNormally = true
+    return result
+  } finally {
+    if (loggingEnabled) {
+      val elapsedNs = queue.taskRunner.backend.nanoTime() - startNs
+      if (completedNormally) {
+        log(task, queue, "finished in ${formatDuration(elapsedNs)}")
+      } else {
+        log(task, queue, "failed in ${formatDuration(elapsedNs)}")
+      }
+    }
+  }
+}
+
+private fun log(task: Task, queue: TaskQueue, message: String) {
+  TaskRunner.logger.fine("${queue.name} $message: ${task.name}")
+}
+
+/**
+ * Returns a duration in the nearest whole-number units like "999 µs" or "1 ms". This rounds 0.5
+ * units away from 0 and 0.499 towards 0. The smallest unit this returns is "µs"; the largest unit
+ * it returns is "s". For values in [-499..499] this returns "0 µs".
+ */
+fun formatDuration(ns: Long): String {
+  return when {
+    ns <= -999_500_000 -> "${(ns - 500_000_000) / 1_000_000_000} s"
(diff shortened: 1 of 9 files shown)


okhttp/src/main/java/okhttp3/internal/concurrent/TaskLogger.kt
```diff
@@ -0,0 +1,73 @@
...
+  task: Task,
+  queue: TaskQueue,
+  messageBlock: () -> String
+) {
+  if (TaskRunner.logger.isLoggable(Level.FINE)) {
+    log(task, queue, messageBlock())
+  }
+}
+
+internal inline fun <T> logElapsed(
```
yschimke: In inlining this worth it?

okhttp/src/main/java/okhttp3/internal/concurrent/TaskLogger.kt
```diff
@@ -0,0 +1,73 @@
...
+  task: Task,
+  queue: TaskQueue,
+  block: () -> T
+): T {
+  var startNs = -1L
+  val loggingEnabled = TaskRunner.logger.isLoggable(Level.FINE)
+  if (loggingEnabled) {
```
yschimke: Maybe a case for inverting this check and returning earlier, save a lot of indenting and checks below.

okhttp/src/main/java/okhttp3/internal/concurrent/TaskLogger.kt
```diff
@@ -0,0 +1,73 @@
...
+    val result = block()
+    completedNormally = true
+    return result
+  } finally {
+    if (loggingEnabled) {
+      val elapsedNs = queue.taskRunner.backend.nanoTime() - startNs
+      if (completedNormally) {
+        log(task, queue, "finished in ${formatTime(elapsedNs)}")
+      } else {
+        log(task, queue, "failed in ${formatTime(elapsedNs)}")
```
yschimke: worth including exn.toString() in the message?

okhttp/src/main/java/okhttp3/internal/concurrent/TaskLogger.kt
```diff
@@ -0,0 +1,73 @@
...
+
+fun formatTime(ns: Long): String {
+  return when {
+    ns <= -1_000_000_000 -> "${ns / 1_000_000_000} s"
+    ns <= -1_000_000 -> "${ns / 1_000_000} ms"
+    ns <= -1_000 -> "${ns / 1_000} µs"
+    ns < 1_000 -> "$ns ns"
+    ns < 1_000_000 -> "${ns / 1_000} µs"
+    ns < 1_000_000_000 -> "${ns / 1_000_000} ms"
+    else -> "after ${ns / 1_000_000_000} s"
```
yschimke: why "after" here?

okhttp/src/test/java/okhttp3/internal/ws/WebSocketHttpTest.java
```diff
@@ -195,8 +191,7 @@
 
     newWebSocket();
 
-    assertThat(logs.take()).isEqualTo("");
-    logger.removeHandler(logs);
+    assertThat(testLogHandler.take()).isEqualTo("");
```
yschimke: why is "" logged?

yschimke:
It would be great to add this to the retroactive logging in OkHttpClientTestRule.  Would be really helpful for debugging our CI test failures.

okhttp/src/main/java/okhttp3/internal/concurrent/TaskLogger.kt
swankjesse: Yeah, saves an allocation of a lambda.

okhttp/src/main/java/okhttp3/internal/concurrent/TaskLogger.kt
swankjesse: Oooh, that's a bug. Earlier versions had different text for negative and positive durations.

okhttp/src/main/java/okhttp3/internal/concurrent/TaskLogger.kt
swankjesse: I did it this way to avoid duplicating `block()` once inlined.

okhttp/src/main/java/okhttp3/internal/concurrent/TaskLogger.kt
swankjesse: This is a case where we’re going to re-throw. Don’t want this logging to be any kind of substitute for application-layer troubleshooting.

okhttp/src/test/java/okhttp3/internal/ws/WebSocketHttpTest.java
swankjesse: Really old `@Ignored` test. Probably worth further investigation. I fixed to this:
```
    assertThat(testLogHandler.take()).isEqualTo("INFO: [WS client] onFailure");
```

swankjesse:
@yschimke Please take another look at the TestLogging.kt class; I made the duration formatting fancier.

Agreed on getting OkHttpClientTestRule to capture logs. But I wanna hold that for a follow-up change.

okhttp/src/main/java/okhttp3/internal/concurrent/TaskLogger.kt
```diff
@@ -0,0 +1,77 @@
...
+private fun log(task: Task, queue: TaskQueue, message: String) {
+  TaskRunner.logger.fine("${queue.name} $message: ${task.name}")
+}
+
+/**
+ * Returns a duration in the nearest whole-number units like "999 µs" or "1 ms". This rounds 0.5
+ * units away from 0 and 0.499 towards 0. The smallest unit this returns is "µs"; the largest unit
+ * it returns is "s". For values in [-499..499] this returns "0 µs".
+ */
+fun formatDuration(ns: Long): String {
```
swankjesse: # 👀

yschimke:
> TestLogging

TaskLogger?

yschimke:
LGTM

PR #5656, opened 2019-12-20
Use our test rule instead of the Maven Surefire listener

swankjesse:
Closes: https://github.com/square/okhttp/issues/4894


swankjesse:
The “big fix” was adding `OkHttpClientTestRule.ensureAllTaskQueuesIdle()` because that method links our tests to the async tasks that they spawn. When a test finishes we know that there’s nothing running left behind. Yay!

See also https://github.com/square/okhttp/pull/5296

okhttp-testing-support/src/main/java/okhttp3/OkHttpClientTestRule.kt
```diff
@@ -75,21 +88,29 @@ class OkHttpClientTestRule : TestRule {
...
-          .withFailMessage("Queue still active after 1000ms")
+          .withFailMessage("Queue still active after 1000 ms")
           .isTrue()
     }
   }
 
   override fun apply(base: Statement, description: Description): Statement {
     return object : Statement() {
       override fun evaluate() {
+        val defaultUncaughtExceptionHandler = Thread.getDefaultUncaughtExceptionHandler()
```
yschimke: Do any of our build environments run tests concurrently in the same process? Gradle, Intellij?

swankjesse: Nope! And if ever we want to, multiple processes is the right way. We have lots of static state in our tests, including system properties and security providers.

Issue #4964, opened 2019-04-18
Http2Stream to have independent errors for request and response streams


sinhaashish:
I am using minio for Object storage. Minio is a golang project using go1.12 as the base go version. 
 The server is HTTP2 enabled.  I use using minio java client which uses `okhttp3` . when i try to store the object in a bucket in  multi threaded mode i get `okhttp3.internal.http2.StreamResetException: stream was reset: NO_ERROR`.
Refer https://github.com/minio/minio/issues/7501 . 

`okhttp3.OkHttpClient` uses  http2.0 by default which causes such issue. 
If i forcefuly set the protocol to 1.1 ,it works fine.
Refer PR https://github.com/minio/minio-java/pull/766.

I find there is some issue with `Go1.12` or `okhttp3` as i found this issue in the community. 
Refer https://github.com/square/okhttp/issues/3955
The only solution provided there was to revert to HTTP1_1 .

Setting the protocol to 1.1 is not the solution as i am not using the functionality of HTTP2.0 enabled server.
Below is the code which replicates the problem 

 ```
 import io.minio.MinioClient;
class PutObjectRunnable implements Runnable {
    MinioClient client;
    String bucketName;
    String filename;

    public PutObjectRunnable(MinioClient client, String bucketName, String filename) {
        this.client = client;
        this.bucketName = bucketName;
        this.filename = filename;
    }

    public void run() {
        StringBuffer traceBuffer = new StringBuffer();

        try {

            client.putObject(bucketName, filename, filename);

        } catch (Exception e) {
            System.err.print(traceBuffer.toString());
            e.printStackTrace();
        }
    }
}

```

 
 

```import io.minio.*;

public class ThreadedPutObject {
    public static void main(String args[]) throws Exception {
	try{
        MinioClient client = new MinioClient("https://play.min.io:9000", "Q3AM3UQ867SPQQA43P2F", "zuf+tfteSlswRu7BJ86wekitnifILbZam1KYY3TG", true);
        client.traceOn(System.out);
        long startTime = System.currentTimeMillis();
		 boolean found = minioClient.bucketExists("my-bucketname");
      if (found) {
        System.out.println("my-bucketname already exists");
      } else {
        // Create bucket 'my-bucketname'.
        minioClient.makeBucket("my-bucketname");
        System.out.println("my-bucketname is created successfully");
      }        
      Thread[] threads = new Thread[7];
      String[] location = new String[]{"<<PATH_TO_OBJECT_1>>",
[...]

PR #5663, opened 2019-12-24
Failing test for NO_ERROR stream reset

swankjesse:
It turns out it'll be a bit difficult to fix this because we don't
currently support reading the response stream after an error.

https://github.com/square/okhttp/issues/4964

### DIFF
--- a/okhttp/src/test/java/okhttp3/DuplexTest.java
+++ b/okhttp/src/test/java/okhttp3/DuplexTest.java
@@ -18,11 +18,14 @@ package okhttp3;
 import java.io.IOException;
 import java.net.HttpURLConnection;
 import java.net.ProtocolException;
+import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.LinkedBlockingQueue;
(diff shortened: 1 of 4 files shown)


okhttp/src/test/java/okhttp3/DuplexTest.java
```diff
@@ -505,6 +510,64 @@
...
 
+  /**
+   * OkHttp currently doesn't implement failing the request body stream independently of failing the
+   * corresponding response body stream. This is necessary if we want servers to be able to stop
+   * inbound data and send an early 400 before the request body completes.
+   *
+   * This test sends a slow request that is canceled by the server. It expects the response to still
+   * be readable after the request stream is canceled.
+   */
+  @Ignore
```
yschimke: Is this shortlived?  If not, rather than ignoring, is it worth flagging with a TODO but stating the test in terms of how the failure is currently unfortunately observed?

swankjesse: It’s hopefully shortlived. Need to do some implementation refactoring to make it work.

Issue #5578, opened 2019-10-28
TTFB and responseHeadersStart being called too soon


toncid:
We are trying to measure the TTFB (_time to first byte_) value per request. It seems that tracking the difference between `requestHeadersEnd` (or `requestBodyEnd`) and `responseHeadersStart` is the way to go.

However, `responseHeadersStart` is triggered immediately for some reason, while the request is still in flight.

Example output:

```
0.000 callStart
...
0.320 connectionAcquired
0.322 requestHeadersStart
0.323 requestHeadersEnd     <--- request sent
0.325 responseHeadersStart  <--- response headers received in 2 ms and then downloaded for 240 ms?
0.561 responseHeadersEnd    <--- actual TTFB?
0.563 responseBodyStart
0.571 responseBodyEnd
0.578 connectionReleased
0.579 callEnd
```

This looks like a bug. However, the same timeline is documented in the sample output on: https://square.github.io/okhttp/events/

Am I missing something here?

PR #5669, opened 2019-12-29
Convert RecordingEventListener to Kotlin

swankjesse:


swankjesse:
I’m working on fixing #5578 and I wanted to add timestamps to this. Converting to Kotlin felt like a natural first step.

PR #5670, opened 2019-12-30
Defer EventListener events until bytes are returned

swankjesse:
This changes the timing of responseHeaderStart and responseBodyStart events to
fire when bytes are received from the server. This is a non-trivial behavior
change and should be documented as such in the release notes. In particular,
the responseFailed event may be fired without a preceding responseHeadersStart
event.

To test this I've added a timestamp to our CallEvent test facet.

Closes: https://github.com/square/okhttp/issues/5578

### DIFF
--- a/okhttp-testing-support/src/main/java/okhttp3/RecordingEventListener.kt
+++ b/okhttp-testing-support/src/main/java/okhttp3/RecordingEventListener.kt
@@ -40,6 +40,7 @@ import okhttp3.CallEvent.ResponseHeadersStart
 import okhttp3.CallEvent.SecureConnectEnd
 import okhttp3.CallEvent.SecureConnectStart
 import org.assertj.core.api.Assertions.assertThat
+import org.assertj.core.data.Offset
 import org.junit.Assert.assertTrue
 import java.io.IOException
 import java.net.InetAddress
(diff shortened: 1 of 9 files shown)


okhttp-testing-support/src/main/java/okhttp3/RecordingEventListener.kt
```diff
@@ -63,21 +69,50 @@ open class RecordingEventListener : EventListener() {
...
+    return result
   }
 
   fun recordedEventTypes() = eventSequence.map { it.name }
 
   fun clearAllEvents() {
-    eventSequence.clear()
+    while (eventSequence.isNotEmpty()) {
+      takeEvent()
+    }
```
JakeWharton: What's the difference here?

okhttp-testing-support/src/main/java/okhttp3/RecordingEventListener.kt
```diff
@@ -87,9 +122,14 @@ open class RecordingEventListener : EventListener() {
...
-    val startEvent = e.closes()
+    val startEvent = e.closes(-1L)
     if (startEvent != null) {
-      assertTrue(eventSequence.contains(startEvent))
+      val noCloses = eventSequence.any { it == e.closes(it.timestampNs) }
+      if (noCloses) {
+        assertTrue(noCloses)
+      } else {
+        fail("close without open")
+      }
```
JakeWharton: I think this is just
```kotlin
assertTrue("close without open", noCloses)
```

okhttp-testing-support/src/main/java/okhttp3/RecordingEventListener.kt
swankjesse: I need to call takeEvent() so the `lastTimestampNs` is updated.

okhttp-testing-support/src/main/java/okhttp3/RecordingEventListener.kt
swankjesse: yep, thanks. (I exploded that into 5 lines to set breakpoints and forgot to roll back!)

PR #5672, opened 2019-12-31
Fix crash on repeated MockWebServer shutdown

swankjesse:
The problem was the awaitIdle() call was scheduling executor jobs after
the ExecutorService had been shut down. This fixes that and defends
against other races.

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/concurrent/TaskQueue.kt
+++ b/okhttp/src/main/java/okhttp3/internal/concurrent/TaskQueue.kt
@@ -18,7 +18,6 @@ package okhttp3.internal.concurrent
 import okhttp3.internal.assertThreadDoesntHoldLock
 import java.util.concurrent.CountDownLatch
 import java.util.concurrent.RejectedExecutionException
-import java.util.concurrent.TimeUnit
 
 /**
  * A set of tasks that are executed in sequential order.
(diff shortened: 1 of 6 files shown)


okhttp/src/test/java/okhttp3/internal/concurrent/TaskRunnerRealBackendTest.kt
```diff
@@ -82,11 +82,22 @@ class TaskRunnerRealBackendTest {
...
 
     assertThat(log.take()).isEqualTo("failing task running")
     assertThat(log.take()).isEqualTo("uncaught exception: java.lang.RuntimeException: boom!")
     assertThat(log.take()).isEqualTo("normal task running")
     assertThat(log).isEmpty()
   }
+
+  @Test fun idleLatchAfterShutdown() {
+    queue.schedule("task") {
+      Thread.sleep(250)
```
yschimke: nit: on it's own 250ms is fine.  As part of a test suite 40 tests taking this approach add 10 seconds to running the tests and is still more flaky then using e.g. CountDownLatch here.  

I've raised https://github.com/square/okhttp/issues/5673 to fix such usages.

okhttp/src/main/java/okhttp3/internal/concurrent/TaskQueue.kt
```diff
@@ -101,25 +100,42 @@ class TaskQueue internal constructor(
...
-      override fun runOnce(): Long {
-        latch.countDown()
-        return -1L
+      // If there's an existing AwaitIdleTask, use it. This is necessary when the executor is
+      // shutdown but still busy as we can't enqueue in that case.
+      val existingTask = activeTask
+      if (existingTask is AwaitIdleTask) {
+        return existingTask.latch
+      }
+      for (futureTask in futureTasks) {
```
yschimke: Not sure about practically, but do you need to check anything but the last task here?  If it's non empty and has other tasks, seems by definition it's not idle.

swankjesse: I think at the edges I’m conflating two concepts:
 * idle: nothing running, nothing scheduled
 * flush: complete all currently-enqueued tasks

In practice they’re the same effect for us, but I’ve implemented flush and labeled it idle. Because true idle is more work to implement for academic benefit I’m going to leave it as-is for now. Unsatisfying!

swankjesse: One other wrench here . . . idle is distinct from shutdown. For example in our test suite we enforce idleness between tests, but share the same task runner between tests.

yschimke: Hopefully this is still a remaining cause of when our tests fail they tend to be flaky in batches, and we can fix a few remaining issues and have it rock solid.

Overall - trending ok https://circleci.com/build-insights/gh/square/okhttp/master

But big room for improvements once the base is solid.

Issue #5690, opened 2020-01-04
Android test failures on Android 5.0 (21)


yschimke:
```
09:23:57 V/InstrumentationResultParser: 3) testConscryptRequest(okhttp.android.test.OkHttpTest)
09:23:57 V/InstrumentationResultParser: org.junit.ComparisonFailure: expected:<org.conscrypt.[ConscryptFileDescriptorSocket]> but was:<org.conscrypt.[KitKatPlatformOpenSSLSocketImplAdapter]>
09:23:57 V/InstrumentationResultParser: at org.junit.Assert.assertEquals(Assert.java:115)
09:23:57 V/InstrumentationResultParser: at org.junit.Assert.assertEquals(Assert.java:144)
09:23:57 V/InstrumentationResultParser: at okhttp.android.test.OkHttpTest.testConscryptRequest(OkHttpTest.kt:142)
```

```
09:23:57 V/InstrumentationResultParser: 1) testRequestUsesPlayProvider(okhttp.android.test.OkHttpTest)
09:23:57 V/InstrumentationResultParser: com.google.android.gms.common.GooglePlayServicesNotAvailableException
09:23:57 V/InstrumentationResultParser: at com.google.android.gms.common.GooglePlayServicesUtilLight.ensurePlayServicesAvailable(Unknown Source)
09:23:57 V/InstrumentationResultParser: at com.google.android.gms.common.GoogleApiAvailabilityLight.verifyGooglePlayServicesIsAvailable(Unknown Source)
09:23:57 V/InstrumentationResultParser: at com.google.android.gms.security.ProviderInstaller.installIfNeeded(Unknown Source)
09:23:57 V/InstrumentationResultParser: at okhttp.android.test.OkHttpTest.testRequestUsesPlayProvider(OkHttpTest.kt:156)
```

```
09:23:57 V/InstrumentationResultParser: 2) testMockWebserverRequest(okhttp.android.test.OkHttpTest)
09:23:57 V/InstrumentationResultParser: java.lang.IllegalArgumentException: List has more than one element.
09:23:57 V/InstrumentationResultParser: at kotlin.collections.CollectionsKt___CollectionsKt.single(_Collections.kt:519)
09:23:57 V/InstrumentationResultParser: at okhttp.android.test.OkHttpTest.testMockWebserverRequest(OkHttpTest.kt:289)
```

PR #5691, opened 2020-01-04
Handle Android 5.0 socket type in tests

yschimke:
https://github.com/square/okhttp/issues/5690


swankjesse:
Weird it uses the word KitKat in the Lollipop release!

JakeWharton:
It usually means KitKat and up. AndroidX does the same.

PR #5692, opened 2020-01-04
Google Play assumption failures handled

yschimke:


yschimke:
https://github.com/square/okhttp/issues/5690

PR #5697, opened 2020-01-04
Fix missing import in Android Test

yschimke:
```
> Task :android-test:compileDebugAndroidTestKotlin FAILED
e: okhttp/android-test/src/androidTest/java/okhttp/android/test/OkHttpTest.kt: (164, 9): Unresolved reference: AssumptionViolatedException
```


yschimke:
Something a bit wacky with imports for junit with the instrumentation tests.  Possibly related to some transient test dependency?  This works...

PR #5699, opened 2020-01-05
Fixup samples

yschimke:


samples/guide/src/test/kotlin/okhttp3/AllMainsTest.kt
```diff
@@ -23,14 +23,18 @@ import java.io.File
 import java.lang.reflect.InvocationTargetException
 
 @RunWith(Parameterized::class)
-@Ignore
+//@Ignore
```
yschimke: TODO: revert before landing

samples/guide/src/main/java/okhttp3/recipes/kt/CustomTrust.kt
```diff
@@ -23,24 +23,19 @@ import java.io.IOException
 import java.io.InputStream
 import java.security.KeyStore
 import java.security.cert.CertificateFactory
-import java.util.Arrays
-import javax.net.ssl.KeyManagerFactory
-import javax.net.ssl.SSLContext
-import javax.net.ssl.TrustManager
-import javax.net.ssl.TrustManagerFactory
-import javax.net.ssl.X509TrustManager
+import javax.net.ssl.*
```
yschimke: TODO: revert

samples/guide/src/main/java/okhttp3/recipes/PostStreamingWithPipe.java
```diff
@@ -32,6 +32,8 @@
   private final OkHttpClient client = new OkHttpClient();
 
   public void run() throws Exception {
+    if (true) return;
```
yschimke: TODO: revert

samples/guide/src/main/java/okhttp3/recipes/CustomTrust.java
```diff
@@ -39,21 +41,30 @@
...
+
   public CustomTrust() {
-    X509TrustManager trustManager;
     SSLSocketFactory sslSocketFactory;
     try {
-      trustManager = trustManagerForCertificates(trustedCertificatesInputStream());
+      KeyAndTrustManagers keyAndTrustManagers = trustManagerForCertificates(trustedCertificatesInputStream());
       SSLContext sslContext = SSLContext.getInstance("TLS");
-      sslContext.init(null, new TrustManager[] { trustManager }, null);
+      sslContext.init(keyAndTrustManagers.keyManagers, keyAndTrustManagers.trustManagers, null);
```
yschimke: @swankjesse Should we change this test to use okhttp-tls?

yschimke:
@swankjesse Did you change your certificates for publicobject.com to Let's Encrypt?

image

In general, how can we make it easier to debug a failure like this?

yschimke:
Root certs from the test

image


image

samples/guide/src/main/java/okhttp3/recipes/CustomTrust.java
swankjesse: I did!

swankjesse: Yes!

swankjesse:
Yep, I changed publicobject to letsencrypt. Good catch.

yschimke:
@swankjesse it really shouldn't require this much expert knowledge to diagnose, I think we can do a lot better to explain these problems.  Just not sure how...

samples/guide/src/main/java/okhttp3/recipes/CustomTrust.java
yschimke: For followup https://github.com/square/okhttp/issues/5701

PR #5732, opened 2020-01-19
Add utility for prior responses as a sequence

yschimke:
For discussion.  Low value, but something to consider e.g.

- Should these sorts of things be public util API
- extension methods (Response.priorResponses)
- or built in to Response

Second question, should we support an easy way to chain responses when the interceptor is essentially forking? e.g. something like Exception.addSuppressed (but without mutability). Currently it's possible but awkward.

### DIFF
--- a/okhttp/src/main/java/okhttp3/Response.kt
+++ b/okhttp/src/main/java/okhttp3/Response.kt
@@ -303,7 +303,7 @@ class Response internal constructor(
   }
 
   /**
-   * Returns the sequence of Responses that
+   * Returns the sequence of Responses that led to this final Response.
    */
   fun responses(): Iterable<Response> = generateSequence(this) {
     it.priorResponse
--- a/okhttp/src/main/java/okhttp3/Response.kt
+++ b/okhttp/src/main/java/okhttp3/Response.kt
@@ -302,10 +302,19 @@ class Response internal constructor(
     checkNotNull(body) { "response is not eligible for a body and must not be closed" }.close()
   }
 
+  /**
+   * Returns the sequence of Responses that
+   */
   fun responses(): Iterable<Response> = generateSequence(this) {
     it.priorResponse
   }.asIterable()
 
+  /**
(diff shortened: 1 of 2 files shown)


okhttp/src/main/java/okhttp3/Response.kt
```diff
@@ -302,6 +302,21 @@ class Response internal constructor(
     checkNotNull(body) { "response is not eligible for a body and must not be closed" }.close()
   }
 
+  /**
+   * Returns the sequence of Responses that led to this final Response.
```
swankjesse: Is this backwards? I think `this` should be last for most uses? Also unsure of the name. response.responses() feels non-obvious. I wanna reach for something like `history()` but that doesn’t feel right either. `progression()` ? That’s too scary-sounding.

okhttp/src/main/java/okhttp3/Response.kt
```diff
@@ -302,6 +302,21 @@ class Response internal constructor(
...
+    it.priorResponse
+  }.asIterable()
+
+  /**
+   * Returns the number of network Responses that were received to get to this final Response.
+   *
+   * The minimum is 1. A 401 response that is then correctly authenticated will have a response
+   * count of 2.
+   */
+  val responseCount = responses().count()
```
swankjesse: I know it’s not as critical anymore, but I don’t love the extra allocations this creates in every Response creation.

yschimke:
Not ready for a public API

swankjesse:
> Should we support an easy way to chain responses when the interceptor is essentially forking?

I think what we’re missing is tags on responses. This could be for forking, but also for any other work the interceptor is doing. Though I struggle to come up with compelling examples.

okhttp/src/main/java/okhttp3/Response.kt
yschimke: This is a great point, I would definitely leave this as a function.

PR #5733, opened 2020-01-19
Handle Conscrypt decode behaviour

yschimke:
Fix for https://app.circleci.com/jobs/github/square/okhttp/8574

### DIFF
--- a/okhttp-tls/src/main/java/okhttp3/tls/HeldCertificate.kt
+++ b/okhttp-tls/src/main/java/okhttp3/tls/HeldCertificate.kt
@@ -47,6 +47,7 @@ import org.bouncycastle.asn1.x509.GeneralName
 import org.bouncycastle.asn1.x509.X509Extensions
 import org.bouncycastle.jce.provider.BouncyCastleProvider
 import org.bouncycastle.x509.X509V3CertificateGenerator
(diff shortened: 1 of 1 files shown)


okhttp-tls/src/main/java/okhttp3/tls/HeldCertificate.kt
```diff
@@ -516,9 +517,17 @@ class HeldCertificate(
     }
 
     private fun decodePem(pem: String): X509Certificate {
-      val certificates = CertificateFactory.getInstance("X.509")
+      val certificateFactory = CertificateFactory.getInstance("X.509")
+      val certificates = certificateFactory
           .generateCertificates(Buffer().writeUtf8(pem).inputStream())
```
yschimke: @swankjesse Any reason we don't use the singular form?

https://docs.oracle.com/javase/7/docs/api/java/security/cert/CertificateFactory.html#generateCertificate(java.io.InputStream)

swankjesse: Looks like it was a mistake. I like your new behavior; seems like it’d catch bugs where there are multiple certificates in a single string.

PR #5736, opened 2020-01-20
Switch to Gradle's Maven Publishing Plugin

swankjesse:
This removes a bunch of low-value stuff from the pom.xml files:

 - name
 - description
 - licence distribution ('repo')
 - developers clause
 - test dependencies

I don't think any of this will be missed, and it shrinks the pom.xml
file to the minimal set of useful stuff.

This also causes us to publish a gradle .module file. This is the
motivation for this change. It'll allow us to ship a Gradle platform,
which is a more capable than a Maven BOM.

### DIFF
--- a/docs/releasing.md
+++ b/docs/releasing.md
@@ -78,8 +78,8 @@ Cutting a Release
 
     ```
     sed -i "" \
-      "s/VERSION_NAME=.*/VERSION_NAME=$RELEASE_VERSION/g" \
-      gradle.properties
+      "s/version = '.*'/version = '$RELEASE_VERSION'/g" \
+      build.gradle
     sed -i "" \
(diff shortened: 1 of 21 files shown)


docs/releasing.md
```diff
@@ -78,8 +78,8 @@ Cutting a Release
 
     ```
     sed -i "" \
-      "s/VERSION_NAME=.*/VERSION_NAME=$RELEASE_VERSION/g" \
-      gradle.properties
+      "s/version = '.*'/version = '$RELEASE_VERSION'/g" \
```
yschimke: This suggests we should consider moving it to it's own file, or calculating it from the branch, tag name.

https://github.com/nemerosa/versioning does the latter.

yschimke:
That's a nice Spring clean.

docs/releasing.md
swankjesse: I think I wanna punt on that for a follow-up. (I’m anxious about the complexity cost / benefit of such things, especially as the versioning thing is more code that itself needs to be kept up to date.)

PR #5742, opened 2020-01-21
Publish a BOM for OkHttp 4.x.

swankjesse:
https://docs.gradle.org/current/userguide/java_platform_plugin.html

This prevents downstream applications from pulling in different
versions of various OkHttp subprojects.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -66,9 +66,28 @@ plugins {
   id 'me.champeau.gradle.japicmp' version '0.2.9'
 }
 
+/** Returns the artifact ID for the project, or null if it is not published. */
+ext.publishedArtifactId = { project ->
+  if (project.name == 'okhttp-logging-interceptor') {
+    return 'logging-interceptor'
+  } else if (project.name == 'mockwebserver'
+          || project.name == 'okcurl'
+          || project.name == 'okhttp'
+          || project.name == 'okhttp-bom'
+          || project.name == 'okhttp-brotli'
+          || project.name == 'okhttp-dnsoverhttps'
+          || project.name == 'okhttp-sse'
+          || project.name == 'okhttp-tls'
+          || project.name == 'okhttp-urlconnection') {
+    return project.name
+  } else {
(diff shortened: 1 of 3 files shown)


swankjesse:
I have to admit that now that I understand these I think they’re great. I regret abandoning the BOM contributions made by external contributors!

okhttp-bom/build.gradle
```diff
@@ -0,0 +1,13 @@
+dependencies {
+  constraints {
+    api project(':mockwebserver')
```
JakeWharton: Can we do this in a loop? And maybe conditionally based on if the `maven-publish` plugin was applied?

```groovy
dependencies {
  constraints {
    rootProject.subprojects { subproject ->
      if (subproject != project) {
        subproject.plugins.withId('maven-publish') {
          api subproject
        }
      }
    }
  }
}
```

I have no idea if this works.

build.gradle
```diff
@@ -87,8 +87,10 @@ allprojects {
   }
 }
 
+/** Configure building for Java+Kotlin projects. */
 subprojects { project ->
   if (project.name == 'android-test') return
+  if (project.ext.artifactId == 'okhttp-bom') return
```
JakeWharton: Why the different mechanism than `name`?

okhttp-bom/build.gradle
swankjesse: Unfortunately it doesn’t work because this runs before all of the subprojects’ plugins are applied. But I got something almost as good that avoids duplication:

```
dependencies {
  constraints {
    rootProject.subprojects { subproject ->
      def artifactId = rootProject.ext.publishedArtifactId(subproject)
      if (artifactId != null && project.name != 'okhttp-bom') {
        api subproject
      }
    }
  }
}
```

build.gradle
swankjesse: No good reason. Fixed to be consistent with above.

Issue #4145, opened 2018-07-12
Support easily configuring the TLS cipher suites in MockWebServer


skingsland:
I'm trying to use MockWebServer to test my OkHttpClient configuration. The real web server my code hits just dropped support for TLS v1.0, so I'm changing my code to use TLS v1.2 and use the server's preferred cipher suites. I'd like to test this change with an in-memory web server that mimics the real server (in that it uses TLS 1.1 or 1.2, and supports certain cipher suites), but there's not an easy way to tell MockWebServer which cipher suites to use.

This isn't easy, because `MockWebServer#useHttps` expects an SSLSocketFactory, and there's no good way to configure the cipher suites that it will use. The way I did it was to wrap the underlying SSLSocketFactory, and call `sslSocket.setEnabledCipherSuites(..)` on the SSLSocket before returning it, like so:

    final class MySSLSocketFactory extends SSLSocketFactory {
        final SSLSocketFactory wrappedSSLSocketFactory;

        MySSLSocketFactory(SSLSocketFactory wrappedSSLSocketFactory) {
            this.wrappedSSLSocketFactory = wrappedSSLSocketFactory;
        }

        @Override
        public String[] getDefaultCipherSuites() {
            return getMyCipherSuites();
        }

        @Override
        public String[] getSupportedCipherSuites() {
            return getMyCipherSuites();
        }

        @Override
        public Socket createSocket(Socket s, String host, int port, boolean autoClose) throws IOException {
            final SSLSocket sslSocket = (SSLSocket) wrappedSSLSocketFactory.createSocket(s, host, port, autoClose);

            // change the supported cipher suites on the socket, *before* it's returned to the client
            sslSocket.setEnabledCipherSuites(getMyCipherSuites());

            return sslSocket;
        }

        // other overloaded createSocket() methods do the same

        private String[] getMyCipherSuites() {
            // TODO: change this to return whatever cipher suites you want the server to use, from CipherSuite
            return new String[] {...};
        }
    }

It's already kind of a hassle to set up a MockWebServer with HTTPS in a test, with the cert, keystore, trustmanager, etc., and this just makes it more complicated. It would be great if the cipher suites (and TLS version while we're at it) were easily configurable, but if not, then at least having a recipe to follow would help a lot.

swankjesse:
If we supported ConnectionSpec in MockWebServer, that would go a long way?
https://square.github.io/okhttp/3.x/okhttp/okhttp3/ConnectionSpec.html

[...]

Issue #2916, opened 2016-10-16
OkHttpServer


swankjesse:
We’re exploring building this. We could reuse the same request, response, and interceptor types.

artem-zinnatullin:
1. In WIP PR https://github.com/square/okhttp/pull/2912/files: `<artifactId>okwebserver</artifactId>`, not `okhttpserver`. Can you please clarify naming?
2. Will `OkHttpServer` result in deletion of `MockWebServer` or it'll be used under the hood of `MockWebServer`?

swankjesse:
TBD.

artem-zinnatullin:
Yeah, I understand, those questions were not like "ANSWER THEM NOW OR I'LL UNSTAR OKHTTP ON GITHUB!!" but more "Please think about that and share some thoughts when you'll have some opinion". 

No pressure! 🐱

JakeWharton:
MWS can't test SSE because body is `Buffer`. This should support.

yschimke:
@swankjesse would you build this and https://github.com/cashapp/misk ?

swankjesse:
Misk is a server framework. It currently uses Jetty though if we wanted to make the investment we could swap out Jetty for something we build ourselves.

yschimke:
I'm more wondering whether we can close this?

PR #5745, opened 2020-01-22
Introduces MockWebServer builder API

bric3:
Hi,

This pull requests proposes a builder API for MockWebServer. The motivation is that the current API has a few limitations in comfort.

If used as a JUnit Rule, each mehtod has to configure the webserver in a repetitive way, while this is useful when testing different _communication_ setups, it can be a bit repetitive if the tests focus is the request/response exchange. Also when using MockWebServer as a JUnit rule it's impossible to configure the port binding. 

```java
@Rule
public MockWebServer server = new MockWebServer();

@Test
public void use_mock_web_server() {
  server.useHttps(serverHandshakeCertificates.sslSocketFactory());
  server.requestClientAuth();
  server.start(8443);
  // ...
}
```

After this PR

```java
@Rule
public MockWebServer server = new MockWebServer.Builder()
    .port(middleware.port())
    .useHttps(serverHandshakeCertificates.sslSocketFactory())
    .requestClientAuth()
    .build();

@Test
public void use_mock_weber_server_without_configuring() {
  // ...
}
```

For example before this PR, try-with-resources had to be used this way

```java
try(MockWebServer server = new MockWebServer())
  server.useHttps(serverHandshakeCertificates.sslSocketFactory());
  server.requestClientAuth();
  server.start(8443);
  // ...
}
```

After 

```java
try(MockWebServer server = new MockWebServer.Builder()
    .port(8443)
    .useHttps(serverHandshakeCertificates.sslSocketFactory())
    .requestClientAuth()
    .build()) {
  // ...
}
```

The proposed change don't break existing API however this lead some drawbacks I'm not quite happy with, e.g. this snippet is legal and won't raise any errors :

```java
MockWebServer server = new MockWebServer.Builder()
    .port(8443)
    .useHttps(serverHandshakeCertificates.sslSocketFactory())
    .requestClientAuth()
    .build()
server.noClientAuth();
server.start(8080);
```

I'm not sure how to address this. I'll be happy to fix any suggestions you have.
By the way I believe I have already signed the ICLA for leakcanary.

Cheers

### DIFF
--- a/mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.kt
+++ b/mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.kt
@@ -100,6 +100,7 @@ class MockWebServer private constructor(
         port: Int = -1,
         sslSocketFactory: SSLSocketFactory?,
         tunnelProxy: Boolean,
+        clientAuth: Int,
         protocols: List<Protocol>
 ) : ExternalResource(), Closeable {
   private val taskRunnerBackend = TaskRunner.RealBackend(
@@ -217,14 +218,16 @@ class MockWebServer private constructor(
       port = 0,
       sslSocketFactory = null,
       tunnelProxy = false,
-      protocols = immutableListOf(Protocol.HTTP_2, Protocol.HTTP_1_1)
+      clientAuth = CLIENT_AUTH_NONE,
+      protocols = emptyList()
   ) {}
 
   init {
     this.portField = port
     this.sslSocketFactory = sslSocketFactory
     this.tunnelProxy = tunnelProxy
-    this.protocols = protocols
+    this.clientAuth = clientAuth
+    this.protocols = if(protocols.isNullOrEmpty()) immutableListOf(Protocol.HTTP_2, Protocol.HTTP_1_1) else protocols
   }
 
   @JvmName("-deprecated_port")
@@ -1150,6 +1153,7 @@ class MockWebServer private constructor(
   class Builder {
     internal var sslSocketFactory: SSLSocketFactory? = null
     internal var tunnelProxy: Boolean = false
+    internal var clientAuth = CLIENT_AUTH_NONE
     internal var port: Int = 0
     var protocols: List<Protocol> = ArrayList()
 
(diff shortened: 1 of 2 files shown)


bric3:
This reminds me that this builder API could be used to introduce `ConnectionSpec` as in #4145 ?

bric3:
Some stage of the build failed, but they don't seem related to the changes of this PR.

ci/circleci: testjdk8 failed with this test
 - okhttp3.WholeOperationTimeoutTest.noTimeout

Locally this comands succeeds `GRADLE_OPTS='-Dorg.gradle.daemon=false -Dokhttp.platform=jdk8 -Dorg.gradle.workers.max=3 -Xmx1G -Dtest.java.home=$HOME/.sdkman/candidates/java/8.0.232-amzn/' ./gradlew --build-cache --parallel --continue test`

ci/circleci: testjdk11 failed with these tests
 - okhttp3.InterceptorTest.interceptorMakesAnUnrelatedAsyncRequest
 - okhttp3.EventListenerTest.requestBodyFailHttp1OverHttps
 - okhttp3.EventListenerTest.requestBodyFailHttp2OverHttps
 - okhttp3.EventListenerTest.secureConnectWithTunnel
 - okhttp3.EventListenerTest.timeToFirstByteHttp1OverHttps
 - okhttp3.CacheTest.varyMatchesAddedRequestHeaderField

Locally this command `GRADLE_OPTS='-Dorg.gradle.daemon=false -Dokhttp.platform=jdk9 -Dorg.gradle.workers.max=3 -Xmx1G ' ./gradlew --build-cache --parallel --continue test` succeeds.

swankjesse:
Neat! At this point I’m not interested in making incremental API improvements to MockWebServer because I don’t want to encourage our users to pay the cost to migrate.

BUT! I am interested in doing a real update to the MockWebServer APIs where we adopt modern API practices: immutable MockResponse objects, no JUnit dependency, streaming, etc.

bric3:
> BUT! I am interested in doing a real update to the MockWebServer APIs where we adopt modern API practices: immutable MockResponse objects, no JUnit dependency, streaming, etc.

Does that mean deprecating MockWebServer in favor of another type/api ?

> Neat! At this point I’m not interested in making incremental API improvements to MockWebServer because I don’t want to encourage our users to pay the cost to migrate.

Fair enough, that's why I didn't want to deprectate the current API. The motivation was really to provide a way to configure the server at declaration before the server starts.

swankjesse:
I’d prefer to avoid deprecating MockWebServer because I don’t want to cause downstream consumers to do the work to upgrade. The current thing works just fine; we just think we could do nicer.

But I do like the idea of introducing a new API, only if we could make it significantly better than what we have today. This is a big project!

For now I’d like to stall on this. I don’t want to make incremental API improvements when we should be making dramatic API improvements.

bric3:
OK, thanks for explaining !
Is this something I could help with ? _I noticed a reference to OkHttpServer in #2916._

yschimke:
I think we agreed to hold off on this for now.  Not a comment on the PR quality.

Issue #4689, opened 2019-03-07
failedRoutes in RouteDatabase causes Full GC


lfz757077613:
1. In my project, I used Okhttp with a large of proxies. Most of the proxies were used only once, so failedRoutes saved too many routes and cause full gc all the time
2. I think that failedRoutes use a time-bound cache instead of a simple set is better

swankjesse:
Wow!

parker89:
I encountered this issue as well, and filed a PR to fix it: https://github.com/square/okhttp/pull/5746

lfz757077613:
> I encountered this issue as well. This PR should fix it: #5746

cool~I hava used [ahc](https://github.com/AsyncHttpClient/async-http-client) for a long time, very good performance. I think you can also try it

yschimke:
Comment by @swankjesse https://github.com/square/okhttp/pull/5746#issuecomment-578387225

Inverts the problem from RouteSelector -> 1xRouteDatabase -> X*Y*Address

to

RouteSelector -> X*RouteDatabase -> Y*Address

X is client settings like SSL Socket Factory
Y is address like host, port etc

swankjesse:
Yes! You explained it much better.

yschimke:
Should we scope the connection pool instead of route database here?

PR #5746, opened 2020-01-22
Expire old routes in the failed routes database

parker89:
Currently the RouteDatabase uses a set to track failed routes. If a route is never attempted again, it can accumulate in this set indefinitely since there is no mechanism to purge old routes. 

Each route requires about 270KB, so even 1000 routes will use 270MB of heap. 

This PR uses a cache with an expiration time of 10 minutes and a max size of 1,000 routes.

For context, we have clients that talk to applications in a cluster. The cluster has many nodes and each node hosts many applications, which are on randomized ports. Since each host/port can be random, the list of failed routes can grow indefinitely. 

This PR addresses this issue https://github.com/square/okhttp/issues/4689

### DIFF
--- a/okhttp/build.gradle
+++ b/okhttp/build.gradle
@@ -40,6 +40,7 @@ dependencies {
   testImplementation deps.junit
   testImplementation deps.assertj
   testImplementation deps.openjsse
+  testImplementation deps.guava
   testCompileOnly deps.jsr305
 }
 
--- a/okhttp/build.gradle
+++ b/okhttp/build.gradle
@@ -29,6 +29,7 @@ dependencies {
   compileOnly deps.openjsse
   compileOnly deps.jsr305
   compileOnly deps.animalSniffer
+  compileOnly deps.guava
 
   testImplementation project(':okhttp-testing-support')
   testImplementation project(':okhttp-tls')
--- a/okhttp-dnsoverhttps/build.gradle
+++ b/okhttp-dnsoverhttps/build.gradle
@@ -13,6 +13,7 @@ dependencies {
   testImplementation deps.conscrypt
   testImplementation deps.junit
   testImplementation deps.assertj
+  testImplementation deps.guava
 }
 
 afterEvaluate { project ->
--- a/okhttp-logging-interceptor/build.gradle
+++ b/okhttp-logging-interceptor/build.gradle
@@ -16,6 +16,7 @@ dependencies {
   testImplementation project(':okhttp-testing-support')
   testImplementation project(':okhttp-tls')
   testImplementation deps.assertj
+  testImplementation deps.guava
 }
 
 afterEvaluate { project ->
--- a/okhttp-sse/build.gradle
+++ b/okhttp-sse/build.gradle
@@ -14,6 +14,7 @@ dependencies {
   testImplementation project(':mockwebserver')
   testImplementation deps.junit
   testImplementation deps.assertj
+  testImplementation deps.guava
   testCompileOnly deps.jsr305
 }
 
--- a/okhttp/src/main/java/okhttp3/internal/connection/RouteDatabase.kt
+++ b/okhttp/src/main/java/okhttp3/internal/connection/RouteDatabase.kt
@@ -33,7 +33,7 @@ class RouteDatabase {
 
   /** Records a failure connecting to [failedRoute]. */
   @Synchronized fun failed(failedRoute: Route) {
-    failedRoutes.put(failedRoute, Object())
+    failedRoutes.put(failedRoute, Any())
   }
 
   /** Records success connecting to [route]. */
--- a/okhttp/src/main/java/okhttp3/internal/connection/RouteDatabase.kt
+++ b/okhttp/src/main/java/okhttp3/internal/connection/RouteDatabase.kt
@@ -15,7 +15,9 @@
  */
 package okhttp3.internal.connection
 
+import com.google.common.cache.CacheBuilder
 import okhttp3.Route
+import java.time.Duration
 
 /**
  * A blacklist of failed routes to avoid when creating a new connection to a target address. This is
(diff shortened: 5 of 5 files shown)


swankjesse:
Where did you get the 270 KiB number?

okhttp/build.gradle
```diff
@@ -29,6 +29,7 @@ dependencies {
   compileOnly deps.openjsse
   compileOnly deps.jsr305
   compileOnly deps.animalSniffer
+  compileOnly deps.guava
```
swankjesse: We are not adding a Guava dependency.

parker89: I assumed it would be fine since it's already part of the build.gradle here: https://github.com/square/okhttp/blob/master/build.gradle#L13

What's the concern with adding it? And do you have any suggestions otherwise?

parker89:
I got 270 KiB from eclipse memory analyzer when investigating why the RouteDatabase on our server was taking up ~3GB. This may be specific to our particular situation though. I can try to get a breakdown of why a `Route` is that large.

parker89:
The reason the `Route` is so large is:
`Route` has an `okhttp3.Address` 
`okhttp3.Address` has variable `sslSocketFactory` of type `sun.security.ssl.SSLSocketFactoryImpl`,
which has variable `context` of type `sun.security.ssl.SSLContextImpl$TLSContext`,
which has a `serverCache` and a `clientCache` of type `sun.security.ssl.SSLSessionContextImpl`

`SSLSessionContextImpl` has 2 `sun.security.util.MemoryCache`s, `sessionCache` and `sessionHostPortCache`, which are initialized by `SSLSessionContextImpl` to have a size of 20,480 by default.
https://docs.oracle.com/en/java/javase/12/docs/api/java.base/javax/net/ssl/SSLSessionContext.html#getSessionCacheSize()

So `sessionCache` and `sessionHostPortCache` are around ~131KB each, even though they only hold a single element (at least on our JVM). (Basically the size of the backing LinkedHashMap is 32,768 slots, since 2^15 is the power of 2 larger than 20,480) 

I suppose we could lower `javax.net.ssl.sessionCacheSize` from the default of 20,480, but that would affect other places on our JVM. 

Adding a time-based cache would improve the behavior here for everyone.

swankjesse:
@parker89 do all of your addresses have the same `SSLSocketFactoryImpl` or are they different?

swankjesse:
If you SSLSocketFactory instances are all different, you aren’t going to get any benefit from the route database anyway! And so our current behavior is definitely bad and needs a fix.

I’m curious – _why_ are they all different?

yschimke:
@swankjesse I think the bigger issue here is perhaps holding onto sslSocketFactory.  We can't really know what platform costs this has.  Would a change along this line be worthwhile?

```
diff --git a/okhttp/src/main/java/okhttp3/Address.kt b/okhttp/src/main/java/okhttp3/Address.kt
index 1911a42f8..97a0c0e79 100644
--- a/okhttp/src/main/java/okhttp3/Address.kt
+++ b/okhttp/src/main/java/okhttp3/Address.kt
@@ -22,6 +22,7 @@ import javax.net.SocketFactory
 import javax.net.ssl.HostnameVerifier
 import javax.net.ssl.SSLSocketFactory
 import okhttp3.internal.toImmutableList
+import java.lang.ref.WeakReference

 /**
  * A specification for a connection to an origin server. For simple connections, this is the
@@ -38,16 +39,16 @@ class Address(
   @get:JvmName("dns") val dns: Dns,

   /** Returns the socket factory for new connections. */
-  @get:JvmName("socketFactory") val socketFactory: SocketFactory,
+  @get:JvmName("socketFactory") val socketFactory: WeakReference<SocketFactory>,

   /** Returns the SSL socket factory, or null if this is not an HTTPS address. */
-  @get:JvmName("sslSocketFactory") val sslSocketFactory: SSLSocketFactory?,
+  @get:JvmName("sslSocketFactory") val sslSocketFactory: WeakReference<SSLSocketFactory>?,

   /** Returns the hostname verifier, or null if this is not an HTTPS address. */
   @get:JvmName("hostnameVerifier") val hostnameVerifier: HostnameVerifier?,

   /** Returns this address's certificate pinner, or null if this is not an HTTPS address. */
-  @get:JvmName("certificatePinner") val certificatePinner: CertificatePinner?,
+  @get:JvmName("certificatePinner") val certificatePinner: WeakReference<CertificatePinner>?,

   /** Returns the client's proxy authenticator. */
   @get:JvmName("proxyAuthenticator") val proxyAuthenticator: Authenticator,
@@ -178,9 +179,7 @@ class Address(
     result = 31 * result + connectionSpecs.hashCode()
     result = 31 * result + proxySelector.hashCode()
     result = 31 * result + Objects.hashCode(proxy)
-    result = 31 * result + Objects.hashCode(sslSocketFactory)
     result = 31 * result + Objects.hashCode(hostnameVerifier)
-    result = 31 * result + Objects.hashCode(certificatePinner)
     return result
   }
```

It has extra allocation cost for normal single client, multiple requests+addresses mode.  So asking before doing anything here.

yschimke:
Alternatively do we want to expire old routes and try them again anyway?

swankjesse:
I think the WeakReference approach is _almost_ perfect. But we don't want the references to be weak in the route itself because it breaks equals and has other problems that'll be difficult to uncover.

Instead we scope the RouteDatabase to these things, and if our RouteDatabase has the wrong ones we toss is and create a new one when the OkHttpClient is created.

yschimke:
@parker89 Did you plan to take a look at this?  How urgent is this for you?

swankjesse:
Another take that avoids Guava: https://github.com/square/okhttp/pull/5756

parker89:
@yschimke This isn't incredibly urgent for us since the it takes around 1 week to accumulate enough `SSLSocketFactoryImpl` instances to cause problems. 

We were planning on looking into why we have so many `SSLSocketFactoryImpl` instances first though.

yschimke:
Closing this PR in favour of https://github.com/square/okhttp/pull/5756

PR #5748, opened 2020-01-22
Bouncycastle Client Auth Fixes

yschimke:


yschimke:
Debug only code

okhttp-testing-support/src/main/java/okhttp3/testing/PlatformRule.kt
```diff
@@ -318,6 +319,7 @@ open class PlatformRule @JvmOverloads constructor(
       } else if (platformSystemProperty == BOUNCYCASTLE_PROPERTY && Security.getProviders()[0].name != "BC") {
         Security.insertProviderAt(BouncyCastleProvider(), 1)
         Security.insertProviderAt(BouncyCastleJsseProvider(), 2)
+        Security.insertProviderAt(BouncyCastlePQCProvider(), 3)
```
yschimke: not required

PR #5751, opened 2020-01-24
Bouncy Castle Provider working with HTTP/2 on JDK 8

yschimke:
Improves client side for BouncyCastle use.  n.b. highlights issues when properly using BouncyCastle in a server.  Fixes for that to come separately.

### DIFF
--- a/okhttp-testing-support/build.gradle
+++ b/okhttp-testing-support/build.gradle
@@ -4,6 +4,7 @@ dependencies {
   api deps.assertj
   api deps.bouncycastledebug
   implementation deps.bouncycastlepkix, {
+    // prefer debug build for our own testing
     exclude module:"bcprov-jdk15on"
   }
   implementation deps.bouncycastletls, {
--- a/okhttp-testing-support/build.gradle
+++ b/okhttp-testing-support/build.gradle
@@ -2,8 +2,13 @@ dependencies {
   api project(':okhttp')
   api deps.junit
   api deps.assertj
-  api deps.bouncycastle
-  api deps.bouncycastletls
+  api deps.bouncycastledebug
+  implementation deps.bouncycastlepkix, {
+    exclude module:"bcprov-jdk15on"
+  }
+  implementation deps.bouncycastletls, {
+    exclude module:"bcprov-jdk15on"
+  }
   api deps.conscrypt
   api deps.corretto
   api deps.openjsse
--- a/okhttp-testing-support/src/main/java/okhttp3/OkHttpDebugLogging.kt
+++ b/okhttp-testing-support/src/main/java/okhttp3/OkHttpDebugLogging.kt
@@ -47,17 +47,5 @@ object OkHttpDebugLogging {
     }
   }
 
-  fun enable(loggerClass: KClass<*>) {
-    val logger = Logger.getLogger(loggerClass.java.name)
-    if (configuredLoggers.add(logger)) {
-      logger.addHandler(ConsoleHandler().apply {
-        level = Level.FINE
-        formatter = object : SimpleFormatter() {
-          override fun format(record: LogRecord) =
-              String.format("[%1\$tF %1\$tT] %2\$s %n", record.millis, record.message)
-        }
-      })
-      logger.level = Level.FINEST
-    }
-  }
+  fun enable(loggerClass: KClass<*>) = enable(loggerClass.java.name)
 }
--- a/okhttp-testing-support/src/main/java/okhttp3/OkHttpDebugLogging.kt
+++ b/okhttp-testing-support/src/main/java/okhttp3/OkHttpDebugLogging.kt
@@ -33,7 +33,21 @@ object OkHttpDebugLogging {
 
   fun enableTaskRunner() = enable(TaskRunner::class)
 
-  private fun enable(loggerClass: KClass<*>) {
+  fun enable(loggerClass: String) {
+    val logger = Logger.getLogger(loggerClass)
+    if (configuredLoggers.add(logger)) {
+      logger.addHandler(ConsoleHandler().apply {
+        level = Level.FINE
+        formatter = object : SimpleFormatter() {
+          override fun format(record: LogRecord) =
+            String.format("[%1\$tF %1\$tT] %2\$s %n", record.millis, record.message)
+        }
+      })
+      logger.level = Level.FINEST
+    }
+  }
+
+  fun enable(loggerClass: KClass<*>) {
     val logger = Logger.getLogger(loggerClass.java.name)
     if (configuredLoggers.add(logger)) {
       logger.addHandler(ConsoleHandler().apply {
@@ -43,7 +57,7 @@ object OkHttpDebugLogging {
               String.format("[%1\$tF %1\$tT] %2\$s %n", record.millis, record.message)
         }
       })
-      logger.level = Level.FINE
+      logger.level = Level.FINEST
     }
   }
 }
--- a/okhttp/src/main/java/okhttp3/internal/platform/BouncyCastlePlatform.kt
+++ b/okhttp/src/main/java/okhttp3/internal/platform/BouncyCastlePlatform.kt
@@ -1,0 +1,96 @@
+/*
+ * Copyright (C) 2019 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okhttp3.internal.platform
+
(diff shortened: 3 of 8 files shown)


yschimke:
Test output 

```
Running Tests with BouncyCastlePlatform
Jan 24, 2020 10:24:30 PM org.bouncycastle.jsse.provider.ProvTrustManagerFactorySpi getDefaultTrustStore
Jan 24, 2020 10:24:30 PM org.bouncycastle.jsse.provider.PropertyUtils getStringSecurityProperty
INFO: Found string security property [jdk.tls.disabledAlgorithms]: SSLv3, RC4, DES, MD5withRSA, DH keySize < 1024, EC keySize < 224, 3DES_EDE_CBC, anon, NULL
Jan 24, 2020 10:24:30 PM org.bouncycastle.jsse.provider.PropertyUtils getStringSecurityProperty
INFO: Found string security property [jdk.certpath.disabledAlgorithms]: MD2, MD5, SHA1 jdkCA & usage TLSServer, RSA keySize < 1024, DSA keySize < 1024, EC keySize < 224
[2020-01-24 22:24:30] Boolean system property [jdk.tls.trustNameService] defaulted to: false 
[2020-01-24 22:24:30] Boolean system property [jsse.enableSNIExtension] defaulted to: true 
[2020-01-24 22:24:31] Boolean system property [org.bouncycastle.jsse.ec.disableChar2] defaulted to: false 
[2020-01-24 22:24:31] Boolean system property [org.bouncycastle.ec.disable_f2m] defaulted to: false 
[2020-01-24 22:24:31] Client notified of selected protocol version: TLSv1.2 
[2020-01-24 22:24:31] Server specified new session: 9ec20e1b4dd811551a5133f9086f1aa9c4d7b2ddf8f2a8574961704b29aea03e 
[2020-01-24 22:24:31] Client notified of selected cipher suite: TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384 
[2020-01-24 22:24:32] Client notified of selected protocol version: TLSv1.2 
[2020-01-24 22:24:32] Server specified new session: b4da168bcc707749314e65a4c5dcabdbba315932c5915fc1611b588733465457 
[2020-01-24 22:24:32] Client notified of selected cipher suite: TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256 
[2020-01-24 22:24:32] Client raised warning(1) close_notify(0) alert: Connection closed 
[2020-01-24 22:24:32] Client raised warning(1) close_notify(0) alert: Connection closed 
```

okhttp/src/test/java/okhttp3/BouncyCastleTest.kt
```diff
@@ -39,38 +32,10 @@ class BouncyCastleTest {
 
   @Before
   fun setUp() {
+    OkHttpDebugLogging.enable("org.bouncycastle.jsse")
     platform.assumeBouncyCastle()
   }
 
-  @Test
-  fun testTlsv13Works() {
-    platform.expectFailureOnJdkVersion(8)
```
yschimke: Test wasn't running previously and so didn't flag issues with client and server running together.

okhttp/src/main/java/okhttp3/internal/platform/BouncyCastlePlatform.kt
```diff
@@ -0,0 +1,96 @@
...
+
+      sslSocket.parameters = sslParameters
+    } else {
+      super.configureTlsExtensions(sslSocket, hostname, protocols)
+    }
+  }
+
+  override fun getSelectedProtocol(sslSocket: SSLSocket): String? =
+      if (sslSocket is BCSSLSocket) {
+        when (val protocol = (sslSocket as BCSSLSocket).applicationProtocol) {
```
yschimke: This cast is because of overload

```
public interface BCSSLSocket
{
    String getApplicationProtocol();
```

And JDK9 method directly on SSLSocket

okhttp-testing-support/build.gradle
```diff
@@ -2,8 +2,13 @@ dependencies {
...
   api deps.junit
   api deps.assertj
-  api deps.bouncycastle
-  api deps.bouncycastletls
+  api deps.bouncycastledebug
+  implementation deps.bouncycastlepkix, {
+    exclude module:"bcprov-jdk15on"
+  }
+  implementation deps.bouncycastletls, {
+    exclude module:"bcprov-jdk15on"
```
swankjesse: could you add a comment to explain this? Dependency excludes are always hiding problems as far as I’m concerned~

okhttp-testing-support/src/main/java/okhttp3/OkHttpDebugLogging.kt
```diff
@@ -33,7 +33,21 @@ object OkHttpDebugLogging {
...
+        formatter = object : SimpleFormatter() {
+          override fun format(record: LogRecord) =
+            String.format("[%1\$tF %1\$tT] %2\$s %n", record.millis, record.message)
+        }
+      })
+      logger.level = Level.FINEST
+    }
+  }
+
+  fun enable(loggerClass: KClass<*>) {
```
swankjesse: change this to call through the above code?

okhttp/src/main/java/okhttp3/internal/platform/BouncyCastlePlatform.kt
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  public override fun trustManager(sslSocketFactory: SSLSocketFactory): X509TrustManager? =
+      throw UnsupportedOperationException(
+          "clientBuilder.sslSocketFactory(SSLSocketFactory) not supported with BouncyCastle")
+
+  override fun configureTlsExtensions(
+    sslSocket: SSLSocket,
+    hostname: String?,
+    protocols: List<@JvmSuppressWildcards Protocol>
```
swankjesse: Curious why `@JvmSuppressWildcards` is necessary

okhttp/src/main/java/okhttp3/internal/platform/Platform.kt
```diff
@@ -196,6 +196,12 @@ open class Platform {
         return "OpenJSSE" == preferredProvider
       }
 
+    private val isBouncyCastlePreferred: Boolean
+      get() {
+        val preferredProvider = Security.getProviders()[0].name
```
swankjesse: I think this is a pretty good policy for Conscrypt because you’re probably only installing it to get access to it’s HTTPS provider. But for BouncyCastle I suspect people install it to fill gaps in the built-in JSSE. In particular, generating private keys!

So I’d prefer BC to require an explicit opt-in.

yschimke: What would that look like? We don't have a public API for this?  System property?

The other consideration is whether to refactor Platform to use the SocketAdapter model everywhere and support all modes at once.

okhttp/src/main/java/okhttp3/internal/platform/BouncyCastlePlatform.kt
yschimke: IIRC we did it to retain exact compatibility with Java in the Kotlin migration.  not certain at the moment.

okhttp-testing-support/build.gradle
yschimke: It's about running with the debug version which does some logging.  But otherwise you get randomly logging or not when debugging.

yschimke:
> Cool!
> 
> Just curious – who are our potential customers for this? Are users interested in the BouncyCastle provider?

Possibly noone.

Motivation

1. I hit this when trying options for graal with so loading issues.  Those unrelated issues might be solved now.
2. Generally I wanted to understand for different crypto libraries if and how broken we are
3.  Work out whether it's an OkHttp bug or Crypto library bug or just a limitation
4. It's a bit of a Sudoku exercise for me personally, interesting to problem solve through to understand the crypto libraries

swankjesse:
I really like the crypto-bug vs. OkHttp bug angle.

okhttp/src/main/java/okhttp3/internal/platform/Platform.kt
swankjesse: Supporting all modes at once would be nice. As we do it we’re potentially risking locking in the wrong platform at application start-up.

swankjesse: For now, I’d prefer a system property. I’m particularly anxious about users upgrading to 4.4 and getting BouncyCastle unexpectedly.

okhttp/src/main/java/okhttp3/internal/platform/BouncyCastlePlatform.kt
swankjesse: SGTM!

okhttp-testing-support/build.gradle
swankjesse: oh, that’s totally interesting. What signifies the debug version?

yschimke: 'bouncycastle': "org.bouncycastle:bcprov-jdk15on:${versions.bouncycastle}",
      'bouncycastledebug': "org.bouncycastle:bcprov-debug-jdk15on:${versions.bouncycastle}",

These two are the same with additional debug.  It's minimal, but given how bad every other SSL library is at debugability, it's something...

yschimke: discussed here https://www.bouncycastle.org/latest_releases.html

PR #5752, opened 2020-01-25
Explain prior responses issue for App Interceptors

yschimke:


yschimke:
For discussion, tests to come.

okhttp/src/main/java/okhttp3/internal/connection/Transmitter.kt
```diff
@@ -122,7 +122,7 @@ class Transmitter(
       if (this.request!!.url.canReuseConnectionFor(request.url) && exchangeFinder!!.hasRouteToTry()) {
         return // Already ready.
       }
-      check(exchange == null)
+      check(exchange == null) { "Transmitted has existing exchange, has an app interceptor left prior responses open?" }
```
octylFractal:
```suggestion
      check(exchange == null) { "Transmitter has existing exchange, has an app interceptor left prior responses open?" }
```
I assume this is what was intended.

yschimke: Thanks

okhttp/src/main/java/okhttp3/internal/connection/Transmitter.kt
```diff
@@ -122,7 +122,7 @@ class Transmitter(
       if (this.request!!.url.canReuseConnectionFor(request.url) && exchangeFinder!!.hasRouteToTry()) {
         return // Already ready.
       }
-      check(exchange == null)
+      check(exchange == null) { "Transmitter has existing exchange, has an app interceptor left prior responses open?" }
```
swankjesse: Wanna use the exact same text we use on line 158?

swankjesse: (possibly fixing that line to match this!)

PR #5754, opened 2020-01-25
Android Integration - app lifecycle and background/network availability 

yschimke:
Draft (WIP) of integration on Android

- app lifecycle
- hook into services to correctly handle background and network events

FYI just starting with what the external API would look like

### DIFF
--- a/okhttp/src/main/java/okhttp3/android/OkHttpAndroidClientFactory.kt
+++ b/okhttp/src/main/java/okhttp3/android/OkHttpAndroidClientFactory.kt
@@ -1,0 +1,28 @@
+package okhttp3.android
+
+import android.content.Context
+import okhttp3.OkHttpClient
+
+class OkHttpAndroidClientFactory(val appCtxt: Context, val client2: OkHttpClient) {
+
+    val client: OkHttpClient
+        get() {
+            return client2
+        }
+
+    fun newClient(init: OkHttpClient.Builder.() -> Unit): OkHttpClient {
+        val builder = client.newBuilder()
+        init(builder)
+        return builder.build()
+    }
+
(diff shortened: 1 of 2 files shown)


yschimke:
Plan to adopt approaches from previous tests

e.g. background events https://github.com/yschimke/OkHttpAndroidApp/blob/master/android/app/src/main/java/ee/schimke/okhttp/android/android/AppForegroundStatus.kt

airplan mode https://github.com/yschimke/OkHttpAndroidApp/blob/master/android/app/src/main/java/ee/schimke/okhttp/android/android/PhoneStatusLiveData.kt

network events

okhttp/src/main/java/okhttp3/android/OkHttpAndroidClientFactory.kt
```diff
@@ -0,0 +1,28 @@
+package okhttp3.android
+
+import android.content.Context
+import okhttp3.OkHttpClient
+
+class OkHttpAndroidClientFactory(val appCtxt: Context, val client2: OkHttpClient) {
+
+    val client: OkHttpClient
```
yschimke: Plan to shut this client down when context is disposed.

okhttp/src/main/java/okhttp3/android/OkHttpAndroidClientFactory.kt
```diff
@@ -0,0 +1,28 @@
...
+        get() {
+            return client2
+        }
+
+    fun newClient(init: OkHttpClient.Builder.() -> Unit): OkHttpClient {
```
yschimke: Will make sure this is OK from java.

PR #5756, opened 2020-01-26
Limit the route database to 128 elements

swankjesse:
We don't have an API to clear this, and in some degenerate situations it
can grow to hundreds of megabytes in size because the SSLSocketFactory
instances held by Route are large.

Closes: https://github.com/square/okhttp/issues/4689

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/Util.kt
+++ b/okhttp/src/main/java/okhttp3/internal/Util.kt
@@ -172,6 +172,17 @@ fun Array<String>.concat(value: String): Array<String> {
 }
 
 /**
+ * Removes the first element of this non-empty set and returns it. If this set was created with
+ * [mutableSetOf] this returns the oldest element in the set.
+ */
+fun <T : Any?> MutableSet<T>.removeFirst(): T {
+  val i = iterator()
+  val result = i.next()
+  i.remove()
+  return result
+}
+
+/**
  * Increments [startIndex] until this string is not ASCII whitespace. Stops at [endIndex].
  */
 fun String.indexOfFirstNonAsciiWhitespace(startIndex: Int = 0, endIndex: Int = length): Int {
--- a/okhttp/src/main/java/okhttp3/internal/connection/RouteDatabase.kt
+++ b/okhttp/src/main/java/okhttp3/internal/connection/RouteDatabase.kt
@@ -16,6 +16,7 @@
 package okhttp3.internal.connection
 
 import okhttp3.Route
+import okhttp3.internal.removeFirst
 
 /**
  * A blacklist of failed routes to avoid when creating a new connection to a target address. This is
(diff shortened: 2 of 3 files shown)


yschimke:
Any thoughts on also scoping the client? Or is this enough.  FWIW I had time on a flight to San Francisco https://github.com/square/okhttp/pull/5758

okhttp/src/main/java/okhttp3/internal/connection/RouteDatabase.kt
```diff
@@ -24,10 +25,19 @@ import okhttp3.Route
  * preferred.
  */
 class RouteDatabase {
+  /**
+   * Don't grow unbounded, especially since the [SSLSocketFactory] instances held by [Route] may
+   * be quite large: 270 KiB according to a report on issue 5746.
+   */
+  private val maxSize = 128
```
yschimke: Kotlin question, should this be a const to avoid per database field allocation?

okhttp/src/main/java/okhttp3/internal/Util.kt
```diff
@@ -171,6 +171,17 @@ fun Array<String>.concat(value: String): Array<String> {
   return result as Array<String>
 }
 
+/**
+ * Removes the first element of this non-empty set and returns it. If this set was created with
```
yschimke: As this could be reused, maybe worth describing the failure case of empty set.

yschimke:
My only concern with this change is that if the issue is size of SSLSocketFactory instances, then under typical cases there is probably only one SSLSocketFactory shared amongst all instances.  So this probably introduces delays into apps with thousands of target hosts.  We maybe relied on this with IPv6 issues because we don't have happy eyeballs.

okhttp/src/main/java/okhttp3/internal/Util.kt
```diff
@@ -171,6 +171,17 @@ fun Array<String>.concat(value: String): Array<String> {
...
+ * [mutableSetOf] this returns the oldest element in the set.
+ */
+fun <T : Any?> MutableSet<T>.removeFirst(): T {
```
yschimke: Probably worth apply this to LinkedHashSet instead of MutableSet, since MutableSet doesn't have the ordering guarantees.

```
 * A generic unordered collection of elements that does not support duplicate elements, and supports
 * adding and removing elements.
```

okhttp/src/main/java/okhttp3/internal/connection/RouteDatabase.kt
```diff
@@ -24,10 +25,19 @@ import okhttp3.Route
...
+
   private val failedRoutes = mutableSetOf<Route>()
```
yschimke: May as well presize to maxSize and use direct constructor `LinkedHashSet(maxSize)`

okhttp/src/main/java/okhttp3/internal/connection/RouteDatabase.kt
```diff
@@ -24,10 +25,19 @@ import okhttp3.Route
...
 
   /** Records a failure connecting to [failedRoute]. */
   @Synchronized fun failed(failedRoute: Route) {
+    if (failedRoutes.size == maxSize) {
```
yschimke: YUNO?

https://stackoverflow.com/a/5601377/1542667

```
public class MaxSizeHashMap<K, V> extends LinkedHashMap<K, V> {
    private final int maxSize;

    public MaxSizeHashMap(int maxSize) {
        this.maxSize = maxSize;
    }

    @Override
    protected boolean removeEldestEntry(Map.Entry<K, V> eldest) {
        return size() > maxSize;
    }
}
```

Issue #5632, opened 2019-12-06
Shaded OkHttps should have different thread names


zhxnlai:
This looks like the normal OkHttpConnectionPool thread. Maybe rename it to prefix with the shading package?
```
"OkHttp ConnectionPool" #82 daemon prio=5 os_prio=0 cpu=214.96ms elapsed=26805.18s tid=0x00007f36e40be800 nid=0x73 in Object.wait()  [0x00007f36e31e4000]
   java.lang.Thread.State: TIMED_WAITING (on object monitor)
	at java.lang.Object.wait(java.base@11.0.3/Native Method)
	- waiting on <no object reference available>
	at java.lang.Object.wait(java.base@11.0.3/Object.java:462)
	at com.launchdarkly.shaded.okhttp3.ConnectionPool$1.run(ConnectionPool.java:67)
```

PR #5757, opened 2020-01-27
Use different thread names if OkHttp is shaded

swankjesse:
Closes: https://github.com/square/okhttp/issues/5632


okhttp/src/main/java/okhttp3/internal/Util.kt
```diff
@@ -532,6 +532,15 @@ internal fun <E> MutableList<E>.addIfAbsent(element: E) {
 @JvmField
 internal val assertionsEnabled = OkHttpClient::class.java.desiredAssertionStatus()
 
+/**
+ * Returns the string "OkHttp" unless the library has been shaded for inclusion in another library.
```
JakeWharton: "or obfuscated with tools like R8 or ProGuard."

PR #5758, opened 2020-01-27
Scope the RouteDB more precisely

yschimke:


swankjesse:
Ugh, hate to collide with you!

I was thinking about this approach and the drawback is that if you customize your client with newBuilder() you could inadvertently prevent sharing of bad routes that would otherwise be safe to share. For example, you might configure your SSLSocketFactory as late as possible to support the Android provider installer mechanism.

yschimke:
That's a one off cost.  You have two instances of RouteDatabase in that case.

swankjesse:
I'm imagining something like this:
```
var client = OkHttpClient()
var sslSocketFactory = ...

fun newCall(request: Request) {
  return client.newBuilder()
    .sslSocketFactory (sslSocketFactory)
    .build()
    .newCall(request)
}
```

This particular one is contrived! But in practice it's possible today to do this and share a RouteDatabase.

yschimke:
Yeah, it's no doubt written that way in some number of clients.  That isn't great example code however, so I'm less swayed by that example, it could be rewritten.

There are generally 2 observable effects at play here 

1) Memory/GC size 
- forever 
- scoped to Client (and implicitly SSLEngine lifecycle) 
- 128 requests (not 128 SSLEngine instances)
2) Repeated timeout (making request on bad route) 
- after forced successful reattempt only 
- forced reattempt plus for each new client instances 
- forced reattempt plus after 129 requests.

Which behaviours are actually desireable?

yschimke:
On this PR, I also need to reset when the connection pool changes to retain existing observed behaviour.

yschimke:
BTW the reason, I think I'm arguing for this PR approach (your suggestion) is that it actually solves the GC problem by getting the lifecycle of related objects lined up.  The 128 approach works, by making an additional functional change that happens to also avoid a memory leak.

But in your example above, if the sslSocketFactory changes on each request, we keep retaining them (memory leak) if we never actually get above 128 hosts.

okhttp/src/main/java/okhttp3/internal/connection/RealConnectionPool.kt
```diff
@@ -46,7 +44,7 @@ class RealConnectionPool(
   }
 
   private val connections = ArrayDeque<RealConnection>()
-  val routeDatabase = RouteDatabase()
+//  val routeDatabase = RouteDatabase()
```
swankjesse: nit: delete

swankjesse:
I completely agree that the right fix is in the scoping, and not to mitigate the leak by capping at 128. But I don’t like this particular band-aid because it feels a bit . . . arbitrary? Nulling out the route database is a surprising requirement of changing the SSLSocketFactory, and seems easy to regress. Your test helps a lot here.

It’s too bad there’s nothing built-in to help with this. Would love for Java to have ephemerons like .Net’s ConditionalWeakTable.

okhttp/src/main/java/okhttp3/internal/connection/RealConnectionPool.kt
swankjesse: fixed myself!

yschimke:
> I completely agree that the right fix is in the scoping, and not to mitigate the leak by capping at 128. But I don’t like this particular band-aid because it feels a bit . . . arbitrary? Nulling out the route database is a surprising requirement of changing the SSLSocketFactory, and seems easy to regress. Your test helps a lot here.
> 
> It’s too bad there’s nothing built-in to help with this. Would love for Java to have ephemerons like .Net’s ConditionalWeakTable.

Agreed, that's why I tried to reframe the problem into what would we want the observable effects be?

Do you want to think it over for a few days? See if you think of a better solution?  This is already better than the WeakReference approach I put up to trigger the discussion.

Or just land this and test it well?

swankjesse:
Lets land this ’cause it’s good enough.

smallmenu:
Today, How to resolve the GC questions of RouteDatabases?   #4689  #5746

yschimke:
Currently this is considered closed because we don't hold the huge sslcontext object if you make a large number of temporary clients. 

If you have a very large number of distinct addresses then it could still be an issue but we could presumably hold a lot if it's just the proxy address that differs.

If you are still concerned could you share more details about your usage. Napkin calculations of number of addresses, how you create clients and measure the retained size?

smallmenu:
> Currently this is considered closed because we don't hold the huge sslcontext object if you make a large number of temporary clients.
> 
> If you have a very large number of distinct addresses then it could still be an issue but we could presumably hold a lot if it's just the proxy address that differs.
> 
> If you are still concerned could you share more details about your usage. Napkin calculations of number of addresses, how you create clients and measure the retained size?

We use okhttp as a http client tool for a spider project，in this project we will crawl a large number of websites(domains) or apps.
Sometimes we use many proxys in order to solve the anti-spider problem. 
So we have the GC problem with the application runing a long time。

`
private static final ConnectionPool DEFAULT_CONNECTION_POOL = new ConnectionPool(5, 15, TimeUnit.SECONDS);

private static final OkHttpClient DEFAULT_CLIENT = new OkHttpClient();

public static OkHttpClient client() {
        return builder(DEFAULT_CONNECTION_POOL, 0).build();
    }

public static OkHttpClient.Builder builder(ConnectionPool connectionPool, long timeoutSecond, OkHttpProxy userProxy) {

        connectionPool = connectionPool != null ? connectionPool : DEFAULT_CONNECTION_POOL;
        long timeout = timeoutSecond > 0 ? timeoutSecond : DEFAULT_TIMEOUT;

        HttpsUtils.SSLParams sslSocketFactory = HttpsUtils.getSslSocketFactory(null, null, null);

        OkHttpClient.Builder builder = DEFAULT_CLIENT.newBuilder();

        if (null != userProxy) {
            if (StringUtils.isNotBlank(userProxy.getHost())) {
                builder.proxy(new Proxy(Proxy.Type.HTTP, new InetSocketAddress(userProxy.getHost(), userProxy.getPort())));
            }

            if (StringUtils.isNotBlank(userProxy.getUsername())) {
                Authenticator proxyAuthenticator = (route, r) -> {
                    String credential = Credentials.basic(userProxy.getUsername(), userProxy.getPassword());
                    return r.request().newBuilder()
                            .header("Proxy-Authorization", credential)
                            .build();
                };
                builder.proxyAuthenticator(proxyAuthenticator);
            }
        }

        builder.connectionPool(connectionPool);
        builder.sslSocketFactory(sslSocketFactory.sSLSocketFactory, sslSocketFactory.trustManager);
        builder.hostnameVerifier((hostname, session) -> true);
        builder.followSslRedirects(true);
        builder.followRedirects(true);
        builder.connectTimeout(timeout, TimeUnit.SECONDS);
        builder.readTimeout(timeout, TimeUnit.SECONDS);
        builder.writeTimeout(timeout, TimeUnit.SECONDS);
        builder.retryOnConnectionFailure(true);

        return builder;
    }
`

yschimke:
That's helpful.

What is the lifetime of your clients?  Do you hold onto them, do they get released?  Since they don't share a ssl context, when you release the client instance for each proxy config, you should get GC of the routes also.

smallmenu:
> That's helpful.
> 
> What is the lifetime of your clients? Do you hold onto them, do they get released? Since they don't share a ssl context, when you release the client instance for each proxy config, you should get GC of the routes also.

Usually，We start many threads with thread pool,  such as codes above，We use the static method  `.client()`  return a static Instance of OkHttpClient which is .newBuilder() and share ConnectionPool.  

Then it will run all the time。after send every request, we also close the response。 

er..  How and When to release the client instance?

Thanks.

yschimke:
Releasing the client in this case, just means throwing it away, the RouteDB should get GCed with it.

If you only create one client and reuse then that is _usually_ ideal, and if you can send us some memory profiling it would help understand your usage.  You might even be able to put your proxy selection logic, behind a single ProxySelector implementation.  

If you are creating a lot of new clients by calling client() repeatedly, then that is likely the bigger memory problem, not the RouteDB.  Each client will hold a heavyweight instance of SSLContext etc.

This is a better question for stackoverflow as it isn't clear that there is a bug to fix. But we'd need a clear reproduction of the bug you are seeing that is representative of your use to treat this as a bug.  To be clear, I'm not saying what we are doing is ideal, just that we haven't got tuneable knobs for everything, and it's hard to make a sensible default e.g. capping RouteDB at 128 entries.

smallmenu:
Thanks.
 I try to write more demos for testing.

Issue #5760, opened 2020-01-30
OkHttp 4.3.1 spams log, didn't find class "org.conscrypt.ConscryptHostnameVerifier" on path


jamesonwilliams:
A trivial use of OkHttp 4.3.1 causes some spam in the device log.

https://gist.github.com/jamesonwilliams/8ab7237083467ed3cb8680115a3e3829

This code:
```
disposable = Single.fromCallable(() ->
    new OkHttpClient().newCall(new Request.Builder()
        .url("https://raw.githubusercontent.com/square/okhttp/master/README.md")
        .get()
        .build())
    .execute()
    .body()
    .string()
)
.subscribeOn(Schedulers.io())
.observeOn(AndroidSchedulers.mainThread())
.subscribe(
    textView::setText,
    failure -> textView.setText(Log.getStackTraceString(failure))
);
```
With this OkHttp:
```
implementation 'com.squareup.okhttp3:okhttp:4.3.1'
```

Causes this spam, as the out-of-the-box default:
```
I: Rejecting re-init on previously-failed class java.lang.Class<okhttp3.internal.platform.ConscryptPlatform$configureTrustManager$1>: java.lang.NoClassDefFoundError: Failed resolution of: Lorg/conscrypt/ConscryptHostnameVerifier;
I:     at okhttp3.internal.platform.android.SocketAdapter okhttp3.internal.platform.android.ConscryptSocketAdapter$Companion.buildIfSupported() (ConscryptSocketAdapter.kt:62)
I:     at void okhttp3.internal.platform.AndroidPlatform.<init>() (AndroidPlatform.kt:44)
I:     at okhttp3.internal.platform.Platform okhttp3.internal.platform.AndroidPlatform$Companion.buildIfSupported() (AndroidPlatform.kt:212)
I:     at okhttp3.internal.platform.Platform okhttp3.internal.platform.Platform$Companion.findPlatform() (Platform.kt:207)
I:     at okhttp3.internal.platform.Platform okhttp3.internal.platform.Platform$Companion.access$findPlatform(okhttp3.internal.platform.Platform$Companion) (Platform.kt:169)
I:     at void okhttp3.internal.platform.Platform.<clinit>() (Platform.kt:170)
I:     at void okhttp3.OkHttpClient.<init>(okhttp3.OkHttpClient$Builder) (OkHttpClient.kt:220)
I:     at void okhttp3.OkHttpClient.<init>() (OkHttpClient.kt:212)
[...]

PR #5763, opened 2020-01-31
Handle NoClassDefFoundError for Conscrypt

yschimke:


yschimke:
I have low confidence on this, but if we are spamming it's likely because we are retrying this?

https://github.com/square/okhttp/issues/5760

Issue #5761, opened 2020-01-30
Cache returns stale response body data on when ETag is the same (Windows only)


bitwiseman:
## Environment
OkHttp 4.2.2 
Only happens on Windows.  On linux and MacOS, the correct (expected) behavior occurs.

## Description 
On Windows, the OkHttp Cache does not update response body data for a request when a later response has the same ETag as a previous response.  

**Steps**
1. Request a URL /a/b
    * Response Header: 404, no ETag
    * Response Body: '{ "message": "Not Found" }'
    * Cache Updated
2. Request a URL /a/b with "If-Modified-Since" header 
    * Response Header: 304, ETag "C"
    * Cache automatically converts - client sees:
        * Response Header: 404, ETag "C"
        * Response Body: '{ "message": "Not Found" }'
3. Request a URL /a/b with "Cache-Control: no-cache" to force retrieval of new data
    * Response Header: 200, ETag "C"
    * Response Body:  '{ data: "..." }'
    * NOTE: Yes, that is the same ETag with different data - not good, but cache should still work
    * Cache Updated? (I think this is where the problem starts).
4. Request a URL /a/b with "If-None-Match" equal to ETag "C"

**Expected Result**
* Response Header: 304, ETag "C"
* Cache automatically converts - client sees:
    * Response Header: 200, ETag "C"
    * Response Body:  '{ data: "..." }'

**Actual Result**
* Response Header: 304, ETag "C"
* Cache automatically converts - client sees:
    * Response Header: 200, ETag "C"
    * Response Body: '{ "message": "Not Found" }'

## Automated Test
I know it is not great to point to a test on an external project but the test fully automated and will repro the issue every time: 
https://github.com/github-api/github-api/blob/967512629821b706e8d06ef9b3e57061f397d073/src/test/java/org/kohsuke/github/extras/okhttp3/GitHubCachingTest.java#L59

To run automated repro:
1. clone the repository on Windows with JDK 8+
2. remove the `assumeFalse()` call
3. run `mvn install`

## Notes
This issue was found and can be reproduced using https://github.com/github-api/github-api .
It is exposed due to a bug in GitHub server responses (see https://github.com/github-api/github-api/issues/669) but it is still a bug in this OkHttp Cache behavior on Windows.

swankjesse:
The bug is that in step 3 the cache should be updated with the new response, but it is not being updated?

bitwiseman:
@swankjesse 
Yes, I believe the bug is in step 3. 
[...]

PR #5764, opened 2020-02-01
Windows test for #5761

bitwiseman:
@swankjesse  
I think this will reproduce #5761, however I don't run windows locally so it might not.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,5 +1,8 @@
 version: 2.1
 
+orbs:
+  windows: circleci/windows@2.4.0
+
 commands:
   runtests:
     description: "Run tests"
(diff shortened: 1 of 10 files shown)


bitwiseman:
Switch testjdk11 to windows.  There are a number failures (DiskLruCacheTest), but not my test which is in CacheTest.  😢

swankjesse:
@bitwiseman does it fail locally on Windows?

bitwiseman:
@swankjesse 
I don't have a local windows machine. I'm working blind. Sorry.  
Separate question, why isn't there already a Windows instance in the CI system?

swankjesse:
We’ve never had a Windows-specific bug!

bitwiseman:
@swankjesse  
Perhaps no one has ever filed one, but you have failing tests - see https://github.com/square/okhttp/pull/5769 .  😄

bitwiseman:
Rebased this change on top of #5769.

yschimke:
Hasn't moved since February, n.b. I have Windows tests running successfully, so if we re-open it should be with CI changes separately from any test changes.

bitwiseman:
@yschimke  Yeah, I'm pretty sure this was superseded but other PRs including a fix.

PR #5765, opened 2020-02-02
Fold together RealCall and Transmitter

swankjesse:
These two were 1:1 with each other and there were a few places in the code
where we had both. Putting them into the same class creates a class that
does has a lot of responsibilities, but I believe it's simpler overall than
having two classes.


okhttp-sse/src/main/java/okhttp3/internal/sse/RealEventSource.kt
```diff
@@ -62,7 +62,7 @@ class RealEventSource(
       }
 
       // This is a long-lived response. Cancel full-call timeouts.
-      Exchange.get(response)?.timeoutEarlyExit()
+      call.timeoutEarlyExit()
```
yschimke: nice!

okhttp/src/main/java/okhttp3/internal/connection/Exchange.kt
```diff
@@ -162,12 +156,12 @@ class Exchange(
    */
   fun detachWithViolence() {
     codec.cancel()
-    transmitter.exchangeMessageDone(this, true, true, null)
+    call.exchangeMessageDone(this, requestDone = true, responseDone = true, e = null)
```
yschimke: love kotlin named params...

PR #5767, opened 2020-02-02
Reorganize RealCall and RealInterceptorChain

swankjesse:
I've attempted to put RealCall's methods into a logical order; it's approximately
the order the methods get called in for a successful call.

For RealInterceptorChain I've created a copy() method inspired-by data classes.
This has the added benefit of shrinking the call stack everywhere.

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/http/RealInterceptorChain.kt
+++ b/okhttp/src/main/java/okhttp3/internal/http/RealInterceptorChain.kt
@@ -17,6 +17,7 @@ package okhttp3.internal.http
 
 import java.io.IOException
 import java.util.concurrent.TimeUnit
+import okhttp3.Call
 import okhttp3.Connection
 import okhttp3.Interceptor
 import okhttp3.Request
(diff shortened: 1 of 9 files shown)


okhttp/src/main/java/okhttp3/internal/http/RealInterceptorChain.kt
```diff
@@ -33,82 +34,76 @@ import okhttp3.internal.connection.RealCall
...
 
-    // If we already have an exchange, confirm that the incoming request will use it.
-    check(this.exchange == null || this.exchange.connection()!!.supportsUrl(request.url)) {
-      "network interceptor ${interceptors[index - 1]} must retain the same host and port"
-    }
-
-    // If we already have an exchange, confirm that this is the only call to chain.proceed().
-    check(this.exchange == null || calls <= 1) {
-      "network interceptor ${interceptors[index - 1]} must call proceed() exactly once"
+    if (exchange != null) {
```
yschimke: For documentation purposes might be worth extracting this condition into an "isNetworkInterceptor" condition

okhttp/src/main/java/okhttp3/internal/http/RealInterceptorChain.kt
```diff
@@ -33,82 +34,76 @@ import okhttp3.internal.connection.RealCall
...
+  internal val connectTimeoutMillis: Int,
+  internal val readTimeoutMillis: Int,
+  internal val writeTimeoutMillis: Int
 ) : Interceptor.Chain {
 
   private var calls: Int = 0
 
-  override fun connection(): Connection? = exchange?.connection()
+  internal fun copy(
+    exchange: Exchange? = this.exchange,
```
yschimke: nit: why reorder these two?

okhttp/src/main/java/okhttp3/internal/http/RealInterceptorChain.kt
swankjesse: I like it. But I tried it, and then it makes dereferencing `exchange` more difficult because Kotlin can’t infer it’s non-null.

okhttp/src/main/java/okhttp3/internal/http/RealInterceptorChain.kt
swankjesse: total oversight. Fixed!

okhttp/src/main/java/okhttp3/internal/http/RealInterceptorChain.kt
yschimke: We can fix when contracts stabilise

Issue #5424, opened 2019-09-06
Handle HTTP 421 (Misdirected Request)


benweet:
HTTP/2 compliant servers may return a 421 code (Misdirected Request) if the client tries to multiplex HTTP requests addressing different sub-domains into one TLS session. In that particular case, the client should handle the error by opening another TLS connection and retrying the HTTP request.

See : https://trac.nginx.org/nginx/ticket/1252

Note : this issue occurs on Kubernetes using an nginx ingress controller configured with a wildcard certificate.

PR #5775, opened 2020-02-09
Formalize the lifecycle of Exchange

swankjesse:
There's four useful events:

1. When we're ready to call network interceptors. At this point
   we expect an exchange to exist, as a handle to the connection
   we've chosen for the network call

2. When we hold an exchange that must later be released.

3. When we release the exchange

4. When we're done calling network interceptors. Only at this point
   do we no longer need to remember what the exchange was and what
   connection it used.

In practice 1 and 2 happen mostly at the same time. Right now 1 is
in RetryAndFollowUpInterceptor and 2 is in ConnectInterceptor, but
we should be able to move those around without pain.

But we have a problem with 3 and 4. If the response has no response
body, then 3 will precede 4. We will release the exchange as soon as
we know there's no streams held. This means that network interceptors,
including our own RetryAndFollowUpInterceptor, do not consistently
have an exchange after they've called proceed(). This is problematic
if they want to make decisions based on what connection was used.

I've added a new member, 'interceptorScopedExchange' that holds the
exchange from 1 and 2 through 4, regardless of when 3 happens.

This is all working towards a solution to handling HTTP 421s on
coalesced connections.

https://github.com/square/okhttp/issues/5424

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/connection/RealCall.kt
+++ b/okhttp/src/main/java/okhttp3/internal/connection/RealCall.kt
@@ -39,7 +39,6 @@ import okhttp3.Response
 import okhttp3.internal.assertThreadDoesntHoldLock
 import okhttp3.internal.assertThreadHoldsLock
 import okhttp3.internal.cache.CacheInterceptor
-import okhttp3.internal.canReuseConnectionFor
 import okhttp3.internal.closeQuietly
 import okhttp3.internal.http.BridgeInterceptor
 import okhttp3.internal.http.CallServerInterceptor
@@ -82,9 +81,6 @@ class RealCall(
   /** Finds an exchange to send the next request and receive the next response. */
   private var exchangeFinder: ExchangeFinder? = null
 
-  /** The request that specified the current [exchangeFinder]. */
-  private var exchangeFinderRequest: Request? = null
-
   // Guarded by connectionPool.
   var connection: RealConnection? = null
   private var exchange: Exchange? = null
@@ -96,6 +92,14 @@ class RealCall(
 
   // Guarded by this.
   private var executed = false
+
+  /**
+   * This is the same value as [exchange], but scoped to the execution of the network interceptors.
+   * The [exchange] field is assigned to null when its streams end, which may be before or after the
+   * network interceptors return.
+   */
+  internal var interceptorScopedExchange: Exchange? = null
+    private set
 
   override fun timeout() = timeout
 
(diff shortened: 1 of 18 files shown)


okhttp/src/main/java/okhttp3/internal/http/RetryAndFollowUpInterceptor.kt
```diff
@@ -59,71 +59,68 @@ class RetryAndFollowUpInterceptor(private val client: OkHttpClient) : Intercepto
...
+        if (call.isCanceled()) {
+          throw IOException("Canceled")
         }
-        continue
-      } catch (e: IOException) {
-        // An attempt to communicate with a server failed. The request may have been sent.
-        val requestSendStarted = e !is ConnectionShutdownException
-        if (!recover(e, call, requestSendStarted, request)) throw e
-        continue
-      } finally {
```
swankjesse: This finally clause is moved to the bottom of the while loop!

okhttp/src/main/java/okhttp3/internal/http/RetryAndFollowUpInterceptor.kt
```diff
@@ -59,71 +59,68 @@ class RetryAndFollowUpInterceptor(private val client: OkHttpClient) : Intercepto
...
-
-      if (call.isCanceled()) {
-        throw IOException("Canceled")
-      }
+      call.enterNetworkInterceptorExchange(request)
 
       var response: Response
-      var success = false
+      var closeActiveExchange = true
       try {
```
swankjesse: This diff goes from `try {} catch {} finally {}` to `try { try {} catch } finally {}`. It is easier to see with whitespace changes hidden.

okhttp/src/main/java/okhttp3/internal/connection/RealCall.kt
```diff
@@ -235,17 +239,15 @@ class RealCall(
...
   /** Finds a new or pooled connection to carry a forthcoming request and response. */
-  internal fun prepareNewExchange(chain: RealInterceptorChain): Exchange {
+  internal fun initExchange(chain: RealInterceptorChain): Exchange {
     synchronized(connectionPool) {
       check(!noMoreExchanges) { "released" }
-      check(exchange == null) {
-        "cannot make a new request because the previous response is still open: " +
-            "please call response.close()"
-      }
+      check(exchange == null)
```
yschimke: why remove the explaining message?  Are we sure this will be caught in previous location?

okhttp/src/main/java/okhttp3/internal/http/RetryAndFollowUpInterceptor.kt
```diff
@@ -59,71 +59,68 @@ class RetryAndFollowUpInterceptor(private val client: OkHttpClient) : Intercepto
...
+        if (++followUpCount > MAX_FOLLOW_UPS) {
+          throw ProtocolException("Too many follow-up requests: $followUpCount")
+        }
 
-      request = followUp
-      priorResponse = response
+        request = followUp
+        priorResponse = response
+      } finally {
+        call.exitNetworkIntercptorExchange(closeActiveExchange)
```
yschimke: typo

okhttp/src/main/java/okhttp3/internal/http1/Http1ExchangeCodec.kt
```diff
@@ -65,18 +65,18 @@ class Http1ExchangeCodec(
...
   private val source: BufferedSource,
   private val sink: BufferedSink
 ) : ExchangeCodec {
   private var state = STATE_IDLE
   private var headerLimit = HEADER_LIMIT.toLong()
 
-  private fun Response.isChunked() =
-      "chunked".equals(header("Transfer-Encoding"), ignoreCase = true)
+  private val Response.isChunked: Boolean
+      get() = "chunked".equals(header("Transfer-Encoding"), ignoreCase = true)
```
yschimke: Separate from this review, but should this be 

```
headers("Transfer-Encoding").any { "chunked".equals(it, ignoreCase = true) }
```

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Transfer-Encoding

> // Several values can be listed, separated by a comma
> Transfer-Encoding: gzip, chunked

okhttp/src/main/java/okhttp3/internal/connection/RealCall.kt
swankjesse: Yeah. By promoting the other check to occur unconditionally this one becomes more of an assert.

okhttp/src/main/java/okhttp3/internal/http/RetryAndFollowUpInterceptor.kt
swankjesse: Fixed!

okhttp/src/main/java/okhttp3/internal/http1/Http1ExchangeCodec.kt
swankjesse: Ooooh, yeah that’s tricky.

PR #5785, opened 2020-02-15
Limit recovery attempts for REFUSED_STREAM errors

swankjesse:
We limit per-connection retries but not per-call retries, so this was
creating large numbers of connections each of which called the server
and accepted yet another REFUSED_STREAM.

Instead we fail sooner with a StreamResetException.

This shows that the ExchangeFinder interface is still somewhat inadequate
to support all of the use cases we have.

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/connection/ExchangeFinder.kt
+++ b/okhttp/src/main/java/okhttp3/internal/connection/ExchangeFinder.kt
@@ -28,6 +28,9 @@ import okhttp3.internal.canReuseConnectionFor
 import okhttp3.internal.closeQuietly
 import okhttp3.internal.http.ExchangeCodec
 import okhttp3.internal.http.RealInterceptorChain
+import okhttp3.internal.http2.ConnectionShutdownException
+import okhttp3.internal.http2.ErrorCode
+import okhttp3.internal.http2.StreamResetException
 
 /**
  * Attempts to find the connections for a sequence of exchanges. This uses the following strategies:
@@ -60,7 +63,9 @@ class ExchangeFinder(
   private val routeSelector: RouteSelector = RouteSelector(
       address, call.client.routeDatabase, call, eventListener)
   private var connectingConnection: RealConnection? = null
-  private var hasStreamFailure = false
+  private var refusedStreamCount = 0
+  private var connectionShutdownCount = 0
+  private var otherFailureCount = 0
   private var nextRouteToTry: Route? = null
 
   fun find(
@@ -78,10 +83,10 @@ class ExchangeFinder(
       )
       return resultConnection.newCodec(client, chain)
     } catch (e: RouteException) {
-      trackFailure()
+      trackFailure(e.lastConnectException)
       throw e
     } catch (e: IOException) {
-      trackFailure()
+      trackFailure(e)
       throw RouteException(e)
     }
   }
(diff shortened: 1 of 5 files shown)


okhttp/src/main/java/okhttp3/internal/connection/RealConnection.kt
```diff
@@ -690,7 +690,7 @@ class RealConnection(
       if (e is StreamResetException) {
         when (e.errorCode) {
           ErrorCode.REFUSED_STREAM -> {
-            // Retry REFUSED_STREAM errors once on the same connection.
+            // Stop using this connection after one REFUSED_STREAM error.
             refusedStreamCount++
             if (refusedStreamCount > 1) {
```
swankjesse: The ExchangeFinder counts now mirror similar counts on RealConnection. We need to track per-connection failures and per-exchange failures separately!

swankjesse:
Ick, needs some internal refactoring to pass tests.

okhttp/src/main/java/okhttp3/internal/connection/RealConnection.kt
```diff
@@ -690,7 +690,7 @@ class RealConnection(
       if (e is StreamResetException) {
         when (e.errorCode) {
           ErrorCode.REFUSED_STREAM -> {
-            // Retry REFUSED_STREAM errors once on the same connection.
+            // Stop using this connection after one REFUSED_STREAM error.
```
yschimke: two? logic seems to suggest, one was ok.

okhttp/src/main/java/okhttp3/internal/connection/ExchangeFinder.kt
```diff
@@ -60,7 +63,9 @@ class ExchangeFinder(
   private val routeSelector: RouteSelector = RouteSelector(
       address, call.client.routeDatabase, call, eventListener)
   private var connectingConnection: RealConnection? = null
-  private var hasStreamFailure = false
+  private var refusedStreamCount = 0
```
yschimke: What do you think about pulling these out into a separate class.  Might help land it e.g. with a simple legacy strategy and the correct ones that you can unit test for?

yschimke: I think I'm skewed here because of frameworks I've used at scale, that were not simple binary healthy/unhealthy had more distinct abstractions here. For a mobile client where biggest change is likely network/route related, and server is most likely hidden behind wildcard host and a reverse proxy to individual servers, maybe this is fine.

Finagle https://github.com/twitter/finagle/blob/develop/finagle-core/src/main/scala/com/twitter/finagle/liveness/FailureAccrualPolicy.scala

RSocket https://github.com/rsocket/rsocket-java/blob/21c39bdaa60c6a916e3288048b268f2b756dd5e7/rsocket-core/src/main/java/io/rsocket/Availability.java

swankjesse: I think for OkHttp we want to mostly track facts like stream failure counts, and let a user-defined class make the decisions.

swankjesse: Conversation continues here: https://github.com/square/okhttp/pull/5787/files

PR #5786, opened 2020-02-16
Don't do proxy selection eagerly if we're pooling

swankjesse:
If we're going to end up using a pooled connection anyway we
should defer proxy selection until we need it.

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/connection/ExchangeFinder.kt
+++ b/okhttp/src/main/java/okhttp3/internal/connection/ExchangeFinder.kt
@@ -57,8 +57,7 @@ class ExchangeFinder(
   private var routeSelection: RouteSelector.Selection? = null
 
   // State guarded by connectionPool.
-  private val routeSelector: RouteSelector = RouteSelector(
-      address, call.client.routeDatabase, call, eventListener)
+  private var routeSelector: RouteSelector? = null
   private var connectingConnection: RealConnection? = null
   private var hasStreamFailure = false
   private var nextRouteToTry: Route? = null
@@ -188,8 +187,13 @@ class ExchangeFinder(
     // If we need a route selection, make one. This is a blocking operation.
     var newRouteSelection = false
     if (selectedRoute == null && (routeSelection == null || !routeSelection!!.hasNext())) {
+      var localRouteSelector = routeSelector
+      if (localRouteSelector == null) {
+        localRouteSelector = RouteSelector(address, call.client.routeDatabase, call, eventListener)
+        this.routeSelector = localRouteSelector
+      }
       newRouteSelection = true
-      routeSelection = routeSelector.next()
+      routeSelection = localRouteSelector.next()
     }
 
     var routes: List<Route>? = null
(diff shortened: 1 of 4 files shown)


okhttp/src/test/java/okhttp3/EventListenerTest.java
```diff
@@ -1258,9 +1254,7 @@ private void timeToFirstByte() throws IOException {
 
     // Confirm the events occur when expected.
     listener.takeEvent(CallStart.class, 0L);
-    listener.takeEvent(ProxySelectStart.class, applicationInterceptorDelay);
-    listener.takeEvent(ProxySelectEnd.class, 0L);
-    listener.takeEvent(ConnectionAcquired.class, 0L);
+    listener.takeEvent(ConnectionAcquired.class, applicationInterceptorDelay);
```
yschimke: That's a lovely change.

okhttp/src/main/java/okhttp3/internal/connection/RealConnection.kt
```diff
@@ -690,7 +690,7 @@ class RealConnection(
       if (e is StreamResetException) {
         when (e.errorCode) {
           ErrorCode.REFUSED_STREAM -> {
-            // Retry REFUSED_STREAM errors once on the same connection.
+            // Stop using this connection after one REFUSED_STREAM error.
```
yschimke: two?

okhttp/src/main/java/okhttp3/internal/connection/ExchangeFinder.kt
```diff
@@ -287,12 +291,21 @@ class ExchangeFinder(
...
         nextRouteToTry = call.connection!!.route()
         return true
       }
-      return (routeSelection?.hasNext() ?: false) || routeSelector.hasNext()
+
+      // If we have a routes left, use 'em.
+      if (routeSelection?.hasNext() == true) return true
+
+      // If we haven't initialized the route selector yet, assume it'll have at least one route.
+      val localRouteSelector = routeSelector ?: return true
```
yschimke: Are these the lines that defer proxy lookup?

okhttp/src/main/java/okhttp3/internal/connection/RealConnection.kt
swankjesse: Yeah, that doesn’t read as well as I wanted. Fixed:

```
            // Stop using this connection on the 2nd REFUSED_STREAM error.
```

okhttp/src/main/java/okhttp3/internal/connection/ExchangeFinder.kt
swankjesse: You got it.

swankjesse: Nope, I read this wrong. Deferring the proxy lookup happens by deferring RouteSelector construction.

okhttp/src/main/java/okhttp3/internal/connection/ExchangeFinder.kt
```diff
@@ -57,8 +57,7 @@ class ExchangeFinder(
   private var routeSelection: RouteSelector.Selection? = null
 
   // State guarded by connectionPool.
-  private val routeSelector: RouteSelector = RouteSelector(
-      address, call.client.routeDatabase, call, eventListener)
```
swankjesse: Deferring proxy selection happens here.

yschimke:
@swankjesse  I assume if we don't find exact match then we still do the DNS check before reusing a pooled connection via connection coalescing rules?

swankjesse:
I got the branch name wrong because we were already deferring DNS. Here's the new sequence:

 1. Look in the pool
 2. Do proxy selection and the first DNS lookup
 3. Look in the pool again. This might hit because we have an IP address
 4. Connect to the resolved IP address
 5. Look in the pool the last time to see if we raced

It's a bit obnoxious that we look in the pool 3 times for unpooled connections! But this is how we maximize the pool’s use while minimizing network activity.

PR #5787, opened 2020-02-16
Limit recovery attempts for REFUSED_STREAM errors

swankjesse:
We limit per-connection retries but not per-call retries, so this was
creating large numbers of connections each of which called the server
and accepted yet another REFUSED_STREAM.

Instead we fail sooner with a StreamResetException.

This shows that the ExchangeFinder interface is still somewhat inadequate
to support all of the use cases we have.

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/connection/ExchangeFinder.kt
+++ b/okhttp/src/main/java/okhttp3/internal/connection/ExchangeFinder.kt
@@ -19,7 +19,6 @@ import java.io.IOException
 import java.net.Socket
 import okhttp3.Address
 import okhttp3.EventListener
-import okhttp3.HttpUrl
 import okhttp3.OkHttpClient
 import okhttp3.Route
 import okhttp3.internal.assertThreadDoesntHoldLock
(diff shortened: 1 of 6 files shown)


okhttp/src/main/java/okhttp3/internal/connection/ExchangeFinder.kt
```diff
@@ -59,7 +62,9 @@ class ExchangeFinder(
   // State guarded by connectionPool.
   private var routeSelector: RouteSelector? = null
   private var connectingConnection: RealConnection? = null
-  private var hasStreamFailure = false
+  private var refusedStreamCount = 0
```
swankjesse: From [PR 5785](https://github.com/square/okhttp/pull/5785):
> What do you think about pulling these out into a separate class. Might help land it e.g. with a simple legacy strategy and the correct ones that you can unit test for?

swankjesse: Love that idea. What I’d like to move towards is giving users the ability to define their own connection pooling policies including load balancing, seeding the connection pool, etc.

okhttp/src/main/java/okhttp3/internal/http/RetryAndFollowUpInterceptor.kt
```diff
@@ -59,8 +59,9 @@ class RetryAndFollowUpInterceptor(private val client: OkHttpClient) : Intercepto
     val call = realChain.call
     var followUpCount = 0
     var priorResponse: Response? = null
+    var newExchangeFinder = true
```
swankjesse: Here’s a potentially consequential behavior change.

We used to retain the same `ExchangeFinder` for all follow-up calls, as long as host+port stuff didn’t change. That means that a route failure for the initial request would impact an eventually-successful redirect or authorization.

With this change we retain only the `Connection`. We don’t remember previous failed routes.

I think it’s both simpler and more robust.

(Though I’m not happy with the name ExchangeFinder, or even Exchange for that matter.)

yschimke:
Main signal is the integration tests continuing to pass.  Are you happy with our coverage here?

swankjesse:
Yeah, I think our tests here are quite thorough. I say that because in making this change I bumped into a lot of such tests!

Issue #5770, opened 2020-02-04
DefaultProxySelector: IllegalArgumentException: protocol = https host = null


drakeet:
Okhttp version: 4.3.1

When using okhttp request an URL `https://>`, it crashes:

```
Fatal Exception: java.lang.IllegalArgumentException: protocol = https host = null
       at sun.net.spi.DefaultProxySelector.select + 183(DefaultProxySelector.java:183)
       at okhttp3.internal.connection.RouteSelector.resetNextProxy + 103(RouteSelector.java:103)
       at okhttp3.internal.connection.RouteSelector.<init> + 53(RouteSelector.java:53)
       at okhttp3.internal.connection.ExchangeFinder.<init> + 61(ExchangeFinder.java:61)
       at okhttp3.internal.connection.Transmitter.prepareToConnect + 134(Transmitter.java:134)
       at okhttp3.internal.http.RetryAndFollowUpInterceptor.intercept + 62(RetryAndFollowUpInterceptor.java:62)
       at okhttp3.internal.http.RealInterceptorChain.proceed + 112(RealInterceptorChain.java:112)
       at okhttp3.internal.http.RealInterceptorChain.proceed + 87(RealInterceptorChain.java:87)
       at okhttp3.RealCall.getResponseWithInterceptorChain + 194(RealCall.java:194)
       at okhttp3.RealCall$AsyncCall.run + 138(RealCall.java:138)
       at java.util.concurrent.ThreadPoolExecutor.runWorker + 1167(ThreadPoolExecutor.java:1167)
       at java.util.concurrent.ThreadPoolExecutor$Worker.run + 641(ThreadPoolExecutor.java:641)
       at java.lang.Thread.run + 764(Thread.java:764)
```

swankjesse:
@drakeet what’s your host name? Does it contain `_` or any other special characters?

swankjesse:
This is a simple workaround if you don’t need the system’s proxy selector:

```
OkHttpClient client = new OkHttpClient.Builder()
    .proxy(Proxy.NO_PROXY)
    .build()
```

drakeet:
@swankjesse the URL is just `https://>`. I have created a hotfix by [SdkEditor](https://github.com/iwhys/sdk-editor-plugin):

![image](https://user-images.githubusercontent.com/5214214/74396915-60670000-4e4e-11ea-9f30-833789109c5f.png)

It works, and if it is acceptable, I can create a PR for this our Okhttp.

[...]

PR #5788, opened 2020-02-16
Don't crash on a null host

swankjesse:
Closes: https://github.com/square/okhttp/issues/5770

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/connection/RouteSelector.kt
+++ b/okhttp/src/main/java/okhttp3/internal/connection/RouteSelector.kt
@@ -94,19 +94,23 @@ class RouteSelector(
 
   /** Prepares the proxy servers to try. */
   private fun resetNextProxy(url: HttpUrl, proxy: Proxy?) {
-    eventListener.proxySelectStart(call, url)
-    proxies = if (proxy != null) {
+    fun selectProxies(): List<Proxy> {
       // If the user specifies a proxy, try that and only that.
-      listOf(proxy)
-    } else {
+      if (proxy != null) return listOf(proxy)
+
+      // If the URI lacks a host (as in "http://</"), don't call the ProxySelector.
+      val uri = url.toUri()
+      if (uri.host == null) return immutableListOf(Proxy.NO_PROXY)
+
       // Try each of the ProxySelector choices until one connection succeeds.
-      val proxiesOrNull = address.proxySelector.select(url.toUri())
-      if (proxiesOrNull != null && proxiesOrNull.isNotEmpty()) {
-        proxiesOrNull.toImmutableList()
-      } else {
-        immutableListOf(Proxy.NO_PROXY)
-      }
+      val proxiesOrNull = address.proxySelector.select(uri)
+      if (proxiesOrNull.isNullOrEmpty()) return immutableListOf(Proxy.NO_PROXY)
+
+      return proxiesOrNull.toImmutableList()
     }
+
+    eventListener.proxySelectStart(call, url)
+    proxies = selectProxies()
     nextProxyIndex = 0
     eventListener.proxySelectEnd(call, url, proxies)
   }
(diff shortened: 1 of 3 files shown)


okhttp/src/main/java/okhttp3/internal/connection/RouteSelector.kt
```diff
@@ -95,17 +95,19 @@ class RouteSelector(
...
-      } else {
-        immutableListOf(Proxy.NO_PROXY)
-      }
+      val proxiesOrNull = address.proxySelector.select(uri)
+      if (proxiesOrNull.isNullOrEmpty()) return@run immutableListOf(Proxy.NO_PROXY)
+
+      return@run proxiesOrNull.toImmutableList()
     }
     nextProxyIndex = 0
     eventListener.proxySelectEnd(call, url, proxies)
```
swankjesse: alternately:
```
 private fun resetNextProxy(url: HttpUrl, proxy: Proxy?) {
    eventListener.proxySelectStart(call, url)
    proxies = if (proxy != null) {
      listOf(proxy)
    } else {
      val uri = url.toUri()
      if (uri.host == null) {
        immutableListOf(Proxy.NO_PROXY)
      } else {
        val proxiesOrNull = address.proxySelector.select(uri)
        if (proxiesOrNull.isNullOrEmpty()) {
          immutableListOf(Proxy.NO_PROXY)
        } else {
          proxiesOrNull.toImmutableList()
        }
      }
    }
    nextProxyIndex = 0
    eventListener.proxySelectEnd(call, url, proxies)
  }
```

okhttp/src/main/java/okhttp3/internal/connection/RouteSelector.kt
```diff
@@ -95,17 +95,19 @@ class RouteSelector(
   /** Prepares the proxy servers to try. */
   private fun resetNextProxy(url: HttpUrl, proxy: Proxy?) {
     eventListener.proxySelectStart(call, url)
-    proxies = if (proxy != null) {
+    proxies = run {
```
yschimke: is it cleaner just to have an inline function than to use run and return@run?

swankjesse: Perhaps? I changed to use a local function.

swankjesse:
Completely agree that we’ve fallen into a bizarre invalid-fest!

In this particular code we merely need to stay alive long enough for the DNS request to return 0 results. Without this the default proxy selector crashes with an IllegalArgumentException, which is really bad for tools like crawlers that could encounter such URLs in the wild.

I don’t expect this particular modeling failure to push itself deeper into our code. Although HttpUrl is too lenient, it’s self-consistent and the only places we get hurt is when we take sketchy URLs like `https://>/` and integrate them into other systems: DNS, proxy selection, cookie stores, etc.

The other idea I had for `https://>/` was to replace its impossible hostname with a well-formed but unresolvable URL, Like `https://-/` or `https://0/`. Ultimately those also feel bad!

okhttp/src/test/java/okhttp3/HttpUrlTest.java
```diff
@@ -440,6 +440,13 @@ HttpUrl parse(String url) {
         .test(Component.HOST);
   }
 
+  /** This one's ugly: the HttpUrl's host is non-empty, but the URI's host is null. */
+  @Test public void hostContainsOnlyStrippedCharacters() throws Exception {
+    HttpUrl url = parse("http://>/");
+    assertThat(url.host()).isEqualTo(">");
```
monkey-mas: Should we add validation in `HttpUrl` that prevents the user from having an _invalid_ hostname?

swankjesse: Yes, but...

We’d need to define what a valid hostname is. That’s difficult, and lots of libraries that have attempted this have failed.

PR #5790, opened 2020-02-16
Add a canceled event to EventListener

swankjesse:


okhttp/src/main/java/okhttp3/EventListener.kt
```diff
@@ -401,6 +401,27 @@ abstract class EventListener {
   ) {
   }
 
+  /**
+   * Invoked when a call is canceled.
```
yschimke: Nice docs!

PR #5793, opened 2020-02-17
Use .use() writing metadata to the cache

swankjesse:


okhttp/src/main/java/okhttp3/Cache.kt
```diff
@@ -559,42 +557,42 @@ class Cache internal constructor(
 
     @Throws(IOException::class)
     fun writeTo(editor: DiskLruCache.Editor) {
-      val sink = editor.newSink(ENTRY_METADATA).buffer()
-      sink.writeUtf8(url).writeByte('\n'.toInt())
-      sink.writeUtf8(requestMethod).writeByte('\n'.toInt())
-      sink.writeDecimalLong(varyHeaders.size.toLong()).writeByte('\n'.toInt())
-      for (i in 0 until varyHeaders.size) {
-        sink.writeUtf8(varyHeaders.name(i))
+      editor.newSink(ENTRY_METADATA).buffer().use { sink ->
```
oldergod: neat

Issue #5547, opened 2019-10-10
Race condition causes retry loop?


swankjesse:
https://github.com/square/okhttp/pull/2179#issuecomment-539877107

> I found this patch may cause endless loop:
> 
> 1. on a mobile phone, user sends one request, in function StreamAllocation::findHealthyConnection(), it returns a candidate
> 2. if this candidate successCount is 0, it is returned without health check (candidate.isHealthy())
> 3. switch the phone's netowrk from 4G to wifi, so the socket is broken, the candidate is unhealthy.
> 4. to send request, SocketException(socket is closed) is thrown to RetryAndFollowUpInterceptor, and will have one more retry.
> 5. StreamAllocation::findHealthyConnection() returns the same candidate.
     ...
     so it causes endless loop.

yschimke:
My main interest is whether we would detect this with anyone using the LoggingEventListener?  Bugs happen, do we make it easy to report and diagnose after the fact?

BTW in practice, switching phone network from 4G to Wifi doesn't break the existing 4G connections on standard Android phones.  It's perfectly valid to choose between a 4G and Wifi network using interface binding on the socket. See [OkHttpAndroidApp/.../AndroidNetworkManager.kt](https://github.com/yschimke/OkHttpAndroidApp/blob/master/android/app/src/main/java/ee/schimke/okhttp/android/factory/AndroidNetworkManager.kt#L212)

tankjj:
> My main interest is whether we would detect this with anyone using the LoggingEventListener? Bugs happen, do we make it easy to report and diagnose after the fact?
> 
> BTW in practice, switching phone network from 4G to Wifi doesn't break the existing 4G connections on standard Android phones. It's perfectly valid to choose between a 4G and Wifi network using interface binding on the socket. See [OkHttpAndroidApp/.../AndroidNetworkManager.kt](https://github.com/yschimke/OkHttpAndroidApp/blob/master/android/app/src/main/java/ee/schimke/okhttp/android/factory/AndroidNetworkManager.kt#L212)

On my android phone(xiaomi MIX 2, OS is android 8.0), switch 4G to Wifi, it will break the connection, I checked the rawSocket of the connection, it is closed. If you do the health check, it'll also return false.

yschimke:
@tankjj thanks. I wonder if it’s Android 9 (Nokia 7.1) or it depends on whether you explicitly bind to a network?

tankjj:
[...]

PR #5795, opened 2020-02-17
Be smarter about connection health checks

swankjesse:
Previously we didn't do any health checks at all until a
connection completed a single exchange. This was awkward
for HTTP/2, which could have multiple exchanges in-flight
before any were health checked.

We were also doing the awkward and expensive read timeout
check on all non-GET requests on pooled connections. Now
we only perform these checks if the connection was idle
for 10 seconds or more.

Closes: https://github.com/square/okhttp/issues/5547

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/connection/RealConnection.kt
+++ b/okhttp/src/main/java/okhttp3/internal/connection/RealConnection.kt
@@ -25,7 +25,6 @@ import java.net.ProtocolException
 import java.net.Proxy
 import java.net.Socket
 import java.net.SocketException
-import java.net.SocketTimeoutException
 import java.net.UnknownServiceException
 import java.security.cert.X509Certificate
 import java.util.concurrent.TimeUnit.MILLISECONDS
@@ -59,6 +58,7 @@ import okhttp3.internal.http2.Http2ExchangeCodec
 import okhttp3.internal.http2.Http2Stream
 import okhttp3.internal.http2.Settings
 import okhttp3.internal.http2.StreamResetException
+import okhttp3.internal.isHealthy
 import okhttp3.internal.platform.Platform
 import okhttp3.internal.tls.OkHostnameVerifier
 import okhttp3.internal.toHostHeader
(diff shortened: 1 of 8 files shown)


okhttp/src/main/java/okhttp3/internal/connection/RealConnection.kt
```diff
@@ -736,6 +729,7 @@ class RealConnection(
   companion object {
     private const val NPE_THROW_WITH_NULL = "throw with null exception"
     private const val MAX_TUNNEL_ATTEMPTS = 21
+    private const val IDLE_CONNECTION_HEALTHY_NS = 1_000_000_000 // 10 seconds.
```
yschimke: Isn't this one second?

swankjesse: Endless screaming.

It is. I wanted to use const which took away TimeUnit.toNanos().

okhttp/src/main/java/okhttp3/internal/connection/RealConnection.kt
```diff
@@ -227,6 +227,8 @@ class RealConnection(
       throw RouteException(ProtocolException(
           "Too many tunnel connections attempted: $MAX_TUNNEL_ATTEMPTS"))
     }
+
+    idleAtNs = System.nanoTime()
```
yschimke: Assume this triggers the behaviour change.

swankjesse: Yep.

okhttp/src/test/java/okhttp3/ConnectionReuseTest.java
```diff
@@ -178,30 +189,6 @@
     assertThat(server.takeRequest().getSequenceNumber()).isEqualTo(0);
   }
 
-  @Test public void staleConnectionNotReusedForNonIdempotentRequest() throws Exception {
```
swankjesse: I moved this test to Kotlin so that I could manipulate `idleAtNs`.

Issue #5708, opened 2020-01-07
Some Web Sockets tests are flaky


swankjesse:
https://circleci.com/gh/square/okhttp/8330?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

closeReasonMaximumLength - okhttp3.internal.ws.WebSocketHttpTest
```
org.junit.ComparisonFailure: expected:<[0]> but was:<[1]>
	at okhttp3.OkHttpClientTestRule.ensureAllConnectionsReleased(OkHttpClientTestRule.kt:91)
	at okhttp3.OkHttpClientTestRule$apply$1.evaluate(OkHttpClientTestRule.kt:121)
```

serverCloseThenClientClose - okhttp3.internal.ws.RealWebSocketTest
```
java.lang.AssertionError: 
Expecting empty but was:<[Closed[1000 Hello!]]>
	at okhttp3.internal.ws.WebSocketRecorder.assertExhausted(WebSocketRecorder.java:163)
	at okhttp3.internal.ws.RealWebSocketTest.tearDown(RealWebSocketTest.java:58)
```

swankjesse:
These flaked on Java 13.

yschimke:
At the moment the websocket tests are 3 of the 4 flaky tests

https://circleci.com/build-insights/gh/square/okhttp/master

yschimke:
Dupe of https://github.com/square/okhttp/issues/6033, which covers more tests

PR #5800, opened 2020-02-18
Fixed flaky RealWebSocketTest.serverCloseThenClientClose().

dave-r12:
https://github.com/square/okhttp/issues/5708


dave-r12:
I was seeing this fail locally consistently with JDK13. This was flaky because the client called `close()` which enqueues a task to write the frame. Since this happens on a background thread it was hit or miss whether the `@After` check would find any events in the listener.

dave-r12:
Test failure seems unrelated:
```
okhttp3.internal.http2.HttpOverHttp2Test > concurrentRequestWithEmptyFlowControlWindow[h2] FAILED
    org.junit.ComparisonFailure: expected:<[0]> but was:<[1]>
```

Looking!

JakeWharton:
No good deed goes unpunished! Fix one, find another.

PR #5802, opened 2020-02-18
Prefer reporting a JUnit Timeout failure in HttpOverHttp2Test.

dave-r12:
I was reviewing a CI test failure reported here: https://circleci.com/gh/square/okhttp/9726

Reviewing the test case everything _seemed_ fine. I noticed that if I breakpoint on the last line of the test and wait 5 seconds it'll timeout. This is expected because we have a timeout rule. But the failure is reporting as `OkHttpClientTestRule.ensureAllConnectionsReleased()`. To me I'd like to see the Timeout failure. It feels like someone might start diving into something that doesn't actually exist. The timeout implies the test didn't complete and we will probably expect leaked connections! 😄 Tricky situation though.

I also made the timeout less aggressive (60 seconds instead of 5.) I think this means we prefer a slightly slower build that likely succeeds rather than a faster build with sporadic timeout failures.

Let me know what you think.

### DIFF
--- a/okhttp/src/test/java/okhttp3/internal/http2/HttpOverHttp2Test.java
+++ b/okhttp/src/test/java/okhttp3/internal/http2/HttpOverHttp2Test.java
@@ -113,8 +113,8 @@ public final class HttpOverHttp2Test {
   private final PlatformRule platform = new PlatformRule();
   private final OkHttpClientTestRule clientTestRule = new OkHttpClientTestRule();
   @Rule public final TestRule chain = RuleChain.outerRule(platform)
-      .around(clientTestRule)
-      .around(new Timeout(5, SECONDS));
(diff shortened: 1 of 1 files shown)


yschimke:
The original intent was that the client test rule will log events (after the fact) for failing tests, so you could debug a timeout exception.  If it's impeding understanding failures, then let's flip and I'll work out another approach for the former case.

dave-r12:
@yschimke cool. Can we get those events on CircleCI? I can't seem to get at the through the GUI.

yschimke:
It comes out for any @Flaky tests or those failing that use a client

e.g. https://9774-5152285-gh.circle-artifacts.com/0/~/reports/okhttp/build/reports/tests/test/classes/okhttp3.WholeOperationTimeoutTest.html

Issue #5816, opened 2020-02-20
ExceptionInitializerError with new Android 11 version (DP1)


arocha55:
OKHTTP 3.14.4
Retrofit 2.7.1
Device: Pixel 3 XL
Android Build: [crosshatch-rpp1.200123.016-factory-5808a1e6.zip](https://developer.android.com/preview/download)

**Note:** Issue is not reproducible on Android 10 or earlier versions.

Found this issue when running my sample app using latest version of Retrofit (2.7.1) and targeting newly released Android 11 (targetSdkVersion = 30).

When attempting to run the app in Android 11, the app immediately crashes with the following exception:

> 2020-02-20 14:26:08.386 22962-22962/com.my.example.retrofit2 E/AndroidRuntime: FATAL EXCEPTION: main
    Process: com.my.example.retrofit2, PID: 22962
    java.lang.ExceptionInInitializerError
        at okhttp3.internal.platform.Platform.get(Platform.java:85)
        at okhttp3.OkHttpClient.newSslSocketFactory(OkHttpClient.java:263)
        at okhttp3.OkHttpClient.<init>(OkHttpClient.java:229)
        at okhttp3.OkHttpClient.<init>(OkHttpClient.java:202)
        at retrofit2.Retrofit$Builder.build(Retrofit.java:614)
        at com.my.example.retrofit2.network.RetrofitInstance.getRetrofitInstance(RetrofitInstance.java:37)
        at com.my.example.retrofit2.main.IntractorImpl.getMarsPhotosArrayList(IntractorImpl.java:21)
        at com.my.example.retrofit2.main.MainPresenterImpl.requestDataFromServer(MainPresenterImpl.java:29)
        at com.my.example.retrofit2.main.MainActivity.onCreate(MainActivity.java:37)
        at android.app.Activity.performCreate(Activity.java:7970)
        at android.app.Activity.performCreate(Activity.java:7959)
        at android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1309)
        at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:3311)
        at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:3480)
        at android.app.servertransaction.LaunchActivityItem.execute(LaunchActivityItem.java:83)
        at android.app.servertransaction.TransactionExecutor.executeCallbacks(TransactionExecutor.java:135)
        at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:95)
[...]

PR #5817, opened 2020-02-21
Fixes for Android Test (including R)

yschimke:
In order to test https://github.com/square/okhttp/issues/5816, run against Android R.

### DIFF
--- a/android-test/build.gradle
+++ b/android-test/build.gradle
@@ -25,7 +25,7 @@ android {
     versionName "1.0"
 
     testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
-    testInstrumentationRunnerArguments(['notClass': 'org.conscrypt.KitKatPlatformOpenSSLSocketImplAdapter'])
+    testInstrumentationRunnerArguments(['notClass': 'org.conscrypt.KitKatPlatformOpenSSLSocketImplAdapter,org.bouncycastle.pqc.crypto.qtesla.QTeslaKeyEncodingTests'])
   }
 }
 
--- a/android-test/src/androidTest/java/okhttp/android/test/OkHttpTest.kt
+++ b/android-test/src/androidTest/java/okhttp/android/test/OkHttpTest.kt
@@ -310,7 +310,8 @@ class OkHttpTest {
     response.use {
       assertEquals(200, response.code)
       assertEquals(Protocol.HTTP_2, response.protocol)
-      assertEquals(TlsVersion.TLS_1_2, response.handshake?.tlsVersion)
+      val tlsVersion = response.handshake?.tlsVersion
+      assertTrue(tlsVersion == TlsVersion.TLS_1_2 || tlsVersion == TlsVersion.TLS_1_3)
       assertEquals("CN=localhost",
           (response.handshake!!.peerCertificates.first() as X509Certificate).subjectDN.name)
     }
@@ -359,6 +360,8 @@ class OkHttpTest {
 
   @Test
   fun testEventListener() {
+    client.connectionPool.evictAll()
+
     val eventListener = RecordingEventListener()
 
     enableTls()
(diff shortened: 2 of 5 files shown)


android-test/src/androidTest/java/okhttp/android/test/OkHttpTest.kt
```diff
@@ -386,7 +389,7 @@ class OkHttpTest {
       assertEquals(200, response.code)
     }
 
-    assertEquals(listOf("CallStart", "ProxySelectStart", "ProxySelectEnd",
+    assertEquals(listOf("CallStart",
```
yschimke: Fix for improved behaviour from a previous change

okhttp/src/main/java/okhttp3/internal/platform/AndroidPlatform.kt
```diff
@@ -200,6 +200,7 @@ class AndroidPlatform : Platform() {
 
     val isSupported: Boolean = when {
       !isAndroid -> false
+      Build.VERSION.SDK_INT >= 30 -> false // graylisted methods are banned
```
yschimke: I think the failure is not related to using AndroidPlatform instead of Android10Platform, but rather still possibly building AndroidPlatform on versions >= R

yschimke:
```
07:11:22 V/InstrumentationResultParser: Time: 3.217
07:11:22 V/InstrumentationResultParser:
07:11:22 V/InstrumentationResultParser: OK (16 tests)
07:11:22 V/InstrumentationResultParser:
07:11:22 V/InstrumentationResultParser:
07:11:22 V/InstrumentationResultParser: INSTRUMENTATION_CODE: -1
```

okhttp/src/main/java/okhttp3/internal/platform/AndroidPlatform.kt
yschimke: This shouldn't be needed but is correct IIUC

android-test/src/androidTest/java/okhttp/android/test/OkHttpTest.kt
```diff
@@ -310,7 +310,8 @@ class OkHttpTest {
     response.use {
       assertEquals(200, response.code)
       assertEquals(Protocol.HTTP_2, response.protocol)
-      assertEquals(TlsVersion.TLS_1_2, response.handshake?.tlsVersion)
+      val tlsVersion = response.handshake?.tlsVersion
+      assertTrue(tlsVersion == TlsVersion.TLS_1_2 || tlsVersion == TlsVersion.TLS_1_3)
```
swankjesse: 🎉

android-test/src/androidTest/java/okhttp/android/test/OkHttpTest.kt
```diff
@@ -359,6 +360,8 @@ class OkHttpTest {
 
   @Test
   fun testEventListener() {
+    client.connectionPool.evictAll()
```
swankjesse: Doesn’t the client rule do this?

okhttp/src/main/java/okhttp3/internal/platform/AndroidPlatform.kt
swankjesse: Could we replace this with a check?

```
check(Build.VERSION.SDK_INT < 30)
```

android-test/build.gradle
```diff
@@ -25,7 +25,7 @@ android {
     versionName "1.0"
 
     testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
-    testInstrumentationRunnerArguments(['notClass': 'org.conscrypt.KitKatPlatformOpenSSLSocketImplAdapter'])
+    testInstrumentationRunnerArguments(['notClass': 'org.conscrypt.KitKatPlatformOpenSSLSocketImplAdapter,org.bouncycastle.pqc.crypto.qtesla.QTeslaKeyEncodingTests'])
```
swankjesse: such a weird thing to have to defend against

yschimke: Yep.

It's in their bcprov Jar.  We depend on that strongly via okhttp-testing-support and then it get's picked up by the instrumentation runner.

okhttp/src/main/java/okhttp3/internal/platform/AndroidPlatform.kt
yschimke: A check seems wrong for an isSupported call.  So will avoid this.

PR #5702, opened 2020-01-05
Set hostname on Android 5.0

yschimke:
Fix for https://github.com/square/okhttp/issues/5700

Android 5.0 requires explicit call to set the hostname.

PR #5820, opened 2020-02-21
Handle Android 11 restrictions (3.14.x)

yschimke:
Splits Android 10 and Android Platform classes. SSLSockets was added in 10, but the existing code only breaks in 11.

Can't use much of the existing reflection code because it is either blacklisted, or available without reflection on known JDK types.


yschimke:
TODO test on device, there isn't an easy way to test this without a separate app build :(

okhttp/src/main/java/okhttp3/internal/platform/Android11Platform.java
```diff
@@ -0,0 +1,99 @@
...
+      }
+
+      return alpnResult;
+    } catch (IllegalAccessException | InvocationTargetException e) {
+      throw new AssertionError(e);
+    }
+  }
+
+  public static @Nullable Platform buildIfSupported() {
+    // Attempt to find Android 5+ APIs.
```
JakeWharton:
```suggestion
    // Attempt to find Android 11+ APIs.
```

yschimke:
I think since it is a branch I'll revert the Android platform parts of the change. And instead only add the R class which subclasses and overrides. 

Easier to show there is no other impact. 

This feels like a half assed bàckport of SocketAdapter change.

yschimke:
Tested on Android R

```

    OkHttpClient client = new OkHttpClient();

    Request request = new Request.Builder().url("https://api.twitter.com/robots.txt").build();
    Call call = client.newCall(request);

    System.out.println(response.body().string());

    System.out.println(Platform.get());
    System.out.println(response.handshake().tlsVersion());
    System.out.println(response.protocol());
    System.out.println(Build.VERSION.CODENAME);
```

Results

```
I/System.out: User-agent: *
    Disallow: /
    Android11Platform
    TLS_1_2
    h2
    R
```

yschimke:
Android 10

```
I/System.out: # Used for Google app indexing. See https://developers.google.com/app-indexing/webmasters/server
    User-agent: Googlebot
    Disallow:
    User-agent: *
    Disallow: /
    Android10Platform
    TLS_1_2
    h2
    29
```

yschimke:
Android 5.0 21

```
02-22 11:02:13.222 2655-2673/com.example.myapplication I/System.out: # Used for Google app indexing. See https://developers.google.com/app-indexing/webmasters/server
02-22 11:02:13.222 2655-2673/com.example.myapplication I/System.out: User-agent: Googlebot
02-22 11:02:13.222 2655-2673/com.example.myapplication I/System.out: Disallow:
02-22 11:02:13.222 2655-2673/com.example.myapplication I/System.out: User-agent: *
02-22 11:02:13.222 2655-2673/com.example.myapplication I/System.out: Disallow: /
02-22 11:02:13.222 2655-2673/com.example.myapplication I/System.out: AndroidPlatform
02-22 11:02:13.222 2655-2673/com.example.myapplication I/System.out: TLS_1_2
02-22 11:02:13.222 2655-2673/com.example.myapplication I/System.out: h2
02-22 11:02:13.222 2655-2673/com.example.myapplication I/System.out: 21
```

.circleci/config.yml
```diff
@@ -2,7 +2,7 @@ version: 2
 jobs:
   build:
     docker:
-      - image: circleci/openjdk:8u171-jdk
+      - image: circleci/openjdk:11.0.6-jdk-stretch
```
yschimke: @swankjesse Any concern with this an effect on you releasing?

swankjesse: Nope, Java 11 is right.

yschimke:
Before this lands I'll try to use non reflection

JakeWharton:
Just remember that API 29 is non-final so while it's probably safe to rely on these new APIs at this point (internal API freeze has happened), it's technically not stable yet and could change. It would be safer to ship reflection and switch to direct linking in a few months.

yschimke:
@JakeWharton Are you sure?

> Android 10 is the tenth major release and the 17th version of the Android mobile operating system. It was released on September 3, 2019.

The issue here isn't so much the new APIs, it's the fact that the legacy reflection APIs are not available. Even the JDK 9 methods look safe

```
    // BEGIN Android-added: Add ALPN-related methods from OpenJDK 9.
    // Also removed references to DTLS in documentation; Android doesn't support DTLS.
    /**
     * Returns the most recent application protocol value negotiated for this
     * connection.
     * <p>
     * If supported by the underlying SSL/TLS implementation,
     * application name negotiation mechanisms such as <a
     * href="http://www.ietf.org/rfc/rfc7301.txt"> RFC 7301 </a>, the
     * Application-Layer Protocol Negotiation (ALPN), can negotiate
     * application-level values between peers.
     * <p>
     * @implSpec
     * The implementation in this class throws
     * {@code UnsupportedOperationException} and performs no other action.
     *
     * @return null if it has not yet been determined if application
     *         protocols might be used for this connection, an empty
     *         {@code String} if application protocols values will not
     *         be used, or a non-empty application protocol {@code String}
     *         if a value was successfully negotiated.
     * @throws UnsupportedOperationException if the underlying provider
     *         does not implement the operation.
     * @since 9
     */
    public String getApplicationProtocol() {
        throw new UnsupportedOperationException();
    }
```

JakeWharton:
This PR is for 11

yschimke:
Yep - but it seems safe to fix for issues with 11, by using 10 APIs.

yschimke:
I've retested with built JARs in 5.0, 10 and 11 preview.  All working, so think remaining issue if build related when switching from a legacy android jar to robolectric android-all.

okhttp/src/main/java/okhttp3/internal/platform/Android10Platform.java
```diff
@@ -0,0 +1,80 @@
...
+@SuppressLint("NewApi")
+class Android10Platform extends AndroidPlatform {
+  Android10Platform(Class<?> sslParametersClass) {
+    super(sslParametersClass, null, null, null, null, null);
+  }
+
+  @SuppressLint("NewApi")
+  @IgnoreJRERequirement
+  @Override public void configureTlsExtensions(
+      SSLSocket sslSocket, String hostname, List<Protocol> protocols) {
```
swankjesse: Might be able to get rid of the hostname thing altogether in our master branch

okhttp/src/main/java/okhttp3/internal/platform/AndroidPlatform.java
```diff
@@ -421,4 +426,13 @@ static CloseGuard get() {
       throw new IllegalStateException("No TLS provider", e);
     }
   }
+
+  static int getSdkInt() {
+    try {
+      return Build.VERSION.SDK_INT;
+    } catch (NoClassDefFoundError ignored) {
+      // fails fatally against robolectric classes
```
swankjesse: Nice comment.

okhttp/src/main/java/okhttp3/internal/platform/Android10Platform.java
swankjesse: (no action to take on this PR)

okhttp/src/main/java/okhttp3/internal/platform/Android10Platform.java
```diff
@@ -0,0 +1,80 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okhttp3.internal.platform;
+
+import android.annotation.SuppressLint;
+import android.net.ssl.SSLSockets;
+import android.os.Build;
```
swankjesse: nit: unused

okhttp/src/main/java/okhttp3/internal/platform/Android10Platform.java
yschimke: It can't go completely.

Compare

Android 10 it's gone completely

https://github.com/square/okhttp/blob/ff0dee1c34fc06084357c93dc480f6ad89fef24a/okhttp/src/main/java/okhttp3/internal/platform/android/AndroidSocketAdapter.kt#L60

Android it's needed for 21-24.

https://github.com/square/okhttp/blob/ff0dee1c34fc06084357c93dc480f6ad89fef24a/okhttp/src/main/java/okhttp3/internal/platform/android/Android10SocketAdapter.kt#L49

yschimke: I was overeager last time and it resulted in you releasing 4.3.1

https://github.com/square/okhttp/pull/5702/files

PR #5525, opened 2019-10-02
[WIP] Use Android 10 API for session tickets

yschimke:

PR #5821, opened 2020-02-22
Upgrade to Android 10 APIs when available

yschimke:
Update to Robolectric and take opportunity to use more public API.


yschimke:
This is disappointing

```
The :okhttp:animalsnifferMain task failed.View task in console log
Failed to check signatures
>  failed to process jar /Users/yschimke/.gradle/caches/modules-2/files-2.1/org.robolectric/android-all/10-robolectric-5803371/873cbe64a7ac6fc2f9f5e2f6a24967fcc7a28ff8/android-all-10-robolectric-5803371.jar : Bad class file /Users/yschimke/.gradle/caches/modules-2/files-2.1/org.robolectric/android-all/10-robolectric-5803371/873cbe64a7ac6fc2f9f5e2f6a24967fcc7a28ff8/android-all-10-robolectric-5803371.jar:libcore/io/Os.class
  > Bad class file /Users/yschimke/.gradle/caches/modules-2/files-2.1/org.robolectric/android-all/10-robolectric-5803371/873cbe64a7ac6fc2f9f5e2f6a24967fcc7a28ff8/android-all-10-robolectric-5803371.jar:libcore/io/Os.class
    > (No message provided)
```

yschimke:
No wonder this was familiar...

https://github.com/square/okhttp/pull/5525

yschimke:
And https://github.com/xvik/gradle-animalsniffer-plugin/issues/8

okhttp/src/main/java/okhttp3/internal/platform/android/Android10SocketAdapter.kt
```diff
@@ -27,16 +28,13 @@ import okhttp3.internal.platform.Platform
...
-  private val socketFactory =
-      SSLCertificateSocketFactory.getDefault(10000) as SSLCertificateSocketFactory
-
   override fun trustManager(sslSocketFactory: SSLSocketFactory): X509TrustManager? = null
 
   override fun matchesSocketFactory(sslSocketFactory: SSLSocketFactory): Boolean = false
 
-  override fun matchesSocket(sslSocket: SSLSocket): Boolean = sslSocket.javaClass.name.startsWith(
-      "com.android.org.conscrypt")
+  override fun matchesSocket(sslSocket: SSLSocket): Boolean = SSLSockets.isSupportedSocket(sslSocket)
```
swankjesse: sweet!

yschimke: Agree, it's just kind of funny that we can't ignore this overall, just in one class at runtime on new devices.

And we have the horror that is DeferredSocketAdapter for dealing with classes that even trying reflection at runtime can fail fatally.  So we sample the socket as it goes past instead.

```
  @Synchronized private fun getDelegate(actualSSLSocketClass: SSLSocket): SocketAdapter? {
    if (!initialized) {
      try {
        var possibleClass: Class<in SSLSocket> = actualSSLSocketClass.javaClass
        while (possibleClass.name != "$socketPackage.OpenSSLSocketImpl") {
          possibleClass = possibleClass.superclass

          if (possibleClass == null) {
            throw AssertionError(
                "No OpenSSLSocketImpl superclass of socket of type $actualSSLSocketClass")
          }
        }

        delegate = AndroidSocketAdapter(possibleClass)
      } catch (e: Exception) {
        Platform.get()
            .log("Failed to initialize DeferredSocketAdapter $socketPackage", Platform.WARN, e)
      }

      initialized = true
    }

    return delegate
  }
```

yschimke:
Blocked on https://github.com/xvik/gradle-animalsniffer-plugin/issues/8

yschimke:
@swankjesse @JakeWharton This may never progress

https://github.com/mojohaus/animal-sniffer

> MojoHaus AnimalSniffer Maven Plugin
> 
> This is the animal-sniffer-maven-plugin.
> 
> Please note this plugin is now in maintenance level as the exact same feature can be now easily achieved using the --release flag from Javac see (http://openjdk.java.net/jeps/247)

And there just isn't any movement on the github issues in months, so I believe them.

yschimke:
A little bit of movement here, no release yet https://github.com/mojohaus/animal-sniffer/commit/71db56b1d247e0446e75e40628d8943297ab00ea

okhttp/src/main/java/okhttp3/internal/platform/AndroidPlatform.kt
```diff
@@ -38,6 +39,7 @@ import okhttp3.internal.tls.CertificateChainCleaner
 import okhttp3.internal.tls.TrustRootIndex
 
 /** Android 5+. */
+@SuppressSignatureCheck
```
yschimke, line 42: This is the highest risk part of the change, means watching Intellij/Android Studio warnings and running tests before releases generally.  We should automate the latter at some point.

swankjesse: I wonder what the work would be to get “warning clean”. Might just end up dumping `@SuppressWarnings` everywhere, which I don’t love.

PR #5834, opened 2020-02-29
Switch Http2Connection to daemon for clients

yschimke:


okhttp/src/main/java/okhttp3/internal/http2/Http2Connection.kt
```diff
@@ -489,9 +489,10 @@ class Http2Connection internal constructor(builder: Builder) : Closeable {
    *
    * @param sendConnectionPreface true to send connection preface frames. This should always be true
    *     except for in tests that don't check for a connection preface.
+   * @param daemon whether to use a Daemon thread for the connection.
    */
   @Throws(IOException::class) @JvmOverloads
-  fun start(sendConnectionPreface: Boolean = true) {
+  fun start(sendConnectionPreface: Boolean = true, daemon: Boolean = true) {
```
yschimke: FWIW I don't see this as an anti-pattern in Kotlin, so would like to leave them.  But in java tests it should be made clearer.

okhttp/src/main/java/okhttp3/internal/http2/Http2Connection.kt
```diff
@@ -500,7 +501,11 @@ class Http2Connection internal constructor(builder: Builder) : Closeable {
         writer.windowUpdate(0, (windowSize - DEFAULT_INITIAL_WINDOW_SIZE).toLong())
       }
     }
-    Thread(readerRunnable, connectionName).start() // Not a daemon thread.
+    // Thread doesn't use client Dispatcher, since it is scoped potentially across clients via
+    // ConnectionPool.
+    Thread(readerRunnable, connectionName).apply {
```
yschimke: An alternative design here would be to add TaskRunner.NON_DAEMON.  and then use the boolean to choose between INSTANCE and NON_DAEMON.

This would add automatic tracking of these threads, so we can check everything is shutdown in OkHttpClientTestRule also.

Worth doing?

mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.kt
```diff
@@ -536,7 +536,7 @@ class MockWebServer : ExternalResource(), Closeable {
             .socket(socket)
             .listener(http2SocketHandler)
             .build()
-        connection.start()
+        connection.start(taskRunner = TaskRunner.NON_DAEMON)
```
swankjesse: Just use `taskRunner` ?

okhttp/src/main/java/okhttp3/internal/concurrent/TaskRunner.kt
```diff
@@ -308,6 +308,9 @@ class TaskRunner(
     @JvmField
     val INSTANCE = TaskRunner(RealBackend(threadFactory("$okHttpName TaskRunner", daemon = true)))
 
+    @JvmField
+    val NON_DAEMON = TaskRunner(RealBackend(threadFactory("$okHttpName TaskRunner", daemon = false)))
```
swankjesse: If you use the MockWebServer’s taskRunner, you don’t need this.

okhttp/src/main/java/okhttp3/internal/http2/Http2Connection.kt
```diff
@@ -605,8 +608,8 @@ class Http2Connection internal constructor(builder: Builder) : Closeable {
    */
   inner class ReaderRunnable internal constructor(
     internal val reader: Http2Reader
-  ) : Runnable, Http2Reader.Handler {
-    override fun run() {
+  ) : Http2Reader.Handler, () -> Unit {
```
swankjesse: oh that’s really neat, I’ve never seen a function type as an implemented interface

yschimke: I'll leave it there, but I was tempted to drop it and also remove Runnable.  But this saves the block allocation I guess.

okhttp/src/test/java/okhttp3/internal/http2/Http2ConnectionTest.java
```diff
@@ -1832,6 +1834,24 @@
...
+
+    Socket socket = peer.openSocket();
+    Http2Connection connection = new Http2Connection.Builder(true, taskRunner)
+        .socket(socket)
+        .pushObserver(IGNORE)
+        .build();
+    connection.start(/* sendConnectionPreface = */ false, taskRunner);
+
+    List<TaskQueue> queues = taskRunner.activeQueues();
+    assertThat(queues).hasSize(1);
```
yschimke: Wanted to add

```
    //assertThat(queues.get(0).getActiveTask$okhttp().getName()).hasToString("OkHttp localhost");
```

But won't work in a Java based test :(

PR #5835, opened 2020-02-29
Allow for users who have disabled certificate checks in dev.

yschimke:
In dev with disabled TrustManager and hostName verifier, just checking whether we can reuse a connection across hosts can cause a failure

### DIFF
--- a/okhttp/src/main/java/okhttp3/Handshake.kt
+++ b/okhttp/src/main/java/okhttp3/Handshake.kt
@@ -22,6 +22,7 @@ import java.security.cert.X509Certificate
 import javax.net.ssl.SSLPeerUnverifiedException
 import javax.net.ssl.SSLSession
 import okhttp3.internal.immutableListOf
+import okhttp3.internal.platform.Platform
 import okhttp3.internal.toImmutableList
 
 /**
@@ -48,8 +49,14 @@ class Handshake internal constructor(
   peerCertificatesFn: () -> List<Certificate>
 ) {
   /** Returns a possibly-empty list of certificates that identify the remote peer. */
-  @get:JvmName("peerCertificates") val peerCertificates: List<Certificate> by lazy(
-      peerCertificatesFn)
+  @get:JvmName("peerCertificates") val peerCertificates: List<Certificate> by lazy {
+    try {
+      peerCertificatesFn()
+    } catch (spue: SSLPeerUnverifiedException) {
+      Platform.get().log("Unable to verify connection certificates", Platform.INFO, spue)
+      listOf<Certificate>()
+    }
+  }
 
   @JvmName("-deprecated_tlsVersion")
   @Deprecated(
--- a/okhttp/src/main/java/okhttp3/Handshake.kt
+++ b/okhttp/src/main/java/okhttp3/Handshake.kt
@@ -22,7 +22,6 @@ import java.security.cert.X509Certificate
 import javax.net.ssl.SSLPeerUnverifiedException
 import javax.net.ssl.SSLSession
 import okhttp3.internal.immutableListOf
-import okhttp3.internal.platform.Platform
 import okhttp3.internal.toImmutableList
 
 /**
@@ -53,7 +52,6 @@ class Handshake internal constructor(
     try {
       peerCertificatesFn()
     } catch (spue: SSLPeerUnverifiedException) {
-      Platform.get().log("Unable to verify connection certificates", Platform.INFO, spue)
       listOf<Certificate>()
     }
   }
@@ -128,11 +126,7 @@ class Handshake internal constructor(
   }
 
   override fun toString(): String {
-    val peerCertificatesString = try {
-      peerCertificates.map { it.name }.toString()
-    } catch (_: SSLPeerUnverifiedException) {
-      "Failed: SSLPeerUnverifiedException"
-    }
+    val peerCertificatesString = peerCertificates.map { it.name }.toString()
     return "Handshake{" +
         "tlsVersion=$tlsVersion " +
         "cipherSuite=$cipherSuite " +
--- a/okhttp/src/main/java/okhttp3/internal/connection/RealConnection.kt
+++ b/okhttp/src/main/java/okhttp3/internal/connection/RealConnection.kt
@@ -46,6 +46,7 @@ import okhttp3.Response
 import okhttp3.Route
 import okhttp3.internal.EMPTY_RESPONSE
 import okhttp3.internal.assertThreadDoesntHoldLock
+import okhttp3.internal.assertThreadHoldsLock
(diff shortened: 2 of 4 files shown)


yschimke:
TODO: discuss how to test

okhttp/src/main/java/okhttp3/internal/connection/RealConnection.kt
```diff
@@ -575,10 +576,17 @@ class RealConnection(
...
-        OkHostnameVerifier.verify(url.host, handshake!!.peerCertificates[0] as X509Certificate)
+    try {
+      // We have a host mismatch. But if the certificate matches, we're still good.
+      return !noCoalescedConnections &&
+          handshake != null &&
+          OkHostnameVerifier.verify(url.host, handshake!!.peerCertificates[0] as X509Certificate)
+    } catch (_: SSLPeerUnverifiedException) {
+      // OkHostnameVerifier isn't guaranteed to work if user has disabled security via
+      // TrustManager and hostnameVerifier
+      noCoalescedConnections = true
```
yschimke: @swankjesse This is only used when we already have the connectionPool lock.  calling the noCoalescedConnections method would fail because of this.  But checking here and failing if we don't have it seems problematic in an exception handler.

Thoughts?

yschimke: Triggers

```
    java.lang.AssertionError: Thread Test worker MUST hold lock on okhttp3.internal.connection.RealConnectionPool@56c6ffac
        at okhttp3.internal.connection.RealConnection.supportsUrl(RealConnection.kt:770)
        at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:87)
        at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:100)
        at okhttp3.internal.cache.CacheInterceptor.intercept(CacheInterceptor.kt:82)
        at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:100)
        at okhttp3.internal.http.BridgeInterceptor.intercept(BridgeInterceptor.kt:83)
        at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:100)
        at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:100)
        at okhttp3.internal.connection.RealCall.getResponseWithInterceptorChain$okhttp(RealCall.kt:197)
        at okhttp3.internal.connection.RealCall.execute(RealCall.kt:148)
        at okhttp3.CacheTest.get(CacheTest.java:2490)
        at okhttp3.CacheTest.varyMatchesAddedRequestHeaderField(CacheTest.java:1605)
```

yschimke:
Something isn't satisfying about this change

Ideally it should

1) Catch the exception, return false
2) Remember that the connection is unsafe (avoid wasting effort next time)
3) Do this safely considering required connection pool lock

image

okhttp/src/main/java/okhttp3/internal/connection/RealConnection.kt
```diff
@@ -564,6 +565,22 @@ class RealConnection(
...
+  /**
+   * The same as [supportsUrl] but catching any issues that would disqualify this live Connection
+   * for use across others urls. In that case noCoalescedConnections is updated, but the method
+   * must be called within the connectionPool lock.
+   */
+  fun safeSupportsUrl(url: HttpUrl) = try {
+    connectionPool.assertThreadHoldsLock()
+
+    supportsUrl(url)
+  } catch (_: SSLPeerUnverifiedException) {
```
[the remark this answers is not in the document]
swankjesse: Would it be better to have an internal method Handshake like `firstPeerCertificate` ? That method might be able to skip chain cleaning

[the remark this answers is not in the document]
yschimke: I think cleaning does sorting though. And without cleaning what if we accept a self signed cert that is for a host we don't have rights to.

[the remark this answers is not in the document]
yschimke: I'll hold off on this one.  It's awkward.

yschimke:
> I don’t love the change, not because the change doesn’t work but because it leaks the fact that peer certificate cleaning is lazy to more callsites. I still find that awkward, especially because it’s supposed to work like a value object.
> 
> Next steps? What happens if we change the cleaner function to catch and return uncleaned certificates? I don’t _think_ there’s an attack vector there.

The awkward situation is that it's correct to do this in the following cases

1. We want to check certificate pinning
2. We want to log the peer certificates
3. We want to coalesce traffic

If you think firstPeerCertificate can be cheaper then let's do that.  But it's not clear to me how.

What about tweaking your suggestion to centralise this behind the lazy function and swallow and return _0 certificates_.  It should work for the first two, and the third should be a simple fix.  It doesn't risk allowing an attack of claiming other hosts with an invalid local certificate.

```
    // We have a host mismatch. But if the certificate matches, we're still good.
    return !noCoalescedConnections &&
        handshake != null &&
        OkHostnameVerifier.verify(url.host, handshake!!.peerCertificates[0] as X509Certificate)
```

swankjesse:
The zero certificates idea is great. Let's do that!

okhttp/src/main/java/okhttp3/Handshake.kt
```diff
@@ -48,8 +48,13 @@ class Handshake internal constructor(
   peerCertificatesFn: () -> List<Certificate>
 ) {
   /** Returns a possibly-empty list of certificates that identify the remote peer. */
-  @get:JvmName("peerCertificates") val peerCertificates: List<Certificate> by lazy(
-      peerCertificatesFn)
+  @get:JvmName("peerCertificates") val peerCertificates: List<Certificate> by lazy {
+    try {
+      peerCertificatesFn()
+    } catch (spue: SSLPeerUnverifiedException) {
+      listOf<Certificate>()
```
yschimke: Removed logging, as it's spammy.

yschimke:
ping @swankjesse

okhttp/src/main/java/okhttp3/Handshake.kt
```diff
@@ -121,11 +126,7 @@ class Handshake internal constructor(
   }
 
   override fun toString(): String {
-    val peerCertificatesString = try {
-      peerCertificates.map { it.name }.toString()
-    } catch (_: SSLPeerUnverifiedException) {
-      "Failed: SSLPeerUnverifiedException"
-    }
+    val peerCertificatesString = peerCertificates.map { it.name }.toString()
```
swankjesse: Nice.

swankjesse:
I’m really happy with where this landed.

PR #4871, opened 2019-04-04
Add WebSocket permessage-deflate

AntonRutkevich:
Resolves #1733 

Hi! 

Here's the draft of the ["permessage-deflate"](https://tools.ietf.org/html/rfc7692#section-7) compression for WebSocket. 

It has at least one unresolved issue discussed in https://github.com/square/okio/issues/570. 
It would be great to get your feedback on the overall approach. 

Thanks!

PR #5839, opened 2020-03-03
WebSocket permessage-deflate extension header parser

swankjesse:
Originally contributed via PR #4871. This differs from that contribution
in a few ways:

 - rename WebSocketOptions to WebSocketExtensions
 - rename properties to track names defined in RFC 7692
 - use several independent JUnit tests instead of @Parameterized
 - use Util.kt to do parsing without intermediate splitting

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/ws/WebSocketExtensions.kt
+++ b/okhttp/src/main/java/okhttp3/internal/ws/WebSocketExtensions.kt
@@ -1,0 +1,172 @@
+/*
+ * Copyright (C) 2020 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okhttp3.internal.ws
+
+import java.io.IOException
+import okhttp3.Response
+import okhttp3.internal.delimiterOffset
+import okhttp3.internal.trimSubstring
+
+/**
+ * Models the contents of a `Sec-WebSocket-Extensions` response header. OkHttp honors one extension
+ * `permessage-deflate` and four parameters, `client_max_window_bits`, `client_no_context_takeover`,
+ * `server_max_window_bits`, and `server_no_context_takeover`.
+ *
+ * Typically this will look like one of the following:
+ *
+ * ```
+ * Sec-WebSocket-Extensions: permessage-deflate
+ * Sec-WebSocket-Extensions: permessage-deflate; client_max_window_bits="15"
+ * Sec-WebSocket-Extensions: permessage-deflate; client_max_window_bits=15
+ * Sec-WebSocket-Extensions: permessage-deflate; client_no_context_takeover
+ * Sec-WebSocket-Extensions: permessage-deflate; server_max_window_bits="15"
+ * Sec-WebSocket-Extensions: permessage-deflate; server_max_window_bits=15
+ * Sec-WebSocket-Extensions: permessage-deflate; server_no_context_takeover
+ * Sec-WebSocket-Extensions: permessage-deflate; server_no_context_takeover;
+ *     client_no_context_takeover
+ * Sec-WebSocket-Extensions: permessage-deflate; server_max_window_bits="15";
+ *     client_max_window_bits="15"; server_no_context_takeover; client_no_context_takeover
+ * ```
+ *
+ * If any other extension or parameter is specified, then [unknownValues] will be true. Such
+ * responses should be refused as their web socket extensions will not be understood.
+ *
+ * Note that [java.util.zip.Deflater] is hardcoded to use 15 bits (32 KiB) for
+ * `client_max_window_bits` and [java.util.zip.Inflater] is hardcoded to use 15 bits (32 KiB) for
+ * `server_max_window_bits`. This harms our ability to support these parameters:
+ *
+ *  * If `client_max_window_bits` is less than 15, OkHttp must close the web socket with code 1010.
+ *    Otherwise it would compress values in a way that servers could not decompress.
+ *  * If `server_max_window_bits` is less than 15, OkHttp will waste memory on an oversized buffer.
+ *
+ * See [RFC 7692, 7.1][rfc_7692] for details on negotiation process.
+ *
+ * [rfc_7692]: https://tools.ietf.org/html/rfc7692#section-7.1
+ */
+data class WebSocketExtensions(
+  /** True if the agreed upon extensions includes the permessage-deflate extension. */
+  @JvmField val perMessageDeflate: Boolean = false,
+
+  /** Should be a value in [8..15]. Only 15 is acceptable by OkHttp as Java APIs are limited. */
+  @JvmField val clientMaxWindowBits: Int? = null,
+
+  /** True if the agreed upon extension parameters includes "client_no_context_takeover". */
+  @JvmField val clientNoContextTakeover: Boolean = false,
+
+  /** Should be a value in [8..15]. Any value in that range is acceptable by OkHttp. */
+  @JvmField val serverMaxWindowBits: Int? = null,
+
+  /** True if the agreed upon extension parameters includes "server_no_context_takeover". */
+  @JvmField val serverNoContextTakeover: Boolean = false,
+
+  /**
+   * True if the agreed upon extensions or parameters contained values unrecognized by OkHttp.
+   * Typically this indicates that the client will need to close the web socket with code 1010.
+   */
+  @JvmField val unknownValues: Boolean = false
+) {
+  companion object {
+    private const val HEADER_WEB_SOCKET_EXTENSION = "Sec-WebSocket-Extensions"
+
+    @Throws(IOException::class)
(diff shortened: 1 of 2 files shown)


okhttp/src/main/java/okhttp3/internal/ws/WebSocketExtensions.kt
```diff
@@ -0,0 +1,133 @@
...
+package okhttp3.internal.ws
+
+import java.io.IOException
+import okhttp3.Response
+import okhttp3.internal.delimiterOffset
+import okhttp3.internal.trimSubstring
+
+/**
+ * Models the contents of a `Sec-WebSocket-Extensions` response header. OkHttp honors one extension
+ * `permessage-deflate` and two of its parameters, `client_no_context_takeover` and
```
AntonRutkevich: Again, for future me. `12` here is just any value 8<=X<15.

window_bits | sender=15 | sender=12
-------------------- | ------- | -------- 
receiver=15  | ok | ok
receiver=12 | failure | ok

For client-server stream, okhttp is a **sender** and can use any window_size, as long as it is supported by server.

For server-client stream, okhttp is a **receiver** and can work with any `server_max_window_bits` received from server. Server won't send `client_max_window_bits` without prior request from client.

So, 
1. There is no reason for okhttp to send `server_max_window_bits` or `client_max_window_bits` to server, as okhttp supports 15 (and therefore, any value when okhttp is reading).
2. okhttp MAY ignore `server_max_window_bits` from server, as it defines the sliding window for server-client stream, and okhttp is fine with it being smaller than 15. However, we MUST NOT fail if we surprisingly receive such parameter from server. [7.1.2.1, last paragraph](https://tools.ietf.org/html/rfc7692#section-7.1.2.1)
3. okhttp MUST fail on `client_max_window_bits` from server, as server MUST NOT send `client_max_window_bits` unless received it from client in request (and we don't send it).

okhttp/src/main/java/okhttp3/internal/ws/WebSocketExtensions.kt
```diff
@@ -0,0 +1,133 @@
...
+ */
+package okhttp3.internal.ws
+
+import java.io.IOException
+import okhttp3.Response
+import okhttp3.internal.delimiterOffset
+import okhttp3.internal.trimSubstring
+
+/**
+ * Models the contents of a `Sec-WebSocket-Extensions` response header. OkHttp honors one extension
```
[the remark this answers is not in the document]
AntonRutkevich: @swankjesse , from what I see, the PR is aligned with the 1..3 from above, so everything is fine here.

okhttp/src/main/java/okhttp3/internal/ws/WebSocketExtensions.kt
AntonRutkevich: @swankjesse , 

looks like support for 2 must be added to avoid failing in cases where we can actually operate normally. 

The logic is fine for 3, but I'd update the reasoning for `client_max_window_bits` in the doc below to something like: "client_max_window_bits is unknown because server MUST NOT return it unless requested by client, and we don't send it".

AntonRutkevich: Though, for 2, I'm not sure how to test it live, eg where we can find a server which will have configurable LZ77 sliding window.

okhttp/src/main/java/okhttp3/internal/ws/WebSocketExtensions.kt
```diff
@@ -0,0 +1,133 @@
...
+ *
+ * Note that the `client_max_window_bits` and `server_max_window_bits` parameters are unknown
+ * because we do not expect servers to return them. (OkHttp is unable to handle any number of bits
+ * other than 15 (32 KiB) due to a lack of configurability in [java.util.zip.Deflater].)
+ *
+ * See [RFC 7692, 7.1][rfc_7692] for details on negotiation process.
+ *
+ * [rfc_7692]: https://tools.ietf.org/html/rfc7692#section-7.1
+ */
+data class WebSocketExtensions(
```
AntonRutkevich: Indeed, with PR separated into several smaller ones, it's much easier to reason about them. 👍

okhttp/src/test/java/okhttp3/internal/ws/WebSocketExtensionsTest.kt
```diff
@@ -0,0 +1,178 @@
...
+ */
+package okhttp3.internal.ws
+
+import okhttp3.Protocol
+import okhttp3.Request
+import okhttp3.Response
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.Test
+
+internal class WebSocketExtensionsTest {
```
yschimke: nit: why internal?

okhttp/src/main/java/okhttp3/internal/ws/WebSocketExtensions.kt
swankjesse: Oooh good catch. We should accept and ignore `server_max_window_bits=8` through `server_max_window_bits=15`. We can't benefit from allocating a smaller buffer due to Java API limitations.

okhttp/src/main/java/okhttp3/internal/ws/WebSocketExtensions.kt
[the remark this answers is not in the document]
swankjesse: We need to track the server's behavior for context takeover by making reset() calls at appropriate times. Otherwise we'll be decompressing with the wrong compression context. So we cannot ignore the server’s context takeover configuration.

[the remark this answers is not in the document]
swankjesse: (I think your #3 should be MUST, not MAY.)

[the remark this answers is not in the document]
AntonRutkevich: Re-read the source and specs. You're right, we need to support `server_no_context_takeover` in response, and apply it to MessageDeflater (eg for client->server stream part). 

For some reason I missed the part that we cannot make context_takeover asymmetrical, as it applies to both reading and writing simultaneously on each party. 

Fixed the comment above, thanks.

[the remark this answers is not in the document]
AntonRutkevich: @swankjesse , please see above, I had to update the comment.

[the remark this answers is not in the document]
AntonRutkevich: However, it might be the case that I miss the point and we have to add the `contextTakeover` parameter to the MessageInflater too, and reset the `inflater` on every message read.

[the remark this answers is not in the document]
swankjesse: I tested this experimentally. If the reset calls don’t align between deflater and inflater I get an exception:

https://gist.github.com/swankjesse/b504eb456474e4c5f47fbbeba1ef0244#file-resettest-kt-L69

okhttp/src/test/java/okhttp3/internal/ws/WebSocketExtensionsTest.kt
swankjesse: No good reason. Fixed.

okhttp/src/main/java/okhttp3/internal/ws/WebSocketExtensions.kt
swankjesse: Updated to support parsing both the client_max_window_bits and server_max_window_bits. It’s simpler to just parse everything and to do validation work separately.

okhttp/src/main/java/okhttp3/internal/ws/WebSocketExtensions.kt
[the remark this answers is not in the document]
AntonRutkevich: It fails indeed.
Thanks for double checking! Turns out I was under incorrect impression on how it works.

We'll have to add `contextTakeover` parameter based on `server_no_context_takeover` option from server response then.

okhttp/src/main/java/okhttp3/internal/ws/WebSocketExtensions.kt
```diff
@@ -0,0 +1,172 @@
...
+ * `client_max_window_bits` and [java.util.zip.Inflater] is hardcoded to use 15 bits (32 KiB) for
+ * `server_max_window_bits`. This harms our ability to support these parameters:
+ *
+ *  * If `client_max_window_bits` is less than 15, OkHttp must close the web socket with code 1010.
+ *    Otherwise it would compress values in a way that servers could not decompress.
+ *  * If `server_max_window_bits` is less than 15, OkHttp will waste memory on an oversized buffer.
+ *
+ * See [RFC 7692, 7.1][rfc_7692] for details on negotiation process.
+ *
+ * [rfc_7692]: https://tools.ietf.org/html/rfc7692#section-7.1
```
AntonRutkevich: Great doc!

okhttp/src/main/java/okhttp3/internal/ws/WebSocketExtensions.kt
```diff
@@ -0,0 +1,172 @@
...
+                  }
+                  else -> {
+                    unexpectedValues = true // Unexpected parameter.
+                  }
+                }
+              }
+            }
+
+            else -> {
+              unexpectedValues = true // Unexpected extension.
```
AntonRutkevich: WDYT about making `unexpectedValues` a data class, and save the message that is currently in comments to it? Along with the content of `Sec-WebSocket-Extensions` header that could not be parsed. 

This way we could throw a meaningful exception to the user.

swankjesse: I don't think we're going to need it in practice. If we want the user to understand the problem, we should just include the unparsed header.

Issue #5840, opened 2020-03-03
Crash on Android 10 with underscore in subdomain


talklittle:
Crash on OkHttp 3.12.9 and 3.12.10 when a URL subdomain contains underscores. Reproducible on API 29 (Android 10) emulator but not API 28 (Android 9).

On OkHttp 3.12.8 it errors but does not crash. Similarly OkHttp 3.12.10 errors but does not crash on Android 9.

Reproduce by making a GET request to an HTTPS URL with underscore in subdomain.

I saw old issue #242 closed saying underscores aren't allowed in hostnames, which may be technically true, but they are sometimes seen in the wild. Examples:

* https://open_nsfw.gitlab.io/ (which was highly circulated on the web a few years ago)
* https://example_underscore_123.s3.amazonaws.com/
* https://instant_regret.reddit.com/

Logcats:

OkHttp 3.12.10 on Android 10:

    --------- beginning of crash
    E/AndroidRuntime: FATAL EXCEPTION: OkHttp Dispatcher
    Process: com.talklittle.testokhttpunderscoresubdomain, PID: 10321
    java.lang.IllegalArgumentException: Invalid input to toASCII: example_underscore.s3.amazonaws.com
        at java.net.IDN.toASCII(IDN.java:115)
        at javax.net.ssl.SNIHostName.<init>(SNIHostName.java:99)
        at com.android.org.conscrypt.Platform.getSSLParameters(Platform.java:153)
        at com.android.org.conscrypt.ConscryptFileDescriptorSocket.getSSLParameters(ConscryptFileDescriptorSocket.java:1117)
        at okhttp3.internal.platform.Android10Platform.configureTlsExtensions(Android10Platform.java:40)
        at okhttp3.internal.connection.RealConnection.connectTls(RealConnection.java:315)
        at okhttp3.internal.connection.RealConnection.establishProtocol(RealConnection.java:284)
        at okhttp3.internal.connection.RealConnection.connect(RealConnection.java:169)
        at okhttp3.internal.connection.StreamAllocation.findConnection(StreamAllocation.java:258)
        at okhttp3.internal.connection.StreamAllocation.findHealthyConnection(StreamAllocation.java:135)
        at okhttp3.internal.connection.StreamAllocation.newStream(StreamAllocation.java:114)
        at okhttp3.internal.connection.ConnectInterceptor.intercept(ConnectInterceptor.java:42)
[...]

PR #5841, opened 2020-03-03
Catch IllegalArgumentException from Android internals

yschimke:
See https://github.com/square/okhttp/issues/5840
https://github.com/google/conscrypt/issues/821

### DIFF
--- a/android-test/src/androidTest/java/okhttp/android/test/OkHttpTest.kt
+++ b/android-test/src/androidTest/java/okhttp/android/test/OkHttpTest.kt
@@ -64,6 +64,8 @@ import javax.net.ssl.X509TrustManager
 import java.util.logging.Logger
 import okhttp3.internal.platform.AndroidPlatform
 import okhttp3.internal.platform.Android10Platform
+import java.io.IOException
+import java.lang.IllegalArgumentException
 
 /**
  * Run with "./gradlew :android-test:connectedCheck" and make sure ANDROID_SDK_ROOT is set.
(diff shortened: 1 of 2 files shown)


okhttp/src/main/java/okhttp3/internal/platform/android/Android10SocketAdapter.kt
```diff
@@ -54,14 +56,19 @@ class Android10SocketAdapter : SocketAdapter {
...
-    // Enable ALPN.
-    sslParameters.applicationProtocols = Platform.alpnProtocolNames(protocols).toTypedArray()
+      // Enable ALPN.
+      sslParameters.applicationProtocols = Platform.alpnProtocolNames(protocols).toTypedArray()
 
-    sslSocket.sslParameters = sslParameters
+      sslSocket.sslParameters = sslParameters
+    } catch (iae: IllegalArgumentException) {
+      // probably java.lang.IllegalArgumentException: Invalid input to toASCII from IDN.toASCII
+      throw IOException("Android internal error", iae)
```
yschimke: @swankjesse Thoughts on wording here?

yschimke: Also swallowing here would allow request to proceed, but with HTTP/1.1

android-test/src/androidTest/java/okhttp/android/test/OkHttpTest.kt
```diff
@@ -501,6 +503,23 @@ class OkHttpTest {
...
+  @Test
+  fun testUnderscoreRequest() {
+    assumeNetwork()
+
+    val request =
+        Request.Builder().url("https://example_underscore_123.s3.amazonaws.com/").build()
+
+    try {
+      client.newCall(request).execute()
+      // Hopefully this passes
```
swankjesse: Good comment.

okhttp/src/main/java/okhttp3/internal/platform/android/Android10SocketAdapter.kt
swankjesse: I like this.

PR #5842, opened 2020-03-03
Catch IllegalArgumentException from Android internals [3.14.x]

yschimke:
See #5840
google/conscrypt#821


yschimke:
Tested on another branch https://github.com/square/okhttp/pull/5843

PR #5848, opened 2020-03-05
Web Socket message inflater and deflater

swankjesse:
Originally contributed via PR #4871. This differs from that contribution
in a few ways:

 - Combined inflater and deflater into one test
 - Always operate on a buffer in-place
 - Support no context takeover in both reading and writing

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/ws/MessageDeflater.kt
+++ b/okhttp/src/main/java/okhttp3/internal/ws/MessageDeflater.kt
@@ -1,0 +1,65 @@
+/*
+ * Copyright (C) 2020 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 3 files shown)


okhttp/src/main/java/okhttp3/internal/ws/MessageDeflater.kt
```diff
@@ -0,0 +1,65 @@
...
+import java.util.zip.Deflater
+import okio.Buffer
+import okio.ByteString
+import okio.ByteString.Companion.decodeHex
+import okio.DeflaterSink
+
+private val EMPTY_DEFLATE_BLOCK = "000000ffff".decodeHex()
+private const val LAST_OCTETS_COUNT_TO_REMOVE_AFTER_DEFLATION = 4
+
+class MessageDeflater(
```
JakeWharton: Nothing about these types are specific to WS or its messages. Should this be a BufferDeflater?

swankjesse: It manipulates the deflated data to match the WS spec and is unlikely to be appropriate in other contexts. I don’t think it’s general purpose!

swankjesse:
FYI @AntonRutkevich !

PR #5851, opened 2020-03-06
Rip out WebSocketWriter's message splitting

swankjesse:
An old version of OkHttp web sockets supported streaming messages,
so a writer could transmit partial bytes of a message without
completing it.

When we promoted Web Sockets as a core feature we dropped support
for streaming a single message. But we still had the mechanism
that implemented it.

This PR removes that mechanism, which should make it easier to
implement compression.

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/ws/WebSocketWriter.kt
+++ b/okhttp/src/main/java/okhttp3/internal/ws/WebSocketWriter.kt
@@ -19,7 +19,6 @@ import java.io.IOException
 import java.util.Random
 import okhttp3.internal.ws.WebSocketProtocol.B0_FLAG_FIN
 import okhttp3.internal.ws.WebSocketProtocol.B1_FLAG_MASK
-import okhttp3.internal.ws.WebSocketProtocol.OPCODE_CONTINUATION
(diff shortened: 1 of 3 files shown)


swankjesse:
FYI @AntonRutkevich !

okhttp/src/main/java/okhttp3/internal/ws/WebSocketWriter.kt
```diff
@@ -140,120 +132,51 @@ internal class WebSocketWriter(
     sink.flush()
   }
 
-  /**
-   * Stream a message payload as a series of frames. This allows control frames to be interleaved
-   * between parts of the message.
-   */
```
AntonRutkevich: Based on this comment, with this PR merged, [control frames](https://tools.ietf.org/html/rfc6455#section-5.5.1) (CLOSE, PING, PONG) will have to wait till a large message (eg 100..200+ Kb) is fully sent. Right? 

Or did it behave the same before, and the removed code is dead anyway?

JakeWharton: It's dead code. The input is a `ByteString` which when written to a `Buffer` was written as a single, large message.

It had the behavior of the `serverLargeBufferedPayloadWrittenAsOneFrame` test.

PR #5854, opened 2020-03-06
[WIP] Prototype Dev Allowlist

yschimke:
Instead of encouraging disabling the trust manager completely, and also breaking all use of peer certificates in Dev.  Try to implement a Dev Allowlist cleanly.

Requirements

- Must only work in Debug builds for Android
- Should minimise overall risk, e.g. still use security for other requests, to avoid prod surprises.


okhttp/src/main/java/okhttp3/OkHttpClient.kt
```diff
@@ -988,6 +990,20 @@ open class OkHttpClient internal constructor(
       writeTimeout(duration.toMillis(), MILLISECONDS)
     }
 
+    @VisibleForTesting
+    fun enableDevWhitelist(vararg hosts: String) = apply {
```
JakeWharton: Please use the term allowlist instead of whitelist

okhttp/src/main/java/okhttp3/OkHttpClient.kt
```diff
@@ -15,6 +15,7 @@
  */
 package okhttp3
 
+import com.android.internal.annotations.VisibleForTesting
```
JakeWharton: This shouldn't be used

android-test/src/androidTest/java/okhttp/android/test/OkHttpTest.kt
```diff
@@ -93,6 +97,11 @@ class OkHttpTest {
   val server = MockWebServer()
   private val handshakeCertificates = localhost()
 
+  @After
```
yschimke: Shouldn't be needed

android-test/src/androidTest/java/okhttp/android/test/OkHttpTest.kt
```diff
@@ -520,6 +529,45 @@ class OkHttpTest {
     }
   }
 
+  @Test
+  fun testDevAllowlist() {
+    assumeNetwork()
+    enableTls()
+
+    println(BuildConfig.DEBUG)
```
yschimke: TODO remove debug

okhttp/src/main/java/okhttp3/android/OkHttpAndroidClientFactory.kt
```diff
@@ -0,0 +1,58 @@
+package okhttp3.android
+
+import android.content.Context
+import okhttp3.OkHttpClient
+import okhttp3.internal.platform.Platform
+import okhttp3.internal.platform.android.BuildConfig
+import okhttp3.internal.tls.AllowlistedTrustManager
+
+class OkHttpAndroidClientFactory private constructor(val client: OkHttpClient) {
```
yschimke: Haven't given the Java API.  This is currently Kotlin first.  I expanded the scope to how do we hook into Android environment in one place, so we can understand things like the lifecycle and DEBUG mode etc.

This might be too far, or something to play with a bit, see what it's like via snapshots with real projects.

swankjesse:
I don’t like this API. It competes with `OkHttpClient.Builder`... if we wanted to add ergonomics to that class... we should just do that directly. Perhaps with extension functions that are only available on Android?

swankjesse:
android.kt

```
fun OkHttpClient.Builder.cache(size: Long) : OkHttpClient.Builder
fun OkHttpClient.Builder.unsafeDebugHost(name: String) : OkHttpClient.Builder
```

yschimke:
I'd like to have a solid integration point with the App context. Was hoping to find something workable that we can build on.

Hence tying together Context, BuildConfig into API.

swankjesse:
Good point on the application Context. Yet I still dislike offering both our Builder and the new Factory thing. It feels clumsy because the APIs are not orthogonal: you can configure the builder directly, or via the context, and it's unclear why they're different.

I'd prefer something where we don't have APIs that replace the Builder; instead I'd like ones that help you to use it.
```
fun Context.okhttpCache(size: Long): Cache
fun OkHttpClient.Builder.unsafeDebugHost(context: Context, name: String) : OkHttpClient.Builder
```

yschimke:
> Good point on the application Context. Yet I still dislike offering both our Builder and the new Factory thing. It feels clumsy because the APIs are not orthogonal: you can configure the builder directly, or via the context, and it's unclear why they're different.
> 
> I'd prefer something where we don't have APIs that replace the Builder; instead I'd like ones that help you to use it.

Let's riff on this PR, get a useful set of functionality for Android, and iterate until I can convince you without twisting your arm and you are legit happy with it.

swankjesse:
That is the recipe that we use to make great things!

okhttp/src/main/java/okhttp3/android/OkHttpAndroidClientFactory.kt
```diff
@@ -0,0 +1,70 @@
...
+    fun newClient(init: OkHttpClient.Builder.() -> Unit): OkHttpClient {
+        val builder = client.newBuilder()
+        init(builder)
+        return builder.build()
+    }
+
+    fun shutdown() {
```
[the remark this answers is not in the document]
yschimke: FWIW part of the reason for shutdown here, is that if you have a Cache you probably also want to close it, since that does some cleanup.  And this method could be hooked int App Lifecycle.

yschimke:
@swankjesse This PR is definitely more for discussion than landing.  The allow list can be split out, but worth discussing how to hook into more android lifecycle etc, so we build something cohesive when we need to be aware of things like backgrounding and doze mode.

okhttp/src/main/java/okhttp3/android/android.kt
```diff
@@ -0,0 +1,53 @@
+package okhttp3.android
+
+import android.content.Context
+import okhttp3.Cache
+import okhttp3.OkHttpClient
+import okhttp3.internal.platform.Platform
+import okhttp3.internal.platform.android.BuildConfig
+import okhttp3.internal.tls.AllowlistedTrustManager
+
+const val MiB = 1024L * 1024L
```
swankjesse: internal or private

okhttp/src/main/java/okhttp3/android/android.kt
```diff
@@ -0,0 +1,53 @@
...
+
+fun setDevMode(buildConfigClass: Class<*>? = null, buildConfig: BuildConfig? = null) {
```
swankjesse: the `static` nature of this feels bad. . . who is responsible for calling this?

If we wanted to do this, I think we should ship an aar and install it automatically? (Should we do that?)

okhttp/src/main/java/okhttp3/internal/concurrent/TaskQueue.kt
```diff
@@ -101,6 +101,11 @@ class TaskQueue internal constructor(
     }, delayNanos)
   }
 
+  fun activeTaskNames(): List<String> = synchronized(taskRunner) {
```
swankjesse: I’m not personally ready for expression function utopia

okhttp/src/main/java/okhttp3/internal/platform/android/BuildConfig.kt
```diff
@@ -0,0 +1,16 @@
+package okhttp3.internal.platform.android
+
+@Suppress("PropertyName")
+class BuildConfig internal constructor(val DEBUG: Boolean) {
```
swankjesse: cute

okhttp/src/main/java/okhttp3/android/android.kt
yschimke: Let's try it.  Don't rush this. Maybe it's not due for 5.0 branch along with other eg MP features. But get it feeling right before we make a bad public API.

yschimke: I think that also solves another issue I had.  Outside of robolectric android-all anything else is mostly unavailable if shipped as an AAR.  This code could live in a AAR building okhttp-android module?

swankjesse: Yep!

Issue #5859, opened 2020-03-08
GET BUG in 4.4.0


MS893:
Hi,

When I perform a GET on this URL : "https://hopteam.hop.fr" using 4.3.1 version, it works fine (response is 200)

If I do the same GET with 4.4.0, responses is 404

Here is my code
`            request = new Request.Builder()
                .url(url_hopteam)
                .build();
            response = client.newCall(request).execute();
`

The client is
`        client = new OkHttpClient.Builder()
            .connectTimeout(28, TimeUnit.SECONDS)
            .writeTimeout(20, TimeUnit.SECONDS)
            .readTimeout(28, TimeUnit.SECONDS)
            .addNetworkInterceptor(new LoggingInterceptor())
            .cookieJar(cookieJar)
            .connectionSpecs(Arrays.asList(ConnectionSpec.MODERN_TLS, ConnectionSpec.COMPATIBLE_TLS))
            .build();
`

In my app build.gradle, I have
`    implementation 'com.squareup.okhttp3:okhttp:4.4.0'`

In the project build.gradle, I have
`        classpath 'com.android.tools.build:gradle:3.5.2'
`

MS893:
![image](https://user-images.githubusercontent.com/12143016/76160078-2a542d80-6127-11ea-8af3-1f1afeef93d4.png)

yschimke:
I suspect this is a server misconfiguration, potentially triggered by a subtle change in OkHttp behaviour with redirects, connection coalescing and DNS ordering.

```
$ okurl --debug https://hopteam.hop.fr 2>&1 | grep Location
11:40:19.114	Location: https://office.hop.fr/opensso/saml2/jsp/spSSOInit.jsp?metaAlias=/MSAD/sp&idpEntityID=http://madfs.hop.fr/adfs/services/trust&NameIDFormat=urn:oasis:names:tc:SAML:2.0:nameid-format:transient&binding=HTTP-POST&goto=https://hopteam.hop.fr/home
11:40:19.514	Location: https://madfs.hop.fr/adfs/ls/?SAMLRequest=...
```

I'll check the certificates to confirm this.

swankjesse:
Did we inadvertently started coalescing http/1 connections? Other scarier regression would be coalescing redirect calls without a 2nd DNS check

yschimke:
I suspect the latter.

yschimke:
```
$ nslookup hopteam.hop.fr | grep Address
Address: 193.84.89.75
[...]

PR #5860, opened 2020-03-08
Don't reuse a connection on redirect if certs match but DNS does not

swankjesse:
We attempt to minimize connection and reconnection work, but in this case we
were overly aggressive about retaining the same connection. In some deployments
services will share certificates but not DNS addresses; when redirecting
between such services we were incorrectly attempting to reuse the connection.

This would have resulted in 404s and other misdirected requests.

Closes: https://github.com/square/okhttp/issues/5859

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/connection/ExchangeFinder.kt
+++ b/okhttp/src/main/java/okhttp3/internal/connection/ExchangeFinder.kt
@@ -19,6 +19,7 @@ import java.io.IOException
 import java.net.Socket
 import okhttp3.Address
 import okhttp3.EventListener
+import okhttp3.HttpUrl
 import okhttp3.OkHttpClient
 import okhttp3.Route
 import okhttp3.internal.assertThreadDoesntHoldLock
@@ -145,7 +146,7 @@ class ExchangeFinder(
       val callConnection = call.connection // changes within this overall method
       releasedConnection = callConnection
       toClose = if (callConnection != null && (callConnection.noNewExchanges ||
-              !callConnection.supportsUrl(address.url))) {
+              !sameHostAndPort(callConnection.route().address.url))) {
         call.releaseConnectionNoEvents()
       } else {
         null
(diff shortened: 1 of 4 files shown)


okhttp/src/main/java/okhttp3/internal/http/RealInterceptorChain.kt
```diff
@@ -84,7 +84,7 @@ class RealInterceptorChain(
     calls++
 
     if (exchange != null) {
-      check(exchange.connection.supportsUrl(request.url)) {
+      check(exchange.finder.sameHostAndPort(request.url)) {
```
swankjesse: This is a similar bug fix. If a network interceptor changes the URL from a.example.com to b.example.com, we want to explode even if the connection is actually capable of that request.

okhttp/src/main/java/okhttp3/internal/connection/ExchangeFinder.kt
```diff
@@ -333,4 +334,14 @@ class ExchangeFinder(
...
+
+  /**
+   * Returns true if the host and port are unchanged from when this was created. This is used to
+   * detect if followups need to do a full connection-finding process including DNS resolution, and
+   * certificate pin checks.
+   */
+  fun sameHostAndPort(url: HttpUrl): Boolean {
+    val routeUrl = address.url
+    return url.port == routeUrl.port && url.host == routeUrl.host
+  }
```
swankjesse: We do this on `ExchangeFinder` and not `Connection` because `ExchangeFinder` tracks current URL, whereas `Connection` tracks whichever URL it initially connected with.

okhttp/src/main/java/okhttp3/internal/http/RealInterceptorChain.kt
yschimke: We could technically get by without this one, the network interceptor that is changing the request may have prior knowledge that the request url change is within the same server?

swankjesse: Yep. But I think it’s safer to just forbid it and wait for that request to come to us. Otherwise we’re also bypassing certificate pinning on b.example.com and that’s awkward.

PR #5867, opened 2020-03-14
Hook up compression in WebSocketReader and WebSocketWriter

swankjesse:
It isn't yet enabled in the calling RealWebSocket, so this is only
visible to tests right now.

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/ws/WebSocketWriter.kt
+++ b/okhttp/src/main/java/okhttp3/internal/ws/WebSocketWriter.kt
@@ -15,9 +15,11 @@
  */
 package okhttp3.internal.ws
 
+import java.io.Closeable
 import java.io.IOException
 import java.util.Random
 import okhttp3.internal.ws.WebSocketProtocol.B0_FLAG_FIN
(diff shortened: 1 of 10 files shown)


swankjesse:
FYI @AntonRutkevich

okhttp/src/main/java/okhttp3/internal/ws/WebSocketWriter.kt
```diff
@@ -39,11 +41,15 @@ import okio.ByteString
...
+class WebSocketWriter(
   private val isClient: Boolean,
   val sink: BufferedSink,
-  val random: Random
-) {
+  val random: Random,
+  private val messageDeflater: MessageDeflater?,
+  private val minimumDeflateSize: Long
+) : Closeable {
+  /** This buffer is holds outbound data for compression and masking. */
```
yschimke: nit: "is hold"

yschimke:
LGTM

swankjesse:
Gonna fix typo in follow-up.

AntonRutkevich:
Much cleaner than the initial version due to previous refactoring. 👍

PR #5868, opened 2020-03-15
Allowlist Sample

yschimke:
An Android Sample for allowing requests to a specific dev server e.g. localhost.  All other traffic is still validated.

### DIFF
--- a/samples/guide/src/main/java/okhttp3/recipes/kt/DevServerJVM.kt
+++ b/samples/guide/src/main/java/okhttp3/recipes/kt/DevServerJVM.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2014 Square, Inc.
+ * Copyright (C) 2020 Square, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


yschimke:
Not sure how to test on JVM :(

```
Skipping all security checks
[CN=localhost]
Skipping all security checks
[CN=www.google.com, O=Google LLC, L=Mountain View, ST=California, C=US, CN=GTS CA 1O1, O=Google Trust Services, C=US]
https://www.google.com/robots.txt
```

yschimke:
Going to test with X509ExtendedTrustManager

samples/guide/src/main/java/okhttp3/recipes/kt/DevServerJVM.kt
```diff
@@ -0,0 +1,151 @@
...
+      delegate.checkServerTrusted(chain, authType, engine)
+    }
+  }
+
+  fun isAllowed(host: String): Boolean = hosts.contains(host)
+
+  override fun getAcceptedIssuers(): Array<X509Certificate> = delegate.acceptedIssuers
+}
+
+class DevServerJVM {
```
swankjesse: supernit: Jvm

yschimke:
This was the point of the issue I raised. What is the API like?

yschimke:
@swankjesse Going to land as it is for now, the work out what the nice public API for this would be.  I don't think it's trivial.

PR #5869, opened 2020-03-15
Get compression working in RealWebSocketTest

swankjesse:
The only thing left is to negotiate extensions, including failing the
call if the server returns unexpected extensions.

### DIFF
--- a/okhttp/src/main/java/okhttp3/OkHttpClient.kt
+++ b/okhttp/src/main/java/okhttp3/OkHttpClient.kt
@@ -248,12 +248,16 @@ open class OkHttpClient internal constructor(
 
   /** Uses [request] to connect a new web socket. */
   override fun newWebSocket(request: Request, listener: WebSocketListener): WebSocket {
+    // Compress messages 1 KiB or larger.
+    val minimumDeflateSize = 1024L
     val webSocket = RealWebSocket(
-        TaskRunner.INSTANCE,
-        request,
-        listener,
-        Random(),
-        pingIntervalMillis.toLong()
+        taskRunner = TaskRunner.INSTANCE,
+        originalRequest = request,
+        listener = listener,
+        random = Random(),
+        pingIntervalMillis = pingIntervalMillis.toLong(),
+        extensions = null, // Always null for clients.
+        minimumDeflateSize = minimumDeflateSize
     )
     webSocket.connect(this)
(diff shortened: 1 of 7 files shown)


okhttp/src/main/java/okhttp3/OkHttpClient.kt
```diff
@@ -248,12 +248,16 @@ open class OkHttpClient internal constructor(
...
-        request,
-        listener,
-        Random(),
-        pingIntervalMillis.toLong()
+        taskRunner = TaskRunner.INSTANCE,
+        originalRequest = request,
+        listener = listener,
+        random = Random(),
+        pingIntervalMillis = pingIntervalMillis.toLong(),
+        extensions = null, // Always null for clients.
```
yschimke: RealWebSocket is an internal class, feels like these two constructor invocations could be factory methods on the companion object? random, ping and extensions both diff.

okhttp/src/main/java/okhttp3/internal/ws/RealWebSocket.kt
```diff
@@ -56,7 +56,14 @@ class RealWebSocket(
   private val originalRequest: Request,
   internal val listener: WebSocketListener,
   private val random: Random,
-  private val pingIntervalMillis: Long
+  private val pingIntervalMillis: Long,
+  /**
+   * For clients this is initially null, and will be assigned to the agreed-upon extensions. For
+   * servers it should be the agreed-upon extensions immediately.
+   */
+  private var extensions: WebSocketExtensions?,
```
yschimke: Should this be a non-nullable lateinit?

okhttp/src/main/java/okhttp3/OkHttpClient.kt
swankjesse: I tried it and I ended up with clumsier code than when I started. I think the problem is that the tests are sort of clients and sort of servers, so we’d end up needing several ways to construct this.

okhttp/src/main/java/okhttp3/internal/ws/RealWebSocket.kt
swankjesse: I think this might be more awkward. Although it’s spiritually lateinit, I can’t assign null to that which makes the client constructor awkward.

PR #5871, opened 2020-03-15
Turn on web socket compression

swankjesse:
Theres a few TODOs outstanding around error cases, and I need to
run integration tests against real-world things. But the core
functionality works.

I used coverage to confirm the context takeover cases are exercised
with and without.

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/ws/RealWebSocket.kt
+++ b/okhttp/src/main/java/okhttp3/internal/ws/RealWebSocket.kt
@@ -144,6 +144,12 @@ class RealWebSocket(
   }
 
   fun connect(client: OkHttpClient) {
+    if (originalRequest.header("Sec-WebSocket-Extensions") != null) {
+      failWebSocket(ProtocolException(
+          "Request header not permitted: 'Sec-WebSocket-Extensions'"), null)
+      return
+    }
+
     val webSocketClient = client.newBuilder()
         .eventListener(EventListener.NONE)
         .protocols(ONLY_HTTP1)
@@ -153,6 +159,7 @@ class RealWebSocket(
         .header("Connection", "Upgrade")
         .header("Sec-WebSocket-Key", key)
         .header("Sec-WebSocket-Version", "13")
+        .header("Sec-WebSocket-Extensions", "permessage-deflate")
         .build()
     call = RealCall(webSocketClient, request, forWebSocket = true)
     call!!.enqueue(object : Callback {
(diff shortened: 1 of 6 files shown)


swankjesse:
I wanna do an OkHttp 4.5 'Release Candidate' with this feature so early adopters can check whether this works before everyone else does.

okhttp/src/main/java/okhttp3/internal/ws/RealWebSocket.kt
```diff
@@ -607,5 +637,15 @@ class RealWebSocket(
...
+
+    /**
+     * The smallest message that will be compressed. We use 1024 because smaller messages already
+     * fit comfortably within a single ethernet packet (1500 bytes) even with framing overhead.
+     *
+     * For tests this must be big enough to realize real compression on test messages like
+     * 'aaaaaaaaaa...'. Our tests check if compression was applied just by looking at the size if
+     * the inbound buffer.
+     */
+    const val DEFAULT_MINIMUM_DEFLATE_SIZE = 1024L
```
AntonRutkevich: @swankjesse , consider the following case: 

An app is sending a stream of coordinates, JSON encoded. It has dozen of fields (lat, lon, accuracy, heading, timestamp, ...), so the size of the uncompressed JSON is ~110..130B.
With context takeover, a stream of such messages gets compressed quite well (~10-20B per message). 

With all framing overhead, on-wire messages are ~230B uncompressed vs ~120B compressed. These are the ~real numbers we got from Wireshark experiments.

So, even for messages much smaller than 1024B, compression has significant effect on traffic consumption. In my understanding (might be wrong), traffic is measured by the actual IP level traffic, not rounded up to ethernet packet size.

With that, what do you think about making this at least somehow configurable via the OkHttp API? As a user, I'd be surprised to see that compression for my messages (all of which are under 1K) isn't working, despite being implemented in OkHttp.

swankjesse:
I think we should expose a configuration parameter on OkHttpClient.Builder! Want to send a PR?

AntonRutkevich:
Yep

AntonRutkevich:
#5890

okhttp/src/main/java/okhttp3/internal/ws/RealWebSocket.kt
```diff
@@ -144,6 +144,12 @@ class RealWebSocket(
   }
 
   fun connect(client: OkHttpClient) {
+    if (originalRequest.header("Sec-WebSocket-Extensions") != null) {
+      failWebSocket(ProtocolException(
+          "Request header not permitted: 'Sec-WebSocket-Extensions'"), null)
+      return
```
sbordet, line 150: @swankjesse I think this code is a little too restrictive to applications that want to specify their own extensions or other extensions that are not just the `permessage-deflate`.
I would rather merge the extensions provided by the application with the internal `permessage-deflate`, unless the application already specifies `permessage-deflate`. Thoughts?
FTR I have a CometD test that was testing this that fails with OkHttp 4.5.0. Thanks!

yschimke: Can you raise a separate issue to track and link to your test case?  Want to make sure this doesn't get lost if we don't immediately fix.

sbordet: @yschimke filed #5953, thanks!

okhttp/src/main/java/okhttp3/internal/ws/RealWebSocket.kt
```diff
@@ -162,15 +169,24 @@ class RealWebSocket(
...
-          extensions = WebSocketExtensions()
         } catch (e: IOException) {
           exchange?.webSocketUpgradeFailed()
           failWebSocket(e, response)
           response.closeQuietly()
           return
         }
 
+        // Apply the extensions. If they're unacceptable initiate a graceful shut down.
+        // TODO(jwilson): Listeners should get onFailure() instead of onClosing() + onClosed(1010).
```
nlgtuankiet, line 180: @swankjesse  
Is there any reason not to use failWebSocket() here and return?
If not I am happy to submit a pr to fix this TODO

Issue #5701, opened 2020-01-05
Change CustomTrust sample to use okhttp-tls


yschimke:
Forked from https://github.com/square/okhttp/pull/5699/files

PR #5872, opened 2020-03-16
insecureForHost API for TrustManager overrides

yschimke:
There is a potentially interesting API behind insecureForHost, that we could decide to make public in okhttp-tls later.  For now, having a correct, minimally insecure implementation of clientBuilder.insecureForHost("localhost") will solve a lot of stackoverflow questions.

I plan to harvest all the stackoverflow points... :)  This was always the long game.


okhttp/src/main/java/okhttp3/OkHttpTrustManager.kt
```diff
@@ -0,0 +1,159 @@
...
+import java.net.Socket
+import java.security.cert.Certificate
+import java.security.cert.X509Certificate
+import javax.net.ssl.SSLEngine
+import javax.net.ssl.X509ExtendedTrustManager
+import javax.net.ssl.X509TrustManager
+import okhttp3.internal.platform.Platform
+import org.codehaus.mojo.animal_sniffer.IgnoreJRERequirement
+
+interface OkHttpTrustManager : X509TrustManager {
```
yschimke: @swankjesse I'm not sold on this.  It's making the public API we wanted to avoid.  This draft is musing on the right API if we want to support it.  Not sure why else you would want to override the TrustManager?

It's probably to have two distinct TrustManagers e.g. from different keystores?

So maybe it's a selecting implementation and we provide a utility like OkHttpTrustManager.INSECURE that would only be used for the devserver.

okhttp/src/main/java/okhttp3/OkHttpTrustManager.kt
```diff
@@ -0,0 +1,159 @@
...
+  fun checkServerTrusted(
+    chain: Array<out X509Certificate>,
+    authType: String,
+    host: String
+  ): List<Certificate>?
+
+  val delegate: X509TrustManager
```
yschimke: Maybe this should change to a Function `(String) -> X509TrustManager` and delete allowedPredicate.

okhttp/src/main/java/okhttp3/OkHttpTrustManager.kt
swankjesse: Do people want to mix a secure and an insecure trust manager? I suspect an `INSECURE_TRUST_MANAGER` constant might be the easiest way to get up and running quickly?

Perhaps even something that doesn’t expose the TrustManager type:

```
     .insecureTrustManager("🛑 NOT FOR PRODUCTION USE 🛑")
     .build()
```

In the above example we’d require that literal string to be passed in; any other string would throw.

yschimke: Give me one more go :) before we get to magic strings...

samples/guide/src/main/java/okhttp3/recipes/kt/DevServer.kt
```diff
@@ -114,7 +51,7 @@ class DevServerAndroid {
   val hosts = arrayOf(server.hostName)
 
   val platformTrustManager = platformTrustManager()
-  val trustManager = AndroidAllowlistedTrustManager(platformTrustManager, *hosts)
+  val trustManager = OkHttpTrustManager.hostOverride(platformTrustManager, "localhost", InsecureTrustManager)
```
yschimke: I like how this sample ended up because we never implement the insecure mode ourselves, but give a clean "as safe as possible" sample given a concrete real need.

okhttp/src/main/java/okhttp3/internal/platform/android/OkHttpTrustManagerAndroid.kt
```diff
@@ -0,0 +1,64 @@
...
+) : OkHttpTrustManager {
+  val delegateMethod = lookupDelegateMethod()
+
+  fun checkServerTrusted(
+    chain: Array<out X509Certificate>,
+    authType: String,
+    host: String
+  ): List<Certificate> {
+
+    TODO()
```
yschimke: Ignore the Android path for now.

samples/guide/src/main/java/okhttp3/recipes/kt/DevServer.kt
```diff
@@ -114,7 +51,7 @@ class DevServerAndroid {
...
   val sslSocketFactory = Platform.get().newSSLContext().apply {
```
yschimke: This sample should be cleaned up to not use Platform.

okhttp/src/main/java/okhttp3/OkHttpTrustManager.kt
swankjesse: Do we need a new interface for this thing? Seems like the existing TrustManager interface is sufficient!

swankjesse:
From this morning:
https://stackoverflow.com/questions/60735480/okhttp3-returns-sslhandshakeexception-handshake-failed

yschimke:
> From this morning:
> https://stackoverflow.com/questions/60735480/okhttp3-returns-sslhandshakeexception-handshake-failed

I think I can crack this one, will take a look later after work.

What do you think about making this extend the work on okhttp-tls?  Make it a utility there?

swankjesse:
I’d love to do a feature that’s very close to your allowlist. Something where I can turn off TLS stuff for some hosts that don’t do it right.

```
    .insecureForHosts("localhost", "192.168.1.30")
    .build()
```

yschimke:
Plaintext http? Or insecure Https?

swankjesse:
Insecure HTTPS.

yschimke:
@swankjesse Take another look on direction

samples/guide/src/main/java/okhttp3/recipes/kt/DevServer.kt
yschimke: @swankjesse How can we avoid Platform here and dealing with sslSocketFactory.  Just using OkHttpTrustManager and heldCertificates.

Should OkHttpTrustManager be keyManager aware?

okhttp-tls/src/main/java/okhttp3/tls/internal/OkHttpTrustManager.kt
```diff
@@ -0,0 +1,118 @@
...
+    val tm = findByHost(engine.peerHost)
+
+    if (tm is X509ExtendedTrustManager) {
+      tm.checkServerTrusted(chain, authType, engine)
+    } else {
+      tm.checkServerTrusted(chain, authType)
+    }
+  }
+
+  override fun getAcceptedIssuers(): Array<X509Certificate> = default.acceptedIssuers
```
yschimke: Should this be a set union of all accepted issuers?

yschimke:
Testing update

```
Checking localhost against InsecureTrustManager 0
Checking localhost against X509TrustManagerImpl 1
Running security checks for localhost using X509TrustManagerImpl
[CN=localhost]
Checking www.google.com against InsecureTrustManager 0
Checking www.google.com against X509TrustManagerImpl 1
Using default X509TrustManagerImpl 88
Running security checks for www.google.com using X509TrustManagerImpl
[CN=www.google.com, O=Google LLC, L=Mountain View, ST=California, C=US]
https://www.google.com/robots.txt
```

samples/guide/src/main/java/okhttp3/recipes/kt/DevServer.kt
swankjesse: It is really convenient to have all the stuff you need for an SSLContext all in one place. Though the API is still tricky to get right.

yschimke:
> I’m imagining the entire API surface area could be one new method on OkHttpClient.Builder `insecureHttpsHost` that takes a hostname. Internally we could delegate or whatever, but the user doesn’t need to see the word trust manager or understand that "trust manager" means "certificate authorities" (or even to know what certificate authorities are!).

You did too good a job convincing me this was a bad idea previously.

I suspect in a lot of these cases the devserver has internal CA issued certificates available in dev builds.  So I was aiming for that case.  But with a fallback also for insecure.  I'll try another loop with just the public API you suggest, since we still need the majority of this code related to bridging both the Android and JVM TrustManager extensions.

FWIW - I've tackled this beast once before, and it wasn't fun, and I wished there was something like okhttp-tls that went slightly further.  In that case it was related to required client auth for Estonian ID smartcard auth.  In that case I'm not even sure my code would work on Android.

https://github.com/yschimke/okurl/blob/0520489d697d49b179010e468a87ef0749ff95be/src/main/kotlin/com/baulsupp/okurl/security/CertificateUtils.kt
https://github.com/yschimke/okurl/tree/master/certificates

https://github.com/yschimke/okurl/blob/0520489d697d49b179010e468a87ef0749ff95be/src/main/kotlin/com/baulsupp/okurl/security/OpenSCUtil.kt

swankjesse:
Today: https://www.reddit.com/r/androiddev/comments/fl7jjm/how_can_i_use_custom_and_default_ssl_certificate/

okhttp/src/main/java/okhttp3/OkHttpClient.kt
```diff
@@ -1027,6 +1028,19 @@ open class OkHttpClient internal constructor(
       pingInterval(duration.toMillis(), MILLISECONDS)
     }
 
+    fun insecureTrustManager(hostName: String): Builder {
+      // TODO could build this up and defer to later, to avoid overwriting other changes
```
yschimke: TrustManagerBridge.Builder could short circuit to use the default only if one isn't provided. On the other hand, this functionality also fixes this weirdness

https://square.github.io/okhttp/4.x/okhttp/okhttp3/-ok-http-client/-builder/ssl-socket-factory/#trustmanagers-on-android-are-weird

yschimke:
@swankjesse minimal public API, but with internal classes until I can convince you on a public API.

okhttp/src/main/java/okhttp3/OkHttpClient.kt
yschimke: Going to refactor this, do this at the last minute when overrides are introduced.

okhttp/src/main/java/okhttp3/OkHttpClient.kt
```diff
@@ -785,6 +785,8 @@ open class OkHttpClient internal constructor(
      *    }
      * ```
      *
+     * n.b. OkHttp will generate a wrapper with this method for you, if you don't.
```
yschimke: Might remove this, it seems problematic.  We either use TrustManagerImpl or we fake with emptyList()

yschimke: Generally we need this because the Default TrustManager on Android is TrustManagerImpl.  No one else should implement that logic.  So if you use overrides we currently have two choices

1) You wrap an existing TrustManagerImpl, and delegate to that.  We can define an optional typed interface to extend that allows for that.
2) You don't e.g. InsecureTrustManager - in which case we shouldn't ask people to implement it, we should just return an emptyList() ourselves.

But we should make it simpler to get right.

yschimke: OkHttpTrustManagerJvm does something vaguely similar, if you do implement X509ExtendedTrustManager we use that, otherwise we drop the socket param and call the check check method.

okhttp/src/main/java/okhttp3/internal/platform/ConscryptPlatform.kt
```diff
@@ -41,7 +41,9 @@ class ConscryptPlatform private constructor() : Platform() {
       SSLContext.getInstance("TLS", provider)
 
   override fun platformTrustManager(): X509TrustManager {
-    return Conscrypt.getDefaultX509TrustManager()
+    return Conscrypt.getDefaultX509TrustManager().also {
```
yschimke: We do this proactively because TrustManagerBridge will wrap this call, so it's hard to do otherwise.

yschimke:
> Today: https://www.reddit.com/r/androiddev/comments/fl7jjm/how_can_i_use_custom_and_default_ssl_certificate/

Answer 

"different hosts completly"

Basically this person wants this fix.

android-test/src/androidTest/java/okhttp/android/test/OkHttpTest.kt
```diff
@@ -517,8 +541,8 @@ class OkHttpTest {
       // Hopefully this passes
     } catch (ioe: IOException) {
       // https://github.com/square/okhttp/issues/5840
-      assertEquals("Android internal error", ioe.message)
-      assertEquals(IllegalArgumentException::class.java, ioe.cause!!.javaClass)
+      val exceptions = generateSequence(ioe as Throwable) { e -> e.cause }.toList()
```
yschimke: Not predictable on all Android versions where this error will surface.

yschimke:
For after 4.5...

okhttp/src/main/java/okhttp3/OkHttpClient.kt
```diff
@@ -224,13 +227,23 @@ open class OkHttpClient internal constructor(
...
+    }
+
     if (builder.sslSocketFactoryOrNull != null || connectionSpecs.none { it.isTls }) {
       this.sslSocketFactoryOrNull = builder.sslSocketFactoryOrNull
       this.certificateChainCleaner = builder.certificateChainCleaner
       this.x509TrustManager = builder.x509TrustManagerOrNull
     } else {
-      this.x509TrustManager = Platform.get().platformTrustManager()
-      Platform.get().configureTrustManager(x509TrustManager)
+      this.x509TrustManager = builder.x509TrustManagerOrNull ?: Platform.get().platformTrustManager()
```
yschimke: Provided we are ok with this line, the whole change *should* be a noop if the new method isn't used.  Consider that in the scope of the change.

android-test/src/androidTest/java/okhttp/android/test/OkHttpTest.kt
swankjesse: Neat.

okhttp/src/main/java/okhttp3/OkHttpClient.kt
```diff
@@ -1049,6 +1071,29 @@ open class OkHttpClient internal constructor(
       this.minWebSocketMessageToCompress = bytes
     }
 
+    /**
+     * Set an override for host to allow insecure connections, typically a devserver with
+     * a self-signed certificate unavailable to the dev build of the client.
+     *
+     * This will override any existing sslSocketFactory with the platform default.
+     *
+     * n.b. not for production use, only available from Android version 26.
```
swankjesse: We can make this work on all versions, can’t we?

okhttp/src/main/java/okhttp3/OkHttpClient.kt
```diff
@@ -1049,6 +1071,29 @@ open class OkHttpClient internal constructor(
...
+     *
+     * @param the exact hostname from the URL for insecure connections.
```
swankjesse: nit: you need the param name here, don’t you?

mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.kt
```diff
@@ -275,7 +275,7 @@ class MockWebServer : ExternalResource(), Closeable {
    *
    * @param tunnelProxy true to expect the HTTP CONNECT method before negotiating TLS.
    */
-  fun useHttps(sslSocketFactory: SSLSocketFactory, tunnelProxy: Boolean) {
+  fun useHttps(sslSocketFactory: SSLSocketFactory, tunnelProxy: Boolean = false) {
```
yschimke: Is this safe?

okhttp/src/main/java/okhttp3/OkHttpClient.kt
yschimke: We can implement it working, but Android only calls the method with an additional host on some versions.  I suspect it's actually version 23 -> 24.

e.g. tied in with

```

  override fun isCleartextTrafficPermitted(hostname: String): Boolean {
    return try {
      val networkPolicyClass = Class.forName("android.security.NetworkSecurityPolicy")
      val getInstanceMethod = networkPolicyClass.getMethod("getInstance")
      val networkSecurityPolicy = getInstanceMethod.invoke(null)
      api24IsCleartextTrafficPermitted(hostname, networkPolicyClass, networkSecurityPolicy)
    } catch (_: ClassNotFoundException) {
      super.isCleartextTrafficPermitted(hostname)
    } catch (_: NoSuchMethodException) {
      super.isCleartextTrafficPermitted(hostname)
    } catch (e: IllegalAccessException) {
      throw AssertionError("unable to determine cleartext support", e)
    } catch (e: IllegalArgumentException) {
      throw AssertionError("unable to determine cleartext support", e)
    } catch (e: InvocationTargetException) {
      throw AssertionError("unable to determine cleartext support", e)
    }
  }
```

yschimke: https://github.com/aosp-mirror/platform_frameworks_base/commit/93962c2f0118bcda78ffde7e0b6be73da7ab2500#diff-9c086e3a85d0f86ba44bf971ebca1171

https://github.com/aosp-mirror/platform_frameworks_base/commit/0f5d386a769cafa3f4faf93f17acb22503b6fd3e#diff-9c086e3a85d0f86ba44bf971ebca1171

yschimke:
@swankjesse Are you ok with this minimal API for now? And that it's effectively a no-op when not activated.

okhttp/src/main/java/okhttp3/OkHttpClient.kt
yschimke: See https://github.com/square/okhttp/pull/5960

okhttp/src/main/kotlin/okhttp3/OkHttpClient.kt
```diff
@@ -1066,6 +1096,29 @@ open class OkHttpClient internal constructor(
       this.minWebSocketMessageToCompress = bytes
     }
 
+    /**
+     * Set an override for host to allow insecure connections, typically a devserver with
```
swankjesse, line 1100: Here’s documentation that I think explains what we need to explain here:

> Configures OkHttp to not authenticate the HTTPS server on to [hostName]. This makes the client vulnerable to man-in-the-middle attacks and should only be used only in private development environments and only to carry test data.
>
> The following protections are removed:
>
> * The server’s TLS certificate **does not need to be signed** by a trusted certificate authority. Instead, OkHttp will trust any well-formed certificate, even if it is self-signed. This is necessary for testing against localhost, or in development environments where a certificate authority is not possible.
> * The server’s TLS certificate **does not need to match** the requested hostname. For example, if the certificate is issued to `example.com` and the request is to `localhost`, the certificate will still be accepted!
>
> Other TLS features are still used but provide no security benefits in absence of the above gaps. For example, an insecure TLS connection is capable of negotiating HTTP/2 with ALPN and it also has a regular-looking handshake.

Dig?

swankjesse: There’s consequences here for our connection pools. We’ll need a test to confirm that the custom SSL objects implement enough equals and hashCode stuff to permit their connections to be pooled, and also enough to prevent their connections from being shared with OkHttpClient instances that don’t have these options.

okhttp/src/main/kotlin/okhttp3/internal/tls/OkHostnameVerifier.kt
```diff
@@ -32,6 +32,8 @@ import okhttp3.internal.toCanonicalHost
  */
 @Suppress("NAME_SHADOWING")
 object OkHostnameVerifier : HostnameVerifier {
+  val Insecure = HostnameVerifier { _, _ -> true }
```
swankjesse, line 35: INSECURE?

okhttp/src/main/kotlin/okhttp3/internal/tls/TrustManagerAndroid.kt
```diff
@@ -0,0 +1,133 @@
...
+import javax.net.ssl.X509TrustManager
+import okhttp3.internal.SuppressSignatureCheck
+
+internal open class TrustManagerOverrideAndroid(
+  val predicate: (String) -> Boolean,
+  val trustManager: AndroidTrustManager
+)
+
+internal class TrustManagerWrapperAndroid(val trustManager: X509TrustManager) : AndroidTrustManager {
+  val delegateMethod by lazy { lookupAndroidDelegateMethod() }
```
swankjesse, line 31: should this be non-lazy to fail fast?

okhttp/src/main/kotlin/okhttp3/internal/tls/TrustManagerBridge.kt
```diff
@@ -0,0 +1,56 @@
...
+package okhttp3.internal.tls
+
+import android.annotation.SuppressLint
+import javax.net.ssl.X509ExtendedTrustManager
+import javax.net.ssl.X509TrustManager
+import okhttp3.internal.platform.Platform
+import org.codehaus.mojo.animal_sniffer.IgnoreJRERequirement
+
+object TrustManagerBridge {
+  class Builder(var default: X509TrustManager? = null) {
```
swankjesse, line 25: why the enclosing object?

okhttp/src/main/kotlin/okhttp3/internal/tls/TrustManagerBridge.kt
```diff
@@ -0,0 +1,56 @@
...
+    private var overrides: MutableList<TrustManagerOverride> = mutableListOf()
+
+    fun default(default: X509TrustManager) = apply {
+      this.default = default
+    }
+
+    fun hostOverride(hostName: String, trustManager: X509TrustManager) = apply {
```
swankjesse, line 32: still useful?

okhttp/src/main/kotlin/okhttp3/internal/tls/TrustManagerJvm.kt
```diff
@@ -0,0 +1,107 @@
...
+import java.net.Socket
+import java.security.cert.X509Certificate
+import javax.net.ssl.SSLEngine
+import javax.net.ssl.X509ExtendedTrustManager
+import javax.net.ssl.X509TrustManager
+import org.codehaus.mojo.animal_sniffer.IgnoreJRERequirement
+
+@IgnoreJRERequirement
+@SuppressLint("NewApi")
+class TrustManagerJvm internal constructor(
```
swankjesse, line 28: why do these need to be separate types? Can’t we implement once for all? Does the Android one break on the JVM?

okhttp/src/main/kotlin/okhttp3/internal/tls/TrustManagerOverride.kt
```diff
@@ -0,0 +1,25 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okhttp3.internal.tls
+
+import javax.net.ssl.HostnameVerifier
+import javax.net.ssl.X509TrustManager
+
+open class TrustManagerOverride(
```
swankjesse, line 21: should this just be a hostname since our configuration is just on or off?

android-test/src/androidTest/java/okhttp/android/test/OkHttpTest.kt
```diff
@@ -327,6 +338,122 @@ class OkHttpTest {
     }
   }
 
+  @Test
+  fun testDevserverSupport() {
```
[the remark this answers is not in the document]
yschimke: In this case, it is to check we are still using the real CA file.  Will add comment and assumeNetwork.

okhttp/src/main/kotlin/okhttp3/OkHttpClient.kt
```diff
@@ -227,18 +231,35 @@ open class OkHttpClient internal constructor(
...
-          .withCertificateChainCleaner(certificateChainCleaner!!)
+      this.hostnameVerifier = OkHostnameVerifier
     } else {
-      this.x509TrustManager = Platform.get().platformTrustManager()
-      this.sslSocketFactoryOrNull = Platform.get().newSslSocketFactory(x509TrustManager!!)
-      this.certificateChainCleaner = CertificateChainCleaner.get(x509TrustManager!!)
-      this.certificatePinner = builder.certificatePinner
-          .withCertificateChainCleaner(certificateChainCleaner!!)
+      if (builder.trustManagerOverrides != null) {
+        this.x509TrustManager = TrustManagerBridge.Builder()
```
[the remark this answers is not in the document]
yschimke: There is a bit of awkward logic to separate out by platform (JVM/Android), but let's try without.

okhttp/src/main/kotlin/okhttp3/internal/tls/TrustManagerJvm.kt
yschimke: Different class hierarchies, different methods, different semantics.  On JVM we need it to implement X509ExtendedTrustManager, on Android we can't even rely on that type, and those methods can never be called.

okhttp/src/main/kotlin/okhttp3/internal/tls/TrustManagerOverride.kt
yschimke: Dumbing it down to hostname for now.  FWIW a lot of this complexity comes from me thinking ahead to far to something more complicated e.g. setting up intranet certificate policies based on wildcards.  Removing all that for now.

okhttp/src/main/kotlin/okhttp3/OkHttpClient.kt
yschimke: +1 I'll add a test

swankjesse:
It might be useful to consider how browsers’ trust manager mechanisms work. By default you get a well known set of root CAs. You can install your own CA for institutional use. And you can click through an angry "TLS is sketchy” warning on dev server's and sites that are misconfigured.

My preference for this feature is to resemble that angry TLS is sketchy page. 

What triggers that page? Untrusted CAs? Hostname mismatches? Expired certs? Incomplete chains? (We should test how expired certs work!)

For institutional our best strategy is to encourage users to add their trusted CAs instead of using this feature. Making it easier to accept a PEM file could be useful here. (I think our custom trust recipe might need an update to use okhttp-tls).

yschimke:
> Making it easier to accept a PEM file could be useful here. (I think our custom trust recipe might need an update to use okhttp-tls).

Already have a tracking issue for that https://github.com/square/okhttp/issues/5701

yschimke:
> My preference for this feature is to resemble that angry TLS is sketchy page.

I like this idea, somewhat like the CertificatePinner failure messages.  I think eventually we should bridge the work of okhttp-tls with failure cases.  Unfortunately the errors from TLS implementations are totally incomprehensible to me.  Poor users.  FWIW if you try to use something like 2FA at work, or European smartcards with Firefox or Chrome, it's similar.  Technically secure, but with esoteric* errors popping up.

* happy to use byzantine, just because it's a cool word.

yschimke:
@swankjesse I think landing this carefully in the minimal devserver form first (I was trying for too broad a solution initially, towards where you are talking) is wise.  Particularly being comfortable that if you don't activate it that it can't break security because it isn't invoked.

Then if we think it's useful work on a nice internal API for the cases you talk about and when ready make it public.

okhttp/src/main/kotlin/okhttp3/OkHttpClient.kt
```diff
@@ -743,6 +768,10 @@ open class OkHttpClient internal constructor(
         level = DeprecationLevel.ERROR
     )
     fun sslSocketFactory(sslSocketFactory: SSLSocketFactory) = apply {
+      check(trustManagerOverrides == null) {
```
swankjesse, line 771: Mostly curious – is this an intrinsic limitation or just something to implement?

okhttp/src/main/kotlin/okhttp3/OkHttpClient.kt
```diff
@@ -1066,6 +1099,42 @@ open class OkHttpClient internal constructor(
...
+     * - The server’s TLS certificate **does not need to match** the requested hostname. For
+     * example, if the certificate is issued to `example.com` and the request is to `localhost`,
+     * the certificate will still be accepted!
+     *
+     * Other TLS features are still used but provide no security benefits in absence of the above
+     * gaps. For example, an insecure TLS connection is capable of negotiating HTTP/2 with ALPN and
+     * it also has a regular-looking handshake.
+     *
+     * n.b. only available from Android version 24, due to use of host specific certificate checks
+     * added in Android.
```
swankjesse, line 1121: Hang on, how do Android developers to development testing on older releases? Can’t we do that too?

okhttp/src/main/kotlin/okhttp3/OkHttpClient.kt
```diff
@@ -1066,6 +1099,42 @@ open class OkHttpClient internal constructor(
...
+      val override = TrustManagerOverride(hostname, OkHostnameVerifier.INSECURE, InsecureTrustManager)
+
+      val overrides = trustManagerOverrides ?: mutableListOf<TrustManagerOverride>().also {
+        this.trustManagerOverrides = it
+      }
+
+      overrides.add(override)
+
+      // invalidate and rebuild sslSocketFactory later
+      this.sslSocketFactoryOrNull = null
```
swankjesse, line 1135: I don’t like how it silently throws away your configured sslSocketFactory if you call these methods in a different order

okhttp/src/main/kotlin/okhttp3/internal/tls/TrustManagerAndroid.kt
```diff
@@ -0,0 +1,133 @@
...
+  override fun checkServerTrusted(
+    chain: Array<X509Certificate>,
+    authType: String,
+    hostname: String
+  ): List<X509Certificate> {
+    if (trustManager is AndroidTrustManager) {
+      return trustManager.checkServerTrusted(chain, authType, hostname)
+    }
+
+    // TODO review security implications here of orEmpty
```
swankjesse, line 53: I suspect it violates a contract? Returning 0 certificates feels like it’ll cause crashes elsewhere

okhttp/src/main/kotlin/okhttp3/internal/tls/AndroidTrustManager.kt
```diff
@@ -0,0 +1,25 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okhttp3.internal.tls
+
+import java.security.cert.X509Certificate
+import javax.net.ssl.X509TrustManager
+
+interface AndroidTrustManager : X509TrustManager {
```
swankjesse, line 21: This exists just so we can delegate around internally without using reflection?

okhttp/src/main/kotlin/okhttp3/internal/tls/TrustManagerAndroid.kt
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  override fun checkClientTrusted(chain: Array<out X509Certificate>, authType: String) {
+    trustManager.checkClientTrusted(chain, authType)
+  }
+
+  override fun getAcceptedIssuers(): Array<X509Certificate> = trustManager.acceptedIssuers
+
+  // Allows to keep using the X509TrustManagerExtensions to clean certificates
+  override fun checkServerTrusted(
```
swankjesse, line 44: What do you think about adding an extension function on TrustManager that does this?

```
fun TrustManager.checkServerTrusted(chain: Array<X509Certificate>, authType: String, hostname: String) {
  // reflectively call the method on 'this' !
}
```

PR #5873, opened 2020-03-17
Compare OkHttp vs. other popular HTTP clients

swankjesse:


samples/compare/build.gradle
```diff
@@ -0,0 +1,16 @@
+apply plugin: 'kotlin-kapt'
```
JakeWharton: org.jetbrains.kotlin.kapt

Also apply .jvm first?

JakeWharton: Wait you're not even using kapt. Just jvm then?

swankjesse: Copy pasta error. Removed!

Issue #5876, opened 2020-03-18
Web Sockets should lazily create the Inflater and Deflater


swankjesse:
We're creating these eagerly when compression is negotiated. We should consider creating them on-demand as it's possible no messages end up using compression due to message size thresholds.

PR #5877, opened 2020-03-19
Don't create inflaters or deflaters until they're needed

swankjesse:
They consume a decent amount of memory and it's possible that
they won't even be used, even if compression is negotiated.

Closes: https://github.com/square/okhttp/issues/5876

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/ws/WebSocketReader.kt
+++ b/okhttp/src/main/java/okhttp3/internal/ws/WebSocketReader.kt
@@ -55,7 +55,8 @@ class WebSocketReader(
   private val isClient: Boolean,
   val source: BufferedSource,
   private val frameCallback: FrameCallback,
-  private val messageInflater: MessageInflater?
+  private val perMessageDeflate: Boolean,
+  private val noContextTakeover: Boolean
 ) : Closeable {
   private var closed = false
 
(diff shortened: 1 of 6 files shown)


okhttp/src/main/java/okhttp3/internal/ws/WebSocketReader.kt
```diff
@@ -69,6 +70,9 @@ class WebSocketReader(
   private val controlFrameBuffer = Buffer()
   private val messageFrameBuffer = Buffer()
 
+  /** Lazily initialized on first use. */
+  private var messageInflater: MessageInflater? = null
```
yschimke: FWIW I still prefer

```
    private val messageInflater = lazy(mode = LazyThreadSafetyMode.NONE) { MessageInflater(noContextTakeover) }
```

JakeWharton: That allocates and causes Kotlin to generate a `KProperty[]` and eagerly initialize all of the `KProperty` instances on classload. Very bad.

yschimke: OK, so not the type of delegate, but the use of a delegate full stop.  Thanks

https://kotlinlang.org/docs/reference/delegated-properties.html#translation-rules

JakeWharton: For now. They're working on eliminating the `KProperty[]` and `KProperty`, but you'll still have to allocate the delegate instance, though that's not too bad.

yschimke:
Is this waiting post 4.5? OR should be an RC2?

swankjesse:
Nah, I’ve just been busy. The last 24 hours has been a very long year.

yschimke:
> Nah, I’ve just been busy. The last 24 hours has been a very long year.

Unacceptable! Sorry if sounded like I was hassling, was just curious.

PR #5886, opened 2020-03-21
Android comparison

yschimke:


yschimke:
FWIW Jetty doesn't work because of MethodHandles being used in bytecode

android-test/src/androidTest/java/okhttp/android/test/compare/ApacheHttpClientTest.kt
```diff
@@ -0,0 +1,51 @@
...
+  @After fun tearDown() {
+    httpClient.close()
+  }
+
+  @Test fun get() {
+    val request = HttpGet("https://google.com/robots.txt")
+
+    httpClient.execute(request).use { response ->
+      assertThat(response.code).isEqualTo(200)
+      // TODO reenable ALPN later
```
swankjesse: what does this mean?

android-test/src/androidTest/java/okhttp/android/test/compare/OkHttpClientTest.kt
```diff
@@ -0,0 +1,48 @@
...
+class OkHttpClientTest {
+  private var client = OkHttpClient()
+
+  @Test fun get() {
+    val request = Request.Builder()
+        .url("https://google.com/robots.txt")
+        .build()
+    client.newCall(request).execute().use { response ->
+      assertThat(response.code).isEqualTo(200)
+      assertThat(response.protocol).isEqualTo(Protocol.HTTP_2)
```
swankjesse: Nice.

android-test/src/androidTest/java/okhttp/android/test/compare/ApacheHttpClientTest.kt
yschimke: It means it didn't work by default, and I couldn't find an example with the Sync client, so I suspect it may just work with the Async client or need code like

https://hc.apache.org/httpcomponents-client-5.0.x/httpclient5/examples/AsyncClientTlsAlpn.java

yschimke: ¯\_(ツ)_/¯

PR #5888, opened 2020-03-21
Give up when ExchangeFinder exhausts all routes.

dave-r12:
https://github.com/square/okhttp/issues/3308

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/connection/ExchangeFinder.kt
+++ b/okhttp/src/main/java/okhttp3/internal/connection/ExchangeFinder.kt
@@ -113,13 +113,23 @@ class ExchangeFinder(
           connectionRetryEnabled = connectionRetryEnabled
       )
 
-      // Confirm that the connection is good. If it isn't, take it out of the pool and start again.
-      if (!candidate.isHealthy(doExtensiveHealthChecks)) {
-        candidate.noNewExchanges()
-        continue
+      // Confirm that the connection is good.
+      if (candidate.isHealthy(doExtensiveHealthChecks)) {
+        return candidate
       }
 
-      return candidate
+      // If it isn't, take it out of the pool.
+      candidate.noNewExchanges()
+
+      // Make sure we have some routes left to try. One example where we may exhaust all the routes
+      // would happen if we made a new connection and it immediately is detected as unhealthy.
+      synchronized(connectionPool) {
+        val routesLeft = routeSelection?.hasNext() ?: true
+        val routesSelectionLeft = routeSelector?.hasNext() ?: true
+        if (nextRouteToTry == null && !routesLeft && !routesSelectionLeft) {
+          throw IOException("exhausted all routes")
+        }
+      }
     }
   }
 
--- a/okhttp/src/test/java/okhttp3/CallTest.java
+++ b/okhttp/src/test/java/okhttp3/CallTest.java
@@ -61,6 +61,7 @@ import okhttp3.CallEvent.ConnectionReleased;
 import okhttp3.CallEvent.ResponseFailed;
 import okhttp3.internal.DoubleInetAddressDns;
 import okhttp3.internal.RecordingOkAuthenticator;
+import okhttp3.internal.Util;
 import okhttp3.internal.Version;
 import okhttp3.internal.http.RecordingProxySelector;
 import okhttp3.internal.io.InMemoryFileSystem;
@@ -3790,6 +3791,20 @@ public final class CallTest {
       assertThat(source.exhausted()).isTrue();
       assertThat(response.trailers()).isEqualTo(Headers.of("trailers", "boom"));
     }
+  }
+
+  @Test public void connectionIsImmediatelyUnhealthy() throws Exception {
+    EventListener listener = new EventListener() {
+      @Override public void connectionAcquired(Call call, Connection connection) {
+        Util.closeQuietly(connection.socket());
+      }
+    };
+
+    client = client.newBuilder()
+        .eventListener(listener)
+        .build();
+
+    executeSynchronously("/").assertFailure(IOException.class);
   }
 
   @Test public void requestBodyThrowsUnrelatedToNetwork() throws Exception {


dave-r12:
There is a (extremely rare?) scenario where we may make a connection to a server that is then flagged as unhealthy. Currently 4.4.0 and master will throw a `NoSuchElementException`.

I ran this test against 4.3.1 and we appear to go an infinite loop. I'm not sure the right action to take in this scenario. Let me know your thoughts.

yschimke:
Not sure about the right fix.  Seems like worth fixing in 4.5, probably commit this behind @Ignore first?

swankjesse:
Great test

yschimke:
This PR is the hit sequel to

image

okhttp/src/main/java/okhttp3/internal/connection/ExchangeFinder.kt
```diff
@@ -113,13 +113,23 @@ class ExchangeFinder(
...
+        return candidate
       }
 
-      return candidate
+      // If it isn't, take it out of the pool.
+      candidate.noNewExchanges()
+
+      // Make sure we have some routes left to try. One example where we may exhaust all the routes
+      // would happen if we made a new connection and it immediately is detected as unhealthy.
+      synchronized(connectionPool) {
```
yschimke: Should we put a 
    connectionPool.assertThreadDoesntHoldLock()

At the top of this method?  I find it really really useful to document the current expected state of locks whenever we then open into a lock.

okhttp/src/main/java/okhttp3/internal/connection/ExchangeFinder.kt
```diff
@@ -113,13 +113,23 @@ class ExchangeFinder(
...
+        val routesLeft = routeSelection?.hasNext() ?: true
```
yschimke: It seems like line 108 must have made routeSelection non null, so can we change to !!

okhttp/src/main/java/okhttp3/internal/connection/ExchangeFinder.kt
```diff
@@ -113,13 +113,23 @@ class ExchangeFinder(
...
+        val routesSelectionLeft = routeSelector?.hasNext() ?: true
+        if (nextRouteToTry == null && !routesLeft && !routesSelectionLeft) {
+          throw IOException("exhausted all routes")
```
yschimke: We can probably be more expressive in this error, it's an interesting case when routes are immediately unhealthy.

yschimke:
Nice PR!

okhttp/src/main/java/okhttp3/internal/connection/ExchangeFinder.kt
```diff
@@ -113,13 +113,23 @@ class ExchangeFinder(
...
-      return candidate
+      // If it isn't, take it out of the pool.
+      candidate.noNewExchanges()
+
+      // Make sure we have some routes left to try. One example where we may exhaust all the routes
+      // would happen if we made a new connection and it immediately is detected as unhealthy.
+      synchronized(connectionPool) {
+        val routesLeft = routeSelection?.hasNext() ?: true
+        val routesSelectionLeft = routeSelector?.hasNext() ?: true
+        if (nextRouteToTry == null && !routesLeft && !routesSelectionLeft) {
```
swankjesse: I’d be tempted to change this to short circuit more aggressively?

```
      // Make sure we have some routes left to try. One example where we may exhaust all the routes
      // would happen if we made a new connection and it immediately is detected as unhealthy.
      synchronized(connectionPool) {
        if (nextRouteToTry != null) continue

        val routesLeft = routeSelection?.hasNext() ?: true
        if (routesLeft) continue

        val routesSelectionLeft = routeSelector?.hasNext() ?: true
        if (routesSelectionLeft) continue

        throw IOException("exhausted all routes")
      }
```

swankjesse:
I’m gonna merge this as-is because it fixes a big bug, and is already great.

Please address feedback in follow-up PRs as necessary!

okhttp/src/main/java/okhttp3/internal/connection/ExchangeFinder.kt
dave-r12: These will be `null` if the call had a previous connection we could use. Here's a test that demonstrates it:

```java
    server.enqueue(new MockResponse().setResponseCode(401));
    server.enqueue(new MockResponse().setResponseCode(200));
    AtomicReference<Connection> ref = new AtomicReference<>();
    EventListener listener = new EventListener() {
      @Override public void connectionAcquired(Call call, Connection connection) {
        ref.set(connection);
      }
    };

    client = client.newBuilder()
        .eventListener(listener)
        .authenticator((route, response) -> {
          Util.closeQuietly(ref.get().socket());
          return response.request();
        })
        .build();

    executeSynchronously("/").assertCode(200);
```

Vinod217:
Hi @swankjesse & others, We are also facing this issue with okhttp3 3.12.8 version, Could you please confirm if this issue fixed in any latest okhttp3 version?

okhttp3- 3.12.8
Retrofit- 2.6.1

Logs:-
Fatal Exception: java.lang.NoSuchMethodError
No virtual method ,??Վ�蕚õ0ޅ�V��BsK촜ϯ��+Г�?3??_S?#*烔H奼�cMݗ�Ӷɞͷtey�Nԣݛ?java/net/URI;)Ljava/util/List; in class Ljava/net/ProxySelector; or its super classes (declaration of 'java.net.ProxySelector' appears in /system/framework/core-oj.jar)
okhttp3.internal.connection.RouteSelector.resetNextProxy (RouteSelector.java:129)
okhttp3.internal.connection.RouteSelector. (RouteSelector.java:63)
okhttp3.internal.connection.StreamAllocation. (StreamAllocation.java:101)
okhttp3.internal.http.RetryAndFollowUpInterceptor.intercept (RetryAndFollowUpInterceptor.java:113)
okhttp3.internal.http.RealInterceptorChain.proceed (RealInterceptorChain.java:147)
okhttp3.internal.http.RealInterceptorChain.proceed (RealInterceptorChain.java:121)
okhttp3.RealCall.getResponseWithInterceptorChain (RealCall.java:257)

Thanks in advance.

Issue #5885, opened 2020-03-21
Normalize ipv6 addresses for HostNameVerification in OkHostNameVerifier


magwan:
In our k8 environment, 
KUBERNETES_SERVICE_HOST is set to ipv6 address in compact form. Ex: FD::0001
In certificate subject alternative name, the same ip is set in normalized form. Ex FD00:0:0:0:0:0:0:1
Due to above different formats, hostname verification is failed even though both are same address.

It will be good, if OkHostNameVerifier normalize the hostname and sans before comparing them.

Reference Implementation from https://github.com/apache/cxf/blob/master/rt/transports/http/src/main/java/org/apache/cxf/transport/https/httpclient/DefaultHostnameVerifier.java 

```
static void matchIPv6Address(final String host, final List<SubjectName> subjectAlts) throws SSLException {
        final String normalisedHost = normaliseAddress(host);
        for (int i = 0; i < subjectAlts.size(); i++) {
            final SubjectName subjectAlt = subjectAlts.get(i);
            if (subjectAlt.getType() == SubjectName.IP) {
                final String normalizedSubjectAlt = normaliseAddress(subjectAlt.getValue());
                if (normalisedHost.equals(normalizedSubjectAlt)) {
                    return;
                }
            }
        }
        throw new SSLPeerUnverifiedException("Certificate for <" + host + "> doesn't match any "
                + "of the subject alternative names: " + subjectAlts);
    }
```

yschimke:
Looks like the correct thing to do

https://tools.ietf.org/html/rfc2818#section-3.1

>    In some cases, the URI is specified as an IP address rather than a
>    hostname. In this case, the iPAddress subjectAltName must be present
>    in the certificate and must exactly match the IP in the URI.

https://tools.ietf.org/html/rfc2459#section-4.2.1.7

   When the subjectAltName extension contains a iPAddress, the address
   MUST be stored in the octet string in "network byte order," as
   specified in RFC 791 [RFC 791]. The least significant bit (LSB) of



Housley, et. al.            Standards Track                    [Page 32]

 
RFC 2459        Internet X.509 Public Key Infrastructure    January 1999

>    each octet is the LSB of the corresponding byte in the network
>    address. For IP Version 4, as specified in RFC 791, the octet string
>    MUST contain exactly four octets.  For IP Version 6, as specified in
>    RFC 1883, the octet string MUST contain exactly sixteen octets [RFC
>    1883].

yschimke:
[...]

PR #5889, opened 2020-03-22
Verify IPv6 hostnames

yschimke:
Fixes https://github.com/square/okhttp/issues/5885

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/tls/OkHostnameVerifier.kt
+++ b/okhttp/src/main/java/okhttp3/internal/tls/OkHostnameVerifier.kt
@@ -59,7 +59,7 @@ object OkHostnameVerifier : HostnameVerifier {
     }
 
     return getSubjectAltNames(certificate, ALT_IPA_NAME).any {
-      ipAddress.equals(it.toCanonicalHost(), ignoreCase = true)
+      ipAddress == it.toCanonicalHost()
     }
   }
 
--- a/okhttp/src/main/java/okhttp3/internal/tls/OkHostnameVerifier.kt
+++ b/okhttp/src/main/java/okhttp3/internal/tls/OkHostnameVerifier.kt
@@ -53,13 +53,10 @@ object OkHostnameVerifier : HostnameVerifier {
 
   /** Returns true if [certificate] matches [ipAddress]. */
   private fun verifyIpAddress(ipAddress: String, certificate: X509Certificate): Boolean {
-    if (assertionsEnabled) {
-      // HttpUrl has already canonicalized the ipAddress
-      check(ipAddress == ipAddress.toCanonicalHost()) { "Non canonical host passed in '$ipAddress'" }
-    }
+    val canonicalIpAddress = ipAddress.toCanonicalHost()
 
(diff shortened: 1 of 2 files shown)


okhttp/src/main/java/okhttp3/internal/tls/OkHostnameVerifier.kt
```diff
@@ -51,8 +52,10 @@ object OkHostnameVerifier : HostnameVerifier {
 
   /** Returns true if [certificate] matches [ipAddress]. */
   private fun verifyIpAddress(ipAddress: String, certificate: X509Certificate): Boolean {
+    val canonicalIpAddress = ipAddress.toCanonicalHost()
```
yschimke: FWIW I didn't consider IP Addresses a host path to optimise too much.  But that's certainly debatable.

okhttp/src/test/java/okhttp3/internal/tls/HostnameVerifierTest.java
```diff
@@ -516,6 +516,42 @@
...
+  @Test public void subjectAltNameWithIPAddresses() throws Exception {
+    // $ cat ./cert.cnf
+    // [req]
+    // distinguished_name=distinguished_name
+    // req_extensions=req_extensions
+    // x509_extensions=x509_extensions
+    // [distinguished_name]
+    // [req_extensions]
+    // [x509_extensions]
+    // subjectAltName=IP:0:0:0:0:0:0:0:1,IP:2a03:2880:f003:c07:face:b00c::2,IP:0::5,IP:192.168.1.1
```
swankjesse: Wanna do an IPv4 mapped address also?
https://github.com/square/okhttp/issues/4451

yschimke: What do you want the behaviour to be for ::ffff:192.168.1.1?

yschimke: Will land that as a follow up.

swankjesse: I think it’s the IPv4 address.

yschimke: What's the reasoning?  Are you self signing certificates but don't want to specify both forms of IPv4 and IPV6 (mapped IPv4) and you expect it to still work for either?

I'm nervous about Postel's law here.  Maybe keeping it explicit is the better option?

swankjesse: The canonical form of certain IPv6 addresses (with colons!) is an IPv4 address (no colons!). It’s weird, but it’s how it’s specified and I’ve come to terms with it.

If we’re canonicalizing IP addresses, we should canonicalize all of ’em and test that it works. I expect the test already passes but I’d like coverage of that case.

yschimke: This ok? https://github.com/square/okhttp/pull/5892

PR #5890, opened 2020-03-22
Make minimum outgoing web socket message compression size configurable

AntonRutkevich:
Any suggestions on naming (and general) are welcome.

### DIFF
--- a/okhttp/src/main/java/okhttp3/OkHttpClient.kt
+++ b/okhttp/src/main/java/okhttp3/OkHttpClient.kt
@@ -212,6 +212,13 @@ open class OkHttpClient internal constructor(
   /** Web socket and HTTP/2 ping interval (in milliseconds). By default pings are not sent. */
   @get:JvmName("pingIntervalMillis") val pingIntervalMillis: Int = builder.pingInterval
 
+  /**
+   * Minimum outbound web socket message size (in bytes) that will be compressed.
+   * The default is 1024 bytes.
+   */
+  @get:JvmName("webSocketMinimumDeflateSizeBytes")
+  val webSocketMinimumDeflateSizeBytes: Long = builder.webSocketMinimumDeflateSize
+
   val routeDatabase: RouteDatabase = builder.routeDatabase ?: RouteDatabase()
 
   constructor() : this(Builder())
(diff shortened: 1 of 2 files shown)


okhttp/src/main/java/okhttp3/OkHttpClient.kt
```diff
@@ -1025,6 +1034,17 @@ open class OkHttpClient internal constructor(
       pingInterval(duration.toMillis(), MILLISECONDS)
     }
 
+    /**
+     * Sets minimum outbound web socket message size (in bytes) that will be compressed.
+     *
+     * Set to 0 to enable compression for all outbound messages.
+     *
+     * [RealWebSocket.DEFAULT_MINIMUM_DEFLATE_SIZE] by default.
```
yschimke: Probably shouldn't refer to an internal class for documentation, since it will encourage use, alternatively we may want this default to be accessible.

yschimke:
Add a test in OkHttpClientTest, IIRC this also acts as our public API validation.

okhttp/src/main/java/okhttp3/OkHttpClient.kt
AntonRutkevich: Done. There are now 3 copies of the 1024 value in plain text though.

okhttp/src/main/java/okhttp3/OkHttpClient.kt
```diff
@@ -1025,6 +1034,21 @@ open class OkHttpClient internal constructor(
...
     }
 
+    /**
+     * Sets minimum outbound web socket message size (in bytes) that will be compressed.
+     *
+     * Set to 0 to enable compression for all outbound messages.
+     *
+     * 1024 by default.
+     */
+    fun webSocketMinimumDeflateSize(bytes: Long) = apply {
```
swankjesse: Rename to `webSocketMinCompressSize`

“Compress” because developers don’t know what deflate is; that algorithm is an implementation detail.
“Min” for consistency with ConnectionPool and Dispatcher which use the short form for max.

Also tempted to rearrange the words, putting 'min' first...

`minWebSocketMessageToCompress`
`minCompressMessageSize`
`minMessageToCompress`
`minMessageSizeToCompress`

Any favorites?

okhttp/src/test/java/okhttp3/OkHttpClientTest.java
```diff
@@ -377,4 +382,15 @@ public void connectFailed(URI uri, SocketAddress socketAddress, IOException e) {
         .build()
         .getRouteDatabase());
   }
+
+  @Test public void webSocketMinimumDeflateSizeNegative() {
```
swankjesse: nice test

swankjesse:
Nice PR!

okhttp/src/main/java/okhttp3/OkHttpClient.kt
AntonRutkevich: Agree.
I'd go with `minWebSocketMessageToCompress`.

PR #5891, opened 2020-03-22
Support 3.12.x through December 31, 2021

swankjesse:
Our original expectations were that KitKat devices would age out in
2020. It's been a garbage year and this expectations no longer hold.
We're extending support an extra year.

We continue to support OkHttp 3.12.x because later releases are not
usable on older devices and JVMs.

OkHttp 3.14.x has no such reason to receive support, and we'll stop
updating it on June 30, 2020. Users of OkHttp 3.14.x should upgrade
to the 4.x series. Android apps should accept Kotlin libraries; it's
the platform's future. And JVM apps are less constrained by binary
size. Here's the size of OkHttp + transitive dependencies, and those
of alternatives:

 * OkHttp 4.4: 2.47 MiB including Kotlin stdlib (1.51 MiB)
 * Apache HTTP Client 5.0: 1.78 MiB including slf4j-api (0.04 MiB)
 * Jetty HTTP Client 9.4: 1.19 MiB

### DIFF
--- a/docs/security.md
+++ b/docs/security.md
@@ -3,11 +3,11 @@ Security Policy
 
 ## Supported Versions
 
-| Version | Supported  | Notes          |
-| ------- | ---------- | -------------- |
-| 4.x     | ✅         |                |
-| 3.14.x  | ✅         |                |
-| 3.12.x  | ✅         | Android 2.3+ (API level 9+) and Java 7+. Platforms may not support TLSv1.2. Until December 31, 2020  |
+| Version | Supported        | Notes          |
+| ------- | ---------------- | -------------- |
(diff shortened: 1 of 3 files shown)


docs/security.md
```diff
@@ -7,7 +7,7 @@ Security Policy
 | ------- | ---------- | -------------- |
 | 4.x     | ✅         |                |
 | 3.14.x  | ✅         |                |
```
yschimke: Let's write a date here rather than just commit logs.

swankjesse: Screen Shot 2020-03-24 at 1 24 54 PM

PR #5901, opened 2020-03-25
Force cancel all events on client

yschimke:
Once a test leaves a client running, all other tests will fail.  But leaving things running can impact other tests, so make another attempt to cleanup.

### DIFF
--- a/android-test/src/androidTest/java/okhttp/android/test/OkHttpTest.kt
+++ b/android-test/src/androidTest/java/okhttp/android/test/OkHttpTest.kt
@@ -167,6 +167,7 @@ class OkHttpTest {
       }
     } finally {
       Security.removeProvider("Conscrypt")
+      client.close()
     }
   }
 
@@ -202,6 +203,7 @@ class OkHttpTest {
       }
     } finally {
       Security.removeProvider("GmsCore_OpenSSL")
+      client.close()
     }
   }
 
@@ -549,6 +551,11 @@ class OkHttpTest {
     } catch (uhe: UnknownHostException) {
       assumeNoException(uhe)
     }
+  }
+
+  fun OkHttpClient.close() {
+    dispatcher.executorService.shutdown()
+    connectionPool.evictAll()
   }
 
   companion object {
--- a/okhttp-testing-support/src/main/java/okhttp3/OkHttpClientTestRule.kt
+++ b/okhttp-testing-support/src/main/java/okhttp3/OkHttpClientTestRule.kt
@@ -20,7 +20,8 @@ import java.util.concurrent.TimeUnit
 import java.util.logging.Logger
 import okhttp3.internal.concurrent.TaskRunner
 import okhttp3.testing.Flaky
-import org.assertj.core.api.Assertions.assertThat
+import org.junit.Assert.assertEquals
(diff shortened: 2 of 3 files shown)


yschimke:
@swankjesse It's internal class, but you might not want to expose.

okhttp/src/main/java/okhttp3/internal/concurrent/TaskRunner.kt
```diff
@@ -242,9 +242,9 @@ class TaskRunner(
     }
   }
 
-  private fun cancelAll() {
+  fun cancelAll() {
     for (i in busyQueues.size - 1 downTo 0) {
-      readyQueues[i].cancelAllAndDecide()
+      busyQueues[i].cancelAllAndDecide()
```
yschimke: Found bug

swankjesse: ooooh

okhttp-testing-support/src/main/java/okhttp3/OkHttpClientTestRule.kt
```diff
@@ -88,15 +89,18 @@ class OkHttpClientTestRule : TestRule {
...
-      assertThat(queue.idleLatch().await(1_000L, TimeUnit.MILLISECONDS))
-          .withFailMessage("Queue still active after 1000 ms")
-          .isTrue()
+    try {
+      for (queue in TaskRunner.INSTANCE.activeQueues()) {
+        assertTrue("Queue still active after 1000 ms",
+            queue.idleLatch().await(1_000L, TimeUnit.MILLISECONDS))
+      }
+    } finally {
+      TaskRunner.INSTANCE.cancelAll()
```
swankjesse: should this be an if/else thing?

```
for (queue in TaskRunner.INSTANCE.activeQueues()) {
  if (!queue.idleLatch().await(1_000L, TimeUnit.MILLISECONDS)) {
    TaskRunner.INSTANCE.cancelAll() 
    fail("Queue still active after 1000 ms")
  }
}
```
that way we don’t cancelAll() when it’s succeeding?

okhttp-testing-support/src/main/java/okhttp3/OkHttpClientTestRule.kt
```diff
@@ -88,15 +89,18 @@ class OkHttpClientTestRule : TestRule {
     testClient?.let {
       val connectionPool = it.connectionPool
       connectionPool.evictAll()
-      assertThat(connectionPool.connectionCount()).isEqualTo(0)
+      assertEquals(0, connectionPool.connectionCount())
```
yschimke: This is due to assertj relying on Duration and failing on Android.

android-test/src/androidTest/java/okhttp/android/test/OkHttpTest.kt
```diff
@@ -167,6 +167,7 @@ class OkHttpTest {
       }
     } finally {
       Security.removeProvider("Conscrypt")
+      client.close()
```
yschimke: FWIW these tests triggered the overall bug.  They have a client build unchained from OkHttpClientTestRule.

PR #5905, opened 2020-03-28
Close Android test properly in case it passes

yschimke:


yschimke:
Passes on Android 21 which means it fails

PR #5909, opened 2020-03-28
Defer class initialization to avoid spamming on Conscrypt errors (not for landing)

yschimke:
Considering as a potential fix for https://github.com/square/okhttp/issues/5760

AndroidPlatform uses ConscryptSocketAdaptor which has

```
    fun buildIfSupported(): SocketAdapter? =
        if (ConscryptPlatform.isSupported) ConscryptSocketAdapter() else null
```

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/platform/ConscryptPlatform.kt
+++ b/okhttp/src/main/java/okhttp3/internal/platform/ConscryptPlatform.kt
@@ -102,7 +102,7 @@ class ConscryptPlatform private constructor() : Platform() {
   companion object {
     val isSupported: Boolean = try {
       // Trigger an early exception over a fatal error, prefer a RuntimeException over Error.
-      Class.forName("org.conscrypt.Conscrypt\$Version")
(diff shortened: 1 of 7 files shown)


yschimke:
The main risk with the change would appear to be that we can load the target class, but not dependencies and thus get to the wrong point.

okhttp/src/main/java/okhttp3/internal/platform/ConscryptPlatform.kt
```diff
@@ -102,7 +102,7 @@ class ConscryptPlatform private constructor() : Platform() {
   companion object {
     val isSupported: Boolean = try {
       // Trigger an early exception over a fatal error, prefer a RuntimeException over Error.
-      Class.forName("org.conscrypt.Conscrypt\$Version")
+      Class.forName("org.conscrypt.Conscrypt\$Version", false, javaClass.classLoader)
```
yschimke: This is the main one to consider changing to avoid the spam.

Issue #5594, opened 2019-11-12
IllegalArgumentException: reason.size() > 123


flocsy:
There's this relatively new code in: okhttp/src/main/java/okhttp3/internal/ws/RealWebSocket.kt:

```
    if (reason != null) {
      reasonBytes = reason.encodeUtf8()
      require(reasonBytes.size <= CLOSE_MESSAGE_MAX) {
        "reason.size() > $CLOSE_MESSAGE_MAX: $reason"
      }
    }
```

The problem is that this causes "IllegalArgumentException: reason.size() > 123" and there's no "nice" way to fix it.

Of course I can do:

```
   int code = 1000;
   String reason = ("disconnected: " + url).substring(0,123);
   webSocket.close(code, reason);
```

but this is not ideal, because that 123 can change. So I would rather do:

```
   String reason = ("disconnected: " + url).substring(0,(int) WebSocketProtocol.CLOSE_MESSAGE_MAX);
```

but this doesn't work because it's internal.

So at least change it to public, but even better solution maybe would be to do this (the substring) in okhttp, and instead of an exception just log a warning.

swankjesse:
123 won't change.

pau101:
It should be acknowledged that it would be erroneous to truncate arbitrary reason strings by 123, as that max length refers to bytes of UTF-8 text, not characters, nor String chars (UTF-16 code units). The error message OkHttp is providing is a bit misleading in that regard.

yschimke:
Yep, agreed.  We shouldn't arbitrarily truncate, it's awkward with partial unicode strings etc.  So clients should do this, reporting the error is the best we can do as an API.  If you feel strongly, submit a PR to improve the message to include encoded bytes or similar. But unlikely we fix this internally if it hasn't gotten attention so far.

Won't fix without a PR.

PR #5913, opened 2020-03-29
Document maximum close reason length

pau101:
This PR adds documentation to the exception which is thrown when the close reason is longer than 123 UTF-8 encoded bytes, addressing concerns from #5594. There are [preexisting tests](https://github.com/square/okhttp/blob/4c1f65d1388ae8d5720530ca9b0094f7343eb1c8/okhttp/src/test/java/okhttp3/internal/ws/WebSocketHttpTest.java#L476) for this behavior yet it is currently undocumented.

### DIFF
--- a/okhttp/src/main/java/okhttp3/WebSocket.kt
+++ b/okhttp/src/main/java/okhttp3/WebSocket.kt
@@ -98,8 +98,8 @@ interface WebSocket {
    *
    * @param code Status code as defined by
    *     [Section 7.4 of RFC 6455](http://tools.ietf.org/html/rfc6455#section-7.4).
-   * @param reason Reason for shutting down or null.
-   * @throws IllegalArgumentException if code is invalid.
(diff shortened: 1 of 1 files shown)


okhttp/src/main/java/okhttp3/WebSocket.kt
```diff
@@ -98,8 +98,8 @@ interface WebSocket {
    *
    * @param code Status code as defined by
    *     [Section 7.4 of RFC 6455](http://tools.ietf.org/html/rfc6455#section-7.4).
-   * @param reason Reason for shutting down or null.
-   * @throws IllegalArgumentException if code is invalid.
+   * @param reason Reason for shutting down, no longer than 123 bytes of UTF-8 encoded data (**not** characters) or null.
```
yschimke: This is quite explicit, but given our draconian failures for a String API, probably warranted.

PR #5914, opened 2020-03-30
Revert "Document maximum close reason length"

abc52090241:
Reverts square/okhttp#5913


swankjesse:
Why?

yschimke:
Reopen with discussion, I can't see any reason apart from this being accidental.

Issue #5819, opened 2020-02-21
Infinite loop in connection retry


danielimre:
For some reason (I haven't been able to reproduce this locally yet), under heavy load some of our okhttp clients might get into an infinite loop (eating up one or more CPU cores) when connection failure retry is enabled. It seems this affects some connections only in the connection pool. 

If I disable connection failure retry then a `SocketException` is thrown at https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/internal/connection/RealConnection.kt#L594 with the following stack trace:

```
java.net.SocketException: Socket is closed at java.net.Socket.setSoTimeout(Unknown Source) 
at sun.security.ssl.BaseSSLSocketImpl.setSoTimeout(Unknown Source) 
at sun.security.ssl.SSLSocketImpl.setSoTimeout(Unknown Source) 
at okhttp3.internal.connection.RealConnection.newCodec$okhttp(RealConnection.kt:594) 
at okhttp3.internal.connection.ExchangeFinder.find(ExchangeFinder.kt:83) 
at okhttp3.internal.connection.RealCall.initExchange$okhttp(RealCall.kt:245) 
at okhttp3.internal.connection.ConnectInterceptor.intercept(ConnectInterceptor.kt:32) 
at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:100) 
at okhttp3.internal.cache.CacheInterceptor.intercept(CacheInterceptor.kt:82) 
at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:100) 
at okhttp3.internal.http.BridgeInterceptor.intercept(BridgeInterceptor.kt:83) 
at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:100) 
at okhttp3.internal.http.RetryAndFollowUpInterceptor.intercept(RetryAndFollowUpInterceptor.kt:74) 
at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:100) 
at okhttp3.internal.connection.RealCall.getResponseWithInterceptorChain$okhttp(RealCall.kt:197)
at okhttp3.internal.connection.RealCall$AsyncCall.run(RealCall.kt:502)
```
In this case a certain percentage of the requests fail constantly suggesting a stuck pooled faulty connection which is never evicted.

The only option is to restart the affected application.

[...]

PR #5916, opened 2020-03-31
Add test demonstrating rogue connection seeder.

dave-r12:
https://github.com/square/okhttp/issues/5819


dave-r12:
Possibly the cause of #5819. Is this a case we want to handle? Seems sorta off the beaten path and lots of things would need fall in place for it to happen in practice.

yschimke:
@dave-r12 impressive repro!  Seems high value to me, what would a fix potentially look like?  Be more responsive to the disconnect? Or to handle and automatically retry if idempotent with HTTP/1.1?

yschimke:
One good aspect for this is that callTimeout does stop this infinite hang.

PR #5920, opened 2020-04-04
Flag connection as unhealthy when underlying socket is closed.

dave-r12:
https://github.com/square/okhttp/issues/5819

### DIFF
--- a/okhttp/src/main/java/okhttp3/internal/connection/RealConnection.kt
+++ b/okhttp/src/main/java/okhttp3/internal/connection/RealConnection.kt
@@ -631,9 +631,11 @@ class RealConnection(
   fun isHealthy(doExtensiveChecks: Boolean): Boolean {
     val nowNs = System.nanoTime()
 
+    val rawSocket = this.rawSocket!!
     val socket = this.socket!!
     val source = this.source!!
-    if (socket.isClosed || socket.isInputShutdown || socket.isOutputShutdown) {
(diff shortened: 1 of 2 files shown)


dave-r12:
I believe this reproduces the issue reported in #5819. This is very subtle. I found that if we close the `rawSocket` in `RealConnection` (using cancel) it is not seen by the application layer `socket` if we are using SSL/TLS. So the health checks pass. When we later call `newCodec` it throws a `SocketException`. Since the connection is considered healthy this continues indefinitely.

okhttp/src/main/java/okhttp3/internal/connection/RealConnection.kt
```diff
@@ -631,9 +631,11 @@ class RealConnection(
   fun isHealthy(doExtensiveChecks: Boolean): Boolean {
     val nowNs = System.nanoTime()
 
+    val rawSocket = this.rawSocket!!
     val socket = this.socket!!
     val source = this.source!!
-    if (socket.isClosed || socket.isInputShutdown || socket.isOutputShutdown) {
+    if (rawSocket.isClosed || socket.isClosed || socket.isInputShutdown ||
```
swankjesse, line 637: nice

yschimke:
Impressive detective work.

yschimke:
I wonder whether this could solve a whole bunch of problems after network changes etc.  I'm glad this made 4.5!

PR #5921, opened 2020-04-05
Add additional headers and format response body to improve readability

StolarczykJakub:
This pull request added to the Interceptor a few changes:
- Formatting Json response body (without any extra dependency)
- Adding additional log headers 
- Selection of json indent spaces

All changes are disabled by default, we are still able to see standard type of logs.
I just wanted to improve UX for developers

### DIFF
--- a/okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.kt
+++ b/okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.kt
@@ -37,7 +37,10 @@ import okio.GzipSource
  * change slightly between releases. If you need a stable logging format, use your own interceptor.
  */
 class HttpLoggingInterceptor @JvmOverloads constructor(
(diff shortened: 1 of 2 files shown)


swankjesse:
We aren't taking new features in the logging interceptor. Its deliberately basic and we encourage you to use a fancier tool if you use it a lot. Some examples are Chucker for Android and Charles Proxy.

Please don't spend time wiritng PRs for OkHttp without getting a green light from project maintainers first. We would have been able to save you some time.

swankjesse:
Also! This code looks reasonable enough. Please consider releasing it as a standalone project!

yschimke:
+1 it would be great to see further innovation here and jitpack.io should make hosting your own library pretty low maintenance

StolarczykJakub:
Thanks for feedback! My intention was to add it to the base so no extra libs needed. As android developer there is no application with networking without okhttp so it would be nice to be able to format those logs, but i understand and respect your decision. Take care :)

PR #5925, opened 2020-04-08
Kotlin 1.3.71 upgrade

yschimke:


yschimke:
Nothing meaningful https://github.com/JetBrains/kotlin/releases/tag/v1.3.71

build.gradle
```diff
@@ -204,6 +204,7 @@ subprojects { project ->
   }
 
   dependencies {
+    //noinspection GradleDynamicVersion
```
yschimke, line 207: + flagged as dynamic version

swankjesse: Which `+` ?

yschimke: com.google.errorprone:javac:9+181-r4173-1

yschimke: I can revert and raise a bug against the gradle project I guess?

Issue #4049, opened 2018-06-07
Expose public APIs for CertificatePinner pins


pooyanmajd:
Hey guys,

CertificatePinner.Builder class has add method, Is it possible to add a get method as well? I need to know if there are any pins already added in the class

yschimke:
@pooyanmajd care to submit a PR?

swankjesse:
This is tough ’cause we don’t currently expose a `Pin` type, and would need to do that, or something similar, to make this API make sense.

pooyanmajd:
Yeah I wanted to create a PR but I figured out we need to expose `Pin` object as well or we can just return a Pair object since Pin is <String, String>

pooyanmajd:
So we should make `Pin` class a public separate class not a nested static one, and return the list of Pins from builder

PR #5927, opened 2020-04-09
Public API for Pins

yschimke:
Addressing https://github.com/square/okhttp/issues/4049


okhttp/src/main/java/okhttp3/CertificatePinner.kt
```diff
@@ -212,16 +213,7 @@ class CertificatePinner internal constructor(
...
-    var result: List<Pin> = emptyList()
-    for (pin in pins) {
-      if (pin.matches(hostname)) {
-        if (result.isEmpty()) result = mutableListOf()
-        (result as MutableList<Pin>).add(pin)
-      }
-    }
-    return result
-  }
+  fun findMatchingPins(hostname: String): List<Pin> = pins.filterList { it.matches(hostname) }
```
swankjesse, line 216: OMG Kotlin nice

okhttp/src/main/java/okhttp3/CertificatePinner.kt
```diff
@@ -247,9 +239,9 @@ class CertificatePinner internal constructor(
     return result
   }
 
-  internal data class Pin(
+  data class Pin internal constructor(
```
swankjesse, line 242: internal constructors on a data class are almost useless thanks to .copy().

swankjesse: ... wanna remove the data class bit here and make this a plain class?

okhttp/src/main/java/okhttp3/CertificatePinner.kt
```diff
@@ -276,6 +268,31 @@ class CertificatePinner internal constructor(
     }
 
     override fun toString(): String = hashAlgorithm + hash.base64()
+
+    companion object {
+      @JvmStatic
+      fun newPin(pattern: String, pin: String): Pin {
```
swankjesse, line 274: make this into the class constructor? If it’s not a data class it’s easy to write that way

okhttp/src/main/java/okhttp3/internal/Util.kt
```diff
@@ -354,7 +390,10 @@ fun Source.skipAll(duration: Int, timeUnit: TimeUnit): Boolean {
  * source is helpful, such as when doing so completes a cache body or frees a socket connection for
  * reuse.
  */
-fun Source.discard(timeout: Int, timeUnit: TimeUnit): Boolean = try {
+fun Source.discard(
```
swankjesse, line 393: what’s with these changes?

okhttp/src/main/java/okhttp3/internal/Util.kt
```diff
@@ -589,3 +636,14 @@ fun Exception.withSuppressed(suppressed: List<Exception>): Throwable = apply {
 
   for (e in suppressed) addSuppressed(e)
 }
+
+inline fun <T> Iterable<T>.filterList(predicate: (T) -> Boolean): List<T> {
```
swankjesse, line 640: dig

okhttp/src/main/java/okhttp3/internal/Util.kt
yschimke: I'll revert.  I think came from the Square code formatter in Intellij?

okhttp/src/main/java/okhttp3/CertificatePinner.kt
```diff
@@ -275,7 +295,7 @@ class CertificatePinner internal constructor(
       }
     }
 
-    override fun toString(): String = hashAlgorithm + hash.base64()
+    override fun toString(): String = hashAlgorithm + "/" + hash.base64()
```
yschimke, line 298: This seems like the better external api, with the slash removed

yschimke:
Any thoughts on API for using a single Pin?  The loop in CertificatePinner is good to cache the tested certificate hash, but as an API we probably want some boolean test method?

okhttp/src/main/java/okhttp3/CertificatePinner.kt
```diff
@@ -247,14 +239,42 @@ class CertificatePinner internal constructor(
     return result
   }
 
-  internal data class Pin(
+  class Pin internal constructor(
```
swankjesse, line 242: Public constructor?

okhttp/src/main/java/okhttp3/CertificatePinner.kt
swankjesse: Definitely.

swankjesse: Template string?

okhttp/src/main/java/okhttp3/CertificatePinner.kt
```diff
@@ -317,29 +357,7 @@ class CertificatePinner internal constructor(
     internal fun X509Certificate.toSha1ByteString(): ByteString =
         publicKey.encoded.toByteString().sha1()
 
-    internal fun X509Certificate.toSha256ByteString(): ByteString =
+    fun X509Certificate.toSha256ByteString(): ByteString =
```
swankjesse, line 360: Rename to sha256Hash and make the other one public as sha1Hash? That's as close as we can get to the Pin API.

okhttp/src/main/java/okhttp3/CertificatePinner.kt
```diff
@@ -165,11 +166,11 @@ class CertificatePinner internal constructor(
 
       for (pin in pins) {
         when (pin.hashAlgorithm) {
-          "sha256/" -> {
+          "sha256" -> {
```
swankjesse, line 169: I think the body of this for loop to be a method, `Pin.matchesCertificate(certificates: List<Certificate>)`.

okhttp/src/main/java/okhttp3/CertificatePinner.kt
```diff
@@ -247,14 +239,42 @@ class CertificatePinner internal constructor(
...
+        }
+        pin.startsWith("sha256/") -> {
+          this.hashAlgorithm = "sha256"
+          this.hash = pin.substring("sha256/".length).decodeBase64() ?: throw IllegalArgumentException("Invalid pin hash: $pin")
+        }
+        else -> throw IllegalArgumentException("pins must start with 'sha256/' or 'sha1/': $pin")
+      }
+    }
+
     fun matches(hostname: String): Boolean {
```
swankjesse, line 278: Rename to matchesHostname so we have matchesHostname and matchesCertificates?

okhttp/src/main/java/okhttp3/CertificatePinner.kt
```diff
@@ -275,12 +293,38 @@ class CertificatePinner internal constructor(
...
+      result = 31 * result + hashAlgorithm.hashCode()
+      result = 31 * result + hash.hashCode()
+      return result
+    }
   }
 
   /** Builds a configured certificate pinner. */
   class Builder {
-    private val pins = mutableListOf<Pin>()
+    val pins = mutableListOf<Pin>()
```
yschimke, line 327: Added to meet original request

okhttp/src/main/java/okhttp3/CertificatePinner.kt
```diff
@@ -133,8 +134,8 @@ import okio.ByteString.Companion.toByteString
  */
 @Suppress("NAME_SHADOWING")
 class CertificatePinner internal constructor(
-  private val pins: Set<Pin>,
-  internal val certificateChainCleaner: CertificateChainCleaner?
+  val pins: Set<Pin>,
```
yschimke, line 137: Added to meet original request

PR #5931, opened 2020-04-10
Emulate Windows file system restrictions in a wrapper

swankjesse:
This should make it possible to get DiskLruCacheTest passing on
Windows even when Windows isn't available.

### DIFF
--- a/okhttp-testing-support/src/main/java/okhttp3/TestUtil.kt
+++ b/okhttp-testing-support/src/main/java/okhttp3/TestUtil.kt
@@ -20,7 +20,8 @@ import java.net.InetSocketAddress
 import java.net.UnknownHostException
 import java.util.Arrays
 import okhttp3.internal.http2.Header
-import org.junit.Assume
+import org.junit.Assume.assumeFalse
+import org.junit.Assume.assumeNoException
 
 object TestUtil {
   @JvmField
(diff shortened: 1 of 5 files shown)


okhttp/src/test/java/okhttp3/internal/io/FileSystemTest.kt
```diff
@@ -0,0 +1,143 @@
...
+        arrayOf(WindowsFileSystem(FileSystem.SYSTEM), true),
+        arrayOf(WindowsFileSystem(InMemoryFileSystem()), true)
+    )
+  }
+
+  @Parameter(0) @JvmField
+  var fileSystem: FileSystem = FileSystem.SYSTEM
+
+  @Parameter(1) @JvmField
+  var windows: Boolean = false
```
JakeWharton, line 53: You can constructor-inject these

```kotlin
class FileSystemTest(val fs: FileSystem, val windows: Boolean)
```

No defaults or lateinit needed.

okhttp-testing-support/src/main/java/okhttp3/TestUtil.kt
```diff
@@ -59,20 +60,16 @@ object TestUtil {
     try {
       InetAddress.getByName("www.google.com")
     } catch (uhe: UnknownHostException) {
-      Assume.assumeNoException(uhe)
+      assumeNoException(uhe)
     }
   }
 
   @JvmStatic
   fun assumeNotWindows() {
```
yschimke, line 68: TIOLI: This might be more natural on PlatformRule?

okhttp/src/test/java/okhttp3/internal/io/FileSystemTest.kt
swankjesse: Ooooh how did I not know about that syntax? Much nicer. Fixed.

okhttp-testing-support/src/main/java/okhttp3/TestUtil.kt
swankjesse: I think we probably want go from _Platform_ to the somewhat-independent parts it’s composed of, like TlsPlatform and VmPlatform. There’s an implied OsPlatform as a third thing that we’ve mostly avoided until now!

yschimke: +1 

The SocketAdaptor is also interesting because it allows for different socket types at the same time, and also cross cuts Android and JDK.

Issue #5698, opened 2020-01-05
Fix bouncy castle platform bugs


yschimke:
After https://github.com/square/okhttp/pull/5695 lands, there are plenty of failing tests that either need test or code fixes.  Previous testing has shown big differences between JDK 8 and JDK 11 for Bouncy Castle.

```
$ ./gradlew -Dokhttp.platform=bouncycastle test
```

yschimke:
2 fixes here https://github.com/square/okhttp/pull/5741

yschimke:
A test fix https://github.com/square/okhttp/pull/5747

yschimke:
IIUC current Failing test relates to BouncyCastle provider using a default non BC KeyManager & TrustManager when used with client auth

```
okhttp3.tls.HandshakeCertificatesTest > clientAndServer FAILED
    java.util.concurrent.ExecutionException: org.bouncycastle.tls.TlsFatalAlert: handshake_failure(40)
        at java.base/java.util.concurrent.FutureTask.report(FutureTask.java:122)
        at java.base/java.util.concurrent.FutureTask.get(FutureTask.java:191)
        at okhttp3.tls.HandshakeCertificatesTest.clientAndServer(HandshakeCertificatesTest.java:105)

        Caused by:
        org.bouncycastle.tls.TlsFatalAlert: handshake_failure(40)
            at org.bouncycastle.tls.AbstractTlsServer.getSelectedCipherSuite(AbstractTlsServer.java:387)
            at org.bouncycastle.jsse.provider.ProvTlsServer.getSelectedCipherSuite(ProvTlsServer.java:291)
            at org.bouncycastle.tls.TlsServerProtocol.sendServerHelloMessage(TlsServerProtocol.java:747)
            at org.bouncycastle.tls.TlsServerProtocol.handleHandshakeMessage(TlsServerProtocol.java:162)
            at org.bouncycastle.tls.TlsProtocol.processHandshakeQueue(TlsProtocol.java:546)
            at org.bouncycastle.tls.TlsProtocol.processRecord(TlsProtocol.java:464)
            at org.bouncycastle.tls.RecordStream.readRecord(RecordStream.java:224)
            at org.bouncycastle.tls.TlsProtocol.safeReadRecord(TlsProtocol.java:687)
            at org.bouncycastle.tls.TlsProtocol.blockForHandshake(TlsProtocol.java:325)
            at org.bouncycastle.tls.TlsServerProtocol.accept(TlsServerProtocol.java:83)
[...]

PR #5936, opened 2020-04-11
Enable BouncyCastle on Android and add some testing

yschimke:
Enables BouncyCastle H2 support on Android and gets tests mostly passing on CI.  MockWebServer and BouncyCastle are not currently working.


okhttp/src/main/java/okhttp3/internal/platform/Platform.kt
```diff
@@ -198,7 +198,8 @@ open class Platform {
 
     private val isBouncyCastlePreferred: Boolean
       get() {
-        return System.getProperty("okhttp.platform") == "bouncycastle"
+        val preferredProvider = Security.getProviders()[0].name
```
yschimke, line 201: Need registration of providers, not just system property.

cc @swankjesse this was deliberate originally and this is a behaviour change.

yschimke:
Builds ok with most of MockWebServer tests commented out https://app.circleci.com/pipelines/github/square/okhttp/2899/workflows/31a40bfd-edce-4ea4-a2c6-9d9724e60f66/jobs/11836

Not turning this on, but want a baseline to start from that isn't a sea of red.

yschimke:
https://github.com/square/okhttp/issues/5698

okhttp-testing-support/build.gradle
```diff
@@ -2,14 +2,9 @@ dependencies {
...
   api deps.assertj
-  api deps.bouncycastledebug
-  implementation deps.bouncycastlepkix, {
-    // prefer debug build for our own testing
-    exclude module:"bcprov-jdk15on"
-  }
-  implementation deps.bouncycastletls, {
-    exclude module:"bcprov-jdk15on"
-  }
+  api deps.bouncycastle
```
yschimke, line 5: I haven't used the debug to date, so giving up on it.  More viable to test in a separate project.

okhttp/src/main/java/okhttp3/OkHttpClient.kt
```diff
@@ -1061,10 +1062,13 @@ open class OkHttpClient internal constructor(
     private fun newSslSocketFactory(trustManager: X509TrustManager): SSLSocketFactory {
       try {
         val sslContext = Platform.get().newSSLContext()
-        sslContext.init(null, arrayOf<TrustManager>(trustManager), null)
+        // no support for bouncycastle SecureRandom on Android
+        // see JcaTlsCryptoProvider.create
+        val random = SecureRandom()
```
yschimke, line 1067: BouncyCastle provider doesn't have a working SecureRandom on Android, so this is required.

https://stackoverflow.com/questions/10259780/does-bouncycastle-have-a-securerandom-service

okhttp/src/test/java/okhttp3/OkHttpClientTest.java
```diff
@@ -50,6 +51,8 @@
 import static org.junit.Assert.fail;
 
 public final class OkHttpClientTest {
+  @Rule public PlatformRule platform = new PlatformRule();
```
yschimke, line 54: This could be removed but is generally useful.  I was adding because of the assumption in Platform that the system property indicated bouncycastle was in use, vs registration as preferred provider.

okhttp-testing-support/src/main/java/okhttp3/testing/PlatformRule.kt
```diff
@@ -313,6 +313,9 @@ open class PlatformRule @JvmOverloads constructor(
   }
 
   fun assumeNotBouncyCastle() {
+    // Most failures are with MockWebServer
+    // org.bouncycastle.tls.TlsFatalAlertReceived: handshake_failure(40)
+    //        at org.bouncycastle.tls.TlsProtocol.handleAlertMessage(TlsProtocol.java:241)
```
swankjesse, line 318: oooh handy comment

okhttp/src/main/java/okhttp3/internal/platform/Platform.kt
swankjesse: I like this behavior better

okhttp/src/main/java/okhttp3/internal/platform/android/BouncyCastleSocketAdapter.kt
```diff
@@ -0,0 +1,68 @@
...
+    }
+  }
+
+  override fun configureTlsExtensions(
+    sslSocket: SSLSocket,
+    hostname: String?,
+    protocols: List<Protocol>
+  ) {
+    // No TLS extensions if the socket class is custom.
+    if (matchesSocket(sslSocket)) {
```
swankjesse, line 51: contracts will be nice here. We might get the cast for free?

okhttp/src/main/java/okhttp3/internal/platform/android/BouncyCastleSocketAdapter.kt
```diff
@@ -0,0 +1,68 @@
...
+      val s = sslSocket as BCSSLSocket
+
+      // Enable ALPN.
```
swankjesse, line 54: copy pasta

okhttp/src/main/java/okhttp3/internal/platform/android/BouncyCastleSocketAdapter.kt
```diff
@@ -0,0 +1,68 @@
...
+      val sslParameters = sslSocket.parameters
+
+      // Enable ALPN.
+      sslParameters.applicationProtocols = Platform.alpnProtocolNames(protocols).toTypedArray()
+
+      sslSocket.parameters = sslParameters
```
swankjesse, line 60: That this line is necessary is probably my least favorite part aspect of Kotlin properties. Looks like a no-op!

okhttp/src/test/java/okhttp3/OkHttpClientTest.java
swankjesse: Should OkHttpClientTestRule do this for all tests? Having one rule that just sets everything up is easy

okhttp/src/main/java/okhttp3/internal/platform/android/BouncyCastleSocketAdapter.kt
yschimke: Yep, but still experimental :(

okhttp/src/test/java/okhttp3/OkHttpClientTest.java
yschimke: +1 I think you've suggested before, I just haven't done it myself.

okhttp/src/main/java/okhttp3/OkHttpClient.kt
yschimke: @swankjesse Any risk with this change?  

https://github.com/bcgit/bc-java/blob/master/tls/src/main/java/org/bouncycastle/tls/crypto/impl/jcajce/JcaTlsCryptoProvider.java#L66

Looks like a reason to do this, but also a reason to only do this if you truly understand what it's doing.  We don't have that context as we don't control the JVM security config, our users do.

Issue #5902, opened 2020-03-26
:android-test:connectedCheck fails to run with duplicate class files


yschimke:
```
Caused by: com.android.tools.r8.utils.AbortException: Error: Program type already present: okhttp3.CallEvent$CallFailed
        at com.android.tools.r8.utils.Reporter.failIfPendingErrors(Reporter.java:101)
        at com.android.tools.r8.utils.Reporter.fatalError(Reporter.java:72)
        at com.android.tools.r8.utils.ExceptionUtils.withCompilationHandler(ExceptionUtils.java:66)
        ... 11 more
```

Seemingly because Intellij and gradle output to different paths

```
find . -name \*CallFailed\*
./okhttp-testing-support/build/classes/java/main/okhttp3/CallEvent$CallFailed.class
./okhttp-testing-support/build/classes/kotlin/main/okhttp3/CallEvent$CallFailed.class
```

yschimke:
Example project, command to build

$ ANDROID_SDK_ROOT=XXXX/sdk ./gradlew :android-test:connectedCheck

yschimke:
The main intersting this about this model is that it has both java and kotlin under the same directory tree

```
$ find okhttp-testing-support/src/main/java
okhttp-testing-support/src/main/java
okhttp-testing-support/src/main/java/okhttp3
okhttp-testing-support/src/main/java/okhttp3/FakeProxySelector.java
okhttp-testing-support/src/main/java/okhttp3/ForwardingRequestBody.java
okhttp-testing-support/src/main/java/okhttp3/FakeSSLSession.java
okhttp-testing-support/src/main/java/okhttp3/OkHttpClientTestRule.kt
okhttp-testing-support/src/main/java/okhttp3/ClientRuleEventListener.kt
okhttp-testing-support/src/main/java/okhttp3/internal
okhttp-testing-support/src/main/java/okhttp3/internal/io
okhttp-testing-support/src/main/java/okhttp3/internal/io/InMemoryFileSystem.java
okhttp-testing-support/src/main/java/okhttp3/internal/duplex
okhttp-testing-support/src/main/java/okhttp3/internal/duplex/AsyncRequestBody.java
okhttp-testing-support/src/main/java/okhttp3/UppercaseRequestInterceptor.java
okhttp-testing-support/src/main/java/okhttp3/ForwardingResponseBody.java
okhttp-testing-support/src/main/java/okhttp3/RecordingEventListener.kt
[...]

PR #5938, opened 2020-04-11
Rename to src/main/kotlin

yschimke:


yschimke:
https://github.com/square/okhttp/issues/5902

PR #5939, opened 2020-04-11
Revert to null SecureRandom

yschimke:
Safer for most providers, but causes failures for bc-java.


yschimke:
Tracking longer term fix here https://github.com/bcgit/bc-java/issues/686

PR #5941, opened 2020-04-12
Fix DiskLruCache to work on Windows

swankjesse:
As originally designed DiskLruCache assumes an inode-like
file system, where it's fine to delete files that are
currently being read or written.

On Windows the file system forbids this, so we must be
more careful when deleting and renaming files. These
operations come up a lot internally in the cache:
 - deleting to evict an entry
 - renaming to commit a dirty file to a clean file

The workaround is simple if unsatisfying: we don't
permit concurrent reads and writes on Windows. We
can have multiple concurrent reders, or a single
writer.

One challenge in this implementation is detecting
whether we're running on Windows or a good operating
system. We deliberately don't look at System properties
here because the OS and file system may disagree, such
as when a Windows machine has an ext4 partition, or when
a Linux machine has an NTFS partition. Instead of detecting
we just attempt an edit and see what happens.

Another challenge in this implementation is what to
do when a file needs to be deleted but cannot be because
it is currently open. In such cases we now mark the
cache entry as a 'zombie'. When the files are later
closed they now check for zombie status and delete the
files if necessary. Note that it is not possible to
store a new cache entry while it is a zombie.

Closes: https://github.com/square/okhttp/issues/5761

### DIFF
--- a/okhttp-testing-support/src/main/kotlin/okhttp3/internal/io/InMemoryFileSystem.kt
+++ b/okhttp-testing-support/src/main/kotlin/okhttp3/internal/io/InMemoryFileSystem.kt
@@ -19,6 +19,7 @@ import java.io.File
 import java.io.FileNotFoundException
 import java.io.IOException
 import java.util.IdentityHashMap
+import okhttp3.TestUtil.isDescendentOf
 import okio.Buffer
 import okio.ForwardingSink
 import okio.ForwardingSource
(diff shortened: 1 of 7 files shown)


okhttp-testing-support/src/main/kotlin/okhttp3/internal/io/InMemoryFileSystem.kt
```diff
@@ -30,70 +31,52 @@ import org.junit.runners.model.Statement
...
-    if (!openResources.isEmpty()) {
-      val builder =
-        StringBuilder("Resources acquired but not closed:")
-      for (resource in openResources) {
-        builder.append("\n * ")
-            .append(resource)
-      }
-      throw IllegalStateException(builder.toString())
+    check(openResources.isEmpty()) {
+      "Resources acquired but not closed:\n * ${openResources.joinToString(separator = "\n * ")}"
```
yschimke, line 56: nice

okhttp/src/main/kotlin/okhttp3/internal/cache/DiskLruCache.kt
```diff
@@ -225,6 +228,8 @@ class DiskLruCache internal constructor(
       }
     }
 
+    civilizedFileSystem = fileSystem.isCivilized(journalFileBackup)
```
yschimke, line 231: This is probably worthy of logging when false.

swankjesse: ooooh I think it’d be pretty boring. It’ll pretty much always be true on Windows, and false elsewhere.

Issue #5657, opened 2019-12-20
MockWebServer shutdown is not idempotent


swankjesse:
We enqueue a task to await idle, and that fails.
```
java.util.concurrent.RejectedExecutionException: Task okhttp3.internal.concurrent.TaskRunner$runnable$1@83b95cb rejected from java.util.concurrent.ThreadPoolExecutor@7848d4ed[Shutting down, pool size = 1, active threads = 1, queued tasks = 0, completed tasks = 2]
	at java.base/java.util.concurrent.ThreadPoolExecutor$AbortPolicy.rejectedExecution(ThreadPoolExecutor.java:2055)
	at java.base/java.util.concurrent.ThreadPoolExecutor.reject(ThreadPoolExecutor.java:825)
	at java.base/java.util.concurrent.ThreadPoolExecutor.execute(ThreadPoolExecutor.java:1355)
	at okhttp3.internal.concurrent.TaskRunner$RealBackend.execute(TaskRunner.kt:298)
	at okhttp3.internal.concurrent.TaskRunner.kickCoordinator$okhttp(TaskRunner.kt:91)
	at okhttp3.internal.concurrent.TaskQueue.awaitIdle(TaskQueue.kt:118)
	at okhttp3.mockwebserver.MockWebServer.shutdown(MockWebServer.kt:451)
	at okhttp3.mockwebserver.MockWebServer.after(MockWebServer.kt:460)
``

**Comment by swankjesse** [comment]
Fixed by https://github.com/square/okhttp/pull/5672

**Comment by swankjesse** [comment]
```
java.util.concurrent.RejectedExecutionException: Task okhttp3.internal.concurrent.TaskRunner$runnable$1@4267357c rejected from java.util.concurrent.ThreadPoolExecutor@1232116e[Shutting down, pool size = 1, active threads = 1, queued tasks = 0, completed tasks = 8]
	at java.base/java.util.concurrent.ThreadPoolExecutor$AbortPolicy.rejectedExecution(ThreadPoolExecutor.java:2055)
	at java.base/java.util.concurrent.ThreadPoolExecutor.reject(ThreadPoolExecutor.java:825)
	at java.base/java.util.concurrent.ThreadPoolExecutor.execute(ThreadPoolExecutor.java:1355)
	at okhttp3.internal.concurrent.TaskRunner$RealBackend.execute(TaskRunner.kt:298)
[...]

PR #5946, opened 2020-04-12
Confirm MockWebServer shutdown is idempotent

yschimke:
https://github.com/square/okhttp/issues/5657


yschimke:
TODO try to break it with a long running task

PR #5947, opened 2020-04-12
Kotlin 1.4 dependency upgrade and language features

yschimke:
Update to Kotlin 1.4 and basic upgrade to functionality.  This also includes okio 2.8.0.


yschimke:
@swankjesse There is a library author mode

https://blog.jetbrains.com/kotlin/2020/06/kotlin-1-4-m2-released/#explicit-api-mode
https://github.com/Kotlin/KEEP/blob/master/proposals/explicit-api-mode.md

swankjesse:
I don't think it helps us? I can't think of a place where explicit `public` or return types will make the code more clear.

yschimke:
@swankjesse we ready?

build.gradle
```diff
@@ -62,9 +60,11 @@ buildscript {
 }
 
 plugins {
-  id 'ru.vyarus.animalsniffer' version '1.5.0'
-  id 'com.github.johnrengelman.shadow' version '5.2.0'
-  id 'me.champeau.gradle.japicmp' version '0.2.9'
+  id "ru.vyarus.animalsniffer" version "1.5.0"
```
swankjesse, line 63: The ' quotes don't do interpolation, the " quotes do. Motivation here?

build.gradle
```diff
@@ -205,20 +196,20 @@ subprojects { project ->
     }
   }
 
-  dependencies {
-    //noinspection GradleDynamicVersion
-    errorproneJavac 'com.google.errorprone:javac:9+181-r4173-1'
-    errorprone 'com.google.errorprone:error_prone_core:2.3.4'
-  }
+//  dependencies {
```
swankjesse, line 199: Delete?

build.gradle
```diff
@@ -205,20 +196,20 @@ subprojects { project ->
...
     sourceCompatibility = JavaVersion.VERSION_1_8
     targetCompatibility = JavaVersion.VERSION_1_8
 
-    options.errorprone {
-      check("MissingFail", CheckSeverity.ERROR)
-      check("MissingOverride", CheckSeverity.ERROR)
-      enabled = JavaVersion.current() < JavaVersion.VERSION_12
-    }
+//    options.errorprone {
+//      check("MissingFail", net.ltgt.gradle.errorprone.CheckSeverity.ERROR)
```
swankjesse, line 209: Delete?

okhttp-testing-support/src/main/kotlin/okhttp3/RecordingEventListener.kt
```diff
@@ -129,7 +129,7 @@ open class RecordingEventListener : EventListener() {
     for (lock in forbiddenLocks) {
       assertThat(Thread.holdsLock(lock))
           .overridingErrorMessage(lock.toString())
-          .isFalse()
+          .isFalse
```
swankjesse, line 132: The IntelliJ waning irks me here, and I prefer the quotes. The is false() call has a critical side-effect, but accessing a property should not have side-effects.

I'd prefer to leave these as functions.

okhttp/src/main/kotlin/okhttp3/internal/Util.kt
```diff
@@ -378,18 +375,18 @@ fun Socket.peerName(): String {
  * @param source the source used to read bytes from the socket.
  */
 fun Socket.isHealthy(source: BufferedSource): Boolean {
-  try {
+  return try {
```
swankjesse, line 378: For what it's worth, the return value here is non-obvious with the finally clause.  IntelliJ is obnoxious by making recommendations that make the code less clear.

okhttp/src/main/kotlin/okhttp3/internal/platform/OpenJSSEPlatform.kt
```diff
@@ -50,7 +50,7 @@ class OpenJSSEPlatform private constructor() : Platform() {
     return trustManagers[0] as X509TrustManager
   }
 
-  public override fun trustManager(sslSocketFactory: SSLSocketFactory): X509TrustManager? =
+  override fun trustManager(sslSocketFactory: SSLSocketFactory): X509TrustManager? =
```
swankjesse, line 53: Thanks for fixing all these unnecessary modifiers.

okhttp/src/test/java/okhttp3/CipherSuiteTest.java
```diff
@@ -31,6 +31,8 @@
     try {
       forJavaName(null);
       fail("Should have thrown");
+    } catch (NullPointerException expected) {
+      // Kotlin 1.4
     } catch (IllegalArgumentException expected) {
```
swankjesse, line 36: Due IAE still happen?

swankjesse (approved):
Yay! The fun interfaces are nice.

build.gradle
yschimke: I think that isn't how my brain works, I wouldn't use different strings based on whether it needs interpolation.  I'd just mainly use the interpolated form whether or not is has a variable part.

okhttp/src/main/kotlin/okhttp3/internal/Util.kt
yschimke: The specific reason I like the return here is that it forces a return expression from all branches of the clause.  Compiler would fail here anyway, but makes it clear that the result of the try catch is the final result of the method.

yschimke: Can I argue this one?

okhttp/src/test/java/okhttp3/CipherSuiteTest.java
yschimke: Cleaning up.

okhttp/src/main/kotlin/okhttp3/internal/Util.kt
swankjesse: Of course. And your argument makes sense.

Issue #5895, opened 2020-03-24
Null Pointer Exception During Certificate Pinning (v. 4.4.1)


rahul-narkhede:
When creating a certificate pinned connection using OkHttp 4.4.1, we face a Nul Pointer Exception at the following line 
https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/internal/connection/RealConnection.kt#L393

Bellow is the stack trace
```
2020-03-24 15:36:44.978 6227-6307/com.example.application E/AndroidRuntime: FATAL EXCEPTION: OkHttp Dispatcher
    Process: com.example.application, PID: 6227
    kotlin.KotlinNullPointerException
        at okhttp3.internal.connection.RealConnection$connectTls$1.invoke(RealConnection.kt:393)
        at okhttp3.internal.connection.RealConnection$connectTls$1.invoke(RealConnection.kt:73)
        at okhttp3.Handshake$peerCertificates$2.invoke(Handshake.kt:53)
        at okhttp3.Handshake$peerCertificates$2.invoke(Handshake.kt:34)
        at kotlin.SynchronizedLazyImpl.getValue(LazyJVM.kt:74)
        at okhttp3.Handshake.peerCertificates(Unknown Source:7)
        at okhttp3.internal.connection.RealConnection$connectTls$2.invoke(RealConnection.kt:399)
        at okhttp3.internal.connection.RealConnection$connectTls$2.invoke(RealConnection.kt:73)
        at okhttp3.CertificatePinner.check$okhttp(CertificatePinner.kt:159)
        at okhttp3.internal.connection.RealConnection.connectTls(RealConnection.kt:398)
        at okhttp3.internal.connection.RealConnection.establishProtocol(RealConnection.kt:325)
        at okhttp3.internal.connection.RealConnection.connect(RealConnection.kt:197)
        at okhttp3.internal.connection.ExchangeFinder.findConnection(ExchangeFinder.kt:235)
        at okhttp3.internal.connection.ExchangeFinder.findHealthyConnection(ExchangeFinder.kt:108)
        at okhttp3.internal.connection.ExchangeFinder.find(ExchangeFinder.kt:76)
        at okhttp3.internal.connection.RealCall.initExchange$okhttp(RealCall.kt:245)
        at okhttp3.internal.connection.ConnectInterceptor.intercept(ConnectInterceptor.kt:32)
        at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:100)
[...]

PR #5949, opened 2020-04-13
Clean up SSL state in OkHttpClient

yschimke:
Attempting a fix for https://github.com/square/okhttp/issues/5895.  At worst it should flush out poor behavior from clients or our own edge cases.

### DIFF
--- a/okhttp/src/main/kotlin/okhttp3/OkHttpClient.kt
+++ b/okhttp/src/main/kotlin/okhttp3/OkHttpClient.kt
@@ -231,9 +231,10 @@ open class OkHttpClient internal constructor(
       this.certificatePinner = CertificatePinner.DEFAULT
     } else if (builder.sslSocketFactoryOrNull != null) {
       this.sslSocketFactoryOrNull = builder.sslSocketFactoryOrNull
-      this.certificateChainCleaner = builder.certificateChainCleaner
+      this.certificateChainCleaner = builder.certificateChainCleaner!!
       this.x509TrustManager = builder.x509TrustManagerOrNull
       this.certificatePinner = builder.certificatePinner
+          .withCertificateChainCleaner(certificateChainCleaner!!)
     } else {
       this.x509TrustManager = Platform.get().platformTrustManager()
       Platform.get().configureTrustManager(x509TrustManager)
@@ -241,9 +242,13 @@ open class OkHttpClient internal constructor(
       Platform.get().configureSslSocketFactory(sslSocketFactoryOrNull)
       this.certificateChainCleaner = CertificateChainCleaner.get(x509TrustManager!!)
       this.certificatePinner = builder.certificatePinner
-          .withCertificateChainCleaner(certificateChainCleaner)
+          .withCertificateChainCleaner(certificateChainCleaner!!)
     }
 
+    verifyClientState()
+  }
+
+  private fun verifyClientState() {
     check(null !in (interceptors as List<Interceptor?>)) {
       "Null interceptor: $interceptors"
     }
@@ -251,10 +256,6 @@ open class OkHttpClient internal constructor(
       "Null network interceptor: $networkInterceptors"
     }
 
-    verifyClientSslState()
-  }
-
-  private fun verifyClientSslState() {
     if (connectionSpecs.none { it.isTls }) {
       check(sslSocketFactoryOrNull == null)
       check(certificateChainCleaner == null)
--- a/okhttp/src/main/kotlin/okhttp3/OkHttpClient.kt
+++ b/okhttp/src/main/kotlin/okhttp3/OkHttpClient.kt
@@ -258,10 +258,10 @@ open class OkHttpClient internal constructor(
       check(x509TrustManager == null)
       check(certificatePinner == CertificatePinner.DEFAULT)
     } else {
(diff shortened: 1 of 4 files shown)


okhttp/src/main/kotlin/okhttp3/internal/platform/ConscryptPlatform.kt
```diff
@@ -41,26 +43,18 @@ class ConscryptPlatform private constructor() : Platform() {
...
-          when {
-            sp != null -> readFieldOrNull(sp, X509TrustManager::class.java, "x509TrustManager")
-            else -> null
-          }
-        } catch (e: Exception) {
-          throw UnsupportedOperationException(
-              "clientBuilder.sslSocketFactory(SSLSocketFactory) not supported on Conscrypt", e)
-        }
-      }
+  override fun trustManager(sslSocketFactory: SSLSocketFactory): X509TrustManager? = null
```
yschimke, line 57: Deliberate behaviour change, if a client opts into Conscrypt, they should be knowledgeable enough to provide a trust manager.  FWIW we don't support private use of our internal APIs, so we should stop pushing this upstream also.

yschimke:
TODO probably useful to add some useful examples of chaing OkHttpClient instances to confirm all transitions.  Any thoughts?

But at least this should flush out weird configs like using a sslSocketFactory with an unrelated trustManager etc.  Hopefully flush out any scenarios where we can end with a broken CertificatePinner.

okhttp/src/main/kotlin/okhttp3/OkHttpClient.kt
```diff
@@ -224,30 +222,46 @@ open class OkHttpClient internal constructor(
...
+  private fun verifyClientState() {
     check(null !in (interceptors as List<Interceptor?>)) {
       "Null interceptor: $interceptors"
     }
     check(null !in (networkInterceptors as List<Interceptor?>)) {
       "Null network interceptor: $networkInterceptors"
     }
+
+    if (connectionSpecs.none { it.isTls }) {
+      check(sslSocketFactoryOrNull == null)
```
swankjesse, line 256: This assertion isn’t very interesting! Seems impossible to violate.

swankjesse: Maybe omit these 4 checks, and invert the if ?
```
     if (connectionSpecs.any { it.isTls }) { 
```

okhttp/src/main/kotlin/okhttp3/OkHttpClient.kt
```diff
@@ -734,7 +748,10 @@ open class OkHttpClient internal constructor(
       }
 
       this.sslSocketFactoryOrNull = sslSocketFactory
-      this.certificateChainCleaner = Platform.get().buildCertificateChainCleaner(sslSocketFactory)
+      this.x509TrustManagerOrNull = Platform.get().trustManager(sslSocketFactory) ?: throw IllegalStateException(
```
swankjesse, line 751: I like that we’ll fail sooner

okhttp/src/main/kotlin/okhttp3/internal/platform/ConscryptPlatform.kt
swankjesse: Yeah, very good idea.

okhttp/src/main/kotlin/okhttp3/OkHttpClient.kt
yschimke: OK, will check other case.  I think I had this mainly for testing, I want to leave it here because carrying these through weird chains of clients, is I think how we got into the wacky state with NPE.

okhttp/src/main/kotlin/okhttp3/internal/platform/ConscryptPlatform.kt
```diff
@@ -41,26 +43,18 @@ class ConscryptPlatform private constructor() : Platform() {
...
   override fun platformTrustManager(): X509TrustManager {
-    return Conscrypt.getDefaultX509TrustManager()
+    val trustManagers = TrustManagerFactory.getInstance(TrustManagerFactory.getDefaultAlgorithm()).apply {
+      init(null as KeyStore?)
+    }.trustManagers!!
+    check(trustManagers.size == 1 && trustManagers[0] is X509TrustManager) {
+      "Unexpected default trust managers: ${trustManagers.contentToString()}"
+    }
+    val x509TrustManager = trustManagers[0] as X509TrustManager
+    Conscrypt.setHostnameVerifier(x509TrustManager) { _, _ -> true }
```
yschimke, line 53: @swankjesse Do we want to do this on Android generally also?  This could possibly help with the weird versioning story.

yschimke:
> TODO probably useful to add some useful examples of chaing OkHttpClient instances to confirm all transitions. Any thoughts?
> 
> But at least this should flush out weird configs like using a sslSocketFactory with an unrelated trustManager etc. Hopefully flush out any scenarios where we can end with a broken CertificatePinner.

@swankjesse Any thoughts on whether we need more tests for weird transition of client/builder chaining?

swankjesse:
No I don’t think so.

PR #5966, opened 2020-04-20
Migrate from Travis to GitHub Action for publishing

JakeWharton:


.github/workflows/build.yml
```diff
@@ -0,0 +1,29 @@
...
+
+on: [push, pull_request]
+
+env:
+  GRADLE_OPTS: "-Dorg.gradle.jvmargs=-Xmx4g -Dorg.gradle.daemon=false -Dkotlin.incremental=false"
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    if: github.ref == 'refs/heads/master'
```
yschimke, line 11: How does this combine with pull_request above? (line 3)

JakeWharton: You would have to make a PR _from_ master which will never occur. The PR trigger is for when we add regular builds to this file and turn down Circle builds.

Issue #5970, opened 2020-04-21
OkHttp 3.x fails with OpenJDK Zulu 8.0.252


rpalcolea:
Hi folks,

We found that this particular piece of code in okhttp 3.x: https://github.com/square/okhttp/blob/okhttp_3.14.x/okhttp/src/main/java/okhttp3/internal/platform/Jdk9Platform.java#L81-L94 breaks.

This checks for `setApplicationProtocols` and `getApplicationProtocol` which are JDK 9. however, these ones got backported to Zulu 8.0.252 causing OkHttp believing that is running on jdk9 and initializing a `Jdk9Platform` which breaks projects

Wondering if a patch could be released for 3.x since there are several libraries and projects out there that haven't migrated to 4.x

Thanks

yschimke:
Can you explain more about why it breaks projects.  If these methods are here, isn't it correct to use this Platform? What do you want it to do?

https://github.com/square/okhttp/blob/okhttp_3.14.x/okhttp/src/main/java/okhttp3/internal/platform/Platform.java#L222

Do you want it to use alpn-boot library, or just not support http/2 via ALPN?

DanielThomas:
Those methods are being used to detect Java 9 ALPN support however they're now available on Java 8, but ALPN support is not, so cause runtime failures due to the `Jdk9Platform` being used:

```
java.lang.AssertionError: failed to get ALPN selected protocol
	at okhttp3.internal.platform.Jdk9Platform.getSelectedProtocol(Jdk9Platform.java:68) ~[carson-0.461.0-all.jar:0.461.0]
```

rahulsom:
a41361efcb0d4ed2e7f09313c9e9fcc3d72e837b is what makes >4.3.0 work with 8u252

yschimke:
@DanielThomas Do you have the root cause exception?

throw new AssertionError("failed to get ALPN selected protocol", e);

I'm also curious why this was backported in what was presumably a very recent release https://docs.azul.com/zulu/zulurelnotes/ZuluReleaseNotes/CA_What'sNew.htm

yschimke:
> [a41361e](https://github.com/square/okhttp/commit/a41361efcb0d4ed2e7f09313c9e9fcc3d72e837b) is what makes >4.3.0 work with 8u252

Yep, but the reflection and the directly compiled code should act the same.

DanielThomas:
[...]

PR #5972, opened 2020-04-21
UnsupportedOperationException from JDK8 252 onwards

yschimke:
Fix for https://github.com/square/okhttp/issues/5970

Activates Jdk9Platform on Java 8 when the new methods are present (builds 252+).  Also handles the UnsupportedOperationException.

### DIFF
--- a/okhttp/src/main/kotlin/okhttp3/internal/platform/Jdk9Platform.kt
+++ b/okhttp/src/main/kotlin/okhttp3/internal/platform/Jdk9Platform.kt
@@ -66,9 +66,9 @@ open class Jdk9Platform : Platform() {
     val isAvailable: Boolean
 
     init {
-      val jdkVersion = System.getProperty("java.specification.version")
+      val jdkVersion: String? = System.getProperty("java.specification.version")
 
-      val majorVersion = jdkVersion.toIntOrNull()
+      val majorVersion = jdkVersion?.toIntOrNull()
 
       isAvailable = if (majorVersion != null) {
         majorVersion >= 9
--- a/okhttp/src/main/kotlin/okhttp3/internal/platform/Jdk9Platform.kt
+++ b/okhttp/src/main/kotlin/okhttp3/internal/platform/Jdk9Platform.kt
@@ -66,8 +66,21 @@ open class Jdk9Platform : Platform() {
     val isAvailable: Boolean
 
     init {
-      val majorVersion: Int = Integer.getInteger("java.specification.version") ?: 8
-      isAvailable = majorVersion >= 9
+      val jdkVersion = System.getProperty("java.specification.version")
+
+      val majorVersion = jdkVersion.toIntOrNull()
+
+      isAvailable = if (majorVersion != null) {
+        majorVersion >= 9
+      } else {
+        try {
+          // also present on JDK8 after build 252.
+          SSLSocket::class.java.getMethod("getApplicationProtocol")
+          true
+        } catch (nsme: NoSuchMethodException) {
+          false
+        }
+      }
     }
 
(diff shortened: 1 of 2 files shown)


yschimke:
Raised by @DanielThomas we need to change activation logic here also.  Otherwise we won't work on 8-252

okhttp/src/main/kotlin/okhttp3/internal/platform/Jdk9Platform.kt
```diff
@@ -40,11 +40,16 @@ open class Jdk9Platform : Platform() {
...
+    try {
+      // SSLSocket.getApplicationProtocol returns "" if application protocols values will not
+      // be used. Observed if you didn't specify SSLParameters.setApplicationProtocols
+      return when (val protocol = sslSocket.applicationProtocol) {
+        null, "" -> null
+        else -> protocol
+      }
+    } catch (e: UnsupportedOperationException) {
+      // https://docs.oracle.com/javase/9/docs/api/javax/net/ssl/SSLSocket.html#getApplicationProtocol--
+      return null
```
monkey-mas, line 52: Maybe we can add test in `Jdk9PlatformTest`?
```java
  @Test
  public void selectedProtocolIsNullWhenSslSocketThrowsExceptionForApplicationProtocol() {
    assertThat(new Jdk9Platform().getSelectedProtocol(
        new SSLSocket() {
          @Override public String getApplicationProtocol() {
            throw new UnsupportedOperationException("Mock exception");
          }

          // Implement abstract methods.
          @Override public String[] getSupportedCipherSuites() {
            return new String[0];
          }

          @Override public String[] getEnabledCipherSuites() {
            return new String[0];
          }

          @Override public void setEnabledCipherSuites(String[] suites) {

          }

          @Override public String[] getSupportedProtocols() {
            return new String[0];
          }

          @Override public String[] getEnabledProtocols() {
            return new String[0];
          }

          @Override public void setEnabledProtocols(String[] protocols) {

          }

          @Override public SSLSession getSession() {
            return null;
          }

          @Override public void addHandshakeCompletedListener(HandshakeCompletedListener listener) {

          }

          @Override
          public void removeHandshakeCompletedListener(HandshakeCompletedListener listener) {

          }

          @Override public void startHandshake() throws IOException {

          }

          @Override public void setUseClientMode(boolean mode) {

          }

          @Override public boolean getUseClientMode() {
            return false;
          }

          @Override public void setNeedClientAuth(boolean need) {

          }

          @Override public boolean getNeedClientAuth() {
            return false;
          }

          @Override public void setWantClientAuth(boolean want) {

          }

          @Override public boolean getWantClientAuth() {
            return false;
          }

          @Override public void setEnableSessionCreation(boolean flag) {

          }

          @Override public boolean getEnableSessionCreation() {
            return false;
          }
        })).isNull();
```

monkey-mas:
@swankjesse Can you also take a look please?

okhttp/src/test/java/okhttp3/internal/platform/Jdk9PlatformTest.java
```diff
@@ -15,22 +15,118 @@
...
   public void testToStringIsClassname() {
     assertThat(new Jdk9Platform().toString()).isEqualTo("Jdk9Platform");
   }
+
+  @Test
+  public void selectedProtocolIsNullWhenSslSocketThrowsExceptionForApplicationProtocol() {
+    platform.assumeJdk9();
+
+    assertThat(new Jdk9Platform().getSelectedProtocol(
+        new SSLSocket() {
```
swankjesse, line 61: `new DelegatingSSLSocket(null)` ?

monkey-mas: > new DelegatingSSLSocket(null) ?
Cool! Didn't realize that!

PR #5978, opened 2020-04-23
Java Net Authenticator

yschimke:
Refactor the okhttp-urlconnection implementation.  Available as 

```
client.newBuilder()
        .proxyAuthenticator(okhttp3.Authenticator.JAVA_NET_AUTHENTICATOR)
        .build()
```

### DIFF
--- a/okhttp/src/main/kotlin/okhttp3/Authenticator.kt
+++ b/okhttp/src/main/kotlin/okhttp3/Authenticator.kt
@@ -116,7 +116,7 @@ interface Authenticator {
       override fun authenticate(route: Route?, response: Response): Request? = null
     }
 
-    /** An authenticator that knows no credentials and makes no attempt to authenticate. */
+    /** An authenticator that uses the java.net.Authenticator global authenticator. */
     @JvmField
     val JAVA_NET_AUTHENTICATOR: Authenticator = JavaNetAuthenticator()
   }
--- a/okhttp/src/main/kotlin/okhttp3/Authenticator.kt
+++ b/okhttp/src/main/kotlin/okhttp3/Authenticator.kt
@@ -15,8 +15,8 @@
  */
 package okhttp3
 
-import okhttp3.internal.authenticator.JavaNetAuthenticator
 import java.io.IOException
+import okhttp3.internal.authenticator.JavaNetAuthenticator
 
 /**
  * Performs either **preemptive** authentication before connecting to a proxy server, or
--- a/okhttp/src/main/kotlin/okhttp3/Authenticator.kt
+++ b/okhttp/src/main/kotlin/okhttp3/Authenticator.kt
@@ -15,6 +15,7 @@
  */
 package okhttp3
 
+import okhttp3.internal.authenticator.JavaNetAuthenticator
 import java.io.IOException
 
 /**
@@ -114,5 +115,9 @@ interface Authenticator {
     private class AuthenticatorNone : Authenticator {
       override fun authenticate(route: Route?, response: Response): Request? = null
     }
+
+    /** An authenticator that knows no credentials and makes no attempt to authenticate. */
+    @JvmField
+    val JAVA_NET_AUTHENTICATOR: Authenticator = JavaNetAuthenticator()
   }
 }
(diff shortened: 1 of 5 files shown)


okhttp/src/main/kotlin/okhttp3/internal/authenticator/JavaNetAuthenticator.kt
```diff
@@ -0,0 +1,94 @@
...
+      }
+    }
+
+    return null // No challenges were satisfied!
+  }
+
+  @Throws(IOException::class)
+  private fun Proxy.connectToInetAddress(url: HttpUrl, dns: Dns): InetAddress {
+    return when (type()) {
+      Proxy.Type.DIRECT -> dns.lookup(url.host).first()
```
yschimke, line 90: Changed from existing java impl.

swankjesse: Nice improvement

okhttp/src/main/kotlin/okhttp3/Authenticator.kt
```diff
@@ -114,5 +115,9 @@ interface Authenticator {
     private class AuthenticatorNone : Authenticator {
       override fun authenticate(route: Route?, response: Response): Request? = null
     }
+
+    /** An authenticator that knows no credentials and makes no attempt to authenticate. */
```
swankjesse, line 119: Copy pasta problem

This has credentials; they’re installed in `java.net.Authenticator` !

okhttp/src/test/java/okhttp3/internal/authenticator/JavaNetAuthenticatorTest.kt
```diff
@@ -0,0 +1,74 @@
+package okhttp3.internal.authenticator
```
swankjesse, line 1: supernit: copyright

okhttp/src/test/java/okhttp3/internal/authenticator/JavaNetAuthenticatorTest.kt
```diff
@@ -0,0 +1,74 @@
...
+        .message("Unauthorized")
+        .build()
+    val authRequest = authenticator.authenticate(route, response)
+
+    assertNotNull(authRequest)
+    assertEquals(
+        "Basic ${RecordingAuthenticator.BASE_64_CREDENTIALS}", authRequest!!.header("Authorization")
+    )
+  }
+}
```
monkey-mas, line 74: nit: maybe we can add another test for non-Basic auth request? I might be wrong though.
```kotlin
  @Test
  fun noSupportForNonBasicAuth() {
        .url("https://server/robots.txt")
        .build()

    val response = Response.Builder()
        .request(request)
        .code(401)
        .header("WWW-Authenticate", "UnsupportedScheme realm=\"User Visible Realm\"")
        .protocol(HTTP_2)
        .message("Unauthorized")
        .build()

    val authRequest = authenticator.authenticate(null, response)
    assertNull(authRequest)
  }
```

okhttp/src/test/java/okhttp3/internal/authenticator/JavaNetAuthenticatorTest.kt
```diff
@@ -0,0 +1,74 @@
...
+    val response = Response.Builder()
+        .request(request)
+        .code(401)
+        .header("WWW-Authenticate", "Basic realm=\"User Visible Realm\"")
+        .protocol(HTTP_2)
+        .message("Unauthorized")
+        .build()
+    val authRequest = authenticator.authenticate(route, response)
+
+    assertNotNull(authRequest)
```
monkey-mas, line 69: supernit: given that `authRequest!!.header("Authorization")` throws NPE when `authRequest` is null, we could drop this non-null assertion.

okhttp/src/test/java/okhttp3/internal/authenticator/JavaNetAuthenticatorTest.kt
yschimke: We should automate this in spotless :)

Would save you a lot of typing on my reviews :)

Issue #5833, opened 2020-02-29
Add WebSocket EventListener


change2:
i use okhttp connect a websocket service ,  i want know websocket heart beat failed can tell my program,but now can'not 
i hope
websocket heartbeat  failed. can throw special exception ?
if failed can log heart beat failed . but now nothing

yschimke:
Heartbeats are not directly exposed, sorry.  They are deliberately not public API.

changealice:
can print some log may let me known heartbeat heathly ?

yschimke:
This is currently disabled so an enhancement we should consider on it's merits.

```
  @Test public void webSocketsDontTriggerEventListener() {
    RecordingEventListener listener = new RecordingEventListener();
```

ShivamArya2357:
I want to work in this issue

yschimke:
> I want to work in this issue

@ShivamArya2357 A few concerns here

1) I know @swankjesse prefers to avoid "licking the cookie". If someone submits a PR, it will be reviewed, but generally we don't tell people not to work on things because someone else has already claimed it.  An inflight PR would generally show that you are actively working on it already.
2) This is a non-trivial PR, so potentially a frustrating experience for a first time contributor and the reviewer.

Those generally combine to discourage first time contributors unfortunately, but for a non-trivial PR the work reviewing iterative versions of the PR is potentially far greater than writing it.

That leaves a gap where you ignore my (friendly) advice, and submit a high quality PR that looks close enough to ready that someone on the project reviews your PR and helps get it ready to land.  That's how the vast majority of external PRs tend to succeed.  Still, there are examples where a high quality PR is still considered against other project aims (e.g. code simplicity) and doesn't land.  So this can be frustrating for you.

swankjesse:
This issue isn’t available for new contributors.

bentemple:
Having this listener would also be really helpful since then we can use it to determine the average latency to the server.

Ideally the listener callback would either contain the total round trip ping/pong time with the server, OR a callback whenever the ping is called, and a callback whenever the pong is received.

baldhead69:
Hi,

Some evolution on this ????

yschimke:
No.  I was hoping to look into it, but haven't started anything. Not sure if anyone else is looking into it.

yschimke:
Closing as low priority.

Issue #5319, opened 2019-08-01
Expose Http2 Frame listener to allow non logging debugging


yschimke:
https://github.com/square/okhttp/issues/5314

Compare OkHttp logging to nghttp2.

yschimke:
Won't fix.

yschimke:
Making a judgement call that adding listeners for individual frames is unlikely to be a near term API we expose.

yschimke:
From worth https://stackoverflow.com/questions/65225514/handling-http-2-goaway-in-okhttp4

worth noting that we drop useful error data from APNs

PR #5991, opened 2020-04-26
Centralise logging [RFC]

yschimke:
I keep hitting weirdness in logging when I try to use it, and wanted to tighten up expectations and consistency over the logging approaches.


okhttp/src/main/kotlin/okhttp3/internal/platform/Platform.kt
```diff
@@ -119,9 +120,12 @@ open class Platform {
     socket.connect(address, connectTimeout)
   }
 
-  open fun log(message: String, level: Int = INFO, t: Throwable? = null) {
-    val logLevel = if (level == WARN) Level.WARNING else Level.INFO
-    logger.log(logLevel, message, t)
+  @Deprecated("Moved to logger field", replaceWith = ReplaceWith("logger.info(message, t)"))
```
swankjesse, line 123: nit: we should never deprecate internal APIs; we can just fix ’em.

android-test/src/androidTest/java/okhttp/android/test/OkHttpTest.kt
```diff
@@ -297,7 +316,7 @@ class OkHttpTest {
       moshi.adapter(HowsMySslResults::class.java).fromJson(response.body!!.string())!!
     }
 
-    Platform.get().log("results $results", Platform.WARN)
+    Platform.get().getLogger(OkHttpTest::class.java.simpleName).warn("results $results")
```
swankjesse, line 319: I don’t think you want simpleName here?

okhttp/src/main/kotlin/okhttp3/internal/platform/Jdk8WithJettyBootPlatform.kt
```diff
@@ -62,7 +62,9 @@ class Jdk8WithJettyBootPlatform(
     try {
       val provider = Proxy.getInvocationHandler(getMethod.invoke(null, sslSocket)) as AlpnProvider
       if (!provider.unsupported && provider.selected == null) {
-        log("ALPN callback dropped: HTTP/2 is disabled. " + "Is alpn-boot on the boot class path?")
+        logger.debug(
```
swankjesse, line 65: did the level here change? I think that’s fine but it should be a separate PR

okhttp/src/main/kotlin/okhttp3/internal/platform/Platform.kt
```diff
@@ -119,9 +120,12 @@ open class Platform {
...
+  fun log(message: String, level: Int = INFO, t: Throwable? = null) {
+    when (level) {
+      INFO -> logger.info(message, t)
+      WARN -> logger.warn(message, t)
```
swankjesse, line 127: did debug disappear?

swankjesse:
Maybe I should have started with this question: what kind of weirdness are you running into?

swankjesse:
It’s too bad java.util.Logging misses the mark so badly.

yschimke:
> It’s a nice abstraction but I really hate to introduce our own logging abstraction.

Agreed.  Awkwardly we have some conflicting requirements.

1. We want it to do the right thing by default by two distinct platforms (Android + Java8+)
2. We don't want additional dependencies.
3. Platform.get().log is a useful way of notifying the developer of things in dev mode.  In JVM using JUL achieves this, on Android System.out.println or Log.i do the same.

yschimke:
> I’m grumpy because I just read this which really resonated with me.
> https://sobolevn.me/2020/03/do-not-log

+100 Hopefully this exact topic comes across as a theme I've been pushing for.  A lot of the code in exploratory android work has been trying to build events out of connections, networks etc.

https://github.com/square/okhttp/issues/5194
https://github.com/square/okhttp/issues/5833
https://github.com/square/okhttp/issues/5319

JSSE is the worst example to be honest.  Such a critical API and failures are only debuggable via turning on logging.

I'll raise a clear ticket to cleanup logging and replace with events and observable state.

yschimke:
> Could we use something like OkHttpDebugLogcat so we exclusively use the JUL logging API instead of creating our own?

This is probably the best short term solution, switch to JUL completely, but hook up our own zero dependency logging by default with AndroidPlatform.

yschimke:
> Maybe I should have started with this question: what kind of weirdness are you running into?

It's not fun trying to handle logging in useful way e.g.

https://github.com/yschimke/okurl/blob/master/src/main/kotlin/com/baulsupp/okurl/util/LoggingUtil.kt#L40

PR #5993, opened 2020-04-26
Multipart reader first steps

swankjesse:
Still to do:
 - API to get the boundary parameter from a MediaType
 - API to create a MultipartReader from a ResponseBody
 - Update to support RFC 2387 features?

### DIFF
--- a/okhttp/src/main/kotlin/okhttp3/MultipartReader.kt
+++ b/okhttp/src/main/kotlin/okhttp3/MultipartReader.kt
@@ -1,0 +1,203 @@
+/*
+ * Copyright (C) 2020 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okhttp3
+
+import java.io.Closeable
+import java.io.IOException
+import java.net.ProtocolException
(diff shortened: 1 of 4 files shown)


okhttp/src/main/kotlin/okhttp3/MultipartReader.kt
```diff
@@ -0,0 +1,203 @@
...
+ * [close].
+ *
+ * [rfc_2046]: http://www.ietf.org/rfc/rfc2046.txt
+ */
+class MultipartReader(
+  private val source: BufferedSource,
+  @get:JvmName("boundary") val boundary: String
+) : Closeable {
+  /** This delimiter typically precedes the first part. */
+  private val dashDashBoundary = Buffer()
```
yschimke, line 65: Shouldn't these be defined on a companion object to avoid allocating and executing on each instance?

okhttp/src/main/kotlin/okhttp3/MultipartReader.kt
```diff
@@ -0,0 +1,203 @@
...
+      // This is a subsequent part or a preamble. Skip until "\r\n--" followed by the boundary.
+      while (true) {
+        val toSkip = currentPartBytesRemaining(maxResult = 8192)
+        if (toSkip == 0L) break
+        source.skip(toSkip)
+      }
+      source.skip(crlfDashDashBoundary.size.toLong())
+    }
+
+    // Read either \r\n or --\r\n to determine of there is another part.
```
yschimke, line 114:
```suggestion
    // Read either \r\n or --\r\n to determine if there is another part.
```

okhttp/src/main/kotlin/okhttp3/internal/http1/HeadersReader.kt
```diff
@@ -0,0 +1,49 @@
...
+  /** Read a single line counted against the header size limit. */
+  fun readLine(): String {
+    val line = source.readUtf8LineStrict(headerLimit)
+    headerLimit -= line.length.toLong()
+    return line
+  }
+
+  /** Reads headers or trailers. */
+  fun readHeaders(): Headers {
+    val result = Builder()
```
yschimke, line 37: nit: I know someone who hates unqualified Builder references

PR #5994, opened 2020-04-27
Add an API to look up media type parameters

swankjesse:
This changes our behavior for repeated charsets. Previously we would
crash; now we return the first value.

### DIFF
--- a/okhttp/src/main/kotlin/okhttp3/MediaType.kt
+++ b/okhttp/src/main/kotlin/okhttp3/MediaType.kt
@@ -38,7 +38,8 @@ class MediaType private constructor(
    */
   @get:JvmName("subtype") val subtype: String,
 
-  private val charset: String?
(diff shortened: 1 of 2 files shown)


okhttp/src/main/kotlin/okhttp3/MediaType.kt
```diff
@@ -47,13 +48,27 @@ class MediaType private constructor(
...
       defaultValue // This charset is invalid or unsupported. Give up.
     }
   }
 
+  /**
+   * Returns the parameter [name] of this media type, or null if this media type does not define
+   * such a parameter.
+   */
+  fun parameter(name: String): String? {
+    for (i in parameterNamesAndValues.indices step 2) {
```
yschimke, line 64: TIL: step 2

yschimke:
Maybe worth interning some standard types from teh internets.

PR #5998, opened 2020-04-27
Update the public suffixes DB

swankjesse:
Also check in the raw public suffixes DB into our test/ directory for
easier diffing


swankjesse:
View changes to the public suffixes list here:
https://github.com/square/okhttp/pull/5998/commits/16ca1d46c66a3a9244d7634447281da1dd96b9c7

PR #6002, opened 2020-04-29
Rewrite the location header in the reverting interceptor

swankjesse:


swankjesse:
I think this is less hacky than changing the status.

yschimke:
> I think this is less hacky than changing the status.

Kind of jealous your rewrites are so much nicer than my first attempt :)

PR #6008, opened 2020-05-01
Confirm android logging disabled

yschimke:


android-test/src/androidTest/java/okhttp/android/test/OkHttpTest.kt
```diff
@@ -599,6 +605,56 @@ class OkHttpTest {
...
+        .execute()
+
+    response.use {
+      assertEquals(200, response.code)
+      assertEquals(Protocol.HTTP_2, response.protocol)
+    }
+
+    // Only logs to the test logger above
+    // Will fail if "adb shell setprop log.tag.okhttp.Http2 DEBUG" is called
+    assertEquals(setOf(OkHttpTest::class.java.name), testHandler.calls.keys)
```
swankjesse, line 655: How did this get added?

swankjesse: Ahhh it's in the test setup!

adriancole: #6011 drifted this I think. can someone with more experience confirm? Basically when this is invoked the level is always FINEST and hasn't to do with adb settings

yschimke: Probably, thanks for the attention here! Is it worth splitting this out into it's own test?

The goal is to make sure we don't accidentally enable debug logging for millions of Android apps filling mobile disks or causing PII incidents etc.

adriancole: > Probably, thanks for the attention here! Is it worth splitting this out into it's own test?

assume this is so that we don't inherit the logging stuff of the test rule? could make sense!

> The goal is to make sure we don't accidentally enable debug logging for millions of Android apps filling mobile disks or causing PII incidents etc.

this will be a good code comment to make over the test once we decide what to do with it

PR #6010, opened 2020-05-01
Android Cache Test 

yschimke:


android-test/src/androidTest/java/okhttp/android/test/OkHttpTest.kt
```diff
@@ -595,6 +598,59 @@ class OkHttpTest {
...
+            assertNull(it.networkResponse)
+          }
+
+      Thread.sleep(5000)
+
+      client.newCall(request)
+          .execute()
+          .use {
+            assertEquals(200, it.code)
+            // TODO understand this?
```
swankjesse, line 641: It means the network was used to check whether the cache was fresh

android-test/src/androidTest/java/okhttp/android/test/OkHttpTest.kt
```diff
@@ -595,6 +598,59 @@ class OkHttpTest {
...
+    val cacheSize = 1L * 1024 * 1024 // 1MB
+    val cache = Cache(ctxt.cacheDir.resolve("testCache"), cacheSize)
+
+    try {
+      client = client.newBuilder()
+          .cache(cache)
+          .build()
+
+      val request = Request.Builder()
+          .url("https://httpbin.org/cache/3")
```
swankjesse, line 616: use MockWebServer here?

yschimke: I've previously preferred real world servers to confirm that things are working end to end.  But given the tests are relatively stable, it's probably a good point to switch from manually running to running in Github.

Will copy the approach here https://twitter.com/colinwhi/status/1244426550918823937

android-test/src/androidTest/java/okhttp/android/test/OkHttpTest.kt
yschimke: I think this always surprises me, that Caches don't stop the network request got valid content. :)

I need to read up more on our heuristic caching.

https://www.mnot.net/blog/2009/02/24/unintended_caching
https://paulcalvano.com/index.php/2018/03/14/http-heuristic-caching-missing-cache-control-and-expires-headers-explained/

yschimke: > It means the network was used to check whether the cache was fresh

Actually I'm convinced it is the opposite case, the cache was fetched, found stale and then a network call was made.

android-test/src/androidTest/java/okhttp/android/test/OkHttpTest.kt
swankjesse: Hell yes. Yay @colinrtwhite !

colinrtwhite: Nice! Glad this was useful for y'all 😄

PR #6015, opened 2020-05-03
Cache Events

yschimke:
Add cache events to Event Listener, mainly to avoid confusing CallStart>CallEnd or CallStart>CallFailed responses due to Cache strategy.

### DIFF
--- a/okhttp/src/main/kotlin/okhttp3/EventListener.kt
+++ b/okhttp/src/main/kotlin/okhttp3/EventListener.kt
@@ -422,6 +422,28 @@ abstract class EventListener {
   ) {
   }
 
+  /**
+   * Invoked when a call fails due to cache rules. e.g. onlyIfCached and no cache entry.
+   */
+  open fun cacheFailure(call: Call, response: Response) {
+  }
+
+  /**
+   * Invoked when a result is served from the cache. n.b The response should be used
+   * to determine exactly how the cache came to be used, e.g. with a conditional cache response
+   * from the network.
+   */
+  open fun cacheMiss(call: Call, response: Response) {
+  }
+
+  /**
+   * Invoked when a response is served from the network. n.b The response should be used
+   * to determine exactly how the cache came to be used, e.g. with a conditional cache response
+   * from the network.
+   */
+  open fun cacheHit(call: Call, response: Response) {
+  }
+
   interface Factory {
     /**
      * Creates an instance of the [EventListener] for a particular [Call]. The returned
--- a/okhttp/src/main/kotlin/okhttp3/EventListener.kt
+++ b/okhttp/src/main/kotlin/okhttp3/EventListener.kt
@@ -430,19 +430,19 @@ abstract class EventListener {
   }
 
   /**
-   * Invoked when a result is served from the cache. The Response is provided since there may
-   * not be the normal event sequences received.
+   * Invoked when a result is served from the cache. The Response provided is the top level
+   * Response and normal event sequences will not be received.
    *
-   * Events will only be received when a Cache is configured for the client.
+   * This event will only be received when a Cache is configured for the client.
    */
   open fun cacheHit(call: Call, response: Response) {
   }
 
   /**
    * Invoked when a response will be served from the network. The Response will be
-   * available via following events.
+   * available from normal event sequences.
    *
-   * Events will only be received when a Cache is configured for the client.
+   * This event will only be received when a Cache is configured for the client.
    */
   open fun cacheMiss(call: Call) {
   }
@@ -452,7 +452,7 @@ abstract class EventListener {
    * cached Response freshness. Will be followed by cacheHit or cacheMiss after the network
    * Response is available.
    *
-   * Events will only be received when a Cache is configured for the client.
+   * This event will only be received when a Cache is configured for the client.
    */
   open fun cacheConditionalHit(call: Call, cachedResponse: Response) {
   }
--- a/okhttp/src/main/kotlin/okhttp3/EventListener.kt
+++ b/okhttp/src/main/kotlin/okhttp3/EventListener.kt
@@ -430,9 +430,8 @@ abstract class EventListener {
   }
 
   /**
-   * Invoked when a result is served from the cache. n.b The response should be used
-   * to determine exactly how the cache came to be used, e.g. with a conditional cache response
-   * from the network.
+   * Invoked when a result is served from the cache. The Response is provided since their may
+   * not be the normal event sequences received.
    *
    * Events will only be received when a Cache is configured for the client.
    */
--- a/okhttp/src/main/kotlin/okhttp3/EventListener.kt
+++ b/okhttp/src/main/kotlin/okhttp3/EventListener.kt
@@ -424,6 +424,8 @@ abstract class EventListener {
 
   /**
    * Invoked when a call fails due to cache rules. e.g. onlyIfCached and no cache entry.
+   *
+   * Events will only be received when a Cache is configured for the client.
    */
   open fun cacheFailure(call: Call, response: Response) {
   }
@@ -432,6 +434,8 @@ abstract class EventListener {
    * Invoked when a result is served from the cache. n.b The response should be used
    * to determine exactly how the cache came to be used, e.g. with a conditional cache response
    * from the network.
+   *
+   * Events will only be received when a Cache is configured for the client.
    */
   open fun cacheMiss(call: Call, response: Response) {
   }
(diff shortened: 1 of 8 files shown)


yschimke:
@swankjesse I tried to implement a cacheFallback test but didn't succeed. e.g. instead of a 304, are there any situations where we should fallback to the cache response with either a Etag, or max-age but we hit the network for heuristic reasons?

okhttp/src/test/java/okhttp3/EventListenerTest.java
```diff
@@ -1443,4 +1449,107 @@ public void redirectUsingNewConnectionEventSequence() throws IOException {
...
+
+    listener.clearAllEvents();
+
+    call = call.clone();
+    response = call.execute();
+    assertThat(response.code()).isEqualTo(200);
+    assertThat(response.body().string()).isEqualTo("abc");
+    response.close();
+
+    assertThat(listener.recordedEventTypes()).containsExactly("CallStart", "CacheHit", "CallEnd");
```
yschimke, line 1541: Without this event you don't get a chance to look at networkResponse or cacheResponse either.

okhttp/src/main/kotlin/okhttp3/EventListener.kt
```diff
@@ -422,6 +422,34 @@ abstract class EventListener {
...
+  }
+
+  /**
+   * Invoked when a result is served from the cache. n.b The response should be used
+   * to determine exactly how the cache came to be used, e.g. with a conditional cache response
+   * from the network.
+   *
+   * Events will only be received when a Cache is configured for the client.
+   */
+  open fun cacheMiss(call: Call, response: Response) {
```
swankjesse, line 440: is `response` the application-layer response? Will it have cacheResponse or networkResponse fields defined?

okhttp/src/main/kotlin/okhttp3/EventListener.kt
```diff
@@ -422,6 +422,34 @@ abstract class EventListener {
...
+  }
+
+  /**
+   * Invoked when a response is served from the network. n.b The response should be used
+   * to determine exactly how the cache came to be used, e.g. with a conditional cache response
+   * from the network.
+   *
+   * Events will only be received when a Cache is configured for the client.
```
swankjesse, line 448: is `response` the application-layer response? Will it have cacheResponse or networkResponse fields defined?

okhttp/src/main/kotlin/okhttp3/EventListener.kt
```diff
@@ -422,6 +422,34 @@ abstract class EventListener {
   ) {
   }
 
+  /**
+   * Invoked when a call fails due to cache rules. e.g. onlyIfCached and no cache entry.
+   *
+   * Events will only be received when a Cache is configured for the client.
+   */
+  open fun cacheFailure(call: Call, response: Response) {
```
swankjesse, line 430: I think this wants to be called unsatisfiableRequest ? You don’t need a cache to get this error!

okhttp3.CallTest#onlyIfCachedReturns504WhenNotCached

swankjesse: or maybe `satisfactionFailure` ?!

okhttp/src/main/kotlin/okhttp3/internal/cache/CacheInterceptor.kt
```diff
@@ -67,14 +71,21 @@ class CacheInterceptor(internal val cache: Cache?) : Interceptor {
           .body(EMPTY_RESPONSE)
           .sentRequestAtMillis(-1L)
           .receivedResponseAtMillis(System.currentTimeMillis())
-          .build()
+          .build().also {
+            // unconditionally log a cache related failure
```
swankjesse, line 75: supernit: https://nedbatchelder.com/blog/201401/comments_should_be_sentences.html

okhttp/src/main/kotlin/okhttp3/internal/cache/CacheInterceptor.kt
```diff
@@ -67,14 +71,21 @@ class CacheInterceptor(internal val cache: Cache?) : Interceptor {
...
+          }
     }
 
     // If we don't need the network, we're done.
     if (networkRequest == null) {
       return cacheResponse!!.newBuilder()
           .cacheResponse(stripBody(cacheResponse))
-          .build()
+          .build().also {
+            if (cache != null) {
```
swankjesse, line 85: this null check looks unnecessary?

okhttp/src/main/kotlin/okhttp3/internal/cache/CacheInterceptor.kt
```diff
@@ -104,7 +115,9 @@ class CacheInterceptor(internal val cache: Cache?) : Interceptor {
         // Content-Encoding header (as performed by initContentStream()).
         cache!!.trackConditionalCacheHit()
         cache.update(cacheResponse, response)
-        return response
+        return response.also {
+          listener.cacheHit(call, it)
```
swankjesse, line 119: should we track conditionalCacheHits separately?

swankjesse: There’s decisions about whether we got a response from the network or not, got a response from the cache or not, and which one won

okhttp/src/test/java/okhttp3/EventListenerTest.java
```diff
@@ -1443,4 +1449,107 @@ public void redirectUsingNewConnectionEventSequence() throws IOException {
...
+        .build());
+    Response response = call.execute();
+    assertThat(response.code()).isEqualTo(200);
+    assertThat(response.body().string()).isEqualTo("abc");
+    response.close();
+
+    assertThat(listener.recordedEventTypes()).containsExactly("CallStart",
+        "ProxySelectStart", "ProxySelectEnd", "DnsStart", "DnsEnd",
+        "ConnectStart", "ConnectEnd", "ConnectionAcquired", "RequestHeadersStart",
+        "RequestHeadersEnd", "ResponseHeadersStart", "ResponseHeadersEnd", "CacheMiss",
```
swankjesse, line 1470: I think cache miss might be in the wrong place here? The cache miss occurred before proxy selection.

(only a conditional miss occurs after response headers?)

okhttp/src/test/java/okhttp3/EventListenerTest.java
```diff
@@ -1443,4 +1449,107 @@ public void redirectUsingNewConnectionEventSequence() throws IOException {
...
+    assertThat(listener.recordedEventTypes()).containsExactly("CallStart", "CacheHit", "CallEnd");
+  }
+
+  @NotNull private Cache enableCache() throws IOException {
+    cache = makeCache();
+    client = client.newBuilder().cache(cache).build();
+    return cache;
+  }
+
+  @NotNull private Cache makeCache() throws IOException {
```
swankjesse, line 1550: no `@NotNull` in OkHttp code.
https://developer.squareup.com/blog/non-null-is-the-default/

okhttp/src/main/kotlin/okhttp3/internal/cache/CacheInterceptor.kt
yschimke: Yep - this accounting is different than the Cache counters.  I wanted to keep this model simple, but you can always look at the network and cache responses, also prior response to understand more.

But critically, without these new events, EventListeners aren't guaranteed to see a Response to work off.  So this seems pretty useful.

yschimke:
> Thanks for driving this. But I’d like to propose perhaps changing which events we publish.

Amen, happy to use this to work out the right events.

okhttp/src/main/kotlin/okhttp3/EventListener.kt
```diff
@@ -422,6 +422,34 @@ abstract class EventListener {
...
+  }
+
+  /**
+   * Invoked when a result is served from the cache. n.b The response should be used
```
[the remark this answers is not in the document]
yschimke: I think I overuse these!!!

And exclamations marks!!

okhttp/src/main/kotlin/okhttp3/EventListener.kt
yschimke: Yep - was figuring on top level, so you can drop back to our existing logic by introspecting the top level Response.

okhttp/src/main/kotlin/okhttp3/EventListener.kt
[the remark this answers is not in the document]
JakeWharton: I have no idea what nb even means.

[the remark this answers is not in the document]
yschimke: From latin and in that category of i.e., e.g., and n.b

image

I need to stop using.

PR #6020, opened 2020-05-05
Add Cache events to LoggingEventListener

yschimke:


okhttp-logging-interceptor/src/test/java/okhttp3/logging/LoggingEventListenerTest.java
```diff
@@ -240,6 +243,27 @@ public void connectFail() {
...
 
+  @Test
+  public void testCacheEvents() {
+    Request request = new Request.Builder().url(url).build();
+    Call call = client.newCall(request);
+    Response response = new Response.Builder().request(request).code(200).message("").protocol(HTTP_2).build();
+
+    EventListener listener = loggingEventListenerFactory.create(call);
+
+    listener.cacheConditionalHit(call, response);
```
yschimke, line 254: There seems little value in testing all combination against here in this test given it's a pass through.

Issue #6025, opened 2020-05-07
MediaType.parse is too lenient on Android


yschimke:
See discussion in https://github.com/square/okhttp/issues/4535

PR #6029, opened 2020-05-07
Mediatype Android fixes

yschimke:
Enforce JDK MediaType rules on Android.

https://docs.oracle.com/javase/7/docs/api/java/nio/charset/Charset.html

Fix for https://github.com/square/okhttp/issues/6025

### DIFF
--- a/android-test/src/androidTest/java/okhttp/android/test/issues/MediaTypeTest.java
+++ b/android-test/src/androidTest/java/okhttp/android/test/issues/MediaTypeTest.java
@@ -1,0 +1,250 @@
+/*
+ * Copyright (C) 2013 Square, Inc.
+ * Copyright (C) 2011 The Guava Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


okhttp/src/main/kotlin/okhttp3/MediaType.kt
```diff
@@ -50,7 +52,12 @@ class MediaType private constructor(
   fun charset(defaultValue: Charset? = null): Charset? {
     val charset = parameter("charset") ?: return defaultValue
     return try {
-      Charset.forName(charset)
+      Charset.forName(charset).also {
+        if (it.name() != charset && !charsetRegex.matches(charset)) {
```
swankjesse, line 56: wanna move the regex creation from where it is now to inside this call?

As is 100% of the time we class load MediaType we’re going to compile that pattern, and likely 0.000% of the time we’ll benefit from that work

android-test/src/androidTest/java/okhttp/android/test/issues/MediaTypeTest.java
```diff
@@ -0,0 +1,250 @@
...
+import static org.junit.Assert.fail;
+
+/**
+ * Test MediaType API and parsing.
+ *
+ * <p>This test includes tests from <a href="https://code.google.com/p/guava-libraries/">Guava's</a>
+ * MediaTypeTest.
+ */
+@RunWith(Parameterized.class)
+public class MediaTypeTest {
```
swankjesse, line 38: duplicating tests is ick

PR #6032, opened 2020-05-08
Custom trust example & API for certificate decoding

yschimke:
https://github.com/square/okhttp/issues/5701


samples/guide/src/main/java/okhttp3/recipes/CustomTrust.java
```diff
@@ -74,185 +153,19 @@ public void run() throws Exception {
         .build();
 
     try (Response response = client.newCall(request).execute()) {
-      if (!response.isSuccessful()) throw new IOException("Unexpected code " + response);
+      if (!response.isSuccessful()) {
+        Headers responseHeaders = response.headers();
```
swankjesse, line 157: is this behavior change desired? printing headers only for failed responses?

okhttp-tls/src/main/kotlin/okhttp3/tls/HeldCertificate.kt
```diff
@@ -516,7 +516,25 @@ class HeldCertificate(
...
+     * cHAwWTATBgcqhkjOPQIBBggqhkjOPQMBBwNCAASda8ChkQXxGELnrV/oBnIAx3dD
+     * ocUOJfdz4pOJTP6dVQB9U3UBiW5uSX/MoOD0LL5zG3bVyL3Y6pDwKuYvfLNhoyAw
+     * HjAcBgNVHREBAf8EEjAQhwQBAQEBgghjYXNoLmFwcDAKBggqhkjOPQQDAgNIADBF
+     * AiAyHHg1N6YDDQiY920+cnI5XSZwEGhAtb9PYWO8bLmkcQIhAI2CfEZf3V/obmdT
+     * yyaoEufLKVXhrTQhRfodTeigi4RX
+     * -----END CERTIFICATE-----
+     * ```
+     */
+    @JvmStatic
+    fun decodeCertificate(pem: String): X509Certificate {
```
swankjesse, line 537: ooooh neat

swankjesse: rename to `decodeCertificatePem` ?

Should this be an extension on String?
https://publicobject.com/2019/05/22/factory-or-extension-function/

swankjesse: Also I don’t think this is the right file for this method. Should we have a file for certificates?

swankjesse: HeldCertificate is intended to model certificates where we have a private key. Mabybe umm `object Certifcates { ... }` ?

samples/guide/src/main/java/okhttp3/recipes/CustomTrust.java
yschimke: Yeah, mainly the interesting thing was the logo rather than headers.  Was there a point to the headers?

yschimke: I thought for the kotlin example it was more intresting to show the handshake peerCertificates.

```
          for (peerCertificate in response.handshake?.peerCertificates.orEmpty()) {
            println((peerCertificate as X509Certificate).subjectDN)
          }
```

swankjesse: Good call.

samples/guide/src/main/java/okhttp3/recipes/kt/CustomTrust.kt
```diff
@@ -16,213 +16,162 @@
...
 
   init {
-    val (keyManagers, trustManagers) = trustManagerForCertificates(trustedCertificatesInputStream())
-    val sslContext = SSLContext.getInstance("TLS")
-    sslContext.init(keyManagers, trustManagers, null)
-    val sslSocketFactory = sslContext.socketFactory
-
+    // This implementation just embeds the PEM files in Java strings; most applications will
+    // instead read this from a resource file that gets bundled with the application.
+    val letsEncryptCertificate = letsEncryptCertificateAuthority.decodeCertificatePem()
```
swankjesse, line 136: worthwhile declaring these as Certificates and putting the decode call inline?

okhttp-tls/src/main/kotlin/okhttp3/tls/Certificates.kt
```diff
@@ -0,0 +1,81 @@
...
+    throw IllegalArgumentException("failed to decode certificate", e)
+  }
+}
+
+/**
+ * Returns the certificate encoded in [PEM format][rfc_7468].
+ *
+ * [rfc_7468]: https://tools.ietf.org/html/rfc7468
+ */
+fun X509Certificate.certificatePem(): String {
```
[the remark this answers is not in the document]
yschimke: These will be public API, I'll land this with a final round of changes, but I suggest you actively play with the API from kotlin and java before we bake 4.7.

PR #6040, opened 2020-05-10
HandshakeCertificates.Builder.addInsecureHost()

swankjesse:
This API continues the work started here:
https://github.com/square/okhttp/pull/5872


android-test/src/androidTest/java/okhttp/android/test/InsecureForHostTest.kt
```diff
@@ -0,0 +1,115 @@
...
+import okhttp3.tls.HandshakeCertificates
+import okhttp3.tls.HeldCertificate
+import okhttp3.tls.internal.TlsUtil.localhost
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.Assert.fail
+import org.junit.Rule
+import org.junit.Test
+import javax.net.ssl.SSLException
+
+class InsecureForHostTest {
```
swankjesse, line 32: TODO before merge: delete this copy of this class and get the core OkHttp tests working everywhere!

samples/guide/src/main/java/okhttp3/recipes/kt/DevServer.kt
```diff
@@ -0,0 +1,66 @@
...
+        .setHeader("Location", "https://www.google.com/robots.txt"))
+  }
+
+  val clientCertificates = HandshakeCertificates.Builder()
+      .addPlatformTrustedCertificates()
+      .addInsecureHost(server.hostName)
+      .build()
+
+  val client = OkHttpClient.Builder()
+      .sslSocketFactory(clientCertificates.sslSocketFactory(), clientCertificates.trustManager)
```
yschimke, line 44: We should make this jump more direct between HandshakeCertificates and OkHttpClient, this seems like an ideal way to configure without needing to understand trustmanagers etc.

```
OkHttpClient.Builder().handshakeCertificates(clientCertificates).build()
```

samples/guide/src/main/java/okhttp3/recipes/kt/DevServer.kt
```diff
@@ -0,0 +1,66 @@
...
+    useHttps(handshakeCertificates.sslSocketFactory(), false)
+
+    enqueue(MockResponse()
+        .setResponseCode(HTTP_MOVED_TEMP)
+        .setHeader("Location", "https://www.google.com/robots.txt"))
+  }
+
+  val clientCertificates = HandshakeCertificates.Builder()
+      .addPlatformTrustedCertificates()
+      .addInsecureHost(server.hostName)
```
yschimke, line 40: I like this part of the change, by using this API we don't need to handle all possible cases, just those supported by okhttp-tls.

okhttp-tls/src/main/kotlin/okhttp3/tls/internal/InsecureExtendedTrustManager.kt
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ * This extends [X509ExtendedTrustManager] to disable verification for a set of hosts.
+ *
+ * Note that the superclass [X509ExtendedTrustManager] isn't available on Android until version 7
+ * (API level 24).
+ */
+@IgnoreJRERequirement
+internal class InsecureExtendedTrustManager(
+  private val delegate: X509ExtendedTrustManager,
```
yschimke, line 34: Much simpler, you assume delegate must be X509ExtendedTrustManager and don't bother with most methods.

samples/guide/src/main/java/okhttp3/recipes/kt/DevServer.kt
```diff
@@ -0,0 +1,66 @@
...
+import okhttp3.mockwebserver.MockResponse
+import okhttp3.mockwebserver.MockWebServer
+import okhttp3.tls.HandshakeCertificates
+import okhttp3.tls.internal.TlsUtil
+
+class DevServer {
+  val handshakeCertificates = TlsUtil.localhost()
+
+  val server = MockWebServer().apply {
+    useHttps(handshakeCertificates.sslSocketFactory(), false)
```
yschimke, line 31: There is a rough edge here, if you use addInsecureHost and then use for a server or MockWebServer it will break in a confusing way.  Not a blocker, just raising.

android-test/src/androidTest/java/okhttp/android/test/InsecureForHostTest.kt
yschimke: I'd argue to keep these separate, land with or without this test, but not blocked on the other :)

yschimke:
Nice simplification.

okhttp-tls/src/main/kotlin/okhttp3/tls/internal/TlsUtil.kt
```diff
@@ -64,7 +67,13 @@ object TlsUtil {
       "Unexpected trust managers: ${result.contentToString()}"
     }
 
-    return result[0] as X509TrustManager
+    val trustManager = result[0] as X509TrustManager
+
+    return when {
+      insecureHosts.isEmpty() -> trustManager
+      AndroidPlatform.isAndroid -> InsecureAndroidTrustManager(trustManager, insecureHosts)
+      else -> InsecureExtendedTrustManager(trustManager as X509ExtendedTrustManager, insecureHosts)
```
yschimke, line 75: Hooking in here is smart! Less to worry about with the lifecycle of reusing clients via newBuilder etc etc.

samples/guide/src/main/java/okhttp3/recipes/kt/DevServer.kt
swankjesse: Agreed. I wanna do this in a follow-up because I think it might take some work to get right!

okhttp-tls/src/main/kotlin/okhttp3/tls/internal/InsecureExtendedTrustManager.kt
swankjesse: Yep. I took code from the DevServerJvm here!

samples/guide/src/main/java/okhttp3/recipes/kt/DevServer.kt
swankjesse: Ooooh interesting, yeah you’re right. Might be worth trying to make this work with client certificates.

okhttp-tls/src/main/kotlin/okhttp3/tls/internal/TlsUtil.kt
swankjesse: Yeah. I originally wanted this in OkHttpClient.Builder but I couldn’t find a nice way to change the behavior of the user’s supplied trust manager.

yschimke: Good call.

android-test/src/androidTest/java/okhttp/android/test/InsecureForHostTest.kt
yschimke: I regret these words... :)

PR #6042, opened 2020-05-10
Github issue template

yschimke:
When adding a test to reproduce an issue, I normally tack in onto CallTest because it has mockwebserver, client and platform integration hooks already setup.

Looking at how to simplify this for people reporting an issue.  e.g. Duplicate the `IssueTemplate` and show your failing test case.


yschimke:
Example event logging for failed test

https://13601-5152285-gh.circle-artifacts.com/0/%7E/reports/okhttp/build/reports/tests/test/classes/okhttp3.issues.Issue5819.html#rogueConnectionSeeder

yschimke:
Example single issue test case

https://github.com/square/okhttp/blob/0c78d8c6cd2db36609fb17dba22dc5052ec88bc7/okhttp/src/test/java/okhttp3/issues/Issue5819.kt

yschimke:
Next steps would be to add in addition to the gist, instructions for submitting a bug in this format.

swankjesse:
What's the goal here? I think if we want better bug reports this might create more friction, not less!

Our best bug reporters will already clone OkHttp and put the rest case in the right place. For example if they're reporting about web sockets, they'll write a web sockets test.

For everyone else I think we want something easy to paste and work in their existing projects. Minimal dependencies on test rules, and optional components is essential to reduce friction.

yschimke:
> What's the goal here?

I'm not attached to it, happy to close.  I wanted to raise the signal we get out of bug reproductions e.g. for a github issue, make it dead simple to create a repro (like the gist we already have) and then have that run against different JDK versions, and have logging output etc.

Closing for now, this isn't a hill for me to die on.

swankjesse:
Yeah agreed that better signal is good! But I think we get it with reduced friction?

Issue #6038, opened 2020-05-09
SSE EventSource 'text/event-stream' MIME type should be implicitly set


RalphSteinhagen:
Firstly, thanks for this great library!

I noticed a slight inconsistency of the SSE EventSource implementation where the `text/event-stream` MIME-type is being checked for the event receiving path 
https://github.com/square/okhttp/blob/e31cb03d86b9c4daf2124c6e62e218b32e72e056/okhttp-sse/src/main/kotlin/okhttp3/internal/sse/RealEventSource.kt#L58-L62
but not the initial request ('Accept' 'header-ist' of the EventSource) as done e.g. in browsers such as Firefox and Chrome etc.  A more detailed description can be found [here](https://html.spec.whatwg.org/multipage/server-sent-events.html).

As a workaround the following is possible:
```Java
OkHttpClient okClient = new OkHttpClient();
Request request = new Request.Builder().url(path).addHeader("Accept", "text/event-stream").build();

EventSource.Factory factory = EventSources.createFactory(okClient);
EventSource sseEventSource = factory.newEventSource(request, EVENT_SOURCE_LISTENER);
```

It would be nice if this could be kept consistent and the MIME-type `text/event-stream` be set implicitly by the EventSource constructor, in the `connect(...)' call, its Factory, or equivalently.

Many thanks in advance for your consideration.

P.S. out of curiosity: why is `EventSourceListener` an abstract class and not an interface.

yschimke:
We use an abstract class so we can add new events without breaking clients built against older versions. 

See a similar changes in EventListener https://github.com/square/okhttp/commit/8cebf9012bbdfdbad6b833effc545ff5b8b8e48b#diff-5ca874acb455fb0ee586729d2fb6a89d

yschimke:
This seems like a sensible change if the Accept header isn't already set in the request.  It clearly shows the intention of the request.  There is some risk for setting it when older clients were not, but it likely shows a flaw in the server implementation and the workaround would be easy.

RalphSteinhagen:
Thanks for the clarification and your consideration.

N.B. the origin for the "text/event-stream": our service can serve replies using different user-defined/preferred protocols for 'GET' (and other) calls for each of its tree-like structured endpoints. Depending on the 'Accept' header the reply is served either as an html (e.g. for web-browsers), binary, json, xml, yaml, ..., via long-polling and/or SSE notifications on content change followed by a user-driven pull using one of the previously mentioned formats.

[...]

PR #6044, opened 2020-05-11
Set implicit Accept for EventSource

yschimke:
Low urgency, but easy enough to implement or close won't fix.

Addresses https://github.com/square/okhttp/issues/6038

### DIFF
--- a/okhttp-sse/src/test/java/okhttp3/internal/sse/EventSourceHttpTest.java
+++ b/okhttp-sse/src/test/java/okhttp3/internal/sse/EventSourceHttpTest.java
@@ -16,6 +16,7 @@
 package okhttp3.internal.sse;
 
 import java.util.concurrent.TimeUnit;
+import javax.annotation.Nullable;
 import okhttp3.OkHttpClient;
 import okhttp3.OkHttpClientTestRule;
 import okhttp3.Request;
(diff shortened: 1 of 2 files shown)


okhttp-sse/src/test/java/okhttp3/internal/sse/EventSourceHttpTest.java
```diff
@@ -108,9 +108,49 @@
...
 
+  @Test public void retainsAccept() {
+    server.enqueue(new MockResponse().setBody(""
+        + "data: hey\n"
+        + "\n").setHeader("content-type", "text/event-stream"));
+
+    EventSource source = newEventSource("text/plain");
+
+    assertThat(source.request().url().encodedPath()).isEqualTo("/");
+    assertThat(source.request().header("Accept")).isEqualTo("text/plain");
```
swankjesse, line 119: alternately you can use RecordedRequest to assert the header made it all the way to the server?

okhttp-sse/src/test/java/okhttp3/internal/sse/EventSourceHttpTest.java
```diff
@@ -108,9 +108,49 @@
...
+    listener.assertClose();
+  }
+
   private EventSource newEventSource() {
-    Request request = new Request.Builder()
-        .url(server.url("/"))
+    return newEventSource(null);
+  }
+
+  private EventSource newEventSource(String accept) {
```
swankjesse, line 145: nit: `@Nullable`

Issue #6033, opened 2020-05-08
Flaky test report


yschimke:
From last 100 failed builds, filtered to master branch 

```
1	okhttp3.DuplexTest.duplexWithAuthChallenge
1	okhttp3.EventListenerTest.authenticatingTunnelProxyConnect
1	okhttp3.EventListenerTest.failedDribbledCallEventSequence
1	okhttp3.EventListenerTest.requestBodyFailHttp2OverHttps
1	okhttp3.EventListenerTest.secureConnectWithTunnel
1	okhttp3.EventListenerTest.successfulSecureConnect
1	okhttp3.EventListenerTest.timeToFirstByteHttp2OverHttps
1	okhttp3.WholeOperationTimeoutTest.timeoutReadingResponse
1	okhttp3.internal.concurrent.TaskRunnerTest.cancelWhileExecutingDoesNotStopUncancelableTask
10	okhttp3.internal.http.CancelTest.cancelAndFollowup[(CANCEL, H2)]
4	okhttp3.internal.http.CancelTest.cancelAndFollowup[(CANCEL, HTTP)]
3	okhttp3.internal.http.CancelTest.cancelAndFollowup[(CANCEL, HTTPS)]
2	okhttp3.internal.http2.Http2ConnectionTest.readTimesOut
1	okhttp3.internal.http2.HttpOverHttp2Test.connectionNotReusedAfterShutdown[h2]
1	okhttp3.internal.http2.HttpOverHttp2Test.connectionWindowUpdateAfterCanceling[h2_prior_knowledge]
1	okhttp3.internal.http2.HttpOverHttp2Test.connectionWindowUpdateOnClose[h2_prior_knowledge]
3	okhttp3.internal.http2.HttpOverHttp2Test.responseHeadersAfterGoaway[h2_prior_knowledge]
2	okhttp3.internal.tls.CertificatePinnerChainValidationTest.pinRootNotPresentInChain
1	okhttp3.internal.ws.RealWebSocketTest.pingOnInterval
1	okhttp3.internal.ws.RealWebSocketTest.unacknowledgedPingFailsConnection
2	okhttp3.internal.ws.RealWebSocketTest.unexpectedPongsDoNotInterfereWithFailureDetection
2	okhttp3.internal.ws.WebSocketHttpTest.clientMaxWindowBitsIncluded
3	okhttp3.internal.ws.WebSocketHttpTest.closeReasonMaximumLength
[...]

PR #6045, opened 2020-05-11
Attempt to minimise WebSocket test flakiness

yschimke:
https://github.com/square/okhttp/issues/6033

Assuming this is due to race conditions immediately after shutdown with websocket, allow for 500ms extra before failing.  Will record which tests see delays, so we can query for this.

### DIFF
--- a/okhttp-testing-support/src/main/kotlin/okhttp3/ClientRuleEventListener.kt
+++ b/okhttp-testing-support/src/main/kotlin/okhttp3/ClientRuleEventListener.kt
@@ -266,7 +266,12 @@ class ClientRuleEventListener(
   }
 
   private fun logWithTime(message: String) {
-    val timeMs = TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - startNs)
+    val timeMs = if (startNs == 0L) {
+      0L
+    } else {
+      TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - startNs)
+    }
+
     logger.invoke("[$timeMs ms] $message")
   }
 }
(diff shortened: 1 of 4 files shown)


okhttp-testing-support/src/main/kotlin/okhttp3/ClientRuleEventListener.kt
```diff
@@ -266,7 +266,12 @@ class ClientRuleEventListener(
   }
 
   private fun logWithTime(message: String) {
-    val timeMs = TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - startNs)
+    val timeMs = if (startNs == 0L) {
```
swankjesse, line 269: What’s going on here? Note that 0L is a valid timestamp!

okhttp/src/test/java/okhttp3/EventListenerTest.java
```diff
@@ -115,6 +114,15 @@
     if (cache != null) {
       cache.delete();
     }
+
+    client.connectionPool().evictAll();
+    if (client.connectionPool().connectionCount() > 0) {
```
swankjesse, line 119: Move this to `OkHttpClientTestRule.ensureAllConnectionsReleased()`

swankjesse: ... actually it has an assertion for this. Maybe replace that assertion with this?

okhttp/src/test/java/okhttp3/internal/ws/WebSocketHttpTest.java
```diff
@@ -100,8 +100,17 @@ private OkHttpClientTestRule configureClientTestRule() {
     platform.assumeNotBouncyCastle();
   }
 
-  @After public void tearDown() {
+  @After public void tearDown() throws InterruptedException {
     clientListener.assertExhausted();
+
+    client.connectionPool().evictAll();
```
swankjesse, line 106: Same, this should be in OkHttpClientTestRule

okhttp-testing-support/src/main/kotlin/okhttp3/ClientRuleEventListener.kt
yschimke: Early cancel.  I've updated and commented, preferring a masking variable that is immutable to startNs?.run or similar.

Issue #5705, opened 2020-01-06
RealWebSocket send throws writerTask NPE


Naitbit:
okhttp 4.3.0 throws NPE each time I try to send command via websocket.
`writerTask` is null in `RealWebSocket::runWriter`
```kotlin
  private fun runWriter() {
    this.assertThreadHoldsLock()

    taskQueue.schedule(writerTask!!) // crashes in this line
  }
```
Stacktrace:
```
Caused by kotlin.KotlinNullPointerException
       at okhttp3.internal.ws.RealWebSocket.runWriter + 398(RealWebSocket.java:398)
       at okhttp3.internal.ws.RealWebSocket.send + 352(RealWebSocket.java:352)
       at okhttp3.internal.ws.RealWebSocket.send + 332(RealWebSocket.java:332)
```

test:
```kotlin
import okhttp3.OkHttpClient
import okhttp3.Request
import okhttp3.WebSocketListener
import okhttp3.mockwebserver.MockWebServer
import org.junit.Rule
import org.junit.Test


class OkHttpBug {
    @get:Rule
    val server = MockWebServer()

    @Test
    fun test() {
        val request = Request.Builder()
            .url(server.url("/path"))
            .header("User-Agent", "Test Case")
            .build()
        val client = OkHttpClient()
        val listener = object : WebSocketListener() {

        }
        val command = """{ "id": "file" }"""
        val webSocket = client.newWebSocket(request, listener)
        webSocket.send(command)
        webSocket.close(1000, "RealWebSocket crashes with runWriter writerTask!! NPE")
    }
}
```

This seems to be 100% reproducible in 4.3.0. Had about 1000 crashes on that version on small amount of sessions.
This (almost) does not happen with the same code on okhttp 4.2.2 (2 crashes in 1000x user count in much longer period)

swankjesse:
Will fix.

swankjesse:
Great bug report. Will try to get a release out tonight.
https://github.com/square/okhttp/pull/5707

swankjesse:
FYI, OkHttp 4.3.1 is released.

simtse:
Hmm @swankjesse we ran into this issue again and we're using OkHttp 4.5.0.
```
kotlin.KotlinNullPointerException
        at okhttp3.internal.ws.RealWebSocket.writePingFrame$okhttp(RealWebSocket:569)
[...]

PR #6047, opened 2020-05-12
Avoid any chance at NPE in RealWebSocket

yschimke:
Pretty sure that with correct locking that this *already* can't be null.  The syncrhonized memory barrier in writePingFrame, which happens via the task runner, should always be after the syncrhonized memory barrier in initReaderAndWriter.

But this subtle change should be a noop but delay the ping if we find ourselves somehow with null.

To address https://github.com/square/okhttp/issues/5705


yschimke:
@dave-r12 You are awesome at these puzzlers, riddle me this, how was this ever null?

PR #6055, opened 2020-05-15
Fix isAndroid check

yschimke:
Fix for OkHttp platform issues in AndroidStudio (e.g. when running as a plugin)

https://github.com/square/retrofit/issues/3389

### DIFF
--- a/okhttp/src/main/kotlin/okhttp3/internal/platform/Platform.kt
+++ b/okhttp/src/main/kotlin/okhttp3/internal/platform/Platform.kt
@@ -184,6 +184,8 @@ open class Platform {
     fun alpnProtocolNames(protocols: List<Protocol>) =
         protocols.filter { it != Protocol.HTTP_1_0 }.map { it.toString() }
 
+    val isAndroid = "Dalvik".equals(System.getProperty("java.vm.name"))
+
     private val isConscryptPreferred: Boolean
       get() {
         val preferredProvider = Security.getProviders()[0].name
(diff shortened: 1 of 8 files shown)


okhttp/src/test/java/okhttp3/internal/platform/PlatformTest.java
```diff
@@ -42,4 +42,12 @@ public static String getJvmSpecVersion() {
   public void testToStringIsClassname() {
     assertThat(new Platform().toString()).isEqualTo("Platform");
   }
+
+  @Test
+  public void testNotAndroid() {
+    platform.assumeNotAndroid();
+
+    // This is tautological so just confirms that it runs.
```
yschimke, line 50: https://13779-5152285-gh.circle-artifacts.com/0/%7E/reports/okhttp/build/reports/tests/test/classes/okhttp3.internal.platform.PlatformTest.html

okhttp/src/main/kotlin/okhttp3/internal/platform/Platform.kt
```diff
@@ -184,6 +184,8 @@ open class Platform {
     fun alpnProtocolNames(protocols: List<Protocol>) =
         protocols.filter { it != Protocol.HTTP_1_0 }.map { it.toString() }
 
+    val isAndroid = "Dalvik".equals(System.getProperty("java.vm.name"))
```
swankjesse, line 187: Any background on how you came to this one? I don’t love it but I can’t think of anything better

swankjesse: `==` instead of `.equals()` ?

yschimke: Took same approach as Retrofit fix. Cc @JakeWharton

JakeWharton: I had a change in OkHttp to implement this fix as well. Was waiting on the Retrofit PR review.

swankjesse: Merged!

JakeWharton: Thanks. Is this going to make it into 3.14.x? Retrofit still depends on that version 😶 .

swankjesse: Yep!

JakeWharton: 🙏

okhttp/src/main/kotlin/okhttp3/internal/platform/Platform.kt
```diff
@@ -204,56 +206,66 @@ open class Platform {
 
     /** Attempt to match the host runtime to a capable Platform implementation. */
     private fun findPlatform(): Platform {
-      val android10 = Android10Platform.buildIfSupported()?.also {
-        AndroidLog.enable()
-      }
+      // This explicit check avoids activating in Android Studio with Android specific classes
+      // available when running plugins inside the IDE.
+      if (isAndroid) {
```
JakeWharton, line 211: I think the body of this can be simplified to
```
if (isAndroid) {
  AndroidLog.enable()
  return Android10Platform.buildIfSupported() ?: AndroidPlatform.buildIfSupported()!!
}
```
Do we ever expect the base Android platform to fail if we detected that we're on it?

okhttp/src/main/kotlin/okhttp3/internal/platform/Platform.kt
yschimke: No offense intended, didn't see a PR in flight so thought it was easy enough to copy the same approach you used with Retrofit.

PR #6060, opened 2020-05-17
Wireshark Testing

yschimke:
Testing outputting to a key log for assist wireshark decoding of TLS.


yschimke:
I think (!) it's working.

wireshark filter = "http2"

image

yschimke:
Damn, just realised afterwards that Netty already did this

https://github.com/netty/netty/pull/8653

yschimke:
image

yschimke:
@swankjesse Is it worth landing this as a sample?

samples/guide/src/main/java/okhttp3/recipes/kt/WiresharkExample.kt
```diff
@@ -0,0 +1,246 @@
...
+ * filter.
+ */
+class WireSharkKeyLoggerListener : EventListener() {
+  var random: String? = null
+  lateinit var currentThread: Thread
+
+  val loggerHandler = object : Handler() {
+    override fun publish(record: LogRecord) {
+      // Try to avoid multi threading issues with concurrent requests
+      if (Thread.currentThread() != currentThread)
```
swankjesse, line 51: supernit: always braces when we wrap

(gotta lock down spotless)

samples/guide/src/main/java/okhttp3/recipes/kt/WiresharkExample.kt
```diff
@@ -0,0 +1,246 @@
...
+      // Plaintext before ENCRYPTION
+      // Plaintext after DECRYPTION
+      val message = record.message
+
+      val parameters = record.parameters
+
+      if (parameters != null && !message.startsWith("Raw") && !message.startsWith("Plaintext")) {
+//        p(record.message)
+//        p(record.parameters[0])
+        val param = parameters[0] as String
```
swankjesse, line 82: param –> parameter

(in my goal to reduce the total number of names wherever possible, param/parameters is nails on chalkboard)

samples/guide/src/main/java/okhttp3/recipes/kt/WiresharkExample.kt
```diff
@@ -0,0 +1,246 @@
...
+    call: Call,
+    connection: Connection
+  ) {
+    if (random != null) {
+      val sslSocket = connection.socket() as SSLSocket
+      val session = sslSocket.session
+
+      val masterSecretHex = session.masterSecret?.encoded?.toByteString()
+          ?.hex()
+//          p(masterSecretHex)
```
swankjesse, line 134: what’s this?

samples/guide/src/main/java/okhttp3/recipes/kt/WiresharkExample.kt
```diff
@@ -0,0 +1,246 @@
...
+}
+
+fun main() {
+  println("TLSv1.2 traffic will be logged automatically")
+  println("Follow instructions at https://github.com/neykov/extract-tls-secrets for TLSv1.3")
+  println("Pid: ${ProcessHandle.current().pid()}")
+
+  Thread.sleep(10000)
+
+  WireSharkKeyLoggerListener.register()
```
swankjesse, line 242: Should the event listener do this automatically? I want to paste the whole thing somewhere, right?

samples/guide/src/main/java/okhttp3/recipes/kt/WiresharkExample.kt
```diff
@@ -0,0 +1,246 @@
...
+        }
+  }
+}
+
+fun main() {
+  println("TLSv1.2 traffic will be logged automatically")
+  println("Follow instructions at https://github.com/neykov/extract-tls-secrets for TLSv1.3")
+  println("Pid: ${ProcessHandle.current().pid()}")
+
+  Thread.sleep(10000)
```
swankjesse, line 240: what’s this for?

swankjesse: (Add a comment)

samples/guide/src/main/java/okhttp3/recipes/kt/WiresharkExample.kt
```diff
@@ -0,0 +1,246 @@
...
+          }
+    }
+  }
+}
+
+class WiresharkExample {
+  val connectionSpec =
+    Builder(ConnectionSpec.RESTRICTED_TLS)
+        .tlsVersions(TLS_1_2)
+//        .cipherSuites(CipherSuite.TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256)
```
swankjesse, line 181: ?

samples/guide/src/main/java/okhttp3/recipes/kt/WiresharkExample.kt
```diff
@@ -0,0 +1,246 @@
...
+        .build()
+
+  var bootstrapClient = OkHttpClient.Builder()
+      .connectionSpecs(listOf(connectionSpec))
+      .addInterceptor(BrotliInterceptor)
+      .eventListenerFactory(WireSharkKeyLoggerListener.Factory())
+      .build()
+
+  val dns = DnsOverHttps.Builder().client(bootstrapClient)
+    .url("https://1.1.1.1/dns-query".toHttpUrl())
```
swankjesse, line 191: supernit: whitespace is wrong

samples/guide/src/main/java/okhttp3/recipes/kt/WiresharkExample.kt
```diff
@@ -0,0 +1,246 @@
...
+import okhttp3.HttpUrl.Companion.toHttpUrl
+import okhttp3.OkHttpClient
+import okhttp3.Request
+import okhttp3.TlsVersion.TLS_1_2
+import okhttp3.brotli.BrotliInterceptor
+import okhttp3.dnsoverhttps.DnsOverHttps
+import okio.ByteString.Companion.toByteString
+
+/**
+ * Logs SSL keys to /tmp/key.log, allowing Wireshark to decode traffic and be examined with http2
```
swankjesse, line 41: Could you write just a bit about what this does and how it works here? The blog post is good but it’s a random blog post and who knows how long it’ll be around

swankjesse:
This is very neat stuff. It’s unfortunate we have to jump through so many hoops to do it.

samples/guide/src/main/java/okhttp3/recipes/kt/WiresharkExample.kt
yschimke: I've linked to my Stackoverflow question and partial answer which has screenshots.

samples/guide/src/main/java/okhttp3/recipes/kt/WiresharkExample.kt
yschimke: It needs to be done ASAP, ideally on the command line via -Djavax.net.debug=

yschimke:
> This is very neat stuff. It’s unfortunate we have to jump through so many hoops to do it.

Agreed, Conscrypt is the right place to do this.  https://github.com/google/conscrypt/issues/847

yschimke:
> This is very neat stuff. It’s unfortunate we have to jump through so many hoops to do it.

Also, I think this example makes using Wireshark really simple and compelling for debugging HTTP/2 errors.  It makes me care a lot less about improving the internal frame logging since Wireshark is obviously the right tool and this works with certificate pinning, with a public webserver, and without MITM proxies.

samples/guide/src/main/java/okhttp3/recipes/kt/WiresharkExample.kt
yschimke: This was because I believe the choice of non ECDHE cipher might not work.  But I haven't tested.  And I think the widely used ciphers will work.

yschimke:
@swankjesse ok to land?

samples/guide/src/main/java/okhttp3/recipes/kt/WiresharkExample.kt
```diff
@@ -0,0 +1,287 @@
...
+import okhttp3.TlsVersion.TLS_1_3
+import okhttp3.brotli.BrotliInterceptor
+import okhttp3.dnsoverhttps.DnsOverHttps
+import okhttp3.internal.SuppressSignatureCheck
+import okio.ByteString.Companion.toByteString
+
+/**
+ * Logs SSL keys to a log file, allowing Wireshark to decode traffic and be examined with http2
+ * filter. The approach is to hook into JSSE log events for the messages between client and server
+ * during handshake, and then take the agreed masterSecret from private fields of the session.
```
swankjesse, line 47: this doc is 👌

samples/guide/src/main/java/okhttp3/recipes/kt/WiresharkExample.kt
```diff
@@ -0,0 +1,287 @@
...
+    val matchResult = randomRegex.find(param)
+
+    return if (matchResult != null) {
+      matchResult.groupValues[1].replace(" ", "")
+    } else {
+      null
+    }
+  }
+
+  override fun secureConnectStart(call: Call) {
```
swankjesse, line 124: This makes OkHttp event listeners look super good

samples/guide/src/main/java/okhttp3/recipes/kt/WiresharkExample.kt
```diff
@@ -0,0 +1,287 @@
...
+    }
+  }
+}
+
+val eventListenerFactory = WireSharkKeyLoggerListener.Factory(File("/tmp/key.log"), verbose = true)
+
+@SuppressSignatureCheck
+class WiresharkExample(private val tlsVersions: List<TlsVersion>) {
+  val connectionSpec =
+    Builder(ConnectionSpec.RESTRICTED_TLS)
```
swankjesse, line 203: supernit: Builder -> ConnectionSpec.Builder

samples/guide/src/main/java/okhttp3/recipes/kt/WiresharkExample.kt
```diff
@@ -0,0 +1,287 @@
...
+        .tlsVersions(*tlsVersions.toTypedArray())
+        .build()
+
+  var bootstrapClient = OkHttpClient.Builder()
+      .connectionSpecs(listOf(connectionSpec))
+      .addInterceptor(BrotliInterceptor)
```
swankjesse, line 209: distraction?

samples/guide/src/main/java/okhttp3/recipes/kt/WiresharkExample.kt
```diff
@@ -0,0 +1,287 @@
...
+      .eventListenerFactory(eventListenerFactory)
+      .build()
+
+  val dns = DnsOverHttps.Builder()
```
swankjesse, line 213: strictly necessary? I think this is also a distraction. Also discourages use for private networks.

samples/guide/src/main/java/okhttp3/recipes/kt/WiresharkExample.kt
```diff
@@ -0,0 +1,287 @@
...
+      .build()
+
+  fun run() {
+    if (tlsVersions.contains(TLS_1_2)) {
+      println("TLSv1.2 traffic will be logged automatically")
+    }
+
+    if (tlsVersions.contains(TLS_1_3)) {
+      println("TLSv1.3 requires an external command run before first traffic is sent")
+      println("Follow instructions at https://github.com/neykov/extract-tls-secrets for TLSv1.3")
```
swankjesse, line 233: The agent is quite interesting. I wonder if we could detect it automatically so you don’t need to do all the work to extract the PID and log

samples/guide/src/main/java/okhttp3/recipes/kt/WiresharkExample.kt
```diff
@@ -0,0 +1,287 @@
...
+      println(
+          "Pid: ${ProcessHandle.current()
```
swankjesse, line 235: This is really nice

samples/guide/src/main/java/okhttp3/recipes/kt/WiresharkExample.kt
yschimke: removed.

samples/guide/src/main/java/okhttp3/recipes/kt/WiresharkExample.kt
yschimke: The agent is awesome - nice work @neykov

If you are TLSv1.3 - the agent is as automatic as you need since you can just run remotely against any running Java app.  OkHttp or otherwise.

But this sample is mainly to make it simple without running additional steps internally.  Not sure how to incrementally improve for now.  Let's try this out ourselves debugging failures against QNAP or Go based servers :)

samples/guide/src/main/java/okhttp3/recipes/kt/WiresharkExample.kt
yschimke: We could probably improve with a command line

```
$ wireshark -o tls.keylog_file:/tmp/key.log -Y http2 -k
```

mrprona92:
@yschimke im working on Android emulator. How i can do the same thing to export key like that.

yschimke:
@mrprona92 I don't think you can. This uses non public or stable features of specific JVMs, that don't exist on Android AFAICT.

mrprona92:
@yschimke thanks for your response. any suggest for me to do this by another way? i using okhttp with ssl handshake, If i export an keylog file from wireshark. Then read it from android side. is that avaiable?

yschimke:
Use some sort of MITM proxy? Sorry, not sure.  Worth asking in a wider forum.

Issue #6062, opened 2020-05-18
InsecureAndroidTrustManager checkServerTrusted fails when using addInsecureHost


marcorighini:
Android, trying to use `okhttp 4.7.0` `addInsecureHost`
```
// clientBuilder: OkHttpClient.Builder
// added some interceptors
...
val clientCertificates = HandshakeCertificates.Builder()
                    .addPlatformTrustedCertificates()
                    .addInsecureHost(proxyHost)
                    .build()
clientBuilder.proxy(Proxy(Proxy.Type.HTTP, InetSocketAddress.createUnresolved(proxyHost, proxyPort)))
clientBuilder.sslSocketFactory(clientCertificates.sslSocketFactory(), clientCertificates.trustManager)
clientBuilder.hostnameVerifier(HostnameVerifier { _, _ -> true })
...
```
but it fails with
```
     Caused by: java.security.cert.CertificateException: java.lang.RuntimeException: java.lang.IllegalArgumentException: Expected receiver of type android.security.net.config.RootTrustManager, but got okhttp3.tls.internal.InsecureAndroidTrustManager
        at com.google.android.gms.org.conscrypt.ConscryptFileDescriptorSocket.verifyCertificateChain(:com.google.android.gms@16089023@16.0.89 (040800-239467275):9)
        at com.google.android.gms.org.conscrypt.NativeCrypto.SSL_do_handshake(Native Method)
        at com.google.android.gms.org.conscrypt.NativeSsl.doHandshake(:com.google.android.gms@16089023@16.0.89 (040800-239467275):7)
        at com.google.android.gms.org.conscrypt.ConscryptFileDescriptorSocket.startHandshake(:com.google.android.gms@16089023@16.0.89 (040800-239467275):14)
        	... 58 more
     Caused by: java.lang.RuntimeException: java.lang.IllegalArgumentException: Expected receiver of type android.security.net.config.RootTrustManager, but got okhttp3.tls.internal.InsecureAndroidTrustManager
        at com.google.android.gms.org.conscrypt.Platform.checkTrusted(:com.google.android.gms@16089023@16.0.89 (040800-239467275):3)
        at com.google.android.gms.org.conscrypt.Platform.checkServerTrusted(:com.google.android.gms@16089023@16.0.89 (040800-239467275):3)
        at com.google.android.gms.org.conscrypt.ConscryptFileDescriptorSocket.verifyCertificateChain(:com.google.android.gms@16089023@16.0.89 (040800-239467275):12)
[...]

PR #6063, opened 2020-05-18
Fix for insecureHost on Android

yschimke:
As raised in https://github.com/square/okhttp/issues/6062 and fix suggested by @mattinger, we had a typo in the Android implementation.

### DIFF
--- a/android-test/src/androidTest/java/okhttp/android/test/OkHttpTest.kt
+++ b/android-test/src/androidTest/java/okhttp/android/test/OkHttpTest.kt
@@ -301,6 +301,8 @@ class OkHttpTest {
       assertEquals(200, response.code)
       assertTrue(socketClass?.startsWith("com.android.org.conscrypt.") == true)
     }
+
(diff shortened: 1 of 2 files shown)


android-test/src/androidTest/java/okhttp/android/test/OkHttpTest.kt
```diff
@@ -217,12 +247,27 @@ class OkHttpTest {
...
+  private fun localhostInsecureRequest() {
+    enableTls()
+
+    server.enqueue(MockResponse().setResponseCode(200))
+
+    val request = Request.Builder().url(server.url("/")).build()
+
+    client.newCall(request).execute().use {
+      assertEquals(200, it.code)
+      assertEquals(listOf("CN=${server.hostName}"), it.handshake?.peerCertificates?.map { (it as X509Certificate).subjectDN.name })
```
yschimke, line 267: I want to confirm why this isn't listOf(), which was what I would expect.

PR #6064, opened 2020-05-18
Update README.md

Ritik2604:
Just correcting some errors


swankjesse:
I don't think this is an improvement.

PR #6066, opened 2020-05-19
Another layer of indirection to defer Conscrypt init

yschimke:
Avoid https://github.com/square/okhttp/issues/5760, by deferring work from building the platform until when it first encounters an unknown socket type.

This is high priority because the log warning causes a lot of bad stackoverflow advice to add conscrypt dependencies, which is not the intention. https://stackoverflow.com/search?q=org.conscrypt.ConscryptHostnameVerifier

### DIFF
--- a/okhttp/src/main/kotlin/okhttp3/internal/platform/android/AndroidSocketAdapter.kt
+++ b/okhttp/src/main/kotlin/okhttp3/internal/platform/android/AndroidSocketAdapter.kt
@@ -94,6 +94,9 @@ open class AndroidSocketAdapter(private val sslSocketClass: Class<in SSLSocket>)
   }
 
   companion object {
+    val playProviderFactory: DeferredSocketAdapter.Factory =
+      factory("com.google.android.gms.org.conscrypt")
+
     /**
      * Builds a SocketAdapter from an observed implementation class, by grabbing the Class
      * reference to perform reflection on at runtime.
--- a/okhttp/src/main/kotlin/okhttp3/internal/platform/android/AndroidSocketAdapter.kt
+++ b/okhttp/src/main/kotlin/okhttp3/internal/platform/android/AndroidSocketAdapter.kt
@@ -25,6 +25,10 @@ import okhttp3.internal.platform.Platform
 
 /**
  * Modern reflection based SocketAdapter for Conscrypt class SSLSockets.
+ *
+ * This is used directly for providers where class name is known e.g. the Google Play Provider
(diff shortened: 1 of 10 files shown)


okhttp/src/main/kotlin/okhttp3/internal/platform/android/AndroidSocketAdapter.kt
```diff
@@ -97,11 +95,51 @@ open class AndroidSocketAdapter(private val sslSocketClass: Class<in SSLSocket>)
...
+        buildIfSupported(sslSocketClass)
       } catch (e: Exception) {
-        Platform.get().log(level = Platform.WARN, message = "unable to load android socket classes", t = e)
+        Platform.get()
+            .log(level = Platform.WARN, message = "unable to load android socket classes", t = e)
         null
       }
     }
+
+    fun buildIfSupported(actualSSLSocketClass: Class<in SSLSocket>): SocketAdapter? {
```
swankjesse, line 106: Docs on what the expected inputs are please

okhttp/src/main/kotlin/okhttp3/internal/platform/android/DeferredSocketAdapter.kt
```diff
@@ -16,25 +16,25 @@
...
+class DeferredSocketAdapter(private val socketAdapterFactory: Factory) : SocketAdapter {
   private var delegate: SocketAdapter? = null
 
   override fun isSupported(): Boolean {
     return true
   }
 
   override fun matchesSocket(sslSocket: SSLSocket): Boolean {
-    return sslSocket.javaClass.name.startsWith(socketPackage)
+    if (socketAdapterFactory.matchesSocket(sslSocket)) {
```
swankjesse, line 33: return socketAdapterFactory.matchesSocket(sslSocket)

yschimke:
@swankjesse I'll double back and check these again to be sure with real devices and simulators.

yschimke:
@swankjesse I attempted to squash merge this and it gave me an error with a Web button to retry, that however merged without squashing, I presume a github bug.  But it's exposed some of my transient branch commits onto master.  Sorry.  Check if you want to revert before the next commit and I'll reapply.

I'm not doing that because I don't want to rewrite history etc with a force push.

swankjesse:
I cleaned up our Git history. The updated merge commit is here:
d8f02f4efe2ee98aad80767a6ff4f607784ce831

Issue #6069, opened 2020-05-19
IllegalArgumentException: Not a Conscrypt trust manager: android.security.net.config.RootTrustManager


Strubbl:
since release 4.7.0 and also for 4.7.1 i get the following exception when using the OkHttpClient.

```
2020-05-19 13:15:39.061 13971-14060/fr.gaulupeau.apps.InThePoche.debug E/ArticleUpdateWorker: updateArticles() exception
    java.lang.IllegalArgumentException: Not a Conscrypt trust manager: android.security.net.config.RootTrustManager
        at org.conscrypt.Conscrypt.toConscrypt(Conscrypt.java:736)
        at org.conscrypt.Conscrypt.setHostnameVerifier(Conscrypt.java:769)
        at okhttp3.internal.platform.ConscryptPlatform.platformTrustManager(ConscryptPlatform.kt:53)
        at okhttp3.OkHttpClient.<init>(OkHttpClient.kt:237)
        at okhttp3.OkHttpClient$Builder.build(OkHttpClient.kt:1069)
        at fr.gaulupeau.apps.Poche.network.WallabagConnection.createClient(WallabagConnection.java:123)
        at fr.gaulupeau.apps.Poche.network.WallabagConnection.createWallabagService(WallabagConnection.java:107)
        at fr.gaulupeau.apps.Poche.network.WallabagConnection.getWallabagService(WallabagConnection.java:38)
        at fr.gaulupeau.apps.Poche.service.workers.BaseNetworkWorker.getWallabagService(BaseNetworkWorker.java:86)
        at fr.gaulupeau.apps.Poche.service.workers.ArticleUpdateWorker.updateArticles(ArticleUpdateWorker.java:76)
        at fr.gaulupeau.apps.Poche.service.workers.ArticleUpdateWorker.update(ArticleUpdateWorker.java:37)
        at fr.gaulupeau.apps.Poche.service.tasks.UpdateArticlesTask.run(UpdateArticlesTask.java:17)
        at fr.gaulupeau.apps.Poche.service.tasks.ActionRequestTask.run(ActionRequestTask.java:25)
        at fr.gaulupeau.apps.Poche.service.-$$Lambda$Eg7UQkyPgCNRHvigqxzb-ridugA.run(Unknown Source:2)
        at fr.gaulupeau.apps.Poche.service.TaskService.run(TaskService.java:158)
[...]

PR #6073, opened 2020-05-20
Fix Android Platform init

yschimke:
The Platform Companion ordering means that a Boolean field and the Platform instance field can race.


yschimke:
Fixes https://github.com/square/okhttp/issues/6069

yschimke:
The tests hide this because of specific ordering

image

PR #6083, opened 2020-05-23
New organization for URLConnectionTest test case

klebertarcisio:
Hi everyone,

I am proposing a new organization for the URLConnectionTest class. I created 7 new classes with the following test cases:

URLConnectionProxyTest
connectVia**Proxy**UsingProxyArg
connectVia**Proxy**UsingProxySystemProperty
connectVia**Proxy**UsingHttpProxySystemProperty
connectViaHttp**Proxy**ToHttpsUsingProxyArgWithNoProxy
connectViaHttp**Proxy**ToHttpsUsingHttpProxySystemProperty
connectViaHttp**Proxy**ToHttpsUsingProxyArg
connectViaHttp**Proxy**ToHttpsUsingProxySystemProperty
connectViaHttp**Proxy**ToHttpsUsingHttpsProxySystemProperty
redirectWith**Proxy**Selector
response305Use**Proxy**

### URLConnectionStreamTest
flushAfter**Stream**TransmittedWithChunkedEncoding
flushAfter**Stream**TransmittedWithFixedLength
flushAfter**Stream**TransmittedWithNoLengthHeaders
input**Stream**AvailableWithChunkedEncoding
input**Stream**AvailableWithContentLengthHeader
input**Stream**AvailableWithNoLengthHeaders

### URLConnectionUploadTest
chunked**Upload**_byteByByte
chunked**Upload**_smallBuffers
chunked**Upload**_largeBuffers
fixedLength**Upload**_byteByByte
fixedLength**Upload**_smallBuffers
fixedLength**Upload**_largeBuffers

### URLGzipConnectionTest
**gzip**AndConnectionReuseWithFixedLength
**gzip**AndConnectionReuseWithChunkedEncoding
**gzip**AndConnectionReuseWithFixedLengthAndTls
**gzip**AndConnectionReuseWithChunkedEncodingAndTls

### URLResponseConnectionTest
**response**307WithGet
**response**307WithHead
**response**307WithOptions
**response**307WithPost
**response**308WithGet
**response**308WithHead
**response**308WithOptions
**response**308WithPost
**response**307WithPostReverted
**response**308WithPostReverted
**response**300MultipleChoiceWithPost
**response**301MovedPermanentlyWithPost
**response**302MovedTemporarilyWithPost
**response**303SeeOtherWithPost
postRedirectToGetWithChunkedRequest
postRedirectToGetWithStreamedRequest

**URLReusedConnectionTest**
**postFailsWith**BufferedRequestForSmallRequest
**postFailsWith**BufferedRequestForLargeRequest
**postFailsWith**ChunkedRequestForSmallRequest
**postFailsWith**ChunkedRequestForLargeRequest
**postFailsWith**FixedLengthRequestForSmallRequest
**postFailsWith**FixedLengthRequestForLargeRequest

### URLZeroLengthConnectionTest
**zeroLength**Post
**zeroLength**Post_HTTP_2
**zeroLength**Put
**zeroLength**Put_HTTP_2

### DIFF
--- a/okhttp/src/test/java/okhttp3/URLConnectionUploadTest.java
+++ b/okhttp/src/test/java/okhttp3/URLConnectionUploadTest.java
@@ -15,53 +15,17 @@
  */
 package okhttp3;
 
-import okhttp3.internal.Internal;
-import okhttp3.internal.RecordingAuthenticator;
-import okhttp3.internal.RecordingOkAuthenticator;
-import okhttp3.internal.platform.Platform;
 import okhttp3.mockwebserver.MockResponse;
 import okhttp3.mockwebserver.MockWebServer;
 import okhttp3.mockwebserver.RecordedRequest;
-import okhttp3.mockwebserver.SocketPolicy;
-import okhttp3.testing.Flaky;
-import okhttp3.testing.PlatformRule;
-import okhttp3.tls.HandshakeCertificates;
 import okio.*;
 import org.junit.*;
-import org.junit.rules.TemporaryFolder;
 
 import javax.annotation.Nullable;
-import javax.net.ServerSocketFactory;
-import javax.net.SocketFactory;
-import javax.net.ssl.*;
 import java.io.IOException;
-import java.io.InputStream;
-import java.net.Authenticator;
-import java.net.*;
-import java.security.KeyStore;
-import java.security.cert.CertificateException;
-import java.security.cert.X509Certificate;
-import java.time.Duration;
 import java.util.*;
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicReference;
-import java.util.zip.GZIPInputStream;
 
-import static java.net.HttpURLConnection.HTTP_MOVED_TEMP;
-import static java.nio.charset.StandardCharsets.UTF_8;
-import static java.util.Arrays.asList;
-import static java.util.Locale.US;
-import static java.util.concurrent.TimeUnit.MILLISECONDS;
-import static java.util.concurrent.TimeUnit.NANOSECONDS;
-import static okhttp3.internal.Internal.addHeaderLenient;
-import static okhttp3.internal.Util.immutableListOf;
-import static okhttp3.internal.Util.userAgent;
-import static okhttp3.internal.http.StatusLine.HTTP_PERM_REDIRECT;
-import static okhttp3.internal.http.StatusLine.HTTP_TEMP_REDIRECT;
-import static okhttp3.mockwebserver.SocketPolicy.*;
-import static okhttp3.tls.internal.TlsUtil.localhost;
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.junit.Assert.fail;
 
 public final class URLConnectionUploadTest {
 
--- a/okhttp/src/test/java/okhttp3/URLConnectionUploadTest.java
+++ b/okhttp/src/test/java/okhttp3/URLConnectionUploadTest.java
@@ -1,0 +1,146 @@
+/*
+ * Copyright (C) 2009 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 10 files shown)


okhttp/src/test/java/okhttp3/URLConnectionUploadTest.java
```diff
@@ -0,0 +1,146 @@
...
+import okhttp3.internal.RecordingOkAuthenticator;
+import okhttp3.internal.platform.Platform;
+import okhttp3.mockwebserver.MockResponse;
+import okhttp3.mockwebserver.MockWebServer;
+import okhttp3.mockwebserver.RecordedRequest;
+import okhttp3.mockwebserver.SocketPolicy;
+import okhttp3.testing.Flaky;
+import okhttp3.testing.PlatformRule;
+import okhttp3.tls.HandshakeCertificates;
+import okio.*;
```
yschimke, line 29: Can you run `./gradlew spotlessApply` to cleanup

okhttp/src/test/java/okhttp3/URLConnectionUploadTest.java
```diff
@@ -0,0 +1,146 @@
...
+
+  @Rule public final OkHttpClientTestRule clientTestRule = new OkHttpClientTestRule();
+  @Rule public final MockWebServer server = new MockWebServer();
+  private OkHttpClient client = clientTestRule.newClient();
+
+  @Before public void setUp() {
+    server.setProtocolNegotiationEnabled(false);
+  }
+
+  @Test public void chunkedUpload_byteByByte() throws Exception {
```
yschimke, line 76: Is it worth parameterizing this test?

yschimke:
No strong opinion from me, I'll defer to others on this.  But I'm curious what drove this change?

swankjesse:
Each test is usually the intersection of 2 or more features. Organizing by feature may make it harder to find tests, not easier. Some tests here do upload and connection reuse, for example.

I'll second Yuri's question: what problem are you trying to solve?

okhttp/src/test/java/okhttp3/URLConnectionUploadTest.java
klebertarcisio: Yes, you're right. I believe that in this case it is not necessary. I just corrected it. I also removed unnecessary imports. Please tell me if you have any more suggestions. I will be happy to help.

klebertarcisio:
Hi, thanks for the answers!

I used a tool to detect smells and this class was classified as **GodClass** because it has more than **500 lines of code** and has a lower **Tight Class Cohesion** than the average of the other classes in the system. 

_Do you think that 500 lines for OkHttp is a reasonable number?_ 

Given the motivations I mentioned, _do you think it is necessary to remove GodClass?_

I looked for similar tests to create new classes. _What would be a better strategy?_

okhttp/src/test/java/okhttp3/URLConnectionUploadTest.java
klebertarcisio: I removed unnecessary imports. Thanks.

yschimke:
@klebertarcisio I don't think overall applying your personal metrics of code quality to a project is a high value use of your time.  

IMHO - This one concern about whether a test class is a GodClass is probably most relevant if you can see that almost all PRs involve changing these tests also.  I think that's the concern about the GodClass in a normal system, everyone ends up changing the same files with risky merges etc, because everything is muddled up there. But I think for metrics driven development, we would value this sort of refactor more for non-test classes.  

CallTest and URLConnectionTest are probably the most likely files to change for any given PR.  But mostly I think we just add new test cases there. On further thought my bigger concern is that a primary use of these tests is as a regression suite. We would lose this history between branches.

Look at `git diff okhttp_3.14.x:okhttp-tests/src/test/java/okhttp3/URLConnectionTest.java master:okhttp/src/test/java/okhttp3/URLConnectionTest.java` 

So I'm going to shut down this avenue, unless another maintainer steps in to actively support the change.

klebertarcisio:
Hi @yschimke, I understood your point, thank you very much for your feedback.

PR #6086, opened 2020-05-24
[WIP] Loom and JDK 15 testing

yschimke:


yschimke:
Testing with 

```
$ ./gradlew test -Dtest.java.home=/Library/Java/JavaVirtualMachines/jdk-15-loom.jdk/Contents/Home
...
okhttp3.internal.connection.ConnectionPoolTest > interruptStopsThread FAILED
    java.lang.AssertionError:
    Expecting empty but was:<[Q10000]>
        at okhttp3.internal.connection.ConnectionPoolTest.interruptStopsThread(ConnectionPoolTest.java:204)

okhttp3.internal.ws.WebSocketHttpTest > wssScheme FAILED
    java.lang.AssertionError: Expected Open but was Failure[java.net.SocketTimeoutException: Read timed out]
        at okhttp3.internal.ws.WebSocketRecorder.assertOpen(WebSocketRecorder.java:169)
        at okhttp3.internal.ws.WebSocketHttpTest.websocketScheme(WebSocketHttpTest.java:917)
        at okhttp3.internal.ws.WebSocketHttpTest.wssScheme(WebSocketHttpTest.java:529)
```

okhttp/src/main/kotlin/okhttp3/internal/concurrent/LoomBackend.kt
```diff
@@ -0,0 +1,72 @@
...
+package okhttp3.internal.concurrent
+
+import java.lang.invoke.MethodHandle
+import java.lang.invoke.MethodHandles
+import java.lang.invoke.MethodType
+import okhttp3.internal.SuppressSignatureCheck
+import okhttp3.internal.notify
+
+@SuppressSignatureCheck
+class LoomBackend() : Backend {
```
yschimke, line 25: We may want to implement this closer to the RealBackend example using

Executors.newUnboundedVirtualThreadExecutor()

Or Thread.Builder?

yschimke:
The Loom stuff is going to take some getting used to.  This noddy experiment with an alternative TaskRunner.Backend is a good example.  The virtual threads don't appear to show up via jstack.  VirtualThread which extends Thread isn't a public class.  It has new states which are presumably interesting like pinning.

```
LoomBackend b = new LoomBackend();

b.execute(() -> {
  String response = example.run("https://httpbin.org/delay/20");
  System.out.println(response);
});
``` 

image

We also need to stop using wait...

> In addition, blocking in native code or attempting to obtain an unavailable monitor when entering synchronized or calling Object.wait, will also block the native carrier thread.

swankjesse:
I believe Loom will support `synchronized` at some point? In the interim we can use `ReentrantLock` from `java.util.concurrent` ?

yschimke:
I might start with HTTP/1.1 and test that it executes without pinning.

yschimke:
Not an active branch yet - working but I think proper support is probably more work on threading models etc.

PR #6089, opened 2020-05-25
Use finer-grained locks in the connection pool

swankjesse:
Previously we were locking the RealConnectionPool for all
connection-related state, including cancelation.

This changes the locks to be per-Connection, leaning heavily
on thread-confined fields in RealCall, Exchange, and
ExchangeFinder.

### DIFF
--- a/docs/concurrency.md
+++ b/docs/concurrency.md
@@ -69,30 +69,26 @@ This is necessary for bookkeeping when creating new streams. Correct framing req
 
 ## Connection Pool
 
-### Background
-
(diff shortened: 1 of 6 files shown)


swankjesse:
I’ve got a benchmark that keeps 128 connections in flight concurrently. It does ~10K concurrent calls on my MacBook Pro, while running OkHttp and MockWebServer side-by-side.

### BEFORE

Lots of contention on the shared RealConnectionPool.

Screen Shot 2020-05-24 at 10 23 28

### AFTER

Much less contention. But note that the throughput isn’t significantly better.

Screen Shot 2020-05-24 at 23 37 57

PR #6093, opened 2020-05-27
Tidy up the code that does connection releasing in RealCall

swankjesse:
Recent refactorings have made it clear that there's three things
we need to be done before we can release the connection:
 - the request body
 - the response body
 - the series of exchanges

This changes the code to invert these booleans. In a follow-up I'd
like to add some defensive code around duplex calls where the request
and response body could race when they close.

### DIFF
--- a/okhttp/src/main/kotlin/okhttp3/Cache.kt
+++ b/okhttp/src/main/kotlin/okhttp3/Cache.kt
@@ -295,9 +295,9 @@ class Cache internal constructor(
   @Throws(IOException::class)
   fun urls(): MutableIterator<String> {
     return object : MutableIterator<String> {
-      val delegate: MutableIterator<DiskLruCache.Snapshot> = cache.snapshots()
-      var nextUrl: String? = null
-      var canRemove = false
+      private val delegate: MutableIterator<DiskLruCache.Snapshot> = cache.snapshots()
+      private var nextUrl: String? = null
+      private var canRemove = false
 
       override fun hasNext(): Boolean {
         if (nextUrl != null) return true
--- a/okhttp/src/main/kotlin/okhttp3/internal/connection/RealCall.kt
+++ b/okhttp/src/main/kotlin/okhttp3/internal/connection/RealCall.kt
@@ -87,10 +87,10 @@ class RealCall(
 
   var connection: RealConnection? = null
     private set
-  private var exchangeRequestDone = false
-  private var exchangeResponseDone = false
+  private var requestBodyOpen = false
+  private var responseBodyOpen = false
   private var timeoutEarlyExit = false
-  private var noMoreExchanges = false
+  private var expectMoreExchanges = true
 
   /**
    * This is the same value as [exchange], but scoped to the execution of the network interceptors.
@@ -215,10 +215,11 @@ class RealCall(
    */
   fun enterNetworkInterceptorExchange(request: Request, newExchangeFinder: Boolean) {
     check(interceptorScopedExchange == null)
-    check(exchange == null) {
+    check(!responseBodyOpen) {
       "cannot make a new request because the previous response is still open: " +
           "please call response.close()"
     }
+    check(!requestBodyOpen)
 
     if (newExchangeFinder) {
       this.exchangeFinder = ExchangeFinder(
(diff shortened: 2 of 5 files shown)


okhttp/src/main/kotlin/okhttp3/Cache.kt
```diff
@@ -295,9 +295,9 @@ class Cache internal constructor(
   @Throws(IOException::class)
   fun urls(): MutableIterator<String> {
     return object : MutableIterator<String> {
-      val delegate: MutableIterator<DiskLruCache.Snapshot> = cache.snapshots()
-      var nextUrl: String? = null
-      var canRemove = false
+      private val delegate: MutableIterator<DiskLruCache.Snapshot> = cache.snapshots()
+      private var nextUrl: String? = null
+      private var canRemove = false
```
swankjesse, line 300: I ran spotbugs and it said these were generating unused accessors

okhttp/src/main/kotlin/okhttp3/internal/connection/RealCall.kt
```diff
@@ -87,10 +87,10 @@ class RealCall(
 
   var connection: RealConnection? = null
     private set
-  private var exchangeRequestDone = false
-  private var exchangeResponseDone = false
+  private var requestBodyOpen = false
+  private var responseBodyOpen = false
```
swankjesse, line 91: these booleans are inverted. I didn’t invert the initial values because it doesn’t matter until we initialize the first exchange.

okhttp/src/main/kotlin/okhttp3/internal/connection/RealCall.kt
```diff
@@ -232,16 +233,17 @@ class RealCall(
 
   /** Finds a new or pooled connection to carry a forthcoming request and response. */
   internal fun initExchange(chain: RealInterceptorChain): Exchange {
-    check(!noMoreExchanges) { "released" }
-    check(exchange == null)
+    check(expectMoreExchanges) { "released" }
+    check(!responseBodyOpen)
+    check(!requestBodyOpen)
```
swankjesse, line 238: these two values being false imply that the exchange is null, but this is more semantic

okhttp/src/main/kotlin/okhttp3/internal/connection/RealCall.kt
```diff
@@ -301,19 +302,14 @@ class RealCall(
    * If the call was canceled or timed out, this will wrap [e] in an exception that provides that
    * additional context. Otherwise [e] is returned as-is.
    */
-  private fun <E : IOException?> maybeReleaseConnection(e: E, incrementSuccesses: Boolean): E {
-    var result = e
+  private fun <E : IOException?> maybeReleaseConnection(e: E): E {
+    if (requestBodyOpen || responseBodyOpen || expectMoreExchanges) return e
```
swankjesse, line 306: correlates to the previous expression that was used twice below,

```
if (exchange == null && noMoreExchanges)
```

okhttp/src/main/kotlin/okhttp3/internal/connection/RealConnection.kt
```diff
@@ -705,43 +699,38 @@ class RealConnection(
...
-            }
-          }
-
-          e.errorCode == ErrorCode.CANCEL && call.isCanceled() -> {
-            // Permit any number of CANCEL errors on locally-canceled calls.
-          }
-
-          else -> {
-            // Everything else wants a fresh connection.
+  @Synchronized internal fun trackFailure(call: RealCall, e: IOException?) {
```
swankjesse, line 702: this diff reads better hiding whitespace changes. I’m just promoting `synchronized` to `@Synchronized`.

okhttp/src/main/kotlin/okhttp3/internal/connection/RealConnection.kt
```diff
@@ -146,25 +146,21 @@ class RealConnection(
    * Returns true if this is an HTTP/2 connection. Such connections can be used in multiple HTTP
    * requests simultaneously.
    */
-  val isMultiplexed: Boolean
+  internal val isMultiplexed: Boolean
     get() = http2Connection != null
 
   /** Prevent further exchanges from being created on this connection. */
-  fun noNewExchanges() {
-    assertThreadDoesntHoldLock()
```
yschimke, line 154: This is the one line that concerns me.  We don't use a static analysis check like GuardedBy, and these particular checks always gave me more confidence that the somewhat complicated and fluid concurrency dance we attempt doesn't have some potential race conditions or theoretical deadlocks.

swankjesse: The assertion is most useful in a few cases:
 - checking a different lock that it's acquiring (possible deadlock)
 - checking no lock before doing I/O

I wanna keep the check in these situations. The check is doing at runtime what I'd prefer a framework or type checker to do for us statically.

In these cases it's a pretty benign check: are we using reentrant capabilities of the lock. I find reentrant synchronization to be quite sloppy, and an invitation for bugs. But there's no particular danger in these changed cases; we’re not doing any deadlock-capable locking or I/O.

Your comment does suggest something we should fix though! We have a check for unexpected locks held in our EventListener. I should update that to check for locks on connections also!

Issue #6100, opened 2020-05-29
Chain.with*Timeout methods do not work with Network Interceptors


ryanhall07:
code to reproduce
```
override fun intercept(chain: Interceptor.Chain): okhttp3.Response {
  chain
    .withReadTimeout(1, TimeUnit.MILLISECONDS)
    .proceed(request)
}

```

exception
```
    java.lang.IllegalStateException: network interceptor misk.client.NetworkInterceptorWrapper@1429341e must call proceed() exactly once
    	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:104)
    	at io.opentracing.contrib.okhttp3.TracingCallFactory$NetworkInterceptor.intercept(TracingCallFactory.java:105)
    	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:100)
    	at okhttp3.internal.connection.ConnectInterceptor.intercept(ConnectInterceptor.kt:34)
    	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:100)
    	at okhttp3.internal.cache.CacheInterceptor.intercept(CacheInterceptor.kt:82)
    	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:100)
    	at okhttp3.internal.http.BridgeInterceptor.intercept(BridgeInterceptor.kt:83)
    	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:100)
    	at okhttp3.internal.http.RetryAndFollowUpInterceptor.intercept(RetryAndFollowUpInterceptor.kt:76)
    	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:100)
    	at datadog.trace.instrumentation.okhttp3.TracingInterceptor.intercept(TracingInterceptor.java:36)
    	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:100)
    	at io.opentracing.contrib.okhttp3.TracingCallFactory$1.intercept(TracingCallFactory.java:59)
    	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:100)
    	at okhttp3.internal.connection.RealCall.getResponseWithInterceptorChain$okhttp(RealCall.kt:197)
    	at okhttp3.internal.connection.RealCall$AsyncCall.run(RealCall.kt:502)
[...]

PR #6102, opened 2020-05-31
Clarify Interceptor timeout docs

yschimke:
Timeout mostly take effect before we have a connection, so rely on being adjusted in the application interceptors.

Instead of fixing the limitation highlighted here https://github.com/square/okhttp/issues/6100, simplify clarify that adjusting timeouts must be done in application interceptors.

An alternative implementation would simply ignore these updates, but that seems worse longer term for support requests "Why is my timeout ignored".  These are only used in ExchangeFinder before networkInterceptors fire.

### DIFF
--- a/docs/interceptors.md
+++ b/docs/interceptors.md
@@ -124,13 +124,14 @@ Each interceptor chain has relative merits.
  * Observe the application's original intent. Unconcerned with OkHttp-injected headers like `If-None-Match`.
  * Permitted to short-circuit and not call `Chain.proceed()`.
  * Permitted to retry and make multiple calls to `Chain.proceed()`.
+ * Can adjust Call timeouts using withConnectTimeout, withReadTimeout, withWriteTimeout.
 
 **Network Interceptors**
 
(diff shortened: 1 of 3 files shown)


yschimke:
TODO add unit tests after agree on behaviour

ryanhall07:
thanks! this would have been very helpful.

docs/interceptors.md
```diff
@@ -124,13 +124,14 @@ Each interceptor chain has relative merits.
...
  * Permitted to retry and make multiple calls to `Chain.proceed()`.
+ * Can adjust Call timeouts using withConnectTimeout, withReadTimeout, withWriteTimeout.
 
 **Network Interceptors**
 
  * Able to operate on intermediate responses like redirects and retries.
  * Not invoked for cached responses that short-circuit the network.
  * Observe the data just as it will be transmitted over the network.
- * Access to the `Connection` that carries the request.
+ * Access to the `Connection` that carries the request..
```
swankjesse, line 134: extra dot?

PR #6103, opened 2020-05-31
[WIP] Testing with multijdk docker image

yschimke:
Quick testing building a combined image

https://github.com/yschimke/cimg-openjdk/blob/master/multi/Dockerfile

Forked from https://github.com/CircleCI-Public/cimg-openjdk

Then always building with 11, and running with 8-242, 8-252, 11, and 14.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -61,7 +61,7 @@ commands:
 jobs:
   compile:
     docker:
-      - image: circleci/openjdk:11.0.7-jdk-buster
+      - image: yschimke/multijdk:latest
 
     environment:
       JVM_OPTS: -Xmx1g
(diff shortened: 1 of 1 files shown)


yschimke:
Full disclosure - I'm a docker amateur. But making a first attempt now that CircleCI published new images, that change some existing assumptions.

.circleci/config.yml
```diff
@@ -124,7 +124,7 @@ jobs:
 
   testjdk8:
     docker:
-      - image: circleci/openjdk:11.0.6-jdk-stretch
```
yschimke, line 127: This was previously left behind because we needed a stretch image which has both JDK 8 in base for running tests, and JDK 11 for. compile

yschimke:
TODO add android SDKs and some setup emulators and use as a single
TODO publish somewhere as a square/okhttp CI image that we could use in CircleCI or github actions.

yschimke:
We should revisit this in the scope of Github builds.

PR #6104, opened 2020-05-31
Avoid BouncyCastle failures

yschimke:
BouncyCastle isn't known to work with MockWebServer.

### DIFF
--- a/okhttp/src/test/java/okhttp3/internal/http/CancelTest.kt
+++ b/okhttp/src/test/java/okhttp3/internal/http/CancelTest.kt
@@ -95,6 +95,8 @@ class CancelTest(mode: Pair<CancelMode, ConnectionType>) {
       platform.assumeHttp2Support()
     }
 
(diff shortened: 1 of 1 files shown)


yschimke:
From https://app.circleci.com/pipelines/github/square/okhttp/3392/workflows/fb8e7f49-2fb8-4192-9224-6e9136ff4816/jobs/14416/steps

yschimke:
> That’s too bad. I wouldn’t have expected those features to interop so poorly.
> 
> Tracking bug upstream?

I closed it off with https://github.com/square/okhttp/issues/5698

There are lots of subtle individual differences causing failures.  So real, some just in terms of our test assumptions.

I made the call that a user trying to use MockWebServer shouldn't also enable bouncycastle. I think us working as a client with BC (which is generally the case) is higher priority than supporting the MWS code as well.

You could disable the assumption and go through and investigate and fix, but we haven't done that over past X years, feels like busy work.

PR #6107, opened 2020-06-01
Add hub4j/github-api to works_with_okhttp.md

bitwiseman:


swankjesse:
Ooooh good library to know about! But I’m gonna close this because I don’t think it’s tenable to include API clients in this list. This is intended to be a list of libraries that extend or complement OkHttp.

If I did API clients it’d be a very long list!

PR #6111, opened 2020-06-03
Allow for domain name trailing dot in PublicSuffixDatabase

yschimke:


okhttp/src/main/kotlin/okhttp3/internal/publicsuffix/PublicSuffixDatabase.kt
```diff
@@ -84,6 +83,19 @@ class PublicSuffixDatabase {
     return domain.split('.').asSequence().drop(firstLabelOffset).joinToString(".")
   }
 
+  private fun splitDomain(domain: String): List<String> {
+    // We use UTF-8 in the list so we need to convert to Unicode.
+    val unicodeDomain = IDN.toUnicode(domain)
+    val domainLabels = unicodeDomain.split('.')
+
+    if (domainLabels.lastOrNull() == "") {
```
swankjesse, line 91: There's no need for OrNull here; split always returns at least one string.

okhttp/src/test/java/okhttp3/HttpUrlTest.java
```diff
@@ -1756,6 +1756,9 @@ HttpUrl parse(String url) {
     assertThat(parse("https://xn--4pvxs.jp").topPrivateDomain()).isNull();
     assertThat(parse("https://localhost").topPrivateDomain()).isNull();
     assertThat(parse("https://127.0.0.1").topPrivateDomain()).isNull();
+
+    // https://github.com/square/okhttp/issues/6109
+    assertThat(parse("http://a./").topPrivateDomain()).isNull();
```
swankjesse, line 1761: Also test these?
```
a..
..a
a..b
.a
.
..

```

swankjesse:
Does this return the wrong answer for say, `https://squareup.com.`

yschimke:
I feel like this is a strong hint :)
I'll check. Thanks!

okhttp/src/test/java/okhttp3/HttpUrlTest.java
yschimke: My interpretation is that hostname ending with '.' is the one special case since it's actually the blessed form of dns, and we mostly just all agreed to drop that trailing '.'

Or is there some special behaviour we want for these?

yschimke:
> Does this return the wrong answer for say, `https://squareup.com.`

Ok, you win this round Mr Swank.

okhttp/src/test/java/okhttp3/HttpUrlTest.java
swankjesse: I really don’t know. I was mostly worried that we’d return `squareup.com` as the top private domain, which seemed wrong. But I’d forgotten about the trailing `.` thing in DNS.

Now I’m curious, What Does Guava Do?

yschimke: The awkward thing with this change is that HttpUrl parsing is our shield, but also known to be non exact.  And as much as it doesn't match URI, it will be painful and cause bugs in various places e.g. calling through to JDK methods like ProxySelector.

PR #6114, opened 2020-06-03
Dependency upgrades

yschimke:
Upgrade to latest release of Gradle and Kotlin


yschimke:
No urgency, but testing against latest versions since clients might use newer kotlin.

PR #6121, opened 2020-06-08
Add Monitoring Connections for Error Status Messages

### LINKED ISSUE #1001: Handling of early 403 FORBIDDEN with Connection: close whilst streaming
We are using okhttp against an HTTP server (Finagle) which will cut off the connection early if the client is not authorised for a particular resource.

```
HTTP/1.1 403 Forbidden\r\n
[Expert Info (Chat/Sequence): HTTP/1.1 403 Forbidden\r\n]
Request Version: HTTP/1.1
Status Code: 403
Response Phrase: Forbidden
Content-Length: 0\r\n
Connection: close\r\n
\r\n
```

The client does not handle this behaviour well, as in it escapes with

```
java.net.SocketException: Connection reset
    at java.net.SocketOutputStream.socketWrite(SocketOutputStream.java:118)
    at java.net.SocketOutputStream.write(SocketOutputStream.java:159)
    at okio.Okio$1.write(Okio.java:78)
    at okio.AsyncTimeout$1.write(AsyncTimeout.java:155)
    at okio.RealBufferedSink.emitCompleteSegments(RealBufferedSink.java:133)
    at okio.RealBufferedSink.write(RealBufferedSink.java:45)
    at com.squareup.okhttp.internal.http.HttpConnection$FixedLengthSink.write(HttpConnection.java:302)
    at okio.RealBufferedSink.emitCompleteSegments(RealBufferedSink.java:133)
    at okio.RealBufferedSink.writeAll(RealBufferedSink.java:83)
    at com.squareup.okhttp.internal.http.URLConnectionTest$2.writeTo(URLConnectionTest.java:2235)
    at com.squareup.okhttp.Call.getResponse(Call.java:202)
    at com.squareup.okhttp.Call.execute(Call.java:80) 
```

I'm not an expert on the [HTTP/1.1](http://www.ietf.org/rfc/rfc2616.txt) spec, but given from what I've read so far, I believe the server is in the right and I'd expect the client to parse the response headers and expose them. The sections I'm referring to are

> 8.1.2 Overall Operation
> ... Persistent connections provide a mechanism by which a client and a server can signal the close of a TCP connection. This signaling takes > place using the Connection header field (section 14.10). Once a close has been signaled, the client MUST NOT send any more requests on that connection.

and further

> 8.1.2.1 Negotiation
> ... If the server chooses to close the connection immediately after sending the
> response, it SHOULD send a Connection header including the connection-token close.
> An HTTP/1.1 client MAY expect a connection to remain open, but would decide to keep it open based on whether the response from a server contains a Connection header with the connection-token close.

I tried to recreate the problem in your test infrastructure, please see [URLConnectionTest#connectionCloseEarlyInResponseWhilstStreaming](https://gist.github.com/tweek-in-a-box/da91dd412b08e12985dc) (based on 7f763c1).

bessbd:
As per https://tools.ietf.org/html/rfc2616#section-8.2.2 An
HTTP/1.1 (or later) client sending a message-body SHOULD
monitor the network connection for an error status while it
is transmitting the request.

Currently, okhttp implements a send request, then read response
if/when request was sent successfully.
This hides early responses such as 413 Payload Too Large
errors from clients.

This commit fixes that by adding a response read in case of an
exception happening while the request is being sent.

This closes #1001 .

### DIFF
--- a/okhttp-tests/src/test/java/okhttp3/internal/http/DisconnectTest.java
+++ b/okhttp-tests/src/test/java/okhttp3/internal/http/DisconnectTest.java
@@ -19,15 +19,25 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.net.HttpURLConnection;
+import java.net.InetSocketAddress;
 import java.net.ServerSocket;
 import java.net.Socket;
+import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 import javax.net.ServerSocketFactory;
 import javax.net.SocketFactory;
+
+import com.sun.net.httpserver.HttpExchange;
+import com.sun.net.httpserver.HttpHandler;
+import com.sun.net.httpserver.HttpServer;
 import okhttp3.DelegatingServerSocketFactory;
 import okhttp3.DelegatingSocketFactory;
+import okhttp3.MediaType;
 import okhttp3.OkHttpClient;
 import okhttp3.OkUrlFactory;
+import okhttp3.Request;
+import okhttp3.RequestBody;
+import okhttp3.Response;
 import okhttp3.mockwebserver.MockResponse;
 import okhttp3.mockwebserver.MockWebServer;
 import okio.Buffer;
@@ -35,6 +45,7 @@ import org.junit.Before;
 import org.junit.Test;
 
 import static okhttp3.TestUtil.defaultClient;
+import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.fail;
 
 public final class DisconnectTest {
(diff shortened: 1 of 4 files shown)


yschimke:
We would need to start with a working version on the 4.x branch.  3.12.x is generally for security fixes, although this PR/fix would be a good candidate for 3.12.x.

Currently it looks like there are some test failures, so re-running to see if those are reliably failing.

```
[ERROR] Failures: 

[ERROR]   CallTest.failingCallsDoNotInterfereWithConnection:2667->access$200:100->makeFailingCall:3443 expected:<[write body fail!]> but was:<[timeout]>

[ERROR]   CallTest.postWithFileNotFound:3384 Expected exception type among [class java.io.FileNotFoundException], got java.net.SocketTimeoutException: timeout

[INFO] 
```

bessbd:
@swankjesse & @yschimke : the fix in this PR seems to break some existing tests and I can't tell what is the expected behavior for those cases. Can you please help me out on what I should change here so that CI passes?

bessbd:
@yschimke : thank you for the quick response!

> We would need to start with a working version on the 4.x branch

Would you be okay with helping me getting this merged and once it's done I'd port it to 4.x? We are currently using okhttp 3 so fixing that is a first priority to me. (But of course we'd want the fix in 4 too so that there won't be a regression upon migrating.)

>  it looks like there are some test failures

Yes, I've noticed the failing tests and they seem to be deterministic, but I can't tell what the expected behavior is there
* `postWithFileNotFound`: is it okay to just change the exception type there or am I breaking something that's intentionally like that?
* `failingCallsDoNotInterfereWithConnection`: same thing here. Do I need to change something in my "production" code or do I just need to adjust the test case?

yschimke:
@bessbd I don't think we would rush this into 3.12.x.  I think we want to get the behaviour right on master (4.x) and then backport it to 3.12.x, maybe 3.14.x.

If we rush something into 3.12.x that doesn't match intended 4.x behaviour then we will end up stuck.

The behaviour change looks sensible, but we have other related bugs with HTTP/2 or other scenarios.

If you need it sooner, I suggest forking 3.12.x and use a custom build temporarily until we can help you work through this PR.

swankjesse:
This is too big of a change to ever land in the 3.12.x branch. If you need it there I recommend shipping your own private build.

bessbd:
I'll start porting this change to 4.x then. Thank you for the feedback, @yschimke and @swankjesse

PR #6135, opened 2020-06-20
Early implementation of an ASN.1 DER decoder

swankjesse:
My goal is to finish the implementation to support an ASN.1 DER encoder
and then to change okhttp-tls to no longer depend on BouncyCastle.

### DIFF
--- a/okhttp-tls/src/main/kotlin/okhttp3/tls/internal/der/DerReader.kt
+++ b/okhttp-tls/src/main/kotlin/okhttp3/tls/internal/der/DerReader.kt
@@ -1,0 +1,301 @@
+/*
+ * Copyright (C) 2020 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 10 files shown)


okhttp-tls/src/main/kotlin/okhttp3/tls/internal/der/DerReader.kt
```diff
@@ -0,0 +1,301 @@
...
+    val tagAndClass = source.readByte().toInt() and 0xff
+    val tagClass = tagAndClass and 0b1100_0000
+    val constructed = (tagAndClass and 0b0010_0000) == 0b0010_0000
+    val tag0 = tagAndClass and 0b0001_1111
+    if (tag0 == 0b0001_1111) {
+      var tagBits = 0L
+      while (true) {
+        val tagN = source.readByte().toInt() and 0xff
+        tagBits += (tagN and 0b0111_1111)
+        if (tagN and 0b1000_0000 == 0b1000_0000) break
```
oldergod, line 130:
```suggestion
        if ((tagN and 0b1000_0000) == 0b1000_0000) break
```

okhttp-tls/src/main/kotlin/okhttp3/tls/internal/der/DerReader.kt
```diff
@@ -0,0 +1,301 @@
...
+      tag = tag0.toLong()
+    }
+
+    // Read the length.
+    val length: Long
+    val length0 = source.readByte().toInt() and 0xff
+    if (length0 == 0b1000_0000) {
+      // Indefinite length.
+      length = -1L
+    } else if (length0 and 0b1000_0000 == 0b1000_0000) {
```
oldergod, line 144:
```suggestion
    } else if ((length0 and 0b1000_0000) == 0b1000_0000) {
```

PR #6137, opened 2020-06-20
add Dependabot

sullis:
https://github.blog/2020-06-01-keep-all-your-packages-up-to-date-with-dependabot/

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+version: 2
+updates:
+  - package-ecosystem: "gradle"
+    directory: "/"
+    schedule:
+      interval: "daily"
+      time: "02:00"
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
+      time: "02:00"


swankjesse:
I don't expect this to be able to edit our Gradle config; it's pretty custom. Would be good to use something uniform if robots are doing the updates.

yschimke:
@sullis Is there example output, or what it would change?  Can you link to other high profile OSS libraries already adopting so we can see the cadence and information in a typical PR?

I have read the blog post.

yschimke:
Example from Apple servicetalk https://github.com/apple/servicetalk/pull/1089

micronaut https://github.com/micronaut-projects/micronaut-core/pull/3548/files

twilio-java https://github.com/twilio/twilio-java/pulls

image

cc @sullis Does this handle indirect version numbers?  I'm assuming no, like Jesse says.

JakeWharton:
Perhaps we should switch to wildcard dependencies (`+`) and use a lock file instead? I would assume it could update lock files (it does for Rust and Ruby).

PR #6139, opened 2020-06-22
DER encoder

swankjesse:
I haven't yet implemented certificate encoding

### DIFF
--- a/okhttp-tls/src/main/kotlin/okhttp3/tls/internal/der/Adapters.kt
+++ b/okhttp-tls/src/main/kotlin/okhttp3/tls/internal/der/Adapters.kt
@@ -20,6 +20,7 @@ import java.math.BigInteger
 import java.text.SimpleDateFormat
 import java.util.Date
 import java.util.TimeZone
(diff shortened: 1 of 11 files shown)


okhttp-tls/src/main/kotlin/okhttp3/tls/internal/der/Adapters.kt
```diff
@@ -134,6 +161,16 @@ internal object Adapters {
...
+    val dateFormat = SimpleDateFormat("yyMMddHHmmssXX").apply {
+      timeZone = utc
+      set2DigitYearStart(Date(-631152000000L)) // 1950-01-01T00:00:00Z.
+    }
+
+    return dateFormat.format(date)
+  }
+
   /**
    * A timestamp like "20191215190210-0800" for 2019-12-15T19:02:10-08:00. This is the same as
```
yschimke, line 175: Isn't Z a better example here than -0800?

oldergod:
SGTM

PR #6145, opened 2020-06-25
Limited support for decoding extensions

swankjesse:


yschimke:
This is looking really nice.

okhttp-tls/src/main/kotlin/okhttp3/tls/internal/der/DerWriter.kt
```diff
@@ -83,6 +96,22 @@ internal class DerWriter(sink: BufferedSink) {
     sink.writeAll(content)
   }
 
+  /**
+   * Create a new namespace for type hints. Type hints from the enclosing type are no longer usable
+   * by the current type's members.
+   */
+  fun pushTypeHint() {
+    typeHintStack.add(null)
```
oldergod, line 104: is this a TODO? Why do you keep a stack of nulls?

swankjesse: I need to grow the stack so I have a place to put something later. It also obscures whatever value is currently at the top of the stack, which is necessary to prevent namespace pollution.

PR #6148, opened 2020-06-27
Move tag management in DerAdapters

swankjesse:
Previously the caller was responsible for decoding the tag on the
adapter it called. This was easy for basic tags, but proved problematic
for tags like CHOICE and ANY and typehint-dependent tags, because the
caller didn't have enough information.

This moves the tag reading and writing into the DerAdapter itself. If
it encounters a tag it can't understand, it assumes it's own value has
been skipped and returns a default.

This turns out to be a bit easier to reason about, and hopefully will
make it easier to implement AlgorithmIdentifier.parameters, which has
values that are optional or not depending on what the type hint is.


yschimke:
LGTM

PR #6149, opened 2020-06-27
Document common security issues on older Android devices

yschimke:


swankjesse:
I like it! And I moved it. I tried editing in the GitHub editor but I couldn’t add a new file without going back to my desktop, so that’s what I did.
https://github.com/square/okhttp/pull/6150

PR #6150, opened 2020-06-27
Document common security issues on older Android devices

swankjesse:
This is ychimke's update, just moved to another file and copy-edited to cover
both Android and JVM handshakes.


yschimke:
LGTM

PR #6151, opened 2020-06-27
Fix bugs so we can roundtrip certificates

swankjesse:
This gets us to the point where the byte-for-byte encoding is equiavalent
to what bouncycastle was producing.

This shook out some bugs:
 - We weren't boxing the version
 - We weren't clearing the constructed bit on boxed extensions
 - We weren't encoding null when we needed to be


yschimke:
LGTM

PR #6152, opened 2020-06-28
Use OkHttp's certificate creation code

swankjesse:
We don't implement the full feature set that Bouncycastle has, but
we also don't need it.

In follow up changes I intend to remove the Bouncycastle dependency
for everything but some test cases.

### DIFF
--- a/okhttp-tls/src/main/kotlin/okhttp3/tls/HeldCertificate.kt
+++ b/okhttp-tls/src/main/kotlin/okhttp3/tls/HeldCertificate.kt
@@ -16,6 +16,7 @@
 package okhttp3.tls
 
 import java.math.BigInteger
+import java.net.InetAddress
 import java.security.GeneralSecurityException
 import java.security.KeyFactory
 import java.security.KeyPair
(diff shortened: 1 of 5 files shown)


okhttp-tls/src/main/kotlin/okhttp3/tls/HeldCertificate.kt
```diff
@@ -315,88 +321,134 @@ class HeldCertificate(
...
-        val encodableAltNames = arrayOfNulls<ASN1Encodable>(altNames.size)
-        for (i in 0 until altNames.size) {
-          val altName = altNames[i]
-          val tag = when {
-            altName.canParseAsIpAddress() -> GeneralName.iPAddress
-            else -> GeneralName.dNSName
+        val extensionValue = altNames.map {
+          when {
+            it.canParseAsIpAddress() -> {
+              generalNameIpAddress to InetAddress.getByName(it).address.toByteString()
```
yschimke, line 424: Is this known to be expanded to the full form always for IPv6?

swankjesse: Yep. The result will always be 4 or 16 bytes.

okhttp-tls/src/main/kotlin/okhttp3/tls/HeldCertificate.kt
```diff
@@ -315,88 +321,134 @@ class HeldCertificate(
...
-
-      // Subject, public & private keys for this certificate's signer. It may be self signed!
-      val signedByKeyPair: KeyPair
-      val signedByPrincipal: X500Principal
+      // Subject keys & identity.
+      val subjectKeyPair = keyPair ?: generateKeyPair()
+      val subjectPublicKeyInfo = CertificateAdapters.subjectPublicKeyInfo.fromDer(
+          subjectKeyPair.public.encoded.toByteString()
+      )
+      val subject: List<List<AttributeTypeAndValue>> = subject()
```
oldergod, line 329: woof

okhttp-tls/src/main/kotlin/okhttp3/tls/HeldCertificate.kt
```diff
@@ -315,88 +321,134 @@ class HeldCertificate(
...
+          )
+      )
+
+      return HeldCertificate(subjectKeyPair, certificate.toX509Certificate())
+    }
+
+    private fun subject(): List<List<AttributeTypeAndValue>> {
+      val result = mutableListOf<List<AttributeTypeAndValue>>()
+
+      if (ou != null) {
```
oldergod, line 382: There's no better name for those?

PR #6153, opened 2020-06-28
Stop using Bouncycastle to do PKCS8 to PKCS1

swankjesse:
PKCS8 is a PrivateKeyInfo
PKCS1 is the private key field of a PKCS8

We need to do a simple unboxing to go from one to the other.

### DIFF
--- a/okhttp-tls/src/main/kotlin/okhttp3/tls/HeldCertificate.kt
+++ b/okhttp-tls/src/main/kotlin/okhttp3/tls/HeldCertificate.kt
@@ -49,7 +49,6 @@ import okhttp3.tls.internal.der.Validity
 import okio.ByteString
 import okio.ByteString.Companion.decodeBase64
 import okio.ByteString.Companion.toByteString
-import org.bouncycastle.asn1.pkcs.PrivateKeyInfo
 import org.bouncycastle.jce.provider.BouncyCastleProvider
 
 /**
@@ -180,8 +179,8 @@ class HeldCertificate(
   }
 
   private fun pkcs1Bytes(): ByteString {
-    val privateKeyInfo = PrivateKeyInfo.getInstance(keyPair.private.encoded)
-    return privateKeyInfo.parsePrivateKey().toASN1Primitive().encoded.toByteString()
+    val decoded = CertificateAdapters.privateKeyInfo.fromDer(keyPair.private.encoded.toByteString())
+    return decoded.privateKey
   }
 
   /** Build a held certificate with reasonable defaults. */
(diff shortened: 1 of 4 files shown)


okhttp-tls/src/main/kotlin/okhttp3/tls/HeldCertificate.kt
```diff
@@ -180,8 +179,8 @@ class HeldCertificate(
   }
 
   private fun pkcs1Bytes(): ByteString {
-    val privateKeyInfo = PrivateKeyInfo.getInstance(keyPair.private.encoded)
-    return privateKeyInfo.parsePrivateKey().toASN1Primitive().encoded.toByteString()
+    val decoded = CertificateAdapters.privateKeyInfo.fromDer(keyPair.private.encoded.toByteString())
+    return decoded.privateKey
```
yschimke, line 183: Is there value in returning PrivateKeyInfo over the private Key bytes?

yschimke: scratch -- that internal class.

okhttp-tls/src/test/java/okhttp3/tls/internal/der/DerCertificatesTest.kt
```diff
@@ -732,6 +732,30 @@ internal class DerCertificatesTest {
...
+        "g1yzK7VijygvBoUrr+rt2lbmKgcUQ/rxu8IIQk0M/xgJqSkXDXuOnboGM7sQSKfJAZUtT7xozvLzV7ECQQCJW59w" +
+        "7NIM0qZ/gIX2gcNZr1B/V3zcGlolTDciRm+fnKGNt2EEDKnVL3swzbEfTCa48IT0QKgZJqpXZERa26UHAkBLXmiP" +
+        "5f5pk8F3wcXzAeVw06z3k1IB41Tu6MX+CyPU+TeudRlz+wV8b0zDvK+EnRKCCbptVFj1Bkt8lQ4JfcnhAkAk2Y3G" +
+        "z+HySrkcT7Cg12M/NkdUQnZe3jr88pt/+IGNwomc6Wt/mJ4fcWONTkGMcfOZff1NQeNXDAZ6941XCsIVAkASOg02" +
+        "PlVHLidU7mIE65swMM5/RNhS4aFjez/MwxFNOHaxc9VgCwYPXCLOtdf7AVovdyG0XWgbUXH+NyxKwboE")
+        .decodeBase64()!!
+
+    val decoded = CertificateAdapters.privateKeyInfo.fromDer(privateKeyInfoByteString)
+
+    assertThat(decoded.version).isEqualTo(0L)
```
yschimke, line 751: nit: 0 makes me nervous as test data.

swankjesse: It means "v1" and I expect it's the only value we're likely to see in practice. The work to get a v2 is annoying because it means we need values for otherwise optional fields, and I expect that'll be annoying to do!

PR #6154, opened 2020-06-29
Honor RFC 5280 when choosing a time format

swankjesse:
We switch to 4-digit years at the last possible moment.

### DIFF
--- a/okhttp-tls/src/main/kotlin/okhttp3/tls/internal/der/CertificateAdapters.kt
+++ b/okhttp-tls/src/main/kotlin/okhttp3/tls/internal/der/CertificateAdapters.kt
@@ -17,6 +17,7 @@ package okhttp3.tls.internal.der
 
 import java.math.BigInteger
 import okio.ByteString
+import okio.IOException
 
 /**
  * ASN.1 adapters adapted from the specifications in [RFC 5280][rfc_5280].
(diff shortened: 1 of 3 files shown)


okhttp-tls/src/main/kotlin/okhttp3/tls/internal/der/CertificateAdapters.kt
```diff
@@ -31,11 +32,43 @@ internal object CertificateAdapters {
...
+    override fun matches(header: DerHeader): Boolean {
+      return Adapters.UTC_TIME.matches(header) || Adapters.GENERALIZED_TIME.matches(header)
+    }
+
+    override fun readValue(reader: DerReader): Long {
+      val peekHeader = reader.peekHeader()
+          ?: throw IOException("expected time but was exhausted at $reader")
+
+      return when {
+        peekHeader.tagClass == Adapters.UTC_TIME.tagClass &&
```
yschimke, line 52: Should you be strict here, Postel's law doesn't seem like a good fit for security certificates.  Fail if the certificate doesn't follow the 2049/50 rule.  It's no longer distinct and the certificate bytes are wrong etc.

swankjesse: Oooooh, there’s a test case I’m missing. I need to confirm I can handle 4-digit dates preceding 2050. Here’s what the spec says:

> CAs conforming to this profile MUST always encode certificate
> validity dates through the year 2049 as UTCTime; certificate validity
> dates in 2050 or later MUST be encoded as GeneralizedTime.
> Conforming applications MUST be able to process validity dates that
> are encoded in either UTCTime or GeneralizedTime.

yschimke: Being lax seems to miss the point :) But yep, no issue with you following the spec.

> 4. Distinguished Encoding Rules
> The Distinguished Encoding Rules for ASN.1, abbreviated DER, are a subset of BER, and give exactly one way to represent any ASN.1 value as an octet string. DER is intended for applications in which a unique octet string encoding is needed, as is the case when a digital signature is computed on an ASN.1 value. DER is defined in Section 8.7 of X.509.

swankjesse: Yeah, I think that’s a lot more optimistic than is achievable in practice.

The date encoding stuff is obvious, and there’s similar things to that like PrintableString vs. IA5String vs. UTF8String. If we were building a production CA I’d want to retain which encoding was used on input. Ick.

PR #6157, opened 2020-06-30
Improve code style for the new DER package

swankjesse:
Improve docs, fix some names, fix some internal APIs.

### DIFF
--- a/okhttp-tls/src/main/kotlin/okhttp3/tls/HeldCertificate.kt
+++ b/okhttp-tls/src/main/kotlin/okhttp3/tls/HeldCertificate.kt
@@ -44,6 +44,11 @@ import okhttp3.tls.internal.der.CertificateAdapters.generalNameDnsName
 import okhttp3.tls.internal.der.CertificateAdapters.generalNameIpAddress
 import okhttp3.tls.internal.der.Extension
 import okhttp3.tls.internal.der.ObjectIdentifiers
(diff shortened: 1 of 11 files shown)


okhttp-tls/src/main/kotlin/okhttp3/tls/HeldCertificate.kt
```diff
@@ -407,11 +412,11 @@ class HeldCertificate(
 
       if (maxIntermediateCas != -1) {
         result += Extension(
-            extnID = ObjectIdentifiers.basicConstraints,
+            id = basicConstraints,
             critical = true,
-            extnValue = BasicConstraints(
+            value = BasicConstraints(
                 ca = true,
-                pathLenConstraint = 3
```
swankjesse, line 414: only behavior change. This was a copy-paste bug and the test happens to be asserting that the sample value is exactly 3!

okhttp-tls/src/main/kotlin/okhttp3/tls/HeldCertificate.kt
```diff
@@ -187,8 +192,8 @@ class HeldCertificate(
   class Builder {
     private var notBefore = -1L
     private var notAfter = -1L
-    private var cn: String? = null
-    private var ou: String? = null
+    private var commonName: String? = null
+    private var organizationalUnit: String? = null
```
swankjesse, line 196: https://github.com/square/okhttp/pull/6152#discussion_r447000106

PR #6160, opened 2020-07-01
[WIP] testing Device Attestation

yschimke:
This is a bit exploratory, probably not something we want to support.  But maybe something to allow supporting *outside* OkHttp by extending okhttp-tls classes?


yschimke:
Will revisit after we can land the support @swankjesse added cleanly

PR #6164, opened 2020-07-02
Add limits to what length values DerReader supports

cketti:
With this change `DerReader` doesn't support reading values with a length greater than `Long.MAX_VALUE`. That shouldn't be a problem in practice.

### DIFF
--- a/okhttp-tls/src/main/kotlin/okhttp3/tls/internal/der/DerReader.kt
+++ b/okhttp-tls/src/main/kotlin/okhttp3/tls/internal/der/DerReader.kt
@@ -134,6 +134,10 @@ internal class DerReader(source: Source) {
         }
 
         var lengthBits = source.readByte().toLong() and 0xff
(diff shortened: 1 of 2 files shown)


okhttp-tls/src/main/kotlin/okhttp3/tls/internal/der/DerReader.kt
```diff
@@ -129,11 +129,16 @@ internal class DerReader(source: Source) {
       (length0 and 0b1000_0000) == 0b1000_0000 -> {
         // Length specified over multiple bytes.
         val lengthBytes = length0 and 0b0111_1111
+        if (lengthBytes > 8) throw IOException("Length encoded with more than 8 bytes is not supported")
+
         var lengthBits = source.readByte().toLong() and 0xff
```
cketti, line 134: Missing check: This byte must never be 0. It must have the highest bit set if lengthBytes == 1.

PR #6170, opened 2020-07-06
Don't crash on unknown GeneralName types.

swankjesse:
We don't have API support for all of them, but we shouldn't crash when an
unsupported name is encountered.

Also encode attributes using UTF-8, not PrintableString. Both are permitted,
but UTF-8 supports more data.


yschimke:
To spec you should look at criticalj right? Isn't that must understand?

PR #6192, opened 2020-07-21
Split simplemockwebserver

tokuhirom:
- split core logic of the MockWebServer from the `org.junit.rules.*`.
- With this patch, Users can use MockWebServer without junit4's runtime dependency.


swankjesse:
I like the idea of this. I'd potentially brake it even further, moving MockWebServer to another module that doesn't depend on JUnit!

yschimke:
@tokuhirom Are you still planning to move this forward?

yschimke:
Folded into https://github.com/square/okhttp/pull/6317

PR #6198, opened 2020-07-23
Avoid DefaultImpls bug

yschimke:
Graal and Kotlin don't like interfaces without any default methods. Continuation of https://github.com/square/okhttp/pull/5631/files

https://youtrack.jetbrains.com/issue/KT-33097


meoyawn:
sad this didn't make it to 4.8.1 😞

@swankjesse are okhttp snapshots still a thing? Can't find 4.9.0 https://oss.sonatype.org/content/repositories/snapshots/com/squareup/okhttp3/logging-interceptor/

JakeWharton:
I'm not sure why snapshots aren't showing up. They seem to be running on CI but definitely don't appear.

PR #6207, opened 2020-08-04
Added new provider

koki-es:
Added new provider alekberg.net


yschimke:
This is just test code, we don't intend to build a full list of providers here.

Are you having other issues with the DoH support?  Please report an issue if you are.

PR #6208, opened 2020-08-04
Add module version check

yschimke:
Test a strict version check, initially only with MockWebServer if two release builds of okhttp jars are used.  Any snapshot or dev builds are allowed and may or may not fail for other incompatibility reasons.  But mixing release builds is enforced as a failure.

Support for actual breaking changes has been pretty light recently, so occasional reports of users confused by version incompatibilities when mixing versions is more notceable.  I hope this PR just removes a class of bug reports and stackoverflow questions, since the error is immediately obvious.

If any snapshot versions are used, then I don't want to stop people working around issues, so strictly doing this with released versions.

### DIFF
--- a/okhttp/src/test/java/okhttp3/OkHttpTest.kt
+++ b/okhttp/src/test/java/okhttp3/OkHttpTest.kt
@@ -35,11 +35,11 @@ class OkHttpTest {
   fun testReleaseVersionIncompatibity() {
     try {
       OkHttp.checkVersion(
-          module = "okhttp-mockwebserver", moduleVersion = "4.20.0", okhttpVersion = "4.21.0"
+          module = "mockwebserver", moduleVersion = "4.20.0", okhttpVersion = "4.21.0"
       )
       fail()
     } catch (ise: IllegalStateException) {
-      assertEquals("com.squareup.okhttp3:okhttp:4.21.0 not compatible with com.squareup.okhttp3:okhttp-mockwebserver:4.20.0", ise.message)
+      assertEquals("com.squareup.okhttp3:okhttp:4.21.0 is not compatible with com.squareup.okhttp3:mockwebserver:4.20.0", ise.message)
     }
   }
 
--- a/okhttp/src/test/java/okhttp3/OkHttpTest.kt
+++ b/okhttp/src/test/java/okhttp3/OkHttpTest.kt
@@ -16,11 +16,45 @@
 package okhttp3
 
 import org.assertj.core.api.Assertions.assertThat
+import org.junit.Assert.assertEquals
+import org.junit.Assert.fail
 import org.junit.Test
 
 class OkHttpTest {
   @Test
   fun testVersion() {
     assertThat(OkHttp.VERSION).matches("[0-9]+\\.[0-9]+\\.[0-9]+(-.+)?")
+  }
+
+  @Test
+  fun testReleaseVersionCompatibity() {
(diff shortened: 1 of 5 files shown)


yschimke:
I thought about moving this into MockWebServer.COMPANION, but I suspect that causes confusing errors during general classloading etc.  So figure direction objection creation is safer.  The test case should fail to be created, or run, rather than the test run terminating while building the list of test cases.

okhttp/src/test/java/okhttp3/OkHttpTest.kt
```diff
@@ -16,11 +16,45 @@
...
     assertThat(OkHttp.VERSION).matches("[0-9]+\\.[0-9]+\\.[0-9]+(-.+)?")
   }
+
+  @Test
+  fun testReleaseVersionCompatibity() {
+    OkHttp.checkVersion(module = "okhttp-mockwebserver", moduleVersion = "4.20.0", okhttpVersion = "4.20.0")
+  }
+
+  @Test
+  fun testReleaseVersionIncompatibity() {
```
swankjesse, line 35: sp. incompatibility

swankjesse: (and below)

okhttp/src/test/java/okhttp3/OkHttpTest.kt
```diff
@@ -16,11 +16,45 @@
...
+  }
+
+  @Test
+  fun testSnapshotVersionCompatibity() {
+    OkHttp.checkVersion(module = "okhttp-mockwebserver", moduleVersion = "4.20.0-SNAPSHOT", okhttpVersion = "4.20.0-SNAPSHOT")
+  }
+
+  @Test
+  fun testSnapshotReleaseVersionIncompatibity() {
+    OkHttp.checkVersion(module = "okhttp-mockwebserver", moduleVersion = "4.20.0", okhttpVersion = "4.21.0-SNAPSHOT")
```
swankjesse, line 53: what makes this case okay?

yschimke:
> Although we see lots of version skew errors in our issues, I assume most version skew is benign and doesn’t harm users in practice.

I'm surprised you are ok with this, maybe less with mockwebserver which is mainly annoying.  But eventually with okhttp-logging-interceptor, we should strictly enforce this and catch during development before an APK gets shipped.

But overall, happy to wait until we promote a BOM more.

okhttp/src/test/java/okhttp3/OkHttpTest.kt
yschimke: Basically to not be annoying to anyone that is testing a dev version of OkHttp.  Probably testing a PR build, or possibly knowingly releases with locally patched lib that we don't control.

But if you are using released tagged versions of your libs, then you don't cross the versions.

image

yschimke:
Will close for now, consider reopening in future.

swankjesse:
One other option is to avoid sharing internal stuff between modules. This is easy in most things but not MockWebServer; the HTTP/2 internals are pretty huge.

yschimke:
Will wait for a signal if you change your mind on this.

yschimke:
Another case https://github.com/square/okhttp/issues/6231

yschimke:
@swankjesse saw https://stackoverflow.com/questions/64351110/no-static-method-delimiteroffset-in-class-lokhttp3-internal-util

Do you have an example for the BOM?

yschimke:
Another example which I presume is down to version skew https://stackoverflow.com/questions/66888631/ionic-android-app-crashes-with-okhttp-error

PR #5836, opened 2020-03-01
Add SuppressedExceptions for multiple Routes

yschimke:
Before we have Happy Eyeballs, we currently throw the exception for the final Route.  Often a failure for an earlier Route may be more actionable, so add to suppressed.

n.b. RouteException already has a suppressed chain so this is a two level tree effectively.

PR #6218, opened 2020-08-11
Change list to mutableList to avoid creating a new list.

TangHuaiZhe:
'+=' of `_Collection.kt` create new list under the hood:

```
/**
 * Returns a list containing all elements of the original collection and then the given [element].
 */
public operator fun <T> Collection<T>.plus(element: T): List<T> {
    val result = ArrayList<T>(size + 1)
    result.addAll(this)
    result.add(element)
    return result
}
```

And `MutableCollection.kt` only adds the specified [element] to this mutable collection.

```
/**
 * Adds the specified [element] to this mutable collection.
 */
@kotlin.internal.InlineOnly
public inline operator fun <T> MutableCollection<in T>.plusAssign(element: T) {
    this.add(element)
}
```

### DIFF
--- a/okhttp/src/main/kotlin/okhttp3/internal/http/RetryAndFollowUpInterceptor.kt
+++ b/okhttp/src/main/kotlin/okhttp3/internal/http/RetryAndFollowUpInterceptor.kt
@@ -61,7 +61,7 @@ class RetryAndFollowUpInterceptor(private val client: OkHttpClient) : Intercepto
     var followUpCount = 0
     var priorResponse: Response? = null
     var newExchangeFinder = true
-    var recoveredFailures = listOf<IOException>()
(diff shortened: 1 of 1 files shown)


JakeWharton:
I think the idea here is that we expect this list to be empty 99.9% of the time and thus `listOf()` is free. When an exception occurs, you're already allocating a bunch of things for the exception's stack trace elements so what's one more for the list.

TangHuaiZhe:
> I think the idea here is that we expect this list to be empty 99.9% of the time and thus `listOf()` is free. When an exception occurs, you're already allocating a bunch of things for the exception's stack trace elements so what's one more for the list.

Thanks for your professional explanation.

The discussion of this commit is in [Add SuppressedExceptions for multiple Routes](https://github.com/square/okhttp/pull/5836).

PR #6225, opened 2020-08-17
Update MockWebServer.kt to fix runtime crash caused by "okhttp3.mockwebserver.MockWebServer" exceeds limit of 23 characters. 

russell-shizhen:
Fix bug https://stackoverflow.com/questions/63387807/java-lang-illegalargumentexception-log-tag-okhttp3-mockwebserver-mockwebserver

### DIFF
--- a/mockwebserver/src/main/kotlin/okhttp3/mockwebserver/MockWebServer.kt
+++ b/mockwebserver/src/main/kotlin/okhttp3/mockwebserver/MockWebServer.kt
@@ -1145,6 +1145,6 @@ class MockWebServer : ExternalResource(), Closeable {
       override fun getAcceptedIssuers(): Array<X509Certificate> = throw AssertionError()
     }
 
-    private val logger = Logger.getLogger(MockWebServer::class.java.name)
+    private val logger = Logger.getLogger(MockWebServer::class.java.simpleName)
   }
 }


yschimke:
I would prefer making this change in the same place as the other runtime long class names.

https://github.com/square/okhttp/blob/b086b8d1835bcb3825d60ceb44bec7dec356405b/okhttp/src/main/kotlin/okhttp3/internal/platform/android/AndroidLog.kt#L91

The comments indicate I hit this before in testing, and then just didn't fix correctly.

yschimke:
@russell-shizhen Just checking, do you want me to make the changes discussed above or should I leave it with you?

russell-shizhen:
@yschimke , 

According to the implementation in side https://github.com/square/okhttp/blob/b086b8d1835bcb3825d60ceb44bec7dec356405b/okhttp/src/main/kotlin/okhttp3/internal/platform/android/AndroidLog.kt#L91 

```Java
private fun loggerTag(loggerName: String): String {
    // We need to handle long logger names before they hit Log.
    // java.lang.IllegalArgumentException: Log tag "okhttp3.mockwebserver.MockWebServer" exceeds limit of 23 characters
    return knownLoggers[loggerName] ?: loggerName.take(23)
  }
```
`loggerName.take(23)` will take a log tag "okhttp3.mockwebserver." (23 chars) which may not be that helpful and sometimes misleading due to its inaccuracy. Maybe some tag string that is self-explained will be more helpful for debugging.

russell-shizhen:
@yschimke , I am not sure about "should I leave it with you", what can I do?

yschimke:
Well either 

a) You intend to make further changes and we can work through the PR
b) I can submit an alternate PR

If you look at the method I sent, it already does what you are suggesting, using known abbreviations when appropriate.  So you could add a specific mapping for MockWebServer also (use a string instead of a class reference).

https://github.com/square/okhttp/blob/b086b8d1835bcb3825d60ceb44bec7dec356405b/okhttp/src/main/kotlin/okhttp3/internal/platform/android/AndroidLog.kt#L63

russell-shizhen:
@yschimke , Thank you for your advices. I have made the PR https://github.com/square/okhttp/pull/6229

PR #6228, opened 2020-08-19
Conscrypt 2.5.0 upgrade

yschimke:
New release unfortunately changes the ConscryptHostnameVerifier interface

https://groups.google.com/d/msgid/conscrypt/fe34caac-4cdb-4e64-b877-ea8cd0ca8b64n%40googlegroups.com?utm_medium=email&utm_source=footer

https://github.com/google/conscrypt/issues/877

### DIFF
--- a/okhttp/src/main/kotlin/okhttp3/internal/platform/ConscryptPlatform.kt
+++ b/okhttp/src/main/kotlin/okhttp3/internal/platform/ConscryptPlatform.kt
@@ -83,9 +83,7 @@ class ConscryptPlatform private constructor() : Platform() {
   override fun newSslSocketFactory(trustManager: X509TrustManager): SSLSocketFactory {
     return newSSLContext().apply {
       init(null, arrayOf<TrustManager>(trustManager), null)
-    }.socketFactory.also {
-      Conscrypt.setUseEngineSocket(it, true)
-    }
+    }.socketFactory
   }
 
   companion object {
--- a/okhttp/src/main/kotlin/okhttp3/internal/platform/ConscryptPlatform.kt
+++ b/okhttp/src/main/kotlin/okhttp3/internal/platform/ConscryptPlatform.kt
@@ -50,7 +50,7 @@ class ConscryptPlatform private constructor() : Platform() {
       "Unexpected default trust managers: ${trustManagers.contentToString()}"
     }
     val x509TrustManager = trustManagers[0] as X509TrustManager
-    Conscrypt.setHostnameVerifier(x509TrustManager) { _, _ -> true }
+    Conscrypt.setHostnameVerifier(x509TrustManager) { _, _, _ -> true }
     return x509TrustManager
   }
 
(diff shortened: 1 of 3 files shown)


okhttp/src/main/kotlin/okhttp3/internal/platform/ConscryptPlatform.kt
```diff
@@ -50,7 +50,7 @@ class ConscryptPlatform private constructor() : Platform() {
       "Unexpected default trust managers: ${trustManagers.contentToString()}"
     }
     val x509TrustManager = trustManagers[0] as X509TrustManager
-    Conscrypt.setHostnameVerifier(x509TrustManager) { _, _ -> true }
+    Conscrypt.setHostnameVerifier(x509TrustManager) { _, _, _ -> true }
```
swankjesse, line 53: I think you can make this work with both versions if you’re careful. Implement an interface in a class that defines both the old and the new method, and apply `override` on only one of the two methods. At runtime things should just work.

okhttp/src/main/kotlin/okhttp3/internal/platform/ConscryptPlatform.kt
```diff
@@ -83,9 +83,7 @@ class ConscryptPlatform private constructor() : Platform() {
   override fun newSslSocketFactory(trustManager: X509TrustManager): SSLSocketFactory {
     return newSSLContext().apply {
       init(null, arrayOf<TrustManager>(trustManager), null)
-    }.socketFactory.also {
-      Conscrypt.setUseEngineSocket(it, true)
```
swankjesse, line 87: no longer necessary?

yschimke: It's the default as per the release notes

yschimke:
cc @prbprbprb Can you take a look also

yschimke:
Tests in CI will kick on on landing e.g. https://app.circleci.com/pipelines/github/square/okhttp/3688/workflows/aebd8a95-ef1c-4229-94a2-a1377b2600a6/jobs/16159

okhttp/src/main/kotlin/okhttp3/internal/platform/ConscryptPlatform.kt
```diff
@@ -50,10 +51,28 @@ class ConscryptPlatform private constructor() : Platform() {
       "Unexpected default trust managers: ${trustManagers.contentToString()}"
     }
     val x509TrustManager = trustManagers[0] as X509TrustManager
-    Conscrypt.setHostnameVerifier(x509TrustManager) { _, _ -> true }
+    // Disabled because OkHttp will run anyway
+    Conscrypt.setHostnameVerifier(x509TrustManager, DisabledHostnameVerifier)
     return x509TrustManager
   }
 
+  internal object DisabledHostnameVerifier : ConscryptHostnameVerifier {
```
swankjesse, line 59: dig

PR #6235, opened 2020-08-24
Document interceptor throwing modes

swankjesse:


okhttp/src/main/kotlin/okhttp3/Interceptor.kt
```diff
@@ -22,6 +22,39 @@ import java.util.concurrent.TimeUnit
...
+ *
+ *  * For synchronous calls made with [Call.execute], the exception is propagated to the caller.
+ *
+ *  * For asynchronous calls made with [Call.enqueue], an [IOException] is propagated to the caller
+ *    indicating that the call was canceled. The interceptor's exception is delivered to the current
+ *    thread's [uncaught exception handler][Thread.UncaughtExceptionHandler]. By default this
+ *    crashes the application on Android and prints a stacktrace on the JVM. (Crash reporting
+ *    libraries may customize this behavior.)
+ *
+ * A good way to signal a failure is with a synthetic HTTP response:
```
yschimke, line 40: Is the idea that IOException should be purely the domain of connectivity issues?

yschimke (approved):

okhttp/src/main/kotlin/okhttp3/Interceptor.kt
swankjesse: Not exclusively, but synthetic failures are a good option to call out.

PR #6240, opened 2020-08-27
When a URL cannot be parsed, add the input to the exception message.

jeffbarge:
This will greatly help with debugging when stacktraces aren't helpful.


swankjesse:
Unfortunately we don‘t like to put URLs or strings-supposed-to-be URLs in stacktraces because they frequently contain secrets.

PR #6245, opened 2020-08-31
Optimize the code here.

ThirdGoddess:
Optimize the loop structure.


swankjesse:
The goal of these examples is to provide clarity. In this case, the optimization is a small distraction.

ThirdGoddess:
Thank you. OK, I see.

PR #6257, opened 2020-09-09
Fix build on windows

yschimke:
See stackoverflow for why this is required, related to the gradle daemon.

https://stackoverflow.com/a/34717160/1542667

I don't currently have any concern that their is a logic bug with string encodings in src/main, but rather that strings in tests are compiled assuming the wrong file encoding, changing the test meaning.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -123,8 +123,6 @@ subprojects { project ->
   sourceCompatibility = JavaVersion.VERSION_1_8
   targetCompatibility = JavaVersion.VERSION_1_8
 
-  compileJava.options.encoding = 'UTF-8'
-
   tasks.withType(JavaCompile) {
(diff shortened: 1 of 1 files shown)


yschimke:
image

windows-1252

yschimke:
Typical errors

image

build.gradle
```diff
@@ -123,6 +123,12 @@ subprojects { project ->
   sourceCompatibility = JavaVersion.VERSION_1_8
   targetCompatibility = JavaVersion.VERSION_1_8
 
+  compileJava.options.encoding = 'UTF-8'
```
JakeWharton, line 126: Delete? The second one applies to `compileJava` as it is a `JavaCompile` instance.

JakeWharton (approved):

PR #6263, opened 2020-09-11
Conscrypt 2.5.1 Upgrade

yschimke:


yschimke:
Fixes a jar signing bug

https://github.com/google/conscrypt/commit/f2137890dd39d972e7a1a54190acf97dcb46dee6

It's ultimately more important what clients deploy with, but we should update anyway.

PR #6265, opened 2020-09-11
Fix the deploy_website script to handle files with spaces in their names

swankjesse:
Dokka now generates such files and it's too bad.

### DIFF
--- a/deploy_website.sh
+++ b/deploy_website.sh
@@ -34,21 +34,23 @@ cd $DIR
 # Dokka filenames like `-http-url/index.md` don't work well with MkDocs <title> tags.
 # Assign metadata to the file's first Markdown heading.
 # https://www.mkdocs.org/user-guide/writing-your-docs/#meta-data
-title_markdown_file() {
-  TITLE_PATTERN="s/^[#]+ *(.*)/title: \1 - OkHttp/"
-  echo "---"                                                     > "$1.fixed"
-  cat $1 | sed -E "$TITLE_PATTERN" | grep "title: " | head -n 1 >> "$1.fixed"
-  echo "---"                                                    >> "$1.fixed"
-  echo                                                          >> "$1.fixed"
-  cat $1                                                        >> "$1.fixed"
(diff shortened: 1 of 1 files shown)


deploy_website.sh
```diff
@@ -34,21 +34,23 @@ cd $DIR
...
+    echo                                                          >> "$1.fixed"
+    cat $1                                                        >> "$1.fixed"
+    mv "$1.fixed" "$1"
+  }
+
+  for MARKDOWN_FILE do
+    echo $MARKDOWN_FILE
+    title_markdown_file $MARKDOWN_FILE
+  done
+' exec-sh {} +
```
swankjesse, line 53: https://unix.stackexchange.com/a/9499

georgetsiga (approved):

deploy_website.sh
```diff
@@ -34,21 +34,23 @@ cd $DIR
...
-  title_markdown_file $MARKDOWN_FILE
-done
+find docs/4.x/ -name '*.md' -exec sh -c '
+  title_markdown_file() {
+    TITLE_PATTERN="s/^[#]+ *(.*)/title: \1 - OkHttp/"
+    echo "---"                                                     > "$1.fixed"
+    cat $1 | sed -E "$TITLE_PATTERN" | grep "title: " | head -n 1 >> "$1.fixed"
+    echo "---"                                                    >> "$1.fixed"
+    echo                                                          >> "$1.fixed"
+    cat $1                                                        >> "$1.fixed"
```
monkey-mas, line 45:
```sh
cat "$1"
```
I think we need to add double quotes to `$1` since `$1` can include a space. I tried the command(`cat $1`) by using a file("foo 2.txt") and got this error.
```sh
cat: .//foo: No such file or directory
cat: .//2.txt: No such file or directory
```

https://superuser.com/questions/606874/cannot-cat-file-which-has-space-in-name-in-linux/606876

monkey-mas (approved):
LGTM with comment! 👍

PR #6275, opened 2020-09-20
Avoid a bad cache entry persistently failing requests, and log the affected url

yschimke:


swankjesse:
Ick. I like that this makes OkHttp more robust to cache problems! But it also seems like it’s sweeping a lot of potential problems under the rug. Like leaked resources that might be better to crash the application.

yschimke:
> Ick. I like that this makes OkHttp more robust to cache problems! But it also seems like it’s sweeping a lot of potential problems under the rug. Like leaked resources that might be better to crash the application.

Agreed, but the corruption will survive restarts and cause crash loops etc, even once you fix say a server change that triggered it.  Closing for now.

swankjesse:
If a server change can corrupt the cache, we’re in big trouble. My best guess is that a file write fails in a way we don’t expect. Or perhaps a developer puts their cache in a directory that gets backed up, and the restored backup is truncated? There’s so many ugly failure modes.

PR #6282, opened 2020-09-27
Added OSGi metadata via build

swimmesberger:
Added bnd plugin to the gradle build so that META-INF/MANIFEST.MF file contains OSGi metadata.
This is analogous to the implementation in okio: https://github.com/square/okio/pull/777

I added some comments for those modules where it was difficult for me to know which packages should be exported or not. There is also the okhttp-urlconnection module which uses the same package name as the root okhttp project which BND/OSGi does not really like. In OSGi each module should export only a uniquely identifiable package. Therefore for this package I had to find a "creative" way around it. Maybe the better solution would be to publish a OSGi specific artifact for that module with a changed package name (e.g. okhttp3.urlconnection). I don't really know how much that module is used and why it has the same package name.


okhttp-tls/build.gradle
```diff
@@ -1,9 +1,13 @@
 apply plugin: 'me.champeau.gradle.japicmp'
 
 jar {
-  manifest {
-    attributes('Automatic-Module-Name': 'okhttp3.tls')
-  }
+  // modify these lines for MANIFEST.MF properties or for specific bnd instructions
+  // are okhttp3.tls.internal.* packages used in dependent libraries?
```
swankjesse, line 5: Lets talk about this. Can OkHttp libraries use each other’s internal packages?

okhttp-urlconnection/build.gradle
```diff
@@ -1,9 +1,19 @@
 apply plugin: 'me.champeau.gradle.japicmp'
 
 jar {
-  manifest {
-    attributes('Automatic-Module-Name': 'okhttp3.urlconnection')
-  }
+  // modify these lines for MANIFEST.MF properties or for specific bnd instructions
+  // because this project uses the same package name as the default okhttp project (which is mostly "invalid" for OSGi)
+  // we have to remove okhttp from the bnd classpath and import it's packages manually
```
swankjesse, line 6: what does OSGi want us to do in this case? Should we just say this package doesn’t exist for OSGi users? It’s pretty negligible in terms of value.

okhttp/build.gradle
```diff
@@ -1,9 +1,22 @@
 apply plugin: 'me.champeau.gradle.japicmp'
 
 jar {
-  manifest {
-    attributes('Automatic-Module-Name': 'okhttp3')
-  }
+  // modify these lines for MANIFEST.MF properties or for specific bnd instructions
+  // internal package needs to be exported too because interceptors use internal packages
```
swankjesse, line 5: interceptors? Or sibling libraries?

okhttp/build.gradle
```diff
@@ -1,9 +1,22 @@
...
+  bnd '''
+  Export-Package: \
+  okhttp3,\
+  okhttp3.internal.*
```
swankjesse, line 9: Is there a concept of a friend package?

okhttp/build.gradle
```diff
@@ -1,9 +1,22 @@
...
+  Import-Package: \
+  android.*;resolution:=optional,\
+  dalvik.system;resolution:=optional,\
+  org.conscrypt;resolution:=optional,\
+  org.bouncycastle.*;resolution:=optional,\
+  org.openjsse.*;resolution:=optional,\
```
swankjesse, line 15: If we introduce new dependencies, is there a test that’ll fail to remind us to update this?

okhttp-tls/build.gradle
swimmesberger: No. In OSGi there is no concept that allows "special" modules to access packages other modules can not. Therefore if any other module needs access to the  okhttp3.tls.internal.* packages we need to export them. If we do not export them, OSGi will detect at resolution time that the package okhttp3.tls.internal is imported from module x but module y does not export it.

okhttp-urlconnection/build.gradle
swimmesberger: There are 2 solutions to this:
- Rename the package name (OSGi preferred way - breaks backwards compatibility)
- Make okttp-urlconnection a fragment bundle which will at runtime basically mean that the root okhttp bundle and the okhttp-urlconnection bundle use the same ClassLoader. This means that externally when both bundles are present it's like "a single bundle". Fragment bundles are not really best practices and should generally be avoided if possible but it might work for this case.

okhttp/build.gradle
swimmesberger: Yeah I meant the sibling libraries.

okhttp/build.gradle
swimmesberger: See okhttp-tls comment

okhttp/build.gradle
swimmesberger: If new dependencies are introduced you have to decide on a case by case basis if it's an optional dependency or not. If you forget to add the library in the list the worst case will be that a OSGi user of that library has to put that library on the classpath to ensure OSGi can resolve all imports. The resolution optional attribute basically means that those packages are not required at runtime.

okhttp-urlconnection/build.gradle
swankjesse: Can we make all of OkHttp’s libraries a fragment bundle and get the `internal` visibility we crave?

Aside from architectural purity, what’s the drawback of fragment bundles?

swankjesse:
Is there a way to run our JUnit tests with OSGi somehow, so our tests fail if we break OSGi’s rules? I’d hate to introduce an `internal`-style dependency and ship a broken OSGi release.

okhttp-urlconnection/build.gradle
swimmesberger: Fragment bundles are kind of niche and behave differently based on the environment on some cases. PDE for example the OSGi build system from eclipse (the eclipse version of "bnd") will fail to honor packages **added** by a fragment module which is not defined on the host module. To make things work for PDE we would additionally need the "Eclipse-ExtensibleAPI: true" attribute in the manifest.

For the Okhttp use-case fragment bundles might work because every sibling module depends on the okhttp core library anyway. Do you know if any popular thirdparty library depend on one or more internal packages? (e.g. retrofit?) If that's the case we would need to export these packages anyway.

swimmesberger:
I could add a bnd [Launchpad](https://bnd.bndtools.org/chapters/315-launchpad-testing.html) test that creates a OSGi environment with every okhttp module and tries to resolve it (resolving is the OSGi phase where OSGi tries to find a Export-Package for every Import-Package). Most OSGi issues will show in the resolving phase. When resolving works the OSGi metadata should be fine.

okhttp-urlconnection/build.gradle
swankjesse: OkHttp’s `internal` stuff is OkHttp-only.

Another option – we can stop reaching across submodules to do internal stuff? That might be simpler.

swankjesse: ... turns out that’s a nonstarter for MockWebServer.

swimmesberger: Getting rid of the internal stuff would be indeed the best solution. For okhttp-urlconnection we would still need a fragment because of the package name issue.

Btw. MockWebServer - I have missed that module, I have to modify the build.gradle for that one.

Maybe a hybrid solution would be possible? For those modules still requiring internal apis we use the fragment approach and for the other modules we create a "normal" OSGi bundle.

swankjesse:
So I’m currently struggling with `okhttp3.internal`. I want MockWebServer to use code from here (low-level HTTP/2 protocol implementation) but I don’t want to publish that package to all OSGi users. As a library author I was expecting OSGi would give me better encapsulation, but it sounds like that isn’t true here because of the `internal` sharing for MockWebServer.

If OSGi doesn’t add encapsulation, what does it add?

swankjesse:
From the Okio bug:

> Not providing OSGi metadata wastes many hours in many enterprises around the world

What are enterprises getting for those hours? Is it encapsulation, just not in this case?

swimmesberger:
For the MockWebServer we should consider using a fragment, it allows us to keep the internal package private and still give the MockWebServer module the possibility to access those packages.

OSGi enforces encapsulation, the issue is getting libraries not written with OSGi in mind to fit into the OSGi corset. When a project is fully engaging in the OSGi world it gives you encapsulation no other ecosystem is giving you but it comes with a cost.
As an example: The model of enforcing versions for imported packages gives OSGi the ability to keep multiple versions of different libraries in a single JVM without any class loading/reference issues, each module knows exactly which version of a library it should load. This adds a lot of flexibility in large projects because every module can be developed independently without taking care special care about library version. Of course to keep things clean it's best practice to keep all modules on the same page but when different teams develop different modules that get's more difficult. OSGi also provides also a very flexible and clean "dependency injection" API with the [DS](https://docs.osgi.org/specification/osgi.cmpn/7.0.0/service.component.html) specification.
Our project for example provides a couple of key services (you can think of microservices in a single JVM) that other modules can use (it's like GRPC/REST microservice just in a single JVM and "simple" method calls). For our use-case (non cloud based) it's a perfect fit because we have basically a microservice way of developing (with the requirement on JVM languages...) a application in a single monolithic application. Which is especially important because we have no way to deploy a set of services on docker or stuff like that because as I said we are not in an cloud environment.

swankjesse:
Yep, good point about the classloading features. 

Can we make `okhttp3.internal` published for now, and switch to fragments later?

JakeWharton:
The internal package is already exposed to jar and java module system users and we break its binary API constantly. I don't think we'd treat removal of it from OSGi as anything but a bug fix.

swankjesse:
Wish list:

1. Nothing fragile. If I inadvertently break OSGi’s rules, it should fail the build.
2. Easy to maintain. I don't want to study bnd documentation.
3. More encapsulation. This is my lowest priority.

I think this means we publish `internal` because it's least fragile.

swimmesberger:
I have added a test to verify the OSGi metadata can be verified. The test fails currently because the dependent okio library used is not OSGi ready yet. Therefore the okio library can't be resolved.
I also only apply the bnd plugin to selected projects therefore mockwebserver, okhttp-hpacktests and okhttp-testing-support do not contain any OSGi metadata. For testing libraries OSGi metadata does not really make sense therefore I excluded them for now.

okhttp/src/test/java/okhttp3/osgi/OkHttpOsgiTest.java
```diff
@@ -0,0 +1,96 @@
...
+import java.io.InputStream;
+import java.net.URISyntaxException;
+import java.net.URL;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.stream.Stream;
+
+/**
+ * Tries to resolve the OSGi metadata specified in the
```
adriancole, line 21: someone died writing? :)

okhttp/src/test/resources/okhttp3/osgi/resolveTest.bndrun
```diff
@@ -0,0 +1,13 @@
+-runfw: org.eclipse.osgi
```
adriancole, line 1: it feels weird to split responsibility between code and text like this. I'm not sure it is bad per se, but I think sometimes of testcontainers and enjoy seeing the setup logic in one markup (even if that is compiled :) )

okhttp/build.gradle
adriancole: in zipkin we do things like this..

export your internal package..
https://github.com/openzipkin/brave/blob/master/brave/bnd.bnd#L9

explicitly import when internal is used
https://github.com/openzipkin/brave/blob/master/context/slf4j/bnd.bnd#L1

block import when something is source retention..
https://github.com/openzipkin/brave/blob/master/instrumentation/http/bnd.bnd#L2

okhttp/build.gradle
adriancole: maybe a move the comment about optional (ex why not `Import-Package:  *`) here?

`Platform` or similar might need an update to mention when adding optional library support, to not forget to update this and wherever the test cases are! It could also be useful to use those paths in the OSGi test

adriancole (commented):
Thanks for the help with this. cc @reta if you don't mind helping review

okhttp/src/test/java/okhttp3/osgi/OkHttpOsgiTest.java
swimmesberger: Hah - fixed it  🤷‍♂️

okhttp/src/test/resources/okhttp3/osgi/resolveTest.bndrun
swimmesberger: I have moved the bnd instructions from the bndrun, cnf/build.bnd into code so everything should be in one place now.

okhttp/build.gradle
swimmesberger: Just to be sure I understand the approach you set the "braveinternal" attribute mandatory which enforces that the "braveinternal" attribute is set for every Import-Package statement? Therefore if someone wants to Import one of the internal packages the "braveinternal" attribute must explicitly set otherwise the bundle would not resolve?

So for my understanding this would be like a "I agree that what I'm importing is not public API and can change at will". I kind of like that approach because it prevents accidently importing a package which you shouldn't because it's internal.

okhttp/build.gradle
swimmesberger: If we add a simple "*" here all the android/ssl packages would be required on the classpath for the bundle to resolve therefore I explicitly define the packages which are not required in every case at runtime.

Hmm... maybe we should add the comment to the build.gradle dependency block instead of the platform class because the dependency block must be touched to add new dependencies. I kind of dislike adding comments to code which are build related.

swimmesberger:
@adriancole thanks for helping out with this

okhttp/build.gradle
adriancole: I think you understand it, yes. We have had bugs where we forgot to explicitly import the internal thing, but I think it is a better optimization anyway.

okhttp/build.gradle
adriancole: I suppose I meant whatever the package-info equiv of kotlin. I suspect any new member of this directory would possibly taint here. https://github.com/square/okhttp/tree/master/okhttp/src/main/kotlin/okhttp3/internal/platform

Not too fussed where the comment goes, it can be in the CONTRIBUTING file.

thx for the clarification on not '*' that part could be a comment inline here in the gradle.

okhttp/build.gradle
swimmesberger: I did hope that bnd does not specify mandatory attributes automatically but I have just tried it and bnd does in fact automatically infer mandatory attributes for Import-Package statements. Therefore if someone is using bnd as a build system he wouldn't even notice that there is an mandatory attribute called "okhttpinternal"/"braveinternal"...

adriancole: trying to understand the impact of this.. third parties outside this repo shouldn't be using the internal code anyway, so should be not impacted by ignorance of them. Only here, we have to make sure the bnds are coherent.. am I missing something?

swimmesberger: We want to ensure that only the okhttp sibling libraries are able to use the internal package of the main okhttp project. Therefore we add a mandatory "okhttpinternal" directive which ensures that for each Import-Package: okhttp3.internal statement the attribute must be explicitly defined. For us it's easy to define that attribute for the siblings okhttp3 libraries which use those internal packages. If a 3rdparty consumer would also import one of the internal packages without defining the attribute the bundle would resolve with an error according to OSGi spec because the attribute is not defined.
Unfortunately if the 3rdparty is also using bnd as a build system (like we in our internal project for example) we could freely use the okhttp3.internal package (even though we shouldn't) and we wouldn't event notice that we should define a mandatory attribute because bnd figures that out automatically.
As an example: I tried to add that attribute to the root okhttp3 project and I didn't have to specify the attribute in the sibling okhttp-sse library manually because bnd automatically inferred the mandatory attribute which kind of destroys the purpose of those whole attribute.
<pre>
Manifest-Version: 1.0
Automatic-Module-Name: okhttp3.sse
Bnd-LastModified: 1601454505191
Bundle-ManifestVersion: 2
Bundle-Name: com.squareup.okhttp3.sse
Bundle-SymbolicName: com.squareup.okhttp3.sse
Bundle-Version: 4.10.0.SNAPSHOT
Created-By: 14.0.2 (AdoptOpenJDK)
Export-Package: okhttp3.sse;uses:="kotlin,kotlin.jvm,okhttp3";version=
 "4.10.0"
Import-Package: kotlin;version="[1.4,2)",kotlin.io;version="[1.4,2)",k
 otlin.jvm;version="[1.4,2)",kotlin.jvm.internal;version="[1.4,2)",okh
 ttp3;version="[4.10,5)",okhttp3.internal;version="[4.10,5)";okhttpint
 ernal=true,okhttp3.internal.connection;version="[4.10,5)";okhttpinter
 nal=true,okio
Private-Package: okhttp3.internal.sse
Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=1.8))"
Tool: Bnd-5.1.2.202007211702
</pre>

adriancole: If this automatically sucks in the internal stuff, it would seem to be a behavioral difference in manifest generation here vs for example maven-bundle-plugin. This might be a different interpretation of what is sane behavior, or simply an accident in one or the other.

In any case, would you agree that the internal attribution is important to annotate even the gradle plugin or default configuration thwarts that guard rail?

cc also @helgoboss for input as it is sometimes hard for project teams to come to a sane approach for things like this if they aren't actively using the tooling.

swimmesberger: I have added the attribute although when bnd is used the benefit is negligible. I did not have to explicitly define the import in the sibling libraries because bnd adds the imports with the property set to true automatically...

okhttp/build.gradle
swimmesberger: Added documentation to the CONTRIBUTING file

swimmesberger:
@adriancole @swankjesse is there anything still missing for this PR? @swankjesse is the a date when we can expect that a new okio version is released which we can use in okhttp so the tests get back to green?

adriancole (approved):
Looks good to me, disclaimer being I'm not an OSGi person.

In projects I've worked on, we used servicemix bundles of okhttp, so it is probably worth pinging someone from there for a second set of eyes cc @jbonofre @WillemJiang @iocanel

For example zipkin almost always uses okhttp for client reporting. With this in, we'd likely switch out re-bundling

yschimke:
@swimmesberger bumping okio deps here https://github.com/square/okhttp/pull/6307, I believe Jesse's plan is RC of okhttp to test this and graal specifically.

swankjesse (approved):

swankjesse:
I attempted to run this locally, and it failed like this:

```
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
Failed to deploy idea_rt.jar
Failed to deploy junit5-rt.jar
Failed to deploy junit-rt.jar
Deployed okio-jvm-2.9.0.jar
Failed to deploy kotlin-stdlib-1.4.10.jar
Failed to deploy android-all-11-robolectric-6757853.jar
Deployed bctls-jdk15on-1.66.jar
Deployed bcprov-jdk15on-1.66.jar
Deployed conscrypt-openjdk-uber-2.5.1.jar
Failed to deploy openjsse-1.1.0.jar
Deployed jsr305-3.0.2.jar
Failed to deploy animal-sniffer-annotations-1.19.jar
Failed to deploy svm-20.2.0.jar
Failed to deploy kotlin-stdlib-common-1.4.10.jar
Failed to deploy annotations-13.0.jar
Failed to deploy graal-sdk-20.2.0.jar
Failed to deploy kotlin-stdlib-jdk8-1.4.10.jar
Failed to deploy kotlin-stdlib-jdk7-1.4.10.jar
Failed to deploy junit-4.13.jar
Deployed assertj-core-3.15.0.jar
Deployed biz.aQute.resolve-5.1.2.jar
Failed to deploy AmazonCorrettoCryptoProvider-1.3.1-linux-x86_64.jar
Deployed dec-0.1.2.jar
Failed to deploy hamcrest-core-1.3.jar
Deployed biz.aQute.repository-5.1.2.jar
Deployed biz.aQute.bndlib-5.1.2.jar
Deployed slf4j-api-1.7.25.jar
Deployed bcpkix-jdk15on-1.66.jar

org.osgi.service.resolver.ResolutionException: Unable to resolve <<INITIAL>>: missing requirement osgi.identity;filter:='(osgi.identity=com.squareup.okhttp3)'

	at biz.aQute.resolve.ResolveProcess.augment(ResolveProcess.java:354)
	at biz.aQute.resolve.ResolveProcess.augment(ResolveProcess.java:211)
	at biz.aQute.resolve.ResolveProcess.resolveRequired(ResolveProcess.java:83)
	at biz.aQute.resolve.RunResolution.resolve(RunResolution.java:83)
	at biz.aQute.resolve.Bndrun.resolve(Bndrun.java:100)
	at biz.aQute.resolve.Bndrun.resolve(Bndrun.java:94)
	at okhttp3.osgi.OkHttpOsgiTest.testMainModuleWithSiblings(OkHttpOsgiTest.java:63)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.BlockJUnit4ClassRunner$1.evaluate(BlockJUnit4ClassRunner.java:100)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:366)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:103)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:63)
	at org.junit.runners.ParentRunner$4.run(ParentRunner.java:331)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:79)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:329)
	at org.junit.runners.ParentRunner.access$100(ParentRunner.java:66)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:293)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:413)
	at com.intellij.junit4.JUnit4IdeaTestRunner.startRunnerWithArgs(JUnit4IdeaTestRunner.java:69)
	at com.intellij.rt.junit.IdeaTestRunner$Repeater.startRunnerWithArgs(IdeaTestRunner.java:33)
	at com.intellij.rt.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:220)
	at com.intellij.rt.junit.JUnitStarter.main(JUnitStarter.java:53)
Caused by: org.apache.felix.resolver.reason.ReasonException: Unable to resolve <<INITIAL>>: missing requirement osgi.identity;filter:='(osgi.identity=com.squareup.okhttp3)'
	at org.apache.felix.resolver.Candidates$MissingRequirementError.toException(Candidates.java:1341)
	at org.apache.felix.resolver.ResolverImpl.doResolve(ResolverImpl.java:434)
	at org.apache.felix.resolver.ResolverImpl.resolve(ResolverImpl.java:421)
	at org.apache.felix.resolver.ResolverImpl.resolve(ResolverImpl.java:375)
	at biz.aQute.resolve.BndResolver.resolve(BndResolver.java:32)
	at biz.aQute.resolve.ResolveProcess.resolveRequired(ResolveProcess.java:81)
	... 30 more
```

swankjesse:
Okay, figured it out! I needed to run tests with Gradle, and I needed to add `com.oracle.svm.core.annotate` to our imported packages.

swankjesse:
Merged manually, including some small style changes. 74f97813705bd91c1d8a960362d4c9c045ac2ce2

Thanks for your patience and endurance on this!

swimmesberger:
Thanks everyone involved! @swankjesse FYI: In the final commit the CONTRIBUTING.md seems to be missing. I can also understand the concerns about using "kotlin-osgi-bundle" but we only use it in the osgiApi configuration which will only be added to the bnd classpath. Therefore bnd could potentially use classes/packages (to calculate stuff) not available in the "default" jar but in practice it simply leads to proper versioned packages which is always a win for OSGi.

swankjesse:
@swimmesberger the CONTRIBUTING change is deliberate. We've never had an external contributor introduce a new dependency and I think that documenting how to do it might imply that it was a thing we would expect or accept. OkHttp is very prudent with it's dependencies (2!) so I don't think we lose much here.

PR #6283, opened 2020-09-28
Add Monitoring Connections for Error Status Messages

bessbd:
As per https://tools.ietf.org/html/rfc2616#section-8.2.2 An
HTTP/1.1 (or later) client sending a message-body SHOULD
monitor the network connection for an error status while it
is transmitting the request.

Currently, okhttp implements a send request, then read response
if/when request was sent successfully.
This hides early responses such as 413 Payload Too Large
errors from clients.

This commit fixes that by adding a response read in case of an
exception happening while the request is being sent.

This closes square#1001 .

Related PR: https://github.com/square/okhttp/pull/6121

### DIFF
--- a/okhttp/src/test/java/okhttp3/internal/http/ThreadInterruptTest.java
+++ b/okhttp/src/test/java/okhttp3/internal/http/ThreadInterruptTest.java
@@ -17,12 +17,19 @@ package okhttp3.internal.http;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.OutputStream;
+import java.net.InetSocketAddress;
 import java.net.ServerSocket;
 import java.net.Socket;
+import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 import javax.annotation.Nullable;
 import javax.net.ServerSocketFactory;
 import javax.net.SocketFactory;
+
+import com.sun.net.httpserver.HttpExchange;
+import com.sun.net.httpserver.HttpHandler;
+import com.sun.net.httpserver.HttpServer;
 import okhttp3.Call;
 import okhttp3.DelegatingServerSocketFactory;
 import okhttp3.DelegatingSocketFactory;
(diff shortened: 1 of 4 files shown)


okhttp/src/test/java/okhttp3/internal/http/ThreadInterruptTest.java
```diff
@@ -17,12 +17,19 @@
...
+import java.net.InetSocketAddress;
 import java.net.ServerSocket;
 import java.net.Socket;
+import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 import javax.annotation.Nullable;
 import javax.net.ServerSocketFactory;
 import javax.net.SocketFactory;
+
+import com.sun.net.httpserver.HttpExchange;
```
swankjesse, line 30: ew

swankjesse (changes requested):
We might fix this, but this isn’t the right fix.

swankjesse:
....in particular, we need tests to confirm the connection is properly released and not re-used. Also, we use MockWebServer in this project!

bessbd:
Hi @swankjesse ,

Thank you for the quick response! Can you please elaborate on 
> but this isn’t the right fix

? Is there something fundamentally off with it or is it more of a "let's use MockWebServer and please add 1-2 more test cases"?

swankjesse:
I think the fix needs to move from `CallServerInterceptor` to `ExchangeCodec`. If we’re gonna do this, we should support HTTP/1 and HTTP/2 and make sure we leave the connections and streams in a healthy state.

yschimke:
@swankjesse Devil's advocate, since it doesn't introduce new API, just improves behaviour in a readable way for HTTP/1.1 should we adopt this short term until a proper full fix covering HTTP/2 also?

swankjesse:
I've run into an HTTP/2 variant of this problem and I'm going to attempt to fix that there. The thing I'm least comfortable with is lifecycle of the connection after we recover. It's broken for writing but not reading, and that isn't something we've modeled.

swankjesse:
Thanks for driving the work on this. I’ve created a PR that includes some ideas from this. Please review!
https://github.com/square/okhttp/pull/6295

PR #6287, opened 2020-09-29
Cache Corruption Test

yschimke:
Base for some corruption tests, but currently passing.

### DIFF
--- a/okhttp/src/test/java/okhttp3/CacheCorruptionTest.kt
+++ b/okhttp/src/test/java/okhttp3/CacheCorruptionTest.kt
@@ -1,3 +1,18 @@
+/*
+ * Copyright (C) 2020 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package okhttp3
 
 import okhttp3.internal.io.InMemoryFileSystem
@@ -63,6 +78,7 @@ class CacheCorruptionTest {
   @Test fun corruptedCipher() {
     val response = testCorruptingCache {
       corruptMetadata {
+        // mess with cipher suite
         it.replace("TLS_", "SLT_")
       }
     }
@@ -95,7 +111,6 @@ class CacheCorruptionTest {
 
     val contents = metadataBuffer!!.peek().readUtf8()
 
-    // mess with cipher suite
     metadataBuffer.clear()
     metadataBuffer.writeUtf8(corruptor(contents))
   }
--- a/okhttp/src/test/java/okhttp3/CacheCorruptionTest.kt
+++ b/okhttp/src/test/java/okhttp3/CacheCorruptionTest.kt
@@ -5,13 +5,12 @@ import okhttp3.mockwebserver.MockResponse
 import okhttp3.mockwebserver.MockWebServer
 import okhttp3.testing.PlatformRule
 import okhttp3.tls.internal.TlsUtil.localhost
-import org.assertj.core.api.Assertions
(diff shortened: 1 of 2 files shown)


okhttp/src/test/java/okhttp3/CacheCorruptionTest.kt
```diff
@@ -0,0 +1,109 @@
...
+    val response1: Response = client.newCall(request).execute()
+    val bodySource = response1.body!!.source()
+    Assertions.assertThat(bodySource.readUtf8()).isEqualTo("ABC")
+
+    val metadataFile = fileSystem.files.keys.find { it.name.endsWith(".0") }
+    val metadataBuffer = fileSystem.files[metadataFile]
+
+    val contents = metadataBuffer!!.peek().readUtf8()
+
+    metadataBuffer.writeUtf8(contents.substring(0, contents.length / 4))
```
adriancole, line 85: worth a comment? my guess is this is where we are corrupting.. and later we are expecting the request to return the same string anyway... because we are resilient to corrupted entries?

okhttp/src/test/java/okhttp3/CacheCorruptionTest.kt
```diff
@@ -0,0 +1,109 @@
...
+    Assertions.assertThat(cache.requestCount()).isEqualTo(2)
+    Assertions.assertThat(cache.networkCount()).isEqualTo(1)
+    Assertions.assertThat(cache.hitCount()).isEqualTo(1)
+  }
+
+  /**
+   * @param delta the offset from the current date to use. Negative values yield dates in the past;
+   * positive values yield dates in the future.
+   */
+  private fun formatDate(delta: Long, timeUnit: TimeUnit): String? {
```
adriancole, line 100: this is handy

adriancole (commented):
drive by notes

okhttp/src/test/java/okhttp3/CacheCorruptionTest.kt
yschimke: copy and pasted from the larger cache test.  I have problems dealing with massive source files, they break my brain so I added this test file to cover any fuzzing or deliberate break the cache scenarios.

okhttp/src/test/java/okhttp3/CacheCorruptionTest.kt
yschimke: Copying the PR notes

> Base for some corruption tests, but currently passing.

I want to try to reproduce every painful cache corruption we can here, make sure the cache always allows for forward progress.

adriancole: sorry I meant a comment to explain this corruption case.

ex. 
```suggestion
    // We expect cached request to serve even when the metadata file is truncated
    metadataBuffer.writeUtf8(contents.substring(0, contents.length / 4))
```

It is a bit more complex I guess to explain how it is able to serve the payload when it is corrupt, though it would be nice to say that if it is short enough to pointer in a sentence.

I've been out a while so have the benefit of low context.. some shared implementation knowledge I'll be able to point out until I too know too much :D

okhttp/src/test/java/okhttp3/CacheCorruptionTest.kt
```diff
@@ -0,0 +1,141 @@
+package okhttp3
```
swankjesse, line 1: gotta spotless our copyright thingies

okhttp/src/test/java/okhttp3/CacheCorruptionTest.kt
```diff
@@ -0,0 +1,141 @@
...
+    assertThat(cache.hitCount()).isEqualTo(0)
+  }
+
+  private fun corruptMetadata(corruptor: (String) -> String) {
+    val metadataFile = fileSystem.files.keys.find { it.name.endsWith(".0") }
+    val metadataBuffer = fileSystem.files[metadataFile]
+
+    val contents = metadataBuffer!!.peek().readUtf8()
+
+    // mess with cipher suite
```
swankjesse, line 98: comment is no longer valid?

swankjesse (approved):

okhttp/src/test/java/okhttp3/CacheCorruptionTest.kt
yschimke: MVP would just be a IFTTT script that comments on my PR when I add a new file matching *.kt.  Also asks me to use complete sentences in comments. Asks me not to inline expressions for function declarations.

PR #6288, opened 2020-09-29
Lints android-test

adriancole:
When complete, this should allow folks running the android-test more success.

Depends on feedback in general and specifically here https://github.com/square/okhttp/pull/6008#discussion_r496553972

### DIFF
--- a/android-test/README.md
+++ b/android-test/README.md
@@ -3,10 +3,17 @@ Android Test
 
 A gradle module for running Android instrumentation tests on a device or emulator.
 
-1. Run an Emulator using Android Studio or from command line.
+1. Add an Emulator named `pixel5`, if you don't already have one
 
 ```
-$ emulator -no-window @pixel5
+$ sdkmanager --install "system-images;android-29;google_apis;x86"
+$ echo "no" | avdmanager --verbose create avd --force --name "pixel5" --device "pixel" --package "system-images;android-29;google_apis;x86" --tag "google_apis" --abi "x86"
+```
+
+2. Run an Emulator using Android Studio or from command line.
+
+```
+$ emulator -no-window -no-snapshot-load @pixel5
 ```
 
 2. Turn on logs with logcat
--- a/android-test/src/androidTest/java/okhttp/android/test/OkHttpTest.kt
+++ b/android-test/src/androidTest/java/okhttp/android/test/OkHttpTest.kt
@@ -196,8 +196,14 @@ class OkHttpTest {
         assertEquals(200, response.code)
         // see https://github.com/google/conscrypt/blob/b9463b2f74df42d85c73715a5f19e005dfb7b802/android/src/main/java/org/conscrypt/Platform.java#L613
         when {
+            // TODO: how do we determine when we expect useEngineSocket to be true, resulting in org.conscrypt.Java8EngineSocket?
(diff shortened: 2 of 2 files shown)


adriancole:
This is me warming back up, so wanted to improve the setup instructions. If anything is worse than before please note!

android-test/src/androidTest/java/okhttp/android/test/OkHttpTest.kt
```diff
@@ -196,8 +196,14 @@ class OkHttpTest {
         assertEquals(200, response.code)
         // see https://github.com/google/conscrypt/blob/b9463b2f74df42d85c73715a5f19e005dfb7b802/android/src/main/java/org/conscrypt/Platform.java#L613
         when {
+            // TODO: how do we determine when we expect useEngineSocket to be true, resulting in org.conscrypt.Java8EngineSocket?
```
adriancole, line 199: this is maybe not worth doing, if the below or similar if branch is ok. Alternatively, we can switch to assertJ and use a startsWith assertion

android-test/README.md
```diff
@@ -3,10 +3,17 @@ Android Test
 
 A gradle module for running Android instrumentation tests on a device or emulator.
 
-1. Run an Emulator using Android Studio or from command line.
+1. Add an Emulator named `pixel5`, if you don't already have one
```
yschimke, line 6: Love this!

android-test/src/androidTest/java/okhttp/android/test/OkHttpTest.kt
yschimke: Are you running these as normal JDK tests in Intellij?  They should hopefully be running on device or in an emulator.

adriancole: sorry I guess I had a pretty horrible comment there. I will put on the test itself that this is going to execute in the emulator and strip notions of JDK.

meanwhile my version of pixel5 fails on this unless I permit this to resolve to Java8EngineSocket. This is exactly the cognitive thing I was hoping you could help with.. is there a way to tell as we do in the branch on `Build.VERSION.SDK_INT` whether this will be Java8EngineSocket or Java8FileDescriptorSocket? I looked at the code and it seems to be effected by a flag `useEngineSocket`. FWIW you can see my emulator is using sdk 29..

I'll update the comments after you teach me more :D thanks for the help

yschimke: I think this shows a bug in the test setup previously, sorry.  I think we just want org.conscrypt.Java8EngineSocket here.

https://groups.google.com/g/conscrypt/c/Ajrexsz_BQc

```
assertEquals("org.conscrypt.Java8EngineSocket", socketClass)
```

yschimke:
Big sincere thanks from me for the critical attention here, I've been lazy in running these.  Ideally we will eventually move CI to github and run these tests on every PR and master nightly.

https://github.com/square/retrofit/blob/master/.github/workflows/build.yml

adriancole:
I will save the logging related glitch to another PR as I want to think more about it..

yschimke (approved):

adriancole:
Thanks for the help @yschimke!

PR #6292, opened 2020-10-01
Graal support via optional module

yschimke:


okcurl/src/test/java/okhttp3/curl/GenerateReflectionConfig.kt
```diff
@@ -0,0 +1,28 @@
...
+ * limitations under the License.
+ */
+package okhttp3.curl
+
+import picocli.CommandLine.Model.CommandSpec.*
+import picocli.codegen.aot.graalvm.ReflectionConfigGenerator
+import java.io.File
+
+/**
+ * Manual process to update reflect-config.json
```
yschimke, line 23: We can move this into okcurl/build.gradle if we care enough.

okhttp-graal/src/main/kotlin/okhttp3/graal/svm.kt
```diff
@@ -0,0 +1,66 @@
...
+import okhttp3.internal.platform.BouncyCastlePlatform
+import okhttp3.internal.platform.ConscryptPlatform
+import okhttp3.internal.platform.Jdk8WithJettyBootPlatform
+import okhttp3.internal.platform.Jdk9Platform
+import okhttp3.internal.platform.OpenJSSEPlatform
+import okhttp3.internal.platform.Platform
+
+@TargetClass(AndroidPlatform::class)
+@Delete
+class TargetAndroidPlatform {
```
yschimke, line 32: The goal of these are to get to the point that "allow-incomplete-classpath" isn't required.  But I didn't set that as complete goal yet.

yschimke: Currently would fail with 

```
com.oracle.graal.pointsto.constraints.UnresolvedElementException: Discovered unresolved type during parsing: android.os.Build$VERSION. To diagnose the issue you can use the --allow-incomplete-classpath option. The missing type is then reported at run time when it is accessed the first time.
```

okcurl/build.gradle
```diff
@@ -1,3 +1,9 @@
+plugins {
+  id "com.palantir.graal" version "0.7.1-20-g113a84d"
+}
+
+import org.apache.tools.ant.taskdefs.condition.Os
```
swankjesse, line 5: ant?!

okcurl/build.gradle
```diff
@@ -35,3 +46,21 @@ dependencies {
...
   mergeServiceFiles()
 }
+
+graal {
+  mainClass "okhttp3.curl.Main"
+  outputName "okcurl"
+  graalVersion "20.2.0"
+  javaVersion "11"
+
+  option "--enable-https"
```
swankjesse, line 56: this is a graal option, not an OkHttp option?!

okcurl/build.gradle
```diff
@@ -35,3 +46,21 @@ dependencies {
...
+  option "--no-fallback"
+  option "--allow-incomplete-classpath"
+
+  if (Os.isFamily(Os.FAMILY_WINDOWS)) {
+    // May be possible without, but autodetection is problematic on Windows 10
+    // see https://github.com/palantir/gradle-graal
+    // see https://www.graalvm.org/docs/reference-manual/native-image/#prerequisites
+    windowsVsVarsPath("C:\\Program Files (x86)\\Microsoft Visual Studio\\2019\\BuildTools\\VC\\Auxiliary\\Build\\vcvars64.bat")
```
swankjesse, line 64: 🤮 too bad

okcurl/src/main/resources/META-INF/native-image/okhttp3/okcurl/reflect-config.json
```diff
@@ -0,0 +1,45 @@
...
+  },
+  {
+    "name" : "okhttp3.curl.Main$VersionProvider",
+    "allDeclaredConstructors" : true,
+    "allPublicConstructors" : true,
+    "allDeclaredMethods" : true,
+    "allPublicMethods" : true
+  },
+  {
+    "name" : "picocli.CommandLine$AutoHelpMixin",
```
swankjesse, line 35: makes me wonder if there’s a codegen-based alternative!

okcurl/src/test/java/okhttp3/curl/GenerateReflectionConfig.kt
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okhttp3.curl
+
+import picocli.CommandLine.Model.CommandSpec.*
```
swankjesse, line 18: nit: wildcard

okcurl/src/test/java/okhttp3/curl/GenerateReflectionConfig.kt
swankjesse: This works for me.

okcurl/src/test/java/okhttp3/curl/GenerateReflectionConfig.kt
```diff
@@ -0,0 +1,28 @@
...
+import picocli.codegen.aot.graalvm.ReflectionConfigGenerator
+import java.io.File
+
+/**
+ * Manual process to update reflect-config.json
+ */
+fun main() {
```
swankjesse, line 25: internal ?

okhttp-graal/README.md
```diff
@@ -0,0 +1,6 @@
+OkHttp Graal Support
+====================
+
+This module is a helper module for building native images
+with Graal https://www.graalvm.org/.  Including this module should
+result in automatic support for native images within the OkHttp libraries.
```
swankjesse, line 6: OMG this is exactly what I want!

😎

okhttp-graal/build.gradle
```diff
@@ -0,0 +1,28 @@
...
+    attributes('Automatic-Module-Name': 'okhttp3.graal')
+  }
+}
+
+dependencies {
+  api project(':okhttp')
+
+  compileOnly("org.graalvm.nativeimage:svm:20.2.0") {
+    // https://youtrack.jetbrains.com/issue/KT-29513
+    exclude group: 'org.graalvm.nativeimage'
```
swankjesse, line 12: does this prevent us from using native images? or something else?

okhttp-graal/src/main/kotlin/okhttp3/graal/svm.kt
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okhttp3.graal
+
+import com.oracle.svm.core.annotate.Delete
```
swankjesse, line 18: TIL svm = substratevm

okhttp-graal/src/main/kotlin/okhttp3/graal/svm.kt
```diff
@@ -0,0 +1,66 @@
...
+}
+
+@TargetClass(OpenJSSEPlatform::class)
+@Delete
+class TargetOpenJSSEPlatform {
+}
+
+@TargetClass(Platform.Companion::class)
+class TargetPlatform {
+  @Substitute
```
swankjesse, line 62: amaze

swankjesse (approved):

swankjesse:
Since it only adds one class and no classpath dependencies, should we just add `svm.kt` to the main OkHttp artifact? That way OkHttp works with graal out of the box, no extra dependency necessary?

okhttp-graal/README.md
```diff
@@ -0,0 +1,6 @@
+OkHttp Graal Support
+====================
+
+This module is a helper module for building native images
```
adriancole, line 4: is it worthwhile to include an example to make an okcurl native image?

okcurl/build.gradle
```diff
@@ -1,3 +1,9 @@
+plugins {
+  id "com.palantir.graal" version "0.7.1-20-g113a84d"
```
schlosna, line 2: Just a heads up this is technically a snapshot build that was accidentally publishing (see https://github.com/palantir/gradle-graal/pull/392 ). The [only changes since 0.7.1 release are updates to build plugins](https://github.com/palantir/gradle-graal/compare/0.7.1...develop), so I think you could use 0.7.1 here if you like until a 0.7.2 or 0.8.0 are cut.


```suggestion
  id "com.palantir.graal" version "0.7.1"
```

okcurl/build.gradle
yschimke: It's apparently still the canonical way I think.

okcurl/build.gradle
yschimke: It makes sense, do you want a keystore and native crypto libraries bundled?

okcurl/src/test/java/okhttp3/curl/GenerateReflectionConfig.kt
yschimke: It's a test class?

okcurl/build.gradle
yschimke: https://stackoverflow.com/questions/11235614/how-to-detect-the-current-os-from-gradle

okcurl/src/main/resources/META-INF/native-image/okhttp3/okcurl/reflect-config.json
yschimke: It's meant to work, but since it was out of scope (okcurl) of the main goal (okhttp) of the PR, I went with this as a first version.  I'll follow up with a better fix, and fix logging in okcurl, but that's low priority I think.

yschimke: https://github.com/remkop/picocli/tree/master/picocli-codegen

okhttp-graal/build.gradle
yschimke: kapt scans a compressed file that graal stores in the maven repository and fails.  This just says that we only want the annotations to compile against, we don't need the graal runtime.

yschimke:
> Since it only adds one class and no classpath dependencies, should we just add `svm.kt` to the main OkHttp artifact? That way OkHttp works with graal out of the box, no extra dependency necessary?

Yep, but if we have issues in non core modules then we will have this setup for each.  Let's go with your plan.

Types of failures when you get things wrong with graal, i.e. we can expect a few bug reports about things we haven't got perfect in other modules.

1) Use reflection or classpath resources but it hasn't retained those.
2) Your code at runtime on another machine uses state from the build machine at runtime.
3) You hardcode something like a UUID or secure random // return 42

adriancole:
Thanks for getting this started!

okcurl/build.gradle
yschimke: Fixed.

okcurl/build.gradle
swankjesse: Yep!

okcurl/src/test/java/okhttp3/curl/GenerateReflectionConfig.kt
swankjesse: Never mind me!

okhttp-graal/build.gradle
JakeWharton: Should we be filing a Kotlin bug for that?

yschimke: It's linked in the bug above

Fixed for 1.4 in https://youtrack.jetbrains.com/issue/KT-41313, so maybe I can remove now.

PR #6293, opened 2020-10-01
Confirm we can read a response that completed before RST_STREAM

swankjesse:
The server can reject the request without breaking the response.


zhxnlai (approved):
💯 💯

PR #6295, opened 2020-10-02
Read the response even if writing the request fails

swankjesse:
Closes: #1001

### DIFF
--- a/okhttp/src/main/kotlin/okhttp3/internal/http/CallServerInterceptor.kt
+++ b/okhttp/src/main/kotlin/okhttp3/internal/http/CallServerInterceptor.kt
@@ -20,6 +20,7 @@ import java.net.ProtocolException
 import okhttp3.Interceptor
 import okhttp3.Response
 import okhttp3.internal.EMPTY_RESPONSE
+import okhttp3.internal.http2.ConnectionShutdownException
 import okio.buffer
 
 /** This is the last interceptor in the chain. It makes a network call to the server. */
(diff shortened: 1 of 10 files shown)


okhttp/src/main/kotlin/okhttp3/internal/http/CallServerInterceptor.kt
```diff
@@ -20,6 +20,7 @@ import java.net.ProtocolException
 import okhttp3.Interceptor
 import okhttp3.Response
 import okhttp3.internal.EMPTY_RESPONSE
+import okhttp3.internal.http2.ConnectionShutdownException
 import okio.buffer
```
swankjesse, line 24: the diffs on this file are worth viewing in 'ignore whitespace' mode because there’s a lot of indentation change

okhttp/src/main/kotlin/okhttp3/internal/http/CallServerInterceptor.kt
```diff
@@ -33,98 +34,118 @@ class CallServerInterceptor(private val forWebSocket: Boolean) : Interceptor {
...
+        exchange.noNewExchangesOnConnection()
+      }
+      if ((code == 204 || code == 205) && response.body?.contentLength() ?: -1L > 0L) {
+        throw ProtocolException(
+            "HTTP $code had non-zero Content-Length: ${response.body?.contentLength()}")
+      }
+      return response
+    } catch (e: IOException) {
+      if (sendRequestException != null) {
+        sendRequestException.addSuppressed(e)
```
yschimke, line 145: Nice!

okhttp/src/test/java/okhttp3/ServerTruncatesRequestTest.kt
```diff
@@ -0,0 +1,268 @@
...
+
+    server.enqueue(mockResponse)
+
+    val call = client.newCall(Request.Builder()
+        .url(server.url("/"))
+        .post(SlowRequestBody)
+        .build())
+
+    call.execute().use { response ->
+      assertThat(response.body!!.string()).isEqualTo("abc")
```
yschimke, line 186: This is nice!

okhttp/src/test/java/okhttp3/EventListenerTest.java
```diff
@@ -1118,7 +1118,7 @@ private void writeChunk(BufferedSink sink) throws IOException {
     assertThat(listener.recordedEventTypes()).containsExactly(
         "CallStart", "ProxySelectStart", "ProxySelectEnd", "DnsStart", "DnsEnd", "ConnectStart",
         "ConnectEnd", "ConnectionAcquired", "RequestHeadersStart", "RequestHeadersEnd",
-        "RequestBodyStart", "RequestFailed", "ConnectionReleased", "CallFailed");
+        "RequestBodyStart", "RequestFailed", "ResponseFailed", "ConnectionReleased", "CallFailed");
```
yschimke, line 1121: Should we have a test for this where it does successfully read the response?

yschimke (approved):

okhttp/src/test/java/okhttp3/ServerTruncatesRequestTest.kt
```diff
@@ -0,0 +1,268 @@
...
+    if (https) {
+      expectedEvents += "SecureConnectStart"
+      expectedEvents += "SecureConnectEnd"
+    }
+    expectedEvents += "ConnectEnd"
+    expectedEvents += "ConnectionAcquired"
+    expectedEvents += "RequestHeadersStart"
+    expectedEvents += "RequestHeadersEnd"
+    expectedEvents += "RequestBodyStart"
+    expectedEvents += "RequestFailed"
```
adriancole, line 103: maybe a trailing comment here highlighting that the important part is we read the response headers and body even on error.

okhttp/src/test/java/okhttp3/EventListenerTest.java
swankjesse: Yep, the new test case includes that.

yschimke:
@swankjesse any chance this fixes issues like https://github.com/  PhilippC/keepass2android/issues/747

bessbd:
I've just verified this fix E2E in a project that uses okhttp. Thank you, @swankjesse

swankjesse:
@ricellis 5.0.0-alpha.3 is out as of yesterday and it has it. I don't think we'll backport to 4.x because it's a riskier change. (We're tracking one regression related to internet disconnects.)

ricellis:
>I don't think we'll backport to 4.x because it's a riskier change.

Ah that's a shame because we missed 3.x to 4.x for the [same reason once before](https://github.com/square/okhttp/pull/6121#issuecomment-640463624), guess this fix has unlucky timing!

Thank you for the update.

yschimke:
I'm assuming that we would want to also continue reading in the case of

> 431 "Request Header Fields Too Large"

Where we didn't complete sending headers?

PR #6304, opened 2020-10-02
Enable GitHub Code Scanning

ddz:
Created a `codeql-analysis.yml` to configure GitHub workflow for code scanning.
* Added Setup Java action to override default Java 8.
* Added custom build command from CircleCI yml commented out in case CodeQL autobuild fails in the future.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,76 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+name: "CodeQL"
+
+on:
+  push:
+    branches: [master]
+  pull_request:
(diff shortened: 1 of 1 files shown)


swankjesse (approved):

yschimke:
Is this posting to here?  Does this support kotlin?

https://lgtm.com/projects/g/square/okhttp/?mode=list
https://lgtm.com/query/project:7870357/lang:java/

swankjesse:
Perhaps this doesn't help with .kt files?
https://help.semmle.com/codeql/supported-languages-and-frameworks.html

ddz:
Yes, doesn't look like Kotlin is supported yet, but [it appears to be on their radar](https://news.ycombinator.com/item?id=24648281). So this scanning will only cover samples, mockwebserver, and some test code until then. It's definitely debatable whether it's worth the additional CI time until then.

yschimke:
> It's definitely debatable whether it's worth the additional CI time until then.

Seems like a good initiative to support and will be nice when it becomes active.

Issue #6299, opened 2020-10-02
R8 warning during the build OkHttp 4.9.0


ahulyk:
I'm getting the following warning during build:
`Missing class: org.conscrypt.ConscryptHostnameVerifier`

My proguard config:
```
# JSR 305 annotations are for embedding nullability information.
-dontwarn javax.annotation.**

# A resource is loaded with a relative path so the package of this class must be preserved.
-keepnames class okhttp3.internal.publicsuffix.PublicSuffixDatabase

# Animal Sniffer compileOnly dependency to ensure APIs are compatible with older versions of Java.
-dontwarn org.codehaus.mojo.animal_sniffer.*

# OkHttp platform used only on JVM and when Conscrypt dependency is available.
-dontwarn okhttp3.internal.platform.ConscryptPlatform
-dontwarn okhttp3.internal.platform.Android10Platform
```

Is it safe enough to add:
`-dontwarn org.conscrypt.ConscryptHostnameVerifier` ?

yschimke:
I assume we are just missing this from 

https://github.com/square/okhttp/blob/master/okhttp/src/main/resources/META-INF/proguard/okhttp3.pro

Ideally we would flag these in our own builds if this is missing but changes slowly over time.

VincentJoshuaET:
![image](https://user-images.githubusercontent.com/19756326/106412007-96e3c380-6481-11eb-8665-f27938f33d03.png)
I still have this issue. It looks like the line is not there on the library's ProGuard file

ahulyk:
> 
> I still have this issue. It looks like the line is not there on the library's ProGuard file

This fix has been added to the 4.10 milestone. Please try 4.10

VincentJoshuaET:
Couldn't this be in 4.9.2 stable? Looks like 4.10 is now 5.x which is still alpha

PR #6308, opened 2020-10-06
R8 warning during the build OkHttp 4.9.0 #6299

ahulyk:
Update Proguard rule to hide the following warning:
`Missing class: org.conscrypt.ConscryptHostnameVerifier`

### DIFF
--- a/okhttp/src/main/resources/META-INF/proguard/okhttp3.pro
+++ b/okhttp/src/main/resources/META-INF/proguard/okhttp3.pro
@@ -9,3 +9,4 @@
 
 # OkHttp platform used only on JVM and when Conscrypt dependency is available.
 -dontwarn okhttp3.internal.platform.ConscryptPlatform
+-dontwarn org.conscrypt.ConscryptHostnameVerifier


okhttp/src/main/resources/META-INF/proguard/okhttp3.pro
```diff
@@ -9,3 +9,4 @@
 
 # OkHttp platform used only on JVM and when Conscrypt dependency is available.
 -dontwarn okhttp3.internal.platform.ConscryptPlatform
+-dontwarn org.conscrypt.ConscryptHostnameVerifier
```
swankjesse, line 12: strange; does dontwarn cover the files that shouldn’t warn, or the imports of those files? This file is only referenced from the file that precedes it

ahulyk: I did an extra check. And I'm getting the following warning:
> Task :app:minify******ReleaseWithR8
AGPBI: {"kind":"warning","text":"Missing class: org.conscrypt.ConscryptHostnameVerifier","sources":[{}],"tool":"R8"}

swankjesse (approved):

swankjesse:
Thanks!

PR #6311, opened 2020-10-08
Only support ALPN Boot on most recent JDK 8 builds

yschimke:
For discussion - may as well simplify builds at this point.  Anyone hitting JDK/JSSE related issues with HTTP/2 will be asked to upgrade and will probably find that they have the backported support for ALPN anyway.  So reducing surface area for CI and build complexity.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -202,15 +202,13 @@ subprojects { project ->
 
   if (platform == "jdk8alpn") {
     // Add alpn-boot on Java 8 so we can use HTTP/2 without a stable API.
-    def alpnBootVersion = alpnBootVersion()
-    if (alpnBootVersion != null) {
-      dependencies {
-        testCompile "org.mortbay.jetty.alpn:alpn-boot:$alpnBootVersion"
-      }
-      def alpnBootJar = configurations.testCompile.find { it.name.startsWith("alpn-boot-") }
-      test {
-        jvmArgs += "-Xbootclasspath/p:${alpnBootJar}"
-      }
+    def alpnBootVersion = '8.1.13.v20181017'
+    dependencies {
+      testCompile "org.mortbay.jetty.alpn:alpn-boot:$alpnBootVersion"
+    }
+    def alpnBootJar = configurations.testCompile.find { it.name.startsWith("alpn-boot-") }
+    test {
+      jvmArgs += "-Xbootclasspath/p:${alpnBootJar}"
     }
   } else if (platform == "conscrypt") {
     dependencies {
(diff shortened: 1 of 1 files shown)


adriancole:
To try to understand, you are saying we retain the alpn boot, but only recent versions of it, vs constrain the JDK to a version that obviates it? https://www.eclipse.org/jetty/documentation/current/alpn-chapter.html#alpn-openjdk8

yschimke:
@adriancole "For discussion"

So facts
- 8u261 and above doesn't need and disables alpn boot, we will never maintain this list
- We don't support running builds across all these versions 
- having an alpn boot in the classpath is now likely harmless as it disables itself (may corrupt method signatures I guess)
- we can test with 3.12.x or 4.9.1 if we want to see how it would have worked

There is probably numerous test failures if you run against an older version with head

```
    case 31..50:
      return '8.1.3.v20150130'
```

Proposal
- Keep simplifying the builds so anyone doing a CI migration has less to deal with

What's your take?

adriancole:
It is an interesting topic, which matrix of build envs are given a warranty by this project?

What you've mentioned, just trimming to what seems like minimum update 191 is implicit. If there was something like maven enforcer, we could also make the build fail in the same change if it were less than that.

Ex. I have a project that technically can use Java 1.7 but it would be crap if you used it and didn't know the minimum patch for it to work.

```
                <requireJavaVersion>
                  <!-- 1.7.0_80 to prevent a compile bug accessing CrossOrigin.
                       JDK 11 is the last version to compile Java 6 bytecode. -->
                  <version>[1.7.0_80,12)</version>
                </requireJavaVersion>
```

TL'DR' I think I would suggest making the JDK version an explicit support thing, and this subordinate to that. 

Ex. We support minimum JDK 8u191 (for some temporary reason because 8k forks and no idea how many can move to 8u261). Put that minimum threshold in the build itself and crash early if not met.

yschimke:
Yeah, I'm happy with that.  Will change after work day.

To be clear we are retaining this functionality for the mythical user who

- Must deploy 4.10.x+ with JDK8u < 191
- Can't upgrade their JDK8 instance
- But must upgrade their OkHttp

adriancole:
ps sorry I didn't do homework. It looks like we already separate out compile JDK (currently 11) from the test one.

https://github.com/square/okhttp/blob/master/.circleci/config.yml#L140-L153

I guess the feature of testing alpn-boot is itself a separate scenario and if we removed this from master, it would mean it is left to trust.

Another way to handle platform things like this would be heavier but more contained. For example, have a testcontainers invocation that includes the "spot tests" we expect to be tricky on certain platforms vs running all tests on all JDK matrixes. This would be a fairly large infrastructure dig, so can see the value of leaving things as they are.

adriancole:
wires kindof crossed, but they heavy way out would be to make a specific testcontainers test for that scenario, pulling a zulu docker image of the old version or something, but the driver of it here could be normal JDK 11.

yschimke:
Internally in the code we do this via assumptions now.  

```
  @Flaky
  @Test public void missingPongsFailsConnection() throws Exception {
    if (protocol == Protocol.HTTP_2) {
      // https://github.com/square/okhttp/issues/5221
      platform.expectFailureOnJdkVersion(12);
    }
```

swankjesse:
Got any recent issues that this would simplify?

I don't love this because it’s our benefits are unclear, but the imposition on the user is nontrivial: many developers don't have direct control over which JDK8 they're deploying on.

I'd rather drop alpn-boot completely, but even that is an ‘aint broke don't fix it’ problem to me.

yschimke:
For discussion -> discussed.

yschimke:
Main simplification would be for whenever someone tackles the CI migration.  And slightly closer to I can understand the build setup because it fits on one screen.

Issue #6296, opened 2020-10-02
Run JUnit tests with GraalVM


swankjesse:
It's quite possible there's gaps on OkHttp when executed on GraalVM. 

We should do the Gradle work to run our test suite on it. It appears possible, though awkward.
https://github.com/spring-projects-experimental/spring-graalvm-native/issues/22

Spring uses JUnit 5 in the above example. If doing our tests on GraalVM requires upgrading to JUnit 5, we should probably do that too.

yschimke:
> Check if that's interesting to use GraalVM capability to create shared libraries + calling them from Java code

Interesting...

yschimke:
Wow

```
Test run finished after 966 ms
[       101 containers found      ]
[         2 containers skipped    ]
[        99 containers started    ]
[         1 containers aborted    ]
[        98 containers successful ]
[         0 containers failed     ]
[      1096 tests found           ]
[        68 tests skipped         ]
[      1025 tests started         ]
[         0 tests aborted         ]
[      1025 tests successful      ]
[         0 tests failed          ]
```

yschimke:
Do you want any help with any aspect of this? exploring best ways to setup reflection with junit? adding support in PlatformRule to use NativeImageDetector.inNativeImage()? fixing tests?

I'll assume no, but happy to chip in when and if you think I can help.

swankjesse:
Yeah I'd love help.

In hacking on this yesterday I learned just how annoying it is to pull one module’s tests into another module's dependencies. My conclusion was to make this into a task we can use from any module, and not a separate module. Which led to this PR:
https://github.com/palantir/gradle-graal/pull/393

The next natural step I see while waiting on the above is to temporarily manually move a few tests into native-image-tests/src/main/java to execute 'em in Graal. Presumably there's gonna be challenges there! Getting CallTest passing will be a nice milestone.

yschimke:
@swankjesse +1 - graal itself doesn't add additional modules, just a target for running nativeImage.  Is that an option here?

nativeImage -> jar
as
graalTest -> test?

swankjesse:
Yep!

yschimke:
POC graal test with okhttp and mockwebserver with only Junit5

https://github.com/yschimke/okhttp/pull/5


```
$  ./native-image-tests/build/graal/ConsoleLauncher --select-class okhttp3.CallKotlinTest

Thanks for using JUnit! Support its development at https://junit.org/sponsoring

╷
[...]

PR #6313, opened 2020-10-09
Initial steps towards running tests in Graalvm

swankjesse:
I want to confirm all of OkHttp works with Graalvm, and I'm hoping
to demonstrate that by getting the entire test suite running inside
Graalvm.

https://github.com/square/okhttp/issues/6296

 * [ ] Configure seeding to identify tests without executing them (ignore all somehow?)
 * [ ] Seed using the Graal from the Graal Gradle plugin, not the local machine
 * [ ] Run OkHttp tests, not SampleTest (will need the vintage engine, since we’re JUnit 4)

If we get that all working, we’re good. For bonus points it’d be nice to upstream this mechanism into the Graal plugin, or something standalone. It’d be quite rad to make Graalvm unit testing not dramatically more complicated than regular unit testing.

### DIFF
--- a/native-image-tests/build.gradle
+++ b/native-image-tests/build.gradle
@@ -1,0 +1,48 @@
+plugins {
+  id "com.palantir.graal" version "0.7.1"
+}
+
+dependencies {
+  implementation deps.assertj
+  implementation "org.junit.jupiter:junit-jupiter-api:5.7.0"
+  implementation "org.junit.jupiter:junit-jupiter-engine:5.7.0"
+  implementation "org.junit.platform:junit-platform-console:1.7.0"
+
+  compileOnly deps.jsr305
+}
+
+animalsniffer {
+  ignoreFailures = true
+}
+
+def graalHome = "/Library/Java/JavaVirtualMachines/graalvm-ce-java11-20.2.0/Contents/Home"
+def graalvmResources = project.file("$buildDir/graalvm/resources")
+def graalvmConfig = project.file("$graalvmResources/META-INF/native-image")
+graalvmConfig.mkdirs()
+
+sourceSets {
+  main.resources.srcDirs += graalvmResources
+}
+
+task seedGraalvmConfig(type: Exec) {
+  dependsOn("classes")
+  commandLine(
+    "$graalHome/bin/java",
+    "-agentlib:native-image-agent=config-output-dir=$graalvmConfig",
+    "-classpath", sourceSets.main.runtimeClasspath.getAsPath(),
+    "org.junit.platform.console.ConsoleLauncher",
+    "--select-class", "okhttp3.SampleTest"
+  )
+}
+
+graal {
+  mainClass "org.junit.platform.console.ConsoleLauncher"
+  outputName "ConsoleLauncher"
+  graalVersion "20.2.0"
+  javaVersion "11"
+
+  option "--enable-https"
+  option "--no-fallback"
+  option "--allow-incomplete-classpath"
+  option "--report-unsupported-elements-at-runtime"
+}
--- a/native-image-tests/README.md
+++ b/native-image-tests/README.md
@@ -1,0 +1,55 @@
+Native Image Tests
+==================
+
+This executes OkHttp's test suite inside a Graalvm image.
+
+Seeding Graalvm Config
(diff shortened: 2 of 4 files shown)


swankjesse:
if you’re trying this out, you’ll need this too:

```
cd /Library/Java/JavaVirtualMachines/graalvm-ce-java11-20.2.0/Contents/Home/
sudo gu install native-image
```

native-image-tests/build.gradle
```diff
@@ -0,0 +1,48 @@
...
+  main.resources.srcDirs += graalvmResources
+}
+
+task seedGraalvmConfig(type: Exec) {
+  dependsOn("classes")
+  commandLine(
+    "$graalHome/bin/java",
+    "-agentlib:native-image-agent=config-output-dir=$graalvmConfig",
+    "-classpath", sourceSets.main.runtimeClasspath.getAsPath(),
+    "org.junit.platform.console.ConsoleLauncher",
```
yschimke, line 33: Instead of this approach of running twice, the technique |I've used is

https://github.com/yschimke/okurl/blob/master/src/main/kotlin/com/baulsupp/okurl/graal/RuntimeReflectionRegistrationFeature.kt

```
@AutomaticFeature
internal class RuntimeReflectionRegistrationFeature : Feature {
  override fun beforeAnalysis(access: BeforeAnalysisAccess) {
    try {
      val pkg = "com.baulsupp.okurl"
      ClassGraph()
//      .verbose() // Log to stderr
        .enableAnnotationInfo()
        .enableMethodInfo()
        .enableClassInfo() // Scan classes, methods, fields, annotations
        .acceptPackages(pkg) // Scan com.xyz and subpackages (omit to scan all packages)
        .scan()
        .use { scanResult ->                    // Start the scan
          for (classInfo in scanResult.getSubclasses(JsonAdapter::class.java.name)) {
            registerMoshiAdapter(classInfo.loadClass())
          }
          for (classInfo in scanResult.getClassesWithMethodAnnotation(FromJson::javaClass.name)) {
            registerAnnotatedMoshiAdapter(classInfo.loadClass())
          }
        }
```


So you could scan under some package, initially okhttp3.sample for @Test annotations, and widen over time.

yschimke:
Also, I suspect you may end up substituting out a lot of PlatformRule, so it returns static true/false responses to the is[Not][Platform] methods.

native-image-tests/build.gradle
yschimke: I'm not against this however, just pointing out options.

yschimke: Digging into this a bit more, I understand the issues with the scan the classpath for annotations approach.  It makes deep assumptions based on the typical uses of Junit.  But Junit 5 has moved past a lot of this, has a clean API for test discovery etc https://junit.org/junit5/docs/5.7.0-RC1/api/org.junit.platform.launcher/org/junit/platform/launcher/core/EngineDiscoveryOrchestrator.html

So maybe the final result of a clean plugin will be a combination of the two.  an @AutomaticFeature that uses EngineDiscoveryOrchestrator instead of just launching the tests, to register each class for reflection needed for later test execution?

native-image-tests/build.gradle
```diff
@@ -0,0 +1,48 @@
...
+  implementation "org.junit.platform:junit-platform-console:1.7.0"
+
+  compileOnly deps.jsr305
+}
+
+animalsniffer {
+  ignoreFailures = true
+}
+
+def graalHome = "/Library/Java/JavaVirtualMachines/graalvm-ce-java11-20.2.0/Contents/Home"
```
yschimke, line 18: It would be nice to check if the palantir plugin has already populated ~/.gradle/caches/com.palantir.graal/20.2.0/11/graalvm-ce-java11-20.2.0

yschimke (approved):
Love this.

native-image-tests/build.gradle
yschimke: Just read "Seed using the Graal from the Graal Gradle plugin, not the local machine"

swankjesse:
I’m going to merge without addressing the feedback above. I’ll be addressing that in a series of follow-up PRs.

PR #6314, opened 2020-10-10
Try to link to existing tests

yschimke:
Not quite working...

```
$ ./native-image-tests/build/graal/ConsoleLauncher
Running Tests with Jdk9Platform

Thanks for using JUnit! Support its development at https://junit.org/sponsoring

╷
├─ JUnit Jupiter ✔
└─ JUnit Vintage ✔
   └─ EventSourceHttpTest ✘ Unsupported constructor jdk.internal.reflect.MethodAccessorGenerator.<init>() is reachable

Failures (1):
  JUnit Vintage:EventSourceHttpTest
    ClassSource [className = 'okhttp3.sse.internal.EventSourceHttpTest', filePosition = null]
    => com.oracle.svm.core.jdk.UnsupportedFeatureError: Unsupported constructor jdk.internal.reflect.MethodAccessorGenerator.<init>() is reachable
       com.oracle.svm.core.util.VMError.unsupportedFeature(VMError.java:86)
       jdk.internal.reflect.MethodAccessorGenerator.<init>(MethodAccessorGenerator.java:64)
       jdk.internal.reflect.ReflectionFactory.generateConstructor(ReflectionFactory.java:513)
       jdk.internal.reflect.ReflectionFactory.newConstructorForSerialization(ReflectionFactory.java:506)
       java.io.ObjectStreamClass.getSerializableConstructor(ObjectStreamClass.java:1531)
       java.io.ObjectStreamClass$2.run(ObjectStreamClass.java:509)
       java.io.ObjectStreamClass$2.run(ObjectStreamClass.java:484)
       java.security.AccessController.doPrivileged(AccessController.java:83)
       java.io.ObjectStreamClass.<init>(ObjectStreamClass.java:484)
       java.io.ObjectStreamClass.lookup(ObjectStreamClass.java:381)
       [...]

Test run finished after 1 ms
[         3 containers found      ]
[         0 containers skipped    ]
[         3 containers started    ]
[         0 containers aborted    ]
[         2 containers successful ]
[         1 containers failed     ]
[         7 tests found           ]
[         0 tests skipped         ]
[         0 tests started         ]
[         0 tests aborted         ]
[         0 tests successful      ]
[         0 tests failed          ]
```


native-image-tests/src/main/resources/META-INF/native-image/native-image-tests/reflect-config.json
```diff
@@ -0,0 +1,215 @@
+[
```
yschimke, line 1: Idea 1: Work out what is needed for junit framework and check that in.

native-image-tests/src/main/kotlin/okhttp3/test/graal/JunitTestRegisterer.kt
```diff
@@ -0,0 +1,50 @@
...
+import org.junit.platform.launcher.core.LauncherFactory
+
+@AutomaticFeature
+class JunitTestRegisterer: Feature {
+  override fun beforeAnalysis(access: Feature.BeforeAnalysisAccess?) {
+//    discoverTests()
+  }
+
+  fun discoverTests() {
+    val request = LauncherDiscoveryRequestBuilder.request()
```
yschimke, line 20: Idea 2 - detect tests cleanly using junit APIs and register those using an automatic feature.

native-image-tests/src/main/kotlin/okhttp3/SampleTest.kt
```diff
@@ -16,7 +16,7 @@
 package okhttp3
 
 import org.assertj.core.api.AssertionsForClassTypes.assertThat
-import org.junit.jupiter.api.Test
+import org.junit.Test
```
yschimke, line 19: Idea 3: Keep using the legacy APIs, instead of upgrading to junit 5 at the same time

native-image-tests/build.gradle
```diff
@@ -16,12 +30,14 @@ animalsniffer {
...
-def graalvmResources = project.file("$buildDir/graalvm/resources")
-def graalvmConfig = project.file("$graalvmResources/META-INF/native-image")
+def graalvmResources = project.file("src/main/resources")
+def graalvmConfig = project.file("$graalvmResources/META-INF/native-image/native-image-tests")
 graalvmConfig.mkdirs()
 
 sourceSets {
   main.resources.srcDirs += graalvmResources
+
+  main.java.srcDirs += project.file("../okhttp-sse/src/test/java")
```
yschimke, line 40: Idea 4: *gradle symlink* the existing classes in, starting with a small fast module but building up.

native-image-tests/src/main/kotlin/okhttp3/SampleTest.kt
swankjesse: Definitely. I think vintage is right. I just didn't want to invest further in JUnit 4.x.

native-image-tests/src/main/resources/META-INF/native-image/native-image-tests/reflect-config.json
swankjesse: Is it more robust to compute this by running a test?

native-image-tests/src/main/resources/META-INF/native-image/native-image-tests/reflect-config.json
```diff
@@ -0,0 +1,215 @@
...
+{
+  "name":"apple.security.AppleProvider",
```
swankjesse, line 3: For example, presumably this is platform-dependent.

swankjesse (approved):

native-image-tests/src/main/resources/META-INF/native-image/native-image-tests/reflect-config.json
yschimke: I was sort of imagining a combined approach.  Each different aspect would be independently correct.

1. Junit 5 launchers we could get the right settings via a one off manual process until Junit 5 supports graal inately.
2. Our tests would be configured overall via an AutomaticFeature separating out detection from running.  So you could use the image to run any set of tests etc.
3. We have correct okhttp main code that works automatically with graal.

native-image-tests/src/main/kotlin/okhttp3/SampleTest.kt
yschimke: I think this bumps the priority of the nice PR to split up MockWebServer.  Then follow up with a rip-the-bandaid approach to move us from Junit 4 to Junit 5.

native-image-tests/src/main/resources/META-INF/native-image/native-image-tests/reflect-config.json
yschimke: Yep - to be clear this comes from your gradle task, not from anything I've done.

yschimke: I can see a lot of benefits to start with the run tests twice approach, but long term a few drawbacks.

1) If the Graal JVM is different than native image, then tests may break when they exercise different code paths anyway.
2) It's 2x slower
3) It may hide problems that we want surfaced, e.g. some of these should maybe be included in the config files baked inside of okhttp-sse jar.

native-image-tests/src/main/kotlin/okhttp3/SampleTest.kt
swankjesse: Agreed on MockWebServer.

yschimke:
@swankjesse I probably won't land this as is.  It was mainly about exploring what options are there.  Should I focus on getting this part working

```
main.java.srcDirs += project.file("../okhttp-sse/src/test/java")
```

And running that test to start with?

native-image-tests/src/main/kotlin/okhttp3/SampleTest.kt
yschimke: https://github.com/square/okhttp/pull/6315 makes me think investing anytime in Junit 4.x here is likely wasted. Going to look at how to progress the MockWebServer split.

PR #6320, opened 2020-10-15
Java Toolchain for splitting compile and CI test runner

yschimke:
https://docs.gradle.org/current/userguide/toolchains.html#header

https://docs.gradle.org/current/samples/sample_jvm_multi_project_with_toolchains.html#header

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -124,7 +124,7 @@ jobs:
 
   testjdk8:
     docker:
-      - image: circleci/openjdk:8u242-jdk-buster
+      - image: circleci/openjdk:11.0.6-jdk-stretch
 
     environment:
       JVM_OPTS: -Xmx1g
@@ -135,11 +135,11 @@ jobs:
 
       - runtests:
           platform: jdk8
-          testjdk: -Dtest.java.version=8
+          testjdk: -Dtest.java.version=8 -Dorg.gradle.java.installations.paths=/usr/lib/jvm/java-8-openjdk-amd64
 
   testjdk8alpn:
     docker:
-      - image: circleci/openjdk:8u242-jdk-buster
+      - image: circleci/openjdk:11.0.6-jdk-stretch
 
     environment:
       JVM_OPTS: -Xmx1g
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -186,6 +192,10 @@ subprojects { project ->
   def platformJavaHome = System.getProperty('test.java.home')
 
   test {
+    javaLauncher = javaToolchains.launcherFor {
+      languageVersion = JavaLanguageVersion.of(14)
```
yschimke, line 196: Will allow overrides for this, defaulting to 11, and fix up CI to use it instead of requiring circleci images to select

.circleci/config.yml
```diff
@@ -135,11 +135,11 @@ jobs:
 
       - runtests:
           platform: jdk8
-          testjdk: -Dtest.java.home=/usr/lib/jvm/java-8-openjdk-amd64
+          testjdk: -Dtest.java.version=8
 
   testjdk8alpn:
     docker:
-      - image: circleci/openjdk:11.0.6-jdk-stretch
```
yschimke, line 142: I suspect I may need to use a custom path and revert this docker image.  Will follow up on this.

org.gradle.java.installations.paths=/custom/path/jdk1.8,/shared/jre11

yschimke:
Test runs are tripping up on compiling (not running)

```
          println("TLSv1.3 requires an external command run before first traffic is sent")
          println("Follow instructions at https://github.com/neykov/extract-tls-secrets for TLSv1.3")
          println("Pid: ${ProcessHandle.current().pid()}")
```


and

```
    val config = "--name=OpenSC\nlibrary=/Library/OpenSC/lib/opensc-pkcs11.so\nslot=$slot\n"

    // May fail with ProviderException with root cause like
    // sun.security.pkcs11.wrapper.PKCS11Exception: CKR_SLOT_ID_INVALID
    val pkcs11 = Security.getProvider("SunPKCS11").configure(config)
    Security.addProvider(pkcs11)
```

build.gradle
```diff
@@ -149,6 +149,12 @@ subprojects { project ->
     options.encoding = 'UTF-8'
   }
 
+  java {
```
swankjesse, line 152: nice

swankjesse (approved):

PR #6324, opened 2020-10-17
Make okhttp easier to find in IntelliJ's recents

swankjesse:


settings.gradle
```diff
@@ -1,4 +1,4 @@
-rootProject.name = 'okhttp-parent'
+rootProject.name = "🍓 okhttp"
```
swankjesse, line 1: open to suggestions on alternatives!

JakeWharton: You're creating an arms race now. How about just OkHttp? You can type and filter the projects so it's only one or two characters away as is.

swankjesse: Not in the ‘open recent’ menu?

yschimke: Probably okhttp, moving okhttp (directory) to okhttp-core and then reverting the artifactId in publishedArtifactId back to okhttp.

yschimke:
Also maybe for yourself https://stackoverflow.com/questions/43558618/groups-of-projects-in-intellij-idea

yschimke (changes requested):
It's like spitting in the face of https://github.com/square/Aardvark developers

PR #6326, opened 2020-10-18
MockWebServer wrapper

swankjesse:
This is step 1 of a plan to break MockWebServer's dependency on JUnit 4.

1. New module, mockwebserverwrapper: implement almost all of the mockwebserver
   API by delegating to the real mockwebserver API.

2. Swap the names. Rename the package okhttp3.mockwebserver to mockwebserver3,
   and rename okhttp3.mockwebserverwrapper to okhttp3.mockwebserver. With this
   rename all users of MockWebServer are now using the wrapper instead.

3. Change mockwebserver3 to not depend on JUnit 4.

4. Migrate callers from the okhttp3.mockwebserver wrapper to mockwebserver3,
   dropping their JUnit dependency along the way. It may take a very long time
   for callers to migrate. That's okay; the wrapper isn't much code to maintain.

Notable breaks in binary compatibility:

 - Duplex is not supported in the wrapper. This was never intended as public
   API.
 - QueueDispatcher cannot be subclassed in the wrapper. Supporting this is
   rather annoying because we need to support reverse wrapping and unwrapping.
   I think it's not worth the effort.

### DIFF
--- a/mockwebserverwrapper/src/main/kotlin/okhttp3/mockwebserverwrapper/Dispatcher.kt
+++ b/mockwebserverwrapper/src/main/kotlin/okhttp3/mockwebserverwrapper/Dispatcher.kt
@@ -1,0 +1,27 @@
+/*
+ * Copyright (C) 2020 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okhttp3.mockwebserverwrapper
+
+abstract class Dispatcher {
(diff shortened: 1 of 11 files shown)


mockwebserverwrapper/src/main/kotlin/okhttp3/mockwebserverwrapper/Dispatcher.kt
```diff
@@ -0,0 +1,27 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okhttp3.mockwebserverwrapper
+
+abstract class Dispatcher {
+  @Throws(InterruptedException::class)
```
yschimke, line 19: Do we want throws in new kotlin public APIs?

yschimke (approved):

adriancole (approved):
sound plan!

mockwebserverwrapper/src/main/kotlin/okhttp3/mockwebserverwrapper/Dispatcher.kt
swankjesse: We do! But this isn’t a new API; it’s a different implementation of an old API.

PR #6330, opened 2020-10-20
Experimental junit5 mockwebserver3 integration

yschimke:
Playing around with a junit5 extension, not sure if you have an existing one you are working on.  Probably worth external review of how to implement.  I haven't even looked around for existing 3rd party ones, just playing with jupiter APIs at the moment.

No strong attachment to this code.


mockwebserver-junit5/src/main/kotlin/mockwebserver3/junit5/MockWebServerExtension.kt
```diff
@@ -0,0 +1,72 @@
...
+    try {
+      server.shutdown()
+    } catch (e: IOException) {
+      logger.log(Level.WARNING, "MockWebServer shutdown failed", e)
+    }
+  }
+
+  companion object {
+    private val logger = Logger.getLogger(MockWebServerExtension::class.java.name)
+    const val MockWebServerKey = "MockWebServer"
```
yschimke, line 70: We might want to build an API that allows multiple servers?

mockwebserver-junit5/src/test/java/mockwebserver3/junit5/ExampleMockWebServerTest.kt
```diff
@@ -0,0 +1,16 @@
+package mockwebserver3.junit5
+
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.extension.RegisterExtension
+import kotlin.test.assertTrue
+
+class ExampleMockWebServerTest {
+  @RegisterExtension
```
swankjesse, line 8: I was thinking we might be able to implement this in a way where the field you declare is a normal MockWebServer instance. (Though it might be more difficult to implement.)

yschimke: Anything here is possible, it's probably more relevant, what is more idiomatic junit5 test code.  We can try out a few options, see what the tests are like.  Will try your suggestion on this branch tonight.

yschimke: Seems like that is the approach using ParameterResolver here

https://github.com/Dev-Mind/mockwebserver/blob/master/src/main/java/fr/devmind/mockwebserver/MockWebServerExtension.java

cc @javamind

adriancole: @anuraaga any feelings one way or another on this?

anuraaga: I see both `@ExtendsWith` with type resolution and `RegisterExtension` fields like this one. The latter has some flexibility by making it easier to configure the test-specific behavior if there's any settings I believe, something like 

```java
private MockWebServerExtension server = new MockWebServerExtension() {
  @Override
  int responseDelay() {
    return 100;
  }
}
```

It's also nice if the extension can be applied beforeAll or beforeEach at user discretion

yschimke: The ExampleMockWebServerParameterTest test shows ExtendsWith. Agree about extending or configuring the instance, as much as we choose to make that public API.

But not sure how to achieve what you are suggesting for beforeAll and beforeEach.  some flag?  Seems a bit clunky and at that point maybe you want your own Extension with your particular lifecycle.

anuraaga: I use this pattern to let one option on the extension control it - it's exposed as a protected method for anonymous subclass overriding, which is admittedly not the most beautiful but seems reasonable at the same time.

https://github.com/line/armeria/blob/master/junit5/src/main/java/com/linecorp/armeria/testing/junit5/common/AbstractAllOrEachExtension.java

With many many tests, the ability to reuse the port across tests in a class does become significant in my experience so whether default or not I do hope it's supported easily enough :)

yschimke: This option doesn't feel in the style of OkHttp.  Also as far as public APIs go, I think we already have a wide one in terms of MockWebServer, adding more at an intermediate layer feels like overkill here.  You can already configure teh mockwebserver before it starts.

mockwebserver-junit5/src/main/kotlin/mockwebserver3/junit5/MockWebServerExtension.kt
yschimke: For later, easier to add new public API later than remove.

mockwebserver/src/main/kotlin/mockwebserver3/MockWebServer.kt
```diff
@@ -187,7 +187,7 @@ class MockWebServer : Closeable {
       field = protocolList
     }
 
-  private var started: Boolean = false
+  var started: Boolean = false
```
yschimke, line 190: This is a public API change - to a new class.

yschimke:
@jwilson What are your plans for a OkHttp release - post mockwebserver cleanup, adoption of enough tests to be confident of nativeImage?

I'd like to land something even experimental here and start migrating some tests to junit 5 so we can do these in parallel.  Thoughts?

mockwebserver-junit5/src/main/kotlin/mockwebserver3/junit5/MockWebServerExtension.kt
```diff
@@ -0,0 +1,93 @@
...
+
+/**
+ * Runs MockWebServer for the duration of a single test method.
+ *
+ * In Java JUnit 5 tests (ie. tests annotated `@org.junit.jupiter.api.Test`), use this by defining
+ * a field with the `@RegisterExtension` annotation.  When used with @ExtendWith, access the
+ * server instance using a parameter of type mockwebserver3.MockWebServer
+ *
+ * The MockWebServer instance is stored in the GLOBAL namespace using the key "MockWebServer".
+ * This can be useful for other plugins that want to observe the MockWebServer in a test.
```
swankjesse, line 38: I don't like this because some tests use 2 instances!

mockwebserver-junit5/src/main/kotlin/mockwebserver3/junit5/MockWebServerExtension.kt
swankjesse: I think this precludes multiple servers?

mockwebserver-junit5/src/test/java/mockwebserver3/junit5/ExampleMockWebServerParameterTest.kt
```diff
@@ -0,0 +1,15 @@
+package mockwebserver3.junit5
+
+import mockwebserver3.MockWebServer
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.extension.ExtendWith
+import org.junit.jupiter.api.extension.RegisterExtension
+import kotlin.test.assertTrue
+
+@ExtendWith(MockWebServerExtension::class)
```
swankjesse, line 9: Can we register our extension using ServiceLoader instead? That way the annotation is unnecessary?

mockwebserver/src/main/kotlin/mockwebserver3/MockWebServer.kt
swankjesse: Needs a whole bunch of @Synchronized things to make this public.

swankjesse: Is this public for end users or just public for testing?

swankjesse:
Agreed on the plan: get tests running under GraalVM, fix up the MockWebServer APIs, and release. I don't like either of the JUnit 5 APIs here though. The `Rule` was so flexible and small; this new extensions API is clunkly to use in comparison. I'd like to find something very small, almost like the old `@Rule` if we can!

adriancole:
FWIW I'd also prefer to decouple and have the junit 5 happen after as it is easier to practice. In zipkin brave we have a number of paths that I'm too knackered to refactor at this moment and even if could would be too rushed to make an immediate design recommendation. MWS is one of the best parts of open source and deserves the best api and integration.

yschimke:
Ok. I'll move it into an internal package, mention it's not for external use and we can take a bit of time to get it right.

My 2c, it's worth forgetting what we had in old MWS rule and lean into junit 5. Probably use annotated parameter injection.

But I'd like to land something internal to unblock graal work. That could be a old school rule for now.  You can auto register extensions via service model.

adriancole:
indeed injected parameter is likely it as yeah the instance sharing thing makes things messy in in jupiter. I've lost many hours on this. that said, there's going to be some finesse needed and probably some cases people won't share instances.. 

I have loved composition of rules and that's super weird as parameter injection... but that's conjecture.. I've not tried composition in parameters.. maybe it isn't too bad to setup a server and redirect to another with two parameters.. things like that.

In any case we should use real diffs that show how it works, and accept with open mind it might look different. having experimental gets the conversation started! thanks

mockwebserver/src/main/kotlin/mockwebserver3/MockWebServer.kt
yschimke: I'll revert.

mockwebserver-junit5/src/test/java/mockwebserver3/junit5/ExampleMockWebServerParameterTest.kt
yschimke: Yes, I've added an example of that now.

mockwebserver-junit5/src/main/kotlin/mockwebserver3/junit5/MockWebServerExtension.kt
yschimke: Good point. Added an example using an annotation on a field that could have configurable annotation params etc.

yschimke:
What other things would we expect this module to handle OkHttpClient injection? Dns? Recording Event Listener? Flaky test handling? Fail on non closed resources?

yschimke:
I picked injecting MockWebServer because it seemed like the thing blocking using okhttp-sse as a test of graal to start with.  I'd still like to keep that in mind.  If we don't consider this as published public API for now, what is a minimal target to get in an experiment with real use so we can work out what an awesome mockwebserver3-junit5 integration would be.

https://github.com/square/okhttp/pull/6315

adriancole:
vote for leak detection (by default!)

mockwebserver-junit4/README.md
```diff
@@ -0,0 +1,4 @@
+MockWebServer for Junit 4
+=========================
+
+Experimental support for mockwebserver3.MockWebServer in Junit 4.
```
swankjesse, line 4: This one isn’t experimental!

mockwebserver-junit5/README.md
```diff
@@ -0,0 +1,5 @@
+MockWebServer for Junit 5
+=========================
+
+Experimental support for mockwebserver3.MockWebServer in Junit 5.
+Do not use.
```
swankjesse, line 5: 👍🏻

mockwebserver-junit5/src/main/kotlin/mockwebserver3/junit5/internal/MockWebServerInstance.kt
```diff
@@ -0,0 +1,3 @@
+package mockwebserver3.junit5.internal
+
+annotation class MockWebServerInstance(val tls: Boolean)
```
swankjesse, line 3: Remove parameter, it’s too arbitrary? Programmatic configuration is easy enough, and where it’s not we should fix that.

mockwebserver-junit5/src/main/kotlin/mockwebserver3/junit5/internal/MockWebServerRuleExtension.kt
```diff
@@ -0,0 +1,110 @@
...
+ *
+ * ```
+ * @JvmField @RegisterExtension val mockWebServer: MockWebServerExtension = MockWebServerExtension()
+ * ```
+ */
+class MockWebServerRuleExtension : BeforeEachCallback, AfterEachCallback, BeforeTestExecutionCallback {
+  var server: MockWebServer? = MockWebServer()
+
+  override fun beforeEach(context: ExtensionContext) {
+    println("Here B")
```
swankjesse, line 57: nit

mockwebserver-junit5/src/main/kotlin/mockwebserver3/junit5/internal/MockWebServerRuleExtension.kt
```diff
@@ -0,0 +1,110 @@
...
+
+    context.requiredTestInstances.allInstances
+      .forEach(Consumer { instance: Any? ->
+        injectInstanceFields(context, instance!!)
+      })
+  }
+
+  private fun injectInstanceFields(context: ExtensionContext, instance: Any) {
+    injectFields(context, instance, instance.javaClass
+    ) { member: Field? ->
```
swankjesse, line 67: nit: weird wrapping

mockwebserver-junit5/src/main/kotlin/mockwebserver3/junit5/internal/MockWebServerRuleExtension.kt
```diff
@@ -0,0 +1,110 @@
...
+      ReflectionUtils.isNotStatic(member)
+    }
+  }
+
+  private fun injectFields(
+    context: ExtensionContext, testInstance: Any, testClass: Class<*>, predicate: Predicate<Field>
+  ) {
+    // TODO fake to make the test work for now
```
swankjesse, line 75: nit

mockwebserver-junit5/src/main/kotlin/mockwebserver3/junit5/internal/MockWebServerRuleExtension.kt
```diff
@@ -0,0 +1,110 @@
...
+    if (testInstance.javaClass.name == "mockwebserver3.junit5.internal.ExampleInjectedMockWebServerTest") {
+      server = testInstance.javaClass.getMethod("getServer").invoke(testInstance) as MockWebServer?
+    }
+
+    AnnotationUtils.findAnnotatedFields(testClass, MockWebServerInstance::class.java, predicate).forEach(
```
swankjesse, line 80: I find this awkward . . . . we create a server, and we also ingest the server field?

mockwebserver-junit5/src/test/java/mockwebserver3/junit5/internal/ExampleMockWebServerParameterTest.kt
```diff
@@ -0,0 +1,14 @@
+package mockwebserver3.junit5.internal
+
+import mockwebserver3.MockWebServer
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.extension.ExtendWith
+import kotlin.test.assertTrue
+
+@ExtendWith(MockWebServerExtension::class)
```
swankjesse, line 8: Is this required? I’d hope the ServiceLoader would make it unnecessary

mockwebserver-junit5/src/test/java/mockwebserver3/junit5/internal/ExampleMockWebServerTest.kt
```diff
@@ -0,0 +1,16 @@
+package mockwebserver3.junit5.internal
+
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.extension.RegisterExtension
+import kotlin.test.assertTrue
+
+class ExampleMockWebServerTest {
+  @RegisterExtension
```
swankjesse, line 8: Can we avoid this?

swankjesse (approved):
Approved to unblock. But I think the API needs work.

I’d like to support these:

```
class FirstTest(
  val server: MockWebServer
) {
  @Test fun test() {
    ...
  }
}
```

```
class SecondTest() {
  @Test fun test(
    val server: MockWebServer
  ) {
    ...
  }
}
```

We do a parameter resolver that’ll automatically shut the server down at test completion. I think we can do it without any `@ExtendWith`,  `@MockWebServerInstance`, or `@RegisterExtension` annotations/

swankjesse:
I added a commit with all my feedback here:
https://github.com/square/okhttp/compare/yschimke-junit5?expand=1

yschimke:
> Approved to unblock. But I think the API needs work.

Agreed - this PR was me trying all available options for discussion, but I sincerely think that the best way forward is something internal for now, and we iterate on it through active use and then publish when ready.

PR #6346, opened 2020-10-28
Use generated certificates in unit tests

yschimke:
We currently embed externally generated certificates in our unit tests, but we can now generate our own certificates so we should.


okhttp/src/main/kotlin/okhttp3/internal/tls/DnsUtils.kt
```diff
@@ -0,0 +1,76 @@
...
+    return c in 'A'..'Z'
+  }
+
+  // https://tools.ietf.org/html/rfc2459#section-4.2.1.11
+  fun normalizeIA5String(s: String): String {
+    var pos = 0
+    var remaining = s.length
+    while (remaining > 0) {
+      // TODO enable if we agree to enforce here
+//      check (isIA5(s[pos])) {
```
yschimke, line 49: @swankjesse if we trust our overall logic, we should enforce this. Tests like verifyMultipleCn will fail and need to be adjusted.

yschimke: This is an addition to the original file which was converted from java.

swankjesse: I don’t think we need this? I also don’t think Apache HTTP components needs it? I think a better solution is to simply require the input argument is already lowercase ASCII?

swankjesse: In particular, `verifyHostname()` is not a public API. We only need this defense if we’re passing in strings ourselves that need this transformation.

swankjesse: Yep, as far as I can tell 100% of the hosts we pass to `verify()` have already passed through `okhttp3.internal.HostnamesKt#toCanonicalHost` which is ASCII-only.

We could potentially document or enforce that constraint?

yschimke: > We could  ... enforce that constraint?

That is this block.  I'll follow up on later PR.

yschimke: verify() is public API though.  While we package OkHostnameVerifier in an internal package, it's a public API exposed very obviously in OkHttpClient via the hostnameVerifier() accessor.

okhttp/src/main/kotlin/okhttp3/internal/tls/OkHostnameVerifier.kt
```diff
@@ -61,7 +60,7 @@ object OkHostnameVerifier : HostnameVerifier {
 
   /** Returns true if [certificate] matches [hostname]. */
   private fun verifyHostname(hostname: String, certificate: X509Certificate): Boolean {
-    val hostname = hostname.toLowerCase(Locale.US)
+    val hostname = DnsUtils.normalizeIA5String(hostname)
```
swankjesse, line 63: should we use `hostname.toCanonicalHost()` instead?

yschimke: It's worse actually, causes us to fail on the tel example also.

```
okhttp3.internal.tls.HostnameVerifierTest > specialKInHostname FAILED
    org.junit.ComparisonFailure: expected:<[fals]e> but was:<[tru]e>
        at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
        at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
        at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
        at okhttp3.internal.tls.HostnameVerifierTest.specialKInHostname(HostnameVerifierTest.java:586)
```

yschimke: Maybe there is a fix where fix toCanonicalHost logic and then we use it, but maybe that's a good second pass PR once the test is in.

swankjesse: What about just removing the `toLowerCase(Locale.US)` call completely? I’m trying to avoid having a bunch of code in here that doesn’t get exercised in practice.

okhttp/src/main/kotlin/okhttp3/internal/tls/DnsUtils.kt
```diff
@@ -0,0 +1,75 @@
...
+   * Secondly it avoids the simple and efficient lowercase ASCII conversion method below
+   * from being applied to other unicode uppercase characters and just moving them to another
+   * undesired unicode character.
+   */
+  private fun isUpper(c: Char): Boolean {
+    return c in 'A'..'Z'
+  }
+
+  // https://tools.ietf.org/html/rfc2459#section-4.2.1.11
+  fun normalizeIA5String(s: String): String {
```
swankjesse, line 47: In Okio this function is called `ByteString.toAsciiLowercase()` (though to use that one we’d need to roundtrip through a ByteString)

swankjesse:
I pushed a commit here:
https://github.com/square/okhttp/compare/master...yschimke-hostnames?expand=1#files_bucket

okhttp/src/main/kotlin/okhttp3/internal/tls/OkHostnameVerifier.kt
yschimke: It's still public API

```


  @Test
  public void thatCatchesErrorsWithBadSession() {
    HostnameVerifier localVerifier = new OkHttpClient().hostnameVerifier();

    // Since this is public API, okhttp3.internal.tls.OkHostnameVerifier.verify is also
    assertThat(verifier).isInstanceOf(OkHostnameVerifier.class);

    SSLSession session = TlsUtil.localhost().sslContext().createSSLEngine().getSession();
    assertThat(localVerifier.verify("\uD83D\uDCA9.com", session)).isFalse();
  }
```

yschimke:
> 
> 
> I pushed a commit here:
> https://github.com/square/okhttp/compare/master...yschimke-hostnames?expand=1#files_bucket

Nice simplification Picasso :) 

I'll land based on your commit.

PR #6349, opened 2020-10-30
Execute CallTest with JUnit 5

swankjesse:


yschimke:
Great cleanup and example towards migration.

PR #6353, opened 2020-10-30
Test quirks of HostnameVerifier

yschimke:


okhttp/src/test/java/okhttp3/internal/tls/HostnameVerifierTest.java
```diff
@@ -646,6 +646,33 @@
...
+        + "MIIBSDCB86ADAgECAhRLR4TGgXBegg0np90FZ1KPeWpDtjANBgkqhkiG9w0BAQsF\n"
+        + "ADASMRAwDgYDVQQDDAdmb28uY29tMCAXDTIwMTAyOTA2NTkwNVoYDzIxMjAxMDA1\n"
+        + "MDY1OTA1WjASMRAwDgYDVQQDDAdmb28uY29tMFwwDQYJKoZIhvcNAQEBBQADSwAw\n"
+        + "SAJBALQcTVW9aW++ClIV9/9iSzijsPvQGEu/FQOjIycSrSIheZyZmR8bluSNBq0C\n"
+        + "9fpalRKZb0S2tlCTi5WoX8d3K30CAwEAAaMfMB0wGwYDVR0RBBQwEoIH4oShLmNv\n"
+        + "bYIH4oSqLmNvbTANBgkqhkiG9w0BAQsFAANBAA1+/eDvSUGv78iEjNW+1w3OPAwt\n"
+        + "Ij1qLQ/YI8OogZPMk7YY46/ydWWp7UpD47zy/vKmm4pOc8Glc8MoDD6UADs=\n"
+        + "-----END CERTIFICATE-----\n");
+
+    assertThat(verifier.verify("���.com", session)).isFalse();
```
swankjesse, line 673: is this supposed to be replacement chars?

swankjesse: potentially worth calling that out, otherwise readers might think that git or IntelliJ screwed up somehow

swankjesse (approved):

yschimke:
The approval gives me nice warm fuzzies, but it's blocked to resolve what we think the behaviour should be.  Or just assert the observed external behaviour since it is JSSE?

swankjesse:
Should reject these in Certifikit? Lets change it to always return false if we see any non-ASCII characters? They shouldn’t be in the certificate and therefore we shouldn’t match if they’re in the hostname arg.

okhttp/src/test/java/okhttp3/internal/tls/HostnameVerifierTest.java
```diff
@@ -332,11 +337,11 @@
...
   /**
-   * Ignored due to incompatibilities between Android and Java on how non-ASCII subject alt names
-   * are parsed. Android fails to parse these, which means we fall back to the CN. The RI does parse
-   * them, so the CN is unused.
+   * Previously ignored due to incompatibilities between Android and Java on how non-ASCII subject
+   * alt names are parsed. Android fails to parse these, which means we fall back to the CN.
+   * The RI does parse them, so the CN is unused.
    */
-  @Test @Ignore public void testWilcardNonAsciiSubjectAlt() throws Exception {
+  @Test public void testWilcardNonAsciiSubjectAlt() throws Exception {
```
yschimke, line 344: @swankjesse I reactivated these tests as part of the change, are you ok with these?

yschimke:
> Lets change it to always return false if we see any non-ASCII characters?

@swankjesse I've landed with this change, overall probably worth a quick post review for style.

okhttp/src/test/java/okhttp3/internal/tls/HostnameVerifierTest.java
swankjesse: Yep!

swankjesse (commented):
LGTM

PR #6361, opened 2020-10-31
Fix toolchain issues breaking jdk8alpn build

yschimke:


yschimke:
See https://app.circleci.com/pipelines/github/square/okhttp/4055/workflows/5338866e-4ff0-474f-b8b9-b43419135ca2/jobs/18127

PR #6367, opened 2020-11-01
Fixing n-i-t launcher

yschimke:
Graal main method to run tests with minimal reflection and automatic settings.
Uses the test list in native-image-tests/src/main/resources/testlist.txt.

### DIFF
--- a/native-image-tests/build.gradle
+++ b/native-image-tests/build.gradle
@@ -59,6 +59,7 @@ graal {
   option "--report-unsupported-elements-at-runtime"
   option "-H:+ReportExceptionStackTraces"
   option "-H:+TraceClassInitialization"
+  option "--initialize-at-build-time=kotlin.text.Charsets"
 
   if (Os.isFamily(Os.FAMILY_WINDOWS)) {
     // May be possible without, but autodetection is problematic on Windows 10
--- a/native-image-tests/build.gradle
+++ b/native-image-tests/build.gradle
@@ -20,7 +20,7 @@ dependencies {
   implementation project(':mockwebserver-deprecated')
   implementation project(':okhttp-urlconnection')
   implementation project(':mockwebserver-junit4')
-  testRuntimeOnly project(':mockwebserver-junit5')
+  implementation project(':mockwebserver-junit5')
   implementation deps.bndResolve
   implementation deps.junit5Api
 
@@ -39,9 +39,12 @@ animalsniffer {
   ignoreFailures = true
 }
 
-sourceSets {
-  main.java.srcDirs += project.file("../okhttp-logging-interceptor/src/test/java")
-  main.java.srcDirs += project.file("../okhttp-sse/src/test/java")
+if (!properties.containsKey('android.injected.invoked.from.ide')) {
+  sourceSets {
+    // Not included in IDE as this confuses Intellij for obvious reasons.
+    main.java.srcDirs += project.file("../okhttp-logging-interceptor/src/test/java")
+    main.java.srcDirs += project.file("../okhttp-sse/src/test/java")
+  }
 }
 
 graal {
@@ -54,6 +57,7 @@ graal {
   option "--no-fallback"
   option "--allow-incomplete-classpath"
   option "--report-unsupported-elements-at-runtime"
+  option "-H:+ReportExceptionStackTraces"
 
   if (Os.isFamily(Os.FAMILY_WINDOWS)) {
     // May be possible without, but autodetection is problematic on Windows 10
--- a/native-image-tests/build.gradle
+++ b/native-image-tests/build.gradle
@@ -58,6 +58,7 @@ graal {
   option "--allow-incomplete-classpath"
   option "--report-unsupported-elements-at-runtime"
   option "-H:+ReportExceptionStackTraces"
+  option "-H:+TraceClassInitialization"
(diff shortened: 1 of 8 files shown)


yschimke:
@swankjesse With RunTests.kt ultimately with enough cleanup we may get to a point that ConsoleLauncher would be cleanly usable instead of this classlist.txt approach, but I've found that reflection and loading all available classes opportunistically with GraalVM is very brittle.  I'd like this to be both simple (no extra prestep like manual run of seed) and predictable (doesn't trigger a lot of interesting code paths accidentally).

native-image-tests/build.gradle
```diff
@@ -20,7 +20,7 @@ dependencies {
   implementation project(':mockwebserver-deprecated')
   implementation project(':okhttp-urlconnection')
   implementation project(':mockwebserver-junit4')
-  testRuntimeOnly project(':mockwebserver-junit5')
+  implementation project(':mockwebserver-junit5')
```
swankjesse, line 23: This one’s interesting. We don’t refer to it at compile time right?

settings.gradle
```diff
@@ -11,10 +11,7 @@ if (properties.containsKey('android.injected.invoked.from.ide') ||
   include ':android-test'
 }
 
-if (!properties.containsKey('android.injected.invoked.from.ide')) {
```
swankjesse, line 14: Nice improvement here

native-image-tests/src/main/kotlin/okhttp3/TestRegistration.kt
```diff
@@ -18,25 +18,43 @@ package okhttp3
...
+    knownTestFile.readLines().forEach {
+      try {
+        val testClass = access.findClassByName(it)
+
+        if (testClass != null) {
+          access.registerAsUsed(testClass)
+          registerTest(testClass)
+        }
+      } catch (e: Exception) {
+        e.printStackTrace()
```
swankjesse, line 53: rethrow? what’s this defending against

native-image-tests/src/main/kotlin/okhttp3/RunTests.kt
```diff
@@ -17,60 +17,87 @@ package okhttp3
...
+    .lines()
+    .filter { it.isNotBlank() }
+    .map {
+      selectClass(Class.forName(it, false, sampleTestClass.classLoader))
+    }
+}
+
+fun buildRequest(selectors: List<DiscoverySelector>): LauncherDiscoveryRequest {
+  val request: LauncherDiscoveryRequest = LauncherDiscoveryRequestBuilder.request()
+    // TODO replace junit.jupiter.extensions.autodetection.enabled with API approach.
```
swankjesse, line 84: what’s this mean? Changing how we configure JUnit here? Or how we did it in our tests?

native-image-tests/src/main/kotlin/okhttp3/GenerateClassList.kt
```diff
@@ -0,0 +1,15 @@
+package okhttp3
+
+import org.junit.jupiter.engine.descriptor.ClassBasedTestDescriptor
+import java.io.File
+
+/**
+ * Run periodically to refresh the known set of working tests.
+ *
+ * TODO use filtering to allow skipping acceptable problem tests
```
swankjesse, line 9: Better to integrate with PlatformTest, like Platform.isGraalvm() ?

native-image-tests/src/main/kotlin/okhttp3/GenerateClassList.kt
```diff
@@ -0,0 +1,15 @@
+package okhttp3
```
swankjesse, line 1: supernit: copyright

swankjesse (approved):

native-image-tests/build.gradle
yschimke: There is no test runtime!

native-image-tests/src/main/kotlin/okhttp3/RunTests.kt
yschimke: https://github.com/junit-team/junit5/issues/2470

I'm this close to a config driven completely without any magic.  But then I nee to set a system property to make it use JAR services mechanism.  I'd prefer someway to do that in code.

native-image-tests/src/main/kotlin/okhttp3/TestRegistration.kt
yschimke: I'll comment. If you throw an exception here then native image building fails half way through siltently without rewriting the binary.  So we report noisily, but keep going and prefer running most tests still.

native-image-tests/src/main/kotlin/okhttp3/GenerateClassList.kt
yschimke: Yep, probably in a follow up. I'd still like to use tags. Go full junit5 native on this.

native-image-tests/src/main/kotlin/okhttp3/GenerateClassList.kt
yschimke: You should automate this

```
if pr.author == yschimke and files.new > 0
  comment("remember copyright")
```

PR #6369, opened 2020-11-01
Fix samples/compare with junit 5

yschimke:


samples/compare/src/test/kotlin/okhttp3/compare/JavaHttpClientTest.kt
```diff
@@ -15,30 +15,35 @@
...
  *
  * https://openjdk.java.net/groups/net/httpclient/intro.html
+ *
+ * Baseline test if we ned to validate OkHttp behaviour against other popular clients.
  */
-@Disabled("Requires Java 11, but OkHttp runs on Java 8+")
-class JavaHttpClientTest(
-  private val server: MockWebServer
-) {
-  private val httpClient = HttpClient.newBuilder()
```
yschimke, line 37: This is the key fix since it isn't available before JDK11

samples/compare/src/test/kotlin/okhttp3/compare/ApacheHttpClientTest.kt
```diff
@@ -28,17 +28,17 @@ import org.junit.jupiter.api.Test
  * Apache HttpClient 5.x.
  *
  * https://hc.apache.org/httpcomponents-client-5.0.x/index.html
+ *
+ * Baseline test if we ned to validate OkHttp behaviour against other popular clients.
  */
-class ApacheHttpClientTest(
-  private val server: MockWebServer
-) {
+class ApacheHttpClientTest {
```
yschimke, line 34: This change was accidental because I made the changes on old copy of master before yours.  But leaving for discussion.

samples/compare/src/test/kotlin/okhttp3/compare/ApacheHttpClientTest.kt
```diff
@@ -28,17 +28,17 @@ import org.junit.jupiter.api.Test
  * Apache HttpClient 5.x.
  *
  * https://hc.apache.org/httpcomponents-client-5.0.x/index.html
+ *
+ * Baseline test if we ned to validate OkHttp behaviour against other popular clients.
```
swankjesse, line 32: ned -> need

swankjesse (approved):

samples/compare/src/test/kotlin/okhttp3/compare/JavaHttpClientTest.kt
```diff
@@ -15,30 +15,35 @@
...
+import org.junit.jupiter.api.extension.RegisterExtension
 
 /**
  * Java HTTP Client.
  *
  * https://openjdk.java.net/groups/net/httpclient/intro.html
+ *
+ * Baseline test if we ned to validate OkHttp behaviour against other popular clients.
  */
-@Disabled("Requires Java 11, but OkHttp runs on Java 8+")
```
swankjesse, line 33: It's interesting that @Ignore was sufficient with JUnit 4 but @Disabled is not on JUnit 5.

swankjesse (approved):

samples/compare/src/test/kotlin/okhttp3/compare/JavaHttpClientTest.kt
yschimke: True, but this seems better anyway

PR #6391, opened 2020-11-04
Allow for dev builds without proguard warnings (for discussion)

yschimke:
Forking off https://github.com/google/conscrypt/issues/881

Should we avoid warning here, or does it add value to stop this working in a release build?


swankjesse:
I don't understand. Why is ProGuard stripping this? Isn't it reachable?

yschimke:
Only via reflection I believe.

PR #6398, opened 2020-11-07
Test fix for cert names

yschimke:


okhttp/src/test/java/okhttp3/internal/tls/HostnameVerifierTest.java
```diff
@@ -373,7 +380,11 @@
         + "-----END CERTIFICATE-----\n");
 
     X509Certificate peerCertificate = ((X509Certificate) session.getPeerCertificates()[0]);
-    assertThat(certificateSANs(peerCertificate)).containsExactly("*.bar.com", "*.������.co.jp");
+    if (platform.isConscrypt()) {
+      assertThat(certificateSANs(peerCertificate)).containsExactly("*.bar.com");
+    } else {
+      assertThat(certificateSANs(peerCertificate)).containsExactly("*.bar.com", "*.������.co.jp");
```
swankjesse, line 386: is this asserting a bug? Shouldn’t the cert library be limiting these to ASCII strings?

swankjesse: (the bug is in JSSE, Coretto, etc.)

yschimke: Yep.

yschimke:
@swankjesse sorry, misread as approved, instead of just a comment.  Let me know if you want me to revert.

Issue #6390, opened 2020-11-04
Confirm client cipher_suites order is maintained


yschimke:
Looks like we use the SSLSocket ordering over the client. 

```
  private fun supportedSpec(sslSocket: SSLSocket, isFallback: Boolean): ConnectionSpec {
    var cipherSuitesIntersection = if (cipherSuitesAsString != null) {
      sslSocket.enabledCipherSuites.intersect(cipherSuitesAsString, CipherSuite.ORDER_BY_NAME)
    } else {
      sslSocket.enabledCipherSuites
    }
```

https://tools.ietf.org/html/rfc5246

>    The cipher suite list, passed from the client to the server in the
   ClientHello message, contains the combinations of cryptographic
   algorithms supported by the client in order of the client's
   preference (favorite choice first).  Each cipher suite defines a key
   exchange algorithm, a bulk encryption algorithm (including secret key
   length), a MAC algorithm, and a PRF.  The server will select a cipher
   suite or, if no acceptable choices are presented, return a handshake
   failure alert and close the connection.  If the list contains cipher
   suites the server does not recognize, support, or wish to use, the
   server MUST ignore those cipher suites, and process the remaining
   ones as usual.

yschimke:
My other question from the spec is what is meant by client? The app developer? OkHttp? The JSSE provider?

yschimke:
From https://stackoverflow.com/questions/64672259/how-can-i-preserve-the-order-of-my-ciphersuite-list-in-okhttp/64676161#64676161

swankjesse:
If we change which order we prefer, we should make sure our released order is stable from what we've been shipping. There's potential performance regressions from changing which cipher suite is used!

yschimke:
A theoretical bug ugly fix https://github.com/yschimke/okhttp/pull/6/files

cc @swankjesse

yschimke:
Is this worth fixing for 4.10? Or should we add test of existing behaviour now, then put out a 4.11.0-RC1 and ask interested parties to test?

swankjesse:
I have a slight preference to hold until 4.11 just cause 4.10 is already pretty big.

yschimke:
JDK 13 now defaults to server cipher suite ordering, which may affect testing.

https://seanjmullan.org/blog/2019/08/05/jdk13

ogolovanov:
Hi.

Its almost 2 years since bugfix by @yschimke ( https://github.com/square/okhttp/pull/6407 ).
[...]

PR #6400, opened 2020-11-07
Test current behaviour of cipher suite ordering

yschimke:
First PR to confirm ordering of existing functionality.  Relates to https://github.com/square/okhttp/issues/6390

### DIFF
--- a/okhttp/src/test/java/okhttp3/CallHandshakeTest.kt
+++ b/okhttp/src/test/java/okhttp3/CallHandshakeTest.kt
@@ -1,0 +1,190 @@
+/*
+ * Copyright (C) 2020 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okhttp3
+
+import mockwebserver3.MockResponse
+import mockwebserver3.MockWebServer
+import okhttp3.testing.PlatformRule
+import okhttp3.tls.internal.TlsUtil.localhost
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.extension.RegisterExtension
+import javax.net.ssl.SSLSocket
+
+class CallHandshakeTest {
+  private lateinit var client: OkHttpClient
+  private lateinit var server: MockWebServer
+  val handshakeCertificates = localhost()
+
+  @RegisterExtension
+  @JvmField
+  val clientTestRule = OkHttpClientTestRule()
+
+  @RegisterExtension
+  @JvmField
+  var platform = PlatformRule()
+
+  private lateinit var handshakeEnabledCipherSuites: List<String>
+  private lateinit var defaultEnabledCipherSuites: List<String>
+  private lateinit var defaultSupportedCipherSuites: List<String>
+
+  val expectedModernTls12CipherSuites = listOf(CipherSuite.TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384)
+  val expectedModernTls13CipherSuites = listOf(CipherSuite.TLS_AES_128_GCM_SHA256)
+
+  @BeforeEach
+  fun setup(server: MockWebServer) {
+    this.server = server
+
+    server.enqueue(MockResponse().setResponseCode(200))
+
+    client = clientTestRule.newClientBuilder()
+      .sslSocketFactory(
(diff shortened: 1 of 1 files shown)


okhttp/src/test/java/okhttp3/CallHandshakeTest.kt
```diff
@@ -0,0 +1,197 @@
...
+
+  @Test
+  fun testDefaultHandshakeCipherSuiteOrderingTls12Modern() {
+    val client = makeClient(ConnectionSpec.MODERN_TLS, TlsVersion.TLS_1_2)
+
+    val handshake = makeRequest(client)
+
+    assertThat(handshake.cipherSuite).isIn(expectedModernTls12CipherSuites)
+
+    // Probably something like
```
swankjesse, line 95: According to [the doc](https://docs.google.com/spreadsheets/d/1C3FdZSlCBq_-qrVwG1KDIzNIB3Hyg_rKAcgmSzOsHyQ/edit#gid=0) it should be this:

TLS_AES_128_GCM_SHA256
TLS_CHACHA20_POLY1305_SHA256
TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256
TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384
TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384
TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256
TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256
TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA
TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA
TLS_RSA_WITH_AES_128_GCM_SHA256
TLS_RSA_WITH_AES_256_GCM_SHA384
TLS_RSA_WITH_AES_128_CBC_SHA
TLS_RSA_WITH_AES_256_CBC_SHA
SSL_RSA_WITH_3DES_EDE_CBC_SHA

okhttp/src/test/java/okhttp3/CallHandshakeTest.kt
```diff
@@ -0,0 +1,197 @@
...
+
+  @RegisterExtension
+  @JvmField
+  var platform = PlatformRule()
+
+  private lateinit var handshakeEnabledCipherSuites: List<String>
+  private lateinit var defaultEnabledCipherSuites: List<String>
+  private lateinit var defaultSupportedCipherSuites: List<String>
+
+  val expectedModernTls12CipherSuites = listOf(TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384, TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA384)
```
swankjesse, line 49: why is this list so short?

okhttp/src/test/java/okhttp3/CallHandshakeTest.kt
```diff
@@ -0,0 +1,197 @@
...
+    // TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256
+    // TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384
+    // TLS_RSA_WITH_AES_256_GCM_SHA384
+    // TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
+    // TLS_RSA_WITH_AES_128_GCM_SHA256
+    // TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA
+    // TLS_RSA_WITH_AES_256_CBC_SHA
+    // TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA
+    // TLS_RSA_WITH_AES_128_CBC_SHA
+    assertThat(handshakeEnabledCipherSuites).containsExactlyElementsOf(expectedConnectionCipherSuites(client))
```
swankjesse, line 136: this isn’t asserting the order. Is that unexpected?

swankjesse (commented):
Good start but this should call out that it’s asserting what we do, not necessarily what we want.

swankjesse (approved):

yschimke:
> 
> 
> Good start but this should call out that it’s asserting what we do, not necessarily what we want.

Consider this a toehold test.

I think two follow up streams

1) Refine our current and general expectations -> TODO in follow ups
2) Fix the bug -> https://github.com/yschimke/okhttp/pull/6/files

okhttp/src/test/java/okhttp3/CallHandshakeTest.kt
yschimke: I'm going to land with it more relaxed and passing then we can define exactly what our current and future expectations are.

okhttp/src/test/java/okhttp3/CallHandshakeTest.kt
yschimke: Given a fixed set of platforms we test against in CI, there are only a certain number of likely results for the first cipher.  But at least if we curate any valid high strength choices here, it's a smoke test if it ever breaks and starts using a weaker one.

okhttp/src/test/java/okhttp3/CallHandshakeTest.kt
yschimke: It is asserting the order according to javadoc

```
   * Same as {@link #containsExactly(Object...)} but handle the {@link Iterable} to array conversion : verifies that
   * actual contains exactly the elements of the given iterable and nothing else <b>in the same order</b>.
```

PR #6401, opened 2020-11-07
Workaround native image params issue

yschimke:


okhttp/src/main/kotlin/okhttp3/internal/cache/DiskLruCache.kt
```diff
@@ -34,12 +25,11 @@ import okhttp3.internal.isCivilized
 import okhttp3.internal.okHttpName
 import okhttp3.internal.platform.Platform
 import okhttp3.internal.platform.Platform.Companion.WARN
-import okio.BufferedSink
-import okio.ForwardingSource
-import okio.Sink
-import okio.Source
-import okio.blackholeSink
-import okio.buffer
+import okio.*
```
swankjesse, line 28: Nit: wildcards

okhttp/src/test/java/okhttp3/internal/http2/HttpOverHttp2Test.java
```diff
@@ -73,12 +54,14 @@
 import okio.BufferedSink;
 import okio.GzipSink;
 import okio.Okio;
+import org.jetbrains.annotations.NotNull;
```
swankjesse, line 57: Don't import this. All our Java values are non-null unless otherwise specified

swankjesse (approved):
Nice

PR #6405, opened 2020-11-08
Use clean assumptions APIs

yschimke:
Junit4 assumptions are not handled cleanly by Junit5, so this may cascade.

Assertj tries to handle this

```


    private static RuntimeException assumptionNotMet(AssertionError assertionError) throws ReflectiveOperationException {
        Class<?> assumptionClass = getAssumptionClass("org.junit.AssumptionViolatedException");
        if (assumptionClass != null) {
            return assumptionNotMet(assumptionClass, assertionError);
        } else {
            assumptionClass = getAssumptionClass("org.opentest4j.TestAbortedException");
            if (assumptionClass != null) {
                return assumptionNotMet(assumptionClass, assertionError);
            } else {
                assumptionClass = getAssumptionClass("org.testng.SkipException");
                if (assumptionClass != null) {
                    return assumptionNotMet(assumptionClass, assertionError);
                } else {
                    throw new IllegalStateException("Assumptions require JUnit, opentest4j or TestNG on the classpath");
                }
            }
        }
    }
```


okhttp/src/test/java/okhttp3/internal/http2/HttpOverHttp2Test.java
```diff
@@ -81,6 +62,20 @@
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.ArgumentsSource;
 
+import javax.annotation.Nullable;
```
swankjesse, line 65: this imports shuffling is annoying. Our tools are fighting us.

swankjesse (approved):

okhttp/src/test/java/okhttp3/internal/http2/HttpOverHttp2Test.java
```diff
@@ -81,6 +62,20 @@
...
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.Authenticator;
+import java.net.HttpURLConnection;
+import java.net.SocketTimeoutException;
+import java.time.Duration;
+import java.util.Arrays;
+import java.util.List;
+import java.util.concurrent.*;
```
swankjesse, line 76: ... especially this.

okhttp/src/test/java/okhttp3/internal/http2/HttpOverHttp2Test.java
yschimke: Agreed.  I don't have one computer, using a Windows one mainly and every non standard setup makes it brittle.

PR #6407, opened 2020-11-08
Fix client cipher suite ordering

yschimke:
```
      // 3 options here for ordering
      // 1) Legacy Platform - based on the Platform/Provider existing ordering in
      // sslSocket.enabledCipherSuites
      // 2) OkHttp Client - based on MODERN_CIPHER_SUITES source code ordering, or caller override.
      // 3) Caller specified, but assuming the visible defaults in MODERN_CIPHER_SUITES are rejigged
      // to match legacy i.e. the platform/provider
      //
      // Opting for 3 here and in MODERN_CIPHER_SUITES via platformOrdered
```


yschimke:
Addresses https://github.com/square/okhttp/issues/6390

yschimke:
Not to land before 4.11

PR #6408, opened 2020-11-08
Fix Android tests

yschimke:
- Junit 5
- bouncy castle fixes
- lets encrypt test

PR #6409, opened 2020-11-10
Android dependency upgrade

yschimke:
Splitting out of https://github.com/square/okhttp/pull/6408 to make these separate actions


yschimke:
See notes on jdk15to18 https://bouncycastle.org/latest_releases.html

swankjesse (approved):

PR #6412, opened 2020-11-12
Regression Test Module

yschimke:
Regression Tests for Android.  Only using public APIs and externally observable behaviour.

Follow ups
- get it working for JVM also
- solve how to get it working across various versions, probably involves converting the tests back to java instead of kotlin for source compatibility.

### DIFF
--- a/regression-test/src/androidTest/java/okhttp/regression/LetsEncryptTest.kt
+++ b/regression-test/src/androidTest/java/okhttp/regression/LetsEncryptTest.kt
@@ -28,9 +28,10 @@ import org.junit.runner.RunWith
 import javax.net.ssl.SSLHandshakeException
 
 /**
- * OkHttp.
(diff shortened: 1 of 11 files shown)


regression-test/src/androidTest/java/okhttp/regression/LetsEncryptTest.kt
```diff
@@ -0,0 +1,56 @@
...
+import okhttp3.Request
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertTrue
+import org.junit.Assert.fail
+import org.junit.Test
+import org.junit.runner.RunWith
+import javax.net.ssl.SSLHandshakeException
+
+/**
+ * OkHttp.
```
swankjesse, line 31: Copy paste?

swankjesse (approved):
This is brilliant

PR #6416, opened 2020-11-12
Simplify config for native image

yschimke:
Embed required options into okhttp jar.


okhttp/src/main/resources/META-INF/native-image/okhttp/okhttp/native-image.properties
```diff
@@ -0,0 +1 @@
+Args = -H:+AddAllCharsets -H:EnableURLProtocols=http,https --enable-https
```
swankjesse, line 1: sweet

swankjesse (approved):

PR #6422, opened 2020-11-14
Switch tests to java for source compatibility

yschimke:
With this change editing the build.gradle allows testing against various versions as a regression test.  Maybe improvements to come later, but for now solve the quick testing of combinations of OkHttp versions and Android versions.


regression-test/src/androidTest/java/okhttp/regression/LetsEncryptTest.java
```diff
@@ -0,0 +1,60 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import javax.net.ssl.SSLHandshakeException;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+/**
+ * Let's Encrypt expiring root test.
```
swankjesse, line 34: 🙏🏻

swankjesse (approved):

PR #6433, opened 2020-11-20
Migrate Batch of Param Tests

yschimke:


okhttp/src/test/java/okhttp3/HttpUrlTest.java
```diff
@@ -17,409 +17,466 @@
...
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ArgumentsSource;
 
-import static java.util.Arrays.asList;
 import static java.util.Collections.singletonList;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.jupiter.api.Assertions.fail;
 
-@RunWith(Parameterized.class)
 public final class HttpUrlTest {
```
swankjesse, line 31: For tests like this should we just do subclasses? That's probably a lot less invasive in the syntax.

swankjesse (approved):

PR #6435, opened 2020-11-21
Support Android in Connection Spec Tests

yschimke:
Ciphers available changes on Android


yschimke:
I can't land this in okhttp main tests because of Build dependency from Android.  So generally getting advice on the diff.

okhttp/src/main/kotlin/okhttp3/internal/platform/Platform.kt
```diff
@@ -273,5 +274,13 @@ open class Platform {
       }
       return result.readByteArray()
     }
+
+    fun androidSdkVersion(): Int? {
```
swankjesse, line 278: move to PlatformRuleTest?

okhttp/src/test/java/okhttp3/ConnectionSpecTest.java
```diff
@@ -224,14 +227,31 @@
...
-        CipherSuite.TLS_RSA_WITH_RC4_128_MD5.javaName(),
+        CipherSuite.TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256.javaName(),
+        CipherSuite.TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA.javaName(),
     });
 
     applyConnectionSpec(tlsSpec, sslSocket, false);
-    assertThat(sslSocket.getEnabledCipherSuites()).containsExactly(
-        CipherSuite.TLS_RSA_WITH_RC4_128_SHA.javaName(),
-        CipherSuite.TLS_RSA_WITH_RC4_128_MD5.javaName());
+    if (Platform.Companion.isAndroid()) {
```
swankjesse, line 235: Rad!

swankjesse (approved):
Nice

PR #6436, opened 2020-11-21
Launcher for quick tests

yschimke:
Use junit5 tests to tag slow tests and run everything but across the project.

image


yschimke:
image

PR #6437, opened 2020-11-21
Launcher for quick tests

yschimke:
Use junit5 tests to tag slow tests and run everything but across the project 
image

### DIFF
--- a/okhttp/src/test/java/okhttp3/KotlinSourceModernTest.kt
+++ b/okhttp/src/test/java/okhttp3/KotlinSourceModernTest.kt
@@ -76,6 +76,8 @@ import okio.BufferedSink
 import okio.BufferedSource
 import okio.ByteString
 import okio.Timeout
+import org.junit.jupiter.api.Assumptions.assumeFalse
+import org.junit.jupiter.api.BeforeEach
 import org.junit.jupiter.api.Disabled
 import org.junit.jupiter.api.Test
 
@@ -95,6 +97,11 @@ import org.junit.jupiter.api.Test
 )
 @Disabled
 class KotlinSourceModernTest {
+  @BeforeEach
+  fun disabled() {
+    assumeFalse(true)
+  }
+
   @Test
   fun address() {
     val address: Address = newAddress()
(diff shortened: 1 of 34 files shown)


swankjesse (commented):
What qualifies a test as fast or slow? Excluding CallTest surprises me; it's big but is it also slow?

Maybe I'd argue we should constrain this to tests that sleep?

yschimke:
image

I'll move from the top of CallTest to individual slow tests, at least one takes 10 seconds.

yschimke:
> What qualifies a test as fast or slow? Excluding CallTest surprises me; it's big but is it also slow?
> 
> Maybe I'd argue we should constrain this to tests that sleep?

Ignoring the why it's slow aspect, I found myself not running tests locally because they take minutes.  Including most of CallTest, and it taking 12 seconds seems like a reasonable compromise.

okhttp/src/test/java/okhttp3/KotlinSourceModernTest.kt
```diff
@@ -95,6 +97,11 @@ import org.junit.jupiter.api.Test
 )
 @Disabled
 class KotlinSourceModernTest {
+  @BeforeEach
+  fun disabled() {
```
swankjesse, line 101: What's this for?

swankjesse (approved):
Any guidelines for authors of new tests? What's the cutoff?

yschimke:
> 
> 
> Any guidelines for authors of new tests? What's the cutoff?

2 second is Slow
> 0.5 is Slowish

But overall goal is that probably more like, "across whole project"

!Slow -> 45 seconds
!Slow & ! Slowish -> 10 seconds

okhttp/src/test/java/okhttp3/KotlinSourceModernTest.kt
yschimke: @Disabled wasn't working in this setup, it was running it anyway.  And timing out and spewing errors from TODO().  There might be some correct fix, but like the Android 5 junit runner which doesn't support automatic registration of extensions, each new integration has a non zero chance of not handling some feature of test running.

Issue #6413, opened 2020-11-12
Network Error Logging Support in OkHttp


tuberd:
Hey OkHttp folks,
Was wondering if there was any plans to support [Network Error Logging](https://www.w3.org/TR/network-error-logging/) natively in OkHttp.  It's supported in [cronet through experimental features](https://source.chromium.org/chromium/chromium/src/+/master:components/cronet/url_request_context_config.cc;l=138;drc=4cd506f8d178ede30acc775708218ed960dfa873) but in talking to a handful of app teams, it would be great to be able to see reachability data via okhttp as well.

Thanks so much!

yschimke:
A few thoughts

1) This doesn't feel like something we would turn on by default for all clients.  
2) Similar complexity additional features tend to be implemented as separate libraries like https://github.com/babylonhealth/certificate-transparency-android
3) Detecting the specific errors is really awkward across so many different platforms (Android versions, JDK versions, Providers).  

On the last point it seems like a lot of work to correctly root cause every error type automatically, this is a look at just one class of errors

https://github.com/square/okhttp/issues/6411

And with Conscrypt - we've asked for ways to cleanly understand causes of handshake failures but it isn't something cleanly supported at the moment https://github.com/google/conscrypt/issues/847 

If we could even do 3) reliably, I think we could wrap more exceptions into ones with something like a public API for errors, even without the reporting aspect.  Make it possible for clients to catch those and explain to end users why the connection is faulty.

swankjesse:
Yeah I think you could probably do this with OkHttp, but I don’t think it’s a good fit within the library.

swankjesse:
But one thing we could do, which would be pretty nice, is to report our errors using the taxonomy here:
https://www.w3.org/TR/network-error-logging/#predefined-network-error-types

yschimke:
Let's reopen and define not as an interceptor but a gradual move to errors containing enough structured data to sometimes report these.

Maybe work with Conscrypt folk to make this happen a layer lower also.

yschimke:
Put an exploration in the PR above.

yschimke:
Also wondering how it could relate to https://www.chromium.org/developers/design-documents/network-stack/netlog

Which are effective as a much more user friendly (cooperative) version of wireshark pcaps. chrome://net-export/

yschimke:
Or if we wait long enough

https://tools.ietf.org/html/draft-marx-qlog-main-schema-02#section-4.3.3
[...]

PR #6440, opened 2020-11-21
NEL Error Logging

yschimke:
A brief exploration of what NEL error logging could look like in OkHttp?

https://github.com/square/okhttp/issues/6413


swankjesse:
I’m reluctant to introduce our own exception types that replace or wrap the built-in exception types. If we do it really well it’ll help users to debug their issues; but it also breaks all of our StackOverflow answers.

It’s also difficult to do fully because of where exceptions are thrown. Some are through Okio, some through JSSE/Conscrypt, and some through the JDK. Wrapping exceptions everywhere is going to be quite difficult to do well.

yschimke:
@swankjesse any thoughts on suppressed exception as a carrier for the detail we want?

swankjesse:
Backing up a bit – the ultimate goal is to improve metrics right? So operators can tell the difference between a TLS failure, a DNS failure, a timeout failure, etc.?

yschimke:
> 
> 
> Backing up a bit – the ultimate goal is to improve metrics right? So operators can tell the difference between a TLS failure, a DNS failure, a timeout failure, etc.?

I'd say site operators (i provide some APIs across multiple platforms used by other developers and I'd like to know how and possibly why it's failing).  

Also as an app developer using OkHttp and Conscrypt (directly or through Android), I'd like to be able to self diagnore problems and report directly without asking for help in stackoverflow or OkHttp forums.

A wise man said

> But one thing we could do, which would be pretty nice, is to report our errors using the taxonomy here:
https://www.w3.org/TR/network-error-logging/#predefined-network-error-types

Hence moving this down into Conscrypt also https://github.com/google/conscrypt/issues/922

Also after the fact there is often more we could have done if details of the failure were accessible without assumptions of the implementation.  For example

```
  try {
    sslSocket.startHandshake()
  } catch (s: SSLHandshakeException) {
    val r = generateSequence<Throwable>(s) { it.cause }.find { it is CertPathValidatorException } as? CertPathValidatorException

    if (r != null) {
      println(r.certPath.certificates.map { (it as X509Certificate).subjectDN })
    }
  }
```

Can we as OkHttp pull this out and make it a clear error solution like certificate pinning failures?

PR #6442, opened 2020-11-22
RFC - would we drop support for alpn-boot *post* 4.9

yschimke:
Expecting some pushback, but we have ALPN support without alpn-boot on all supported JVMs since it's been backported. It's effectively dead code.  

I contend 3 things
- JDK 8 now supports ALPN without this library - it's a noop after 8u252
- Someone who is on JDK8 strictly < 252 is unlikely to also be using OkHttp > 4.9 with kotlin dependencies
- OkHttp 3.12.x will continue to support people with JDK 7+ and alpn-boot


swankjesse:
I don't think users have that much control over their JVM version. Imagine a Gradle plugin that uses OkHttp: it runs on whatever JAVA_HOME is, and developers don't upgrade their JVMs without prompting.

yschimke:
> 
> 
> I don't think users have that much control over their JVM version. Imagine a Gradle plugin that uses OkHttp: it runs on whatever JAVA_HOME is, and developers don't upgrade their JVMs without prompting.

Is there ever going to be a better time to do it? "If you run JDK 8, but are stuck on 242 or earlier, then either don't update past OkHttp 4.9 or drop to 3.12.x, or upgrade to 252".  

Also Isn't that a terrible example, will it even work with alpn-boot confusion? https://www.eclipse.org/jetty/documentation/current/alpn-chapter.html.  Does it work in those cases?

I suspect the annoyed users will be those deploying frequently a modern app onto older software and system stack.  e.g. running a WAR inside WebSphere on 1.8.0u45

yschimke:
This discussion for me is driven by a few concerns

- We are supporting more use cases in code (native image, OSGi, Android tests), thankfully we are also choosing to not support some like RoboVM.
- We have more platforms to run in CI.
- There is non zero maintenance cost e.g. jdk8alpn CI is breaking with some byzantine* build failure.

* using big words I hardly understand makes me feel smart :)

swankjesse:
I don't want to orphan anyone on 4.x, otherwise it's another release to backport security fixes to. Right now we're backwards compatible and there's no reason to not upgrade; if we introduce one that costs us too.

yschimke:
> 
> 
> I don't want to orphan anyone on 4.x, otherwise it's another release to backport security fixes to. Right now we're backwards compatible and there's no reason to not upgrade; if we introduce one that costs us too.

OK, what about a sunset date? end of June 2021.

PR #6443, opened 2020-11-22
Experiment with test concurrency

yschimke:
Some documentation in https://junit.org/junit5/docs/snapshot/user-guide/#writing-tests-parallel-execution

image


yschimke:
- Run with concurrency per test method in IDE.  runs in 3 seconds.
- Adjusting the "Not Slow" IDE runner to avoid tests that are flaky with concurrency
- Running CallTest concurrently via annotation

yschimke:
Closing because I forgot about the impact of the Platform test, running Platform.resetForTests()

PR #6447, opened 2020-11-24
Test Pinning with cross signed certificates

yschimke:
> Mr. Schimke, what you've just coded is one of the most insanely idiotic things I have ever read. At no point in your rambling, incoherent test cases were you even close to anything that could be considered a rational thought. Everyone in this room is now dumber for having read it. I award you no points, and may God have mercy on your soul. 

As suspected, it's a real mess when there are multiple paths to a successful chain. I'm assuming someone stable, but awkward because you need to pin against both roots.

On Android we might also want to confirm that when connecting with 4 trusted certs we get the full chain, which appears to happen on the JDK.

### DIFF
--- a/okhttp-tls/src/test/java/okhttp3/tls/CrossSigningTest.java
+++ b/okhttp-tls/src/test/java/okhttp3/tls/CrossSigningTest.java
@@ -18,6 +18,8 @@ package okhttp3.tls;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
+import java.security.Provider;
+import java.security.Security;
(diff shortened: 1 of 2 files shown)


yschimke:
image

yschimke:
Doesn't run by default but is high enough value as a check between now and January

yschimke:
@swankjesse Want to land this and test under different providers? Or park/close unless we re-implement with MWS?

swankjesse (approved):
This makes my head hurt.

yschimke:
@swankjesse It's probably better as a point in time PR. Will require updating the cert every 3 months, and the chain of trust is still in flux throughout January https://letsencrypt.org/certificates/

PR #6449, opened 2020-11-25
Patch doc shareinstance / new builder

afaucogney:
Add extra explication and context to better understand newBuilder impact.

inform that the configuration of the original instance is kept
inform that a configuration of the original instance may be overriden

### DIFF
--- a/okhttp/src/main/kotlin/okhttp3/OkHttpClient.kt
+++ b/okhttp/src/main/kotlin/okhttp3/OkHttpClient.kt
@@ -89,7 +89,7 @@ import org.codehaus.mojo.animal_sniffer.IgnoreJRERequirement
  * ```
  *
  * This example shows a call with a short 500 millisecond read timeout and a 1000 millisecond 
- * write timout. Original configuration is kept, but can be overriden.
+ * write timeout. Original configuration is kept, but can be overriden.
  *
  * ```
  * OkHttpClient eagerClient = client.newBuilder()
--- a/okhttp/src/main/kotlin/okhttp3/OkHttpClient.kt
+++ b/okhttp/src/main/kotlin/okhttp3/OkHttpClient.kt
@@ -77,9 +77,16 @@ import org.codehaus.mojo.animal_sniffer.IgnoreJRERequirement
  *
  * You can customize a shared OkHttpClient instance with [newBuilder]. This builds a client that
  * shares the same connection pool, thread pools, and configuration. Use the builder methods to
- * configure the derived client for a specific purpose.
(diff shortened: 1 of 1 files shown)


okhttp/src/main/kotlin/okhttp3/OkHttpClient.kt
```diff
@@ -77,9 +77,19 @@ import org.codehaus.mojo.animal_sniffer.IgnoreJRERequirement
...
+ * This example shows the single instance with default configurations.
+ * 
+ * ```
+ * public final OkHttpClient client = new OkHttpClient.Builder()
+ *     .readTimeout(1000, TimeUnit.MILLISECONDS)
+ *     .writeTimeout(1000, TimeUnit.MILLISECONDS)
+ *     .build();
+ * ```
+ *
+ * This example shows a call with a short 500 millisecond read timeout and a 1000 millisecond
```
yschimke, line 91: doesn't match above. 500 vs 1000.

okhttp/src/main/kotlin/okhttp3/OkHttpClient.kt
```diff
@@ -77,9 +77,19 @@ import org.codehaus.mojo.animal_sniffer.IgnoreJRERequirement
...
+ * 
+ * ```
+ * public final OkHttpClient client = new OkHttpClient.Builder()
+ *     .readTimeout(1000, TimeUnit.MILLISECONDS)
+ *     .writeTimeout(1000, TimeUnit.MILLISECONDS)
+ *     .build();
+ * ```
+ *
+ * This example shows a call with a short 500 millisecond read timeout and a 1000 millisecond 
+ * write timout. Original configuration is kept, but can be overriden.
```
yschimke, line 92:
```suggestion
 * write timeout. Original configuration is kept, but can be overriden.
```

okhttp/src/main/kotlin/okhttp3/OkHttpClient.kt
afaucogney: This phrase is for the example at bottom. 1000 write timeout inherited from the original and 500 overridden with the new builder configuration. Isn't it ? What's your point ? @yschimke

yschimke: I read in the context of the one above.  makes sense.

yschimke:
@swankjesse do we need the CLA for doc comments? https://github.com/square/okhttp/blob/master/CONTRIBUTING.md

afaucogney:
I have done the CLA since the beginning

PR #6464, opened 2020-12-09
Add BOM mention to Releases section

vbuberen:
It is hard to find evidence of OkHttp BOM existing, since it is mentioned only in [release notes to 4.4.0 version](https://github.com/square/okhttp/blob/master/CHANGELOG.md#version-440). For example, I didn't know about it until needed to read through changes for a few versions. Today had a conversation with a friend who also wasn't aware about BOM being available.

To help more people discover OkHttp BOM I updated the README to mention it clearly, so it will be easy to find.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,6 +107,18 @@ implementation("com.squareup.okhttp3:okhttp:4.9.0")
 
 Snapshot builds are [available][snap]. [R8 and ProGuard][r8_proguard] rules are available.
 
+Also, we have a [bill of materials (BOM)][bom] available to help you keep OkHttp artifacts up to date and be sure about versions compatibility.
+
+```kotlin
+    dependencies {
+       // define a BOM and its version
+       implementation(platform("com.squareup.okhttp3:okhttp-bom:4.9.0"))
+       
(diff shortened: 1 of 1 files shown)


swankjesse (approved):

README.md
```diff
@@ -107,6 +107,18 @@ implementation("com.squareup.okhttp3:okhttp:4.9.0")
...
 Snapshot builds are [available][snap]. [R8 and ProGuard][r8_proguard] rules are available.
 
+Also, we have a [bill of materials (BOM)][bom] available to help you keep OkHttp artifacts up to date and be sure about versions compatibility.
+
+```kotlin
+    dependencies {
+       // define a BOM and its version
+       implementation(platform("com.squareup.okhttp3:okhttp-bom:4.9.0"))
+       
+       // define any required OkHttp artifacts without verion
```
yschimke, line 117: typo: version

vbuberen: Good catch. I will fix it. Wanted to add the same snippet for snapshot verisons

yschimke: Already fixed it. Not sure why I mentioned it to be honest. Thanks for adding this.

Issue #6116, opened 2020-06-05
Customize KeyStore for HandshakeCertificates


MGaetan89:
Hello,

I am trying to establish a connection with my server by using two way SSL.
For that, I'm using `HandshakeCertificates` as follow:
```kotlin
val deviceCertificate = keyStore.getCertificate(CERTIFICATE_ALIAS) as X509Certificate
val publicKey = deviceCertificate.publicKey
val privateKey = getKeyStore().getKey(KEY_ALIAS, null) as PrivateKey
val keyPair = KeyPair(publicKey, privateKey)
val heldCertificate = HeldCertificate(keyPair, deviceCertificate)

handshakeBuilder.heldCertificate(heldCertificate)
```

Which I then set on my OkHttpClient with:
```kotlin
val handshake = handshakeBuilder.build()

clientBuilder.sslSocketFactory(handshake.sslSocketFactory(), handshake.trustManager)
```

But when I run my app, it crashes with:
```
java.security.KeyStoreException: java.lang.NullPointerException: Attempt to invoke virtual method 'int java.lang.String.length()' on a null object reference
        at com.android.org.bouncycastle.jcajce.provider.keystore.bc.BcKeyStoreSpi.engineSetKeyEntry(BcKeyStoreSpi.java:689)
        at java.security.KeyStore.setKeyEntry(KeyStore.java:1179)
        at okhttp3.tls.internal.TlsUtil.newKeyManager(TlsUtil.kt:95)
        at okhttp3.tls.HandshakeCertificates$Builder.build(HandshakeCertificates.kt:175)
        at com.myapp.common.security.data.CommonSecurityProvider$Companion.getSslSocketFactoryConfig(CommonSecurityProvider.kt:112)
        at com.myapp.common.network.di.CommonNetworkModuleKt$commonNetworkModule$1$7.invoke(CommonNetworkModule.kt:64)
        at com.myapp.common.network.di.CommonNetworkModuleKt$commonNetworkModule$1$7.invoke(Unknown Source:4)
        at org.koin.core.instance.InstanceFactory.create(InstanceFactory.kt:50)
        	... 139 more
```

From my investigation, it could be because OkHttp is using its own `KeyStore` when building the `HandshakeCertificates`, while my app is using `AndroidKeyStore`. Is there a way to customize this? Or am I doing something wrong?

[...]

Issue #5684, opened 2020-01-02
BC HandshakeCertificates NullPointerException


sonnyCZE:
Hello,

I have probably found bug in ```HandshakeCertificates.Builder```.  When I am trying to build object with client certificate the BouncyCastle ```BcKeyStoreSpi.engineSetKeyEntry``` throws NPE. I am using okhttp and okhttp-tls version 4.1.0 but even older versions acts the same way. I am able to reproduce this on multiple devices (Android 9, 8.1, 7.0). Also I have already tried to disable proguard but still it is throwing the NPE.

Code
```
PrivateKey privateKey = KeyChain.getPrivateKey(context, certAlias);
X509Certificate[] certChain = KeyChain.getCertificateChain(context, certAlias);
X509Certificate mdmCert = certChain[0];

HandshakeCertificates.Builder builder = new HandshakeCertificates.Builder();
builder.addPlatformTrustedCertificates();

for (X509Certificate cert : certChain) {
    builder.addTrustedCertificate(cert);
}

builder.heldCertificate(        //if this is commented then everything is ok
        new HeldCertificate(
new KeyPair(
        mdmCert.getPublicKey(),
        privateKey
),
mdmCert
        ),
        certChain
);

HandshakeCertificates clientCertificates = builder.build(); //this throws exception
```

Stacktrace
```
Caused by: java.security.KeyStoreException: java.lang.NullPointerException: Attempt to invoke virtual method 'int java.lang.String.length()' on a null object reference
    at com.android.org.bouncycastle.jcajce.provider.keystore.bc.BcKeyStoreSpi.engineSetKeyEntry(BcKeyStoreSpi.java:685)
    at java.security.KeyStore.setKeyEntry(KeyStore.java:1179)
    at okhttp3.tls.internal.TlsUtil.newKeyManager(TlsUtil.kt:85)
    at okhttp3.tls.HandshakeCertificates$Builder.build(HandshakeCertificates.kt:144)
```

swankjesse:
Could you please provide the full stacktrace? An executable test case would also be nice.

sonnyCZE:
Right now I have only certificates from our clients so I can't share it. But the problem is with multiple certificates from multiple clients. Thank you

Complete code:
```
import android.annotation.SuppressLint;
import android.content.Context;
import android.os.AsyncTask;
import android.os.Bundle;
import android.security.KeyChain;
import android.security.KeyChainException;

import java.security.KeyPair;
import java.security.PrivateKey;
[...]

Issue #5172, opened 2019-06-06
NPE building HandshakeCertificates when Proguard is enabled


nimeacuerdo:
Hi there,

I'm using the following code with version **3.13.1** to create a `HandshakeCertificates` that is going to be used to configure `MockWebserver` to enable https in acceptance tests where our api is mocked this way, running on **Android P** emulators.

```
HandshakeCertificates serverCertificates = new HandshakeCertificates.Builder()
		.heldCertificate(certificate)
		.build();

mockWebServer.useHttps(serverCertificates.sslSocketFactory(), false);
```

as described in [this example gist](https://github.com/square/okhttp/blob/master/samples/guide/src/main/java/okhttp3/recipes/HttpsServer.java) you published.

Everything works fine (configuring the certificate in client side, handshaking of the connection, etc), without **Proguard** but when we enable it the following error occurs while initialising test code:

```
java.lang.NullPointerException: Attempt to invoke interface method 'java.lang.String java.security.PublicKey.getAlgorithm()' on a null object reference
        at java.security.KeyStore$PrivateKeyEntry.<init>(KeyStore.java:576)
        at java.security.KeyStore$PrivateKeyEntry.<init>(KeyStore.java:526)
        at java.security.KeyStoreSpi.engineGetEntry(KeyStoreSpi.java:485)
        at java.security.KeyStore.getEntry(KeyStore.java:1560)
        at com.android.org.conscrypt.KeyManagerImpl.<init>(KeyManagerImpl.java:71)
        at com.android.org.conscrypt.KeyManagerFactoryImpl.engineGetKeyManagers(KeyManagerFactoryImpl.java:114)
        at javax.net.ssl.KeyManagerFactory.getKeyManagers(KeyManagerFactory.java:305)
        at okhttp3.tls.internal.TlsUtil.newKeyManager(SourceFile:101)
        at okhttp3.tls.HandshakeCertificates$Builder.build(SourceFile:164)
```

so I don't know if there is something I should be excluding explicitly in our config in order to avoid this problem.

The certificate used to configure `HandshakeCertificates` is provided here:

```
@Provides
@Singleton
HeldCertificate provideHeldCertificate() {
	return new HeldCertificate.Builder()
			.addSubjectAlternativeName("localhost")
			.build();
}
```

swankjesse:
Yuck. If you can figure out why, please share.

nimeacuerdo:
Got it working by adding the following to the test-specific proguard configuration:

[...]

PR #6468, opened 2020-12-11
Fail on unencodable keys

yschimke:
Fail early and predictably on unencodable keys.

Meant to catch

https://github.com/square/okhttp/issues/5172
https://github.com/square/okhttp/issues/6116
https://github.com/square/okhttp/issues/5684


yschimke:
Can't work out a good test here, positive tests should be shown.  But will have a hard time testing android.security.KeyChain on Android, and testing this on JDK is possibly tough without something like an OpenSC device/PKCS11

yschimke:
ping @swankjesse

swankjesse (approved):

PR #6471, opened 2020-12-13
Android Client Auth demo

yschimke:


regression-test/src/androidTest/java/okhttp/regression/keys/ClientAuthAndroidTest.java
```diff
@@ -0,0 +1,175 @@
...
+                new String[]{"RSA", "DSA"},  // List of acceptable key types. null for any
+                null,  // issuer, null for any
+                "server.cryptomix.com",  // host name of server requesting the cert, null if unavailable
+                443,  // port of server requesting the cert, -1 if unavailable
+                "mykey"); // alias to preselect, null if unavailable
+
+        PrivateKey pk = KeyChain.getPrivateKey(InstrumentationRegistry.getInstrumentation().getTargetContext(), "mykey");
+        X509Certificate[] chain = KeyChain.getCertificateChain(InstrumentationRegistry.getInstrumentation().getTargetContext(), "mykey");
+
+        assertNotNull(pk);
```
yschimke, line 154: Fails here if you have commented out choosePrivateKeyAlias

regression-test/src/androidTest/java/okhttp/regression/keys/ClientAuthAndroidTest.java
```diff
@@ -0,0 +1,175 @@
...
+    @Test
+    public void keyStoreCertificate() throws IOException, NoSuchAlgorithmException, KeyManagementException, KeyStoreException, KeyChainException, InterruptedException {
+        TrustManagerFactory trustManagerFactory = TrustManagerFactory.getInstance(TrustManagerFactory.getDefaultAlgorithm());
+        trustManagerFactory.init((KeyStore) null);
+        X509TrustManager trustManager = (X509TrustManager) trustManagerFactory.getTrustManagers()[0];
+
+        CountDownLatch latch = new CountDownLatch(1);
+
+        String[] selectedAlias = new String[1];
+        KeyChain.choosePrivateKeyAlias(null, // TODO launch with an activity
```
yschimke, line 140: Fails here as I'll need to launch an activity to test at least first time.  Maybe some permission is available to allow managing keys and certs.

PR #6482, opened 2020-12-18
Simplify Kotlin Examples with main.kts scripts

yschimke:
Thought experiment about using kotlin scripting, which works with nice completion in basically any Intellij or Android Studio project.  So these samples are trivial to cut and paste and have running without a surrounding project including okhttp.

Also run straight from the command line with `chmod u+x script.main.kts`

https://github.com/Kotlin/kotlin-script-examples/blob/master/jvm/main-kts/MainKts.md


samples/guide/kts/post.main.kts
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env -S kotlinc-jvm -nowarn -script
+
+@file:Repository("https://jcenter.bintray.com")
```
yschimke, line 3: may not need this anymore...

yschimke:
@swankjesse thoughts on this? Runs as a bash shell (if you have kotlin installed), runs in Intellij, completely self contained (doesn't use libraries from containing project).

yschimke:
Went with a stack overflow Q&A https://stackoverflow.com/questions/66364993/how-to-make-a-reproduction-of-a-bug-for-okhttp/66364994#66364994

Issue #6363, opened 2020-10-31
Conform lockingSourceCount is correct


swankjesse:
From @QingShiLuoGu on https://github.com/square/okhttp/issues/5761#issuecomment-719928420

> When we call Entry.newSource(), the "lockingSourceCount" will increase, but not in Editor.newSource(). If we read the cache through Editor.newSource(), will this problem happen again?

PR #6490, opened 2020-12-23
Add a test case  about  "Conform lockingSourceCount is correct" 

QingShiLuoGu:
just add a failing test case for #6363  so so you can fix it.


edith2727 (approved):

edith2727 (approved):
activar stripe

AFZAL4911 (commented):
Hello dear
How is collect

yschimke:
This should now be cleanly reproducible via Okio file system 

See https://github.com/square/okhttp/blob/ac63910c010af12686b7a86a07f401592dfc4fb9/okhttp/src/test/java/okhttp3/internal/cache/DiskLruCacheTest.kt#L51

Do you want to update the test and I'll take a look?

QingShiLuoGu:
Do I need  fork the latest and then pull request agian?

QingShiLuoGu:
hello, do I need fork the latest version and then pull request again?
Or can I just edit this commit?  @yschimke

QingShiLuoGu:
I think it has been updated,please check it.  @yschimke

yschimke:
@swankjesse any opinion on this test?

I'm assuming that DiskLruCache.Snapshot, which is closeable, should have a checkNotClosed() method, and fail fast if you try to use it after closing it.  But it breaks a bunch of other tests.  

image

swankjesse:
@yschimke yeah that reads to me like a small bug? The snapshot should not be used after it is closed, though I think there's a valid use case to using a snapshot after it's enclosing cache is closed.

This class isn't public API so we can improve it's behavior where that’s useful. But we don't really need to be super aggressive about enforcing invariants unless we're worried we might inadvertently violate them.

yschimke:
Going to close this one out unless we can reproduce with a test case using public API.  We can fix this, but it likely just incurs a tiny performance hit on each cache usage.

PR #6495, opened 2020-12-30
Fail earlier on cache corruptions

yschimke:
Fail earlier on cache corruption, log warning instead of failing terminally.

### DIFF
--- a/okhttp/src/main/kotlin/okhttp3/Cache.kt
+++ b/okhttp/src/main/kotlin/okhttp3/Cache.kt
@@ -505,7 +505,7 @@ class Cache internal constructor(
         val urlLine = source.readUtf8LineStrict()
         // Choice here is between failing with a correct RuntimeException
         // or mostly silently with an IOException
-        url = urlLine.toHttpUrlOrNull() ?: throw IOException("Cache Failure for $urlLine").also {
+        url = urlLine.toHttpUrlOrNull() ?: throw IOException("Cache corruption for $urlLine").also {
           Platform.get().log("cache corruption", WARN, it)
         }
         requestMethod = source.readUtf8LineStrict()
--- a/okhttp/src/main/kotlin/okhttp3/Cache.kt
+++ b/okhttp/src/main/kotlin/okhttp3/Cache.kt
@@ -15,16 +15,7 @@
  */
 package okhttp3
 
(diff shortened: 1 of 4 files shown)


okhttp/src/main/kotlin/okhttp3/HttpUrl.kt
```diff
@@ -1258,7 +1258,7 @@ class HttpUrl internal constructor(
         this.scheme = base.scheme
       } else {
         throw IllegalArgumentException(
-            "Expected URL scheme 'http' or 'https' but no colon was found")
+            "Expected URL scheme 'http' or 'https' but no scheme was found for $input")
```
yschimke, line 1261: Happy to drop this for PII reasons if there is concern, although as a corrupted url, the debug value seems higher than normal.

okhttp/src/main/kotlin/okhttp3/Cache.kt
```diff
@@ -490,9 +500,14 @@ class Cache internal constructor(
...
      */
     @Throws(IOException::class) constructor(rawSource: Source) {
-      try {
+      rawSource.use {
         val source = rawSource.buffer()
-        url = source.readUtf8LineStrict()
+        val urlLine = source.readUtf8LineStrict()
+        // Choice here is between failing with a correct RuntimeException
+        // or mostly silently with an IOException
+        url = urlLine.toHttpUrlOrNull() ?: throw IOException("Cache Failure for $urlLine").also {
```
yschimke, line 508: 3 options here

1) runtime exception - fatal...
2) io exception only
3) io exception and log a warning

okhttp/src/main/kotlin/okhttp3/HttpUrl.kt
swankjesse: Log up to 6 characters? Like `www.ex...` ?

okhttp/src/main/kotlin/okhttp3/Cache.kt
swankjesse: change `cache failure` to `cache corruption` ?

swankjesse (approved):

Thomas-Vos:
Could this fix be backported to 4.9.x, please?

PR #6500, opened 2021-01-01
Okio Filesystem

yschimke:
Update to Okio 3.0 alpha and use new FileSystem API.


okhttp/src/main/kotlin/okhttp3/internal/cache/DiskLruCache.kt
```diff
@@ -20,13 +20,14 @@ import okhttp3.internal.cache.DiskLruCache.Editor
 import okhttp3.internal.closeQuietly
 import okhttp3.internal.concurrent.Task
 import okhttp3.internal.concurrent.TaskRunner
-import okhttp3.internal.io.FileSystem
 import okhttp3.internal.isCivilized
 import okhttp3.internal.okHttpName
 import okhttp3.internal.platform.Platform
 import okhttp3.internal.platform.Platform.Companion.WARN
 import okio.*
```
swankjesse, line 27: off-topic: IntelliJ has really become bad at Kotlin imports

okhttp/src/main/kotlin/okhttp3/internal/cache/DiskLruCache.kt
```diff
@@ -214,10 +228,11 @@ class DiskLruCache(
       if (fileSystem.exists(journalFile)) {
         fileSystem.delete(journalFileBackup)
       } else {
-        fileSystem.rename(journalFileBackup, journalFile)
+        fileSystem.atomicMove(journalFileBackup, journalFile)
       }
     }
 
+    // TODO move into API?
```
swankjesse, line 235: I don’t think so. It’s a pretty weird use case to need to know this.

okhttp/src/main/kotlin/okhttp3/internal/cache/DiskLruCache.kt
```diff
@@ -359,8 +376,13 @@ class DiskLruCache(
       } else {
         entry.currentEditor = null
         for (t in 0 until valueCount) {
-          fileSystem.delete(entry.cleanFiles[t])
-          fileSystem.delete(entry.dirtyFiles[t])
+          // TODO safer to try to delete and then cache error
```
swankjesse, line 379: cache error or catch error?

okhttp/src/main/kotlin/okhttp3/internal/cache/DiskLruCache.kt
```diff
@@ -397,10 +419,12 @@ class DiskLruCache(
     }
 
     if (fileSystem.exists(journalFile)) {
-      fileSystem.rename(journalFile, journalFileBackup)
+      fileSystem.atomicMove(journalFile, journalFileBackup)
+    }
+    fileSystem.atomicMove(journalFileTmp, journalFile)
+    if (fileSystem.exists(journalFileBackup)) {
```
swankjesse, line 425: presumably this logic can be simplified by recognizing two cases:
 - journalFile exists on line 421 in which case the backup will exist here
 - journalFile doesn’t exist on line 421, in which case the backup shouldn’t exist here?

okhttp/src/main/kotlin/okhttp3/internal/cache/DiskLruCache.kt
```diff
@@ -613,7 +640,9 @@ class DiskLruCache(
     entry.currentEditor?.detach() // Prevent the edit from completing normally.
 
     for (i in 0 until valueCount) {
-      fileSystem.delete(entry.cleanFiles[i])
+      if (fileSystem.exists(entry.cleanFiles[i])) {
+        fileSystem.delete(entry.cleanFiles[i])
```
swankjesse, line 644: I think you want `deleteIfExists` that caches a `NoSuchFileException`

okhttp/src/main/kotlin/okhttp3/internal/cache/DiskLruCache.kt
```diff
@@ -697,7 +726,11 @@ class DiskLruCache(
   @Throws(IOException::class)
   fun delete() {
     close()
-    fileSystem.deleteContents(directory)
+    if (fileSystem.exists(directory)) {
+      fileSystem.deleteRecursively(directory)
+      // TODO check this logic
+      fileSystem.createDirectory(directory)
```
swankjesse, line 732: This isn’t right because the permissions and other directory metadata might change. Instead you gotta loop over `directory` children and delete each

okhttp/src/test/java/okhttp3/CacheCorruptionTest.kt
```diff
@@ -52,25 +62,29 @@ class CacheCorruptionTest(
...
   private val cookieManager = CookieManager()
 
-  @BeforeEach fun setUp() {
+  @BeforeEach
+  fun setUp() {
     platform.assumeNotOpenJSSE()
     platform.assumeNotBouncyCastle()
     server.protocolNegotiationEnabled = false
-    cache = buildCache(File("/cache/"), Int.MAX_VALUE.toLong(), fileSystem)
+    val loggingFileSystem = LoggingFilesystem(fileSystem)
```
swankjesse, line 70: NEAT

okhttp/src/test/java/okhttp3/CacheCorruptionTest.kt
```diff
@@ -101,30 +116,41 @@ class CacheCorruptionTest(
...
   private fun corruptMetadata(corruptor: (String) -> String) {
-    val metadataFile = fileSystem.files.keys.find { it.name.endsWith(".0") }
-    val metadataBuffer = fileSystem.files[metadataFile]
+    val metadataFile = fileSystem.allPaths.find {
+      it.name.endsWith(".0")
+    }
 
-    val contents = metadataBuffer!!.peek().readUtf8()
+    if (metadataFile != null) {
+      val contents = fileSystem.source(metadataFile).buffer().use {
```
swankjesse, line 124: now this is just `fileSystem.read(metadataFile) { readUtf8() }`

okhttp/src/test/java/okhttp3/CacheCorruptionTest.kt
```diff
@@ -101,30 +116,41 @@ class CacheCorruptionTest(
...
+        it.readUtf8()
+      }
 
-    metadataBuffer.clear()
-    metadataBuffer.writeUtf8(corruptor(contents))
+      fileSystem.sink(metadataFile).buffer().use {
```
swankjesse, line 128: `fileSystem.write(metadataFile) { writeUtf8(corruptor(contents)) }`

okhttp/src/test/java/okhttp3/CallTest.java
```diff
@@ -104,7 +104,7 @@
 @Timeout(30)
 public final class CallTest {
   @RegisterExtension final PlatformRule platform = new PlatformRule();
-  @RegisterExtension final InMemoryFileSystem fileSystem = new InMemoryFileSystem();
+  final FakeFileSystem fileSystem = new FakeFileSystem();
```
swankjesse, line 107: Call `checkNoOpenFiles()` in tearDown ?

swankjesse (commented):
# RAD  🔥💙👑🌠

okhttp/src/main/kotlin/okhttp3/internal/cache/DiskLruCache.kt
yschimke: Is that a planned helper method in okio?

okhttp/src/main/kotlin/okhttp3/internal/cache/DiskLruCache.kt
yschimke: I think they are converging on .editorconfig.  They allowed using it for a few simple things like import length, but I believe it's presence in newer versions is very impactful.

IMHO The solution is ditching the square styles import and diving into .editorconfig

okhttp/src/main/kotlin/okhttp3/internal/cache/DiskLruCache.kt
yschimke: Not convinced, but trusting your judgement.  Specifically it seems like "shouldn't exist here" is a weak assumption.

okhttp/src/test/java/okhttp3/CacheCorruptionTest.kt
yschimke: I hope we can come up with ways to make it simple to debug problems with the Cache.  And eventually make it safe to use the Cache, as in fail early if you are doing something broken, or with a clear error at runtime.  I like the FileSystem API for this reason.

All towards less bug reports and stackoverflow questions with bizarre errors.

okhttp/src/main/kotlin/okhttp3/Cache.kt
```diff
@@ -365,14 +365,18 @@ class Cache internal constructor(
   }
 
   @get:JvmName("directory") val directory: File
+    get() = cache.directory.toFile()
+
+  @get:JvmName("directoryPath") val directoryPath: Path
```
yschimke, line 370: New public API method.  cc @swankjesse

okhttp-testing-support/src/main/kotlin/okhttp3/okio/LoggingFilesystem.kt
```diff
@@ -0,0 +1,66 @@
...
+
+import okio.ExperimentalFileSystem
+import okio.FileSystem
+import okio.ForwardingFileSystem
+import okio.Path
+import okio.Sink
+import okio.Source
+
+@OptIn(ExperimentalFileSystem::class)
+class LoggingFilesystem(fileSystem: FileSystem) : ForwardingFileSystem(fileSystem) {
```
swankjesse, line 26: You could probably override `onPathParameter()` for all the single-argument methods
```
      override fun onPathParameter(path: Path, functionName: String, parameterName: String): Path {
        println("$functionName($parameterName=$path)")
        return path
      }
```

okhttp/src/main/kotlin/okhttp3/internal/Util.kt
```diff
@@ -542,6 +544,45 @@ fun FileSystem.isCivilized(file: File): Boolean {
...
+  } catch (fnfe: FileNotFoundException) {
+    return
+  }
+}
+
+/**
+ * Tolerant delete, try to clear as many files as possible even after a failure.
+ */
+@OptIn(ExperimentalFileSystem::class)
+fun FileSystem.deleteContents(directory: Path) {
```
swankjesse, line 561: 👍🏻

okhttp/src/main/kotlin/okhttp3/internal/cache/DiskLruCache.kt
```diff
@@ -90,6 +102,18 @@ class DiskLruCache(
   /** Used for asynchronous journal rebuilds. */
   taskRunner: TaskRunner
 ) : Closeable, Flushable {
+  internal val fileSystem: FileSystem = object : ForwardingFileSystem(fileSystem) {
+    override fun sink(file: Path): Sink {
+      file.parent?.let {
+        // TODO from okhttp3.internal.io.FileSystem
```
swankjesse, line 108: This is good because it’s directly equivalent. In follow up I think we should do this once in the init function.

okhttp/src/main/kotlin/okhttp3/internal/cache/DiskLruCache.kt
```diff
@@ -250,12 +274,12 @@ class DiskLruCache(
 
   @Throws(IOException::class)
   private fun readJournal() {
-    fileSystem.source(journalFile).buffer().use { source ->
-      val magic = source.readUtf8LineStrict()
-      val version = source.readUtf8LineStrict()
-      val appVersionString = source.readUtf8LineStrict()
-      val valueCountString = source.readUtf8LineStrict()
-      val blank = source.readUtf8LineStrict()
+    fileSystem.read(journalFile) {
```
swankjesse, line 277: ooooh nice

okhttp/src/main/kotlin/okhttp3/internal/cache/DiskLruCache.kt
```diff
@@ -375,32 +399,34 @@ class DiskLruCache(
   internal fun rebuildJournal() {
     journalWriter?.close()
 
-    fileSystem.sink(journalFileTmp).buffer().use { sink ->
-      sink.writeUtf8(MAGIC).writeByte('\n'.toInt())
-      sink.writeUtf8(VERSION_1).writeByte('\n'.toInt())
-      sink.writeDecimalLong(appVersion.toLong()).writeByte('\n'.toInt())
-      sink.writeDecimalLong(valueCount.toLong()).writeByte('\n'.toInt())
-      sink.writeByte('\n'.toInt())
+    fileSystem.write(journalFileTmp) {
```
swankjesse, line 402: This makes me very happy with the read() and write() functions.

okhttp/src/main/kotlin/okhttp3/internal/cache/DiskLruCache.kt
```diff
@@ -519,14 +545,15 @@ class DiskLruCache(
       if (success && !entry.zombie) {
         if (fileSystem.exists(dirty)) {
           val clean = entry.cleanFiles[i]
-          fileSystem.rename(dirty, clean)
+          fileSystem.atomicMove(dirty, clean)
           val oldLength = entry.lengths[i]
-          val newLength = fileSystem.size(clean)
+          // TODO check null behaviour
```
swankjesse, line 550: This is consistent with what java.io.File did.

swankjesse (approved):

PR #6503, opened 2021-01-07
Gracefully handle null Conscrypt version

### LINKED ISSUE #6502: Crash when Conscrypt VERSION is null
This is a bit of an edge case, but we have a number of people crashing in OkHttp 4.9.0 due to Conscrypt's `version()` call returning `null`. While I unfortunately don't have a repro, looking at Conscrypt's code, it seems this is a valid possibility, as per the code in [Conscrypt.java line 112](https://github.com/google/conscrypt/blob/85f6ce9f27ae1989657d70b428b6c10c69ce6068/common/src/main/java/org/conscrypt/Conscrypt.java#L112). That having been said, I realize this is an edge case, especially since these devices (Samsung/Motorola) are actually returning `true` for conscrypt being enabled, but `null` for the version.

Would it be possible to treat a `null` version as a signal for Conscrypt not being supported? Happy to open a PR if this is reasonable.

```
Caused by java.lang.NullPointerException: Attempt to invoke virtual method 'int org.conscrypt.Conscrypt$Version.major()' on a null object reference
       at okhttp3.internal.platform.ConscryptPlatform$Companion.atLeastVersion(ConscryptPlatform.java:129)
       at okhttp3.internal.platform.ConscryptPlatform.<clinit>(ConscryptPlatform.java:115)
       at okhttp3.internal.platform.android.ConscryptSocketAdapter$Companion$factory$1.matchesSocket(ConscryptSocketAdapter.java:58)
       at okhttp3.internal.platform.android.DeferredSocketAdapter.matchesSocket(DeferredSocketAdapter.java:37)
       at okhttp3.internal.platform.AndroidPlatform.configureTlsExtensions(AndroidPlatform.java:85)
       at okhttp3.internal.connection.RealConnection.connectTls(RealConnection.java:375)
       at okhttp3.internal.connection.RealConnection.establishProtocol(RealConnection.java:337)
       at okhttp3.internal.connection.RealConnection.connect(RealConnection.java:209)
```

ahmedre:
In certain cases, Conscrypt returns a null version. Treat Conscrypt as
not available in these cases, since the version information is unknown.
Fixes #6502.

### DIFF
--- a/okhttp/src/main/kotlin/okhttp3/internal/platform/ConscryptPlatform.kt
+++ b/okhttp/src/main/kotlin/okhttp3/internal/platform/ConscryptPlatform.kt
@@ -124,7 +124,7 @@ class ConscryptPlatform private constructor() : Platform() {
     fun buildIfSupported(): ConscryptPlatform? = if (isSupported) ConscryptPlatform() else null
 
     fun atLeastVersion(major: Int, minor: Int = 0, patch: Int = 0): Boolean {
-      val conscryptVersion = Conscrypt.version()
+      val conscryptVersion = Conscrypt.version() ?: return false
 
       if (conscryptVersion.major() != major) {
         return conscryptVersion.major() > major


ahmedre:
I wanted to write a test for this, but wasn't sure if there was a cleaner way than reflectively setting the version to null or finding a way to supply an empty `conscrypt.properties` to Conscrypt so that it returns null.

yschimke (approved):

Issue #5855, opened 2020-03-06
Feature Request: Offer More Flexible Websocket Pong Failure Policy


commonsguy:
Right now, OkHttp's websocket pong rule is: it must receive a pong for every ping, and that pong must be received before the time comes for the next ping (see #3227 and [this code](https://github.com/square/okhttp/blob/parent-4.4.0/okhttp/src/main/java/okhttp3/internal/ws/RealWebSocket.kt#L486-L508)). I'm sure that works in many cases, but it is a fairly rigid policy. For example, right now I am dealing with an IoT device where I have to ping frequently, but I might not get pongs all the time.

At the moment, I am trying to work around this by:

- Identify a websocket failure due to a missed pong, by the icky approach of examining the exception handed to `onFailure()` and seeing if it is a `SocketTimeoutException` and has `didn't receive pong` in the message

- Open a fresh websocket and hope the IoT device can deal (spoiler alert: it can't deal all that well...)

The current implementation is fine as a default, but it would be useful if we could have an asymmetric policy, with the ping interval at least somewhat independent of the pong timeout period. In my case, it would be something like:

- Send pings every 2 seconds
- Fail only if we do not receive a pong within 40 seconds

(or something like that)

Right now, we are using another websocket library. We are able to implement what we need there because all the guts are exposed, so we can send extra ping frames. We have configured the library's automatic ping logic to have a long period (e.g., 40 seconds) to prevent an early websocket failure due to a missed pong, with the extra ping frames satisfying the IoT device's need for more frequent pings. I am trying to switch over to OkHttp, because we're going to need some of what it offers (particularly in the SSL/TLS area) eventually.

swankjesse:
Hm. My aspirations here are to make the pings + pongs dynamic and based on the other activity on the web socket. 

Perhaps something like “send a ping after 10 seconds of inactivity” instead of “send a ping every 10 seconds”, and backing off from 10 seconds to 30 to 60 to 120, so that a web socket that doesn’t get much use also doesn’t need many pings. In such cases we could probably adjust the pong timeout to be based on a fixed deadline (say 30 seconds in high-latency environments, and 2 seconds in low-latency ones) rather than on the transmission time of the next ping.

commonsguy:
[...]

PR #6504, opened 2021-01-07
adding ping max failures

m4ce:
I am sure this PR is nowhere near ready for approval. I'd like to get some feedback on the approach. We currently have a situation where the websocket backend we are connected is at times very busy and doesn't seem to prioritize pong frames. We keep receiving data but pong frames seem to be quite delayed, or even worse never sent. The idea is to have a maximum number of consecutive ping failures, defaulted to 1 which doesn't change the existing behaviour. We don't really want to increase the ping timeout as that might still lead to a socket failure, which we are trying to avoid. We'd rather try to ping a few times more before bailing out.


okhttp/src/main/kotlin/okhttp3/OkHttpClient.kt
```diff
@@ -220,6 +220,9 @@ open class OkHttpClient internal constructor(
   /** Web socket and HTTP/2 ping interval (in milliseconds). By default pings are not sent. */
   @get:JvmName("pingIntervalMillis") val pingIntervalMillis: Int = builder.pingInterval
 
+  /** Ping max number of consecutive failure. By default is 3 failures. */
```
yschimke, line 223: commetn doesn't match code below (1)

okhttp/src/main/kotlin/okhttp3/OkHttpClient.kt
```diff
@@ -505,6 +509,7 @@ open class OkHttpClient internal constructor(
     internal var readTimeout = 10_000
     internal var writeTimeout = 10_000
     internal var pingInterval = 0
+    internal var pingMaxFailures = 1
```
yschimke, line 512: Should this be set in the constructor

```
internal constructor(okHttpClient: OkHttpClient) : this() {
```

yschimke:
I suspect some headwind with this change.  The discussion in the related issue was tending towards a change that made pings sending more flexible.  There are reason for this like avoiding pings for websockets with obvious traffic, and backing off for dormant websockets.

https://github.com/square/okhttp/issues/5855#issue-576940619

This change adds public API which is the most contentious thing around allowing for faulty servers.  Tolerating missing pongs seems awkward generally because if its because the network connection is faulty, then the websocket likely isn't healthy.

yschimke:
In your concrete case is it the server that is unhealthy and not responding? Or it is the network, including congestion and extending the timeout?

m4ce:
@yschimke, thanks for reviewing the PR. As far as we can see, the server occasionally misses to reply to our ping within the ping interval. I know for a fact that we are receiving messages from the other side, so the connection is actually healthy. I can see messages coming in right up to moment okhttp fails the connection because of the missed pong reply. Don't see any congestion on the host and at the network level. I believe some websocket backends deprioritize ping/pong frames, almost like routers do with ICMP, where the traffic is generally classified as low prio. I will investigate more though. I thought of dropping this here to kick off a discussion and see if such feature could make sense.

yschimke:
I think the deprioritization you are describing is good evidence for the suggestion by @swankjesse to avoid sending while we have active traffic.

m4ce:
@yschimke - yes, as a matter of fact, that would've been the next thing I wanted to discuss. Within my team, we initially assumed that okhttp was only sending the ping if the connection has been idle for a given amount of time. But we soon realized that it pings at a given interval regardless of activity and fails hard after the first failed attempt. Maybe the sweet spot lies with a combination of max failures and last activity?

okhttp/src/main/kotlin/okhttp3/OkHttpClient.kt
m4ce: done, ty

okhttp/src/main/kotlin/okhttp3/OkHttpClient.kt
m4ce: done, ty

m4ce:
@yschimke @swankjesse - I added the logic to check the last activity. I am not a kotlin dev, let me know if things could be done better and in a more efficient way. I am eager to discuss this more if you think it's worth it..

PR #6515, opened 2021-01-18
assertj 3.18.1

sullis:


yschimke:
Is there some background to this? No strong objection to upgrading if it's blocking you.  I often bump the versions, so not against this. But please explain the context for needing this version?

Otherwise it's just a change that could cause some issues for OkHttp maintainers chasing down some surprising difference.

PR #6516, opened 2021-01-18
Intellij export of editorconfig from square/java-code-styles

yschimke:
Exported from preferences menu, removed less common types


JakeWharton:
This is pretty horrific, especially with regard to long-term maintainability. In reality we probably only needs like 6-10 lines.

yschimke:
@JakeWharton 100% agree.  Going to ask some questions to try to imagine what the target we are actually heading for looks like.

Wants

1) Without additional setup (for example importing github.com/square/java-code-styles) Intellij formats correctly
2) Our gradle check and apply tasks (whichever they are) match intellij styles enough that we don't end up with IDE fighting CI
3) Noisy sticky formatting features like single import and indentation/spaces are absolutely

How

1) .editorconfig with minimal entries to make Intellij match our formatting.
2) simple gradle task (spotlessApply*?) to fix before submitting a PR. 

I don't want to try to much here if we don't agree what the solution looks like.  Any pointers to projects in square that have nailed this, let me know.

* I think this broke with some of the sourcepath tricks we used to get conscrypt and android testing working without duplicating tests.

swankjesse:
Agreed with Wants. I don't even know if we care to require spotless before a PR; accidental differences are low-cost but enforcement is high-cost. I'd rather we get the tools to fix whitespace for us and try to avoid humans ever spending time on it.

Diffing the SQ styles vs. IntelliJ defaults might help. Anything that's the same presumably doesn't need to be in the .editorconfig file!

yschimke:
Will rework in a smaller project like certifikit, before transposing here.

yschimke:
Interesting starting point from ktfmt https://github.com/facebookincubator/ktfmt/blob/master/docs/editorconfig-section.txt

Issue #6323, opened 2020-10-16
FormBody parameters should use URL encoding strict mode


fangzhengjin:
The default URL encoding of the FromBody parameter uses a non-strict mode, which will encode spaces as %20, but now some applications no longer recognize %20 as spaces, which will cause decoding errors.
I think the default encoding mode should be changed to strict mode, and spaces should be encoded as +.

fangzhengjin:
I don’t know which development language is used by the service I requested and how to decode the URL, but the data I sent can be decoded correctly only when spaces are encoded as +, but httpclient does encode spaces as +.

And I also saw the description of the encoding of spaces in Wikipedia, the encoding of spaces should use + instead of %20.
[wikipedia](https://en.wikipedia.org/wiki/Percent-encoding#The_application/x-www-form-urlencoded_type)
> The encoding used by default is based on an early version of the general URI percent-encoding rules, with a number of modifications such as newline normalization and **replacing spaces with + instead of %20**.

yschimke:
Is spec now https://url.spec.whatwg.org/#urlencoded-serializing

fangzhengjin:
It seems that the W3C recommendation is also "+": http://www.w3.org/TR/html4/interact/forms.html#h-17.13.4.1
In jdk, URLEncoder will encode most special characters, such as spaces, %, +, #, etc., but OKhttp does not encode these characters.

fangzhengjin:
I would recommend to percent-encode all character except "unreserved" defined in RFC-3986, p.2.3

> unreserved = ALPHA / DIGIT / "-" / "." / "_" / "~"

swankjesse:
Yep, you're absolutely right. How'd we get that wrong?! Can fix.

swankjesse:
The spec we should follow here is W3’s.
https://www.w3.org/TR/2014/REC-html5-20141028/single-page.html#url-encoded-form-data

oday-xiexin:
> Yep, you're absolutely right. How'd we get that wrong?! Can fix.

I have also encountered this problem and hope it can be solved as soon as possible

yschimke:
@swankjesse Is this a good one for an external contributor to submit a PR?

swankjesse:
Let’s just do it. It isn’t very big!

oday-xiexin:
> Let’s just do it. It isn’t very big!

Has this problem not been fixed yet?

oday-xiexin:
[...]

PR #6518, opened 2021-01-19
url encoded form data uses '+' for ' ' according to w3

emrecosar:
url-encoded form data should encode ` `   into `+` for percent-encoding when content-type is `application/x-www-form-urlencoded `

Also, `FORM_ENCODE_SET` updated with `+` sign to align with w3 character set to be encoded.

Reference: [W3 url encoded form data](https://www.w3.org/TR/html4/interact/forms.html#h-17.13.4.1)

Issue: #6323

### DIFF
--- a/okhttp/src/main/kotlin/okhttp3/FormBody.kt
+++ b/okhttp/src/main/kotlin/okhttp3/FormBody.kt
@@ -92,12 +92,12 @@ class FormBody internal constructor(
     fun add(name: String, value: String) = apply {
       names += name.canonicalize(
           encodeSet = FORM_ENCODE_SET,
-          plusIsSpace = true,
+          plusIsSpace = false,
           charset = charset
       )
       values += value.canonicalize(
           encodeSet = FORM_ENCODE_SET,
-          plusIsSpace = true,
+          plusIsSpace = false,
           charset = charset
       )
     }
--- a/okhttp/src/main/kotlin/okhttp3/FormBody.kt
+++ b/okhttp/src/main/kotlin/okhttp3/FormBody.kt
@@ -92,12 +92,12 @@ class FormBody internal constructor(
     fun add(name: String, value: String) = apply {
       names += name.canonicalize(
           encodeSet = FORM_ENCODE_SET,
-          plusIsSpace = false,
+          plusIsSpace = false, // plus is encoded as `%25`, space is encoded as plus
           charset = charset
       )
       values += value.canonicalize(
           encodeSet = FORM_ENCODE_SET,
-          plusIsSpace = false,
+          plusIsSpace = false, // plus is encoded as `%25`, space is encoded as plus
           charset = charset
       )
     }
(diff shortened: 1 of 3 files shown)


emrecosar:
My intention was to remove all `plusIsSpace` parameter and then. encode space character to plus character only on form encoding is in place but that would make a lot of changes. What do you think?

okhttp/src/main/kotlin/okhttp3/FormBody.kt
```diff
@@ -92,12 +92,12 @@ class FormBody internal constructor(
     fun add(name: String, value: String) = apply {
       names += name.canonicalize(
           encodeSet = FORM_ENCODE_SET,
-          plusIsSpace = true,
+          plusIsSpace = false,
```
yschimke, line 95: This probable warrants an inline comment because the parameter name diverts from the intention. 

> @param plusIsSpace true to encode '+' as "%2B" if it is not already encoded.

But plusIsSpace, we just don't want the additional encoding.

yschimke (approved):

yschimke (changes requested):
Comment at least around plusIsSpace = false, when plus is space.

emrecosar:
> Comment at least around plusIsSpace = false, when plus is space.

Thanks, I added a comment.
What do you think about my earlier question on completely remove `plusIsSpace` parameter?
Only the FormBody uses it for this special cases and my idea is just to use encodeSet control instead of another variable. I know it's changing the interface but will help to reduce parameter and complexity.

yschimke:
> 
> 
> > Comment at least around plusIsSpace = false, when plus is space.
> 
> Thanks, I added a comment.
> What do you think about my earlier question on completely remove `plusIsSpace` parameter?
> Only the FormBody uses it for this special cases and my idea is just to use encodeSet control instead of another variable. I know it's changing the interface but will help to reduce parameter and complexity.

Let's land as is with the comment, then make a second separate PR to discuss cleanup.  This can land as is.

yschimke (approved):

swankjesse:
Thanks!

zmwxiaoming:
how can i stay old version and solve this problem

yschimke:
Generally for any new patches or additional functionality you want on OkHttp 3.x, you will need to create a fork and apply each patch you want.

Our supported versions are documented here https://square.github.io/okhttp/security/security/

patriksevallius:
Does this fix exist in a released version?

yschimke:
Yep, should be in 4 and 5 alpha

patriksevallius:
I meant a stable release, I would prefer not to use an alpha version.

yschimke:
Sorry, I was unclear. Use 4.12.0 then.

patriksevallius:
But 4.12.0 doesn't look like it has the fix: https://github.com/square/okhttp/blob/parent-4.12.0/okhttp/src/main/kotlin/okhttp3/FormBody.kt

Both add() and addEncoded() has plusIsSpace = true.

PR #6527, opened 2021-01-24
Use === to compare encode sets

swankjesse:
Also fix a tiny typo in a comment


yschimke (approved):
I considered this but held off as the first line in String equals does this check. Any advice on when to microoptimise like this?

yschimke:
In this case its also safe as even a trivial change to the set will kill this branch.

swankjesse:
Yeah we're cheating here because we want the form use case, not its encode set. If another caller had the same encode set it might not want this behavior.

PR #6529, opened 2021-01-28
Gradle upgrade, more specific toolchain

yschimke:
Update to Gradle 6.8.1.  Use specific toolchain vendor, and move repositories out of submodules to settings.


swankjesse (approved):
Love it

Issue #6509, opened 2021-01-14
Workaround for Conscrypt concurrency issue


prbprbprb:
Android issue [177450597](https://issuetracker.google.com/177450597) looks to be due to a concurrency issue in Conscrypt where two threads race to close an `SSLEngine`-based `SSLSocket` and access to the underlying BIO is unsynchronized.   

We already have a PR google/conscrypt#942 in-flight which addresses this, but realistically we're not going to get that onto the majority of Android 10 devices, so the bug is going to be around for some time, so maybe okhttp could consider quietly swallowing any NPE in `closeQuietly()`?

swankjesse:
Yeah, good idea. Want to submit a PR? Otherwise we’ll do. 

Also I assume we should backport the fix to 3.12.x?

prbprbprb:
I'll have a look, it'll be good for me :)

Even if we mitigate it in `closeQuietly()`, the other thread in the race will still see exceptions some of the time, but I guess that's less of an issue because it's getting cancelled anyway (or returned to an executor, I don't know the architecture :)

yschimke:
This also worried me... https://github.com/square/okhttp/issues/6521

Is OkHttp failing in a bad state when the underlying SSL layer sends runtime errors?

swankjesse:
Yeah good question. I think it's quite surprising that a runtime crash in the connection pool cleanup task could hang a call.

PR #6533, opened 2021-01-30
Silently ignore 'bio == null' NullPointerExceptions

swankjesse:
Closes: https://github.com/square/okhttp/issues/6509


yschimke (approved):

yschimke:
I think I broke this build on master, it's really awkward to get passing builds with very different behaviour on JDK8, JDK8 alpn, JDK 11 etc.  Then JDK 8 now supports ALPN and TLS 1.3.  Does my head in.

PR #6546, opened 2021-02-04
Kotlin 1.4.30 with JVM IR

yschimke:
https://blog.jetbrains.com/kotlin/2021/02/the-jvm-backend-is-in-beta-let-s-make-it-stable-together/

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -14,7 +14,7 @@ buildscript {
       'jsoup': '1.13.1',
       'junit': '4.13',
       'junit5': '5.7.0',
-      'kotlin': '1.4.21',
+      'kotlin': '1.4.30',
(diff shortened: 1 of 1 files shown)


yschimke:
Testing build only, not aware of any driving reason for this.

swankjesse (approved):

yschimke:
It's alpha, are you approving to land? I think it's used with Jetpack Compose projects, so probably safe.  

https://kotlinlang.org/docs/reference/whatsnew14.html#new-jvm-ir-backend
https://kotlinlang.org/docs/reference/evolution/components-stability.html

yschimke:
Will avoid for now, but nothing to report to Intellij either.  Meant to be faster.

PR #6550, opened 2021-02-06
Public Cache API and test

yschimke:
Make Cache API public, and test capturing events.

### DIFF
--- a/okhttp/src/test/java/okhttp3/CacheTest.java
+++ b/okhttp/src/test/java/okhttp3/CacheTest.java
@@ -16,7 +16,6 @@
 
 package okhttp3;
 
-import java.io.File;
 import java.io.IOException;
 import java.net.CookieManager;
 import java.net.HttpURLConnection;
@@ -25,6 +24,7 @@ import java.security.Principal;
 import java.security.cert.Certificate;
 import java.text.DateFormat;
 import java.text.SimpleDateFormat;
+import java.util.ArrayList;
 import java.util.Date;
 import java.util.Iterator;
 import java.util.List;
@@ -44,6 +44,8 @@ import okhttp3.tls.HandshakeCertificates;
 import okio.Buffer;
 import okio.BufferedSink;
 import okio.BufferedSource;
+import okio.FileSystem;
+import okio.ForwardingFileSystem;
 import okio.GzipSink;
 import okio.Okio;
 import okio.Path;
@@ -84,6 +86,7 @@ public final class CacheTest {
     platform.assumeNotBouncyCastle();
 
     server.setProtocolNegotiationEnabled(false);
+    fileSystem.emulateUnix();
     cache = new Cache(Path.get("/cache/"), Integer.MAX_VALUE, fileSystem);
     client = clientTestRule.newClientBuilder()
         .cache(cache)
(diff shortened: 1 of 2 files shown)


yschimke:
This API doesn't address the handling of the unexplained runtime exceptions due to file corruption/race conditions/multi process etc.

But it raises a few things to consider

1) Should we have unit tests for corruption cases.
2) Should we build to provably recover or consider an API to report them? (corrupted count or onCorruption).
3) Should we provide a sample to write your own decorated filesystem that avoids the multi instance or process cases? e.g. a filesystem that writes out the process name and pid and clears up under normal uses and logs if there is a mismatch? 

3 is a half baked idea, but how do we help clients be aware that something is broken?

swankjesse:
I'd love to figure out the definitive cause here.

If it's multiple processes overlapping? I bet there's probably some patterns here we haven't considered to detect it, and then we'd need a strategy to recover if that happens. Perhaps not cache on the 2nd instance? Yuck.

If it's multiple instances in the same process? We should probably detect this through a static variable and throw an exception?

yschimke:
What about combining the processes and instances.  An Android gist with a FileSystem that tries to enforce uniqueness by soft-locking on "appname:instancecounter", and open and close write and delete the market file.  Report by default, but optional fail on startup during dev?

yschimke:
@swankjesse are you for exposing this in 5.0?

swankjesse:
Yes!

yschimke:
OK, holding off for approval.  This change is just that minimal public API.

```
class Cache(
   directory: Path,
   maxSize: Long,
   fileSystem: FileSystem
)
```

And a useful test with example logging that should be easy to build new integrating cases off.

yschimke:
@swankjesse ping

okhttp/src/test/java/okhttp3/CacheTest.java
```diff
@@ -2610,6 +2613,51 @@ private RecordedRequest assertConditionallyCached(MockResponse response) throws
...
+      }
+    };
+    Path path = Path.get("/cache");
+    Cache c = new Cache(path, 100000L, loggingFileSystem);
+
+    assertThat(c.directoryPath()).isEqualTo(path);
+
+    c.size();
+
+    assertThat(events).containsExactly("metadataOrNull:/cache/journal.bkp",
```
swankjesse, line 2642: Love it.

okhttp/src/test/java/okhttp3/CacheTest.java
```diff
@@ -2610,6 +2613,51 @@ private RecordedRequest assertConditionallyCached(MockResponse response) throws
...
+            "metadataOrNull:/cache/journal",
+            "atomicMove:/cache/journal.tmp",
+            "atomicMove:/cache/journal",
+            "appendingSink:/cache/journal");
+
+    events.clear();
+
+    c.size();
+
+    assertThat(events).containsExactly();
```
swankjesse, line 2658: isEmpty() ?

okhttp/src/test/java/okhttp3/CacheTest.java
```diff
@@ -2610,6 +2613,51 @@ private RecordedRequest assertConditionallyCached(MockResponse response) throws
...
+    fileSystem.createDirectories(cache.directoryPath());
+
+    FileSystem loggingFileSystem = new ForwardingFileSystem(fileSystem) {
+      @java.lang.Override
+      public Path onPathParameter(Path path, java.lang.String functionName, java.lang.String parameterName) {
+        events.add(functionName + ":" + path);
+        return path;
+      }
+
+      @java.lang.Override
```
swankjesse, line 2629: java.lang ??

swankjesse (approved):

PR #6551, opened 2021-02-06
Make it more difficult to accidentally log sensitive headers

swankjesse:


okhttp/src/main/kotlin/okhttp3/internal/Util.kt
```diff
@@ -249,6 +249,14 @@ fun String.canParseAsIpAddress(): Boolean {
   return VERIFY_AS_IP_ADDRESS.matches(this)
 }
 
+/** Returns true if we should void putting this this header in an exception or toString(). */
```
yschimke, line 252:
```suggestion
/** Returns true if we should void putting this header in an exception or toString(). */
```

okhttp/src/main/kotlin/okhttp3/internal/Util.kt
```diff
@@ -249,6 +249,14 @@ fun String.canParseAsIpAddress(): Boolean {
...
+fun isSensitiveHeader(name: String): Boolean {
```
yschimke, line 253: Should this be a public API feature?

yschimke (approved):

yschimke:
Failing job looks like it's just circleci being woefully under provisioned recently https://app.circleci.com/pipelines/github/square/okhttp/4508/workflows/b41ede5e-98f7-4d2d-9dd4-79e655f4838f/jobs/20469

okhttp/src/main/kotlin/okhttp3/internal/Util.kt
swankjesse: I don’t think so. It’s the sort of thing you don’t know you need until it’s too late.

PR #6555, opened 2021-02-08
Option to log http headers in one line

dverbovskiy:
Option to log http headers in one line for saving quota of logline/second


yschimke:
This looks like a useful change, but we generally don't release incremental improvements to the logging interceptors.  Instead we encourage people to clone the class and add the functionality you need.  There are lots of little features like this that have been requested, and we don't aim to make everyone happy with a configurable implementation.

Issue #4425, opened 2018-11-30
Reading HTTP Response When Request Isn't Fully Sent


dzelemba:
If I send a large POST request using OkHttp, but the server responds quickly with an error (e.g. permission denied) I get a

java.net.SocketException: Broken pipe (Write failed)( status = 272 )

exception.

Command line tools like curl seem to handle this case just fine and are able to return the HTTP Response instead of aborting because the write failed. Can okhttp do the same?

Note: also asked here https://stackoverflow.com/questions/53451467/reading-http-response-when-request-isnt-fully-sent.

swankjesse:
Tough one. I’m not sure.

bstsnail:
I have the same problem, still don't know how to get the response

yschimke:
Won't fix.

drtrang:
> Won't fix.

Why? It confuses me.

dzelemba:
For anyone struggling with this issue, we did end up getting it to work using Jetty and HTTP 2.0: https://github.com/GoogleCloudPlatform/healthcare-dicom-dicomweb-adapter/blob/master/dicom_util/src/main/java/com/google/cloud/healthcare/DicomWebClientJetty.java#L65

swankjesse:
Related: https://github.com/square/okhttp/pull/6121

jan-berge-ommedal:
@swankjesse I have been testing the latest alpha-release and was happy to see early server responses handled much better than before! The fix works great when the server decides to end the exchange early.

However, i have another use-case where the request body source stream might fail client-side during transmission. In this case, an exception is thrown because of:

https://github.com/square/okhttp/blob/master/okhttp/src/main/kotlin/okhttp3/internal/http/CallServerInterceptor.kt#L86
```
	if (!exchange.hasFailure) {
        throw e // Don't attempt to read the response; we failed to send the request.
	}
```
even if we actually did manage to partially send the request.

It would be nice to still be able to get the server response code (typically 4xx/5xx) + headers/trailers in this case. I started sketching on a way to fix this: https://github.com/square/okhttp/compare/master...jan-berge-ommedal:partial_requests

Could something like this be fitted into okhttp?

yschimke:
@jan-berge-ommedal rather than on this issue, why not put up a proper PR.

jan-berge-ommedal:
PR up @yschimke, sorry for the late response

PR #6585, opened 2021-03-12
read server response if request was partially transferred

jan-berge-ommedal:
Putting up a PR as suggested by @yschimke in https://github.com/square/okhttp/issues/4425#issuecomment-786830186

I am attempting to read the server response if the client did start an exchange but fails to transmit the whole request body. This might happen even if there is no issue with the server connection.

The difficult part was to determine if we did actually start the exchange or not, as okhttp seemingly waits for complete segments before emitting. Please advice if there are better ways to do this than adding RequestBodySink.inProgress().

Also, when not throwing the IOException it seems as if the exchange is not properly cleaned up, probably because catch clauses in RealCall.kt and RetryAndFollowUpInterceptor.kt are not activated. This is detected by OkHttpClientTestRule.ensureAllConnectionsReleased(). Would always closing RequestBodySink inside be the right way to solve this? https://github.com/square/okhttp/pull/6585/files#r593421813

### DIFF
--- a/okhttp/src/main/kotlin/okhttp3/internal/http/CallServerInterceptor.kt
+++ b/okhttp/src/main/kotlin/okhttp3/internal/http/CallServerInterceptor.kt
@@ -37,6 +37,7 @@ class CallServerInterceptor(private val forWebSocket: Boolean) : Interceptor {
     var invokeStartEvent = true
     var responseBuilder: Response.Builder? = null
     var sendRequestException: IOException? = null
+    var requestInProgress = false
     try {
       exchange.writeRequestHeaders(request)
 
@@ -58,9 +59,15 @@ class CallServerInterceptor(private val forWebSocket: Boolean) : Interceptor {
             requestBody.writeTo(bufferedRequestBody)
           } else {
             // Write the request body if the "Expect: 100-continue" expectation was met.
-            val bufferedRequestBody = exchange.createRequestBody(request, false).buffer()
-            requestBody.writeTo(bufferedRequestBody)
-            bufferedRequestBody.close()
+            val requestBodySink = exchange.createRequestBody(request, false)
+            val bufferedRequestBody = requestBodySink.buffer()
+            try {
+              requestBody.writeTo(bufferedRequestBody)
+              bufferedRequestBody.close()
+            } finally {
+              requestInProgress = requestBodySink.inProgress();
+              // TODO move bufferedRequestBody.close() here?
+            }
           }
         } else {
           exchange.noRequestBody()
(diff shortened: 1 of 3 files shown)


okhttp/src/main/kotlin/okhttp3/internal/http/CallServerInterceptor.kt
```diff
@@ -58,9 +59,15 @@ class CallServerInterceptor(private val forWebSocket: Boolean) : Interceptor {
...
-            requestBody.writeTo(bufferedRequestBody)
-            bufferedRequestBody.close()
+            val requestBodySink = exchange.createRequestBody(request, false)
+            val bufferedRequestBody = requestBodySink.buffer()
+            try {
+              requestBody.writeTo(bufferedRequestBody)
+              bufferedRequestBody.close()
+            } finally {
+              requestInProgress = requestBodySink.inProgress();
+              // TODO move bufferedRequestBody.close() here?
```
jan-berge-ommedal, line 69: closing the RequestBodySink here fixes ServerTruncatesRequestTest.readResponseWhenPartialRequest(). And it does make sense to always close the RequestBodySink right?

yschimke:
This looks good to me.  New test only, no other tests and logically makes sense.

okhttp/src/main/kotlin/okhttp3/internal/http/CallServerInterceptor.kt
yschimke: Or put it inside a `.use {}` block?

jan-berge-ommedal: Would be nice, but `close()` completes the request body and currently this affects `inProgress()`. So i think `inProgress()` must be called before `close()` when the transfer fails.

But if i reimplement it as `didReceiveBytes()`, i can use `.use{}` :+1:

jan-berge-ommedal:
After thinking a bit more about this PR, i am having a little cold feet.

In essence, what this PR does is changing the behavior when experiencing an io error unrelated to the exchange with the server. Previously the exchange would be violently cancelled, now we will pretend everything went well in order to give the server a chance to respond. A properly implemented server should always verify that the request was indeed valid, but it might be safer to cancel with violence. The server might react differently in these two scenarios and i am not sure that the latter should be the default. 

We could make this configurable of course, but it is fairly easy to solve this case by swallowing io exceptions in a request body implementation.

yschimke:
@swankjesse Closing this out unless you feel strongly enough that we'd stand behind any observable change.

PR #6586, opened 2021-03-13
Test interaction between pins and non-CA certificates

swankjesse:
Our path building is very naive and it's possible to cause OkHttp to
create chains that are different from the chains previously verified
by the SSL engine.

Our preference is to use the platform's chain where possible. This includes
Android API 17+, which has an API that returns the actual chain that
was verified, not the (potentially larger) set of certificates served
by the server.

We should further lock down our chain processing. I'm reluctant to do
full graph building because it's a lot of ugly code and potentially
expensive. For now, a test that attempts to detect gaps between what
the TLS layer lets through and what our chain cleaner computes.

### DIFF
--- a/okhttp/src/test/java/okhttp3/internal/tls/CertificatePinnerChainValidationTest.java
+++ b/okhttp/src/test/java/okhttp3/internal/tls/CertificatePinnerChainValidationTest.java
@@ -27,6 +27,9 @@ import javax.net.ssl.SSLSocketFactory;
 import javax.net.ssl.TrustManager;
 import javax.net.ssl.X509KeyManager;
 import javax.net.ssl.X509TrustManager;
+import mockwebserver3.MockResponse;
+import mockwebserver3.MockWebServer;
+import mockwebserver3.SocketPolicy;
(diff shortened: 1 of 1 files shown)


yschimke (approved):

yschimke:
@swankjesse I think this shows we have a gap when you opt into conscrypt on JDK.

```
CertificatePinnerChainValidationTest > signersMustHaveCaBitSet() FAILED
    org.opentest4j.AssertionFailedError: expected connection failure but got Response{protocol=h2, code=200, message=, url=https://localhost:44305/}
```

Should we 

a) address in BasicCertificateChainCleaner, even though that works for JSSO
b) implement a new Conscrypt ChainCleaner for JDK.
c) identify any missing hooks from Conscrypt we should be using?

swankjesse:
I think option `c)` is the least bad? Thanks for opening a tracking bug.

yschimke:
@swankjesse the Conscrypt report google/conscrypt#1018 (repeating myself)

jj-crypto (approved):

PR #6590, opened 2021-03-18
Keep taskRunner's cancelAll safe.

JarvisGG:


yschimke (approved):
Looks like an appropriate change.  We generally only call from test runners since it is internal API (by package convention, not by visibility)

Since this is not public API, can you expand on whether it is related to existing behaviour in tests or you need this for some particular reason?

JarvisGG:
> Looks like an appropriate change. We generally only call from test runners since it is internal API (by package convention, not by visibility)
> 
> Since this is not public API, can you expand on whether it is related to existing behaviour in tests or you need this for some particular reason?

Just found out this method was a risk operation.😄

okhttp/src/main/kotlin/okhttp3/internal/concurrent/TaskRunner.kt
```diff
@@ -243,14 +243,16 @@ class TaskRunner(
...
   fun cancelAll() {
-    for (i in busyQueues.size - 1 downTo 0) {
-      busyQueues[i].cancelAllAndDecide()
-    }
-    for (i in readyQueues.size - 1 downTo 0) {
-      val queue = readyQueues[i]
-      queue.cancelAllAndDecide()
-      if (queue.futureTasks.isEmpty()) {
-        readyQueues.removeAt(i)
+    synchronized(this) {
```
swankjesse, line 246: Please replace `synchronized`  with this:
```
    this.assertThreadHoldsLock()
```
Our only production caller holds a lock when it calls this; I don’t want re-entrant synchronized blocks if I can avoid it.

Possibly also requires us to change `OkHttpClientTestRule` to acquire the lock before making this call.

In either case, DO NOT call APIs in `okhttp3.internal`; they’re undocumented, subject to change, and not for end users’ use.

swankjesse (changes requested):

okhttp/src/main/kotlin/okhttp3/internal/concurrent/TaskRunner.kt
JarvisGG: Ok.

swankjesse (approved):

swankjesse:
Thank you!

PR #6593, opened 2021-03-19
Upgrade Kotlin to 1.4.31

adoroszlai:
## What changes were proposed in this pull request?

Upgrade Kotlin to 1.4.31 (latest release), mostly to get a fix for CVE-2020-29582.

## How was this patch tested?

```
$ ./gradlew clean check
...
BUILD SUCCESSFUL in 3m 55s
130 actionable tasks: 129 executed, 1 up-to-date
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -15,7 +15,7 @@ buildscript {
       'jnrUnixsocket': '0.28',
       'jsoup': '1.13.1',
       'junit': '4.13',
-      'kotlin': '1.4.10',
+      'kotlin': '1.4.31',
       'moshi': '1.9.2',
       'okio': '2.8.0',
       'ktlint': '0.38.0',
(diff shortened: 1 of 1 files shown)


yschimke:
@adoroszlai will also require an update in buildSrc/src/main/kotlin/deps.kt:

yschimke (changes requested):
~Needs buildSrc/src/main/kotlin/deps.kt: update~

Just realised this is for 4.9.x.  I'm not certain we should drive updates for CVEs unless they directly affect us.  Do we create temporary files?

yschimke (commented):
Not blocking anymore, but less sure about a maintenance release for this.

adoroszlai:
Thanks @yschimke for the review.

> I'm not certain we should drive updates for CVEs unless they directly affect us. Do we create temporary files?

I don't think it really affects okhttp: cannot find related code in kotlin files.  Please feel free to close this PR.

PR #6596, opened 2021-03-20
Changed expected exception check by JUnit API usage

eas5:
This is a test refactoring.

**Problem:**
The Exception Handling test smell occurs when the test outcome is manually determined through pass or fail method calls, which are dependent on the production method throwing an exception, generally inside a try/catch block.

**Solution:**
We propose using JUnit's exception handling to automatically pass/fail the test instead of writing custom exception handling code or throwing an exception. In this particular case, JUnit 5 assertThrows() was used to properly handle the expected exception.

**Result:**
_Before:_
```
try {
    builder.maxAge(-1, TimeUnit.SECONDS);
    fail();
}
```

_After:_
```
assertThrows(IllegalArgumentException.class, () ->{
  builder.maxAge(-1, TimeUnit.SECONDS);
});
```

### DIFF
--- a/okhttp/src/test/java/okhttp3/CacheControlTest.java
+++ b/okhttp/src/test/java/okhttp3/CacheControlTest.java
@@ -19,7 +19,7 @@ import java.util.concurrent.TimeUnit;
 import org.junit.jupiter.api.Test;
 
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.junit.jupiter.api.Assertions.fail;
+import static org.junit.jupiter.api.Assertions.assertThrows;
 
 public final class CacheControlTest {
   @Test public void emptyBuilderIsEmpty() throws Exception {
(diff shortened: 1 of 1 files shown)


JakeWharton:
I am fundamentally in disagreement that using try/fail/catch is a problem.

If you just want to say this is a refactoring that reduces the boilerplate of that pattern with a helper method then I'm on board. It's literally the same exact pattern wrapped in a method. So if it's a smell, you're just spraying perfume to mask the odor rather than actually taking a shower or whatever this metaphor is breaking down but you get the point.

I usually prefer a custom `assertThrows` that returns a Truth `ThrowableSubject` since it's much easier to assert on the type with instanceOf/exactlyInstanceOf as well as the message, causal chain, etc.

eas5:
You are absolutely right. There is no evident problem at a first glance.

Indeed, the central idea of this proposal is to make the exception catching a responsibility of the test framework, which already provides _the same implementation wrapped in a method_ in its API. The boilerplate reduction would happen by centralizing the exception handling (through whatever pattern would be better suited to the task), potentially simplifying future maintenance activities both in test cases and test framework.

In addition, without the need of using try/catch blocks, test code _might_ be more straightforward and comprehensive.

Thank you very much for your comments.

JakeWharton:
Yeah. I'm supportive for that reason!

yschimke:
What is proposed here? I really like this, but think one file is pointless, but changing this across at least a module makes sense.

I like it because it's possible to forget the fail() and such the cleanup is much safer.

JakeWharton:
Error prone enforces the fail call so it's solely a LOC win.

yschimke:
@eas5 Is there a plan to refactor all instances of this pattern or just this one in particular? If just this one, a bit of context around why this one is a problem would be helpful.

yschimke:
Doing this change as one specific cleanup at a time doesn't seem worthwhile

PR #6325, opened 2020-10-17
EA Testing of JDK 16 - seems to work!

yschimke:
This branch is just a canary in the coal mine.  The actual PR is likely just a version bump to the CI config.

https://wiki.shibboleth.net/confluence/display/DEV/Java+16

Schedule:

    2020-12-10 Rampdown Phase One
    2021-01-14 Rampdown Phase Two
    2021-02-04 Initial Release Candidate
    2021-02-18 Final Release Candidate
    2021-03-16 General Availability

Proposed to Target Java 16

    JEP 380: Unix-Domain Socket Channels
    JEP 395: Records

getting ahead of things https://jdk.java.net/16/release-notes

Build 21

Refine ZipOutputStream.putNextEntry() to recalculate ZipEntry's compressed size (JDK-8253952)
core-libs/java.util.jar

Prior to JDK 16, ZipOutputStream.putnextEntry() would not recalculate the compressed size for a compressed (DEFLATED) entry. This could result in the ZipException, "invalid entry compressed size", being thrown if the current ZLIB implementation used when ZipOutputStream.putNextEntry() is called differs from the implementation at the time when the entry was added to the original ZIP file.

Starting with JDK 16, if the compressed size has not been explicitly set with the ZipEntry.setCompressedSize(long) method when writing a compressed(DEFLATED) entry, then the compressed size will be set to the actual compressed size after deflation.

Build 18
Removal of Legacy Elliptic Curves (JDK-8235710)
security-libs/javax.crypto

[...]

PR #6597, opened 2021-03-21
JDK16 build instead of JDK15

yschimke:


yschimke:
Previous testing in https://github.com/square/okhttp/pull/6325/files

yschimke:
Possibly some fixes required before gradle 7 https://melix.github.io/blog/2021/03/gradle-java16.html

swankjesse (approved):

PR #6598, opened 2021-03-21
Testing JDK16 Unix Sockets

yschimke:


yschimke:
@swankjesse how does the "This is named "tunneling" because it assumes the ultimate destination has a hostname and port." part work?

yschimke:
No plan to ship, just evaluating for JDK16.

PR #6606, opened 2021-03-26
Testing with Okio alpha.2

yschimke:


swankjesse:
I think I like the `emulateWindows()` API.

swankjesse:
Thanks!

Issue #6609, opened 2021-03-30
Support creating a RequestBody from a FileDescriptor


nic0lette:
In Android 11, changes related to Scoped Storage have made it so it isn't always possible to create a `File` reference to content, such as content that's retrieved via the Storage Access Framework. It _is_ possible to get a `FileDescriptor` (or `FileInputStream`) to the data, however.

Since `FileDescriptor` is part of the standard Java API, including it would allow Android developers who need to work within the Scoped Storage changes without adding a dependency on the Android SDK.

JakeWharton:
It's just for `FileDescriptor`, and came in on Slack since it's becoming a fairly ubiquitous type in Android APIs as they eliminate the ability to use `File`. Jesse said yes, and requested it be moved to an issue.

PR #6614, opened 2021-04-02
RequestBody for FileDescriptor

yschimke:
https://github.com/square/okhttp/issues/6609

### DIFF
--- a/okhttp/src/main/kotlin/okhttp3/RequestBody.kt
+++ b/okhttp/src/main/kotlin/okhttp3/RequestBody.kt
@@ -15,19 +15,17 @@
  */
 package okhttp3
 
-import com.android.okhttp.okio.Okio
-import java.io.File
-import java.io.IOException
-import java.nio.charset.Charset
-import kotlin.text.Charsets.UTF_8
 import okhttp3.MediaType.Companion.toMediaTypeOrNull
 import okhttp3.internal.checkOffsetAndCount
 import okio.BufferedSink
 import okio.ByteString
-import okio.buffer
 import okio.source
+import java.io.File
 import java.io.FileDescriptor
 import java.io.FileInputStream
+import java.io.IOException
+import java.nio.charset.Charset
+import kotlin.text.Charsets.UTF_8
 
 abstract class RequestBody {
 
--- a/okhttp/src/main/kotlin/okhttp3/RequestBody.kt
+++ b/okhttp/src/main/kotlin/okhttp3/RequestBody.kt
@@ -15,6 +15,7 @@
  */
 package okhttp3
 
+import com.android.okhttp.okio.Okio
 import java.io.File
 import java.io.IOException
 import java.nio.charset.Charset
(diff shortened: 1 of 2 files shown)


okhttp/src/main/kotlin/okhttp3/RequestBody.kt
```diff
@@ -133,6 +135,23 @@ abstract class RequestBody {
...
     }
 
+    /** Returns a new request body that transmits this. */
+    @JvmStatic
+    @JvmName("create")
+    fun FileDescriptor.toRequestBody(contentType: MediaType? = null): RequestBody {
+      return object : RequestBody() {
+        override fun contentType() = contentType
+
+        override fun isOneShot(): Boolean = true
```
JakeWharton, line 145: Is this true? We can't read it twice?

okhttp/src/main/kotlin/okhttp3/RequestBody.kt
```diff
@@ -15,15 +15,17 @@
...
 import okhttp3.internal.checkOffsetAndCount
 import okio.BufferedSink
 import okio.ByteString
 import okio.source
+import java.io.File
+import java.io.FileDescriptor
+import java.io.FileInputStream
+import java.io.IOException
+import java.nio.charset.Charset
+import kotlin.text.Charsets.UTF_8
```
JakeWharton, line 28: Seems like we need spotless up in here enforcing import order.

JakeWharton (approved):

okhttp/src/main/kotlin/okhttp3/RequestBody.kt
yschimke: In theory you could reset the FileInputStream via a FileChannel, but it would seem quite a stretch.  Consider other frequent FileDescriptors like Pipes and STDIN.  It doesn't make sense in those cases.

At the point you want to reset it you should probably implement your own.

```
    /**
     * A handle to the standard input stream. Usually, this file
     * descriptor is not used directly, but rather via the input stream
     * known as <code>System.in</code>.
     *
     * @see     java.lang.System#in
     */
    public static final FileDescriptor in = new FileDescriptor(0);
```

yschimke: There is a test that confirms it fails, which doesn't rely on isOneShot.

okhttp/src/main/kotlin/okhttp3/RequestBody.kt
yschimke: +1 

I bounce around various projects and computers at times with different conventions so I'm much more in favour of getting ktfmt and .editorconfig with IntelliJ just working such that we don't have these format wars.  If you find a perfect setup on some project, let me know otherwise I'll try to find some time to work through the sort of setup.

I suspect Square folk get to a stable point with java-code-styles but for me this is a constant annoyance.

swankjesse: I’m hoping we standardize on a very short `.editorconfig`.

swankjesse (commented):
Love it.

PR #6615, opened 2021-04-03
Simplify our secondary gradle modules

yschimke:
Being automatic was a lofty goal that makes the experience in IDE unpredictable.  Particularly while IDE and command line are sharing daemons. Switch to a simpler model.

- By default build in IDE and command line/daemon without Android and Graal builds
- Allow -PandroidBuild or -PgraalBuild to include additional modules and overlapping source paths

This isn't a great situation, but OkHttp has so many different builds to consider
- JDK vs Android vs Graal
- JDK 8, 11, 16
- Platform - JSSE, Conscrypt, etc

And ultimately it seems really tough to get something that works automatically within Intellij without finding weirdness with plugins or on branch switches when reimporting gradle.

### DIFF
--- a/android-test/build.gradle
+++ b/android-test/build.gradle
@@ -31,10 +31,6 @@ repositories {
   }
 }
 
-def isIDE = properties.containsKey('android.injected.invoked.from.ide') ||
-        (System.getenv("XPC_SERVICE_NAME") ?: "").contains("intellij") ||
-        System.getenv("IDEA_INITIAL_DIRECTORY") != null
-
 android {
   compileOptions {
     sourceCompatibility JavaVersion.VERSION_1_8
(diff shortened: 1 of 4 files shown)


yschimke:
An example of the issues I've managed to let slip through

```
> Task :native-image-tests:compileKotlin FAILED
e: /home/circleci/project/okhttp/src/test/java/okhttp3/JSSETest.kt: (73, 27): Cannot access 'exchange': it is internal in 'Response'
e: /home/circleci/project/okhttp/src/test/java/okhttp3/JSSETest.kt: (73, 38): Cannot access 'connection': it is internal in 'Exchange'
```

yschimke:

swankjesse (commented):
LGTM

PR #6616, opened 2021-04-03
Show failing test to report to openjdk

yschimke:


yschimke:
Memory cache seems very unpredictable sun.security.util.MemoryCache in SSLSessionContextImpl.  But very relevant that it has a high miss rate.

yschimke:
Raised https://bugs.java.com/bugdatabase/view_bug.do?bug_id=JDK-8264944

PR #6623, opened 2021-04-10
Gradle 7.0 upgrade

yschimke:
https://docs.gradle.org/7.0/release-notes.html

- Also including the change to mostly avoid native image and android builds unless explicitly opted in.  Will try to bring them back in CI first.
- Also fixes a couple of issues triggering with animal sniffer and process resources


yschimke:
@swankjesse requesting post submit review - want to have smaller target PR for testing with build caching.  This is not strictly necessary but makes forward progress possible and isn't affecting deployed code (src/main/java).

swankjesse (commented):
LGTM!

PR #6642, opened 2021-04-18
Replace part of http with https

Goooler:


docs/interceptors.md
```diff
@@ -74,24 +74,24 @@ OkHttpClient client = new OkHttpClient.Builder()
...
+    .url("https://www.publicobject.com/helloworld.txt")
     .header("User-Agent", "OkHttp Example")
     .build();
 
 Response response = client.newCall(request).execute();
 response.body().close();
 ```
 
-When we run this code, the interceptor runs twice. Once for the initial request to `http://www.publicobject.com/helloworld.txt`, and another for the redirect to `https://publicobject.com/helloworld.txt`.
+When we run this code, the interceptor runs twice. Once for the initial request to `https://www.publicobject.com/helloworld.txt`, and another for the redirect to `https://publicobject.com/helloworld.txt`.
```
yschimke, line 85: Suspect this nuance of http redirect to https is worth review.

docs/changelog_2x.md
```diff
@@ -583,7 +583,7 @@ in addition to synchronous blocking calls.
     getFollowSslRedirects()**. We reserve the word _protocol_ for the HTTP
     version being used (HTTP/1.1, HTTP/2). The old name of this method was
     misleading; it was always used to configure redirects between `https://` and
-    `http://` schemes.
+    `https://` schemes.
```
yschimke, line 586: incorrect

LICENSE.txt
```diff
@@ -1,7 +1,7 @@
 
                                  Apache License
                            Version 2.0, January 2004
-                        http://www.apache.org/licenses/
+                        https://www.apache.org/licenses/
```
yschimke, line 4: IANAL - but I really don't think we should be changing the text of the license in any way here.  Maybe break other downstream tooling for all we know.  Flag a license change and review of OkHttp usage in some companies?

android-test/src/androidTest/java/okhttp/android/test/OkHttpTest.kt
```diff
@@ -5,7 +5,7 @@
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *      http://www.apache.org/licenses/LICENSE-2.0
+ *      https://www.apache.org/licenses/LICENSE-2.0
```
yschimke, line 8: Can we leave these out.  The Apache license text is still http, and www.apache.org is probably in browsers HSTS preload files anyway.

docs/interceptors.md
```diff
@@ -41,18 +41,18 @@ OkHttpClient client = new OkHttpClient.Builder()
...
+    .url("https://www.publicobject.com/helloworld.txt")
     .header("User-Agent", "OkHttp Example")
     .build();
 
 Response response = client.newCall(request).execute();
 response.body().close();
 ```
 
-The URL `http://www.publicobject.com/helloworld.txt` redirects to `https://publicobject.com/helloworld.txt`, and OkHttp follows this redirect automatically. Our application interceptor is called **once** and the response returned from `chain.proceed()` has the redirected response:
+The URL `https://www.publicobject.com/helloworld.txt` redirects to `https://publicobject.com/helloworld.txt`, and OkHttp follows this redirect automatically. Our application interceptor is called **once** and the response returned from `chain.proceed()` has the redirected response:
```
yschimke, line 52: Again this changes the meaning of the documentation.  There are two redirects here http->https and www.publicobject.com -> publicobject.com.  Both are intentional.

yschimke:
Thanks for doing the change, I appreciate it.  However I don't think it's a simple search and replace for a project that is a http client.  Hence my review comments.

Goooler:
> Thanks for doing the change, I appreciate it. However I don't think it's a simple search and replace for a project that is a http client. Hence my review comments.

Sorry, I have not noticed that before doing.

android-test/src/androidTest/java/okhttp/android/test/OkHttpTest.kt
Goooler: Reverted now.

docs/interceptors.md
```diff
@@ -41,7 +41,7 @@ OkHttpClient client = new OkHttpClient.Builder()
     .build();
 
 Request request = new Request.Builder()
-    .url("http://www.publicobject.com/helloworld.txt")
+    .url("https://www.publicobject.com/helloworld.txt")
```
yschimke, line 44: Sorry, I missed this in previous review. Given documentation below, I think this line should be reverted.

docs/interceptors.md
```diff
@@ -74,7 +74,7 @@ OkHttpClient client = new OkHttpClient.Builder()
...
```
yschimke, line 77: Sorry, I missed this in previous review. Given documentation below, I think this line should be reverted.

docs/interceptors.md
```diff
@@ -52,7 +52,7 @@ response.body().close();
 The URL `http://www.publicobject.com/helloworld.txt` redirects to `https://publicobject.com/helloworld.txt`, and OkHttp follows this redirect automatically. Our application interceptor is called **once** and the response returned from `chain.proceed()` has the redirected response:
 
 ```
-INFO: Sending request http://www.publicobject.com/helloworld.txt on null
+INFO: Sending request https://www.publicobject.com/helloworld.txt on null
```
yschimke, line 55: Sorry, I missed this in previous review. Given documentation below, I think this line should be reverted.

yschimke (commented):
A big thanks for working through this.

Goooler:
> A big thanks for working through this.

You are so kind, thx.

swankjesse:
Follow-up here: https://github.com/square/okhttp/pull/6654

PR #6644, opened 2021-04-22
Remove openjdk16 jobs from PR

yschimke:


yschimke:
Last few builds have been stable, so dropping this

PR #6647, opened 2021-04-24
ALPN builds

yschimke:
8.0.242 is the last alpn-boot relevant JDK.  Want a setup that could mostly work locally or in CI to replicate compiling against 11, then running tests against 8.0.242.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -122,7 +122,7 @@ jobs:
         uses: actions/setup-java@v2
         with:
           distribution: 'adopt'
-          java-version: 1.8.0_242
+          java-version: 8.0.242
 
       - name: Run Checks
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -121,10 +122,10 @@ jobs:
         uses: actions/setup-java@v2
         with:
           distribution: 'adopt'
-          java-version: 11
+          java-version: 8.0.242
```
yschimke, line 125: This makes the compile fail.  I'll change to try setting up 8.0.242, then 11, then use a custom toolchain location to point at 8.0.242

https://docs.gradle.org/current/userguide/toolchains.html

swankjesse (commented):
LGTM

PR #6667, opened 2021-05-07
Do not return a nullable header value if a default value is set

bramhaag:
This PR splits up the `header(name: String, defaultValue: String? = null): String?` method in `Request` into two methods:
 - `fun header(name: String): String?`
 - `fun header(name: String, defaultValue: String): String`

This way, the method does not return a nullable type when a default value is set. 

I deprecated the original method to prevent introducing breaking API changes.

### DIFF
--- a/okhttp/src/main/kotlin/okhttp3/Response.kt
+++ b/okhttp/src/main/kotlin/okhttp3/Response.kt
@@ -166,6 +166,13 @@ class Response internal constructor(
 
   fun header(name: String, defaultValue: String): String = header(name) ?: defaultValue
 
+  @JvmName("-deprecated_header")
+  @Deprecated(
+    message = "functionality covered by header(String) and header(String, String)",
+    replaceWith = ReplaceWith(expression = "header(name)"),
+    level = DeprecationLevel.ERROR)
+  fun header(name: String, defaultValue: String?): String? = header(name) ?: defaultValue
+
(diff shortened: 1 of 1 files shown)


okhttp/src/main/kotlin/okhttp3/Response.kt
```diff
@@ -162,8 +162,16 @@ class Response internal constructor(
...
   fun headers(name: String): List<String> = headers.values(name)
 
-  @JvmOverloads
-  fun header(name: String, defaultValue: String? = null): String? = headers[name] ?: defaultValue
+  fun header(name: String): String? = headers[name]
+
+  fun header(name: String, defaultValue: String): String = header(name) ?: defaultValue
+
+  @JvmName("-deprecated_header")
+  @Deprecated(
```
yschimke, line 170: I could be misunderstanding the experience for java and kotlin callers with source compatibility.  But this seems like effectively a new method for most people, so weird that it is deprecated.

yschimke:
Overall, I'm not convinced the value is worth the complexity here.

bramhaag:
You are right: these changes do introduce compatibility problems that I did not think about, both on the Java and Kotlin side. 

I made this PR because I was annoyed by the fact that I still had to do null checks even if I set a non-null default value, but introducing this functionality in this way without introducing breaking changes does not seem possible.

Issue #6674, opened 2021-05-11
HttpMethod.permitsRequestBody("OPTIONS") returns true


kanesee:
Whole issue is in the title. Currently, the method implementation is 
```
  public static boolean permitsRequestBody(String method) {
    return !(method.equals("GET") || method.equals("HEAD"));
  }

```

This doesn't account for OPTIONS method which should not have a request body according to spec

luis33338:
> Todo el problema está en el título. Actualmente, la implementación del método es
> 
> ```
>   public static boolean permitsRequestBody(String method) {
>     return !(method.equals("GET") || method.equals("HEAD"));
>   }
> ```
> 
> Esto no tiene en cuenta el método OPTIONS, que no debería tener un cuerpo de solicitud de acuerdo con las especificaciones.

  public static boolean permitsRequestBody(String method) {
    return !(method.equals("GET") || method.equals("HEAD"));
  }

carlonzo:
out of curiosity, what are the specs you refer to?
reading https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html mentions that `OPTIONS` may have a request body.

kanesee:
I might have been referring to these specs
https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/OPTIONS

But w3 is probably the more standard source, so I leave it up to you to decide whether to close this without resolution.
I made my own method for my purposes so this is not a blocker for me anymore.

yschimke:
Think this is now https://datatracker.ietf.org/doc/html/rfc7231#section-4.3.7

> A client that generates an OPTIONS request containing a payload body
   MUST send a valid Content-Type header field describing the
   representation media type.  Although this specification does not
   define any use for such a payload, future extensions to HTTP might
   use the OPTIONS body to make more detailed queries about the target
   resource.

PR #6676, opened 2021-05-13
fix: issue #6674 fix. add OPTION method to bad methods for having body

aboots:
fix issue #6674
add OPTION method to not allowed methods with body


yschimke:
Closing based on https://datatracker.ietf.org/doc/html/rfc7231#section-4.3.7

PR #6657, opened 2021-05-02
Fix OpenJSSE build

yschimke:
OpenJSSE supports JSSE 11 features on JDK 8.

PR #6698, opened 2021-06-04
Build optimisations

robmoore-i:
Hey! 👋

I've made some optimizations to your build with help from Gradle Enterprise. This PR contains some build script changes which enable local builds to benefit from incremental build optimizations.

Let me know if you have any concerns or questions about these changes.

## Optimisation for incremental builds

After these changes, on my machine, when I run `./gradlew clean check` and then `./gradlew check`, the first check takes 1m 9s and the second check takes 1s. When I then make a test code change in the `mockwebserver` subproject and rerun `./gradlew check` it takes 15s, because it re-runs (only!) the `mockwebserver`.

### Addressing implicit dependencies between tasks

The two tasks `:okcurl:copyResourcesTemplates` and `:okhttp:copyJavaTemplates` both copy files, such that the `sourcesJar` tasks can use them. This dependency between tasks is not made explicit anywhere, which means that Gradle doesn't have the information it needs to ensure that incremental builds are safe. This makes these tasks and their dependents (such as the test tasks) unable to benefit from incremental build optimization, which has a significant impact on build performance.

### Runtime classpath normalization

The normalization included in the root `build.gradle` tells all projects that they don't need to rebuild & retest themselves if the only thing that has changed on their runtime classpath is the `Bnd-LastModified` timestamp in the `META-INF/METADATA.MF` file of their dependencies' JARs.

Regarding the classpath modifications made by the OsgiTest (the changes in `okhttp/build.gradle`), I have included a verbose comment in the build script to describe what is going on there.

## Removal of deprecation warning

The below code is no longer necessary :)

```
kotlinDslPluginOptions {
    experimentalWarning.set(false)
}
```

## (Removed from PR) Applying Gradle properties to leverage available build-accelerating features

Adding 'org.gradle.configureondemand=true' tells Gradle to use [configuration on demand](https://docs.gradle.org/current/userguide/multi_project_configuration_and_execution.html#sec:configuration_on_demand). **Update**: Removed this because it's not recommended to use configuration on demand for Kotlin DSL builds.

Adding 'org.gradle.parallel=true' tells Gradle to [run builds in parallel](https://docs.gradle.org/current/userguide/performance.html#parallel_execution). **Update**: Removed this because doesn't work well for publishing artefacts to Sonatype.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,4 +1,3 @@
 org.gradle.jvmargs='-Dfile.encoding=UTF-8'
 android.enableJetifier=true
 android.useAndroidX=true
-org.gradle.parallel=true
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,6 +1,5 @@
 org.gradle.jvmargs='-Dfile.encoding=UTF-8'
 android.enableJetifier=true
 android.useAndroidX=true
-org.gradle.caching=true
 org.gradle.configureondemand=true
 org.gradle.parallel=true
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,3 +1,6 @@
 org.gradle.jvmargs='-Dfile.encoding=UTF-8'
 android.enableJetifier=true
 android.useAndroidX=true
+org.gradle.caching=false
+org.gradle.configureondemand=true
+org.gradle.parallel=true
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,5 +1,4 @@
 org.gradle.jvmargs='-Dfile.encoding=UTF-8'
 android.enableJetifier=true
 android.useAndroidX=true
-org.gradle.configureondemand=true
 org.gradle.parallel=true
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,6 +1,6 @@
 org.gradle.jvmargs='-Dfile.encoding=UTF-8'
 android.enableJetifier=true
 android.useAndroidX=true
-org.gradle.caching=false
+org.gradle.caching=true
 org.gradle.configureondemand=true
 org.gradle.parallel=true
--- a/.gitignore
+++ b/.gitignore
@@ -2,8 +2,6 @@
 .project
 .settings
 .gradle
-build-cache
-tmp
 eclipsebin
 
 bin
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,8 @@
 .project
 .settings
 .gradle
+build-cache
+tmp
 eclipsebin
 
 bin
--- a/build.gradle
+++ b/build.gradle
@@ -39,6 +39,14 @@ allprojects {
       configurations.findAll { it.canBeResolved }.files
     }
   }
+
+  normalization {
+    runtimeClasspath {
+      metaInf {
+        ignoreAttribute("Bnd-LastModified")
+      }
+    }
+  }
 }
 
 ext.applyOsgi = { project ->
--- a/buildSrc/build.gradle.kts
+++ b/buildSrc/build.gradle.kts
@@ -18,10 +18,6 @@ plugins {
     `kotlin-dsl`
 }
 
-kotlinDslPluginOptions {
-    experimentalWarning.set(false)
-}
-
 repositories {
     mavenCentral()
 }
--- a/okcurl/build.gradle
+++ b/okcurl/build.gradle
@@ -20,6 +20,9 @@ sourceSets {
 processResources {
   dependsOn 'copyResourcesTemplates'
 }
+sourcesJar {
+    dependsOn 'copyResourcesTemplates'
+}
 task copyResourcesTemplates(type: Copy) {
   from 'src/main/resources-templates'
   into "$buildDir/generated/resources-templates"
--- a/okhttp/build.gradle
+++ b/okhttp/build.gradle
@@ -29,12 +29,25 @@ sourceSets {
 
 normalization {
     runtimeClasspath {
-        // - These files are generated during test execution by the test:
-        // okhttp/src/test/java/okhttp3/osgi/OsgiTest.java
-        // - The underlying index.xml file contains a timestamp property,
-        // which results in the :okhttp:test task not being up-to-date.
-        // - This normalization prevents the test task from having to
-        // re-execute despite nothing of relevance changing.
+        /*
+           - The below two ignored files are generated during test execution
+           by the test: okhttp/src/test/java/okhttp3/osgi/OsgiTest.java
+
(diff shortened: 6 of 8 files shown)


gradle.properties
```diff
@@ -1,3 +1,5 @@
 org.gradle.jvmargs='-Dfile.encoding=UTF-8'
 android.enableJetifier=true
 android.useAndroidX=true
+org.gradle.configureondemand=true
```
TangHuaiZhe, line 4: Looks that `configureondemand` is not recommend by Gradle when using Kotlin based Gradle build scripts:
> Enabling the incubating `configuration on demand` feature is not recommended as it can lead to very hard-to-diagnose problems.

https://docs.gradle.org/current/userguide/migrating_from_groovy_to_kotlin_dsl.html#before_you_start_migrating

gradle.properties
```diff
@@ -1,3 +1,5 @@
...
+org.gradle.parallel=true
```
JakeWharton, line 5: The concurrent task execution which this enables causes problems for artifact publishing to Sonatype which sees the concurrent uploads as separate projects. If someone wants to use this unstable feature, they can enable it locally rather than us forcing it on them.

robmoore-i: Thanks, I wasn't aware of that. I'll remove this configuration.

gradle.properties
robmoore-i: Thanks for finding that - I'll remove this configuration. The configuration time is quite low so this isn't much of a saving anyway (just a few seconds).

gradle.properties
ldaley: Separately, if you want to be able to publish in parallel you can use https://github.com/gradle-nexus/publish-plugin/.

yschimke:
Thanks for focusing on this. This should be really impactful.

JakeWharton (approved):

gradle.properties
JakeWharton: Thanks for the link. One complication is we need to retain the ability to deploy to our internal Artifactory with a simple env var change ideally.

robmoore-i:
It looks like there are some failing tests outstanding from the most recently merged PR: https://github.com/square/okhttp/pull/6657/checks

yschimke:
> It looks like there are some failing tests outstanding from the most recently merged PR: https://github.com/square/okhttp/pull/6657/checks

These are long standing, slowly getting fixed amongst.  CircleCI was effectively broken, timing out for a long time.  Moving to github actions has been gradual and has flagged some tests that were way too particular (cipher suites etc) and subject to evolving JDK situation, a lot of things have been backported to JDK 8 and 11 recently breaking assumptions.

I'll land as is and fix forward.

yschimke:
Link to normalization docs https://docs.gradle.org/current/userguide/build_cache_concepts.html#normalization

gradle.properties
ldaley: That indeed complicates things. To get the best of both, you'd need to one way or another have different logic. I guess this might be further complicated by the requirement to not have square-private publishing be part of the “open” source repository.

If I were tasked with implementing this, I'd probably use the nexus-publish plugin to publish to OSSRH as it's the only reliable way I have found to avoid ending up with multiple staging repositories, and use the Artifactory plugin to upload to Artifactory. Its primary benefit is that it uses finer grained parallelism for uploading than the stock plugins. 

If you don't want to commit such Square specifics to the repo, the Artifactory functionality would need to be wrapped in a plugin that is only injected in the Square environment one way or another. I imagine such a thing could be used by any other open Square projects that build with Gradle. 

This wouldn't really facilitate arbitrarily changing the destination repo _address_ with an env var, but you could use that to rewire what a `publish` lifecycle task does (i.e. publish to public or private) based on such an external input. Alternatively, you could just have `publishPublic` and `publishPrivate` lifecycle tasks etc. Or, you could have have dual Sonatype/Artifactory capabilities, and allow use of arbitrary destination repos along those lines, controlled via env var.

Or of course, stick with the current lowest-common-denominator approach at the expense of build time.

ldaley:
Just a comment on this:

> If someone wants to use this unstable feature, they can enable it locally rather than us forcing it on them.

(from https://github.com/square/okhttp/pull/6698#discussion_r645502847)

The suitability of using `--parallel` is really a build specific thing and not really environmentally sensitive. The majority of builds work reliably with `--parallel`, with significant benefit for larger builds. It is “incubating” and not enabled-by-default as it is not guaranteed safe for _all_ builds; in particular builds that use custom logic with shared mutable memory structures across projects. It's very unlikely that this project has that dynamic, and use of `--parallel` would be safe (should the publishing issues be resolved). 

FTR, this uncertainty around `--parallel` and potential for hazard is actively being worked on by Gradle, particularly as part of the “[configuration caching](https://docs.gradle.org/current/userguide/configuration_cache.html)” initiative.

I'm not necessarily advocating for a change in position on use of `--parallel` for this project. As this is a well regarded and influential project, I wanted to provide some extra context for anyone coming across this comment.

JakeWharton:
FWIW I do use `--parallel` just about everywhere via a Bash alias which also locates the wrapper entrypoint, but I personally prefer to keep "raw" `./gradlew` to be as vanilla as possible so that I can always drop down to get fairly stable behavior.

ldaley:
> FWIW I do use --parallel just about everywhere via a Bash alias which also locates the wrapper entrypoint

Good to know, thanks. 

You can also achieve the same thing by putting `org.gradle.parallel=true` in `~/.gradle/gradle.properties` and then use `--no-parallel` to disable it ad-hoc for a particular invocation.

yschimke:
@ldaley @robmoore-i is it worth putting some of this in the readme for common project setup. Any developer or CI improvements we can do with Gradle would be great. I spend a lot of hopping between PRs, and I'd like to optimise these flows where we can.  Even to the point of a Gradle build cache populated by CI jobs, mind if I tag either of you on CLs in this space.

As a project we have a really awkward set of targets to build for, so would benefit a lot from build optimisation.  I did some for Circle CI, then lost it all with the github move.

ldaley:
@yschimke for the time being, please feel free to tag me. 

I'll also initiate some discussion internally @ Gradle Inc. on some other options and will get back to you on that.

yschimke:
@robmoore-i This seems like a hugely impactful change.  Thanks!!!!!!

yschimke:
```
$  ./gradlew test

BUILD SUCCESSFUL in 1s
101 actionable tasks: 101 up-to-date
```

robmoore-i:
Thanks for coming back here to share @yschimke! I'm glad to see this 😊

swankjesse (commented):
LGTM

Issue #6704, opened 2021-06-12
okHttp and graalvm not working with latest stable (4.9.1) - only RC?


Excpt0r:
Hey,
I was playing around with okHttp and graalvm, so I'm not experienced with both.
I had problems with some classes/unresolved types i.e.:
OpenJSSEPlatform, ConscryptPlatform, BouncyCastlePlatform

As the okHttp website states it is ready for graalvm, I thought the fault is with my configuration and studied stackoverflow and okcurl.
But at the end, using okhttp version 4.10.0-RC1 instead of the latest stable 4.9.0 just did the trick.
The relevant change seems to be okhttp/src/main/kotlin/okhttp3/internal/graal/svm.kt

Hope this helps some people, and maybe it would be worth mentioning this here:
https://square.github.io/okhttp/#graalvm-native-image

yschimke:
@Excpt0r yep, good point. We are adding support in 5.0.0-alpha.2 now. I'll put a note.

yschimke:
@Excpt0r OkHttp 5 will follow on from Okio 3.0, so it's not imminent.

PR #6705, opened 2021-06-13
Document that GraalVM support is on 5.0.0-alpha.2 and later.

yschimke:
From https://github.com/square/okhttp/issues/6704


yschimke:
@swankjesse requesting post review, doc only.

swankjesse (commented):
LGTM

PR #6714, opened 2021-06-13
Conditionally run extra builds on PRs

yschimke:


swankjesse:
LGTM

This is so cool I had no idea!

yschimke:
Yep - it's signal to noise management, need to be able to run extra jobs when relevant, we want to test most scenarios in CI, but PRs shouldn't suffer flaky builds.

Issue #6146, opened 2020-06-25
Expose Websocket Ping/Pong Round Trip Time via the WebsocketListener


bentemple:
The ping / pong round trip time is important and useful information that can pretty trivially be exposed via the WebsocketListener but currently isn't. Can we expose this information in a method such as `onPong(roundTripTime: Long)`?

There have been some discussions around the ping/pong mechanism:
https://github.com/square/okhttp/issues/2239
https://github.com/square/okhttp/issues/3197

OkHttp does now support configuring the ping timeout, but it would also be valuable and useful information to know how long the round-trip-time of each ping request takes. Websockets are often used in real time applications, and knowing your latency to the server can have a range of uses across an app to provide better synchronization with the server.

yschimke:
Think this is a dupe of https://github.com/square/okhttp/issues/5833, we wouldn't do one without the other I think.

Please mention your requirements there.

PR #6730, opened 2021-06-28
Customization of ping-pong handling for latency monitoring

smallmonk:
The change allows customisation of ping frame and receiving handling of pong frame for #6146

### DIFF
--- a/okhttp/src/main/kotlin/okhttp3/OkHttpClient.kt
+++ b/okhttp/src/main/kotlin/okhttp3/OkHttpClient.kt
@@ -267,6 +267,11 @@ open class OkHttpClient internal constructor(
   /** Prepares the [request] to be executed at some point in the future. */
   override fun newCall(request: Request): Call = RealCall(this, request, forWebSocket = false)
 
+  /** Uses [request] to connect a new web socket. */
+  override fun newWebSocket(request: Request, listener: WebSocketListener): WebSocket {
+    return newWebSocket(request, listener, null)
+  }
+
   fun newWebSocket(request: Request, listener: WebSocketListener, pingPayloadFactory: WebSocketPingPayloadFactory?): WebSocket {
     val webSocket = RealWebSocket(
       taskRunner = TaskRunner.INSTANCE,
@@ -280,11 +285,6 @@ open class OkHttpClient internal constructor(
     )
     webSocket.connect(this)
     return webSocket
-  }
-
-  /** Uses [request] to connect a new web socket. */
-  override fun newWebSocket(request: Request, listener: WebSocketListener): WebSocket {
-    return newWebSocket(request, listener, null)
   }
 
   open fun newBuilder(): Builder = Builder(this)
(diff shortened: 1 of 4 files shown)


okhttp/src/main/kotlin/okhttp3/OkHttpClient.kt
```diff
@@ -269,14 +269,19 @@ open class OkHttpClient internal constructor(
 
   /** Uses [request] to connect a new web socket. */
   override fun newWebSocket(request: Request, listener: WebSocketListener): WebSocket {
+    return newWebSocket(request, listener, null)
+  }
+
+  fun newWebSocket(request: Request, listener: WebSocketListener, pingPayloadFactory: WebSocketPingPayloadFactory?): WebSocket {
```
yschimke, line 275: The change overall seems relatively safe and clean. I think the main concern overall is the new API here and in WebSocketListener.  I wonder whether a subinterface specific to ping-pong handling could be a good way to control this instead.  If the listener implements that we will use it. It would minimise surface area and not expose this as top level API and we could mark that interface as experimental?

@swankjesse @JakeWharton

yschimke:
Overall, I would not land in this form and previously @swankjesse expressed enough concern to push back on specific submissions in this area or customising when we send ping pongs.  So expect some headwind to this PR. see https://github.com/square/okhttp/issues/5833 . We are really careful about new public API, especially when it's obvious afterwards that we solved in the wrong approach or have to deprecate an early attempt at a solution.

But it's useful to prompt the discussion again.

smallmonk:
Thanks for sharing your consideration. I understand the complexity that once the API is released, it is hard to take it back later even if it has issue.  For okhttp, the impact is great as okhttp is one of the default library added to Android Studio.
Whoever has the need could cherry-pick this pull request for their own customisation.

swankjesse:
What’s the motivation? My preference is for pings to stay implementation details in OkHttp, just as they are in browser-based web socket APIs.
https://developer.mozilla.org/en-US/docs/Web/API/WebSocket

smallmonk:
Just for sharing on my case, the round trip latency of web socket connection is requested to be displayed on the screen so that app user is aware of the connection quality.  There are 3 ways I could think of:
(1) Send the timestamp in a web socket frame to server and update server implementation to provide the same frame without any change (or latency is computed from server and sent to client).  This requires explicit server implementation for latency computation.

(2) The computation of round trip latency could be implemented within the library. Let assume that the common way to compute latency is to (i) send the timestamp in ping frame to server and (ii) compute the time difference upon receiving the pong frame.  The latency computation implementation within the library might cause issue to the server implementation as some web socket server implementation might have issue to respond to non-empty ping frame.

(3) Provide API to allow the developer to customise the behaviour of ping-pong for latency computation.  Server change is not required in this case.  The side-effect for the proposed implementation is that inexperienced might not understand the rationale of the customisation and could misuse the API which causes server issue like case (2). Currently, I am not sure what is the good solution to avoid the misuse.

In browser implementation, since case (3) is not possible, I will have to choose case (1) instead. 
I would say that this change is not a must for Android developer but it empower the developer to have one more option for latency implementation.

yschimke:
> Whoever has the need could cherry-pick this pull request for their own customisation.

+1, I'll close this for now.  You have a workaround and we wouldn't want to rush in new public API for this as is.  We are very careful with API changes because of the balance in supporting existing users including major SDK etc that can't change easily.

Issue #6620, opened 2021-04-06
Want "mockWebServer.dispatcher.clear()"


lathspell:
I have a JUnit file with several test functions that share a MockWebServer instance. The reason that they share it is that I need to declare it statically so that I can use Spring's `@DynamicPropertySource` to inject the MockWebServer's port number as Spring environment variable.
Anyway, Sometimes I enqueue a response and then test function A fails and does not take the response off the queue as expected. Then the next test function B is executed, I enqueue another response but the next call to the mock retrieves the first response in the queue - that was intended for the test function A.

What I propose is a `mockWebServer.dispatcher.clear()` method that empties the response queue. I could call this in a `@BeforeEach` annotated JUnit method.

JakeWharton:
Seems reasonable for QueueDispatcher but not Dispatcher.

lathspell:
In `MockWebServer.enqueue(response: MockResponse)` the "dispatcher" is cast to QueueDispatcher to calll its "enqueue()" method. I would propose the same for "clear()".

ansgarkonermann:
I also ran into this. Clearing the mock server response queue would truly be a valuable feature enhancement for me.

yschimke:
To be clear on the status of this, enhancements for MockWebServer isn't a big focus for the OkHttp maintainers, but as this seems like a reasonable request, it's probably something that would be accepted given a nice simple clean PR.  @lathspell Any interest in working on this?

lathspell:
see https://github.com/square/okhttp/pull/6736

mtorba:
Any news for this issue? A solution has been **merged** since July 2021, and yet it's still not part of any public release (as far as I can tell)

yschimke:
Released in 5 alphas.

garashis:
Can someone paste the how to use solution in 5 alphas

datadidit:
Will this be in a non alpha release 4.11.0 is newer than your latest alpha but doesn't have this change ported to it.

DavidTheProgrammer:
I'm literally in this same predicament. I'm considering creating the mockwebserver with a static port and counting on that for my tests, but that seems error prone.

PR #6736, opened 2021-07-01
Added Dispatched.clear() method for testing

lathspell:
Regarding https://github.com/square/okhttp/issues/6620

The main patch is `mockwebserver/src/main/kotlin/mockwebserver3/QueueDispatcher.kt` with the test in `mockwebserver/src/test/java/mockwebserver3/MockWebServerTest.java`.

Everything else was just fixed to satisfy the interface class. If you want a smaller patch or do not want a change in the interface I could also make `clear()` with a default implementation that does nothing. The only drawback would be that new implementations of the Dispatcher interface would then miss to override the clear method. But on the other hand, who would ever want to add another implementation anyway. Your choice.

### DIFF
--- a/mockwebserver/src/main/kotlin/mockwebserver3/QueueDispatcher.kt
+++ b/mockwebserver/src/main/kotlin/mockwebserver3/QueueDispatcher.kt
@@ -59,6 +59,10 @@ open class QueueDispatcher : Dispatcher() {
     responseQueue.add(response)
   }
 
+  open fun clear() {
+    responseQueue.clear()
+  }
+
   override fun shutdown() {
     responseQueue.add(DEAD_LETTER)
   }
--- a/mockwebserver/src/test/java/mockwebserver3/MockWebServerTest.java
+++ b/mockwebserver/src/test/java/mockwebserver3/MockWebServerTest.java
@@ -257,6 +257,15 @@ public final class MockWebServerTest {
     server.url("/b").url().openConnection().getInputStream(); // Should succeed.
   }
 
+  @Test public void clearDispatcherQueue() throws Exception {
+    server.enqueue(new MockResponse().setBody("A"));
+    ((QueueDispatcher) server.getDispatcher()).clear();
+    server.enqueue(new MockResponse().setBody("B"));
+
+    InputStream in = server.url("/a").url().openConnection().getInputStream();
+    assertThat(in.read()).isEqualTo('B');
+  }
+
   /**
    * Throttle the request body by sleeping 500ms after every 3 bytes. With a 6-byte request, this
    * should yield one sleep for a total delay of 500ms.


mockwebserver-deprecated/src/main/kotlin/okhttp3/mockwebserver/Dispatcher.kt
```diff
@@ -19,6 +19,8 @@ abstract class Dispatcher {
   @Throws(InterruptedException::class)
   abstract fun dispatch(request: RecordedRequest): MockResponse
 
+  abstract fun clear()
```
yschimke, line 22: What about an open method with an empty body here?

yschimke:
```
MockWebServerTest > testClear() FAILED
    java.util.concurrent.TimeoutException: testClear() timed out after 30 seconds
```

mockwebserver-deprecated/src/main/kotlin/okhttp3/mockwebserver/Dispatcher.kt
swankjesse: I think this likely only makes sense on QueueDispatcher . . . could we just require callers to downcast to that type? I’d rather the callers downcast than have a method that doesn’t make sense on another dispatcher type.

lathspell: Fine for me, as long as it is documented so that users know about it.

lathspell: I reduced the Merge Request to just an `open fun clear()` in `QueueDispatcher`.

yschimke (approved):
LGTM

mockwebserver/src/main/kotlin/mockwebserver3/QueueDispatcher.kt
```diff
@@ -59,6 +59,10 @@ open class QueueDispatcher : Dispatcher() {
     responseQueue.add(response)
   }
 
+  open fun clear() {
```
GeorgiGeorgiev09, line 62: Can you please implement this feature to maven lib please !

seanstallbaumchamp: do we have an eta on when this will be publicly available?

perryh-luna:
Ping @yschimke do you know why this hasn't been included in a release yet? It's been merged for a while. Thank you!

swankjesse:
Released.
https://github.com/square/okhttp/blob/master/CHANGELOG.md#version-500-alpha11

syl-oh:
Also, you need to import it via `import mockwebserver3.MockWebServer;` and not `import okhttp3.mockwebserver.MockWebServer;`

PR #6750, opened 2021-07-11
Android 30 CloseGuard

yschimke:
Avoids using grey list methods.


yschimke:
I'm curious about our historic usage.  We never use the explicit close. We use the reference leaking to trigger our call to warnIfOpen which is always true because we don't call close.

yschimke:
Also our usage looks different with https://developer.android.com/reference/android/util/CloseGuard referring to reachabilityFence()

okhttp/src/main/kotlin/okhttp3/internal/platform/android/Android10CloseGuard.kt
```diff
@@ -0,0 +1,31 @@
...
+package okhttp3.internal.platform.android
+
+import okhttp3.internal.platform.CloseGuard
+
+class Android10CloseGuard : CloseGuard {
+  override fun createAndOpen(closer: String): Any? = android.util.CloseGuard().apply { open(closer) }
+
+  override fun warnIfOpen(closeGuardInstance: Any?): Boolean {
+    val instance = closeGuardInstance as? android.util.CloseGuard
+      ?: throw IllegalArgumentException("Invalid closeguard $closeGuardInstance")
```
JakeWharton, line 25: Don't think this provides meaningful improvement on the exception that would be thrown by removing the `?` on the `as`.

JakeWharton (approved):

okhttp/src/main/kotlin/okhttp3/internal/platform/Android10Platform.kt
```diff
@@ -56,6 +57,14 @@ class Android10Platform : Platform() {
       // No TLS extensions if the socket class is custom.
       socketAdapters.find { it.matchesSocket(sslSocket) }?.getSelectedProtocol(sslSocket)
 
+  @SuppressLint("NewApi")
+  override fun getStackTraceForCloseable(closer: String): CloseGuard = CloseGuard().apply { open(closer) }
```
swankjesse, line 61: This file is named `Android10` (API 29) but it references a class from API 30

yschimke: Doh. Will fix.

Is there a better approach now? i.e. remove this and rely on LeakCanary or similar? As a developer only tool, maybe we could drop support for the legacy mode and only support on Android 30+?

swankjesse: So here's a weird observation: what do we get by using CloseGuard over the behavior in the base class?

I think when we introduced this it was attempting to take advantage of CloseGuard when a connection is leaked. But there's nothing difficult or proprietary about CloseGuard; it just hooks into the GC in a way we can do ourselves.

So I wonder... what happens if we drop all the overrides of this stuff and just rely on the base class behavior? I think it is particularly nice with the Logger.FINE option.

yschimke: 1) Hook into StrictMode system

```
D/StrictMode: StrictMode policy violation: android.os.strictmode.LeakedClosableViolation: A resource was acquired at attached stack trace but never released. See java.io.Closeable for information on avoiding resource leaks.
        at android.os.StrictMode$AndroidCloseGuardReporter.report(StrictMode.java:1910)
        at dalvik.system.CloseGuard.warnIfOpen(CloseGuard.java:286)
```

2) Maybe cheaper than a stacktrace?

But... we don't use it properly https://developer.android.com/reference/android/util/CloseGuard.  We never call close, because we only call warnIfOpen when we've decided using the reference checks.

```
guard.close();
          ...;
          if (Build.VERSION.SDK_INT >= 28) {
              Reference.reachabilityFence(this);
          }
```

okhttp/src/main/kotlin/okhttp3/internal/platform/Android10Platform.kt
```diff
@@ -56,6 +57,23 @@ class Android10Platform : Platform() {
       // No TLS extensions if the socket class is custom.
       socketAdapters.find { it.matchesSocket(sslSocket) }?.getSelectedProtocol(sslSocket)
 
+  override fun getStackTraceForCloseable(closer: String): Any? {
+    return if (Build.VERSION.SDK_INT >= 30) {
+      CloseGuard().apply { open(closer) }
+    } else {
+      super.getStackTraceForCloseable(closer)
```
yschimke, line 64: Falling back to old behaviour which is gated on logging levels ( > FINE )

yschimke:
@swankjesse Going to land as is, we discussed the form and it's a nice cleanup not relying on reflection that is now actively deprecated.  Will follow up with any further cleanups, for example we could use the full features of CloseGuard instead of our own GC detection.

okhttp/src/main/kotlin/okhttp3/internal/platform/Android10Platform.kt
```diff
@@ -56,6 +57,23 @@ class Android10Platform : Platform() {
...
+    }
+  }
+
+  override fun logCloseableLeak(message: String, stackTrace: Any?) {
+    if (Build.VERSION.SDK_INT >= 30) {
+      (stackTrace as CloseGuard).warnIfOpen()
+    } else {
+      // Unable to report via CloseGuard. As a last-ditch effort, send it to the logger.
```
swankjesse, line 72: Great comment, yeah that’s why this exists!

okhttp/src/main/kotlin/okhttp3/internal/platform/android/CloseGuard.kt
```diff
@@ -1,75 +0,0 @@
...
-package okhttp3.internal.platform.android
-
-import java.lang.reflect.Method
-
-/**
- * Provides access to the internal dalvik.system.CloseGuard class. Android uses this in
- * combination with android.os.StrictMode to report on leaked java.io.Closeable's. Available since
- * Android API 11.
- */
-internal class CloseGuard(
```
swankjesse, line 25: Yay red!

swankjesse (commented):
LGTM

jj-crypto (approved):

PR #6761, opened 2021-07-25
Clean Android Regression Test

yschimke:
Provides a clean test which is easily adaptable to reproduce failures on specific Android SDKs, devices or OkHttp versions (other than master).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -312,15 +312,13 @@ jobs:
 
   testandroidregression:
     runs-on: macos-latest
-    timeout-minutes: 30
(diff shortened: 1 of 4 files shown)


yschimke:
```
> Task :regression-test:connectedDebugAndroidTest
Starting 6 tests on test(AVD) - 10
```

yschimke:
@swankjesse requesting post review, test only and will only run when a label is applied to a PR.

swankjesse (commented):
I like it!

swankjesse:
LGTM

PR #6766, opened 2021-07-27
Cookie.newBuilder added.

yschimke:


okhttp/src/test/java/okhttp3/CookieTest.java
```diff
@@ -423,6 +423,25 @@
...
+            .expiresAt(MAX_DATE)
+            .build();
+    assertThat(cookie.name()).isEqualTo("a");
+    assertThat(cookie.value()).isEqualTo("b");
+    assertThat(cookie.expiresAt()).isEqualTo(MAX_DATE);
+    assertThat(cookie.domain()).isEqualTo("example.com");
+    assertThat(cookie.path()).isEqualTo("/");
+    assertThat(cookie.secure()).isFalse();
+    assertThat(cookie.httpOnly()).isFalse();
+    // can't be unset
```
yschimke, line 440: Worth discussing this, strictly speaking, this isn't worse than currently chaining these calls on a fresh Builder, but is not ideal either.

okhttp/src/main/kotlin/okhttp3/Cookie.kt
```diff
@@ -262,6 +264,20 @@ class Cookie private constructor(
     private var persistent = false
     private var hostOnly = false
 
+    constructor()
+
+    internal constructor(cookie: Cookie) {
```
swankjesse, line 269: Can this be private?

okhttp/src/test/java/okhttp3/CookieTest.java
swankjesse: 👍🏻

swankjesse: That change is independent!

swankjesse (approved):

okhttp/src/main/kotlin/okhttp3/Cookie.kt
yschimke: No, Cookie can't access Cookie.Builder(Cookie) if it is.

PR #6780, opened 2021-07-28
Fix ho™ail.com example.

jairus1988:
If an attacker is leveraging the fact that ™ might render as tm, they
would use ho™ail.com as the domain to masquerade hotmail.com.


swankjesse (commented):
LGTM

jj-crypto (approved):

PR #6809, opened 2021-08-01
Allow ClientAuthTest changes on JDK8 after security updates.

yschimke:
With recent JDK 8 changes, some of the JDK 9 behaviour is also seen on JDK 8 also.

This removes some checks last edited here https://github.com/square/okhttp/pull/5263/files for conscrypt.


swankjesse (commented):
LGTM

PR #6810, opened 2021-08-01
Avoid task logger flakes due to shared logger.

yschimke:
Seeing this a lot recently, bleedover from another test with delayed logging.

```
TaskRunnerTest > cancelReturnsTruePreventsNextExecution() FAILED
    java.lang.AssertionError: 
    Expecting:
      <["FINE: Q10000 scheduled after 100 µs: task",
        "FINE: Q10001 finished run in 498 ms: MockWebServer /127.0.0.1:50232",
        "FINE: Q10000 canceled              : task"]>
    to contain exactly (and in same order):
      <["FINE: Q10000 scheduled after 100 µs: task",
        "FINE: Q10000 canceled              : task"]>
    but some elements were not expected:
      <["FINE: Q10001 finished run in 498 ms: MockWebServer /127.0.0.1:50232"]>
        at okhttp3.internal.concurrent.TaskRunnerTest.cancelReturnsTruePreventsNextExecution(TaskRunnerTest.kt:184)
```

### DIFF
--- a/okhttp/src/test/java/okhttp3/internal/concurrent/TaskFaker.kt
+++ b/okhttp/src/test/java/okhttp3/internal/concurrent/TaskFaker.kt
@@ -16,6 +16,7 @@
 package okhttp3.internal.concurrent
 
 import java.util.concurrent.Executors
+import java.util.logging.Logger
 import okhttp3.internal.assertionsEnabled
 import org.assertj.core.api.Assertions.assertThat
 
@@ -24,7 +25,7 @@ import org.assertj.core.api.Assertions.assertThat
  * deterministic. All tasks are executed on-demand on the test thread by calls to [runTasks] and
  * [advanceUntil].
  */
-class TaskFaker {
+class TaskFaker() {
   @Suppress("NOTHING_TO_INLINE")
   internal inline fun Any.assertThreadHoldsLock() {
     if (assertionsEnabled && !Thread.holdsLock(this)) {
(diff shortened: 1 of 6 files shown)


yschimke:
@swankjesse Not happy with this fix, but the flakyness or asserting on shared logging infra is awkward otherwise. In reality this change is essentially a new field in TaskRunner

```
internal val logger: Logger = TaskRunner.logger
```

okhttp/src/test/java/okhttp3/internal/concurrent/TaskFaker.kt
```diff
@@ -45,6 +46,8 @@ class TaskFaker {
   @Suppress("PLATFORM_CLASS_MAPPED_TO_KOTLIN", "NOTHING_TO_INLINE")
   internal inline fun Any.notify() = (this as Object).notify()
 
+  internal val logger = Logger.getLogger("TaskFaker." + instance++)
```
swankjesse, line 49: neat

okhttp/src/test/java/okhttp3/internal/concurrent/TaskFaker.kt
```diff
@@ -196,4 +199,8 @@ class TaskFaker {
 
   /** Returns true if no tasks have been scheduled. This runs the coordinator for confirmation. */
   fun isIdle() = taskRunner.activeQueues().isEmpty()
+
+  companion object {
```
swankjesse, line 203: nit: private?

swankjesse (commented):
LGTM

PR #6819, opened 2021-08-03
Rewrite Http2 connectionNotReusedAfterShutdown test to be less flaky

yschimke:


yschimke:
Failures on CI are connectionWindowUpdateOnClose and concurrentRequestWithEmptyFlowControlWindow

amirlivneh:
What's the root cause for the failures?

yschimke:
It's long term flaky, see https://github.com/square/okhttp/issues/6033. SocketPolicy describes these tests relying on disconnect as flaky and suggests introducing delays.

In recent CI jobs it's been a timeout, I believe because it hangs when it's burned the second response with a failed request that gets retried.  So I added a third response to make that obvious and added the delays until I'm sure the disconnect has worked (hopefully 0, or 100ms, or 2100ms).

swankjesse (commented):
LGTM

jj-crypto (approved):

PR #6841, opened 2021-08-28
Fixed leading spaces for code blocks in recipes.md

BrianOdisho:
Markdown coverts text with four leading spaces into a code block. Fenced code blocks in this file had four leading spaces in addition to the being wrapped with ```, because of this the code blocks weren't being rendered properly in the markdown preview.

Removing the two out of the four leading spaces fixed the issue.

### DIFF
--- a/docs/recipes.md
+++ b/docs/recipes.md
@@ -10,108 +10,108 @@ Download a file, print its headers, and print its response body as a string.
 The `string()` method on response body is convenient and efficient for small documents. But if the response body is large (greater than 1 MiB), avoid `string()` because it will load the entire document into memory. In that case, prefer to process the body as a stream.
 
 === "Kotlin"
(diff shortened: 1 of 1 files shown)


yschimke (approved):

BrianOdisho:
@yschimke should I be concerned about the `testopenjdk11` check failing (it looks unrelated)?

Goooler:
> @yschimke should I be concerned about the `testopenjdk11` check failing (it looks unrelated)?

Don't worry.

JakeWharton:
Much of the code inside the fences is still indented by 2 spaces. Can we clean that up while we're here?

BrianOdisho:
Good catch @JakeWharton, done.

JakeWharton:
Thanks!

Issue #6853, opened 2021-09-10
Couldn't do get request after turning off internet


kate-fin:
Hello.

I do POST request for getting token, next do GET request to get data. I turn off my internet connection, after that turn on it and do GET request again.

Link to the project: https://github.com/kate-fin/OkHttpDoesntWorkApplication.git
You just need click the button, turn off internet, turn it on and click button again 

Log of error:
E/AndroidRuntime: FATAL EXCEPTION: OkHttp Dispatcher
    Process: com.example.okhttpdoesntworkapplication, PID: 30701
    java.lang.IllegalStateException: state: 0
        at okhttp3.internal.http1.Http1ExchangeCodec.readResponseHeaders(Http1ExchangeCodec.kt:173)
        at okhttp3.internal.connection.Exchange.readResponseHeaders(Exchange.kt:110)
        at okhttp3.internal.http.CallServerInterceptor.intercept(CallServerInterceptor.kt:93)
        at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
        at okhttp3.internal.connection.ConnectInterceptor.intercept(ConnectInterceptor.kt:34)
        at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
        at okhttp3.internal.cache.CacheInterceptor.intercept(CacheInterceptor.kt:95)
        at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
        at okhttp3.internal.http.BridgeInterceptor.intercept(BridgeInterceptor.kt:83)
        at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
        at okhttp3.internal.http.RetryAndFollowUpInterceptor.intercept(RetryAndFollowUpInterceptor.kt:76)
        at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
        at okhttp3.logging.HttpLoggingInterceptor.intercept(HttpLoggingInterceptor.kt:222)
        at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
        at okhttp3.internal.connection.RealCall.getResponseWithInterceptorChain$okhttp(RealCall.kt:201)
        at okhttp3.internal.connection.RealCall$AsyncCall.run(RealCall.kt:517)
[...]

PR #6856, opened 2021-09-12
Test for socket closed before request headers sent

yschimke:
Fix by allowing for reading response, before successfully completing sending request.

To address https://github.com/square/okhttp/issues/6853

My personal theory is that Android suppresses close events when apps go background or airplane mode, so they don't wake up in response in IOException and start a flood of requests.


okhttp/src/test/java/okhttp3/ServerTruncatesRequestTest.kt
```diff
@@ -151,22 +161,72 @@ class ServerTruncatesRequestTest(
...
 
-  @Test fun serverTruncatesRequestButTrailersCanStillBeReadHttp2() {
+  @Test
+  fun serverTruncatesRequestButTrailersCanStillBeReadHttp2() {
     enableProtocol(Protocol.HTTP_2)
     serverTruncatesRequestButTrailersCanStillBeRead(http2 = true)
   }
 
   private fun serverTruncatesRequestButTrailersCanStillBeRead(http2: Boolean) {
     val mockResponse = MockResponse()
```
yschimke, line 224: I can revert these, but not certain it's wrong.

okhttp/src/test/java/okhttp3/ServerTruncatesRequestTest.kt
```diff
@@ -151,22 +161,72 @@ class ServerTruncatesRequestTest(
...
+      var socket: SSLSocket? = null
+      var closed = false
+
+      override fun connectionAcquired(call: Call, connection: Connection) {
+          socket = connection.socket() as SSLSocket
+      }
+
+      override fun requestHeadersStart(call: Call) {
+        if (closed) {
+          throw IOException("fake socket failure")
```
yschimke, line 181: This isn't great but gets the failure to occur at the point we see it on Android.

```
  @Throws(IOException::class)
  fun writeRequestHeaders(request: Request) {
    try {
      eventListener.requestHeadersStart(call)
      codec.writeRequestHeaders(request)
      eventListener.requestHeadersEnd(call, request)
    } catch (e: IOException) {
      eventListener.requestFailed(call, e)
      trackFailure(e)
      throw e
    }
  }
```

yschimke: An alternative would be to create a DelegatingSocket or DelegatingSSLSocket and set a flag in requestHeadersStart, then cause the next write to fail.

okhttp/src/main/kotlin/okhttp3/internal/http/CallServerInterceptor.kt
```diff
@@ -37,8 +37,11 @@ class CallServerInterceptor(private val forWebSocket: Boolean) : Interceptor {
     var invokeStartEvent = true
     var responseBuilder: Response.Builder? = null
     var sendRequestException: IOException? = null
+
+    // Send request headers, or if failed then avoid any attempt to read an early response.
+    exchange.writeRequestHeaders(request)
```
yschimke, line 42: I'm not happy with the possibly redundant code in the catch block below. cc @swankjesse

jj-crypto (approved):

okhttp/src/test/java/okhttp3/ServerTruncatesRequestTest.kt
swankjesse: This is so nice and simple though.

swankjesse (commented):
I think the right fix is changing the state machine... I believe we’re probably still broken in a bunch of cases here, such as when writing the body fails.

I’d like to land the test case only on this, and try a different fix.

okhttp/src/test/java/okhttp3/ServerTruncatesRequestTest.kt
swankjesse: Thoughts on closing the socket here instead of throwing the exception? That way the exception comes from the right place in the code.

yschimke:
> I think the right fix is changing the state machine... I believe we’re probably still broken in a bunch of cases here, such as when writing the body fails.
> 
> I’d like to land the test case only on this, and try a different fix.

Perfect, happy for you to take it over :)

yschimke:
@swankjesse re-read your comments, will check in just the test disabled, and we can do the correct fix in a follow up.

swankjesse (approved):
Love it

PR #6861, opened 2021-09-14
DelegatingSSLSocket without reflection

yschimke:
Uses JDK9 methods without reflection since we now build with JDK 11.


yschimke:
Replaced by https://github.com/square/okhttp/pull/6862

Issue #6201, opened 2020-07-28
java.lang.UnsupportedOperationException when using OkHttp 3.12.12 on Android 10 (API 29)


marosseleng:
Hi all,

I am developing an Android app, where I use my small library to handle network operations out of main apps, so it could be reused. The setup is as follows:

**Network library:**

- OhHttp 3.12.12
- Retrofit 2.6.4
- conscrypt 2.2.1
- minSdkVersion 15 (that's why we have to stay with 3.12)
- targetSdkVersion 29

**The app:**
- minSdkVersion 15
- targetSdkVersion 29

When I execute a request on Android 10, it fails with the following stacktrace:

    2020-07-28 15:06:09.056 14386-14430/com.example.testapp W/System.err: java.lang.UnsupportedOperationException
    2020-07-28 15:06:09.057 14386-14430/com.example.testapp W/System.err:     at javax.net.ssl.SSLSocket.getApplicationProtocol(SSLSocket.java:1467)
    2020-07-28 15:06:09.057 14386-14430/com.example.testapp W/System.err:     at okhttp3.internal.platform.Android10Platform.getSelectedProtocol(Android10Platform.java:63)
    2020-07-28 15:06:09.057 14386-14430/com.example.testapp W/System.err:     at okhttp3.internal.connection.RealConnection.connectTls(RealConnection.java:347)
    2020-07-28 15:06:09.057 14386-14430/com.example.testapp W/System.err:     at okhttp3.internal.connection.RealConnection.establishProtocol(RealConnection.java:284)
    2020-07-28 15:06:09.057 14386-14430/com.example.testapp W/System.err:     at okhttp3.internal.connection.RealConnection.connect(RealConnection.java:169)
    2020-07-28 15:06:09.057 14386-14430/com.example.testapp W/System.err:     at okhttp3.internal.connection.StreamAllocation.findConnection(StreamAllocation.java:258)
    2020-07-28 15:06:09.057 14386-14430/com.example.testapp W/System.err:     at okhttp3.internal.connection.StreamAllocation.findHealthyConnection(StreamAllocation.java:135)
    2020-07-28 15:06:09.057 14386-14430/com.example.testapp W/System.err:     at okhttp3.internal.connection.StreamAllocation.newStream(StreamAllocation.java:114)
    2020-07-28 15:06:09.057 14386-14430/com.example.testapp W/System.err:     at okhttp3.internal.connection.ConnectInterceptor.intercept(ConnectInterceptor.java:42)
[...]

PR #6862, opened 2021-09-14
Allow Custom SSLSocketFactory without ALPN support

yschimke:
Be tolerant of UnsupportedOperationException from sslSocket.getApplicationProtocol().

This change will need to be remade on 3.12.12 after confirming on main. To address https://github.com/square/okhttp/issues/6201


okhttp-testing-support/src/main/kotlin/okhttp3/DelegatingSSLSession.kt
```diff
@@ -0,0 +1,114 @@
...
+
+import java.security.Principal
+import java.security.cert.Certificate
+import javax.net.ssl.SSLPeerUnverifiedException
+import javax.net.ssl.SSLSession
+import javax.net.ssl.SSLSessionContext
+import javax.security.cert.X509Certificate
+
+/** An [SSLSession] that delegates all calls.  */
+abstract class DelegatingSSLSession(protected val delegate: SSLSession?) : SSLSession {
```
swankjesse, line 26: Why nullable? It’d be simpler to throw in the `get()` function if you wanna return null sometimes.

okhttp/src/main/kotlin/okhttp3/internal/platform/android/Android10SocketAdapter.kt
```diff
@@ -39,11 +39,19 @@ class Android10SocketAdapter : SocketAdapter {
   override fun isSupported(): Boolean = Companion.isSupported()
 
   @SuppressLint("NewApi")
-  override fun getSelectedProtocol(sslSocket: SSLSocket): String? =
+  override fun getSelectedProtocol(sslSocket: SSLSocket): String?  {
+    return try {
+      // SSLSocket.getApplicationProtocol returns "" if application protocols values will not
+      // be used. Observed if you didn't specify SSLParameters.setApplicationProtocols
```
swankjesse, line 45: nice comment

swankjesse (approved):
Love it

okhttp-testing-support/src/main/kotlin/okhttp3/DelegatingSSLSession.kt
swankjesse: (this is consistent with the original Java. Merging as-is; you can follow-up if you like!)

PR #6881, opened 2021-10-13
Fix logging interceptor name in bom

yschimke:


yschimke:
It's only a problem on 3.12.x

https://github.com/square/okhttp/blob/okhttp_3.12.x/bom/pom.xml
https://github.com/square/okhttp/blob/master/okhttp-bom/build.gradle.kts

swankjesse:
LGTM

jj-crypto (approved):

PR #6912, opened 2021-11-14
[WIP] IDNA upgrade on Android

yschimke:


okhttp/src/main/kotlin/okhttp3/internal/platform/Android10Platform.kt
```diff
@@ -16,20 +16,17 @@
...
 import javax.net.ssl.X509TrustManager
 import okhttp3.Protocol
 import okhttp3.internal.SuppressSignatureCheck
-import okhttp3.internal.platform.android.AndroidSocketAdapter
-import okhttp3.internal.platform.android.Android10SocketAdapter
-import okhttp3.internal.platform.android.AndroidCertificateChainCleaner
-import okhttp3.internal.platform.android.BouncyCastleSocketAdapter
-import okhttp3.internal.platform.android.ConscryptSocketAdapter
-import okhttp3.internal.platform.android.DeferredSocketAdapter
+import okhttp3.internal.platform.android.*
```
yschimke, line 29: will revert

okhttp/src/main/kotlin/okhttp3/internal/platform/Android10Platform.kt
```diff
@@ -81,6 +87,26 @@ class Android10Platform : Platform() {
   override fun buildCertificateChainCleaner(trustManager: X509TrustManager): CertificateChainCleaner =
       AndroidCertificateChainCleaner.buildIfSupported(trustManager) ?: super.buildCertificateChainCleaner(trustManager)
 
+  override fun IDNtoASCII(domain: String): String {
+    val output = StringBuilder()
+    val info = IDNA.Info()
+    idna.nameToASCII(domain, output, info)
+    if (info.hasErrors()) {
```
yschimke, line 94: Should we throw on info.hasErrors? Or conditionally based on the errors?

yschimke:
Failing on

```
okhttp.android.test.OkHttpTest > testUnderscoreRequest[test(AVD) - 10] FAILED 
	java.lang.IllegalArgumentException: Invalid URL host: "example_underscore_123.s3.amazonaws.com"
	at okhttp3.HttpUrl$Builder.parse$okhttp(HttpUrl.kt:1338)
```

okhttp/src/main/kotlin/okhttp3/internal/platform/Android10Platform.kt
j-bernard: You should throw when there is an error. As far as I know ICU4j correctly handles valid labels.

j-bernard:
Underscore are not legal in hostnames, hence the test failure.
But it seems to be a tricky question, see this issue in a Python IDNA package https://github.com/kjd/idna/issues/17.

j-bernard:
Here is a domain that should be correctly converted with the update while it was not before, if you want to include it in your tests:
| U-Label  | A-Label |
| ------------- | ------------- |
| fußball.test  | xn--fuball-cta.test  |

yschimke:
> Here is a domain that should be correctly converted with the update while it was not before, if you want to include it in your tests:
> U-Label 	A-Label
> fußball.test 	xn--fuball-cta.test

fußball.de (exists!) isn't the best example because lowercase works anyway. fußball -> fussball

yschimke:
> fußball.de (exists!) isn't the best example because lowercase works anyway. fußball -> fussball

Maybe that is why it's a good test? https://mail.python.org/pipermail/python-dev/2017-December/151503.html

yschimke:
Firefox redirects to www.xn--fuball-cta.de and fails because the cert is for fussball.de

```
$ cft --host www.xn--fuball-cta.de
Error:  Hostname www.xn--fuball-cta.de not verified:
    certificate: sha256/IcNO01b1bwMEB0NO2fJwlgRHqaXf5cq7MgG+sCSN8/g=
    DN: CN=*.fussball.de, O=DFB GmbH, L=Frankfurt am Main, ST=Hessen, C=DE
    subjectAltNames: [*.fussball.de, fussball.de]
```

yschimke:
There seems a very real risk that we shouldn't land this because it isn't clear that it's implemented correctly in the wild and it it causes as many new problems as it solves...

j-bernard:
If your check .de registry, they provide an [IDN converter](https://www.denic.de/en/service/tools/idn-web-converter/) that does the right conversion. There [FAQ](https://www.denic.de/en/faqs/faqs-about-idns-ss/) also contains useful information.
So, fußball.de and fussball.de are different domains, but in that case it happens that the same registrar got both and performed a redirection from fußball.de to fussball.de.
The fact that Chrome is the most used web browser and does not correctly perform IDNA conversion does not help making the things right, and thus we have that kind of website that needs to set up workarounds to be reached by most users.

yschimke:
Thanks for the link.  I'll add a test for überall-ist.de from that link.

okhttp/src/main/kotlin/okhttp3/internal/platform/Android10Platform.kt
```diff
@@ -43,6 +40,15 @@ class Android10Platform : Platform() {
       DeferredSocketAdapter(BouncyCastleSocketAdapter.factory)
   ).filter { it.isSupported() }
 
+  var idna = IDNA.getUTS46Instance(
+    IDNA.NONTRANSITIONAL_TO_ASCII
+      or IDNA.NONTRANSITIONAL_TO_UNICODE
```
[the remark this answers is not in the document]
yschimke: It seems a common pattern, and the tests should confirm it works.  Any examples in this "transitional" category?

jj-crypto (approved):

yschimke:
@jj-crypto curious about the approval? Are you affected by this issue? Any example domains that are affecting you?

okhttp/src/main/kotlin/okhttp3/internal/platform/Android10Platform.kt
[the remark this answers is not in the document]
j-bernard: the ß character is one impacted by the transitional flag.

yschimke:
@j-bernard Not sure it's safe to land, this actively breaks on "_", while Android tolerates it.  It's arguably going to cause more failures than new successes.

```
     Caused by: java.lang.IllegalArgumentException: nameToASCII error [DISALLOWED]
        at xxx.ICUIDNConverter.toASCII(BBCRadioWearActivity.kt:34)
```

j-bernard:
Makes sense. The best would be that Android fixes it.

PR #6779, opened 2021-07-28
Work around Graal Issue

yschimke:
See https://github.com/oracle/graal/issues/3466
and build https://github.com/square/okhttp/pull/6777/checks?check_run_id=3181507303

https://github.com/square/okhttp/issues/6715

PR #6917, opened 2021-11-20
Reference the bug that's preventing an inline function

swankjesse:
Follow-up to https://github.com/square/okhttp/pull/6779/files


yschimke:
@swankjesse Thanks!

jj-crypto (approved):

PR #6918, opened 2021-11-20
Simplify Animal Sniffer config

yschimke:


okhttp-tls/src/main/kotlin/okhttp3/tls/internal/TlsUtil.kt
```diff
@@ -30,6 +30,7 @@ import okhttp3.tls.HandshakeCertificates
 import okhttp3.tls.HeldCertificate
 import org.codehaus.mojo.animal_sniffer.IgnoreJRERequirement
 
+@IgnoreJRERequirement
```
yschimke, line 33: This duplicates newTrustManager but testing to see a clean build to start with.

swankjesse (approved):

PR #6774, opened 2021-07-28
Avoid flaky tests due to DNS.

yschimke:
See https://github.com/square/okhttp/issues/4836

We don't actually use the requestUrl in our tests, so since it deals with DNS, I'm trying to avoid calling it eagerly.

PR #6919, opened 2021-11-20
Change how we get a hostname for RecordedRequest.requestUrl

swankjesse:
Previously we did a reverse DNS lookup. But the correct strategy
is to look at the inbound ':authority' (HTTP/2) or 'Host' (HTTP/1)
header.

See: https://github.com/square/okhttp/pull/6774

### DIFF
--- a/mockwebserver/src/main/kotlin/mockwebserver3/RecordedRequest.kt
+++ b/mockwebserver/src/main/kotlin/mockwebserver3/RecordedRequest.kt
@@ -60,7 +60,6 @@ class RecordedRequest @JvmOverloads constructor(
    */
   val failure: IOException? = null
 ) {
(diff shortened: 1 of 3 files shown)


yschimke:
Wonder if it's a flaky test?

```
ClientAuthTest > missingClientAuthFailsForNeeds() FAILED
    java.lang.AssertionError: 
    Expecting actual not to be empty
        at okhttp3.internal.tls.ClientAuthTest.missingClientAuthFailsForNeeds(ClientAuthTest.java:218)
```

yschimke (approved):
Nice simplification!  Much clearer the intent.

swankjesse:
I don’t believe that test failure is related; the client auth stuff doesn’t use RecordedRequest on the server side.

PR #6920, opened 2021-11-21
Better exception logging when exhausting routes

yschimke:


swankjesse (approved):
Thanks!

swankjesse:
Hmmmm... It is empty. I don't like the ‘exhausted all routes’ error because it has less information than the failure message of any particular route. I wonder where we’re throwing that away?

yschimke:
Agreed, but it looks like it has never had suppressed errors since it first got added.  Probably worth adding.

yschimke:
Making this PR about adding those suppression in.

yschimke:
@swankjesse This one is hurting in CI.  It seems like we get "exhausted all routes" when we get a connection, but health checks fails. Probably the SSL connection fails after we think we've connected.  

But there isn't an easy way to get an exception from this.

I suggest removing that check for now.  We already chain suppressed exceptions in the RouteException, but this isn't the same error flow.  So it can never pass.

```
      assertThat(expected.getMessage()).isEqualTo("exhausted all routes");
      assertThat(expected.getSuppressed()).isNotEmpty();
    }
``` 

```

      // Confirm that the connection is good.
      if (candidate.isHealthy(doExtensiveHealthChecks)) {
        return candidate
      }

      // If it isn't, take it out of the pool.
      candidate.noNewExchanges()

      // Make sure we have some routes left to try. One example where we may exhaust all the routes
      // would happen if we made a new connection and it immediately is detected as unhealthy.
      if (nextRouteToTry != null) continue

      val routesLeft = routeSelection?.hasNext() ?: true
      if (routesLeft) continue

      val routesSelectionLeft = routeSelector?.hasNext() ?: true
      if (routesSelectionLeft) continue

      throw IOException("exhausted all routes")
```

yschimke:
If we want to keep this, then we can record each IOException in RealConnection at the point we call listener.connectionFailed(), and also record that we have an unhealthy connection (the server closed the connection on us).

swankjesse:
Works for me.

swankjesse:
(Removing the check works for me. Improving the exception is a separate task that we can do later.)

maldito-dot:
> https://github.com/top-gg/docs/tree/master/content%2Fapi%2Fbot.mdx

@maldito-dot

maldito-dot (commented):
@monstro805

maldito-dot (commented):
@maldito-dot

yschimke:
No longer relevant

PR #6924, opened 2021-11-23
Make EventSource.cancel() fail the call immediately

swankjesse:
EventSource.cancel() did nothing if it was created by EventSources.processResponse.
All EventSources would continue to process buffered events rather than
breaking out sooner.


yschimke (approved):
LGTM

yschimke:
Seems like a flaky test

https://github.com/square/okhttp/runs/4487781366?check_suite_focus=true
https://github.com/square/okhttp/runs/4386097408?check_suite_focus=true

PR #6937, opened 2021-11-27
Remove suppression check that can never pass.

yschimke:


yschimke:
@swankjesse Going to land as it makes the CI builds quite flaky, when it hits this exception catch clause it always fails, and it varies by provider and I think is quite racy.  Will follow up with the correct fix in https://github.com/square/okhttp/pull/6920

But want CI passing for now.

swankjesse (commented):
LGTM

PR #6940, opened 2021-11-28
Backport of Cache Corruption Fix

yschimke:
Backport of #6495


yschimke:
@swankjesse some risk here as the test for the fix was based on the Cache use of Okio FileSystem.

swankjesse (approved):

Issue #6821, opened 2021-08-03
Update snapshot and release publishing after gradle 7 upgrade


yschimke:
Need to use similar approach to okio, replacing uploadArchives with publish.  Currently blocked on signing key.

See okio for modern config https://github.com/square/okio/blob/079c09ea4c5b7dea947596debc478efd26d36922/buildSrc/src/main/kotlin/okio-publishing.gradle.kts

Goooler:
I think we can upgrade Dokka 1.5.0 with this change.

yschimke:
I'm not sure these are related, think they can land separately.

Previous exploration which was blocked at the time, hopefully we can crib the config from similar square projects

https://github.com/square/okhttp/issues/6450
https://github.com/Kotlin/dokka/issues/1592 (fixed)
https://github.com/square/okhttp/pull/6342

Goooler:
I've converted all gradle files to kts but root `build.gradle`, hopefullly to convert it after this.

Goooler:
May refer to [kotlinpoet](https://github.com/square/kotlinpoet) (with latest [gradle-maven-publish-plugin](https://github.com/vanniktech/gradle-maven-publish-plugin) and `Dokka`), which I think it's better than okio.

yschimke:
@Goooler whats the current best advice for this? You seem to have a clearer idea than I do.

Goooler:
It seems like publishing to mavenLocal is ok.

![image](https://user-images.githubusercontent.com/10363352/143808543-8a20adb7-7bb5-4a21-8dc1-4e893c5a3554.png)

Goooler:
Just copy publishing artifacts & website job from okio, anything we left ?

Goooler:
@yschimke Should we add github cache actions on CI? See https://github.com/actions/cache

PR #6941, opened 2021-11-28
Fix publishing of docs

yschimke:
Taken from https://github.com/square/okio/blob/master/.github/workflows/build.yml#L91

To address https://github.com/square/okhttp/issues/6821

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,7 +29,8 @@ jobs:
           java-version: 11
 
       - name: Upload Artifacts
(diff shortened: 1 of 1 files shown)


yschimke:
Won't run on master - so hard to test.

yschimke:
Actually just noticed from the issue https://github.com/square/okio/blob/079c09ea4c5b7dea947596debc478efd26d36922/buildSrc/src/main/kotlin/okio-publishing.gradle.kts

Issue #6943, opened 2021-11-29
Run build failed on apple sillicon chips


Goooler:
I run the project on my new MacBook pro with m1 pro, Adopt JDK download failed.

```
Execution failed for task ':okhttp:compileKotlin'.
> Error while evaluating property 'kotlinJavaToolchainProvider.javaVersion' of task ':okhttp:compileKotlin'
   > Failed to calculate the value of property 'providedJvm$kotlin_gradle_plugin'.
      > Unable to download toolchain matching these requirements: {languageVersion=11, vendor=ADOPTOPENJDK, implementation=vendor-specific}
         > Unable to download toolchain. This might indicate that the combination (version, architecture, release/early access, ...) for the requested JDK is not available.
            > Could not read 'https://api.adoptopenjdk.net/v3/binary/latest/11/ga/mac/aarch64/jdk/hotspot/normal/adoptopenjdk' as it does not exist.

```

JakeWharton:
Adopt doesn't exist anymore anyway. Send a PR to replace with Zulu.

Goooler:
Replaced it with 
```kotlin
  configure<JavaPluginExtension> {
    toolchain {
      languageVersion.set(JavaLanguageVersion.of(11))
      vendor.set(JvmVendorSpec.AZUL)
    }
  }
```
but still failed
```
Execution failed for task ':okhttp:compileKotlin'.
> Error while evaluating property 'kotlinJavaToolchainProvider.javaVersion' of task ':okhttp:compileKotlin'
   > Failed to calculate the value of property 'providedJvm$kotlin_gradle_plugin'.
      > No compatible toolchains found for request filter: {languageVersion=11, vendor=AZUL, implementation=vendor-specific} (auto-detect true, auto-download true)

```

Goooler:
And kapt `info.picocli:picocli-codegen` failed, filed a bug [here](https://github.com/remkop/picocli/issues/1480).

yschimke:
Seems we might be blocked here?

https://github.com/gradle/gradle/issues/19140 - Support "Zulu OpenJDK Discovery API" for auto provisioning toolchains
https://github.com/gradle/gradle/issues/16596 - Support auto-download of Java toolchains for more vendors than just AdoptOpenJDK

yschimke:
Seems like this might be close

https://docs.gradle.org/current/userguide/toolchains.html#sec:provisioning

[...]

PR #6944, opened 2021-11-30
Changing 'adopt' to 'zulu'

geertjanw:
Because Adopt is no longer current.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -25,7 +25,7 @@ jobs:
       - name: Configure JDK
         uses: actions/setup-java@v2
         with:
-          distribution: 'adopt'
+          distribution: 'zulu'
           java-version: 11
 
       - name: Upload Artifacts


yschimke:
This isn't a complete fix, see https://github.com/square/okhttp/issues/6943 and https://github.com/square/okhttp/blob/6d11a31ab5eddf92149b7a045f92745a93eeae34/build.gradle.kts#L82

geertjanw:
Ah, I see, makes sense. Happy to fix this, if you find this direction interesting and relevant to you.

yschimke:
Definitely, but I think the issue is only adoptopenjdk is automatically downloaded, which we rely on.  Are there ways around that?

geertjanw:
You could download Zulu automatically instead.

yschimke:
If it's that simple a fix PR would be appreciated, existing attempt is logged in https://github.com/square/okhttp/issues/6943

geertjanw:
Cool! Will do.

yschimke:
Testing a wider fix here https://github.com/square/okhttp/pull/7028

PR #6947, opened 2021-12-01
The version shouldn't be const

swankjesse:
Downstream projects should get the current version, even if they're
not recompiled.

### DIFF
--- a/okhttp/src/main/kotlin/okhttp3/internal/Util.kt
+++ b/okhttp/src/main/kotlin/okhttp3/internal/Util.kt
@@ -670,4 +670,4 @@ internal inline fun <T> Iterable<T>.filterList(predicate: T.() -> Boolean): List
   return result
 }
 
-const val userAgent: String = "okhttp/${OkHttp.VERSION}"
+val userAgent: String = "okhttp/${OkHttp.VERSION}"
--- a/okhttp/src/main/java-templates/okhttp3/OkHttp.kt
+++ b/okhttp/src/main/java-templates/okhttp3/OkHttp.kt
@@ -31,5 +31,5 @@ object OkHttp {
    *
    * [semver]: https://semver.org
    */
-  const val VERSION = "$projectVersion"
+  val VERSION = "$projectVersion"
 }


oldergod (approved):

yschimke:
Ouch

`e: /home/runner/work/okhttp/okhttp/okhttp/src/main/kotlin/okhttp3/internal/Util.kt: (673, 31): Only 'const val' can be used in constant expressions`

yschimke (approved):

okhttp/src/main/kotlin/okhttp3/internal/Util.kt
```diff
@@ -670,4 +670,4 @@ internal inline fun <T> Iterable<T>.filterList(predicate: T.() -> Boolean): List
   return result
 }
 
-const val userAgent: String = "okhttp/${OkHttp.VERSION}"
+val userAgent: String = "okhttp/${OkHttp.VERSION}"
```
JakeWharton, line 673: This one should be a const though. Reference `$projectVersion` directly?

swankjesse: By templating more code? I don't think that's worth the tiny benefit. If you've got a different idea I'm game.

JakeWharton: Generate the file with the non-const and const rather than templating:

https://github.com/cashapp/zipline/blob/85116a14c55dd2ed28138e1f483b43224943978f/zipline/build.gradle.kts#L17-L37
https://github.com/cashapp/zipline/blob/85116a14c55dd2ed28138e1f483b43224943978f/zipline/build.gradle.kts#L84

swankjesse: Yeah, or generate a templated file with two symbols, one const and one not. Is const actually useful?

We save one string concatenation and pay for it with more code loaded into the heap for the lifetime of the process.

swankjesse:
Need to rebase on multiplatform stuff.

Issue #6699, opened 2021-06-07
HttpLogginInterceptor with bodies but without headers


bartek-wesolowski:
I'd like to use `HttpLoggingInterceptor` to log requests and responses with their bodies. I don't care about headers. I'd like there to be a way to do it via a new log level or a flag, e.g. `loggingInterceptor.logHeaders = false`. I believe I'm not the only one in need of this functionality:
https://stackoverflow.com/questions/56922533/okhttp3-httplogginginterceptor-log-body-without-logging-header

swankjesse:
Our aspirations with the logging interceptor is to include something basic and small. If you need something more capable, I recommend either writing your own interceptor or using a tool like Chucker or Charles.

bartek-wesolowski:
Adding an additional if statement wouldn't make `HttpLoggingInterceptor` much more complicated, but I guess it's your call. I can still copy-paste the code and make the necessary adjustments myself, but I was thinking that more people needed the same functionality.

Also, `isProbablyUtf8()` in `utf8.kt` is internal, so that's another thing you need to copy-paste if you want to extend `HttpLoggingInterceptor`.

swankjesse:
@bartek-wesolowski agreed that it would nbe a small fix. I'm attempting to convince you & others that there are better tools for this job than the logging interceptor. I use Charles for this stuff and it's wonderful. It even works with protocol buffers!

swankjesse:
Yep!

yschimke:
I added a nub of an article with two links.  If you have a recent screenshot of Charles with realistic traffic, maybe edit with a screenshot if you feel like it.  https://stackoverflow.com/a/67954642/1542667

Issue #6085, opened 2020-05-24
Exclude sensitive information from body


raderio:
We need a way to hide(replace with "****") some sensitive information from the body.

Is there something similar to `redactHeader` from `HttpLoggingInterceptor` but for the body properties?

yschimke:
We generally don't add incremental new features to the http logging interceptor. Even if clearly useful like this.

I'd suggest you fork this module and add the features you need to a local codebase. If you feel strongly it should be in a shared utility library instead of your codebase, you could even publish it as a library on GitHub.  There are lots of individual requests that all make sense for this area, which is why we don't add them. The change rate of the public API for this would be much faster than the rest of the core of OkHttp.

PR #6948, opened 2021-12-02
Logging body limit for HttpLoggingInterceptor

mikekulasinski-kulsoft:
Sometimes the message passed to `HttpLoggingInterceptor.Logger#log` is too big and some clients might get OOM when parsing the message. 

Solution is to add a body size limit that will only log body that has less than X bytes and for anything above that limit it will just print a message `<-- END HTTP (too long body X-byte omitted)`

### DIFF
--- a/okhttp-logging-interceptor/src/main/kotlin/okhttp3/logging/HttpLoggingInterceptor.kt
+++ b/okhttp-logging-interceptor/src/main/kotlin/okhttp3/logging/HttpLoggingInterceptor.kt
@@ -43,6 +43,12 @@ class HttpLoggingInterceptor @JvmOverloads constructor(
 
   @Volatile private var headersToRedact = emptySet<String>()
 
+  /**
+   * Maximum body size in bytes that will get logged otherwise the body will
+   * be omitted. Works only for level [Level.BODY]
+   */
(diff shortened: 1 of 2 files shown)


yschimke:
We are holding a pretty consistent line on this.  The cadence of changes to OkHttp and potential to cause issues on upgrades make it an awkward public API to maintain within OkHttp itself.

https://github.com/square/okhttp/issues/6085#issuecomment-633201058

> 
> We generally don't add incremental new features to the http logging interceptor. Even if clearly useful like this.
> 
> I'd suggest you fork this module and add the features you need to a local codebase. If you feel strongly it should be in a shared utility library instead of your codebase, you could even publish it as a library on GitHub. There are lots of individual requests that all make sense for this area, which is why we don't add them. The change rate of the public API for this would be much faster than the rest of the core of OkHttp.

Also see https://github.com/square/okhttp/issues/6699 where there is a link to a https://github.com/bartek-wesolowski/customizable-okhttp-logging-interceptor if this is helpful.

mikekulasinski-kulsoft:
OK Thanks for explanation, the GitHub project you pointed out is perfect. Thanks!

PR #6960, opened 2021-12-15
Add ability to make assertions on raw data sent to MockWebServer

nowkarol:
I've added RawSocketClient to test current MockWebServer behavior while using EXPECT_CONTINUE and (in future) CONTINUE_ALWAYS. HttpURLConnection is not sufficient because it hides 100 Continue response and don't wait for it but sends request body immediately.

### DIFF
--- a/mockwebserver/src/test/java/mockwebserver3/MockWebServerTest.java
+++ b/mockwebserver/src/test/java/mockwebserver3/MockWebServerTest.java
@@ -32,6 +32,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
 import javax.net.ssl.HttpsURLConnection;
+import mockwebserver3.rawsocketclient.RawSocketClient;
 import okhttp3.Handshake;
 import okhttp3.Headers;
 import okhttp3.HttpUrl;
@@ -485,20 +486,31 @@ public final class MockWebServerTest {
   }
 
   @Test public void http100Continue() throws Exception {
-    server.enqueue(new MockResponse().setBody("response"));
+    server.enqueue(new MockResponse().setSocketPolicy(SocketPolicy.EXPECT_CONTINUE));
+    RawSocketClient client = new RawSocketClient(server.getPort());
+    client.sendInUtf8(
+      "POST / HTTP/1.1\r\n" +
+      "Expect: 100-Continue\r\n" +
+      "Content-Length: 0\r\n" +
+      "\r\n");
 
-    URL url = server.url("/").url();
-    HttpURLConnection connection = (HttpURLConnection) url.openConnection();
-    connection.setDoOutput(true);
-    connection.setRequestProperty("Expect", "100-Continue");
-    connection.getOutputStream().write("request".getBytes(UTF_8));
+    RecordedRequest requestWithExpect = server.takeRequest();
+    assertThat(requestWithExpect.getBody().readUtf8()).isEqualTo("");
 
-    InputStream in = connection.getInputStream();
-    BufferedReader reader = new BufferedReader(new InputStreamReader(in, UTF_8));
-    assertThat(reader.readLine()).isEqualTo("response");
+    String response = client.readResponseUtf8String();
 
-    RecordedRequest request = server.takeRequest();
-    assertThat(request.getBody().readUtf8()).isEqualTo("request");
+    // MockWebServer currently doesn't wait for announced request
+    assertThat(response).isEqualTo(
+      "HTTP/1.1 100 Continue\r\n" +
+      "Content-Length: 0\r\n" +
+      "\r\n" +
+      "HTTP/1.1 200 OK\r\n" +
+      "Content-Length: 0\r\n" +
+      "\r\n");
+
+    client.sendInUtf8("announced request");
+    // cannot make MockServer assertion on this request because it has no headers
+    client.close();
   }
 
   @Test public void testH2PriorKnowledgeServerFallback() {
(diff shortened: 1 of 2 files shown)


mockwebserver/src/test/java/mockwebserver3/rawsocketclient/RawSocketClient.java
```diff
@@ -0,0 +1,42 @@
...
+
+  public void sendInUtf8(String stringToSend) throws IOException {
+    socket.getOutputStream().write(stringToSend.getBytes(UTF_8));
+  }
+
+  public String readResponseUtf8String() throws IOException {
+    InputStream in = socket.getInputStream();
+    byte[] array = new byte[MAX_RESPONSE_CHUNK_SIZE_IN_BYTES];
+    int bytesRead = in.read(array);
+    return new String(trimArrayTo(array, bytesRead), UTF_8);
```
yschimke, line 29: can be done via

`String(array, 0, bytesRead, UTF_8)`

mockwebserver/src/test/java/mockwebserver3/rawsocketclient/RawSocketClient.java
```diff
@@ -0,0 +1,42 @@
+package mockwebserver3.rawsocketclient;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.Socket;
+
+import static java.net.InetAddress.getLoopbackAddress;
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+public class RawSocketClient implements AutoCloseable {
```
yschimke, line 10: Not sure about RawSocketClient as a top level class.  If it's required for one specific test then it should probably be in that test.  If it's intended for reuse then I'm concerned about the maintenance of the tests, ideally we could operate at a higher level above a raw socket.

I don't have a clear answer, just giving my initial thoughts.

nowkarol: I would like to use it in other tests. It would probably require some hairy reflection to wrap socket used by other clients by sniffer.

I think it is important to be able to make assertions during exchanges, for example to make sure that server responds after receiving request. It would require to make assertions on whole conversation, something like automatic tcpdump during tests. That probably would be hard and not so readable.

I think it should be used only:
- to test MockWebServerTest during improper client behavior so we can make sure that it is recognized and it will be visible if OkHttp behaves this way
- to test exchange with multiple requests/responses with are hidden by clients (like with 1xx codes)

mockwebserver/src/test/java/mockwebserver3/rawsocketclient/RawSocketClient.java
nowkarol: right, done

mockwebserver/src/test/java/mockwebserver3/MockWebServerTest.java
```diff
@@ -485,20 +486,31 @@ public void tearDown() throws Exception {
...
+    RawSocketClient client = new RawSocketClient(server.getPort());
+    client.sendInUtf8(
+      "POST / HTTP/1.1\r\n" +
+      "Expect: 100-Continue\r\n" +
+      "Content-Length: 0\r\n" +
+      "\r\n");
+
+    RecordedRequest requestWithExpect = server.takeRequest();
+    assertThat(requestWithExpect.getBody().readUtf8()).isEqualTo("");
+    Thread.sleep(100); // wait for whole response
```
nowkarol, line 499: I've realized that response is flushed partially so there is race condition here :( I could change `readResponseUtf8String` method to read everything until socket closes but it would disable ability to make assertions during exchanges. This ability is not used yet but will be needed. For example after changing MockWebServer so it waits for request body after Expect: 100-Continue

swankjesse:
I don’t think this is a MockWebServer, this is a MockSocket. It’s a neat API, but I don’t think it belongs in this project.

swankjesse:
(I misunderstood what was going on here, so I’m gonna retract my quick-close. I misread the new class as being public API, but it’s test API.)

swankjesse:
That said, I’d prefer to not build framework around this. 

OkHttpClient is our spec-compliant HTTP client. It’s expected to behave reasonably with ‘tricky’ HTTP servers.
MockWebServer is our tricky HTTP server for testing. It has mechanisms to speak weird and slow HTTP.

This introduces a low-level tricky HTTP client to test MockWebServer. In theory we could write tons of tests on how MockWebServer handles lots of slow/attacker/fuzzy clients. But that’s a giant expansion of responsibility for MockWebServer.

I’d rather we test this specifically with OkHttp.

nowkarol:
In order to implement 102 support I need to make changes to MockWebServer. To ensure that changes are right I need to test them. Using OkHttp to test tool which is used to test OkHttp doesn’t seem right. Probably that’s why in such tests `HttpUrlConnection` client is used but it hides details of 1xx exchanges.

Adding ability to return interim 102 response from MockWebServer seems pretty similar to existing Socket policies `EXPECT_CONTINUE` and `CONTINUE_ALWAYS`. I think that they don't behave as they supposed to do:
server doesn’t wait for request body but returns 200 right after 100 so it doesn’t follow RFC.

In case of `CONTINUE_ALWAYS` OkHttp doesn’t change it’s behavior, it just assumes that after unexpected 100 server will follow with 200 (that is what MockWebServer does so tests pass).
Aligning with server response (despite it’s not correct) would require to resend body. Even if this is fine it would be nice to test how will OKHttp react to server which expect to get body second time.

So for me MockWebServer behaves tricky when I think it should not. RawSocketClient adds possibility to write tests which catch such cases.

swankjesse:
I agree that it's annoying to test MockWebServer against itself. Historically we confirm interop if new features against other HTTP clients or servers manually, then commit known-working things to OkHttp and MockWebServer. We are exposed to potential regressions this way on PRs that update both client and server. This hasn’t been a problem in practice.

nowkarol:
Ok I won't insist

PR #6962, opened 2021-12-16
Use Kotlin's regex API in MediaType

swankjesse:
Preparing to promote this class to commonMain


okhttp/src/jvmTest/java/okhttp3/MediaTypeTest.kt
```diff
@@ -0,0 +1,280 @@
...
+import okhttp3.MediaType.Companion.toMediaTypeOrNull
+import okhttp3.internal.platform.Platform.Companion.isAndroid
+
+/**
+ * Test MediaType API and parsing.
+ *
+ * This test includes tests from [Guava's](https://code.google.com/p/guava-libraries/)
+ * MediaTypeTest.
+ */
+open class MediaTypeTest {
```
yschimke, line 34: Do we lose any public API protection from moving these to kotlin? I thought we had some API only java tests, but I think we remove them?

yschimke: Mainly about the different preferred forms for kotlin and java.  `MediaType.parse(string)` vs `string.toMediaType()`

okhttp/src/jvmMain/kotlin/okhttp3/MediaType.kt
```diff
@@ -107,31 +106,30 @@ class MediaType private constructor(
     @JvmStatic
     @JvmName("get")
```
yschimke, line 107: Why not move at the same time to common? There seems to be some additional imports required when moving.

```
import kotlin.jvm.JvmName
import kotlin.jvm.JvmOverloads
```

okhttp/build.gradle.kts
```diff
@@ -25,6 +25,8 @@ kotlin {
     }
     val commonTest by getting {
```
yschimke, line 26: I get the following error trying to move a single file to commonMain. What should commonMain be?

```
Execution failed for task ':okhttp:copyKotlinTemplates'.
> Could not copy file '/Users/yschimke/IdeaProjects/okhttp/okhttp/src/commonMain/kotlinTemplates/okhttp3/Headers.kt' to '/Users/yschimke/IdeaProjects/okhttp/okhttp/build/generated/sources/kotlinTemplates/okhttp3/Headers.kt'.
   > Missing property (line) for Groovy template expansion. Defined keys [projectVersion].
```

buildSrc/src/main/kotlin/deps.kt
```diff
@@ -71,6 +71,9 @@ object Dependencies {
   const val kotlinReflect = "org.jetbrains.kotlin:kotlin-reflect:${Versions.kotlin}"
   const val kotlinStdlibOsgi = "org.jetbrains.kotlin:kotlin-osgi-bundle:${Versions.kotlin}"
   const val kotlinJunit5 = "org.jetbrains.kotlin:kotlin-test-junit5:${Versions.kotlin}"
+  const val kotlinTest = "org.jetbrains.kotlin:kotlin-test-common:${Versions.kotlin}"
```
JakeWharton, line 74: You don't need all these dependencies anymore. Just kotlin-test on commonTest.

okhttp/build.gradle.kts
yschimke: Ignore, my stupidity.

okhttp/src/jvmMain/kotlin/okhttp3/MediaType.kt
yschimke: I think I appreciate now, how much easier it is to incrementally migrate file by file! Just removing the java.* imports.

okhttp/src/jvmTest/java/okhttp3/MediaTypeTest.kt
yschimke, line 34: why "open"?

yschimke: n.m. MediaTypeGetTest

okhttp/src/jvmTest/java/okhttp3/MediaTypeTest.kt
```diff
@@ -0,0 +1,280 @@
...
+  protected open fun parse(string: String): MediaType = string.toMediaType()
```
yschimke, line 35: I think this should be `string.toMediaTypeOrNull()`, since it was

```
  protected MediaType parse(String string) {
    return MediaType.parse(string);
  }
```
while 

```
    @JvmStatic
    @JvmName("get")
    fun String.toMediaType(): MediaType {
```

and 

```
    @JvmStatic
    @JvmName("parse")
    fun String.toMediaTypeOrNull(): MediaType? {
```

okhttp/src/jvmTest/java/okhttp3/MediaTypeTest.kt
swankjesse: I don't think so. We had two strategies for API compatibility: japicmp (Java source and binary compatibility) and [KotlinSourceModernTest](https://github.com/square/okhttp/blob/master/okhttp/src/jvmTest/java/okhttp3/KotlinSourceModernTest.kt) (Kotlin source compatibility). These tests jobs are testing the behavior, not the API.

okhttp/src/jvmMain/kotlin/okhttp3/MediaType.kt
swankjesse: Yeah, it's gonna be a bit of a slog.

buildSrc/src/main/kotlin/deps.kt
swankjesse: I want that to be true but it didn't work for me on my first attempt. I'll try it again.

okhttp/src/jvmTest/java/okhttp3/MediaTypeTest.kt
swankjesse: Yep, good call!

swankjesse: (I switched it cause I had to get rid of `!!` in the source.)

buildSrc/src/main/kotlin/deps.kt
swankjesse: It doesn’t work as-is. I might need to upgrade things? Later.
```
> Task :okhttp:compileTestKotlinJvm FAILED
e: /Users/jwilson/Projects/okhttp/okhttp/src/jvmTest/java/okhttp3/MediaTypeGetTest.kt: (18, 20): Unresolved reference: assertEquals
e: /Users/jwilson/Projects/okhttp/okhttp/src/jvmTest/java/okhttp3/MediaTypeGetTest.kt: (19, 20): Unresolved reference: assertFailsWith
e: /Users/jwilson/Projects/okhttp/okhttp/src/jvmTest/java/okhttp3/MediaTypeTest.kt: (21, 20): Unresolved reference: Test
```

yschimke (approved):

PR #6964, opened 2021-12-16
Use Gradle 7.3.2. Log4shell mitigation.

autonomousapps:
Gradle 7.3.2 adds dependency constraints to the _build_ classpath to reject known-bad versions of log4j.

See also https://blog.gradle.org/log4j-vulnerability.

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.3-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.3.2-all.zip
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists


yschimke:
Is this about build dependencies or plugins? We don't have a dependency on log4j, even indirectly, so not sure this is required.  

Also

> Note that the Gradle Build Tool itself is not impacted by this vulnerability as it does not use Log4j. Gradle uses SLF4J and a custom logging implementation not susceptible to the vulnerable string substitution.

autonomousapps:
I don't see any disadvantage to using the latest 7.x patch.

It's about build dependencies and plugins (which are the same thing in this context). You may not have any transitive dependency on log4j now, but that doesn't stop it sneaking in in the future. E.g., the Shadow plugin 7.0.0 has a dependency on log4j 2.14.1.

yschimke:
Sure.  We update gradle regularly, so no object for that reason.  I think it's probably best to avoid PRs with names suggesting a fix for a serious industry wide security that don't affect that project.  It is likely to trigger extra work for clients downstream who assume this is needed.  

Users of OkHttp have always been safe from this PR. I don't see an easy exploit for our CI system since we don't accept input, and it would be a github action DOS rather than hacking us. 

It looks like we should technically update to shadow 7.1.1.  Do you want to add that to this PR?

https://github.com/johnrengelman/shadow/blob/402c14fab1fb241c82898082e7ca58417c75c6f6/src/docs/changes/README.md

yschimke:
Should we wait for a further patch?

https://twitter.com/BlueSpaceCanary/status/1471695226188099589?t=m6Y15Ydy2pSyOO5Hnq5mPg&s=19

autonomousapps:
I'm not sure what your goal is with the condescending replies, but it's your repo, your choice.

I was applying a template to dozens of PRs yesterday as part of my official duties as an employee of Block (née Square), Inc. The PR was a courtesy, nothing more.

yschimke:
Apologies. It seemed like a automated thing whether we had log4j dependencies or not. I was a bit frustrated to get a PR fixing an issue that didn't seem to exist. The shadow plugin you highlighted shows it a real issue for CI.

Sorry for being rude. 

But the question about waiting for a further fix wasn't sarcastic, I literally meant it seems like the situation is continuing.

autonomousapps:
> But the question about waiting for a further fix wasn't sarcastic, I literally meant it seems like the situation is continuing.

Understood, and thank you for saying something. From what little I've read, while the latest iteration of log4j still has vulnerabilities, they're significantly less severe / harder to exploit than the original, so this is still a forward step. I don't know if Gradle intends yet another patch on this subject, but if it concerned you, the blog post I linked at top discusses how to mitigate manually, which would let you tune the version of log4j you considered acceptable for your project.

yschimke:
7.3.3-rc1 updates the checks for 2.7.0

https://docs.gradle.org/7.3.3-rc-1/release-notes.html

autonomousapps:
7.3.3 dropped soon after :D https://docs.gradle.org/7.3.3/release-notes.html

PR #6965, opened 2021-12-17
Use expect/actual for Charset

yschimke:
Expect / Actual for Charset and StandardCharsets

Looked at whether we could use 

ktor-io (want to avoid dependency on ktor)

https://github.com/ktorio/ktor/blob/0081f943b434bdd0afd82424b389629e89a89461/ktor-io/jvm/src/io/ktor/utils/io/charsets/CharsetJVM.kt
https://github.com/ktorio/ktor/blob/0081f943b434bdd0afd82424b389629e89a89461/ktor-io/common/src/io/ktor/utils/io/charsets/Encoding.kt

or kotlix-io (experimental)

https://github.com/Kotlin/kotlinx-io/blob/c5bba0114e33c2afd8e4b1ab214aad94ee416d00/core/commonMain/src/kotlinx/io/text/Charset.kt
https://github.com/Kotlin/kotlinx-io/blob/c5bba0114e33c2afd8e4b1ab214aad94ee416d00/core/jvmMain/src/kotlinx/io/text/JvmCharsets.kt

kotlin.text.Charsets seems safe since they are just java.nio.charset.Charset instances

```
public object Charsets {
    /**
     * Eight-bit UCS Transformation Format.
     */
    @JvmField
    public val UTF_8: Charset = Charset.forName("UTF-8")
```


okhttp/src/commonMain/kotlin/okhttp3/-CommonPlatform.kt
```diff
@@ -0,0 +1,21 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package okhttp3
+
+expect abstract class Charset
+
+expect val Charset.name: String
```
yschimke, line 21: @swankjesse Should we uplift this okio? Or is there a reason that okio doesn't consider Charset outside of jvmMain?

It's pretty central given charset is often negotiated in the response, so it needs to be a first class concept.

swankjesse: I think we skip Charset completely and make multiplatform OkHttp UTF-8 only.

yschimke: Internally we need it, we might get a response with non UTF-8 encoding, and need to provide it as bytes with a Charset identifier.  So even if we avoid conversion APIs, I don't think we can avoid this completely.

Or close this and stick with no Charset in common?

swankjesse: Close this, no Charset in common. As far as I know there's no generic charset encoder/decoder available in JS or native anyway, so we can't do ResponseBody.string() properly unless it's either ASCII or UTF-8. Or if we ship a dependency on ICU which I'd prefer to not.

PR #6966, opened 2021-12-17
Move basic enum types to common and use expect/actual for IOException

yschimke:


okhttp/src/commonMain/kotlin/okhttp3/-CommonPlatform.kt
```diff
@@ -0,0 +1,21 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package okhttp3
+
+expect open class IOException(message: String?, cause: Throwable?) : Exception {
```
yschimke, line 19: Is there a way to share these with Okio?

okhttp/src/commonMain/kotlin/okhttp3/TlsVersion.kt
```diff
@@ -15,6 +15,9 @@
  */
 package okhttp3
```
yschimke, line 16: What is the pattern for 

```

  @JvmName("-deprecated_javaName")
  @Deprecated(
      message = "moved to val",
      replaceWith = ReplaceWith(expression = "javaName"),
  fun javaName(): String = javaName
```

How to only implement in JVM and maintain absolute backwards compatibility?

okhttp/src/commonMain/kotlin/okhttp3/Protocol.kt
```diff
@@ -15,7 +15,7 @@
  */
 package okhttp3
```
yschimke, line 16: What to do with

```
  @Deprecated("OkHttp has dropped support for SPDY. Prefer {@link #HTTP_2}.")
  SPDY_3("spdy/3.1"),
```

okhttp/src/commonMain/kotlin/okhttp3/-CommonPlatform.kt
swankjesse: Yes, we should delete this and use `okio.IOException` instead.

okhttp/src/commonMain/kotlin/okhttp3/TlsVersion.kt
swankjesse: We need to create 3 copies of the class:
 * commonMain defines the shared API with expect functions, plus an internal file with shared implementations
 * jvmMain defines actuals, delegating to the shared implementations. Plus JVM-only APIs
 * nonJvmMain is the same as jvmMain, but no JVM-only APIs

Kotlin multiplatform needs a lot of boilerplate if there’s any API differences

okhttp/src/commonMain/kotlin/okhttp3/Protocol.kt
swankjesse: It stays as is. We don't break binary compatibility.

yschimke: I'm assuming enums need to be complete with the same set for all languages, otherwise you can't define their values in commonMain.

okhttp/src/commonMain/kotlin/okhttp3/TlsVersion.kt
yschimke: I got this working, which was really helpful.  Then realised we probably don't want to bring across javaName to other platforms.

okhttp/build.gradle.kts
```diff
@@ -19,14 +19,20 @@ kotlin {
 
   sourceSets {
     val commonMain by getting {
-      kotlin.srcDir("$buildDir/generated/sources/kotlinTemplates")
```
yschimke, line 22: damn

okhttp/src/nonJvmMain/kotlin/okhttp3/TlsVersion.kt
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okhttp3
+
+actual enum class TlsVersion {
```
yschimke, line 18: This loses any string name like javaName, but in truth we don't have one.

swankjesse (approved):

PR #6978, opened 2022-01-02
Gradle upgrade, more specific toolchain

maldito-dot:
@maldito-dot


yschimke:
This PR is against a very old version.

Issue #6910, opened 2021-11-12
Add support for IDNA 2008 (RFC 5891)


j-bernard:
I'm aware of issue https://github.com/square/okhttp/issues/1615 but I think there is an easier way to fix this.

There is an trustworthy IDN implementation that actually offers IDNA 2008 compatibility, [ICU4J](https://unicode-org.github.io/icu/userguide/icu4j/), see [IDNA class](https://unicode-org.github.io/icu-docs/apidoc/released/icu4j/com/ibm/icu/text/IDNA.html).

The Android platform already replaced the java.net.IDN methods with ICU4J however they kept using the static methods that are deprecated as only compliant with IDNA 2003.

See my  [bug report](https://issuetracker.google.com/issues/206015971) on Android issue tracker for more details and snippet to correctly use ICU4J with the right flags.

yschimke:
If this is part of the public Android APIs, we could potentially use the better implementation on Android, while leaving JDK as is.

How common are these domains? What are some examples that are currently broken?

<img width="666" alt="image" src="https://user-images.githubusercontent.com/231923/141505696-0a8a8f2a-4e81-429a-9aaa-331da214dd7e.png">

j-bernard:
See https://unicode.org/reports/tr46/#IDNAComparison for differences between the new and old standards.

I don't have statistics about these domains but at least one pretty common German character (Eszet: ß) is impacted.
I found a domain that seems to have been created especially to illustrate the problems of non-compliance with IDNA 2008, see [this post](https://www.lookout.net/2011/06/risks-of-using-or-s-in-domain-names.html): http://straße.de/. You won't get the same page with Google Chrome (IDNA 2003) and Firefox (IDNA 2008).

yschimke:
@j-bernard Care to take a look at https://github.com/square/okhttp/pull/6912

What flags should I be passing to getUTS46Instance?

j-bernard:
@yschimke thanks, I'll have a look at the PR.

You should pass the following flags
`IDNA.CHECK_BIDI, IDNA.CHECK_CONTEXTJ, IDNA.CHECK_CONTEXTO, IDNA.USE_STD3_RULES` and depending on the way you convert `IDNA.NONTRANSITIONAL_TO_ASCII`  or `IDNA.NONTRANSITIONAL_TO_UNICODE`.

Android doc seems to say API level 24 but I did not conduct any test on old versions.

yschimke:
[...]

PR #6995, opened 2022-01-08
First draft of a Punycode decoder

swankjesse:
Still to do:
 - Punycode encoder
 - Handle overflows, including tests
 - Handle wrong character types, including tests

### DIFF
--- a/okhttp/src/commonMain/kotlin/okhttp3/internal/Punycode.kt
+++ b/okhttp/src/commonMain/kotlin/okhttp3/internal/Punycode.kt
@@ -1,0 +1,161 @@
+/*
+ * Copyright (C) 2022 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


yschimke:
Are you sure you want to write this yourself? The eoncoding might be possible but the rules seem nasty.

For JS wouldn't we prefer to use

https://nodejs.org/api/url.html#the-whatwg-url-api
https://github.com/mathiasbynens/punycode.js

Or is it possible to support encoding in kotlin, but outsource the checks?

swankjesse:
@yschimke I want something that'll work in native, JVM, and JS. I haven't yet looked too hard at nameprep, which is where I expect the scariest dragons are.

yschimke:
If you are implementing it from base principles, it opens up questions like https://github.com/square/okhttp/issues/6910

This seems ambitious, do you really want to maintain the correctness of this? 

Will you only use this on platforms without an existing IDNA implementation? Would you allow using ICU libraries if installed?

PR #6996, opened 2022-01-08
Headers multiplatform

yschimke:
Minimal API surface area is common.  In part because of complexities of types like date, but also lack of sorted map for KMP.


okhttp/src/commonMain/kotlin/okhttp3/internal/-HeadersCommon.kt
```diff
@@ -0,0 +1,184 @@
+/*
+ * Copyright (C) 2022 Square, Inc.
```
yschimke, line 2: What are the copyright considerations given above is 

> Licensed to the Apache Software Foundation

yschimke:
@swankjesse The most relevant thing on this review for me is which methods should be part of the common interface.  I've dropped a lot of them, especially anything related to build from a response or unsafe variants.  Figuring it is easier to add back than remove.

okhttp/src/commonMain/kotlin/okhttp3/internal/-HeadersCommon.kt
swankjesse: ASF is a legal entity that owns all the code originally forked from Apache Harmony.
Easily confused with the Apache license, which is a grant from a legal entity.

(If the ASF owns copyright on things, let’s not accidentally lose track of that fact.)

okhttp/src/commonMain/kotlin/okhttp3/internal/-HeadersCommon.kt
```diff
@@ -0,0 +1,192 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package okhttp3.internal
+
+import okhttp3.Headers
+
+// TODO find out why [] doesn't fail and returns undefined
```
swankjesse, line 21: I think it’s just how JavaScript arrays work

yschimke: So should I leave it with the existing behaviour?

okhttp/src/nonJvmMain/kotlin/okhttp3/Headers.kt
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  actual fun values(name: String): List<String> = commonValues(name)
+
+  /**
+   * Returns the number of bytes required to encode these headers using HTTP/1.1. This is also the
+   * approximate size of HTTP/2 headers before they are compressed with HPACK. This value is
+   * intended to be used as a metric: smaller headers are more efficient to encode and transmit.
+   */
+  fun byteCount(): Long {
```
swankjesse, line 64: Promote to common?

swankjesse: I think this is duplicated otherwise.

okhttp/src/nonJvmMain/kotlin/okhttp3/Headers.kt
```diff
@@ -0,0 +1,118 @@
...
+
+  actual val size: Int
+    get() = namesAndValues.size / 2
+
+  actual fun name(index: Int): String = commonName(index)
+
+  actual fun value(index: Int): String = commonValue(index)
+
+  actual fun names(): Set<String> {
+    // TODO check ok with unsorted set
```
swankjesse, line 53: Yuck yuck yuck. Particularly ugly ’cause the order is shuffled from hashing, right?

Can we put an orderedSet in `internal`? I hate nondeterministic things.

swankjesse: Ahhh, no so the issue is that this Set is case-sensitive and that’s incorrect. `Cookie` and `cookie` both show up in the result, and that’s strictly incorrect.

swankjesse: Failing test or tracking bug for this?

swankjesse (approved):

okhttp/src/nonJvmMain/kotlin/okhttp3/Headers.kt
yschimke: I didn't implement it yet as we aren't to the point of needing it for event listener metrics.

okhttp/src/nonJvmMain/kotlin/okhttp3/Headers.kt
yschimke:
```
    return (0 until size step 2).map { name(it) }.distinctBy { it.lowercase() }.toSet()
```

okhttp/src/jsTest/kotlin/okhttp3/HeadersJsTest.kt
```diff
@@ -0,0 +1,41 @@
...
+
+import assertk.assertThat
+import assertk.assertions.isEqualTo
+import kotlin.test.Test
+
+class HeadersJsTest {
+  @Test
+  fun nameIndexesAreStrict() {
+    val headers = Headers.headersOf("a", "b", "c", "d")
+    assertThat(headers.name(-1)).isEqualTo(undefined)
```
swankjesse, line 27: I think this should throw and not return undefined.

okhttp/src/commonMain/kotlin/okhttp3/internal/-HeadersCommon.kt
swankjesse: Your original behavior that defended against undefined is my preference.

PR #6998, opened 2022-01-09
ResponseBody multiplatform

yschimke:
Removed most of the quick accessors, `body.string()` is arguably the most useful but requires agreeing how to handle charset, while caller deals with this in the case of `body.source().readUtf8()`


okhttp/src/jvmTest/java/okhttp3/ResponseBodyJvmTest.kt
```diff
@@ -0,0 +1,491 @@
+/*
```
yschimke, line 1: Ignore this as it's in https://github.com/square/okhttp/pull/6997

buildSrc/src/main/kotlin/deps.kt
```diff
@@ -96,4 +96,5 @@ object Dependencies {
   const val checkStyle = "com.puppycrawl.tools:checkstyle:${Versions.checkStyle}"
   const val signatureAndroid21 = "net.sf.androidscents.signature:android-api-level-21:5.0.1_r2@signature"
   const val signatureJava18 = "org.codehaus.mojo.signature:java18:1.0@signature"
+  const val assertk = "com.willowtreeapps.assertk:assertk:0.25"
```
swankjesse, line 99: oooh neat. Repo & sponsor look legit.

okhttp/src/commonMain/kotlin/okhttp3/ResponseBody.kt
```diff
@@ -0,0 +1,128 @@
...
+  abstract fun contentLength(): Long
+
+  abstract fun source(): BufferedSource
+
+  override fun close()
+
+  companion object {
+    /**
+     * Returns a new response body that transmits this string. If [contentType] is non-null and
+     * lacks a charset, this will use UTF-8.
```
swankjesse, line 112: There’s a catch here. We won’t honor charsetString on non-JVM platforms because we don’t have an encoder.

Should we crash in such cases?

swankjesse: (I think we probably crash on the JVM if we’re given an unknown charset. We should emulate that behavior on other platforms, throwing an exception to indicate that the charset is unknown.)

okhttp/src/commonMain/kotlin/okhttp3/internal/-ResponseBodyCommon.kt
```diff
@@ -0,0 +1,72 @@
...
+import okio.BufferedSource
+import okio.ByteString
+import okio.IOException
+import okio.use
+
+internal fun ResponseBody.commonBytes() = commonConsumeSource(BufferedSource::readByteArray) { it.size }
+
+internal fun ResponseBody.commonByteString() = commonConsumeSource(BufferedSource::readByteString) { it.size }
+
+internal inline fun <T : Any> ResponseBody.commonConsumeSource(
```
swankjesse, line 32: Take it or leave it: is the amount of code saved here worth the abstraction? The original code was really simple, this is pretty fancy.

swankjesse: TIL: this abstraction predates this refactor. I had no idea. I don’t care for it!

swankjesse: FYI @JakeWharton !

okhttp/src/commonTest/kotlin/okhttp3/ResponseBodyTest.kt
```diff
@@ -0,0 +1,110 @@
...
+  companion object {
+    @JvmOverloads
+    fun body(hex: String, charset: String? = null): ResponseBody {
+      val mediaType = if (charset == null) null else "any/thing; charset=$charset".toMediaType()
+      return hex.decodeHex().toResponseBody(mediaType)
+    }
+  }
+}
+
+abstract class ForwardingSource(
```
swankjesse, line 102: Okio includes this now

okhttp/src/jvmMain/kotlin/okhttp3/ResponseBody.kt
```diff
@@ -120,8 +52,9 @@ abstract class ResponseBody : Closeable {
    * may trigger an [OutOfMemoryError]. Prefer to stream the response body if this is a
    * possibility for your response.
    */
+  // TODO consider which of these should be part of common API
```
swankjesse, line 55: I think both of them

okhttp/src/jvmMain/kotlin/okhttp3/ResponseBody.kt
```diff
@@ -162,6 +79,7 @@ abstract class ResponseBody : Closeable {
    *
    * Otherwise the response bytes are decoded as UTF-8.
    */
+  // TODO consider which of these should be part of common API
```
swankjesse, line 82: Not this

okhttp/src/jvmMain/kotlin/okhttp3/ResponseBody.kt
```diff
@@ -182,14 +100,15 @@ abstract class ResponseBody : Closeable {
    * may trigger an [OutOfMemoryError]. Prefer to stream the response body if this is a
    * possibility for your response.
    */
+  // TODO consider which of these should be part of common API
```
swankjesse, line 103: I think we include this in the common API, but throw an exception if the charset is specified but is not UTF-8.

swankjesse (approved):

okhttp/src/commonTest/kotlin/okhttp3/ResponseBodyTest.kt
yschimke: Only in jvm as far as I can tell?

okhttp/src/commonMain/kotlin/okhttp3/internal/-ResponseBodyCommon.kt
yschimke: I'll follow up on this, there is a fair amount of logic there.

okhttp/src/commonTest/kotlin/okhttp3/ResponseBodyTest.kt
yschimke: https://github.com/square/okio/pull/1083

okhttp/src/jvmMain/kotlin/okhttp3/ResponseBody.kt
yschimke: It's a bit nasty that UnsupportedEncodingException is an IOException, likely to cause retries rather than early failure.

yschimke: On JVM we currently fallback to default, which seems safe there, but not on non JVM.

```
  fun charset(defaultValue: Charset? = null): Charset? {
    val charset = parameter("charset") ?: return defaultValue
    return try {
      Charset.forName(charset)
    } catch (_: IllegalArgumentException) {
      defaultValue // This charset is invalid or unsupported. Give up.
    }
  }
```

swankjesse: Hmmm... falling back to UTF-8 universally feels bad to me. I don't really like our JVM behavior either.

I see what you're saying regarding IOExceptions. But for non-JVM... I still think our least bad option is to throw an IOException. I don't think a RuntimeException is appropriate either; the client shouldn't crash if it can't understand the server.

PR #6999, opened 2022-01-09
CacheControl MultiPlatform

yschimke:
To Discuss:

Dates

Also depends on Headers, which hasn't landed yet.


okhttp/src/commonMain/kotlin/okhttp3/CacheControl.kt
```diff
@@ -0,0 +1,172 @@
...
+    fun noStore(): Builder
+
+    /**
+     * Sets the maximum age of a cached response. If the cache response's age exceeds [maxAge], it
+     * will not be used and a network request will be made.
+     *
+     * @param maxAge a non-negative integer. This is stored and transmitted with [TimeUnit.SECONDS]
+     *     precision; finer precision will be lost.
+     */
+//    fun maxAge(maxAge: Int, timeUnit: TimeUnit): Builder
```
yschimke, line 117: I'm happy to land this and iterate, but marking these as required.

okhttp/src/jvmTest/java/okhttp3/CacheControlJvmTest.kt
```diff
@@ -0,0 +1,208 @@
...
+import okhttp3.CacheControl.Companion.parse
+import okhttp3.Headers.Companion.headersOf
+import org.assertj.core.api.Assertions
+import org.junit.jupiter.api.Test
+import java.lang.Exception
+import java.lang.IllegalArgumentException
+import java.util.concurrent.TimeUnit
+
+class CacheControlJvmTest {
+    @Test
```
swankjesse, line 27: what’s with the spacing?

okhttp/src/jvmTest/java/okhttp3/CacheControlJvmTest.kt
```diff
@@ -0,0 +1,208 @@
...
+            .noCache()
+            .noStore()
+            .maxAge(1, TimeUnit.SECONDS)
+            .maxStale(2, TimeUnit.SECONDS)
+            .minFresh(3, TimeUnit.SECONDS)
+            .onlyIfCached()
+            .noTransform()
+            .immutable()
+            .build()
+        Assertions.assertThat(cacheControl.toString()).isEqualTo(
```
swankjesse, line 40: static import assertThat?

okhttp/src/jvmTest/java/okhttp3/CacheControlJvmTest.kt
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    @Test
+    fun parseCacheControlAndPragmaAreCombined() {
+        val headers = headersOf("Cache-Control", "max-age=12", "Pragma", "must-revalidate", "Pragma", "public")
+        val cacheControl = parse(headers)
+        Assertions.assertThat(cacheControl.toString()).isEqualTo("max-age=12, public, must-revalidate")
+    }
+
+    // Testing instance equality.
```
swankjesse, line 135: nit: equality is kind of the wrong word here, we’re testing instance identity?

swankjesse: This comment made more sense when it was on an `@SuppressWarnings` thing. Right now the comment is pretty unnecessary; the `isSameAs` does it all.

swankjesse: ALSO the redundant string construction that motivated the warning is missing? I think that’s intended to confirm we get the same instance, working against `String.intern()` on the JVM?

okhttp/src/nonJvmMain/kotlin/okhttp3/CacheControl.kt
```diff
@@ -0,0 +1,98 @@
...
+  actual val isPublic: Boolean,
+  actual val mustRevalidate: Boolean,
+  actual val maxStaleSeconds: Int,
+  actual val minFreshSeconds: Int,
+  actual val onlyIfCached: Boolean,
+  actual val noTransform: Boolean,
+  actual val immutable: Boolean,
+  private var headerValue: String?
+) {
+  override fun toString(): String {
```
swankjesse, line 33: promote to common?

okhttp/src/nonJvmMain/kotlin/okhttp3/CacheControl.kt
```diff
@@ -0,0 +1,98 @@
...
+    actual fun noTransform() = apply {
+      this.noTransform = true
+    }
+
+    actual fun immutable() = apply {
+      this.immutable = true
+    }
+
+    actual fun build(): CacheControl {
+      return CacheControl(noCache, noStore, maxAgeSeconds, -1, false, false, false, maxStaleSeconds,
```
swankjesse, line 88: this is a really good place to use named parameters.

okhttp/src/nonJvmMain/kotlin/okhttp3/CacheControl.kt
```diff
@@ -0,0 +1,98 @@
...
+          minFreshSeconds, onlyIfCached, noTransform, immutable, null)
+    }
+  }
+
+  actual companion object {
+    actual val FORCE_NETWORK = Builder()
```
swankjesse, line 94: promote to common.

@egorand had good advice when we were doing this in Okio. EVERYTHING that is shared goes in a commonXxx function, no matter how small. I like that ’cause it’s a simple rule, and because it guarantees we don’t miss a platform when changing things.

swankjesse (approved):

swankjesse:
For dates . . . . unsure.
For Durations . . . kotlin.time.Duration is stable in Kotlin 1.6. Let’s upgrade to Kotlin 1.6 and use Duration everywhere.

swankjesse:
Also `DurationUnit` is in Kotlin 1.6!

okhttp/src/jvmTest/java/okhttp3/CacheControlJvmTest.kt
yschimke: Not sure, I have the square format configured.  The source was 2, if I format it's 4.  Intellij quirk?

okhttp/src/jvmTest/java/okhttp3/CacheControlJvmTest.kt
yschimke: Seems useful, particularly if these are from header compression at http layer.  I'll remove comment but keep test.

yschimke:
> Also `DurationUnit` is in Kotlin 1.6!

I'll follow up with this, land this first so that PR focuses on dates.

PR #7001, opened 2022-01-09
Load in Stringprep data for RFC 3454

swankjesse:


swankjesse:
This confirms we can load the data from the RFC. We still need to:
 * Decide which RFC to use 
 * Pick a more efficient format for the data, both for memory size & CPU efficiency
 * Implement algorithms that honor this data.

See also:
https://datatracker.ietf.org/doc/html/rfc3454
https://datatracker.ietf.org/doc/html/rfc3491

yschimke:
Impressive work.

oldergod (approved):

swankjesse:
Okay this is ready to land. The code is all currently in `jvmTest`, which is a safe place to iterate until it’s complete enough to land in nonJvmMain. The biggest blocker on really getting it into nonJvmMain is including resources in a non-JVM artifact.

oldergod (approved):

PR #7002, opened 2022-01-09
CacheControl time unit

yschimke:


okhttp/src/jvmMain/kotlin/okhttp3/internal/-JvmPlatform.kt
```diff
@@ -0,0 +1,23 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package okhttp3.internal
+
+import java.util.concurrent.TimeUnit as JavaTimeUnit
```
yschimke, line 19: It seems a shame that kotlin doesn't handle this issue for us.  But if we want both platforms to have their natural time unit implementation, without doubling all methods, it probably means this.

We can also define expect values for the enum as well, I assume.

okhttp/src/commonMain/kotlin/okhttp3/internal/-CommonPlatform.kt
```diff
@@ -0,0 +1,21 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package okhttp3.internal
+
+expect enum class TimeUnit
```
swankjesse, line 19: I don’t think it’s reasonable for us to put public APIs in our `internal` package

okhttp/src/jvmMain/kotlin/okhttp3/internal/-JvmPlatform.kt
swankjesse: I think we’re actually stuck doubling our methods.

As-is I don’t think callers can use this API in multiplatform code, unless they exclusively use `okhttp3.internal.TimeUnit` in their implementations.

swankjesse:
I see several bad options:

 * We accept `okhttp3.internal.TimeUnit`. This is bad because anyone writing multiplatform code needs to import our internal API.
 * We create a new public API, `okhttp3.TimeUnit`, and alias it to `java.util.concurrent.TimeUnit` on JVM and `kotlin.time.DurationUnit` on non-JVM. This is bad because multiplatform callers needs to use OkHttp’s type instead of the existing multiplatform type (``kotlin.time.DurationUnit`) throughout their code. Callers can’t even cast from one to the other in multiplatform code, because on the JVM the cast would fail.
 * We use `kotlin.time.DurationUnit` only. This is bad because it’s not backwards-compatible with our shipping releases.
 * We use `java.util.concurrent.TimeUnit` only. This is bad because it isn’t available on JS and native.

Therefore I think our least worst option is to double the number of methods we have! The ugliest one is on our `InterceptorChain` interface, where adding new methods is not backwards-compatible. I think it’s worth doing and breaking backwards-compatibility there. I don’t expect that interface is widely-implemented.

okhttp/src/commonMain/kotlin/okhttp3/internal/-CommonPlatform.kt
yschimke: That was a mistake, but no longer relevant.  Agree with you on this being public API.  We aren't the owners of TimeUnit.  I think okio.IOException is justifiable.  But we aren't OkTime.

swankjesse (approved):

PR #7003, opened 2022-01-09
Fix indexing

yschimke:


okhttp/src/nonJvmMain/kotlin/okhttp3/Headers.kt
```diff
@@ -50,7 +50,7 @@ actual class Headers internal actual constructor(
   actual fun value(index: Int): String = commonValue(index)
 
   actual fun names(): Set<String> {
-    return (0 until size step 2).map { name(it) }.distinctBy { it.lowercase() }.toSet()
```
yschimke, line 53: Fixing a bug - now covered by a test.

swankjesse (approved):

PR #7006, opened 2022-01-10
Handshake discussion starter

yschimke:


okhttp/src/commonMain/kotlin/okhttp3/Handshake.kt
```diff
@@ -0,0 +1,46 @@
...
+  /** Returns the cipher suite used for the connection. */
+  val cipherSuite: CipherSuite
+
+  /** Returns a possibly-empty list of certificates that identify this peer. */
+  val localCertificates: List<Certificate>
+
+  /** Returns a possibly-empty list of certificates that identify the remote peer. */
+  val peerCertificates: List<Certificate>
+
+  override fun equals(other: Any?): Boolean
```
swankjesse, line 41: Is specifying these useful?

swankjesse (commented):
My read is this proposes using Java’s built-in certificate type on the JVM, and finding or creating something else for non-JVM platforms.

Just pondering alternatives, what about creating our own multiplatform certificate type? We could promote code from Certifikit into OkHttp:

```
data class Certificate(
  /** This is a integer enum. Use 0L for v1, 1L for v2, and 2L for v3. */
  val version: Long,
  val serialNumber: BigInteger,
  val signature: AlgorithmIdentifier,
  val issuer: List<List<AttributeTypeAndValue>>,
  val validity: Validity,
  val subject: List<List<AttributeTypeAndValue>>,
  val subjectPublicKeyInfo: SubjectPublicKeyInfo,
  val issuerUniqueID: BitString?,
  val subjectUniqueID: BitString?,
  val extensions: List<Extension>,
  val signatureAlgorithm: AlgorithmIdentifier,
  val signatureValue: BitString
)
```

The big drag here is introducing a ton of new public APIs into OkHttp for not only the certificate, but all its other baggage: SubjectPublicKeyInfo, BitString, AlgorithmIdentifier, AttributeTypeAndValue, etc.

Another drag with this is the complexity of deciding when to transcode a JVM certificate into this format in JVM OkHttp.

Which all makes me reluctant, mostly because we’re dramatically growing OkHttp , either directly or by introducing a dependency on some new oktls thing.

yschimke:
I like the Certifikit idea, but probably via extension methods and maybe provided as an optional module.

I think on every platform, the native certificates are the ones we should default to.  Maybe we have a common Certificate, which is effectively Any.  Then on all platforms we return the native Certificate format, but with the specific platform return type.

By default some minimal accessors common for operating on certificates (getting DN, getting bytes, validating).

And a okhttp-certifikit module with deep accessors for cross platform Certificate handling, always deferred, always done by callers, not us internally?

yschimke:
Generally by default, let's do everything in native formats, with native code, not write it ourselves.

But provide a way for multiplatform code to use us for the non critical parts (post native handshake code).

swankjesse:
I’m not completely sold on relying on native things everywhere. Perhaps we say 'iOS, JVM, and Android only’ ?

But the native things on Linux and Windows will be very different, and will require a lot of adapters etc. to use well. I’m still hurting from Okio’s FileSystem and the unique features of each platform.

yschimke:
Makes sense, although I don't know what the JS implementation is yet, but trying to get enough of the Call tree supported so we can start experimenting with fetch or similar.

yschimke:
JS is likely 2 environments.

a) Node - https://nodejs.org/api/tls.html

b) Browser - Fetch (no obvious access to certs)

For node this shows accessing certificates

```
        val response = suspendCoroutine<IncomingMessage> {
          https.get("https://api.twitter.com/robots.txt") { msg ->
            it.resume(msg)
          }
        }
        println(response.headers.date)

        val tlsSocket = response.socket as TLSSocket

        var cert = tlsSocket.getPeerCertificate(true)

        while (cert != null) {
          val raw = cert.raw.unsafeCast<ByteArray>()
          println(raw.toByteString(0, raw.size))
          println(cert.serialNumber)
          println(cert.subject.CN)
          val issuer = cert.issuerCertificate
          if (issuer.serialNumber == cert.serialNumber) {
            break
          }
          cert = issuer
        }
```

yschimke:
Nodejs support is very early, so we would probably take these definitions into our codebase? https://github.com/Kotlin/kotlinx-nodejs

swankjesse:
I’d like to focus on JVM and native ’cause I think those are most likely for us to get adoption. Browser JS is potentially a third useful target (shared UI things!) but I think it’s a very big stretch for NodeJS support to be adopted.

yschimke:
Closing for now, Handshake isn't blocking forward progress.  It can be JVM only to start with, added on a few platforms while firming up the API.

Issue #506, opened 2014-02-01
Happy Eyeballs


swankjesse:
http://en.wikipedia.org/wiki/Happy_Eyeballs

sriharim25:
@dlubarov , @JakeWharton do you folks happen to be working on bringing Happy Eyeballs support to OkHttp? There are more DHCPv6-based IPv6 networks every day, especially in corporate settings, and the ability to pick the better transport in parallel will greatly benefit user experience.

swankjesse:
It’s possible to implement this manually by making two HTTP requests yourself and picking the one that returns first. Doing it inside of OkHttp is more difficult. For example, what should the interceptors see? We might be able to do further work later to unblock this, but for now it seems very difficult to do well.

mikelambert:
So this is causing issues for IPv6 http requests with React Native, linked above.

It sounds like the workaround suggested above is to do two requests in the above layer, with one forced to IPv4. Curious, how does one force IPv4 in an okhttp request?

Do I need to resolve the DNS to the IPv4 and IPv6 address and add in Host headers manually? Or is there some easier way?

swankjesse:
Implement the Dns interface and strip the addresses you don't want. Be careful to not return an empty list; that case needs to throw an exception.

ericmguimaraes:
Hey guys, by any chance you still planing on doing that? I'm facing the same problem here and the workaround doesn't really look good. It would be great to have a more sophisticated approach built-in the lib.

yschimke:
Linking directly to RFC - https://tools.ietf.org/html/rfc8305

medusar:
So, what is the plan? Will okhttp support Happy Eyeballs ?

dbrgn:
In the [docs](https://square.github.io/okhttp/), I read:

> OkHttp perseveres when the network is troublesome: it will silently recover from common connection problems. If your service has multiple IP addresses OkHttp will attempt alternate addresses if the first connect fails. This is necessary for IPv4+IPv6 and services hosted in redundant data centers.

However, when I run this code in an API 21 (5.0) emulator...

```java
public class MainActivity extends AppCompatActivity {

    @SuppressLint("StaticFieldLeak")
    @Override
    protected void onCreate(Bundle savedInstanceState) {
        super.onCreate(savedInstanceState);
        setContentView(R.layout.activity_main);

        new AsyncTask<Void, Void, Void>() {
            @Override
            protected Void doInBackground(Void... voids) {
[...]

PR #7009, opened 2022-01-12
Add an initial test for Happy Eyeballs

swankjesse:
https://github.com/square/okhttp/issues/506

### DIFF
--- a/okhttp/src/jvmTest/java/okhttp3/HappyEyeballsTest.kt
+++ b/okhttp/src/jvmTest/java/okhttp3/HappyEyeballsTest.kt
@@ -1,0 +1,242 @@
+/*
+ * Copyright (C) 2022 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okhttp3
+
+import java.io.IOException
+import java.net.Inet4Address
+import java.net.Inet6Address
+import java.net.InetAddress
+import kotlin.test.assertFailsWith
+import mockwebserver3.MockResponse
+import mockwebserver3.MockWebServer
+import org.assertj.core.api.Assertions.assertThat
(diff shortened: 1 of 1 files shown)


okhttp/src/jvmTest/java/okhttp3/HappyEyeballsTest.kt
```diff
@@ -0,0 +1,242 @@
...
+
+  private val listener = RecordingEventListener()
+  private lateinit var client: OkHttpClient
+  private lateinit var url: HttpUrl
+
+  /**
+   * This is mutable and order matters. By default, it contains [IPv4, IPv6]. Tests may manipulate
+   * it to prefer IPv6.
+   */
+  private var dnsResults = listOf<InetAddress>()
```
yschimke, line 60: If we want to do this right, we may want to allow for a more specific subclass of Dns interface.

```
interface AsyncDns: Dns {
  fun lookupHosts(): List<Future<List<InetAddress>>>
} 
```

Since 

> When a client has both IPv4 and IPv6 connectivity and is trying to
   establish a connection with a named host, it needs to send out both
   AAAA and A DNS queries.  Both queries SHOULD be made as soon after
   one another as possible, with the AAAA query made first and
   immediately followed by the A query.

   Implementations SHOULD NOT wait for both families of answers to
   return before attempting connection establishment.  If one query
   fails to return or takes significantly longer to return, waiting for
   the second address family can significantly delay the connection
   establishment of the first one.  Therefore, the client SHOULD treat
   DNS resolution as asynchronous.  Note that if the platform does not
   offer an asynchronous DNS API, this behavior can be simulated by
   making two separate synchronous queries on different threads, one per
   address family.

yschimke: Likewise in Android?  https://developer.android.com/reference/android/net/DnsResolver#query(android.net.Network,%20java.lang.String,%20int,%20int,%20java.util.concurrent.Executor,%20android.os.CancellationSignal,%20android.net.DnsResolver.Callback%3C?%20super%20java.util.List%3Cjava.net.InetAddress%3E%3E)

yschimke (approved):

jamesonwilliams:
cc: @EricKuck

okhttp/src/jvmTest/java/okhttp3/HappyEyeballsTest.kt
swankjesse: I think they’re independent phases. 

Look up which hosts to attempt first, then attempt to connect to 2 of them in parallel.

swankjesse: Ahh now I see what you're saying, we can use new APIs to query each set independently. Yep, let's do that too.

PR #7010, opened 2022-01-13
Support HTTP3 draft as h3-29 as an ALPN token

MegatronKing:
It is better to support HTTP3 draft like h3-29. We can remove this code when drafts are no longer used.

### DIFF
--- a/okhttp/src/commonMain/kotlin/okhttp3/Protocol.kt
+++ b/okhttp/src/commonMain/kotlin/okhttp3/Protocol.kt
@@ -121,8 +121,10 @@ enum class Protocol(private val protocol: String) {
         HTTP_2.protocol -> HTTP_2
         SPDY_3.protocol -> SPDY_3
         QUIC.protocol -> QUIC
-        HTTP_3.protocol -> HTTP_3
-        else -> throw IOException("Unexpected protocol: $protocol")
+        else -> {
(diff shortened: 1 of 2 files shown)


okhttp/src/commonMain/kotlin/okhttp3/Protocol.kt
```diff
@@ -121,8 +121,10 @@ enum class Protocol(private val protocol: String) {
         HTTP_2.protocol -> HTTP_2
         SPDY_3.protocol -> SPDY_3
         QUIC.protocol -> QUIC
-        HTTP_3.protocol -> HTTP_3
-        else -> throw IOException("Unexpected protocol: $protocol")
+        else -> {
+          // Support HTTP3 draft like h3-29
+          if (protocol.startsWith(HTTP_3.protocol)) HTTP_3 else throw IOException("Unexpected protocol: $protocol")
```
swankjesse, line 126: Wanna add a quick test case?

swankjesse (approved):

okhttp/src/commonMain/kotlin/okhttp3/Protocol.kt
MegatronKing: Added

swankjesse (approved):

swankjesse:
Thanks.

PR #7017, opened 2022-01-16
Change OkHttpClient.routeDatabase to be internal

swankjesse:
This is the only behavior change in the PR. The rest is converting
a test from Java to Kotlin (so it can see this internal member) and
getting a TaskRunner instance into the ExchangeFinder, hopefully
useful for implementing Happy Eyeballs.

Also fix up some formatting problems.


okhttp/src/jvmMain/kotlin/okhttp3/OkHttpClient.kt
```diff
@@ -227,7 +227,8 @@ open class OkHttpClient internal constructor(
   @get:JvmName("minWebSocketMessageToCompress")
   val minWebSocketMessageToCompress: Long = builder.minWebSocketMessageToCompress
 
-  val routeDatabase: RouteDatabase = builder.routeDatabase ?: RouteDatabase()
+  internal val routeDatabase: RouteDatabase = builder.routeDatabase ?: RouteDatabase()
```
swankjesse, line 230: This is the behavior change. This should never have been non-internal; it was a mistake to have an `internal` type in a public API.

okhttp/src/jvmTest/java/okhttp3/OkHttpClientTest.java
```diff
@@ -1,405 +0,0 @@
...
-      builder.addInterceptor(null);
-      fail();
-    } catch (NullPointerException expected) {
-    }
-  }
-
-  @Test public void nullNetworkInterceptor() {
-    OkHttpClient.Builder builder = new OkHttpClient.Builder();
-    try {
-      builder.addNetworkInterceptor(null);
```
swankjesse, line 185: I removed this test (and a couple similar) because Kotlin doesn’t allow us to express it. We’re letting the type system do something Java required us to do manually.

yschimke (approved):

yschimke:
One comment about internal vs package "okhttp3.internal", when I try to understand behaviour of connections, but maybe in the future Routes, especially for Happy Eyeballs, I tend to use some of these internal APIs.

No objection, but what visibility would you expose for happy eyeballs? Just the existing connection events in EventListener?

swankjesse:
I think we should make it opt-in initially. That way if we make a correctness or performance regression in some edge cases the user has recourse. Then opt-out later!

Issue #7016, opened 2022-01-16
Http2 Connection becomes stalled after interrupt during call.execute


yschimke:
Reproduction in https://github.com/square/okhttp/pull/7014/files

With 1 Good thread - downloads work.

With 1 Good thread + 9 Cancelling (interrupt) threads - Download gets stuck, and other threads get SocketTimeout on following request

With 1 Good thread + 9 cancel + Interceptor setting noNewExchanges - Overall systems stays healthy, but download stays stuck.

yschimke:
Relates to https://github.com/square/okhttp/issues/3146

yschimke:
It looks like this is meant to be covered

```
      val dataStream = getStream(streamId)
      if (dataStream == null) {
        writeSynResetLater(streamId, ErrorCode.PROTOCOL_ERROR)
        updateConnectionFlowControl(length.toLong())
        source.skip(length.toLong())
        return
      }
```

yschimke:
It seems like it could be related to interrupts vs close.

This appears to fix the MediaTest.

```
          try {
            response.body!!.source().skip(8000001)
          } catch (ioe: InterruptedIOException) {
            response.close()
          }
```          

I had thought that an interrupt was meant to cancel the call, but it doesn't seem like it does.

yschimke:
I assume it's the leaked responses that don't get closed, but without being consumed are effectively blocking the entire connection.

yschimke:
Also tangentially relevant https://github.com/coil-kt/coil/blob/main/CHANGELOG.md#200-alpha01---october-11-2021

> Important: Coil now has its own disk cache implementation and no longer relies on OkHttp for disk caching.

    This change was made to:
        Better support thread interruption while decoding images. This improves performance when image requests are started and stopped in quick succession.
        
> You should not use OkHttp's Cache with Coil 2.0 as it can be corrupted if it's interrupted while writing to it.

yschimke:
Gaaahhh!!!

The fix above calling `response.close()` only works after the response has been returned.  If the interrupt happens during call.execute, it still get's into a bad state.

yschimke:
I guess we can't even know the state of the Connection, maybe partially written output, so we'd need to close the whole connection.  As I can do that externally to fix this, I guess the optimal change is to use enqueue, and handle the interrupt separately.

```
Exception in thread "Thread-11" java.io.InterruptedIOException: interrupted
	at okio.Timeout.throwIfReached(Timeout.kt:98)
[...]

PR #7022, opened 2022-01-18
Document cancellation behaviour

yschimke:
Relating to https://github.com/square/okhttp/issues/7016.

Document the issue with interrupts.

### DIFF
--- a/docs/calls.md
+++ b/docs/calls.md
@@ -45,6 +45,15 @@ Calls are executed in one of two ways:
 
 Calls can be canceled from any thread. This will fail the call if it hasn’t yet completed! Code that is writing the request body or reading the response body will suffer an `IOException` when its call is canceled.
 
+## Interrupts vs Cancellations
+
+OkHttp is not currently interrupt safe.  An interrupt in an execute call can result in a headers
+frame being partially written to the connection.  See https://github.com/square/okhttp/issues/7016.
+After this the connection is in an unknown state.
+
+Instead `call.cancel` or `body.close` should be used to cleanly cancel the request.
+If an interrupt happens during reading the response body, then the response must still be closed.
+
 ## Dispatch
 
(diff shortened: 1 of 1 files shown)


swankjesse (approved):

swankjesse:
I think we should fix this rather than advising our users work around it?

Working around it is really hard for anyone canceling Futures that might transitively be performing OkHttp calls.

yschimke:
@swankjesse +1Given the exception bubbles up from newStream to the caller through execute(), any pointers for how to address?

```
Exception in thread "Thread-11" java.io.InterruptedIOException: interrupted
	at okio.Timeout.throwIfReached(Timeout.kt:98)
	at okio.OutputStreamSink.write(JvmOkio.kt:54)
	at okio.AsyncTimeout$sink$1.write(AsyncTimeout.kt:99)
	at okio.RealBufferedSink.flush(RealBufferedSink.kt:267)
	at okhttp3.internal.http2.Http2Writer.flush(Http2Writer.kt:120)
	at okhttp3.internal.http2.Http2Connection.newStream(Http2Connection.kt:268)
	at okhttp3.internal.http2.Http2Connection.newStream(Http2Connection.kt:225)
	at okhttp3.internal.http2.Http2ExchangeCodec.writeRequestHeaders(Http2ExchangeCodec.kt:76)
	at okhttp3.internal.connection.Exchange.writeRequestHeaders(Exchange.kt:63)
	at okhttp3.internal.http.CallServerInterceptor.intercept(CallServerInterceptor.kt:41)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)```
```

a) Remove the stream + send rst_stream - and hope connection is healthy?
b) Assume connection is damaged and set noNewExchanges?
c) try to protect against this, avoid using the same user provided thread for an connection IO write?
d) try to minimise the window, make this effectively uninterruptible? Maybe requires support in Okio?
e) other?

Issue #3107, opened 2017-01-14
Okio will clear the Thread interrupt status


wanpg:
I do not know whether the following question is a bug or not.

I am using Okhttp version 3.4.2.

I have this thread handling code like this.
```java
public abstract class CustomThread<T> extends Thread {

  abstract T execute();
    abstract void onResult(T data);

    @Override
    public void run() {
        T object = null;
        try {
            if (!Thread.currentThread().isInterrupted()) {
                object = execute();
            }
        } catch (Exception e) {
            e.printStackTrace();
        }
        try {
            if (!Thread.currentThread().isInterrupted()) {
                onResult(object);
            }
        } catch (Exception e) {
            e.printStackTrace();
        }
    }
}
```

I want to call okhttp in `abstract T execute ()` for network requests.

When I invoke `CustomThread.interrupt ()` to interrupt the thread, Okhttp probability will call the following method.

```java
/**
 * Throws an {@link InterruptedIOException} if the deadline has been reached or if the current
 * thread has been interrupted. This method doesn't detect timeouts; that should be implemented to
 * asynchronously abort an in-progress operation.
 */
public void throwIfReached() throws IOException {
    if (Thread.interrupted()) {
        throw new InterruptedIOException("thread interrupted");
    }

    if (hasDeadline && deadlineNanoTime - System.nanoTime() <= 0) {
        throw new InterruptedIOException("deadline reached");
    }
}
```

`Thread.interrupted()` will clear the Thread interrupt status. Resulting in `CustomThread` should not be invoked `onResult (object);` has been invoked

I would like to know is that your subsequent version can be changed to call `Thread.currentThread (). IsInterrupted ()` to determine whether the thread interrupted?

swankjesse:
From the [Java documentation](https://docs.oracle.com/javase/tutorial/essential/concurrency/interrupt.html):

> _“By convention, any method that exits by throwing an InterruptedException clears interrupt status when it does so. However, it's always possible that interrupt status will immediately be set again, by another thread invoking interrupt.”_

alexoro:
Hi.

> By convention, any method that exits by throwing an InterruptedException clears interrupt status when it does so 
But it does not throw InterruptedException, it throws InterruptedIOException, which extends IOException.
Also there's SocketTimeoutException (and anything else) which extends InterruptedIOException.
In result we are left in situation when thread-cancelation and socket-errors finish in same way.

swankjesse:
[...]

PR #7023, opened 2022-01-19
Improve interrupts (for discussion)

yschimke:


yschimke:
From https://github.com/square/okhttp/pull/7022

This is a little of 

a) Remove the stream + send rst_stream - and hope connection is healthy?
d) try to minimise the window,

okhttp/src/jvmMain/kotlin/okhttp3/internal/http2/Http2Connection.kt
```diff
@@ -237,6 +237,11 @@ class Http2Connection internal constructor(builder: Builder) : Closeable {
     val stream: Http2Stream
     val streamId: Int
 
+    // trigger interrupts before we write to the buffered output
+    // since flush is likely where we get notified.
+    if (Thread.interrupted())
+      throw InterruptedIOException()
```
yschimke, line 243: Should we reset this?

yschimke: Can change to 

```
    if (Thread.currentThread().isInterrupted)
      throw InterruptedIOException()
```

okhttp/src/jvmMain/kotlin/okhttp3/internal/http/CallServerInterceptor.kt
```diff
@@ -79,6 +80,10 @@ class CallServerInterceptor(private val forWebSocket: Boolean) : Interceptor {
         exchange.finishRequest()
       }
     } catch (e: IOException) {
+      if (e is InterruptedIOException) {
+        exchange.cancel()
+        throw e // Interrupted externally, so fail
```
swankjesse, line 85: Nice comment.

okhttp/src/jvmMain/kotlin/okhttp3/internal/http2/Http2Connection.kt
swankjesse: Is there a test that would differentiate this? I suspect it’s probably not necessary as we’re not doing any blocking calls between here and flush().

swankjesse: An interrupted thread should be in one of these states:

 1. Doing non-blocking work with the `Thread.interrupted()` bit set to true
 2. Throwing an InterruptedException or InterruptedIOException

Your code the idiomatic way to go from 1 to 2.

yschimke: I don't think it was the flush here I was worried about.  It was all the stream accounting work between here and that flush.

I wanted to minimise the race window, but failing before we create the stream.

yschimke: > Is there a test that would differentiate this? I suspect it’s probably not necessary as we’re not doing any blocking calls between here and flush().

I'll see how to create a test here.

okhttp/src/jvmMain/kotlin/okhttp3/internal/http2/Http2Connection.kt
```diff
@@ -237,6 +237,12 @@ class Http2Connection internal constructor(builder: Builder) : Closeable {
     val stream: Http2Stream
     val streamId: Int
 
+    // trigger interrupts before we write to the buffered output
+    // since flush is likely where we get notified.
+    if (Thread.currentThread().isInterrupted) {
```
yschimke, line 242: I've changed this to not clear the status, since we expect this to fail up to a higher level.

okhttp/src/jvmTest/java/okhttp3/internal/http2/Http2ConnectionTest.kt
```diff
@@ -1932,6 +1933,38 @@ class Http2ConnectionTest {
     assertThat(queues).hasSize(1)
   }
 
+  @Test fun interruptDuringNewStream() {
```
yschimke, line 1936: @swankjesse advice on this test

crazysouthern (approved):

okhttp/src/jvmMain/kotlin/okhttp3/internal/http2/Http2Connection.kt
```diff
@@ -237,6 +237,12 @@ class Http2Connection internal constructor(builder: Builder) : Closeable {
     val stream: Http2Stream
     val streamId: Int
 
+    // trigger interrupts before we write to the buffered output
+    // since flush is likely where we get notified.
+    if (Thread.interrupted()) {
+      throw InterruptedIOException()
+    }
+
```
yschimke, line 245: @swankjesse I do wonder whether we should have a try catch around everything here and send a RST_STREAM if we think we may have created a stream.  

Or tear down the entire connection at this point to recover?

okhttp/src/jvmMain/kotlin/okhttp3/internal/http2/Http2Connection.kt
yschimke: https://github.com/square/okhttp/issues/3107

okhttp/src/jvmMain/kotlin/okhttp3/internal/http2/Http2Connection.kt
yschimke: Made this change to tear down the connection.

okhttp/src/jvmMain/kotlin/okhttp3/internal/http2/Http2Connection.kt
```diff
@@ -237,38 +237,53 @@ class Http2Connection internal constructor(builder: Builder) : Closeable {
...
+      if (flushHeaders) {
+        writer.flush()
       }
-    }
 
-    if (flushHeaders) {
-      writer.flush()
+      return stream
+    } catch (iioe: InterruptedIOException) {
```
yschimke, line 280: @swankjesse Alternatives here

a) If interrupted - don't call flush, instead write a RST_STREAM, and throw an InterruptedIOException. Keep connection healthy but generally make sure the 
b) Do this work via taskrunner so it can't be interrupted. Catch the interruption in say CountDownLatch in the caller thread, and then cancel the stream.

swankjesse (commented):
I don’t like having to handle interrupts throughout the code. It seems... arbitrary? 

And the testing strategy here also seems like it’s precisely targeting a specific rare situation... useful for regression testing but unsuited to directing us to put interrupt recovery in all the places we might need it.

We should figure out what our invariants are; perhaps this:
 * If the user thread doing a Call.execute() is interrupted, that call is canceled and resources are not leaked
 * If the dispatcher thread doing a Call.enqueue() is interrupted, that call is canceled and resources are not leaked
 * If any OkHttp-internal thread is interrupted... what do we do? Perhaps we should enter a minimum-level-of-service mode: stop populating the cache; stop pooling connections; etc. ? Ie. assume that OkHttp is being closed?

Or we could even even just cancel every Call after an OkHttp-internal thread is interrupted. I think our behavior needs to be undefined there because we can’t do our job if our threads are taken away from us.

swankjesse:
(I suspect our most interesting bug in this space is not degrading the cache when a Call thread is interrupted.)

yschimke:
OkHttp don't have good guarantees for the stack of the Okio Source, Sink after an interrupts.

> If the user thread doing a Call.execute() is interrupted, that call is canceled and resources are not leaked

This is tough, that call might have written to the Http/2 connection.  That's the dangerous bit.  I think in most other cases the  Interrupt is interrupting from a separate Buffer read, that can't break the Http/2 connection state.

> If any OkHttp-internal thread is interrupted.

I like this.

> I think our behavior needs to be undefined

That's the problem. This is primarily for a client (ExoPlayer) designed specifically to use Interrupts as a signal, 
that's a completely valid choice when they don't use us directly. There is an OkHttp adapter which introduces this.

I think we should define it, so that anything not matching that definition is a bug.

Or we state, interrupts are 100% *not* supported and we change the ExoPlayer adapter to use enqueue and a CountDown latch or similar.

Are most other interrupts sensibly handled?

Http/1.1 -> Interrupt kills the connection.
Http/2 -> after the stream is established, reads and writes are protected through per stream buffers?

swankjesse:
Version 2 of that comment is a doc!
https://github.com/square/okhttp/pull/7185/files

PR #7025, opened 2022-01-21
Okio Path examples

yschimke:
Evaluating https://twitter.com/RunIhorRun/status/1482138045960830984

> Do you have a good comprehensive example for integrating with Retrofit (OkHttp) to upload and download a file?

Seems like there isn't currently a clean way from Path to RequestBody


samples/guide/src/main/java/okhttp3/recipes/kt/PostPath.kt
```diff
@@ -0,0 +1,60 @@
...
+import okhttp3.OkHttpClient
+import okhttp3.Request
+import okhttp3.RequestBody.Companion.asRequestBody
+import okio.Path.Companion.toPath
+import okio.buffer
+import okio.fakefilesystem.FakeFileSystem
+
+class PostPath {
+  private val client = OkHttpClient()
+  private val fileSystem = FakeFileSystem()
```
swankjesse, line 29: This is the huge payoff. Fake files in tests for the win.

PR #7029, opened 2022-01-22
Tidy up ExchangeFinder to split pooling and reuse logic

swankjesse:
Each exchange has 3 places for a connection
 - the connection on the call (from a previous redirect, auth challenge, etc.)
 - a connection from the pool
 - a fresh connection

We query the pool after each blocking step to give ourselves the maximum
chance or connection reuse.

This changes code structure without changing behavior.


yschimke (approved):

swankjesse:

PR #7030, opened 2022-01-22
Move Connection.isHealthy() checks to where reuse happens

swankjesse:
One behavior change here is we won't fire connectionAcquired()
followed by connectionReleased() if we take an connection from the
pool only to immediately find it's unhealthy.

I'm working towards moving the loop over connect attempts and
fails from RetryAndFollowUpInterceptor and into ExchangeFinder.
(We'll still need some recovery in RetryAndFollowUpInterceptor,
but not connection recovery.)

### DIFF
--- a/okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/ExchangeFinder.kt
+++ b/okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/ExchangeFinder.kt
@@ -73,7 +73,7 @@ class ExchangeFinder(
     chain: RealInterceptorChain
   ): ExchangeCodec {
     try {
-      val resultConnection = findHealthyConnection(
+      val resultConnection = findConnection(
         connectTimeout = chain.connectTimeoutMillis,
         readTimeout = chain.readTimeoutMillis,
         writeTimeout = chain.writeTimeoutMillis,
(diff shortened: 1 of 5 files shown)


okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/ExchangeFinder.kt
```diff
@@ -92,57 +92,14 @@ class ExchangeFinder(
...
-  @Throws(IOException::class)
-  private fun findHealthyConnection(
-    connectTimeout: Int,
-    readTimeout: Int,
-    writeTimeout: Int,
-    pingIntervalMillis: Int,
-    connectionRetryEnabled: Boolean,
-    doExtensiveHealthChecks: Boolean
-  ): RealConnection {
-    while (true) {
```
swankjesse, line 107: Removing this loop FEELS GOOD. The previous behavior was that this would loop at most N + 1 times for N unhealthy pooled/reused connections and 1 new route to attempt. By not returning these until after the health check, we save the loop.

okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/ExchangeFinder.kt
```diff
@@ -208,14 +165,23 @@ class ExchangeFinder(
...
     val candidate = call.connection ?: return null
 
-    var toClose: Socket? = null
-    synchronized(candidate) {
-      if (candidate.noNewExchanges || !sameHostAndPort(candidate.route().address.url)) {
-        toClose = call.releaseConnectionNoEvents()
+    // Make sure this connection is healthy & eligible for new exchanges. If it's no longer needed
+    // then we're on the hook to close it.
+    val healthy = candidate.isHealthy(doExtensiveHealthChecks)
+    val toClose: Socket? = synchronized(candidate) {
```
swankjesse, line 175: We used to release the call's ownership of the connection when the route didn't match. Now we do it when the route didn’t match or the health check fails.

okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/ExchangeFinder.kt
```diff
@@ -259,6 +225,7 @@ class ExchangeFinder(
     }
 
     // List available IP addresses for the current proxy. This may block in Dns.lookup().
+    if (!newRouteSelector.hasNext()) throw IOException("exhausted all routes")
```
swankjesse, line 228: I'm not 100% convinced this is strictly necessary, but it’s better to fail this way than with a `NoSuchElementException`. (I hope to delete this later.)

okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/ExchangeFinder.kt
```diff
@@ -301,7 +270,6 @@ class ExchangeFinder(
   }
 
   fun trackFailure(e: IOException) {
-    nextRouteToTry = null
```
swankjesse, line 304: I don’t think this made any sense. We were clearing out a stashed route when an unrelated route failed. For example, if we coalesced connections this would clear out the original route that we coalesced from. (This could be the cause of our difficult-to-reproduce `exhausted all routes` errors.)

okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/ExchangeFinder.kt
```diff
@@ -355,6 +323,7 @@ class ExchangeFinder(
 
     synchronized(connection) {
       if (connection.routeFailureCount != 0) return null
+      if (!connection.noNewExchanges) return null // This route is still in use.
```
swankjesse, line 326: This replaces what we did in trackFailure(): rather than removing a duplicated route we never add it.

okhttp/src/jvmTest/java/okhttp3/CallTest.kt
```diff
@@ -3900,7 +3900,7 @@ open class CallTest(
       .build()
     executeSynchronously("/")
       .assertFailure(IOException::class.java)
-      .assertFailure("exhausted all routes")
+      .assertFailure("Socket closed", "Socket is closed")
```
swankjesse, line 3903: We no longer health check new connections.

swankjesse:
# Behavior Change

We no longer recover if there’s a single route and it fails with a refused_stream immediately. I think this is an improvement.

yschimke:
> One behavior change here is we won't fire connectionAcquired()
> followed by connectionReleased()

Was this ever clear why it was released? I'm not concerned about the events not being seen for a call in event listener, but overall with connection pool behaviour and soon happy eyeballs, it is maybe worth considering how we make this behavior visible.

yschimke (approved):
Change makes sense, but relying on passing test for ultimate review.  

Overall seems like a nice simplification of logic that got incrementally more complicated.

swankjesse:

PR #7031, opened 2022-01-23
Move the connect failure recovery loop

swankjesse:
Previously we did all IOException recovery in RetryAndFollowUpInterceptor.
With this PR we move recovering from connect failures into ExchangeFinder.
This allows us to delete the RouteException class as it's no longer
necessary.

The user-visible consequence of this change is we now segregate
suppressed exceptions into TCP layer and HTTP layer.

This also introduces a new Plan sealed class to represent the
two different ways a connection is found: reuse and connect.


okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/ExchangeFinder.kt
```diff
@@ -266,7 +237,63 @@ class ExchangeFinder(
...
+  /** Establish a new connection. */
+  internal inner class ConnectPlan(
+    route: Route,
+    val routes: List<Route>? = null,
+  ) : Plan() {
+    val connection = RealConnection(client.taskRunner, connectionPool, route)
+    var complete = false
+    var failure: IOException? = null
+
+    fun connect() {
```
swankjesse, line 259: This is all working towards making it so we can run this function in a background thread.

okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/RealCall.kt
```diff
@@ -115,7 +116,7 @@ class RealCall(
 
   @Volatile private var canceled = false
   @Volatile private var exchange: Exchange? = null
-  @Volatile var connectionToCancel: RealConnection? = null
+  internal val connectionsToCancel = CopyOnWriteArrayList<RealConnection>()
```
swankjesse, line 119: Expecting we'll soon support more than one of these at a time.

okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/ExchangeFinder.kt
yschimke: So for Happy Eyeballs, do you intend to have a HappyEyeballsConnectPlan? or multiple ConnectPlans for each possible route? Where do implementations of things like delays for additional results come in, or preferring routes?

okhttp/src/jvmTest/java/okhttp3/CallKotlinTest.kt
```diff
@@ -248,4 +249,26 @@ class CallKotlinTest(
...
+    client = client.newBuilder()
+        .dns(DoubleInetAddressDns()) // Two routes so we get two failures.
+        .build()
+
+    val request = Request.Builder().url(server.url("/")).build()
+    try {
+      client.newCall(request).execute()
+      fail("")
+    } catch (expected: IOException) {
+      expected.assertSuppressed {
```
yschimke, line 267: Nice!

okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/RouteException.kt
```diff
@@ -1,33 +0,0 @@
-/*
```
yschimke, line 1: There are a small number of references of this type in github, I know it's internal but is it worth deprecating instead of removing.  I'll leave call to you.  But you could make the argument that the exception types thrown from a public API are public API also.

okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/ExchangeFinder.kt
swankjesse: My best idea is we have a single thread, the call thread, act as coordinator.

Then all calls to Socket.connect() need to be on a TaskRunner.

The greatest challenge is that the coordinator needs to also do the DNS lookups, at least unless I extract that out.

okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/RouteException.kt
swankjesse: I believe that in all recent releases, this is thrown and caught always within OkHttp. End users who are catching it should be able to stop cause that is now dead code.

I don't like breaking binary compatibility, but this is in internal and we can't be paralyzed there.

swankjesse: Can you share your search results? I am bad at GitHub search! Everything I found was just decompiles of OkHttp itself.

yschimke: Let's do it then.

It's mostly as you describe https://github.com/search?p=1&q=%22import+okhttp3.internal.connection.RouteException%22&type=Code

https://github.com/NBXXF/xxf_android/blob/777162e3412bcf75ab1f41361d6fce240cb1952f/xxf/src/main/java/com/xxf/arch/rxjava/transformer/filter/ErrorIgnoreNetFilter.kt
https://github.com/nilam24/SmartPantry/blob/19368abfacd8288c155656c3eda053f9e36d772a/app/src/main/java/com/example/src/smartpantry/ApiClient.java

Nothing big, like the Google Maps SDK or similar.

okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/ExchangeFinder.kt
yschimke: Because of sync call or just code structure? We have a lot of better options for DNS. Android DnsResolver, OkHttp DoH, other JVM options. Dns.SYSTEM would be a waste for Happy Eyeballs.

swankjesse: Yeah, in this case it's just the concurrency that's tricky. We have a bunch of state on which routes we need to try and it's currently pretty gnarly.

yschimke (approved):

swankjesse:

PR #7033, opened 2022-01-25
Rename ExchangeFinder to RealRoutePlanner

swankjesse:
This introduces a new RoutePlanner interface that I'm intending
to use to make it possible to unit test all the interesting races
in happy eyeballs.

ExchangeFinder is the current sequential algorithm, and I think
I just need to create a new thing, HappyEyeballsExchangeFinder,
that implements a parallel search.

This commit is intended as a pure refactor with no behavior
changes.


oldergod (approved):

swankjesse:

PR #7035, opened 2022-01-26
HappyEyeballs

swankjesse:
This doesn't yet introduce any mechanism to enable or disable
happy eyeballs.

It also doesn't sort IP addresses to alternate IPv6, IPv4
for best success.

It also doesn't limit how many connections are attempted
simultaneously.

It also lacks an appropriate number of tests.

### DIFF
--- a/okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/RoutePlanner.kt
+++ b/okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/RoutePlanner.kt
@@ -53,11 +53,11 @@ interface RoutePlanner {
 
   fun trackFailure(e: IOException)
 
-  /**
-   * Returns true if the current route has a failure that retrying could fix, and that there's
-   * a route to retry on.
-   */
-  fun retryAfterFailure(): Boolean
+  /** Returns true if this planner has received any failures. */
+  fun hasFailure(): Boolean
+
+  /** Returns true if this planner has more routes to try. */
+  fun hasMoreRoutes(): Boolean
 
   /**
    * Returns true if the host and port are unchanged from when this was created. This is used to
@@ -78,5 +78,7 @@ interface RoutePlanner {
     fun connect()
 
     fun handleSuccess(): RealConnection
+
+    fun cancel()
   }
 }
--- a/okhttp-logging-interceptor/src/test/java/okhttp3/logging/LoggingEventListenerTest.java
+++ b/okhttp-logging-interceptor/src/test/java/okhttp3/logging/LoggingEventListenerTest.java
@@ -17,20 +17,21 @@ package okhttp3.logging;
 
 import java.io.IOException;
 import java.net.UnknownHostException;
+import mockwebserver3.MockResponse;
+import mockwebserver3.MockWebServer;
+import mockwebserver3.SocketPolicy;
 import mockwebserver3.junit5.internal.MockWebServerExtension;
 import okhttp3.Call;
 import okhttp3.EventListener;
 import okhttp3.HttpUrl;
 import okhttp3.MediaType;
 import okhttp3.OkHttpClient;
-import okhttp3.TestUtil;
-import okhttp3.testing.PlatformRule;
+import okhttp3.OkHttpClientTestRule;
 import okhttp3.Request;
 import okhttp3.RequestBody;
 import okhttp3.Response;
-import mockwebserver3.MockResponse;
-import mockwebserver3.MockWebServer;
-import mockwebserver3.SocketPolicy;
+import okhttp3.TestUtil;
+import okhttp3.testing.PlatformRule;
 import okhttp3.tls.HandshakeCertificates;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
@@ -49,6 +50,7 @@ public final class LoggingEventListenerTest {
   private static final MediaType PLAIN = MediaType.get("text/plain");
 
   @RegisterExtension public final PlatformRule platform = new PlatformRule();
+  @RegisterExtension public final OkHttpClientTestRule clientTestRule = new OkHttpClientTestRule();
   private MockWebServer server;
 
   private final HandshakeCertificates handshakeCertificates = localhost();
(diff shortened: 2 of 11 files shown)


swankjesse:
# This pull request is not appropriate to land.

okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/RoutePlanner.kt
```diff
@@ -53,11 +53,11 @@ interface RoutePlanner {
 
   fun trackFailure(e: IOException)
 
-  /**
-   * Returns true if the current route has a failure that retrying could fix, and that there's
-   * a route to retry on.
-   */
-  fun retryAfterFailure(): Boolean
+  /** Returns true if the current route has a failure that retrying could fix. */
```
yschimke, line 56: hasRetryableFailure based on the comment?

okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/RoutePlanner.kt
```diff
@@ -53,11 +53,11 @@ interface RoutePlanner {
...
+  fun hasFailure(): Boolean
+
+  /** Returns true if the there's a route to retry on. */
```
yschimke, line 59: awkward wording.

okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/HappyEyeballsExchangeFinder.kt
```diff
@@ -0,0 +1,135 @@
...
+    if (exception != null) {
+      trackFailure(exception)
+      return null
+    }
+
+    return completed.plan.handleSuccess()
+  }
+
+  private fun trackFailure(exception: IOException) {
+    routePlanner.trackFailure(exception)
```
yschimke, line 122: Not needed at the moment, since it's only a counter, but should trackFailure be told the route also?

yschimke: https://datatracker.ietf.org/doc/html/rfc6555#section-4.2 Talks about stateful implementations

yschimke: Although changing mobile networks makes this complicated, would only apply until the network changes.

okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/HappyEyeballsExchangeFinder.kt
```diff
@@ -0,0 +1,135 @@
...
+    connectsInFlight += plan
+
+    // Already connected? Enqueue the result immediately.
+    if (plan.isConnected) {
+      connectResults.put(ConnectResult(plan, null))
+      return
+    }
+
+    // Connect asynchronously.
+    val taskName = "$okHttpName connect ${routePlanner.address.url.redact()}"
```
yschimke, line 91: is host better here than redacted url?

swankjesse: As is here what I have isn't good enough. I need more Route information, specifically the destination IP. A thread dump show what’s connecting including both hostname and IP address.

```
OkHttp connect cash.app 1.2.3.4
OkHttp connect cash.app 1234:5678::9abc
```

swankjesse: And maybe more if there's a proxy...

```
OkHttp connect cash.app 1.2.3.4
OkHttp connect cash.app 1234:5678::9abc
OkHttp connect cash.app via proxy proxy.isp.net 1234:5678::9abc
```

okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/RoutePlanner.kt
swankjesse: Fixed:
```
  /** Returns true if this planner has received any failures. */
```

okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/RoutePlanner.kt
swankjesse: Fixed:
```
  /** Returns true if this planner has more routes to try. */
```

okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/HappyEyeballsExchangeFinder.kt
swankjesse: I’ll add it when we need it!

okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/HappyEyeballsExchangeFinder.kt
swankjesse: https://github.com/square/okhttp/pull/7037

yschimke:
> # This pull request is not appropriate to land.

```
@github-actions
build / checkjdk11 (pull_request) Cancelled after 360m — checkjdk11 
```

Shipit! :)

swankjesse:
Yeah, the 360m thing was a nice reminder of how thorough our tests are.
We throw a deliberate uncaught exception in URLConnectionTest.connectHttpsViaSocketFactory(). I fixed it by relaying the uncaught exception back to the ExchangeFinder!

yschimke:
@swankjesse Does it help to have some async + IP filtered Dns implementations for this? I can add a callback API to DoH, and trivially support and Android async implementation. 

Just wondering if there is anything I can contribute to here?

Or is first target, not a full happy eyeballs, just enough to make things work with broken IPv6?

swankjesse:
You got it. First target is getting things working with broken IPv6. If you're up for it, we need a function to reorder a list of InetAddress objects according to RFC 8305 section 4.

https://datatracker.ietf.org/doc/html/rfc8305#section-4

There's a lot of stuff there but I think to start all we want to do is interleave IPv6 and IPv4 addresses:
 * starting with IPv6
 * preserving order otherwise

Given this input:
```
10.0.0.6
10.0.0.1
10.0.0.4
::ab
::fc
```

And produces this output:
```
::ab
10.0.0.6
::fc
10.0.0.1
10.0.0.4
```

If you wanna take a stab at this, please do!

yschimke:
> If you wanna take a stab at this, please do!

On it.

swankjesse:
@yschimke this is ready to land! I’ll create a checklist on the tracking bug for remaining follow-ups.

yschimke (approved):

swankjesse:

PR #7037, opened 2022-01-27
Create a nice toString() for Route

swankjesse:
I'm going to use this to name threads when connecting via
Happy Eyeballs.

### DIFF
--- a/okhttp/src/jvmMain/kotlin/okhttp3/Route.kt
+++ b/okhttp/src/jvmMain/kotlin/okhttp3/Route.kt
@@ -17,17 +17,16 @@ package okhttp3
 
 import java.net.InetSocketAddress
 import java.net.Proxy
+import okhttp3.internal.toCanonicalHost
 
 /**
  * The concrete route used by a connection to reach an abstract origin server. When creating a
  * connection the client has many options:
  *
- *  * **HTTP proxy:** a proxy server may be explicitly configured for the client.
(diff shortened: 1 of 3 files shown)


okhttp/src/jvmMain/kotlin/okhttp3/Route.kt
```diff
@@ -87,5 +86,42 @@ class Route(
...
-  override fun toString(): String = "Route{$socketAddress}"
+  /**
+   * Returns a string with the URL hostname, socket IP address, and socket port, like one of these:
+   *
+   *  * `example.com:80 at 1.2.3.4:8888`
+   *  * `example.com:443 via proxy [::1]:8888`
+   *
+   * This omits duplicate information when possible.
+   */
+  override fun toString(): String {
```
yschimke, line 97: Do we use this at least once, maybe more. Is it worth making this lazy and avoid doing twice?!

yschimke (approved):

okhttp/src/jvmMain/kotlin/okhttp3/Route.kt
swankjesse: We're currently using it zero times. With happy eyeballs we'll use it once to name a task that ultimately names a thread.

I don't think it benefits from caching in our current use cases. Caching stuff like this is most useful if we use the string form programmatically, like if we were using the string as a HashMap key.

Also, I reserve the right to change my mind on this if YourKit tells me to!

swankjesse:

PR #7038, opened 2022-01-28
Create TestValueFactory to reduce test case boilerplate

swankjesse:
This is bad because it moves test logic out of tests. But it's also
good because it lowers the cost of writing unit tests with common
types like RealConnection and Route.


okhttp/src/jvmTest/java/okhttp3/internal/RecordingOkAuthenticator.java
```diff
@@ -1,73 +0,0 @@
...
-import java.util.ArrayList;
-import java.util.List;
-import javax.annotation.Nullable;
-import okhttp3.Authenticator;
-import okhttp3.Challenge;
-import okhttp3.Request;
-import okhttp3.Response;
-import okhttp3.Route;
-
-public final class RecordingOkAuthenticator implements Authenticator {
```
swankjesse, line 28: Moved to okhttp-testing-support and converted to Kotlin.

yschimke (approved):
Not sure I agree about the downside, fake objects are a well known pattern, and usually best described outside of the test using consistent predictable logic.

swankjesse:

PR #7040, opened 2022-01-29
First basic version of InetAddress ordering

yschimke:


yschimke:
Lot of questions. But if goal is just to get first two requests on ipv6 and ipv4, then it's a minimal implementation.

Why did you select ipv6 in 3rd place?

```
::ab
10.0.0.6
::fc
```

- How to wire into FastFallbackExchangeFinder but not ExchangeFinder (to preserve legacy behaviour)
- API to dns collection and ordering? Sequence?
- Should this be Stateful service? Lot of rules in the spec regarding remembering, but specifically only on same network.
- How much of the advanced priorities to implement?

swankjesse:
I’m looking at section 4 from RFC 8305

> Next, the client SHOULD modify the ordered list to interleave address
> families.  Whichever address family is first in the list should be
> followed by an address of the other address family; that is, if the
> first address in the sorted list is IPv6, then the first IPv4 address
> should be moved up in the list to be second in the list.  An
> implementation MAY want to favor one address family more by allowing
> multiple addresses of that family to be attempted before trying the
> other family.  The number of contiguous addresses of the first
> address family will be referred to as the "First Address Family
> Count" and can be a configurable value.  This is performed to avoid
> waiting through a long list of addresses from a given address family
> if connectivity over that address family is impaired.

My interpretation is we write a function that follows these rules:

 * While we only have addresses from one family, proceed through those in order
 * Once we have addresses from both families, interleave!
 * Prefer IPv6

swankjesse:
I think we change `RouteSelector` to only order the addresses if `client.fastFallback` is set. That’s most conservative for upgrading from existing systems because you have to opt-in to get the behavior change.

swankjesse:
I think the API signature you have now is perfect for our current needs. List in, List out. 

If later we get APIs to do IPv4 and IPv6 independently and asynchronously, I think the API we want probably looks more complicated:

```
  /**
   * Returns a new list that combines two lists of IP addresses, in the order
   * they should be connected to.
   * 
   * The result implements these rules in precedence order:
   * 
   *   * The result contains all elements of [alreadyKnown] and [newlyDiscovered].
   *   * Elements in [alreadyKnown] in `[0..processedIndex)` retain their existing
   *     position. These elements represent addresses we've already initiated
   *     connections to and we can’t reprioritize past decisions
   *   * address families are interleaved
   *   * IPv6 is preferred and precedes IPv4
   *   * addresses retain their original order
   *
   * When two rules disagree, the earlier rule wins.
   */
  fun combineAndOrder(
    alreadyKnown: List<InetAddress>,
    processedIndex: Int,
    newlyDiscovered: List<InetAddress>)
  } : List<InetAddress>
```

swankjesse:
I don’t think we have the data to implement any of the advanced priorities yet. Lets ship concurrent attempts first!

okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/RouteSelector.kt
```diff
@@ -160,13 +160,15 @@ class RouteSelector(
...
       val addresses = address.dns.lookup(socketHost)
       if (addresses.isEmpty()) {
         throw UnknownHostException("${address.dns} returned no addresses for $socketHost")
       }
 
-      eventListener.dnsEnd(call, socketHost, addresses)
+      // Try each address for best behavior in mixed IPv4/IPv6 environments.
+      val prioritisedAddresses = InetAddressOrder.reorder(addresses)
+
+      eventListener.dnsEnd(call, socketHost, prioritisedAddresses)
```
swankjesse, line 171: I think we should notify the EventListener with exactly what the network returned? Seems more useful in debugging than returning our processed result?

swankjesse (approved):
This is wonderful

yschimke:
> This is wonderful

This would be the easiest interview question ever, interleave two lists, so your enthusiasm is appreciated but surprising...

swankjesse:

Issue #7041, opened 2022-01-30
Kotlin code sample missing for Authenticator in Recipes section


ColtonIdle:
Kotlin:
<img width="441" alt="2022-01-30 at 01 22 45" src="https://user-images.githubusercontent.com/31751141/151689138-3b083714-42bf-4495-b956-37d102b9fcce.png">

Java:
<img width="442" alt="2022-01-30 at 01 22 37" src="https://user-images.githubusercontent.com/31751141/151689128-b04f67b7-4031-445f-a3e2-7d5011d9e7e9.png">

ColtonIdle:
Thanks for the quick response @yschimke 

For what it's worth, I tried copying the kotlin code today for making a sync and async request and those have a few issues. It looks like the apis were updated. Not sure if it's worth opening up another issue, but it'd be nice to make sure those samples compile.

yschimke:
@ColtonIdle not sure what the issues with those examples are, they worked fine for me.  Could you grab a screenshot in your IDE?

PR #7043, opened 2022-01-30
Fix Authenticate docs to include more kotlin

yschimke:
The Content Tab feature means that only java had any code samples for responseCount.


yschimke:
@swankjesse Bit worried about the effect of the whitespace changes.  Is there an easy way to test this in CI?

yschimke:
To verify https://github.com/square/okhttp/issues/7041#issuecomment-1027036179

PR #7046, opened 2022-01-31
Rewrite TaskFaker to permit mutiple stalled tasks

swankjesse:
In order to deterministically test FastFallbackExchangeFinder I need
a test facet that supports several test coordinating:
 - the call thread is waiting for either a connection to connect
   or for the next-connection timeout (250 ms) to elapse
 - the in-flight connection needs to connect

TaskFaker wasn't up to the task for this because it assumed at most
two threads, the test thread and the task thread.

With this change TaskFaker can coordinate multiple task threads
that each run sequentially, as directed by a test thread.

### DIFF
--- a/okhttp-testing-support/src/main/kotlin/okhttp3/internal/concurrent/TaskFaker.kt
+++ b/okhttp-testing-support/src/main/kotlin/okhttp3/internal/concurrent/TaskFaker.kt
@@ -16,14 +16,28 @@
 package okhttp3.internal.concurrent
 
 import java.util.concurrent.Executors
+import java.util.concurrent.Semaphore
+import java.util.concurrent.atomic.AtomicBoolean
 import java.util.logging.Logger
 import okhttp3.OkHttpClient
 import org.assertj.core.api.Assertions.assertThat
 
 /**
  * Runs a [TaskRunner] in a controlled environment so that everything is sequential and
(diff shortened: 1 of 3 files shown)


yschimke (approved):
Not sure I have much signal on this review.  You may want to request other reviews.  Looks good to me, and it's passing and used in some key tests, so I have confidence in it.

swankjesse:
Thanks for the review!

This is the 3rd attempt at this, and the first two failed catastrophically ’cause my design didn’t fit the problem.

 * In the first design I wanted to use a queue to keep track of what was stalled; that didn’t work because I could get into infinite loops draining the queue for tasks not ready to run. I’d have needed a marker and it would have gotten nasty.
 * Attempt two was like this, but I used `ReentrantLock` instead of Java’s built in `synchronized` ’cause I wanted the locking to be fair. That would have worked if I was willing to change the production code to use `Lock` instead of `synchronized` also, but that seemed like a riskier change.

In those attempts I got nowhere near tests running well. They’d crash or deadlock or some combination of both.

With this the design fits the tests’ expectations and so most of the tests passed on the first shot. I had to add the `acquiredTaskRunnerLock` thing which was initially broken, but that was mostly it.

So having tried and failed a few times I’ve gained confidence that our tests are extremely thorough here, and that gives me confidence that this works. Let’s see if I’m still feeling that way after using it to test `FastFallbackExchangeFinder`.

swankjesse:

PR #7047, opened 2022-02-01
Tests for FastFallbackExchangeFinder

swankjesse:
This needed more test facets in TaskFaker.

I'm pretty happy with how the tests read.

### DIFF
--- a/okhttp/src/jvmTest/java/okhttp3/internal/concurrent/TaskRunnerTest.kt
+++ b/okhttp/src/jvmTest/java/okhttp3/internal/concurrent/TaskRunnerTest.kt
@@ -402,16 +402,15 @@ class TaskRunnerTest {
     assertThat(log).isEmpty()
 
     taskFaker.advanceUntil(100.µs)
-    assertThat(log).containsExactly(
+    assertThat(log).containsExactlyInAnyOrder(
         "one:run@100000 parallel=true",
         "two:run@100000 parallel=true",
         "three:run@100000 parallel=true"
     )
 
-    taskFaker.runTasks()
     taskFaker.assertNoMoreTasks()
 
-    assertThat(testLogHandler.takeAll()).containsExactly(
+    assertThat(testLogHandler.takeAll()).containsExactlyInAnyOrder(
         "FINE: Q10000 scheduled after 100 µs: task one",
         "FINE: Q10001 scheduled after 100 µs: task two",
         "FINE: Q10002 scheduled after 100 µs: task three",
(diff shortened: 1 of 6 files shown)


okhttp/src/jvmTest/java/okhttp3/internal/concurrent/TaskRunnerTest.kt
```diff
@@ -396,16 +396,15 @@ class TaskRunnerTest {
...
 
     taskFaker.advanceUntil(100.µs)
-    assertThat(log).containsExactly(
+    assertThat(log).containsExactlyInAnyOrder(
         "one:run@100000 parallel=true",
         "two:run@100000 parallel=true",
         "three:run@100000 parallel=true"
     )
 
-    taskFaker.runTasks()
```
swankjesse, line 405: I was able to get rid of this by introducing `TaskFaker.isRunningAllTasks`. This call was only introduced today and I wasn’t happy with it!

swankjesse (commented):
For follow-up, tests still to write:

 * [x] Multiple calls into the exchange finder
 * [x] Connect fails with an exception
 * [x] Exhaust the exchange finder
 * [ ] Failures tracked

yschimke:
Concerning build failure

```

EventListenerTest[jvm] > requestBodyMultipleFailuresReportedOnlyOnce()[jvm] FAILED
    java.lang.AssertionError: uncaught exception thrown during test
        at okhttp3.OkHttpClientTestRule.afterEach(OkHttpClientTestRule.kt:202)
...
        Caused by:
        java.lang.OutOfMemoryError: Java heap space
            at okio.Segment.<init>(Segment.kt:62)
            at okio.SegmentPool.take(SegmentPool.kt:88)
            at okio.Buffer.writableSegment$okio(Buffer.kt:1444)
            at okio.InputStreamSource.read(JvmOkio.kt:92)
```

okhttp/src/jvmTest/java/okhttp3/internal/connection/FastFallbackExchangeFinderTest.kt
```diff
@@ -15,25 +15,316 @@
...
+    assertThat(takeEvent()).isEqualTo("plan 0 connecting...")
+    assertThat(takeEvent()).isNull()
+
+    taskFaker.advanceUntil(240.ms)
+    assertThat(takeEvent()).isEqualTo("plan 0 connected")
+    assertThat(takeEvent()).isNull()
+  }
+
+  @Test
+  fun firstPlanConnectedBeforeSecondPlan() {
```
yschimke, line 78: These tests are really nice, and show the logic really well.

yschimke (approved):
Low risk change, 1 line to change to non test code.

swankjesse:
I looked into the build failure. When I redid TaskFaker I caused threads to be retained when they weren’t previously. In particular anything using TaskFaker with the disk cache was permanently leaking a Thread because we had a cleanup task queued but never runnable.

I added a commit to this PR which was also suffering leaks in tests:
https://github.com/square/okhttp/pull/7048

swankjesse:

PR #7054, opened 2022-02-05
[testing] Exploration of Async DNS on Android and also DoH

yschimke:
Not any sort of suggested or refined API.  But checking out the options for AsyncDns on both Android (DnsResolver on API 29+) and JVM via DoH.

### DIFF
--- a/okhttp/src/jvmMain/kotlin/okhttp3/internal/platform/android/AndroidDns.kt
+++ b/okhttp/src/jvmMain/kotlin/okhttp3/internal/platform/android/AndroidDns.kt
@@ -1,0 +1,69 @@
+/*
+ * Copyright (C) 2022 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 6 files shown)


okhttp/src/jvmMain/kotlin/okhttp3/internal/platform/android/AndroidDns.kt
```diff
@@ -0,0 +1,69 @@
...
+
+@SuppressLint("NewApi") // API 29!
+class AndroidDns : AsyncDns {
+  val resolver = DnsResolver.getInstance()
+  val executor = Executors.newSingleThreadExecutor()
+
+  override fun query(hostname: String, callback: AsyncDns.Callback) {
+    val executing = AtomicInteger(2)
+
+    resolver.query(
```
yschimke, line 34: Split by type, otherwise the results are nicely interleaved.

> Send a DNS query with the specified name and query type, get back a set of InetAddresses with rfc6724 sorting style asynchronously. The answer will be provided asynchronously through the provided [Callback](https://developer.android.com/reference/android/net/DnsResolver.Callback).

yschimke: This is the first part of Happy Eyeballs ordering - which is helpfully useful.

> First, the client MUST sort the addresses received up to this point using Destination Address Selection ([[RFC6724], Section 6](https://datatracker.ietf.org/doc/html/rfc6724#section-6)).

yschimke:
@swankjesse Hey, I don't suggest changing your design for a future RFC (HTTPS records), but maybe worth considering how that would fit in, if we want to use the IPv4 and IPv6 address hints, or ALPN, HSTS, or AltSvc flags from the DNS phase.

yschimke:
Half of this is now in https://github.com/square/okhttp/pull/7157

PR #7057, opened 2022-02-06
Fix typo in event ordering.

yschimke:
If handshake fails we won't have a secureConnectEnd.


yschimke:
@swankjesse post review

oldergod (commented):
LGTM

Issue #7045, opened 2022-01-31
OkHttp does not close connection when server sends a FIN, ACK


tejasjadhav:
We've recently encountered an issue on production wherein an HTTP server, which is shutting down, sends back a `FIN, ACK` packet, but OkHttp still continues sending traffic on that connection instead of closing it and started a new connection.

Here is a screenshot of one such incident on Wireshark on my local setup (`10.101.84.85` is the server, `172.17.0.6` is the client),
![Screenshot 2022-01-31 at 6 27 11 PM](https://user-images.githubusercontent.com/10093001/151797917-9896ce18-f5c7-45f4-86fb-4e46cfaefee9.png)

My test setup is as follows,
* Hosted on Kubernetes.
* Client side running using OkHttp 4.5.0 (Java 18)
* Server side running a Golang HTTP server (Golang 1.18)
* Client side uses Kubernetes FQDN to talk to the server.
* Client maintains a connection pool of 10 connections.
* Client side code: https://gist.github.com/tejasjadhav/d5a4b4efca8e7400236ce18e86dcb00a#file-app-java
* Server side code: https://gist.github.com/tejasjadhav/d5a4b4efca8e7400236ce18e86dcb00a#file-main-go

### Expectation
OkHttp client should remove the connection from the pool when server sends back `FIN, ACK` packet.

### Actual
OkHttp does not terminate the connection for which server sends back `FIN, ACK` packet and still sends traffic on that connection, which results in a connection resets and failures on subsequent calls.

### Steps to reproduce
1. Create an OkHttp client with a connection pool to an HTTP server.
2. Make 1-2 requests using that client to initialise the pool.
3. Shutdown the server. Observe that `FIN, ACK` packets are being sent by the server on shutdown.
4. While the server is shutting down, send 1-2 requests using the OkHttp client.
5. The requests should fail with an error `unexpected end of stream`. Server would have sent an `RST` in response.

### Environment
* OkHttp 4.5.0
* Java 18

yschimke:
Any chance you can test with 4.9.3? We have had various fixes in this area, but not specifically aware of anything that would fix it.

tejasjadhav:
@yschimke tried it on OkHttp 4.9.3. Still seeing the same behavior

![Screenshot 2022-01-31 at 9 44 44 PM 1](https://user-images.githubusercontent.com/10093001/151830823-579148a3-7dcf-4733-ba26-ef7f34172ebd.png)

yschimke:
[...]

PR #7058, opened 2022-02-06
Test to verify half close behaviour

yschimke:
Verifying https://github.com/square/okhttp/issues/7045


yschimke:
@swankjesse test only - verifies that we act correctly on a half closed socket. A bit redundant, until it's not...

PR #7063, opened 2022-02-06
Remove external link icon

### LINKED ISSUE #7060: Remove external link icon in website
The indicator for external sites makes the docs look cluttered, and it is broken in light mode. Let’s figure out how to remove it?

Screen Shot 2022-02-06 at 12 38 14
Screen Shot 2022-02-06 at 12 38 10

Andre601:
Closes #7060 

Removes the CSS responsible for adding the External link icon to URLs pointing away from the Docs.

I initially tried to alter the CSS in such a way that a differently colored icon would be used depending on the color scheme, but this was impossible with my basic skills, so I went ahead and removed it instead.


swankjesse:
Thanks for this.

PR #7065, opened 2022-02-08
Move code from RealConnection to RealConnectPlan

swankjesse:
Also promote RealConnectPlan to a top-level type, ConnectPlan.

Each RealConnection is now created only once its ready to be used
to carry exchanges. We do the actual connect work in ConnectPlan.


yschimke:
Tests taking 4 hours suggests an issue.

Overall looks like a sensible refactoring of RealConnection.

swankjesse:
The first timeouts were a basic blunder; I didn't hand off the HTTP2 connection object from ConnectPlan to RealCinnection.

The next problem was I didn't fix FastFallbackExchangeFinder to cycle through the retry plans when those were returned.

With those fixed, tests pass.

I saw some tests flake but I think it's a test fragility problem and not a correctness problem. In particular FastFallbackExchangeFinderTest assumes IPv6 will complete first, which is likely but not guaranteed. I don't think this is a new flake in this PR. But it'll need some work.

Next step is splitting ConnectPlan into two steps: TCP handshake and then everything else. With that I should be able to promote the TCP handshake to be the only subject of the race.

oldergod (approved):

swankjesse:

PR #7073, opened 2022-02-11
fix DoH error when using ip address as hostname

TaoZang:
Fix the issue that when network trying to do a DNS lookup with an IP numeric hostname, the DnsOverHttps will throw an UnknownHostException instead of returning the InetAddress with this IP address.
For example, If user runs a http proxy, the DnsOverHttps module will try to lookup the local Proxy address such as 127.0.0.x.


okhttp-dnsoverhttps/src/main/kotlin/okhttp3/dnsoverhttps/DnsOverHttps.kt
```diff
@@ -67,7 +67,11 @@ class DnsOverHttps internal constructor(
       }
     }
 
-    return lookupHttps(hostname)
+    return try {
+      listOf(InetAddressUtil.forString(hostname))
```
yschimke, line 71: Can you add a test?  

We need to avoid an exception in most calls here.  Maybe reverse this logic (try hostname first) and in the normal case call lookupHttps, and catch and exception, check if it's a ip address and then return forString.

okhttp-dnsoverhttps/src/main/kotlin/okhttp3/dnsoverhttps/InetAddressUtil.kt
```diff
@@ -0,0 +1,130 @@
...
+package okhttp3.dnsoverhttps
+
+import java.lang.IllegalArgumentException
+import java.lang.NumberFormatException
+import java.lang.StringBuilder
+import java.net.InetAddress
+import java.net.UnknownHostException
+import java.util.*
+
+object InetAddressUtil {
```
yschimke, line 25: This looks like a lot of supporting code.  Are there existing methods that do this, or would a simple regex as a first parse be simpler (\d+\.\d+...) or (\d+:...)?

okhttp-dnsoverhttps/src/main/kotlin/okhttp3/dnsoverhttps/DnsOverHttps.kt
TaoZang: I think it would be better if we try converting by forString first because we can avoid a potential https connection if the hostname is an numerical IP address.

okhttp-dnsoverhttps/src/main/kotlin/okhttp3/dnsoverhttps/InetAddressUtil.kt
TaoZang: I implement the InetAddressUtil kotlin version refer to the com.google.common.net.InetAddresses class. I just concern about that a simple regex may cause some bugs due to the complex format of IPv4/IPv6. A common parse algorithm would be better.

okhttp-dnsoverhttps/src/main/kotlin/okhttp3/dnsoverhttps/DnsOverHttps.kt
TaoZang: If most of the calls will be hostname, maybe call the lookupHttps first would be better. I will update the code.

TaoZang:
> I'll backport this to 4.9.x when we have a fix, but until that happens you might want to have your own Dns wrapper that does this logic yourself.
> 
> Thanks for the report and PR.

Thank you for your review:)

okhttp-dnsoverhttps/src/main/kotlin/okhttp3/dnsoverhttps/InetAddressUtil.kt
TaoZang: I also updated the copyright just same as the raw implement file.

yschimke:
Looking at the logic of InetAddress.getAllByName, would the following be simpler?

```
  override fun lookup(hostname: String): List<InetAddress> {
    if (isIpAddress(hostname)) {
      return listOf(InetAddress.getByName(hostname))
    }
    ...
  private fun isIpAddress(hostname: String): Boolean {
    return hostname[0].digitToIntOrNull(16) != null
      || hostname[0] == ':' || hostname[0] == '['
  }
```

TaoZang:
> Looking at the logic of InetAddress.getAllByName, would the following be simpler?
> 
> ```
>   override fun lookup(hostname: String): List<InetAddress> {
>     if (isIpAddress(hostname)) {
>       return listOf(InetAddress.getByName(hostname))
>     }
>     ...
>   private fun isIpAddress(hostname: String): Boolean {
>     return hostname[0].digitToIntOrNull(16) != null
>       || hostname[0] == ':' || hostname[0] == '['
>   }
> ```
> 
> getAllByName shortcircuits and avoids Dns if the hostname is an ip address. Using similar logic in isIpAddress.

It seems that the getAllByName function also needs some support functions such as textToNumericFormatV4/textToNumericFormatV6 from sun.net.util.IPAddressUtil class, I thought it might be btter to avoid import a class from sun.net package here.

TaoZang:
> Please add some tests to DnsOverHttpsTest.

I just added some tests in DnsOverHttpsText.

TaoZang:
I just removed the complex support class. 
Because the getAllName will shortcircuits and directly convert the hostname to InetAddress, so we only check the failed looking-up hostname when it is possible a valid IP address and convert it to InetAddress by getAllName.

okhttp-dnsoverhttps/src/test/java/okhttp3/dnsoverhttps/DnsOverHttpsTest.java
```diff
@@ -84,6 +84,18 @@
...
   }
 
+  @Test public void getIPAddressHost() throws Exception {
+    dns = buildLocalhost(bootstrapClient, true);
+
+    String hostIPv4 = "127.0.0.1";
+    List<InetAddress> resultIPv4 = dns.lookup(hostIPv4);
+    assertThat(resultIPv4).contains(address(hostIPv4));
+
+    String hostIPv6 = "2a03:2880:f029:11:face:b00c:0:2";
```
yschimke, line 94: Can you add tests for [::1] and ::1

okhttp-dnsoverhttps/src/main/kotlin/okhttp3/dnsoverhttps/DnsOverHttps.kt
```diff
@@ -67,7 +69,31 @@ class DnsOverHttps internal constructor(
...
+        } catch (e: UnknownHostException) {
+          throw bestFailure
+        }
+      } else {
+        throw bestFailure
+      }
+    }
+  }
+
+  private fun checkIsIpAddress(hostname: String): Boolean {
```
swankjesse, line 87: Should use our existing function, `canParseAsIpAddress()` from `okhttp3.internal`. We definitely don’t want to functions to convert IP addresses to strings.

okhttp-dnsoverhttps/src/main/kotlin/okhttp3/dnsoverhttps/DnsOverHttps.kt
```diff
@@ -67,7 +69,31 @@ class DnsOverHttps internal constructor(
...
+    if (ipv4Regex.matches(hostname)) {
+      return true
+    }
+
+    if (hostname[0].digitToIntOrNull(16) != null || hostname[0] == ':' || hostname[0] == '[') {
```
swankjesse, line 92: If the hostname is `[`, that’s malformed. We do not pass the square braces to `Dns`.

swankjesse (commented):
I’m unconvinced that this is the right layer for this fix. I think it’d be better if we never call `Dns.lookup()` when we have an IP address.

okhttp-dnsoverhttps/src/test/java/okhttp3/dnsoverhttps/DnsOverHttpsTest.java
TaoZang: Sure, just addes test cases for these addresses.

okhttp-dnsoverhttps/src/main/kotlin/okhttp3/dnsoverhttps/DnsOverHttps.kt
TaoZang: Sure, I missed that there is a existing function.

okhttp-dnsoverhttps/src/main/kotlin/okhttp3/dnsoverhttps/DnsOverHttps.kt
TaoZang: I thought that the getAllByName will strip the square braces so I just pass the string to getAllByName directly.
Due to the canParseAsIpAddress will not handle `[`, I will strip the square braces and use the existing one to check if the hostname is a IP address.

okhttp-dnsoverhttps/src/main/kotlin/okhttp3/dnsoverhttps/DnsOverHttps.kt
```diff
@@ -67,7 +69,24 @@ class DnsOverHttps internal constructor(
...
-    return lookupHttps(hostname)
+    return try {
+      lookupHttps(hostname)
+    } catch (bestFailure: UnknownHostException) {
+      val address = if (hostname.startsWith("[") && hostname.endsWith("]")) {
+        hostname.substring(1, hostname.length - 1)
+      } else {
+        hostname
+      }
+      if (address.canParseAsIpAddress()) {
```
yschimke, line 80: Nice find, I forgot this was a thing.  Can you invert this now?

Also if that function doesn't worry about [], we can probably skip that here also.

See https://github.com/square/okhttp/blob/3ad1912f783e108b3d0ad2c4a5b1b89b827e4db9/okhttp/src/jvmTest/java/okhttp3/HttpUrlTest.java#L514

Which confirms the HttpUrl strips the [].

TaoZang: Do you mean that the HttpUrl will handle the square braces before calling dns lookup, so we don't need to check and strips the `[]` here and just remove the test case for hostname "[::1]"?

swankjesse:
At a risk of repeating myself, what happens if we never call any Dns implementation when the input is already an IP address?

yschimke:
> At a risk of repeating myself, what happens if we never call any Dns implementation when the input is already an IP address?

Yep - this makes sense. @TaoZang sorry to give you bad advice, and thanks for you patience.  What do you think about make this change to avoid using Dns for IP Addresses.

TaoZang:
> Yep - this makes sense. @TaoZang sorry to give you bad advice, and thanks for you patience. What do you think about make this change to avoid using Dns for IP Addresses.



> At a risk of repeating myself, what happens if we never call any Dns implementation when the input is already an IP address?

Yeah, maybe I just misunderstood your meaning before. Avoid to call DNS lookup in higher layer when the input is already an IP addres seems be make more sense. I will try to change the fix by this way.

okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/RouteSelector.kt
```diff
@@ -159,15 +161,20 @@ class RouteSelector(
     if (proxy.type() == Proxy.Type.SOCKS) {
       mutableInetSocketAddresses += InetSocketAddress.createUnresolved(socketHost, socketPort)
     } else {
-      eventListener.dnsStart(call, socketHost)
+      val addresses = if (socketHost.canParseAsIpAddress()) {
+        listOf(InetAddress.getByName(socketHost))
+      } else {
+        eventListener.dnsStart(call, socketHost)
```
yschimke, line 167: This seems like a nice benefit of this change.

okhttp-dnsoverhttps/src/main/kotlin/okhttp3/dnsoverhttps/DnsOverHttps.kt
```diff
@@ -53,6 +53,7 @@ class DnsOverHttps internal constructor(
   @get:JvmName("resolvePrivateAddresses") val resolvePrivateAddresses: Boolean,
   @get:JvmName("resolvePublicAddresses") val resolvePublicAddresses: Boolean
 ) : Dns {
+
```
yschimke, line 56: nit: Can you revert this line.

yschimke (approved):

yschimke (changes requested):
Please add a test that confirms this behaviour.  Something in EventListenerTest like

```


  @Test public void successfulCallEventSequenceForIpAddress() throws IOException {
      .setBody("abc"));

    String ipAddress = InetAddress.getLocalHost().getHostAddress();

    Call call = client.newCall(new Request.Builder()
      .url(server.url("/").newBuilder().host(ipAddress).build())
      .build());
    assertThat(response.code()).isEqualTo(200);
    assertThat(response.body().string()).isEqualTo("abc");
    response.body().close();

    assertThat(listener.recordedEventTypes()).containsExactly("CallStart",
      "ProxySelectStart", "ProxySelectEnd",
      "ConnectStart", "ConnectEnd", "ConnectionAcquired", "RequestHeadersStart",
      "RequestHeadersEnd", "ResponseHeadersStart", "ResponseHeadersEnd", "ResponseBodyStart",
      "ResponseBodyEnd", "ConnectionReleased", "CallEnd");
  }
```

TaoZang:
> Please add a test that confirms this behaviour. Something in EventListenerTest like
> 
> ```
> 
> 
>   @Test public void successfulCallEventSequenceForIpAddress() throws IOException {
>     server.enqueue(new MockResponse()
>       .setBody("abc"));
> 
>     String ipAddress = InetAddress.getLocalHost().getHostAddress();
> 
>     Call call = client.newCall(new Request.Builder()
>       .url(server.url("/").newBuilder().host(ipAddress).build())
>       .build());
>     Response response = call.execute();
>     assertThat(response.code()).isEqualTo(200);
>     assertThat(response.body().string()).isEqualTo("abc");
>     response.body().close();
> 
>     assertThat(listener.recordedEventTypes()).containsExactly("CallStart",
>       "ProxySelectStart", "ProxySelectEnd",
>       "ConnectStart", "ConnectEnd", "ConnectionAcquired", "RequestHeadersStart",
>       "RequestHeadersEnd", "ResponseHeadersStart", "ResponseHeadersEnd", "ResponseBodyStart",
>       "ResponseBodyEnd", "ConnectionReleased", "CallEnd");
>   }
> ```

Sure, it seems a reasonable test case to cover this change.

yschimke (approved):

JoshuaJamesOng:
> I'll backport this to 4.9.x when we have a fix, but until that happens you might want to have your own Dns wrapper that does this logic yourself.
> 
> Thanks for the report and PR.

@yschimke I see this change was accepted into the `5.x` alphas. Are there any plans to backport this to `4.x`, now?

yschimke:
Trying on https://github.com/square/okhttp/pull/7648

PR #7075, opened 2022-02-12
Adopt Version Catalog and Updates plugin

yschimke:
In progress, but validating it works and simplifies routine updates.

Updating to releases would be with 

```
$ ./gradlew versionCatalogUpdate
```


buildSrc/src/main/kotlin/artifacts.kt
```diff
@@ -45,7 +45,7 @@ private fun Project.applyOsgi(
   val osgi = project.sourceSets.create("osgi")
   val osgiApi = project.configurations.getByName(osgiApiConfigurationName)
   project.dependencies {
-    osgiApi(Dependencies.kotlinStdlibOsgi)
+    osgiApi("org.jetbrains.kotlin:kotlin-osgi-bundle:1.6.10")
```
yschimke, line 48: Can refer to from buildSrc.

build.gradle.kts
```diff
@@ -89,28 +94,29 @@ subprojects {
 
   val checkstyleConfig: Configuration by configurations.creating
   dependencies {
-    checkstyleConfig(Dependencies.checkStyle) {
+    checkstyleConfig("com.puppycrawl.tools:checkstyle:9.2") {
```
yschimke, line 97: Can move into an afterEvaluate to support?

gradle/libs.versions.toml
```diff
@@ -0,0 +1,167 @@
+[versions]
+biz-aQute-bnd = "6.1.0"
```
yschimke, line 2: These include version bumps see https://github.com/square/okhttp/pull/7075/commits/f817079572c98fc9033677ee4aefb944cecdd5a0

gradle/libs.versions.toml
```diff
@@ -0,0 +1,167 @@
...
+kotlin-platform-android = "kotlin-platform-android:1.6.10"
+kotlin-platform-common = "kotlin-platform-common:1.6.10"
+kotlin-platform-js = "kotlin-platform-js:1.6.10"
+kotlin-platform-jvm = "kotlin-platform-jvm:1.6.10"
+kotlin-scripting = "kotlin-scripting:1.6.10"
+kotlin2js = "kotlin2js:1.6.10"
+ltgt-errorprone = "net.ltgt.errorprone:2.0.2"
+palantir-graal = "com.palantir.graal:0.10.0"
+vanniktech-maven-publish = "com.vanniktech.maven.publish:0.18.0"
+vanniktech-maven-publish-base = "com.vanniktech.maven.publish.base:0.18.0"
```
Goooler, line 167: The plugins block could be removed.

build.gradle.kts
```diff
@@ -1,25 +1,28 @@
...
+    classpath(libs.gradleplugin.palantir.graal)
+    classpath(libs.gradleplugin.aqute.bnd)
+    classpath(libs.gradleplugin.johnrengelman.shadow)
+    classpath(libs.gradleplugin.japicmp)
+    classpath(libs.gradleplugin.animalsniffer)
+    classpath(libs.gradleplugin.errorprone)
+    classpath(libs.gradleplugin.spotless)
+    classpath(libs.gradleplugin.vanniktech)
+    classpath(libs.gradleplugin.benmanes.versions)
+    classpath(libs.gradleplugin.littlerobots.vcu)
```
Goooler, line 25: Suggest removing these 2 plugins after migration.

yschimke: Why? They simplify the update process. For me this is more valuable than just refactoring our dependency files.

Goooler: Once you finish the migration, the `littlerobots.vcu` is less helpful, if you want to check updates, [using a Gradle init script](https://github.com/ben-manes/gradle-versions-plugin#using-a-gradle-init-script) might be a better solution; if you want to update automatically, can use [renovate](https://www.whitesourcesoftware.com/free-developer-tools/renovate/).

yschimke: Is there a good sample of the former? It appears to have the same effect, but requires editing files that we currently currently .gitignore.

JakeWharton: We're doing renovate on cashapp org. It's... fine. Will hold us over until dependabot decides to wake up and actually support other ecosystems.

Goooler: > Is there a good sample of the former? It appears to have the same effect, but requires editing files that we currently currently .gitignore.

detekt/detekt/pull/4294

yschimke: OK, I'll remove the plugins, and wait for someone in Block to configure for this repo.

I can see them coming through now https://github.com/yschimke/okhttp/pulls

JakeWharton: Involuntary shudder reading "someone at Block"... 🤮

swankjesse (approved):

build.gradle.kts
Goooler: Yeah.
```kotlin
  afterEvaluate {
    val checkstyleConfig: Configuration by configurations.creating
    dependencies {
      checkstyleConfig(libs.checkStyle) {
        isTransitive = false
      }
    }
    ...
  }
```

gradle/libs.versions.toml
```diff
@@ -0,0 +1,102 @@
...
+com-squareup-moshi = "1.13.0"
+com-squareup-okio = "3.0.0"
+conscrypt = "2.5.2"
+de-mannodermaus-junit5 = "1.3.0"
+graalvm = "22.0.0.2"
+info-picocli = "4.6.3"
+junit5 = "5.8.2"
+junit5Android = "1.3.0"
+moshi = "1.13.0"
+okio = "3.0.0"
```
Goooler, line 13:
```suggestion
```


Duplicated with:
```toml
com-squareup-moshi = "1.13.0"
com-squareup-okio = "3.0.0"
```

build.gradle.kts
Goooler: @yschimke Missed removing this?

yschimke: Yep - fixed here https://github.com/square/okhttp/pull/7081/files

PR #7077, opened 2022-02-14
Move assertk to tests dependencies

EricKuck:
Assertk was previously being brought in as a runtime dependency. It's only used in tests, so let's move it over.


yschimke:
Thanks!

PR #7085, opened 2022-02-17
Keep 'compileKotlin' and 'compileJava' task use the same JVM target 11

TangHuaiZhe:
Fix a compile warning.
image

The warning log can be found in https://github.com/square/okhttp/runs/5214030818?check_suite_focus=true


TangHuaiZhe:
Closed for https://github.com/gradle/gradle/issues/18935

PR #7093, opened 2022-02-19
Change Fast Fallback to race TCP only

swankjesse:
It will only attempt one TCP connection at a time.

### DIFF
--- a/okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/FastFallbackExchangeFinder.kt
+++ b/okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/FastFallbackExchangeFinder.kt
@@ -35,8 +35,14 @@ internal class FastFallbackExchangeFinder(
 ) {
   private val connectDelayMillis = 250L
 
-  /** Plans currently being connected, and that will later be added to [connectResults]. */
-  private var connectsInFlight = CopyOnWriteArrayList<Plan>()
+  /**
+   * Plans currently being connected, and that will later be added to [connectResults]. This is
+   * mutated by the call thread only. If is accessed by background connect threads.
+   */
+  private val tcpConnectsInFlight = CopyOnWriteArrayList<Plan>()
+
+  /** Plans we started, and canceled, because another plan won a race. */
+  private val deferredPlans = ArrayDeque<Plan>()
 
   /**
    * Results are posted here as they occur. The find job is done when either one plan completes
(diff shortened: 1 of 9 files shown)


okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/FastFallbackExchangeFinder.kt
```diff
@@ -35,132 +35,139 @@ internal class FastFallbackExchangeFinder(
...
+        launchTcpConnect()
 
-        val connection = awaitConnection()
-        if (connection != null) return connection
+        var connectResult = awaitTcpConnect() ?: continue
 
-        morePlansExist = morePlansExist && routePlanner.hasMoreRoutes()
-      }
+        if (connectResult.isSuccess) {
+          // We have a connected TCP connection. Cancel the other racing connects that all lost.
```
yschimke, line 70: At least in the comment, maybe worth clarifying, this is cancel and defer.

okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/FastFallbackExchangeFinder.kt
```diff
@@ -35,132 +35,139 @@ internal class FastFallbackExchangeFinder(
...
+          cancelInFlightConnects()
 
-      throw firstException!!
-    } finally {
-      for (plan in connectsInFlight) {
-        plan.cancel()
+          // Finish connecting. We won't have to if the winner is from the connection pool.
+          if (!connectResult.plan.isReady) {
```
yschimke, line 74: Is it better if this doesn't need to be checked? Can connectTlsEtc just shortcircuit and return this?

okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/FastFallbackExchangeFinder.kt
```diff
@@ -35,132 +35,139 @@ internal class FastFallbackExchangeFinder(
...
+          if (firstException == null) {
+            firstException = throwable
+          } else {
+            firstException.addSuppressed(throwable)
+          }
+        }
+
+        val nextPlan = connectResult.nextPlan
+        if (nextPlan != null) {
+          deferredPlans.addFirst(nextPlan)
```
yschimke, line 96: Since this is somewhat derivitive, another ConnectSpec or a tunnel, is the point to put it after other more different routes?

okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/FastFallbackExchangeFinder.kt
```diff
@@ -35,132 +35,139 @@ internal class FastFallbackExchangeFinder(
...
 
-  /** Plans currently being connected, and that will later be added to [connectResults]. */
-  private var connectsInFlight = CopyOnWriteArrayList<Plan>()
+  /**
+   * Plans currently being connected, and that will later be added to [connectResults]. This is
+   * mutated by the call thread only. If is accessed by background connect threads.
+   */
+  private val tcpConnectsInFlight = CopyOnWriteArrayList<Plan>()
+
+  /** Plans we started, and canceled, because another plan won a race. */
```
yschimke, line 44: Maybe worth a comment that this is always a clean plan, cancelled and retryed.

The comment here suggests it's almost the same plan.

```
      plan.cancel()
      val retry = plan.retry() ?: continue
```

yschimke: Or is that a feature of the current strategy, and potentially these could be in progress ones, that we are holding for some time?

okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/ReusePlan.kt
```diff
@@ -0,0 +1,34 @@
...
+
+  override fun connectTcp() = error("already connected")
+
+  override fun connectTlsEtc() = error("already connected")
+
+  override fun handleSuccess() = connection
+
+  override fun cancel() = error("unexpected cancel")
+
+  override fun retry() = error("unexpected retry")
```
yschimke, line 33: I'm curious about some of these. Should 

- retry return null
- cancel do nothing
- connectTcp, connectTlsEtc - return this

okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/FastFallbackExchangeFinder.kt
```diff
@@ -35,132 +35,139 @@ internal class FastFallbackExchangeFinder(
...
+        deferredPlans.isNotEmpty()
+      ) {
         if (routePlanner.isCanceled()) throw IOException("Canceled")
 
-        launchConnect()
+        launchTcpConnect()
 
-        val connection = awaitConnection()
-        if (connection != null) return connection
+        var connectResult = awaitTcpConnect() ?: continue
```
yschimke, line 67: Do you intend to make this loop and wait connectDelayMillis more nuanced in future? Basically the "more nuanced implementation" from the spec. If so, what about pulling up connectDelayMillis and passing it into awaitTcpConnect() as a param?

okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/FastFallbackExchangeFinder.kt
```diff
@@ -35,132 +35,139 @@ internal class FastFallbackExchangeFinder(
...
     try {
-      while (morePlansExist || connectsInFlight.isNotEmpty()) {
+      while (routePlannerHasMoreRoutes ||
+        tcpConnectsInFlight.isNotEmpty() ||
+        deferredPlans.isNotEmpty()
+      ) {
         if (routePlanner.isCanceled()) throw IOException("Canceled")
 
-        launchConnect()
+        launchTcpConnect()
```
yschimke, line 65: Do you always want to start a new connection on each go around of the loop? If you already have 2-3 on the go, you may not have waited the minimum 100ms "Minimum Connection Attempt Delay", and may either want to delay or check if other plans have tcp connected also.

okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/FastFallbackExchangeFinder.kt
swankjesse: Doen.

swankjesse: Done.

okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/FastFallbackExchangeFinder.kt
swankjesse: Yeah I like that, but when I tried to change to it it gets a bit sloppy. For example, in the FakePlan it’ll need to identify whether to sleep a 2nd time.

okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/FastFallbackExchangeFinder.kt
swankjesse: Exactly. I added this comment:
```
          // Try this plan's successor before deferred plans because it won the race!
```

okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/FastFallbackExchangeFinder.kt
swankjesse: Yeah this doc is wrong. I changed to this:
```
  /**
   * These are retries of plans that were canceled when they lost a race. If the race's winner ends
   * up not working out, this is what we'll attempt first.
   */
```

okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/ReusePlan.kt
swankjesse: At the moment these code paths are unreachable. Neither `ReusePlan` nor `FailedPlan` are ever async, and so we never need to worry about them being canceled or retried.

Admittedly this is coupled to the implementation of `FastFallbackExchangeFinder`. I slightly prefer making this presumably-unreachable code explode than trying to make it do something reasonable if its unexpectedly executed.

okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/FastFallbackExchangeFinder.kt
swankjesse: Great idea. Done.

okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/FastFallbackExchangeFinder.kt
swankjesse: Ooooh this is a great find. Lemme write a test case for that.

yschimke (approved):
Looks good, test passing, good test coverage and readable design.  A nice first class domain model for connection establishment.

swankjesse:

PR #7097, opened 2022-02-20
Stop tracking failures in RoutePlanner

swankjesse:
This model doesn't work now that we're pulling routes concurrently.
Instead rely on the exchange and connection to track failures.

### DIFF
--- a/okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/SequentialExchangeFinder.kt
+++ b/okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/SequentialExchangeFinder.kt
@@ -1,0 +1,66 @@
+/*
+ * Copyright (C) 2015 Square, Inc.
+ *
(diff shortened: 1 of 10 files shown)


okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/SequentialExchangeFinder.kt
```diff
@@ -0,0 +1,66 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okhttp3.internal.connection
+
+import java.io.IOException
+
+/** Attempt routes one at a time until one connects. */
+internal class SequentialExchangeFinder(
```
swankjesse, line 21: Renamed from ExchangeFinder, which is now an interface.

yschimke (approved):

swankjesse:

PR #7099, opened 2022-02-20
Don't leak a connection that's canceled before it connects

swankjesse:
We had a bug where calling ConnectPlan.cancel() before ConnectPlan.connectTcp()
would drop the cancel. This is because the rawSocket wasn't created yet to
cancel.

Also document fast fallback on our documentation site.

### DIFF
--- a/docs/features/connections.md
+++ b/docs/features/connections.md
@@ -16,9 +16,9 @@ They're also concrete: each URL identifies a specific path (like `/square/okhttp
 
 ### [Addresses](https://square.github.io/okhttp/4.x/okhttp/okhttp3/-address/)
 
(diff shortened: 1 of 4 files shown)


yschimke:
Seems like the terminal state for a socket should be, either specifically closed, ended normally after use, or exception.

We don't know what a custom socket factory may do, so is it worth closing in the cancel case?

yschimke (approved):

swankjesse:
The nice thing about raw TCP sockets is we have a single API for canceling and closing: `Socket.close()`.

swankjesse:

yschimke:
> The nice thing about raw TCP sockets is we have a single API for canceling and closing: `Socket.close()`.

Yep - from the review I missed that the calling function was doing the cleanup.

```
      if (!success) {
        rawSocket?.closeQuietly()
      }
```

PR #7102, opened 2022-02-21
Update Response.request docs to track behavior

swankjesse:
Also change internal interceptors to explicitly retain the requests
that were forwarded down the chain.


yschimke:
Do we have any tests covering this?

yschimke (approved):

swankjesse:
@yschimke lemme add one. We have lots of tests that fail if we return the wrong response, but nothing precise.

PR #7106, opened 2022-02-24
Add Failsafe to the "works with OkHttp" page

jhalterman:
This PR adds [Failsafe](https://failsafe.dev/okhttp/) to the "works with OkHttp" page. I'm not positive what the scope of that page is meant to be, so hopefully this fits, but if not that's cool. 

Also, I'm not sure what the process is for generating actual docs to the gh-pages branch, but if that's something I should do, I can followup with another PR.


swankjesse (approved):
Thanks!

PR #7107, opened 2022-02-24
Response KMP

yschimke:

PR #7108, opened 2022-02-24
Call KMP

yschimke:
Builds on https://github.com/square/okhttp/pull/7107 so needs to land afterwards.

### DIFF
--- a/okhttp/src/commonMain/kotlin/okhttp3/Call.kt
+++ b/okhttp/src/commonMain/kotlin/okhttp3/Call.kt
@@ -15,14 +15,14 @@
  */
 package okhttp3
 
-import java.io.IOException
+import okio.IOException
 import okio.Timeout
 
 /**
  * A call is a request that has been prepared for execution. A call can be canceled. As this object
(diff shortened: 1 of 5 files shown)


okhttp/src/commonMain/kotlin/okhttp3/Call.kt
```diff
@@ -15,14 +15,14 @@
...
-import java.io.IOException
+import okio.IOException
 import okio.Timeout
 
 /**
  * A call is a request that has been prepared for execution. A call can be canceled. As this object
  * represents a single request/response pair (stream), it cannot be executed twice.
  */
-interface Call : Cloneable {
+interface Call {
```
yschimke, line 25: kotlin.Cloneable doesn't work in common, and it's purely a marker interface, so while an API change, the actual implementations will all implement this, see RealCall.

yschimke:
Actually will need to drop execute as we can't support it on JS, so have expect and actual interfaces.

yschimke:
@swankjesse yay

API check failed

	public final fun parse (Lokhttp3/Headers;)Lokhttp3/CacheControl;

PR #7133, opened 2022-02-28
Spotless with ratchet

yschimke:
Configure spotless, with ratchet on modified files.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -175,7 +175,7 @@ subprojects {
 
   apply(plugin = "com.diffplug.spotless")
   configure<SpotlessExtension> {
-    ratchetFrom("master")
+    ratchetFrom("origin/master")
     kotlin {
       target("**/*.kt")
       ktlint(libs.versions.ktlint.get()).userData(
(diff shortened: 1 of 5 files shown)


yschimke:
Testing with a single change to a local file.

```
Execution failed for task ':okhttp:spotlessKotlinCheck'.
> The following files had format violations:
      src/commonMain/kotlin/okhttp3/Protocol.kt
          @@ -15,10 +15,9 @@
           ·*/
           package·okhttp3
```

yschimke:
Breaking on https://github.com/diffplug/spotless/issues/710

yschimke:
@swankjesse landing to test on master.

JakeWharton:
Lgtm

yschimke:
Thanks for the setup advice @JakeWharton

swankjesse (commented):
LGTM

Goooler (commented):
LGTM

build.gradle.kts
```diff
@@ -173,6 +173,23 @@ subprojects {
     }
   }
 
+  apply(plugin = "com.diffplug.spotless")
```
swankjesse, line 176: We should be loading the config from `.editorconfig`, rather than partially duplicating it here. As-is the two disagree on where `java.*` imports go.

Goooler: But it seems haven't support yet. See https://github.com/diffplug/spotless/issues/142

Issue #2604, opened 2016-06-05
Multipart - how to disable Content-Length for a part?


rafakob:
Question:
How to disable `Content-Length` header for a part? I need to do this because an API that I'm using does not allow it (and I can't do anything about it).

I'm sure that the problem is that header, because I've tried to upload a file manually using `HttpURLConnection` and it works great without `Content-Length`, and it fails immediately when I add it.

I've tried to add interceptor to the request (both `addInterceptor` and `addNetworkInterceptor`) in order to remove unwanted header. But it is not here (I can see only `Content-Disposition`). Is there another way to disable it?

Mentioning
https://github.com/square/okhttp/issues/2138

dave-r12:
@rafakob I think the workaround would be to copy/paste MultipartBody into your project and remove the code that adds the Content-Length.

rafakob:
But Retrofit's `@Part` annotation doesn't allow any other custom type. Could you explain me a little bit more how to do this?

swankjesse:
You might not be able to use Retrofit for this.

rafakob:
Ok, so I've commented out lines 157-159 in my `CustomMultipartBody.java` and manually performed request (without Retrofit). It works! Maybe not so elegant but definitely better than using `HttpURLConnection` because I can use my global `OkHttpClient` instance with all interceptors, cookies, authorization etc. 

Anyway, did anyone else report similar issues? If so, maybe it would be a good idea to make that header optional?

swankjesse:
You’re the only one. You should petition the owner of the server who has this limitation to fix their server so we don’t have to fix our client.

cjeker:
This is not a server issue but actually a problem on the client. Non of the RFC handling multipart/form-data do allow Content-Length as a header field for individual parts.
See also:
https://www.ietf.org/rfc/rfc2388.txt
https://www.ietf.org/rfc/rfc2045.txt
https://www.ietf.org/rfc/rfc2046.txt

https://www.ietf.org/rfc/rfc1867.txt has this:

>    The HTTP protocol may require a content-length for the overall
>    transmission. Even if it were not to do so, HTTP clients are
>    encouraged to supply content-length for overall file input so that a
>    busy server could detect if the proposed file data is too large to be
>    processed reasonably ....

And the HTTP spec mentions the following:

>       Note: HTTP's use of Content-Length for message framing differs
>       significantly from the same field's use in MIME, where it is an
>       optional field used only within the "message/external-body"
>       media-type.

[...]

PR #7139, opened 2022-03-04
Remove Content-Length in MultipartRequest Bodies

sdandois:
Solves #2604.

This removes the Content-Length header in a part of a multipart request. 

I've tested with Chrome and it doesn't add the Content-Length header.

### DIFF
--- a/okhttp/src/jvmMain/kotlin/okhttp3/MultipartBody.kt
+++ b/okhttp/src/jvmMain/kotlin/okhttp3/MultipartBody.kt
@@ -139,7 +139,7 @@ class MultipartBody internal constructor(
       }
 
       val contentLength = body.contentLength()
-      if (contentLength == -1L) {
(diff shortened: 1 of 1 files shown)


JakeWharton:
That issue is not open.

yschimke:
Closing - the linked issue has workarounds to fix this for servers that are incorrectly rejecting this.

sdandois:
> Closing - the linked issue has workarounds to fix this for servers that are incorrectly rejecting this.

@JakeWharton, thanks for the response. I've updated the description with a new issue describing my case, which is slightly different than the one in the previous closed issue.

yschimke:
Are you able to run vanilla OkHttp with the rewritten MultipartBody class from your project?

Unless there is a specific opinionated decision to make this change, I think having your own MultipartBody is probably the correct workaround here for a faulty server.

Issue #5874, opened 2020-03-17
Support HTTP protocol upgrades


swankjesse:
We should implement user-requested protocol upgrades as specified by [RFC 7230 section 6.7](https://tools.ietf.org/html/rfc7230#section-6.7).

### Use Cases

 * [Docker connection hijacking](https://docs.docker.com/engine/api/v1.22/#32-hijacking) see also [docker-java #1347](https://github.com/docker-java/docker-java/pull/1347)
 * Our own web sockets
 * Third-party web sockets implementations on OkHttp

### Eligibility

An HTTP/1 call is upgraded if all of the following are true:

 - The caller’s request includes an `Upgrade` header 
 - The caller’s request includes a `Connection` header with the value `upgrade`
 - The server’s response includes an `Upgrade` header 
 - The server’s response includes a `Connection` header with the value `upgrade`
 - The response code is 101.

### On Upgrade

A successful upgrade changes the behavior of the HTTP response:

 - The `ResponseBody` is null.
 - The response has a non-null `Streams` object that carries the input and output stream. Note that the source and sink timeouts should work properly!

```
interface Streams {
  val source: Source
  val sink: Sink
  fun cancel()
}
```

```
class Response {
  ...
  /** Non-null if this response is a successful upgrade ... */
  @get:JvmName("streams") val streams: Streams?
}
```

A successful upgrade has these side-effects:

 - The call timeout is immediately completed. (`RealCall.timeoutEarlyExit()`)
 - The connection is forbidden from carrying new exchanges. (`RealConnection.noNewExchanges()`)
 - The socket’s read timeout is disabled. (`Socket.setSoTimeout()`) 
 - If the call is asynchronous (`Call.enqueue()`), the call counts against Dispatcher limits until `onResponse()` returns.

I’ve used the class name `Streams` instead of `UpgradedConnection` or something feature-specific because I think we might be able to reuse this type for `CONNECT` calls.

### Web Sockets

Can we migrate our internal web sockets code to use this? Ideally yes, though that shouldn’t block this from being released.

### Event Listeners

Ideally we have well defined behavior for `EventListeners` on an upgraded connection. We need to decide whether to count bytes of the upgraded connection for the benefit of listeners.

yschimke:
What are the cancellation & close rules?  Half closed supported?

swankjesse:
Source.close(): releases resources and cause writes to fail.
Sink.close(): flushes buffered data, releases resources, and sends EOF to reader.
[...]

PR #7141, opened 2022-03-05
Experiment with Early Hints

yschimke:
For discussion, not connected between Http2Connection/Stream that receives the EarlyHints headers, and Exchange/EventListener for reporting them.

Early Hints should be delivered before the Response headers are available, and the early hints may not match actual headers, so we can't delay and merge them to keep same API.  So using the EventListener seems like the right place to use this.

The challenge I have is that normal flow is driven in CallServerInterceptor, and Exchange is the place where we have the Http2ExchangeCodec and EventListener.  Not clear how to pass

Other questions: For the full set of 100 responses https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/100, is there anything else to consider?

100 Continue - implemented in CallServerInterceptor
101 Switching Protocols - not implemented, we use ALPN, be good to add https://github.com/square/okhttp/issues/5874
103 - This PR.


okhttp/src/jvmTest/java/okhttp3/EarlyHintsTest.kt
```diff
@@ -0,0 +1,52 @@
...
+  val eventListener = RecordingEventListener()
+
+  private val client = clientTestRule.newClientBuilder()
+    .eventListenerFactory(clientTestRule.wrap(eventListener))
+    .build()
+
+  @Test
+  fun testCloudflare() {
+    val request = Request.Builder().url("https://tradingstrategy.ai")
+      .header("User-Agent", "curl/7.77.0")
```
yschimke, line 38: Required to allow Early Hints. I guess User Agent filtering is making sure OkHttp doesn't get broken?

yschimke:
Should HTTP/1.1 early hints be fully supported, or just tested to confirm it's not breaking?

gesellix:
For #5874, support for "101 Switching Protocols" for HTTP/1 would be great.

yschimke:
@gesellix Yep - but not sure it's incremental enough to just throw it in with this. Maybe it is and we can handle these cleanly in a common approach, but switching from HTTP/1.1 to HTTP/2 seems a lot more involved.

gesellix:
@yschimke yes, I certainly didn't want to put any pressure on this. See my comment only as reaction on your thoughts about "is there anything else to consider?" and me trying to keep the awareness about the other issue. Status 103 isn't the same as 101, so this is fine with me :)

yschimke:
Its a good time to look to complete this area.

swankjesse:
An alternative: use the existing EventListener.responseHeaders() function for this, but also for all 1xx responses.

That means users also receive other 1xx codes for free, and the Response type of that API means they get the status code.

yschimke:
Should I just handle it like the continue logic, read headers multiple times until not a 1xx

yschimke:
And create a Response to pass to the listener?

okhttp/src/jvmMain/kotlin/okhttp3/internal/http/CallServerInterceptor.kt
```diff
@@ -103,8 +103,10 @@ class CallServerInterceptor(private val forWebSocket: Boolean) : Interceptor {
           .receivedResponseAtMillis(System.currentTimeMillis())
           .build()
       var code = response.code
-      if (code == 100) {
-        // Server sent a 100-continue even though we did not request one. Try again to read the
+      while (code in 100..199) {
+        exchange.eventListener.responseHeadersEnd(exchange.call, response)
+
+        // Server sent a 1xx so report and continue. Try again to read the
```
yschimke, line 109: Seems like I shouldn't ignore a "Switching Protocols" here. 🤔

swankjesse: I think it's a future-proof default. What does curl do with HTTP 105??

swankjesse: I assume this is broken for 101?

swankjesse:
I think this might benefit from a bigger uglier loop to handle combinations of 103 and 100? Seems possible that different middlewares might give us a 103 followed by a 100, or vice versa.

yschimke:
> I think this might benefit from a bigger uglier loop to handle combinations of 103 and 100? Seems possible that different middlewares might give us a 103 followed by a 100, or vice versa.

I think this leaves us with 2 options, and implementing 1 or 2.

a) A minimal change to put tight loops around reading the response headers, and looping with 103 and spitting out to the listener.
b) A bigger change - to cleanly support 1xx overall. I think refactoring the bigger loop is part of this, and it should include Switching Protocols rather than try to get a cleaner change before we consider this path.

Should we bother with a) first? Or aim just for b?)?

yschimke:
> An alternative: use the existing EventListener.responseHeaders() function for this, but also for all 1xx responses.
> 
> That means users also receive other 1xx codes for free, and the Response type of that API means they get the status code.

This feels like a breaking change, which is ok for 5.0, because clients need to know to ignore responseBody sometimes as it may not relate to the response they are waiting for.

swankjesse:
I think `b` is the place to go first. Also we might need some help from MockWebServer here. Maybe this?

```
class MockResponse {
  ...
  val informationalResponses = List<MockResponse>
}
```

yschimke:
Yep, makes sense.

PR #7142, opened 2022-03-05
Coroutines module.

yschimke:
Both useful for general coroutines, but effectively the
multiplatform API for call. Done this way to avoid
kotlinx.coroutines.core being a core dependency on JVM.


okhttp-ktx/src/jvmMain/kotlin/okhttp3/JvmCallExtensions.kt
```diff
@@ -0,0 +1,41 @@
...
+ *
+ */
+
+package okhttp3
+
+import kotlin.coroutines.resumeWithException
+import kotlinx.coroutines.suspendCancellableCoroutine
+import okio.IOException
+
+suspend fun Call.awaitResponse(): Response {
```
yschimke, line 24: This could be shared but for two issues

a) There is no natural home, okhttp doesn't have coroutines
b) okhttp-ktx is too high for JS.

okhttp/src/jsMain/kotlin/okhttp3/-JsPlatform.kt
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+package okhttp3
+
+import kotlin.coroutines.resumeWithException
+import kotlinx.coroutines.suspendCancellableCoroutine
+import okio.IOException
+
+suspend fun Call.jsExecuteAsync(): Response {
+  return suspendCancellableCoroutine { continuation ->
```
yschimke, line 25: This should be in nonJvmMain, but I can't reference the suspendCancellableCoroutine from that module.  It looks like it is defined in kotlix-coroutines-core-common, but only works for jvm and js.

JakeWharton:
This should probably just be named okhttp-coroutines. Ktx is named as such for Kotlin extensions to Java libraries. You're now a Kotlin library and 99% of Kotlin niceties can go in the main library. Instead, what you've built is an integration library with a third-party library.

yschimke:
The sample code and test made me realise that ResponseBody.string() is an unsafe pattern.  Should we remove from the expect class, and make clients use directly the source and provide an stringAsync() method?

gradle/libs.versions.toml
```diff
@@ -65,6 +65,7 @@ kotlin-test-annotations = { module = "org.jetbrains.kotlin:kotlin-test-annotatio
 kotlin-test-common = { module = "org.jetbrains.kotlin:kotlin-test-common", version.ref = "org-jetbrains-kotlin" }
 kotlin-test-js = { module = "org.jetbrains.kotlin:kotlin-test-js", version.ref = "org-jetbrains-kotlin" }
 kotlin-test-junit = { module = "org.jetbrains.kotlin:kotlin-test-junit", version.ref = "org-jetbrains-kotlin" }
+kotlinx-coroutines-core = "org.jetbrains.kotlinx:kotlinx-coroutines-test:1.6.0"
```
Goooler, line 68:
```suggestion
kotlinx-coroutines-core = "org.jetbrains.kotlinx:kotlinx-coroutines-core:1.6.0"
```

swankjesse:
Let's not support request or response body streaming in JS.

When you have a response body, it's body is always fully buffere. It'd be equivalent to the terminal interceptor waiting for the complete response body before returning.

yschimke:
> Let's not support request or response body streaming in JS.
> 
> When you have a response body, it's body is always fully buffere. It'd be equivalent to the terminal interceptor waiting for the complete response body before returning.

Anything I should change to the existing Response and ResponseBody class in multiplatform.  Will do in a separate PR.

gradle/libs.versions.toml
```diff
@@ -65,6 +65,8 @@ kotlin-test-annotations = { module = "org.jetbrains.kotlin:kotlin-test-annotatio
 kotlin-test-common = { module = "org.jetbrains.kotlin:kotlin-test-common", version.ref = "org-jetbrains-kotlin" }
 kotlin-test-js = { module = "org.jetbrains.kotlin:kotlin-test-js", version.ref = "org-jetbrains-kotlin" }
 kotlin-test-junit = { module = "org.jetbrains.kotlin:kotlin-test-junit", version.ref = "org-jetbrains-kotlin" }
+kotlinx-coroutines-core = "org.jetbrains.kotlinx:kotlinx-coroutines-core:1.6.0"
+kotlinx-coroutines-test = "org.jetbrains.kotlinx:kotlinx-coroutines-test:1.6.0"
```
Goooler, line 69:
```suggestion
kotlinx-coroutines-core = { module = "org.jetbrains.kotlinx:kotlinx-coroutines-core", version.ref = "org-jetbrains-coroutines" }
kotlinx-coroutines-test = { module = "org.jetbrains.kotlinx:kotlinx-coroutines-test", version.ref = "org-jetbrains-coroutines" }
```

Goooler: Need to add `org-jetbrains-coroutines = "1.6.0"` in versions block.

okhttp-coroutines/README.md
```diff
@@ -0,0 +1,14 @@
+OkHttp Coroutines
+==========
+
+Komfortable APIs for Kotlin clients.
```
swankjesse, line 4: I can’t

okhttp-coroutines/src/jvmMain/kotlin/okhttp3/JvmCallExtensions.kt
```diff
@@ -0,0 +1,41 @@
...
+ *
+ */
+
+package okhttp3
+
+import kotlinx.coroutines.suspendCancellableCoroutine
+import okio.IOException
+import kotlin.coroutines.resumeWithException
+
+suspend fun Call.awaitResponse(): Response {
```
swankjesse, line 24: Are `executeAsync` and `awaitResponse` synonyms? Why not just use `executeAsync` only?

swankjesse: (I expect _await_ blocks waiting for something that’s already running, and doesn’t launch anything?)

okhttp/src/nonJvmMain/kotlin/okhttp3/Call.kt
```diff
@@ -16,9 +16,38 @@
...
+   * Note that transport-layer success (receiving a HTTP response code, headers and body) does not
+   * necessarily indicate application-layer success: `response` may still indicate an unhappy HTTP
+   * response code like 404 or 500.
+   *
+   * @throws IOException if the request could not be executed due to cancellation, a connectivity
+   *     problem or timeout. Because networks can fail during an exchange, it is possible that the
+   *     remote server accepted the request before the failure.
+   * @throws IllegalStateException when the call has already been executed.
+   */
+  suspend fun executeAsync(): Response
```
swankjesse, line 49: Should we have blocking execute() on native?

swankjesse (approved):
Good stuff

okhttp/src/nonJvmMain/kotlin/okhttp3/Call.kt
yschimke: I think we need some concrete example clients to try this with, before adding more API.

There are different sweet spots that I think we will find get used

1) Multiplatform clients wanting a common base, client code in commonMain.
2) Individual platform specific calls even in overall multiplatform projects, client code in jvmMain, jsMain but maybe some common Interceptors.

Issue #7152, opened 2022-03-10
unexpected cancel in ReusePlan.cancel


swankjesse:
I’d assumed this was unreachable and I was wrong!

```
Fatal Exception: java.lang.IllegalStateException: unexpected cancel
       at okhttp3.internal.connection.ReusePlan.cancel(ReusePlan.kt:31)
       at okhttp3.internal.connection.ReusePlan.cancel(ReusePlan.kt:19)
       at okhttp3.internal.connection.FastFallbackExchangeFinder.cancelInFlightConnects(FastFallbackExchangeFinder.kt:174)
       at okhttp3.internal.connection.FastFallbackExchangeFinder.find(FastFallbackExchangeFinder.kt:109)
       at okhttp3.internal.connection.RealCall.initExchange$okhttp(RealCall.kt:267)
       at okhttp3.internal.connection.ConnectInterceptor.intercept(ConnectInterceptor.kt:32)
       at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
       at okhttp3.internal.cache.CacheInterceptor.intercept(CacheInterceptor.kt:96)
       at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
       at okhttp3.internal.http.BridgeInterceptor.intercept(BridgeInterceptor.kt:83)
       at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
       at okhttp3.internal.http.RetryAndFollowUpInterceptor.intercept(RetryAndFollowUpInterceptor.kt:75)
       at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
       at <one of our custom interceptors>

```

PR #7154, opened 2022-03-10
Never put pooled connections in the result queue

swankjesse:
In a race this could cause us to prefer a new connection over
a pooled connection, which violates our invariant that pooled
connections are always used if they're found.

Closes: https://github.com/square/okhttp/issues/7152


yschimke (approved):

yschimke:
Just spotless failing.

PR #7155, opened 2022-03-10
Turn off Spotless style enforcement

swankjesse:
The ratchet thing doesn't work for me 'cause I can't get it
to reformat what I've changed.

Also it's needlessly inconsistent with .editorconfig. We should
wait until that's fixed and possibly try again.


swankjesse:
In particular, this doesn’t work on a branch that `spotlessCheck` is failing on.

```
 ./gradlew spotlessApply
```

yschimke (approved):

PR #7157, opened 2022-03-12
Android module and Async Dns

yschimke:
Primarily an Android specific implementation of AsyncDns. 
This is intended to unblock the Happy Eyeballs, fast follows implementation to handle IPv4 and IPv6 separately.

AsyncDns should be considered an experimental public API.

@swankjesse I don't know how to do this in okhttp core using Flow. Bringing in kotlinx.coroutines is probably awkward for JVM clients already frustrated at getting kotlin, and more likely to introduce versioning issues that stdlib (which has been perfect).

Also an optional okhttp-android module, to allow specific implementations that are network aware*. Published as an AAR file which might help with android specific optimisations like baseline profiles.

Android apps may often need to understand network conditions, choose the appropriate network, and when doing so
should use the appropriate socket factory, dns, and even proxy.


okhttp/src/jvmMain/kotlin/okhttp3/AsyncDns.kt
```diff
@@ -0,0 +1,119 @@
...
+interface AsyncDns {
+  /**
+   * Query dns records for `hostname`, in the order they are received.
+   */
+  fun query(hostname: String, callback: Callback)
+
+  /**
+   * Callback to receive results from the Dns Queries.
+   */
+  interface Callback {
```
yschimke, line 47: There are a lot of subtle considerations in this API design, ordering, repeated methods, empty results.

okhttp/src/jvmMain/kotlin/okhttp3/AsyncDns.kt
```diff
@@ -0,0 +1,119 @@
...
+ *  * Android DnsResolver
+ *  * OkHttp DnsOverHttps
+ *  * dnsjava Resolver
+ *
+ * Implementations of this interface must be safe for concurrent use.
+ *
+ * The interface may be extended by specific AsyncDns implementations with additional records types,
+ * so as SCVB/HTTPS records, which may not be generally applicable.
+ */
+interface AsyncDns {
```
yschimke, line 38: I'd love an Experimental annotation on this.

yschimke:
Two remaining topics

1) When to pull over the changes for DnsOverHttps
2) When to consider the design for other Dns records like Svcb, Https (draft RFC https://datatracker.ietf.org/doc/draft-ietf-dnsop-svcb-https/)

okhttp-android/src/androidTest/kotlin/okhttp3/android/AndroidDnsTest.kt
```diff
@@ -0,0 +1,178 @@
...
+        latch.countDown()
+      }
+
+      override fun onError(dnsClass: AsyncDns.DnsClass, e: IOException) {
+        exception = e
+        latch.countDown()
+      }
+    })
+
+    latch.await()
```
yschimke, line 114: To discuss: This shouldn't be an infinite wait.  If say 10 seconds, does it need to be configurable?

Should existing results be sent? Then onComplete sent? How to avoid events after that?

crazysouthern (approved):

crazysouthern (approved):

swankjesse (commented):
It gets complicated using one instance to deal with either 1 or 2 responses.

What if we make it so each request gets a single response, and let the caller pick what’s in that response? If the caller wants one callback we can give them that, and if they want two callbacks we can give them that too?

```
interface AsyncDns {
  fun lookup(hostname: String, callback: Callback)

  interface Callback {
    fun onFailure(hostname: String, e: IOException)
    fun onResponse(hostname: String, response: List<InetAddress>)
  }

  companion object {
    /** Unions the results from [SYSTEM_IPV4_ONLY] and [SYSTEM_IPV6_ONLY]. */
    val SYSTEM: AsyncDns
    val SYSTEM_IPV4_ONLY: AsyncDns
    val SYSTEM_IPV6_ONLY: AsyncDns
  }
}
```

okhttp-android/src/androidTest/kotlin/okhttp3/android/AndroidAsyncDnsTest.kt
```diff
@@ -0,0 +1,192 @@
...
+      .build()
+    return@lazy HandshakeCertificates.Builder()
+      .addPlatformTrustedCertificates()
+      .heldCertificate(heldCertificate)
+      .addTrustedCertificate(heldCertificate.certificate)
+      .build()
+  }
+
+  private val client = OkHttpClient.Builder()
+    .dns(dns.asDns())
```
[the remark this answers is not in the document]
swankjesse: Makes sense now that I’ve ready below!

yschimke:
> What if we make it so each request gets a single response, and let the caller pick what’s in that response? If the caller wants one callback we can give them that, and if they want two callbacks we can give them that too?

Makes sense, I'll simplify to this API.  But I'll leave the union implementation and the predefined constants out for now.

```
    /** Unions the results from [SYSTEM_IPV4_ONLY] and [SYSTEM_IPV6_ONLY]. */
    val SYSTEM: AsyncDns
    val SYSTEM_IPV4_ONLY: AsyncDns
    val SYSTEM_IPV6_ONLY: AsyncDns
```

okhttp-android/api/okhttp-android.api
```diff
@@ -0,0 +1,12 @@
+public final class okhttp3/android/AndroidAsyncDns : okhttp3/AsyncDns {
+	public static final field Companion Lokhttp3/android/AndroidAsyncDns$Companion;
+	public fun <init> (Lokhttp3/AsyncDns$DnsClass;Landroid/net/Network;)V
+	public synthetic fun <init> (Lokhttp3/AsyncDns$DnsClass;Landroid/net/Network;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
+	public fun query (Ljava/lang/String;Lokhttp3/AsyncDns$Callback;)V
+}
+
+public final class okhttp3/android/AndroidAsyncDns$Companion {
```
yschimke, line 8: These are not nice from java, but as these are Android specific, I think Kotlin is a good assumption.

okhttp-android/src/androidTest/kotlin/okhttp3/android/AndroidAsyncDnsTest.kt
```diff
@@ -0,0 +1,188 @@
...
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Disabled
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.extension.ExtendWith
+import org.opentest4j.TestAbortedException
+
+/**
+ * Run with "./gradlew :android-test:connectedCheck" and make sure ANDROID_SDK_ROOT is set.
+ */
+@ExtendWith(MockWebServerExtension::class)
```
swankjesse, line 45: Yay!

okhttp-android/src/main/kotlin/okhttp3/android/AndroidAsyncDns.kt
```diff
@@ -0,0 +1,66 @@
...
+import android.os.Build
+import androidx.annotation.RequiresApi
+import java.net.InetAddress
+import java.net.UnknownHostException
+import java.util.concurrent.Executors
+import okhttp3.AsyncDns
+
+/**
+ * Dns implementation based on android.net.DnsResolver, which submits two requests for
+ * A and AAAA records, and returns the addresses or exception from each before returning.
```
swankjesse, line 31: Is this comment out of date?

okhttp-android/src/main/kotlin/okhttp3/android/AndroidAsyncDns.kt
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * @param network network to use, if not selects the default network.
+ */
+@RequiresApi(Build.VERSION_CODES.Q)
+class AndroidAsyncDns(
```
swankjesse, line 36: Hide public constructor?

swankjesse: (Hmmm . . . . actually for `Network` this is useful . . .  hmmmm)

okhttp/src/jvmMain/kotlin/okhttp3/AsyncDns.kt
```diff
@@ -0,0 +1,114 @@
...
+ * The main implementations will typically be implemented using specific Dns libraries such as
+ *  * Android DnsResolver
+ *  * OkHttp DnsOverHttps
+ *  * dnsjava Resolver
+ *
+ * Implementations of this interface must be safe for concurrent use.
+ */
+interface AsyncDns {
+  /**
+   * Query dns records for `hostname`, in the order they are received.
```
swankjesse, line 37: nit: DNS should be in SCREAMY CASE in documentation

okhttp/src/jvmMain/kotlin/okhttp3/AsyncDns.kt
```diff
@@ -0,0 +1,114 @@
...
+  fun query(hostname: String, callback: Callback)
+
+  /**
+   * Callback to receive results from the Dns Queries.
+   */
+  interface Callback {
+    /**
+     * Return addresses for a dns question,
+     * typically but always a single class of IPv4 (A) or IPv6 (AAAA).
+     * May be empty.
```
swankjesse, line 48: We should specify what empty means . . . it means the host is unreachable?

okhttp/src/jvmMain/kotlin/okhttp3/AsyncDns.kt
```diff
@@ -0,0 +1,114 @@
...
+     * Returns an error for the dns query.
+     */
+    fun onFailure(hostname: String, e: IOException)
+  }
+
+  /**
+   * Class of Dns addresses, such that clients that treat these differently, such
+   * as attempting IPv6 first, can make such decisions.
+   */
+  enum class DnsClass(val type: Int) {
```
swankjesse, line 62: Need this be a public API? (I think it has to be??)

swankjesse (approved):
For a follow-up I think we could make it so the main OkHttp artifact’s `Platform` class directs us to use this code when running on the Android platform?

okhttp-android/src/main/kotlin/okhttp3/android/AndroidAsyncDns.kt
yschimke: Yep - I actually liked this bit. There are bugs in stackoverflow because android/wear networking returns differing/empty results across Wifi, LTE, Bluetooth paired connections (on Wear).

The choice is implicit currently with Dns.SYSTEM, and this makes it explicit.

okhttp/src/jvmMain/kotlin/okhttp3/AsyncDns.kt
yschimke: Yep - part of public API.

Issue #7167, opened 2022-03-18
HTTP 1xx followed by empty response body is broken


swankjesse:
We’re incorrectly interpreting the HTTP 2xx response that follows as trailers instead of headers in `Http2Stream.receiveHeaders`:

```
      if (!hasResponseHeaders || !inFinished) {
        hasResponseHeaders = true
        headersQueue += headers
      } else {
        ...
      }
```

We might be able to fix by using the presence of the `:status` header to determine if its headers or trailers.

swankjesse:
We should have caught this in `serverRespondsWithUnsolicited100Continue_HTTP2` but MockWebServer incorrectly doesn’t send the 100 response when it isn’t preceded by an `Expect` header.

PR #7168, opened 2022-03-18
Don't interpret headers as trailers after a 100 response

swankjesse:
This only occurs when the response body is empty.

Closes: https://github.com/square/okhttp/issues/7167

### DIFF
--- a/okhttp/src/jvmMain/kotlin/okhttp3/internal/http2/Http2Stream.kt
+++ b/okhttp/src/jvmMain/kotlin/okhttp3/internal/http2/Http2Stream.kt
@@ -282,7 +282,10 @@ class Http2Stream internal constructor(
 
     val open: Boolean
     synchronized(this) {
-      if (!hasResponseHeaders || !inFinished) {
+      if (!hasResponseHeaders ||
+        headers[Header.RESPONSE_STATUS_UTF8] != null ||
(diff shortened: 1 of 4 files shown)


okhttp/src/jvmMain/kotlin/okhttp3/internal/http2/Http2Stream.kt
```diff
@@ -282,7 +282,10 @@ class Http2Stream internal constructor(
 
     val open: Boolean
     synchronized(this) {
-      if (!hasResponseHeaders || !inFinished) {
+      if (!hasResponseHeaders ||
```
yschimke, line 285: Agree about not using inFinished. Is there a cleanup that should be considered for the hasResponseHeaders?  Should it be false until the first non 100?

yschimke (approved):
Makes sense.

okhttp/src/jvmMain/kotlin/okhttp3/internal/http2/Http2Stream.kt
swankjesse: I don't think it matters in practice.

In my first attempt at this I forgot that receiveHeaders is called by both clients and servers, so I handle either `:status` or `:method`.

PR #7172, opened 2022-03-20
Include an okhttp-android baseline.

yschimke:
Evaluating the benefit of a baseline profile for okhttp.

Apps that use baseline profiles, but test the activities rather than network requests would benefit.

Square precedent in https://github.com/square/workflow-kotlin/tree/main/workflow-core/src/main


yschimke:
Supporting docs

https://developer.android.com/studio/profile/baselineprofiles
https://android-developers.googleblog.com/2022/01/improving-app-performance-with-baseline.html

yschimke:
@swankjesse any thoughts, I'd like to validate the improvement using future alpha releases in downstream apps.  No useful benchmark here to compare.

Or maybe internally in Square?

swankjesse:
I like it. And the rationale for testing is good.

Can you add a link somewhere to show how to rebuild this? Can we get GitHub actions to do it?

swankjesse (approved):

yschimke:
I'll follow up with instructions.  Should I try to land https://github.com/square/okhttp/pull/7171?

It's potentially awkward because Intellij doesn't support AGP 7.1.x

swankjesse:
Yeah, I think it'd be nice to have some Android UI that runs OkHttp. I wonder if it'd be worthwhile to have a test field where you can paste a URL?

oceyblue8:
Hey what are we doing herw

Issue #7178, opened 2022-03-24
java.io.IOException: java.util.zip.DataFormatException: invalid distance too far back


steve-chancetop:
Hello, guys. I found an issue when I integrated Java project with kotlin project. This issue occurred when a java client calling a kotlin api (javalin). This issue confused me a lot. Would you guys give some explain of this.
Error trace as
```
Caused by: java.io.IOException: java.util.zip.DataFormatException: invalid distance too far back
	at okio.InflaterSource.readOrInflate(InflaterSource.kt:99)
	at okio.InflaterSource.read(InflaterSource.kt:49)
	at okio.GzipSource.read(GzipSource.kt:69)
	at okio.Buffer.writeAll(Buffer.kt:1642)
	at okio.RealBufferedSource.readByteArray(RealBufferedSource.kt:236)
	at okhttp3.ResponseBody.bytes(ResponseBody.kt:124)
```

steve-chancetop:
Thank you for your updates. This is not easy to reproduce. But I'd prefer to provide it, if there is any more detail to reproduce it.

yschimke:
Thanks, please reopen with more detail.

PR #7181, opened 2022-03-24
Remove superfluous throw

realdadfish:


JakeWharton (approved):

swankjesse (commented):
I don't think this throw is superfluous; it tells both the reader and the compiler that control flow doesn't continue beyond invalidLengths. In particular, a statement inserted on line 983 was unreachable; now it's not.

realdadfish:
The Nothing return type already told the compiler the same thing and therefor it spit out a warning on the call sites. TBH I would have found it more readable to just create the Exception in the method and not throw it inside there, but guessed this might be some kind of code style to follow.

yschimke:
@swankjesse that was my first thought, but then seeing the Nothing response, I figured inside the IDE it's still clear it can't proceed.

Your call, want to revert?

swankjesse:
Ahhh... returning nothing is enough, and my claim is wrong! Apologies for that! No action necessary.

Kinou90100:
#7183
#7181
#7178
#7023
#6518
@Christoc90

Issue #7186, opened 2022-03-26
IllegalStateException "Check failed" with Fast Fallback on 5.0.0-alpha.6 on Android


talklittle:
OkHttp 5.0.0-alpha.6

Fast fallback enabled.

Unfortunately I don't know how to reproduce the issue. Got a few crash logs, Android 11 and 12, and not specific to any device manufacturer.

Two slightly different stack traces. One mentions `RealRoutePlanner.planReusePooledConnection` and the other does not.

```
Fatal Exception: java.lang.IllegalStateException: Check failed.
       at okhttp3.internal.connection.RealCall.acquireConnectionNoEvents(RealCall.java:89)
       at okhttp3.internal.connection.RealConnectionPool.callAcquirePooledConnection(RealConnectionPool.java:52)
       at okhttp3.internal.connection.RealRoutePlanner.planReusePooledConnection$okhttp(RealRoutePlanner.java:32)
       at okhttp3.internal.connection.ConnectPlan.handleSuccess(ConnectPlan.java:21)
       at okhttp3.internal.connection.FastFallbackExchangeFinder.find(FastFallbackExchangeFinder.java:138)
       at okhttp3.internal.connection.RealCall.initExchange$okhttp(RealCall.java:29)
       at okhttp3.internal.connection.ConnectInterceptor.intercept(ConnectInterceptor.java:11)
       at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:166)
       at okhttp3.internal.cache.CacheInterceptor.intercept(CacheInterceptor.java:194)
       at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:166)
       at okhttp3.internal.http.BridgeInterceptor.intercept(BridgeInterceptor.java:167)
       at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:166)
       at okhttp3.internal.http.RetryAndFollowUpInterceptor.intercept(RetryAndFollowUpInterceptor.java:34)
       at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:166)
       at okhttp3.internal.connection.RealCall.getResponseWithInterceptorChain$okhttp(RealCall.java:113)
       at okhttp3.internal.connection.RealCall$AsyncCall.run(RealCall.java:33)
       at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
[...]

PR #7190, opened 2022-03-28
Prefer reusing the call connection over deferred plans

swankjesse:
I still need to figure out a test for this.

Closes: https://github.com/square/okhttp/issues/7186


yschimke (approved):
Looks reasonable and tests are good.  But I'll have to trust you on the logic.

I don't specifically follow this "Prefer reusing the call connection over deferred plans" as it seems to mostly prefer successors to race winners.

swankjesse:
Lemme merge this and ship a new alpha. I’ll also try to set up a test, though my test infra doesn’t really embrace it.

PR #7196, opened 2022-03-29
Ignore early hints for 4.9.x

unlucku:
Does the rfc mention if :status: has to be the first response header or not? If it does not, then a for loop to look for the status might be needed. 

Alternatively this change can be put in Http2Connection.headers and we can call .get(":status:") on the headers but I think that might end up requiring more changes than it should.

### DIFF
--- a/okhttp/src/main/kotlin/okhttp3/internal/http2/Http2Reader.kt
+++ b/okhttp/src/main/kotlin/okhttp3/internal/http2/Http2Reader.kt
@@ -65,7 +65,7 @@ class Http2Reader(
       source = continuation,
       headerTableSizeSetting = 4096
   )
-
+  private const val EARLY_HINTS_CODE = "103"
   @Throws(IOException::class)
   fun readConnectionPreface(handler: Handler) {
     if (client) {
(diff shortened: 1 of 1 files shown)


swankjesse:
We’re gonna revamp 1xx handling in 5.0.0. I think we get that working with our new 1xx testing stuff in MockWebServer, then backport to 4.9.x.

FYI @yschimke

okhttp/src/main/kotlin/okhttp3/internal/http2/Http2Reader.kt
```diff
@@ -65,7 +65,7 @@ class Http2Reader(
       source = continuation,
       headerTableSizeSetting = 4096
   )
-
+  private const val EARLY_HINTS_CODE = "103"
```
yschimke, line 68: This needs to be a ByteString.  

```suggestion
  private const val EARLY_HINTS_CODE = "103"
```

yschimke:
I don't think the plan is to land this, but it didn't fix the issue, so probably indicates we need to be able to test it to land the eventual fix.

okhttp/src/main/kotlin/okhttp3/internal/http2/Http2Reader.kt
```diff
@@ -146,7 +146,10 @@ class Http2Reader(
     headerBlockLength = lengthWithoutPadding(headerBlockLength, flags, padding)
     val headerBlock = readHeaderBlock(headerBlockLength, padding, flags, streamId)
 
-    handler.headers(endStream, streamId, -1, headerBlock)
+    // Ignore early hints
+	  if (!(headerBlockLength != 0 && headerBlock[0].value.equals(EARLY_HINTS_CODE))) {
```
yschimke, line 150: It doesn't seem like ":status" has to be the first header. Will just check it's the ":status" but otherwise leave as is.

yschimke:
Rebased onto 4.10.x and fixed the bug plus adding a test

https://github.com/square/okhttp/pull/7443

okhttp/src/main/kotlin/okhttp3/internal/http2/Http2Reader.kt
yschimke: Mentioned in description.

PR #7205, opened 2022-04-02
Make Response.body non-null

swankjesse:
This is a source-incompatible API change. In particular, Response.body(null)
is no longer source-compatible for Kotlin source.

The upside is tremendous: no need for callers to use !! on Response.body
on every single API call. In the rare cases where a Response doesn't have
a body we use a runtime error. This is unlikely to cause problems in
practice; users don't have reason to read the response body on supporting
responses.

### DIFF
--- a/okhttp/src/commonMain/kotlin/okhttp3/internal/-ResponseCommon.kt
+++ b/okhttp/src/commonMain/kotlin/okhttp3/internal/-ResponseCommon.kt
@@ -19,6 +19,7 @@ package okhttp3.internal
 import kotlin.jvm.JvmOverloads
 import okhttp3.CacheControl
 import okhttp3.Headers
+import okhttp3.MediaType
 import okhttp3.Protocol
 import okhttp3.Request
 import okhttp3.Response
@@ -30,9 +31,44 @@ import okhttp3.internal.http.HTTP_MULT_CHOICE
 import okhttp3.internal.http.HTTP_PERM_REDIRECT
 import okhttp3.internal.http.HTTP_SEE_OTHER
 import okhttp3.internal.http.HTTP_TEMP_REDIRECT
-
 import okio.Buffer
 import okio.IOException
+import okio.Source
+import okio.Timeout
+import okio.buffer
+
+class UnreadableResponseBody(
+  private val mediaType: MediaType?,
+  private val contentLength: Long,
+): ResponseBody(), Source {
+  override fun contentType() = mediaType
+  override fun contentLength() = contentLength
+  override fun source() = buffer()
+
+  override fun read(sink: Buffer, byteCount: Long): Long {
+    throw IllegalStateException("""
+      |Unreadable ResponseBody! These Response objects have bodies that are stripped:
+      | * Response.cacheResponse
+      | * Response.networkResponse
+      | * Response.priorResponse
+      | * EventSourceListener
+      | * WebSocketListener
+      |(It is safe to call contentType() and contentLength() on these response bodies.)
+      """.trimMargin()
+    )
+  }
+
+  override fun timeout() = Timeout.NONE
+
+  override fun close() {
+  }
+}
+
+fun Response.stripBody(): Response {
+  return newBuilder()
+    .body(UnreadableResponseBody(body.contentType(), body.contentLength()))
+    .build()
+}
 
 val Response.commonIsSuccessful: Boolean
   get() = code in 200..299
@@ -44,7 +80,7 @@ fun Response.commonHeader(name: String, defaultValue: String?): String? = header
 
 @Throws(IOException::class)
 fun Response.commonPeekBody(byteCount: Long): ResponseBody {
-  val peeked = body!!.source().peek()
+  val peeked = body.source().peek()
   val buffer = Buffer()
   peeked.request(byteCount)
   buffer.write(peeked, minOf(byteCount, peeked.buffer.size))
@@ -70,7 +106,7 @@ val Response.commonCacheControl: CacheControl
   }
 
 fun Response.commonClose() {
-  checkNotNull(body) { "response is not eligible for a body and must not be closed" }.close()
+  body.close()
 }
 
 fun Response.commonToString(): String =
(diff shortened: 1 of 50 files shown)


yschimke:
Is it binary compatible, I assume yes, doesn't change the signature?

okhttp/src/commonMain/kotlin/okhttp3/internal/-ResponseCommon.kt
```diff
@@ -30,9 +31,44 @@ import okhttp3.internal.http.HTTP_MULT_CHOICE
...
+
+  override fun timeout() = Timeout.NONE
+
+  override fun close() {
+  }
+}
+
+fun Response.stripBody(): Response {
+  return newBuilder()
+    .body(UnreadableResponseBody(body.contentType(), body.contentLength()))
```
yschimke, line 69: This is really nice.

okhttp/src/commonMain/kotlin/okhttp3/internal/-ResponseCommon.kt
```diff
@@ -30,9 +31,44 @@ import okhttp3.internal.http.HTTP_MULT_CHOICE
...
 import okhttp3.internal.http.HTTP_SEE_OTHER
 import okhttp3.internal.http.HTTP_TEMP_REDIRECT
-
 import okio.Buffer
 import okio.IOException
+import okio.Source
+import okio.Timeout
+import okio.buffer
+
+class UnreadableResponseBody(
```
yschimke, line 40: Should this type be public API, and possible stripBody? for others to use?

How does someone who previously checked the presence of a body before using it (I don't have a convincing example right now, because optionally consuming a body is weird), now check whether it will throw?

swankjesse:
It's binary compatible only because the JVM doesn't know about non-nullable objects.

okhttp/src/commonMain/kotlin/okhttp3/internal/-ResponseCommon.kt
```diff
@@ -30,9 +31,44 @@ import okhttp3.internal.http.HTTP_MULT_CHOICE
...
+  private val mediaType: MediaType?,
+  private val contentLength: Long,
+): ResponseBody(), Source {
+  override fun contentType() = mediaType
+  override fun contentLength() = contentLength
+  override fun source() = buffer()
+
+  override fun read(sink: Buffer, byteCount: Long): Long {
+    throw IllegalStateException("""
```
yschimke, line 49: I agree it's a logic bug, so IllegalStateException is warranted.  My first reaction, was IOException is safer, but as I'm typing this I realise that would be confusing and probably handled in app code with a retry.

okhttp/src/jvmMain/kotlin/okhttp3/internal/cache/CacheInterceptor.kt
```diff
@@ -68,7 +68,6 @@ class CacheInterceptor(internal val cache: Cache?) : Interceptor {
           .protocol(Protocol.HTTP_1_1)
           .code(HTTP_GATEWAY_TIMEOUT)
           .message("Unsatisfiable Request (only-if-cached)")
-          .body(EMPTY_RESPONSE)
```
yschimke, line 71: This doesn't appear like a straight conversion, if the default is `EMPTY_BYTE_ARRAY.toResponseBody()`, doesn't this give the Response a body effectively?

yschimke (approved):
Great cleanup.  Anything else bugging you in 4.9.x?

okhttp/src/commonMain/kotlin/okhttp3/internal/-ResponseCommon.kt
swankjesse: Yeah very good question. I think it's better without a public API cause we don't want users to think they need to check this.

I think this is similar to the keySet of a MutableMap: you can't insert into it, but it'd be strange to be in a situation where you were even attempting that.

swankjesse:
Yeah!

We should add public constructors for Request and Response that short-circuit the Builders. Named parameters in Kotlin make this safe in a way it isn't in Java.

Also possibly defaulting more stuff in both those constructors and builders.

yschimke:
> Yeah!
> 
> We should add public constructors for Request and Response that short-circuit the Builders. Named parameters in Kotlin make this safe in a way it isn't in Java.
> 
> Also possibly defaulting more stuff in both those constructors and builders.

I thought Kotlin constructors with default parameters are still brittle if you add a new param, even if you give it a default.

It would be nice if there was a @Since() param with a @VersionedOverloads that generated the deprecated forms...

swankjesse:
For Request and Response the types map closely to the on-the-wire data, so we aren't taking a big risk by locking in the set of parameters.

yschimke:
> For Request and Response the types map closely to the on-the-wire data, so we aren't taking a big risk by locking in the set of parameters.

True, I might be concerned about multiplatform differences, features we add over time into the common.

```
  // fun challenges(): List<Challenge>

  // fun <T> tag(type: Class<out T>): T? = type.cast(tags[type])
```

And in practice I think a lot of user code just mutates via `newBuilder()`.

swankjesse:
Yep! I think the `Builder` remains primary, we just add a convenience with the core stuff.

okhttp/src/commonMain/kotlin/okhttp3/internal/-ResponseCommon.kt
JakeWharton: Is there a reason you left the type public instead of internal?

swankjesse: Nope. Fixed. https://github.com/square/okhttp/pull/7209

okhttp/src/jvmMain/kotlin/okhttp3/internal/cache/CacheInterceptor.kt
yschimke: @swankjesse want to flag this question.

swankjesse: Yeah the default is empty, not unreadable.

PR #7208, opened 2022-04-03
Add a proper Kotlin constructor for Request

swankjesse:
This turns out to be very useful throughout our test suite.


okhttp/src/jvmTest/java/okhttp3/RequestTest.java
```diff
@@ -1,371 +0,0 @@
...
-import java.util.UUID;
-import okio.Buffer;
-import okio.ByteString;
-import org.junit.jupiter.api.Test;
-
-import static java.nio.charset.StandardCharsets.UTF_8;
-import static org.assertj.core.api.Assertions.assertThat;
-import static org.junit.jupiter.api.Assertions.fail;
-
-public final class RequestTest {
```
swankjesse, line 31: Converted to Kotlin.

okhttp/src/jvmTest/java/okhttp3/RequestTest.java
```diff
@@ -1,371 +0,0 @@
...
-    } catch (IllegalArgumentException expected) {
-    }
-    try {
-      builder.addHeader("", "Value");
-      fail();
-    } catch (IllegalArgumentException expected) {
-    }
-  }
-
-  @Test public void headerForbidsNullArguments() {
```
swankjesse, line 193: This particular test dropped.

okhttp/src/jvmTest/java/okhttp3/RequestTest.kt
```diff
@@ -0,0 +1,448 @@
...
+import okhttp3.MediaType.Companion.toMediaType
+import okhttp3.RequestBody.Companion.asRequestBody
+import okhttp3.RequestBody.Companion.toRequestBody
+import okio.Buffer
+import okio.ByteString.Companion.encodeUtf8
+import org.assertj.core.api.Assertions.assertThat
+import org.assertj.core.api.Assertions.fail
+import org.junit.jupiter.api.Test
+
+class RequestTest {
```
swankjesse, line 33: Converted from Java. The first 4 tests are new.

yschimke:
```

RequestTest[jvm] > multipleTags()[jvm] FAILED
    java.lang.AssertionError: 
    Expecting actual:
      20170815L
    and actual:
      20170815L
    to refer to the same object
        at okhttp3.RequestTest.multipleTags(RequestTest.kt:427)
```

okhttp-android/src/androidTest/kotlin/okhttp3/android/AndroidAsyncDnsTest.kt
```diff
@@ -94,7 +95,7 @@ class AndroidAsyncDnsTest(val server: MockWebServer) {
 
   @Test
   fun testRequestInvalid() {
-    val call = client.newCall(Request.Builder().url("https://google.invalid/").build())
+    val call = client.newCall(Request("https://google.invalid/".toHttpUrl()))
```
yschimke, line 98: I think I know the answer here...

But what about

```
"https://google.invalid/".toHttpUrl().toRequest()
"https://google.invalid/".toHttpUrl().toRequest(method="POST")
```

with a default for method?

okhttp/src/jvmMain/kotlin/okhttp3/Request.kt
```diff
@@ -32,18 +32,52 @@ import okhttp3.internal.commonPut
...
+   * Constructs a new request.
+   *
+   * Use [Builder] for more fluent construction, including helper methods for various HTTP methods.
+   *
+   * @param method defaults to "GET" if [body] is null, and "POST" otherwise.
+   */
+  constructor(
+    url: HttpUrl,
+    headers: Headers = Headers.headersOf(),
+    method: String = "\u0000", // Sentinel value chooses based on what the body is.
```
yschimke, line 65: Cute, was a bit against it at first, but it hits the sweet spot for usage.

yschimke: It also removes the comment I came here to make, that HttpMethods should be public API, since you need to pass in here.  But not an issue since this can be left out mostly.

yschimke (approved):
My concerns with the Constructor were just around additional params, but it seems to satisfy the typical cases, and use the Builder for anything more exotic (tags, ).

yschimke:
> ```
> RequestTest[jvm] > multipleTags()[jvm] FAILED
>     java.lang.AssertionError: 
>     Expecting actual:
>       20170815L
>     and actual:
>       20170815L
>     to refer to the same object
>         at okhttp3.RequestTest.multipleTags(RequestTest.kt:427)
> ```


    val longTag: Long? = 20170815L

okhttp-android/src/androidTest/kotlin/okhttp3/android/AndroidAsyncDnsTest.kt
swankjesse: It feels just slightly like a layering violation. I don't think `.toXxx()` methods are as obvious when they take ‘important’ parameters?

(I can't necessary defend this position while justifying `toHttpUrl()` and `toRequestBody()`, but this one feels off.)

okhttp/src/jvmMain/kotlin/okhttp3/Request.kt
swankjesse: I considered `null` as the sentinel, but that lets users pass null which I don’t like.

I considered making the `when` expression the default value, but that puts the parameters in a surprising order.

This is ugly, but users shouldn’t see it.

swankjesse: Ooooh think of the poor C programmers, strlen on this guy isn't fun.

KikilLove22 (commented):
okhttp/src/jvmMain/kotlin/okhttp3/Cache.kt

PR #7218, opened 2022-04-06
Support request priority

### LINKED ISSUE #404: Implement RFC 9113 stream priorities for HTTP/2
At the moment we don't honor stream priority, and we don't allow app developers to prioritize their streams.

We should.

tom5079:
Currently, `Dispatcher` executes calls First-In-First-Out.  
This PR adds a way for users to customize what calls to execute next.  
resolves #404.

### DIFF
--- a/okhttp/src/jvmMain/kotlin/okhttp3/Dispatcher.kt
+++ b/okhttp/src/jvmMain/kotlin/okhttp3/Dispatcher.kt
@@ -18,6 +18,7 @@ package okhttp3
 import java.util.ArrayDeque
 import java.util.Collections
 import java.util.Deque
+import java.util.LinkedList
 import java.util.concurrent.ExecutorService
 import java.util.concurrent.SynchronousQueue
 import java.util.concurrent.ThreadPoolExecutor
@@ -98,8 +99,20 @@ class Dispatcher() {
       return executorServiceOrNull!!
     }
 
+
+  @get:Synchronized
+  @set:Synchronized
+  @get:JvmName("priorityOrder")
+    /**
+     * A comparator that determines the priority of the async calls.
+     * A call with the highest priority will be executed first.
+     *
+     * When [priorityOrder] is null or returns 0, calls that are enqueued first have priority.
+     */
+  var priorityOrder: Comparator<Call>? = null
+
   /** Ready async calls in the order they'll be run. */
-  private val readyAsyncCalls = ArrayDeque<AsyncCall>()
+  private val readyAsyncCalls = LinkedList<AsyncCall>()
 
   /** Running asynchronous calls. Includes canceled calls that haven't finished yet. */
   private val runningAsyncCalls = ArrayDeque<AsyncCall>()
@@ -113,13 +126,26 @@ class Dispatcher() {
 
   internal fun enqueue(call: AsyncCall) {
     synchronized(this) {
-      readyAsyncCalls.add(call)
-
       // Mutate the AsyncCall so that it shares the AtomicInteger of an existing running call to
       // the same host.
       if (!call.call.forWebSocket) {
         val existingCall = findExistingCallWithHost(call.host)
         if (existingCall != null) call.reuseCallsPerHostFrom(existingCall)
+      }
+
+      val priorityOrder = priorityOrder
+
+      if (priorityOrder == null) readyAsyncCalls.addLast(call) else {
+        val i = readyAsyncCalls.listIterator()
+
+        while (i.hasNext()) {
+          if (priorityOrder.compare(i.next().call, call.call) < 0) {
+            i.previous()
+            break
+          }
+        }
+
+        i.add(call)
       }
     }
     promoteAndExecute()
@@ -164,6 +190,12 @@ class Dispatcher() {
     val executableCalls = mutableListOf<AsyncCall>()
     val isRunning: Boolean
     synchronized(this) {
+      val priorityOrder = priorityOrder
+
+      if (priorityOrder != null) {
+        readyAsyncCalls.sortWith { a, b -> priorityOrder.compare(b.call, a.call) } // sort in descending order
+      }
+
       val i = readyAsyncCalls.iterator()
       while (i.hasNext()) {
         val asyncCall = i.next()
--- a/okhttp/src/jvmMain/kotlin/okhttp3/Dispatcher.kt
+++ b/okhttp/src/jvmMain/kotlin/okhttp3/Dispatcher.kt
@@ -99,20 +99,17 @@ class Dispatcher() {
       return executorServiceOrNull!!
     }
 
-
+  /**
+   * Orders calls by order of execution. When two calls have equal priority order they are executed
+   * in the order that they were enqueued in.
+   */
   @get:Synchronized
   @set:Synchronized
   @get:JvmName("priorityOrder")
-    /**
-     * A comparator that determines the priority of the async calls.
-     * A call with the highest priority will be executed first.
-     *
-     * When [priorityOrder] is null or returns 0, calls that are enqueued first have priority.
-     */
   var priorityOrder: Comparator<Call>? = null
 
   /** Ready async calls in the order they'll be run. */
-  private val readyAsyncCalls = LinkedList<AsyncCall>()
+  private val readyAsyncCalls = mutableListOf<AsyncCall>()
 
   /** Running asynchronous calls. Includes canceled calls that haven't finished yet. */
   private val runningAsyncCalls = ArrayDeque<AsyncCall>()
@@ -126,26 +123,13 @@ class Dispatcher() {
 
   internal fun enqueue(call: AsyncCall) {
     synchronized(this) {
+      readyAsyncCalls.add(call)
+
       // Mutate the AsyncCall so that it shares the AtomicInteger of an existing running call to
       // the same host.
       if (!call.call.forWebSocket) {
         val existingCall = findExistingCallWithHost(call.host)
         if (existingCall != null) call.reuseCallsPerHostFrom(existingCall)
-      }
-
-      val priorityOrder = priorityOrder
-
-      if (priorityOrder == null) readyAsyncCalls.addLast(call) else {
-        val i = readyAsyncCalls.listIterator()
-
-        while (i.hasNext()) {
-          if (priorityOrder.compare(i.next().call, call.call) < 0) {
-            i.previous()
-            break
-          }
-        }
-
-        i.add(call)
       }
     }
     promoteAndExecute()
@@ -190,10 +174,8 @@ class Dispatcher() {
     val executableCalls = mutableListOf<AsyncCall>()
     val isRunning: Boolean
     synchronized(this) {
-      val priorityOrder = priorityOrder
-
-      if (priorityOrder != null) {
-        readyAsyncCalls.sortWith { a, b -> priorityOrder.compare(b.call, a.call) } // sort in descending order
+      priorityOrder?.let { comparator ->
+        readyAsyncCalls.sortWith { a, b -> comparator.compare(b.call, a.call) } // sort in descending order
       }
 
       val i = readyAsyncCalls.iterator()
--- a/okhttp/src/jvmMain/kotlin/okhttp3/Dispatcher.kt
+++ b/okhttp/src/jvmMain/kotlin/okhttp3/Dispatcher.kt
@@ -175,7 +175,7 @@ class Dispatcher() {
     val isRunning: Boolean
     synchronized(this) {
       priorityOrder?.let { comparator ->
-        readyAsyncCalls.sortWith { a, b -> comparator.compare(b.call, a.call) } // sort in descending order
+        readyAsyncCalls.sortWith { a, b -> comparator.compare(a.call, b.call) }
       }
 
       val i = readyAsyncCalls.iterator()
--- a/okhttp/src/jvmTest/java/okhttp3/DispatcherTest.java
+++ b/okhttp/src/jvmTest/java/okhttp3/DispatcherTest.java
@@ -54,7 +54,7 @@ public final class DispatcherTest {
     float p2 = 0;
     if (t2 instanceof Priority) p2 = ((Priority) t2).priority;
 
-    return Float.compare(p1, p2);
+    return Float.compare(p2, p1);
   };
 
   @BeforeEach public void setUp() throws Exception {
--- a/okhttp/src/jvmTest/java/okhttp3/DispatcherTest.java
+++ b/okhttp/src/jvmTest/java/okhttp3/DispatcherTest.java
@@ -2,6 +2,7 @@ package okhttp3;
 
 import java.io.IOException;
 import java.io.InterruptedIOException;
+import java.util.Comparator;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.atomic.AtomicBoolean;
 import org.junit.jupiter.api.BeforeEach;
@@ -26,6 +27,35 @@ public final class DispatcherTest {
       .dispatcher(dispatcher)
       .eventListenerFactory(clientTestRule.wrap(listener))
       .build();
+
+  static class Priority {
+    private final float priority;
+
+    Priority() {
+      priority = 0;
+    }
+
+    Priority(float priority) {
+      this.priority = priority;
+    }
+
+    float priority() {
+      return priority;
+    }
+  }
+
+  final Comparator<Call> priorityOrder = (c1, c2) -> {
+    Object t1 = c1.request().tag();
+    Object t2 = c2.request().tag();
+
+    float p1 = 0;
+    if (t1 instanceof Priority) p1 = ((Priority) t1).priority;
+
+    float p2 = 0;
+    if (t2 instanceof Priority) p2 = ((Priority) t2).priority;
+
+    return Float.compare(p1, p2);
+  };
 
   @BeforeEach public void setUp() throws Exception {
     dispatcher.setMaxRequests(20);
@@ -311,6 +341,72 @@ public final class DispatcherTest {
         .containsExactly("CallStart", "CallStart", "CallFailed");
   }
 
+  @Test public void priorityAsyncCallDispatcher() throws Exception {
+    dispatcher.setMaxRequests(3);
+    dispatcher.setPriorityOrder(priorityOrder);
+
+    client.newCall(newRequest("http://a/1")).enqueue(callback);
+    client.newCall(newRequest("http://a/2")).enqueue(callback);
+    client.newCall(newRequest("http://a/3")).enqueue(callback);
+    client.newCall(newRequest("http://a/4")).enqueue(callback);
+    client.newCall(newRequest("http://a/5")).enqueue(callback);
+    client.newCall(newRequest("http://b/1", -1)).enqueue(callback);
+    client.newCall(newRequest("http://b/2", Float.MAX_VALUE)).enqueue(callback);
+    client.newCall(newRequest("http://b/3", 1)).enqueue(callback);
+
+    executor.assertJobs("http://a/1", "http://a/2", "http://a/3");
+
+    executor.finishJob("http://a/1");
+    executor.assertJobs("http://a/2", "http://a/3", "http://b/2");
+
+    executor.finishJob("http://a/2");
+    executor.assertJobs("http://a/3", "http://b/2", "http://b/3");
+
+    executor.finishJob("http://a/3");
+    executor.assertJobs("http://b/2", "http://b/3", "http://a/4");
+
+    executor.finishJob("http://b/2");
+    executor.assertJobs("http://b/3", "http://a/4", "http://a/5");
+
+    executor.finishJob("http://b/3");
+    executor.assertJobs("http://a/4", "http://a/5", "http://b/1");
+
+    executor.finishJob("http://a/4");
+    executor.assertJobs("http://a/5", "http://b/1");
+  }
+
+  @Test public void priorityAsyncCallDispatcherWithMaxRequestsPerHost() {
+    dispatcher.setMaxRequests(3);
+    dispatcher.setMaxRequestsPerHost(2);
+    dispatcher.setPriorityOrder(priorityOrder);
+
+    client.newCall(newRequest("http://a/1")).enqueue(callback);
+    client.newCall(newRequest("http://a/2")).enqueue(callback);
+    client.newCall(newRequest("http://a/3")).enqueue(callback);
+    client.newCall(newRequest("http://a/4")).enqueue(callback);
+    client.newCall(newRequest("http://b/1")).enqueue(callback);
+    client.newCall(newRequest("http://b/2", 1)).enqueue(callback);
+    client.newCall(newRequest("http://b/3", 1)).enqueue(callback);
+    client.newCall(newRequest("http://b/4", 1)).enqueue(callback);
+
+    executor.assertJobs("http://a/1", "http://a/2", "http://b/1");
+
+    executor.finishJob("http://a/1");
+    executor.assertJobs("http://a/2", "http://b/1", "http://b/2");
+
+    executor.finishJob("http://a/2");
+    executor.assertJobs("http://b/1", "http://b/2", "http://a/3");
+
+    executor.finishJob("http://b/2");
+    executor.assertJobs("http://b/1", "http://a/3", "http://b/3");
+
+    executor.finishJob("http://b/1");
+    executor.assertJobs("http://a/3", "http://b/3", "http://b/4");
+
+    executor.finishJob("http://b/4");
+    executor.assertJobs("http://a/3", "http://b/3", "http://a/4");
+  }
+
   private Thread makeSynchronousCall(Call call) {
     Thread thread = new Thread(() -> {
       try {
@@ -329,5 +425,9 @@ public final class DispatcherTest {
 
   private Request newRequest(String url, String tag) {
     return new Request.Builder().url(url).tag(tag).build();
+  }
+
+  private Request newRequest(String url, float priority) {
+    return new Request.Builder().url(url).tag(new Priority(priority)).build();
   }
 }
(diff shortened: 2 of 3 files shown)


okhttp/src/jvmMain/kotlin/okhttp3/Dispatcher.kt
```diff
@@ -98,8 +108,21 @@ class Dispatcher() {
       return executorServiceOrNull!!
     }
 
-  /** Ready async calls in the order they'll be run. */
-  private val readyAsyncCalls = ArrayDeque<AsyncCall>()
+  private var asyncCallDispatcherOrNull: AsyncCallDispatcher? = null
```
yschimke, line 111: Why null here? Can you move this logic into the constructor?

yschimke, line 111: Why null here? Can you move this logic into the constructor?

okhttp/src/jvmMain/kotlin/okhttp3/Dispatcher.kt
```diff
@@ -164,12 +196,27 @@ class Dispatcher() {
     val executableCalls = mutableListOf<AsyncCall>()
     val isRunning: Boolean
     synchronized(this) {
-      val i = readyAsyncCalls.iterator()
-      while (i.hasNext()) {
-        val asyncCall = i.next()
+      while (readyAsyncCalls.isNotEmpty()) {
+        val filteredCalls = readyAsyncCalls.mapNotNull {
+          if (it.callsPerHost.get() < maxRequestsPerHost) it.call else null
```
yschimke, line 201: This feels awfully close to request priorities.  Should this limit be a part of the AsycCallDispatcher functionality, allowing to execute high priority requests above the standard limit?

okhttp/src/jvmMain/kotlin/okhttp3/Dispatcher.kt
```diff
@@ -164,12 +196,27 @@ class Dispatcher() {
...
+        }
+
+        if (filteredCalls.isEmpty()) break // Max host capacity for all requests.
+
+        val nextCall = asyncCallDispatcher.nextCall(filteredCalls)
```
yschimke, line 206: I worry about putting user logic in this hot path.  Potentially that needs to process through large list of items, on every pass through promoteAndExecute.  I wonder whether a more constrained version without the same risk, is to return a priority for each item, with a default returning 0 for all.  These have a priority queue for readyAsyncCalls.

yschimke, line 206: I worry about putting user logic in this hot path.  Potentially that needs to process through large list of items, on every pass through promoteAndExecute.  I wonder whether a more constrained version without the same risk, is to return a priority for each item, with a default returning 0 for all.  These have a priority queue for readyAsyncCalls.

yschimke:
Before addressing any of these comments, I suggest waiting for a signal from @swankjesse.

This looks useful and inline with functionality we should support, but he may have plans in this area.  We generally advise reaching out before starting on a PR, to avoid frustration this way.

yschimke:
Before addressing any of these comments, I suggest waiting for a signal from @swankjesse.

This looks useful and inline with functionality we should support, but he may have plans in this area.  We generally advise reaching out before starting on a PR, to avoid frustration this way.

okhttp/src/jvmMain/kotlin/okhttp3/Dispatcher.kt
```diff
@@ -111,6 +134,15 @@ class Dispatcher() {
     this.executorServiceOrNull = executorService
   }
 
+  constructor(asyncCallDispatcher: AsyncCallDispatcher) : this() {
```
yschimke, line 137: rather than this constructor and the next, how about a single one with default null param values?  This pattern doesn't scale when we add a 3rd or 4th constructor param.

yschimke, line 137: rather than this constructor and the next, how about a single one with default null param values?  This pattern doesn't scale when we add a 3rd or 4th constructor param.

okhttp/src/jvmMain/kotlin/okhttp3/Dispatcher.kt
```diff
@@ -164,12 +196,27 @@ class Dispatcher() {
...
+
+        if (filteredCalls.isEmpty()) break // Max host capacity for all requests.
+
+        val nextCall = asyncCallDispatcher.nextCall(filteredCalls)
+
+        val i = readyAsyncCalls.iterator()
+
+        var asyncCall: AsyncCall = i.next()
+
+        while (asyncCall.call != nextCall && i.hasNext()) {
```
yschimke, line 212: This looks accidentally quadratic. We should avoid this second pass.  

Not sure what the answer is, allowing it? a more optimised filteredCalls? state on AsyncCall to check?

tom5079: I think quadratic runtime here is not a big deal because the outer loop runs until either `readyAsyncCalls` is empty or `runningAsyncCalls.size >= maxRequests`. `maxRequests` is 64 by default so even in the worst case the outer loop will only run 64 times.

Also I think the run time added by this check is fairly negligible compared to the network response time.

tom5079: Also I think achieving linear time is impossible because we have to convert `AsyncCall` to `Call` anyways unless you are okay with exposing `AsyncCall`.

okhttp/src/jvmMain/kotlin/okhttp3/Dispatcher.kt
tom5079: I am not sure what limit you are talking about.  
If you mean `mapNotNull` part, it's for enforcing `maxRequestsPerHost`.

okhttp/src/jvmMain/kotlin/okhttp3/Dispatcher.kt
tom5079: I agree with you, but isn't returning priority for each item essentially the same as just returning next call to execute? Determining the priority takes just as long as just returning the call.

yschimke: It qould mean you could do it once on insertion. Or is dynamic priority important here?

okhttp/src/jvmMain/kotlin/okhttp3/Dispatcher.kt
yschimke: Limit being maxRequestsPerHost. Essentially this limit causes priorities to break. Probably unavoidable, but I figured worth discussing whether high priority requests should override the limit?

okhttp/src/jvmMain/kotlin/okhttp3/Dispatcher.kt
tom5079: Yes, at least for my use case. I am trying to load a bunch of images and prioritize images that are visible to the user and that requires dynamic priority. Maybe we can combine both methods to get an optimal result for everyone?

okhttp/src/jvmMain/kotlin/okhttp3/Dispatcher.kt
tom5079: I don't think it shouldn't and you can see it in the (test case)[https://github.com/square/okhttp/pull/7218/files#diff-451df978a3ed8616ddb81d91165eab2cfd3bddda7bdbf7c68dcda8aac187206bR371]. I think it should work like there is a queue for every host.

okhttp/src/jvmMain/kotlin/okhttp3/Dispatcher.kt
tom5079: Hmm, how about running `promoteAndExecute` on another thread? (possibly using coroutines)  
running this every time user calls `enqueue` sounds quite problematic.

yschimke: I'd suggest waiting for other input, particularly @swankjesse. The more we tweak this the more frustrating it will be if it requires a different approach.

swankjesse:
Could we accomplish what you need by adding this API?


```
class Dispatcher {
  ...
  var priorityOrder: Comparator<Call>?
}
```

Internally we could sort by priority order, either at insert time or at start time. Sorting a nearly-sorted list is fast so we could even do both.

okhttp/src/jvmMain/kotlin/okhttp3/Dispatcher.kt
yschimke: OK, makes sense.

yschimke:
> Could we accomplish what you need by adding this API?
> 
> ```
> class Dispatcher {
>   ...
>   var priorityOrder: Comparator<Call>?
> }
> ```
> 
> Internally we could sort by priority order, either at insert time or at start time. Sorting a nearly-sorted list is fast so we could even do both.

I like this, it is probably possible to get into a weird state with sorting based on an unstable comparator.  Is there a sensible way to avoid this for comparators that are unstable like this? That they only change within a lock on themselves, so we can sort a collection without the results changing during it?

Or am I over worrying about this?

tom5079:
> I like this, it is probably possible to get into a weird state with sorting based on an unstable comparator.
I think you can solve this by copying an instance to val before sorting

@swankjesse yes, I think that would work.

tom5079:
@yschimke the comparator is accessed inside a synchronized block so I don't think it can change during the comparison

@swankjesse yes, that would work 👍

tom5079:
On a second thought, I'm not sure if it is going to be any faster.
```kotlin
fun promoteAndExecute() {
   // ...
    while (readyAsyncCalls.isNotEmpty() && runningAsyncCalls.size >= this.maxRequests) {
      val priorityOrder = priorityOrder

      if (priorityOrder != null) {
        Collections.sort(readyAsyncCalls, Comparator { a, b -> priorityOrder.compare(a.call, b.call) })
      }

      val nextCall = readyAsyncCalls.removeLast()
    
      // enqueue call
    }
}
```
I'm not sure this is what you intended, but `Collections.sort` copies entire list to array, sorts it, and then copies sorted array to the list which is not that different from what we have. It still takes linear time at minimum, and probably more for average cases.  

It alleviates @yschimke's concern of putting user code in a hot path though. How about we provide default code for `AsyncCallDispatcher`?

```kotlin
open class AsyncCallDispatcher(priorityOrder: Comparator<Call>?) {
    open fun nextCall(list: List<Call>): Call {
        // default logic that determines call with max priority
    }
}
```

yschimke:
> On a second thought, I'm not sure if it is going to be any faster.

Can you sort once in promoteAndExecute instead of on each loop? And conditionally on a priority order being configured?

tom5079:
@yschimke got it, I'll change it that way.

tom5079:
One question: I'm getting this weird lint error. Does anyone know how to fix this?
image

okhttp/src/jvmTest/java/okhttp3/DispatcherTest.java
```diff
@@ -27,6 +32,24 @@ public final class DispatcherTest {
       .eventListenerFactory(clientTestRule.wrap(listener))
       .build();
 
+  final @Nullable Comparator<Call> priorityOrder = (c1, c2) -> {
+    final String priorityTag = "priority";
+
+    Object t1 = c1.request().tag();
```
yschimke, line 38: For this example can we use a typed tag? This form is a single tag per request, better to show with a separate priority tag.

okhttp/src/jvmMain/kotlin/okhttp3/internal/-UtilJvm.kt
```diff
@@ -322,3 +325,12 @@ internal inline fun Any.assertThreadDoesntHoldLock() {
...
   }
 }
+
+internal fun threadPoolExecutor(
+  corePoolSize: Int = 0,
+  maximumPoolSize: Int = Int.MAX_VALUE,
+  keepAliveTime: Long = 60,
+  timeUnit: TimeUnit = TimeUnit.SECONDS,
+  workingQueue: BlockingQueue<Runnable> = SynchronousQueue(),
+  threadFactory: ThreadFactory = threadFactory("$okHttpName Dispatcher", false)
```
yschimke, line 335: This feels wrong with Dispatcher in the thread name. Can you revert this part of the change and try to focus the PR on just new functionality?

yschimke:
> One question: I'm getting this weird lint error. Does anyone know how to fix this? 

What does the lint say? Can you try `dispatcher.priorityOrder = priorityOrder`

okhttp/src/jvmMain/kotlin/okhttp3/Dispatcher.kt
```diff
@@ -86,41 +86,49 @@ class Dispatcher() {
...
-    }
+  @set:Synchronized
+  @get:JvmName("priorityOrder")
+    /**
+     * A comparator that determines the priority of the async calls.
+     * A call with the highest priority will be executed first.
+     *
+     * When [priorityOrder] is null or returns 0, calls that are enqueued first have priority.
+     */
+  var priorityOrder: Comparator<Call>? = null
```
yschimke, line 98: @swankjesse Dispatcher is already a mutable object, this follows that pattern.  Do we want to continue that?

Could we deprecate the mutable features and move towards a Dispatcher Builder (in a follow up PR?)

tom5079:
> What does the lint say? Can you try dispatcher.priorityOrder = priorityOrder

The lint has no message, and I can't use `dispatcher.priorityOrder = priorityOrder` because I'm calling it from Java.

okhttp/src/jvmMain/kotlin/okhttp3/Dispatcher.kt
```diff
@@ -86,41 +86,49 @@ class Dispatcher() {
...
 
   /** Ready async calls in the order they'll be run. */
-  private val readyAsyncCalls = ArrayDeque<AsyncCall>()
+  private val readyAsyncCalls = LinkedList<AsyncCall>()
```
yschimke, line 101: I wonder whether this might be simpler with a very small DispatchQueue model.

say

```
interface DispatchQueue {
  fun insert(call: AsyncCall)
  fun iterator(): Iterator<AsyncCall>
  fun sort()
}

StandardDispatchQueue: DispatchQueue {
  private val calls = ArrayDeque<AsyncCall>()

  fun insert(call: AsyncCall) {
    calls.add(call)
  }

  fun iterator(): Iterator<AsyncCall> {
    return calls.iterator()
  }

  fun sort() {
    // Nothing
  }
}

PriorityDispatchQueue(private val comparator = Comparator<Call>): DispatchQueue {
  private var calls = LinkedList<AsyncCall>()

  fun insert(call: AsyncCall) {
    val pos = l.binarySearch(call, comparator)
    if (pos >= 0) {
      // walk forward to find position
      l.add(finalPos, call)
    } else {
      l.add(-pos - 1, call)
    }
  }

  fun iterator(): Iterator<AsyncCall> {
    return calls.iterator()
  }

  fun sort() {
    calls.sort(comparator)
  }
}
```

Two reasons

1) We can remain reasonably confident that the existing behaviour remains, and is no more expensive.
2) We can optimise and test this code separately, and avoid walking the entire list if big.  If small, it's hopefully as efficient. Currently the Dispatcher assumes all the new complexity, it would be nice to leave it as simple or simpler than before.

Thoughts? Code above is rough and probably doesn't even compile.

yschimke:
@tom5079 I hope you don't mind the back and forth reviews.  This is new public API and important to get it right.

But thanks for following up on this, it looks like a really compelling feature for 5.0.

okhttp/src/jvmMain/kotlin/okhttp3/internal/-UtilJvm.kt
tom5079: should I revert the constructor and `executorServiceOrNull` too? or just this function?

tom5079:
@yschimke yea, I kind of expected that my first version probably won't be accepted. I'm more than happy to work though this, all I need is a decent priority api.

tom5079:
I'm thinking maybe I should change this to draft PR. Any thoughts?

yschimke:
> I'm thinking maybe I should change this to draft PR. Any thoughts?

Makes sense, but it's not going to accidentally land :) Your call.

okhttp/src/jvmMain/kotlin/okhttp3/Dispatcher.kt
tom5079: @yschimke I think you can't binarySearch over LinkedList, only linear search works. Even if you copy the whole list to array and find the index, you need to traverse to that index anyways.

okhttp/src/jvmMain/kotlin/okhttp3/internal/-UtilJvm.kt
yschimke: Let's wait for @swankjesse, Dispatcher is different than our other types.  If we start adding new mutable fields, we should discuss where we are heading.

For now, maybe just move this to an internal method in Dispatcher.

okhttp/src/jvmTest/java/okhttp3/DispatcherTest.java
yschimke: @swankjesse as a follow up PR, what do you think about a simple batteries included Priority model with Comparator? Or leave as IKEA flatpack for now?

okhttp/src/jvmMain/kotlin/okhttp3/Dispatcher.kt
tom5079: Maybe I am not fully understanding how much this code should be able to handle. What is the worst case scenario this code has to handle? Until now I assumed maybe 1000 incoming calls and 100 maxRequests.

yschimke: > y understanding how much this code should be able to handle. What is the worst case scenario this code has to handle? Until now I assumed maybe 1000 incoming ca

Good point.  I just failed this interview :)

Maybe that's a good microbenchmark.  submit a 1000 requests and see how often comparator is called?

yschimke: Your usage, my usage doesn't necessarily represent all usage. I'd be concerned about apps using this with huge workloads and it spending any significant time here.  Also why I'm nervous about more open design like `nextCall(allCalls)`.

tom5079: > Also why I'm nervous about more open design like nextCall(allCalls).

yes, I totally understand you. I just wanted to know what use case you are considering and align our thoughts.

okhttp/src/jvmMain/kotlin/okhttp3/Dispatcher.kt
swankjesse: Yeah Dispatcher is mutable, and I’m not necessarily in a hurry to change that. I can imagine changing  `maxRequests` and `maxRequestsPerHost` dynamically in a running program.

swankjesse:
I added a commit here.
https://github.com/square/okhttp/compare/master...jwilson.0410.priority_ideas

okhttp/src/jvmTest/java/okhttp3/DispatcherTest.java
swankjesse: HTTP/2 priority is extremely powerful and in theory we could integrate that all the way to the server. If we support this, we also wanna leave room for that.

yschimke: OK separate tag, that maybe uses this feature.

tom5079:
Little question: Why are we sorting the list at all? I think we can we just scan through the array once and find n calls to execute (n being the number of slots left for request).

Something like this:
```
val promotedCalls = mutable list (probably LinkedList)

for calls in readyCalls:
    // if call has higher priority than call with lowest priority in promotedCalls, push to promotedCalls

// remove calls, append to executablecalls, etc..
```

swankjesse:
Yep. Sometimes we'll need to inspect a lot of calls before we find a runnable one.

yschimke (approved):

okhttp/src/jvmMain/kotlin/okhttp3/Dispatcher.kt
```diff
@@ -98,8 +99,17 @@ class Dispatcher() {
...
+   * in the order that they were enqueued in.
+   */
+  @get:Synchronized
+  @set:Synchronized
+  @get:JvmName("priorityOrder")
+  var priorityOrder: Comparator<Call>? = null
+
   /** Ready async calls in the order they'll be run. */
-  private val readyAsyncCalls = ArrayDeque<AsyncCall>()
+  private val readyAsyncCalls = mutableListOf<AsyncCall>()
```
yschimke, line 112: The only nagging doubt is the cost of remove for ArrayList vs ArrayDeque.  ArrayDeque is more optimal for elements near the start, which is typical... 

Removing the first element is O(1) vs O(n)

yschimke:
Failing on

```
Execution failed for task ':okhttp:jvmApiCheck'.
> API check failed for project okhttp.
  --- /home/runner/work/okhttp/okhttp/okhttp/api/okhttp.api
  +++ /home/runner/work/okhttp/okhttp/okhttp/build/api/okhttp.api
  @@ -470,7 +470,6 @@
   	public final fun -deprecated_executorService ()Ljava/util/concurrent/ExecutorService;
   	public fun <init> ()V
   	public fun <init> (Ljava/util/concurrent/ExecutorService;)V
  -	public synthetic fun <init> (Ljava/util/concurrent/ExecutorService;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
   	public final fun cancelAll ()V
   	public final fun executorService ()Ljava/util/concurrent/ExecutorService;
   	public final fun getIdleCallback ()Ljava/lang/Runnable;
  
   You can run :okhttp:apiDump task to overwrite API declarations
```

tom5079:
@yschimke fixed

okhttp/src/jvmMain/kotlin/okhttp3/Dispatcher.kt
tom5079: @yschimke I have to investigate but I've seen that sort function copies entire list to an array, sorts it, and then copies it back to original list.  
Maybe we can keep using array, but instrad of using iterator, we can use takeFirst() n and copy the rest of the array back.

tom5079: https://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/java/util/Collections.java#l164

yschimke: Yep - so it's better than a naive sort on a LinkedList, but it's still doing a lot more work when we consider we have a mostly sorted list already.

```
     * <p>The implementation was adapted from Tim Peters's list sort for Python
     * (<a href="http://svn.python.org/projects/python/trunk/Objects/listsort.txt">
     * TimSort</a>).  It uses techniques from Peter McIlroy's "Optimistic
     * Sorting and Information Theoretic Complexity", in Proceedings of the
     * Fourth Annual ACM-SIAM Symposium on Discrete Algorithms, pp 467-474,
     * January 1993.
     *
     * <p>This implementation dumps the specified list into an array, sorts
     * the array, and iterates over the list resetting each element
     * from the corresponding position in the array.  This avoids the
     * n<sup>2</sup> log(n) performance that would result from attempting
     * to sort a linked list in place.
```

yschimke: I'm not sure in practice whether this will be noticeable.  I wonder what performance testing square has, that would flush this out on Android and JVM.

If it's a concern and blocking, a possible way forward is the small domain model I suggested that would let us continue to use the current ArrayDeque when no priority comparator is set, and only take the hit when it's set.

```
 interface DispatchQueue {
  fun insert(call: AsyncCall)
  fun iterator(): Iterator<AsyncCall>
  fun sort()
}
```

But let's see if we are happy to land this for now, before changing anything.

tom5079:
If we are going to force lower number to be prioritized, why don't we just let the user to return a integer value instead like `priority: (Call) -> Int`?  
I think it is more intuitive this way.

swankjesse:
Yep, that's probably closer to what we need.

One challenge is whether we allow calls to be permanently starved if they have low priority and the system is busy.

I'm seduced by expressing priority as a time delta. Suppose urgent requests get a time delta of -10 seconds. This would cause them to be executed before normal requests enqueued 9 seconds ago and after normal requests enqueued 11 seconds ago. At the opposite end, speculative requests could be given a time delta of +1 hour, and will only be executed if there's nothing else happening, or if an hour elapses.

okhttp/src/jvmMain/kotlin/okhttp3/Dispatcher.kt
tom5079: @yschimke maybe something like this?

https://github.com/tom5079/okhttp/commit/c5d7106a8274b2875d6b603f0bbe9f170dad0072

tom5079:
How about using a custom Priority DSL then?  
```kotlin
request.priority {
    priority(priorityValue)
    mustExecuteIn(100.milliseconds)
}
```

tom5079:
I guess user can implement the feature you mentioned by themselves with current API if they want to. Embedding how the elapsed time affects priority from our side might limit the customizability in my opinion.

swankjesse:
Relevant:
https://httpwg.org/specs/rfc9113.html#PriorityHere

tom5079:
I ended up implementing this as a ktor interceptor, I'm closing this PR.

wdrj:
> I ended up implementing this as a ktor interceptor, I'm closing this PR.
Could you share your demo？

wdrj:
> I ended up implementing this as a ktor interceptor, I'm closing this PR.

hi，@tom5079  could you share your demo？

tom5079:
@wdrj here
https://github.com/tom5079/PupilSources/blob/master/hitomi/src/main/java/xyz/quaver/pupil/sources/hitomi/HitomiPlugin.kt

wdrj:
> @wdrj here https://github.com/tom5079/PupilSources/blob/master/hitomi/src/main/java/xyz/quaver/pupil/sources/hitomi/HitomiPlugin.kt

thanks

PR #7225, opened 2022-04-08
4.10: Version bump of kotlin, okio.

yschimke:
Bump Kotlin to both fix a CVE (without any clear attack), and also to provide a 4.x version of OkHttp with the major upgrade hit taken early.

This should be a 4.10 build not 4.9.4

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -302,6 +302,12 @@ jobs:
       - name: Validate Gradle Wrapper
         uses: gradle/wrapper-validation-action@v1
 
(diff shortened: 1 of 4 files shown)


swankjesse (approved):

swankjesse:
I think we need to create a new branch to merge this into...

swankjesse:
Changed the target to newly-created `okhttp_4.10.x`.

swankjesse (approved):

yschimke:
Some predictable test failures including a 404 from a network using test.

PR #7227, opened 2022-04-09
Kotlin APIs for Request tags

swankjesse:
Including multiplatform.

### DIFF
--- a/okhttp/src/jvmMain/kotlin/okhttp3/Request.kt
+++ b/okhttp/src/jvmMain/kotlin/okhttp3/Request.kt
@@ -16,6 +16,7 @@
 package okhttp3
 
 import java.net.URL
+import kotlin.reflect.KClass
 import okhttp3.HttpUrl.Companion.toHttpUrl
 import okhttp3.internal.canonicalUrl
 import okhttp3.internal.commonAddHeader
@@ -30,6 +31,7 @@ import okhttp3.internal.commonPatch
 import okhttp3.internal.commonPost
 import okhttp3.internal.commonPut
 import okhttp3.internal.commonRemoveHeader
+import okhttp3.internal.commonTag
 import okhttp3.internal.toImmutableMap
 
 actual class Request internal actual constructor(builder: Builder) {
@@ -45,7 +47,7 @@ actual class Request internal actual constructor(builder: Builder) {
   @get:JvmName("body")
   actual val body: RequestBody? = builder.body
 
-  internal val tags: Map<Class<*>, Any> = builder.tags.toImmutableMap()
+  internal actual val tags: Map<KClass<*>, Any> = builder.tags.toImmutableMap()
 
   internal actual var lazyCacheControl: CacheControl? = null
 
@@ -82,6 +84,11 @@ actual class Request internal actual constructor(builder: Builder) {
 
   actual fun headers(name: String): List<String> = commonHeaders(name)
 
+  @JvmName("reifiedTag")
+  actual inline fun <reified T : Any> tag(): T? = tag(T::class)
+
+  actual fun <T : Any> tag(type: KClass<T>): T? = type.java.cast(tags[type])
+
   /**
    * Returns the tag attached with `Object.class` as a key, or null if no tag is attached with
    * that key.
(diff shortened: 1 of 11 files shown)


okhttp/src/jvmMain/kotlin/okhttp3/internal/-UtilJvm.kt
```diff
@@ -223,12 +223,10 @@ fun <T> immutableListOf(vararg elements: T): List<T> {
...
-/** Returns an immutable copy of this. */
-internal fun <K, V> Map<K, V>.toImmutableMap(): Map<K, V> {
-  return if (isEmpty()) {
-    emptyMap()
-  } else {
-    Collections.unmodifiableMap(LinkedHashMap(this))
+internal actual fun <K, V> Map<K, V>.toImmutableMap(): Map<K, V> {
+  return when {
+    isEmpty() -> emptyMap()
+    else -> Collections.unmodifiableMap(LinkedHashMap(this))
```
yschimke, line 229: Is this because they could have grabbed the tags from the builder and downcasted it? How do you decide to protect against this case?

okhttp/src/jvmTest/java/okhttp3/KotlinSourceModernTest.kt
```diff
@@ -957,6 +957,8 @@ class KotlinSourceModernTest {
     val header: String? = request.header("")
     val headersForName: List<String> = request.headers("")
     val body: RequestBody? = request.body
+    var stringTag: String? = request.tag(String::class)
+    stringTag = request.tag<String>()
```
yschimke, line 961: I like it.

okhttp/src/jvmMain/kotlin/okhttp3/Request.kt
```diff
@@ -82,6 +84,11 @@ actual class Request internal actual constructor(builder: Builder) {
 
   actual fun headers(name: String): List<String> = commonHeaders(name)
 
+  @JvmName("reifiedTag")
```
yschimke, line 87: Is this intended to be called, or just moved out of the way?

yschimke (approved):

okhttp/src/jvmMain/kotlin/okhttp3/Request.kt
swankjesse: Yep, this is for calling. It needs a name to avoid colliding with the parameterless signature.

okhttp/src/jvmMain/kotlin/okhttp3/internal/-UtilJvm.kt
swankjesse: now that you mention it, this is probably overkill. It's to prevent Java callers mutating the map, but I don't think we expose the map directly anyway. Can simplify.

okhttp/src/jvmMain/kotlin/okhttp3/Request.kt
yschimke: Should you test for this as new public Java API?

BTW not seeing it in the jvmApiCheck

```
> API check failed for project okhttp.
  --- /home/runner/work/okhttp/okhttp/okhttp/api/okhttp.api
  +++ /home/runner/work/okhttp/okhttp/okhttp/build/api/okhttp.api
  @@ -1011,6 +1011,7 @@
   	public final fun newBuilder ()Lokhttp3/Request$Builder;
   	public final fun tag ()Ljava/lang/Object;
   	public final fun tag (Ljava/lang/Class;)Ljava/lang/Object;
  +	public final fun tag (Lkotlin/reflect/KClass;)Ljava/lang/Object;
   	public fun toString ()Ljava/lang/String;
   	public final fun url ()Lokhttp3/HttpUrl;
   }
  @@ -1034,6 +1035,7 @@
   	public fun removeHeader (Ljava/lang/String;)Lokhttp3/Request$Builder;
   	public fun tag (Ljava/lang/Class;Ljava/lang/Object;)Lokhttp3/Request$Builder;
   	public fun tag (Ljava/lang/Object;)Lokhttp3/Request$Builder;
  +	public final fun tag (Lkotlin/reflect/KClass;Ljava/lang/Object;)Lokhttp3/Request$Builder;
   	public fun url (Ljava/lang/String;)Lokhttp3/Request$Builder;
   	public fun url (Ljava/net/URL;)Lokhttp3/Request$Builder;
   	public fun url (Lokhttp3/HttpUrl;)Lokhttp3/Request$Builder;
  
   You can run :okhttp:apiDump task to overwrite API declarations
```

PR #7228, opened 2022-04-09
The version shouldn't be const

swankjesse:
Downstream projects should get the current version, even if they're
not recompiled.

### DIFF
--- a/okhttp/src/commonMain/kotlin/okhttp3/OkHttp.kt
+++ b/okhttp/src/commonMain/kotlin/okhttp3/OkHttp.kt
@@ -15,6 +15,8 @@
  */
 package okhttp3
 
+import okhttp3.internal.CONST_VERSION
+
 object OkHttp {
   /**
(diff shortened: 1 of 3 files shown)


swankjesse:
Successor to this PR: https://github.com/square/okhttp/pull/6947

okhttp/src/commonMain/kotlin/okhttp3/OkHttp.kt
```diff
@@ -15,6 +15,8 @@
  */
 package okhttp3
 
+import okhttp3.internal.CONST_VERSION
+
 object OkHttp {
```
swankjesse, line 20: Note taht this is moved from templates to regular code.

yschimke (approved):
Pending API fix

```
   public final class okhttp3/OkHttp {
   	public static final field INSTANCE Lokhttp3/OkHttp;
  -	public static final field VERSION Ljava/lang/String;
  +	public final fun getVERSION ()Ljava/lang/String;
   }
```

swankjesse:
Oooh I added `@JvmField` to prevent an API break. Yay API diffs!

PR #7236, opened 2022-04-12
Update works_with_okhttp.md

kullfar:
add ScribeJava to "Works with OkHttp"

as an example
https://github.com/scribejava/scribejava/blob/master/scribejava-apis/src/test/java/com/github/scribejava/apis/examples/GitHubAsyncOkHttpExample.java


yschimke (approved):
Nice!

Issue #7164, opened 2022-03-17
Provide a clean way for Bridging Interceptors to get a cancel signal


yschimke:
Application Interceptor that use an alternate transport and short circuit the interceptor chain don't have a clean way to handle cancellation.

They can either poll the call, or possibly set a specific EventListener and handle the cancel event to get this signal.

yschimke:
Also wondering when implementing an Interceptor in this new world.  Is this a useful base class?

```
abstract class CoroutineInterceptor() : Interceptor {
  final override fun intercept(chain: Interceptor.Chain): Response {
    return runBlocking() {
      interceptSuspend(chain)
    }
  }

  abstract suspend fun interceptSuspend(chain: Interceptor.Chain): Response
}
```

And should it cancel the runBlocking call automatically when the call is cancelled?

Not an issue if you are blocking on the call or response, but if you are doing other work around the interceptor, including another forked request (auth) then the cancellation would be helpful.

yschimke:
I suspect for bridging interceptors currently they end up looking like

```
  override fun intercept(chain: Interceptor.Chain): Response {
    return runBlocking(Dispatchers.IO) {
      makeRequest(engine, chain.request())
    }
  }
}

suspend fun makeRequest(engine: Engine, request: Request) =
  suspendCancellableCoroutine<Response> { continuation ->
```

yschimke:
I'm out of my depth writing a test for this, hence why it would be useful to have a solid library implementation to simplify these cases.

swankjesse:
Definitely runBlocking is wrong here. If you want to use coroutines in the call to `proceed()` I think you're in for a bad time.

yschimke:
Updated mine to runBlocking without IO.  Not sure why runBlocking is wrong here, it achieves the same thing as the standard blocking call, uses the same thread, but allows you to do asynchronous operations elegently.

runBlocking:

 ```
 * The default [CoroutineDispatcher] for this builder is an internal implementation of event loop that processes continuations
 * in this blocked thread until the completion of this coroutine.
 * See [CoroutineDispatcher] for the other implementations that are provided by `kotlinx.coroutines`.
```

swankjesse:
This post gets into it.
https://www.billjings.com/posts/title/foot-marksmanship-with-runblocking/

Coroutines are cooperative multitasking. Running a blocking call or runBlocking inside a coroutine in uncooperative and bad non-local things can happen.

yschimke:
I don't agree with all of that

[...]

PR #7241, opened 2022-04-15
Call cancellation.

yschimke:
#fixes https://github.com/square/okhttp/issues/7164

### DIFF
--- a/okhttp/src/commonMain/kotlin/okhttp3/Call.kt
+++ b/okhttp/src/commonMain/kotlin/okhttp3/Call.kt
@@ -62,6 +62,16 @@ expect interface Call {
    */
   fun clone(): Call
 
+  /**
+   * Registers a callback for cancellation of the Call.
+   * Typically used in an interceptor that may be in a blocking
+   * call and may need to proactively cancel other work.
+   *
+   * Most cancellation should be handled via existing
+   * [execute], [enqueue] and executeAsync behaviour.
+   */
+  actual fun onCancel(cancelFn: () -> Unit)
+
   fun interface Factory {
     fun newCall(request: Request): Call
   }
--- a/okhttp/src/commonMain/kotlin/okhttp3/Call.kt
+++ b/okhttp/src/commonMain/kotlin/okhttp3/Call.kt
@@ -70,7 +70,7 @@ expect interface Call {
    * Most cancellation should be handled via existing
    * [execute], [enqueue] and executeAsync behaviour.
    */
-  actual fun onCancel(cancelFn: () -> Unit)
+  fun onCancel(cancelFn: () -> Unit)
 
   fun interface Factory {
     fun newCall(request: Request): Call
--- a/okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/RealCall.kt
+++ b/okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/RealCall.kt
@@ -118,7 +118,7 @@ class RealCall(
   @Volatile private var exchange: Exchange? = null
   internal val plansToCancel = CopyOnWriteArrayList<RoutePlanner.Plan>()
 
-  @Volatile private var onCancels: List<() -> Unit> = listOf()
+  private var onCancels: List<() -> Unit> = listOf()
 
   override fun timeout(): Timeout = timeout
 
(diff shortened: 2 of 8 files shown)


yschimke:
This form is safer than supporting some blocking operation, since a call may never actually get executed or cancelled.

okhttp/src/commonMain/kotlin/okhttp3/Call.kt
```diff
@@ -62,6 +62,16 @@ expect interface Call {
...
 
+  /**
+   * Registers a callback for cancellation of the Call.
+   * Typically used in an interceptor that may be in a blocking
+   * call and may need to proactively cancel other work.
+   *
+   * Most cancellation should be handled via existing
+   * [execute], [enqueue] and executeAsync behaviour.
+   */
+  actual fun onCancel(cancelFn: () -> Unit)
```
swankjesse, line 73: should this be `Runnable` for our Java language friends?

okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/RealCall.kt
```diff
@@ -118,6 +118,8 @@ class RealCall(
   @Volatile private var exchange: Exchange? = null
   internal val plansToCancel = CopyOnWriteArrayList<RoutePlanner.Plan>()
 
+  @Volatile private var onCancels: List<() -> Unit> = listOf()
```
swankjesse, line 121: why volatile?

okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/RealCall.kt
```diff
@@ -144,10 +146,20 @@ class RealCall(
     }
 
     eventListener.canceled(this)
+
+    onCancels.forEach { onCancel ->
```
swankjesse, line 150: should this be synchronized?

swankjesse: nit: I think `forEach` is strictly worse than regular loops for step debugging

okhttp/src/commonMain/kotlin/okhttp3/Call.kt
yschimke: It's in common, I could typealias it, but maybe a single use "fun interface" is better?

yschimke: I think adding a supporting class here is overkill, anyone using this from java can return Unit.Instance, or null.  Or you feel it's important to be a pleasant API?

swankjesse: Alternately, would it make sense to let callers provide an entire EventListener here? That gives us and callers lots of power and keeps our API compact?

okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/RealCall.kt
yschimke: This class handles concurrency pretty well, with this I didn't really put enough thought into how to do this right.  but race conditions between people registering listeners and a cancelation from another thread probably deserves some consideration, or documentation of the exact rules to exclude race conditions.

I'm going to make a stab at it, but please review for correctness and roughly optimal behaviour in typical cases...

okhttp/src/commonMain/kotlin/okhttp3/Call.kt
yschimke: Yep - let's give it a go!

yschimke: It will make this cancellation callback JVM only. EventListener isn't multiplatform friendly (proxy, inetaddress, HttpUrl).

I'll try this and see how it goes.

We can some KMP later, as we get more idea of what it needs to be.

yschimke: we can have something like `chain.withEventListener()` and that can register an additional listener in RealCall.eventListener, which is essentially the event listener holder.

But it becomes complicated.  what happens if the call was cancelled just as your interceptor is run.  do we special case cancel and call end to always be called?

yschimke:
I won't try to rush this in, but I'll create some tests with the current code so as we discuss the right options we can see how it affects those.

yschimke:
We could also support it now but minimise the blast radius, by putting it on `Chain.onCancel()`, so it's Jvm only.

yschimke:
@swankjesse what do you think about minimising the blast radius of the change, a new method on chain, would be usable for Interceptors, without a new public API for call cancellations for all clients.

Most clients can use the Callback.onFailure to check for cancellation.

Issue #7232, opened 2022-04-10
Provide a way for Bridging Interceptor to provide trailers.


yschimke:
Not currently possible without an Exchange

```
  /**
   * Returns the trailers after the HTTP response, which may be empty. It is an error to call this
   * before the entire HTTP response body has been consumed.
   */
  @Throws(IOException::class)
  fun trailers(): Headers = checkNotNull(exchange) { "trailers not available" }.trailers()
```

PR #7242, opened 2022-04-15
[for discussion] Draft of trailer support

yschimke:
#fixes https://github.com/square/okhttp/issues/7232

### DIFF
--- a/okhttp/src/jvmMain/kotlin/okhttp3/Response.kt
+++ b/okhttp/src/jvmMain/kotlin/okhttp3/Response.kt
@@ -39,6 +39,7 @@ import okhttp3.internal.commonProtocol
 import okhttp3.internal.commonRemoveHeader
 import okhttp3.internal.commonRequest
 import okhttp3.internal.commonToString
+import okhttp3.internal.commonTrailers
 import okhttp3.internal.connection.Exchange
 import okhttp3.internal.http.parseChallenges
 import okio.Buffer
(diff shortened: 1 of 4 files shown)


okhttp/src/jvmMain/kotlin/okhttp3/Response.kt
```diff
@@ -298,6 +303,8 @@ actual class Response internal constructor(
 
     actual open fun priorResponse(priorResponse: Response?) = commonPriorResponse(priorResponse)
 
+    actual open fun trailers(trailersFn: (() -> Headers)): Builder = commonTrailers(trailersFn)
```
yschimke, line 306: This allows creating a Response for a bridging interceptor, without an Exchange.

swankjesse: Nice.

swankjesse (approved):
This is awesome.

yschimke:
Thanks, I'll add some tests and land.

yschimke:
@swankjesse this needs another review, when implementing the tests I realised it's really brittle by default in the world where not all Responses are built by OkHttp Http1, Http2 Codecs.

So I made the default empty headers, but the Http1, Http2 codecs set the explosive handler initially and then update later in initExchange.

yschimke:
@swankjesse Are you happy with the default trailers being `Headers.of()`, so that Call.Factory implementations don't make exploding objects by default?

yschimke:
@swankjesse Are you happy with the default trailers being `Headers.of()`, so that Call.Factory implementations don't make exploding objects by default?

yschimke:
ping @swankjesse

swankjesse (approved):

PR #7253, opened 2022-04-23
[master] Avoid incorrect cache cause IllegalArgumentException

liangjingkanji:


yschimke:
I assume something went wrong here.

PR #7254, opened 2022-04-23
[4.10.x] Avoid incorrect cache cause IllegalArgumentException

liangjingkanji:

```kotlin
    // ...
    val response = entry.response(snapshot) // throw IllegalArgumentException
}


fun response(snapshot: DiskLruCache.Snapshot): Response {
      val contentType = responseHeaders["Content-Type"]
      val contentLength = responseHeaders["Content-Length"]
      val cacheRequest = Request.Builder()
          .url(url)
          .method(requestMethod, null) // method POST must have a request body
          .headers(varyHeaders)
          .build()
    // ....
    }
```

### DIFF
--- a/okhttp/src/main/kotlin/okhttp3/Cache.kt
+++ b/okhttp/src/main/kotlin/okhttp3/Cache.kt
@@ -167,6 +167,9 @@ class Cache internal constructor(
   constructor(directory: File, maxSize: Long) : this(directory, maxSize, FileSystem.SYSTEM)
 
   internal fun get(request: Request): Response? {
+    if (request.method != "GET") {
+      return null
+    }
     val key = key(request.url)
     val snapshot: DiskLruCache.Snapshot = try {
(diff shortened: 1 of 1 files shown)


yschimke:
This looks safe and correct.  But please add a test to CacheTest that demonstrates the problem.

liangjingkanji:
This is unexpected, It's not good to write unit tests in this case

yschimke:
@liangjingkanji How do you manage to get it? Can you provide a stacktrace?

liangjingkanji:
I keep a cache file of Post request method

yschimke:
Also, it's probably better to get the PR done against master first.  I closed https://github.com/square/okhttp/pull/7253, because it wasn't applying cleanly and had a lot of files.  But let's review on that one first.

liangjingkanji:
Ok, I will merge to master

PR #7255, opened 2022-04-23
Avoid incorrect cache cause IllegalArgumentException

liangjingkanji:

```kotlin
    // ...
    val response = entry.response(snapshot) // throw IllegalArgumentException
}


fun response(snapshot: DiskLruCache.Snapshot): Response {
      val contentType = responseHeaders["Content-Type"]
      val contentLength = responseHeaders["Content-Length"]
      val cacheRequest = Request.Builder()
          .url(url)
          .method(requestMethod, null) // method POST must have a request body
          .headers(varyHeaders)
          .build()
    // ....
    }
```

### DIFF
--- a/okhttp/src/jvmMain/kotlin/okhttp3/Cache.kt
+++ b/okhttp/src/jvmMain/kotlin/okhttp3/Cache.kt
@@ -171,6 +171,9 @@ class Cache(
   )
 
   internal fun get(request: Request): Response? {
+    if (request.method != "GET") {
+      return null
+    }
     val key = key(request.url)
     val snapshot: DiskLruCache.Snapshot = try {
       cache[key] ?: return null


swankjesse:
How does this occur in practice? The cache class is internal API and shouldn't be used outside if OkHttp’s own callsites, which all guard the HTTP method.

liangjingkanji:
You can't tell if the user modified the cache file, For example, I cached Post

hingbong:
> You can't tell if the user modified the cache file, For example, I cached Post

When you use HTTP, you should know what are included in HTTP RFC
[Method Definitions](https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9.5)

liangjingkanji:
@hingbong But I think OkHttp should handle caching more safely

liangjingkanji:
Inevitably there will be insufficient capacity of back-end colleagues

swankjesse:
Can you add a test case to CacheTest that fails without this fix?

liangjingkanji:
This is unexpected user file operations, It's not good to write unit tests in this case

swankjesse:
We deliberately don't support this. If users are editing cache files, OkHttp’s cache behavior is undefined.

liangjingkanji:
Ok

Issue #4530, opened 2019-01-04
Client-side load balancing


swankjesse:
We’re using OkHttp for server-to-server connections on high-latency HTTP/2 connections. I’d like some load-balancing features:

 * **Multiple HTTP/2 connections to the same hostname + IP address.** We currently have code that aggressively tries to deduplicate connections to the same destination. But our destinations are virtual server-side load balancers that front application servers. We want to cycle out these application servers without clients having to wait for TLS handshakes.

 * **Preallocated connections.** Seeding the connection pool before we need it, so when we do we don’t have to pay for a TLS handshake.

 * **Client-side connection TTLs.** Long-lived connections can hide problems. By forcing connections to be recreated at a regular interval (say, 1 hour) we can detect performance problems (thundering herds, etc.) before they bite.

yschimke:
+1 to preallocated connections, I've done horrors like this before

https://github.com/yschimke/okurl/blob/master/src/test/kotlin/commands/uberprices.kts#L18

yschimke:
How intelligent do you plan the load balancing strategies to be? Or round robin. Do you want resilience by connecting to multiple backends when multiple ips are available.

swankjesse:
I think we probably want to make the load balancing strategies pluggable. I’m generally not into pluggable things (why not just include a good implementation in the box!) but in this case I think being pluggable gives us options we can’t build otherwise.

One strategy I really like is from what gRPC calls an external load balancer.
https://grpc.io/grpc/core/md_doc_load-balancing.html
https://grpc.io/blog/loadbalancing

esiqveland:
I am interested in seeing this happen, primarily for increased resilience by always being connected to multiple backends, spreading requests among them.

1. gRPC does load balancing per request. Is this something that could fit into okhttp?

2. The current `Dns` interface only lets you return a `List<InetAddress>`. This makes it a lot less practical to implement something like a DNS SRV lookup for a service, which results in an address and a port. Now changing Dns is probably a long shot, but a (address, port) pair is something I want you to consider when looking into client side load balancing.

swankjesse:
@esiqveland 
1. Yes, I wanna do per-call load balancing.
2. I’d like the load balancer to have the option of taking advice from a DNS service.

yschimke:
Is this still a 4.1 target?

swankjesse:
Dropped to backlog. 4.1 should be small!

yschimke:
[...]

PR #7280, opened 2022-05-15
Add a load balancer sample.

yschimke:
Uses a delegating Call.Factory.

### DIFF
--- a/samples/guide/build.gradle.kts
+++ b/samples/guide/build.gradle.kts
@@ -8,10 +8,16 @@ dependencies {
   implementation(projects.mockwebserver)
   implementation(projects.okhttpTestingSupport)
   implementation(projects.okhttpTls)
+  implementation(projects.okhttpCoroutines)
   implementation(libs.animalsniffer.annotations)
   implementation(libs.squareup.moshi)
   implementation(libs.squareup.okio.fakefilesystem)
   kapt(libs.squareup.moshi.compiler)
+}
+
+task("execute", JavaExec::class) {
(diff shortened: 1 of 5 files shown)


ZahraDress:
#

swankjesse:
Is this an admission of OkHttpClient not doing enough internally?

samples/guide/build.gradle.kts
```diff
@@ -8,12 +8,18 @@ dependencies {
...
   implementation(projects.okhttpTestingSupport)
   implementation(projects.okhttpTls)
+  implementation(projects.okhttpCoroutines)
   implementation(libs.animalsniffer.annotations)
   implementation(libs.squareup.moshi)
   implementation(libs.squareup.okio.fakefilesystem)
   kapt(libs.squareup.moshi.compiler)
 }
 
+task("execute", JavaExec::class) {
```
yschimke, line 18: I'll remove

yschimke:
Maybe, ok, yes.

But also that https://github.com/square/okhttp/issues/4530 isn't implemented

And that other "weird" requirements like implementing something like Happy Eyeballs over multiple networks might be a thing. Or that you want to send different requests over different networks (Bluetooth for small quick operations, LTE for large downloads).

yschimke:
@swankjesse What do you think about this sample? Seems like a pattern we should allow for even if we don't implement client side loadbalancing ourselves. This would be valid in apps that want to use multiple connections to backends.

yschimke:
Or we close it off...

yschimke:
Closing, this PR is probably appropriate for now.

PR #7292, opened 2022-05-29
Add enableTls helper

yschimke:
Slight cleanup of logic into OkHttpClientTestRule


yschimke:
Closing - it doesn't materially improve things.  Would basically require making OkHttpClientTestRule public API if we wanted to get real benefit from it.

Issue #7289, opened 2022-05-26
H2_PRIOR_KNOWLEDGE + proxy should force CONNECT


swankjesse:
Currently when using `H2_PRIOR_KNOWLEDGE` with a proxy we’re assuming the proxy speaks HTTP/2. Attempting the same with curl ignores the prior knowledge config:

```
$ curl -vv --http2-prior-knowledge --proxy localhost:61871 http://android.com/foo
*   Trying 127.0.0.1:61871...
* Connected to localhost (127.0.0.1) port 61871 (#0)
* Ignoring HTTP/2 prior knowledge due to proxy
> GET http://android.com/foo HTTP/1.1
```

I think this behavior is awkward. In particular I don’t expect proxies in the wild to accept HTTP/2 prior knowledge.

We should instead do a `CONNECT` tunnel when using `H2_PRIOR_KNOWLEDGE` with a proxy. This is similar to the following curl command:

```
 curl -vv --http2-prior-knowledge --proxy localhost:61672 --proxytunnel http://android.com/foo
```

This will be the first time OkHttp or MockWebServer uses `CONNECT` without TLS, which will require a (small) amount of refactoring. In a local test it didn’t seem too awkward to get MockWebServer to process the curl command above.

PR #7313, opened 2022-06-08
Change how tunnels are configured with MockWebServer

swankjesse:
This is the MockWebServer side of changing how HTTP/2 prior knowledge
works with proxies.

https://github.com/square/okhttp/issues/7289

### DIFF
--- a/mockwebserver/src/main/kotlin/mockwebserver3/MockResponse.kt
+++ b/mockwebserver/src/main/kotlin/mockwebserver3/MockResponse.kt
@@ -25,6 +25,9 @@ import okio.Buffer
 
 /** A scripted response to be replayed by the mock web server. */
 class MockResponse : Cloneable {
+  var inTunnel = false
+    private set
+
   var informationalResponses: List<MockResponse> = listOf()
     private set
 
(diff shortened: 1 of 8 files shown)


mockwebserver/src/main/kotlin/mockwebserver3/SocketPolicy.kt
```diff
@@ -51,12 +51,6 @@ enum class SocketPolicy {
    */
   DISCONNECT_AT_END,
 
-  /**
-   * Wrap the socket with SSL at the completion of this request/response pair. Used for CONNECT
-   * messages to tunnel SSL over an HTTP proxy.
-   */
-  UPGRADE_TO_SSL_AT_END,
```
swankjesse, line 58: This is an OkHttp 5 backwards-incompatible change!

yschimke: Seems reasonable, it's not a dependency of libraries, more likely pulled into app in tests.

5 is the time to do it.

mockwebserver/src/main/kotlin/mockwebserver3/MockResponse.kt
```diff
@@ -362,6 +365,18 @@ class MockResponse : Cloneable {
     webSocketListener = listener
   }
 
+  /**
+   * Configures this response to be served as a response to an HTTP CONNECT request, either for
+   * doing HTTPS through an HTTP proxy, or HTTP/2 prior knowledge through an HTTP proxy.
+   *
+   * When a new connection is received, all in-tunnel responses are served before the connection is
```
yschimke, line 372: Why is it assumed there are multiple tunnel requests? 401?

yschimke (approved):
Seems reasonable, but consider it a low signal review.

mockwebserver/src/main/kotlin/mockwebserver3/MockResponse.kt
swankjesse: 407!

PR #7317, opened 2022-06-08
Delete the unused tunnelProxy argument

swankjesse:


yschimke (approved):
I'm OK with this, but will affect a reasonable number of apps.

Should be default it, or deprecate that form?

https://adambennett.dev/2021/09/mockwebserver-https/

swankjesse:
This only impacts the new `mockwebserver3` package in OkHttp 5. Thoughts?

yschimke:
Shipit!

Issue #2138, opened 2015-12-21
Multipart should not add "Content-Length"


TanTan-TT:
I found following request will cause server error
**The body of your POST request is not well-formed multipart/form-data**

```
--be1563be-4245-4ac5-8a95-498ba5b57ba8
content-disposition: form-data; name="name"
Content-Length: 22

ic_menu_white_24dp.png
...
--d4653e4a-9a5d-42e5-8a49-fcafbe9c673c
content-disposition: form-data; name="file"; filename="ic_menu_white_24dp.png"
Content-Type: image/png
Content-Length: 92
```

In [okhttp code](https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/MultipartBody.java#L160),

``` java
     long contentLength = body.contentLength();
      if (contentLength != -1) {
        sink.writeUtf8("Content-Length: ")
            .writeDecimalLong(contentLength)
            .write(CRLF);
      } else if (countBytes) {
        // We can't measure the body's size without the sizes of its components.
        byteCountBuffer.clear();
        return -1L;
      }
```

And I google the [rfc1867](https://www.ietf.org/rfc/rfc1867.txt), in the document, there is not description for adding "Content-Length", so maybe it is a bug for okhttp.

JakeWharton:
Sounds like a server bug. Any number of headers are allowed in the part bodies so if the length isn't used it should simply be ignored or if it is used the number if accurate so the server is using it incorrectly.

TanTan-TT:
@JakeWharton Are you sure for that? Because I did not found any content that content-length can be included in request in rfc1867

JakeWharton:
It also does not forbid it. That RFC is just adds the 'form-data' subtype to the existing multipart type. The MIME RFC 1341 states:
> The only header fields that have defined meaning for body parts are those the names of which begin with "Content-". All other header fields are generally to be ignored in body parts.

So while it's redundant with the part boundaries, the server software should not crash either.

st0rm23:
some oss provider will request strict format of multipart/form-data while the okhttp will fail to pass.
the temporary solution is copy an own MuiltipartBody from source code and change following code:
            if (contentLength != -1) {
                sink.writeUtf8("Content-Length: ")
                        .writeDecimalLong(contentLength)
                        .write(CRLF);
            } else if (countBytes) {
                // We can't measure the body's size without the sizes of its components.
                byteCountBuffer.clear();
                return -1L;
            }
to 
[...]

Issue #7140, opened 2022-03-04
MultipartBody with Content-Type at the top, but not in the parts


sdandois:
I'm communicating with a server in an embedded device that doesn't have a robust server implementation.  I need to upload a multipart form with a file on it. It works fine from Google Chrome, but doesn't work using this client.

I've realized that the problem was because the server doesn't recognize the header Content-Type inside a Body Part. This is not that uncommon, as it's mentioned in #2604.

However, the workarounds present in there are of no use for me. They work by returning `-1` in the `contentLength` method. If you do that, you also can't know the contentLength of the whole request.

If the `contentLength` of the whole request is unknown, instead of the `Content-Length` header the `Transfer-Encoding: chunked` header is used.

I want to have the Content-Length header at the top of the request, but don't want to have it inside a body part.

I've managed to solve this by rewriting the MultipartBody class, deleting the place where the `Content-Length` header is written. Since the most popular HTTP client in the world (Google Chrome) doesn't use this header, I thought it would be nice to upstream the change. 

Here is my pull: https://github.com/square/okhttp/pull/7139

swankjesse:
Please copy MultipartBody into your project and customize to taste.

PR #7319, opened 2022-06-09
Drop the Content-Length header from multipart parts

swankjesse:
This causes our users grief:
https://github.com/square/okhttp/issues/2604
https://github.com/square/okhttp/issues/2138

I've resisted removing it because it seemed like it'd
break compatibility in a significant way. But with OkHttp 5
it's a good time to break compatibility.

As far as I can tell the behavior of including Content-Length
in multipart bodies is not performed by web browsers or other
HTTP clients.


yschimke (approved):
Seems simple enough. Good call

tcsullivan:
FYI, #7140 also reported this issue with this exact fix proposed in #7139.

As well, the previous behavior is forbidden by [RFC7578](https://www.rfc-editor.org/rfc/rfc7578.html#section-4.8).

Thanks for implementing the fix.

PR #7320, opened 2022-06-09
Support non-standard SOURCETABLE

cloewen8:
Allows OkHttp to be used for Networked Transport of RTCM via Internet Protocol
(Ntrip) requests. NTRIP is an HTTP-based protocol, intended for transmitting navigation data over the internet.

The original NTRIP revision is based on HTTP 1.1 and includes non-standard headers and a custom protocol version. This pull request adds support and a test for the unhandled protocol version (SOURCETABLE), allowing OkHttp to be used with NTRIP.

This change is already being used in a commercial project. My goal is to contribute it, in the hopes it will be useful to other developers and improve NTRIP reliability.

This change is non-breaking and is based on a similar [change made for ICY](https://github.com/square/okhttp/pull/459). Later revisions no longer use this protocol, but they are not yet in widespread use. Alternatively, support for arbitrary HTTP versions could be introduced to keep OkHttp lean and extensible (a breaking change).


yschimke:
Hard to argue with this based on the presence of ICY. 

Do we know other protocols in similar situation?

Do we know how common the use of this protocol is?

How does it work on other platforms?

swankjesse:
I think if we get one request like this every 5 years we just go with it. If ever we get too many protocols in the decoder we can maybe check for spaces instead of specific terms.

cloewen8:
Greatly appreciate this change being merged, thank you!

To answer the questions.

> Do we know other protocols in similar situation?

Haven't seen any other protocols.

> Do we know how common the use of this protocol is?

NTRIP hosts are available as free and paid services globally. RTK2Go is the most popular with "50,000 to 150,000 connections everyday." A list of publicly available NTRIP hosts (casters) can be found here: https://ntrip-list.com/

It is difficult to estimate exactly how common the protocol is. Public statistics are limited. The standard is officially maintained by the [Radio Technical Commission for Maritime Services (RTCM)](https://www.rtcm.org/index.php).

> How does it work on other platforms?

Most implementation examples use a tcp socket directly. The official documentation uses the Posix sys_socket.h.

NTRIP works by having a client (a mobile device with Internet access) connect to a caster (an NTRIP server) which may have several connections to Base Stations that provide satellite data.

An initial request is used to get a list of casters (called Data Streams or more commonly Mount Points)  that can be connected to for positioning data.

- The client sends initial and periodic fix data and periodically receives high accuracy positioning data.
- The caster receives the fix data and low-accuracy satellite data and sends high-accuracy positioning data.

Issue #7248, opened 2022-04-20
OkHttpClient POST request fails when Request body size beyond certain limit


saravr:
I am making a sequence of POST requests with each subsequent request passing the next token received the previous response in the request body of the new request. I am noticing that if POST HTTP request body size is beyond a certain limit, the request times out after certain number of requests in the sequence. Typical request body size is about 14KB. With OkHttpClient event listener I notice that (for the failing call)

    override fun requestBodyStart(call: Call)

is invoked but not

    override fun requestBodyEnd(call: Call, byteCount: Long)

Instead override fun requestFailed(call: Call, ioe: IOException) is called with Socket timeout exception.

It works if make this call before each request (but looks like a hack)

    okHttpClient.connectionPool.evictAll()
Not sure if OkHTTPClient maintains some state within. Charles is showing Failure status with IO: Stream cancelled by CLIENT message.

I am using

    implementation "com.squareup.okhttp3:okhttp:4.9.3"

Backend is not in my control and I cannot optimize the request body size. Apparently, a truncated-body request reaches server which doesn't response and hence the timeout. Is there a way to fix this issue. Note: If I replace OkHttp with Volley I do not see this issue.

swankjesse:
Hmmm... not enough information to suggest any fixes here. I'm curious if your calling code is blocked waiting for the server to read? 

Can you isolate this into a standalone test case?

saravr:
Here is my test code. Possibly I can try using a fake server to create a standalone test.

```
object NetworkModule {
    private lateinit var webservice: Webservice
    private lateinit var okHttpClient: OkHttpClient

    fun initRetrofit() {
        val loggingInterceptor = HttpLoggingInterceptor()
        loggingInterceptor.setLevel(HttpLoggingInterceptor.Level.BODY)

        val builder = OkHttpClient.Builder()
        builder.addInterceptor(loggingInterceptor)
        //builder.eventListener(MyEventListener())
        okHttpClient = builder.build()

        val retrofit = Retrofit.Builder()
            .addConverterFactory(GsonConverterFactory.create())
            .baseUrl(BASE_URL)
            .client(okHttpClient)
            .build()

        webservice = retrofit.create(Webservice::class.java)
    }

    private var next: String? = ""

    suspend fun getData() {
        do {
            val data = webservice.getData(getHeaders(), makeRequest(next))
            next = data.items?.result?.get(0)?.meta?.next
            val itemCount = data.items?.result?.get(0)?.items?.size
            
[...]

PR #7325, opened 2022-06-12
Include more information in WINDOW_UPDATE frame logs

swankjesse:
As a part of investigating this issue:
https://github.com/square/okhttp/issues/7248


okhttp/src/jvmTest/java/okhttp3/internal/http2/FrameLogTest.kt
```diff
@@ -0,0 +1,173 @@
...
+import okhttp3.internal.http2.Http2.TYPE_PING
+import okhttp3.internal.http2.Http2.TYPE_PUSH_PROMISE
+import okhttp3.internal.http2.Http2.TYPE_SETTINGS
+import okhttp3.internal.http2.Http2.formatFlags
+import okhttp3.internal.http2.Http2.frameLog
+import okhttp3.internal.http2.Http2.frameLogWindowUpdate
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+
+class FrameLogTest {
```
swankjesse, line 35: Automatically converted to Kotlin

okhttp/src/jvmTest/java/okhttp3/internal/http2/FrameLogTest.kt
```diff
@@ -0,0 +1,173 @@
...
+      .isEqualTo("<< 0x00000003    22 HEADERS       END_HEADERS")
+    assertThat(frameLog(true, 3, 226, TYPE_DATA, FLAG_END_STREAM))
+      .isEqualTo("<< 0x00000003   226 DATA          END_STREAM")
+    assertThat(frameLog(false, 0, 8, TYPE_GOAWAY, FLAG_NONE))
+      .isEqualTo(">> 0x00000000     8 GOAWAY        ")
+  }
+
+  /** Window update frames have special formatting.  */
+  @Test
+  fun windowUpdateFrames() {
```
swankjesse, line 61: New test function

okhttp/src/jvmMain/kotlin/okhttp3/internal/http2/Http2Reader.kt
```diff
@@ -308,6 +311,14 @@ class Http2Reader(
     if (length != 4) throw IOException("TYPE_WINDOW_UPDATE length !=4: $length")
```
yschimke, line 311: It seems a shame to not log the faulty frame in these two exception cases. Can we assume a rule that we intend to log each and every frame that comes in or goes out?

yschimke (approved):
Looks good.

okhttp/src/jvmMain/kotlin/okhttp3/internal/http2/Http2Reader.kt
swankjesse: Fixed.

PR #7326, opened 2022-06-12
[Experiment] Ktor Call.Factory

yschimke:
Flush out bugs and implementation questions for KMP Call.Factory.
If I remember right, the Plan of Records, it we won't supply this, but it's valid to confirm it's possible.

Found 
- Bug in resolvedContentType with wrong content type string built
 
Limitations
- Currently reads all request and response bodies before continuing. 

Open Questions
- How to handle, large, streaming or delayed request and response body.
- Are there other platforms (native?) we should use this to test on.


okhttp-ktor/src/jsMain/kotlin/okhttp3/ktor/types.kt
```diff
@@ -0,0 +1,37 @@
...
+
+package okhttp3.ktor
+
+import io.ktor.client.statement.HttpResponse
+import io.ktor.client.statement.bodyAsChannel
+import io.ktor.utils.io.core.readBytes
+import okhttp3.RequestBody
+import okhttp3.ResponseBody
+
+actual suspend fun HttpResponse.toHttpResponseBody(): ResponseBody {
```
yschimke, line 26: I think I need the most help here. `buffer.write(bytes)` below is blocking, so at least that should be sorted out.

okhttp-ktor/src/jsMain/kotlin/okhttp3/ktor/types.kt
```diff
@@ -0,0 +1,37 @@
...
+  val bytes = this.bodyAsChannel().readRemaining().readBytes()
+  val buffer = okio.Buffer()
+  buffer.write(bytes)
+  return HttpResponseBody(this, buffer)
+}
+
+actual suspend fun RequestBody.toHttpRequestBody(): Any {
+  val buffer = okio.Buffer()
+  this.writeTo(buffer)
```
yschimke, line 35: Also blocking.

yschimke: The Ktor doc examples aren't great here, using readBytes()

https://ktor.io/docs/request.html#upload_file

And "This function accepts different types of payloads, including plain text, arbitrary class instances, form data, byte arrays, and so on"

swankjesse:
Yeah I think ktor is the wrong abstraction. It's a multiplatform API that wraps various engines. We're better off wrapping other engines directly for maximum power and performance.

swankjesse:
My best idea for JS is to never stream the request or response body.

yschimke:
> My best idea for JS is to never stream the request or response body.

Sure but even this ends up with an blocking operation putting bytes into a Buffer. Is there an immediate way to get a BufferedSource from a  byte array?

image

yschimke:
> Yeah I think ktor is the wrong abstraction. It's a multiplatform API that wraps various engines. We're better off wrapping other engines directly for maximum power and performance.

I like this approach, but I think it's useful here just to validate it possible and provide a template.  I can close this PR off once it's the best it can be, but still have something to point people to.

I also think it's best if we provide this option, prove it works, but then let others who need this build and maintain the implementation.

swankjesse:
Agreed!

yschimke:
What would the best API be for JS? Is that a lot of work that we should take advantage of Ktor here?

Would this fly as a JS reference implementation?

DRSchlaubi:
> What would the best API be for JS? Is that a lot of work that we should take advantage of Ktor here?
> 
> Would this fly as a JS reference implementation?

Probabbly `fetch` since it is supported in browser and nodejs, it's also what ktor does: https://github.com/ktorio/ktor/tree/main/ktor-client/ktor-client-core/js/src/io/ktor/client/fetch

PR #7329, opened 2022-06-13
Remove usage of toDuration()

staktrace:
The toDuration() function was only introduced in Kotlin 1.6. If using
okhttp3 as part of a gradle 7.4.2 plugin it will run with Kotlin 1.5.31
in the runtime classpath, which results in a NoSuchMethodError.

### DIFF
--- a/okhttp/src/commonMain/kotlin/okhttp3/internal/-CacheControlCommon.kt
+++ b/okhttp/src/commonMain/kotlin/okhttp3/internal/-CacheControlCommon.kt
@@ -18,9 +18,9 @@
 
 package okhttp3.internal
 
+import kotlin.time.Duration
 import kotlin.time.DurationUnit
 import kotlin.time.ExperimentalTime
-import kotlin.time.toDuration
 import okhttp3.CacheControl
 import okhttp3.Headers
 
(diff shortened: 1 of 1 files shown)


yschimke (approved):

staktrace:
So this is failing a test that is annotated as `@Flaky`. The test in question is not failing for me locally, so presumably it's failing because of flakiness.

staktrace:
I'm also unable to re-run the test jobs (presumably not enough permissions on this repo) so I'm kinda stuck here. Any help would be appreciated. @yschimke @swankjesse

yschimke:
No problem, there are some flaky tests. But because of some choices we can't rerun them automatically.

I'll land tomorrow if no other reviews.

staktrace:
Thank you! I've confirmed that the latest `5.0.0-SNAPSHOT` release doesn't have the problem any more. Would it be possible to push a new `alpha.9` release with the fix?

yschimke:
cc @swankjesse 

BTW since I struggled to test this, I really appreciate you following through with this and confirming it.  Thanks!

yschimke:
thanks @swankjesse 5.0.0-alpha.9 is available.

staktrace:
Oho! So some more digging produced the discovery that I need to set `https://s01.oss.sonatype.org/content/repositories/snapshots` as the snapshots repository, rather than `https://oss.sonatype.org/content/repositories/snapshots`. I had assumed the latter would redirect to the former (or perhaps to one of a pool of load-balanced servers) but apparently they are distinct. In particular the 5.0.0-SNAPSHOT version on the latter is [20211217.184251-6](https://oss.sonatype.org/content/repositories/snapshots/com/squareup/okhttp3/okhttp/5.0.0-SNAPSHOT/maven-metadata.xml) while the s01 server has [20220616.185128-116](https://s01.oss.sonatype.org/content/repositories/snapshots/com/squareup/okhttp3/okhttp/5.0.0-SNAPSHOT/maven-metadata.xml). So when I tested with `SNAPSHOT` I was getting a stale snapshot build.

yschimke:
So working now?

staktrace:
No, unfortunately. I'm still getting the `NoSuchMethodError: 'double kotlin.time.Duration$Companion.convert(double, kotlin.time.DurationUnit, kotlin.time.DurationUnit)'`. Only difference is I can consistently reproduce it with the alpha.9 and the SNAPSHOT (and even with local builds) now. At this point my best guess is that it has something to do with the `@ExperimentalTime` annotation that is on the function in [1.6.21](https://github.com/JetBrains/kotlin/blob/v1.6.21/libraries/stdlib/src/kotlin/time/Duration.kt#L62-L65) but wasn't on the function in [1.5.31](https://github.com/JetBrains/kotlin/blob/v1.5.31/libraries/stdlib/src/kotlin/time/Duration.kt#L58-L60).

staktrace:
I took the kotlin-stdlib-1.5.31.jar file, extracted the `kotlin/time/Duration$Companion.class` file, and ran it through `javap -v`. This showed me signature of the function is this:

```
  public final double convert(double, java.util.concurrent.TimeUnit, java.util.concurrent.TimeUnit);
```

So in the kotlin 1.5.31 runtime stdlib, the function signature uses `java.util.concurrent.TimeUnit` instead of `kotlin.time.DurationUnit`, but okhttp3 is compiled against 1.6.21 and so is looking for a function with `kotlin.time.DurationUnit`. This whole experimental time API is a bit of a shitshow...

JakeWharton:
This whole Gradle-ships-an-old-ass-version-of-Kotlin-on-the-buildscript-classpath-preventing-normal-dependency-resolution-semantics is a bit of a shit show.

Let's place blame where blame is due.

yschimke:
I guess I don't see a serious problem here.

For JVM - you can use the TimeUnit forms of these methods.
For KMP - You probably are using Kotlin 1.6, so not a problem.

Is there a way we can mark methods as only being Kotlin 1.6?

staktrace:
I'm not super familiar with multiplatform Kotlin stuff, but if I understood what you're saying correctly we need to modify the jvm-specific codepath. I put up a PR at https://github.com/square/okhttp/pull/7343 that seems to work for me.

yschimke:
the PR looks good.

But what I meant was that you can avoid calling this code path in gradle apps. But then I found your original stacktrace, which shows it isn't your code. So I was wrong.

```
Caused by: java.lang.NoSuchMethodError: 'long kotlin.time.DurationKt.toDuration(int, kotlin.time.DurationUnit)'
        at okhttp3.internal._CacheControlCommonKt.commonMaxStale(-CacheControlCommon.kt:59)
        at okhttp3.CacheControl$Builder.maxStale(CacheControl.kt:158)
        at okhttp3.internal._CacheControlCommonKt.commonForceCache(-CacheControlCommon.kt:83)
        at okhttp3.CacheControl.<clinit>(CacheControl.kt:210)
```

I think your PR looks like the right way to go.

My alternative would be moving this code to expect/actuals

https://github.com/square/okhttp/blob/e5154391f890567b36e4d3e816575ebec98896c1/okhttp/src/commonMain/kotlin/okhttp3/internal/-CacheControlCommon.kt#L83

PR #7343, opened 2022-06-20
Improve runtime compatibility with kotlin 1.5.31

staktrace:
I tried to fix this in https://github.com/square/okhttp/pull/7329 but that didn't work. This reverts that attempt and applies a different fix that seems to work better. See individual commit messages for details.

### DIFF
--- a/okhttp/src/commonMain/kotlin/okhttp3/internal/-CacheControlCommon.kt
+++ b/okhttp/src/commonMain/kotlin/okhttp3/internal/-CacheControlCommon.kt
@@ -18,9 +18,9 @@
 
 package okhttp3.internal
 
-import kotlin.time.Duration
 import kotlin.time.DurationUnit
 import kotlin.time.ExperimentalTime
+import kotlin.time.toDuration
 import okhttp3.CacheControl
 import okhttp3.Headers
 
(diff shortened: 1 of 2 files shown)


yschimke (approved):

okhttp/src/commonMain/kotlin/okhttp3/internal/-CacheControlCommon.kt
```diff
@@ -50,20 +50,20 @@ internal fun CacheControl.commonToString(): String {
 
 internal fun CacheControl.Builder.commonMaxAge(maxAge: Int, timeUnit: DurationUnit) = apply {
   require(maxAge >= 0) { "maxAge < 0: $maxAge" }
-  val maxAgeSecondsDouble = Duration.convert(maxAge.toDouble(), timeUnit, DurationUnit.SECONDS)
-  this.maxAgeSeconds = maxAgeSecondsDouble.commonClampToInt()
+  val maxAgeSecondsLong = maxAge.toDuration(timeUnit).inWholeSeconds
```
yschimke, line 53: Nice, so this is reverting back to before these PRs.

okhttp/src/jvmMain/kotlin/okhttp3/CacheControl.kt
```diff
@@ -153,11 +150,18 @@ actual class CacheControl internal actual constructor(
 
     actual fun immutable() = commonImmutable()
 
-    actual fun maxAge(maxAge: Int, timeUnit: DurationUnit) = commonMaxAge(maxAge, timeUnit)
+    // We are compiling with kotlin 1.6 but need to run with older versions at runtime. For
+    // maximum compat we therefore need to handle both the case where DurationUnit is typealiased
+    // to TimeUnit (as it was pre-1.6) and where it is a distinct wrapper enum (in 1.6). We also
+    // can't use durationUnit.toTimeUnit() because that doesn't exist in the typealiased case.
+    // This function should work in either case.
+    internal fun toJavaTimeUnit(durationUnit: DurationUnit) = TimeUnit.valueOf(durationUnit.name)
```
yschimke, line 158: This doesn't look to bad, a lookup in a Hashmap, from a constant name (at least on JVM).

https://github.com/frohoff/jdk8u-dev-jdk/blob/da0da73ab82ed714dc5be94acd2f0d00fbdfe2e9/src/share/classes/java/lang/Class.java#L3339

staktrace:
Confirmed that (finally!) this fixed the issue we were having.

yschimke:
@staktrace a big hearty thanks for your efforts

staktrace:
Here we go again! This PR fixed the issue we were having, but there's a new (but I guess) related issue.

```
Caused by: java.lang.NoClassDefFoundError: kotlin/time/DurationUnit
	at okhttp3.internal._CacheControlCommonKt.commonForceCache(-CacheControlCommon.kt:83)
	at okhttp3.CacheControl.<clinit>(CacheControl.kt:214)
	at okhttp3.Request.cacheControl(Request.kt:115)
	at okhttp3.internal.cache.CacheStrategy$Factory.compute(CacheStrategy.kt:134)
	at okhttp3.internal.cache.CacheInterceptor.intercept(CacheInterceptor.kt:52)
[snip]
Caused by: java.lang.ClassNotFoundException: kotlin.time.DurationUnit
```

This one didn't get caught by the tests for our plugin (which is the thing that uses okhttp3) but started failing in repos that used the plugin. I'll try to construct a minimal test case for this that ideally we can check in, although who knows if there will be more problems after this that the test case doesn't capture.

staktrace:
I made a standalone reproducer and filed https://github.com/square/okhttp/issues/7361

PR #7354, opened 2022-06-26
Make pom.xml depend on the JVM artifact

swankjesse:
This should prevent Maven users from needing to change
their dependencies from okhttp to okhttp-jvm. More importantly,
it should also prevent them from inadvertently getting two
different OkHttp JVM artifacts on their classpaths at once.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,5 +1,7 @@
 import com.vanniktech.maven.publish.MavenPublishBaseExtension
 import com.vanniktech.maven.publish.SonatypeHost
+import groovy.util.Node
+import groovy.util.NodeList
 import java.net.URL
 import kotlinx.validation.ApiValidationExtension
 import org.gradle.api.tasks.testing.logging.TestExceptionFormat
@@ -209,6 +211,7 @@ subprojects {
   }
 
   plugins.withId("com.vanniktech.maven.publish.base") {
+    val publishingExtension = extensions.getByType(PublishingExtension::class.java)
     configure<MavenPublishBaseExtension> {
       publishToMavenCentral(SonatypeHost.S01)
       signAllPublications()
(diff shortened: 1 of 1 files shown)


build.gradle.kts
```diff
@@ -234,6 +237,31 @@ subprojects {
...
+          val dependencies = (root["dependencies"] as NodeList).firstOrNull() as Node?
+            ?: root.appendNode("dependencies")
+          for (child in dependencies.children().toList()) {
+            dependencies.remove(child as Node)
+          }
+          dependencies.appendNode("dependency").apply {
+            appendNode("groupId", jvmPublication.groupId)
+            appendNode("artifactId", jvmPublication.artifactId)
+            appendNode("version", jvmPublication.version)
+            appendNode("scope", "compile")
```
swankjesse, line 261: I don’t think I want this line. Shouldn’t the scope be compile + runtime?

swankjesse:
We end up with this:

```
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  <!-- This module was also published with a richer model, Gradle metadata,  -->
  <!-- which should be used instead. Do not delete the following line which  -->
  <!-- is to indicate to Gradle or any Gradle module metadata file consumer  -->
  <!-- that they should prefer consuming it instead. -->
  <!-- do_not_remove: published-with-gradle-metadata -->
  <modelVersion>4.0.0</modelVersion>
  <artifactId>okhttp</artifactId>
  <version>5.0.0-SNAPSHOT</version>
  <name>okhttp</name>
  <description>Square’s meticulous HTTP client for Java and Kotlin.</description>
  <url>https://square.github.io/okhttp/</url>
  <licenses>
    <license>
      <name>The Apache Software License, Version 2.0</name>
      <url>https://www.apache.org/licenses/LICENSE-2.0.txt</url>
      <distribution>repo</distribution>
    </license>
  </licenses>
  <developers>
    <developer>
      <name>Square, Inc.</name>
    </developer>
  </developers>
  <scm>
    <connection>scm:git:https://github.com/square/okhttp.git</connection>
    <developerConnection>scm:git:ssh://git@github.com/square/okhttp.git</developerConnection>
    <url>https://github.com/square/okhttp</url>
  </scm>
  <dependencies>
    <dependency>
      <artifactId>okhttp-jvm</artifactId>
      <version>5.0.0-SNAPSHOT</version>
      <scope>compile</scope>
    </dependency>
  </dependencies>
</project>
```

build.gradle.kts
swankjesse: I was thinking compileOnly, not compile. This is fine.

staktrace (approved):

Issue #7351, opened 2022-06-22
okio as a dependency changed artifactId in 4.10.x release


honnix:
Since 4.10.0 release, `okio`, as a dependency, has been changed to `okio-jvm` from `okio`. https://mvnrepository.com/artifact/com.squareup.okhttp3/okhttp/4.10.0 vs https://mvnrepository.com/artifact/com.squareup.okhttp3/okhttp/4.9.3

This is problematic because build tools like maven and gradle will treat them as two different artifacts and pull both of them as dependencies.

As an example:
```xml
  <dependencies>
    <dependency>
      <groupId>com.squareup.okio</groupId>
      <artifactId>okio</artifactId>
      <version>2.9.0</version>
    </dependency>
    <dependency>
      <groupId>com.squareup.okhttp3</groupId>
      <artifactId>okhttp</artifactId>
      <version>4.10.0</version>
    </dependency>
  </dependencies>
```

Will result in both `okio-2.9.0.jar` and `okio-jvm-3.0.0.jar` being pulled in as dependencies, while otherwise build tools will only pick up one of those (version resolution is a different story and things like maven-enforcer-plugin will not help in this case). The problem gets harder to be discovered if `okio` is introduced as a transitive dependency by another artifact.

We also noticed this was reverted in 5.x, see https://mvnrepository.com/artifact/com.squareup.okhttp3/okhttp/5.0.0-alpha.9. So maybe this was introduced accidentally?

yschimke:
Fixed in https://github.com/square/okhttp/pull/7354

honnix:
@yschimke Thanks for looking into this.

Could you please help me understand how #7354 fixes this issue? As I understood #7354 creates a dependency from `okhttp` to `okhttp-jvm` so current Maven users depending on `okhttp` will not need to change their dependency when upgrading to later `okhttp` version.

However, since `okhttp-jvm` depends on `okio-jvm`, Maven users having both `okio` and `okhttp` as dependencies will still end up with pulling in both `okio` and `okio-jvm` into classpath, is the right?

Also I see the PR was made towards the main branch so 4.x will not get it.

Let me try to understand the path moving forwards. Is it true starting from `okhttp` 5.x, using `okhttp-jvm` together with `okio-jvm` is the correct way for JVM platform? If that is the case I think #7354 might not help create the awareness.

[...]

PR #7358, opened 2022-06-27
[4.x] Update Okio to 3.2.0

swankjesse:
This version does the right thing for Kotlin Multiplatform
and Maven.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -17,7 +17,7 @@ buildscript {
       'junit': '4.13',
       'kotlin': '1.6.20',
       'moshi': '1.13.0',
-      'okio': '3.0.0',
+      'okio': '3.2.0',
       'ktlint': '0.38.0',
       'picocli': '4.2.0',
       'openjsse': '1.1.0'


swankjesse:
Note that this is on the 4.x branch

honnix:
Just to confirm, this will end up with depending on `okio`, not `okio-jvm`, right?

swankjesse:
@honnix with this change the new `pom.xml` for OkHttp is:

```
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  <!-- This module was also published with a richer model, Gradle metadata,  -->
  <!-- which should be used instead. Do not delete the following line which  -->
  <!-- is to indicate to Gradle or any Gradle module metadata file consumer  -->
  <!-- that they should prefer consuming it instead. -->
  <!-- do_not_remove: published-with-gradle-metadata -->
  <modelVersion>4.0.0</modelVersion>
  <artifactId>okhttp</artifactId>
  <version>4.11.0-SNAPSHOT</version>
  <name>okhttp</name>
  <description>Square’s meticulous HTTP client for Java and Kotlin.</description>
  <url>https://square.github.io/okhttp/</url>
  <licenses>
    <license>
      <name>The Apache Software License, Version 2.0</name>
      <url>http://www.apache.org/licenses/LICENSE-2.0.txt</url>
    </license>
  </licenses>
  <developers>
    <developer>
      <name>Square, Inc.</name>
    </developer>
  </developers>
  <scm>
    <connection>scm:git:https://github.com/square/okhttp.git</connection>
    <developerConnection>scm:git:ssh://git@github.com/square/okhttp.git</developerConnection>
    <url>https://github.com/square/okhttp</url>
  </scm>
  <dependencies>
    <dependency>
      <groupId>com.squareup.okio</groupId>
      <artifactId>okio-jvm</artifactId>
      <version>3.2.0</version>
      <scope>compile</scope>
    </dependency>
    <dependency>
      <groupId>org.jetbrains.kotlin</groupId>
      <artifactId>kotlin-stdlib</artifactId>
      <version>1.6.20</version>
      <scope>compile</scope>
    </dependency>
  </dependencies>
</project>
```

If you have anything else in your project that depends on `com.squareup.okio:okio:3.2.0`, that transitively depends on `com.squareup.okio:okio-jvm:3.2.0` which is where all the code is.

There’s one hazard remaining, which is if your code depends on `com.squareup.okio:okio:2.x.y` directly, since those 2.x (and 1.x) artifacts contain code. We could try to find a way for this to depend on `com.squareup.okio:okio:3.2.0` to force everything to the latest. Hmm...

yschimke:
I don't think I understand the effects half as well as you. Your call.

honnix:
The idea I think is, as long as okhttp depends on okio (non -jvm one), no matter what okio version user explicitly depends on, it would still be the same artifact and it's up to user to decide which version to choose (usually done by maven enforcer). If we merge this as is, users will need to upgrade to okio 3.2.0 as well.

yschimke:
related https://github.com/square/okhttp/issues/7351

yschimke:
@swankjesse What's the status of this, if we need a newer okio, does that also bump Kotlin to 1.7.10 for 4.11.0?

PR #7374, opened 2022-07-10
Fix android tests

yschimke:


yschimke:
@swankjesse requesting post review, a little bit concerned about the new exception.

Issue #7368, opened 2022-07-04
Ipv6 addresses uses a canonical host name for validation. So, it throws SSLPeerUnverifiedException.


SwatiJainGoyal:
When the IPv6 address of the machine is  like - `2007:db8:ffff:ccd:0:0:abcd:1`, it is canonicalized to `2007:db8:ffff:ccd::abcd:1` which results in javax.net.ssl.SSLPeerUnverifiedException as the exact IPs do not match. 

This is there in okhttp/src/jvmMain/kotlin/okhttp3/internal/tls/OkHostnameVerifier.kt - verifyIpAddress method, due to which the Peers could not be verified. 

The error message is:- 
"Error Occur while getting service endpoints status : javax.net.ssl.SSLPeerUnverifiedException: Hostname `2007:db8:ffff:ccd::abcd:1` not verified:\n certificate: sha256/8tyEWtc7zCo59hvlgpiPrkvNo=\n DN: CN=kube-apiserver\n subjectAltNames: [`2007:db8:ffff:ccd:0:0:abcd:1`,  kubernetes, kubernetes.default, kubernetes.default.svc, kubernetes.default.svc.cluster.local, nodelocal-api.eccd.local, seliicbl00125]

yschimke:
I'll take a look. Previous discussion here https://github.com/square/okhttp/issues/5885

yschimke:
Both the expected host, but also the certificate ip addresses should be canonicalized.

https://github.com/square/okhttp/blob/6ad9fb12b62ca05c811d3db3c3578a688d72b776/okhttp/src/jvmMain/kotlin/okhttp3/internal/tls/OkHostnameVerifier.kt#L61-L67

Do you have a sample certificate I can test against? The code is meant to handle this case.

SwatiJainGoyal:
[apiserver.zip](https://github.com/square/okhttp/files/9043214/apiserver.zip)

SwatiJainGoyal:
Any update on the issue?

yschimke:
Sorry, no.  Got distracted.  I'll try to look in the next week.

yschimke:
you have a typo in your certificate or the host 2007 vs 2001.

But when that is excluded, everything is passing.

yschimke:
From the test

```
[...]

PR #7384, opened 2022-07-16
Ipv6 addresses using a canonical host name for validation

yschimke:


yschimke:
For https://github.com/square/okhttp/issues/7368

Issue #7146, opened 2022-03-07
MockWebServer extension is starting new server when I call @BeforeEach and @AfterEach


garcia-jj:
The MockWebServerExtension class is starting a new MockWebServer for every test, beforeEach, beforeAll, afterEach and beforeAll methods, as I can see in [this class](https://github.com/square/okhttp/blob/master/mockwebserver-junit5/src/main/kotlin/mockwebserver3/junit5/internal/MockWebServerExtension.kt).

Since the beforeEach, beforeAll, afterEach and beforeAll methods are usually used just to configure the tests, I want to suggest to use the same server instance when run before/after methods, creating a new fresh instance only for test methods.

What do you think? 

```java
import org.junit.jupiter.api.BeforeEach;
import org.junit.jupiter.api.Test;
import org.junit.jupiter.api.extension.ExtendWith;

import mockwebserver3.MockWebServer;
import mockwebserver3.junit5.internal.MockWebServerExtension;

@ExtendWith(MockWebServerExtension.class)
public class MyServerTest {

	@BeforeEach
	public void setup(final MockWebServer server) {
		System.out.println("Port:" + server.getPort());
	}

	@Test
	public void myTest(final MockWebServer server) throws Exception {
		System.out.println("Port:" + server.getPort());

	}
}
```

yschimke:
Thanks, as a workaround pass in the MockWebServer in the test constructor.

But I think you are right, without any other annotations we should probably reuse the instance.

Seem like in https://junit.org/junit5/docs/current/api/org.junit.jupiter.api/org/junit/jupiter/api/extension/ParameterResolver.html we can check the https://junit.org/junit5/docs/current/api/org.junit.jupiter.api/org/junit/jupiter/api/extension/ParameterContext.html for an annotation that separates instances.

garcia-jj:
Thank you so much, @yschimke. Adding a constructor as you suggested solves my case. I'll looking forward this enhacement for more updates.

swankjesse:
I'd prefer to not reuse the instance by default. We don't reuse it that in OkHttp’s test suite and it's never been a performance problem.

JakeWharton:
OkHttp's test suite does not behave this way because it uses JUnit 4 where the instance is shared by the before, test, and after methods.

This issue proposes bringing it in line with that behavior rather than creating a new instance for each lifecycle method of a single test execution.

garcia-jj:
[...]

PR #7392, opened 2022-07-24
Fix lifecycle for Junit 5 extension.

yschimke:
#fixes https://github.com/square/okhttp/issues/7146

Takes a more opinionated view on lifecycle, namely.

- removing support for constructor injection (see https://github.com/junit-team/junit5/issues/1746)
- not supporting BeforeAll, AfterAll
- requiring a MockWebServerInstance annotation for distinct instances


yschimke:
@swankjesse requesting post review.  Reviews from anyone else very welcome, especially on Junit5 usage.

mockwebserver-junit5/src/test/java/mockwebserver3/junit5/internal/ExtensionLifecycleTest.kt
```diff
@@ -0,0 +1,58 @@
...
+import okhttp3.Request
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.AfterEach
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.extension.ExtendWith
+import org.junit.jupiter.api.extension.RegisterExtension
+
+@ExtendWith(MockWebServerExtension::class)
+class ExtensionLifecycleTest {
```
swankjesse, line 30: This is great

mockwebserver-junit5/src/main/kotlin/mockwebserver3/junit5/internal/MockWebServerInstance.kt
```diff
@@ -15,20 +15,6 @@
...
-  fun testOne() {
-    assertThat(server.started).isTrue()
-  }
-
-  @Test
-  fun testTwo() {
-    assertThat(server.started).isTrue()
-  }
-}
+annotation class MockWebServerInstance(
```
swankjesse, line 18: Is this a JUnit best practice? I don’t love having to define a metaprogramming API here

swankjesse (commented):
I don’t like how we have to do an annotation and book-keeping to support multiple MockWebServer instances, especially since almost nobody is going to need it.

mockwebserver-junit5/src/main/kotlin/mockwebserver3/junit5/internal/MockWebServerInstance.kt
yschimke: I think so, but I can switch to Named?

https://www.vogella.com/tutorials/JUnit5Extensions/article.html

PR #7398, opened 2022-07-27
Bump AGP to 7.2.1

yschimke:
This seems to work nicely with Intellij 2022.2.

So nice to get to a newer version.


Goooler (commented):
Can bump #7375 also?

Issue #7105, opened 2022-02-22
Code example box in the docs


Deceptio:
The code examples are difficult to read because they are in the "normal" text
[OkHttpClient](https://square.github.io/okhttp/4.x/okhttp/okhttp3/-ok-http-client/)
![image](https://user-images.githubusercontent.com/42783844/155110519-2f7ea8a4-dc13-4024-b2cd-92ba37aad61b.png)

[authenticator](https://square.github.io/okhttp/4.x/okhttp/okhttp3/-authenticator/)
![image](https://user-images.githubusercontent.com/42783844/155110414-0d1509c3-357a-4f35-a24d-90dcdb36bbdb.png)

Maybe you could make a codebox like on this page
[Recipes](https://square.github.io/okhttp/recipes/)
![image](https://user-images.githubusercontent.com/42783844/155110654-b56bbad2-f0b6-4e56-b9cd-773d63c90755.png)

rattanshiv12:
Hi, I'm interested in working on this issue. This would be my first time contributing. Can it be assigned to me?

yschimke:
Yes, thanks, please take a look.  I can point you at some of the documentation processes after I get home.

rattanshiv12:
@yschimke Sure, that would be great, thank you. Also, I was wondering how I can locally test my changes and view the documentation as it would appear on https://square.github.io/okhttp/4.x/okhttp/okhttp3/-ok-http-client/

yschimke:
Take a look at this test script, which should generate a local site directory https://github.com/square/okhttp/pull/7414

rattanshiv12:
@yschimke Hey again. Thanks for the test script. It is super helpful for rendering the docs. So I have been trying to get the code blocks to work properly in the inline KDocs markdown, but it doesn't seem to work as expected. For example this snippet: 
'''
// The singleton HTTP client.
 public final OkHttpClient client = new OkHttpClient();
'''
does not render as code. (using single quotes instead of back-ticks here since the comments support markdown)

The same thing with the code indented:
'''
 <4 SPACES>// The singleton HTTP client.
 <4 SPACES>public final OkHttpClient client = new OkHttpClient();
'''
does render as a code block, but now the issue I see is that the rendered code block appends a '\' character at the end of each line. I have tried many different syntaxes as well as adding 2 spaces at the end of each line to stop this behavior but it won't seem to work. Just wondering if you have any advice here as its a very specific issue that I don't see anyone else having with KDoc and Dokka.

[...]

PR #7414, opened 2022-08-06
Add a test script for generating docs

yschimke:
For https://github.com/square/okhttp/issues/7105


yschimke:
@swankjesse Is it worth refactoring this, putting it as another github CI build workflow and uploading the build artifcats so we can check docs on each master build?

Issue #7410, opened 2022-08-05
StackOverflow when shutting down dispatcher executor with hundreds of enqueued calls


JakeWharton:
```kotlin
fun main() {
	val okhttp = OkHttpClient()
	val callback = object : Callback {
		override fun onFailure(call: Call, e: IOException) {}
		override fun onResponse(call: Call, response: Response) {
			response.close()
		}
	}
	repeat(10_000) {
		okhttp.newCall(Request.Builder().url("https://example.com").build()).enqueue(callback)
	}
	okhttp.dispatcher.executorService.shutdown()
}
```
Fails with:
```
Exception in thread "OkHttp Dispatcher" java.lang.StackOverflowError
	at java.base/java.util.ArrayDeque$DeqIterator.remove(ArrayDeque.java:714)
	at okhttp3.Dispatcher.promoteAndExecute(Dispatcher.kt:174)
	at okhttp3.Dispatcher.finished(Dispatcher.kt:213)
	at okhttp3.Dispatcher.finished$okhttp(Dispatcher.kt:198)
	at okhttp3.internal.connection.RealCall$AsyncCall.executeOn(RealCall.kt:507)
	at okhttp3.Dispatcher.promoteAndExecute(Dispatcher.kt:184)
	at okhttp3.Dispatcher.finished(Dispatcher.kt:213)
	at okhttp3.Dispatcher.finished$okhttp(Dispatcher.kt:198)
	at okhttp3.internal.connection.RealCall$AsyncCall.executeOn(RealCall.kt:507)
	at okhttp3.Dispatcher.promoteAndExecute(Dispatcher.kt:184)
	at okhttp3.Dispatcher.finished(Dispatcher.kt:213)
	at okhttp3.Dispatcher.finished$okhttp(Dispatcher.kt:198)
```

PR #7415, opened 2022-08-06
Avoid failing on shutdown executor

yschimke:
Testing of https://github.com/square/okhttp/issues/7410

Consider this fix for discussion.  Open to suggestions.

Issue happens because RealCall.executeOn has rejected execute call and falls back to recursive handling.

### DIFF
--- a/okhttp/src/jvmMain/kotlin/okhttp3/Dispatcher.kt
+++ b/okhttp/src/jvmMain/kotlin/okhttp3/Dispatcher.kt
@@ -15,9 +15,7 @@
  */
 package okhttp3
 
-import java.util.ArrayDeque
-import java.util.Collections
-import java.util.Deque
+import java.util.*
 import java.util.concurrent.ExecutorService
 import java.util.concurrent.SynchronousQueue
 import java.util.concurrent.ThreadPoolExecutor
@@ -182,7 +180,16 @@ class Dispatcher() {
     // Avoid resubmitting if we can't logically progress
     // particularly because RealCall handles a RejectedExecutionException
     // by executing on the same thread.
-    if (!executorService.isShutdown) {
+    if (executorService.isShutdown) {
+      synchronized(this) {
+        for (i in 0 until executableCalls.size) {
+          val asyncCall = executableCalls[i]
+          asyncCall.callsPerHost.decrementAndGet()
+          runningAsyncCalls.remove(asyncCall)
+        }
+        idleCallback?.run()
+      }
+    } else {
       for (i in 0 until executableCalls.size) {
         val asyncCall = executableCalls[i]
         asyncCall.executeOn(executorService)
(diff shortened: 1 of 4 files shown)


yschimke:
Pretty sure this is not the right fix, since Executor could be shutdown at any point.  So will need to change AsyncCall.executeOn.

yschimke:
Since promoteAndExecute is in the recursive path, this should work.  But still open for suggestions here.

okhttp/src/jvmMain/kotlin/okhttp3/Dispatcher.kt
```diff
@@ -15,9 +15,7 @@
  */
 package okhttp3
 
-import java.util.ArrayDeque
-import java.util.Collections
-import java.util.Deque
+import java.util.*
```
JakeWharton, line 18: Revert star import. Seems like we need kotlinter/ktlint setup in here!

okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/RealCall.kt
```diff
@@ -27,16 +27,7 @@ import java.util.concurrent.atomic.AtomicBoolean
...
-import okhttp3.Call
-import okhttp3.Callback
-import okhttp3.CertificatePinner
-import okhttp3.EventListener
-import okhttp3.HttpUrl
-import okhttp3.Interceptor
-import okhttp3.OkHttpClient
-import okhttp3.Request
-import okhttp3.Response
+import okhttp3.*
```
JakeWharton, line 30: Revert

okhttp/src/jvmTest/java/okhttp3/DispatcherCleanupTest.kt
```diff
@@ -0,0 +1,36 @@
...
+  fun testFinish() {
+    val okhttp = OkHttpClient()
+    val callback = object : Callback {
+      override fun onFailure(call: Call, e: IOException) {}
+      override fun onResponse(call: Call, response: Response) {
+        response.close()
+      }
+    }
+    repeat(10_000) {
+      okhttp.newCall(Request.Builder().url("https://example.com").build()).enqueue(callback)
```
JakeWharton, line 32: Probably needs MockWebServer so as not to be dependent on the internet.

JakeWharton (approved):
Just a slight approve as I'm out my depth here.

okhttp/src/jvmMain/kotlin/okhttp3/Dispatcher.kt
yschimke: Sorry, new laptop.

yschimke:
@swankjesse happy to revisit after your review. In the mean time, give it some time to bake.  Should only have any effect, after Executor is shutdown, so not likely to cause problems in normal use.

PR #7416, opened 2022-08-07
Okcurl multiplatform build

yschimke:
Still working with nativeImage, but technically a multiplatform build.

Towards being able to run a native or JS multiplatform okcurl.


okcurl/src/jvmTest/kotlin/okhttp3/curl/MainTest.kt
```diff
@@ -18,26 +18,32 @@ package okhttp3.curl
...
+import assertk.assertThat
+import assertk.assertions.isEqualTo
+import assertk.assertions.isNull
+import assertk.assertions.startsWith
+//import kotlin.test.Test
+//import kotlin.test.fail
 
 class MainTest {
-  @Test fun simple() {
+//  @Test
```
yschimke, line 29: Can't work out why the build is failing on this package of all things...

martinbonnin: Adding `kotlin("test")` to `commonTest` works for me (see https://github.com/yschimke/okhttp/pull/18). Not really sure how it works. I guess it's a special maven artifact that pulls the various implementations for different platforms

yschimke:
Thanks @martinbonnin I'm getting closer.  Also working around this

https://github.com/palantir/gradle-graal/issues/129

and now onto

```
$ okcurl/build/graal/okcurl                
Exception in thread "main" java.lang.NoClassDefFoundError: kotlin.jvm.internal.Intrinsics
        at okhttp3.curl.MainCommandLineKt.main(MainCommandLine.kt)
```

yschimke:
Working with some workarounds, going to land as okcurl isn't a library module, and helps test the multiplatform libraries.

Issue #7430, opened 2022-08-30
java.net.ProtocolException: expected 0 bytes but received 8192


vellrya:
Hello. Long story short:

```
    val url = "https://www.ozon.ru/"
    val requestBuilder = Request.Builder().url(url)
    requestBuilder.header("User-Agent", "Mozilla/5.0 (Linux; Android 13) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/104.0.5112.97 Mobile Safari/537.36")
    val response = OkHttpClient.Builder().build().newCall(requestBuilder.get().build()).execute()
    println(response.protocol)
    println(response.header("Content-Length"))
    println(response.headers)
    println(response.body!!.contentLength())
    println(response.body!!.contentType())
    val body = response.body
    val string = body!!.string()
    body.close()
    println(string)
```

> h2
> null
> 0
> null
> Exception in thread "main" java.net.ProtocolException: expected 0 bytes but received 8192
> 	at okhttp3.internal.connection.Exchange$ResponseBodySource.read(Exchange.kt:290)
> 	at okio.RealBufferedSource.select(RealBufferedSource.kt:229)
> 	at okhttp3.internal.Util.readBomAsCharset(Util.kt:265)
> 	at okhttp3.ResponseBody.string(ResponseBody.kt:187)
> 	at a.b.c.main(Hello.kt:377)

This site works when force Protocol.HTTP_1_1 in OkHttpClient builder, also works perfectly in browser with http2 (either from real device or from PC browser devtools with mobile UA).
![image](https://user-images.githubusercontent.com/13716114/187449616-346425cc-8bf2-4559-b314-66970c015093.png)

To reproduce this error, you can use any mobile user-agent (I tried 3 different).
Tried on OkHttp 4.10.0 / 5.0.0-alpha.10

yschimke:
Thanks, should be relatively easy to repro with a test server.

yschimke:
So it's sending an early hints (103).

Issue: https://github.com/square/okhttp/issues/3671
and a PR https://github.com/square/okhttp/pull/7196

yschimke:
Closing as a duplicate of https://github.com/square/okhttp/issues/3671

PR #7443, opened 2022-09-10
Ignore Early Hints (4.10.x)

yschimke:
Ignore for now to solve https://github.com/square/okhttp/issues/7430

Initial PR from https://github.com/square/okhttp/pull/7196

### DIFF
--- a/okhttp/src/main/kotlin/okhttp3/internal/http2/Http2Reader.kt
+++ b/okhttp/src/main/kotlin/okhttp3/internal/http2/Http2Reader.kt
@@ -65,7 +65,7 @@ class Http2Reader(
       source = continuation,
       headerTableSizeSetting = 4096
   )
-
+  private const val EARLY_HINTS_CODE = "103"
   @Throws(IOException::class)
   fun readConnectionPreface(handler: Handler) {
     if (client) {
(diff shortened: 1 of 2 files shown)


okhttp/src/main/kotlin/okhttp3/internal/http2/Http2Reader.kt
```diff
@@ -146,9 +146,15 @@ class Http2Reader(
     headerBlockLength = lengthWithoutPadding(headerBlockLength, flags, padding)
     val headerBlock = readHeaderBlock(headerBlockLength, padding, flags, streamId)
 
-    handler.headers(endStream, streamId, -1, headerBlock)
+    // Ignore early hints
```
swankjesse, line 149: I would have expected this in `CallServerInterceptor`. Any benefit to doing it here?

okhttp/src/test/java/okhttp3/internal/http2/Http2Test.java
```diff
@@ -646,6 +646,39 @@ public void pushPromise(int streamId, int promisedStreamId, List<Header> headerB
     });
   }
 
+  @Test public void ignoresEarlyHints() throws IOException {
```
swankjesse, line 649: I would have expected a regular MockWebServer test, not an HTTP/2 implementation detail test. Is there something this approach makes better?

PR #7445, opened 2022-09-10
Remove source from MP ResponseBody

yschimke:
Don't advertise this as supported since it's tricky on JS platform.


JakeWharton:
I still think you want this, but it's fully buffered. It can even use a covariant return type to expose Buffer directly.

PR #7446, opened 2022-09-10
Bump to Kotlin 1.7.10

yschimke:
Following https://github.com/square/okio/pull/1155


Goooler (commented):
Need to remove this flag?
https://github.com/square/okhttp/blob/04d142972d316e89b5dbe9215e36b7e29210ca97/build.gradle.kts#L121

Issue #7297, opened 2022-05-30
Manipulate Interceptors on a Call?


oldergod:
We were looking at some Wire stuff, and Wire operates on a `Call.Factory` so it cannot introduce new interceptors to an existing Call.

It’d be interesting/powerful to be able to add interceptors directly on a `Call` instance.

yschimke:
application or network interceptors? both?

oldergod:
Both

yschimke:
This is a weird mixing of layers, but I can see why it's powerful.  Particularly if we increasingly support Call.Factory as API and also as implementation.

But in that regard it's why it's mixing the layers.  Call.Factory and dependencies are fairly tight.  It would pull in Interceptor, and transitively Chain, Connection.

I guess we could ignore this except on jvm.

What is the contract, that you can add final application or network interceptors?  Any need to be in other positions?

yschimke:
This seems doable, but Call is a pretty central public API that is actively implemented.  So wondering what the APIs are like, going to throw in EventListeners here also.

Can't work out how to do this nicely with a clone, since RealCall is heavyweight.

```
  interface Call {
    fun clone(xxx): Call
  }
```

So maybe

```
  interface Call {
    fun addEventListener(eventListener: EventListener)
    fun addInterceptor(interceptor: Interceptor)
    fun addNetworkInterceptor(interceptor: Interceptor)
  }
```

Or

```
  interface Call {
    // must be called before execution
    fun reconfigure(fn: CallOptions.() -> Unit)
  }
  
  interface CallOptions {
    fun addEventListener(eventListener: EventListener)
    var interceptors: List<Interceptor>
    var networkInterceptors: List<Interceptor>
  }
```

Or `interface ConfigurableCall: Call`

yschimke:
closing as not planned

PR #7447, opened 2022-09-10
Allow Interceptors to add EventListeners.

yschimke:
#fixes https://github.com/square/okhttp/issues/7164


okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/RealCall.kt
```diff
@@ -67,7 +67,7 @@ class RealCall(
 ) : Call, Cloneable {
   private val connectionPool: RealConnectionPool = client.connectionPool.delegate
 
-  internal val eventListener: EventListener = client.eventListenerFactory.create(this)
+  internal val eventListener: EventListenerList = EventListenerList(client.eventListenerFactory.create(this))
```
yschimke, line 70: Are we ok with this extra List allocation when not normally needed?

okhttp/src/jvmMain/kotlin/okhttp3/internal/http/RealInterceptorChain.kt
```diff
@@ -82,6 +83,11 @@ class RealInterceptorChain(
     return copy(writeTimeoutMillis = checkDuration("writeTimeout", timeout.toLong(), unit))
   }
 
+  override fun withEventListener(eventListener: EventListener): Interceptor.Chain {
+    call.eventListener.addListener(eventListener)
```
yschimke, line 87: Does this need synchronization?

I thin maybe just documenting that this can only be called during the Interceptor.intercept when nothing else should be going on.

yschimke:
@oldergod related to https://github.com/square/okhttp/issues/7297 somewhat, I wonder if we want to allow adding an EventListener to a Call before it's executed or cancelled?

okhttp-testing-support/src/main/kotlin/okhttp3/RecordingEventListener.kt
```diff
@@ -55,7 +55,9 @@ import org.assertj.core.api.Assertions.assertThat
 import org.assertj.core.data.Offset
 import org.junit.jupiter.api.Assertions.assertTrue
 
-open class RecordingEventListener : EventListener() {
+open class RecordingEventListener(
+  private val enforceOrder: Boolean = true
```
oldergod, line 59: Maybe this needs some doc about the order it's gonna enforce and maybe when it's useful or not

okhttp/src/jvmMain/kotlin/okhttp3/Interceptor.kt
```diff
@@ -100,5 +100,12 @@ fun interface Interceptor {
     fun writeTimeoutMillis(): Int
 
     fun withWriteTimeout(timeout: Int, unit: TimeUnit): Chain
+
+    /**
+     * Add an EventListener to the Call instance, that will receive all
+     * subsequent events. The chain is not mutated, since the Call is shared
+     * when copying a Chain.
```
oldergod, line 107:
```suggestion
     * Add an [EventListener] to the [Call] instance, that will receive all
     * subsequent events. The [Chain] is not mutated, since the call is shared
     * when copying a chain.
```

oldergod (commented):
LGTM but needs a Jesse review before merging

okhttp/src/jvmMain/kotlin/okhttp3/internal/http/RealInterceptorChain.kt
yschimke: I guess it does for cancellation.

yschimke:
@swankjesse any thoughts on this?

yschimke:
@swankjesse ping, it would be good to get this into an alpha for bridge integrations.

swankjesse:
I’m not sold on this.

I think the top-level goal is reasonable, but this implementation isn’t consistent with our other `with` methods on interceptors – this collects all events; not the events downstream of `proceed`. In particular I think the _with_ method would at minimum need to be _add_.

But then I don’t like other side-effects this can trigger. Like if the interceptor runs 2x, it’ll add the listener 2x.

I think if we really wanted this behavior, we’d be better off adding interceptors to `Call` objects. And we practically already have that already in `OkHttpClient`.

Perhaps we start with some use cases and build the API to address them, rather than taking interceptors and trying to make them more powerful.

yschimke:
Ok, will rework.

PR #7452, opened 2022-09-17
Apply Client Ordering of CipherSuites.

yschimke:
Follow up to https://github.com/square/okhttp/pull/6407

For 5.x only, apply the client specified cipher ordering, which defaults to those selected by MODERN_TLS.

This may effect performance of connections, but clients can and should override if required.  Specifically it is likely to be a difference in ordering from the JDK defaults, but those have changed with JDK releases, and these should be kept inline
with modern secure browsers.


yschimke:
@swankjesse if we announce this change, we could also provide a utility to adopt the JDK platform defaults, for the supported set including ordering.

Thoughts?

yschimke:
Possibly relevant https://go.dev/blog/tls-cipher-suites

yschimke:
@swankjesse Thoughts on this?

okhttp/src/jvmTest/java/okhttp3/CipherSuiteTest.java
```diff
@@ -111,7 +111,7 @@ public class CipherSuiteTest {
         forJavaName("TLS_FAKE_NEW_CIPHER"));
   }
 
-  @Test public void applyIntersectionRetainsSslPrefixes() throws Exception {
+  @Test public void applyIntersectionRetainsTlsPrefixes() throws Exception {
```
swankjesse, line 114: What is happening here?

swankjesse (approved):

okhttp/src/jvmTest/java/okhttp3/CallHandshakeTest.kt
```diff
@@ -228,15 +236,15 @@ class CallHandshakeTest {
 
     if (cipherSuites.contains(TLS_CHACHA20_POLY1305_SHA256.javaName)) {
       assertThat(cipherSuites).containsExactlyElementsOf(listOf(
-        TLS_AES_256_GCM_SHA384,
         TLS_AES_128_GCM_SHA256,
+        TLS_AES_256_GCM_SHA384,
```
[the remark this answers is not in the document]
yschimke: > Same as containsExactly(Object...) but handles the Iterable to array conversion : verifies that actual contains exactly the elements of the given Iterable and nothing else in the same order.

okhttp/src/jvmTest/java/okhttp3/CipherSuiteTest.java
yschimke: Yep, it needs to switch because it follows a different set.

PR #7453, opened 2022-09-18
[4.10.x] Read the response even if writing the request fails

yschimke:
Addresses https://github.com/square/okhttp/issues/1001

Backported from an early 5.x


workmanw:
@yschimke I realize this still has to be reviewed, but thank you a ton for your effort! Truly appericate it.

swankjesse (approved):

PR #7456, opened 2022-09-18
Add a ConnectionListener

yschimke:
Given that we have happy eyeballs, and are considering features like client side load balancing, having visibility into connections seems warranted.

Also specifically for Android, dealing with different networks makes this important.


okhttp/src/jvmTest/java/okhttp3/ConnectionListenerTest.kt
```diff
@@ -0,0 +1,268 @@
...
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Tag
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.Timeout
+import org.junit.jupiter.api.extension.RegisterExtension
+
+@Flaky // STDOUT logging enabled for test
+@Timeout(30)
+@Tag("Slow")
+class ConnectionListenerTest {
```
yschimke, line 42: This is currently just a subset of EventListenerTest without too much thought of the interesting cases.

okhttp/src/jvmMain/kotlin/okhttp3/ConnectionListener.kt
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Listener for connection events. Extend this class to monitor the new connections and closes.
+ *
+ * All event methods must execute fast, without external locking, cannot throw exceptions,
+ * attempt to mutate the event parameters, or be reentrant back into the client.
+ * Any IO - writing to files or network should be done asynchronously.
+ */
+abstract class ConnectionListener {
+  open fun connectionOpening(route: Route) {}
```
yschimke, line 29: Comments to come, but for now working out what the interesting events and use cases would be.

- Track the open connections
- Track connection usage, do we keep connection open too long, how many requests on a single connection.
- Route failures causing delays.

Presumably request failures can be tracked via an EventListener already.
-

okhttp/src/jvmMain/kotlin/okhttp3/ConnectionListener.kt
```diff
@@ -0,0 +1,40 @@
...
+  open fun connectFailed(route: Route, failure: IOException) {}
+  open fun connectionOpened(connection: Connection) {}
+  open fun connectionClosed(connection: Connection) {}
+  open fun connectionAcquired(connection: Connection, call: Call) {}
```
yschimke, line 33: TODO - should we identify reasons connections are not eligible? Any parts of RealConnection.isEligible that would not be expected.

okhttp/src/jvmMain/kotlin/okhttp3/ConnectionListener.kt
swankjesse: Rename to `connectStart` ?
Include the motivating `Call` ?

okhttp/src/jvmMain/kotlin/okhttp3/ConnectionListener.kt
```diff
@@ -0,0 +1,40 @@
...
+ * Listener for connection events. Extend this class to monitor the new connections and closes.
+ *
+ * All event methods must execute fast, without external locking, cannot throw exceptions,
+ * attempt to mutate the event parameters, or be reentrant back into the client.
+ * Any IO - writing to files or network should be done asynchronously.
+ */
+abstract class ConnectionListener {
+  open fun connectionOpening(route: Route) {}
+  open fun connectFailed(route: Route, failure: IOException) {}
+  open fun connectionOpened(connection: Connection) {}
```
swankjesse, line 31: `connectEnd`  ? I’m attempting (foolish?) symmetry with EventListener

okhttp/src/jvmMain/kotlin/okhttp3/ConnectionListener.kt
swankjesse: I don’t think it’s that interesting?

okhttp/src/jvmMain/kotlin/okhttp3/ConnectionListener.kt
```diff
@@ -0,0 +1,40 @@
...
+  open fun connectionClosed(connection: Connection) {}
+  open fun connectionAcquired(connection: Connection, call: Call) {}
+  open fun connectionReleased(connection: Connection, call: Call) {}
```
swankjesse, line 34: I wonder if we should have `successCount` here?

swankjesse: Or perhaps something else where we track possibly-non-fatal exceptions, like RealConnection.trackFailure() ?

okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/RealConnection.kt
```diff
@@ -90,6 +92,13 @@ class RealConnection(
    * Once true this is always true. Guarded by this.
    */
   var noNewExchanges = false
+    set(value) {
```
swankjesse, line 95: I expect this probably breaks our rules around not calling EventListeners inside a synchronized block?

swankjesse (commented):
I think this is a great idea. Let’s ship it in 5.0.

okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/RealConnection.kt
swankjesse: (but why didn’t the test fail?)

okhttp/src/jvmMain/kotlin/okhttp3/ConnectionListener.kt
yschimke: Yep - let's add in a follow up.  We should definitely expose this information here, but worth dedicated focus.

okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/RealConnection.kt
yschimke: I've added tests that catch this, will fix before landing.

okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/RealConnection.kt
```diff
@@ -90,6 +92,15 @@ class RealConnection(
...
+      val previousValue = field
+      field = value
+      if (!previousValue) {
+        queueEvent {
+          connectionListener.noNewExchanges(this)
```
yschimke, line 100: This is the one that feels unfortunate... Relies on `connection.synchronizedWithEvents`

okhttp/src/jvmMain/kotlin/okhttp3/ConnectionListener.kt
swankjesse: Good call. Maybe it aught to be a property on the connection.

okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/RealConnection.kt
swankjesse: Yeah I think we should just pull up the event-notifying side-effect into the callsite

swankjesse (commented):
I’d like to find a way to avoid the event queueing. . . might not be possible? Lemme take a stab at it.

swankjesse:
I _think_ it’s possible 'cause we already plumbed through the `Socket` so we could close it without holding a lock.

okhttp-testing-support/src/main/kotlin/okhttp3/RecordingConnectionListener.kt
```diff
@@ -0,0 +1,137 @@
...
+
+  override fun connectStart(route: Route, call: Call) = logEvent(ConnectionEvent.ConnectStart(System.nanoTime(), route, call))
+
+  override fun connectFailed(route: Route, call: Call, failure: IOException) = logEvent(ConnectionEvent.ConnectFailed(System.nanoTime(), route, call, failure))
+
+  override fun connectEnd(connection: Connection) {
+    logEvent(ConnectionEvent.ConnectEnd(System.nanoTime(), connection))
+  }
+
+  override fun connectionClosed(connection: Connection) = logEvent(ConnectionEvent.ConnectionClosed(System.nanoTime(), connection))
```
swankjesse, line 120: We should assert that Start & Close events are balanced.

okhttp-testing-support/src/main/kotlin/okhttp3/RecordingConnectionListener.kt
```diff
@@ -0,0 +1,137 @@
...
+
+  override fun connectionAcquired(connection: Connection, call: Call) {
+    logEvent(ConnectionEvent.ConnectionAcquired(System.nanoTime(), connection, call))
+  }
+
+  override fun connectionReleased(connection: Connection, call: Call) {
```
swankjesse, line 126: We should assert that Acquire & Release events are balanced.

okhttp/src/jvmMain/kotlin/okhttp3/ConnectionListener.kt
```diff
@@ -0,0 +1,66 @@
...
+
+  /**
+   * Invoked when a call no longer uses a connection.
+   */
+  open fun connectionReleased(connection: Connection, call: Call) {}
+
+  /**
+   * Invoked when a connection is marked for no new exchanges.
+   */
+  open fun noNewExchanges(connection: Connection) {}
```
swankjesse, line 61: FYI, with my pushed commits we don’t send this event if we’re definitely going to close the connection immediately

okhttp/src/jvmTest/java/okhttp3/ConnectionListenerTest.kt
```diff
@@ -52,11 +52,11 @@ open class ConnectionListenerTest {
   private val listener = RecordingConnectionListener()
   private val handshakeCertificates = localhost()
 
-  open val fastFallback: Boolean = true
+  open val fastFallback: Boolean get() = true
```
yschimke, line 55: What's the difference here?

okhttp/src/jvmTest/java/okhttp3/RouteFailureTest.kt
```diff
@@ -222,6 +223,7 @@ class RouteFailureTest {
       "ConnectionAcquired",
       "NoNewExchanges",
       "ConnectionReleased",
+      "ConnectionClosed",
```
yschimke, line 226: Nice!

okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/RealConnectionPool.kt
```diff
@@ -118,6 +127,7 @@ class RealConnectionPool(
     connection.assertThreadHoldsLock()
 
     connections.add(connection)
+//    connection.queueEvent { connectionListener.connectEnd(connection) }
```
swankjesse, line 130: What to do here? I think we need a test to check symmetry before we can drop this comment

okhttp-testing-support/src/main/kotlin/okhttp3/RecordingConnectionListener.kt
```diff
@@ -95,6 +95,11 @@ open class RecordingConnectionListener : ConnectionListener() {
   }
 
   private fun logEvent(e: ConnectionEvent) {
+    if (e.connection != null) {
```
yschimke, line 98: Nice.

okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/RealConnectionPool.kt
```diff
@@ -252,7 +265,6 @@ class RealConnectionPool(
       Platform.get().logCloseableLeak(message, callReference.callStackTrace)
 
       references.removeAt(i)
-      connection.noNewExchanges = true
```
swankjesse, line 255: Very small behavior change here - if we leak a Call that doesn‘t immediately mark the connection with noNewExchanges. That’s fine.

okhttp-testing-support/src/main/kotlin/okhttp3/RecordingConnectionListener.kt
yschimke: I can add. Probably as a fast follow.

okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/RealConnectionPool.kt
yschimke: OK, I can add, unless you plan to.

okhttp/src/jvmTest/java/okhttp3/ConnectionListenerTest.kt
swankjesse: Avoids a potential problem accessing state from the superclass constructor.

okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/RealConnectionPool.kt
swankjesse: Lets do follow-up. This PR is big enough!

Issue #7207, opened 2022-04-02
Improve Flaky handling


yschimke:
Current Flaky annotation just causes additional logging to help diagnose

```
/**
 * Annotation marking a test as flaky, and requires extra logging and linking against
 * a known github issue.  This does not ignore the failure.
 */
annotation class Flaky
```

We should improve this.

1) Allow retries of Flaky tests in Junit5 runner?
2) Allow not failing the build because of Flaky test?
3) Collect stats on Flaky tests (github API?) and track

yschimke:
In short term - I think the flaky tests are hampering PR flow.  I'm going to split them out to a separate run.

Retries with a junit5 runner isn't easy it seems.

PR #7462, opened 2022-09-24
Rerun flaky tests

yschimke:
Trying out https://junit-pioneer.org/docs/retrying-test/

#fixes https://github.com/square/okhttp/issues/7207


swankjesse:
LGTM

I made a follow-up here:
https://github.com/square/okhttp/pull/7666

Issue #7366, opened 2022-07-03
OkHttp 5 blockers


yschimke:
- [x] HttpUrl multiplatform (@swankjesse )
- [x] ~Suspend JS support~
- [x] Status Code fixes https://github.com/square/okhttp/issues/2257
- [x] MockWebServer immutable 
- [x] MockWebserver natively support duplex
- [ ] Cancel listeners for Interceptors https://github.com/square/okhttp/issues/7164
- [x] Update cipher suites
- [x] Loom

yschimke:
@swankjesse is Loom support in or out? https://github.com/square/okhttp/issues/6385

Even just confirming it's supported, if not a distinct TaskRunner.Backend?

swankjesse:
I think Loom is out only because I don't expect it'll need API changes.

yschimke:
I do think we'll need a new module with JDK 19+ dependencies, and probably change the implementation, potentially changes, maybe minor API changes, to TaskRunner.

yschimke:
Going to mark https://github.com/square/okhttp/pull/7393 as done based on confirmation of current set against Chrome.

yschimke:
@swankjesse Adding back Loom as I think https://github.com/square/okhttp/pull/7367 showed we need to remove wait()/notify() and use Conditions instead in okhttp and okio.

yschimke:
Also for KMP, I'm looking at using okcurl as a cross platform CLI buildable with all of graal, native, js. https://github.com/square/okhttp/pull/7416

yschimke:
HttpUrl supported, but without IDN.

yschimke:
Suspend JS support is likely just the current example shown for ktor https://github.com/square/okhttp/pull/7326/files#diff-772553bc9e67a5b9842ab416490ff3fd80a8e7bd878ac641a9c0d644bce96492

Response will be read as part of the suspending executeAsync method.

justinadkins:
Is there any news on when we can expect OkHttp 5 to be released?

gesellix:
https://github.com/square/okhttp/issues/5874 (HTTP 1.1 Upgrades) would be great to be included. It might not be a release blocker, though.

yschimke:
I suspect https://github.com/square/okhttp/issues/7164 could be a 5.1 feature if not in 5.0

Issue #2257, opened 2016-01-17
Support HTTP status 102 Processing


rfc2822:
There's a HTTP interim status code **102 Processing**. It's [defined in RFC 2518](http://tools.ietf.org/html/rfc2518#section-10.1) and can be found in the [HTTP Status Code Registry](http://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml).

Funnily, it has been [dropped in the revised WebDAV RFC 4918](http://tools.ietf.org/html/rfc4918#section-21.4), so I don't know whether it's worth dealing with it (RFC 4918 is from 2007 and 2518 from 1999!). However, it's still in the status code registry, and I have just encountered a quite large WebDAV service which uses this status code ([original case](https://forums.bitfire.at/topic/939/multiple-calendars-not-syncing/)).

It could be exactly treated as status code 100, so I guess adding it in Http1xStream would be enough?

```
// HTTP_PROCESSING: 102

 if (statusLine.code != HTTP_CONTINUE || statusLine.code != HTTP_PROCESSING)
// instead of
 if (statusLine.code != HTTP_CONTINUE)
```

What do you think about that?

swankjesse:
Works for me.

rfc2822:
Cool. :sunglasses: RFC 2616 even says:

> **A client MUST be prepared to accept one or more 1xx status responses prior to a regular response**, even if the client does not expect a 100 (Continue) status message. Unexpected 1xx status responses MAY be ignored by a user agent.

So maybe not only 102 should be added, but all 10x status codes. But as RFC 2616 is only a rough recommendation, I don't know whether this would be wise.

swankjesse:
The spec we care about is here:
http://tools.ietf.org/html/rfc7231#section-6.2

I think it’s a good idea to generalize our current `100` behavior to handle all 1xx codes.

JakeWharton:
There's an issue tracking 100's as #1337.

JakeWharton:
Closing as a dupe of #1337 and renaming it to say "1XX".

guillerodriguez:
Just came across this. See my comment in issue #1337. It was closed as a dupe of #675, which only added support for response code 100, but not for 102 (which was the original topic for this one).

swankjesse:
Agreed!

yschimke:
@nowkarol You seem to have put a lot of thought into how to make the change.  Please submit a PR.

[...]

PR #7479, opened 2022-10-17
Ignore Early Hints (4.9.x)

yschimke:
(cherry picked from commit b565eecf167918253145bce2c348b0bb0ef58b0c)

### DIFF
--- a/okhttp/src/main/kotlin/okhttp3/internal/http/CallServerInterceptor.kt
+++ b/okhttp/src/main/kotlin/okhttp3/internal/http/CallServerInterceptor.kt
@@ -18,8 +18,10 @@ package okhttp3.internal.http
 import java.io.IOException
 import java.net.ProtocolException
 import okhttp3.Interceptor
+import okhttp3.Protocol
 import okhttp3.Response
 import okhttp3.internal.EMPTY_RESPONSE
+import okhttp3.internal.connection.Exchange
 import okhttp3.internal.http2.ConnectionShutdownException
 import okio.buffer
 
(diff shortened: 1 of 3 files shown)


yschimke:
@swankjesse CallTest can't be reused because of addInformationResponse which was added in 5

okhttp/src/main/kotlin/okhttp3/internal/http/CallServerInterceptor.kt
```diff
@@ -148,4 +149,15 @@ class CallServerInterceptor(private val forWebSocket: Boolean) : Interceptor {
       throw e
     }
   }
+
+  private fun shouldIgnoreAndWaitForRealResponse(code: Int, exchange: Exchange): Boolean = when {
```
reschke, line 153: That seems to do the wrong thing for all 1xx codes except 100 and 103.

reschke (changes requested):

yschimke:
The intent was to get something in quick to fix an obviously surfaced bug affecting cloudflare customers.

Then a bigger fix in https://github.com/square/okhttp/issues/2257

It's in the blocker list for 5, https://github.com/square/okhttp/issues/7366

Is there a simple fix here we can make to expand it to all other codes including future 1xx and codes we aren't expecting?

reschke:
Just treat all 1xx the same, except for the one that is somewhat special (101).

yschimke:
I'll close this, make a minor update for 5.x/main.  Then backport.

PR #7493, opened 2022-10-28
Allow chaining the redactHeader() calls on HttpLoggingInterceptor

dmitry-cherkas:
Make the following syntax possible:
```
OkHttpClient.Builder()
                .addInterceptor(
                    HttpLoggingInterceptor()
                        .setLevel(Level.BODY)
                        .redactHeader(HttpHeaders.AUTHORIZATION)
                ).build()
```

Currently one needs to introduce a local variable to hold the `HttpLoggingInterceptor` reference:

```
        val loggingInterceptor = HttpLoggingInterceptor()
            .setLevel(Level.BODY)
        loggingInterceptor.redactHeader(HttpHeaders.AUTHORIZATION)

        OkHttpClient.Builder()
            .addInterceptor(loggingInterceptor)
            .build()
```

Also, `setLevel()` already supports chaining calls, so it's logical to make `redactHeader()` support chaining as well.

### DIFF
--- a/okhttp-logging-interceptor/api/logging-interceptor.api
+++ b/okhttp-logging-interceptor/api/logging-interceptor.api
@@ -6,7 +6,7 @@ public final class okhttp3/logging/HttpLoggingInterceptor : okhttp3/Interceptor 
 	public final fun getLevel ()Lokhttp3/logging/HttpLoggingInterceptor$Level;
 	public fun intercept (Lokhttp3/Interceptor$Chain;)Lokhttp3/Response;
 	public final fun level (Lokhttp3/logging/HttpLoggingInterceptor$Level;)V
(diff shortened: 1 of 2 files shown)


yschimke:
You can do this with 

```
OkHttpClient.Builder()
                .addInterceptor(
                    HttpLoggingInterceptor()
                        .setLevel(Level.BODY)
                        .apply {
                           redactHeader(HttpHeaders.AUTHORIZATION)
                        }
                ).build()
```                

So I'm not sure it warrants changing a public API. It should be safe, but since the workaround is simple, I'm inclined to leave it.

swankjesse:
Yep. We don’t break binary-compatibility in OkHttp so this specific change is a nonstarter.

PR #7516, opened 2022-11-13
Use Date.from().

Isira-Seneviratne:
* Use the `Date.from(Instant)` method.

### DIFF
--- a/okhttp/src/jvmMain/kotlin/okhttp3/Headers.kt
+++ b/okhttp/src/jvmMain/kotlin/okhttp3/Headers.kt
@@ -45,8 +45,6 @@ import okhttp3.internal.commonValues
 import okhttp3.internal.headersCheckName
 import okhttp3.internal.http.toHttpDateOrNull
 import okhttp3.internal.http.toHttpDateString
-import okhttp3.internal.http2.Header
-import org.codehaus.mojo.animal_sniffer.IgnoreJRERequirement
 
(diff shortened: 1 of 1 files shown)


yschimke:
Thanks, have you signed the CLA?

Isira-Seneviratne:
> Thanks, have you signed the CLA?

No, I haven't. I'll do that, thanks.

swankjesse:
Ugh, Date.from() needs API 26, and we support API 21+.
https://developer.android.com/reference/java/util/Date#from(java.time.Instant)

swankjesse:
Oooh found this:
https://github.com/yschimke/okhttp/commit/47bca5f551ba21e0720ba3b6f66e0381306206ba

Isira-Seneviratne:
> Ugh, Date.from() needs API 26, and we support API 21+.
> https://developer.android.com/reference/java/util/Date#from(java.time.Instant)

The whole of `java.time` (including Instant) requires API 26+, so the current implementation still gives those warnings. 

However, it's available via desugaring on earlier versions: https://developer.android.com/studio/write/java8-support-table

Isira-Seneviratne:
> Ugh, Date.from() needs API 26, and we support API 21+.
> https://developer.android.com/reference/java/util/Date#from(java.time.Instant)

The method (and `java.time`) is backported for older versions by desugaring.

PR #7525, opened 2022-11-27
Actively set hostname only on Android 5

### LINKED ISSUE #7524: Android: okhttp not respect `SSLParameters.setServerNames` parameter
android 5 (API 21): SNI disabled default
android 6 (API 23): add `jsse.enableSNIExtension` property, default true
android 7 (API 24): add `SSLParameters.setServerNames()`
android 10 (API 29): hidden api, okhttp use Android10SocketAdapter

https://github.com/square/okhttp/blob/d5c180498d2d7df6a6d76f2eb924e04befe64b05/okhttp/src/jvmMain/kotlin/okhttp3/internal/platform/android/AndroidSocketAdapter.kt#L55-L58

For devices under Android 10(StandardAndroidSocketAdapter), okhttp always setHostname before TLS handshake, which causes SNI to be forced enabled and overrides any previous configuration. 
https://cs.android.com/android/platform/superproject/+/android-5.0.0_r1.0.1:external/conscrypt/src/main/java/org/conscrypt/OpenSSLSocketImpl.java;l=855

test app: https://github.com/vvb2060/okhttpsni

android 10 (Android10SocketAdapter)
peerHost is cloudflare-dns.com, same as `serverName`: 
```
old SNI: [type=host_name (0), value=sni.cloudflaressl.com]
hostname: sni.cloudflaressl.com
peerHost:cloudflare-dns.com
cert: cloudflare-dns.com
cert: *.cloudflare-dns.com
cert: one.one.one.one
```

android 8 (StandardAndroidSocketAdapter)
SSL handshake aborted:
```
Caused by: javax.net.ssl.SSLProtocolException: SSL handshake aborted: ssl=0x759043da00: Failure in SSL library, usually a protocol error
error:10000410:SSL routines:OPENSSL_internal:SSLV3_ALERT_HANDSHAKE_FAILURE (external/boringssl/src/ssl/tls_record.cc:579 0x759041b2a0:0x00000001)
error:1000009a:SSL routines:OPENSSL_internal:HANDSHAKE_FAILURE_ON_CLIENT_HELLO (external/boringssl/src/ssl/handshake_client.cc:893 0x7585c740ab:0x00000000)
	at com.android.org.conscrypt.NativeCrypto.SSL_do_handshake(Native Method)
	at com.android.org.conscrypt.SslWrapper.doHandshake(SslWrapper.java:374)
	at com.android.org.conscrypt.ConscryptFileDescriptorSocket.startHandshake(ConscryptFileDescriptorSocket.java:217)
	... 17 more
```

vvb2060:
Fix #7524

https://cs.android.com/android/platform/superproject/+/android-6.0.0_r1:external/conscrypt/src/main/java/org/conscrypt/SSLParametersImpl.java;l=539

`jsse.enableSNIExtension` property is backported, so users can configure SNI themselves if need.

### DIFF
--- a/okhttp/src/jvmMain/kotlin/okhttp3/internal/platform/android/AndroidSocketAdapter.kt
+++ b/okhttp/src/jvmMain/kotlin/okhttp3/internal/platform/android/AndroidSocketAdapter.kt
@@ -15,6 +15,7 @@
  */
 package okhttp3.internal.platform.android
 
+import android.os.Build
 import java.lang.reflect.InvocationTargetException
 import java.lang.reflect.Method
 import javax.net.ssl.SSLSocket
@@ -52,9 +53,12 @@ open class AndroidSocketAdapter(private val sslSocketClass: Class<in SSLSocket>)
         // Enable session tickets.
         setUseSessionTickets.invoke(sslSocket, true)
 
-        if (hostname != null) {
-          // This is SSLParameters.setServerNames() in API 24+.
-          setHostname.invoke(sslSocket, hostname)
+        // SNI is enabled by default since SDK_INT 23
+        if (Build.VERSION.SDK_INT < 23 && hostname != null) {
+          // Backport option
+          if (System.getProperty("jsse.enableSNIExtension", "true") == "true") {
+            setHostname.invoke(sslSocket, hostname)
+          }
         }
 
         // Enable ALPN.


yschimke:
We haven't tended to adopt the configuration idioms of different Android and JVM versions. For example we try to enable newer TLS versions on the JVM before they were the default.  See https://github.com/square/okhttp/blob/3ad1912f783e108b3d0ad2c4a5b1b89b827e4db9/okhttp/src/jvmMain/kotlin/okhttp3/internal/platform/Jdk9Platform.kt#L76 

Your change affects isSniEnabledByDefault, but in getUseSni, the useSni flag is also checked. setHostname causes this to be set, so it seems consistent.

https://cs.android.com/android/platform/superproject/+/android-6.0.0_r1:external/conscrypt/src/main/java/org/conscrypt/SSLParametersImpl.java;l=343

```
    /**
     * This method enables Server Name Indication
     *
     * @param hostname the desired SNI hostname, or null to disable
     */
    public void setHostname(String hostname) {
        sslParameters.setUseSni(hostname != null);
        peerHostname = hostname;
    }
```

Do you have servers where this isn't working on legacy android where we are sending SNI? But would work without? I assume sni.cloudflaressl.com requires SNI, so I'd like to understand for a real world scenario what is broken and what will be fixed by this PR.

Thanks

vvb2060:
#7524 see test app

vvb2060:
The problem is not `setUseSni` , but okhttp doesn't respect SSLParameters.setServerNames parameter.  sni being forced to enable is just a side effect.

yschimke:
Thanks, I'll check

vvb2060:
https://github.com/square/okhttp/pull/5702

I believe this patch will not cause a regression. okhttp doesn't force set hostname can save a lot of trouble.

yschimke:
If it results in use not using SNI in more cases, then it's likely to cause a lot of requests to fail. SNI is pretty critical to modern web infra.

But let me run the sample you provided.

yschimke:
OK, I think I'm up to speed.  However my comment 

> We haven't tended to adopt the configuration idioms of different Android and JVM versions.

Still applies.

Your change has two changes

1) don't call sslSocket.setHostname on 24+, since we assume android does this for us, and it may override a custom ssl socket factory.
2) don't call on <=23 if jsse.enableSNIExtension is false as user may be configuring it off or setting manually.

My two objections are 

1) using jsse.enableSNIExtension, we generally want the best TLS connection the platform can support
2) Disabling SNI in 2022

So the change seems too broad.

Similar requirements have come up before for disabling SNI, and also domain fronting.  We supported domain fronting explicitly as a feature, but stopped at allowing disabling SNI.

I'm still looking at your reproduction, but wanted to give initial thoughts. apologies if I've misunderstood, this timeline of different android versions is hard to keep in my head.

vvb2060:
I'm not trying to disable SNI, I'm trying to okhttp respect the parameters I set.

vvb2060:
I used `SSLParameters.setServerNames()` to set SNI to another host, but okhttp reset it before the handshake and I didn't have a chance to intervene.

yschimke:
When this came up previously the workaround was to set the SNI host in the request url, but then use a custom DNS, to redirect to the actual host you need.

Would this work for your use case?

vvb2060:
You've stopped actively enabling SNI in Android 10+, I don't understand why you're so resistant to stop actively setting incorrect hostname in Android 6+. That doesn't disable SNI, does it?

Also, please stop the discussion around SNI enable or disable, the issue is okhttp not respecting the user's hostname settings, I think it's a behavioral error, not a feature request.

vvb2060:
I added `jsse.enableSNIExtension` not to disable SNI, but to make okhttp stop setting hostname. In Android 5, okhttp can't know what hostname the user wants, and as you said before, there won't be a new API, so simply disable this behavior and give the user a chance to set it themselves.

vvb2060:
> So the change seems too broad.

Android 10+: no changes
Android 6-9: Because the platform has enabled SNI, SNI is still enabled after okhttp stops setting hostname. And users can configure the hostname by reflection or SSLParameters.setServerNames()
Android5: okhttp will still set the hostname and enable SNI by default, if the user wants another hostname, they can set jsse.enableSNIExtension property to stop the behavior.

In summary, nothing has changed from before, and the problem of not respecting user configuration has been fixed.

yschimke:
> I added `jsse.enableSNIExtension` not to disable SNI, but to make okhttp stop setting hostname. In Android 5, okhttp can't know what hostname the user wants, and as you said before, there won't be a new API, so simply disable this behavior and give the user a chance to set it themselves.

I hope this hostname is correctly coming from the request url host.

```
  private fun connectTls(sslSocket: SSLSocket, connectionSpec: ConnectionSpec) {
    val address = route.address
    var success = false
    try {
      if (connectionSpec.supportsTlsExtensions) {
        Platform.get().configureTlsExtensions(sslSocket, address.url.host, address.protocols)
      }
```

Which also points out you can probably disable this via the deprecated flag in ConnectionSpec. It may be we should un-deprecate this and make it a feature to allow clients complete control of SSL session.

But I'm still not convinced about OkHttp supporting legacy setups via a platform flag jsse.enableSNIExtension, we usually done the opposite, ignored platform flags and enabled all supported modern features.

yschimke:
> You've stopped actively enabling SNI in Android 10+, I don't understand why you're so resistant to stop actively setting incorrect hostname in Android 6+. That doesn't disable SNI, does it?
> 
> Also, please stop the discussion around SNI enable or disable, the issue is okhttp not respecting the user's hostname settings, I think it's a behavioral error, not a feature request. Or, add a note in documentation that okhttp ignores `SSLParameters.setServerNames()` until Android 10

Good point. Can we change the PR to respect this on high enough android versions? And not use the jsse.enableSNIExtension flag?

yschimke:
And apologies if this is frustrating for you, not my intention, I just want to make the right change, given I'll need to support this for lots of apps and devices.  Thanks for the PR and your patience walking me through it.

vvb2060:
> Can we change the PR to respect this on high enough android versions?

No need to set anything, when the user used SSLParameters.setServerNames(), the hostname has changed. As long as okhttp does not reset it, everything is as expected.

> I hope this hostname is correctly coming from the request url host.


SSLParameters.setServerNames() is a public API and set hostname is a function that works properly in other Java platforms and Android 10+.

> And not use the jsse.enableSNIExtension flag?

I agree that the jsse.enableSNIExtension property is dirty, it comes from Android 6+ and allows the user to disable hostname that platform auto added for TLS handshake, which is exactly the same as okhttp isn't it? 
When the user actively set hostname, jsse.enableSNIExtension is ignored. But in okhttp, since there is no new API, using this property to avoid breaking the user's configuration I think is the logical thing to do.
Maybe it is possible to use reflection to know if the user has a custom hostname? But this would complicate the code and it would be better to use an existing property.

yschimke:
I'm doing some local testing with https://github.com/square/okhttp/pull/7526/files, which is ported from your repo, thanks.

yschimke:
@vvb2060 can you take a look at https://github.com/square/okhttp/pull/7526, this approach will only call setHostname on API 21+22.  All other versions should rely on the platform.

But it also doesn't add special meaning to jsse.enableSNIExtension, we ignore this and set the hostname forcing it to be used even when jsse.enableSNIExtension==false.

Also take a look at the `SniOverrideTest.getWithDns` test, I think it shows the simpler cleaner way to implement what you are trying to do without a custom socket factory.

vvb2060:
https://github.com/vvb2060/okhttpsni/commit/4ff0ed7afed0e0c9eb2d9f0871c859e1a746796d 
I made some changes to configure the hostname without using reflection in API 17-29. Also modified the test url to avoid disable hostnameVerifier.

vvb2060:
`SniOverrideTest.getWithDns` cannot customize hostname, for example `uuid.one.one.one`

vvb2060:
Since API 21-22 has public API custom hostname, I think okhttp should not reset the hostname already configured by the user.


Android 5.0 api 21：
https://cs.android.com/android/platform/superproject/+/android-5.0.0_r1.0.1:external/conscrypt/src/main/java/org/conscrypt/SSLParametersImpl.java;l=100
```
boolean useSni;
```



Android 5.0 api 22：
https://cs.android.com/android/platform/superproject/+/android-5.1.0_r1:external/conscrypt/src/main/java/org/conscrypt/SSLParametersImpl.java;l=100
```
private Boolean useSni;
```

If user use `SSLCertificateSocketFactory.setHostname()`, `useSni` will be set, just check its value and skip rerset hostname.

yschimke:
> `SniOverrideTest.getWithDns` cannot customize hostname, for example `uuid.one.one.one`

In this example there is 2 configured hosts

1) the sni host taken from the url
2) the target address taken from Dns

And if you want a separate host header to use for domain fronting.

What else do you need to customise?

vvb2060:
Chinese GFW will censor SNI, one.one.one.one is in the blacklist, to avoid TCP reset, I need right IP, right Host, and custom sni.

yschimke:
The dns example should work if you add a host header. In fact it should already work for this with no changes.

vvb2060:
?

vvb2060:
Sorry, I didn't understand.  
When sni is `sni.cloudflaressl.com`, the TLS handshake fails, which is the current behavior in Android 5-9, because okhttp resets the hostname.  
When sni is `one.one.one.one` or `cloudflare-dns.com`, it will receive TCP reset sent by ISP, which is the behavior of `SniOverrideTest.getWithDns`.

vvb2060:
Also, you seem reluctant to disable SNI. SNI has caused serious privacy breaches and internet censorship. I don't think feature requests to disable SNI should be ignored until okhttp supports ECH.

yschimke:
I think we were working directly away from that. As I mentioned above, we have a flag (connectionSpec.supportsTlsExtensions) that is now deprecated for that.

https://github.com/square/okhttp/pull/7525#issuecomment-1328256367

But you should raise a separate issue for that.

vvb2060:
My mistake, I have no problem anymore, your PR is great.

yschimke:
Nice, so the request doesn't need special headers set in advance. 

Actually the Interceptor could rewrite the host in the url also, at the same time it copies original host to the header

PR #7534, opened 2022-11-30
Make it more difficult to accidentally log sensitive headers (#6551)

tristantarrant:
This is a backport of #6551 to the `3.14.x` branch. 

https://github.com/square/okhttp/issues/6551-3.14.x


yschimke:
3.x is not supported.  So you'll need to build your own version and host in a local maven repository.

https://square.github.io/okhttp/security/security/

tristantarrant:
Thanks, I hadn't checked the security policy. I'll just continue with migration to the JDK http client instead.

PR #7549, opened 2022-12-11
HttpUrl multi platform

yschimke:
A first stab at getting it technically working. Encoding isn't inplemented, but starting with API and enough for a typical url.

Feedback on the right API subset, and how to structure code welcome.


okhttp/src/commonMain/kotlin/okhttp3/internal/-HttpUrlCommon.kt
```diff
@@ -0,0 +1,904 @@
...
+package okhttp3.internal
+
+import kotlin.jvm.JvmStatic
+import okhttp3.HttpUrl
+import okhttp3.internal.HttpUrlCommon.canonicalize
+import okhttp3.internal.HttpUrlCommon.writePercentDecoded
+import okio.Buffer
+
+internal expect object HttpUrlCommon {
+  internal fun Buffer.writePercentDecoded(
```
swankjesse, line 25: I wouldn’t have expected this to need an `expect fun` ?

okhttp/src/commonMain/kotlin/okhttp3/internal/-HttpUrlCommon.kt
```diff
@@ -0,0 +1,904 @@
...
+    encoded: String,
+    pos: Int,
+    limit: Int,
+    plusIsSpace: Boolean
+  )
+
+  internal fun String.canonicalize(
```
swankjesse, line 32: same

okhttp/src/jvmMain/kotlin/okhttp3/internal/-HttpUrlJvm.kt
```diff
@@ -0,0 +1,197 @@
...
+          val b = encodedCharBuffer.readByte().toInt() and 0xff
+          writeByte('%'.code)
+          writeByte(HEX_DIGITS[b shr 4 and 0xf].code)
+          writeByte(HEX_DIGITS[b and 0xf].code)
+        }
+      } else {
+        // This character doesn't need encoding. Just copy it over.
+        writeUtf8CodePoint(codePoint)
+      }
+      i += Character.charCount(codePoint)
```
swankjesse, line 79: Is this the reason it isn’t multipatform? Okio _almost_ does this: 
https://square.github.io/okio/3.x/okio/okio/okio/utf8-size.html

swankjesse: I’d prefer we create our own multiplatform implementation of `Character.charCount()` vs. duplicate this whole function?

swankjesse: . . . or is the issue the Charset parameter?

It’d be interesting for OkHttp to define a `Charset` that only permits UTF-8 on non-JVM platforms!

okhttp/src/nonJvmMain/kotlin/okhttp3/internal/-HostnamesNonJvm.kt
```diff
@@ -0,0 +1,58 @@
...
+    return null
+  }
+}
+
+internal fun inet4AddressToAscii(address: ByteArray): String {
+  return address.joinToString(".")
+}
+
+fun idnToAscii(host: String): String {
+  // TODO implement properly
```
swankjesse, line 56: Yep!

okhttp/src/nonJvmMain/kotlin/okhttp3/internal/-HttpUrlNonJvm.kt
```diff
@@ -0,0 +1,148 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okhttp3.internal
+
+import okhttp3.internal.CommonHttpUrl.isPercentEncoded
+import okhttp3.internal.NonJvmHttpUrl.writeCanonicalized
+import okio.Buffer
+
+internal object NonJvmHttpUrl {
```
swankjesse, line 22: Reviewer request: any meaningful changes to any of these big functions? I’d prefer to skip a thorough review for anything that’s just moved around

swankjesse (approved):

okhttp/src/jvmMain/kotlin/okhttp3/internal/-HttpUrlJvm.kt
yschimke: Should this go down to okio?

okhttp/src/nonJvmMain/kotlin/okhttp3/internal/-HttpUrlNonJvm.kt
yschimke: No.

I'll refactor in a follow up that will try to make the same common implementation for all of these including Charset as you suggest.  But would like to land this so I have a working implementation to start with.

okhttp/src/commonMain/kotlin/okhttp3/internal/-HttpUrlCommon.kt
yschimke: Called by `String.percentDecode` in common code.

okhttp/src/commonMain/kotlin/okhttp3/internal/-HttpUrlCommon.kt
yschimke: 23 call locations in common code.

yschimke:
Going to land, while it's not how I want to to eventually be.  I want to be able to iterate on improving it with a working base.  And also be able to use in real KMP test implementations.

PR #7551, opened 2022-12-12
Use DateTimeFormatter.

Isira-Seneviratne:
Use a static `DateTimeFormatter` variable instead of a `SimpleDateFormat` `ThreadLocal` variable, as `DateTimeFormatter` is thread-safe.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ Further examples are on the [OkHttp Recipes page][recipes].
 Requirements
 ------------
 
-OkHttp works on Android 5.0+ (API level 21+) and Java 8+.
+OkHttp works on Android 5.0+ (API level 21+) and Java 8+. On Android projects with a `minSdk` below 26 (Android 8.0), [core library desugaring](https://developer.android.com/studio/write/java8-support#library-desugaring) is required.
 
 OkHttp depends on [Okio][okio] for high-performance I/O and the [Kotlin standard library][kotlin]. Both are small libraries with strong backward-compatibility.
 
--- a/okhttp/src/jvmMain/kotlin/okhttp3/internal/http/dates.kt
+++ b/okhttp/src/jvmMain/kotlin/okhttp3/internal/http/dates.kt
@@ -18,6 +18,10 @@ package okhttp3.internal.http
 import java.text.DateFormat
 import java.text.ParsePosition
 import java.text.SimpleDateFormat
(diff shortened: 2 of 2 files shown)


yschimke:
Is something driving this? Or is it a cleanup?

Isira-Seneviratne:
> Is something driving this? Or is it a cleanup?

Yeah, I used an existing static field in `DateTimeFormatter`, sorry about the additional commit. I removed that to make my first PR simpler.

Isira-Seneviratne:
> What is the context behind the change? Does it fix a bug you have observed?

It uses an existing static `DateTimeFormatter` variable, so that a separate `SimpleDateFormat` variable does not have to be defined.

> Will it fail for apps that haven't applied the library-desugaring?

Yes, if the Android version is below 8.0.

yschimke:
Why is RFC 1123 the one used?

What about https://www.rfc-editor.org/rfc/rfc7231#section-7.1.1 ?

Should we prefer two digit dates?

```
   day             =   ([FWS] 1*2DIGIT FWS) / obs-day
```

Isira-Seneviratne:
> Why is RFC 1123 the one used?

My bad, the RFC 7231 format was similar to the RFC 1123 format. I'll update the change.

yschimke:
Since it seems to introduce a potential build change for users, it seems like a bit of pain for no improvement. So I'm tended to close this PR, sorry.

Can you add a test demonstrating a problem with the current logic?
Or demonstrate the new logic is more performant in benchmarks?

```
STANDARD_DATE_FORMAT = object : ThreadLocal<DateFormat>
```

This should be uncontended because of the ThreadLocal, so I'm not convinced it is a performance issue.

Isira-Seneviratne:
> ```
> STANDARD_DATE_FORMAT = object : ThreadLocal<DateFormat>
> ```
> 
> This should be uncontended because of the ThreadLocal, so I'm not convinced it is a performance issue.

Yeah, `DateTimeFormatter` is thread-safe, unlike `SimpleDateFormat`.

yschimke:
Since the ThreadLocal handles that, I'm going to close this out.

yschimke:
Reopening - since we already have some other dependencies like this.

yschimke:
Actually, nope.  

The existing ones are specific methods you avoid if you don't opt into java.time.

PR #7566, opened 2022-12-24
Enable fastFallback by default

yschimke:


yschimke:
Will flush out test failures.

yschimke:
@swankjesse as discussed on Slack, enabling by default before alpha11.

swankjesse (commented):
LGTM!

PR #7570, opened 2022-12-24
Do not merge: Retry publishing 5.0.0-alpha.11

swankjesse:


.github/workflows/build.yml
```diff
@@ -16,7 +16,7 @@ env:
 jobs:
   publish:
     runs-on: ubuntu-latest
-    if: github.repository == 'square/okhttp' && github.ref == 'refs/heads/master'
+    if: github.repository == 'square/okhttp'
```
swankjesse, line 19: Trick to force a publish attempt

PR #7571, opened 2022-12-24
Fix the broken connect timeout test

swankjesse:


swankjesse:
FYI @knownitwhy !

yschimke (approved):
LGTM. 5 seems high, but well within the 10 second default.

PR #7583, opened 2022-12-28
Create MockResponseBody

swankjesse:
We'd previously only ever used Buffer to model the type of the MockResponse body. This has proven inadequate, particularly for features like duplex responses.

Most of the complexity in this PR is changing how throttling and disconnect-during-body work when we don't necessarily have a Buffer for the response body.

This is a first step, more capability to follow.


yschimke (approved):
Nice change. Are you considering mockwebserver more open for investment in future?

swankjesse:
With OkHttp 5.x we’ve got a new public API package, mockwebserver3, which presents a rare opportunity to make breaking changes. I’m making 'em. I hope we don’t need that much more investment in the future; just removing some bad things. But I’m open to ideas!

PR #7588, opened 2022-12-30
[Discussion] Add more connection failure tracking and tests.

yschimke:
Unlikely to be correct, but if the ConnectPlan is the right place, then a few more questions.

- Are there more places for `routeDatabase.failed`?
- Where should `routeDatabase.connected` be called?
- Which exceptions indicate serious problems?

RealConnection.trackFailure doesn't look right for HTTP/2.


okhttp/src/jvmTest/java/okhttp3/CallTest.kt
```diff
@@ -1022,6 +1022,8 @@ open class CallTest {
     executeSynchronously(request)
       .assertCode(200)
       .assertBody("success!")
+
+    assertThat(client.routeDatabase.failedRoutes.single().proxy.address()).isEqualTo(server.toProxyAddress().address())
```
yschimke, line 1026: This is the only failing test when `routeDatabase.failed` is not called.  Seems like we are missing coverage.

yschimke:
Only failure from disabling the RouteDatabase tracking prior to this

```
CallTest[jvm] > readTimeoutFails()[jvm] FAILED
    java.lang.AssertionError: 
    Expecting actual not to be null
        at okhttp3.RecordedResponse.assertCode(RecordedResponse.java:61)
        at okhttp3.CallTest.readTimeoutFails(CallTest.kt:1023)
```

okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/ConnectPlan.kt
```diff
@@ -132,6 +132,7 @@ class ConnectPlan(
       success = true
       return ConnectResult(plan = this)
     } catch (e: IOException) {
+      client.routeDatabase.failed(route)
```
swankjesse, line 135: This will also capture cases where the call was canceled. Hmm.

okhttp/src/jvmTest/java/okhttp3/RouteFailureTest.kt
```diff
@@ -217,30 +170,34 @@ class RouteFailureTest {
...
-      .fastFallback(true)
+      .fastFallback(fastFallback)
       .apply {
         retryOnConnectionFailure = true
       }
       .build()
 
     executeSynchronously(request)
-      .assertFailureMatches("stream was reset: REFUSED_STREAM")
+      .assertCode(200)
```
swankjesse, line 197: ooooh does your fix cause this to connect to a good host? That is PRETTY RAD

okhttp/src/jvmTest/java/okhttp3/CallTest.kt
swankjesse: Or perhaps, the coverage we do have (in RouteFailureTest) is overly pessimistic??

okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/ConnectPlan.kt
swankjesse: And looking at the other call to routeDatabase.failed(), I think we should notify the proxy selector.

This is from `RealConnection.kt`:

```
    // Tell the proxy selector when we fail to connect on a fresh connection.
    if (failedRoute.proxy.type() != Proxy.Type.DIRECT) {
      val address = failedRoute.address
      address.proxySelector.connectFailed(
        address.url.toUri(), failedRoute.proxy.address(), failure
      )
    }
```

Perhaps the nicest structure is to move that snippet out of `RealConnection` and into `RouteDatabase`, so the two are always in sync.

yschimke: Yep, I wanted this to prompt discussion.  I'm not clear that the current exception filtering is obviously right.

I'll split it out and use the opportunity to make a test.

PR #7595, opened 2022-12-30
Public API for duplex streams in MockWebServer

swankjesse:
I'm not 100% on the name 'Stream' for the source+sink pair. It's tempting to use 'Socket', though I think that's an implementation name and this is an abstraction that uses a different implementation.

I've chosen Stream specifically 'cause it's the word used in the HTTP/2 spec. My biggest gripe with it is that it's bidirectional in the HTTP/2 spec, but Java InputStream and OutputStream are not bidirectional.


mockwebserver/src/main/kotlin/mockwebserver3/MockResponse.kt
```diff
@@ -152,7 +148,37 @@ class MockResponse {
...
+        if (value != null) {
+          body_ = null
+          webSocketListener_ = null
+        }
+      }
+    var webSocketListener: WebSocketListener?
+      get() = webSocketListener_
+      set(value) {
+        webSocketListener_ = value
+        if (value != null) {
```
yschimke, line 177: This seems confusing. Specifically a null not clearing the other ones.

Might be better to fail if this is non null and other fields also? Make them explicitly clear other fields first?

But your call.

mockwebserver/src/main/kotlin/mockwebserver3/Stream.kt
```diff
@@ -0,0 +1,39 @@
...
+ */
+package mockwebserver3
+
+import okio.BufferedSink
+import okio.BufferedSource
+
+/**
+ * A bidirectional sequence of data frames exchanged between client and server.
+ */
+interface Stream {
```
yschimke, line 24: I'll keep nerd sniping with suggestions this should be a SocketLike api in okio.

mockwebserver/src/main/kotlin/mockwebserver3/StreamHandler.kt
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2018 Square, Inc.
+ * Copyright (C) 2022 Block, Inc.
```
yschimke, line 2: Global search and replace? Or just new files in the year of our lord 2023?

yschimke (approved):

mockwebserver/src/main/kotlin/mockwebserver3/Stream.kt
swankjesse: Agreed. The giant worm can we’ll find ourselves in is TLS. That’s okay.

mockwebserver/src/main/kotlin/mockwebserver3/StreamHandler.kt
swankjesse: Block for new things, old things stay put.

mockwebserver/src/main/kotlin/mockwebserver3/MockResponse.kt
swankjesse: Yeah, I was trying to avoid awkwardness when doing a manual copy:

```
val a: MockResponse = ...
val b = MockResponse.Builder()
  .body(a.body)
  .webSocketListener(a.webSocketListener)
  .streams(a.streams)
```

I’ve probably overthunk it. I’ll change to make null not-special.




```

**Comment by swankjesse** [comment]
Test failure:

```

DuplexTest[jvm] > duplexWithAuthChallenge()[jvm] FAILED
    java.util.concurrent.ExecutionException: okhttp3.internal.http2.StreamResetException: stream was reset: CANCEL
        at java.base/java.util.concurrent.FutureTask.report(FutureTask.java:122)
        at java.base/java.util.concurrent.FutureTask.get(FutureTask.java:205)
        at mockwebserver3.internal.duplex.MockStreamHandler.awaitSuccess(MockStreamHandler.kt:111)
        at okhttp3.DuplexTest.duplexWithAuthChallenge(DuplexTest.java:415)

        Caused by:
        okhttp3.internal.http2.StreamResetException: stream was reset: CANCEL
            at app//okhttp3.internal.http2.Http2Stream.checkOutNotClosed$okhttp(Http2Stream.kt:675)
            at app//okhttp3.internal.http2.Http2Stream$FramingSink.emitFrame(Http2Stream.kt:585)
            at app//okhttp3.internal.http2.Http2Stream$FramingSink.flush(Http2Stream.kt:608)
            at app//okio.RealBufferedSink.flush(RealBufferedSink.kt:269)
            at app//mockwebserver3.internal.duplex.MockStreamHandler$sendResponse$1$1.invoke(MockStreamHandler.kt:71)
            at app//mockwebserver3.internal.duplex.MockStreamHandler$sendResponse$1$1.invoke(MockStreamHandler.kt:69)
            at app//mockwebserver3.internal.duplex.MockStreamHandler.serviceStreamTask$lambda$9(MockStreamHandler.kt:99)
            at java.base@11.0.17/java.util.concurrent.FutureTask.run(FutureTask.java:264)
            at app//mockwebserver3.internal.duplex.MockStreamHandler.handle(MockStreamHandler.kt:87)
            at app//mockwebserver3.MockWebServer$Http2SocketHandler.writeResponse(MockWebServer.kt:1071)
            at app//mockwebserver3.MockWebServer$Http2SocketHandler.onStream(MockWebServer.kt:933)
```

PR #7598, opened 2022-12-31
Use lazy delegate for standard date format.

Isira-Seneviratne:
Use Kotlin's lazy delegate to initialize the standard date format.

### DIFF
--- a/okhttp/src/jvmMain/kotlin/okhttp3/internal/http/dates.kt
+++ b/okhttp/src/jvmMain/kotlin/okhttp3/internal/http/dates.kt
@@ -29,13 +29,11 @@ internal const val MAX_DATE = 253402300799999L
  * Most websites serve cookies in the blessed format. Eagerly create the parser to ensure such
  * cookies are on the fast path.
  */
-private val STANDARD_DATE_FORMAT = object : ThreadLocal<DateFormat>() {
-  override fun initialValue(): DateFormat {
-    // Date format specified by RFC 7231 section 7.1.1.1.
-    return SimpleDateFormat("EEE, dd MMM yyyy HH:mm:ss 'GMT'", Locale.US).apply {
-      isLenient = false
-      timeZone = UTC
-    }
+private val STANDARD_DATE_FORMAT by lazy {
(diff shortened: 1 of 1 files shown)


okhttp/src/jvmMain/kotlin/okhttp3/internal/http/dates.kt
```diff
@@ -29,13 +29,11 @@ internal const val MAX_DATE = 253402300799999L
...
  * cookies are on the fast path.
  */
-private val STANDARD_DATE_FORMAT = object : ThreadLocal<DateFormat>() {
-  override fun initialValue(): DateFormat {
-    // Date format specified by RFC 7231 section 7.1.1.1.
-    return SimpleDateFormat("EEE, dd MMM yyyy HH:mm:ss 'GMT'", Locale.US).apply {
-      isLenient = false
-      timeZone = UTC
-    }
+private val STANDARD_DATE_FORMAT by lazy {
```
yschimke, line 32: This looks like its removing the thread safety that ThreadLocal provided.

yschimke (commented):
Is there some context here, is it showing up in profiling?

I'm inclined to close this PR, but maybe there is context I'm missing.

okhttp/src/jvmMain/kotlin/okhttp3/internal/http/dates.kt
Isira-Seneviratne: Kotlin's lazy delegates are thread-safe by default.

Isira-Seneviratne:
> Is there some context here, is it showing up in profiling?
> 
> I'm inclined to close this PR, but maybe there is context I'm missing.

Kotlin's lazy delegates provide similar functionality for lazily-initialized values, and are more concise.

okhttp/src/jvmMain/kotlin/okhttp3/internal/http/dates.kt
yschimke: I believe its thread safe for construction, but you will then share an instance. I'm not clear how lazy replaces a ThreadLocal.

PR #7599, opened 2022-12-31
Downgrade Jetty to 9.4.50.v20221201

swankjesse:
This version runs on JDK 1.8.


yschimke:
Ooopss, I'm a bit to approve happy. Good catch.

PR #7602, opened 2022-12-31
Get more TLS stuff passing on BouncyCastle

swankjesse:
There's something up with ECDSA on BouncyCastle, and rather than figuring it out I've just switched to RSA signatures with that provider.


swankjesse:
Needs to be rebased  on https://github.com/square/okhttp/pull/7599

okhttp-testing-support/src/main/kotlin/okhttp3/testing/PlatformRule.kt
```diff
@@ -323,6 +327,13 @@ open class PlatformRule @JvmOverloads constructor(
     }
   }
 
+  fun localhostHandshakeCertificates(): HandshakeCertificates {
+    return when {
+      isBouncyCastle() -> localhostHandshakeCertificatesWithRsa2048
```
yschimke, line 332: Nice

yschimke (approved):

PR #7605, opened 2022-12-31
Use mapNotNull() extension.

Isira-Seneviratne:
Use Kotlin's `mapNotNull` extension in `Cookie.parseAll()`.

### DIFF
--- a/okhttp/src/jvmMain/kotlin/okhttp3/Cookie.kt
+++ b/okhttp/src/jvmMain/kotlin/okhttp3/Cookie.kt
@@ -608,20 +608,8 @@ class Cookie private constructor(
     /** Returns all of the cookies from a set of HTTP response headers. */
     @JvmStatic
     fun parseAll(url: HttpUrl, headers: Headers): List<Cookie> {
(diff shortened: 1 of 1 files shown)


yschimke:
@swankjesse no objection from me.

But I seem to remember we avoided (early) kotlin loops because of allocations, is that a concern here? the current implementation seems allocation free for the no headers case which is the majority.

JakeWharton:
The empty case will allocate an iterator, a backing mutable list, and an unmodifiable wrapper. Before it allocated nothing.

The non-empty case adds an additional iterator allocation to the previous code.

yschimke:
Going to close this, while I don't have a strong objection, I'd be just as likely to accept the reverse as an optimisation.

yschimke:
@Isira-Seneviratne could you raise an issue to discuss future changes before going ahead? I feel bad closing these and you having lost some time on them.

Isira-Seneviratne:
> @Isira-Seneviratne could you raise an issue to discuss future changes before going ahead? I feel bad closing these and you having lost some time on them.

Sure.

Issue #7603, opened 2022-12-31
Tests flake on GitHub actions because hostnames are fetched from DNS


swankjesse:
I [ran tests](https://github.com/square/okhttp/actions/runs/3813128891/jobs/6486697956) on Windows. That failed for a surprising reason:

```
MockWebServerTest > requestUrlReconstructed() FAILED
    org.opentest4j.AssertionFailedError: 
    expected: "MiningMadness.com"
     but was: "miningmadness.com"
        at java.base@11.0.17/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
        at java.base@11.0.17/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
        at java.base@11.0.17/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
        at app//okhttp3.mockwebserver.MockWebServerTest.requestUrlReconstructed(MockWebServerTest.java:471)
```

Perhaps my GitHub actions worker’s IP address has a DNS record, maybe some madness related to mining? We’re doing `InetAddress.getCanonicalHostName()` and that might do a DNS lookup. (Yuck.)

yschimke:
Github has a hosts file with thousands of additional hosts blackholed to local host.

yschimke:
https://github.com/square/okhttp/issues/6813

PR #7607, opened 2023-01-01
Avoid InetAddress.getCanonicalHostName().

swankjesse:
Just hard-coding 'localhost' is simpler and avoids problems where the host machine's IP address has additional DNS registrations.

Closes: https://github.com/square/okhttp/issues/7603


yschimke:
Is this safe for https://github.com/square/okhttp/pull/3547 still?

yschimke (approved):

swankjesse:
Yep. Now we get the hostname from the Host header.

PR #7608, opened 2023-01-01
Get rid of MwsDuplexAccess

swankjesse:
Its unnecessary now that we have a public API type for streams


yschimke (approved):
This is nice.

PR #7611, opened 2023-01-02
Compile all on checkjdk11

yschimke:
Flag things like https://github.com/square/okhttp/pull/7600 earlier.


yschimke:
@swankjesse request post review

PR #7612, opened 2023-01-02
Check for matching ConnectionEvent

yschimke:
Refactor `CallEvent.closes` and take same approach for ConnectionEvent.

### DIFF
--- a/okhttp-testing-support/src/main/kotlin/okhttp3/OkHttpClientTestRule.kt
+++ b/okhttp-testing-support/src/main/kotlin/okhttp3/OkHttpClientTestRule.kt
@@ -173,7 +173,9 @@ class OkHttpClientTestRule : BeforeEachCallback, AfterEachCallback {
       // a test timeout failure.
       val waitTime = (entryTime + 1_000_000_000L - System.nanoTime())
       if (!queue.idleLatch().await(waitTime, TimeUnit.NANOSECONDS)) {
-        TaskRunner.INSTANCE.cancelAll()
+        synchronized (TaskRunner.INSTANCE) {
+          TaskRunner.INSTANCE.cancelAll()
+        }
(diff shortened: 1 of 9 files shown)


okhttp-testing-support/src/main/kotlin/okhttp3/OkHttpClientTestRule.kt
```diff
@@ -173,7 +173,9 @@ class OkHttpClientTestRule : BeforeEachCallback, AfterEachCallback {
       // a test timeout failure.
       val waitTime = (entryTime + 1_000_000_000L - System.nanoTime())
       if (!queue.idleLatch().await(waitTime, TimeUnit.NANOSECONDS)) {
-        TaskRunner.INSTANCE.cancelAll()
+        synchronized (TaskRunner.INSTANCE) {
```
yschimke, line 176: The fail wasn't happening, because cancelAll fails.

swankjesse: oooh good catch

okhttp-testing-support/src/main/kotlin/okhttp3/RecordingConnectionListener.kt
```diff
@@ -106,15 +114,34 @@ open class RecordingConnectionListener : ConnectionListener() {
         .isFalse()
     }
 
+    if (enforceOrder) {
```
swankjesse, line 117: Nice.

swankjesse (approved):
This is a net improvement, but it doesn’t check as much as we could.

In particular, this confirms every close has a matching open, but not that every open has a matching close.

swankjesse:
(I think we probably need a `@Rule` for that!)

PR #7618, opened 2023-01-02
Fix JavaHttpClientTest tests on JDK 19

yschimke:
Invert the logic for testing Content Length of JavaHttpClientTest.


swankjesse (commented):
LGTM

swankjesse:
Ahhh.

PR #7630, opened 2023-01-04
Make tolerant of JDK 19 quirks

yschimke:
See failure https://github.com/square/okhttp/actions/runs/3834111360/jobs/6526194017

```
CertificatePinnerChainValidationTest[jvm] > unrelatedPinnedIntermediateCertificateInChain()[jvm] FAILED
    org.opentest4j.AssertionFailedError: JDK with version from 19 expected to fail with exception that ANYTHING
        at app//org.junit.jupiter.api.AssertionUtils.fail(AssertionUtils.java:38)`
```

### DIFF
--- a/okhttp/src/jvmTest/java/okhttp3/internal/tls/CertificatePinnerChainValidationTest.java
+++ b/okhttp/src/jvmTest/java/okhttp3/internal/tls/CertificatePinnerChainValidationTest.java
@@ -15,21 +15,8 @@
  */
 package okhttp3.internal.tls;
 
-import java.security.GeneralSecurityException;
-import java.security.SecureRandom;
-import java.security.cert.X509Certificate;
-import java.util.Collections;
-import javax.net.ssl.KeyManager;
-import javax.net.ssl.SSLContext;
-import javax.net.ssl.SSLHandshakeException;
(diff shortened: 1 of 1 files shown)


okhttp/src/jvmTest/java/okhttp3/internal/tls/CertificatePinnerChainValidationTest.java
```diff
@@ -182,7 +184,9 @@ public void setup(MockWebServer server) {
     // https://github.com/square/okhttp/issues/4729
     platform.expectFailureOnConscryptPlatform();
     platform.expectFailureOnCorrettoPlatform();
-    platform.expectFailureFromJdkVersion(19);
+
+    // unpredictable as to whether it passes or fails? Environmental?
+    Assumptions.assumeTrue(PlatformVersion.INSTANCE.getMajorVersion() < 19);
```
swankjesse, line 189: What happens if you remove this completely? Perhaps JDK 19 just changed?

yschimke: I'll try. But flipped in a day or so.

PR #7634, opened 2023-01-05
[4.x] Backport 1XX handling from master

yschimke:
In progress, test is difficult without informational responses.

### DIFF
--- a/okhttp/src/test/java/okhttp3/InformationalResponseCodeTest.kt
+++ b/okhttp/src/test/java/okhttp3/InformationalResponseCodeTest.kt
@@ -1,3 +1,18 @@
+/*
+ * Copyright (C) 2023 Block, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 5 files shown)


yschimke:
@swankjesse I think a mostly manual test is the best I can do.

okhttp/src/test/java/okhttp3/InformationalResponseCodeTest.kt
```diff
@@ -0,0 +1,56 @@
...
+ */
+package okhttp3
+
+import okhttp3.testing.PlatformRule
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.Ignore
+import org.junit.Rule
+import org.junit.Test
+
+@Ignore("For manual testing only")
```
yschimke, line 24: Disabled as we don't want to test an external site.

And we don't want to backport changes to the mockwebserver in 5.x

swankjesse (approved):

PR #7648, opened 2023-01-11
[4.x] Cherry pick: fix DoH error when using ip address as hostname

yschimke:
From https://github.com/square/okhttp/pull/7073/files


yschimke:
@swankjesse ran tests locally, but any idea why this is not running in CI?

https://github.com/square/okhttp/blob/okhttp_4.10.x/.github/workflows/build.yml

swankjesse:
I added a label and that kicked it?

Possibly needed one of these to trigger?
```
  pull_request:
    types: [opened, labeled, unlabeled, synchronize]
```

swankjesse (approved):

PR #7651, opened 2023-01-14
Revert "Use Date.from(). (#7516)"

swankjesse:
This reverts commit 55c3013b81c3c985d2797ae7dc48401b1fe72ae9.

### DIFF
--- a/okhttp/src/jvmMain/kotlin/okhttp3/Headers.kt
+++ b/okhttp/src/jvmMain/kotlin/okhttp3/Headers.kt
@@ -45,6 +45,8 @@ import okhttp3.internal.commonValues
 import okhttp3.internal.headersCheckName
 import okhttp3.internal.http.toHttpDateOrNull
 import okhttp3.internal.http.toHttpDateString
+import okhttp3.internal.http2.Header
+import org.codehaus.mojo.animal_sniffer.IgnoreJRERequirement
 
 @Suppress("NAME_SHADOWING")
 actual class Headers internal actual constructor(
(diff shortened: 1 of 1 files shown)


okhttp/src/jvmMain/kotlin/okhttp3/Headers.kt
```diff
@@ -180,25 +184,35 @@ actual class Headers internal actual constructor(
...
+    }
 
     /**
      * Add a header with the specified name and formatted instant. Does validation of header names
      * and value.
      */
-    fun add(name: String, value: Instant) = add(name, Date.from(value))
+    @IgnoreJRERequirement
+    fun add(name: String, value: Instant) = apply {
+      add(name, Date(value.toEpochMilli()))
```
JakeWharton, line 197: Date.from is available if you have already have access to Instant

JakeWharton (commented):
Adding the annotations makes sense, but why change the implementation? Both Instant and Date.from are available on the same API level.

okhttp/src/jvmMain/kotlin/okhttp3/Headers.kt
swankjesse: Great point. So maybe this is not a problem at all.

(I assume Animal Sniffer is still broken, as the `@IgnoreJRERequirement` annotations should be required?)

JakeWharton: Yep. It should be failing since the signature is set to API 21 or something.

swankjesse:
Obsoleted by https://github.com/square/okhttp/pull/7672

PR #7653, opened 2023-01-15
Update readme with additional optional requirements for Android.

Isira-Seneviratne:
Mention in the readme that core library desugaring is required if the `java.time` methods are used on API levels below 26.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,8 @@ Requirements
 
 OkHttp works on Android 5.0+ (API level 21+) and Java 8+.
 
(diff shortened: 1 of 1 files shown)


yschimke (approved):

swankjesse (changes requested):
No, we do *not* want our users to desugar. The current behavior is a bug!

swankjesse:
(These APIs are only for people on API 26+; nobody needs to desugar.)

Issue #7654, opened 2023-01-16
Update to newer `com.squareup.okio:okio-jvm` version in `v4`


mind-overflow:
The latest available stable OkHttp release (`v4.10.0`) depends on `com.squareup.okio:okio-jvm` `v3.0.0`:  
```
[INFO] |  +- com.squareup.okhttp3:okhttp:jar:4.10.0:compile
[INFO] |  |  +- com.squareup.okio:okio-jvm:jar:3.0.0:compile
[INFO] |  |  |  +- org.jetbrains.kotlin:kotlin-stdlib-jdk8:jar:1.5.31:compile
[INFO] |  |  |  |  \- org.jetbrains.kotlin:kotlin-stdlib-jdk7:jar:1.5.31:compile
[INFO] |  |  |  \- org.jetbrains.kotlin:kotlin-stdlib-common:jar:1.5.31:compile
[INFO] |  |  \- org.jetbrains.kotlin:kotlin-stdlib:jar:1.6.20:compile
[INFO] |  |     \- org.jetbrains:annotations:jar:13.0:compile
```  
`okio-jvm` `v3.0.0` imports the library `org.jetbrains.kotlin:kotlin-stdlib-jdk7` `v1.5.31`, that is deemed as vulnerable in CVE-2022-24329.  
Previous issues (#7217) have already mentioned this CVE, which was supposedly fixed in okhttp `v4.10.0`. However, due to the dependency changes (from `com.squareup.okio:okio` `v2.8.0` to `com.squareup.okio:okio-jvm` `v3.0.0`) the issue persists, because now the concerned library `org.jetbrains.kotlin:kotlin-stdlib-jdk7` `v1.5.31` is imported by `okio-jvm`, which previously wasn't.  
  
`com.squareup.okio:okio-jvm` `v3.3.0` imports a more recent version of `org.jetbrains.kotlin:kotlin-stdlib-jdk7`, so it doesn't get flagged. However, there is no `v4` version of okhttp that imports `okio-jvm` `v3.3.0`.  
   
Code analysis tools like SonarQube with OWASP Dependency-Check recognize this vulnerable library and mark okhttp as vulnerable too. I know this is a false positive, but having multiple projects come up with the same issue is not ideal when updating a single dependency can solve it - plus, it's generally never bad to switch to newer versions when possible, if they don't have breaking changes. Given that okhttp `v5` is in alpha development, it would be nice to have a non-flagged `v4` version available to use in the meanwhile.  
  
Thank you!

[...]

PR #7660, opened 2023-01-18
Add a Kotlin dependency to work around CVE-2022-24329

yschimke:
fixes https://github.com/square/okhttp/issues/7654

### DIFF
--- a/okhttp/build.gradle
+++ b/okhttp/build.gradle
@@ -24,6 +24,8 @@ task copyJavaTemplates(type: Copy) {
 dependencies {
   api deps.okio
   api deps.kotlinStdlib
+  // Only imported to avoid broken vulnerability scanners for CVE-2022-24329
+  api deps.kotlinStdlibJdk8
   compileOnly deps.android
   compileOnly deps.bouncycastle
   compileOnly deps.bouncycastletls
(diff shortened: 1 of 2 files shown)


yschimke:
That didn't work because okio is kotlin 1.8. Not sure I want to bring that in for OkHttp 4.x at this time.

So took a different approach.

okhttp/build.gradle
```diff
@@ -24,6 +24,8 @@ task copyJavaTemplates(type: Copy) {
 dependencies {
   api deps.okio
   api deps.kotlinStdlib
+  // Only imported to avoid broken vulnerability scanners for CVE-2022-24329
```
swankjesse, line 27: Got a link to a vulnerability report?

swankjesse:
Found it. https://github.com/square/okhttp/issues/7654

okhttp/build.gradle
swankjesse:
```suggestion
  // Only imported to avoid broken vulnerability scanners for CVE-2022-24329
  // https://github.com/square/okhttp/issues/7654
```

swankjesse (approved):

PR #7661, opened 2023-01-19
fix Cookie.expiresAt doc

fsbarata:
The Cookie.expiresAt doc is currently worded in a way that implies persistent is the case of expiring at the end of the session.

As it says in the Cookie.persistent javadoc, it is the contrary of a session expiring cookie.


swankjesse (commented):
Thanks!

fsbarata:
My pleasure!

PR #7664, opened 2023-01-21
Use JvmSynthetic instead of dash JvmName.

yschimke:
https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.jvm/-jvm-synthetic/


yschimke:
Fails with

```
e: file:///Users/yschimke/IdeaProjects/okhttp/okhttp/src/jvmMain/kotlin/okhttp3/Address.kt:38:3 Platform declaration clash: The following declarations have the same JVM signature (dns()Lokhttp3/Dns;):
    fun `<get-dns>`(): Dns defined in okhttp3.Address
    fun dns(): Dns defined in okhttp3.Address
```

PR #7665, opened 2023-01-21
Experiment with Dispatcher Executor

yschimke:


swankjesse:
Discussed a bit offline.

I’m reluctant to lean in to ExecutorService APIs here, because I think it’s an area filled with footguns. In particular I’m worried that OkHttp sharing a common thread pool with unrelated systems potentially starves OkHttp of threads. Also OkHttp can potentially starve those other systems of threads!

For Loom I think we should detect a stable API (JDK 20?) and enable it silently.

Issue #7652, opened 2023-01-14
Animal Sniffer not working


swankjesse:
It didn’t catch an API incompatibility with `Date.from()`. Perhaps a conflict with Multiplatform?

https://github.com/square/okhttp/pull/7651

PR #7672, opened 2023-01-29
Restore AnimalSniffer support

swankjesse:
It wasn't working because we weren't configuring the dependency type correctly.

https://github.com/xvik/gradle-animalsniffer-plugin/issues/64

Closes: https://github.com/square/okhttp/issues/7652

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -111,8 +111,19 @@ subprojects {
 
   val signature: Configuration by configurations.getting
   dependencies {
-    signature(rootProject.libs.signature.android.apilevel21)
-    signature(rootProject.libs.codehaus.signature.java18)
+    // No dependency requirements for testing-support.
+    if (project.name == "okhttp-testing-support") return@dependencies
+
+    if (project.name == "mockwebserver3-junit5") {
+      // JUnit 5's APIs need java.util.function.Function and java.util.Optional from API 24.
+      signature(rootProject.libs.signature.android.apilevel24) { artifact { type = "signature" } }
+    } else {
+      // Everything else requires Android API 21+.
+      signature(rootProject.libs.signature.android.apilevel21) { artifact { type = "signature" } }
+    }
+
+    // OkHttp requires Java 8+.
+    signature(rootProject.libs.codehaus.signature.java18) { artifact { type = "signature" } }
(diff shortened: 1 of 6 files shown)


samples/tlssurvey/src/main/kotlin/okhttp3/survey/ssllabs/UserAgentCapabilities.kt
```diff
@@ -4,7 +4,7 @@ import com.squareup.moshi.JsonClass
 
 
 @JsonClass(generateAdapter = true)
-data class UserAgentCapabilities(
+class UserAgentCapabilities(
```
swankjesse, line 7: As a data class it ends up generating a class that calls `Integer.hashCode(int)`

build.gradle.kts
```diff
@@ -111,8 +111,19 @@ subprojects {
...
+
+    if (project.name == "mockwebserver3-junit5") {
+      // JUnit 5's APIs need java.util.function.Function and java.util.Optional from API 24.
+      signature(rootProject.libs.signature.android.apilevel24) { artifact { type = "signature" } }
+    } else {
+      // Everything else requires Android API 21+.
+      signature(rootProject.libs.signature.android.apilevel21) { artifact { type = "signature" } }
+    }
+
+    // OkHttp requires Java 8+.
```
yschimke, line 125: Maybe worth a comment that we compile against 18 (in theory) or 11 (in practice) but then run against 8+?

As is the comment is confusing because 8+ and java18 contradict.

yschimke (approved):

build.gradle.kts
JakeWharton: I believe 18 here is actually 1.8

samples/tlssurvey/src/main/kotlin/okhttp3/survey/ssllabs/UserAgentCapabilities.kt
JakeWharton: That's desugared to API 1 since like AGP 3.0, but probably easier to just avoid if it's not in public API

JakeWharton (approved):

PR #7685, opened 2023-02-07
Remove outdated notice from README

Jawnnypoo:
This date has long since past, so lets just keep it as the general warning to not use this version.


yschimke:
Thanks.

PR #7691, opened 2023-02-13
[RFC] Experiment with Android focused Client Builder

yschimke:
There a a bunch of things all android app developers should be thinking of.

- Cache
- Enforcing HTTPS
- logging
- tracing

Consider an API to make more of this the default


yschimke:
Will continue in https://github.com/square/okhttp/pull/7750

PR #7694, opened 2023-02-14
Refine doc and comments about how to run android test

TangHuaiZhe:
Refine doc and comments about how to run android test.
We need add an parameter `-PandroidBuild=true` so Gradle can know about `android-test ` project.


yschimke (approved):
Thanks

PR #7716, opened 2023-02-18
Avoid failing on Loom.

yschimke:


yschimke:
@swankjesse requesting post review.  Want green builds, and we already exclude this in various scenarios for the same error.

I am surprised that loom is stricter than JDK 19.

swankjesse (commented):
LGTM

PR #7717, opened 2023-02-18
Don't publish junit results for check job

yschimke:
Currently shows a failed build after check finishes, but before the next job with tests completes.


swankjesse (commented):
LGTM

PR #7722, opened 2023-02-22
Hide the internal okhttp immutable methods

Jawnnypoo:
In Android Studio/IntelliJ, if you are looking to convert a list to an immutable list, you start to type `toImmu` and then it shows:

Screenshot 2023-02-22 at 2 51 51 PM

Many times, I find myself accidentally selecting the okhttp3 internal implementation. 

My first assumption is that you all are using an internal implementation instead of kotlinx so as to not add another dependency within okhttp. My second assumption, which I found to be true through testing, is that these functions cannot actually use the `internal` modifier. 

So, the solution to me would be to have some prefix that others are unlikely to type, such as `okToImmutableList`. I'll admit though that when actually calling this function within these OkHttp libraries does look a bit ugly. 

I am open to suggestions on other prefixes, or other ways to solve this! Just a pretty annoying thing that I am sure a lot of folks run in to with trying to create an immutable list in their Kotlin projects that use OkHttp.

### DIFF
--- a/mockwebserver/src/main/kotlin/mockwebserver3/MockWebServer.kt
+++ b/mockwebserver/src/main/kotlin/mockwebserver3/MockWebServer.kt
@@ -76,10 +76,10 @@ import okhttp3.internal.http2.ErrorCode
 import okhttp3.internal.http2.Header
 import okhttp3.internal.http2.Http2Connection
 import okhttp3.internal.http2.Http2Stream
-import okhttp3.internal.immutableListOf
+import okhttp3.internal.okImmutableListOf
 import okhttp3.internal.platform.Platform
 import okhttp3.internal.threadFactory
-import okhttp3.internal.toImmutableList
+import okhttp3.internal.okToImmutableList
 import okhttp3.internal.ws.RealWebSocket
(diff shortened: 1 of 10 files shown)


yschimke:
@swankjesse I'm tempted to accept any PRs that avoid these external distractions.

yschimke:
What about if we move these from top level into something like

```
object OkHttpInternal {
  fun <T> List<T>.toImmutableList(): List<T> {
    return Collections.unmodifiableList(toMutableList())
  }
}
```

JakeWharton:
How about mark them as `internal` and then suppress the error in other modules which need to use them.

yschimke:
> How about mark them as `internal` and then suppress the error in other modules which need to use them.

Yep, that works with

`@file:Suppress("INVISIBLE_MEMBER", "INVISIBLE_REFERENCE")`

yschimke:
@Jawnnypoo Do you want to try this approach? Or happy for me to make it?

yschimke (changes requested):
Let's change to internal methods, and suppress when used from other modules.

Issue #7724, opened 2023-02-27
okhttp-sse RealCall CastException


ywang-nr:
**Summary of the problem**

1. New Relic is trying to add support for okhttp-sse and found out one part of the code for RealCall causes cast exception.
> The New Relic Android agent provides performance monitoring instrumentation, we use ASM to perform byte-code injection to collect data from performance issues to tiny errors within the code.

> When integrate both New Relic android agent and okhttp-sse library, we found a crash in `RealEventSource` class file at this line--> `call = client.newCall(request) as RealCall`

**Why are we getting the crash and potential solutions:**
1. RealCall cannot be implemented as it's a final class - Solution 1: Can we adjust RealCall class to be public?
2. Casting newCall as RealCall causes the crash `call = client.newCall(request) as RealCall` - Solution 2: Can we adjust this part of the casting to use a public class?

Above are what I have found out so far, any insights are appreciated. Thank you.

yschimke:
Is there a New Relic bug, or PR you can point us at? Or a stack trace for the class cast exception?

Is it this? https://github.com/newrelic/newrelic-java-agent/blob/3e19dce8fc731279583e8507e5ff47dd9bea0ab6/instrumentation/okhttp-4.4.0/src/main/java/com/nr/agent/instrumentation/okhttp44/internal/RealCall_Instrumentation.java

On 2), I'm a bit sympathetic to OkHttp itself not assuming newCall returns RealCall https://github.com/square/okhttp/blob/okhttp_4x/okhttp-sse/src/main/kotlin/okhttp3/internal/sse/RealEventSource.kt#L41 but I'm not sure what the right fix would be.

I think having a Call.Factory that returns other call implementations is a supported part of the API.  In this case we use OkHttp directly, so casting was considered safe.  But I think we should fix this.

But it would be good to work through this, we aren't against monitoring tools. One way might be to have a public API that let's us unwrap types like this. Say DelegatedCall, with an unwrap method?

In terms of making RealCall public, this is precisely why the API is controlled, we are strict about not breaking our public API, and tools that use internal APIs will break frequently, because we do change those internals. It's been vital in letting us evolve the project.  So it won't be made public API or non final. 

cc @swankjesse @JakeWharton thoughts?

ywang-nr:
Thank you for taking the time to review this. 

**To answer your questions first:**
1. There isn't a bug or PR I can point to you at the moment, but I can provide a sample app to reproduce the issue. Sample app and logs are attached. 
[...]

PR #7728, opened 2023-03-04
Avoid assuming RealCall

yschimke:
Two aims

- 1 avoid needing to cast to RealCall within OkHttp
- Support a clean decorator mechanism for libraries that decorate Call.Factory of WebSocket.Factory

### DIFF
--- a/okhttp-sse/api/okhttp-sse.api
+++ b/okhttp-sse/api/okhttp-sse.api
@@ -21,3 +21,14 @@ public final class okhttp3/sse/EventSources {
 	public static final fun processResponse (Lokhttp3/Response;Lokhttp3/sse/EventSourceListener;)V
 }
 
+public final class okhttp3/sse/EventSources$EventSourceRequest {
+	public fun <init> (Lokhttp3/sse/EventSourceListener;)V
+	public final fun component1 ()Lokhttp3/sse/EventSourceListener;
+	public final fun copy (Lokhttp3/sse/EventSourceListener;)Lokhttp3/sse/EventSources$EventSourceRequest;
+	public static synthetic fun copy$default (Lokhttp3/sse/EventSources$EventSourceRequest;Lokhttp3/sse/EventSourceListener;ILjava/lang/Object;)Lokhttp3/sse/EventSources$EventSourceRequest;
+	public fun equals (Ljava/lang/Object;)Z
+	public final fun getListener ()Lokhttp3/sse/EventSourceListener;
+	public fun hashCode ()I
+	public fun toString ()Ljava/lang/String;
+}
+
--- a/okhttp-sse/src/main/kotlin/okhttp3/sse/EventSources.kt
+++ b/okhttp-sse/src/main/kotlin/okhttp3/sse/EventSources.kt
@@ -42,6 +42,10 @@ object EventSources {
     eventSource.processResponse(response)
   }
 
+  /**
+   * Request tag that indicates the request is a SSE request,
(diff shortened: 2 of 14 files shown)


ywang-nr:
Hi, @yschimke, thank you for putting up the PR! Would you have an ETA when the PR will be merged and released? Thank you.

ywang-nr:
@yschimke Could you help to resolve the branch conflict? Clients are trying to use this branch to test.

Also could you provide an ETA in mind to merge this branch? Thanks.

yschimke:
Yep, will do.

 In order to land, it needs more feedback from maintainers and possibly API changes.

ywang-nr:
> Yep, will do.
> 
> In order to land, it needs more feedback from maintainers and possibly API changes.

That's great, Thank you for the help!

In terms of "more feedback from maintainers and possibly API changes", are you talking about on okhttp side or is there anything NR can help?

swankjesse:
It’s on me.

apurvCoin:
@swankjesse @yschimke   Please help and resolve the merge conflicts .. and expedite this. its really blocking to integrate NR usage  with various 3rd party libraries 
@ywang-nr Is there any workaround to get this working . the latest nr-android-agent also not working with its new feature .

swankjesse:
(NR = NewRelic)

ywang-nr:
@swankjesse Would you be open to have a call with NR and our customers to troubleshoot a work-around?

**Here are more background:**
Upon receiving customer's feedback and after our own investigation, NR has reached out on Feb 27th [here](https://github.com/square/okhttp/issues/7724#issuecomment-1556160177)

Thanks to @yschimke drafted the PR but it hasn't been merged since May. We have asked our clients to try out the PR but it's hard to coordinate with 3rd party libraries, which we have multiple cases open with salesforce-android-sdk.

It would be really helpful if we can get an idea whether this PR will be merged or not, an ETA would be the best. Otherwise, we can get an idea how a potential work-around can be made on NR side. Thank you.

JustinRush-NR:
Another vote here from a New Relic engineering manager: Anything we can do to help get this PR further vetted and merged to unblock multiple parties?

swankjesse:
I’ve got an alternative approach for this.

Currently `RealEventSource` must downcast to `RealCall` because our `Call` interface isn’t rich enough to express `timeoutEarlyExit()`, a function that cancels the call timeout’s timer when we promote a short-lived HTTP call into a long-lived SSE stream. (We can’t get rid of the timeout entirely, because we need the setup HTTP call to fail if it takes too long.)

And `timeoutEarlyExit()` is necessary because Okio’s `Timeout` doesn’t have a mechanism to asynchronously cancel a timeout via its public API. Instead, we hack around it in `RealCall` by calling `exit()` via two different code paths and keeping track which one we used.

@oldergod had a great insight here: `Timeout` should have a cancel API! Then `RealEventSource` can use that directly, as can everything else that’s using `timeoutEarlyExit()`. We don’t need to downcast, and we make it possible for external users to build stuff like `RealEventSource` through our public API alone.

Next steps:
 * Introduce `Timeout.cancel()` API in Okio
 * Ship an Okio release
 * Adopt that new API in OkHttp

JustinRush-NR:
@swankjesse Thanks! Any guess at a timeline or is there another PR yet for that approach?

swankjesse:
@JustinRush-NR I’m working on it currently. https://github.com/square/okio/pull/1391

swankjesse:
More! https://github.com/square/okhttp/pull/8141

PR #7734, opened 2023-03-05
Avoid exposing internal methods

yschimke:
Testing an alternative to https://github.com/square/okhttp/pull/7722


okhttp-logging-interceptor/src/main/kotlin/okhttp3/logging/HttpLoggingInterceptor.kt
```diff
@@ -211,7 +212,7 @@ class HttpLoggingInterceptor @JvmOverloads constructor(
           }
         }
 
-        val charset: Charset = requestBody.contentType().charset()
+        val charset: Charset = requestBody.contentType().charsetNonNull()
```
swankjesse, line 215: Isn’t !! the way to go here?

okhttp/src/jvmMain/kotlin/okhttp3/internal/internal.kt
```diff
@@ -85,13 +85,13 @@ fun MediaType?.chooseCharset(): Pair<Charset, MediaType?> {
   return charset to finalContentType
 }
 
-fun MediaType?.charset(defaultValue: Charset = Charsets.UTF_8): Charset {
+internal fun MediaType?.charsetNonNull(defaultValue: Charset = Charsets.UTF_8): Charset {
```
swankjesse, line 88: I think this should be `charsetOrUtf8`

swankjesse (approved):

okhttp-logging-interceptor/src/main/kotlin/okhttp3/logging/HttpLoggingInterceptor.kt
yschimke: I'm not clear, we know it's non null here.  Will follow up if you are convinced.

PR #7747, opened 2023-03-23
"java.nio.Files#delete" should be preferred

rafche:
When java.io.File#delete fails, this boolean method simply returns false with no indication of the cause. On the other hand, when java.nio.file.Files#delete fails, this void method returns one of a series of exception types to better indicate the cause of the failure. And since more information is generally better in a debugging situation, java.nio.file.Files#delete is the preferred option.


yschimke:
Are you calling the right method? File vs Files?

yschimke:
Is this a real issue you have hit? Or just improving the sample?

rafche:
hi, i just want to improve, it 's not necessary

yschimke:
Not compiling, and not clear we need this for the sample.

PR #7758, opened 2023-04-07
Fix invalid port assignment in HttpUrl.Builder.effectivePort and authority

lvcca:
Throw IllegalArgumentException if invalid port is used in HttpUrl.Builder.effectivePort() and authority

### DIFF
--- a/okhttp/src/commonMain/kotlin/okhttp3/internal/-HttpUrlCommon.kt
+++ b/okhttp/src/commonMain/kotlin/okhttp3/internal/-HttpUrlCommon.kt
@@ -602,7 +602,18 @@ internal object CommonHttpUrl {
   }
 
   internal fun HttpUrl.Builder.effectivePort(): Int {
-    return if (port != -1) port else HttpUrl.defaultPort(scheme!!)
+    if (port >= 1 && port <= 65535){
+      return port
+    }
+    else {
+      val defaultPort: Int = HttpUrl.defaultPort(scheme!!)
+      if (defaultPort != -1){
+        return defaultPort
+      }
+      else {
+        throw IllegalArgumentException("Invalid port: $port")
+      }
+    }
   }
   internal fun HttpUrl.Builder.commonToString(): String {
     return buildString {
@@ -761,7 +772,13 @@ internal object CommonHttpUrl {
               }
             } else {
               host = input.percentDecode(pos = pos, limit = portColonOffset).toCanonicalHost()
-              port = HttpUrl.defaultPort(scheme!!)
+              val defaultPort: Int = HttpUrl.defaultPort(scheme!!)
+              if (defaultPort != -1){
+                port = defaultPort
+              }
+              else {
+                throw IllegalArgumentException("Invalid port: $port")
+              }
             }
             require(host != null) {
               "$INVALID_HOST: \"${input.substring(pos, portColonOffset)}\""


okhttp/src/commonMain/kotlin/okhttp3/internal/-HttpUrlCommon.kt
```diff
@@ -761,7 +772,13 @@ internal object CommonHttpUrl {
               }
             } else {
               host = input.percentDecode(pos = pos, limit = portColonOffset).toCanonicalHost()
-              port = HttpUrl.defaultPort(scheme!!)
+              val defaultPort: Int = HttpUrl.defaultPort(scheme!!)
```
yschimke, line 775: I'm not sure this is needed since we have already validated scheme is http or https.

okhttp/src/commonMain/kotlin/okhttp3/internal/-HttpUrlCommon.kt
```diff
@@ -602,7 +602,18 @@ internal object CommonHttpUrl {
...
+    if (port >= 1 && port <= 65535){
+      return port
+    }
+    else {
+      val defaultPort: Int = HttpUrl.defaultPort(scheme!!)
+      if (defaultPort != -1){
+        return defaultPort
+      }
+      else {
+        throw IllegalArgumentException("Invalid port: $port")
```
yschimke, line 614: Can you add a test for this case to HttpUrlTest?

okhttp/src/commonMain/kotlin/okhttp3/internal/-HttpUrlCommon.kt
```diff
@@ -602,7 +602,18 @@ internal object CommonHttpUrl {
   }
 
   internal fun HttpUrl.Builder.effectivePort(): Int {
-    return if (port != -1) port else HttpUrl.defaultPort(scheme!!)
+    if (port >= 1 && port <= 65535){
```
yschimke, line 605: I'm not sure I agree with this, why change a bad port like 65599 to 80/443?

yschimke:
Can you give more context? What are you hitting, why will this fix it?

Is it that some clients you don't control are hitting bad URLs, crashing your process?

lvcca (commented):
These changes add port validation in areas that could have assigned a -1 port. This issue was experienced in one of my projects in production.

okhttp/src/commonMain/kotlin/okhttp3/internal/-HttpUrlCommon.kt
yschimke: Shouldn't it fail immediately if the port is outside that range, and only use defaultPort for -1?

```
when (port) {
  -1 -> default port
  in 1..65535 -> port
  else -> throw
}
```

yschimke: But please add a test, so we can see it's fixing what you think is broken.  I'm worried if you can't reproduce it in a test.

okhttp/src/commonMain/kotlin/okhttp3/internal/-HttpUrlCommon.kt
yschimke: Sure, then please add a test that exercises these paths.

okhttp/src/commonMain/kotlin/okhttp3/internal/-HttpUrlCommon.kt
yschimke: Or post the stacktrace, and start with a bug report, instead of a PR.

PR #7761, opened 2023-04-10
ConnectionSpec: Fix custom specifications

moonfruit:
`ConnectionSpec` now support arbitrary input strings allows users to enable suites and versions that are not yet known to the library, but are supported by the platform. But this function is not implemented correctly.

This function is the key to how it works:
https://github.com/square/okhttp/blob/afcc2df23aa855220f0d532b404ec393d927520c/okhttp/src/jvmMain/kotlin/okhttp3/ConnectionSpec.kt#L94-L104

`apply()` function uses `specToApply.tlsVersions != null` to check if an arbitrary version is supported.
This is how `tlsVersions` is defined:
https://github.com/square/okhttp/blob/afcc2df23aa855220f0d532b404ec393d927520c/okhttp/src/jvmMain/kotlin/okhttp3/ConnectionSpec.kt#L74-L77

In this function, it converts TLS version string to `TlsVersion`. But an arbitrary input string cannot be converted to `TlsVersion`. It actually throws an exception. The same happens with `cipherSuites`.

And in `toStirng()`, it also uses `tlsVersions` and `cipherSuites`. This makes a `ConnectionSpec` with arbitrary input strings cannot be printed. If you create an exception described with `ConnectionSpec`, it fails again. It makes to know what happened.

I also add two string compatible getters for `tlsVersions` and `cipherSuites` to Java to help debug `ConnectionSpec`.

### DIFF
--- a/okhttp/src/jvmMain/kotlin/okhttp3/ConnectionSpec.kt
+++ b/okhttp/src/jvmMain/kotlin/okhttp3/ConnectionSpec.kt
@@ -16,6 +16,7 @@
 package okhttp3
 
 import java.util.Arrays
+import java.util.Collections
 import java.util.Objects
 import javax.net.ssl.SSLSocket
 import okhttp3.ConnectionSpec.Builder
@@ -71,6 +72,15 @@ class ConnectionSpec internal constructor(
    * Returns the TLS versions to use when negotiating a connection. Returns null if all of the SSL
    * socket's enabled TLS versions should be used.
    */
+  @get:JvmName("tlsVersionsAsString") val tlsVersionsAsStringList: List<String>?
+    get() {
+      return tlsVersionsAsString?.let { Collections.unmodifiableList(it.asList()) }
+    }
+
+  /**
+   * Returns the TLS versions to use when negotiating a connection. Returns null if all of the SSL
+   * socket's enabled TLS versions should be used.
+   */
   @get:JvmName("tlsVersions") val tlsVersions: List<TlsVersion>?
     get() {
       return tlsVersionsAsString?.map { TlsVersion.forJavaName(it) }
@@ -94,7 +104,7 @@ class ConnectionSpec internal constructor(
   internal fun apply(sslSocket: SSLSocket, isFallback: Boolean) {
     val specToApply = supportedSpec(sslSocket, isFallback)
 
-    if (specToApply.tlsVersions != null) {
+    if (specToApply.tlsVersionsAsString != null) {
       sslSocket.enabledProtocols = specToApply.tlsVersionsAsString
     }
 
@@ -190,9 +200,10 @@ class ConnectionSpec internal constructor(
   override fun toString(): String {
     if (!isTls) return "ConnectionSpec()"
 
+    val tlsVersionsToUse = tlsVersionsAsString?.map { TlsVersion.parseJavaName(it)?.toString() ?: it }
     return ("ConnectionSpec(" +
         "cipherSuites=${Objects.toString(cipherSuites, "[all enabled]")}, " +
-        "tlsVersions=${Objects.toString(tlsVersions, "[all enabled]")}, " +
+        "tlsVersions=${Objects.toString(tlsVersionsToUse, "[all enabled]")}, " +
         "supportsTlsExtensions=$supportsTlsExtensions)")
   }
 
--- a/okhttp/src/jvmTest/java/okhttp3/ConnectionSpecTest.java
+++ b/okhttp/src/jvmTest/java/okhttp3/ConnectionSpecTest.java
@@ -21,8 +21,11 @@ import okhttp3.testing.PlatformVersion;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.RegisterExtension;
 
+import javax.net.ssl.HandshakeCompletedListener;
+import javax.net.ssl.SSLSession;
 import javax.net.ssl.SSLSocket;
 import javax.net.ssl.SSLSocketFactory;
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Set;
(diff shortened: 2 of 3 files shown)


okhttp/src/jvmMain/kotlin/okhttp3/ConnectionSpec.kt
```diff
@@ -67,6 +76,15 @@ class ConnectionSpec internal constructor(
       level = DeprecationLevel.ERROR)
   fun cipherSuites(): List<CipherSuite>? = cipherSuites
 
+  /**
+   * Returns the TLS versions to use when negotiating a connection. Returns null if all of the SSL
+   * socket's enabled TLS versions should be used.
+   */
+  @get:JvmName("tlsVersionsAsString") val tlsVersionsAsStringList: List<String>?
+    get() {
+      return tlsVersionsAsString?.let { Collections.unmodifiableList(it.asList()) }
```
yschimke, line 85: I'm not sure the intent is to support arbitrary TlsVersion here. Let me check when we have previously discussed.

But I'm not necessarily against it.

Like with cipherSuite, is we do want to support, we should probably make the types open.

okhttp/src/jvmMain/kotlin/okhttp3/ConnectionSpec.kt
```diff
@@ -51,6 +51,15 @@ class ConnectionSpec internal constructor(
   private val tlsVersionsAsString: Array<String>?
 ) {
 
+  /**
+   * Returns the cipher suites to use for a connection. Returns null if all of the SSL socket's
+   * enabled cipher suites should be used.
+   */
+  @get:JvmName("cipherSuitesAsString") val cipherSuitesAsStringList: List<String>?
```
yschimke, line 58: This is new public api, which seems unfortunate.

From kotlin

```
connectionSpec.cipherSuitesAsStringList
```

From java

```
tlsSpec.cipherSuitesAsString()
```

okhttp/src/jvmMain/kotlin/okhttp3/ConnectionSpec.kt
```diff
@@ -51,6 +51,15 @@ class ConnectionSpec internal constructor(
...
+    get() {
+      return cipherSuitesAsString?.let { Collections.unmodifiableList(it.asList()) }
```
yschimke, line 60: I wonder why we need this, since `CipherSuite.forJavaName()` should create them on the fly.

See https://github.com/square/okhttp/blob/afcc2df23aa855220f0d532b404ec393d927520c/okhttp/src/jvmMain/kotlin/okhttp3/ConnectionSpec.kt#L60

and 

https://github.com/square/okhttp/blob/afcc2df23aa855220f0d532b404ec393d927520c/okhttp/src/jvmMain/kotlin/okhttp3/CipherSuite.kt#L426

okhttp/src/jvmTest/java/okhttp3/ConnectionSpecTest.java
```diff
@@ -182,10 +186,109 @@ public final class ConnectionSpecTest {
   @Test public void tls_stringCiphersAndVersions() throws Exception {
     // Supporting arbitrary input strings allows users to enable suites and versions that are not
     // yet known to the library, but are supported by the platform.
-    new ConnectionSpec.Builder(ConnectionSpec.MODERN_TLS)
+    ConnectionSpec tlsSpec = new ConnectionSpec.Builder(true)
```
yschimke, line 189: This should be passing already without changes above.

```
    assertThat(tlsSpec.cipherSuites()).isEqualTo(asList(CipherSuite.forJavaName("MAGIC-CIPHER")));
```

okhttp/src/jvmMain/kotlin/okhttp3/ConnectionSpec.kt
yschimke: With this change, anyone using the current tlsVersions will get a nasty surprise.

`tlsSpec.tlsVersions()`

failing with

```
java.lang.IllegalArgumentException: Unexpected TLS version: TLS9k
	at okhttp3.TlsVersion$Companion.forJavaName(TlsVersion.kt:47)
	at okhttp3.ConnectionSpec.tlsVersions(ConnectionSpec.kt:85)
```

okhttp/src/jvmMain/kotlin/okhttp3/ConnectionSpec.kt
```diff
@@ -191,8 +209,8 @@ class ConnectionSpec internal constructor(
     if (!isTls) return "ConnectionSpec()"
 
     return ("ConnectionSpec(" +
-        "cipherSuites=${Objects.toString(cipherSuites, "[all enabled]")}, " +
-        "tlsVersions=${Objects.toString(tlsVersions, "[all enabled]")}, " +
+        "cipherSuites=${cipherSuitesAsString?.let { it.contentToString() } ?: "[all enabled]"}, " +
+        "tlsVersions=${tlsVersionsAsString?.let { it.contentToString() } ?: "[all enabled]"}, " +
```
yschimke, line 213: I'm not sure how much we care, but the test below indicates it is behaviour changing.

yschimke:
TlsVersion is an enum class.  If we do want to support this, I think we should change it to be a regular String wrapper class, and understands what would break in the public API.

cc @swankjesse

yschimke:
In future, I'd suggest discussing this sort of PR first, in case it doesn't get accepted.

moonfruit:
I actually have an application scenario right now. This is why I make this pull request. In China there is a special communication encryption standard called TLCP or GMSSL. Some third party libraries like [Kona](https://github.com/Tencent/TencentKonaSMSuite) implement the JSSE interface.
But I can't use it with `okhttp`, but [Apache Http Client](https://hc.apache.org/index.html) has no problem.

yschimke:
If you change TlsVersion to a wrapper instead of enum, does it work without other code changes?

This sounds like something we should support in OkHttp 5

moonfruit:
Sorry, I have not noticed that `CipherSuite` is not an `enum`. I change my commit to remove modifications for `CipherSuite`. And I make `toString()` the same as original output for normal `TlsVersion`.

It would be nice if `TlsVersion` could be like `CipherSuite`, but if you want to maintain the existing API and support arbitrary `TlsVersion`s, all you need to do is change the `apply()` and `toString()` functions. Even the getter for `tlsVersionsAsString` can be excluded.

This modification now allows the API to remain the same for standard `TlsVersion` but to support arbitrary `TlsVersion`s at the same time. 

Maybe in the future, `okhttp` can change `TlsVersion` to a string wrapper. But for now, small changes only can support arbitrary `TlsVersion`s. I think it's worth to accept it.

moonfruit:
Yes, `tlsVersions` will fail for arbitrary `TlsVersion`s.
But if we don't change anything, it still fail for arbitrary `TlsVersion`s.
Nothing changes.

moonfruit:
Whether this is accepted or not, this will fail. So, nothing actually changes.
```java
ConnectionSpec connectionSpec = new ConnectionSpec.Builder(true)
        .tlsVersions("ABC")
        .build();
System.out.println(connectionSpec.tlsVersions())
```

moonfruit:
Please review this again. With this small changes, `okhttp` remains the API as before, and it supports arbitrary `TlsVersion`s .

yschimke:
Doesn't it still explode here?

```
      return tlsVersionsAsString?.map { TlsVersion.forJavaName(it) }
```

moonfruit:
Yes, it still throws an exception for unusual TLS versions. But it will work fine for TLS versions defined in `TlsVersion'.
With or without this change, this remains the same.
What's different is that if you specify unusual TLS versions, it will connect successfully. But now it fails.

yschimke:
Yes, but before it signaled an error and the overall request failed.  Now it's successful but fails as soon as you try to look at the results using the existing public API.

moonfruit:
This is what I wanted. This change is a compromise to the existing situation. Yes, `tlsVersions()` will fail, but `toString()` will return normally. I think this will be sufficient for debugging in most cases. This way, even special `TlsVersion`s can successfully make connections instead of rejecting them outright. This gives those who need this scenario a chance to use `okhttp`.
If, like me, you find this change less than ideal, adding a new public API: a java getter for `tlsVersionsAsString` will solve the problem. This will allow the pull request to be merged into 4.x, minimizing the need for 4.x to support special `TlsVersion`s as well.

moonfruit:
I add `tleVersionsAsString()` back, Is this OK? @yschimke

yschimke:
No, it's not new public API that we want to support.  

I discussed with @swankjesse, we agreed that if we want to enable this case, we should open up TlsVersion, implementing it similar to CipherSuite, and hopefully almost completely backwards compatible.  

Do you want me to apply on top of your PR? Or would you like to try making the change yourself?

yschimke:
@moonfruit we are trying to help you support this setup.  Won't a TlsVersion that supports TLCP and GMSSL solve your problem?

If it won't help then we shouldn't do it.

I understand your frustration, but OkHttp is widely used, so new public API is something we take seriously, and also the behaviour for existing clients. Say if a logging interceptor is printing out the TlsVersion.

yschimke:
> we should open up TlsVersion, implementing it similar to CipherSuite, and hopefully almost completely backwards compatible.

So it should be possible to write

```
  val t = TlsVersion("TLCPv1.1")
```

yschimke:
@moonfruit no urgency, just checking if you are still planning to work on this PR.

moonfruit:
No, I don't work on this PR anymore since it cannot be accepted.
I'm working on another PR to make a non-enum `TlsVersion` to solve this in 5.0+.

PR #7765, opened 2023-04-16
Depend on okhttp-icu 0.2.0

swankjesse:


okhttp/build.gradle.kts
```diff
@@ -52,6 +52,7 @@ kotlin {
       dependencies {
         dependsOn(sourceSets.commonMain.get())
         implementation(libs.kotlinx.coroutines.core)
+        implementation(libs.squareup.okhttp.icu)
```
swankjesse, line 55: Deliberately only adding the dependency to nonJvmMain. No new transitive dependencies for JVM platform users!

yschimke: As discussed in slack, this is sensible. But maybe we should have a hook to allow overriding the system?

yschimke (approved):

okhttp/build.gradle.kts
yschimke: I guess I'm imagining someone implements our normalizer using https://unicode-org.github.io/icu/userguide/icu4j/, and decides to take the hit.

PR #7789, opened 2023-04-29
Implement a compact IDNA mapping table

swankjesse:
This is optimized for fast lookups and small code size. It also avoids using a resource file, as that's not an option on Kotlin/JS.

The data is encoded with 3 ASCII strings, as described in the new IdnaMappingTable class docs. I've got tests that confirm the compact data is consistent with the simple implementation for all code points.


okhttp/src/commonMain/kotlin/okhttp3/internal/idn/IdnaMappingTable.kt
```diff
@@ -0,0 +1,265 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okhttp3.internal.idn
+
+import okio.BufferedSink
+
+/**
+ * An IDNA mapping table optimized for small code and data size.
```
swankjesse, line 21: This doc is worth reading before reviewing the rest!

okhttp/src/jvmTest/java/okhttp3/internal/idn/IdnaMappingTableTest.kt
```diff
@@ -46,11 +58,49 @@ class IdnaMappingTableTest {
     assertThat("™".map()).isEqualTo("tm")
   }
 
+  /** Confirm the compact table satisfies is documented invariants. */
+  @Test fun validateCompactTableInvariants() {
```
swankjesse, line 62: TODO: confirm the output sizes are smaller than the ranges we use to index into them.

sections - at most 16,384 bytes (14 bits, which is what we binary search on.)
ranges - at most 65,536 bytes. (14 bits * 4, which is what we index on with a precision of 4 bytes.)
mappings - at most 16,384 chars (14 bits, which is what we index on.)

okhttp-idna-mapping-table/src/main/kotlin/okhttp3/internal/idn/GenerateIdnaMappingTableCode.kt
```diff
@@ -0,0 +1,98 @@
...
+
+import com.squareup.kotlinpoet.ClassName
+import com.squareup.kotlinpoet.FileSpec
+import com.squareup.kotlinpoet.KModifier
+import com.squareup.kotlinpoet.PropertySpec
+import java.io.File
+import okio.FileSystem
+import okio.Path.Companion.toPath
+
+fun main(vararg args: String) {
```
swankjesse, line 28: This generates a ~100 KiB .kt file. It’s overly long 'cause we use 6-byte escape sequences like `\u0000` instead of 1-byte literals.

okhttp/build.gradle.kts
```diff
@@ -80,6 +80,7 @@ kotlin {
       }
     }
     getByName("jvmTest") {
+      kotlin.srcDir("$buildDir/generated/sources/idnaMappingTable")
```
swankjesse, line 83: jvmTest for now, but should soon be in the nonJvmMain production source set.

okhttp/src/commonMain/kotlin/okhttp3/internal/idn/IdnaMappingTable.kt
```diff
@@ -0,0 +1,265 @@
...
+ */
+internal class IdnaMappingTable internal constructor(
+  val sections: String,
+  val ranges: String,
+  val mappings: String,
+) {
+  /**
+   * Returns true if the [codePoint] was applied successfully. Returns false if it was disallowed.
+   */
+  fun map(codePoint: Int, sink: BufferedSink): Boolean {
```
yschimke, line 114: Should this be on an interface?

yschimke: I guess I'm wondering if we need to load this file in most apps? What does loading a 100KiB class at runtime on android look like?

Would a fast ascii implementation that is used until we hit a non ascii domain be worthwhile? Or I'm overthinking it?

yschimke (approved):
LGTM - but it's a tough review, I think the tests including existing ones should be the proof we need.

okhttp-idna-mapping-table/src/main/kotlin/okhttp3/internal/idn/MappedRange.kt
```diff
@@ -0,0 +1,79 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okhttp3.internal.idn
+
+import okio.ByteString
+
+internal sealed interface MappedRange {
```
yschimke, line 20: Looking at the data, there does appear to be a common pattern of contiguous monotonically increasing ranges. Would we gain by compressing these down to an initial value plus offset?

```
0041          ; mapped                 ; 0061          # 1.1  LATIN CAPITAL LETTER A
...
005A          ; mapped                 ; 007A          # 1.1  LATIN CAPITAL LETTER Z
```

Also

```
1D777         ; mapped                 ; 03B8          # 3.1  MATHEMATICAL SANS-SERIF BOLD SMALL THETA
...
1D77B         ; mapped                 ; 03BC          # 3.1  MATHEMATICAL SANS-SERIF BOLD SMALL MU
```

okhttp/src/commonMain/kotlin/okhttp3/internal/idn/IdnaMappingTable.kt
swankjesse: A fast ASCII one is a good idea!

okhttp-idna-mapping-table/src/main/kotlin/okhttp3/internal/idn/MappedRange.kt
swankjesse: Yes! I’ve got mergeAdjacentRanges() for this. It also handles merging ranges that are different in the source table but equivalent for our use.

yschimke: The generated source seem to have A, B, C etc. Is that working?

yschimke: image

okhttp/src/jvmTest/java/okhttp3/internal/idn/IdnaMappingTableTest.kt
swankjesse: done.

okhttp-idna-mapping-table/src/main/kotlin/okhttp3/internal/idn/MappedRange.kt
swankjesse: Ahhhh, I misunderstood your suggestion.

I tried it, and the ranges string shrinks from 32,612 bytes to 30,068 bytes once I limit the mappings to only those where the delta fits in 14 bits.

But by increasing the maximum delta to 18 bits, I can shrink it much more, to 24,756 bytes. To do this, I need to borrow some bits from `b1`, and conveniently there’s some available. I can use the range from 64..118 that was unassigned.

```diff
 0..63 : Length of the UTF-16 sequence that this range maps to. The offset is b2b3.
+ 64..79 : Offset by a fixed negative offset. The bottom 4 bits of the offset are the top 4 bits of the offset.
+ 80..95 : Offset by a fixed positive offset. The bottom 4 bits of the offset are the top 4 bits of the offset.
   119 : Ignored.
   120 : Valid.
   121 : Disallowed
   122 : Mapped inline to the sequence: [b2].
   123 : Mapped inline to the sequence: [b2a].
   124 : Mapped inline to the sequence: [b2, b3].
   125 : Mapped inline to the sequence: [b2a, b3].
   126 : Mapped inline to the sequence: [b2, b3a].
   127 : Mapped inline to the sequence: [b2a, b3a].
```

I’ll hold this for a follow-up.

PR #7797, opened 2023-05-04
Make okhttp-testing-support a multiplatform project

swankjesse:


yschimke (approved):
Looks great, but seems like it's all still jvmMain, which bits do you need?

swankjesse:
I’m not quite sure what I’ll need! Just wanted to avoid mixing renames with edits to keep reviews and history simple.

PR #7798, opened 2023-05-04
Share more hostnames manipulation code

swankjesse:


okhttp/src/nonJvmMain/kotlin/okhttp3/internal/-HostnamesNonJvm.kt
```diff
@@ -17,44 +17,7 @@ package okhttp3.internal
...
-    } else {
-      result
-    }
-  } catch (_: IllegalArgumentException) {
-    return null
-  }
-}
-
-internal fun inet4AddressToAscii(address: ByteArray): String {
-  return address.joinToString(".")
```
swankjesse, line 54: I think this might have printed `-1.-1.-1.-1` instead of `255.255.255.255` because we’re using signed bytes. Sigh.

okhttp/src/commonMain/kotlin/okhttp3/internal/-HostnamesCommon.kt
```diff
@@ -225,4 +225,89 @@ internal fun inet6AddressToAscii(address: ByteArray): String {
...
+ */
+internal fun canonicalizeInetAddress(address: ByteArray): ByteArray {
+  return when {
+    isMappedIpv4Address(address) -> address.sliceArray(12 until 16)
+    else -> address
+  }
+}
+
+/** Returns true for IPv6 addresses like `0000:0000:0000:0000:0000:ffff:XXXX:XXXX`. */
+private fun isMappedIpv4Address(address: ByteArray): Boolean {
```
swankjesse, line 242: New code! So I don’t use `InetAddress.getByAddress()` to perform IPv6 to IPv4 mapping.

okhttp/src/nonJvmMain/kotlin/okhttp3/internal/-HostnamesNonJvm.kt
yschimke: In my defense, it doesn't look like it's even called anywhere :)

yschimke (approved):
Shit is getting real

okhttp/src/nonJvmMain/kotlin/okhttp3/internal/-HostnamesNonJvm.kt
swankjesse: All good. Getting the tests moved over is an important step!

PR #7800, opened 2023-05-05
Do IDNA mapping on Kotlin/JS

swankjesse:


okhttp/src/nonJvmMain/kotlin/okhttp3/internal/-HostnamesNonJvm.kt
```diff
@@ -16,8 +16,27 @@
...
+      "disallowed code point: $codePoint"
+    }
+  }
+
+  // 2. Normalize, from bufferB to bufferA.
+  val normalized = SYSTEM_NORMALIZER.normalizeNfc(bufferB.readUtf8())
+  bufferA.writeUtf8(normalized)
+
+  // 3. For each label, convert/validate Punycode.
+  // TODO.
```
swankjesse, line 39: So we’ve got a Punycode _decoder_ but not an _encoder_. Gah. One more step.

okhttp-testing-support/src/commonMain/kotlin/okhttp3/TestUtilCommon.kt
```diff
@@ -0,0 +1,29 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okhttp3
+
+import okio.FileSystem
+import okio.Path
+import okio.Path.Companion.toPath
+
+val okHttpRoot: Path
```
yschimke, line 22: I read your blog on this, it's a shame that this isn't something you can push down to okio testing.

Is there any small hook in gradle we can use to change this? Gradle.propertiea to define that affect launch options?

okhttp/src/nonJvmMain/kotlin/okhttp3/internal/-HostnamesNonJvm.kt
```diff
@@ -16,8 +16,27 @@
...
 
 internal actual fun idnToAscii(host: String): String {
-  // TODO implement properly
-  return SYSTEM_NORMALIZER.normalizeNfc(host).lowercase()
+  val bufferA = Buffer().writeUtf8(host)
+  val bufferB = Buffer()
+
+  // 1. Map, from bufferA to bufferB.
+  while (!bufferA.exhausted()) {
+    val codePoint = bufferA.readUtf8CodePoint()
```
yschimke, line 28: Can this fail by exhausting the string on bad input, is that caught in the common logic calling this?

yschimke (approved):

okhttp/src/nonJvmMain/kotlin/okhttp3/internal/-HostnamesNonJvm.kt
swankjesse: Good question!

This is all a hack to get around the fact that `kotlin.String` can’t iterate code points in multiplatform. (What a fail!)

This could throw an `EOFException` if the UTF-8 is malformed, but it won’t be 'cause we just encoded it above.

So really it’ll make forward progress thru the very last byte.

okhttp-testing-support/src/commonMain/kotlin/okhttp3/TestUtilCommon.kt
yschimke: I guess I'd be scanning the "classpath", whatever that means, or other env variables to any way.  From $rootModule, can you find your own test :)

swankjesse: I couldn’t find anything in the iOS simulator that traces back to the repo directory.

PR #7804, opened 2023-05-07
[in progress] Add a Flow control strategy

yschimke:


yschimke:
@swankjesse probably far enough along that your input would be welcome.

yschimke:
Next up, create some tests for the strategies.

Also, debug why this change is working for TwoRequestTest even with the existing strategy? I am changing existing behaviour somehow...

yschimke:
Will simplify all the connection methods to one with a param

PR #7805, opened 2023-05-07
Adopt Punycode in HttpUrl on non-JVM platforms

swankjesse:


okhttp/src/commonMain/kotlin/okhttp3/internal/-HostnamesCommon.kt
```diff
@@ -292,22 +292,18 @@ internal fun String.toCanonicalHost(): String? {
...
-      null
-    } else {
-      result
-    }
-  } catch (_: IllegalArgumentException) {
-    return null
+  val result = idnToAscii(host) ?: return null
+  if (result.isEmpty()) return null
+
+  return if (result.containsInvalidHostnameAsciiCodes()) {
```
yschimke, line 298: Are these two check needed?

Or should be postconditions via check?

okhttp/src/commonTest/kotlin/okhttp3/HttpUrlTest.kt
```diff
@@ -471,14 +471,12 @@ open class HttpUrlTest {
 
   @Test
   fun hostnameLowercaseCharactersMappedDirectly() {
-    if (!isJvm) return // TODO: this test is broken on non-JVM platforms.
```
yschimke, line 474: Nice!

yschimke (approved):
LGTM

Take it with a grain of salt, few beers in watching arsenal game with my son.

okhttp/src/commonMain/kotlin/okhttp3/internal/-HostnamesCommon.kt
swankjesse: Yeah, they’re needed. Java’s IDN ToAscii returns non-ascii characters if it doesn’t have a mapping!

PR #7808, opened 2023-05-08
Confirm labels are NFC after punycode decoding

swankjesse:


okhttp/src/commonTest/kotlin/okhttp3/internal/idn/PunycodeTest.kt
```diff
@@ -109,28 +109,28 @@ class PunycodeTest {
   }
 
   @Test fun nonBasicCodePointInPrefix() {
-    assertEquals("xn--cåt-n3h", Punycode.decode("xn--cåt-n3h"))
+    assertNull(Punycode.decode("xn--cåt-n3h"))
```
swankjesse, line 112: Behavior changed! If it isn‘t valid punycode this function now returns null.

swankjesse: (Previously it returned the input.)

okhttp/src/commonTest/kotlin/okhttp3/HttpUrlTest.kt
```diff
@@ -1838,4 +1839,23 @@ open class HttpUrlTest {
...
+    // café can be NFC (é is one code point) or NFD (e plus ´ as two code points).
+    val hostNfc = "café.com"
+    val hostNfcPunycode = "xn--caf-dma.com"
+    val hostNfd = "café.com"
+    val hostNfdPunycode = "xn--cafe-yvc.com"
+    assertEquals(hostNfcPunycode, "http://$hostNfc/".toHttpUrl().host)
+    assertEquals(hostNfcPunycode, "http://$hostNfcPunycode/".toHttpUrl().host)
+    assertEquals(hostNfcPunycode, "http://$hostNfd/".toHttpUrl().host)
+    if (isJvm) return // TODO: the rest of this test is broken on JVM platforms.
+    assertInvalid("http://$hostNfdPunycode/", """Invalid URL host: "$hostNfdPunycode"""")
```
swankjesse, line 1859: java.net.IDN won’t decode hostNfdPunycode from punycode (presumably 'cause it’s NFD), but it will decode hostNfcPunycode. So the behavior is kind of there.

But java.net.URL handles both.

yschimke (approved):

PR #7810, opened 2023-05-08
Add flow control observability

yschimke:
Add an internal API for receiving HTTP/2 flow control information, definitely not complete.

This shows

Connection - unacked on connection - this is not true in the first, as connection doesn't bump until consumed via a stream
3 - The unacked on stream 3 (the stalled request)
3-buffer - The data sitting inside the Buffer to be read
5 & 5-buffer - same as 3.

image

As an example used before and after this fix https://github.com/square/okhttp/pull/7801/files

image

### DIFF
--- a/okhttp/src/jvmMain/kotlin/okhttp3/internal/http2/Http2Stream.kt
+++ b/okhttp/src/jvmMain/kotlin/okhttp3/internal/http2/Http2Stream.kt
@@ -382,7 +382,7 @@ class Http2Stream internal constructor(
             } else if (readBuffer.size > 0L) {
               // Prepare to read bytes. Start by moving them to the caller's buffer.
               readBytesDelivered = readBuffer.read(sink, minOf(byteCount, readBuffer.size))
-              readBytes.increase(total = readBytesDelivered)
+              readBytes.update(total = readBytesDelivered)
 
               val unacknowledgedBytesRead = readBytes.unacknowledged
               if (errorExceptionToDeliver == null &&
@@ -390,9 +390,8 @@ class Http2Stream internal constructor(
                 // Flow control: notify the peer that we're ready for more data! Only send a
                 // WINDOW_UPDATE if the stream isn't in error.
                 connection.writeWindowUpdateLater(id, unacknowledgedBytesRead)
-                readBytes.increase(acknowledged = unacknowledgedBytesRead)
+                readBytes.update(acknowledged = unacknowledgedBytesRead)
               }
-              connection.flowControlListener.receivingStreamWindowChanged(id, readBytes, readBuffer.size)
             } else if (!finished && errorExceptionToDeliver == null) {
               // Nothing to do. Wait until that changes then try again.
               waitForIo()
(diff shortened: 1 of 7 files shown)


yschimke:
Janky google sheet https://docs.google.com/spreadsheets/d/1Xk0pBLJlYgY9XNcpGFDlZkme0kaYVydBUAHn2BeCrvs/edit?usp=sharing

okhttp/src/jvmMain/kotlin/okhttp3/internal/http2/flowcontrol/WindowCounter.kt
```diff
@@ -0,0 +1,37 @@
+package okhttp3.internal.http2.flowcontrol
+
+import java.util.concurrent.atomic.AtomicLong
+
+class WindowCounter(
+  val streamId: Int
+) {
+  /** The total number of bytes consumed. */
+  private val _total: AtomicLong = AtomicLong(0)
```
swankjesse, line 9: Why atomics?

yschimke: I don't think I'm clear on all the locking that goes on in all cases. I originally had a data class that was updated. Might switch back to that and define locking rules around writes.

okhttp/src/jvmMain/kotlin/okhttp3/internal/http2/Http2Stream.kt
```diff
@@ -386,16 +382,17 @@ class Http2Stream internal constructor(
...
+              val unacknowledgedBytesRead = readBytes.unacknowledged
               if (errorExceptionToDeliver == null &&
                   unacknowledgedBytesRead >= connection.okHttpSettings.initialWindowSize / 2) {
                 // Flow control: notify the peer that we're ready for more data! Only send a
                 // WINDOW_UPDATE if the stream isn't in error.
                 connection.writeWindowUpdateLater(id, unacknowledgedBytesRead)
-                readBytesAcknowledged = readBytesTotal
+                readBytes.increase(acknowledged = unacknowledgedBytesRead)
               }
+              connection.flowControlListener.receivingStreamWindowChanged(id, readBytes, readBuffer.size)
```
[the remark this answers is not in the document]
yschimke: Yep. I am becoming more convinced for libraries like exoplayer we should make the flow control visible in some form.  But happy to agree to remove this before 5.0 unless we find that nice API.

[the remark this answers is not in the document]
yschimke: I think this can move, out of synchronized, but inside the loop.

okhttp/src/jvmMain/kotlin/okhttp3/internal/http2/flowcontrol/WindowCounter.kt
yschimke: Removed, no need.

PR #7811, opened 2023-05-09
Do our own IDN logic on the JVM

swankjesse:
This adds a dependency on the ASCII-data IDN mapping table originally written for Kotlin/JS and multiplatform.


okhttp/src/commonMain/kotlin/okhttp3/internal/idn/IdnaMappingTable.kt
```diff
@@ -249,17 +249,5 @@ inline fun binarySearch(position: Int, limit: Int, compare: (Int) -> Int): Int {
...
-      val compareResult = compare(low)
-      when {
-        compareResult < 0 -> -low - 1 // insertionPoint is before this element.
-        compareResult > 0 -> -low - 2 // insertionPoint is after this element.
-        else -> return low // Match.
-      }
-    }
-    else -> -low - 1 // insertionPoint is before the first element.
-  }
+  return -low - 1 // insertionPoint is before the first element.
```
swankjesse, line 252: Much simpler binary search by using inclusive low and inclusive high

okhttp/src/commonTest/kotlin/okhttp3/WebPlatformToAsciiTest.kt
```diff
@@ -48,28 +49,6 @@ class WebPlatformToAsciiTest {
...
-    "xn--a.ß",
-    "xn--zca.ß",
-    "ශ්‍රී",
-
-    // OkHttp does not reject invalid Punycode.
-    "xn--",
-    "xn--ls8h=",
-
-    // OkHttp returns `xn--mgba3gch31f`, not `xn--mgba3gch31f060k`.
-    "نامه‌ای",
```
swankjesse, line 67: Yay!

okhttp/src/commonTest/kotlin/okhttp3/WebPlatformToAsciiTest.kt
```diff
@@ -48,28 +49,6 @@ class WebPlatformToAsciiTest {
 
     // OkHttp doesn't reject a U+200D. https://www.rfc-editor.org/rfc/rfc5892.html#appendix-A.2
     "xn--1ug.example",
-  )
-
-  val knownFailuresJvm = commonKnownFailures + setOf(
-    // OkHttp incorrectly does transitional processing, so it maps 'ß' to 'ss'
```
swankjesse, line 54: Yay! This is probably the bugfix with the most user impact

okhttp/src/nonJvmMain/kotlin/okhttp3/internal/-NormalizeNonJvm.kt
```diff
@@ -13,5 +13,9 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 package okhttp3.internal
+
+import com.squareup.okhttpicu.SYSTEM_NORMALIZER
+
+internal actual fun normalizeNfc(string: String): String =
```
swankjesse, line 20: I avoided depending on okhttp-icu from the main OkHttp module. The JVM gives us everything we need and I’m reluctant to add an entire jar dependency for something that’s a pass-through to this method.

okhttp/src/commonTest/kotlin/okhttp3/WebPlatformToAsciiTest.kt
```diff
@@ -40,6 +40,7 @@ class WebPlatformToAsciiTest {
 
     // OkHttp does not reject invalid Punycode.
     "xn--a",
+    "xn--a.ß",
```
yschimke, line 43: So this case is now broken on all, not just JVM?

yschimke: Ahhh, nope was defined in both

yschimke (approved):

okhttp/src/commonMain/kotlin/okhttp3/internal/-HostnamesCommon.kt
```diff
@@ -294,16 +296,40 @@ internal fun String.toCanonicalHost(): String? {
...
+
+  // 3. For each label, convert/validate Punycode.
+  val decoded = Punycode.decode(bufferA.readUtf8()) ?: return null
+
+  // 4.1 Validate.
+
+  // Must be NFC.
+  if (decoded != normalizeNfc(decoded)) return null
+
+  // TODO: Must not begin with a combining mark.
```
yschimke, line 327: Will these TODOs correct the failing cases?

swankjesse: Yes, but I’m not 100% certain we want to do these TODOs. Emitting the correct ASCII domain name for an input Unicode domain name is important and worth the effort. But to enforce things like joiners we need to haul in even more unicode data.

Issue #7381, opened 2022-07-13
java.lang.IllegalStateException: Unbalanced enter/exit


adamparslew:
Apologies if this has already been addressed.  I have encountered the following problem:

- I attempt to create a websocket to a given URL but encountering a 503 during handshaking (rather than 101) resulting in a callback to `onFailure()`.
- Within the callback I attempt to create a new websocket to the same URL, which gives rise to a second call to `onFailure()`:
```
java.io.IOException: canceled due to java.lang.IllegalStateException: Unbalanced enter/exit
	at okhttp3.internal.connection.RealCall$AsyncCall.run(RealCall.kt:530)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
	at java.base/java.lang.Thread.run(Thread.java:833)
	Suppressed: java.lang.IllegalStateException: Unbalanced enter/exit
		at okio.AsyncTimeout$Companion.scheduleTimeout(AsyncTimeout.kt:225)
		at okio.AsyncTimeout$Companion.access$scheduleTimeout(AsyncTimeout.kt:201)
		at okio.AsyncTimeout.enter(AsyncTimeout.kt:54)
		at okio.AsyncTimeout$source$1.read(AsyncTimeout.kt:331)
		at okio.RealBufferedSource.indexOf(RealBufferedSource.kt:427)
		at okio.RealBufferedSource.readUtf8LineStrict(RealBufferedSource.kt:320)
		at okhttp3.internal.http1.HeadersReader.readLine(HeadersReader.kt:29)
		at okhttp3.internal.http1.Http1ExchangeCodec.readResponseHeaders(Http1ExchangeCodec.kt:178)
		at okhttp3.internal.connection.Exchange.readResponseHeaders(Exchange.kt:106)
		at okhttp3.internal.http.CallServerInterceptor.intercept(CallServerInterceptor.kt:79)
		at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
		at okhttp3.internal.connection.ConnectInterceptor.intercept(ConnectInterceptor.kt:34)
		at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
		at okhttp3.internal.cache.CacheInterceptor.intercept(CacheInterceptor.kt:95)
[...]

PR #7815, opened 2023-05-11
Fix websocket reconnect race condition

yschimke:
Attempting a fix for https://github.com/square/okhttp/issues/7381

The reason why it happens when fast reconnecting is that when we fail because of a 401, we release the Connection back to the ConnectionPool.  The new connection started from the onFailure callback is very likely to get the same connection, and so two threads are now operating on the same socket.  

```
    call!!.enqueue(object : Callback {
        val exchange = response.exchange
        val streams: Streams
        try {
          checkUpgradeSuccess(response, exchange)
          streams = exchange!!.newWebSocketStreams()
        } catch (e: IOException) {
          exchange?.webSocketUpgradeFailed() // releases the connection back to the ConnectionPool
          failWebSocket(e, response)
          response.closeQuietly() // Drains the Response data
          return
        }
```


yschimke:
@swankjesse requesting post review, would like to test with a published snapshot.

From slack "fix looks good"

yschimke:
Damn, I took a success as a pass but it was really a lucky run of a flaky test

swankjesse (commented):
LGTM

PR #7839, opened 2023-05-20
Android testing with GMD

yschimke:


.github/workflows/build.yml
```diff
@@ -409,35 +409,20 @@ jobs:
...
-        with:
-          api-level: ${{ matrix.api-level }}
-          force-avd-creation: false
-          emulator-options: -no-window -gpu swiftshader_indirect -noaudio -no-boot-anim -camera-back none
-          disable-animations: false
-          script: echo "Generated AVD snapshot for caching."
+      - name: Accept licenses
+        run: |
+          mkdir -p /Users/runner/Library/Android/sdk
+          tar xf .github/android_licenses.tar.gz -C /Users/runner/Library/Android/sdk
```
swankjesse, line 415: This song and dance works for me, and it’s also bizarre.

Who is accepting these licenses? Square, Inc.? Does accepting it both mechanically and publicly obligate us to anything?

swankjesse: I asked some internal people if we have Opinions on stuff like this.

yschimke: When you run various gradle plugins, it writes these files, it just doesn't know about previews

yschimke: I guess I should be using a step like the following with channel set.  Will test.

```
yes | "$ANDROID_HOME"/cmdline-tools/latest/bin/sdkmanager --licenses || true
```

yschimke:
I think we'll need fixes from GMD in 8.4? so maybe to wait for this.

Issue #5322, opened 2019-08-01
OkHttpClient.Builder overrides with kotlin.time.Duration


JakeWharton:
Once they're stable, or we could propagate their experimental nature with their experimental annotation.

yschimke:
+1, but I think we should generally wait until they are stable.  I assume we hide from java?  

I think API usage by a common library like OkHttp, will result in user pain and in general experimental features probably don't compose well through transient libraries. Once they are broadly adopted but experimental, it's going to be an awkward upgrade no matter what.

JakeWharton:
They're automatically hidden from Java since `Duration` is an `inline class`.

yschimke:
Is this useful as an extension? Or does it mainly make sense when internal representation uses this and the legacy API is the converter?

We could stick useful extensions in okhttp-coroutines ( make it modern kotlin).

JakeWharton:
An extension is okay, but putting it in a separate artifact doesn't really make sense.

yschimke:
Does it still work with apiVersion if we land https://github.com/square/okhttp/pull/7267

Or do you prefer shading it like okio like your suggestion in https://github.com/square/okio/pull/960

PR #7842, opened 2023-05-21
Add Kotlin Duration overrides

yschimke:
#fixes https://github.com/square/okhttp/issues/5322

### DIFF
--- a/okhttp/src/jvmMain/kotlin/okhttp3/OkHttpClient.kt
+++ b/okhttp/src/jvmMain/kotlin/okhttp3/OkHttpClient.kt
@@ -1015,9 +1015,8 @@ open class OkHttpClient internal constructor(
      *
      * The default value is 0 which imposes no timeout.
      */
-    @IgnoreJRERequirement
     fun callTimeout(duration: KotlinDuration) = apply {
-      callTimeout(duration.inWholeMilliseconds, MILLISECONDS)
+      callTimeout = checkDuration("duration", duration)
     }
 
     /**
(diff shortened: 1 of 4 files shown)


JakeWharton (approved):
Why is the annotation for animal sniffer needed?

yschimke:
Because I'm a lazy programmer, who copies and pastes.  will remove.

okhttp/src/jvmMain/kotlin/okhttp3/OkHttpClient.kt
```diff
@@ -1004,6 +1005,21 @@ open class OkHttpClient internal constructor(
       callTimeout(duration.toMillis(), MILLISECONDS)
     }
 
+    /**
+     * Sets the default timeout for complete calls. A value of 0 means no timeout, otherwise values
+     * must be between 1 and [Integer.MAX_VALUE] when converted to milliseconds.
```
swankjesse, line 1010: Ick, converting 499 microseconds to no timeout is not good. You’re asking for a very small timeout and we’re giving you none at all.

swankjesse: checkDuration gets it right. It rejects arguments smaller than 1ms.

https://github.com/square/okhttp/blob/ba3287b7486ec999d4a1960c8bd7eada85e0df3a/okhttp/src/jvmMain/kotlin/okhttp3/internal/-UtilJvm.kt#L100

PR #7843, opened 2023-05-21
Always https interceptor

yschimke:
Interceptor that enforces HTTPS, a week form of HSTS.

TODO tests and apply based on the Android NetworkSecurityPolicy.

Breaking up the bigger okhttp-android PR.


swankjesse (commented):
I’m unsure about this.

Promoting `http:` requests to `https:` in an interceptor means not fixing the insecure URLs that originate these requests. Would you consider one that fails the request if the scheme is wrong? Or is that redundant 'cause it’ll fail anyway?

I also don’t like how this interacts with redirects.

yschimke:
The use case is apps with external content. Switch to Https from links in a podcast etc.

Can close. Easy to do externally. Less useful if not something to default on.

PR #7844, opened 2023-05-21
Add a configurable AndroidLogging API

yschimke:
Breaking up the larger okhttp-android PR.

Allow android clients to use Android Log instead of Java logging.


okhttp-android/src/main/kotlin/okhttp3/android/AndroidLogging.kt
```diff
@@ -0,0 +1,16 @@
+package okhttp3.android
+
+import android.util.Log
+import okhttp3.logging.HttpLoggingInterceptor
+import okhttp3.logging.LoggingEventListener
+
+fun LoggingEventListener.Companion.AndroidLogging(
```
swankjesse, line 7: Why an extension function? What does it look like for Java language callers?

swankjesse: I’m also partial to Timber, but I don’t think it’s necessarily worth the trouble to offer a Timber integration.

yschimke: I kind of figure Kotlin is a safe choice on Android.  Can refactor. The point of the PR is to support logging with the feature set of Android. I don't mind the API.

The extension function makes it appear to be part of the logging library, not a secondary thing to discover.

okhttp-android/src/main/kotlin/okhttp3/android/AndroidLogging.kt
```diff
@@ -0,0 +1,16 @@
+package okhttp3.android
```
swankjesse, line 1: supernit: copyright header

swankjesse (approved):

yschimke:
@swankjesse added some tests and minimal docs.

PR #7845, opened 2023-05-21
[Blocked] Add tracing implementations of listeners

yschimke:
Be opinionated about the observability of OkHttp with the ubiquitous tracing libraries, which are already used extensively in standards like WorkManager.


yschimke:
@swankjesse seems an easier proposition now.  Thoughts?

yschimke:
Alternative would be an overridable id generator, and id typed as Any or some OkHttp Id?

yschimke:
@swankjesse in 2026, should we just default to tracing on in android libraries?

PR #7864, opened 2023-05-27
Fix doc publishing

yschimke:
Another attempt to fix https://github.com/square/okhttp/actions/runs/5099519647/jobs/9167263082


swankjesse (commented):
LGTM

PR #7872, opened 2023-05-29
[for discussion] Network listener

yschimke:
Starting with the NetworkObserver from Coil https://github.com/coil-kt/coil/blob/1bb21950e0e49afb4b9901c7dcc4a868e7d3055d/coil-base/src/main/java/coil/network/NetworkObserver.kt#L4

Things we should do
- noNewExchanges we we lose internet connectivity, or a particular network
- logging of the network conditions (internet, types)
- fast fail if no network

Things we could do

- Refresh connections - replace with new network
- request a network if possible


yschimke:
Closing for now, will reconsider and reopen later.

PR #7882, opened 2023-06-07
Add new inline mapping strategy to IdnaMappingTable

aerb:
This implements an optimization suggested for the IdnaMappingTable described in https://github.com/square/okhttp/pull/7789#discussion_r1181021018


yschimke (commented):
What's the impact on mapping file size?

aerb:
> What's the impact on mapping file size?

I got the exact same result as @swankjesse in this comment https://github.com/square/okhttp/pull/7789#discussion_r1181344918 

The mapping field is reduced to 24,756 bytes

aerb:
The file size of `IdnaMappingTableInstance.kt` went from  109744 to 68536 bytes.

Issue #7886, opened 2023-06-16
Suggest Make HandShakeCertificates using privateKey and certificates(server / client)


LeeKwonju:
When you request network response using mTLS,

you can use this way "curl --key {client private Key} --cert {client cert} --cacert {server cert} {target URL} "

And when we use WebView,  there is ClientCertRequest in android.webkit (using mTLS in webView)
using function `proceed(privateKey: PrivateKey, certificates: Array X509Certificate) `we can implement mTLS in WebView.
Same with WebView I think if we can implement mTLS with only private key and certifcates. it would be comfortable.

but When we try to implement mTLS using HandShakeCertificates with PrivateKey from Os, Certificates. It makes error 
at  `keyStore.setKeyEntry("private", heldCertificate.keyPair.private, password, chain)( 94 lines in tlsUtil in Okhttp-tls library)`
related issues : https://github.com/square/okhttp/issues/6463, https://github.com/square/okhttp/issues/5684, https://github.com/square/okhttp/issues/6116

But if we make X509keyManager, that only offer certificate Chain and private key, we can implement mTLS without error
Like your exmple code https://gist.github.com/yschimke/4a1eda31c584a64549c790d887f275e0

LeeKwonju:
Example PR is https://github.com/square/okhttp/pull/7887

PR #7887, opened 2023-06-16
Add HandshakeCertificates constructor from privateKey and certificates

LeeKwonju:
https://github.com/square/okhttp/issues/7886
example code about suggest

### DIFF
--- a/okhttp-tls/src/main/kotlin/okhttp3/tls/HandshakeCertificates.kt
+++ b/okhttp-tls/src/main/kotlin/okhttp3/tls/HandshakeCertificates.kt
@@ -16,22 +16,16 @@
 @file:Suppress("INVISIBLE_MEMBER", "INVISIBLE_REFERENCE")
 package okhttp3.tls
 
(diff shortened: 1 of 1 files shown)


yschimke:
OkHttp is very particular about our public API.  But also about security, what concerns me about this is us providing an implementation of X509KeyManager.  We are not cryptographers.

If we already expose the right hooks to make this possible for others to do, then I think that's the right API for us.

Your PR shows it is.  I'd suggest you publish this as a library on top of OkHttp, but it's not something we will put in the core library.  As it is, you still need to know to use this, presumably on BC Provider. There are other examples where we've decided not to implement the functionality, and @swankjesse has provided a gist instead, this feels like a similar one.

Are you blocked on this, does it need to be in the core library?

LeeKwonju:
There is no special reason it should be in the core library. it was just for comfy.

Arbitrarily Implementating of X509KeyManager could be dangerous.
so, I agree with you. i'll close it

Issue #7904, opened 2023-06-20
Calling `enqueue()` from OkHttp’s Dispatcher shouldn’t count against thread limits


swankjesse:
With coroutines it’s pretty common to use `Call.enqueue()` in the implementation of an `Authenticator` or `Interceptor`.

This ends badly because if the Dispatcher’s is at its configured limit there’s no thread to serve the enqueued call, and the overall network stack can deadlock!

We’ve documented this a bit, but it’s still a hazard. Especially when features like `suspend` on Wire gRPC and Retrofit make it very easy to use a background thread without even knowing it.

Rather than broadly warning everyone of the dangers of this antipattern, we should just make it so enqueued calls made on OkHttp’s Dispatcher thread don’t count against dispatcher limits. Those calls should be performed immediately. This will avoid deadlocks and remove a footgun from our API.

yschimke:
Is there a good pathological case, or common case such as Retrofit to test against?

I'll take a look at this on the weekend, mainly playing around.  Not licking the cookie, more interested to help get a POC up to jam on, and hopefully make it into the next release.

yschimke:
Current approach uses a Call.Factory attached to the interceptor.

Next approach I'll try is checking if the thread is owned by the current clients Dispatcher, and if so do the same.

PR #7908, opened 2023-06-24
Suspend support in interceptors

yschimke:
A quick riff on https://github.com/square/okhttp/issues/7904 to provoke discussion

Detection is by two mechanisms

1) Use a thread local set in AsyncCall.executeOn, if the interceptor strictly on the same thread, this works.  But not swapping dispatchers.
2) Use chain.callFactory which never actually enqueues.

We should look at what is possible on SuspendingInterceptor, if we say that this support relies on using it, and all frameworks (wire, retrofit) are suspending, then maybe we can carry some information  in the CoroutineContext to detect this, and pick it back up again in executeAsync?

### DIFF
--- a/okhttp-coroutines/src/jvmMain/kotlin/okhttp3/SuspendingInterceptor.kt
+++ b/okhttp-coroutines/src/jvmMain/kotlin/okhttp3/SuspendingInterceptor.kt
@@ -1,0 +1,11 @@
+package okhttp3;
+
+import kotlinx.coroutines.runBlocking
+
+abstract class SuspendingInterceptor: Interceptor {
+  override fun intercept(chain: Interceptor.Chain): Response = runBlocking {
+    interceptAsync(chain)
+  }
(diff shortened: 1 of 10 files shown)


okhttp-coroutines/src/jvmMain/kotlin/okhttp3/SuspendingInterceptor.kt
```diff
@@ -0,0 +1,11 @@
+package okhttp3;
+
+import kotlinx.coroutines.runBlocking
+
+fun interface SuspendingInterceptor: Interceptor {
+  override fun intercept(chain: Interceptor.Chain): Response = runBlocking {
+    interceptAsync(chain)
```
agustinn1233, line 7: withTimeoutOrNull podria aplicar?

yschimke:
Closing until I get the initiative for another attempt

Issue #7913, opened 2023-06-28
HTTP/2 RST_STREAM is sent twice after timeout for the same stream


danielkec:
okhttp client sends h2 RST_STREAM twice after timeout for the same stream. That is prohibited by the spec:

[RFC 9113 §5.1](https://httpwg.org/specs/rfc9113.html#StreamStates)
`A stream also enters the "closed" state after an endpoint either sends or receives a RST_STREAM frame.`
...
`An endpoint MUST NOT send frames other than PRIORITY on a closed stream.`

Reproducer:
https://github.com/danielkec/okrepro

![image](https://github.com/square/okhttp/assets/1773630/09c8b96f-4c31-4853-824e-eb6bf09c7422)

danielkec:
Probably a duplicate of #6171

yschimke:
The repro is really useful, thanks.  I'll take a look on the weekend.

yschimke:
3.14.1 is no longer a supported version.  You should use 4.11.0 if you want a supported version.  I'll verify your current test using 4.11.0.

MdEmon001:
<a href="https://viddownloader.online/">Online video downloader</a> is one of the best free video downloaders to help you download online video links on your PC, phone, or any device.
Try it now - https://viddownloader.online/

soulbird:
any updates on this?

yschimke:
closing as dupe of https://github.com/square/okhttp/issues/6171

PR #7921, opened 2023-07-01
Demonstrate reset duplicates

yschimke:
Test and eventually fix for https://github.com/square/okhttp/issues/7913


yschimke:
Appears to only fail in H2_PRIOR_KNOWLEDGE

image

yschimke:
Bad repro, since it's recording both sides....

PR #7918, opened 2023-07-01
AGP 8

yschimke:
Bump to AGP 8 - for general hygiene, assuming that we mostly use IJ 2023.2.

- EAP will resolve to a final soonish.
- Stable will just skip okhttp-android module, which is ok.
- CI will run android code

PR #7937, opened 2023-07-13
Update Gradle to 8.2.1

jobhh:
Gradle 8.2.1 was released on July 10th with some bugfixes: https://gradle.org/releases/
Small addition to #7918, where Gradle was already updated to 8.2.


yschimke:
Thanks for the PR, but we have automated version bumps, so I'd rather not start accepting PRs for this sort of update.

jobhh:
No problem. Thanks for the reply :)

Issue #7936, opened 2023-07-12
103 Early Hints leads to endless wait for headers


stanislaw89:
Steps to reproduce.

Execute the code snipper below multiple times (it reproduces in ~ 1/4 of runs) using `v4.11.0`
```
fun main() {
  val client = OkHttpClient().newBuilder()
    .connectTimeout(5000, TimeUnit.MILLISECONDS)
    .readTimeout(5000, TimeUnit.MILLISECONDS)
    .writeTimeout(5000, TimeUnit.MILLISECONDS)
    .callTimeout(10000, TimeUnit.MILLISECONDS)
    .build()

  val request = Request.Builder()
    .url("https://theornateoracle.com/cart.php?action=add&product_id=456")
    .addHeader(
      "User-Agent", "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_6) " +
      "AppleWebKit/537.36 (KHTML, like Gecko) Chrome/53.0.2785.143 Safari/537.36"
    )
    .cacheControl(CacheControl.FORCE_NETWORK)
    .get()
    .build()

  val call = client.newCall(request)
  println("Executing...")
  call.execute()
  println("Done")
}
```

The main thread hangs forever in [waitForIo of Http2Stream#takeHeaders](https://github.com/square/okhttp/blob/parent-4.11.0/okhttp/src/main/kotlin/okhttp3/internal/http2/Http2Stream.kt#L140).
I investigated a bit, and it looks like it only happens for `103 Early Hints` response codes (inside the [CallServerInterceptor#shouldIgnoreAndWaitForRealResponse](https://github.com/square/okhttp/blob/parent-4.11.0/okhttp/src/main/kotlin/okhttp3/internal/http/CallServerInterceptor.kt#L107) `if`). 
Also, I figured out that the [Http2Stream.StreamTimeout](https://github.com/square/okhttp/blob/parent-4.11.0/okhttp/src/main/kotlin/okhttp3/internal/http2/Http2Stream.kt#L669) is called after some time, but it does not unblock the main thread.

yschimke:
Yep - there is a race condition, and when it blocks we are treating them as trailers.  I'll try to find the right fix.

yschimke:
I could reproduce with 5.x, so if blocked use that until a release.

yschimke:
I haven't reproduced on 5.x, and that server now doesn't return a 103.  

I have some alternate servers, but I think this particular error relates to a 103 and then a response without a body.

[...]

PR #7938, opened 2023-07-15
[4.x] Don't hang on a 103 response

yschimke:
Race condition means that a 103 might be considered to complete a request, such that the the actual headers are considered trailers.

#fixes https://github.com/square/okhttp/issues/7936

### DIFF
--- a/okhttp/src/main/kotlin/okhttp3/internal/http2/Http2Stream.kt
+++ b/okhttp/src/main/kotlin/okhttp3/internal/http2/Http2Stream.kt
@@ -260,12 +260,12 @@ class Http2Stream internal constructor(
       if (this.errorCode != null) {
         return false
       }
-      if (source.finished && sink.finished) {
-        return false
-      }
       this.errorCode = errorCode
       this.errorException = errorException
       notifyAll()
+      if (source.finished && sink.finished) {
+        return false
+      }
     }
     connection.removeStream(id)
     return true
--- a/okhttp/src/main/kotlin/okhttp3/internal/http2/Http2Stream.kt
+++ b/okhttp/src/main/kotlin/okhttp3/internal/http2/Http2Stream.kt
@@ -18,6 +18,8 @@ package okhttp3.internal.http2
 import okhttp3.Headers
 import okhttp3.internal.EMPTY_HEADERS
 import okhttp3.internal.assertThreadDoesntHoldLock
+import okhttp3.internal.http.CallServerInterceptor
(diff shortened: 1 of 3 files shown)


okhttp/src/main/kotlin/okhttp3/internal/http2/Http2Stream.kt
```diff
@@ -283,7 +285,10 @@ class Http2Stream internal constructor(
     val open: Boolean
     synchronized(this) {
       if (!hasResponseHeaders || !inFinished) {
-        hasResponseHeaders = true
+        val status = headers[Header.RESPONSE_STATUS_UTF8]?.toIntOrNull()
```
swankjesse, line 288: Do you mind retargeting `master`, then cherry-picking this back to okhttp_4x ?

I’m a little surprised to see status codes handled at this layer

okhttp/src/test/java/okhttp3/InformationalResponseCodeTest.kt
```diff
@@ -30,19 +30,22 @@ class InformationalResponseCodeTest {
     recordFrames = true
   }
 
-  private var client = clientTestRule.newClient()
+  private var client = clientTestRule.newClientBuilder()
+    .followRedirects(false)
+    .build()
 
   @Test
   fun test103() {
```
swankjesse, line 38: Does master have this test?

yschimke:
I'll check and update this weekend.

okhttp/src/test/java/okhttp3/InformationalResponseCodeTest.kt
yschimke: No, this is a manual test, see line 24.  The problem is mockwebserver in 4.x doesn't have support for informational responses.  So tests on master are better.

okhttp/src/main/kotlin/okhttp3/internal/http2/Http2Stream.kt
yschimke: I think it went here so I could make the minimal change.

I can try again to apply some of the changes directly, it's been a while, so it's a bit vague why it broken down.

swankjesse: I’m still not a fan of handling status codes at the protocol layer. Revert this part and keep the fix for the infinite wait?

yschimke: This feels unfortunate.  I really would like to handle this properly in 4.12.

PR #7947, opened 2023-07-22
Bump 4.x dependency versions

yschimke:
Including okio 3.4.0.

4.x is the production branch, so should get general improvements, worth doing together in a 4.12.


yschimke:
@swankjesse requesting post review, happy to revert any bumps we think are unwise.

swankjesse (commented):
# LGTM

Thanks!

PR #7966, opened 2023-08-01
Log total time of request when buffering body

tyilo:
Currently only the time to receive the headers is logged, which is often a much smaller duration than the total duration of receiving the headers and the body.


yschimke:
Seems like a reasonable change. I'll wait for any other view points before landing.

We don't actively enhance these logging classes, there is an expectation that apps copy and paste reuse this code if you need further tweaks. So don't wait on us to land and release this.

yschimke (approved):

yschimke:
@swankjesse can follow up with any other feedback.

swankjesse (commented):
LGTM

PR #8022, opened 2023-09-14
Add SameSite support to Cookie

amygbui:
# Description
Add SameSite support to the Cookie class with a default value of "lax" when not explicitly set, which is the default browser behavior according to the MDN spec.

MDN Docs: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie

# Motivation
$Pay needs to set different SameSite policies on staging vs production. Right now we're appending `samesite=lax` to the stringified cookie, however built-in support would be much nicer.

### DIFF
--- a/okhttp/src/jvmMain/kotlin/okhttp3/Cookie.kt
+++ b/okhttp/src/jvmMain/kotlin/okhttp3/Cookie.kt
@@ -362,9 +362,7 @@ class Cookie private constructor(
     }
 
     fun sameSite(sameSite: String) = apply {
-      val validSameSites = listOf("lax", "strict", "none")
-      require(validSameSites.contains(sameSite.lowercase())) { "SameSite must be one of: ${validSameSites.joinToString(", ")}" }
-
+      require(sameSite.trim() == sameSite) { "sameSite is not trimmed" }
       this.sameSite = sameSite
     }
 
--- a/okhttp/src/jvmMain/kotlin/okhttp3/Cookie.kt
+++ b/okhttp/src/jvmMain/kotlin/okhttp3/Cookie.kt
@@ -96,18 +96,29 @@ class Cookie private constructor(
   @get:JvmName("hostOnly") val hostOnly: Boolean, // True unless 'domain' is present.
 
   /**
-   * Returns a non-empty string with the cookie's policy for whether the cookie is sent with
-   * cross-site requests, providing some protection against cross-site request forgery (CSRF) attacks.
-   * Same-site requests means requests originating from the same site that set the cookie. Cross-site
-   * means requests originating from site that's different from the one that set the cookie.
+   * Returns a string describing whether this cookie is sent for cross-site calls.
    *
-   * The possible values are:
-   * - `Strict`: Cookie will only be sent for same-site requests. The scheme and domain
-   *     must match the one on the cookie.
-   * - `Lax`: Cookie is not sent on cross-site requests. Cookie won't be sent when going from origin
-   *     site to an external site, but is sent when going from an external site to the origin site.
-   * - `None`: Browser will send the cookie with both cross-site and same-site requests. The Secure
-   *     attribute must be set when setting this value.
+   * Two URLs are on the same site if they share a [top private domain][HttpUrl.topPrivateDomain].
+   * Otherwise, they are cross-site URLs.
+   *
+   * When a URL is requested, it may be in the context of another URL.
+   *
+   *  * **Embedded resources like images and iframes** in browsers use the context as the page in
+   *    the address bar and the subject is the URL of an embedded resource.
+   *
+   *  * **Potentially-destructive navigations such as HTTP POST calls** use the context as the page
+   *    originating the navigation, and the subject is the page being navigated to.
+   *
+   * The values of this attribute determine whether this cookie is sent for cross-site calls:
+   *
+   *  - "Strict": the cookie is omitted when the subject URL is an embedded resource or a
+   *    potentially-destructive navigation.
+   *
+   *  - "Lax": the cookie is omitted when the subject URL is an embedded resource. It is sent for
+   *    potentially-destructive navigation. This is the default value.
+   *
+   *  - None: the cookie is always sent. The "Secure" attribute must also be set when setting this
+   *    value.
    */
   @get:JvmName("sameSite") val sameSite: String,
 ) {
--- a/okhttp/src/jvmMain/kotlin/okhttp3/Cookie.kt
+++ b/okhttp/src/jvmMain/kotlin/okhttp3/Cookie.kt
@@ -117,10 +117,10 @@ class Cookie private constructor(
    *  - "Lax": the cookie is omitted when the subject URL is an embedded resource. It is sent for
    *    potentially-destructive navigation. This is the default value.
    *
-   *  - None: the cookie is always sent. The "Secure" attribute must also be set when setting this
+   *  - "None": the cookie is always sent. The "Secure" attribute must also be set when setting this
    *    value.
    */
-  @get:JvmName("sameSite") val sameSite: String,
+  @get:JvmName("sameSite") val sameSite: String?,
 ) {
 
   /**
@@ -272,7 +272,9 @@ class Cookie private constructor(
         append("; httponly")
       }
 
-      append("; samesite=").append(sameSite)
+      if (sameSite != null) {
+        append("; samesite=").append(sameSite)
+      }
 
       return toString()
     }
@@ -294,7 +296,7 @@ class Cookie private constructor(
     private var httpOnly = false
     private var persistent = false
     private var hostOnly = false
-    private var sameSite = "Lax"
+    private var sameSite: String? = null
 
     internal constructor(cookie: Cookie) : this() {
       this.name = cookie.name
@@ -367,7 +369,7 @@ class Cookie private constructor(
     }
 
     fun build(): Cookie {
-      if (sameSite.lowercase() == "none") {
+      if (sameSite.equals("None", ignoreCase = true)) {
         require(secure) { "Secure attribute is required for this SameSite policy" }
       }
 
@@ -447,7 +449,7 @@ class Cookie private constructor(
       var httpOnly = false
       var hostOnly = true
       var persistent = false
-      var sameSite = "Lax"
+      var sameSite: String? = null
 
       var pos = cookiePairEnd + 1
       val limit = setCookie.length
--- a/okhttp/src/jvmMain/kotlin/okhttp3/Cookie.kt
+++ b/okhttp/src/jvmMain/kotlin/okhttp3/Cookie.kt
@@ -367,9 +367,6 @@ class Cookie private constructor(
     }
 
     fun build(): Cookie {
-      if (sameSite.equals("None", ignoreCase = true)) {
-        require(secure) { "Secure attribute is required for this SameSite policy" }
-      }
 
       return Cookie(
         name ?: throw NullPointerException("builder.name == null"),
--- a/okhttp/src/jvmMain/kotlin/okhttp3/Cookie.kt
+++ b/okhttp/src/jvmMain/kotlin/okhttp3/Cookie.kt
@@ -294,7 +294,7 @@ class Cookie private constructor(
     private var httpOnly = false
     private var persistent = false
     private var hostOnly = false
-    private var sameSite = "lax"
+    private var sameSite = "Lax"
 
     internal constructor(cookie: Cookie) : this() {
       this.name = cookie.name
@@ -359,18 +359,18 @@ class Cookie private constructor(
       this.httpOnly = true
     }
 
-    fun samesite(sameSite: String) = apply {
+    fun sameSite(sameSite: String) = apply {
       val validSameSites = listOf("lax", "strict", "none")
       require(validSameSites.contains(sameSite.lowercase())) { "SameSite must be one of: ${validSameSites.joinToString(", ")}" }
-
-      if (sameSite.lowercase() == "none") {
-        require(secure) { "Secure attribute is required for this SameSite policy" }
-      }
 
       this.sameSite = sameSite
     }
 
     fun build(): Cookie {
+      if (sameSite.lowercase() == "none") {
+        require(secure) { "Secure attribute is required for this SameSite policy" }
+      }
+
       return Cookie(
         name ?: throw NullPointerException("builder.name == null"),
         value ?: throw NullPointerException("builder.value == null"),
@@ -447,7 +447,7 @@ class Cookie private constructor(
       var httpOnly = false
       var hostOnly = true
       var persistent = false
-      var sameSite: String? = null
+      var sameSite = "Lax"
 
       var pos = cookiePairEnd + 1
       val limit = setCookie.length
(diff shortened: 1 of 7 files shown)


okhttp/src/jvmMain/kotlin/okhttp3/Cookie.kt
```diff
@@ -326,25 +349,38 @@ class Cookie private constructor(
       this.httpOnly = true
     }
 
+    fun samesite(sameSite: String) = apply {
+      val validSameSites = immutableListOf("lax", "strict", "none")
```
JakeWharton, line 353: Since this type isn't exposed anywhere, there's no need to add additional allocations to ensure it's immutable to external manipulation.
```suggestion
      val validSameSites = listOf("lax", "strict", "none")
```

JakeWharton (commented):
Mostly LGTM, but will leave actual approval to a more active maintainer

okhttp/src/jvmMain/kotlin/okhttp3/Cookie.kt
```diff
@@ -243,6 +261,8 @@ class Cookie private constructor(
         append("; httponly")
       }
 
+      append("; samesite=").append(sameSite)
```
swankjesse, line 264: Omit if `"None"`

okhttp/src/jvmMain/kotlin/okhttp3/Cookie.kt
```diff
@@ -263,6 +283,7 @@ class Cookie private constructor(
     private var httpOnly = false
     private var persistent = false
     private var hostOnly = false
+    private var sameSite = "lax"
```
swankjesse, line 286: This is documented as `Lax` with a capital `L` ??

okhttp/src/jvmMain/kotlin/okhttp3/Cookie.kt
```diff
@@ -93,7 +93,23 @@ class Cookie private constructor(
...
+   *
+   * The possible values are:
+   * - `Strict`: Cookie will only be sent for same-site requests. The scheme and domain
+   *     must match the one on the cookie.
+   * - `Lax`: Cookie is not sent on cross-site requests. Cookie won't be sent when going from origin
+   *     site to an external site, but is sent when going from an external site to the origin site.
+   * - `None`: Browser will send the cookie with both cross-site and same-site requests. The Secure
+   *     attribute must be set when setting this value.
+   */
+  @get:JvmName("sameSite") val sameSite: String,
```
yschimke, line 112: Is String the right type for this?  It may be is we want to handle even invalid values.  But figured worth asking.

okhttp/src/jvmMain/kotlin/okhttp3/Cookie.kt
yschimke: +1 - seems like we should prefer the typical case.

swankjesse: I believe the default should be `"None"`

okhttp/src/jvmMain/kotlin/okhttp3/Cookie.kt
```diff
@@ -93,7 +93,23 @@ class Cookie private constructor(
...
+   * Same-site requests means requests originating from the same site that set the cookie. Cross-site
+   * means requests originating from site that's different from the one that set the cookie.
+   *
+   * The possible values are:
+   * - `Strict`: Cookie will only be sent for same-site requests. The scheme and domain
+   *     must match the one on the cookie.
+   * - `Lax`: Cookie is not sent on cross-site requests. Cookie won't be sent when going from origin
+   *     site to an external site, but is sent when going from an external site to the origin site.
+   * - `None`: Browser will send the cookie with both cross-site and same-site requests. The Secure
+   *     attribute must be set when setting this value.
```
swankjesse, line 110:
```suggestion
   * Returns a string describing whether this cookie is sent for cross-site calls.
   *
   * Two URLs are on the same site if they share a [top private domain][HttpUrl.topPrivateDomain].
   * Otherwise, they are cross-site URLs.
   *
   * When a URL is requested, it may be in the context of another URL.
   *
   *  * **Embedded resources like images and iframes** in browsers use the context as the page in
   *    the address bar and the subject is the URL of an embedded resource.
   *
   *  * **Potentially-destructive navigations such as HTTP POST calls** use the context as the page
   *    originating the navigation, and the subject is the page being navigated to.
   *
   * The values of this attribute determine whether this cookie is sent for cross-site calls:
   *
   *  - "Strict": the cookie is omitted when the subject URL is an embedded resource or a
   *    potentially-destructive navigation.
   *
   *  - "Lax": the cookie is omitted when the subject URL is an embedded resource. It is sent for
   *    potentially-destructive navigation. This is the default value.
   *
   *  - None: the cookie is always sent. The "Secure" attribute must also be set when setting this
   *    value.
```

I wanna read the spec to confirm my understanding is correct. Also I wanna get the right terms; I don’t think 'Context' and 'Subject' are right, but there probably are words that are.

okhttp/src/jvmMain/kotlin/okhttp3/Cookie.kt
```diff
@@ -326,25 +348,38 @@ class Cookie private constructor(
       this.httpOnly = true
     }
 
+    fun samesite(sameSite: String) = apply {
```
swankjesse, line 351: `sameSite`

okhttp/src/jvmMain/kotlin/okhttp3/Cookie.kt
```diff
@@ -326,25 +348,38 @@ class Cookie private constructor(
...
+      val validSameSites = listOf("lax", "strict", "none")
+      require(validSameSites.contains(sameSite.lowercase())) { "SameSite must be one of: ${validSameSites.joinToString(", ")}" }
```
swankjesse, line 353: We can’t be this strict here. New values may come before OkHttp is updated.

okhttp/src/jvmMain/kotlin/okhttp3/Cookie.kt
```diff
@@ -401,6 +436,7 @@ class Cookie private constructor(
       var httpOnly = false
       var hostOnly = true
       var persistent = false
+      var sameSite: String? = null
```
swankjesse, line 439: default to `"lax"` here so we don’t need to fix it later?

okhttp/src/jvmMain/kotlin/okhttp3/Cookie.kt
```diff
@@ -492,8 +531,18 @@ class Cookie private constructor(
         path = if (lastSlash != 0) encodedPath.substring(0, lastSlash) else "/"
       }
 
+      // If SameSite is absent or not a valid option, use the default value.
+      if (sameSite == null) {
+        sameSite = "lax"
+      }
+
+      // If SameSite is set to `None`, Secure must be true
```
swankjesse, line 539: I don’t think we should do this. The parsed `Cookie`’s job is to represent what was encoded only. If the user wants to break the rules, that’s not our business at this layer.

swankjesse (commented):
I think the test case needs updates. I’d like to read the spec a bit more before landing this.

We might also want to offer APIs to actually implement the behavior for this. Not exactly sure where though, because we don’t currently have a 'context' URL for outbound calls.

okhttp/src/jvmMain/kotlin/okhttp3/Cookie.kt
yschimke: Closing based on other comments.

okhttp/src/jvmMain/kotlin/okhttp3/Cookie.kt
amygbui: Updated, though I'm concerned the default may change in the future and we'll be providing an incorrect default.

okhttp/src/jvmMain/kotlin/okhttp3/Cookie.kt
amygbui: What is the recommendation? Don't validate what's passed in, or add a warning log that says the value passed in is not a recognized value? For parsing, I think it makes sense to parse whatever fields people set even if it breaks the rules. Do we apply the same philosophy for cookie construction?

okhttp/src/jvmMain/kotlin/okhttp3/Cookie.kt
amygbui: If someone intentionally set the value to `"Lax"`, it seems unexpected to omit it from the cookie string.

swankjesse: I think we should make `sameSite` nullable so we can distinguish between explicitly setting `Lax` vs. not

okhttp/src/jvmMain/kotlin/okhttp3/Cookie.kt
```diff
@@ -263,6 +294,7 @@ class Cookie private constructor(
     private var httpOnly = false
     private var persistent = false
     private var hostOnly = false
+    private var sameSite = "Lax"
```
swankjesse, line 297: ... and I think the default should be null, not `Lax`

okhttp/src/jvmMain/kotlin/okhttp3/Cookie.kt
```diff
@@ -326,25 +359,38 @@ class Cookie private constructor(
...
 
+    fun sameSite(sameSite: String) = apply {
+      val validSameSites = listOf("lax", "strict", "none")
+      require(validSameSites.contains(sameSite.lowercase())) { "SameSite must be one of: ${validSameSites.joinToString(", ")}" }
+
+      this.sameSite = sameSite
+    }
+
     fun build(): Cookie {
+      if (sameSite.lowercase() == "none") {
```
swankjesse, line 370:
```suggestion
      if (sameSite.equalsIgnoreCase("None")) {
```

This avoids an allocation

okhttp/src/jvmTest/java/okhttp3/CacheTest.java
```diff
@@ -2015,9 +2015,9 @@ private RecordedRequest assertClientSuppliedCondition(MockResponse seed, String
 
     HttpUrl url = server.url("/");
     assertThat(get(url).body().string()).isEqualTo("A");
-    cookieJar.assertResponseCookies("a=FIRST; path=/");
+    cookieJar.assertResponseCookies("a=FIRST; path=/; samesite=Lax");
```
swankjesse, line 2018: revert, let’s make `null` the default

okhttp/src/jvmTest/java/okhttp3/CookieTest.kt
swankjesse, line 1: Please add tests for parsing and building a cookie with various SameSite attributes, including absent, normal, and unexpected

okhttp/src/jvmMain/kotlin/okhttp3/Cookie.kt
swankjesse: Yeah my bad. Because we’re going to differentiate between absent and Lax I think we should make absent the default.

okhttp/src/jvmMain/kotlin/okhttp3/Cookie.kt
amygbui: Updated using Kotlin syntax: `String.equals(value, ignoreCase = true)`

okhttp/src/jvmTest/java/okhttp3/CookieTest.kt
```diff
@@ -546,6 +550,69 @@ class CookieTest {
...
+  }
+
+  @Test fun trimLeadingAndTrailingWhitespaceFromSameSite() {
+    assertThat(parse(url, "a=b; SameSite= ")!!.sameSite).isEqualTo("")
+    assertThat(parse(url, "a= b; SameSite= Lax")!!.sameSite).isEqualTo("Lax")
+    assertThat(parse(url, "a=b ; SameSite=Lax ;")!!.sameSite).isEqualTo("Lax")
+    assertThat(parse(url, "a=\r\t \nb\n; \rSameSite=\n \tLax")!!.sameSite).isEqualTo("Lax")
+  }
+
+  @Test fun builderSameSiteTrimmed() {
```
amygbui, line 570: I want to call out that passing an empty string in is a valid operation. Should we require a non-empty string? This issue also applies to the `name` and `value` fields.

okhttp/src/jvmMain/kotlin/okhttp3/Cookie.kt
```diff
@@ -326,25 +361,36 @@ class Cookie private constructor(
...
 
+    fun sameSite(sameSite: String) = apply {
+      require(sameSite.trim() == sameSite) { "sameSite is not trimmed" }
+      this.sameSite = sameSite
+    }
+
     fun build(): Cookie {
+      if (sameSite.equals("None", ignoreCase = true)) {
+        require(secure) { "Secure attribute is required for this SameSite policy" }
+      }
```
swankjesse, line 372:
```suggestion
```

I’m anxious about this 'cause I’d like this to work for all cookies:

```kotlin
val cookie: Cookie = ...
val copy = cookie.newBuilder().build()
```

In this case we’re strict on `build()` in a way that we can’t / shouldn’t be on `parse()`.

okhttp/src/jvmTest/java/okhttp3/CookieTest.kt
```diff
@@ -546,6 +550,69 @@ class CookieTest {
     assertThat(cookie.domain).isEqualTo("::1")
   }
 
+  @Test fun emptySameSite() {
```
swankjesse, line 553: this works for me.

okhttp/src/jvmTest/java/okhttp3/CookieTest.kt
swankjesse: No, I don’t think empty strings are special.

okhttp/src/jvmTest/java/okhttp3/CookieTest.kt
```diff
@@ -546,6 +550,69 @@ class CookieTest {
...
+    val cookie = Cookie.Builder()
+      .name("a")
+      .value("b")
+      .domain("example.com")
+      .sameSite(sameSite)
+      .build()
+    assertThat(cookie.sameSite).isEqualTo(sameSite)
+  }
+
+  @Test fun builderSameSiteNoneRequiresSecure() {
```
swankjesse, line 601:
```suggestion
    /** Note that we permit building a cookie that doesn’t follow the rules. */
    @Test fun builderSameSiteNoneDoesNotRequire() {
```

okhttp/src/jvmTest/java/okhttp3/CookieTest.kt
```diff
@@ -546,6 +550,69 @@ class CookieTest {
...
+    val cookieBuilder = Cookie.Builder()
+      .name("a")
+      .value("b")
+      .domain("example.com")
+      .sameSite("None")
+
+    // assertThrows<IllegalArgumentException> {
+    //   cookieBuilder.build()
+    // }
```
swankjesse, line 610:
```suggestion
```

okhttp/src/jvmTest/java/okhttp3/CookieTest.kt
```diff
@@ -546,6 +550,69 @@ class CookieTest {
...
+
+    val cookie = cookieBuilder.secure().build()
```
swankjesse, line 612:
```suggestion
    val cookie = cookieBuilder.build()
```

swankjesse (approved):
Nice tests! 🚀

PR #8025, opened 2023-09-16
Disable javadoc temporarily

yschimke:
Failing currently https://github.com/square/okhttp/actions/runs/6198721220/job/16829730445

```
    Reason: Task ':logging-interceptor:dokkaJavadocJar' uses this output of task ':mockwebserver:dokkaGfm' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed.
    
    Possible solutions:
      1. Declare task ':mockwebserver:dokkaGfm' as an input of ':logging-interceptor:dokkaJavadocJar'.
      2. Declare an explicit dependency on ':mockwebserver:dokkaG
      3. 
```

### DIFF
--- a/mockwebserver-deprecated/build.gradle.kts
+++ b/mockwebserver-deprecated/build.gradle.kts
@@ -25,5 +25,5 @@ dependencies {
 }
 
 mavenPublishing {
-  configure(KotlinJvm(javadocJar = JavadocJar.Dokka("dokkaGfm")))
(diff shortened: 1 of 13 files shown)


yschimke:
@swankjesse can revert if we have a fix, but not sure what that is.

swankjesse (commented):
LGTM

PR #8036, opened 2023-09-25
Add new method in Headers.kt to allow non ASCII chars in both header …

en-milie:
I'm working on a fuzzing tool called CATS: https://github.com/Endava/cats that use okhttp as underlying http transport library. Multiple HTTP headers fuzzers will send different non-ASCII characters in both header names and values. Other fuzzers are prefixing/suffixing header values with spaces. The current `addUnsafeNonAscii()` doesn't allow non-ASCII in header values and also does trimming of values. This PR introduces an additional method that allows non-ASCII in both header names and values and doesn't do any value trimming.

### DIFF
--- a/okhttp/src/main/kotlin/okhttp3/Headers.kt
+++ b/okhttp/src/main/kotlin/okhttp3/Headers.kt
@@ -263,20 +263,11 @@ class Headers private constructor(
     }
 
     /**
-     * Add a header with the specified name and value. Does validation of header names, allowing
-     * non-ASCII values.
+     * Add a header with the specified name and value. Does not validate headers names and values, allowing
+     * non-ASCII characters.
      */
     fun addUnsafeNonAscii(name: String, value: String) = apply {
-      checkName(name)
       addLenient(name, value)
-    }
-
-    /**
-     * Add a header with the specified name and value. Doesn't validate neither header names or values,
-     * allowing non-ASCII characters in both.
-     */
-    fun addUnsafeNonAsciiNameOrValue(name: String, value: String) = apply {
-      addLenientNoTrim(name, value)
     }
 
     /**
@@ -329,15 +320,6 @@ class Headers private constructor(
     internal fun addLenient(name: String, value: String) = apply {
       namesAndValues.add(name)
       namesAndValues.add(value.trim())
-    }
-
-    /**
-     * Add a field with the specified value without any validation or value trimming. Only appropriate for headers
-     * from the remote peer or cache.
-     */
-    private fun addLenientNoTrim(name: String, value: String) = apply {
-      namesAndValues.add(name)
-      namesAndValues.add(value)
     }
 
     fun removeAll(name: String) = apply {
--- a/okhttp/src/main/kotlin/okhttp3/Headers.kt
+++ b/okhttp/src/main/kotlin/okhttp3/Headers.kt
@@ -272,6 +272,14 @@ class Headers private constructor(
     }
 
     /**
+     * Add a header with the specified name and value. Doesn't validate neither header names or values,
+     * allowing non-ASCII characters in both.
+     */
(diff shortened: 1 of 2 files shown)


swankjesse:
I’m sorry to say that OkHttp is absolutely not designed to be used as a fuzzer at the HTTP layer, and you’re going to have a bad time if you want to use it like that. We want to send well-formed HTTP requests, and want to avoid APIs that create malformed HTTP requests, even if you’re doing so deliberately.

That said, it is totally valid to use OkHttp to fuzz at the application layer. For example, to test for SQL injection or with request bodies that use invalid JSON.

On this PR specifically – most servers split the HTTP layer (implemented by nginx, Jetty, etc.) from the application layer. Are you interested in fuzzing a specific layer?

swankjesse:
... I just read the linked GitHub project and it looks like you’re working at the application layer. That’s good news for both of us.

I don’t want to brick-wall this PR, but I do want to advise that we’re disinterested in following this to its logical conclusion. For example, OkHttp is not going to be very helpful if you wanna send say, malformed HTTP/2 frames.

okhttp/src/main/kotlin/okhttp3/Headers.kt
```diff
@@ -271,6 +271,14 @@ class Headers private constructor(
       addLenient(name, value)
     }
 
+    /**
+     * Add a header with the specified name and value. Doesn't validate neither header names or values,
+     * allowing non-ASCII characters in both.
+     */
+    fun addUnsafeNonAsciiNameOrValue(name: String, value: String) = apply {
```
swankjesse, line 278: Lets just change `addUnsafeNonAscii()` to not call `checkName()` and be done with it?

swankjesse: Do you absolutely need trimming? I think that if you wanna go down the path of fuzzing at the HTTP protocol layer you might be better doing that with a raw TCP socket instead of a high-level client like OkHttp.

en-milie: That might be the ultimate solution, using raw TCP socket connections. Currently though, with the existing fuzzers, OkHttp is good enough (with the changes from this PR). I'm currently using a fork, but it will be great to use the official version.
There are fuzzers which prefix/suffix header values with spaces, thus having a method that doesn't do trimming is important.

If `checkName()` is removed maybe we can have 2 methods `addUnsafeNonAscii()` that does the trimming and add another one `addUnsafeNonAsciiNoTrim()`?

swankjesse: I’d really like to keep trimming mandatory in OkHttp. I don’t like the consequences of permitting even more weird `Headers` instances.

en-milie: Ok, I can appreciate keeping things stricter. In this case, would it be ok then just to remove `checkName()`?

swankjesse: Yep! That’s a good compromise.

en-milie: New method removed and also removed call to `checkName()`

swankjesse (approved):

swankjesse:
Mind updating `HeadersJvmTest.kt` ?

en-milie:
Yes, apologies, missed that.

swankjesse (approved):

swankjesse (commented):
Arg, this is targeting OkHttp 4.x. We only accept new PRs against `master`, and cherry-pick those to `main`.

swankjesse:
(Sorry for not flagging that sooner.)

en-milie:
I will probably close this for now. I reached additional cases which will require OkHttp to relax some constraints, which is probably not compatible with how OkHttp was designed. I'll might raise a future PR.

PR #8042, opened 2023-10-04
Making max message size for Websockets configurable

lauhon:
Hi, in react-native, which uses okhttp for networking on android, there is a mismatch between the max message size that can be sent via websockets.

On iOS there is seemingly no limit to the size a message can have. On Android we have this hard cap to 16MiB

This PR will enable teams to override this default setting of 16MiB if they want to.

(Other libraries, such as [ws](https://github.com/websockets/ws/blob/master/doc/ws.md#class-websocketserver) also provide an API for defining the max message size. )

### DIFF
--- a/okhttp/src/jvmMain/kotlin/okhttp3/OkHttpClient.kt
+++ b/okhttp/src/jvmMain/kotlin/okhttp3/OkHttpClient.kt
@@ -547,7 +547,7 @@ open class OkHttpClient internal constructor(
     internal var minWebSocketMessageToCompress = RealWebSocket.DEFAULT_MINIMUM_DEFLATE_SIZE
     internal var routeDatabase: RouteDatabase? = null
     internal var taskRunner: TaskRunner? = null
-    internal var maxQueueSize = RealWebSocket.DEFAULT_MAX_QUEUE_SIZE
+    internal var maxQueueSize: Long = RealWebSocket.DEFAULT_MAX_QUEUE_SIZE
 
     internal constructor(okHttpClient: OkHttpClient) : this() {
       this.dispatcher = okHttpClient.dispatcher
--- a/okhttp/src/jvmMain/kotlin/okhttp3/OkHttpClient.kt
+++ b/okhttp/src/jvmMain/kotlin/okhttp3/OkHttpClient.kt
@@ -232,7 +232,8 @@ open class OkHttpClient internal constructor(
   val minWebSocketMessageToCompress: Long = builder.minWebSocketMessageToCompress
 
   /**
-   * Maximum outbound web socket message size (in bytes) that will be sent
+   * The maximum number of bytes to enqueue. Rather than enqueueing beyond this limit we tear down
+   * the web socket! It's possible that we're writing faster than the peer can read.
    */
   @get:JvmName("maxQueueSize")
   val maxQueueSize: Long = builder.maxQueueSize
--- a/okhttp/src/jvmMain/kotlin/okhttp3/OkHttpClient.kt
+++ b/okhttp/src/jvmMain/kotlin/okhttp3/OkHttpClient.kt
@@ -231,6 +231,12 @@ open class OkHttpClient internal constructor(
   @get:JvmName("minWebSocketMessageToCompress")
   val minWebSocketMessageToCompress: Long = builder.minWebSocketMessageToCompress
 
+  /**
+   * Maximum outbound web socket message size (in bytes) that will be sent
+   */
+  @get:JvmName("maxQueueSize")
+  val maxQueueSize: Long = builder.maxQueueSize
+
   internal val routeDatabase: RouteDatabase = builder.routeDatabase ?: RouteDatabase()
   internal val taskRunner: TaskRunner = builder.taskRunner ?: TaskRunner.INSTANCE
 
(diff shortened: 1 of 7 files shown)


yschimke:
Nice reference to `ws`

> maxPayload {Number} The maximum allowed message size in bytes. Defaults to 100 MiB (104857600 bytes).

okhttp/src/jvmMain/kotlin/okhttp3/internal/ws/RealWebSocket.kt
```diff
@@ -63,7 +63,8 @@ class RealWebSocket(
    */
   private var extensions: WebSocketExtensions?,
   /** If compression is negotiated, outbound messages of this size and larger will be compressed. */
-  private var minimumDeflateSize: Long
+  private var minimumDeflateSize: Long,
+  private var maxQueueSize: Long? = DEFAULT_MAX_QUEUE_SIZE
```
yschimke, line 67: Why nullable Long??

lauhon: Thanks for reviewing.

Mistake on my part. Im still coding at the moment.

lauhon:
Hey @yschimke, i pushed again and now the pipeline **should** go through.

yschimke (approved):

okhttp/src/jvmMain/kotlin/okhttp3/OkHttpClient.kt
```diff
@@ -231,6 +231,12 @@ open class OkHttpClient internal constructor(
   @get:JvmName("minWebSocketMessageToCompress")
   val minWebSocketMessageToCompress: Long = builder.minWebSocketMessageToCompress
 
+  /**
+   * Maximum outbound web socket message size (in bytes) that will be sent
```
yschimke, line 235: nit: It's not the limit for a single message, rather the queue.

Previous docs

``` 
     * The maximum number of bytes to enqueue. Rather than enqueueing beyond this limit we tear down
     * the web socket! It's possible that we're writing faster than the peer can read.
```

lauhon: I restored the previous documentation 👍

lauhon:
Hey @yschimke !

There seems to be a fail in the testopenjdk11 step. Locally the step is running successfully, could you help me out here?

Also it seems the PR should be approved by a second maintainer, do you have someone in mind we can ask to review this?

yschimke:
Yep, probably my fault. Sorry

swankjesse:
This is what I wrote on [react-native #39651](https://github.com/facebook/react-native/issues/39651#issuecomment-1748722058):

> I think this might end badly. OkHttp’s cap is intended to limit queueing on the web socket. If you need to send a large payload, you should do it on a regular HTTP post, and send control messages on the web socket?
>
> As is this will disrupt the latency of the web socket while the oversized message is being transmitted.

yschimke:
@swankjesse the prior art was convincing for me.  Given the default remains the same, isn't it a valid footgun to give developers to match iOS and Node ws?

okhttp/src/jvmMain/kotlin/okhttp3/internal/ws/RealWebSocket.kt
```diff
@@ -402,8 +403,8 @@ class RealWebSocket(
     if (failed || enqueuedClose) return false
 
     // If this frame overflows the buffer, reject it and close the web socket.
-    if (queueSize + data.size > MAX_QUEUE_SIZE) {
-      close(CLOSE_CLIENT_GOING_AWAY, null)
+    if (queueSize + data.size > maxQueueSize) {
+      close(CLOSE_CLIENT_GOING_AWAY, "Message to large to send safely, aborting...")
```
lauhon, line 407: Hey @swankjesse & @yschimke.

If I understand corretly this PR will never be merged like that correct?

In that case, I think it would make sense to at least keep this specific change, as it will improve developer experience if they run into the same problem as me.

lauhon: Im building an app that uses websockets to implement [a specific multi party encryption flow](https://github.com/unboundsecurity/blockchain-crypto-mpc#63-system-flow)

This requires peers to send messages back and forth which can be up to 25MB.
It's totally possible to send the large messages via HTTP requests and control the rest of the flow with websockets, but for a POC I just wanted to rely on websockets. This worked on iOS and failed on android because of the limit built into okhttp.

lauhon: (And because there is no error message as of now, it took me a while before I found out that the size of the message is the reason for the socket to close)

JakeWharton: > It's totally possible to send the large messages via HTTP requests and control the rest of the flow with websockets

While I understand going websocket-only for the proof-of-concept, I definitely would migrate to separate HTTP requests as soon as possible. Any message sent on a web socket blocks all others so they should be very small and very fast to process.

However, I agree we should do better reporting the limitation so it's clear what's happening.

swankjesse:
I’m making the call that we’re not going to make this configurable.

Keeping a hard limit on the web socket backlog gives us the best chance of limiting latency for enqueued messages. It also prevents us from fully-buffering very large messages in memory.

Will document the limitation here:
https://github.com/square/okhttp/issues/8123

PR #8053, opened 2023-10-15
[4.x] Fix bad merge

yschimke:


yschimke:
@swankjesse test only, requesting post review.

okhttp-testing-support/src/main/kotlin/okhttp3/internal/io/InMemoryFileSystem.kt
```diff
@@ -119,4 +119,7 @@ class InMemoryFileSystem : FileSystem, TestRule {
   }
 
   override fun toString() = "InMemoryFileSystem"
+  fun allPaths(): MutableSet<File> {
```
swankjesse, line 122: MutableSet –> Set

okhttp-testing-support/src/main/kotlin/okhttp3/internal/io/InMemoryFileSystem.kt
```diff
@@ -119,4 +119,7 @@ class InMemoryFileSystem : FileSystem, TestRule {
...
+    return files.keys
```
swankjesse, line 123: files.keys is a live view; `files.keys.toSet()` is a necessary defensive copy, if that’s what we want

swankjesse (commented):
LGTM

PR #8054, opened 2023-10-15
[4.x] Avoid hanging after error race condition

yschimke:
Cherry pick of https://github.com/square/okhttp/pull/7954

### DIFF
--- a/okhttp/src/main/kotlin/okhttp3/internal/http2/Http2Stream.kt
+++ b/okhttp/src/main/kotlin/okhttp3/internal/http2/Http2Stream.kt
@@ -258,12 +258,12 @@ class Http2Stream internal constructor(
       if (this.errorCode != null) {
         return false
       }
-      if (source.finished && sink.finished) {
-        return false
-      }
       this.errorCode = errorCode
       this.errorException = errorException
       notifyAll()
+      if (source.finished && sink.finished) {
+        return false
+      }
     }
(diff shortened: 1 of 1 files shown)


swankjesse (approved):

okhttp/src/main/kotlin/okhttp3/internal/http2/Http2Stream.kt
```diff
@@ -258,12 +258,12 @@ class Http2Stream internal constructor(
...
       }
-      if (source.finished && sink.finished) {
-        return false
-      }
       this.errorCode = errorCode
       this.errorException = errorException
       notifyAll()
+      if (source.finished && sink.finished) {
+        return false
+      }
```
swankjesse, line 266:
```suggestion
      notifyAll()
      if (source.finished && sink.finished) {
        return false
      }
      this.errorCode = errorCode
      this.errorException = errorException
```

This way we don’t assign an error code after the stream is finished. I don’t like that, cause that error code isn’t transmitted!

swankjesse: Actually, lets not let OkHttp 4.x and 5.x diverge needlessly here. I’ll merge this as-is and then consider doing a 2nd PR to OkHttp 5.x if I can come up with a test.

PR #8055, opened 2023-10-15
[4.x] Add test for 103 handling

yschimke:


mockwebserver/src/main/kotlin/okhttp3/mockwebserver/MockResponse.kt
```diff
@@ -347,6 +349,10 @@ class MockResponse : Cloneable {
 
   override fun toString() = status
 
+  fun addInformationalResponseInternal(informationalResponse: MockResponse): MockResponse = apply {
```
yschimke, line 352: This is not great, but seems better to have some limited support without the bigger change of backporting.

cc @swankjesse

okhttp/src/test/java/okhttp3/CallTest.java
```diff
@@ -2812,6 +2812,43 @@ public void cancelWhileRequestHeadersAreSent_HTTP_2() throws Exception {
...
+                    .addInformationalResponseInternal(
+                            new MockResponse().setResponseCode(103).addHeader("Link", "</style.css>; rel=preload; as=style")
+                    )
+    );
+    Request request = new Request.Builder()
+            .url(server.url("/"))
+            .post(RequestBody.create("abc", MediaType.get("text/plain")))
+            .build();
+    executeSynchronously(request)
+            .assertFailure(SocketTimeoutException.class);
```
yschimke, line 2829: Showing we don't handle this well.

cc @swankjesse

swankjesse (approved):

mockwebserver/src/main/kotlin/okhttp3/mockwebserver/MockWebServer.kt
```diff
@@ -1026,6 +1026,13 @@ class MockWebServer : ExternalResource(), Closeable {
         readBody = true
       }
 
+      peek.informationalResponses.forEach {
```
swankjesse, line 1029: FWIW, I dislike `forEach` as an alternative to `for(...)`. It’s needlessly complex to step-debug!

okhttp/src/test/java/okhttp3/CallTest.java
swankjesse: 👍🏻

swankjesse (approved):

PR #8061, opened 2023-10-17
Make a defensive copy in InMemoryFileSystem

swankjesse:


okhttp-testing-support/src/main/kotlin/okhttp3/internal/io/InMemoryFileSystem.kt
```diff
@@ -35,6 +35,9 @@ class InMemoryFileSystem : FileSystem, TestRule {
   private val openSources = IdentityHashMap<Source, File>()
   private val openSinks = IdentityHashMap<Sink, File>()
 
+  val allPaths: Set<File>
```
yschimke, line 38: Good catch

yschimke (approved):

PR #8077, opened 2023-10-26
Fix: Staled cache response used when system time adjusted backward

luuillu:
OkHttp would incorrectly use a cached response for a second request if the server's first response did not contain caching headers, if the system time was adjusted backward between the two requests。

### DIFF
--- a/okhttp/src/jvmMain/kotlin/okhttp3/internal/cache/CacheStrategy.kt
+++ b/okhttp/src/jvmMain/kotlin/okhttp3/internal/cache/CacheStrategy.kt
@@ -272,9 +272,9 @@ class CacheStrategy internal constructor(
         apparentReceivedAge
       }
 
-      val responseDuration = receivedResponseMillis - sentRequestMillis
-      val residentDuration = nowMillis - receivedResponseMillis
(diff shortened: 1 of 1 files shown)


okhttp/src/jvmMain/kotlin/okhttp3/internal/cache/CacheStrategy.kt
```diff
@@ -274,7 +274,7 @@ class CacheStrategy internal constructor(
 
       val responseDuration = receivedResponseMillis - sentRequestMillis
       val residentDuration = nowMillis - receivedResponseMillis
-      return receivedAge + responseDuration + residentDuration
+      return maxOf(0, receivedAge + responseDuration + residentDuration)
```
swankjesse, line 277: I think it’s more straightforward to clamp each of the durations to be non-negative instead of doing that to the overall result.

luuillu: Thanks for your review, I have recommitted the file

swankjesse (approved):

PR #8105, opened 2023-11-14
Close the socket when an exception is thrown

luozhenyu:
If an exception is thrown when calling `connect`, the socket and its fd will leak.

https://github.com/square/okhttp/blob/30780c879bd0d28b49f264fac2fe05da85aef3ad/samples/unixdomainsockets/src/main/java/okhttp3/unixdomainsockets/UnixDomainSocketFactory.java#L39-L43


samples/unixdomainsockets/src/main/java/okhttp3/unixdomainsockets/UnixDomainSocketFactory.java
```diff
@@ -38,7 +38,12 @@ public UnixDomainSocketFactory(File path) {
 
   @Override public Socket createSocket(String host, int port) throws IOException {
     Socket result = createSocket();
-    result.connect(new InetSocketAddress(host, port));
+
+    try {
+      result.connect(new InetSocketAddress(host, port));
+    } catch (IOException e) {
+      result.close();
```
yschimke, line 45: This is swallowing the exception and continuing to use it.

yschimke:
From https://github.com/jnr/jnr-unixsocket/blob/master/src/main/java/jnr/unixsocket/UnixSocket.java

It looks correct that we should be closing. But I think you'll need to rethrow the exception.

samples/unixdomainsockets/src/main/java/okhttp3/unixdomainsockets/UnixDomainSocketFactory.java
luozhenyu: > This is swallowing the exception and continuing to use it.

You are right. I corrected it.

yschimke (commented):
Can you add a very basic test that tests it is closed after an exception.

yschimke:
Thanks for the fix. Are you hitting this?

yschimke (approved):

swankjesse (approved):
Thanks!

Issue #8116, opened 2023-11-26
Crash because of recursive access of Platform.get() due to logger on exception in initialization


werner77:
We see the following crash log in our crashlytics of our Android application:

```
Caused by java.lang.NullPointerException
Attempt to invoke virtual method 'java.lang.Class java.lang.Object.getClass()' on a null object reference

okhttp3.internal.platform.android.StandardAndroidSocketAdapter$Companion.buildIfSupported (StandardAndroidSocketAdapter.java:61)
okhttp3.internal.platform.android.StandardAndroidSocketAdapter$Companion.buildIfSupported$default (StandardAndroidSocketAdapter.java:52)
okhttp3.internal.platform.AndroidPlatform.<init> (AndroidPlatform.kt:47)
okhttp3.internal.platform.AndroidPlatform$Companion.buildIfSupported (AndroidPlatform.kt:160)
okhttp3.internal.platform.Platform$Companion.findAndroidPlatform (Platform.java:227)
okhttp3.internal.platform.Platform$Companion.findPlatform (Platform.java:220)
okhttp3.internal.platform.Platform$Companion.access$findPlatform (Platform.java:177)
okhttp3.internal.platform.Platform.<clinit> (Platform.kt:178)
okhttp3.OkHttpClient.<init> (OkHttpClient.kt:237)
okhttp3.OkHttpClient$Builder.build (OkHttpClient.kt:1069)
com.uxcam.internals.ch.a (ch.java:18)
com.uxcam.internals.af.a (af.java:2)
java.lang.Thread.run (Thread.java:923)
```

The Platform constructor invoked via Platform.get() calls StandardAndroidSocketAdapter$Companion.buildIfSupported which has in it's try/catch statement a call to Platform.get() causing a recursive instantiation.

werner77:
This is version 4.12.0

yschimke:
Do you know why it's failing?

vladvlasov256:
In our case, it fails specifically on HTC D816w with Android 7.1:

```
Caused by java.lang.NullPointerException
Attempt to invoke virtual method 'void ik.j.k(java.lang.String, int, java.lang.Throwable)' on a null object reference

okhttp3.internal.platform.android.StandardAndroidSocketAdapter$Companion.buildIfSupported (StandardAndroidSocketAdapter.java:1)
okhttp3.internal.platform.android.StandardAndroidSocketAdapter$Companion.buildIfSupported$default (StandardAndroidSocketAdapter.java:1)
[...]

PR #8117, opened 2023-11-26
Fix Android logging call

yschimke:
From https://github.com/square/okhttp/issues/8116

Fix on master, before backporting.

### DIFF
--- a/okhttp/src/jvmMain/kotlin/okhttp3/internal/platform/android/AndroidLog.kt
+++ b/okhttp/src/jvmMain/kotlin/okhttp3/internal/platform/android/AndroidLog.kt
@@ -66,7 +66,7 @@ object AndroidLog {
     this["okhttp3.mockwebserver.MockWebServer"] = "okhttp.MockWebServer"
   }.toMap()
 
-  internal fun androidLog(loggerName: String, logLevel: Int, message: String, t: Throwable?) {
+  internal fun androidLog(loggerName: String, logLevel: Int, message: String, t: Throwable? = null) {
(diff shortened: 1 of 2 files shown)


JakeWharton (approved):

yschimke:
Test manually with a similar call in Platform

```
    private fun findAndroidPlatform(): Platform {
      AndroidLog.enable()
      AndroidLog.androidLog(loggerName = OkHttpClient::class.java.name, logLevel = Platform.WARN, message = "XXXX")
      return Android10Platform.buildIfSupported() ?: AndroidPlatform.buildIfSupported()!!
    }
```

okhttp/src/jvmMain/kotlin/okhttp3/internal/platform/android/AndroidLog.kt
```diff
@@ -66,7 +66,7 @@ object AndroidLog {
     this["okhttp3.mockwebserver.MockWebServer"] = "okhttp.MockWebServer"
   }.toMap()
 
-  internal fun androidLog(loggerName: String, logLevel: Int, message: String, t: Throwable?) {
+  internal fun androidLog(loggerName: String, logLevel: Int, message: String, t: Throwable? = null) {
```
swankjesse, line 69: The default value is always specified?

swankjesse: Yep.

Nobler:
Hi, is the fix released or not?

Issue #6719, opened 2021-06-18
java.lang.NullPointerException: Deflater has been closed


Layneboy:
java.lang.NullPointerException: Deflater has been closed
	at java.util.zip.Deflater.ensureOpen(Deflater.java:551)
	at java.util.zip.Deflater.deflate(Deflater.java:426)
	at java.util.zip.Deflater.deflate(Deflater.java:352)
	at okio.DeflaterSink.long okio.Buffer.size()()
                      void deflate(boolean)
                      void okio.Buffer.setSize$okio(long)
                      void deflate(boolean)
                      void deflate(boolean)
	at okio.DeflaterSink.void write(okio.Buffer,long)(:2)
	at okhttp3.internal.ws.WebSocketWriter.void okhttp3.internal.ws.MessageDeflater.deflate(okio.Buffer)(:7)
                                        void writeMessageFrame(int,okio.ByteString)
	at okhttp3.internal.ws.RealWebSocket.boolean writeOneFrame$okhttp()(:7)
	at okhttp3.internal.ws.RealWebSocket$WriterTask.long runOnce()()
	at okhttp3.internal.concurrent.TaskRunner.void runTask(okhttp3.internal.concurrent.Task)(:3)
                                           void access$runTask(okhttp3.internal.concurrent.TaskRunner,okhttp3.internal.concurrent.Task)
	at okhttp3.internal.concurrent.TaskRunner$runnable$1.void run()(:7)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1133)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:607)
	at java.lang.Thread.run(Thread.java:761)

yschimke:
Can you provide a bug reproduction, it isn't clear from the stacktrace what type of behaviour will trigger this bug.

IvanShafran:
Hello!

I'm getting precisely the same problem on version 4.9.1. Unfortunately, I can't reproduce this. My guess was about the concurrency issue during simultaneous shutdown and writing. More specifically, shutdown should be caused by some failure because I have another problem that may be connected to this: https://github.com/square/okhttp/issues/6723

I had run a stress test with lots of send operations and fail during onMessage callback, but I got nothing from this. Everything was fine.

Is there any chance that two WebSockets with the same endpoint may share Deflater? RealWebSocket code shows that it creates it for every instance. But maybe I missed something.

[...]

PR #8121, opened 2023-12-02
Fix a concurrency problem with web socket cleanup

swankjesse:
We were seeing crashes in Deflater.deflate with a root cause of closing the deflater while it was in use.

The underlying problem is that web socket streams were closed when the web socket was canceled, but not by the thread that owned those streams.

This moves stream cleanup to always run on the owning thread:
 - writer cleanup happens on the task queue
 - reader cleanup happens after loopReader

This also introduces a new function, Streams.cancel() so a failed writer can break an active reader, and vise-versa. Previously we were using close() to do this job.

There's a lot of test changes but it's mostly in adding the TaskFaker facet to RealWebSocketTest. The test now runs with a fake clock, with the deterministic benefits that brings.

See also: https://github.com/square/okhttp/issues/6719

### DIFF
--- a/mockwebserver/src/main/kotlin/mockwebserver3/MockWebServer.kt
+++ b/mockwebserver/src/main/kotlin/mockwebserver3/MockWebServer.kt
@@ -761,6 +761,10 @@ class MockWebServer : Closeable {
     val connectionClose = CountDownLatch(1)
     val streams = object : RealWebSocket.Streams(false, source, sink) {
       override fun close() = connectionClose.countDown()
+
+      override fun cancel() {
+        socket.closeQuietly()
+      }
     }
     val webSocket = RealWebSocket(
       taskRunner = taskRunner,
@@ -771,16 +775,13 @@ class MockWebServer : Closeable {
       extensions = WebSocketExtensions.parse(webSocketResponse.headers),
       minimumDeflateSize = 0L // Compress all messages if compression is enabled.
     )
-    webSocketResponse.webSocketListener.onOpen(webSocket, fancyResponse)
     val name = "MockWebServer WebSocket ${request.path!!}"
     webSocket.initReaderAndWriter(name, streams)
     try {
-      webSocket.loopReader()
+      webSocket.loopReader(fancyResponse)
 
       // Even if messages are no longer being read we need to wait for the connection close signal.
       connectionClose.await()
-    } catch (e: IOException) {
-      webSocket.failWebSocket(e, null)
     } finally {
       source.closeQuietly()
     }
--- a/okhttp/src/jvmMain/kotlin/okhttp3/internal/ws/RealWebSocket.kt
+++ b/okhttp/src/jvmMain/kotlin/okhttp3/internal/ws/RealWebSocket.kt
@@ -87,11 +87,7 @@ class RealWebSocket(
   /** Names this web socket for observability and debugging. */
   private var name: String? = null
 
-  /**
-   * The streams held by this web socket. This is non-null until all incoming messages have been
-   * read and all outgoing messages have been written. It is closed when both reader and writer are
-   * exhausted, or if there is any failure.
-   */
+  /** The streams held by this web socket. This is closed when both reader and writer are closed. */
   private var streams: Streams? = null
 
   /** Outgoing pongs in the order they should be written. */
@@ -145,8 +141,7 @@ class RealWebSocket(
 
   fun connect(client: OkHttpClient) {
     if (originalRequest.header("Sec-WebSocket-Extensions") != null) {
-      failWebSocket(ProtocolException(
-          "Request header not permitted: 'Sec-WebSocket-Extensions'"), null)
+      failWebSocket(ProtocolException("Request header not permitted: 'Sec-WebSocket-Extensions'"))
       return
     }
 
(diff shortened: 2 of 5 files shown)


okhttp/src/jvmTest/java/okhttp3/internal/ws/RealWebSocketTest.java
```diff
@@ -444,17 +467,24 @@ public boolean processNextFrame() throws IOException {
...
-    @Override public void close() throws IOException {
-      getSource().close();
-      getSink().close();
+    @Override public void close() {
       if (closed) {
         throw new AssertionError("Already closed");
       }
+      try {
+        getSource().close();
+      } catch (IOException ignored) {
```
swankjesse, line 476: TODO: convert to Kotlin and get `closeQuietly()` !

mockwebserver/src/main/kotlin/mockwebserver3/MockWebServer.kt
```diff
@@ -771,16 +775,13 @@ class MockWebServer : Closeable {
...
     val name = "MockWebServer WebSocket ${request.path!!}"
     webSocket.initReaderAndWriter(name, streams)
     try {
-      webSocket.loopReader()
+      webSocket.loopReader(fancyResponse)
 
       // Even if messages are no longer being read we need to wait for the connection close signal.
       connectionClose.await()
-    } catch (e: IOException) {
-      webSocket.failWebSocket(e, null)
```
swankjesse, line 783: Moved into loopReader()

okhttp/src/jvmMain/kotlin/okhttp3/internal/ws/RealWebSocket.kt
```diff
@@ -188,18 +183,13 @@ class RealWebSocket(
         }
 
         // Process all web socket messages.
-        try {
-          val name = "$okHttpName WebSocket ${request.url.redact()}"
-          initReaderAndWriter(name, streams)
-          listener.onOpen(this@RealWebSocket, response)
-          loopReader()
-        } catch (e: Exception) {
-          failWebSocket(e, null)
```
swankjesse, line 197: moved into loopReader()

okhttp/src/jvmMain/kotlin/okhttp3/internal/ws/RealWebSocket.kt
```diff
@@ -304,15 +300,54 @@ class RealWebSocket(
...
+      }
+
+      streamsToClose = when {
+        writer == null -> streams
+        else -> null
+      }
+    }
+
+    if (!failed && streamsToClose != null && receivedCloseCode != -1) {
+      listener.onClosed(this, code, reason!!)
```
swankjesse, line 346: This is now published either when the reader exits, or when the writer writes Close, whichever happens last

okhttp/src/jvmMain/kotlin/okhttp3/internal/ws/RealWebSocket.kt
```diff
@@ -356,35 +391,13 @@ class RealWebSocket(
...
-        this.streams = null
-        readerToClose = this.reader
-        this.reader = null
-        writerToClose = this.writer
-        this.writer = null
-        this.taskQueue.shutdown()
-      }
     }
 
-    try {
```
swankjesse, line 377: all this stuff moved to finishReader()

okhttp/src/jvmMain/kotlin/okhttp3/internal/ws/RealWebSocket.kt
```diff
@@ -542,9 +554,8 @@ class RealWebSocket(
 
       return true
     } finally {
-      streamsToClose?.closeQuietly()
-      readerToClose?.closeQuietly()
```
swankjesse, line 546: Bugfix 1: the writer is closing the reader here. Racy and bad!

okhttp/src/jvmMain/kotlin/okhttp3/internal/ws/RealWebSocket.kt
```diff
@@ -560,40 +571,66 @@ class RealWebSocket(
...
+
       taskQueue.shutdown()
     }
 
     try {
       listener.onFailure(this, e, response)
     } finally {
-      streamsToClose?.closeQuietly()
-      readerToClose?.closeQuietly()
-      writerToClose?.closeQuietly()
```
swankjesse, line 596: Bugfix 2: the reader is closing the writer here. Racy and bad!

okhttp/src/jvmMain/kotlin/okhttp3/internal/ws/RealWebSocket.kt
```diff
@@ -560,40 +571,66 @@ class RealWebSocket(
...
+      streamsToCancel?.cancel()
+
+      // If the caller is the writer thread, close it on this thread.
```
yschimke, line 629: Does this logic get simpler with some `taskQueue.executeOrImmediate()` method that is inlined if already on the task queue? It's the same code as line 616, we just don't want the overhead of execute.

But maybe that's just a flourish?

swankjesse: One gotcha is the synchronized statement. I wanna decide what to do in the synchronized block, but do it outside of that block. So the simple abstraction is difficult!

PR #8124, opened 2023-12-03
Update the IntelliJ PNG resource file

swankjesse:
This is still a placeholder, but I like it better.


swankjesse:
Icon source: https://speakerdeck.com/swankjesse/http-in-a-hostile-world-droidcon-toronto-2019?slide=42

PR #8132, opened 2023-12-07
Update README.md

julian-goettingen:
Doc-change that would have avoided my mistake that I made a bug-report about.

Thank you for not only maintaining the library and providing it for free, but for also answering questions about it.


swankjesse:
Thanks!

PR #8140, opened 2023-12-17
Upgrade Kotlin to 1.9.21, Okio to 3.7.0

swankjesse:


okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/RealCall.kt
```diff
@@ -71,7 +71,7 @@ class RealCall(
 
   private val timeout = object : AsyncTimeout() {
     override fun timedOut() {
-      cancel()
+      this@RealCall.cancel()
```
swankjesse, line 74: # 🙀

JakeWharton (approved):

okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/RealCall.kt
yschimke: Good catch

yschimke (approved):

PR #8141, opened 2023-12-17
Change SSE to use the OkHttp public API only

swankjesse:
Previously we prevented end-users from using their own implementations of Call.Factory because we casted down to RealCall in RealEventSource.

With this change we're implementing SSE without depending on any OkHttp implementation details.

This also introduces a new function in EventSources to create an EventSource.Factory from a Call.Factory, and hides the previous implementation that required a concrete OkHttpClient.

Finally this fixes SSE to publish the same EventListener events as regular HTTP calls.

### DIFF
--- a/okhttp-sse/src/main/kotlin/okhttp3/sse/internal/RealEventSource.kt
+++ b/okhttp-sse/src/main/kotlin/okhttp3/sse/internal/RealEventSource.kt
@@ -18,12 +18,9 @@ package okhttp3.sse.internal
 import java.io.IOException
 import okhttp3.Call
 import okhttp3.Callback
-import okhttp3.EventListener
-import okhttp3.OkHttpClient
 import okhttp3.Request
 import okhttp3.Response
 import okhttp3.ResponseBody
-import okhttp3.internal.connection.RealCall
 import okhttp3.internal.stripBody
 import okhttp3.sse.EventSource
 import okhttp3.sse.EventSourceListener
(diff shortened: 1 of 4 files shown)


JakeWharton (approved):

okhttp-sse/src/main/kotlin/okhttp3/sse/internal/RealEventSource.kt
```diff
@@ -64,7 +58,7 @@ internal class RealEventSource(
       }
 
       // This is a long-lived response. Cancel full-call timeouts.
-      call?.timeoutEarlyExit()
+      call?.timeout()?.cancel()
```
yschimke, line 61: This is much better

okhttp-sse/src/test/java/okhttp3/sse/internal/EventSourceHttpTest.java
```diff
@@ -177,6 +181,41 @@ public void cancelInEventShortCircuits() throws IOException {
...
+
+    EventSource source = newEventSource();
+
+    assertThat(source.request().url().encodedPath()).isEqualTo("/");
+
+    listener.assertOpen();
+    listener.assertEvent(null, null, "hey");
+    listener.assertClose();
+
+    assertThat(eventListener.recordedEventTypes()).containsExactly(
```
yschimke, line 199: Yeah, this seems reasonable, if not particularly useful.

yschimke (approved):

PR #8149, opened 2023-12-19
Switch back to the kotlin JVM plugin

swankjesse:
This does a ton of file moves from jvmMain to main, and jvmTest to test.


swankjesse:
# ⚠️
Ugh, I did a stacked PRs fail on this one. Viewing only the first 2 commits in the diff view should help.

PR #8151, opened 2023-12-20
Even more assertk

swankjesse:
When we split tests into multiplatform we switched them from AssertJ to assertk. I'm migrating everything over to assertk so that I can merge tests back together without changing assertion APIs.

### DIFF
--- a/okhttp-testing-support/src/test/kotlin/okhttp3/OkHttpClientTestRuleTest.kt
+++ b/okhttp-testing-support/src/test/kotlin/okhttp3/OkHttpClientTestRuleTest.kt
@@ -15,7 +15,8 @@
  */
 package okhttp3
 
-import org.assertj.core.api.Assertions.assertThat
+import assertk.assertThat
+import assertk.assertions.hasMessage
 import org.junit.jupiter.api.Test
 import org.junit.jupiter.api.extension.BeforeEachCallback
 import org.junit.jupiter.api.extension.ExtensionContext
@@ -46,7 +47,7 @@ class OkHttpClientTestRuleTest {
       fail("")
     } catch (expected: AssertionError) {
       assertThat(expected).hasMessage("uncaught exception thrown during test")
-      assertThat(expected.cause).hasMessage("boom!")
+      assertThat(expected.cause!!).hasMessage("boom!")
     }
   }
 }
--- a/okhttp/src/test/java/okhttp3/CallTest.kt
+++ b/okhttp/src/test/java/okhttp3/CallTest.kt
@@ -15,6 +15,22 @@
  */
 package okhttp3
 
+import assertk.assertThat
+import assertk.assertions.contains
+import assertk.assertions.containsExactly
+import assertk.assertions.doesNotContain
+import assertk.assertions.hasMessage
+import assertk.assertions.isCloseTo
+import assertk.assertions.isEmpty
+import assertk.assertions.isEqualTo
+import assertk.assertions.isFalse
+import assertk.assertions.isLessThan
+import assertk.assertions.isNotEmpty
+import assertk.assertions.isNotNull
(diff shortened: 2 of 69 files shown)


okhttp-testing-support/src/test/kotlin/okhttp3/OkHttpClientTestRuleTest.kt
```diff
@@ -46,7 +47,7 @@ class OkHttpClientTestRuleTest {
       fail("")
     } catch (expected: AssertionError) {
       assertThat(expected).hasMessage("uncaught exception thrown during test")
-      assertThat(expected.cause).hasMessage("boom!")
+      assertThat(expected.cause!!).hasMessage("boom!")
```
yschimke, line 50: surprised there isnt a nice way to do this with clean messages not NPE.

yschimke: Ahh, it's Throwable, not string.  https://github.com/willowtreeapps/assertk/blob/5ff8a09f63543a18365dee1bb48dc9c4684aab9e/assertk/src/commonMain/kotlin/assertk/assertions/throwable.kt

okhttp/src/test/java/okhttp3/CallTest.kt
```diff
@@ -2855,7 +2869,7 @@ open class CallTest {
     server.enqueue(MockResponse())
     executeSynchronously("/")
     val recordedRequest = server.takeRequest()
-    assertThat(recordedRequest.headers["User-Agent"]).matches(userAgent)
+    assertThat(recordedRequest.headers["User-Agent"]!!).isEqualTo(userAgent)
```
yschimke, line 2872: Shouldn't this be supported without !!

https://github.com/willowtreeapps/assertk/blob/5ff8a09f63543a18365dee1bb48dc9c4684aab9e/assertk/src/commonMain/kotlin/assertk/assertions/string.kt#L6

okhttp/src/test/java/okhttp3/internal/UtilTest.kt
```diff
@@ -49,33 +52,37 @@ class UtilTest {
...
-        assertThatIllegalArgumentException().isThrownBy { checkDuration("timeout", 1, TimeUnit.NANOSECONDS) }
-            .withMessage("timeout too small")
-        assertThatIllegalArgumentException().isThrownBy {
-            checkDuration(
-                "timeout",
-                1L + Int.MAX_VALUE.toLong(),
-                TimeUnit.MILLISECONDS
-            )
-        }.withMessage("timeout too large")
+        assertThat(assertThrows<IllegalStateException> {
```
yschimke, line 55: This isn't a simple migration.

yschimke (approved):
Nothing blocking

JakeWharton (approved):
Good stuff.

Follow ups we should do:
- `try`/`fail` --> `assertFailure`
- `assertThat(...!!)` --> `assertThat(...).isNotNull()`
- `assertThat(assertThrows<..>(...))` --> `assertFailure(...).isInstanceOf<..>()`

okhttp-testing-support/src/test/kotlin/okhttp3/OkHttpClientTestRuleTest.kt
JakeWharton: It would be with a `isNotNull()` which refines an `Assert<T?>` into an `Assert<T>`

okhttp/src/test/java/okhttp3/internal/UtilTest.kt
JakeWharton: See my other comment. There's an `assertFailure` we can use. Hopefully with structural search/replace.

PR #8152, opened 2023-12-21
Convert some tests to Kotlin

swankjesse:


swankjesse:
I split it into two commits to make it easier to compare the effects of converting to Kotlin.
https://github.com/square/okhttp/pull/8152/commits/d1489568d5e3bb601f3f9c58970f43d2d2bf108f?diff=split&w=1#

I’d like to replace all uses of try/fail with assertFailure throughout OkHttp’s tests. But a prerequisite to that is converting tests that use the old pattern to Kotlin!

yschimke (approved):
I'm hoping that was mainly mechanical. I skimmed and patterned match by eye the dark green sections of the diffs. But could have missed something you edited :)

Issue #8154, opened 2023-12-21
5.0.0-alpha.12: java.lang.IllegalStateException: Unable to load /okhttp3/internal/publicsuffix/a.gz resource from the classpath.


ahagelstein:
I have created a code sample which shows the issue.

- Android Studio Hedgehog | 2023.1.1
- OkHttp: 5.0.0-alpha.12
- Kotlin: 1.9.21
- Version of Gradle Plugin: id 'com.android.application' version '8.2.0'
- Version of Gradle: gradle-8.5-bin.zip
- OS: Windows 11 22H2

Steps to Reproduce:

Compile and run the following code in release mode (with minifyEnabled = true and shrinkResources = true).

```
class MainActivity : ComponentActivity() {

    override fun onCreate(savedInstanceState: Bundle?) {

        super.onCreate(savedInstanceState)

        PublicSuffixDatabase.get().getEffectiveTldPlusOne("com")
    }
}
```

Result:

```
Caused by: java.lang.IllegalStateException: Unable to load /okhttp3/internal/publicsuffix/a.gz resource from the classpath.
                                                                                                    	at w3.a.a(Unknown Source:479)
                                                                                                    	at de.telekom.testapp.MainActivity.onCreate(Unknown Source:12)
                                                                                                    	at android.app.Activity.performCreate(Activity.java:6975)
                                                                                                    	at android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1213)
                                                                                                    	at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2770)
                                                                                                    	at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2892) 
                                                                                                    	at android.app.ActivityThread.-wrap11(Unknown Source:0) 
                                                                                                    	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1593) 
                                                                                                    	at android.os.Handler.dispatchMessage(Handler.java:105) 
                                                                                                    	at android.os.Looper.loop(Looper.java:164) 
                                                                                                    	at android.app.ActivityThread.main(ActivityThread.java:6541) 
                                                                                                    	at java.lang.reflect.Method.invoke(Native Method) 
                                                                                                    	at com.android.internal.os.Zygote$MethodAndArgsCaller.run(Zygote.java:240) 
                                                                                                    	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:767) 
```

[...]

PR #8156, opened 2023-12-23
Test with release build & update proguard config.

yschimke:
Fix for https://github.com/square/okhttp/issues/8154

Keep the DB package name since it's now a literal.

Also add a test of a release build of an app.

### DIFF
--- a/android-test-app/src/main/kotlin/okhttp/android/testapp/TestApplication.kt
+++ b/android-test-app/src/main/kotlin/okhttp/android/testapp/TestApplication.kt
@@ -1,0 +1,21 @@
+/*
+ * Copyright (C) 2023 Block, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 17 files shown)


yschimke:
@swankjesse I'll fix the tests in follow up

Seem regexy, maybe your other change?

https://github.com/square/okhttp/actions/runs/7308237700/job/19914706476#step:8:443

android-test-app/src/main/kotlin/okhttp/android/testapp/TestApplication.kt
```diff
@@ -0,0 +1,21 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okhttp.android.testapp
+
+import android.app.Application
+
+class TestApplication: Application() {
```
JakeWharton, line 20: Why was this added? Seems unneeded.

JakeWharton (approved):

android-test-app/src/main/kotlin/okhttp/android/testapp/TestApplication.kt
yschimke: Not strictly necessary, but I wanted the full structure. Not sure whether to delete now, might leave and reconsider.

yschimke:
Will fix those tests in follow ups.

swankjesse (commented):
LGTM

PR #8159, opened 2023-12-24
Bump CI build from JDK 19 to 21 (LTS)

yschimke:


yschimke:
Secretly hoping it fixes https://github.com/square/okhttp/actions/runs/7313047455/job/19924201870

and we get off to the new year with clean builds.

okhttp/src/test/java/okhttp3/HttpUrlTest.kt
```diff
@@ -560,7 +561,8 @@ open class HttpUrlTest {
         ']'.code
       )
       .override(
-        Encoding.IDENTITY,
+        // java.net.URL got stricter
+        if (PlatformVersion.majorVersion >= 21)  Encoding.SKIP else Encoding.IDENTITY,
```
yschimke, line 565: @swankjesse I'm not sure this is right, is there a better way to make assertions here? That it now fails for URL?

yschimke:
@swankjesse requesting post review.  Particularly over how to handle JDK 21 URI strictness.

yschimke:
Landing to get builds green.

okhttp/src/test/java/okhttp3/HttpUrlTest.kt
swankjesse: Lemme attempt a follow-up. I think we need to fix our code.

swankjesse (commented):
LGTM

okhttp/src/test/java/okhttp3/HttpUrlTest.kt
swankjesse: https://github.com/square/okhttp/pull/8363

PR #8161, opened 2023-12-24
Avoid checking file metadata in graal tests

yschimke:
Failing in https://github.com/square/okhttp/actions/runs/7313504687/job/19925152167


yschimke:

Likely I should dig, but we know the code is working, just not this arbitrary test that looks for a resource with Okio FileSystem.RESOURCES.

swankjesse (commented):
LGTM !

PR #8173, opened 2024-01-03
Resolve JDKs if necessary

swankjesse:
Without this, developers need to prepare their environments or they might see failures like this:

    Execution failed for task ':okhttp-idna-mapping-table:test'.
    > Error while evaluating property 'javaLauncher' of task ':okhttp-idna-mapping-table:test'.
       > Failed to calculate the value of task ':okhttp-idna-mapping-table:test' property 'javaLauncher'.
          > No matching toolchains found for requested specification: {languageVersion=21, vendor=any, implementation=vendor-specific} for MAC_OS on aarch64.
             > No locally installed toolchains match and toolchain download repositories have not been configured.

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -1,5 +1,19 @@
 rootProject.name = "okhttp-parent"
 
+plugins {
+  id("org.gradle.toolchains.foojay-resolver") version("0.7.0")
+}
+
+toolchainManagement {
+  jvm {
+    javaRepositories {
+      repository("foojay") {
+        resolverClass.set(org.gradle.toolchains.foojay.FoojayToolchainResolver::class.java)
+      }
+    }
+  }
+}
+
 include(":mockwebserver")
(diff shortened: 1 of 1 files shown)


yschimke (approved):

settings.gradle.kts
```diff
@@ -1,5 +1,19 @@
...
+
+toolchainManagement {
+  jvm {
+    javaRepositories {
+      repository("foojay") {
+        resolverClass.set(org.gradle.toolchains.foojay.FoojayToolchainResolver::class.java)
+      }
+    }
+  }
+}
```
JakeWharton, line 15:
```suggestion
plugins {
  id("org.gradle.toolchains.foojay-resolver-convention") version "0.7.0"
}
```

JakeWharton (approved):

settings.gradle.kts
swankjesse: oh so much better

PR #8175, opened 2024-01-04
Introduce @ExperimentalOkHttpApi

swankjesse:
I'm intending to use this to release OkHttp 5.0 before all of the new APIs are final.

### DIFF
--- a/okhttp/src/main/kotlin/okhttp3/ExperimentalOkHttpApi.kt
+++ b/okhttp/src/main/kotlin/okhttp3/ExperimentalOkHttpApi.kt
@@ -1,0 +1,44 @@
+/*
+ * Copyright (C) 2024 Block, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okhttp3
+
+/**
+ * Marks declarations that are experimental and subject to change without following SemVer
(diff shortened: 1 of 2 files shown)


swankjesse:
(For a follow-up PR: annotating everything that I consider experimental!)

okhttp/src/main/kotlin/okhttp3/ExperimentalOkHttpApi.kt
```diff
@@ -0,0 +1,44 @@
...
+  AnnotationTarget.FIELD,
+  AnnotationTarget.LOCAL_VARIABLE,
+  AnnotationTarget.VALUE_PARAMETER,
+  AnnotationTarget.CONSTRUCTOR,
+  AnnotationTarget.FUNCTION,
+  AnnotationTarget.PROPERTY_GETTER,
+  AnnotationTarget.PROPERTY_SETTER,
+  AnnotationTarget.TYPEALIAS
+)
+@RequiresOptIn(level = RequiresOptIn.Level.WARNING)
```
JakeWharton, line 43: I'm tempted to say this should be error.

JakeWharton (approved):

okhttp/src/main/kotlin/okhttp3/ExperimentalOkHttpApi.kt
```diff
@@ -0,0 +1,44 @@
...
+ * limitations under the License.
+ */
+package okhttp3
+
+/**
+ * Marks declarations that are experimental and subject to change without following SemVer
+ * conventions. Both binary and source-incompatible changes are possible, including complete removal
+ * of the experimental API.
+ *
+ * Do not use these APIs in modules that may be executed using a version of OkHttp different from
```
yschimke, line 23: Do not use in published libraries without also marking experimental?

yschimke (approved):

okhttp/src/main/kotlin/okhttp3/ExperimentalOkHttpApi.kt
swankjesse: Done.

okhttp/src/main/kotlin/okhttp3/ExperimentalOkHttpApi.kt
swankjesse: Even better: Do not use these APIs in published libraries.

PR #8177, opened 2024-01-06
Switch to assertFailsWith

swankjesse:


mockwebserver-deprecated/src/test/java/okhttp3/mockwebserver/MockWebServerTest.kt
```diff
@@ -500,11 +501,10 @@ class MockWebServerTest {
     val connection = url.openConnection() as HttpURLConnection
     assertThat(connection.getResponseCode()).isEqualTo(HttpURLConnection.HTTP_OK)
     val refusedConnection = url.openConnection() as HttpURLConnection
-    try {
+    assertFailsWith<ConnectException> {
       refusedConnection.getResponseCode()
-      fail<Any>("Second connection should be refused")
-    } catch (e: ConnectException) {
-      assertThat(e.message!!).contains("refused")
+    }.also { expected ->
```
swankjesse, line 506: This idiom isn’t perfect, but it was easy enough to Structurally Replace and it makes the diffs pretty easy to review

mockwebserver/src/test/java/mockwebserver3/MockWebServerTest.kt
```diff
@@ -496,11 +497,10 @@ class MockWebServerTest {
...
     assertThat(connection.responseCode).isEqualTo(HttpURLConnection.HTTP_OK)
     val refusedConnection = url.openConnection() as HttpURLConnection
-    try {
+    assertFailsWith<ConnectException> {
       refusedConnection.responseCode
-      fail<Any?>("Second connection should be refused")
-    } catch (e: ConnectException) {
-      assertThat(e.message!!).contains("refused")
+    }.also { expected ->
+      assertThat(expected.message!!).contains("refused")
```
swankjesse, line 503: I also renamed `e` to `expected` wherever I could

okhttp-brotli/src/test/java/okhttp3/brotli/BrotliInterceptorTest.kt
```diff
@@ -21,6 +21,7 @@ import assertk.assertions.hasMessage
 import assertk.assertions.isEmpty
 import assertk.assertions.isEqualTo
 import java.io.IOException
+import kotlin.test.assertFailsWith
```
swankjesse, line 24: I’m using assertFailsWith instead of assertFailure because assertFailure needs a 2nd assertion to specify the exception type

JakeWharton: Yeah I argued for an overload that took a reified exception type when I added `assertFailure` and lost (or, well, didn't press it too hard when given push-back). 

Maybe file a feature request and we can try again?

okhttp-coroutines/build.gradle.kts
```diff
@@ -20,8 +20,9 @@ dependencies {
   api(libs.squareup.okio)
   api(libs.kotlin.stdlib)
 
-  testImplementation(libs.kotlin.test.common)
   testImplementation(libs.kotlin.test.annotations)
+  testImplementation(libs.kotlin.test.common)
+  testImplementation(libs.kotlin.test.junit)
```
JakeWharton, line 25: The need for all three of these is very strange. You should really only need "common" as the Gradle module metadata will resolve both the annotations and the junit implementation automatically in almost all cases.

There are a few exceptions, maybe in the MWS modules, where you have to be explicit. But 99% of modules should get the automatic behavior.

JakeWharton (approved):

JakeWharton:
Saw some out-of-order imports. Probably need a `spotApply` run/

okhttp-coroutines/build.gradle.kts
swankjesse: Yeah, I’m frustrated with this one. For reasons unclear to me omitting `libs.kotlin.test.junit` causes compile to fail.

```
> Task :okhttp-brotli:compileTestKotlin FAILED
e: file:///Volumes/Development/okhttp/okhttp-brotli/src/test/java/okhttp3/brotli/BrotliInterceptorTest.kt:24:20 Unresolved reference: assertFailsWith
```

swankjesse:
The whole repo needs some Spotless love. It’s currently misconfigured and broken. I’ve got that on my queue for demultiplatform cleanup.

okhttp-brotli/src/test/java/okhttp3/brotli/BrotliInterceptorTest.kt
swankjesse: https://github.com/willowtreeapps/assertk/issues/511

PR #8178, opened 2024-01-06
Attempt to fix the android-test build

swankjesse:


swankjesse:
Failing on CI like so:

```

> Task :android-test:compileDebugAndroidTestKotlin FAILED
e: file:///Users/runner/work/okhttp/okhttp/okhttp-brotli/src/test/java/okhttp3/brotli/BrotliInterceptorTest.kt:24:15 Unresolved reference: test
```

JakeWharton (approved):

PR #8179, opened 2024-01-07
Prepare to enable Spotless

swankjesse:
I'm not super keen on any of these changes, but I am very keen on having automatic formatting.


okhttp/src/test/java/okhttp3/CacheTest.kt
```diff
@@ -319,8 +318,8 @@ class CacheTest {
       .build()
     val request = Request.Builder().url(server.url("/")).build()
     val response1 = client.newCall(request).execute()
-    val `in` = response1.body.source()
-    assertThat(`in`.readUtf8()).isEqualTo("ABC")
+    val source = response1.body.source()
+    assertThat(source.readUtf8()).isEqualTo("ABC")
```
JakeWharton, line 322: This shouldn't be needed with ktlint 1.1 anymore

JakeWharton (approved):

okhttp/src/test/java/okhttp3/CacheTest.kt
swankjesse: Good to know.

PR #8180, opened 2024-01-07
Reformat with Spotless

swankjesse:


.editorconfig
```diff
@@ -2,9 +2,10 @@ root = true
...
 [*]
 indent_size = 2
+ij_continuation_indent_size = 2
 charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
 
 [*.{kt, kts}]
-kotlin_imports_layout = ascii
+ij_kotlin_imports_layout = *
```
swankjesse, line 11: I don’t know the difference between these. I’m just doing what Zipline does.

android-test/src/androidTest/java/okhttp/android/test/OkHttpTest.kt
```diff
@@ -184,22 +189,29 @@ class OkHttpTest {
 
       var socketClass: String? = null
 
-      val clientCertificates = HandshakeCertificates.Builder()
-        .addPlatformTrustedCertificates()
-        .addInsecureHost(server.hostName)
-        .build()
+      val clientCertificates =
+        HandshakeCertificates.Builder()
```
swankjesse, line 193: I don’t understand why it wrapped the constructor call?

swankjesse:
I like the nice manual formatting better.

But I want to put robots in charge of formatting, so I’m willing to accept the robots’ dumb formatting rules.

android-test/src/androidTest/java/okhttp/android/test/OkHttpTest.kt
JakeWharton: Looks like it started doing https://github.com/google/google-java-format/wiki/The-Rectangle-Rule

mockwebserver-deprecated/src/main/kotlin/okhttp3/mockwebserver/MockResponse.kt
```diff
@@ -86,62 +87,81 @@ class MockResponse : Cloneable {
...
+      headersBuilder.add(header)
+    }
 
-  fun addHeader(name: String, value: Any) = apply {
+  fun addHeader(
+    name: String,
+    value: Any,
+  ) = apply {
     headersBuilder.add(name, value.toString())
   }
```
JakeWharton, line 129: Seems like a bug that it wraps the expression body onto its own line if the function signature is a single line, but if the function signature wraps it does not wrap the expression body.

JakeWharton (approved):

PR #8190, opened 2024-01-13
Handle DnsResolver direct exception on API 34

yschimke:
Failing on master build https://github.com/square/okhttp/actions/runs/7461208679/job/20300867655

Also seen here https://github.com/Ehviewer-Overhauled/Ehviewer/issues/1311


yschimke:
Can't see an easy way to add a test android.net.DnsResolver is final, and no shadow for robolectric.

JakeWharton:
Robolectric lets you provide your own shadows if you need one it doesn't provide. Should go into `@Config`.

https://robolectric.org/extending/#using-a-custom-shadows

yschimke:
Need to make a few more tests

yschimke:
@JakeWharton care to take another look?

JakeWharton:
I can on Monday

okhttp-android/src/main/kotlin/okhttp3/android/AndroidAsyncDns.kt
```diff
@@ -40,38 +40,47 @@ class AndroidAsyncDns(
...
+            callback.onFailure(
+              hostname,
+              UnknownHostException(e.message).apply {
+                initCause(e)
+              },
+            )
+          }
+        },
+      )
+    } catch (e: Exception) {
```
swankjesse, line 76: Any idea what kinds of exceptions we’ll be potentially hiding here? We’re attributing an arbitrary failure like an IllegalStateException as an unknown host

swankjesse: I saw from the attached but report it’s gonna happen when the DNS isn't available.

```
IllegalArgumentException: Network.fromNetworkHandle refusing to instantiate NETID_UNSET Network.
```

I agree with this PR, we’ve got to be resilient to this API that doesn’t offer us a ton of good options.

yschimke: I'll add a comment.

yschimke: https://github.com/square/okhttp/pull/8322/files

PR #8191, opened 2024-01-13
Allow constructor injection of MockWebServer

yschimke:


yschimke:
@swankjesse interestingly, some failures on an earlier build were familiar flaky tests.  maybe a clue on why it's failing.

https://github.com/square/okhttp/actions/runs/7512760479/job/20453782860

JakeWharton (approved):

yschimke:
Will test on android as I think that's why we needed the extend with.

swankjesse (commented):
LGTM

android-test/build.gradle.kts
```diff
@@ -20,7 +20,8 @@ android {
     testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
     testInstrumentationRunnerArguments += mapOf(
       "runnerBuilder" to "de.mannodermaus.junit5.AndroidJUnit5Builder",
-      "notPackage" to "org.bouncycastle"
+      "notPackage" to "org.bouncycastle",
+      "configurationParameters" to "junit.jupiter.extensions.autodetection.enabled=true"
```
swankjesse, line 24: Do our end-users need to do this? JUnit 5 is such a mess.

TWiStErRob: don't need to, users have the option of
 * using the extension directly by annotating test class.
 * using the extension directly by annotating an annotation to collect extensions
 * setting this property to magically apply extension to all tests

TWiStErRob: Oh, just realized that this is in internal package, but it's public API, as users can choose how to use the extensions (you choose to use the atuodetection in this projecet). Could this be made actual API? @swankjesse

swankjesse: I think we should change our MockWebServer JUnit 5 extension API to use an annotation on each annotated class. It’s two imports instead of JUnit 4’s single import, but it’s self-contained in the test class.

PR #8199, opened 2024-01-15
Add WebSocket.maxQueueSize and setMaxQueueSize methods

mrexodia:
Discussion: #3619

### DIFF
--- a/okhttp/src/main/kotlin/okhttp3/WebSocket.kt
+++ b/okhttp/src/main/kotlin/okhttp3/WebSocket.kt
@@ -64,6 +64,17 @@ interface WebSocket {
   fun queueSize(): Long
 
   /**
+   * Returns the maximum size in bytes of the messages enqueued to be transmitted to the server.
+   */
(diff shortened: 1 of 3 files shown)


yschimke:
Failing on

```
> Task :spotlessKotlin FAILED
Step 'ktlint' found problem in 'okhttp/src/main/kotlin/okhttp3/internal/ws/RealWebSocket.kt':
Error on line: 103, column: 3
rule: standard:no-consecutive-comments
a KDoc may not be preceded by a KDoc
java.lang.AssertionError: Error on line: 103, column: 3
rule: standard:no-consecutive-comments
a KDoc may not be preceded by a KDoc
```

yschimke:
See discussion here also https://github.com/square/okhttp/pull/8042

mrexodia:
Ah I didn’t see that. I do hope you will reconsider though, 16MiB is extremely small and making it configurable with the proper warnings is a very small change as you can see. I’ve been using 100MiB and there are no adverse effects I have observed at all.

Issue #8083, opened 2023-10-29
Update Cache docs to mention that it's not thread safe and should be a singleton


odbol:
Quite a few people have been running into some concurrency issues with Cache (see https://github.com/square/okhttp/issues/8080, https://github.com/square/okhttp/issues/3938, https://github.com/square/okhttp/issues/6280, https://github.com/voghDev/PdfViewPager/issues/175, etc).

Most likely for all of them they are either creating more than one Cache object on the same directory/file, or accessing it from different threads. But it's not clear from the docs that this would cause a problem.

The doc comment for Cache should be updated to mention that it is not thread-safe and that only one Cache object per directory should be created (i.e. make it a singleton).

yschimke:
As well as some docs, we should consider something like checking a static set of locks. ExoPlayer SimpleCache does this.

https://github.com/androidx/media/blob/3f366109b7e5cbf7b7e5b46e8b923e167e9ae893/libraries/datasource/src/main/java/androidx/media3/datasource/cache/SimpleCache.java#L83C38-L83C57

yschimke:
@odbol I don't think it's lack of thread safety, I think it's not safe with multiple cache instances on the same directory.

PR #8209, opened 2024-01-20
[Blocked] Exclusive Cache Lock

yschimke:
Implement a strict CacheLock that works within the process, and also via file system locking if supported.

We can expect a bunch of failures in apps that aren't strictly using a single Cache instance, but should be.

Follow up to https://github.com/square/okhttp/issues/8083

Todo

- [ ] Blocked on https://github.com/square/okio/issues/1464 https://github.com/square/okio/pull/1677


okhttp/src/main/kotlin/okhttp3/internal/cache/CacheLock.kt
```diff
@@ -0,0 +1,44 @@
...
+      fileSystemLock(directory)
+    } else {
+      inMemoryLock(directory)
+    }
+  }
+
+  @SuppressLint("NewApi")
+  fun inMemoryLock(directory: Path): Closeable {
+    // TODO solution for Android N?
+    val existing = openCaches.putIfAbsent(directory, Exception("CacheLock($directory)"))
```
swankjesse, line 26: Is `putIfAbsent()` atomic on synchronized maps? I assume it’s not?

okhttp/src/main/kotlin/okhttp3/internal/cache/CacheLock.kt
```diff
@@ -0,0 +1,44 @@
+package okhttp3.internal.cache
```
swankjesse, line 1: nit: copyright

okhttp/src/main/kotlin/okhttp3/internal/cache/CacheLock.kt
```diff
@@ -0,0 +1,44 @@
...
+    if (existing != null) {
+      throw IllegalStateException("Cache already open at '$directory'", existing)
+    }
+    return okio.Closeable {
+      openCaches.remove(directory)
+    }
+  }
+
+  @SuppressLint("NewApi")
+  fun fileSystemLock(directory: Path): Closeable {
```
swankjesse, line 36: Does this work?!

okhttp/src/main/kotlin/okhttp3/internal/cache/CacheLock.kt
yschimke:
```
        @Override
        public V putIfAbsent(K key, V value) {
            synchronized (mutex) {return m.putIfAbsent(key, value);}
        }
```

okhttp/src/main/kotlin/okhttp3/internal/cache/CacheLock.kt
yschimke: I think it does now. I have tested manually with two processes.

okhttp/src/main/kotlin/okhttp3/internal/cache/CacheLock.kt
swankjesse: Oh neat. Today I learned!

okhttp/src/main/kotlin/okhttp3/internal/cache/CacheLock.kt
```diff
@@ -0,0 +1,80 @@
...
+import org.codehaus.mojo.animal_sniffer.IgnoreJRERequirement
+
+internal object CacheLock {
+  private val openCaches = Collections.synchronizedMap(mutableMapOf<Path, Exception>())
+
+  fun openLock(
+    fileSystem: FileSystem,
+    directory: Path,
+  ): Closeable {
+    return if (fileSystem == FileSystem.SYSTEM && !Platform.isAndroid) {
```
swankjesse, line 35: Ooooh this is a pretty loud request for Okio to support locks!

okhttp/src/main/kotlin/okhttp3/internal/cache/DiskLruCache.kt
```diff
@@ -242,6 +245,8 @@ class DiskLruCache(
 
     civilizedFileSystem = fileSystem.isCivilized(journalFileBackup)
 
+    cacheLock = openLock(fileSystem, directory)
```
swankjesse, line 248: Is fileSystem ever an instance of `FileSystem.SYSTEM` I think we might wrap it ?

okhttp/src/test/java/okhttp3/CacheLockTest.kt
```diff
@@ -0,0 +1,82 @@
...
+import okio.Closeable
+import okio.FileSystem
+import okio.Path.Companion.toOkioPath
+import org.junit.jupiter.api.AfterEach
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.assertThrows
+import org.junit.jupiter.api.io.TempDir
+
+class CacheLockTest {
```
swankjesse, line 30: Wanna add the big complicated test that forks a process so we can be confident that it works across processes?

okhttp/src/test/java/okhttp3/CacheLockTest.kt
```diff
@@ -0,0 +1,82 @@
...
+  @Test
+  fun testCacheLockDifferentPath() {
+    openCache(tempDir / "a")
+
+    openCache(tempDir / "b")
+  }
+
+  private fun openCache(directory: okio.Path): Cache {
+    return Cache(directory, 10_000, FileSystem.SYSTEM).apply {
+      // force early LRU initialisation
```
swankjesse, line 77: I’m a bit torn on this. You don’t get a failure immediately; you get it eventually.

This is probably fine; this new feature is advisory.

okhttp/src/main/kotlin/okhttp3/internal/cache/CacheLock.kt
```diff
@@ -0,0 +1,80 @@
...
+  @IgnoreJRERequirement // only called on JVM
+  fun fileSystemLock(
+    memoryLock: Closeable,
+    directory: Path,
+  ): Closeable {
+    val lockFile = directory / "lock"
+    lockFile.toFile().createNewFile()
+    val channel = FileChannel.open(lockFile.toNioPath(), StandardOpenOption.APPEND)
+
+    checkNotNull(channel.tryLock()) {
```
swankjesse, line 71: I’m anxious that this is going to fail for reasons related to the host file system’s lack of support, and not because the file is positively locked.

Would it be horrible to probe the file system if a lock fails to see if _EVERY_ lock is going to fail? Maybe we attempt to lock a file named with a random token or a timestamp? If that also fails we know it’s the file system’s fault?

swankjesse (commented):
This is a good feature but I’d like a lot more integration tests before we put this into end users’ hands

okhttp/src/test/java/okhttp3/CacheLockTest.kt
yschimke: Yeah, this currently works until it doesn't. Then it ends up as 40+ bugs raised.

okhttp/src/main/kotlin/okhttp3/internal/cache/CacheLock.kt
yschimke: A big +1 to okio solving this in a "Swanky" new API

yschimke: But I'm not blocked on that.

okhttp/src/test/java/okhttp3/CacheLockTest.kt
yschimke: I put a single test in, will use it to flush out issues across platforms. Let me know what you think.

okhttp/src/test/java/okhttp3/CacheLockTest.kt
yschimke: We can't really do IO operations on the main thread, so this should always be the case.

Which reminds me, for SSL init on main thread https://github.com/square/okhttp/pull/8248

okhttp/src/main/kotlin/okhttp3/internal/cache/CacheLock.kt
yschimke: I'll need to think more about this case.

okhttp/src/test/java/okhttp3/CacheLockTest.kt
```diff
@@ -0,0 +1,126 @@
...
+      @Suppress("Since15")
+      ProcessHandle.current().info().command().get().toPath()
+    } else {
+      System.getenv("JAVA_HOME").toPath() / "bin/java"
+    }
+
+    assertThat(FileSystem.SYSTEM.exists(javaExe))
+
+    val process =
+      ProcessBuilder().command(javaExe.toString(), "src/test/java/okhttp3/LockTestProgram.java", (lockFile.toString()))
```
yschimke, line 99: This was what I came up with.  It nicely gives us the same editing experience.

But I guess it's Java 11 only?

I can skip the test on Java 8, or run the compile class file?

okhttp/src/test/java/okhttp3/CacheLockTest.kt
swankjesse: I agree with ‘We can’t really do I/O operations on the main thread’, but I don’t agree that OkHttpClient is being initialized on the main thread. We’re an I/O library!

okhttp/src/test/java/okhttp3/CacheLockTest.kt
```diff
@@ -0,0 +1,126 @@
...
+import okio.Path.Companion.toOkioPath
+import okio.Path.Companion.toPath
+import org.junit.jupiter.api.AfterEach
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.assertThrows
+import org.junit.jupiter.api.extension.RegisterExtension
+import org.junit.jupiter.api.io.TempDir
+
+@org.junit.jupiter.api.parallel.Isolated
```
swankjesse, line 34: Neato.

Import ?

swankjesse (commented):
I like the PR!

But I’m still anxious about how it’ll work on real-world file systems. We probably should confirm it does something reasonable on Mac, Windows, Linux and Android? (Why is this off-by-default on Android?)

yschimke:
Yep, next stage is forcing it through.

I forget why it's not on Android, I'll enable again and see. Or comment for future self.

yschimke:
OK, running ok on Android.

Most tests failing validly since they don't close Cache instances with

```
 DiskLruCacheTest > recoverFromInitializationFailure(Pair) > [3] (FakeFileSystem, true) FAILED
    java.lang.AssertionError: Expected an exception of class java.io.IOException to be thrown, but was java.lang.IllegalStateException: Cache already open at '/cache' in same process
        at org.junit.Assert.fail(Assert.java:89)
        at kotlin.test.junit.JUnitAsserter.fail(JUnitSupport.kt:64)
        at kotlin.test.AssertionsKt__AssertionsImplKt.checkResultIsFailure(AssertionsImpl.kt:30)
        at kotlin.test.AssertionsKt.checkResultIsFailure(Unknown Source)
        at okhttp3.internal.cache.DiskLruCacheTest.recoverFromInitializationFailure(DiskLruCacheTest.kt:140)

        Caused by:
        java.lang.IllegalStateException: Cache already open at '/cache' in same process
            at okhttp3.internal.cache.CacheLock.inMemoryLock(CacheLock.kt:51)
            at okhttp3.internal.cache.CacheLock.openLock(CacheLock.kt:39)
            at okhttp3.internal.cache.DiskLruCache.initialize(DiskLruCache.kt:248)
            at okhttp3.internal.cache.DiskLruCache.get(DiskLruCache.kt:454)
            at okhttp3.internal.cache.DiskLruCacheTest.recoverFromInitializationFailure(DiskLruCacheTest.kt:141)

            Caused by:
            java.lang.Exception: CacheLock(/cache)
                at okhttp3.internal.cache.CacheLock.inMemoryLock(CacheLock.kt:49)
                at okhttp3.internal.cache.CacheLock.openLock(CacheLock.kt:39)
                at okhttp3.internal.cache.DiskLruCache.initialize(DiskLruCache.kt:248)
                at okhttp3.internal.cache.DiskLruCacheTest.createNewCacheWithSize(DiskLruCacheTest.kt:84)
                at okhttp3.internal.cache.DiskLruCacheTest.createNewCache(DiskLruCacheTest.kt:76)
                at okhttp3.internal.cache.DiskLruCacheTest.setUp(DiskLruCacheTest.kt:101)
                at okhttp3.internal.cache.DiskLruCacheTest.recoverFromInitializationFailure(DiskLruCacheTest.kt:124)
```

I'll fix tonight.

swankjesse:
Excellent!

okhttp/src/main/kotlin/okhttp3/internal/cache/CacheLock.kt
yschimke: https://github.com/square/okio/issues/1464

okhttp/src/main/kotlin/okhttp3/internal/cache/CacheLock.kt
yschimke: Made an attempt.  Let me know if it's suitable.

yschimke:
Working on Windows, but tests are brittle :(

```
FileOperatorTest > write() FAILED
    java.io.IOException: Failed to delete temp directory C:\Users\RUNNER~1\AppData\Local\Temp\junit13626620294798743510. The following paths could not be deleted (see suppressed exceptions for details): <root>, test
        at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:184)
        at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
        at java.base/java.util.stream.SortedOps$RefSortingSink.end(SortedOps.java:395)
        at java.base/java.util.stream.Sink$ChainedReference.end(Sink.java:261)
        at java.base/java.util.stream.Sink$ChainedReference.end(Sink.java:261)
        at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:510)
        at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
        at java.base/java.util.stream.ForEachOps$ForEachOp.evaluateSequential(ForEachOps.java:151)
        at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.evaluateSequential(ForEachOps.java:174)
        at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
        at java.base/java.util.stream.ReferencePipeline.forEach(ReferencePipeline.java:596)
        at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
        at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
```

okhttp/src/test/java/okhttp3/CacheLockTest.kt
yschimke: I think the dominant DI libraries will all be initializing OkHttp on the main thread. You have to do extra work to avoid that.

okhttp/src/main/kotlin/okhttp3/internal/cache/CacheLock.kt
yschimke: Also, even if Cache remains unsafe on older devices, having it fail obviously on recent devices will make sure you use it safely, improving all devices.

Issue #8193, opened 2024-01-14
Switch MockWebServer extension to use Named


yschimke:
I think so, but I can switch to Named?

https://www.vogella.com/tutorials/JUnit5Extensions/article.html

_Originally posted by @yschimke in https://github.com/square/okhttp/pull/7392#discussion_r1451709802_

yschimke:
replaced with StopStart

PR #8210, opened 2024-01-20
MockWebServer Named annotations

yschimke:
https://github.com/square/okhttp/issues/8193


swankjesse (commented):
I think this is still too complicated.

What about this:
 * each test gets one MockWebServer
 * if you need multiple, just use setUp() and tearDown()

I don’t think using multiple MockWebServer instances is very common!

yschimke:
Yep, we are probably the only ones using multiple.  So it's a feature for ourselves.

That said, I find it elegant because it's no API, no change normally, but available when you need it.

yschimke:
I'll prepare a separate PR for comparison before closing this.

PR #8224, opened 2024-01-28
Tests for canonicalization

yschimke:


okhttp/src/test/java/okhttp3/internal/HostnamesTest.kt
```diff
@@ -95,4 +115,44 @@ class HostnamesTest {
...
+    ).isEqualTo("129.144.52.38")
+
+    assertThat("::FFFF:129.144.52.38".toCanonicalHost()).isEqualTo("129.144.52.38")
+
+    // Hostnames
+    assertThat("WwW.GoOgLe.cOm".toCanonicalHost()).isEqualTo("www.google.com")
+    assertThat("localhost".toCanonicalHost()).isEqualTo("localhost")
+    assertThat("☃.net".toCanonicalHost()).isEqualTo("xn--n3h.net")
+
+    // Expected?
```
swankjesse, line 152: I think so for all of these.

swankjesse (approved):

Issue #7738, opened 2023-03-07
Brotli DoS


swankjesse:
From a Bugcrowd submission ([Block-internal link](https://jira.sqprod.co/browse/VULN-69637), [Block-internal Bugcrowd link](https://tracker.bugcrowd.com/squareopensource/submissions/aebfae75-5aaf-4df9-877a-5a2d1dc01eaa)) we’ve received, OkHttp is vulnerable to a Brotli decompression bomb DoS.

Receive a request body like this one can exhaust OkHttp:
https://github.com/bones-codes/bombs/raw/master/http/10GB/10GB.html.br.bz2

yschimke:
There isn't an easy API for BrotliInputStream to enforce this, and it's hard to ask users to do it.

https://github.com/square/okhttp/blob/master/okhttp-brotli/src/main/kotlin/okhttp3/brotli/internal/brotli.kt

We could change BrotliInterceptor (deprecate and add) to an instance class, with some reasonable default maximum size, or some ratio compared to compressed.

```
object BrotliInterceptor : Interceptor {
```

Any good patterns suggested?

yschimke:
@swankjesse I'm tempted to fix in Okio via some `DecompressionLimit(ratio: Float?, limit: Long?)` type, and insert around and inside (if ratio is applied).  

Apply to GzipSource and a new BrotliSource.

Also make it possible to create BrotliInterceptor instances and pass in a non default DecompressionLimit, but put something sensible but tolerant in for the object form.

Is it worth contributing the a Brotli Source to okio? 

Envoy has some 100x ratio check https://github.com/envoyproxy/envoy/commit/d4c39e635603e2f23e1e08ddecf5a5fb5a706338#diff-88b327a1e72d55d1bb686b3b1f28f594b6b08139968304e6804a808fbb375ff0R26

Another library has max size https://github.com/python-pillow/Pillow/pull/674/files, same for golang? https://stackoverflow.com/questions/56629115/how-to-protect-service-from-gzip-bomb/56629623#56629623

yschimke:
No current fix is in place in OkHttp or Okio.  

cc @swankjesse do you want to move quickly, or accept as a risk since clients are connecting to known servers and deliberately opting into Brotli.

The current options would be
1) Stop using Brotli - it's an extra optional dependency anyway, not on by default.
2) Ensure you only use for trusted servers.

[...]

PR #8227, opened 2024-01-29
Defend against brotli decompression bombs

swankjesse:
This uses the same ratio technique as Envoy. It's a quick hack and it gets out of the way.

Closes: https://github.com/square/okhttp/issues/7738

### DIFF
--- a/okhttp-brotli/src/main/kotlin/okhttp3/brotli/internal/DecompressionBombChecker.kt
+++ b/okhttp-brotli/src/main/kotlin/okhttp3/brotli/internal/DecompressionBombChecker.kt
@@ -20,7 +20,7 @@ import okio.ForwardingSource
 import okio.IOException
 import okio.Source
 
-/** Fails decompression if the ratio is within a reasonable limit. */
+/** Fails decompression if the ratio is too high. */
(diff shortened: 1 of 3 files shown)


okhttp-brotli/src/main/kotlin/okhttp3/brotli/internal/DecompressionBombChecker.kt
```diff
@@ -0,0 +1,66 @@
...
+ * limitations under the License.
+ */
+package okhttp3.brotli.internal
+
+import okio.Buffer
+import okio.ForwardingSource
+import okio.IOException
+import okio.Source
+
+/** Fails decompression if the ratio is within a reasonable limit. */
```
yschimke, line 23: Isn't?

yschimke (approved):

okhttp-brotli/src/main/kotlin/okhttp3/brotli/internal/DecompressionBombChecker.kt
swankjesse: Fixed!

Issue #8211, opened 2024-01-20
All cacheKey override on Request.


or-else:
Signed URLs, such as https://docs.aws.amazon.com/AmazonS3/latest/userguide/using-presigned-url.html, include a transient token as a query parameter. Okhttp provides no facility to filter out transient query parameters (or, in fact, no facility at all to customize cache lookups) for the purpose of caching. Such objects are effectively never cached by okhttp. Furthermore, nearly every class in okhttp is `final` making it impossible to add such functionality by subclassing.

Javascript in browsers has this facility. You should too.

See this asked (and ignored) at stackoverflow: https://stackoverflow.com/questions/40783613/how-to-custom-cache-in-okhttp3

yschimke:
Can you link to how browsers remove the transient params?

yschimke:
As a kludgy workaround, you can use application and network interceptors to modify what gets evaluated for caching, and what gets sent on the wire.

But we should consider a proper cache key API. Or learn from browsers on transient params?

```
  data class OriginalUrl(val url: HttpUrl)

  val client = OkHttpClient.Builder().eventListener(object : EventListener() {
      override fun cacheMiss(call: Call) {
        println("cacheMiss ${call.request().url}")
      }

      override fun cacheHit(call: Call, response: Response) {
        println("cacheHit ${call.request().url}")
      }

      override fun cacheConditionalHit(call: Call, cachedResponse: Response) {
        println("cacheConditionalHit ${call.request().url}")
      }
    }).addInterceptor {
      val request = it.request()
      val url = request.url
      // only strip the key param for caching, not other params
      val newRequest = if (url.queryParameter("key") != null) {
        request.newBuilder().tag<OriginalUrl>(OriginalUrl(url)).url(url.newBuilder().removeAllQueryParameters("key").build()).build()
      } else {
        request
      }
      it.proceed(newRequest).newBuilder().request(request).build()
    }.addNetworkInterceptor {
      val request = it.request()
      val originalUrl = request.tag<OriginalUrl>()?.url
      // if we modified the url for caching, send the real one
      if (originalUrl != null) {
        val networkRequest = request.newBuilder().url(originalUrl).build()
        it.proceed(networkRequest).newBuilder().request(request).build()
      } else {
        it.proceed(request)
      }
    }.cache(Cache("/cache".toPath(), Long.MAX_VALUE, FakeFileSystem())).build()

  // cache miss - first request
[...]

Issue #8113, opened 2023-11-24
Cache support for DnsOverHttps.Builder().post(true)


calbot:
The Cache class key uses the request url and therefore cannot cache DNS over HTTPS requests when the DNS query address is sent in a Post request with `DnsOverHttps.Builder().post(true)`. When the get form of the DNS request is used it is base64 encoded so it isn't really private which domain is being requested so it would be good to be able to have both privacy of DoH and caching.

yschimke:
@swankjesse I can't see a good way to handle this. We can do a complicated interceptor dance to fool caching.  But I think I'm preferring some internal Tag for CacheKey?

swankjesse:
@yschimke that’s a good idea. Or this?

```
class RequestBody {
  ...
  /** 
   * Returns a non-null key if this request can be cached. If non-null,
   * this key is used in combination to the request URL to create the
   * ultimate cache key.
   */
  open fun cacheKey(): ByteString? {
    return null;
  }
}
```

yschimke:
Yeah that works. And is more constrained. The tag approach would allow mucking with any cache key

yschimke:
A similar requirement...

https://github.com/square/okhttp/issues/8211

But wouldn't work with a RequestBody.

PR #8234, opened 2024-02-03
Cache URL Override

yschimke:
A minimal change to allow a cacheUrlOverride on request for the following examples

- Define the canonical resource this request represents, such as the origin server for a CDN request
- Remove transient query params such as S3 tokens
- A narrow window of POST support - by declaring a POST should otherwise be treated as a GET, and define an appropriate unique url to represent the body if relevant.

It is implemented almost entirely within CacheInterceptor, mutating the request/response that Cache sees to avoid a broader change such as supporting POSTs.

Addresses https://github.com/square/okhttp/issues/8211 also https://github.com/square/okhttp/issues/8113

### DIFF
--- a/okhttp-dnsoverhttps/src/main/kotlin/okhttp3/dnsoverhttps/DnsOverHttps.kt
+++ b/okhttp-dnsoverhttps/src/main/kotlin/okhttp3/dnsoverhttps/DnsOverHttps.kt
@@ -186,16 +186,24 @@ class DnsOverHttps internal constructor(
     throw unknownHostException
   }
 
-  private fun getCacheOnlyResponse(request: Request): Response? {
-    if (!post && client.cache != null) {
+  private fun getCacheOnlyResponse(
+    request: Request,
+  ): Response? {
+    if (client.cache != null) {
       try {
         // Use the cache without hitting the network first
         // 504 code indicates that the Cache is stale
-        val preferCache =
+        val onlyIfCached =
           CacheControl.Builder()
             .onlyIfCached()
             .build()
-        val cacheRequest = request.newBuilder().cacheControl(preferCache).build()
+
+        var cacheUrl = request.url
+
+        val cacheRequest = request.newBuilder()
+          .cacheControl(onlyIfCached)
+          .cacheUrlOverride(cacheUrl)
+          .build()
 
         val cacheResponse = client.newCall(cacheRequest).execute()
 
@@ -247,7 +255,12 @@ class DnsOverHttps internal constructor(
       val query = DnsRecordCodec.encodeQuery(hostname, type)
 
       if (post) {
-        url(url).post(query.toRequestBody(DNS_MESSAGE))
+        url(url)
+          .cacheUrlOverride(
+            url.newBuilder()
+              .addQueryParameter("hostname", hostname).build()
+          )
+          .post(query.toRequestBody(DNS_MESSAGE))
       } else {
         val encoded = query.base64Url().replace("=", "")
         val requestUrl = url.newBuilder().addQueryParameter("dns", encoded).build()
@@ -313,7 +326,8 @@ class DnsOverHttps internal constructor(
         this.bootstrapDnsHosts = bootstrapDnsHosts
       }
 
-    fun bootstrapDnsHosts(vararg bootstrapDnsHosts: InetAddress): Builder = bootstrapDnsHosts(bootstrapDnsHosts.toList())
+    fun bootstrapDnsHosts(vararg bootstrapDnsHosts: InetAddress): Builder =
+      bootstrapDnsHosts(bootstrapDnsHosts.toList())
 
     fun systemDns(systemDns: Dns) =
       apply {
--- a/okhttp-dnsoverhttps/src/main/kotlin/okhttp3/dnsoverhttps/DnsOverHttps.kt
+++ b/okhttp-dnsoverhttps/src/main/kotlin/okhttp3/dnsoverhttps/DnsOverHttps.kt
@@ -186,9 +186,7 @@ class DnsOverHttps internal constructor(
     throw unknownHostException
   }
 
-  private fun getCacheOnlyResponse(
-    request: Request,
-  ): Response? {
+  private fun getCacheOnlyResponse(request: Request): Response? {
     if (client.cache != null) {
       try {
         // Use the cache without hitting the network first
@@ -200,10 +198,11 @@ class DnsOverHttps internal constructor(
 
         var cacheUrl = request.url
 
-        val cacheRequest = request.newBuilder()
-          .cacheControl(onlyIfCached)
-          .cacheUrlOverride(cacheUrl)
-          .build()
+        val cacheRequest =
+          request.newBuilder()
+            .cacheControl(onlyIfCached)
+            .cacheUrlOverride(cacheUrl)
+            .build()
 
         val cacheResponse = client.newCall(cacheRequest).execute()
 
@@ -258,7 +257,7 @@ class DnsOverHttps internal constructor(
         url(url)
           .cacheUrlOverride(
             url.newBuilder()
-              .addQueryParameter("hostname", hostname).build()
+              .addQueryParameter("hostname", hostname).build(),
           )
           .post(query.toRequestBody(DNS_MESSAGE))
       } else {
@@ -326,8 +325,7 @@ class DnsOverHttps internal constructor(
         this.bootstrapDnsHosts = bootstrapDnsHosts
       }
 
-    fun bootstrapDnsHosts(vararg bootstrapDnsHosts: InetAddress): Builder =
-      bootstrapDnsHosts(bootstrapDnsHosts.toList())
+    fun bootstrapDnsHosts(vararg bootstrapDnsHosts: InetAddress): Builder = bootstrapDnsHosts(bootstrapDnsHosts.toList())
 
     fun systemDns(systemDns: Dns) =
       apply {
--- a/okhttp-dnsoverhttps/src/test/java/okhttp3/dnsoverhttps/DnsOverHttpsTest.kt
+++ b/okhttp-dnsoverhttps/src/test/java/okhttp3/dnsoverhttps/DnsOverHttpsTest.kt
@@ -31,9 +31,12 @@ import java.util.concurrent.TimeUnit
 import mockwebserver3.MockResponse
 import mockwebserver3.MockWebServer
 import okhttp3.Cache
+import okhttp3.Call
 import okhttp3.Dns
+import okhttp3.EventListener
 import okhttp3.OkHttpClient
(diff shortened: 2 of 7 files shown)


okhttp-dnsoverhttps/src/main/kotlin/okhttp3/dnsoverhttps/DnsOverHttps.kt
```diff
@@ -187,15 +187,22 @@ class DnsOverHttps internal constructor(
...
         // Use the cache without hitting the network first
         // 504 code indicates that the Cache is stale
-        val preferCache =
+        val onlyIfCached =
           CacheControl.Builder()
             .onlyIfCached()
             .build()
-        val cacheRequest = request.newBuilder().cacheControl(preferCache).build()
+
+        var cacheUrl = request.url
```
swankjesse, line 199: If its a POST does this URL still include the domain name?

okhttp-dnsoverhttps/src/main/kotlin/okhttp3/dnsoverhttps/DnsOverHttps.kt
```diff
@@ -247,7 +254,12 @@ class DnsOverHttps internal constructor(
       val query = DnsRecordCodec.encodeQuery(hostname, type)
 
       if (post) {
-        url(url).post(query.toRequestBody(DNS_MESSAGE))
+        url(url)
+          .cacheUrlOverride(
+            url.newBuilder()
+              .addQueryParameter("hostname", hostname).build(),
```
swankjesse, line 260: Ooooh it does now!

swankjesse: This is supercool

okhttp-dnsoverhttps/src/test/java/okhttp3/dnsoverhttps/DnsOverHttpsTest.kt
```diff
@@ -167,30 +168,79 @@ class DnsOverHttpsTest {
   // 3. successful network response
   // 4. successful stale cached GET response
   // 5. unsuccessful response
-  // TODO how closely to follow POST rules on caching?
```
swankjesse, line 170: YAY

okhttp-dnsoverhttps/src/test/java/okhttp3/dnsoverhttps/DnsOverHttpsTest.kt
```diff
@@ -167,30 +168,79 @@ class DnsOverHttpsTest {
...
+          .build(),
+      )
+    }
+
+    var result = cachedDns.lookup("google.com")
+    assertThat(result).containsExactly(address("157.240.1.18"))
+    var recordedRequest = server.takeRequest()
+    assertThat(recordedRequest.method).isEqualTo("POST")
+    assertThat(recordedRequest.path)
+      .isEqualTo("/lookup?ct")
```
swankjesse, line 232: Off-topic: we’re naughty for having a field called `path` that contains a path and a query. Thanks for confirming that we use the right URL for these!

okhttp/src/main/kotlin/okhttp3/Request.kt
```diff
@@ -316,6 +321,17 @@ class Request internal constructor(builder: Builder) {
       tag: T?,
     ) = commonTag(type.kotlin, tag)
 
+    /**
+     * Override the [Request.url] for caching, if it is either polluted with
+     * transient query params, or has a canonical URL possibly for a CDN.
+     *
+     * If set, this will also allow caching for POST requests.
```
swankjesse, line 328: # 💯

swankjesse:
```suggestion
     * Note that POST requests will not be sent to the server if this URL is set
     * and matches a cached response.
```

okhttp/src/main/kotlin/okhttp3/internal/cache/CacheInterceptor.kt
```diff
@@ -285,3 +288,17 @@ class CacheInterceptor(internal val cache: Cache?) : Interceptor {
...
 }
+
+private fun Request.requestForCache(): Request {
+  val cacheUrlOverride = cacheUrlOverride
+
+  return if (cacheUrlOverride != null && (method == "GET" || method == "POST")) {
+    newBuilder()
+      .get()
+      .url(cacheUrlOverride)
+      .cacheUrlOverride(null)
```
swankjesse, line 299: nice

okhttp/src/test/java/okhttp3/CacheTest.kt
```diff
@@ -1002,11 +1002,22 @@ class CacheTest {
     testRequestMethod("POST", false)
   }
 
+  @Test
+  fun requestMethodPostIsNotCachedUnlessOverriden() {
```
swankjesse, line 1006:
```suggestion
  fun requestMethodPostIsNotCachedUnlessOverridden() {
```

okhttp/src/test/java/okhttp3/CacheTest.kt
```diff
@@ -3250,6 +3267,48 @@ CLEAN $urlKey ${entryMetadata.length} ${entryBody.length}
     )
   }
 
+  @Test
+  fun getHasCorrectResponseIsCorrect() {
```
swankjesse, line 3271:
```suggestion
  fun getHasCorrectResponse() {
```

okhttp/src/test/java/okhttp3/CacheTest.kt
```diff
@@ -3250,6 +3267,48 @@ CLEAN $urlKey ${entryMetadata.length} ${entryBody.length}
...
+    val request = Request(server.url("/abc"))
+
+    val response = testBasicCachingRules(request)
+
+    assertThat(response.request.url).isEqualTo(request.url)
+    assertThat(response.cacheResponse!!.request.url).isEqualTo(request.url)
+  }
+
+  @Test
+  fun postWithOverrideResponseIsCorrect() {
```
swankjesse, line 3281:
```suggestion
  fun postWithOverrideResponse() {
```

swankjesse (approved):
This is a great feature. Very small impact on the public API, very big impact on what people can do with OkHttp.

okhttp/api/okhttp.api
```diff
@@ -1048,6 +1049,7 @@ public class okhttp3/Request$Builder {
 	public fun addHeader (Ljava/lang/String;Ljava/lang/String;)Lokhttp3/Request$Builder;
 	public fun build ()Lokhttp3/Request;
 	public fun cacheControl (Lokhttp3/CacheControl;)Lokhttp3/Request$Builder;
+	public final fun cacheUrlOverride (Lokhttp3/HttpUrl;)Lokhttp3/Request$Builder;
```
swankjesse, line 1052: Note that we’re not adding `OkHttpExperimentalApi` on this, and so we’re committing to this exact API and name. I’m okay with that.

yschimke: Should we reconsider as a policy for new API?

okhttp-dnsoverhttps/src/main/kotlin/okhttp3/dnsoverhttps/DnsOverHttps.kt
yschimke: I'll take a look.

Endeavour233: > I'll take a look.

Thank you so much.

Endeavour233: > I'll take a look.

Any update? Thanks

yschimke: You are right about the cache miss in that method.

That method is there, and works for get, in order to avoid a conditional get when we already have a result, delaying resolution.

I'm tempted to live with that performance regression, but open for discussions. So that PR simplifies the code to remove that.

https://github.com/square/okhttp/pull/8383

yschimke: The tests probably won't fail because of the change, unless I calculate a delay past the heuristic cache checks, so I should follow up with an additional test in that PR.

Endeavour233: > You are right about the cache miss in `getCacheOnlyResponse`.
> 
> The `getCacheOnlyResponse` method is there, and works correctly for get, in order to avoid a conditional network request when we already have a result, delaying resolution. I'll double check this from PR history.
> 
> I'm tempted to live with that performance regression, but open for discussions. So this PR simplifies the code to remove that.
> 
> #8383

Thanks for your explanation. However, I don't think `getCacheOnlyResponse` could help avoiding ‘unnecessary’ conditional request. Firstly, if `CacheStrategy` figures out that we need a conditional request to verify that our cache is still available, I think we should send one. And, what a `onlyIfCached` CacheControl can do is to enforce a response with code HTTP_GATEWAY_TIMEOUT when a conditional request(or any network request) is wanted,(see `CacheStrategy#compute()`) so that we would still abandon the cache and make a request given current implementation.

yschimke: > Firstly, if CacheStrategy figures out that we need a conditional request to verify that our cache is still available, I think we should send one

Yes, that's the conclusion I think I came to. Please leave comments on that PR if there is anything to change further.

BTW thanks for the really thorough analysis here, while it was technically working, it's much simpler code now, and I obviously didn't understand the flow as well as I thought I did. But digging in like you have to find issues and improve the logic, is really valuable. So, thanks.

Endeavour233: > > Firstly, if CacheStrategy figures out that we need a conditional request to verify that our cache is still available, I think we should send one
> 
> Yes, that's the conclusion I think I came to. Please leave comments on that PR if there is anything to change further.
> 
> BTW thanks for the really thorough analysis here, while it was technically working, it's much simpler code now, and I obviously didn't understand the flow as well as I thought I did. But digging in like you have to find issues and improve the logic, is really valuable. So, thanks.

Off the topic: thanks for your contribution to okhttp. And you are my role model. Your enthusiasm in programming, pretty readable code with high performance inspires me a lot. Thank you.

PR #8240, opened 2024-02-05
fix:HttpLoggingInterceptor bodyIsStreaming add other stream type assert

Felix8029:


yschimke:
I don't think application/octet-stream indicates a streaming response.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types

> Generic binary data (or binary data whose true type is unknown) is application/octet-stream.

Are you trying to just not output binary data here?

I guess you could add to `bodyHasUnknownEncoding` or add a `bodyHasBinaryData`. 

Could you start with what you are trying to achieve?

JakeWharton:
It also might not be binary data, just data of unknown mime type. Probably best to rely on the regular binary detection still.

yschimke:
Closing, happy to discuss if I've misunderstood something.

Issue #7476, opened 2022-10-14
Provide way on HttpLoggingInterceptor to redact query params


DavidRigglemanININ:
Currently, support is available for redacting headers on the logging interceptor. The docs even warn that you could be leaking sensitive info if at the level of BODY or HEADERS. However, even BASIC has the potential to leak sensitive info if information is encoded in the url via query params (AWS is one example of where this can happen). So it'd be great if there was an easy way to specify a query param that should be redacted from logging. Thanks for the consideration!

yschimke:
https://github.com/square/okhttp/pull/7458

DavidRigglemanININ:
Unless I'm missing something, the issue you linked is about headers, not query params, which is my issue. Or maybe you were just linking an issue to show how it was done for headers?

yschimke:
You are right, I was getting confused.  Probably worth discussing them together, even if we implement separately or at all.

yschimke:
OK, just summarising what we have now

`internal.isSensitiveHeader()` - checks for known headers like Authorization.

Request PR https://github.com/square/okhttp/pull/7458 - adding check for `isSensitiveHeader`

`HttpLoggingInterceptor`
`redactHeader()` - adds a new header to be redacted.

I think we should treat this seriously, but I don't think we can know the full set of query params or even headers to redact.

I'd suggest we add a redacted for of Request.toString(redactedHeaders, redactedParams) that is public API, with the current defaults.  And make this configurable in HttpLoggingInterceptor.

@swankjesse thoughts?

swankjesse:
If I had to implement this, I think I’d want a `UrlRedacter` interface that converted a URL to a String.

But I continue to believe that HttpLoggingInterceptor is not designed for production use. Speculative logging is too noisy. For development I love Chucker; for production I’d like something that does spans like Datadog or Chronosphere.

yschimke:
Closing as low priority, HttpLoggingInterceptor can be copy and pasted and is not part of OkHttp core.

PR #8242, opened 2024-02-06
Hide the value of sensitive query parameters in log

MinhDang685:
For logging interceptor: add an option to hide the value of sensitive query parameters in the request and response URL. 
Inspired by the implementation of header redaction in `okhttp3.logging.HttpLoggingInterceptor`

**To use**
Call the method `redactQueryParams(the list of query parameters to hide value, separated by commas)` from a `okhttp3.logging.HttpLoggingInterceptor` instance. For example, to hide the value of "user" and "password" query parameters, use `interceptor.redactQueryParams("user", "password")`.
- Original URL: `http://localhost:8000/api/login?user=user1&authentication=basic&password=confidential_password"`
- With the method applied: `http://localhost:8000/api/login?user=██&authentication=basic&password=██"`

### DIFF
--- a/okhttp-logging-interceptor/src/main/kotlin/okhttp3/logging/HttpLoggingInterceptor.kt
+++ b/okhttp-logging-interceptor/src/main/kotlin/okhttp3/logging/HttpLoggingInterceptor.kt
@@ -323,6 +323,9 @@ class HttpLoggingInterceptor
     }
 
     private fun sanitizeUrl(url: HttpUrl): String {
+      if (queryParamsNameToRedact.isEmpty()) {
+        return url.toString()
+      }
       val params = ArrayList<Pair<String, String>>()
       for (parameterName in url.queryParameterNames) {
         params.add(
--- a/okhttp-logging-interceptor/src/main/kotlin/okhttp3/logging/HttpLoggingInterceptor.kt
+++ b/okhttp-logging-interceptor/src/main/kotlin/okhttp3/logging/HttpLoggingInterceptor.kt
@@ -22,6 +22,7 @@ import java.nio.charset.Charset
 import java.util.TreeSet
 import java.util.concurrent.TimeUnit
 import okhttp3.Headers
+import okhttp3.HttpUrl
 import okhttp3.Interceptor
 import okhttp3.OkHttpClient
 import okhttp3.Response
@@ -45,6 +46,8 @@ class HttpLoggingInterceptor
     private val logger: Logger = Logger.DEFAULT,
   ) : Interceptor {
     @Volatile private var headersToRedact = emptySet<String>()
+
+    @Volatile private var queryParamsNameToRedact = emptySet<String>()
 
     @set:JvmName("level")
     @Volatile
@@ -132,6 +135,13 @@ class HttpLoggingInterceptor
       headersToRedact = newHeadersToRedact
     }
 
+    fun redactQueryParams(vararg name: String) {
+      val newQueryParamsNameToRedact = TreeSet(String.CASE_INSENSITIVE_ORDER)
+      newQueryParamsNameToRedact += queryParamsNameToRedact
+      newQueryParamsNameToRedact.addAll(name)
+      queryParamsNameToRedact = newQueryParamsNameToRedact
+    }
+
     /**
      * Sets the level and returns this.
      *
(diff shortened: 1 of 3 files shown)


yschimke:
This is against the resolution of https://github.com/square/okhttp/issues/7476

> But I continue to believe that HttpLoggingInterceptor is not designed for production use. Speculative logging is too noisy. For development I love Chucker; for production I’d like something that does spans like Datadog or Chronosphere.

As it's a functional PR, I'm inclined to accept it.

Will wait for a final take by @swankjesse

yschimke (approved):

okhttp-logging-interceptor/src/test/java/okhttp3/logging/HttpLoggingInterceptorTest.kt
```diff
@@ -903,6 +903,51 @@ class HttpLoggingInterceptorTest {
       .assertNoMoreLogs()
   }
 
+  @Test
+  fun sensitiveQueryParamsAreRedacted() {
```
yschimke, line 907: Can you add a test for duplicate query params, and also retaining the order.

okhttp-logging-interceptor/src/main/kotlin/okhttp3/logging/HttpLoggingInterceptor.kt
```diff
@@ -312,6 +322,32 @@ class HttpLoggingInterceptor
       return response
     }
 
+    private fun sanitizeUrl(url: HttpUrl): String {
```
yschimke, line 325: Actually, I think this needs rework.

queryParameterNames is a unique Set.

Maybe something like

```
    private fun sanitizeUrl(url: HttpUrl): String {
      if (queryParamsNameToRedact.isEmpty() || url.querySize == 0) {
        return url.toString()
      }

      val newUrl = url.newBuilder().query(null).apply {
        for (i in 0 until url.querySize) {
          val parameterName = url.queryParameterName(i)
          val newValue = if (parameterName in queryParamsNameToRedact) "██" else url.queryParameterValue(i)

          addEncodedQueryParameter(parameterName, newValue)
        }
      }.toString()
    }
```

MinhDang685: Thanks for your suggestion; this one is definitely better than my string concatenation solution.

okhttp-logging-interceptor/src/main/kotlin/okhttp3/logging/HttpLoggingInterceptor.kt
```diff
@@ -312,6 +322,20 @@ class HttpLoggingInterceptor
       return response
     }
 
+    internal fun sanitizeUrl(url: HttpUrl): String {
```
swankjesse, line 325: Rename to redactUrl please

okhttp-logging-interceptor/src/test/java/okhttp3/logging/HttpLoggingInterceptorTest.kt
```diff
@@ -903,6 +903,99 @@ class HttpLoggingInterceptorTest {
...
+        .newCall(
+          request()
+            .build(),
+        )
+        .execute()
+    response.body.close()
+    val sanitizedUrl = networkInterceptor.sanitizeUrl(url)
+    val sanitizedUrlPattern = sanitizedUrl.replace("?", """\?""")
+    applicationLogs
+      .assertLogEqual("--> GET $sanitizedUrl")
```
swankjesse, line 990: Please use the word `redactedUrl` here.

okhttp-logging-interceptor/src/test/java/okhttp3/logging/HttpLoggingInterceptorTest.kt
```diff
@@ -903,6 +903,99 @@ class HttpLoggingInterceptorTest {
...
+      .assertLogMatch(Regex("""<-- 200 OK $sanitizedUrlPattern \(\d+ms, \d+-byte body\)"""))
+      .assertNoMoreLogs()
+    networkLogs
+      .assertLogEqual("--> GET $sanitizedUrl http/1.1")
+      .assertLogMatch(Regex("""<-- 200 OK $sanitizedUrlPattern \(\d+ms, \d+-byte body\)"""))
+      .assertNoMoreLogs()
+  }
+
+  @Test
+  fun preserveQueryParamsAfterSanitized() {
```
swankjesse, line 951: Please use the word `Redacted` instead of `Sanitzed` throughout!

swankjesse (approved):
Works for me. I’ve got some naming changes I’d like you to make before I merge it!

okhttp-logging-interceptor/src/test/java/okhttp3/logging/HttpLoggingInterceptorTest.kt
```diff
@@ -903,6 +903,101 @@ class HttpLoggingInterceptorTest {
       .assertNoMoreLogs()
   }
 
+  @Suppress("INVISIBLE_MEMBER", "INVISIBLE_REFERENCE")
```
yschimke, line 906: @MinhDang685 we run these tests in android also, which means they need to workaround internal methods. So I added this.

Issue #8231, opened 2024-01-31
Should SSLContext construction be deferred during OkHttpClient construction


yschimke:
https://www.zacsweers.dev/dagger-party-tricks-deferred-okhttp-init/

> OkHttpClient initialization can take upwards of 100ms on some Android devices in the wild due to its use of TrustManagerFactory. Initializing the Cache is also not free, as it creates an internal Executor and may incur some disk IO if you want to prepare a cache directory for it first.

swankjesse:
Is the best solution to make OkHttp lazy internally? Or to encourage our users to instantiate their OkHttp clients lazily? I think making OkHttp lazy potentially introduces other problems by deferring failures

yschimke:
I'm not sure. But it's 

A) a real problem
B) the default behaviour and common.
C) something we put on every developer to do if we don't.

yschimke:
@swankjesse what would this look like with https://github.com/square/okhttp/pull/7846

I'm not sure it solves anything. You should be able to access the OkHttpClient on main and submit requests using a safe suspend method. But this won't currently be the case.

yschimke:
Should see if it's possible after https://github.com/square/okhttp/pull/8800 lands

PR #8248, opened 2024-02-10
Avoid expensive SSL init on main

yschimke:
Fixes https://github.com/square/okhttp/issues/8231

Avoids expensive work as seen in traces here https://github.com/coil-kt/coil/issues/2097

### DIFF
--- a/okhttp/src/main/kotlin/okhttp3/OkHttpClient.kt
+++ b/okhttp/src/main/kotlin/okhttp3/OkHttpClient.kt
@@ -205,14 +205,15 @@ open class OkHttpClient internal constructor(
   @get:JvmName("socketFactory")
   val socketFactory: SocketFactory = builder.socketFactory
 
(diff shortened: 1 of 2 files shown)


yschimke:
Should land (or not) after https://github.com/square/okhttp/pull/8339

yschimke:
Unfortunately https://github.com/square/okhttp/pull/8800 can't work to solve this as it would still initialise the OkHttpClient instance

yschimke:
@swankjesse I guess I feel pretty strongly that OkHttp is doing the wrong thing. If you can't call `OkHttpClient.Builder().build()` on main thread, then it's really hard to use correctly.

Options

1) I could make a deferring `Call.Factory` but that forces people to adopt, and also doesn't work for libraries taking OkHttpClient directly.

2) Make Android and JVM expect/actual so we can handle differently with Android _deferring when created on Main_.
2b) as above but optional, perhaps with a  https://github.com/square/okhttp/pull/8800

Issue #8171, opened 2024-01-02
Constructor Injection does not work with mockwebserver3-junit5


snv:
While the [readme](https://github.com/square/okhttp/blob/master/mockwebserver-junit5/README.md) claims that constructor injection should work, and offers an example, this in fact does not work (and is not covered by any test [in the project itself](https://github.com/square/okhttp/tree/master/mockwebserver-junit5/src/test/java/mockwebserver3/junit5/internal)).

Following sample-test
```java
import mockwebserver3.MockWebServer;
import mockwebserver3.junit5.internal.MockWebServerExtension;
import org.junit.jupiter.api.Test;
import org.junit.jupiter.api.extension.ExtendWith;

@ExtendWith(MockWebServerExtension.class)
public class MockWebServerInjectionTest {
  private final MockWebServer server;

  MockWebServerInjectionTest(MockWebServer server) {
    this.server = server;
  }

  @Test
  void server_is_set(){
    assert server != null;
  }
}
```
with `com.squareup.okhttp3:mockwebserver3-junit5:5.0.0-alpha.12`
and `org.junit.jupiter:junit-jupiter:5.9.0`
results in
`org.junit.jupiter.api.extension.ParameterResolutionException: No ParameterResolver registered for parameter [mockwebserver3.MockWebServer arg0] in constructor [...MockWebServerInjectionTest(mockwebserver3.MockWebServer)].`

I also tried to use the annotation `@MockWebServerInstance` for field injection into the test class without success.

I assume the `mockwebserver3.junit5.internal.MockWebServerExtension` is a bit too restrictive in its selection of injection-points.

For context:
I can not simply inject into the test-method itself, because i need to use the mockwebserver-instance inside a test-class-local producer-method for a CDI-Bean with [WeldJunit5](https://github.com/weld/weld-testing/tree/master/junit5)

yschimke:
I'll update the docs.

The issue is the model of Junit 5 parameters is different for constructors (class) and test and lifecycle methods (class/test method).

It's simplest to only support the latter to avoid confusion of people using the API correctly, but getting weird results.

This log won't make sense, but just recording for my notes.

```
resource id  [engine:junit-jupiter]/[ExtensionLifecycleTest]
init 
resource id  [engine:junit-jupiter]/[ExtensionLifecycleTest]/[method:testClient2(mockwebserver3.MockWebServer)]
[...]

PR #8249, opened 2024-02-10
More typical Junit5 extension scopes

yschimke:
A second attempt at https://github.com/square/okhttp/issues/8171

Now following examples from TestInfo more strictly.

There are two classes of instances

a) constructor, before all, after all
b) test method, before each, before all


yschimke:

I think this might affect flakyness of some tests, as the scope of parameterised tests wasn't right.

Landing and happy to revert it fix forward.

swankjesse (commented):
LGTM

YES!

mockwebserver-junit5/src/main/kotlin/mockwebserver3/junit5/internal/MockWebServerExtension.kt
```diff
@@ -43,27 +43,12 @@ import org.junit.jupiter.api.extension.ParameterResolver
...
-        }
-
-      return store.serversForTest
-    }
-
-  private val ExtensionContext.Store.serversForTest: ServersForTest
-    get() {
-      return getOrComputeIfAbsent(storeKey) {
+    get() =
+      getStore(namespace).getOrComputeIfAbsent(this.uniqueId) {
```
swankjesse, line 49: I like this a lot more

PR #8263, opened 2024-02-25
Create devcontainer.json

CharliesAngel1:


yschimke:
This need some context and discussion.

So it's this https://code.visualstudio.com/docs/devcontainers/create-dev-container

Why is this the right config for OkHttp, should it have some gradle configuration?

Can you point to similar projects that have this config?

yschimke:
Why is it this and not https://github.com/jenkins-infra/plugin-health-scoring/blob/devcontainer/.devcontainer%2Fdevcontainer.json

PR #8276, opened 2024-03-04
Extract an interface for a ConnectionUser

swankjesse:
This is working towards supporting pre-warmed connections in ConnectionPool. We want to create connections without a Call object.


okhttp/src/main/kotlin/okhttp3/internal/connection/RealConnectionPool.kt
```diff
@@ -71,7 +71,7 @@ class RealConnectionPool(
   }
 
   /**
-   * Attempts to acquire a recycled connection to [address] for [call]. Returns the connection if it
+   * Attempts to acquire a recycled connection to [address] for [connectionUser]. Returns the connection if it
    * was acquired, or null if no connection was acquired. The acquired connection will also be
    * assigned to [RealCall.connection].
```
yschimke, line 76: nit: doesn't always hold. This is through `connectionUser.acquireConnectionNoEvents()` which may no longer be a call.

yschimke:
I can see where it's going. But it's hard to reason about the interface without the new usage.

ean5533:
@yschimke Here's an incomplete example of what we're trying to accomplish, which is enabled by this refactor: https://github.com/square/okhttp/commit/63a69885808f8ff75e7ce351e4dd1e6df1854ec5

okhttp/src/main/kotlin/okhttp3/internal/connection/CallConnectionUser.kt
```diff
@@ -0,0 +1,140 @@
...
+import okhttp3.Protocol
+import okhttp3.Route
+import okhttp3.internal.http.RealInterceptorChain
+
+/**
+ * A connection user that is a specific [RealCall].
+ */
+internal class CallConnectionUser(
+  private val call: RealCall,
+  private val poolConnectionListener: ConnectionListener,
```
yschimke, line 21: Should the concrete instances on this interface be responsible for notifying the connection pool?

Seems like logic should match regardless of the impl.

Maybe pull out or wrap with a poolconnectionuser?

#tioli

PR #8287, opened 2024-03-15
Support minimum connection pool size

ean5533:
This PR adds support for configuring the connection pool to proactively create connections to specified hosts. This is useful for "prewarming" the connection pool before it starts being used.

The main public API change is adding `ConnectionPool.setPolicy(Address, AddressPolicy)`. A policy specifies how many _concurrent streams_ should be supported by the pool (which might be satisfied by a single http/2 connection or by N http/1.1 connections).

The main internal change is adding a new task queue to `RealConnectionPool` which checks to see if it needs to create new connections to satisfy existing policies. This task is enqueued any time a policy changes, a connection is closed, or an http/2 connection's settings decrease the number of concurrent streams supported.

### DIFF
--- a/okhttp/src/main/kotlin/okhttp3/ConnectionPool.kt
+++ b/okhttp/src/main/kotlin/okhttp3/ConnectionPool.kt
@@ -110,6 +110,7 @@ class ConnectionPool internal constructor(
     delegate.evictAll()
   }
 
+  @ExperimentalOkHttpApi
   fun setPolicy(
     address: Address,
     policy: AddressPolicy,
--- a/okhttp/src/main/kotlin/okhttp3/ConnectionPool.kt
+++ b/okhttp/src/main/kotlin/okhttp3/ConnectionPool.kt
@@ -92,4 +92,25 @@ class ConnectionPool internal constructor(
   fun evictAll() {
     delegate.evictAll()
   }
+
+  fun setPolicy(
+    address: Address,
+    policy: AddressPolicy,
+  ) {
+    delegate.setPolicy(address, policy)
+  }
+
+  /**
+   * A policy for how the pool should treat a specific address.
+   *
+   * @param minimumConcurrentCalls How many concurrent calls should be possible to make at any time.
+   *    The pool will passively try to pre-emptively open connections to satisfy this minimum.
+   * @param backoffDelayMillis How long to wait to retry pre-emptive connection attempts that fail.
+   * @param backoffJitterMillis How much jitter to introduce in connection retry backoff delays
+   */
+  data class AddressPolicy(
+    val minimumConcurrentCalls: Int,
+    val backoffDelayMillis: Long = 60 * 1000,
+    val backoffJitterMillis: Int = 100,
+  )
 }
--- a/okhttp/src/main/kotlin/okhttp3/ConnectionPool.kt
+++ b/okhttp/src/main/kotlin/okhttp3/ConnectionPool.kt
@@ -19,6 +19,7 @@ package okhttp3
 import java.util.concurrent.TimeUnit
 import okhttp3.internal.concurrent.TaskRunner
 import okhttp3.internal.connection.RealConnectionPool
+import okhttp3.internal.connection.RouteDatabase
 
 /**
  * Manages reuse of HTTP and HTTP/2 connections for reduced network latency. HTTP requests that
@@ -39,6 +40,14 @@ class ConnectionPool internal constructor(
     timeUnit: TimeUnit = TimeUnit.MINUTES,
     taskRunner: TaskRunner = TaskRunner.INSTANCE,
     connectionListener: ConnectionListener = ConnectionListener.NONE,
+    readTimeoutMillis: Int = 10_000,
+    writeTimeoutMillis: Int = 10_000,
+    socketConnectTimeoutMillis: Int = 10_000,
+    socketReadTimeoutMillis: Int = 10_000,
+    pingIntervalMillis: Int = 10_000,
+    retryOnConnectionFailure: Boolean = true,
+    fastFallback: Boolean = true,
+    routeDatabase: RouteDatabase = RouteDatabase(),
   ) : this(
     RealConnectionPool(
       taskRunner = taskRunner,
@@ -46,6 +55,14 @@ class ConnectionPool internal constructor(
       keepAliveDuration = keepAliveDuration,
       timeUnit = timeUnit,
       connectionListener = connectionListener,
+      readTimeoutMillis = readTimeoutMillis,
+      writeTimeoutMillis = writeTimeoutMillis,
+      socketConnectTimeoutMillis = socketConnectTimeoutMillis,
+      socketReadTimeoutMillis = socketReadTimeoutMillis,
+      pingIntervalMillis = pingIntervalMillis,
+      retryOnConnectionFailure = retryOnConnectionFailure,
+      fastFallback = fastFallback,
+      routeDatabase = routeDatabase,
     ),
   )
 
--- a/okhttp/src/main/kotlin/okhttp3/ConnectionPool.kt
+++ b/okhttp/src/main/kotlin/okhttp3/ConnectionPool.kt
@@ -18,7 +18,11 @@ package okhttp3
 
 import java.util.concurrent.TimeUnit
 import okhttp3.internal.concurrent.TaskRunner
+import okhttp3.internal.connection.FastFallbackExchangeFinder
+import okhttp3.internal.connection.ForceConnectRoutePlanner
+import okhttp3.internal.connection.PoolConnectionUser
 import okhttp3.internal.connection.RealConnectionPool
+import okhttp3.internal.connection.RealRoutePlanner
 import okhttp3.internal.connection.RouteDatabase
 
 /**
@@ -55,14 +59,28 @@ class ConnectionPool internal constructor(
       keepAliveDuration = keepAliveDuration,
       timeUnit = timeUnit,
       connectionListener = connectionListener,
-      readTimeoutMillis = readTimeoutMillis,
-      writeTimeoutMillis = writeTimeoutMillis,
-      socketConnectTimeoutMillis = socketConnectTimeoutMillis,
-      socketReadTimeoutMillis = socketReadTimeoutMillis,
-      pingIntervalMillis = pingIntervalMillis,
-      retryOnConnectionFailure = retryOnConnectionFailure,
-      fastFallback = fastFallback,
-      routeDatabase = routeDatabase,
+      exchangeFinderFactory =  { pool, address, user ->
+        FastFallbackExchangeFinder(
+          ForceConnectRoutePlanner(
+            RealRoutePlanner(
+              taskRunner = taskRunner,
+              connectionPool = pool,
+              readTimeoutMillis = readTimeoutMillis,
+              writeTimeoutMillis = writeTimeoutMillis,
+              socketConnectTimeoutMillis = socketConnectTimeoutMillis,
+              socketReadTimeoutMillis = socketReadTimeoutMillis,
+              pingIntervalMillis = pingIntervalMillis,
+              retryOnConnectionFailure = retryOnConnectionFailure,
+              fastFallback = fastFallback,
+              address = address,
+              routeDatabase = routeDatabase,
+              connectionUser = user,
+            ),
+          ),
+          taskRunner,
+        )
+      }
+
     ),
   )
 
--- a/okhttp/src/main/kotlin/okhttp3/ConnectionPool.kt
+++ b/okhttp/src/main/kotlin/okhttp3/ConnectionPool.kt
@@ -145,6 +145,7 @@ class ConnectionPool internal constructor(
     /**
      * How many concurrent calls should be possible to make at any time.
      * The pool will routinely try to pre-emptively open connections to satisfy this minimum.
+     * Connections will still be closed if they idle beyond the keep-alive but will be replaced.
      */
     @JvmField val minimumConcurrentCalls: Int = 0,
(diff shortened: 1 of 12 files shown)


okhttp/src/main/kotlin/okhttp3/internal/connection/RealConnectionPool.kt
```diff
@@ -286,7 +303,82 @@ class RealConnectionPool(
...
   }
 
+  fun setPolicy(
+    address: Address,
+    policy: ConnectionPool.AddressPolicy,
+  ) {
+    // TODO: it's quite awkward that we have a new policy object that doesn't contain the max
+    require(policy.minimumConcurrentCalls <= maxIdleConnections) {
+      "minimumConcurrentCalls must be <= maxIdleConnections"
+    }
```
ean5533, line 313: The comment on `maxIdleConnections` says:

> The maximum number of idle connections for each address.

Which sounds like something that should also be a part of our new AddressPolicy.

Except, as far as I can tell from looking at the code, that's **not** what `maxIdleConnections` does. Instead, it seems to be a cap on the total number of connections in the pool, regardless of address.

So, how do we reconcile this?

okhttp/src/main/kotlin/okhttp3/internal/connection/RealConnectionPool.kt
```diff
@@ -286,7 +303,82 @@ class RealConnectionPool(
...
+      // This policy is unsatisfied -- open a connection!
+      relevantConnections.isEmpty() -> {
+        try {
+          val exchangeFinder = buildExchangeFinder(address)
+          val connection = exchangeFinder.find()
+          put(connection)
+
+          return 0 // run again immediately to create more connections if needed
+        } catch (ex: Exception) {
+          // TODO do something with exception?
```
ean5533, line 343: I couldn't find any prior art for what to do with exceptions that happen in some background thread like this. Do we log it somehow?

okhttp/src/main/kotlin/okhttp3/internal/connection/PoolConnectionUser.kt
```diff
@@ -0,0 +1,105 @@
...
+import okhttp3.Protocol
+import okhttp3.Route
+
+/**
+ * A user that is a connection pool creating connections in the background
+ * without an intent to immediately use them.
+ *
+ * Most of this class's functionality is currently stubbed out, but will be revisited in the future.
+ */
+class PoolConnectionUser : ConnectionUser {
```
ean5533, line 19: This class is currently a giant no-op. We intend to think more carefully about what the "right" thing to do here is, but first we want to make sure this feature works at all.

okhttp/src/main/kotlin/okhttp3/internal/connection/RealConnectionPool.kt
yschimke: I think generally we catch IOExceptions and handle. 

Unexpected runtime exceptions are allowed to crash the VM or whatever.

Known runtime exceptions in our dependencies, even bugs, are caught and handled.

okhttp/src/main/kotlin/okhttp3/internal/connection/ForceConnectRoutePlanner.kt
```diff
@@ -0,0 +1,22 @@
+package okhttp3.internal.connection
+
+import okhttp3.Address
+import okhttp3.HttpUrl
+
+/**
+ * A RoutePlanner that will always generate a ConnectPlan, ignoring any connection pooling
+ */
+class ForceConnectRoutePlanner(private val delegate: RealRoutePlanner) : RoutePlanner {
```
yschimke, line 9: I don't think we should allow this on Android by default. Network traffic while the app is backgrounded (no activity, service, or workmanager) should be avoided.

Or if network is lost.

I guess either
- document and failfast on android
- add some predicate that can let it work on android, such as isForegrounded.

ean5533: Got it, I'll figure something out

okhttp/src/test/java/okhttp3/internal/connection/ConnectionPoolTest.kt
```diff
@@ -192,6 +200,47 @@ class ConnectionPoolTest {
...
+    val address = routePlanner.address
+    val pool = factory.newConnectionPool(routePlanner = routePlanner)
+
+    // Set the policy, wait for connection to be created
+    routePlanner.addPlan()
+    pool.setPolicy(address, ConnectionPool.AddressPolicy(1))
+    // TODO why does runTasks() spin forever if I don't first call runNextTask() twice?
+    factory.taskFaker.runNextTask()
+    factory.taskFaker.runNextTask()
+    factory.taskFaker.runTasks()
```
ean5533, line 208: Could use some guidance on this

okhttp/src/main/kotlin/okhttp3/internal/connection/RealConnectionPool.kt
yschimke: oldestConnectionsEvictedIfIdleLimitExceeded test confirms it's across addresses.

okhttp/src/test/java/okhttp3/internal/connection/ConnectionPoolTest.kt
yschimke: Sounds like a bug, Jesse might know.  Otherwise will require debugging.

okhttp/src/main/kotlin/okhttp3/internal/connection/RealConnectionPool.kt
ean5533: I've updated the comment, and also updated `cleanup` to consider connections "in use" if they are needed to satisfy a policy

okhttp/src/main/kotlin/okhttp3/ConnectionPool.kt
```diff
@@ -39,13 +43,42 @@ class ConnectionPool internal constructor(
     timeUnit: TimeUnit = TimeUnit.MINUTES,
     taskRunner: TaskRunner = TaskRunner.INSTANCE,
     connectionListener: ConnectionListener = ConnectionListener.NONE,
+    readTimeoutMillis: Int = 10_000,
```
swankjesse, line 46: I wonder about these defaults, and that they’re duplicated here and in OkHttpClient.Builder.

Not a dealbreaker, just a thing I’m contemplating

okhttp/src/main/kotlin/okhttp3/ConnectionPool.kt
```diff
@@ -92,4 +125,34 @@ class ConnectionPool internal constructor(
...
+   */
+  class AddressPolicy(
+    /**
+     * How many concurrent calls should be possible to make at any time.
+     * The pool will routinely try to pre-emptively open connections to satisfy this minimum.
+     *
+     * This feature must not be used on Android because it may result in network traffic while
+     * the app is backgrounded. On Android, leave this set to 0.
+     */
+    @JvmField val minimumConcurrentCalls: Int = 0,
```
swankjesse, line 152: I wonder if we should use the word ‘Streams’ or ‘Connections’

swankjesse: Maybe Calls is the existing thing we have for streams anyway? That might be fine

okhttp/src/main/kotlin/okhttp3/ConnectionPool.kt
```diff
@@ -92,4 +125,34 @@ class ConnectionPool internal constructor(
...
+  /**
+   * A policy for how the pool should treat a specific address.
+   */
+  class AddressPolicy(
+    /**
+     * How many concurrent calls should be possible to make at any time.
+     * The pool will routinely try to pre-emptively open connections to satisfy this minimum.
+     *
+     * This feature must not be used on Android because it may result in network traffic while
+     * the app is backgrounded. On Android, leave this set to 0.
```
swankjesse, line 150: I don’t know about this comment. It might be true, but Android is not necessarily always deployed in the exact same way. It’s possible for an embedded Android to have 24-7 electricity and this would be a reasonable behavior there.

swankjesse (approved):

okhttp/src/main/kotlin/okhttp3/ConnectionPool.kt
ean5533: I'll move the code comment and we can choose to address it (or not) in the public docs

okhttp/src/main/kotlin/okhttp3/OkHttpClient.kt
```diff
@@ -547,7 +586,7 @@ open class OkHttpClient internal constructor(
 
   class Builder() {
     internal var dispatcher: Dispatcher = Dispatcher()
-    internal var connectionPool: ConnectionPool = ConnectionPool()
+    internal var connectionPool: ConnectionPool? = null
```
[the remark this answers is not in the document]
ean5533: Hm, that's interesting. I'm not sure what to do about this, because the connection pool now needs a reference to the `routeDatabase` and several other fields from the Builder, whose final values aren't known until `build()` is called.

How important is this use case? Notably, `client.newBuilder()` was already going to cause a new connection pool to be created anyway, so this only affects cases where a single builder instance gets `build()` called on it multiple times.

okhttp/src/main/kotlin/okhttp3/internal/connection/RealConnectionPool.kt
```diff
@@ -286,7 +313,75 @@ class RealConnectionPool(
...
+          return policy.backoffDelayMillis.jitterBy(policy.backoffJitterMillis) * 1_000_000
+        }
+      }
+    }
+
+    return MIN_CONNECTION_CHECK_INTERVAL_NANOS
+  }
+
+  private fun Long.jitterBy(amount: Int): Long {
+    return this + ThreadLocalRandom.current().nextInt(amount * -1, amount)
```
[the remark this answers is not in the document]
ean5533: Looks like API 21, so we should be good

okhttp/src/main/kotlin/okhttp3/OkHttpClient.kt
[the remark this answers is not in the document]
yschimke: Jesse, you are talking about this case right?

```
val clientBuilder = OkHttpClient.Builder()

val client1 = clientBuilder.build()
val client2 = clientBuilder.build()
val client3 = client1.newBuilder().build()
```

I'd expect 1 & 3 to share a connection pool.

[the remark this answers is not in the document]
ean5533: Ok, I've fixed the code to cache the connection pool that it creates if one wasn't set by the caller. This should be good now.

ean5533:
Summary of most recent changes:

* Only schedule minConnections task when a policy or connection changes
* Rejiggered `policies` to be thread safe via copy-on-write

Things that have **not** been resolved:

* `PoolConnectionUser` is a big no-op so it doesn't call `ConnectionListener` yet when connections are created/failed
* There's no way for a user to tell when a policy has been applied (so no way to wait for the client to be "ready")

I think both of these things can be fixed in followup PRs since they only enhance this new feature and don't affect existing functionality.

swankjesse (commented):
Hopefully you don’t mind I think this needs a full extra round.

Also I’d like to pair on creating a bunch of tests for this. Some candidates:
 * we adjust the limit down dynamically, and connections are closed
 * we adjust the limit down dynamically, and connections are opened
 * HTTP/2 settings change with a new higher allocation limit, and we close connections
 * HTTP/2 settings change with a new lower allocation limit, and we open connections
 * a connection is opened by the opener, closed by duration limit, and then opened again

okhttp/src/main/kotlin/okhttp3/internal/connection/RealConnectionPool.kt
```diff
@@ -187,12 +204,23 @@ class RealConnectionPool(
...
+        val satisfiablePolicy =
+          policies.entries.firstOrNull {
+            it.value.unsatisfiedCountCleanupTask > 0 && connection.isEligible(it.key, null)
+          }
+
         if (pruneAndGetAllocationCount(connection, now) > 0) {
+          // If the connection is in use, keep searching.
+          inUseConnectionCount++
+        } else if (satisfiablePolicy != null) {
+          // If the connection helps satisfy a policy, keep searching.
```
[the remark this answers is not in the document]
ean5533: Whoops, I missed this comment from your previous review. You're right, will fix

okhttp/src/main/kotlin/okhttp3/internal/connection/RealConnectionPool.kt
```diff
@@ -286,6 +314,91 @@ class RealConnectionPool(
...
+   * If not, we create one and then schedule the task to run again immediately.
+   */
+  private fun ensureMinimumConnections(
+    address: Address,
+    state: MinimumConnectionState,
+  ): Long {
+    // This policy does not require minimum connections, don't run again
+    if (state.policy.minimumConcurrentCalls < 1) return -1
+
+    state.unsatisfiedCountMinTask = state.policy.minimumConcurrentCalls
```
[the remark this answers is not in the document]
ean5533: Can do, was just providing symmetry with `state.unsatisfiedCountCleanupTask`

ean5533:
Alright, I'm about to merge this. We've addressed all of the concurrency and allocation concerns. The last remaining problem is that the unit tests are commented out; they currently fail due to problems in `TaskFaker.runTasks()`. Fixing those problems is non-trivial, so I'm going to fast follow this PR with another one that addresses those problems and uncomments the tests.

PR #8289, opened 2024-03-16
Add loom pinning tests

yschimke:


build.gradle.kts
```diff
@@ -169,8 +169,7 @@ subprojects {
 
     if (platform == "loom") {
       jvmArgs(
-        "-Djdk.tracePinnedThread=full",
-        "--enable-preview"
+        "-Djdk.tracePinnedThreads=short",
```
yschimke, line 172: not: the s was missing.

yschimke:
```
    okhttp3.internal.http2.Http2Writer.flush(Http2Writer.kt:125) <== monitors:1
    okhttp3.internal.http2.Http2Connection.newStream(Http2Connection.kt:271) <== monitors:2
Thread[#62,ForkJoinPool-1-worker-3,5,CarrierThreads]
    okhttp3.internal.http2.Http2Stream$FramingSource.read(Http2Stream.kt:395) <== monitors:1
```

yschimke:
@swankjesse Going to land even though it has a failing test for loom. The break is real, and I'd like to fix.

requesting post review, tests and CI only.

container-tests/src/test/java/okhttp3/containers/BasicLoomTest.kt
```diff
@@ -0,0 +1,97 @@
...
+
+@Testcontainers
+@Isolated
+class BasicLoomTest {
+  @JvmField
+  @RegisterExtension
+  val platform = PlatformRule()
+
+  @Container
+  val mockServer: MockServerContainer = MockServerContainer(MOCKSERVER_IMAGE)
```
swankjesse, line 38: Why isn’t this using MockWebServer?!

container-tests/src/test/java/okhttp3/containers/BasicLoomTest.kt
```diff
@@ -0,0 +1,97 @@
...
+
+    client =
+      OkHttpClient.Builder()
+        .trustMockServer()
+        .dispatcher(Dispatcher(newVirtualThreadPerTaskExecutor()))
+        .build()
+
+    executor = newVirtualThreadPerTaskExecutor()
+
+    System.setOut(PrintStream(capturedOut))
```
swankjesse, line 59: What’s this for?

swankjesse (commented):
LGTM

container-tests/src/test/java/okhttp3/containers/BasicLoomTest.kt
yschimke: I'll put a comment. I only wanted to fail on the client side. Not to test our mockwebserver is loom compliant also.

container-tests/src/test/java/okhttp3/containers/BasicLoomTest.kt
yschimke: I don't have a programmatic way to detect loom blocking. Just capture this warning.

container-tests/src/test/java/okhttp3/containers/BasicLoomTest.kt
swankjesse: Makes sense. I’d like to also fix MockWebServer to be compliant!

container-tests/src/test/java/okhttp3/containers/BasicLoomTest.kt
swankjesse: Gotcha.

Issue #8284, opened 2024-03-14
Virtual thread pinning with 5.0.0-alpha.12


Packsolite:
https://github.com/square/okhttp/blob/bad2ebc296c8094370cf94291f571dd618ba7400/okhttp/src/main/kotlin/okhttp3/internal/connection/RealCall.kt#L376

This line causes virtual thread pinning in the latest alpha (5.0.0-alpha.12).
Reason is the use of a ReentrantLock (in TaskQueue.schedule) inside a synchronized block when using a connection pool with a `maxIdleConnection`.

Log output with JVM argument `-Djdk.tracePinnedThreads=full`:
```
[13:36:10] [ForkJoinPool-2-worker-10/INFO]: Thread[#137,ForkJoinPool-2-worker-10,5,CarrierThreads]
[13:36:10] [ForkJoinPool-2-worker-10/INFO]: java.base/java.lang.VirtualThread$VThreadContinuation.onPinned(VirtualThread.java:183)
[13:36:10] [ForkJoinPool-2-worker-10/INFO]: java.base/jdk.internal.vm.Continuation.onPinned0(Continuation.java:393)
[13:36:10] [ForkJoinPool-2-worker-10/INFO]: java.base/java.lang.VirtualThread.park(VirtualThread.java:582) 
[13:36:10] [ForkJoinPool-2-worker-10/INFO]: java.base/java.lang.System$2.parkVirtualThread(System.java:2639)
[13:36:10] [ForkJoinPool-2-worker-10/INFO]: java.base/jdk.internal.misc.VirtualThreads.park(VirtualThreads.java:54)
[13:36:10] [ForkJoinPool-2-worker-10/INFO]: java.base/java.util.concurrent.locks.LockSupport.park(LockSupport.java:219)
[13:36:10] [ForkJoinPool-2-worker-10/INFO]: java.base/java.util.concurrent.locks.AbstractQueuedSynchronizer.acquire(AbstractQueuedSynchronizer.java:754)
[13:36:10] [ForkJoinPool-2-worker-10/INFO]: java.base/java.util.concurrent.locks.AbstractQueuedSynchronizer.acquire(AbstractQueuedSynchronizer.java:990)
[13:36:10] [ForkJoinPool-2-worker-10/INFO]: java.base/java.util.concurrent.locks.ReentrantLock$Sync.lock(ReentrantLock.java:153)
[...]

PR #8290, opened 2024-03-17
Move RealCall and RealConnection to loom safe locks

yschimke:
Main fix for https://github.com/square/okhttp/issues/8284

replace synchronized with ReentrantLock.withLock, and wait/notify with lock.newCondition.signal/await.

Relies on test results from BasicLoomTest

### DIFF
--- a/okhttp-testing-support/src/main/kotlin/okhttp3/TestValueFactory.kt
+++ b/okhttp-testing-support/src/main/kotlin/okhttp3/TestValueFactory.kt
@@ -31,6 +31,7 @@ import javax.net.SocketFactory
 import javax.net.ssl.HostnameVerifier
 import javax.net.ssl.HttpsURLConnection
 import javax.net.ssl.SSLSocketFactory
+import kotlin.concurrent.withLock
 import okhttp3.internal.RecordingOkAuthenticator
 import okhttp3.internal.concurrent.TaskFaker
 import okhttp3.internal.concurrent.TaskRunner
@@ -93,7 +94,7 @@ class TestValueFactory : Closeable {
         socket = Socket(),
         idleAtNs = idleAtNanos,
       )
-    synchronized(result) { pool.put(result) }
+    result.lock.withLock { pool.put(result) }
     return result
   }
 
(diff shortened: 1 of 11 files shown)


swankjesse (approved):
What are the best practices when `synchronized` is only used to guard state, and never used for a blocking or slow call? From our concurrency rules, a bunch of these might be unnecessary?

```
### Locks

We have 3 different things that we synchronize on.

#### Http2Connection

This lock guards internal state of each connection. This lock is never held for blocking operations. That means that we acquire the lock, read or write a few fields and release the lock. No I/O and no application-layer callbacks.

#### Http2Stream

This lock guards the internal state of each stream. As above, it is never held for blocking operations. When we need to hold an application thread to block a read, we use wait/notify on this lock. This works because the lock is released while `wait()` is waiting.

#### Http2Writer

Socket writes are guarded by the Http2Writer. Only one stream can write at a time so that messages are not interleaved. Writes are either made by application-layer threads or the do-stuff-later pool.

...

#### Per-Connection Locks

Each connection has its own lock. The connections in the pool are all in a `ConcurrentLinkedQueue`. Due to data races, iterators of this queue may return removed connections. Callers must check the connection's `noNewExchanges` property before using connections from the pool.

The connection lock is never held while doing I/O (even closing a socket) to prevent contention.

A lock-per-connection is used to maximize concurrency.
```

(We definitely need the Http2Writer part of this)

swankjesse:
Maybe to frame it more precisely:
 - what’s the runtime performance difference for locks vs. synchronized on physical threads?
 - what’s the runtime performance difference for locks vs. synchronized on virtual threads?

Maybe we can find benchmarks on this.

yschimke:
I'll create a tracking bug to benchmark. And review and rescope the locks.

I'll land this as the tests show it's a real problem unless we change the locks.

swankjesse:
I don’t think we need to do any benchmarking work. For my own sake I just want an intuition on which locks to use where.

yschimke:
@swankjesse Are we saying that the locks are not needed for RealCall, RealConnection, Dispatcher? That these should be reworked?

swankjesse:
I think there’s two very different ways we use `synchronized`:

1. To guard in-memory state. For the accesses we acquire the lock, read or write some fields, and release the lock. There’s unlikely to be any contention for these, and if there is it’s likely to be very short-lived.
2. To guard I/O. For these accesses we acquire the lock, perform some potentially-slow I/O operation, and release the lock. These ones are likely to have contention.

I think we definitely want to change 2 to Loom-safe locks. I am unconvinced that there’s a benefit for changing 1. But there is a potential memory + performance cost.

yschimke:
A dissenter https://medium.com/@apusingh1967/how-synchronized-beats-reentrantlock-simplified-for-a-long-suffering-java-programmer-c83d069a6fc0

PR #8302, opened 2024-03-22
Add an Android HttpEngine regression test

yschimke:


yschimke:

Just a test we can compare any reported bugs against Android HttpEngine

regression-test/src/androidTest/java/okhttp/regression/compare/AndroidHttpEngineTest.kt
```diff
@@ -0,0 +1,198 @@
...
+          override fun onReadCompleted(
+            request: UrlRequest,
+            info: UrlResponseInfo,
+            byteBuffer: ByteBuffer,
+          ) {
+            println("onReadCompleted ${info.headers.asMap}")
+            byteBuffer.flip()
+            buffer.write(byteBuffer)
+            byteBuffer.clear()
+            request.read(byteBuffer)
```
swankjesse, line 148: Off topic: I find this API to be completely confusing

swankjesse (commented):
LGTM

regression-test/src/androidTest/java/okhttp/regression/compare/AndroidHttpEngineTest.kt
yschimke: Yeah, agreed. I struggled to write this and I've used it before.

It is possibly optimal with zero copy transfer etc. and you can either consume like I have here as response bytes are received, or out of line as a caller requests them.

PR #8304, opened 2024-03-24
Test using Http/3

yschimke:
Comparing results between HttpClient h3 and OkHttp.

Enable HTTP/3


regression-test/src/androidTest/java/okhttp/regression/compare/AndroidHttpEngineTest.kt
```diff
@@ -57,10 +57,9 @@ class AndroidHttpEngineTest {
     }
   val engine =
     HttpEngine.Builder(context)
-      .setEnableHttp2(true)
```
yschimke, line 60: default to true

yschimke:
@swankjesse test only

swankjesse (commented):
LGTM! Neat!

Issue #3154, opened 2017-02-05
java.lang.IllegalArgumentException: method GET must not have a request body.


hex6ng:
What kind of issue is this?

 - [ ] Question. This issue tracker is not the place for questions. If you want to ask how to do
       something, or to understand why something isn't working the way you expect it to, use Stack
       Overflow. https://stackoverflow.com/questions/tagged/okhttp

 - [x] Bug report. If you’ve found a bug, spend the time to write a failing test. Bugs with tests
       get fixed. Here’s an example: https://gist.github.com/swankjesse/981fcae102f513eb13ed

 - [ ] Feature Request. Start by telling us what problem you’re trying to solve. Often a solution
       already exists! Don’t send pull requests to implement new features without first getting our
       support. Sometimes we leave features out on purpose to keep the project small.

I'm trying to do a get request to Elasticsearch with a query body but OkHttp doesn't allow that.  The query works with curl:

```sh
curl -XGET 'localhost:9200/fncd/event/_search?scroll=1m' -d '
{
    "query": {
	    "range" : {
	        "lastUpdated" : {
	            "gte": 1486184400000
	        }
	    }
    }
}
'
```

swankjesse:
OkHttp doesn’t support bodies for `GET` requests. For one reason, it makes caching too difficult to reason about.

dmscheck:
Will the change in PR #3038 fix this?  Is there a timeframe on when that will be released?

swankjesse:
It’s unlikely we’ll ever support request bodies with GET.

No further action to take on this.

alvico:
Is there any reason why you did not plan to support this? It will be quite useful to use it, for instance when testing elasticsearch queries like in the example of the question. (Had the same problem xD)

swankjesse:
OkHttp is an opinionated HTTP client. By constraining what we support to what the specs allow, we’re able to keep things relatively simple and offer great features like smart caching.

Elasticsearch is wrong for exposing APIs that use GET requests with bodies. They should fix their API.

fengshuzi:
please

DCKcode:
@swankjesse Could you please reopen this? While it's true that in the distant past the HTTP spec did not permit GET requests with a body, any such passages indicating this were specifically removed from RFC7230-7237. There is no argument based in the HTTP RFCs not to allow this.

See also: https://stackoverflow.com/a/983458/884848

swankjesse:
> A payload within a GET request message has no defined semantics;
[...]

PR #8306, opened 2024-03-25
Allow GET requests to have a request body

mo7ty:
Although it is not recommended, if agreed between client and server, GET requests may have content.

* As per [RFC 9110 - HTTP Semantics - section 9.3.1. GET](https://datatracker.ietf.org/doc/html/rfc9110#section-9.3.1):
_(...) A client [SHOULD NOT](https://datatracker.ietf.org/doc/html/rfc2119#section-4) generate content in a GET request unless it
is made directly to an origin server that has previously indicated,
in or out of band, that such a request has a purpose and will be
adequately supported. (...)_
( != [_must not_](https://datatracker.ietf.org/doc/html/rfc2119#section-2))

### DIFF
--- a/okhttp/src/main/kotlin/okhttp3/internal/http/HttpMethod.kt
+++ b/okhttp/src/main/kotlin/okhttp3/internal/http/HttpMethod.kt
@@ -34,7 +34,7 @@ object HttpMethod {
     )
 
   @JvmStatic // Despite being 'internal', this method is called by popular 3rd party SDKs.
-  fun permitsRequestBody(method: String): Boolean = !(method == "GET" || method == "HEAD")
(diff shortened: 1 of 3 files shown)


yschimke:
Prior discussion in https://github.com/square/okhttp/issues/3154

mo7ty:
Isn't desirable for OkHttp to be compliant with [RFC 9110 - HTTP Semantics](https://datatracker.ietf.org/doc/html/rfc9110)?

Or is it just technically not possible?

Is there any workaround besides using a non-compliant method?
(ref [_method token is case-sensitive_](https://datatracker.ietf.org/doc/html/rfc9110#section-9.1-4))

PR #8323, opened 2024-04-01
Don't depend on external OkHttp from within OkHttp

swankjesse:
IntelliJ grabs the wrong one which is quite confusing.


yschimke (approved):
I was deliberately using it for comparison.

But not worth the pain.

swankjesse:
Yeah, I agree with the intention!

PR #8325, opened 2024-04-01
Don't test MockServer's HTTP proxy of an HTTP/2 call

swankjesse:
This test is disabled, and it refers to a closed bug.


container-tests/src/test/java/okhttp3/containers/BasicProxyTest.kt
```diff
@@ -107,27 +106,6 @@ class BasicProxyTest {
     }
   }
 
-  @Test
-  @Disabled("https://github.com/square/okhttp/issues/8233")
```
swankjesse, line 111: I think the way to revive this is to figure out how to get MockServer to do an HTTP/2 connection over an HTTP proxy. 

(We implement this in MockWebServer and test that in HttpOverHttp2Test.http2WithProxy(), so I don’t think it’s a problem with OkHttp.)

PR #8330, opened 2024-04-05
Don't leak response bodies in executeAsync

swankjesse:
Also make callers opt in to an unstable coroutines API. If the resource
cleanup coroutines API changes, we'll have to change this API.

Remove the OkHttp experimental API. This is a good enough API as far
as OkHttp is concerned.

### DIFF
--- a/okhttp-coroutines/src/main/kotlin/okhttp3/JvmCallExtensions.kt
+++ b/okhttp-coroutines/src/main/kotlin/okhttp3/JvmCallExtensions.kt
@@ -20,10 +20,10 @@ package okhttp3
 import kotlin.coroutines.resumeWithException
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.suspendCancellableCoroutine
+import okhttp3.internal.closeQuietly
 import okio.IOException
 
-@OptIn(ExperimentalCoroutinesApi::class)
-@ExperimentalOkHttpApi
+@ExperimentalCoroutinesApi // resume with a resource cleanup.
 suspend fun Call.executeAsync(): Response =
   suspendCancellableCoroutine { continuation ->
     continuation.invokeOnCancellation {
@@ -42,7 +42,9 @@ suspend fun Call.executeAsync(): Response =
           call: Call,
           response: Response,
         ) {
-          continuation.resume(value = response, onCancellation = { call.cancel() })
+          continuation.resume(response) {
+            response.closeQuietly()
+          }
         }
       },
     )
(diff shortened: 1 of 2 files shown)


okhttp-coroutines/src/main/kotlin/okhttp3/JvmCallExtensions.kt
```diff
@@ -42,7 +42,9 @@ suspend fun Call.executeAsync(): Response =
           call: Call,
           response: Response,
         ) {
-          continuation.resume(value = response, onCancellation = { call.cancel() })
+          continuation.resume(response) {
+            response.closeQuietly()
```
swankjesse, line 46: I am naughty and did both a formatting change and a behavior change here.

Used to be cancel -- asynchronously interrupt a call without releasing any resources
Now it’s close -- synchronously release resources

okhttp-coroutines/src/main/kotlin/okhttp3/JvmCallExtensions.kt
```diff
@@ -20,10 +20,10 @@ package okhttp3
 import kotlin.coroutines.resumeWithException
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.suspendCancellableCoroutine
+import okhttp3.internal.closeQuietly
 import okio.IOException
 
-@OptIn(ExperimentalCoroutinesApi::class)
-@ExperimentalOkHttpApi
+@ExperimentalCoroutinesApi // resume with a resource cleanup.
```
swankjesse, line 26: https://github.com/Kotlin/kotlinx.coroutines/issues/4088

yschimke (approved):

okhttp-coroutines/src/main/kotlin/okhttp3/JvmCallExtensions.kt
Endeavour233: Should Retrofit borrow this improvement in `retrofit2/KotlinExtensions.kt`? @JakeWharton

JakeWharton: Retrofit doesn't have this problem. We never pass a live resource across the coroutine boundary. Our underlying request has already been completely consumed and closed before we resume.

Endeavour233: > Retrofit doesn't have this problem. We never pass a live resource across the coroutine boundary. Our underlying request has already been completely consumed and closed before we resume.

Got it, thanks

PR #8338, opened 2024-04-06
Change parameter order for Cache constructor

swankjesse:
Put FileSystem before Path, as is convention.

Also put this new constructor in the public API.


JakeWharton (approved):
Neither Zipline nor Redwood put the FileSystem before the Path. File bugs on both to update?

swankjesse:
Good idea.
https://github.com/cashapp/redwood/issues/1942
https://github.com/cashapp/zipline/issues/1283

yschimke (approved):

PR #8339, opened 2024-04-06
Add a StrictMode noteSlowCall for SSL init

yschimke:
Apps may want to defer SSL init off the main thread, so warn if they have enabled Strict Mode.

Alternative or precursor to https://github.com/square/okhttp/pull/8248, should land which ever way that goes.


yschimke:
@swankjesse if we don't do https://github.com/square/okhttp/pull/8248, then we should be noisy when it's used on main.

yschimke:
@swankjesse any objections to this?

android-test/src/androidTest/java/okhttp/android/test/StrictModeTest.kt
```diff
@@ -0,0 +1,69 @@
...
+    assertThat(violations).isEmpty()
+  }
+
+  private fun applyStrictMode() {
+    StrictMode.setThreadPolicy(
+      ThreadPolicy.Builder()
+        .detectCustomSlowCalls()
+        .penaltyListener({ it.run() }) {
+          violations.add(it)
+        }
```
JakeWharton, line 65: Hmmm. When was this added? I would assume after our minSdk, so it should really have an `@SdkSuppress(minSdk=something)` on the class.

JakeWharton (approved):

android-test/src/androidTest/java/okhttp/android/test/StrictModeTest.kt
yschimke: I'll check, also why it's not failing if so.

yschimke: Ahhh, just the tests. And not running because of https://github.com/mannodermaus/android-junit5/issues/225 and a minSdk 26 requirement for tests.

JakeWharton: Ah. JUnit 5 continues to prove that it's the wrong choice for any project.

PR #8354, opened 2024-04-10
Rename mockwebserver-deprecated to mockwebserver-2015

swankjesse:
I'm in no rush to deprecate this old API. We have to finish the new API first.


swankjesse:
This interferes with publishing. Ugh.

PR #8355, opened 2024-04-10
Fix public APIs for kotlin.time.Duration

swankjesse:
Use this as our preferred API for accepting a duration in OkHttpClient and CacheControl.

Also hide these functions from the Java API.

### DIFF
--- a/okhttp/src/main/kotlin/okhttp3/CacheControl.kt
+++ b/okhttp/src/main/kotlin/okhttp3/CacheControl.kt
@@ -186,23 +186,26 @@ class CacheControl internal constructor(
      * @param maxAge a non-negative duration. This is stored and transmitted with [TimeUnit.SECONDS]
      *     precision; finer precision will be lost.
      */
-    fun maxAge(maxAge: Duration) = apply {
-      val maxAgeSeconds = maxAge.inWholeSeconds
-      require(maxAgeSeconds >= 0) { "maxAge < 0: $maxAgeSeconds" }
-      this.maxAgeSeconds = maxAgeSeconds.commonClampToInt()
-    }
+    fun maxAge(maxAge: Duration) =
+      apply {
+        val maxAgeSeconds = maxAge.inWholeSeconds
+        require(maxAgeSeconds >= 0) { "maxAge < 0: $maxAgeSeconds" }
+        this.maxAgeSeconds = maxAgeSeconds.commonClampToInt()
+      }
 
-    fun maxStale(maxStale: Duration) = apply {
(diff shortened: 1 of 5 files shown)


okhttp/src/main/kotlin/okhttp3/CacheControl.kt
```diff
@@ -186,26 +183,26 @@ class CacheControl internal constructor(
...
-     * @param maxAge a non-negative integer. This is stored and transmitted with [TimeUnit.SECONDS]
+     * @param maxAge a non-negative duration. This is stored and transmitted with [TimeUnit.SECONDS]
      *     precision; finer precision will be lost.
      */
-    @ExperimentalOkHttpApi
-    fun maxAge(
-      maxAge: Int,
-      timeUnit: DurationUnit,
-    ) = commonMaxAge(maxAge, timeUnit)
+    @JvmSynthetic // Don't expose @JvmInline types in the Java API.
```
JakeWharton, line 189: You shouldn't need any of these annotations as functions accepting value types should already have their names mangled to include a hypen which is an invalid character in Java method names. The API dump should confirm.

okhttp/src/main/kotlin/okhttp3/CacheControl.kt
```diff
@@ -186,26 +183,26 @@ class CacheControl internal constructor(
...
+    fun maxAge(maxAge: Duration) = apply {
+      require(maxAge.inWholeSeconds >= 0) { "maxAge < 0: ${maxAge.inWholeSeconds}" }
```
JakeWharton, line 191: If you do
```suggestion
      require(!maxAge.isNegative()) { "maxAge < 0: ${maxAge.inWholeSeconds}" }
```
you avoid having to do an internal conversion solely to perform the check

JakeWharton (approved):

okhttp/src/main/kotlin/okhttp3/CacheControl.kt
swankjesse: Fixed.

okhttp/src/main/kotlin/okhttp3/CacheControl.kt
swankjesse: Fixed to call inWholeSeconds once.

madd-44:
> Use this as our preferred API for accepting a duration in OkHttpClient and CacheControl.
> 
> Also hide these functions from the Java API.

>Sorry didn't know what to sorry

madd-44:
Is your freanx hacking by phone brother.  Talk to him plz

PR #8363, opened 2024-04-14
Assert these chars are skipped in hostnames: " < > ^ ` { | }

swankjesse:


yschimke:
Test only, widening the checks.  Will merge after it passes.

PR #8367, opened 2024-04-17
Revert "Move RealCall and RealConnection to loom safe locks"

swankjesse:
Reverts square/okhttp#8290

Unfortunately I think this PR is missing some `synchronized` blocks that needed to be converted to `withLock {}` blocks. Mixing and matching the two definitely doesn’t work, and our tests aren’t sophisticated enough to cover that.

I think this is a good idea; we just need to be particularly precise with the implementation. Every occurrence of a lock needs to be fixed together; otherwise we split the lock.


yschimke:
Damn sorry, I merged against the pool warmer changes, thought I got them and obviously missed some

chenchenfang:
So when can you use virtual threads?

PR #8371, opened 2024-04-17
Move to centralised loom safe locks

yschimke:
Mostly https://github.com/square/okhttp/pull/8290

But with a centralised lock definition.  Draft for discussion, but possibly we could encode the rules from docs/contribute/concurrency.md, so we fail fast if we ever break them.

### DIFF
--- a/okhttp/src/main/kotlin/okhttp3/Dispatcher.kt
+++ b/okhttp/src/main/kotlin/okhttp3/Dispatcher.kt
@@ -223,9 +223,10 @@ class Dispatcher() {
   }
 
   /** Used by [Call.execute] to signal it is in-flight. */
-  internal fun executed(call: RealCall) = this.withLock {
-    runningSyncCalls.add(call)
-  }
+  internal fun executed(call: RealCall) =
+    this.withLock {
+      runningSyncCalls.add(call)
+    }
 
   /** Used by [AsyncCall.run] to signal completion. */
   internal fun finished(call: AsyncCall) {
@@ -256,14 +257,16 @@ class Dispatcher() {
   }
 
   /** Returns a snapshot of the calls currently awaiting execution. */
-  fun queuedCalls(): List<Call> = this.withLock {
-    return Collections.unmodifiableList(readyAsyncCalls.map { it.call })
-  }
+  fun queuedCalls(): List<Call> =
+    this.withLock {
+      return Collections.unmodifiableList(readyAsyncCalls.map { it.call })
+    }
 
   /** Returns a snapshot of the calls currently being executed. */
-  fun runningCalls(): List<Call> = this.withLock {
-    return Collections.unmodifiableList(runningSyncCalls + runningAsyncCalls.map { it.call })
-  }
+  fun runningCalls(): List<Call> =
+    this.withLock {
+      return Collections.unmodifiableList(runningSyncCalls + runningAsyncCalls.map { it.call })
+    }
 
   fun queuedCallsCount(): Int = this.withLock { readyAsyncCalls.size }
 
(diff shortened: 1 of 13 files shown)


okhttp/src/main/kotlin/okhttp3/internal/connection/Locks.kt
```diff
@@ -0,0 +1,78 @@
...
+    assertThreadDoesntHoldLock()
+    return lock.withLock(action)
+  }
+
+  inline fun <T> Http2Writer.withLock(action: () -> T): T {
+    contract { callsInPlace(action, InvocationKind.EXACTLY_ONCE) }
+
+    // TODO remove
+    assertThreadDoesntHoldLock()
+    // TODO can we assert we don't have the connection lock?
```
yschimke, line 74: What's a good structure to handle this sort of lock hierarchy?

yschimke:
Catching the ones I missed

```
Http2ConnectionTest > sendGoAway() FAILED
    java.lang.AssertionError: Thread Test worker MUST NOT hold lock on okhttp3.internal.http2.Http2Connection@1e253c9d
        at okhttp3.internal.http2.Http2Connection.isHealthy(Http2Connection.kt:1205)
        at okhttp3.internal.http2.Http2ConnectionTest.sendGoAway(Http2ConnectionTest.kt:1397)
```

okhttp/src/main/kotlin/okhttp3/Dispatcher.kt
```diff
@@ -82,29 +88,31 @@ class Dispatcher() {
    * This means that if you are doing synchronous calls the network layer will not truly be idle
    * until every returned [Response] has been closed.
    */
-  @set:Synchronized
-  @get:Synchronized
   var idleCallback: Runnable? = null
+    get() = this.withLock { field }
```
swankjesse, line 92: Maybe for follow up: should we do a property delegate so we can have syntax like this?
```
var idleCallback: Runnable? by lock.property(null)
```

(or similar)

okhttp/src/main/kotlin/okhttp3/internal/connection/Locks.kt
```diff
@@ -0,0 +1,78 @@
...
+import okhttp3.internal.http2.Http2Stream
+import okhttp3.internal.http2.Http2Writer
+
+/**
+ * Centralisation of central locks according to docs/contribute/concurrency.md
+ */
+internal object Locks {
+  inline fun <T> Dispatcher.withLock(action: () -> T): T {
+    contract { callsInPlace(action, InvocationKind.EXACTLY_ONCE) }
+    // TODO remove
```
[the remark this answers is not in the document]
swankjesse: Why do all the work to have a ReentrantLock and then forbid its use?

These checks are not fast enough to include.
https://publicobject.com/2019/11/18/kotlins-assert-is-not-like-javas-assert/

okhttp/src/main/kotlin/okhttp3/internal/connection/RealCall.kt
```diff
@@ -95,7 +100,7 @@ class RealCall(
   internal var interceptorScopedExchange: Exchange? = null
     private set
 
-  // These properties are guarded by this. They are typically only accessed by the thread executing
+  // These properties are guarded by lock. They are typically only accessed by the thread executing
```
swankjesse, line 103: `[lock]` to tell the IDE this is a variable?

okhttp/src/main/kotlin/okhttp3/internal/http2/Http2Stream.kt
```diff
@@ -43,7 +44,10 @@ class Http2Stream internal constructor(
   inFinished: Boolean,
   headers: Headers?,
 ) {
-  // Internal state is guarded by this. No long-running or potentially blocking operations are
+  internal val lock: ReentrantLock = ReentrantLock()
+  val condition: Condition = lock.newCondition()
+
+  // Internal state is guarded by lock. No long-running or potentially blocking operations are
```
swankjesse, line 50: `[lock]`

swankjesse (approved):
This is great!

okhttp/src/main/kotlin/okhttp3/internal/connection/Locks.kt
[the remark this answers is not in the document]
yschimke: It's the opposite, this was a quick check to fail if we also had synchronization on the owner, not a concurrent Lock.

PR #8372, opened 2024-04-18
Move coroutines extensions to okhttp3.coroutines

swankjesse:
This has the side-effect of fixing our packaging of this module. Previously the okhttp-coroutines.jar contained the code from the :okhttp artifact, not the :okhttp-coroutines artifact. I suspect something in one of our build steps is incorrectly using the package names (only) as a cache key, and it should instead be using the package contents.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,6 +1,5 @@
 @file:Suppress("UnstableApiUsage")
 
-import com.diffplug.gradle.spotless.KotlinExtension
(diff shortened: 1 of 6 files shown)


swankjesse:
https://github.com/bndtools/bnd/issues/6089

JakeWharton (approved):
Breaks the Java module system, too, so a good change regardless of the OSGi nonsense.

yschimke (approved):

Issue #8286, opened 2024-03-15
How to cleanly pin a network


yschimke:
WorkManager provides a required network for scheduled jobs with network constraints

https://developer.android.com/reference/androidx/work/WorkerParameters#getNetwork()

Using this specific network is possible with a custom Dns and SocketFactory.  However the naive implementation would probably result in defeating connection pooling in some cases.

Is there a supported pattern for this, if Dns and SocketFactory both support equality checks, would pooling be correctly retained?

cc @swankjesse

swankjesse:
How about this:

```
/**
 * Returns a DNS that uses this network’s `getAllByName` function.
 * 
 * Different objects returned from this will be equal if their networks are equal.
 */
val Network.dns: Dns
```

And maybe also this:

```
/**
 * Returns a client that uses [network] for its socket factory and DNS. The
 * client will only be able to make calls to hosts that are reachable on [network].
 */
fun OkHttpClient.Builder.network(network: Network)
```

Plus `@JvmName` stuff to make the Java APIs reasonable.

Do we need to hide these in the `okhttp-android` module? It’d be nice to do it in the main `okhttp` module.

yschimke:
@swankjesse for discussion in https://github.com/square/okhttp/pull/8376, I think I want to move away from recreating the OkHttpClient instance, and instead configure Dns, SocketFactory and Address matcher? that listen for a Request tag (NetworkPin) that callers or interceptors set.

appdev:
> @swankjesse for discussion in #8376, I think I want to move away from recreating the OkHttpClient instance, and instead configure Dns, SocketFactory and Address matcher? that listen for a Request tag (NetworkPin) that callers or interceptors set. 对于 #8376 中的讨论，我想我不想重新创建 OkHttpClient 实例，而是配置 Dns、SocketFactory 和地址匹配器？侦听调用者或拦截器设置的请求标记 (NetworkPin)。

Hello, have you solved this problem?

yschimke:
No, the approach requires a new client.

yschimke:
in progress PR I'll come back to in https://github.com/square/okhttp/pull/8376

PR #8376, opened 2024-04-20
Network pinning on Android

yschimke:
Cleanly pinning network on Android - https://github.com/square/okhttp/issues/8286

```
    OkHttpClient
      .Builder()
      .addInterceptor(AndroidNetworkPinning())
      .build()

// Substitute in some preferred network here
val activeNetwork = connectivityManager.activeNetwork

  .url(server.url("/"))
  .tag<Network>(activeNetwork)
  .build()
```

### DIFF
--- a/android-test/src/androidDeviceTest/java/okhttp/android/test/AndroidNetworkPinningTest.kt
+++ b/android-test/src/androidDeviceTest/java/okhttp/android/test/AndroidNetworkPinningTest.kt
@@ -1,0 +1,102 @@
+/*
+ * Copyright (C) 2025 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okhttp.android.test
+
+import android.content.Context
+import android.net.ConnectivityManager
+import android.net.Network
+import android.os.Build
+import androidx.test.core.app.ApplicationProvider
+import androidx.test.filters.SdkSuppress
+import mockwebserver3.MockResponse
+import mockwebserver3.MockWebServer
+import mockwebserver3.junit5.StartStop
+import okhttp3.OkHttpClient
+import okhttp3.Request
+import okhttp3.android.AndroidNetworkPinning
+import okhttp3.internal.platform.PlatformRegistry
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Assertions.assertNotEquals
+import org.junit.jupiter.api.Assumptions.assumeTrue
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Tag
+import org.junit.jupiter.api.Test
+
+@Tag("Slow")
+@SdkSuppress(minSdkVersion = Build.VERSION_CODES.Q)
+class AndroidNetworkPinningTest {
+
+  val applicationContext = ApplicationProvider.getApplicationContext<Context>()
+  val connectivityManager = applicationContext.getSystemService(ConnectivityManager::class.java)
+
+  private var client: OkHttpClient =
+    OkHttpClient
+      .Builder()
+      .addInterceptor(AndroidNetworkPinning())
+      .addInterceptor {
+        it
+          .proceed(it.request())
+          .newBuilder()
+          .header("used-dns", it.dns.javaClass.simpleName)
+          .build()
+      }.build()
+
+  @StartStop
+  private val server = MockWebServer()
+
+  @BeforeEach
+  fun setup() {
+    // Needed because of Platform.resetForTests
+    PlatformRegistry.applicationContext = applicationContext
+  }
+
+  @Test
+  fun testDefaultRequest() {
+    server.enqueue(MockResponse(200, body = "Hello"))
+
+    val request = Request.Builder().url(server.url("/")).build()
+
+    val response = client.newCall(request).execute()
+
+    response.use {
+      assertEquals(200, response.code)
+      assertNotEquals("AndroidDns", response.header("used-dns"))
+    }
+  }
+
+  @Test
+  fun testPinnedRequest() {
+    server.enqueue(MockResponse(200, body = "Hello"))
+
+    val network = connectivityManager.activeNetwork
+
+    assumeTrue(network != null)
+
+    val request =
+      Request
+        .Builder()
+        .url(server.url("/"))
+        .tag<Network>(network)
+        .build()
+
+    val response = client.newCall(request).execute()
+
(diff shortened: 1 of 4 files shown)


okhttp-android/src/main/kotlin/okhttp3/android/AndroidAsyncDns.kt
```diff
@@ -47,6 +48,7 @@ class AndroidAsyncDns(
 
   override fun query(
     hostname: String,
+    originatingCall: Call?,
```
yschimke, line 51: Not sure if I want Call or Request here?

There are implementations of Dns that would benefit from the Request info including tags to make decisions.
Our strategy APIs that are oblivious to that end up needing awkward workarounds that interceptors can't fix.

okhttp-android/src/main/kotlin/okhttp3/android/NetworkPin.kt
```diff
@@ -0,0 +1,22 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okhttp3.android
+
+import android.net.Network
+import okhttp3.ExperimentalOkHttpApi
+
+@ExperimentalOkHttpApi
+data class NetworkPin(val network: Network?)
```
yschimke, line 22: That a request is pinned to a network is meaningful for stats, retries.

There is an alternative where we don't recreate the OkHttp client, but the Android Dns and SocketFactory are aware of this tag.

I think I'd like to move that way.

okhttp-android/src/main/kotlin/okhttp3/android/NetworkSelection.kt
```diff
@@ -0,0 +1,47 @@
...
+import androidx.annotation.RequiresApi
+import javax.net.SocketFactory
+import okhttp3.ExperimentalOkHttpApi
+import okhttp3.OkHttpClient
+import okhttp3.android.internal.NetworkPinInterceptor
+
+@ExperimentalOkHttpApi
+object NetworkSelection {
+  @RequiresApi(Build.VERSION_CODES.Q)
+  fun OkHttpClient.Builder.withNetwork(network: Network?): OkHttpClient.Builder {
```
swankjesse, line 29: I like this part of the API.

What happens if we attempt this in the main OkHttp module? We’d have a runtime dependency on `Network`, but you wouldn’t call this API if you don’t have it

okhttp-android/src/main/kotlin/okhttp3/android/internal/NetworkPinInterceptor.kt
```diff
@@ -0,0 +1,30 @@
...
+ * limitations under the License.
+ */
+package okhttp3.android.internal
+
+import android.net.Network
+import okhttp3.Interceptor
+import okhttp3.Response
+import okhttp3.android.NetworkPin
+
+internal class NetworkPinInterceptor(val network: Network) : Interceptor {
```
swankjesse, line 23: I am not sold on this

okhttp/src/main/kotlin/okhttp3/AsyncDns.kt
```diff
@@ -39,27 +39,39 @@ interface AsyncDns {
    */
   fun query(
     hostname: String,
+    originatingCall: Call?,
     callback: Callback,
   )
 
+  /** Returns a [Dns] that blocks until all async results are available. */
+  open fun asBlocking(): Dns {
```
swankjesse, line 47: This all feels like a round-peg in a square-hole . . . building an async API and only ever using it in a blocking way.

I think we should build a regular `Dns` implementation for Android’s DNS stuff, and eject AsyncDns from our API until we’re ready to take advantage of it being async

okhttp-android/src/main/kotlin/okhttp3/android/NetworkPin.kt
swankjesse: This is currently unused as far as any behavior is concerned. I don’t think we need this!

(If users want to keep stats on networks, they can do that with EventListeners and Interceptors that are also scoped to the OkHttpClient object.)

okhttp/src/main/kotlin/okhttp3/AsyncDns.kt
yschimke: I was using it here, so you can have per-network Dns.  I don't think that's possible without the async android resolver API.

okhttp-android/src/main/kotlin/okhttp3/android/NetworkSelection.kt
yschimke: I am ok with either. 

I guess ultimately we should choose either

a) we build android specific bits in main module, but you just don't use them if not needed.
b) we encourage every android app to include okhttp and okhttp-android and use those bits.

I'm not sure.

okhttp-android/src/main/kotlin/okhttp3/android/NetworkPin.kt
yschimke: Removed. I'll build a separate case, but this matches Cronet/HttpEngine functionality. 

Although not on a Request basis, which was where this was heading.

okhttp/src/main/kotlin/okhttp3/AsyncDns.kt
yschimke: Ejecting the AsyncDns API for now.  it's internal and moved to okhttp-android.

okhttp-android/src/main/kotlin/okhttp3/android/AndroidAsyncDns.kt
yschimke: not public API anymore

okhttp/api/okhttp.api
```diff
@@ -27,33 +27,6 @@ public final class okhttp3/Address {
 	public final fun url ()Lokhttp3/HttpUrl;
 }
 
-public abstract interface class okhttp3/AsyncDns {
```
yschimke, line 30: @swankjesse no longer public API anywhere

okhttp/src/main/kotlin/okhttp3/AsyncDns.kt
yschimke: Resolving as internal API.

We should address an Async DNS public API when we want to take Happy Eyeballs further, or further non blocking internals.

okhttp-android/src/main/kotlin/okhttp3/android/NetworkSelection.kt
yschimke: It's an extension method now in the okhttp core module.

yschimke:
Will rework based on https://github.com/square/okhttp/pull/8800

okhttp-android/src/main/kotlin/okhttp3/android/NetworkSelection.kt
yschimke: I ended up with it in the main module. But using Call.Decorator instead. You don't need different client instances, just to tag the request with a Network.

okhttp/src/main/kotlin/okhttp3/AsyncDns.kt
yschimke: I ended up with an internal only AndroidDns, that uses IPv4 and IPv6 in a single callback, using a CompletableFuture. It's very simple and hidden now.

yschimke:
@swankjesse can you take another look, I think it's simple and clean enough we could land as a feature.

PR #8377, opened 2024-04-20
[Draft - not for 5.0] Experiment with Proxy model and support for Https proxies

yschimke:


yschimke:
@swankjesse it's getting ugly. But java.net.Proxy seems fundamentally broken.

okhttp/src/main/kotlin/okhttp3/internal/connection/ConnectPlan.kt
```diff
@@ -256,10 +257,23 @@ class ConnectPlan(
...
-        Proxy.Type.DIRECT, Proxy.Type.HTTP -> route.address.socketFactory.createSocket()!!
-        else -> Socket(route.proxy)
+      when (proxyAddress) {
+        Proxy.Direct -> {
+          route.address.socketFactory.createSocket()!!
+        }
+        is Proxy.Http -> {
+          if (proxyAddress.server.isHttps) {
+            // TODO check
+            route.address.sslSocketFactory!!.createSocket()!!
```
yschimke, line 269: @swankjesse I think this is the meat of the fix.

PR #8383, opened 2024-04-22
Simplify the DnsOverHttps logic

yschimke:
Rely on OkHttpClient caching, instead of a cache only request.

Possibly a slight performance regression, making extra conditional requests, but it's unclear that that is not intended.


swankjesse (approved):
Love it

PR #8386, opened 2024-04-24
Allow Addresses to have a max calls per connection

peckb1:
### Overview
This adds the ability for the Address Policy to also contain a maximum number of connections per Stream.

Additionally creates a ConnectionPoolPolicy which can have default options set up for all connections in the pool.

### Context
In some environments it seems that OkHttp's HTTP/2 gets worse the more streams that are packed onto a single connection. 
By adding in the ability for users to alter their client's AddressPolicy to set a maximum number of connections per stream, they can control how much (if any) packing they want OkHttp to perform.

### DIFF
--- a/okhttp/src/main/kotlin/okhttp3/ConnectionPool.kt
+++ b/okhttp/src/main/kotlin/okhttp3/ConnectionPool.kt
@@ -148,15 +148,14 @@ class ConnectionPool internal constructor(
      * Connections will still be closed if they idle beyond the keep-alive but will be replaced.
      */
     @JvmField val minimumConcurrentCalls: Int = 0,
-    /**
-     * The maximum number of concurrent calls per connection.
-     *
-     * Set this value to 1 to disable HTTP/2 connection coalescing
-     */
-    @JvmField val maximumConcurrentCallsPerConnection: Int = Int.MAX_VALUE,
     /** How long to wait to retry pre-emptive connection attempts that fail. */
     @JvmField val backoffDelayMillis: Long = 60 * 1000,
     /** How much jitter to introduce in connection retry backoff delays */
     @JvmField val backoffJitterMillis: Int = 100,
+    /**
+     * The maximum number of concurrent calls per connection.
+     * Set this value to 1 to disable HTTP/2 connection coalescing
+     */
+    @JvmField val maximumConcurrentCallsPerConnection: Int = Int.MAX_VALUE,
   )
 }
--- a/okhttp/src/main/kotlin/okhttp3/ConnectionPool.kt
+++ b/okhttp/src/main/kotlin/okhttp3/ConnectionPool.kt
@@ -178,5 +178,11 @@ class ConnectionPool internal constructor(
      * Set this value to 1 to disable HTTP/2 connection coalescing
      */
     @JvmField val maximumConcurrentCallsPerConnection: Int = Int.MAX_VALUE,
-  )
+  ) {
+    init {
+      require(maximumConcurrentCallsPerConnection > 0) {
+        "maximumConcurrentCallsPerConnection is not allowed to be less than one (1)."
+      }
+    }
+  }
 }
--- a/okhttp/src/main/kotlin/okhttp3/ConnectionPool.kt
+++ b/okhttp/src/main/kotlin/okhttp3/ConnectionPool.kt
@@ -139,6 +139,15 @@ class ConnectionPool internal constructor(
   }
 
   /**
+   * Sets a default AddressPolicy that applies to all addresses.
+   * Overwrites any existing default policy.
+   */
+  @ExperimentalOkHttpApi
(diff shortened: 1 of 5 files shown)


okhttp/src/main/kotlin/okhttp3/internal/connection/RealConnection.kt
```diff
@@ -368,6 +384,18 @@ class RealConnection(
     }
   }
 
+  private fun getMaximumAllocationLimit(): Int {
+    // if we have not negotiated a max per streams yet, don't check for the policy override
+    val negotiatedMaxCurrentStreams = lastMaxConcurrentStreamsFromSettings ?: return 1
```
yschimke, line 389: Isn't this just defaulted to 1 for Http/1 and Http2Connection.DEFAULT_SETTINGS.getMaxConcurrentStreams() for Http/2?

should it be a lateinit and start sets to one or the other?

peckb1: maybe?

`private var lastMaxConcurrentStreamsFromSettings = 1` would avoid the early return, but then we're checking against the policy.

Would it be better to have the null and early return or check the policy every time, even if we're not in HTTP/2 land?

ean5533 (approved):
Seems reasonable to me, but I recommend waiting for a review from a more experienced maintainer

okhttp/src/main/kotlin/okhttp3/internal/connection/RealConnection.kt
```diff
@@ -26,7 +26,7 @@ import java.util.concurrent.TimeUnit.MILLISECONDS
 import java.util.concurrent.locks.ReentrantLock
 import javax.net.ssl.SSLPeerUnverifiedException
 import javax.net.ssl.SSLSocket
-import kotlin.concurrent.withLock
```
peckb1, line 29: no longer need when the references from `lock.withLock` were changed to `this.withLock`
 - though that extension function still forwards to `lock.withLock` so same concurrent method used one step further.

okhttp/src/main/kotlin/okhttp3/ConnectionPool.kt
```diff
@@ -152,5 +161,28 @@ class ConnectionPool internal constructor(
     @JvmField val backoffDelayMillis: Long = 60 * 1000,
     /** How much jitter to introduce in connection retry backoff delays */
     @JvmField val backoffJitterMillis: Int = 100,
+    /**
+     * The maximum number of concurrent calls per connection for a given address.
+     * Set this value to 1 to disable HTTP/2 connection coalescing
+     */
+    @JvmField val maximumConcurrentCallsPerConnection: Int? = null,
```
yschimke, line 168: nit: I think we use the `max` prefix elsewhere.
```suggestion
    @JvmField val maxConcurrentCallsPerConnection: Int? = null,
```

okhttp/src/main/kotlin/okhttp3/ConnectionPool.kt
```diff
@@ -152,5 +161,28 @@ class ConnectionPool internal constructor(
...
   )
+
+  /**
+   * A default policy for all connections. Overridable by [AddressPolicy]
+   * for specific addresses.
+   */
+  class ConnectionPoolPolicy(
+    /**
+     * The maximum number of concurrent calls per connection.
+     * Set this value to 1 to disable HTTP/2 connection coalescing
```
yschimke, line 178: I'm not sure "connection coalescing" is the right term here. 

From https://daniel.haxx.se/blog/2016/08/18/http2-connection-coalescing/

> Connection coalescing means that the browser tries to determine which of the remote hosts that it can reach over the same TCP connection.

okhttp/src/main/kotlin/okhttp3/ConnectionPool.kt
```diff
@@ -152,5 +161,28 @@ class ConnectionPool internal constructor(
...
     @JvmField val backoffJitterMillis: Int = 100,
+    /**
+     * The maximum number of concurrent calls per connection for a given address.
+     * Set this value to 1 to disable HTTP/2 connection coalescing
+     */
+    @JvmField val maximumConcurrentCallsPerConnection: Int? = null,
   )
+
+  /**
+   * A default policy for all connections. Overridable by [AddressPolicy]
```
yschimke, line 172: AddressPolicy is internal, so this effectively makes a public feature expose an internal one.

I'd like @swankjesse take on this.

I guess I wonder whether this indicates rather than two tiered policy (ConnectionPoolPolicy then AddressPolicy), that a single ConnectionPoolPolicy interface that can answer either by global or per address logic.

yschimke:
AddressPolicy went internal after this PR, in order to ship 5.0. 

So worth discussion (including @swankjesse)  about the public API for these features before too much rework on this PR.

yschimke:
Can you fix the merge conflicts?

Issue #8388, opened 2024-04-26
TaskRunner's concurrent performance exhibits a significant decline in case of poor networks conditions


bringyou:
Our server needs to call some external API, and the stablity of API providers is quite poor and not reliable.
And we found that during instances of network fluctuation, okHttp's internal `TaskRunner` will cause so many threads keep waiting for condition,  **finally resulting in all threads that have attempted HTTP calls becoming suspended**. We can check the jstack pic below
<img width="1375" alt="image" src="https://github.com/square/okhttp/assets/25130803/79790731-555d-4eb2-b411-606194d9a330">
After debugging we found that all `OkHttpClient` uses `TaskRunner.INSTANCE` [by the default](https://github.com/square/okhttp/blob/master/okhttp/src/main/kotlin/okhttp3/OkHttpClient.kt#L266).
And nearly every method in `TaskRunner` [acquire a instance level lock](https://github.com/square/okhttp/blob/master/okhttp/src/main/kotlin/okhttp3/internal/concurrent/TaskRunner.kt).
**In another word,  all instances of `OkHttpClient` configured as default will competing for the same lock instance** in some scenarios, especially when external's network is poor and not reliable.
Is there any plan to improve the concurrent performance of `TaskRunner`? e.g.  use atomic values rather than ReentrantLock

bringyou:
How to reproduce: calling a external API which can only serve limited tcp and takes a long time to create each tcp

yschimke:
Can you show a stack trace or the thread that holds the lock?

yschimke:
Are you sure those stacktraces aren't just the threads waiting for work to do?

Can you point to other profiling showing the slowdown?

As far as locks vs synchronised. I looked at this recently https://github.com/square/okhttp/issues/8366

It seems most detailed investigations show for contended work using Locks is the more performant option. So before switching to synchronized, do you have some links or benchmarks showing the benefit of synchronized?

swankjesse:
This particular lock is not expected to be held for more than a few cycles – every acquire reads & writes a few lock-guarded fields and then releases the lock.

But under highly-concurrent use, the fact that we’re using locking at all is likely to cause some contention.

Can you tell me how many concurrent calls are you making?

If it’s not very many (say, 10 or fewer) and you’re seeing this problem, then it’s likely we’ve got a bug in our implementation. We’re holding the lock longer than we’re expecting to.

[...]

PR #8389, opened 2024-04-27
Centralise more locks

yschimke:
Investigating https://github.com/square/okhttp/issues/8388 - trying to ensure we have a central place to reason about our locks.


yschimke:
The idea is with a central place of inlined locking primitives, I can experiment like

https://github.com/square/okhttp/blob/44dc78f53ee188268f9e8a2a57c9f2537be9b7a9/okhttp/src/main/kotlin/okhttp3/internal/connection/Locks.kt

In ways we will never land, but without a lot of refactoring to do so.

swankjesse (approved):

PR #8391, opened 2024-04-28
Start fewer threads in TaskRunner

swankjesse:
We've got a race where we'll start a thread when we need one, even if we've already started a thread. This changes TaskRunner's behavior to never add a thread if we're still waiting for a recently-added one to start running.

This is intended to reduce the number of threads contenting for the TaskRunner lock as reported in this issue:

https://github.com/square/okhttp/issues/8388

### DIFF
--- a/okhttp/src/main/kotlin/okhttp3/internal/concurrent/TaskRunner.kt
+++ b/okhttp/src/main/kotlin/okhttp3/internal/concurrent/TaskRunner.kt
@@ -52,6 +52,17 @@ class TaskRunner(
   private var coordinatorWaiting = false
   private var coordinatorWakeUpAt = 0L
 
+  /**
+   * When we need a new thread to run tasks, we call [Backend.execute]. A few microseconds later we
+   * expect a newly-started thread to call [Runnable.run]. We shouldn't request new threads until
+   * the already-requested ones are in service, otherwise we might create more threads than we need.
+   *
+   * We use [executeCallCount] and [runCallCount] to defend against starting more threads than we
+   * need. Both fields are guarded by [lock].
+   */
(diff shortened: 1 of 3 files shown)


yschimke (approved):
Nice, very readable.

 I'd curious to see if it affects the blocking on the SynchronousQueue I observed.

okhttp/src/main/kotlin/okhttp3/internal/concurrent/TaskRunner.kt
```diff
@@ -76,7 +92,7 @@ class TaskRunner(
               // If the task is crashing start another thread to service the queues.
               if (!completedNormally) {
                 lock.withLock {
```
Bennett-Lynch, line 94: Is it possible to remove the lock usage in favor of a single `AtomicInteger` which tracks outstanding threads requested? `startAnotherThread` logic could then use CAS semantics to decrement if positive and only start a thread if the value was modified.

Bennett-Lynch:
Hi @swankjesse. Would it please be possible to initiate a 4.x release that includes this fix? The latest 4.x release is still 4.12.0 from 2023-10-16. Thank you.

DPICART:
Hi, sorry to dig up an old topic. 
I am experiencing exactly the same issue (#8388) when benchmarking my API. 

My API doesn’t directly use OkHttp 4—it’s a transitive dependency from another library. Because of this, I’m unable to upgrade to version 5.

Is it possible to have a new 4.x release that includes this fix, or is there a workaround available? Thank you.

yschimke:
No, but you can upgrade to 5.x in your project. You dont need a 4.x release.

Just make sure all artifacts are the same version.

PR #8394, opened 2024-04-29
Acquire fewer locks in TaskRunner

swankjesse:
Previously each run did this:

 - acquire a lock to take a task
 - acquire a lock to finish a task
 - if crashed, acquire a lock to start a new thread

So to run 10 tasks without any crashes, we'd acquire the lock 20 times.

With this update, we do this:

 - acquire a lock to take the first task
 - acquire a lock to release task N and take task N + 1

So to run 10 tasks without any crashes, we now acquire the lock 11 times.

### DIFF
--- a/okhttp/src/main/kotlin/okhttp3/internal/concurrent/TaskRunner.kt
+++ b/okhttp/src/main/kotlin/okhttp3/internal/concurrent/TaskRunner.kt
@@ -83,9 +83,10 @@ class TaskRunner(
         try {
           while (true) {
             currentThread.name = task.name
-            val delayNanos = logger.logElapsed(task, task.queue!!) {
-              task.runOnce()
-            }
+            val delayNanos =
+              logger.logElapsed(task, task.queue!!) {
+                task.runOnce()
+              }
 
             // A task ran successfully. Update the execution state and take the next task.
             task = withLock {
--- a/okhttp/src/main/kotlin/okhttp3/internal/concurrent/TaskRunner.kt
+++ b/okhttp/src/main/kotlin/okhttp3/internal/concurrent/TaskRunner.kt
@@ -73,7 +73,8 @@ class TaskRunner(
   private val runnable: Runnable =
     object : Runnable {
       override fun run() {
-        var task: Task = withLock {
+        var task: Task =
+          withLock {
             runCallCount++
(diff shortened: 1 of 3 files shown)


okhttp/src/main/kotlin/okhttp3/internal/concurrent/TaskRunner.kt
```diff
@@ -130,25 +134,10 @@ class TaskRunner(
...
-    currentThread.name = task.name
-
-    var delayNanos = -1L
-    try {
-      delayNanos = task.runOnce()
-    } finally {
-      lock.withLock {
-        afterRun(task, delayNanos)
-      }
-      currentThread.name = oldName
```
swankjesse, line 145: We also change the thread name fewer times!

okhttp/src/test/java/okhttp3/internal/concurrent/TaskRunnerTest.kt
```diff
@@ -95,11 +95,11 @@ class TaskRunnerTest {
     assertThat(testLogHandler.takeAll()).containsExactly(
       "FINE: Q10000 scheduled after 100 µs: task",
       "FINE: Q10000 starting              : task",
-      "FINE: Q10000 run again after  50 µs: task",
       "FINE: Q10000 finished run in   0 µs: task",
+      "FINE: Q10000 run again after  50 µs: task",
```
swankjesse, line 99: I like this new order better

yschimke: +1 more logical

yschimke (approved):

yschimke:
Does this failure need an update?

```
ConnectionPoolTest > interruptStopsThread() FAILED
    org.opentest4j.AssertionFailedError: expected to be empty but was:<[Q10505]>
        at app//okhttp3.internal.connection.ConnectionPoolTest.interruptStopsThread(ConnectionPoolTest.kt:203)
```

swankjesse:
I think there’s non-determinism that I need to fix in that test, as I can’t reproduce that failure locally! Lemme do that first, then rebase this on top.

ean5533 (approved):

PR #8401, opened 2024-05-09
Allow nullable executorService parameter in Dispatcher class constructor

linleo345:
Allow nullable executorService parameter in Dispatcher.kt constructor to maintain backwards compatibility with OkHttp 3.x migration to 4.x and beyond.

For some more context: 
In my internal company's library, there were some code snippets using OkHttp 3.x that passed null into the Dispatcher constructor for the `executorService` parameter. This worked with the Java implementation of OkHttp library, but with the migration to Kotlin in 4.x and the `executorService` parameter not being nullable in Dispatcher constructor, this caused some projects to fail after upgrading from OkHttp 3.x to 4.x


linleo345:
Can I get a re-run of the failed check? Looks like it failed testopenjdk11 action due to the flaky test issue in connectionPreWarmingHttp2

swankjesse (approved):

PR #8404, opened 2024-05-10
Enabling logging for Brotli encoding

orafaaraujo:
This change aims to enable the log of the response body when using Brotli content-encoding.


orafaaraujo:
Ok, I noticed we don't have a `BrotliSource` to call to do the job for us. Closing for now.

PR #8410, opened 2024-05-14
Add a socket abstraction

ean5533:
## What

This PR adds a bunch of abstractions over `Socket` and related factories. We named the abstraction classes things like `OkioSocket` which telegraphs our intent to eventually upstream these into Okio. For now they'll just live in `okhttp3.internal`.

## Why

This is the first step toward making our test infrastructure better. This will allow us to create fake sockets that integrate with TaskFaker which yield on read/write so that other tasks can get a shot.

As a motivating use case, this will allow us to fix the flaky `ConnectPoolTest.connectionPreWarmingHttp2`.

## What's next

* Implement a bunch of fakes that play nicely with `TaskFaker` and use them in `MockHttp2Peer`, `MockWebServer`, etc
* Maybe: Use composition instead of inheritance in `OkioSslSocket`


mockwebserver/api/mockwebserver3.api
```diff
@@ -107,7 +107,7 @@ public final class mockwebserver3/MockWebServer : java/io/Closeable {
 	public final fun getProtocolNegotiationEnabled ()Z
 	public final fun getProtocols ()Ljava/util/List;
 	public final fun getRequestCount ()I
-	public final fun getServerSocketFactory ()Ljavax/net/ServerSocketFactory;
+	public final fun getServerSocketFactory ()Lokhttp3/internal/socket/OkioServerSocketFactory;
```
ean5533, line 110: How do we feel about breaking changes to MockWebServer public methods?

okhttp/api/okhttp.api
```diff
@@ -409,6 +409,7 @@ public final class okhttp3/ConnectionSpec {
 	public fun equals (Ljava/lang/Object;)Z
 	public fun hashCode ()I
 	public final fun isCompatible (Ljavax/net/ssl/SSLSocket;)Z
+	public final fun isCompatible (Lokhttp3/internal/socket/OkioSslSocket;)Z
```
ean5533, line 412: Note: I tossed a `ExperimentalOkHttpApi` on here

mockwebserver/src/main/kotlin/mockwebserver3/RecordedRequest.kt
```diff
@@ -74,9 +74,9 @@ class RecordedRequest(
   val handshakeServerNames: List<String>
 
   init {
-    if (socket is SSLSocket) {
+    if (socket is OkioSslSocket) {
       try {
-        this.handshake = socket.session.handshake()
+        this.handshake = socket.session?.handshake()
```
yschimke, line 79: Should we keep this as non-nullable, blow up if called in wrong state? or `socket.handshake` directly?

For discussion, just this is a chance to iron out the API a bit. But can be done later.

mockwebserver/src/test/java/mockwebserver3/RecordedRequestTest.kt
```diff
@@ -38,7 +39,7 @@ class RecordedRequestTest {
         localAddress = InetAddress.getByAddress("127.0.0.1", byteArrayOf(127, 0, 0, 1)),
         localPort = 80,
       )
-    val request = RecordedRequest("GET / HTTP/1.1", headers, emptyList(), 0, Buffer(), 0, socket)
+    val request = RecordedRequest("GET / HTTP/1.1", headers, emptyList(), 0, Buffer(), 0, RealOkioSocket(socket))
```
yschimke, line 42: Worth a TODO to introduce a FakeOkioSocket?

mockwebserver/src/main/kotlin/mockwebserver3/MockWebServer.kt
```diff
@@ -232,7 +234,7 @@ class MockWebServer : Closeable {
    * Serve requests with HTTPS rather than otherwise.
    */
   fun useHttps(sslSocketFactory: SSLSocketFactory) {
-    this.sslSocketFactory = sslSocketFactory
+    this.sslSocketFactory = RealOkioSslSocketFactory(sslSocketFactory)
```
yschimke, line 237: is it worth us having a more opinionated extenion method for these?

`fun SslSocketFactory.asOkioSocketFactory(): OkioSslSocketFactory = ...`

okhttp-testing-support/src/main/kotlin/okhttp3/internal/concurrent/TaskFaker.kt
```diff
@@ -331,6 +331,9 @@ class TaskFaker : Closeable {
         try {
           runnable.run()
           require(currentTask == this) { "unexpected current task: $currentTask" }
+        } catch (e: Throwable) {
+          e.printStackTrace()
```
yschimke, line 335: is this intentional to keep?

okhttp/src/main/kotlin/okhttp3/Address.kt
```diff
@@ -59,6 +61,44 @@ class Address(
    */
   @get:JvmName("proxySelector") val proxySelector: ProxySelector,
 ) {
+  /** Returns the socket factory for new connections. */
+  @get:JvmName("socketFactory")
+  val socketFactory: SocketFactory
+    get() = (okioSocketFactory as RealOkioSocketFactory).delegate
```
yschimke, line 67: I think this hurts some of the benefits of having simpler decoration, delegation of an interface.  Should we have clean ways to unwrap and wrap without knowing the concrete type?

```
interface OkioSocketFactory {
...
  val platformSocketFactory: SocketFactory?
...
}

interface OkioSslSocketFactory: OkioSocketFactory {
...
  override val platformSocketFactory: SslSocketFactory?
...
}
```

okhttp/src/main/kotlin/okhttp3/ConnectionSpec.kt
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2014 Square, Inc.
+ * Copyright (C) 2024 Square, Inc.
```
yschimke, line 2: deliberate?

okhttp/src/main/kotlin/okhttp3/ConnectionSpec.kt
```diff
@@ -148,6 +157,22 @@ class ConnectionSpec internal constructor(
...
+   * For cipher suites, at least one of the [required cipher suites][cipherSuites] must match the
+   * socket's enabled cipher suites. If there are no required cipher suites the socket must have at
+   * least one cipher suite enabled.
+   *
+   * For protocols, at least one of the [required protocols][tlsVersions] must match the socket's
+   * enabled protocols.
+   */
+  @ExperimentalOkHttpApi
+  fun isCompatible(socket: OkioSslSocket): Boolean {
+    return isCompatible((socket as RealOkioSslSocket).delegate)
```
yschimke, line 173: worth a TODO? feels like something we may want supported in the API in future.

okhttp/src/main/kotlin/okhttp3/internal/socket/OkioSslSocketFactory.kt
```diff
@@ -0,0 +1,53 @@
...
+
+import javax.net.ssl.SSLSocket
+import javax.net.ssl.SSLSocketFactory
+
+interface OkioSslSocketFactory {
+  fun createSocket(socket: OkioSocket): OkioSslSocket
+
+  fun createSocket(
+    socket: OkioSocket,
+    host: String,
```
yschimke, line 27: Should we combine these to InetSocketAddress?

okhttp/src/main/kotlin/okhttp3/internal/connection/ConnectPlan.kt
```diff
@@ -393,17 +398,17 @@ class ConnectPlan(
       // Success! Save the handshake and the ALPN protocol.
       val maybeProtocol =
         if (connectionSpec.supportsTlsExtensions) {
-          Platform.get().getSelectedProtocol(sslSocket)
+          Platform.get().getSelectedProtocol(sslSocket.delegate)
```
yschimke, line 401: Probably a TODO, but if you look at SocketAdapter, ConscryptSocketAdapter, Android10SocketAdapter, these are ideally something that would be hidden behind the OkioSslSocketFactory API, not through Platform.

okhttp/src/main/kotlin/okhttp3/internal/connection/RealConnection.kt
```diff
@@ -313,7 +316,9 @@ class RealConnection(
     rawSocket?.closeQuietly()
   }
 
-  override fun socket(): Socket = socket!!
+  override fun socket(): Socket =
+    (socket as? RealOkioSslSocket)?.delegate
```
yschimke, line 320: I think this hints at being able to unwrap an OkioSocket to an optional platform `Socket?`

okhttp/src/main/kotlin/okhttp3/internal/socket/OkioSocket.kt
```diff
@@ -0,0 +1,103 @@
...
+
+  var soTimeout: Int
+  val isClosed: Boolean
+  val isInputShutdown: Boolean
+  val isOutputShutdown: Boolean
+  val peerName: String
+  val localPort: Int
+  val inetAddress: InetAddress?
+  val localAddress: InetAddress
+  val remoteSocketAddress: SocketAddress?
```
yschimke, line 43: Can we simplify these into two InetSocketAddress vals?

okhttp/src/main/kotlin/okhttp3/internal/socket/OkioSocket.kt
```diff
@@ -0,0 +1,103 @@
...
+
+interface OkioSocket : Closeable {
+  val source: BufferedSource
+  val sink: BufferedSink
+
+  var soTimeout: Int
+  val isClosed: Boolean
+  val isInputShutdown: Boolean
+  val isOutputShutdown: Boolean
+  val peerName: String
```
yschimke, line 39: I think this was only an internal function, do we want to make it top level API?

okhttp/src/main/kotlin/okhttp3/internal/socket/OkioSocket.kt
```diff
@@ -0,0 +1,103 @@
...
+import okio.Closeable
+import okio.buffer
+import okio.sink
+import okio.source
+
+interface OkioSocket : Closeable {
+  val source: BufferedSource
+  val sink: BufferedSink
+
+  var soTimeout: Int
```
yschimke, line 35: Should we avoid this API here? Does it map to Okio timeout? or should we leave it for code that knows it's a platform socket?

okhttp/src/main/kotlin/okhttp3/internal/socket/OkioSocketFactory.kt
```diff
@@ -0,0 +1,45 @@
...
+import java.net.Proxy
+import java.net.Socket
+import javax.net.SocketFactory
+
+interface OkioSocketFactory {
+  fun createSocket(): OkioSocket
+
+  fun createSocket(proxy: Proxy): OkioSocket
+
+  fun createSocket(
```
yschimke, line 28: Should remove this and assume clients follow OkHttp's use of `createSocket()` + `connect()`

okhttp/src/main/kotlin/okhttp3/internal/socket/OkioSocketFactory.kt
```diff
@@ -0,0 +1,45 @@
...
+package okhttp3.internal.socket
+
+import java.net.Proxy
+import java.net.Socket
+import javax.net.SocketFactory
+
+interface OkioSocketFactory {
+  fun createSocket(): OkioSocket
+
+  fun createSocket(proxy: Proxy): OkioSocket
```
yschimke, line 26: I wonder if we want this here? Could we say that a proxy is a different OkioSocketFactory? What would we lose here?

okhttp/src/main/kotlin/okhttp3/internal/socket/OkioSocket.kt
```diff
@@ -0,0 +1,103 @@
...
+  val remoteSocketAddress: SocketAddress?
+
+  fun connect(address: InetSocketAddress)
+
+  fun connect(
+    address: InetSocketAddress,
+    connectTimeout: Int,
+  )
+
+  fun shutdownOutput()
```
yschimke, line 52: Would we lose much by expecting clients to do `sink.close()` ?

okhttp/src/main/kotlin/okhttp3/internal/socket/OkioSocket.kt
```diff
@@ -0,0 +1,103 @@
...
+  val localPort: Int
+  val inetAddress: InetAddress?
+  val localAddress: InetAddress
+  val remoteSocketAddress: SocketAddress?
+
+  fun connect(address: InetSocketAddress)
+
+  fun connect(
+    address: InetSocketAddress,
+    connectTimeout: Int,
```
yschimke, line 49: Should connectTimeout be a property of the OkioSocket instead? or handled via Okio Timeout?

okhttp/src/main/kotlin/okhttp3/internal/socket/OkioSslSocketFactory.kt
yschimke: Also looking at UnixDomainSocketFactory, I wonder whether the type of Address should be more flexible. For UnixDomainSocketFactory, it's really a File, or Path, and we have a bunch of lying methods.

Not sure this is possible...

okhttp/src/test/java/okhttp3/URLConnectionTest.kt
```diff
@@ -2937,12 +2942,14 @@ class URLConnectionTest {
...
     val socketBufferSize = 4 * 1024
     server.serverSocketFactory =
-      object : DelegatingServerSocketFactory(getDefault()) {
-        override fun configureServerSocket(serverSocket: ServerSocket): ServerSocket {
-          serverSocket.receiveBufferSize = socketBufferSize
-          return serverSocket
-        }
-      }
+      RealOkioServerSocketFactory(
+        object : DelegatingServerSocketFactory(getDefault()) {
```
yschimke, line 2946: possibly a TODO? It would be nice to not need `DelegatingServerSocketFactory` and friends because of your beautiful new API.

yschimke: I'm not suggesting interceptors :) More that decorating could be made simpler?

samples/unixdomainsockets/src/main/java/okhttp3/unixdomainsockets/ClientAndServer.java
```diff
@@ -34,7 +35,9 @@ public void run() throws Exception {
     socketFile.delete(); // Clean up from previous runs.
 
     MockWebServer server = new MockWebServer();
-    server.setServerSocketFactory(new UnixDomainServerSocketFactory(socketFile));
+    RealOkioServerSocketFactory serverSocketFactory =
+      new RealOkioServerSocketFactory(new UnixDomainServerSocketFactory(socketFile));
```
yschimke, line 39: possibly worth a TODO? Feels like this new API supporting other socket types would be a win?

yschimke (commented):
I like it, but it's disruptive enough including for legacy 4.x APIs that you probably want @swankjesse input before addressing any of my suggestions.

mockwebserver/src/main/kotlin/mockwebserver3/MockWebServer.kt
```diff
@@ -107,7 +109,7 @@ class MockWebServer : Closeable {
   private val taskRunner = TaskRunner(taskRunnerBackend)
   private val requestQueue = LinkedBlockingQueue<RecordedRequest>()
   private val openClientSockets =
-    Collections.newSetFromMap(ConcurrentHashMap<Socket, Boolean>())
+    Collections.newSetFromMap(ConcurrentHashMap<OkioSocket, Boolean>())
```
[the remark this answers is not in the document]
ean5533: This is a great question. My instinct is to document that equals/hashcode are referenced-based, because I think that's the only reasonable way to think about socket equality. I think I would be surprised if sockets were considered equal based on anything than their unique state, at least by default.

okhttp/src/main/kotlin/okhttp3/internal/socket/OkioServerSocketFactory.kt
```diff
@@ -0,0 +1,32 @@
...
+
+package okhttp3.internal.socket
+
+import javax.net.ServerSocketFactory
+
+interface OkioServerSocketFactory {
+  fun newServerSocket(): OkioServerSocket
+}
+
+class RealOkioServerSocketFactory(
```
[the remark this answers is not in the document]
ean5533: > Also the name OkioServerSocketFactory is awkward once you later drop Okio, should we think about the name already?

I don't have strong feelings about this, but I think the `Okio*` names might be ideal specifically because it disambiguates them from the stdlib equivalents. I can't think of any other name I'd pick. That said, maybe a better name will reveal itself as we flesh them out?

mockwebserver/src/main/kotlin/mockwebserver3/RecordedRequest.kt
ean5533: Hmm, yeah. Looking at places that read `handshake`, basically everyone expects it to be non-null. From a very quick spot check, it looks like `socket.session` will effectively always be non-null, and so will `socket.session.handshake()`. So I think I'd be in favor of tightening up the constraints here. But probably in a second PR.

okhttp/src/main/kotlin/okhttp3/ConnectionSpec.kt
ean5533: No, that's weird. I set up automatic file headers in IJ, it must have decided to update some existing files sometimes... 🤔 

Anyway, will revert

mockwebserver/src/main/kotlin/mockwebserver3/MockWebServer.kt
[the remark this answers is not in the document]
yschimke: Maybe, although a socket is a 4-tuple of host:port, host:port.

And equality could break with decoration used this way.

Probably documenting reference equality is right, just bouncing ideas.

okhttp/src/main/kotlin/okhttp3/internal/socket/OkioServerSocketFactory.kt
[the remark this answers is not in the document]
yschimke: Yep, works for me.

okhttp-testing-support/src/main/kotlin/okhttp3/internal/concurrent/TaskFaker.kt
yschimke: Maybe a TaskFaker UncaughtExceptionHandler?

okhttp/src/main/kotlin/okhttp3/Address.kt
yschimke: I agree it should be possible, but I think a way to break out when the native implementation has features you need without assuming Real* will come in handy.

And we demonstrate we want it here in a few places.

okhttp/src/main/kotlin/okhttp3/internal/connection/ConnectPlan.kt
yschimke: I think this socket adapter experience was why I kept suggestion an okio socket. This change looks like a good start.

okhttp/src/main/kotlin/okhttp3/internal/socket/OkioSocket.kt
ean5533: Good call. At most this should just be another extension function `OkioSocket.peerName()`. However I think we don't need it at all, we've abstracted away most of the places it was used. Will remove.

okhttp/src/main/kotlin/okhttp3/internal/socket/OkioSocket.kt
ean5533: Hmm. In theory any implementation of `OkioSocket` could enforce this timeout, even without a platform socket under the hood, by modifying whatever is being buffered by `source`. But I'm not sure if that makes for a good API or not 🤔

okhttp/src/main/kotlin/okhttp3/internal/socket/OkioSocketFactory.kt
ean5533: 👍 I'm not opposed to this

yschimke:
Closing for now - needs major rework and rebase onto Okio Socket.

Issue #7553, opened 2022-12-14
routeDatabase.failed is not called when we fail to establish a connection


moshegood:
When we have two ways of connecting to a destination, say IPv4 and IPv6.
If the IPv4 connection then fails (say EOF due to server dying), it gets marked in the routeDatabase as failed.
If the IPv6 connection then fails to connect (say the server doesn't support IPv6), it does not get marked in the routeDatabase as failed.
This means that ALL future connection attempts will be over IPv6. Which will all fail. Even though the IPv4 connection has come back up and is now stable.

What we should do:
also mark routes where we fail to establish a connection as failed. Then the IPv6 (in our example) route would also be marked as failed, and given that all routes have failed, they will all be tried again. Including the IPv4 connection which would then work when the server has recovered.

yschimke:
This surprises me.  Is this on 4.10? It's probably changed quite substantially in 5.0 because of the happy eyeballs support.

Can you make a reproduction? I'd expect it just to postpone the IPv4 in RouteSelector, try IPv6 and then IPv4 again.

We should definitely have more tests for that, I'll try to add one to CallTest in 5.0 to confirm my understanding.

moshegood:
This is on 4.10.
I can reproduce. The error is `java.net.ConnectException: Failed to connect to localhost/[0:0:0:0:0:0:0:1]:8080` but this route doesn't show up in the failed list.

peacemaker123456:
@yschimke any progress here? Or a possible workaround?
I am coming from https://github.com/square/okhttp/issues/7696
I was relying on the onFailed callback of the proxySelector to update my proxy list. 
Kind regards

yschimke:
I believe fixed

PR #8415, opened 2024-05-21
Call proxy selector on failure

yschimke:
For discussion, I looked into this previously but can't remember or find why I didn't fix. It's marked for 5.0 release.

Report any route connect failures to the proxy selector if we used it and got a proxy.

https://github.com/square/okhttp/issues/7553


swankjesse (approved):
Nice catch

PR #8420, opened 2024-05-24
update javadoc for `Response#body`

Endeavour233:
Previous doc shows that body is always null for cacheResponse, networkResponse and priorResponse. It's confusing since the body property can't be null. This pull request updates the doc to show that we'll get an unreadable body from the listed responses.


yschimke:
@Endeavour233 Thanks

PR #8428, opened 2024-06-02
chore: update a deprecated usage of resume function Kotlinx coroutines in okhttp-coroutines

ellet0:
It seems that the project uses a modern version of Kotlin

https://github.com/square/okhttp/blob/54238b4c713080c3fd32fb1a070fb5d6814c9a09/gradle/libs.versions.toml#L13

Which already supports using the new function

image

It seems the old function was experimental before Kotlin 1.9.0 which might be one reason why `executeAsync` requires `@ExperimentalCoroutinesApi` annotation

### DIFF
--- a/okhttp-coroutines/src/main/kotlin/okhttp3/coroutines/ExecuteAsync.kt
+++ b/okhttp-coroutines/src/main/kotlin/okhttp3/coroutines/ExecuteAsync.kt
@@ -45,7 +45,7 @@ suspend fun Call.executeAsync(): Response =
           call: Call,
           response: Response,
         ) {
-          continuation.resume(response) {
+          continuation.resume(response) { _, _, _ ->
             response.closeQuietly()
           }
         }


JakeWharton:
Should we remove the annotation too then?

ellet0:
> Should we remove the annotation too then?

The version of the okhttp-coroutines module says it's in alpha ([5.0.0-alpha.11](https://central.sonatype.com/artifact/com.squareup.okhttp3/okhttp-coroutines/5.0.0-alpha.11))

it works just like expected for me, I'm not sure what's missing and why this annotation is used in the first place, my guess is it was required on older versions of Kotlinx coroutines, see this issue: https://github.com/Kotlin/kotlinx.coroutines/issues/4088

It was created by a developer from Square (the same community/organization that developed OkHttp) and it seems to be fixed, I'm still not sure why it's not removed, maybe they use the experimental annotation from Kotlinx coroutines as a way to indicate the functionality is not production ready, subject to change, miss something, have bugs or they have some other reasons

For me, it was working perfectly fine, the code seemed to compile successfully without annotating the function and requiring to mark the functions that are using `executeAsync()`

**Not related issue**: using this module makes the bundle size a little bit bigger (compared to the size of the current project I'm working on), there might be unused dependencies or some libraries that are meant to be used in tests only or some other imports, I tried to use this extension function by copying the code to a blank project vs using the extension function from [okhttp-coroutines](https://github.com/square/okhttp/tree/master/okhttp-coroutines)
image

The size is smaller when copying the code of the extension function inside this module, since now the extension function `executeAsync` is stable, it's possible to publish the `okhttp` with `okhttp-coroutines` with the same version

yschimke:
Doesn't this require bumping to the 1.9.0-RC of Kotlinx coroutines

https://github.com/Kotlin/kotlinx.coroutines/releases/tag/1.9.0-RC

> Promoted CancellableContinuation.resume with an onCancellation lambda to stable, providing extra arguments to the lambda

ellet0:
> Doesn't this require bumping to the 1.9.0-RC of Kotlinx coroutines
> 
> https://github.com/Kotlin/kotlinx.coroutines/releases/tag/1.9.0-RC
> 
> > Promoted CancellableContinuation.resume with an onCancellation lambda to stable, providing extra arguments to the lambda
> 

It does, though this is the minimum supported version by Kotlin 2.0.0. Most projects are already working on migrating to use K2.

Otherwise, we would need to wait for a while before increasing the minimum version.

This change might need to be done to OkHttp 5.0.0 instead. Closing this PR.

yschimke:
We should follow Okio, even if not strictly neccesary.  I think they are on 1.9.21

Pending PR https://github.com/square/okio/pull/1483

PR #8430, opened 2024-06-04
docs(readme): include a note about using okhttp-coroutines module

ellet0:
The `okhttp-coroutines` module is published with a different version than OkHttp and its other modules and won't work with the OkHttp BOM platform, The current `README.md` might suggest that this module is under early development and has not yet been published.

This PR updates the `README.md` file in the `okhttp-coroutines`, I noticed this module is not yet mentioned in `README.md` of the root project folder or at OkHttp docs website, except this link: https://square.github.io/okhttp/5.x/okhttp-coroutines/index.html

At the moment `5.0.0-alpha.14` is hardcoded in this PR, this repository might not have a script that will replace the hardcoded value in this folder yet. Let me know if I should replace it with something else like `<latest-version-here>` and include a GitHub badge that will show the latest version in Maven Central instead.

### DIFF
--- a/okhttp-coroutines/README.md
+++ b/okhttp-coroutines/README.md
@@ -21,3 +21,13 @@ Cancellation if implemented sensibly in both directions.
 Cancelling a coroutine scope, will cancel the call.
 Cancelling a call, will throw a CancellationException
 but not cancel the scope if caught.
+
+## Releases
+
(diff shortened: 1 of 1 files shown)


JakeWharton:
The coroutines module is published with all of the other modules and they all have the same version. This module is new for 5.x, which means there is no stable 4.x version. But you can see all of the modules are published with this 5.0 alpha here: https://search.maven.org/search?q=g:com.squareup.okhttp3

Issue #8451, opened 2024-06-22
Reenable connectionPreWarmingHttp2


yschimke:
Disabling test for now, as it's making it hard to see CI results.

We should reenable.

Blocked on https://github.com/square/okhttp/pull/8410

PR #8452, opened 2024-06-22
Disable connectionPreWarmingHttp2 for now

yschimke:
Want to clear out a bunch of dependency PRs which will be less noisy with this disabled for now

Tracking reenabling in https://github.com/square/okhttp/issues/8451


yschimke:
Landing to make some progress on housekeeping

PR #8503, opened 2024-08-07
Update isHealthy check to be safer

seljabali:
Fixes https://github.com/square/okhttp/issues/8447

### DIFF
--- a/okhttp/src/main/kotlin/okhttp3/internal/connection/RealConnection.kt
+++ b/okhttp/src/main/kotlin/okhttp3/internal/connection/RealConnection.kt
@@ -321,9 +321,9 @@ class RealConnection(
 
     val nowNs = System.nanoTime()
 
-    val rawSocket = this.rawSocket!!
-    val socket = this.socket!!
-    val source = this.source!!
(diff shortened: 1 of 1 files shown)


prbprbprb:
But maybe masks an underlying bug? Those values shouldn't be null at this point.

seljabali:
> But maybe masks an underlying bug? Those values shouldn't be null at this point.

Question is: Should isHealthy() return false or throw an exception when these values are null? 

At face value, it should return false for this method, however throw an exception elsewhere when these values are non-null.

prbprbprb:
My reading of the code is that these are preconditions, hence the use of `!!` in Kotlin whereas in Java we would use something like `Objects.requireNotNull()`.  Because if the preconditions aren't met, something has already failed previously.

Caveat: I'm not an okhttp expert but happened to be looking at a similar exception in okhttp2 earlier today.

yschimke:
Shouldn't be possible after https://github.com/square/okhttp/pull/8514/files, or at least this can't be applied.

That said, it's possible I just moved a logic bug up higher, but this fix isn't the right one anymore.

Issue #8472, opened 2024-07-11
Incorrect retries on connection that still succeeded


speekha:
We've wanted to upgrade OkHttp v4.x for quite a while now, but we can't because of some change regarding the connection pool and its impact on the retry policy. In some cases, we can see a connection going to a server and being answered (we use proxyman to track network flows, and we can see both request and response going through), but for some reason, OkHttp decides to silently drop that connection even though it was successful and retry it (not sure if the connection was deemed stale and thus closed). So we see a second request going through in proxyman (which ends up with an error from the server in some cases because it was already processed the first time, authentication for instance), and in our code, the only result that we receive is that second one (the server error).

We've encountered the problem with all v 4.x versions that we've tested so far, but never had the problem with v3.x (currently 3.14.9). The exact conditions are hard to reproduce, but if you disable the retryOnConnectionFailure, you end up with an IOException. This means that for some reason, the socket was closed by OkHttp v4.x, in a way that did not happen in OkHttp v3.x, leading to this exception:

Error response
java.io.IOException: unexpected end of stream on https://www.xxxxxxxxxx.fr/...
    at okhttp3.internal.http1.Http1ExchangeCodec.readResponseHeaders(Http1ExchangeCodec.kt:210)
    at okhttp3.internal.connection.Exchange.readResponseHeaders(Exchange.kt:110)
    at okhttp3.internal.http.CallServerInterceptor.intercept(CallServerInterceptor.kt:93)
    at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
    at okhttp3.internal.connection.ConnectInterceptor.intercept(ConnectInterceptor.kt:34)
    at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
    at okhttp3.internal.cache.CacheInterceptor.intercept(CacheInterceptor.kt:95)
    at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
    at okhttp3.internal.http.BridgeInterceptor.intercept(BridgeInterceptor.kt:83)
    at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
    at okhttp3.internal.http.RetryAndFollowUpInterceptor.intercept(RetryAndFollowUpInterceptor.kt:76)
    at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
[...]

PR #8504, opened 2024-08-10
Removing the limitation of 256KB for header size

gahfy:
This allows users to not be restricted to the limitation of 256KB for the size of the header, in order to solve #8472 :

### Python3 server code

```python3
import socket
import random
import string

def generate_random_string(length):
    return ''.join(random.choices(string.ascii_letters + string.digits, k=length))

random_string = ''.join(random.choices(string.ascii_letters + string.digits, k=256 * 1024 - 48))

# Création d'un socket TCP
with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as server_socket:
    server_socket.bind(('127.0.0.1', 8080))
    server_socket.listen()

    while True:
        conn, addr = server_socket.accept()
        with conn:
            response = (
                "HTTP/1.1 200 OK\n"
                "Content-Type: text/plain\n"
                f"MyHeader:{random_string}\n"
                "Content-Length: 14\n"
                "\n"
                "Erreur de requête"
            )
            conn.sendall(response.encode('utf-8'))
```

## Kotlin Client code after/before the PR (Comment/Uncomment lines 10/11)

```kotlin
import okhttp3.OkHttpClient;
import okhttp3.Request;
import okhttp3.Response;
import okhttp3.ResponseBody;
import okhttp3.internal.http1.HeadersReader;

public class OkHttpTest {
  private static final String ENDPOINT = "http://127.0.0.1:8080";

  public static void main(String... args) {
    OkHttpClient client = new OkHttpClient.Builder().build();
    //OkHttpClient client = new OkHttpClient.Builder().headerSizeLimit(HeadersReader.HEADER_NO_LIMIT).build();

    try {
        .url(ENDPOINT)
        .build();
      try (Response response = client.newCall(request).execute()) {
        ResponseBody body = response.body();
        System.out.println(body.string());
      }
    } catch(Exception e) {
      e.printStackTrace();
    }
  }
}
```

Before, you will get an `EOFException`, after, everything works fine

### DIFF
--- a/okhttp/src/main/kotlin/okhttp3/OkHttpClient.kt
+++ b/okhttp/src/main/kotlin/okhttp3/OkHttpClient.kt
@@ -1394,7 +1394,7 @@ open class OkHttpClient internal constructor(
     /**
      * Sets the maximum length (in characters, excluding line breaks) of the header.
      *
-     * Set to -1 to allow any size fot the header.
+     * Set to [HeadersReader.HEADER_NO_LIMIT] to allow any size fot the header.
      *
      * [HeadersReader.HEADER_LIMIT] by default.
      */
--- a/okhttp/src/main/kotlin/okhttp3/OkHttpClient.kt
+++ b/okhttp/src/main/kotlin/okhttp3/OkHttpClient.kt
@@ -37,6 +37,7 @@ import okhttp3.internal.checkDuration
 import okhttp3.internal.concurrent.TaskRunner
 import okhttp3.internal.connection.RealCall
 import okhttp3.internal.connection.RouteDatabase
+import okhttp3.internal.http1.HeadersReader
 import okhttp3.internal.immutableListOf
 import okhttp3.internal.platform.Platform
 import okhttp3.internal.proxy.NullProxySelector
@@ -264,6 +265,7 @@ open class OkHttpClient internal constructor(
 
   internal val routeDatabase: RouteDatabase = builder.routeDatabase ?: RouteDatabase()
   internal val taskRunner: TaskRunner = builder.taskRunner ?: TaskRunner.INSTANCE
+  internal val headerSizeLimit: Long = builder.headerSizeLimit
 
   @get:JvmName("connectionPool")
   val connectionPool: ConnectionPool =
@@ -626,6 +628,7 @@ open class OkHttpClient internal constructor(
     internal var minWebSocketMessageToCompress = RealWebSocket.DEFAULT_MINIMUM_DEFLATE_SIZE
     internal var routeDatabase: RouteDatabase? = null
     internal var taskRunner: TaskRunner? = null
+    internal var headerSizeLimit: Long = HeadersReader.HEADER_LIMIT
 
     internal constructor(okHttpClient: OkHttpClient) : this() {
       this.dispatcher = okHttpClient.dispatcher
@@ -661,6 +664,7 @@ open class OkHttpClient internal constructor(
       this.minWebSocketMessageToCompress = okHttpClient.minWebSocketMessageToCompress
       this.routeDatabase = okHttpClient.routeDatabase
       this.taskRunner = okHttpClient.taskRunner
+      this.headerSizeLimit = okHttpClient.headerSizeLimit
     }
 
     /**
@@ -1385,6 +1389,22 @@ open class OkHttpClient internal constructor(
         }
 
         this.minWebSocketMessageToCompress = bytes
+      }
+
+    /**
+     * Sets the maximum length (in characters, excluding line breaks) of the header.
+     *
+     * Set to -1 to allow any size fot the header.
+     *
+     * [HeadersReader.HEADER_LIMIT] by default.
+     */
+    fun headerSizeLimit(characters: Long) =
+      apply {
+        require(characters >= -1) {
+          "headerSizeLimit must me either positive or equals to -1: $characters"
+        }
+
+        this.headerSizeLimit = characters
       }
 
     fun build(): OkHttpClient = OkHttpClient(this)
--- a/okhttp/src/main/kotlin/okhttp3/OkHttpClient.kt
+++ b/okhttp/src/main/kotlin/okhttp3/OkHttpClient.kt
@@ -1401,7 +1401,7 @@ open class OkHttpClient internal constructor(
     fun headerSizeLimit(characters: Long) =
       apply {
         require(characters >= -1) {
-          "headerSizeLimit must me either positive or equals to -1: $characters"
+          "headerSizeLimit must be either positive or equals to -1: $characters"
         }
 
         this.headerSizeLimit = characters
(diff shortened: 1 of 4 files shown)


okhttp/src/main/kotlin/okhttp3/OkHttpClient.kt
```diff
@@ -1387,6 +1391,22 @@ open class OkHttpClient internal constructor(
...
+     * Sets the maximum length (in characters, excluding line breaks) of the header.
+     *
+     * Set to [HeadersReader.HEADER_NO_LIMIT] to allow any size fot the header.
+     *
+     * [HeadersReader.HEADER_LIMIT] by default.
+     */
+    fun headerSizeLimit(characters: Long) =
+      apply {
+        require(characters >= -1) {
+          "headerSizeLimit must me either positive or equals to -1: $characters"
```
yschimke, line 1404: Typo: me -> be

yschimke (commented):
Seems reasonable. Let me confirm with specs and also my past decisions.

okhttp/src/main/kotlin/okhttp3/OkHttpClient.kt
gahfy: Thanks for noticing. Done

okhttp/src/main/kotlin/okhttp3/internal/http1/HeadersReader.kt
```diff
@@ -19,15 +19,26 @@ import okhttp3.Headers
...
+  val source: BufferedSource,
+  private var headerLimit: Long = HEADER_LIMIT
+) {
 
   /** Read a single line counted against the header size limit. */
   fun readLine(): String {
-    val line = source.readUtf8LineStrict(headerLimit)
-    headerLimit -= line.length.toLong()
+    val line = if(headerLimit == HEADER_NO_LIMIT)
+        source.readUtf8LineStrict()
```
yschimke, line 32: Given we now have two codepaths here, and it was already confusing. Should we catch and re throw in all cases with context such as remaining header length/"NO_LIMIT"?

yschimke:
In https://github.com/square/okhttp/pull/3613 we removed a system property lookup. Not sure why we didn't move to a config.

yschimke:
My one concern about this is that the current default is way above internet infrastructure. So likely to break on proxies or other clients.

okhttp/src/main/kotlin/okhttp3/internal/http1/HeadersReader.kt
```diff
@@ -19,15 +19,26 @@ import okhttp3.Headers
...
+      else
+        source.readUtf8LineStrict(headerLimit)
+    if(headerLimit != HEADER_NO_LIMIT) {
+      if(line.length.toLong() > headerLimit) {
+        headerLimit = HEADER_NO_LIMIT - 1L
```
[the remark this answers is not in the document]
gahfy: Yes, and I'm not sure about the behavior then of `source.readUtf8LineStrict(headerLimit)` with a negative number in the case of an empty line, which wouldn't break the rule limit. So putting it strictly negative will handle that scenario.

okhttp/src/main/kotlin/okhttp3/internal/http1/HeadersReader.kt
gahfy: I'm not sure to understant what you mean here ? If there is no limit and we do `source.readUtf8LineStrict(headerLimit)`, it will throw an exception, and the next `source.readUtf8LineStrict()` will be run on the next line if I understant it well.
So how would you perform the completion of the line in the `catch` clause ?

gahfy:
> My one concern about this is that the current default is way above internet infrastructure. So likely to break on proxies or other clients.

I agree with that, that is why in the #8472 issue, I suggested also to not change anything, and just update the documentation. Btw, very long headers do not break any RFC 2616 rule. Maybe making sure that we can handle any response which does not break any rule would make the library also usable by people with that weird server response.

okhttp/src/main/kotlin/okhttp3/internal/http1/HeadersReader.kt
yschimke: It's probably worth catching and rethrowing and explaining issue so it's clear why

- newline not found within limit 
- newline now found by end of stream

gahfy: Not sure to really understand what you mean. Can you write 1 or 2 line of pseudo code, to make sure I understand it well (no need to make something correct, it does not need to compile either, it's just in order to get an idea of what you mean)

yschimke: I'm mainly concerned about the confusion with this limit and the limit used by Okio.

https://github.com/square/okio/blob/597a65527c987d9f221dd5607c49e49f934df2de/okio/src/jvmTest/kotlin/okio/ReadUtf8LineTest.kt#L66

```
  throw EOFException(
    "\\n not found: limit=" + minOf(buffer.size, limit) +
      " content=" + data.readByteString().hex() + '…'.toString(),
  )
```

So if I've set the header length limit to 1024, but I might get 0, 10, or 1024 as the limit in the exception message. I'd be confused if I saw a different limit here.

Maybe something like

```
    val line = try {
      if (headerLimit == HEADER_NO_LIMIT) source.readUtf8LineStrict()
      else source.readUtf8LineStrict(headerLimit)
    } catch (eofe: EOFException) {
      throw EOFException(buildString {
        append("Could not read a complete header")
        if (headerLimit != HEADER_NO_LIMIT) {
          append(" within $headerLimit bytes")
        }
      }).apply {
        initCause(eofe)
      }
    }
``

But I'm open to suggestions.

yschimke:
I don't know about the documentation vs config choice.

But if we want to move forward with this, could you add some tests of header limits that show the messages that will be seen.

swankjesse:
Obviously we could just bump this limit!

Unfortunately the rest of our implementation assumes headers are small enough to keep in memory. It assumes we don’t need to worry too much about the performance implications of loading headers in memory for the HTTP response cache, or when echoing headers back to the server as cookies or redirects. To support very large headers efficiently, we’d have to potentially redesign some of this stuff. It also potentially creates a DoS attack vector.

I’m sorry that this limit caused you so much grief, and I’m sorry that it isn’t particularly visible in our documentation. I’m especially sorry for being an absentee maintainer and wasting your time both reporting this bug and contributing an excellent fix.

But I don’t want this to be configurable. If you have an HTTP server that returns more than 256 KiB of response headers, it’s broken in other major user agents like [Chrome](https://source.chromium.org/chromium/chromium/src/+/main:net/http/http_stream_parser.h;l=159?q=ERR_RESPONSE_HEADERS_TOO_BIG&ss=chromium
). You’re also likely to run into problems elsewhere.

Please work around this problem by moving this data into the response body.

Issue #8512, opened 2024-08-17
Provide Option to Disable Thread Name Modification in OkHttp


HollisMeynell:
We heavily rely on thread names for tracking purposes within our logging system. By default, OkHttp changes the thread name to include the URL of the request being made. While this is beneficial for debugging, it introduces potential privacy concerns. The URLs being accessed can contain sensitive information, and including these in thread names may inadvertently leak this information into logs, which are often stored and monitored for long periods.

We would greatly appreciate it if OkHttp could introduce a configuration option that allows developers to disable this behavior. This would allow us to maintain the original thread names and avoid potential privacy risks associated with exposing URLs in logs.

Proposed Solution: add logic [here](https://github.com/square/okhttp/blob/babbf643cdf28df52022868996d2af9e91fc1c77/okhttp/src/main/kotlin/okhttp3/internal/connection/RealCall.kt#L528)

yschimke:
Seems like a valid request.  I'm not sure the right way to achieve this.

It feels overkill for public API on OkHttpClient or Call.

cc @swankjesse any suggestions?

efemoney:
You can supply your own dispatcher & therefore your own Executor that is just an exact copy of what OkHttp does today but with the renamed thread. 

This is what we do for instance.

swankjesse:
OkHttp redacts the URLs in its thread names . . . it’ll contain the hostname but not the path or query. Is that still a problem?

PR #8513, opened 2024-08-17
Centralise thread name logic

yschimke:
Use a single utility function in main code for setting the thread name, in case we want to intercept this.

Possibly before any fix for https://github.com/square/okhttp/issues/8512


HollisMeynell:
It seems that just adding an `if` statement should be enough, and it shouldn't have any other impact,if needed, I can submit a pull request to add this configuration.

swankjesse (approved):

PR #1901, opened 2015-10-07
Add support for WebDAV HTTP methods (RFC 4918)

rfc2822:
The tests are still successful. Is there anything else required to get WebDAV support into okhttp?

PR #8550, opened 2024-10-08
Add QUERY method support

### LINKED ISSUE #8380: QUERY support
https://twitter.com/jonchurch/status/1781781750995181847?t=qH-g4VFbBrulPM_JlyZl5A&s=19

For the http get with body community

desiderantes:
Fixes #8380.

This PR adds a new public method, `query(RequestBody)`, to perform a request using the QUERY method. It also enables cache support for QUERY requests.


yschimke:
It seems like it's still a draft https://datatracker.ietf.org/doc/draft-ietf-httpbis-safe-method-w-body/

But overall, we've added verbs for WebDav https://github.com/square/okhttp/pull/1901

yschimke:
@swankjesse any objection?

desiderantes:
Just wanted to note that Node already shipped [support for it](https://github.com/nodejs/node/issues/51562)

swankjesse (commented):
We need to validate that GET and QUERY calls to the same URL are cached independently

okhttp/src/main/kotlin/okhttp3/Cache.kt
```diff
@@ -228,9 +228,10 @@ class Cache internal constructor(
       return null
     }
 
-    if (requestMethod != "GET") {
-      // Don't cache non-GET responses. We're technically allowed to cache HEAD requests and some
-      // POST requests, but the complexity of doing so is high and the benefit is low.
+    if (requestMethod !in CACHEABLE_METHODS) {
```
swankjesse, line 231: Rather annoyingly the previous check was very fast, and this check is quite slow. Checking presence in a set needs to allocate an iterator. I don’t like that this might matter?

swankjesse: Doing uglier code is faster:
```suggestion
    if (requestMethod != "GET" && requestMethod != "QUERY") {
```

yschimke:
rebased this, seems worth doing.

desiderantes:
> rebased this, seems worth doing.

Thank you!

okhttp/src/jvmTest/kotlin/okhttp3/CacheTest.kt
```diff
@@ -499,6 +501,82 @@ class CacheTest {
     assertThat(recordedRequest3.exchangeIndex).isEqualTo(2)
   }
 
+  @Test
+  fun getAndQueryRedirectToCachedResultIndependently() {
```
yschimke, line 505: nice thorough test

martinbonnin:
Hey everyone, sorry I'm late to the party. Shouldn't this include the request body as part of the cache key somewhere?

Spec says this:

```
The cache key for a QUERY request (see Section 2 of [HTTP-CACHING]) MUST incorporate the request content.
```

Which I don't think is the case as of today?

yschimke:
@martinbonnin Damn, good catch. I'll revert the PR and we can fix in a new PR.  Sorry @desiderantes!

okhttp/src/jvmTest/kotlin/okhttp3/CacheTest.kt
```diff
@@ -499,6 +501,82 @@ class CacheTest {
...
+    val response2 = client.newCall(request2).execute()
+    assertThat(response2.body.string()).isEqualTo("ABC")
+    val recordedRequest2 = server.takeRequest()
+    assertThat(recordedRequest2.requestLine).isEqualTo("GET /bar HTTP/1.1")
+
+    val request3 =
+      Request
+        .Builder()
+        .url(server.url("/baz"))
+        .query(RequestBody.EMPTY)
```
yschimke, line 562: Good to test that caching includes the request body.

desiderantes:
> @martinbonnin Damn, good catch. I'll revert the PR and we can fix in a new PR.  Sorry @desiderantes!

No worries, I'll send another PR tomorrow with the changes

yschimke:
I may not revert, if we can fix forward. Will keep the review tighter.

Issue #8551, opened 2024-10-11
deflatedBytes is not empty when inflate is called in MessageInflater


Fischiii:
When using binary data over WebSockets we trigger the require statement in the inflate method of the MessageInflater. This happens with release 5.0.0-alpha.12 and 5.0.0.alpha.14. We are using OKhttp on Android.

This only seems to happen with a certain Package size and composition. We attached the packet we used to replicate this a 100% of the time. This is the package before it is being deflated. We can reproduce this issue with multiple server setups. If we return the package below on a server that understands per message deflate we will get the Exception every time:

[bytestream.txt](https://github.com/user-attachments/files/17339145/bytestream.txt)

The failure is always that the inflate method is called but deflatedBytes.size is > 0. Since it is a require call that it must be 0 we fail here. 

![Image](https://github.com/user-attachments/assets/5ec34b78-a55c-4fba-a8a7-cbecc8f1561f)

We assume that the actual issue is on the inflate before the actual failure as this is apparently "finished" while still leaving 4 bytes in the stream.

The last inflate call in `WebSocketReader` before we trigger the require condition

![Image](https://github.com/user-attachments/assets/3055c5e3-ff65-4915-b645-b6ac90b80df0)

Before the last `readOrInflate` call before we trigger the require condition.

![Image](https://github.com/user-attachments/assets/62849e96-a4d8-4cb8-9284-9e7b4e8e8c70)

In the last call of `readOrInflateBefore` we trigger the require condition.

![Image](https://github.com/user-attachments/assets/babfdb43-f7d6-4edb-a2b9-1530fe7f9a15)

The end of the last inflate call before we trigger the require condition. Note that deflatedBytes size is 4 here which is the size we will trigger the require condition with.

![Image](https://github.com/user-attachments/assets/ffaa78f0-bc46-4b19-afea-fd66fa8a53fc)

![Image](https://github.com/user-attachments/assets/c86b85bf-5bf8-4bfd-9433-3387357ae764)

We haven not been able to pin down the why this happens exactly, but we can we can reproduce this a 100% of the time so we can easily provide further information if needed.

yschimke:
cc @swankjesse 

I wonder whether it's triggering the codepath changed in https://github.com/square/okhttp/issues/8078 ?

Fischiii:
@yschimke I just want to add that we have been using the version with the fix  successfully for some time. Just one litte change in the provided payload and the package works. We must be running into some edge case here.

[...]

PR #8582, opened 2024-11-19
Tests for inflator failure.

Fischiii:
[#8551 ](https://github.com/square/okhttp/issues/8551)

Added tests to reproduce the issue. 

Added one test with the broken deflated message:
"inflater left with deflated bytes"

Added another test with a similar, but working message:
"inflater left with deflated bytes working message"

The first test results in the exact same behaviour as we see within okhttp when we discovered the issue: [#8551 ](https://github.com/square/okhttp/issues/8551)


yschimke:
@Fischiii a big thanks for doing this.

Fischiii:
@yschimke merged

swankjesse:
I think I’ve isolated this? 
https://github.com/square/okhttp/pull/8610/files

yschimke:
@swankjesse since the spec specifically has additional data blocks, are we sure it's not just a differ nice between the wp c padding and what our vanilla compression expects?

yschimke:
Closing, there are merge conflicts, and I suggest to focus on https://github.com/square/okhttp/pull/8610

PR #8586, opened 2024-11-20
Update BridgeInterceptor.kt

GeorgCantor:
Use of joinToString

### DIFF
--- a/okhttp/src/main/kotlin/okhttp3/internal/http/BridgeInterceptor.kt
+++ b/okhttp/src/main/kotlin/okhttp3/internal/http/BridgeInterceptor.kt
@@ -110,11 +110,7 @@ class BridgeInterceptor(private val cookieJar: CookieJar) : Interceptor {
   }
 
   /** Returns a 'Cookie' HTTP request header with all cookies, like `a=b; c=d`. */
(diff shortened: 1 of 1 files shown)


okhttp/src/main/kotlin/okhttp3/internal/http/BridgeInterceptor.kt
```diff
@@ -110,11 +110,7 @@ class BridgeInterceptor(private val cookieJar: CookieJar) : Interceptor {
...
   /** Returns a 'Cookie' HTTP request header with all cookies, like `a=b; c=d`. */
-  private fun cookieHeader(cookies: List<Cookie>): String =
-    buildString {
-      cookies.forEachIndexed { index, cookie ->
-        if (index > 0) append("; ")
-        append(cookie.name).append('=').append(cookie.value)
-      }
-    }
+  private fun cookieHeader(cookies: List<Cookie>): String {
+    return cookies.joinToString("; ") { "${it.name}=${it.value}" }
```
JakeWharton, line 114: Unfortunately this creates a nested `StringBuilder` for each name/value pair before adding it to an enclosing one, whereas the old code only uses a single one for all pairs.

Issue #8580, opened 2024-11-18
Allow customization of public suffix DB loading behavior by Platform


Esteth:
On some platforms, Java Resources are expensive to load and encode compared to a platform's native solution for bundling data. The public suffix file is loaded using Java Resources.

This is the case on Android, where Android Resources are faster to load and smaller to store than Java Resources as used in PublicSuffixDatabase.kt

Ideally, each Platform could choose to override the loading behavior for the public suffix database, falling back to the default behavior with java resources since it's supported on all java platforms.

yschimke:
We have an android module, so we could make it conditional on that.  I'll take a look.

But we would need to prove it's noticeably faster to be worth the effort.

Esteth:
I'm virtually certain there are memory wins to be had because AFAIK using Java Resources on android require holding the EOCD zip table for the whole APK in memory in perpetuity once they've been accesed. Loading that table from disk and unpacking it can also take some time.

For small apps this is a small cost, but for large apps that table can be large (1MB+) because it contains strings for all the files in the APK among other things.

JakeWharton:
It's only 40KiB. It could fit into a `ByteArray` of some class...

yschimke:
released in 5.0

PR #8589, opened 2024-11-24
Embed the public suffix database list directly inside a class

yschimke:
Attempt a fix for https://github.com/square/okhttp/issues/8580

Removes the need for proguard for the database. At the cost of more costly code generation at build time.

Point at which we parse the file

```
java.lang.Exception
	at okhttp3.internal.publicsuffix.EmbeddedPublicSuffixList.<clinit>(EmbeddedPublicSuffixList.kt:35)
	at okhttp3.internal.publicsuffix.PublicSuffixDatabase.<clinit>(PublicSuffixDatabase.kt:157)
	at okhttp3.Cookie$Companion.parse$okhttp(Cookie.kt:561)
	at okhttp3.Cookie$Companion.parse(Cookie.kt:450)
	at okhttp3.CookiesTest.acceptOriginalServerMatchesSubdomain(CookiesTest.kt:306)
```

Generated source looks like, and both the source and class file are 57KB.

```
  override val bytes: ByteString = "Y2l0eS5rYXdhc2FraS5qcApjaXR5LmtpdGFreXVzaHUuanAKY2l0eS5rb2JlLmpwCmNpdHkubmFnb3lhLmpwCmNpdHkuc2FwcG9yby5qcApjaXR5LnNlbmRhaS5qcApjaXR5Lnlva29oYW1hLmpwCnd3dy5jawo=".decodeBase64()!!
```


okhttp/build.gradle.kts
```diff
@@ -9,12 +12,48 @@ plugins {
...
+
   filteringCharset = Charsets.UTF_8.toString()
+
+  // TODO replace with KotlinPoet?
+  val databaseGz = project.file("src/test/resources/okhttp3/internal/publicsuffix/PublicSuffixDatabase.gz")
+  println("loading $databaseGz")
+  val (publicSuffixListBytes, publicSuffixListExceptionBytes) = DataInputStream(GZIPInputStream(FileInputStream(databaseGz))).use {
+    val totalBytes = it.readInt()
+    val publicSuffixListBytes = it.readNBytes(totalBytes)
+    println("read $totalBytes")
```
yschimke, line 38: temp logging to ensure we aren't doing this frequently

okhttp/build.gradle.kts
```diff
@@ -9,12 +12,48 @@ plugins {
...
+}
+
 val copyKotlinTemplates = tasks.register<Copy>("copyKotlinTemplates") {
   from("src/main/kotlinTemplates")
   into("$buildDir/generated/sources/kotlinTemplates")
-  expand("projectVersion" to project.version)
+
   filteringCharset = Charsets.UTF_8.toString()
+
+  // TODO replace with KotlinPoet?
```
yschimke, line 32: Suggestions welcome!

yschimke:
Seems to kill the kotlin compiler

Maybe https://youtrack.jetbrains.com/issue/KT-72429/StackOverflowError-when-compiling-large-files

or

https://stackoverflow.com/a/60934758/1542667

yschimke:
I shouldn't have uncompressed it.  Will compare the compressed version

okhttp/build.gradle.kts
```diff
@@ -9,12 +13,27 @@ plugins {
   id("binary-compatibility-validator")
 }
 
-// Build & use okhttp3/internal/-InternalVersion.kt
+fun ByteArray.toByteStringExpression(): String {
+  return "\"${Base64.getEncoder().encodeToString(this@toByteStringExpression)}\".decodeBase64()!!"
```
JakeWharton, line 17: Seems like `byteArrayOf` would save even more space since it would be stored raw instead of as a string that is then encoded to bytes for the classfile.

JakeWharton (commented):
I'll review the rest on desktop. Too dense for mobile.

okhttp/build.gradle.kts
yschimke: Yep, I'll try that. I was hitting limits of Kotlin? compiler before, with the source file like "...12, 15, -34, -56, 12..."

But let me try again

yschimke: Source back to 150KB, Kotlin compiler is noticeably struggling with the build. When the compile completes I'll report on the class file.

okhttp/build.gradle.kts
yschimke: Since KotlinPoet generates source, this doesn't seem like a big win.

okhttp/build.gradle.kts
yschimke: I've given up on that for now.

JakeWharton: Probably worth filing a Kotlin issue if it's not fast. Although this project is stuck on old Kotlin with the old compiler, I believe.

Could also just generate package-private Java with a single static field.

yschimke:
```
C:\Users\yuri\IdeaProjects\okhttp\okhttp\build\generated\sources\kotlinTemplates\okhttp3\internal\publicsuffix\PublicSuffixDatabaseGZ.java:29: error: code too large
  public static final byte[] compressedBytes = new byte[] {31,-117,8,0,0,0,0,0,0,0,-116,-67,-33,-114,-60,-88,-46,39,-72,123,91,-113,50,82,-90,78,119,-97,-17,-100,-39,-57
```

yschimke: If I split to 10 methods, it works but they don't get optimised as a single array for some reason?

If I split to 10 constants, still the same error.

JakeWharton: Sure. I'll look tomorrow

okhttp/src/main/kotlin/okhttp3/internal/publicsuffix/PublicSuffixDatabase.kt
```diff
@@ -338,7 +239,7 @@ class PublicSuffixDatabase internal constructor(
             low = mid + end + 1
           } else {
             // Found a match.
-            match = String(this, mid, publicSuffixLength)
+            match = this.substring(mid, mid + publicSuffixLength).string(Charsets.UTF_8)
```
JakeWharton, line 242: Seems like `string` should have an overload which takes start/end indices to avoid this double copy. Or we should have added [`ByteString.slice`](https://github.com/square/okio/issues/668)!

okhttp/src/main/kotlinTemplates/okhttp3/internal/publicsuffix/EmbeddedPublicSuffixList.kt
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * A implementation of I/O for PublicSuffixDatabase.gz by directly encoding
+ * the relevant byte arrays in a class file.
+ */
+internal object EmbeddedPublicSuffixList: PublicSuffixList {
+  override fun ensureLoaded() {
+  }
+
+  val compressedBytes: ByteString = $publicSuffixListBytes
```
JakeWharton, line 38: How about a nullable `var` and setting to null once read?

okhttp/src/main/kotlinTemplates/okhttp3/internal/publicsuffix/EmbeddedPublicSuffixList.kt
```diff
@@ -0,0 +1,56 @@
...
+
+  override val bytes: ByteString
+
+  override val exceptionBytes: ByteString
+
+  init {
```
JakeWharton, line 44: Not done in `ensureLoaded`?

JakeWharton (approved):

okhttp/src/main/kotlinTemplates/okhttp3/internal/publicsuffix/EmbeddedPublicSuffixList.kt
yschimke: I figured the JVM classloader would be a nice simple mutex. Ther s so much locking code in the other impl

okhttp/src/main/kotlin/okhttp3/internal/publicsuffix/PublicSuffixDatabase.kt
yschimke: Will raise a PR.

swankjesse (commented):
I particularly don’t like this. We shouldn’t need to degrade good user’s runtime performance and artifact sizes because bad users are clumsy and trip on every rake.

swankjesse:
Is there a way to load resources on Android that isn’t horrible for performance?!

yschimke:
I can benchmark 
- resources
- assets
- embedded

Although my money is on this being better for performance than resources, just maybe at a small runtime memory cost.

swankjesse:
Can we do assets on Android and resources on the JVM? OkHttp 5 can be multiplatform with just Android and JVM platforms, right?

yschimke:
This went a bit pear shaped...

https://github.com/square/okhttp/pull/8599

But exploring the space :)

yschimke:
I'll check the multiplatform build option next.

yschimke:
hopefully go with multiplatform following something like

https://jeroenmols.com/blog/2021/03/17/share-code-kotlin-multiplatform/

And avoid the com.android.library plugin

https://kotlinlang.org/docs/whatsnew2020.html#deprecated-compatibility-with-kotlin-multiplatform-gradle-plugin-and-gradle-java-plugins

yschimke:
Made an attempt https://github.com/square/okhttp/pull/8600

We would need that other PR above in order to get a Context to load from assets.  So that other change is worth benchmarking first.

swankjesse (commented):
LGTM!

PR #8599, opened 2024-11-30
Experiment with configuring Android to use assets

yschimke:
for loading the PublicSuffixDatabase.

TODO - benchmark the 3 options


swankjesse:
I don’t think benchmarking is _that_ useful here. Simpler to strictly use assets on Android and resources on the JVM, which we know is a good fast configuration.

(Though _how_ to use Assets when we don’t have a Context is tricky)

yschimke:
@swankjesse I'll take another stab since https://github.com/square/okhttp/pull/8600 landed.

Should be now possible.

PR #8610, opened 2024-12-06
Discard trailing data after a deflated message

swankjesse:
We can recover from this and it seems like the least worst of our options.

Closes: https://github.com/square/okhttp/issues/8551


okhttp/src/main/kotlin/okhttp3/internal/ws/MessageInflater.kt
```diff
@@ -55,8 +58,21 @@ class MessageInflater(
     do {
       inflaterSource.readOrInflate(buffer, Long.MAX_VALUE)
     } while (inflater.bytesRead < totalBytesToRead && !inflater.finished())
+
+    // The inflater data was self-terminated and there's unexpected trailing data. Tear it all down
+    // so we don't leak that data into the input of the next message.
+    if (inflater.bytesRead < totalBytesToRead) {
+      deflatedBytes.clear()
+      inflaterSource.close()
+      this.inflaterSource = null
```
swankjesse, line 67: I’m gonna claim it’s really weird if you have a stream-terminating message followed by another message and `noContextTakeover = false`. But perhaps I should write a test for this? I’m not exactly sure what reasonable behavior is.

swankjesse: Also - I think the server in the case we’re addressing here is probably buggy. We might be better to follow @Fischiii’s lead and just reject the bogus data.

yschimke: I think that was my change to make his PR pass

okhttp/src/test/java/okhttp3/internal/ws/MessageDeflaterInflaterTest.kt
```diff
@@ -136,6 +139,41 @@ internal class MessageDeflaterInflaterTest {
     assertThat(buffer.readUtf8()).isEqualTo("Hello inflation!")
   }
 
+  /**
+   * It's possible a self-terminating deflated message will contain trailing data that won't be
```
yschimke, line 143: Is it worth tying this back to the padding bytes in the spec?

yschimke (approved):

okhttp/src/main/kotlin/okhttp3/internal/ws/MessageInflater.kt
Fischiii: @swankjesse @yschimke in case it helps. The server used is a Kestrel (.net version 7) with the WebsocketSharp.Standard2 Framework for the WebwSocket connection. Data transferred is binary form Protobuffer.

PR #8634, opened 2024-12-27
Merge okhttp-android into okhttp.

yschimke:
Since :okhttp is kmp, it isn't required to be separate. And also the okhttp-android/jvm artifact naming conflicts with it.


yschimke:
@swankjesse we will need this before any releases because of the conflict in artifact naming.

yschimke:
Moved to https://github.com/square/okhttp/pull/8635

PR #8636, opened 2024-12-27
Use Assets or Resources based on platform.

yschimke:
Assets uncompressed on Android. Assuming that the AAR file will compress.

Use Resources, compressed on JVM.

Includes ~an Androidx Startup Initializer~ a ContentProvider to grab the app context.


yschimke:
Androidx Startup approach fails with

https://github.com/square/okhttp/actions/runs/12520208963/job/34925534823?pr=8636

```
Execution failed for task ':okhttp:animalsnifferMain'.
> Could not resolve all files for configuration ':okhttp:compileClasspath'.
   > Could not resolve androidx.startup:startup-runtime:1.2.0.
     Required by:
         project :okhttp
      > No matching variant of androidx.startup:startup-runtime:1.2.0 was found. The consumer was configured to find a library for use during compile-time, compatible with Java 8, preferably in the form of class files, preferably optimized for standard JVMs, and its dependencies declared externally but:
```

okhttp/build.gradle.kts
```diff
@@ -3,10 +3,9 @@
 import aQute.bnd.gradle.BundleTaskExtension
 import com.vanniktech.maven.publish.JavadocJar
 import com.vanniktech.maven.publish.KotlinMultiplatform
-import org.gradle.kotlin.dsl.compileOnly
+import org.gradle.internal.fingerprint.classpath.impl.ClasspathFingerprintingStrategy.compileClasspath
```
yschimke, line 6: oops

JakeWharton:
androidx.startup is the right choice, though. The cost of each content provider is non-zero. Animal sniffer needs changed to run on the extracted jars from the aars, or just dropped from use on the Android target.

yschimke:
@JakeWharton trying https://github.com/gmazzo/gradle-aar2jar-plugin

PR #8648, opened 2025-01-09
FIX: Fixed the issue where the response called by onFailure is closed…

whosmyqueen:
… when response.isSuccessful is equal to false


whosmyqueen:
I get the return value of Response in onFailure in EventSourceListener, which indicates that Response is closed.

yschimke:
> indicates that Response is closed.

What are you checking that says it's closed and why does it matter? Is the response error incomplete for you?

Please raise a bug rather than a fix before it's clear there is a problem.

PR #8651, opened 2025-01-17
Create a friends configuration to use internals.

yschimke:
From https://www.liutikas.net/2025/01/12/Kotlin-Library-Friends.html we should avoid using

@file:Suppress("INVISIBLE_MEMBER", "INVISIBLE_REFERENCE")

So instead lets just be friends.


yschimke:
This is annoying https://github.com/square/okhttp/pull/8651/files#diff-970d7a7163dea8a6aebc4e054b84492894d60a955e261de0eaab23c342d96e53

and https://github.com/square/okhttp/pull/8651/files#diff-13b0055b3788ccf725cab8dd6074f62feaf2b5c2c7402a3d42f4f68b2b7ff194

yschimke:
One reference left in okhttp.android.test.AndroidAsyncDnsTest

To follow up

mockwebserver-deprecated/build.gradle.kts
```diff
@@ -15,7 +15,7 @@ tasks.jar {
 }
 
 dependencies {
-  api(projects.okhttp)
+  "friends"(projects.okhttp)
```
JakeWharton, line 18: Seems like this requires two configurations: `implementationFriend` and `apiFriend`. Otherwise you are changing the visibility of the transitive dependency, as evidence by all the unrelated changes below where you are having to add the dependency yourself when it should be automatically available.

PR #8665, opened 2025-01-30
Fix poor buffering case for MultipartReader

yschimke:
100mb Body is fully written, reader wants 1MB chunks, but get 8kb at a time. Unfortunately internally we do indexOf over the full 100mb.

```
outer reading 1048576
indexOf over 104857611 with maxResult 8192
inner 8192
outer read 8192
outer reading 1048576
indexOf over 104849419 with maxResult 8192
inner 8192
outer read 8192
```


yschimke:
@swankjesse any suggestions for the right way to test and fix this?

yschimke:
I'm not sure if this is actually a bug in Okio RealBufferedSource that it assumes that checking segment size is optimal?

yschimke:
500ms with the fix
I stopped after 2 minutes without the fix

yschimke:
@swankjesse requesting post review (as agreed)

PR #8666, opened 2025-02-01
Add retryDecision event to track retry logic in EventListener

### LINKED ISSUE #8663: How to track retries in the EventListener callbacks?
Hi Team,

I want to capture timing metrics like **connect, SSL, DNS, and request/response times** for each request, including retries. However, since OkHttp reuses the same Call object for retries and doesn't always trigger callFailed(), it isn't easy to differentiate retries from the original request.

Observations:
- `callStart()` is only triggered once per Call. And the retries may start at `connectStart()`, `requestHeadersStart()`, etc - depending upon the state of the original request.
- `callFailed()` isn't always called before a retry, making failure tracking tricky.
- The same `Call` instance is used across retries.

Question:
I want to track the metrics for each request separately like below - 
`{
  "attempt": 1,
  "dnsLookupTimeMs": 50,
  "connectTimeMs": 120,
  "sslHandshakeTimeMs": 80,
  // failure occurs during response body read
},
{
  "attempt": 2, // retry request which succeeds
   // Reused connection, so no new DNS, connect and SSL lookup
  responseTime: 200,
  requestTotalTime: 500
}
`
How can I achieve this?

yschimke:
Introduced a new `retryDecision` event to improve observability of retry logic. This includes detailed reasons for retry decisions, enhancing debugging. Updated relevant tests, interceptors, and event handling to accommodate the new event.

It also cleanly signals the boundary before a prior request and the follow up.

fixes #8663


yschimke:
@swankjesse requesting post review. This seems like useful context for event listeners.

ddas09:
@yschimke, will this `retryDecision` event be invoked when we retry requests using a Custom RetryInterceptor? Or this is only for the inbuilt auto-retries?

yschimke:
Currently only for inbuilt. But maybe it should be considered for all secondary requests.

Technically your interceptor could call event listener (I think).

What is hard is app interceptors can execute 0 to n requests, and even concurrently. 

I guess whenever chain.proceed is called after already returning?

ddas09:
That’s correct.

swankjesse:
I took another step on this:
https://github.com/square/okhttp/pull/8796

PR #8676, opened 2025-02-17
[4.x] Fix poor buffering case for MultipartReader

yschimke:
* Demonstrate poor buffering case
* Fix for repeated reads of small byteCount from large part

(cherry picked from commit 3cc87c3fc4bee09642cc22a561060da7af2b939c)


yschimke:
@swankjesse requesting post review (as agreed)

PR #8720, opened 2025-04-05
`Empty` constant of Headers, RequestBody, ResponseBody

### LINKED ISSUE #8670: Provide a public API like RequestBody.Empty
I'm using version 5.0.0-alpha14, and found that we have a public `commonEmptyRequestBody` in `-UtilCommon.kt`.

```kotlin
val commonEmptyHeaders: Headers = Headers.headersOf()
val commonEmptyRequestBody: RequestBody = EMPTY_BYTE_ARRAY.toRequestBody()
val commonEmptyResponse: ResponseBody = EMPTY_BYTE_ARRAY.toResponseBody()
```

I think we can have them `internal` and public them like this:
```kotlin
// with @JvmField

Headers.Empty
RequestBody.Empty
ResponseBody.Empty
````
This type might fit Kotlin more. Also we can easier access them in Java like `Headers.Companion.Empty`

### PR DESCRIPTION
**Pull request by MukjepScarlet**
removed informal internal constants

closes #8670

### DIFF
--- a/okhttp/src/commonJvmAndroid/kotlin/okhttp3/RequestBody.kt
+++ b/okhttp/src/commonJvmAndroid/kotlin/okhttp3/RequestBody.kt
@@ -19,8 +19,11 @@ import java.io.File
 import java.io.FileDescriptor
 import java.io.FileInputStream
 import java.io.IOException
-import okhttp3.internal.*
 import okhttp3.internal.chooseCharset
+import okhttp3.internal.commonContentLength
+import okhttp3.internal.commonIsDuplex
+import okhttp3.internal.commonIsOneShot
+import okhttp3.internal.commonToRequestBody
 import okio.BufferedSink
 import okio.ByteString
 import okio.FileSystem
@@ -107,7 +110,7 @@ abstract class RequestBody {
 
       override fun contentLength() = 0L
 
-      override fun writeTo(sink: BufferedSink) { /** write nothing */ }
+      override fun writeTo(sink: BufferedSink) {}
     }
 
     /**
--- a/okhttp/src/commonJvmAndroid/kotlin/okhttp3/RequestBody.kt
+++ b/okhttp/src/commonJvmAndroid/kotlin/okhttp3/RequestBody.kt
@@ -19,11 +19,8 @@ import java.io.File
 import java.io.FileDescriptor
 import java.io.FileInputStream
 import java.io.IOException
+import okhttp3.internal.*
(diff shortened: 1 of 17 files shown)

### CODE REVIEW

File: okhttp/src/commonJvmAndroid/kotlin/okhttp3/RequestBody.kt [hunk 1]
```diff
@@ -19,11 +19,8 @@ import java.io.File
 import java.io.FileDescriptor
 import java.io.FileInputStream
 import java.io.IOException
+import okhttp3.internal.*
```
yschimke, line 22: can you expand this?

MukjepScarlet: Okay? But more than 5 extensions are imported

okhttp/src/commonJvmAndroid/kotlin/okhttp3/ResponseBody.kt
```diff
@@ -213,6 +213,20 @@ abstract class ResponseBody : Closeable {
   }
 
   companion object {
+    /** Empty response body. */
+    @JvmField
+    val Empty: ResponseBody = EmptyResponseBody()
+
+    private class EmptyResponseBody : ResponseBody() {
+      private val source = Buffer()
```
yschimke, line 221: This seems dangerous, since it will be externally accessible and is mutable, also not thread safe.

I'd suggest implementing an EmptySource, and `.buffer()` each time you return it?

MukjepScarlet: Create a `Buffer` instance for each `ResponseBody.Empty.source()` invocation?

MukjepScarlet: It's returned as `BufferedSource`, but if the user cast it to `Buffer` then they can write contents to it...

yschimke: Yep, but at least that won't be seen by other callers.  We should probably ask upstream for Source.Empty and friends.

MukjepScarlet: One trouble of EmptyBufferedSource is that it also has a Buffer member, if we want user can never read data from it, we still need a `EmptyBuffer`

yschimke:
run `./gradlew okhttp:apiDump` to update api files

PR #8735, opened 2025-04-10
catching Throwables in RealEventSource

### LINKED ISSUE #8463: RealEventSource lost control on Error
As shown in the code snippet, if an Error (such as OutOfMemoryError or StackOverflowError, which are common) is thrown at line 78, the catch block at line 81 will not catch this Error, and neither the onFailure nor onClose hooks will be called. Consequently, the processing flow will terminate silently. 
Therefore, it is recommended to catch Throwable instead.

okhttp_bug

MukjepScarlet:
closes #8463

### DIFF
--- a/okhttp-sse/src/main/kotlin/okhttp3/sse/internal/RealEventSource.kt
+++ b/okhttp-sse/src/main/kotlin/okhttp3/sse/internal/RealEventSource.kt
@@ -80,7 +80,7 @@ internal class RealEventSource(
           while (!canceled && reader.processNextEvent()) {
           }
         }
-      } catch (e: Exception) {
+      } catch (e: Throwable) {
         val exception =
           when {
             canceled -> IOException("canceled", e)


MukjepScarlet:
Defination of `listener.onFailure`:

```kotlin
public open fun onFailure(
    eventSource: EventSource,
    t: Throwable?,
    response: Response?
): Unit
```

okhttp-sse/src/main/kotlin/okhttp3/sse/internal/RealEventSource.kt
```diff
@@ -80,7 +80,7 @@ internal class RealEventSource(
           while (!canceled && reader.processNextEvent()) {
           }
         }
-      } catch (e: Exception) {
+      } catch (e: Throwable) {
```
AlexCue987, line 83: Is there a reason we want to catch/handle something we cannot recover from, such as an `OutOfMemoryError`.

MukjepScarlet: Because the `onFailure` will re-throw it, with param of `Throwable?` instead of `Exception?`

yschimke: I responded on the original bug, I don't think we should be surfacing these as if the app can handle and continue. The state of OkHttpClient is probably unknown at this point.

PR #8748, opened 2025-04-24
replace Regex.matchAtPolyfill with Kotlin stdlib function

MukjepScarlet:
It's no longer experimental now.
[doc](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.text/-regex/#605132161%2FFunctions%2F447362634)


okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/-MediaTypeCommon.kt
```diff
@@ -45,16 +45,15 @@ private val PARAMETER = Regex(";\\s*(?:$TOKEN=(?:$TOKEN|$QUOTED))?")
  * @throws IllegalArgumentException if this is not a well-formed media type.
  */
 internal fun String.commonToMediaType(): MediaType {
-  val typeSubtype: MatchResult =
-    TYPE_SUBTYPE.matchAtPolyfill(this, 0)
+  val typeSubtype = TYPE_SUBTYPE.matchAt(this, 0)
```
JakeWharton, line 48: Seems like Spotless wants
```suggestion
  val typeSubtype =
    TYPE_SUBTYPE.matchAt(this, 0)
```

JakeWharton (approved):

PR #8749, opened 2025-04-24
fix: IOException is swallowed in ResponseBody.close and Response.close

MukjepScarlet:


yschimke:
Can you add some context?

These methods use closeQuietly, so should not be throwing an exception.

PR #8765, opened 2025-05-10
Fix test with latest JDKs

yschimke:
Change in https://bugs.openjdk.org/browse/JDK-8335135

> HttpInputStream does not throw IOException when response is truncated


mockwebserver/src/test/java/mockwebserver3/MockWebServerTest.kt
```diff
@@ -439,6 +439,7 @@ class MockWebServerTest {
 
   @Test
   fun disconnectResponseHalfway() {
+    println(System.getProperty("java.version"))
```
JakeWharton, line 442:
```suggestion
```

JakeWharton (approved):

PR #8769, opened 2025-05-13
Drop the findbugs dependency

swankjesse:
We don't need it.


yschimke (approved):

okhttp/build.gradle.kts
```diff
@@ -231,33 +226,6 @@ tasks.named<Jar>("jvmJar").configure {
...
     )
   }
-
-//  val bundleExtension = extensions.create(
```
yschimke, line 235: This was the new hack I mentioned.

JakeWharton (approved):
All bugs have been found. Dependency no longer required.

PR #8796, opened 2025-05-23
Split EventListener.retryDecision into retryDecision and followUpDecision

swankjesse:
retryDecision has an exception, and is related to connectivity failures.

followUpDecision has an intermediate response, and is related to redirects and auth challenges.

This drops the reason strings, which have 'logging brain' and I want to push our users towards 'metrics brain'.

### DIFF
--- a/okhttp-sse/src/test/java/okhttp3/sse/internal/EventSourceHttpTest.kt
+++ b/okhttp-sse/src/test/java/okhttp3/sse/internal/EventSourceHttpTest.kt
@@ -257,6 +257,7 @@ class EventSourceHttpTest {
       "RequestHeadersEnd",
       "ResponseHeadersStart",
       "ResponseHeadersEnd",
+      "FollowUpDecision",
       "ResponseBodyStart",
       "ResponseBodyEnd",
       "ConnectionReleased",


okhttp/src/commonJvmAndroid/kotlin/okhttp3/EventListener.kt
```diff
@@ -454,14 +455,58 @@ abstract class EventListener {
...
+   * Client configuration may be used to make follow-up decisions, such as:
+   *
+   *  * [OkHttpClient.followRedirects] must be true to follow redirects.
+   *  * [OkHttpClient.followSslRedirects] must be true to follow redirects that add or remove HTTPS.
+   *  * [OkHttpClient.authenticator] must respond to an authorization challenge.
+   *
+   * @param networkResponse the intermediate response that may require a follow-up request.
+   * @param nextRequest the follow-up request that will be made. Null if no follow-up will be made.
+   */
+  open fun followUpDecision(
```
yschimke, line 506: Yep, nice. When documented properly, those are clearly two separate things.

yschimke (approved):
It feels a shame to lose the why. But `reason: String` is not a structured why, so agree with this on principle.

PR #8797, opened 2025-05-23
Stop double-compressing the PublicSuffixDatabase

squarejesse:
It's compressed as a .gz file that's also in a .jar file.

Previously the compressed size was 42,502 bytes, and the total okhttp.jar file was 884,193.

With this change, the total okhttp.jar file is 884,154 bytes.

### DIFF
--- a/okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/publicsuffix/ResourcePublicSuffixList.kt
+++ b/okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/publicsuffix/ResourcePublicSuffixList.kt
@@ -16,7 +16,6 @@
 package okhttp3.internal.publicsuffix
 
 import okio.FileSystem
-import okio.GzipSource
 import okio.Path
 import okio.Path.Companion.toPath
 import okio.Source
--- a/okhttp/src/jvmTest/kotlin/okhttp3/internal/publicsuffix/PublicSuffixListGenerator.kt
+++ b/okhttp/src/jvmTest/kotlin/okhttp3/internal/publicsuffix/PublicSuffixListGenerator.kt
@@ -34,7 +34,6 @@ import okio.FileSystem
 import okio.Path
 import okio.Path.Companion.toPath
 import okio.buffer
-import okio.gzip
 
 /**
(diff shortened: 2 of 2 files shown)


JakeWharton (approved):

okhttp/src/jvmTest/resources/okhttp3/internal/publicsuffix/NOTICE
```diff
@@ -1,4 +1,4 @@
-Note that PublicSuffixDatabase.gz is compiled from The Public Suffix List:
+Note that PublicSuffixDatabase is compiled from The Public Suffix List:
```
yschimke, line 1: Should .list suffix be present?

okhttp/src/jvmMain/resources/META-INF/native-image/okhttp/okhttp/resource-config.json
```diff
@@ -1,5 +1,5 @@
 {
   "resources": [
-    {"pattern": "okhttp3/internal/publicsuffix/PublicSuffixDatabase.gz"}
+    {"pattern": "okhttp3/internal/publicsuffix/PublicSuffixDatabase"}
```
yschimke, line 3: .list?

yschimke (approved):

okhttp/src/jvmMain/resources/META-INF/native-image/okhttp/okhttp/resource-config.json
yschimke: I hope that graal tests will catch this if I didn't. Turning on for this CL.

okhttp/src/jvmTest/resources/okhttp3/internal/publicsuffix/NOTICE
yschimke: Still a mix of .list and .dat.

swankjesse: Yeah, in that directory it’s `public_suffix_list.dat` which is the exact file we get from the publicsuffix.org. Elsewhere it’s our custom transformation of that file for good performance at runtime.

PR #8800, opened 2025-05-26
[WIP] Call.Decorator

yschimke:
Demonstrate a possible Call.Decorator API that handles some of the following cases, without forcing clients to deal exclusively with Call.Factory.

- Intercept Call creation on the callers thread - for example associating trace information
- Modifying the request - adding tags 
- Platform specific checks, such as Android Main thread or insecure URL checks
- Switching between clients, such as Network Pinning

Todo

- [ ] Agree correct API
- [ ] Add tests

### DIFF
--- a/android-test/src/androidTest/java/okhttp/android/test/OffMainThread.kt
+++ b/android-test/src/androidTest/java/okhttp/android/test/OffMainThread.kt
@@ -1,0 +1,46 @@
+/*
+ * Copyright (C) 2025 Block, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okhttp.android.test
+
+import android.os.Looper
+import okhttp3.Call
+import okhttp3.Request
+import okhttp3.Response
+
+object OffMainThread : Call.Decorator {
+  override fun newCall(
+    chain: Call.Factory,
+    request: Request,
+  ): Call = StrictModeCall(chain.newCall(request))
+
+  private class StrictModeCall(
+    private val delegate: Call,
+  ) : Call by delegate {
+    init {
+      println("StrictModeCall")
+    }
+
+    override fun execute(): Response {
+      if (Looper.getMainLooper().isCurrentThread) {
+        throw IllegalStateException("Network on main thread")
+      }
+
+      return delegate.execute()
+    }
+
+    override fun clone(): Call = StrictModeCall(delegate.clone())
+  }
+}
--- a/okhttp/src/commonJvmAndroid/kotlin/okhttp3/Call.kt
+++ b/okhttp/src/commonJvmAndroid/kotlin/okhttp3/Call.kt
@@ -15,6 +15,7 @@
  */
 package okhttp3
 
+import okhttp3.Interceptor.Chain
 import okio.IOException
 import okio.Timeout
 
@@ -95,5 +96,9 @@ interface Call : Cloneable {
 
   fun interface Factory {
     fun newCall(request: Request): Call
+  }
+
+  fun interface Decorator {
+    fun newCall(chain: Call.Factory, request: Request): Call
   }
 }
--- a/okhttp/src/commonJvmAndroid/kotlin/okhttp3/Call.kt
+++ b/okhttp/src/commonJvmAndroid/kotlin/okhttp3/Call.kt
@@ -15,7 +15,6 @@
  */
 package okhttp3
 
-import okhttp3.Interceptor.Chain
 import okio.IOException
 import okio.Timeout
 
@@ -99,6 +98,9 @@ interface Call : Cloneable {
   }
 
   fun interface Decorator {
-    fun newCall(chain: Call.Factory, request: Request): Call
+    fun newCall(
+      chain: Call.Factory,
+      request: Request,
+    ): Call
   }
 }
--- a/okhttp/src/commonJvmAndroid/kotlin/okhttp3/OkHttpClient.kt
+++ b/okhttp/src/commonJvmAndroid/kotlin/okhttp3/OkHttpClient.kt
@@ -150,7 +150,6 @@ open class OkHttpClient internal constructor(
    * the connection is established (if any) until after the response source is selected (either the
    * origin server, cache, or both).
    */
-  @get:JvmName("interceptors")
   val callDecorators: List<Call.Decorator> =
     builder.callDecorators.toImmutableList()
 
@@ -369,18 +368,24 @@ open class OkHttpClient internal constructor(
 
   /** Prepares the [request] to be executed at some point in the future. */
   override fun newCall(request: Request): Call {
-    if (callDecorators.isEmpty())
-      return RealCall(this, request, forWebSocket = false)
+    if (callDecorators.isNotEmpty()) {
+      return DecoratedCallFactory().newCall(request)
+    }
 
-    return newCall(request, decoratorOrder = callDecorators.listIterator())
+    return RealCall(this, request, forWebSocket = false)
   }
 
-  internal fun newCall(request: Request, decoratorOrder: ListIterator<Call.Decorator>): Call {
-    val decorator = decoratorOrder.next()
+  private inner class DecoratedCallFactory(
+    private val index: Int = 0,
+  ) : Call.Factory {
+    init {
+      println("DecoratedCallFactory index: $index")
+    }
 
-    val
-
-    return decorator.newCall(chain, request)
+    override fun newCall(request: Request): Call {
+      val next = if (index > callDecorators.lastIndex) this@OkHttpClient else DecoratedCallFactory(index + 1)
+      return callDecorators[index].newCall(next, request)
+    }
   }
 
   /** Uses [request] to connect a new web socket. */
@@ -654,6 +659,7 @@ open class OkHttpClient internal constructor(
       this.dispatcher = okHttpClient.dispatcher
       this.connectionPool = okHttpClient.connectionPool
       this.interceptors += okHttpClient.interceptors
+      this.callDecorators += okHttpClient.callDecorators
       this.networkInterceptors += okHttpClient.networkInterceptors
       this.eventListenerFactory = okHttpClient.eventListenerFactory
       this.retryOnConnectionFailure = okHttpClient.retryOnConnectionFailure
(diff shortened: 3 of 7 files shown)


okhttp/src/commonJvmAndroid/kotlin/okhttp3/Call.kt
```diff
@@ -96,4 +96,11 @@ interface Call : Cloneable {
   fun interface Factory {
     fun newCall(request: Request): Call
   }
+
+  fun interface Decorator {
```
yschimke, line 118: TODO comment

okhttp/src/commonJvmAndroid/kotlin/okhttp3/OkHttpClient.kt
```diff
@@ -145,6 +145,14 @@ open class OkHttpClient internal constructor(
   val interceptors: List<Interceptor> =
     builder.interceptors.toImmutableList()
 
+  /**
+   * Returns an immutable list of interceptors that observe the full span of each call: from before
```
yschimke: TODO comment

okhttp/src/commonJvmAndroid/kotlin/okhttp3/OkHttpClient.kt
```diff
@@ -359,7 +367,26 @@ open class OkHttpClient internal constructor(
...
+    }
+
+    return RealCall(this, request, forWebSocket = false)
+  }
+
+  private inner class DecoratedCallFactory(
+    private val index: Int = 0,
+  ) : Call.Factory {
+    init {
+      println("DecoratedCallFactory index: $index")
```
yschimke: TODO remove

harrytmthy (commented):
Hi @yschimke, I noticed the current API is a bit ambiguous about whether decorators are intended to be "wrappers" or can "short-circuit" the call. If one decorator short-circuits (i.e. returns a call without calling `chain.newCall()`), it effectively skips all decorators after it, and the rest of the chain is unaware.

Short-circuiting might be intentional in cases like:

```kotlin
if (noNetworkAvailable()) {
  return FailingCall("No network")
}
return chain.newCall(request)
```

This ambiguity could lead to some issues:
- **Ordering becomes critical and fragile:** Decorators that must always run (e.g. security, metrics, tracing) need to be registered first, but this isn't obvious or enforced.
- **Invisible dependencies:** One decorator might assume another has run (e.g. metrics before auth).
- **Debugging becomes tricky:** "Why isn't tracing working?" → "Oh, the offline checker short-circuited the chain."

Would it be worth documenting this explicitly? Something like:

> Decorators can optionally short-circuit the call chain. If so, be aware that any decorators added after them will not be invoked.

Alternatively, if short-circuiting is rare or discouraged, it might make sense to split the interface (e.g. `CallWrapper` vs `CallInterceptor`), though that might be overkill for now.

yschimke:
@harrytmthy yep, definitely worth documenting.

Interceptors document exactly this, so I'll add.

In some ways this is the point, while interceptors surround the execution of the Call.

App interceptor - 0..n executions
Network interceptor - exactly 1 ( or failure)

This Call.Decorator are configured centrally in order to allow to short circuit, modify the original request or even redirect the call to another client instance.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/OkHttpClient.kt
```diff
@@ -359,7 +367,24 @@ open class OkHttpClient internal constructor(
...
+  }
+
+  private inner class DecoratedCallFactory : Call.Factory {
+    var index: Int = 0
+
+    override fun newCall(request: Request): Call =
+      if (index > callDecorators.lastIndex) {
+        RealCall(this@OkHttpClient, request, forWebSocket = false)
+      } else {
+        callDecorators[index++].newCall(this, request)
```
yschimke: This mutability might be an issue if someone implements a decorator that the chain twice.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/OkHttpClient.kt
```diff
@@ -359,7 +367,24 @@ open class OkHttpClient internal constructor(
...
+
+    return RealCall(this, request, forWebSocket = false)
+  }
+
+  private inner class DecoratedCallFactory : Call.Factory {
+    var index: Int = 0
+
+    override fun newCall(request: Request): Call =
+      if (index > callDecorators.lastIndex) {
+        RealCall(this@OkHttpClient, request, forWebSocket = false)
```
yschimke: I think this should be some wrapper returned, that utilises the result of the first decorator.

then passes it through the interceptor chain.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/OkHttpClient.kt
```diff
@@ -359,7 +367,24 @@ open class OkHttpClient internal constructor(
   }
 
   /** Prepares the [request] to be executed at some point in the future. */
-  override fun newCall(request: Request): Call = RealCall(this, request, forWebSocket = false)
+  override fun newCall(request: Request): Call {
+    if (callDecorators.isNotEmpty()) {
+      return DecoratedCallFactory().newCall(request)
+    }
+
+    return RealCall(this, request, forWebSocket = false)
```
pyricau: Suggestion: delete DecoratedCallFactory, add a new `decoratedCallFactory` field:


```kotlin
  private val decoratedCallFactory by lazy {
    var previousCallFactory = Call.Factory { 
      request -> RealCall(this, request, forWebSocket = false)
    }
    for (callDecorator in callDecorators.reversed()) {
      val next = previousCallFactory
      previousCallFactory = Call.Factory { request -> callDecorator.newCall(next, request) }
    }
    previousCallFactory
  }
  ```
  
  Then you can update `newCall()` to:
  
  ```kotlin
  override fun newCall(request: Request): Call = decoratedCallFactory.newCall(request)
  ```

Comment (yschimke) [reply okhttp/src/commonJvmAndroid/kotlin/okhttp3/OkHttpClient.kt]: Yeah, that's much better. Much nicer, safer, cleaner.

Kind of annoying really...

Comment (pyricau) [reply okhttp/src/commonJvmAndroid/kotlin/okhttp3/OkHttpClient.kt]: ok time to come forward: I actually tried a few time to get here and couldn't, so I wrote a recursive impl instead, then asked an LLM to make me a non recursive version which then helped me clean up and come up with this one

File: android-test/src/androidTest/java/okhttp/android/test/OffMainThread.kt [hunk 7]
```diff
@@ -0,0 +1,42 @@
...
+    chain: Call.Factory,
+    request: Request,
+  ): Call = StrictModeCall(chain.newCall(request))
+
+  private class StrictModeCall(
+    private val delegate: Call,
+  ) : Call by delegate {
+    override fun execute(): Response {
+      if (Looper.getMainLooper().isCurrentThread) {
+        throw IllegalStateException("Network on main thread")
```
pyricau, line 33: Not a strong opinion, but I've seen `IllegalThreadStateException` used in our codebase.. though idk that's meant to "indicate that a thread is not in an appropriate state for the requested operation" so not sure. And weird that exception extends IllegalArgumentException

android-test/src/androidTest/java/okhttp/android/test/OffMainThread.kt
```diff
@@ -0,0 +1,42 @@
...
+  override fun newCall(
+    chain: Call.Factory,
+    request: Request,
+  ): Call = StrictModeCall(chain.newCall(request))
+
+  private class StrictModeCall(
+    private val delegate: Call,
+  ) : Call by delegate {
+    override fun execute(): Response {
+      if (Looper.getMainLooper().isCurrentThread) {
```
pyricau: What's OkHttp's min Android API? `Looper.isCurrentThread` is min 23. I saw a bunch of config for min API 21 in the okhttp repo but not sure what the actual target is.

If you need support for older APIs, you could use `if (Looper.getMainLooper() === Looper.myLooper())`

okhttp/src/commonJvmAndroid/kotlin/okhttp3/OkHttpClient.kt
JakeWharton: This is called a "fold", and if you are doing it from the end to the start it's a "foldRight".

```
private val decoratedCallFactory by lazy {
  callDecorators.foldRight(
    initial = {
      Call.Factory { request -> RealCall(this, request, forWebSocket = false)
    },
    operation = { callDecorator, next ->
      Call.Factory { request -> callDecorator.newCall(next, request) }
    },
  )
}
```
This also uses a single list iterator allocation (which can iterate backwards) rather than allocating a new List with the reversed decorators and then allocating a regular iterator.

It also does an `isEmpty()` check to avoid the iterator allocation altogether in the common case of zero decorators.

I would also consider dropping the `lazy`. The common case is for this to be list to be empty or maybe contain a single element, so the eager allocation burden is basically teh same as allocating the lazy itself.

pyricau: Neat! I initially started exploring one of the kotlin collections functional extensions but couldn't remember what it was I was looking for..

good point on lazy too.

android-test/src/androidTest/java/okhttp/android/test/OffMainThread.kt
yschimke: I'll update. Note, this is just a sample. Not shipped library code.

yschimke:
@pyricau I'm thinking I might need access to the OkHttpClient in my decorator to spawn new derived instances. 

Will the current API meet your needs?

android-test/src/androidTest/java/okhttp/android/test/OffMainThread.kt
pyricau: oooh I missed that, good point.

pyricau:
> @pyricau I'm thinking I might need access to the OkHttpClient in my decorator to spawn new derived instances.
> 
> Will the current API meet your needs?

It does yes, I only need access to the request.

Technically you can actually access the client, by forwarding `newCall()` then retrieving the client from RealCall.. not sure it's ideal though.

yschimke:
@pyricau in my case, I need to fork off new instances of OkHttpClient, so can't let the chain proceed normally.

https://github.com/square/okhttp/pull/8376/files#diff-24d566d17e597cae3dc5320b45de67927c45576c79e80c069114a4d0969100e5

```
  fun OkHttpClient.Builder.withNetwork(network: Network?): OkHttpClient.Builder {
    return if (network == null) {
      dns(Dns.ANDROID)
        .socketFactory(SocketFactory.getDefault())
    } else {
      dns(Dns.forNetwork(network))
        .socketFactory(AndroidSocketFactory(network))
    }
  }
```

yschimke:
@pyricau this is the case I want to get working https://github.com/square/okhttp/pull/8376

So I'll introduce a Chain interface that maps more closely to the Interceptor Chain.

yschimke:
@harrytmthy I think you were right, so I updated the docs to say 

> A Decorator that changes the OkHttpClient should typically retain later decorators in the new client.

example from https://github.com/square/okhttp/pull/8376

```
  private fun OkHttpClient.withNetwork(network: Network): OkHttpClient =
    newBuilder()
      .dns(AndroidDns(network))
      .socketFactory(network.socketFactory)
      .apply {
        // Keep decorators after this one in the new client
        val indexOfThisDecorator = callDecorators.indexOf(this@AndroidNetworkPinning)
        callDecorators.subList(0, indexOfThisDecorator + 1).clear()
      }.build()
```

yschimke:
Replacing by next attempt https://github.com/square/okhttp/pull/9148

PR #8802, opened 2025-05-27
Use SHOUTY_CASE for static constants

swankjesse:
OkHttp has lots of constants defined from before it was a Kotlin library, and I'd rather the library be self-consistent than consistent with Kotlin idioms.

As a bonus we get to remain consistent with Java idioms. I'd love a Kotlin mechanism to use @JvmField and @JvmName at the same time, but I can't figure out how.

### DIFF
--- a/okhttp/src/commonJvmAndroid/kotlin/okhttp3/RequestBody.kt
+++ b/okhttp/src/commonJvmAndroid/kotlin/okhttp3/RequestBody.kt
@@ -101,17 +101,9 @@ abstract class RequestBody {
   open fun isOneShot(): Boolean = commonIsOneShot()
 
   companion object {
-    /** Empty request body. */
+    /** Empty request body with no content-type. */
     @JvmField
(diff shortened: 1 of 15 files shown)


okhttp/src/commonJvmAndroid/kotlin/okhttp3/RequestBody.kt
```diff
@@ -101,17 +101,9 @@ abstract class RequestBody {
   open fun isOneShot(): Boolean = commonIsOneShot()
 
   companion object {
-    /** Empty request body. */
+    /** Empty request body with no content-type. */
     @JvmField
-    val Empty: RequestBody = EmptyRequestBody()
-
-    private class EmptyRequestBody : RequestBody() {
```
swankjesse, line 108: Not defining an extra class to keep OkHttp small

JakeWharton (approved):

PR #8809, opened 2025-05-27
Fix a failing test by adding a FollowUpDecision event

swankjesse:


JakeWharton (approved):
Why don't these tests run by default?

PR #8813, opened 2025-05-27
Start using BufferedSource.indexOf() with a toIndex

swankjesse:
This avoids the need for the new FixedLengthSource class.


okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/FixedLengthSource.kt
```diff
@@ -1,80 +0,0 @@
...
-@JvmOverloads
-internal fun Source.limit(
-  byteCount: Long,
-  onReadExhausted: (eof: Boolean) -> Unit = {},
-): Source {
-  require(byteCount >= 0) { "byteCount < 0: $byteCount" }
-  return FixedLengthSource(this, byteCount, onReadExhausted, truncate = true)
-}
-
-internal class FixedLengthSource(
```
JakeWharton, line 41: This type already existed anyway https://github.com/square/okhttp/blob/master/okhttp%2Fsrc%2FcommonJvmAndroid%2Fkotlin%2Fokhttp3%2Finternal%2Fhttp1%2FHttp1ExchangeCodec.kt and should have been reused not copied.

JakeWharton (approved):

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/FixedLengthSource.kt
swankjesse: Yup.

yschimke (approved):

PR #8816, opened 2025-05-28
fix immutableListOf array

MukjepScarlet:


okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/-UtilJvm.kt
```diff
@@ -249,14 +249,24 @@ internal inline fun <T> Set<T>.unmodifiable(): Set<T> = Collections.unmodifiable
 internal inline fun <K, V> Map<K, V>.unmodifiable(): Map<K, V> = Collections.unmodifiableMap(this)
 
 /** Returns an immutable copy of this. */
-internal inline fun <reified T> List<T>.toImmutableList(): List<T> = this.toTypedArray().toImmutableList()
+internal inline fun <reified T> List<T>.toImmutableList(): List<T> =
```
JakeWharton, line 252: This is a lot of code to inline. We can eliminate the need for `inline`+`reified` by doing something like `ArrayList(this).unmodifiable()` instead.

MukjepScarlet: Hi! I ran some benchmarks of these wrap methods:
```kotlin
    inline fun <reified T> List<T>.unmodifiable1(): List<T> = Collections.unmodifiableList(this.toTypedArray().asList())

    fun <T> List<T>.unmodifiable2(): List<T> = Collections.unmodifiableList(ArrayList(this))

    @Suppress("UNCHECKED_CAST")
    fun <T> List<T>.unmodifiable3(): List<T> = Collections.unmodifiableList((this as java.util.Collection<*>).toArray().asList()) as List<T>
```
```kotlin
    // --- Conversion Performance ---
    @Benchmark
    fun creation1(): List<Int> {return sourceList.unmodifiable1()}

    @Benchmark
    fun creation2(): List<Int> {return sourceList.unmodifiable2()}

    @Benchmark
    fun creation3(): List<Int> {return sourceList.unmodifiable3()}

    // --- Access Performance ---
    @Benchmark
    fun access1(): Int {return unmodifiable1[500]}

    @Benchmark
    fun access2(): Int {return unmodifiable2[500]}

    @Benchmark
    fun access3(): Int {return unmodifiable3[500]}
```
Result:
```
Benchmark                          Mode  Cnt    Score    Error  Units
ListConversionBenchmark.access1    avgt   10    1.987 ±  0.195  ns/op
ListConversionBenchmark.access2    avgt   10    2.351 ±  0.114  ns/op
ListConversionBenchmark.access3    avgt   10    1.904 ±  0.089  ns/op
ListConversionBenchmark.creation1  avgt   10  837.129 ± 72.220  ns/op
ListConversionBenchmark.creation2  avgt   10  447.069 ± 84.947  ns/op
ListConversionBenchmark.creation3  avgt   10  435.755 ± 41.612  ns/op
```
The third one is best, and doesn't need to inline anything (but it looks weird), I want to ask for your opinion.

JakeWharton: Sure, third is fine. Just leave a one-line comment maybe saying it's for performance reasons.

dsingley:
👍🏻 I believe that this PR will fix an `UnsupportedOperation` error I've encountered in my Java project that transitively depends on `okhttp-tls` https://github.com/dsingley/jwt/pull/110

```
Error:  Errors: 
Error:    IntegrationTests.setUp:43 » UnsupportedOperation This function has a reified type parameter and thus can only be inlined at compilation time, not called directly.
Error:    KeysTest$CreateSSLSocketFactory.should_supportClientAuthentication:116 » UnsupportedOperation This function has a reified type parameter and thus can only be inlined at compilation time, not called directly.
```

PR #8817, opened 2025-05-28
Change the locking mechanism to 'synchronized'

squarejesse:
We migrated to ReentrantReadWriteLock to better support Loom, but now we can use synchronized and still support Loom. This old mechanism may be faster than ReentrantReadWriteLock.

### DIFF
--- a/okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/concurrent/Lockable.kt
+++ b/okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/concurrent/Lockable.kt
@@ -1,0 +1,79 @@
+/*
+ * Copyright (C) 2025 Block, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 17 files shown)


okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/concurrent/Lockable.kt
```diff
@@ -0,0 +1,79 @@
...
+ * limitations under the License.
+ */
+@file:Suppress("PLATFORM_CLASS_MAPPED_TO_KOTLIN", "NOTHING_TO_INLINE")
+
+package okhttp3.internal.concurrent
+
+import okhttp3.internal.assertionsEnabled
+
+/**
+ * Marker interface for objects that use the JVM's `synchronized` mechanism and th related
```
yschimke, line 23:
```suggestion
 * Marker interface for objects that use the JVM's `synchronized` mechanism and the related
```

yschimke (approved):

PR #8829, opened 2025-05-30
TrailersSource, a new public API for HTTP trailers

swankjesse:
This doesn't change much.

The new interface replaces a Kotlin lambda that had similar behavior, though that lambda wasn't specified as thoroughly.

We now permit calls to Response.trailers() before the entire response body is consumed. But when such calls are made the response body is discarded first. This could be a bit of a surprise to users, especially users who do an additional layer of buffering over their ResponseBody because it could yield latent failures.

But I think it's better behavior overall, especially since it's possible data binding layers might not consume the entire response body in all cases. For example, a JSON library might not read to the very end of the response if it has received a terminating '}' byte.

The riskiest part of this change is when ResponseBodySource self-reports as complete. Previously it would self-report as complete when it returned as many bytes as promised in the Content-Length. With this change it will now only report itself complete when it receives a definitive EOF on the stream, signaled by a -1 byte count on a read. This is because only when the EOF is received can we be sure that the trailers are received, and we must not unhook the Exchange from the Call until that happens. The previous behavior could make Call.cancel() no-op even if we were blocked waiting on trailers.

### DIFF
--- a/okhttp/src/commonJvmAndroid/kotlin/okhttp3/Response.kt
+++ b/okhttp/src/commonJvmAndroid/kotlin/okhttp3/Response.kt
@@ -28,6 +28,7 @@ import okhttp3.internal.connection.Exchange
 import okhttp3.internal.http.HTTP_PERM_REDIRECT
 import okhttp3.internal.http.HTTP_TEMP_REDIRECT
 import okhttp3.internal.http.parseChallenges
+import okhttp3.internal.skipAll
 import okio.Buffer
 
 /**
@@ -109,7 +110,7 @@ class Response internal constructor(
    */
   @get:JvmName("receivedResponseAtMillis") val receivedResponseAtMillis: Long,
   @get:JvmName("exchange") internal val exchange: Exchange?,
-  private var trailersFn: (() -> Headers),
+  private var trailersSource: TrailersSource,
 ) : Closeable {
   internal var lazyCacheControl: CacheControl? = null
 
@@ -176,11 +177,24 @@ class Response internal constructor(
   fun headers(): Headers = headers
 
   /**
-   * Returns the trailers after the HTTP response, which may be empty. It is an error to call this
-   * before the entire HTTP response body has been consumed.
+   * Returns the trailers after the HTTP response, which may be empty. This blocks until the
+   * trailers are available to read.
+   *
+   * It is not safe to call this concurrently with code that is processing the response body. If you
+   * call this without consuming the complete response body, any remaining bytes in the response
+   * body will be discarded before trailers are returned.
+   *
+   * If [Call.cancel] is called while this is blocking, this call will immediately throw.
+   *
+   * @throws IllegalStateException if the response is closed.
+   * @throws IOException if the trailers cannot be loaded, such as if the network connection is
+   *     dropped.
    */
   @Throws(IOException::class)
-  fun trailers(): Headers = trailersFn()
+  fun trailers(): Headers {
+    body.source().skipAll()
+    return trailersSource.get()
+  }
 
   /**
    * Peeks up to [byteCount] bytes from the response body and returns them as a new response
@@ -329,7 +343,7 @@ class Response internal constructor(
     internal var sentRequestAtMillis: Long = 0
     internal var receivedResponseAtMillis: Long = 0
     internal var exchange: Exchange? = null
-    internal var trailersFn: (() -> Headers) = { Headers.headersOf() }
+    internal var trailersSource: TrailersSource = TrailersSource.EMPTY
 
     constructor() {
       headers = Headers.Builder()
@@ -349,7 +363,7 @@ class Response internal constructor(
       this.sentRequestAtMillis = response.sentRequestAtMillis
       this.receivedResponseAtMillis = response.receivedResponseAtMillis
       this.exchange = response.exchange
-      this.trailersFn = response.trailersFn
+      this.trailersSource = response.trailersSource
     }
 
     open fun request(request: Request) =
@@ -445,9 +459,9 @@ class Response internal constructor(
       }
 
     @ExperimentalOkHttpApi
-    open fun trailers(trailersFn: (() -> Headers)): Builder =
+    open fun trailers(trailersSource: TrailersSource): Builder =
       apply {
-        this.trailersFn = trailersFn
+        this.trailersSource = trailersSource
       }
 
     open fun sentRequestAtMillis(sentRequestAtMillis: Long) =
@@ -462,7 +476,7 @@ class Response internal constructor(
 
     internal fun initExchange(exchange: Exchange) {
       this.exchange = exchange
-      this.trailersFn = { exchange.trailers() }
+      this.trailersSource = TrailersSource { exchange.trailers() }
     }
 
     open fun build(): Response {
@@ -481,7 +495,7 @@ class Response internal constructor(
         sentRequestAtMillis,
         receivedResponseAtMillis,
         exchange,
-        trailersFn,
+        trailersSource,
       )
     }
   }
--- a/okhttp/src/commonJvmAndroid/kotlin/okhttp3/TrailersSource.kt
+++ b/okhttp/src/commonJvmAndroid/kotlin/okhttp3/TrailersSource.kt
@@ -1,0 +1,30 @@
+/*
+ * Copyright (C) 2025 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okhttp3
+
+import okio.IOException
+
+@ExperimentalOkHttpApi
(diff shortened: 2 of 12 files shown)


okhttp/src/commonJvmAndroid/kotlin/okhttp3/Response.kt
```diff
@@ -176,11 +177,24 @@ class Response internal constructor(
...
-   * Returns the trailers after the HTTP response, which may be empty. It is an error to call this
-   * before the entire HTTP response body has been consumed.
+   * Returns the trailers after the HTTP response, which may be empty. This blocks until the
+   * trailers are available to read.
+   *
+   * It is not safe to call this concurrently with code that is processing the response body. If you
+   * call this without consuming the complete response body, any remaining bytes in the response
+   * body will be discarded before trailers are returned.
+   *
+   * If [Call.cancel] is called while this is blocking, this call will immediately throw.
```
swankjesse, line 187: keeping this promise was tricky!

okhttp/src/commonJvmAndroid/kotlin/okhttp3/TrailersSource.kt
```diff
@@ -0,0 +1,30 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okhttp3
+
+import okio.IOException
+
+@ExperimentalOkHttpApi
+fun interface TrailersSource {
+  /** Returns the trailers, blocking if they need to be loaded. */
```
swankjesse, line 22: There’s still a bug in the way this is wired up.

`Response.trailers()` discards the response body automatically, and then calls this function. What SHOULD be happening is this function itself should discard the response body. It’s not wired up that way because it’s a bigger change, but I might attempt that in a follow-up.

Not urgent for 5.0!

okhttp/src/jvmTest/kotlin/okhttp3/TrailersTest.kt
```diff
@@ -0,0 +1,361 @@
...
+        .addHeader("h1", "v1")
+        .trailers(headersOf("t1", "v2"))
+        .body(protocol, "Hello")
+        .build(),
+    )
+
+    val call = client.newCall(Request(server.url("/")))
+    call.execute().use { response ->
+      assertThat(response.header("h1")).isEqualTo("v1")
+      assertThat(response.trailers()).isEqualTo(headersOf("t1", "v2"))
```
swankjesse, line 126: * [ ] assert response body is discarded

okhttp/src/jvmTest/kotlin/okhttp3/TrailersTest.kt
```diff
@@ -0,0 +1,361 @@
...
+    trailersAndCache(Protocol.HTTP_1_1)
+  }
+
+  @Test
+  @Disabled
+  fun trailersAndCacheHttp2() {
+    trailersAndCache(Protocol.HTTP_2)
+  }
+
+  private fun trailersAndCache(protocol: Protocol) {
```
swankjesse, line 142: TOTALLY DOESN’T WORK

okhttp/src/jvmTest/kotlin/okhttp3/TrailersTest.kt
```diff
@@ -0,0 +1,361 @@
...
+      val source = response.body.source()
+      assertThat(response.header("h1")).isEqualTo("v1")
+      assertThat(source.readUtf8()).isEqualTo("Hello")
+      assertThat(response.trailers()).isEqualTo(headersOf("t1", "v2"))
+    }
+  }
+
+  @Test
+  fun delayBeforeTrailersHttp1() {
+    delayBeforeTrailers(Protocol.HTTP_1_1)
```
swankjesse, line 174: This manual no-burst stuff is the worst. I haven’t done it here 'cause I’m afraid of how our MockWebServer JUnit 5 thing will interact with Burst.

swankjesse: (More precisely, Burst is broken when you do fancy things with JUnit 5)

okhttp/src/jvmTest/kotlin/okhttp3/TrailersTest.kt
```diff
@@ -0,0 +1,361 @@
...
+    call.execute().use { response ->
+      val source = response.body.source()
+      assertThat(response.header("h1")).isEqualTo("v1")
+      assertThat(source.readUtf8(5)).isEqualTo("Hello")
+      val trailersDelay =
+        measureTime {
+          val trailers = response.trailers()
+          assertThat(trailers).isEqualTo(headersOf("t1", "v2"))
+        }
+      assertThat(trailersDelay).isGreaterThan(250.milliseconds)
```
swankjesse, line 206: Defending against weirdness like the client assuming there’s no trailers 'cause there’s no trailers yet

okhttp/src/jvmTest/kotlin/okhttp3/TrailersTest.kt
```diff
@@ -0,0 +1,361 @@
...
+      val source = response.body.source()
+      assertThat(response.header("h1")).isEqualTo("v1")
+      assertThat(source.readUtf8(5)).isEqualTo("Hello")
+      call.cancelLater(500.milliseconds)
+      val trailersDelay =
+        measureTime {
+          val exception =
+            assertFailsWith<IOException> {
+              val trailers = response.trailers()
+              println(trailers)
```
swankjesse, line 309: * [ ] fix

okhttp/src/jvmTest/kotlin/okhttp3/TrailersTest.kt
swankjesse: (Follow-up)

okhttp/src/commonJvmAndroid/kotlin/okhttp3/TrailersSource.kt
yschimke: What's the actual bug apart from code in the wrong place? Is it observable?

yschimke (approved):

okhttp/src/jvmTest/kotlin/okhttp3/TrailersTest.kt
swankjesse: I don’t know if we can fix it anywhere. Burst kind of assumes it can do its own naughty business with parameters, and it might not have enough information to leave the JUnit-5 managed ones alone.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/-UtilJvm.kt
```diff
@@ -175,6 +175,13 @@ internal fun Source.skipAll(
   }
 }
 
+@Throws(IOException::class)
+internal fun BufferedSource.skipAll() {
```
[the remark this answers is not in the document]
swankjesse: Maybe? It’s a weird operation to need.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/http1/Http1ExchangeCodec.kt
```diff
@@ -192,7 +192,6 @@ class Http1ExchangeCodec(
           .code(statusLine.code)
           .message(statusLine.message)
           .headers(headersReader.readHeaders())
-          .trailers { error("trailers not available") }
```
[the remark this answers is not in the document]
swankjesse: Yeah, definitely.

swankjesse:
To fix the tests I needed a to push a commit with non-trivial changes to `Http1ExchangeCodec`.

@yschimke Please take a second look?

okhttp/src/jvmTest/kotlin/okhttp3/TrailersTest.kt
```diff
@@ -0,0 +1,465 @@
...
+        .Builder()
+        .trailers(headersOf("t1", "v2"))
+        .body(protocol, "Hello")
+        .build(),
+    )
+
+    val call = client.newCall(Request(server.url("/")))
+    call.execute().use { response ->
+      assertThat(response.body.source().readUtf8()).isEqualTo("Hello")
+      response.body.source().close()
```
yschimke, line 324: Does this close change anything after the body was already consumed?

okhttp/src/jvmTest/kotlin/okhttp3/TrailersTest.kt
```diff
@@ -0,0 +1,465 @@
...
+  fun readTrailersAfterEarlyEofAndCloseHttp1() {
+    readTrailersAfterEarlyEofAndClose(Protocol.HTTP_1_1)
+  }
+
+  @Test
+  fun readTrailersAfterEarlyEofAndCloseHttp2() {
+    readTrailersAfterEarlyEofAndClose(Protocol.HTTP_2)
+  }
+
+  private fun readTrailersAfterEarlyEofAndClose(protocol: Protocol) {
```
yschimke, line 310: Maybe missing something obvious, but what makes this early EOF?

yschimke (approved):

okhttp/src/jvmTest/kotlin/okhttp3/TrailersTest.kt
swankjesse: Just good form.

okhttp/src/jvmTest/kotlin/okhttp3/TrailersTest.kt
swankjesse: This test name is trying too hard to differentiate itself from its neighbours. The caller is reading the end of the stream and closing it, as opposed to calling close() _after_ reading the trailers.

PR #8837, opened 2025-06-03
Finalize the new API for MockResponse

swankjesse:
It's a value object with a bunch of fields, and a builder.

### DIFF
--- a/okhttp/src/jvmTest/kotlin/okhttp3/CallTest.kt
+++ b/okhttp/src/jvmTest/kotlin/okhttp3/CallTest.kt
@@ -3971,8 +3971,8 @@ open class CallTest {
         .Builder()
         .code(HttpURLConnection.HTTP_PROXY_AUTH)
         .headers(headersOf("Proxy-Authenticate", "Basic realm=\"localhost\""))
-        .body("proxy auth required")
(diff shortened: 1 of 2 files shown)


mockwebserver/src/test/java/mockwebserver3/MockResponseSniTest.kt
```diff
@@ -197,7 +197,12 @@ class MockResponseSniTest {
         ).proxy(server.toProxyAddress())
         .build()
 
-    server.enqueue(MockResponse(inTunnel = true))
```
swankjesse, line 200: inTunnel is a weird property to include in the default overloads

mockwebserver/src/main/kotlin/mockwebserver3/MockResponse.kt
```diff
@@ -33,7 +33,6 @@ import okhttp3.internal.http2.Settings
 import okio.Buffer
 
 /** A scripted response to be replayed by the mock web server. */
-@ExperimentalOkHttpApi
```
swankjesse, line 36: This is the goal of the PR !

yschimke (approved):

PR #8838, opened 2025-06-04
Move gzip() from RequestBody to Request.Builder

squarejesse:
On Request.Builder the one function can add the header and apply compression. Otherwise the caller needs to keep those calls in sync.

One severe drawback of this approach is the calls to Request.Builder are now ordered. If you call gzip() before you call post(), it'll crash. Worse, if you call post() and then gzip() and then post() again
with a different body, it'll silently not compress the new body. I don't love this drawback but I think the mitigations aren't worth the effort.

### DIFF
--- a/okhttp/src/commonJvmAndroid/kotlin/okhttp3/RequestBody.kt
+++ b/okhttp/src/commonJvmAndroid/kotlin/okhttp3/RequestBody.kt
@@ -24,9 +24,7 @@ import okhttp3.internal.chooseCharset
 import okio.BufferedSink
 import okio.ByteString
 import okio.FileSystem
-import okio.GzipSink
 import okio.Path
-import okio.buffer
 import okio.source
 
 abstract class RequestBody {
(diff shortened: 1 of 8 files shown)


okhttp/src/commonJvmAndroid/kotlin/okhttp3/RequestBody.kt
```diff
@@ -257,35 +255,5 @@ abstract class RequestBody {
...
-     *
-     * ```
-     * val request = Request.Builder().url("...")
-     *  .addHeader("Content-Encoding", "gzip")
-     *  .post(uncompressedBody.gzip())
-     *  .build()
-     * ```
-     */
-    @JvmStatic
-    @ExperimentalOkHttpApi
```
squarejesse, line 273: Dealing with this annotation is what brought me here

JakeWharton:
Should it be on `Request` maybe? That way you have to do something to `toBuilder()` to mess it up?

squarejesse:
Yeah, `Request.gzip()` is tempting. But I would like that as an additional API, not as a replacement for this...

I think it’s analogous to `MutableList` and `List`. We’ve got `MutableList.sort()` which is in-place, and `List.sorted()` which returns a copy.

We could live without `MutableList.sort()`, but it would be less efficient.

JakeWharton (approved):

PR #8844, opened 2025-06-05
Add a test and fix for non-robolectric android test.

yschimke:
If Robolectric not active, then unit tests for Android will fail annoyingly.


yschimke:
@swankjesse requesting post review. Test only.

PR #8849, opened 2025-06-10
Add support for HTTP/1.1 protocol upgrades

gesellix:
Based on the discussion in https://github.com/square/okhttp/issues/5874

See [RFC 7230 section 6.7](https://datatracker.ietf.org/doc/html/rfc7230#section-6.7) and [RFC 9110 section 15.2.2](https://tools.ietf.org/html/rfc9110#name-101-switching-protocols).

The first two commits are not related and can be submitted as separate pull request.

This is only a draft to continue the discussion started in https://github.com/square/okhttp/issues/5874 and not ready for an actual merge, yet.

### DIFF
--- a/okhttp/src/jvmTest/kotlin/okhttp3/internal/http/HttpUpgradesTest.kt
+++ b/okhttp/src/jvmTest/kotlin/okhttp3/internal/http/HttpUpgradesTest.kt
@@ -96,12 +96,13 @@ class HttpUpgradesTest {
   @Test
   fun upgradeRefusedByServer() {
     server.enqueue(MockResponse(body = "normal request"))
-    val requestWithUpgrade = Request
-      .Builder()
-      .url(server.url("/"))
-      .header("Connection", "upgrade")
-      .header("Upgrade", "tcp")
-      .build()
+    val requestWithUpgrade =
+      Request
+        .Builder()
+        .url(server.url("/"))
+        .header("Connection", "upgrade")
+        .header("Upgrade", "tcp")
+        .build()
     val response = client.newCall(requestWithUpgrade).execute()
     response.body.string()
     assertThat(response.code).isEqualTo(200)
@@ -123,15 +124,17 @@ class HttpUpgradesTest {
             "Content-Type",
             "text/plain; charset=UTF-8",
           ),
-        ).socketHandler(      MockSocketHandler())
-        .build())
+        ).socketHandler(MockSocketHandler())
+        .build(),
+    )
     val request = Request(server.url("/"))
-    val requestWithUpgrade = Request
-      .Builder()
-      .url(server.url("/"))
-      .header("Connection", "upgrade")
-      .header("Upgrade", "tcp")
-      .build()
+    val requestWithUpgrade =
+      Request
+        .Builder()
+        .url(server.url("/"))
+        .header("Connection", "upgrade")
+        .header("Upgrade", "tcp")
+        .build()
     assertConnectionReused(request, requestWithUpgrade)
   }
 


okhttp/src/jvmTest/kotlin/okhttp3/CallTest.kt
```diff
@@ -4840,6 +4848,149 @@ open class CallTest {
     }
   }
 
+  // https://github.com/square/okhttp/issues/5874
+  // call.timeoutEarlyExit()
+  // https://docs.docker.com/engine/api/v1.43/#tag/Container/operation/ContainerAttach
+  // https://learn.microsoft.com/en-us/windows/win32/ipc/named-pipe-client
+  // docker run --rm -it --name attach alpine:edge top -b
+  @Test
+  fun upgradeConnection() {
```
gesellix: I'm not happy with this one, maybe you have better ideas how to test the connection upgrade?

gesellix:
This is now rebased on the current master to use the new okio.Socket.

mockwebserver/src/main/kotlin/mockwebserver3/MockWebServer.kt
```diff
@@ -849,6 +856,23 @@ class MockWebServer : Closeable {
 
     writeHeaders(sink, response.headers)
 
+    if (response.socketHandler != null) {
+      response.socketHandler.handle(
+        object : okio.Socket {
+          override val source: BufferedSource
```
yschimke: I'm not clear who can access this as a BufferedSource for an anonymous object impl. Wouldn't clients still have to Buffer again to be sure?

mockwebserver/src/main/kotlin/mockwebserver3/internal/duplex/MockSocketHandler.kt
```diff
@@ -82,7 +82,7 @@ class MockSocketHandler : SocketHandler {
   @JvmOverloads
   fun sendResponse(
     s: String,
-    responseSent: CountDownLatch = CountDownLatch(0),
+    responseSent: CountDownLatch = CountDownLatch(1),
```
yschimke: Is this fixing an unrelated bug?

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/connection/RealConnection.kt
```diff
@@ -312,6 +314,23 @@ class RealConnection internal constructor(
...
+    noNewExchanges()
+    return object : okio.Socket {
+      override val source: BufferedSource
+        get() = socket.source().buffer()
+
+      override val sink: BufferedSink
+        get() = socket.sink().buffer()
+
+      override fun cancel() {
+        socket.sink().flush()
```
yschimke: Seems to contradict the API

https://github.com/square/okio/blob/bcd25f94b63723a9e90cb74d6b76fff16acfed34/okio/src/commonMain/kotlin/okio/Socket.kt#L44

```
Fail any in-flight and future operations.
```

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Response.kt
```diff
@@ -78,6 +79,10 @@ class Response internal constructor(
    * all instances of [ResponseBody].
    */
   @get:JvmName("body") val body: ResponseBody,
+  /**
+   * Non-null if this response is a successful upgrade ...
+   */
+  @get:JvmName("socket") val socket: Socket?,
```
yschimke, line 84: I'm a bit unclear why this is enough to consider upgrades handled. Is this only for TCP? or do you think that all protocols would be handled as a Socket?

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Upgrade#upgrade_header_with_multiple_protocols

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/http/CallServerInterceptor.kt
```diff
@@ -125,14 +125,27 @@ class CallServerInterceptor(
...
           response.stripBody()
         } else {
-          response
-            .newBuilder()
-            .body(exchange.openResponseBody(response))
-            .build()
+          if (code == HTTP_SWITCHING_PROTOCOLS &&
+            "upgrade".equals(response.request.header("Connection"), ignoreCase = true) &&
+            "upgrade".equals(response.header("Connection"), ignoreCase = true) &&
+            "tcp".equals(response.request.header("Upgrade"), ignoreCase = true) &&
```
yschimke: The PR title feels like it's about all possible protocols, but in practice seems tied to solving TCP specifically for docker?

Is there a spec for this TCP upgrade?

yschimke:
I'm struggling to find any implementations of the TCP upgrade in popular clients like Curl, Netty, Requests etc.

Can you link to any?

And wondering whether this really needs a more thorough implementation of the spec, which allows for negotation.

Example from Netty https://github.com/netty/netty/blob/42d83b8081ab3d018f5d7611973e5d361cdd28d1/codec-http/src/main/java/io/netty/handler/codec/http/HttpClientUpgradeHandler.java#L40

mockwebserver/src/main/kotlin/mockwebserver3/MockWebServer.kt
gesellix: I guess I don't understand your question. Are you implying that this should be changed to a non-buffered Source?

mockwebserver/src/main/kotlin/mockwebserver3/internal/duplex/MockSocketHandler.kt
gesellix: According to the Javadoc, a CountDownLatch with a zero value doesn't have any effect. I didn't experience a bug.

This is unrelated to the current pull request, though. I referred to this one as one of the two commits which are both unrelated and can easily be moved to a dedicated pull request.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/connection/RealConnection.kt
gesellix: You're right, I didn't check the api thoroughly. I'm going to remove this one.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Response.kt
gesellix: I must admit that I only considered TCP upgrades and ignored other options. I suppose that the `socket` will only be necessary (and non-null) for upgrades which are not supported through another implementation (e.g. HTTP2). Returning the plain Socket might be considered as a generic way of supporting any protocol. That said: I'm unsure if this would make sense for OkHttp in general or if you prefer to keep supported protocols explicitly stated anywhere.
So, would it be enough to improve the Javadoc to be more specific?

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/http/CallServerInterceptor.kt
gesellix: You're right, I wasn't aware of the more general possibilities for a connection upgrade, and the case for Websockets hasn't changed with my pull request.

My current aim is in fact to add support for TCP upgrades, because my use case is driven by the Docker engine api.
There's no spec for the TCP upgrade per se, at least not known to me.

Given the issue at https://github.com/square/okhttp/issues/5874 I would suggest to start with support for `tcp` and `websocket`. Other protocols might be added in the future.

mockwebserver/src/main/kotlin/mockwebserver3/MockWebServer.kt
yschimke: Yeah, are you relying on this being buffered here? Or also when used it's applied

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Response.kt
yschimke: I think Jesse should weigh in. I don't think this is a simple feature to add especially given your TCP  case is a poorly specced docker variant of a more general feature.

gesellix:
> I'm struggling to find any implementations of the TCP upgrade in popular clients like Curl, Netty, Requests etc.
> 
> Can you link to any?
> 
> And wondering whether this really needs a more thorough implementation of the spec, which allows for negotation.
> 
> Example from Netty https://github.com/netty/netty/blob/42d83b8081ab3d018f5d7611973e5d361cdd28d1/codec-http/src/main/java/io/netty/handler/codec/http/HttpClientUpgradeHandler.java#L40

There is a use case _with_ the Apache Http Client, so not exactly an example _in_ a client: https://github.com/docker-java/docker-java/blob/20f08311d1613d8819c4548f34046b0557a43117/docker-java-transport-httpclient5/src/main/java/com/github/dockerjava/httpclient5/HijackingHttpRequestExecutor.java
The official Docker Clients ([docker-client](https://github.com/moby/moby/blob/cab4ac834e8bf36aa38a2ca49599773df6e6805a/client/hijack.go#L53), [docker-py](https://github.com/docker/docker-py/blob/6e6a273573fe77f00776b30de0685162a102e43f/docker/api/container.py#L53-L62)) implement the TCP upgrade described at https://docs.docker.com/reference/api/engine/version/v1.50/#tag/Container/operation/ContainerAttach

Looking at https://datatracker.ietf.org/doc/html/rfc9110#field.upgrade there's at least the `Expect` header mentioned. Is this what you mean by negotiation?

As far as I understand the Netty implementation, it handles the connection upgrade more or less by "passing through" and only checking response headers for the requested protocol. I didn't dig deeper into how exactly it would allow the client to handle the upgraded connection.
Maybe OkHttp could select some of well-known or "popular" protocols (e.g. Websocket) for specific upgrade handling and fall back to the more generic Socket for unknown or non-officially supported protocols?

mockwebserver/src/main/kotlin/mockwebserver3/MockWebServer.kt
gesellix: I don't rely on this being buffered, so I commited a cleanup 👍

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/connection/RealConnection.kt
gesellix: This has been removed in RealConnection as well as in the MockWebServer.

gesellix: ... is it correct and sufficient to cancel the Exchange?

yschimke:
Thanks, that's useful.

Yeah there are maybe three paths 

1) special case each transport we support (TCP, websocket)
2) websocket or okio.Socket (TCP + other)
3) design a proper upgrade API

gesellix:
Given the release of OkHttp 5, and this one probably changing the api, is there still a chance to have some kind of minimal implementation for the use case of OkHttp as client for a Docker Engine and the tcp upgrade? I guess there's no other chance of using OkHttp without changing its api? I'm not afraid of maintaining some kind of extension or module so that you wouldn't have to officially support such a feature, but I have no idea how I could implement such a module without maintaining a full OkHttp fork by myself.
A minimal support for specific use cases of the HTTP/1.1 protcol upgrade would be really great. I'm such a big fan of OkHttp :)

yschimke:
@gesellix I think it's worth supporting, but I'd like some other opinions on it also.

I guess what is complicated here is the PR is a partial implementation that covers a non-standardised but valid case. While the standardized cases aren't supported.

gesellix:
So, how much of a standard use case (which ones?) would be required to put some weight or clarity into it? Maybe this is not the right timing to push such a feature and I assume that you don't have other users asking for anything with a connection upgrade (on top of existing implementations for Websocket and HTTP/2).

Do you have an indication which standard use cases make sense? I could try to prepare the required changes either with this pull request or in a separate one. I might only need some guidance on the interfaces and testing if that would work for you.

yschimke:
If we have a clean mechanism, and this is currently the only supported protocol, but we can add more.  That is probably ok.

But if the public API of things like Response change in a way that has a bunch of assumptions that only hold for this TCP docker case, it would be a mistake.

@swankjesse if you have opinions, feel free to weigh in.

swankjesse:
I’d like to ship this.

I’d like to review this PR and the RFCs to figure out exactly what’s best for our next step and I intend to do so.

gesellix:
Thanks for looking into it.
I can merge the current main branch or rebase on top of it to resolve the conflicts, if you like.

gesellix:
The branch has been rebased now (I had to make it twice to make spotless happy).

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/http/CallServerInterceptor.kt
```diff
@@ -128,27 +128,40 @@ class CallServerInterceptor(
...
+        if (forWebSocket && code == HTTP_SWITCHING_PROTOCOLS) {
           // Connection is upgrading, but we need to ensure interceptors see a non-null response body.
           response.stripBody()
         } else {
-          val responseBody = exchange.openResponseBody(response)
-          response
-            .newBuilder()
-            .body(responseBody)
-            .trailers(
-              object : TrailersSource {
```
gesellix: Oh wait, my rebase has gone wrong. I'm going to fix this.

gesellix: ... actually, it looked only strange in the GitHub diff view due to the code formatting. Everything's fine.

swankjesse:
I think running `./gradlew apiDump` will fix the test failure.

mockwebserver/src/main/kotlin/mockwebserver3/MockWebServer.kt
```diff
@@ -865,6 +873,22 @@ public class MockWebServer : Closeable {
...
+          override val source: Source
+            get() = socket.source()
+          override val sink: Sink
+            get() = socket.sink()
+
+          override fun cancel() {
+            socket.closeQuietly()
+          }
+        },
+      )
```
swankjesse: Rather unfortunately, the code to go from a java.net.Socket to an Okio Socket is a lot more involved to support independent `Socket.close()` calls. 

So you gotta use `Socket.asOkioSocket()` here.

```suggestion
      response.socketHandler.handle(
        socket.asOkioSocket()
      )
```

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Response.kt
swankjesse: My understanding is that on a protocol upgrade, OkHttp should hand off the `okio.Socket` to the application layer, and it can do whatever new protocol it wants on top. Our responsibility for the upgrade ends and handing off the socket.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/connection/RealConnection.kt
```diff
@@ -312,6 +314,22 @@ class RealConnection internal constructor(
...
+      override val source: Source
+        get() = socket.source()
+
+      override val sink: Sink
+        get() = socket.sink()
+
+      override fun cancel() {
+        exchange.cancel()
+      }
+    }
```
swankjesse:
```suggestion
    return socket.asOkioSocket()
```

okhttp/src/jvmTest/kotlin/okhttp3/CallTest.kt
yschimke: I wonder if you can test any of this with container-tests?

https://github.com/square/okhttp/blob/master/container-tests/src/test/java/okhttp3/containers/SocksProxyTest.kt#L81

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/http/CallServerInterceptor.kt
```diff
@@ -128,27 +128,40 @@ class CallServerInterceptor(
...
-                override fun peek() = exchange.peekTrailers()
+          if (code == HTTP_SWITCHING_PROTOCOLS &&
```
swankjesse: We probably also need _something_ here to confirm it’s an HTTP/1 connection and not an HTTP/2 connection. We’ll implement that in a follow-up PR!

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/http/CallServerInterceptor.kt
swankjesse: I actually think we might make things easy for ourselves by omitting the `tcp` check here. We can handle all upgrades this way, and let the caller decide what they want to do with their socket based on the contents of the `Upgrade` header.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/http/HttpHeaders.kt
```diff
@@ -225,7 +225,7 @@ fun Response.promisesBody(): Boolean {
   }
 
   val responseCode = code
-  if ((responseCode < HTTP_CONTINUE || responseCode >= 200) &&
+  if ((responseCode < HTTP_CONTINUE || responseCode == HTTP_SWITCHING_PROTOCOLS || responseCode >= 200) &&
```
swankjesse, line 228: I don’t think I agree that a protocol change is a body? This is used to drive stuff like the gzip interceptor, which seems inappropriate here?

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/http/CallServerInterceptor.kt
gesellix: According to https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Upgrade the `Upgrade` header is forbidden for HTTP/2. I didn't verify with the RFC, yet.
How strictly does OkHttp enforce usage conform to HTTP standards?

Also, the Websocket upgrade can be regarded as special case of the common HTTP/1.1 connection upgrade - is there a chance to upgrade HTTP/1.0 to Websocket?

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/http/CallServerInterceptor.kt
gesellix: I removed the `tcp` check as you suggested. I also changed the connection upgrade check to make the similarities between Websocket and generic upgrades more visible. This might help with the next steps/discussion.

mockwebserver/src/main/kotlin/mockwebserver3/MockWebServer.kt
gesellix: TIL. I changed the implementation as you suggested.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/http/HttpHeaders.kt
gesellix: The protocol change itself doesn't promise a body, but would depend on the actual protocl _after_ a successful upgrade.
I think this change was some kind of workaround to run into the "newUnknownLengthSource(response.request.url)" case at https://github.com/gesellix/okhttp/blob/0984bc3fc5d41c4814b953194d10e5ee77361f7d/okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/http1/Http1ExchangeCodec.kt#L145
Maybe there's a better way?

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Response.kt
gesellix: Websockets might be the only exception?

okhttp/src/jvmTest/kotlin/okhttp3/CallTest.kt
gesellix: Thanks, I wasn't aware of those. I'll look into it!

gesellix:
The pull request will be continued at https://github.com/square/okhttp/pull/8969

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/http/CallServerInterceptor.kt
gesellix: HTTP/2 verification is adressed in https://github.com/square/okhttp/pull/8969

PR #8857, opened 2025-06-14
Stabilize the JUnit 4 API

swankjesse:


mockwebserver/src/main/kotlin/mockwebserver3/MockWebServer.kt
```diff
@@ -400,7 +400,6 @@ public class MockWebServer : Closeable {
     }
   }
 
-  @Throws(IOException::class)
```
swankjesse, line 404: Throwing `IOException` here just isn’t useful!

mockwebserver/src/main/kotlin/mockwebserver3/MockWebServer.kt
```diff
@@ -409,12 +408,12 @@ public class MockWebServer : Closeable {
...
 
     // Cause acceptConnections() to break out.
-    serverSocket.close()
+    serverSocket.closeQuietly()
 
     // Await shutdown.
     for (queue in taskRunner.activeQueues()) {
       if (!queue.idleLatch().await(5, TimeUnit.SECONDS)) {
-        throw IOException("Gave up waiting for queue to shut down")
+        throw AssertionError("Gave up waiting for queue to shut down")
```
swankjesse, line 417: If this occurs in practice . . . the system is unhealthy. I think crashing hard like this is okay.

yschimke (approved):

PR #8858, opened 2025-06-15
Fix a liveness bug in MockWebServer

swankjesse:
It was possible to configure delays that would be honored after MockWebServer was shut down. With this fix any delay will be abandoned once the server is closed.

### DIFF
--- a/mockwebserver/src/main/kotlin/mockwebserver3/MockWebServer.kt
+++ b/mockwebserver/src/main/kotlin/mockwebserver3/MockWebServer.kt
@@ -66,7 +66,7 @@ import mockwebserver3.SocketPolicy.StallSocketAtStart
 import mockwebserver3.internal.RecordedRequest
 import mockwebserver3.internal.ThrottledSink
 import mockwebserver3.internal.TriggerSink
-import mockwebserver3.internal.sleepNanos
+import mockwebserver3.internal.sleepWhileOpen
 import okhttp3.Headers
 import okhttp3.Headers.Companion.headersOf
 import okhttp3.HttpUrl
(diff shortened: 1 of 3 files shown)


mockwebserver/src/main/kotlin/mockwebserver3/internal/SleepNanos.kt
```diff
@@ -16,9 +16,20 @@
...
+  var ms = nanos / 1_000_000L
   val ns = nanos - (ms * 1_000_000L)
+
+  while (ms > 100) {
+    Thread.sleep(100)
+    if (isClosed) throw InterruptedIOException("socket closed")
+    ms -= 100L
+  }
+
   if (ms > 0L || nanos > 0) {
```
JakeWharton:
```suggestion
  if (ms > 0L || ns > 0) {
```

JakeWharton (approved):

mockwebserver/src/main/kotlin/mockwebserver3/internal/SleepNanos.kt
swankjesse: Nice catch!

mockwebserver/src/main/kotlin/mockwebserver3/MockWebServer.kt
```diff
@@ -899,6 +899,7 @@ public class MockWebServer : Closeable {
     if (policy.throttlePeriodNanos > 0L) {
       result =
         ThrottledSink(
+          socket = socket,
```
yschimke, line 902: Not for this review, but should we be aiming for all operations like this to be on okio socket?

yschimke (approved):

mockwebserver/src/main/kotlin/mockwebserver3/MockWebServer.kt
swankjesse: Maybe yeah!

PR #8860, opened 2025-06-16
Decompose the request line into individual properties

squarejesse:
Replace the 'path' property with the URL's encodedPath property.

Rename the 'requestUrl' property to 'url'.

### DIFF
--- a/mockwebserver/src/main/kotlin/mockwebserver3/internal/RecordedRequestFactory.kt
+++ b/mockwebserver/src/main/kotlin/mockwebserver3/internal/RecordedRequestFactory.kt
@@ -55,12 +55,12 @@ internal fun RecordedRequest(
   }
 
   val requestUrl =
-    if (requestLine.method == "CONNECT") {
-      "${socket.scheme}://${requestLine.target}/".toHttpUrl()
-    } else {
-      requestLine.target.toHttpUrlOrNull()
-        ?: requestUrl(socket, requestLine, headers)
+    when (requestLine.method) {
+      "CONNECT" -> "${socket.scheme}://${requestLine.target}/".toHttpUrlOrNull()
+      else -> null
     }
+      ?: requestLine.target.toHttpUrlOrNull()
+      ?: requestUrl(socket, requestLine, headers)
 
   return RecordedRequest(
     sequenceNumber = sequenceNumber,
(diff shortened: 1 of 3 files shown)


mockwebserver/src/main/kotlin/mockwebserver3/internal/RecordedRequestFactory.kt
```diff
@@ -52,49 +54,83 @@ internal fun RecordedRequest(
...
 }
+
+internal fun decodeRequestLine(requestLine: String?): RequestLine {
+  val parts = when {
+    requestLine != null -> requestLine.split(' ', limit = 3)
+    else -> return DEFAULT_REQUEST_LINE
+  }
+
+  if (parts.size != 3) {
+    throw ProtocolException("unexpected request line: $requestLine")
```
yschimke, line 89: Is this making it stricter, failing on bad input it would have worked around previously?

mockwebserver/src/main/kotlin/mockwebserver3/RecordedRequest.kt
```diff
@@ -42,10 +42,24 @@ public class RecordedRequest(
...
+   *
+   * For HTTP proxy requests this will be either an absolute-form string like
+   * `http://example.com/index.html` (HTTP proxy) or an authority-form string like
+   * `example.com:443` (HTTPS proxy).
+   *
+   * For OPTIONS requests, this may be an asterisk, `*`.
+   */
+  public val target: String,
+  /** A string like `HTTP/1.1`. */
+  public val version: String,
```
yschimke, line 60: Is httpVersion more immediately obvious?

Should we use types here?

mockwebserver/src/main/kotlin/mockwebserver3/internal/RecordedRequestFactory.kt
```diff
@@ -52,49 +54,83 @@ internal fun RecordedRequest(
...
+  }
```
oldergod, line 90: 👍

oldergod (approved):
I like the result

mockwebserver/src/main/kotlin/mockwebserver3/internal/RecordedRequestFactory.kt
squarejesse: Same strictness. Previously we special-cased `""` to discard malformed input, and now we just pass `DEFAULT_REQUEST_LINE`

mockwebserver/src/main/kotlin/mockwebserver3/RecordedRequest.kt
squarejesse: It’s an open-ended string and could be `ICY` or something if it’s a Shoutcast client? I think a String is a good fit for MockWebServer. I don’t think it’s a good fit for OkHttp!

PR #8861, opened 2025-06-17
Change some MockWebServer ergonomics

swankjesse:
Now calling MockWebServer.port and similar functions will not implicity start the server. It is instead necessary to explicitly call the start() function.

The problem with the old behavior was that reading the port field could fail with an IOException, which is weird behavior for reading a property.

### DIFF
--- a/mockwebserver-deprecated/src/main/kotlin/okhttp3/mockwebserver/MockWebServer.kt
+++ b/mockwebserver-deprecated/src/main/kotlin/okhttp3/mockwebserver/MockWebServer.kt
@@ -24,7 +24,6 @@ import java.util.logging.Level
 import java.util.logging.Logger
 import javax.net.ServerSocketFactory
 import javax.net.ssl.SSLSocketFactory
-import okhttp3.ExperimentalOkHttpApi
 import okhttp3.HttpUrl
 import okhttp3.Protocol
 import org.junit.rules.ExternalResource
@@ -32,7 +31,6 @@ import org.junit.rules.ExternalResource
 class MockWebServer :
   ExternalResource(),
   Closeable {
-  @ExperimentalOkHttpApi
   val delegate = mockwebserver3.MockWebServer()
 
   val requestCount: Int by delegate::requestCount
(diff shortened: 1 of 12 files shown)


mockwebserver-deprecated/src/main/kotlin/okhttp3/mockwebserver/MockWebServer.kt
```diff
@@ -84,27 +78,14 @@ class MockWebServer :
   }
 
   @JvmName("-deprecated_port")
-  @Deprecated(
```
swankjesse, line 87: UN-deprecating stuff from the old MockWebServer. This whole package is obsolete and these deprecation warnings aren’t benefiting anyone.

mockwebserver/api/mockwebserver3.api
```diff
@@ -111,14 +112,12 @@ public final class mockwebserver3/MockWebServer : java/io/Closeable {
 	public final fun setProtocolNegotiationEnabled (Z)V
 	public final fun setProtocols (Ljava/util/List;)V
 	public final fun setServerSocketFactory (Ljavax/net/ServerSocketFactory;)V
-	public final fun setStarted (Z)V
```
swankjesse, line 114: This was bananas

mockwebserver/src/main/kotlin/mockwebserver3/MockWebServer.kt
```diff
@@ -337,26 +317,46 @@ public class MockWebServer : Closeable {
...
   @Throws(IOException::class)
-  private fun start(inetSocketAddress: InetSocketAddress) {
+  private fun start(socketAddress: InetSocketAddress) {
     check(!closed) { "shutdown() already called" }
-    if (started) return
-    started = true
 
-    this._inetSocketAddress = inetSocketAddress
+    val alreadyStartedAddress = socketAddress_
+    if (alreadyStartedAddress != null) {
```
swankjesse, line 328: Start used to silently ignore calls after it was started, even if the target address had changed

yschimke:
LGTM but GitHub erroring on approve

PR #8862, opened 2025-06-18
New @StartStop extension for JUnit 5

swankjesse:
This is a new take on our JUnit 5 extension. Rather than creating and managing the MockWebServer instance in the extension, we let the user create it and all the extension does is manage the lifecycle.

Note that this annotation doesn't require any external configuration - it doesn't need a system-property to opt-into 'automatic' extensions, and it doesn't require a class-level extension either.

### DIFF
--- a/mockwebserver-junit5/src/main/kotlin/mockwebserver3/junit5/StartStop.kt
+++ b/mockwebserver-junit5/src/main/kotlin/mockwebserver3/junit5/StartStop.kt
@@ -1,0 +1,42 @@
+/*
+ * Copyright (C) 2025 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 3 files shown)


mockwebserver-junit5/src/main/kotlin/mockwebserver3/junit5/StartStop.kt
```diff
@@ -0,0 +1,42 @@
...
+ * @StartStop public final MockWebServer server = new MockWebServer();
+ * ```
+ *
+ * Or for Kotlin:
+ *
+ * ```kotlin
+ * @StartStop val server = MockWebServer()
+ * ```
+ */
+@Target(AnnotationTarget.FIELD)
```
yschimke, line 38: Should other form be supported? `AnnotationTarget.PROPERTY`, `AnnotationTarget.PROPERTY_GETTER`?

Would require changing how we get the field.

yschimke: I'm inclined to leave as is, but just asking the question.

swankjesse: I think field does what we want for Java reflection.

PR #8863, opened 2025-06-18
Switch to @StartStop for JUnit 5 tests

swankjesse:


mockwebserver/src/test/java/mockwebserver3/MockWebServerTest.kt
```diff
@@ -71,11 +71,10 @@ class MockWebServerTest {
   @RegisterExtension
   var platform = PlatformRule()
 
-  private lateinit var server: MockWebServer
+  private val server = MockWebServer()
```
yschimke, line 74: This pattern now feels unsafe, since `server.stop()` must be called.

But existing problem.

yschimke (approved):

PR #8867, opened 2025-06-19
Stop using experimental coroutines APIs

swankjesse:
They're stable now.

### DIFF
--- a/okhttp-coroutines/src/main/kotlin/okhttp3/coroutines/ExecuteAsync.kt
+++ b/okhttp-coroutines/src/main/kotlin/okhttp3/coroutines/ExecuteAsync.kt
@@ -18,7 +18,6 @@
 package okhttp3.coroutines
 
 import kotlin.coroutines.resumeWithException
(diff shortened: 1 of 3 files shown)


okhttp-coroutines/src/main/kotlin/okhttp3/coroutines/ExecuteAsync.kt
```diff
@@ -45,8 +43,8 @@ suspend fun Call.executeAsync(): Response =
           call: Call,
           response: Response,
         ) {
-          continuation.resume(response) {
-            response.closeQuietly()
+          continuation.resume(response) { _, value, _ ->
+            value.closeQuietly()
```
swankjesse, line 47: This one’s not deprecated

https://github.com/Kotlin/kotlinx.coroutines/blob/f4f519b36734238ec686dfaec1e174086691781e/kotlinx-coroutines-core/common/src/CancellableContinuation.kt#L304

yschimke (approved):

PR #8870, opened 2025-06-19
 Replace SocketPolicy with a new type, SocketEffect

swankjesse:
SocketPolicy encapsulated two things:
 - a trigger (request start, response body, etc)
 - an effect (close the socket, close the stream, stall, etc.)

It also had some special cases for effects like failing the TLS handshake.

With this PR there's a new class, SocketEffect, that determines what to do. It's assigned to one of 5 fields for different triggers.


mockwebserver/src/test/java/mockwebserver3/MockWebServerTest.kt
```diff
@@ -298,7 +295,7 @@ class MockWebServerTest {
     server.enqueue(
       MockResponse
         .Builder()
-        .socketPolicy(DisconnectAtStart)
+        .onRequestStart(CloseSocket())
```
squarejesse, line 298: I think this kind of diff shows the main benefit of this PR. Previously we had a sparse support for available triggers and effects. With this change we now have full coverage.

okhttp/src/jvmTest/kotlin/okhttp3/URLConnectionTest.kt
```diff
@@ -369,17 +366,43 @@ class URLConnectionTest {
...
+        .Builder()
+        .body("This connection won't pool properly")
+        .onResponseEnd(ShutdownConnection)
+        .build(),
+    )
   }
 
   @Test
   fun serverShutdownInput() {
-    testServerClosesOutput(ShutdownInputAtEnd)
```
squarejesse, line 377: Previously we needed this constant in the public API for just a couple uses.

okhttp/src/jvmTest/kotlin/okhttp3/internal/ws/WebSocketHttpTest.kt
```diff
@@ -769,7 +768,6 @@ class WebSocketHttpTest {
           upgradeResponse(request)
             .body(Buffer().write("81".decodeHex())) // Truncated frame.
             .removeHeader("Content-Length")
-            .socketPolicy(KeepOpen)
```
squarejesse, line 772: (This was the old default behavior)

mockwebserver-deprecated/src/main/kotlin/okhttp3/mockwebserver/DeprecationBridge.kt
```diff
@@ -66,19 +56,38 @@ internal fun MockResponse.wrap(): mockwebserver3.MockResponse {
...
+      result.onResponseEnd(
+        CloseSocket(
+          closeSocket = false,
+          shutdownOutput = true,
+        ),
+      )
+    SocketPolicy.STALL_SOCKET_AT_START -> result.onRequestStart(SocketEffect.Stall)
+    SocketPolicy.NO_RESPONSE -> result.onResponseStart(SocketEffect.Stall)
+    SocketPolicy.RESET_STREAM_AT_START -> result.onRequestStart(CloseStream(http2ErrorCode))
+  }
```
squarejesse, line 89: The above code is a mapping from the old `SocketPolicy` to the new `SocketEffect`

PR #8872, opened 2025-06-20
Change RecordedRequest.chunkSizes to be nullable

swankjesse:
Now that it's all Kotlin we can usefully distinguish between null (not chunked) and empty list (chunked with zero bytes).

Also remove the experimental annotation from RecordedRequest and delete the experimental annotation.


okhttp/src/commonJvmAndroid/kotlin/okhttp3/ExperimentalOkHttpApi.kt
yschimke, line 1: HUGE!!!

yschimke (approved):

Issue #6212, opened 2020-08-05
MockWebServer: Clear requests before each test


GabrielBB:
Can we have a method to clear requests? Something like this:

```kotlin
    @BeforeEach
    fun cleanRequests() {
        mockServer.clearRequests();
    }
```

My tests are taking requests from previous tests. I wouldn't want to  shutdown entirely

JakeWharton:
Are you using the test rule? There should be a new instance for every test.

GabrielBB:
Yes, I'm using `@TestInstance(TestInstance.Lifecycle.PER_CLASS)`

JakeWharton:
Can you switch to PER_METHOD? Assuming such a thing exists.

swankjesse:
Yeah, MockWebServer instances are not expected to be reused.

GabrielBB:
Will a clear method be enough to be reusable ? I can open PR, why not I guess ?

swankjesse:
No, I don't think it's a good fit for a PR. The clear method encourages a pattern we don't support well.

nightswimmings:
What if we need  an static MockWebServer in order to register @DynamicPropertySource properties with the mockWebServer.getPort()

yschimke:
Can't you change your logic to update the DynamicPropertyRegistry as each test runs with the new port/url?

nightswimmings:
DynamicPropertyRegistry method must be static as per documentation, so I guess it does not reevaluate within the class. I think that would imply restarting the whole spring context since the mechanism is similar to that of @SpringBootTest(properties=X). So we'd need the mockrestserver instance to be static as well, I am not sure it makes sense to reinstantiate it globally and start/shutwoning on each test, I find the clearRequests() approach 
more comfy for that scenario

yschimke:
The method is static, but it provides a Supplier that gets called on demand.  I'm not saying it's ideal for you, just that it doesn't appear that you are blocked as you could update the port each time a new MockWebServer is started for your test.

nightswimmings:
I don't understand the Supplier part but I am interested.. :)

I have this currently, how would you exactly do it?

```
static MockWebServer mockWebServer;

    @DynamicPropertySource
    static void properties(DynamicPropertyRegistry propsRegistry) {
         propsRegistry.add("spring.cloud.gateway.routes[0].uri", () -> "http://localhost:" + mockWebServer.getPort())
    }
    
    @BeforeAll
    static void beforeAll() throws IOException {
[...]

PR #8880, opened 2025-06-24
Add a reset method to MockWebServer

mmollaverdi:
This is useful for test suites which start the `MockWebServer` once at the beginning, shut it down at the end, and reuse it across tests. This can be called in a `BeforeEach` or `AfterEach` method  for the request queue and count to be reset in between tests.


yschimke:
Prior context on https://github.com/square/okhttp/issues/6212

And https://github.com/square/okhttp/pull/3390

So we likely won't accept the PR

PR #8897, opened 2025-07-02
feat: add curl() method to Request for generating cURL commands

hamza-badar:
This commit introduces a new `toCurl()` method in the `Request` class that generates a cURL command equivalent for the HTTP request. This is useful for debugging, logging, and reproducing requests outside of the application.

Key features:
- Includes HTTP method (`-X`), headers (`-H`), and request body (`--data`) if present.
- Handles escaping of special characters in body content.
- Appends the request URL.
- Provides KDoc consistent with the existing codebase.
- Added unit tests for:
  - GET requests with headers.
  - POST requests with complex JSON bodies containing nested objects and arrays.

Example generated cURL:

`curl -X POST \
  -H "Content-Type: application/json" \
  -H "Authorization: Bearer xyz789" \
  --data "{\"user\":{\"id\":123,\"name\":\"John Doe\"},\"roles\":[\"admin\",\"editor\"],\"active\":true}" \
  "https://api.example.com/users" `


okhttp/src/commonJvmAndroid/kotlin/okhttp3/Request.kt
```diff
@@ -423,4 +423,40 @@ class Request internal constructor(
...
+    
+    for ((name, value) in headers) {
+      curl.append(" -H ")
+        .append("\"").append(name).append(": ").append(value).append("\"")
+    }
+    
+    body?.let { requestBody ->
+      val buffer = okio.Buffer()
+      requestBody.writeTo(buffer)
+      val bodyString = buffer.readUtf8()
```
JakeWharton: This doesn't handle binary payloads.

okhttp/src/jvmTest/kotlin/okhttp3/RequestTest.kt
```diff
@@ -643,6 +643,65 @@ class RequestTest {
...
   }
 
+   @Test
+  fun curlGet() {
+    val request = Request.Builder()
+      .url("https://example.com")
+      .header("Authorization", "Bearer abc123")
+      .build()
+    val curl = request.curl()
+    assert(curl.contains("curl"))
```
JakeWharton: Please use the assertion library as other tests

okhttp/src/jvmTest/kotlin/okhttp3/RequestTest.kt
```diff
@@ -643,6 +643,65 @@ class RequestTest {
     }
   }
 
+   @Test
```
JakeWharton: Indent incorrect

okhttp/src/jvmTest/kotlin/okhttp3/RequestTest.kt
JakeWharton: Also seems like you should just be testing the entire string

yschimke:
Some prior art at https://github.com/mrmike/Ok2Curl

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Request.kt
hamza-badar: Thanks @JakeWharton, I have handled the binary payload and now this implementation avoids writing request bodies to files, and emits a placeholder for binary content to ensure cross-platform compatibility and side-effect free operation, aligning with OkHttp’s design principles

okhttp/src/jvmTest/kotlin/okhttp3/RequestTest.kt
hamza-badar: Updated, Thanks!

okhttp/src/jvmTest/kotlin/okhttp3/RequestTest.kt
hamza-badar: Updated, Thanks!

hamza-badar:
> Some prior art at https://github.com/mrmike/Ok2Curl

Thanks @yschimke  for pointing that out! I’m aware of Ok2Curl — it’s a great tool and definitely prior art for this feature. However, the goal of this curl() implementation is to:
	•	Avoid introducing additional dependencies into OkHttp core.
	•	Provide a lightweight, self-contained utility that aligns with OkHttp’s philosophy of minimalism and portability.
	•	Avoid side effects (no logging, no interceptors, no I/O).

Ok2Curl is a fantastic external option for projects needing richer cURL logging features, but this curl() method provides a simple, dependency-free API directly on Request for users who want quick cURL generation without extra setup.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Request.kt
```diff
@@ -423,4 +423,59 @@ class Request internal constructor(
...
+        contentType.contains("form-urlencoded")
+      ) {
+        val bodyString = buffer.readUtf8()
+        curl
+          .append(" --data ")
+          .append("\"")
+          .append(bodyString.replace("\"", "\\\""))
+          .append("\"")
+      } else {
+        curl.append(" --data-binary \"[binary body omitted]\"")
```
JakeWharton: I would think that we would want to hex-encode binary data rather than drop it. Plus we need to use an algorithm to detect binary data rather than base it on the content type.

okhttp/src/jvmTest/kotlin/okhttp3/RequestTest.kt
```diff
@@ -643,6 +643,97 @@ class RequestTest {
...
+        .url("https://api.example.com/upload")
+        .post(body)
+        .addHeader("Content-Type", "application/octet-stream")
+        .build()
+
+    val curl = request.curl()
+    assertThat(
+      curl,
+    ).isEqualTo(
+      "curl -X POST -H \"Content-Type: application/octet-stream\" --data-binary \"[binary body omitted]\" \"https://api.example.com/upload\"",
```
yschimke: If the goal is that someone could run these, would it be worthwhile letting this be customised via optional param?

```
--data-binary @filename
--data-binary @-
```

JakeWharton: I think I'm already somewhat opposed to this feature in general, but definitely opposed to any options on it. If you can't do a simple `toCurl()` and get a thing that works all the time then we shouldn't do it.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Request.kt
```diff
@@ -423,4 +423,59 @@ class Request internal constructor(
...
+   * This includes the HTTP method, headers, request body (if present), and URL.
+   *
+   * Example:
+   * ```
+   * curl -X POST -H "Authorization: Bearer token" --data "{\"key\":\"value\"}" "https://example.com/api"
+   * ```
+   *
+   * @return a cURL command string representing this request.
+   */
+  fun curl(): String {
```
JakeWharton: This should probably be named `toCurl`, since it's not a property of the request itself.

okhttp/src/jvmTest/kotlin/okhttp3/RequestTest.kt
yschimke: @- would work as a baked in default. Better than a string you need to edit

yschimke:
This feels exactly like the type of thing that should be in a library. There is no need for it to be built into OkHttp.

I suggest submitting a PR to the Ok2Curl project to work on Request also.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Request.kt
```diff
@@ -423,4 +423,59 @@ class Request internal constructor(
...
+    val curl = StringBuilder("curl")
```
MukjepScarlet, line 459: It might be better to use `Buffer` here, because we would append request body

swankjesse (commented):
I like this idea because curl arguments are kind of a universal format for an HTTP request. I can do ‘copy as curl’ in Chrome, or IntelliJ’s HTTP tool, or whatever.

I request some changes:

1. Bound the command size, probably to something that’s a typical limit for a Linux command line. If the command is too large, we gotta truncate it somehow? (What does Chrome do?)
2. Support binary responses, and differentiate text vs. binary by looking at the response body data (instead of the content-type)

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Request.kt
hamza-badar: Updated, can you please have look once more time @JakeWharton.

okhttp/src/jvmTest/kotlin/okhttp3/RequestTest.kt
hamza-badar: Updated, can you please have a look at it @yschimke.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Request.kt
hamza-badar: Updated, Thanks.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Request.kt
hamza-badar: Updated, Thanks.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Request.kt
```diff
@@ -430,4 +432,100 @@ class Request internal constructor(
...
+   * Detects binary data by checking for non-printable characters in a buffer.
+   */
+  private fun isBinaryData(buffer: Buffer): Boolean {
+    val peekBuffer = buffer.clone()
+    var totalBytes = 0
+    var binaryCount = 0
+
+    while (!peekBuffer.exhausted() && totalBytes < 4096) { // limit to first 4KB for performance
+      val b = peekBuffer.readByte().toInt() and 0xFF
+      if ((b < 0x20 && b !in listOf(0x09, 0x0A, 0x0D)) || b > 0x7E) {
```
MukjepScarlet: `List<Integer>` is not good...
`intArrayOf` or `byteArrayOf` should be here, or directly `b != 0x09 || b != 0x0A || b != 0x0D`

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Request.kt
```diff
@@ -430,4 +432,100 @@ class Request internal constructor(
...
+    }
+
+    // Append body if present
+    body?.let { requestBody ->
+      val buffer = Buffer()
+      requestBody.writeTo(buffer)
+
+      // Clone so we can read multiple times without consuming
+      val peekBuffer = buffer.clone()
+      val isBinary = isBinaryData(peekBuffer)
```
MukjepScarlet, line 483: it's cloned twice

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Request.kt
```diff
@@ -430,4 +432,100 @@ class Request internal constructor(
...
+        .append(name)
+        .append(": ")
+        .append(value)
+        .append("\"")
+    }
+
+    // Append body if present
+    body?.let { requestBody ->
+      val buffer = Buffer()
+      requestBody.writeTo(buffer)
```
MukjepScarlet, line 479: This action has side-effect on the `RequestBody`, please point it out on the function doc

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Request.kt
```diff
@@ -430,4 +432,100 @@ class Request internal constructor(
...
+   * This includes the HTTP method, headers, request body (if present), and URL.
+   *
+   * Example:
+   * ```
+   * curl -X POST -H "Authorization: Bearer token" --data "{\"key\":\"value\"}" "https://example.com/api"
+   * ```
+   *
+   * @return a cURL command string representing this request.
+   */
+  fun toCurl(
```
MukjepScarlet, line 455: `@JvmOverloads`

okhttp/api/jvm/okhttp.api
```diff
@@ -1008,6 +1008,7 @@ public final class okhttp3/Request {
 	public final fun body ()Lokhttp3/RequestBody;
 	public final fun cacheControl ()Lokhttp3/CacheControl;
 	public final fun cacheUrlOverride ()Lokhttp3/HttpUrl;
+	public final fun toCurl ()Ljava/lang/String;
```
MukjepScarlet: Different with actual code?

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Request.kt
```diff
@@ -430,4 +432,100 @@ class Request internal constructor(
...
+            curl.append(" --data-binary \"")
+            val hexBuffer = buffer.clone()
+            while (!hexBuffer.exhausted()) {
+              val b = hexBuffer.readByte().toInt() and 0xFF
+              curl.append("%02x".format(b))
+            }
+            curl.append("\"")
+          }
+          BinaryMode.FILE -> {
+            val fileName = binaryFileName ?: "request_body.bin"
```
MukjepScarlet: Default param?

okhttp/src/jvmTest/kotlin/okhttp3/RequestTest.kt
```diff
@@ -643,6 +643,169 @@ class RequestTest {
...
+        .url("https://api.example.com/users")
+        .post(body)
+        .addHeader("Content-Type", "application/json")
+        .addHeader("Authorization", "Bearer xyz789")
+        .build()
+
+    val curl = request.toCurl()
+    assertThat(curl)
+      .isEqualTo(
+        "curl -X POST -H \"Content-Type: application/json\" -H \"Authorization: Bearer xyz789\" --data \"{\n" +
```
MukjepScarlet, line 710: Use multi line string here

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Request.kt
hamza-badar: Updated, Thanks.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Request.kt
hamza-badar: Updated, Thanks.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Request.kt
hamza-badar: Updated, Thanks.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Request.kt
hamza-badar: Updated, Thanks.

okhttp/api/jvm/okhttp.api
hamza-badar: Updated, Thanks.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Request.kt
hamza-badar: Updated, Thanks.

okhttp/src/jvmTest/kotlin/okhttp3/RequestTest.kt
hamza-badar: I don't think it's much of a concern.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Request.kt
```diff
@@ -430,4 +432,106 @@ class Request internal constructor(
...
+    }
+
+    curl.append(" \"").append(url).append("\"")
+    return curl.toString()
+  }
+
+  /**
+   * Detects binary data by checking for non-printable characters in a buffer.
+   */
+  private fun isBinaryData(peekBuffer: Buffer): Boolean {
```
swankjesse, line 522: This says `isBinaryData()` but it’s actually checking if it’s non-ASCII.

For example, it returns true for this string of regular UTF-8 characters:

```
    val binaryData = isBinaryData(Buffer().writeUtf8("Слава Україні!"))
```

Please switch to use our existing function, `isProbablyUtf8()`. Also please a parameter to that function to change its scope from 16 codepoints / 64 bytes to 4096 codepoints / 16384 bytes ?

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Request.kt
```diff
@@ -430,4 +432,106 @@ class Request internal constructor(
...
+
+      // Clone so we can read multiple times without consuming
+      val peekBuffer = buffer.clone()
+      val isBinary = isBinaryData(peekBuffer)
+
+      if (isBinary) {
+        when (binaryMode) {
+          BinaryMode.HEX -> {
+            curl.append(" --data-binary \"")
+            val hexBuffer = buffer.clone()
```
swankjesse, line 489: We’ve got a function for this, `buffer.readByteString().hex()`

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Request.kt
```diff
@@ -430,4 +432,106 @@ class Request internal constructor(
...
+          }
+          BinaryMode.STDIN -> {
+            curl.append(" --data-binary @-")
+          }
+          BinaryMode.OMIT -> {
+            curl.append(" --data-binary \"[binary body omitted]\"")
+          }
+        }
+      } else {
+        val bodyString = buffer.readString(StandardCharsets.UTF_8)
```
swankjesse, line 507: `buffer.readUtf8()`

swankjesse (approved):

swankjesse:
This is great! I hope you don’t mind I’m going to merge now and address some PR feedback in my own PR!

swankjesse:
Follow-ups: https://github.com/square/okhttp/pull/9112

Issue #8826, opened 2025-05-30
v5.0.0-alpha16 - No virtual method getEventListenerFactory


AlexanderEggers:
After updating from **v5.0.0-alpha14** to **v5.0.0-alpha16**, the **release** build of my android app is crashing with the following error:

> NoSuchMethodError: No virtual method getEventListenerFactory$okhttp()Lokhttp3/EventListener$Factory; in class LJf/K; or its super classes (declaration of 'Jf.K' appears in /data/app/~~F8MgBq3JZu9FbBw49qQ8DQ==/com.myapp.release-N05Wr0CV4hKEqlty98x18g==/base.apk)

This works fine using debug builds, so likely an R8 issue if I have to guess.

WebTiger89:
Using `com.squareup.okhttp3:okhttp-jvm` instead of `com.squareup.okhttp3:okhttp` solved it for me, see [https://github.com/square/okhttp/blob/master/CHANGELOG.md#version-500-alpha15](https://github.com/square/okhttp/blob/master/CHANGELOG.md#version-500-alpha15)

yschimke:
That should be handled automatically for an android Gradle build or I assume maven.

Also if you need this workaround, make sure to use the android artifact on android builds

AlexanderEggers:
> Using `com.squareup.okhttp3:okhttp-jvm` instead of `com.squareup.okhttp3:okhttp` solved it for me, see https://github.com/square/okhttp/blob/master/CHANGELOG.md#version-500-alpha15

I don't believe this is a workaround I want to consider given that I really want to use the android artifact, not just the jvm.

yschimke:
I don't have an interesting repro, mainly because I can't get it failing with a debug or release build. or with the OkHttp android test app.

```
    @Singleton
    @Provides
    fun okhttpClient(
        cache: Cache,
        alwaysHttpsInterceptor: Interceptor,
    ): OkHttpClient {
        return OkHttpClient.Builder().followSslRedirects(false)
            .addInterceptor(alwaysHttpsInterceptor)
            .eventListenerFactory(LoggingEventListener.Factory())
            .cache(cache)
            .build()
            .apply { println("Using okhttp client " + this.eventListenerFactory) }
    }
```

```
Using okhttp client A3.e@b4f5a98
```

Can anyone provide a repro github project?

AlexanderEggers:
> I don't have an interesting repro, mainly because I can't get it failing with a debug or release build. or with the OkHttp android test app.
[...]

PR #8898, opened 2025-07-03
Fix a NoSuchMethodError loading OkHttp on Android

squarejesse:
The Sentry SDK assumes things in our public API that it shouldn't be assuming. Oh well.

Closes: https://github.com/square/okhttp/issues/8826


swankjesse:
I’ve got a thread with them and I’m asking for some extra validation in their code rewriter.

yschimke (approved):

PR #8899, opened 2025-07-03
Fix breaking change from alpha.17 release

staktrace:
The semantics of the old path property included the query string, and should be mapped from the new `target` which is the equivalent field in the new recorded request. Using url.encodedPath only provides the path component and omits the query component, resulting in a breaking change.


staktrace:
The test change in this PR fails without the fix, thereby capturing the regression.

mockwebserver-deprecated/src/main/kotlin/okhttp3/mockwebserver/DeprecationBridge.kt
```diff
@@ -112,7 +112,7 @@ internal fun mockwebserver3.RecordedRequest.unwrap(): RecordedRequest =
     sequenceNumber = exchangeIndex,
     failure = failure,
     method = method,
-    path = url.encodedPath,
+    path = target,
```
swankjesse, line 115: thank you!

swankjesse (approved):

PR #8900, opened 2025-07-03
Remove disclaimer about GraalVM

swankjesse:


yschimke:
@swankjesse I'll take a look at the build failures.

yschimke:
@swankjesse There are some junit failures I need to sort out. But nativeImage building is working and the okcurl binary runs fine.

yschimke:
Should be working, but with some caveats.

- StartStop doesn't work currently, have to manually start
- There is extra config for Junit until native build tools 0.11.0

None of these are problems with the 5.0 release

yschimke:
@swankjesse let's do it?

Issue #8902, opened 2025-07-03
Crash reading trailers in Retrofit Call


swankjesse:
We’ve observed this crash with Retrofit 3 and OkHttp 5.0.0:

```
java.lang.IllegalStateException: Cannot read raw response body of a converted body.
	at retrofit2.OkHttpCall$NoContentResponseBody.source(OkHttpCall.java:300)
	at okhttp3.Response.trailers(Response.kt:195)
	at misk.client.ClientMetricsInterceptorNoSummaryTest.responseCodes(ClientMetricsInterceptorNoSummaryTest.kt:107)
	at java.lang.reflect.Method.invoke(Method.java:568)
```

I believe the core problem is this recently-changed function in OkHttp 5:

```kotlin
  @Throws(IOException::class)
  fun trailers(): Headers {
    val source = body.source()
    if (source.isOpen) {
      source.skipAll()
    }
    return trailersSource.get()
  }
```

I think it’s fine that we exhaust the response body before attempting to read the trailers. But I also believe that behavior should be within the implementation of `trailersSource` and not in the code that calls that function.

Ie. we should change all of our built-in `TrailersSource.get` implementations to do this:

```kotlin
  override fun get(): Headers {
    val source = bodySource // Somehow!
    if (source.isOpen) {
      source.skipAll()
    }
  }
```

That way a custom `ResponseBody` won’t interact awkwardly with `Response`.

An alternative solution would be a custom function on `ResponseBody` to do this work. That way the custom Retrofit `ResponseBody` could override that function.

PR #8904, opened 2025-07-03
Don't force a response body read on all trailers

swankjesse:
We only need to read the response body if its coming from a real server.

Closes: https://github.com/square/okhttp/issues/8902


okhttp/src/jvmTest/kotlin/okhttp3/TrailersTest.kt
```diff
@@ -419,6 +422,70 @@ open class TrailersTest {
...
+        .protocol(Protocol.HTTP_1_1)
+        .code(200)
+        .message("OK")
+        .body(
+          object : ResponseBody() {
+            override fun contentType(): MediaType? = null
+
+            override fun contentLength(): Long = -1L
+
+            override fun source(): BufferedSource = error("unexpected call")
```
yschimke, line 481: Nice test

yschimke (approved):

Issue #8901, opened 2025-07-03
Fix the graal build failures


yschimke:
@swankjesse I'll take a look at the build failures.

_Originally posted by @yschimke in https://github.com/square/okhttp/issues/8900#issuecomment-3032278919_

yschimke:
OK, this is running junit tests. But the okcurl build step already passed.

```
Produced artifacts:
 /home/runner/work/okhttp/okhttp/okcurl/build/native/nativeCompile/okcurl (executable)
 /tmp/native-image-build-output.json (build_info)
========================================================================================================================
Finished generating 'okcurl' in 2m 35s.
[native-image-plugin] Native Image written to: /home/runner/work/okhttp/okhttp/okcurl/build/native/nativeCompile
```

yschimke:
And running, even on windows

```
C:\Users\yuri\workspace\okhttp\okcurl\build\native\nativeCompile\okcurl.exe https://publicobject.com/robots.txt
User-agent: *
Disallow: /glazedlists/wiki/
```

yschimke:
Seems related to errors I'm getting https://github.com/oracle/graal/issues/6957

yschimke:
Actually https://github.com/junit-team/junit-framework/wiki/Upgrading-to-JUnit-5.13

yschimke:
StartStop is failing, so removed for now.

PR #8906, opened 2025-07-04
Fix Graal Tests

yschimke:
Addressing https://github.com/square/okhttp/issues/8901

Two current fixes/workarounds

- Apply https://github.com/junit-team/junit5/wiki/Upgrading-to-JUnit-5.13
- Reduce back to 2 tests to get something passing.  Doesn't include tests from other modules.

We could use a snapshot of 0.11.0 from https://github.com/graalvm/native-build-tools/blob/ecbe2327fc3b093cdf50fbf49ccdb33e92a80502/docs/src/docs/asciidoc/changelog.adoc#release-094


yschimke:

There is still follow up to enable more and fix StartStop.

swankjesse (commented):
LGTM

Issue #8910, opened 2025-07-04
Update mockwebserver docs


yschimke:
https://github.com/square/okhttp/tree/master/mockwebserver shows the old mutable APIs

yschimke:
Also currently linking to old artifacts

PR #8911, opened 2025-07-04
Link to new mockwebserver artifacts

yschimke:
one fix from https://github.com/square/okhttp/issues/8910


yschimke:
https://repo1.maven.org/maven2/com/squareup/okhttp3/mockwebserver3/5.0.0/
vs
https://repo1.maven.org/maven2/com/squareup/okhttp3/mockwebserver/5.0.0/

swankjesse (approved):

swankjesse:
Nice catch.

yschimke:
I think docs are still referring to old idioms

PR #8917, opened 2025-07-05
Add a test of maven dependencies

yschimke:


yschimke:
@swankjesse not sure we need this long term, but at least shows us what we are discussing.

.github/workflows/maven.yml
```diff
@@ -0,0 +1,37 @@
...
+
+env:
+  GRADLE_OPTS: "-Dorg.gradle.jvmargs=-Xmx4g -Dorg.gradle.daemon=false -Dkotlin.incremental=false"
+
+jobs:
+  test_maven:
+    permissions:
+      checks: write # for actions/upload-artifact
+    runs-on: ubuntu-latest
+    if: github.ref == 'refs/heads/master' || contains(github.event.pull_request.labels.*.name, 'maven')
```
JakeWharton: Would prefer this to always run

.github/workflows/maven.yml
```diff
@@ -0,0 +1,37 @@
...
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
```
JakeWharton: This is actually slower than the default: https://github.com/actions/checkout?tab=readme-ov-file#fetch-all-history-for-all-tags-and-branches

maven-tests/.gitignore
```diff
@@ -0,0 +1,2 @@
+build
```
JakeWharton: Should already be covered by the root ignore

maven-tests/pom.xml
```diff
@@ -0,0 +1,98 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>com.squareup.okhttp3</groupId>
+  <artifactId>maven-tests</artifactId>
+  <version>5.0.0-SNAPSHOT</version>
```
JakeWharton: Probably better as 1.0.0-SNAPSHOT or something as nothing will ever update this version.

maven-tests/pom.xml
```diff
@@ -0,0 +1,98 @@
...
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <maven.compiler.source>17</maven.compiler.source>
+    <maven.compiler.target>17</maven.compiler.target>
+  </properties>
+
+  <dependencies>
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <version>4.12</version>
```
JakeWharton:
```suggestion
      <version>4.13.2</version>
```

maven-tests/pom.xml
```diff
@@ -0,0 +1,98 @@
...
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>com.squareup.okhttp3</groupId>
+      <artifactId>okhttp-jvm</artifactId>
+      <version>5.0.0</version>
```
JakeWharton: This is really testing anything useful since its detection would only occur after release when Renovate bumps this version.

We need to trigger the Gradle build to install to maven local and then use a range snapshot version here to be consuming the actual in-repo version of OkHttp.

maven-tests/src/main/java/com/squareup/okhttp3/maventest/SampleHttpClient.java
```diff
@@ -0,0 +1,39 @@
...
+
+public class SampleHttpClient {
+  private final OkHttpClient client;
+
+  public SampleHttpClient() {
+    client = new OkHttpClient.Builder().build();
+  }
+
+  public void makeCall() throws IOException {
+    try (Response response = client.newCall(new Request(HttpUrl.get("https://google.com/robots.txt"), Headers.EMPTY, "GET", null)).execute()) {
```
JakeWharton: Can we inject the URL in the ctor so the test can use MockWebServer?

maven-tests/src/site/site.xml
JakeWharton, line 1: Can be deleted

settings.gradle.kts
```diff
@@ -52,6 +53,10 @@ include(":samples:tlssurvey")
 include(":samples:unixdomainsockets")
 include(":container-tests")
 
+if (mavenBuild.toBoolean()) {
+  include(":maven-tests")
```
JakeWharton: I'm not sure why we want to build these sources from Gradle and Maven. As I said above, would prefer to always build with Maven which should eliminate the need to ever build with Gradle.

.github/workflows/maven.yml
```diff
@@ -0,0 +1,37 @@
...
+    types: [opened, labeled, unlabeled, synchronize]
+
+permissions:
+  contents: read
+
+env:
+  GRADLE_OPTS: "-Dorg.gradle.jvmargs=-Xmx4g -Dorg.gradle.daemon=false -Dkotlin.incremental=false"
+
+jobs:
+  test_maven:
```
swankjesse: Would you consider adding this to the existing `build.yml` ? I don’t think Maven is different-enough from the other checks in that file, and it’s fewer YML files to keep up-to-date for things like `on` and `env`

maven-tests/build.gradle.kts
```diff
@@ -0,0 +1,23 @@
+plugins {
```
swankjesse: What’s this file for?

maven-tests/src/main/java/com/squareup/okhttp3/maventest/SampleHttpClient.java
swankjesse: Or even a test that uses MockWebServer normally? It can be a regular MockWebServer test

settings.gradle.kts
swankjesse: My preference is for CI to always build with Maven, and locally I don’t have to think about it?

.github/workflows/maven.yml
yschimke: https://github.com/square/okhttp/pull/8929

yschimke:

swankjesse (commented):
LGTM

swankjesse:
Thanks for setting this up.

Issue #8918, opened 2025-07-07
Calling `topPrivateDomain()` from a non-primary process on Android crashes


chimbori:
```
Failed to read public suffix list
java.io.IOException: Platform applicationContext not initialized
 at okhttp3.internal.publicsuffix.AssetPublicSuffixList.listSource(AssetPublicSuffixList.kt:28)
 at okhttp3.internal.publicsuffix.BasePublicSuffixList.readTheList(BasePublicSuffixList.kt:47)
 at okhttp3.internal.publicsuffix.BasePublicSuffixList.readTheListUninterruptibly(BasePublicSuffixList.kt:95)
 at okhttp3.internal.publicsuffix.BasePublicSuffixList.ensureLoaded(BasePublicSuffixList.kt:68)
 at okhttp3.internal.publicsuffix.PublicSuffixDatabase.findMatchingRule(PublicSuffixDatabase.kt:82)
 at okhttp3.internal.publicsuffix.PublicSuffixDatabase.getEffectiveTldPlusOne(PublicSuffixDatabase.kt:53)
 at okhttp3.HttpUrl.topPrivateDomain(HttpUrl.kt:780)
 at com.chimbori.hermitcrab.web.BrowserFragment.onViewCreated$lambda$47(BrowserFragment.kt:444)
 … TRIMMED

Process: com.chimbori.hermitcrab:incognito, PID: 12549
java.lang.IllegalStateException: Unable to load PublicSuffixDatabase.list resource.
 at okhttp3.internal.publicsuffix.BasePublicSuffixList.ensureLoaded(BasePublicSuffixList.kt:77)
 at okhttp3.internal.publicsuffix.PublicSuffixDatabase.findMatchingRule(PublicSuffixDatabase.kt:82)
 at okhttp3.internal.publicsuffix.PublicSuffixDatabase.getEffectiveTldPlusOne(PublicSuffixDatabase.kt:53)
 at okhttp3.HttpUrl.topPrivateDomain(HttpUrl.kt:780)
 at com.chimbori.hermitcrab.web.BrowserFragment.onViewCreated$lambda$47(BrowserFragment.kt:444)
 … TRIMMED
```

[...]

PR #8919, opened 2025-07-07
Make okhttp3.internal.platform.PlatformInitializer multiprocess

yschimke:
At least some evidence this is supported, see 

https://cs.android.com/android/platform/superproject/main/+/main:packages/apps/Car/SystemUI/tests/AndroidManifest.xml;l=58

Relates to https://github.com/square/okhttp/issues/8918

- [ ] Add a test confirming this works

### DIFF
--- a/okhttp/src/androidMain/AndroidManifest.xml
+++ b/okhttp/src/androidMain/AndroidManifest.xml
@@ -6,6 +6,7 @@
   <application>
     <provider
       android:name="androidx.startup.InitializationProvider"
+      android:multiprocess="true"
       android:authorities="${applicationId}.androidx-startup"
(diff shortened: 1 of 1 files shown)


manastungare:
2010-09-13 by Dianne Hackborn: https://groups.google.com/d/msg/android-developers/u9UMJtALSXw/FTt1xsmILhMJ (the group has been shut down for spam, but there are references to this thread from StackOverflow and other places).

> Don't use this, it is some old cruft from pre-1.0 design that doesn't work and should be ignored these days. Just pretend like the attribute doesn't exist. :}

Also, b/175725770 requesting for this attribute to be deprecated and removed.

Issue #8916, opened 2025-07-04
peekTrailers() API


swankjesse:
I was reviewing some callers upgrading from OkHttp 4 to OkHttp 5 and I was surprised to see this:

```kotlin
val okhttp3.Response.grpcStatus: GrpcStatus
  get() =
    runCatching { trailers() }.getOrNull()?.get("grpc-status")?.toIntOrNull()?.let { GrpcStatus.get(it) }
      ?: header("grpc-status")?.toIntOrNull()?.let { GrpcStatus.get(it) }
      ?: GrpcStatus.OK
```

The `runCatching()` here is calling `trailers()` and ignoring its `IllegalStateException` if trailers aren’t available.

The OkHttp 5 behaviour change is there – now instead of throwing it blocks. I assumed changing code from throwing to blocking was safe, and I was wrong!

Anyway, if you were okay catching exceptions we used to have an API with semantics like `getTrailersIfTheyAreReady()`. OkHttp 5 replaces that with a function with semantics like `awaitTrailers()`.

I’m tempted to add a new function that has semantics similar to what we had in OkHttp 4.x:

```kotlin
class Response {
  ...

  /**
   * Returns the trailers on this response if they're available
   * immediately. This will be empty if the trailers are available,
   * but have no data.
   *
   * This returns null if the trailers are not available yet.
   *
   * Typically this will be null for HTTP/1 calls until the response
   * is read up to the last chunk. It will be null for HTTP/2 calls
   * until the thread that services the network has processed the
   * inbound frame that includes the trailers.
   * 
   * @throws IOException if the response is unreadable
   */
  @Throws(IOException::class)
  fun peekTrailers(): Headers? {
    ...
  }
}
```

yschimke:
> Look, my setup works for me. Just add an option to reenable the trailers IllegalStateException. 

https://xkcd.com/1172/

PR #8921, opened 2025-07-07
New Response.peekTrailers() API

swankjesse:
This restores behavior we had in OkHttp 4.x that I broke when I changed Response.trailers() to block
until trailers were available.

Closes: https://github.com/square/okhttp/issues/8916

### DIFF
--- a/okhttp/src/commonJvmAndroid/kotlin/okhttp3/Response.kt
+++ b/okhttp/src/commonJvmAndroid/kotlin/okhttp3/Response.kt
@@ -193,6 +193,26 @@ class Response internal constructor(
   fun trailers(): Headers = trailersSource.get()
 
   /**
+   * Returns the trailers after the HTTP response, if they are available to read immediately. Unlike
+   * [trailers], this doesn't block if the trailers are not immediately available, and instead
+   * returns null.
+   *
+   * This will typically return null until [ResponseBody.source] has buffered the last byte of the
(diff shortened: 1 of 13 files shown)


okhttp/src/commonJvmAndroid/kotlin/okhttp3/Response.kt
```diff
@@ -192,6 +192,26 @@ class Response internal constructor(
...
+   * returns null.
+   *
+   * This will typically return null until [ResponseBody.source] has buffered the last byte of the
+   * response body. Call `body.source().request(1024 * 1024)` to block until either that's done, or
+   * 1 MiB of response data is loaded into memory. (You could use any size here, though large values
+   * risk exhausting memory.)
+   *
+   * This returns an empty value if the trailers are available, but have no data.
+   *
+   * It is not safe to call this concurrently with code that is processing the response body.
```
yschimke, line 207: Does this include reading the body in chunks? or is it fine to interleave?

I realise this isn't really useful as trailers terminate the stream. But maybe interesting if you think the body has been received, just not consumed yet.

yschimke (approved):

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Response.kt
swankjesse: As long as there’s only one thread it’s safe.

Issue #8920, opened 2025-07-07
Fix readme for mock web server v3


emartynov:
[README](https://github.com/square/okhttp/tree/master/mockwebserver) says:

```
// Shut down the server. Instances cannot be reused.
  server.shutdown();
```

while there is only method `close` available.

PR #8923, opened 2025-07-07
Update code samples for mockwebserver3

swankjesse:
Closes: https://github.com/square/okhttp/issues/8920


yschimke:
And https://github.com/square/okhttp/issues/8910 ?

Issue #8913, opened 2025-07-04
Empty jar for release 5.0.0?


eduramiba:
It seems the jar for the new release: https://repo1.maven.org/maven2/com/squareup/okhttp3/okhttp/5.0.0/okhttp-5.0.0.jar is empty 🤔 without any classes.

I see recent issues where people are using the 5.0.0 version so maybe I am doing something wrong?

Thanks!
Eduardo

yschimke:
It's this from https://github.com/square/okhttp/blob/master/CHANGELOG.md

> OkHttp is now packaged as separate JVM and Android artifacts. This allows us to offer platform-specific features and optimizations. If your build system handles [Gradle module metadata](https://docs.gradle.org/current/userguide/publishing_gradle_module_metadata.html), this change should be automatic.

okhttp will pull in either okhttp-jvm or okhttp-android automatically.

yschimke:
As a last resort you could depend on okhttp-jvm or okhttp-android directly. But it's more likely to lead to confusion of what you are including if your build system is not aware of the module metadata.

eduramiba:
Thanks for the clarification! I will depend on okhttp-jvm for maven

Andre601:
Encountered the same issue here.

The fact that OkHttp only mentions and displays Gradle examples doesn't help the people who are using OkHttp through Maven, which doesn't seem to have this module metadata awareness... Or at least not without any tweaks to make.

It should be considered to update your docs and changelogs to mention this small fact here a bit more clearly (i.e. by actually mentioning the dependencies you would need to include here.
Not everyone is using Gradle, so only focusing on that is kinda leaving people in the dark that don't use Maven.

swankjesse:
Can fix.

swankjesse:
We’ll assume Maven users are not targeting Android?

yufu775:
maven : Have others encountered this?

* Invalid
```
<dependency>
    <groupId>com.squareup.okhttp3</groupId>
    <artifactId>okhttp-jvm</artifactId>
    <version>5.0.0</version>
</dependency>

<dependency>
    <groupId>com.squareup.okhttp3</groupId>
    <artifactId>logging-interceptor</artifactId>
    <version>5.0.0</version>
</dependency>
```
* valid `5.0.0-alpha.16`

yschimke:
Verified okhttp-jvm instead of okhttp works in maven with https://github.com/square/okhttp/pull/8917

```
[...]

PR #8938, opened 2025-07-12
Update docs for maven workaround

yschimke:
Documenting workaround for https://github.com/square/okhttp/issues/8913


yschimke:

We can see confusion, so I'd like some docs in place that we can update as we do the constraints solution.

PR #8939, opened 2025-07-12
Run most builds on PRs

yschimke:
Seems that the filtered set is not a popular approach. Assuming we have mostly non flaky builds, run more things on all builds (main, branches, prs).

### DIFF
--- a/okcurl/src/test/kotlin/okhttp3/curl/OkcurlTest.kt
+++ b/okcurl/src/test/kotlin/okhttp3/curl/OkcurlTest.kt
@@ -16,9 +16,18 @@
 package okhttp3.curl
 
 import com.github.ajalt.clikt.core.main
+import kotlin.test.BeforeTest
 import kotlin.test.Test
+import okhttp3.TestUtil.assumeNotWindows
 
 class OkcurlTest {
+  @BeforeTest
+  fun skipWindows() {
+    // Failing with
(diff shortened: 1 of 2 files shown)


yschimke:

Disabled one test on windows that includes a System.exit in clikt library.

```
    inline fun <T : BaseCliktCommand<T>> main(command: T, parseAndRun: T.() -> Unit) {
        try {
            command.parseAndRun()
        } catch (e: CliktError) {
            command.echoFormattedHelp(e)
            command.currentContext.exitProcess(e.statusCode)
        }
    }
```

okcurl/src/test/kotlin/okhttp3/curl/OkcurlTest.kt
```diff
@@ -16,9 +16,18 @@
 package okhttp3.curl
 
 import com.github.ajalt.clikt.core.main
+import kotlin.test.BeforeTest
 import kotlin.test.Test
+import okhttp3.TestUtil.assumeNotWindows
 
 class OkcurlTest {
```
JakeWharton, line 23: This whole test seems redundant with `MainTest` which already parses, binds, and runs the `Main` command many times.

yschimke: https://github.com/square/okhttp/pull/8948

PR #8943, opened 2025-07-15
New zstd interceptor

swankjesse:


okhttp-zstd/src/main/kotlin/okhttp3/zstd/ZstdInterceptor.kt
```diff
@@ -0,0 +1,75 @@
...
+      when {
+        encoding.equals("zstd", ignoreCase = true) -> body.source().zstdDecompress()
+        encoding.equals("gzip", ignoreCase = true) -> body.source().gzip()
+        else -> return response
+      }
+
+    return response
+      .newBuilder()
+      .removeHeader("Content-Encoding")
+      .removeHeader("Content-Length")
```
JakeWharton, line 71: Calling buffer here seems wrong. You'd only get the bytes from the first read. Either it's steaming and needs to pass the source, or it's buffered and we know the length.

swankjesse: We don’t offer an API, `Source.asResponseBody()`, only `BufferedSource.asResponseBody()`. Perhaps an oversight?

The motivation is that if the API was `Source.asResponseBody()`, then the implementation itself would need to call `buffer()`, which would result in 2 rounds of buffering if it was already buffered.

Moshi has a similar API with `JsonReader.of(BufferedSource)` for the same reason.

swankjesse: (In either case the length is irrelevant, this buffer() function just converts a Source to a BufferedSource.)

JakeWharton: Ah, this is wrapping in a buffered source. This function is poorly named in hindsight since it could equally be accessing the buffer of a buffered source (noun) or performing the wrapping to add a buffer (verb). We probably should have chosen buffered as the extension form.

okhttp-zstd/README.md
```diff
@@ -0,0 +1,20 @@
+OkHttp ZStandard (zstd) Integration
```
yschimke: facebook/zstd used Zstandard

okhttp-zstd/src/main/kotlin/okhttp3/zstd/ZstdInterceptor.kt
```diff
@@ -0,0 +1,75 @@
...
+    val originalRequest = chain.request()
+
+    if (originalRequest.header("Accept-Encoding") != null) {
+      return chain.proceed(originalRequest)
+    }
+
+    val request =
+      originalRequest
+        .newBuilder()
+        .header("Accept-Encoding", "zstd,gzip")
```
yschimke, line 47: Choosing between zstd or brotli, makes me wonder whether we should support a configurable (default to gzip only) DecompressionInterceptor

okhttp-zstd/src/main/kotlin/okhttp3/zstd/ZstdInterceptor.kt
```diff
@@ -0,0 +1,75 @@
+/*
+ * Copyright (C) 2025 Square, Inc.
```
yschimke, line 2: Square? Not Block?

yschimke (approved):

okhttp-zstd/build.gradle.kts
```diff
@@ -0,0 +1,31 @@
...
+  "Automatic-Module-Name: okhttp3.zstd",
+  "Bundle-SymbolicName: com.squareup.okhttp3.zstd"
+)
+
+dependencies {
+  api(projects.okhttp)
+  implementation(libs.zstd.kmp.okio)
+
+  testImplementation(projects.okhttpTestingSupport)
+  testImplementation(projects.mockwebserver3Junit5)
```
JakeWharton:
```suggestion
```
Seems unused

okhttp-zstd/src/test/java/okhttp3/zstd/ZstdInterceptorTest.kt
```diff
@@ -0,0 +1,167 @@
...
+import okhttp3.ResponseBody.Companion.toResponseBody
+import okhttp3.zstd.ZstdInterceptor.decompress
+import okio.Buffer
+import okio.ByteString
+import okio.ByteString.Companion.EMPTY
+import okio.ByteString.Companion.encodeUtf8
+import okio.Sink
+import okio.buffer
+import okio.gzip
+import org.junit.jupiter.api.Test
```
JakeWharton, line 39:
```suggestion
import org.junit.Test
```
Once you remove the unused MWS JUnit 5 dep you'll probably have to switch this

okhttp-zstd/src/main/kotlin/okhttp3/zstd/ZstdInterceptor.kt
swankjesse: I don’t think so, but only because that configurable thing would be a big artifact.

okhttp-zstd/src/main/kotlin/okhttp3/zstd/ZstdInterceptor.kt
swankjesse: Same same.

okhttp-zstd/src/main/kotlin/okhttp3/zstd/ZstdInterceptor.kt
yschimke: I didn't mean combining all the compression libs.

I meant pulling out a Decompression Interceptor with all the header logic. That accepts multiple compressions strategies/factories.

`add interceptor(DecompressionInterceptor.Builder().withGzip().withZstd().build())`

Gzip built in, but zstd an extension function from this lib

okhttp-zstd/build.gradle.kts
squarejesse: Fixed.

okhttp-zstd/src/test/java/okhttp3/zstd/ZstdInterceptorTest.kt
squarejesse: Fixed.

Issue #7007, opened 2022-01-11
OkHttp using stale Connection after introducing IDLE_CONNECTION_HEALTHY_NS


Avishayy:
[Commit that introduced the optimization](https://github.com/square/okhttp/commit/19cb19ab4ac31aa789bc94759d13898f64f93ce3)

[Relevant issue on react-native](https://github.com/facebook/react-native/issues/32842#issuecomment-1009023135)

tl;dr on above issue: After upgrading react-native for my app, the OkHttp version changed from 3.12.1 to 4.9.1, the connection is being shutdown by the peer and `candidate.isHealthy()` deems the connection as healthy because it hasn't been idle for more than 10 seconds (which wasn't present on 3.12.1), but this is wrong because the connection is already closed and the next read will cause an `EOFException`

I'm not entirely sure if `react-native` was supposed to check if the peer shut the connection down or if the assumption in the above commit is bad, but someone did something wrong here, I'm not proficient enough to get a good answer on that, so I hope to get an answer here (and potentially fix the underlying issue).

yschimke:
Thanks for the thorough investigation on this.

Avishayy:
You're welcome, I'd love to hear your thoughts on this issue.

swankjesse:
@Avishayy any luck with `retryOnConnectionFailure (true)` ? 

One sneaky option we could try is being more aggressive about performing expensive checks if we know we're not going to retry if it fails.

Avishayy:
It does work with `retryOnConnectionFailure(true)`, and I like you idea.

swankjesse:
After some brief investigation, I think it’s best to fix this on the server. If it’s going to close a socket within 10 seconds of the HTTP response, it should send the `Connection: close` response header. That way OkHttp won’t pool the connection.

jacob-pro:
Would it be possible for OkHttp to make the `IDLE_CONNECTION_HEALTHY_NS` configurable?

- I am not able to make any changes to the HTTP server behaviour, so need to fix this in the client side.
- `retryOnConnectionFailure (true)` doesn't help because the request bodies have `isOneShot=true`

yschimke:
OkHttp generally doesn't expose all tuning knobs, preferring correct sensible default when that exists.

I'll look if there is a simple clean workaround using connection pool events.

yschimke:
@swankjesse thoughts on checking isOneShot?

PR #8944, opened 2025-07-16
doExtensiveHealthChecks based on isOneShot instead of method

jacob-pro:
This PR replaces the implementation of the `doExtensiveHealthChecks()` condition, which is used to decide whether to health-check pooled connections prior to re-use.

Currently the socket health check is enabled for all non-GET requests, afterwards the deep health check is only enabled for requests where isOneShot()=true. 

See comment: https://github.com/square/okhttp/issues/7007#issuecomment-2694879063

The docs for isOneShot say:

> Returns true if this body expects at most one call to writeTo and can be transmitted at most once. This is typically used when writing the request body is destructive and it is not possible to recreate the request body after it has been sent.
>
> By default OkHttp will attempt to retransmit request bodies when the original request fails due to any of:
>
> A stale connection. The request was made on a reused connection and that reused connection has since been closed by the server.

## Why

The extensive health check is expensive, and requires a 1ms read timeout check, it should be avoided where possible.

See similar PR: https://github.com/square/okhttp/pull/5795

> We were also doing the awkward and expensive read timeout
check on all non-GET requests on pooled connections. Now
we only perform these checks if the connection was idle
for 10 seconds or more.


jacob-pro:
@yschimke @swankjesse

yschimke:
I'll defer to @swankjesse 

My concerns would be
- whether it's a net performance loss for some conditions, such as slow mobile networks, because of sending expensive but also retryable POSTs that fail
- whether the existing logic is because GET is assume idempotent, but it's not clear it's about that.

yschimke:
Failing on formatting

 Run './gradlew :spotlessApply' to fix these violations.

jacob-pro:
Thanks @yschimke I have applied the formatting changes

swankjesse:
Unfortunately, `isOneShot()` is concerned with producing the response body (can we re-create the request body bytes if we need to re-transmit) rather than a defensive-check (is the action idempotent). I’d prefer to not conflate the two, though I may need to improve the documentation!

> The extensive health check is expensive, and requires a 1ms read timeout check, it should be avoided where possible.

Agreed that it’s expensive!

We currently don’t bother for connections used within the last 10 seconds.

```kotlin
    val idleDurationNs = withLock { nowNs - idleAtNs }
    if (idleDurationNs >= IDLE_CONNECTION_HEALTHY_NS && doExtensiveChecks) {
      return socket.isHealthy(source)
    }
```

Is this adequate? Would bumping it to 20 or 30 seconds help?

(Could you tell me a bit more about your use case?)

jacob-pro:
@swankjesse maybe I am missing something, but I don't understand what the relevance of idempotency is here?

First `chain.request.method != "GET"` is not very good test for idempotency in the first place, what about [PUT or DELETE](https://httpwg.org/specs/rfc9110.html#idempotent.methods) ?

Idempotency determines whether a request is safe to be retried, but that is not how it is used here, consider what would happen if you had:
- A non-idempotent POST
- On a connection last used 9s ago
- The connection has gone unhealthy
- The healthcheck would still be skipped due to idleDurationNs
- If idempotency was actually a concern then the behaviour would be unsafe depending on a 9s or 11s idle?

My understanding of the check is this should be a cost-benefit analysis on whether the extensive health check is likely to be worthwhile:

- There is low-value in doing the expensive health check on connections that were recently used (<10s) because they are unlikely to have become unhealthy
- Similarly there would be low-value in doing the expensive health check when the request could easily be retried by the `RetryAndFollowUpInterceptor`

https://github.com/square/okhttp/blob/d2dd1806973f7355b07a45fb1ef6a1bcfc094fdb/okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/http/RetryAndFollowUpInterceptor.kt#L145-L146

swankjesse:
I don’t think we can reliably tell whether a call on a reused connection failed before the server processed the request, or after the server processed the request. If it failed after the server processed the request, we have a correctness problem, not a performance problem.

So OkHttp is trying to be conservative here. 10 seconds is probably way too conservative, and we could probably bump it to 60 seconds without problem, but also I think the net number of health checks saved by that change would be negligible.

PR #8959, opened 2025-07-21
Update README.md: mention AndroidX startup

rfc2822:
okhttp 5.x requires AndroidX Startup.

When startup is disabled like in example 1 of https://developer.android.com/develop/background-work/background-tasks/persistent/configuration/custom-configuration#remove-default, it will have strange results like "IllegalStateException: Unable to load PublicSuffixDatabase.list resource" on PublicSuffixDatabase.get().getEffectiveTldPlusOne(domain).

This can be confusing and cause problems like https://github.com/bitfireAT/davx5-ose/issues/1602

See also https://github.com/square/okhttp/issues/8918

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,6 +98,8 @@ OkHttp works on Android 5.0+ (API level 21+) and Java 8+.
 
 OkHttp depends on [Okio][okio] for high-performance I/O and the [Kotlin standard library][kotlin]. Both are small libraries with strong backward-compatibility.
 
+On Android, okhttp uses [AndroidX Startup](https://developer.android.com/jetpack/androidx/releases/startup). Make sure that you don't disable Startup in the manifest.
+
 We highly recommend you keep OkHttp up-to-date. As with auto-updating web browsers, staying current
 with HTTPS clients is an important defense against potential security problems. [We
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -98,6 +98,8 @@ OkHttp works on Android 5.0+ (API level 21+) and Java 8+.
 
 OkHttp depends on [Okio][okio] for high-performance I/O and the [Kotlin standard library][kotlin]. Both are small libraries with strong backward-compatibility.
 
+On Android, okhttp uses [AndroidX Startup](https://developer.android.com/jetpack/androidx/releases/startup). Make sure that you don't disable Startup in the manifest.
```
yschimke, line 101: It's not all functionality, we should probably specify it's CookieJar or topPrivateDomain.

And we might need a public API workaround as a follow up.

rfc2822: How to proceed?

Shall I create an issue of the PR? Just wanted to mention that this can really cause confusing problems.

yschimke:
Yeah, create an issue. 

While we discuss the public API, I'll add warnings to logs when this problem surfaces.

yschimke:
Took a stab at a public API and README https://github.com/square/okhttp/pull/8965

Issue #8960, opened 2025-07-21
ClassCastException: InterruptedIOException cannot be cast to Void


swankjesse:
I added a timeout to the crawler and ran it, and I got a surprising crash.

```
java.lang.ClassCastException: class java.io.InterruptedIOException cannot be cast to class java.lang.Void (java.io.InterruptedIOException and java.lang.Void are in module java.base of loader 'bootstrap')
	at okhttp3.internal.connection.Exchange$ResponseBodySource.close(Exchange.kt:347)
	at okio.RealBufferedSource.close(RealBufferedSource.kt:493)
	at okhttp3.internal._UtilCommonKt.closeQuietly(-UtilCommon.kt:284)
	at okhttp3.ResponseBody.close(ResponseBody.kt:195)
	at okhttp3.Response.close(Response.kt:343)
	at okhttp3.sample.Crawler.fetch(Crawler.java:138)
	at okhttp3.sample.Crawler.drainQueue(Crawler.java:81)
	at okhttp3.sample.Crawler.lambda$parallelDrainQueue$0(Crawler.java:62)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
	at java.base/java.lang.Thread.run(Thread.java:1583)
```

neugartf:
Can confirm this happens in the wild for us with OpenTelemetry calls.

neugartf:
Any timeline for the next release? :)

swankjesse:
@neugartf it’s out!

PR #8961, opened 2025-07-21
Avoid a ClassCastException in ResponseBodySource.close

squarejesse:
Closes: https://github.com/square/okhttp/issues/8960


yschimke (approved):
LGTM

Is test infeasible?

squarejesse:
I triggered with a race while running the crawler with a pretty short timeout. It’ll happen on a race condition.

I don’t think I can do a unit test for this, but we could put some kind of fuzz-testing into our quality practice. (Not exactly sure what that looks like right now.)

MertNYuksel:
Hi, I noticed that the fix for this issue has already been merged, thank you! I was wondering if you could share when it might be included in a release?

swankjesse:
(it was available in a release on October 7.)

PR #8965, opened 2025-07-22
Add OkHttp.initialiseApplicationContext

yschimke:
A fallback for apps Disabling the AndroidX Startup Initialiser.

### DIFF
--- a/okhttp/src/androidMain/kotlin/okhttp3/InitialiseApplicationContext.kt
+++ b/okhttp/src/androidMain/kotlin/okhttp3/InitialiseApplicationContext.kt
@@ -1,0 +1,29 @@
+/*
+ * Copyright (C) 2025 Block, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okhttp3
+
+import android.content.Context
+import okhttp3.internal.platform.PlatformRegistry
+
+/**
+ * Configure the ApplicationContext. Not needed unless the AndroidX Startup Initialiser is disabled, or running
+ * a robolectric test.
+ *
+ * The functionality that will fail with a valid Context is primarily Cookies and URL Domain handling.
+ */
+fun OkHttp.initialiseApplicationContext(context: Context) {
+  PlatformRegistry.applicationContext = context.applicationContext
+}
--- a/okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/publicsuffix/BasePublicSuffixList.kt
+++ b/okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/publicsuffix/BasePublicSuffixList.kt
@@ -38,6 +38,8 @@ internal abstract class BasePublicSuffixList : PublicSuffixList {
   override lateinit var bytes: ByteString
   override lateinit var exceptionBytes: ByteString
 
+  private var readFailure: IOException? = null
+
   @Throws(IOException::class)
   private fun readTheList() {
     var publicSuffixListBytes: ByteString?
@@ -74,9 +76,11 @@ internal abstract class BasePublicSuffixList : PublicSuffixList {
       }
     }
 
-    check(::bytes.isInitialized) {
+    if (!::bytes.isInitialized) {
       // May have failed with an IOException
-      "Unable to load $path resource."
+      throw IllegalStateException("Unable to load $path resource.").apply {
+        initCause(readFailure)
+      }
     }
   }
 
@@ -99,6 +103,7 @@ internal abstract class BasePublicSuffixList : PublicSuffixList {
           interrupted = true
         } catch (e: IOException) {
           Platform.get().log("Failed to read public suffix list", Platform.Companion.WARN, e)
+          readFailure = e
           return
         }
       }
--- a/android-test/src/test/kotlin/okhttp/android/test/DisabledInitialiserTest.kt
+++ b/android-test/src/test/kotlin/okhttp/android/test/DisabledInitialiserTest.kt
@@ -16,10 +16,15 @@
  */
 package okhttp.android.test
 
-import androidx.test.core.app.ApplicationProvider
-import okhttp3.OkHttp
-import okhttp3.initialiseApplicationContext
-import org.junit.Before
+import assertk.all
+import assertk.assertFailure
+import assertk.assertions.cause
+import assertk.assertions.hasClass
+import assertk.assertions.hasMessage
+import assertk.assertions.isNotNull
+import java.io.IOException
+import okhttp3.HttpUrl.Companion.toHttpUrl
+import org.junit.Test
 import org.junit.runner.RunWith
 import org.robolectric.RobolectricTestRunner
 import org.robolectric.annotation.Config
@@ -28,11 +33,16 @@ import org.robolectric.annotation.Config
 @Config(
   sdk = [21, 26, 30, 33, 35],
 )
-class RobolectricOkHttpClientTest : BaseOkHttpClientUnitTest() {
-  @Before
-  fun setContext() {
-    // This is awkward because Robolectric won't run our initializers and we don't want test deps
-    // https://github.com/robolectric/robolectric/issues/8461
-    OkHttp.initialiseApplicationContext(ApplicationProvider.getApplicationContext())
+class DisabledInitialiserTest {
+  @Test
+  fun testWithoutContext() {
+    val httpUrl = "https://www.google.co.uk".toHttpUrl()
+    assertFailure { httpUrl.topPrivateDomain() }.all {
+      hasMessage("Unable to load PublicSuffixDatabase.list resource.")
+      cause().isNotNull().all {
+        hasMessage("Platform applicationContext not initialized.")
+        hasClass<IOException>()
+      }
+    }
   }
 }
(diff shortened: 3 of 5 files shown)


okhttp/src/androidMain/kotlin/okhttp3/InitialiseApplicationContext.kt
```diff
@@ -0,0 +1,29 @@
...
+package okhttp3
+
+import android.content.Context
+import okhttp3.internal.platform.PlatformRegistry
+
+/**
+ * Configure the ApplicationContext. Not needed unless the AndroidX Startup Initialiser is disabled, or running
+ * a robolectric test.
+ *
+ * The functionality that will fail with a valid Context is primarily Cookies and URL Domain handling.
```
swankjesse: And it’ll be more soon!

okhttp/src/androidMain/kotlin/okhttp3/InitialiseApplicationContext.kt
```diff
@@ -0,0 +1,29 @@
...
+ */
+fun OkHttp.initialiseApplicationContext(context: Context) {
```
swankjesse: Rename to `OkHttp.initialize` ? The suffix `ApplicationContext` is redundant, and we use USA spelling

okhttp/src/androidMain/kotlin/okhttp3/InitialiseApplicationContext.kt
```diff
@@ -0,0 +1,29 @@
...
+  PlatformRegistry.applicationContext = context.applicationContext
```
swankjesse: Should this no-op if it’s already initialized? Changing this retroactively scares me a bit

swankjesse (approved):

okhttp/src/androidMain/kotlin/okhttp3/InitialiseApplicationContext.kt
JakeWharton: Can we make it a proper static on `OkHttp` for the benefit of Java callers?

swankjesse: Good idea, though it will need some expect/actual effort.

yschimke: I think it's a signature change for OkHttp.VERSION

yschimke: Fixed with JVM and Android specific JvmField on a now deprecated VERSION. And added new Version in common,

yschimke:
@swankjesse different enough, that it's worth a re-review. So not assuming previous approval.

okhttp/src/androidMain/kotlin/okhttp3/OkHttp.android.kt
```diff
@@ -0,0 +1,47 @@
...
+import okhttp3.internal.platform.Platform
+import okhttp3.internal.platform.PlatformRegistry
+
+actual object OkHttp {
+  @JvmStatic
+  actual val Version: String = CONST_VERSION
+
+  @Suppress("MayBeConstant") // Non-const so external callers get the runtime version.
+  @JvmField
+  @Deprecated(level = DeprecationLevel.WARNING, message = "Replaced with Version", replaceWith = ReplaceWith("Version"))
```
swankjesse: Why deprecate this?

okhttp/src/androidMain/kotlin/okhttp3/OkHttp.android.kt
```diff
@@ -0,0 +1,47 @@
...
+package okhttp3
+
+import android.content.Context
+import okhttp3.internal.CONST_VERSION
+import okhttp3.internal.platform.Platform
+import okhttp3.internal.platform.PlatformRegistry
+
+actual object OkHttp {
+  @JvmStatic
+  actual val Version: String = CONST_VERSION
```
swankjesse: Isn’t `VERSION` sufficient?

okhttp/src/androidMain/kotlin/okhttp3/OkHttp.android.kt
yschimke: I guess no need. But it's not available via common now.

okhttp/src/androidMain/kotlin/okhttp3/OkHttp.android.kt
```diff
@@ -0,0 +1,47 @@
...
+  /**
+   * Configure the ApplicationContext. Not needed unless the AndroidX Startup [Initializer] is disabled, or running
+   * a robolectric test.
+   *
+   * The functionality that will fail without a valid Context is primarily Cookies and URL Domain handling, but
+   * may expand in the future.
+   */
+  fun initialize(applicationContext: Context) {
+    if (PlatformRegistry.applicationContext != null) {
+      Platform.get().log("Ignoring OkHttp.initialize as applicationContext already initialised")
```
swankjesse: Why log? It seems like the kind of log message that isn’t actionable enough. I’d rather we either fail with an exception (so the developer can fix it) or silently fail (because who cares?)

okhttp/src/androidMain/kotlin/okhttp3/OkHttp.android.kt
yschimke: Not sure how to have expect JvmField.

okhttp/src/androidMain/kotlin/okhttp3/internal/publicsuffix/AssetPublicSuffixList.kt
```diff
@@ -24,8 +25,22 @@ internal class AssetPublicSuffixList(
...
   override fun listSource(): Source {
-    val assets =
-      PlatformRegistry.applicationContext?.assets ?: throw IOException("Platform applicationContext not initialized")
+    val assets = PlatformRegistry.applicationContext?.assets
+
+    if (assets == null) {
+      if (Build.FINGERPRINT == null) {
+        throw IOException(
+          "Platform applicationContext not initialized. " +
+            "Possibly running Android unit test without Robolectric.",
```
swankjesse: Any idea what action we should recommend callers take here?

okhttp/src/commonJvmAndroid/kotlin/okhttp3/OkHttp.kt
```diff
@@ -15,12 +15,9 @@
...
 package okhttp3
 
-import kotlin.jvm.JvmField
-import okhttp3.internal.CONST_VERSION
-
-object OkHttp {
+expect object OkHttp {
   /**
-   * This is a string like "4.5.0-RC1", "4.5.0", or "4.6.0-SNAPSHOT" indicating the version of
+   * This is a string like "5.0.0", "5.0.0-alpha.762", or "5.3.0-SNAPSHOT" indicating the version of
```
swankjesse, line 20: Nice

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/publicsuffix/BasePublicSuffixList.kt
```diff
@@ -99,6 +103,7 @@ internal abstract class BasePublicSuffixList : PublicSuffixList {
           interrupted = true
         } catch (e: IOException) {
           Platform.get().log("Failed to read public suffix list", Platform.Companion.WARN, e)
```
swankjesse: Is this log unnecessary now? We’ll throw the right exception when attempting to access it?

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/publicsuffix/PublicSuffixDatabase.kt
```diff
@@ -244,5 +244,9 @@ class PublicSuffixDatabase internal constructor(
       }
       return match
     }
+
+    internal fun resetForTests() {
```
swankjesse, line 248: I don’t like that we need this. Seems like we screwed up our dependency injection.

okhttp/src/androidMain/kotlin/okhttp3/OkHttp.android.kt
swankjesse: Ooooh, ouch.

swankjesse: I think it just works if you put `@JvmField` on the actuals?

```kotlin
expect object OkHttp {
  @JvmStatic
  val VERSION: String
}
```

```kotlin
actual object OkHttp {
  @Suppress("MayBeConstant") // Non-const so external callers get the runtime version.
  @JvmField
  actual val VERSION: String = CONST_VERSION
}
```

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/publicsuffix/PublicSuffixDatabase.kt
swankjesse: (I don’t have an alternative to recommend)

okhttp/src/androidMain/kotlin/okhttp3/OkHttp.android.kt
yschimke: Ok, will try. Thanks

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/publicsuffix/PublicSuffixDatabase.kt
yschimke: I think it's that apps have a single go at this. It either works or it fails once and permanently.

But we want to test

okhttp/src/androidMain/kotlin/okhttp3/OkHttp.android.kt
yschimke: This has a warning

```
[ACTUAL_ANNOTATIONS_NOT_MATCH_EXPECT] Annotation `@JvmStatic()` is missing on actual declaration.
All annotations from expect 'val VERSION: String get(): String' must be present with the same arguments on actual 'field:val VERSION: String get(): String', otherwise they might behave incorrectly.
```

That sounds ominous.

But seems to work in common test and the API files.

JakeWharton: Seems like you can completely drop `@JvmStatic` there

yschimke: https://github.com/square/okhttp/pull/8971

I swear this didn't work for me. But it is now.

PR #8967, opened 2025-07-27
Add reusable CompressionInterceptor.kt

yschimke:
Refactor Brotli and Zstd to use this, but also
allow trivial configuration of zstd,br,gzip.


yschimke:
@swankjesse you didn't seem convinced of the value, but I figure maybe through the medium of PR, it will change your mind.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/CompressionInterceptor.kt
```diff
@@ -0,0 +1,85 @@
...
+import okio.buffer
+
+/**
+ * Transparent Compressed response support.
+ *
+ * Adds Accept-Encoding to request and checks (and strips) Content-Encoding in
+ * responses. n.b. this replaces the transparent gzip compression in BridgeInterceptor so should generally include
+ * gzip in the list of algorithms.
+ */
+open class CompressionInterceptor(
```
yschimke, line 35: @swankjesse the name is a bit weird, but I wonder whether this should support configurable request compression also?

okhttp/src/commonJvmAndroid/kotlin/okhttp3/CompressionInterceptor.kt
```diff
@@ -0,0 +1,85 @@
...
+import okhttp3.internal.http.promisesBody
+import okio.BufferedSource
+import okio.GzipSource
+import okio.Source
+import okio.buffer
+
+/**
+ * Transparent Compressed response support.
+ *
+ * Adds Accept-Encoding to request and checks (and strips) Content-Encoding in
```
swankjesse: I don’t like public-API comments referencing internal implementation details

okhttp/src/commonJvmAndroid/kotlin/okhttp3/CompressionInterceptor.kt
```diff
@@ -0,0 +1,85 @@
...
+      .removeHeader("Content-Encoding")
+      .removeHeader("Content-Length")
+      .body(decompressedSource.asResponseBody(body.contentType(), -1))
+      .build()
+  }
+
+  interface DecompressionAlgorithm {
+    val encoding: String
+
+    fun BufferedSource.decompress(): Source
```
swankjesse: Replace with a single receiver function? This is awkward to call

okhttp/src/commonJvmAndroid/kotlin/okhttp3/CompressionInterceptor.kt
```diff
@@ -0,0 +1,85 @@
...
+          .build()
+
+      val response = chain.proceed(request)
+
+      decompress(response)
+    } else {
+      chain.proceed(chain.request())
+    }
+
+  fun decompress(response: Response): Response {
```
swankjesse: Document that this returns the response as-is if it doesn’t have a response body, or if the response body isn't compressed with one of the configured algorithms?

okhttp/src/commonJvmAndroid/kotlin/okhttp3/CompressionInterceptor.kt
```diff
@@ -0,0 +1,85 @@
...
+open class CompressionInterceptor(
+  vararg val algorithms: DecompressionAlgorithm,
+) : Interceptor {
+  override fun intercept(chain: Interceptor.Chain): Response =
+    if (chain.request().header("Accept-Encoding") == null) {
+      val request =
+        chain
+          .request()
+          .newBuilder()
+          .header("Accept-Encoding", algorithms.joinToString(separator = ",") { it.encoding })
```
swankjesse: Precompute `algorithms.joinToString(separator = ",") { it.encoding })` and stash it in a private property?

okhttp-zstd/src/main/kotlin/okhttp3/zstd/ZstdInterceptor.kt
```diff
@@ -33,43 +30,11 @@ import okio.gzip
  *
  * This replaces the transparent gzip compression in OkHttp.
  */
-object ZstdInterceptor : Interceptor {
-  override fun intercept(chain: Interceptor.Chain): Response {
-    val originalRequest = chain.request()
+object ZstdInterceptor : CompressionInterceptor(Zstd, Gzip)
```
swankjesse: We can remove this object from the public API. We haven’t published a release with it yet, and with your new API the caller can do something almost as good:

```kotlin
  .addInterceptor(CompressionInterceptor(Zstd, Gzip))
  .build()
```

okhttp/src/commonJvmAndroid/kotlin/okhttp3/CompressionInterceptor.kt
```diff
@@ -0,0 +1,85 @@
...
+
+/**
+ * Transparent Compressed response support.
+ *
+ * Adds Accept-Encoding to request and checks (and strips) Content-Encoding in
+ * responses. n.b. this replaces the transparent gzip compression in BridgeInterceptor so should generally include
+ * gzip in the list of algorithms.
+ */
+open class CompressionInterceptor(
+  vararg val algorithms: DecompressionAlgorithm,
```
swankjesse, line 36: What should happen if algorithms is empty?
Should we require that it includes Gzip ?

swankjesse (approved):
I still don’t really love it, but lets do it. Giving people who are already using `br` a path to doing `br` + `zstd` is good.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/CompressionInterceptor.kt
yschimke: Or empty sets identity encoding to disable compression?

yschimke: You only have yourself to blame, for tempting me to read a spec

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Accept-Encoding

I think it's a good thing?

okhttp/src/commonJvmAndroid/kotlin/okhttp3/CompressionInterceptor.kt
```diff
@@ -0,0 +1,150 @@
...
+ * Transparent Compressed response support.
+ *
+ * The algorithm map will be turned into a heading such as "Accept-Encoding: br;q=1.0, gzip;q=0.8, *;q=0.1"
+ *
+ * If [algorithms] is empty returns "identity" disabling any later compression.
+ *
+ * See https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Accept-Encoding
+ */
+open class CompressionInterceptor(
+  val algorithms: Map<DecompressionAlgorithm, Double?>,
```
yschimke: An alternative to this Map, would be for DecompressionAlgorithm to have a priority, defaulting to null. But also a `withPriority()` method to clone and assign.

yschimke:
@swankjesse definitely enough changes to need another review

okhttp-zstd/src/main/kotlin/okhttp3/zstd/ZstdInterceptor.kt
```diff
@@ -16,60 +16,16 @@
...
 /**
- * Transparent Zstandard response support.
- *
- * This must be installed as an application interceptor.
- *
- * Adds `Accept-Encoding: zstd,gzip` to request and checks (and strips) for `Content-Encoding: zstd`
- * in responses.
- *
- * This replaces the transparent gzip compression in OkHttp.
+ * Support for Zstandard encoding. Use via the [CompressionInterceptor].
```
swankjesse, line 24: 👍🏻

okhttp/src/commonJvmAndroid/kotlin/okhttp3/CompressionInterceptor.kt
swankjesse: I preferred the varargs. We don’t need a priority if we have a sequence

okhttp/src/commonJvmAndroid/kotlin/okhttp3/CompressionInterceptor.kt
```diff
@@ -0,0 +1,150 @@
...
+    vararg algorithms: DecompressionAlgorithm,
+  ) : this(
+    algorithms.associateWithTo(LinkedHashMap()) {
+      null
+    },
+  )
+
+  internal val acceptEncoding =
+    if (algorithms.isEmpty()) {
+      "identity"
```
swankjesse: Or omit the header?

okhttp/src/commonJvmAndroid/kotlin/okhttp3/CompressionInterceptor.kt
```diff
@@ -0,0 +1,150 @@
...
+     */
+    val Identity =
+      object : DecompressionAlgorithm {
+        override val encoding: String = "identity"
+
+        override fun decompress(compressedSource: BufferedSource): Source = compressedSource
+      }
+
+    /**
+     * Declare support for any compression encoding scheme. However the response will not be decompressed.
```
swankjesse: Why?

okhttp/src/commonJvmAndroid/kotlin/okhttp3/CompressionInterceptor.kt
```diff
@@ -0,0 +1,150 @@
...
+     */
+    val Gzip =
+      object : DecompressionAlgorithm {
+        override val encoding: String = "gzip"
+
+        override fun decompress(compressedSource: BufferedSource): Source = GzipSource(compressedSource)
+      }
+
+    /**
+     * Request the response body to be uncompressed.
```
swankjesse: Isn’t this implied?

swankjesse (commented):
I think I preferred the previous commit!

okhttp/src/commonJvmAndroid/kotlin/okhttp3/CompressionInterceptor.kt
yschimke: There is a spec, should we aim to support it?

Or it's basically not used, so we just support what we need for zstd?

okhttp/src/commonJvmAndroid/kotlin/okhttp3/CompressionInterceptor.kt
yschimke: I figured if you set this, you want to disable later automatic compression?

swankjesse: For that use-case we should just get ’em to do it directly by setting `Accept-Encoding: identity` in a header

okhttp/src/commonJvmAndroid/kotlin/okhttp3/CompressionInterceptor.kt
swankjesse: Yeah I don’t think we need the full power of that spec.

yschimke: Will revert to previous commit and land.

okhttp-brotli/api/okhttp-brotli.api
```diff
@@ -1,5 +1,8 @@
-public final class okhttp3/brotli/BrotliInterceptor : okhttp3/Interceptor {
+public final class okhttp3/brotli/BrotliInterceptor : okhttp3/CompressionInterceptor {
 	public static final field INSTANCE Lokhttp3/brotli/BrotliInterceptor;
-	public fun intercept (Lokhttp3/Interceptor$Chain;)Lokhttp3/Response;
+}
+
+public final class okhttp3/brotli/BrotliInterceptorKt {
```
swankjesse, line 5: Would you consider making this a member of `BrotliInterceptor` instead? Having `Kt` in the class name is our finger in the eye of Java language users

okhttp-zstd/api/okhttp-zstd.api
```diff
@@ -1,5 +1,4 @@
-public final class okhttp3/zstd/ZstdInterceptor : okhttp3/Interceptor {
-	public static final field INSTANCE Lokhttp3/zstd/ZstdInterceptor;
-	public fun intercept (Lokhttp3/Interceptor$Chain;)Lokhttp3/Response;
+public final class okhttp3/zstd/ZstdInterceptorKt {
```
swankjesse, line 1: What should this symbol name be from Java? Perhaps `Zstd.INSTANCE` ?

okhttp/api/android/okhttp.api
```diff
@@ -326,6 +326,22 @@ public final class okhttp3/CipherSuite$Companion {
...
 
+public class okhttp3/CompressionInterceptor : okhttp3/Interceptor {
+	public static final field Companion Lokhttp3/CompressionInterceptor$Companion;
+	public fun <init> ([Lokhttp3/CompressionInterceptor$DecompressionAlgorithm;)V
+	public final fun getAlgorithms ()[Lokhttp3/CompressionInterceptor$DecompressionAlgorithm;
+	public fun intercept (Lokhttp3/Interceptor$Chain;)Lokhttp3/Response;
+}
+
+public final class okhttp3/CompressionInterceptor$Companion {
+	public final fun getGzip ()Lokhttp3/CompressionInterceptor$DecompressionAlgorithm;
```
swankjesse, line 337: Should this be `@JvmStatic` so Java users don’t need to use the Companion class?

yschimke: @JvmField?

PR #8970, opened 2025-07-28
Fix RequestBody events on upgraded connections

gesellix:
See https://docs.docker.com/reference/api/engine/version/v1.51/#tag/Container/operation/ContainerAttach

The following RFC paragraph might be relevant: https://datatracker.ietf.org/doc/html/rfc9110#section-7.8-15
> A client cannot begin using an upgraded protocol on the connection until it has completely sent the request message (i.e., the client can't change the protocol it is sending in the middle of a message). If a server receives both an Upgrade and an [Expect](https://datatracker.ietf.org/doc/html/rfc9110#field.expect) header field with the "100-continue" expectation ([Section 10.1.1](https://datatracker.ietf.org/doc/html/rfc9110#field.expect)), the server MUST send a [100 (Continue)](https://datatracker.ietf.org/doc/html/rfc9110#status.100) response before sending a [101 (Switching Protocols)](https://datatracker.ietf.org/doc/html/rfc9110#status.101) response.


okhttp/src/jvmTest/kotlin/okhttp3/internal/http/HttpUpgradesTest.kt
```diff
@@ -210,17 +217,32 @@ class HttpUpgradesTest {
...
+      "RequestBodyEnd",
+      "ResponseHeadersStart",
+      "ResponseHeadersEnd",
+      "RequestBodyStart",
+      "FollowUpDecision",
+      "ResponseBodyStart",
+      "ResponseBodyEnd",
+      "ConnectionReleased",
+      "CallEnd",
+      "RequestBodyEnd",
```
gesellix: The `RequestBodyEnd` after the `CallEnd` one looks strange to me.

squarejesse: Also the repeated RequestBodyEnd suggests we should rework this

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Request.kt
```diff
@@ -82,8 +82,8 @@ class Request internal constructor(
 
   init {
     val connectionHeader = headers["Connection"]
-    require(body == null || !"upgrade".equals(connectionHeader, ignoreCase = true)) {
-      "expected a null request body with 'Connection: upgrade'"
+    require(body == null || !"upgrade".equals(connectionHeader, ignoreCase = true) || body.contentLength() <= 0) {
```
gesellix: Or even `body.contentLength() == 0`

okhttp/src/jvmTest/kotlin/okhttp3/internal/http/HttpUpgradesTest.kt
```diff
@@ -201,6 +210,8 @@ class HttpUpgradesTest {
       "ResponseHeadersEnd",
       "RequestBodyStart",
       "FollowUpDecision",
+      "SocketSinkStart",
+      "SocketSinkEnd",
```
gesellix: Not sure if these events should appear after the RequestBodyEnd event.

okhttp/src/jvmTest/kotlin/okhttp3/internal/http/HttpUpgradesTest.kt
```diff
@@ -210,7 +221,36 @@ class HttpUpgradesTest {
...
+      "ConnectionAcquired",
+      "RequestHeadersStart",
+      "RequestHeadersEnd",
+      "RequestBodyStart",
+      "RequestBodyEnd",
+      "ResponseHeadersStart",
+      "ResponseHeadersEnd",
+      "FollowUpDecision",
+      "SocketSinkStart",
+      "SocketSinkEnd",
```
gesellix: Not sure if these events should appear after the RequestBodyEnd event.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/EventListener.kt
```diff
@@ -290,6 +290,25 @@ abstract class EventListener {
   ) {
   }
 
+  /**
+   * Invoked just prior to sending bytes on the socket. Will only be invoked for upgraded connections.
+   *
+   * The connection is implicit, and will generally relate to the last [connectionAcquired] event.
```
swankjesse: Can we strengthen this guarantee? Maybe we should pass the connection itself to this method?

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Request.kt
swankjesse: It’s gotta be `body.contentLength() == 0`. If we permit `body.contentLength() == -1` that is not an empty body.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/connection/Exchange.kt
```diff
@@ -248,7 +248,10 @@ class Exchange(
     private val contentLength: Long,
   ) : ForwardingSink(delegate) {
     private var completed = false
+
+    // TODO bytesReceived or bytesWritten?
     private var bytesReceived = 0L
+    private var invokeStartEvent = true
```
swankjesse: I think this class needs to know if it’s being used for the socket or a regular request body

okhttp/src/jvmTest/kotlin/okhttp3/internal/http/HttpUpgradesTest.kt
swankjesse: My preference is that if it’s an upgrade request, we emit a `Content-Length: 0` header but ignore the request body, and emit zero events related to it

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/connection/Exchange.kt
```diff
@@ -143,11 +147,10 @@ class Exchange(
   fun peekTrailers(): Headers? = codec.peekTrailers()
 
   fun upgradeToSocket(): Socket {
+    isSocket = true
```
gesellix, line 150: Might this one fail due to race conditions?

okhttp/src/commonJvmAndroid/kotlin/okhttp3/EventListener.kt
gesellix: Similar to the checks which ensure that *End happens after *Start, leveraging this [generic check](https://github.com/square/okhttp/blob/f84fb612a455a49e3fe2c7a4516d0f2badaee708/okhttp-testing-support/src/main/kotlin/okhttp3/RecordingEventListener.kt#L148-L150)?

swankjesse: Yeah that’s a great idea

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/connection/Exchange.kt
swankjesse: Nope. There’s only one thread operating on the exchange when this happens.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/EventListener.kt
gesellix: I passed the connection to both `socketSinkStart` and `socketSourceStart` now, the `RecordingEventListener` ensures the correct connection for exactly those events. I didn't want to mix the connection check with the closes check, but the dedicated check also doesn't make me happy. WDYT?

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/connection/Exchange.kt
gesellix: `isSocket` will become `true` on `upgradeToSocket()`, so `invokeStartEvent` will become `true`, too.

okhttp/src/jvmTest/kotlin/okhttp3/internal/http/HttpUpgradesTest.kt
gesellix: Now neither `RequestBodyStart` nor `RequestBodyEnd` events are emitted for upgraded events.
I don't think we need the `Content-Length: 0` header, or am I missing something?

gesellix:
We won't aim at adding socket specific events for the time being. This pull request is now reduced to only cleanup some of the request body events when using an upgraded connection.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/connection/Exchange.kt
```diff
@@ -185,6 +188,9 @@ class Exchange(
     codec.carrier.trackFailure(call, e)
   }
 
+  // TODO maybe rename, so that a socket use case (say: no body) would still make sense,
+  // or call a socket specific variant instead?
+
```
swankjesse:
```suggestion
```

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/connection/Exchange.kt
```diff
@@ -232,7 +238,10 @@ class Exchange(
     private val contentLength: Long,
   ) : ForwardingSink(delegate) {
     private var completed = false
+
+    // TODO bytesReceived or bytesWritten?
```
swankjesse:
```suggestion
```

swankjesse: Valid todo, but let’s just make the change.

PR #8972, opened 2025-07-28
New MockWebServerSocket helper class

swankjesse:
This is work towards switching to use okio.Socket
as the implementation of web sockets.


mockwebserver/src/main/kotlin/mockwebserver3/MockWebServer.kt
```diff
@@ -855,31 +849,30 @@ public class MockWebServer : Closeable {
       // Even if messages are no longer being read we need to wait for the connection close signal.
       connectionClose.await()
     } finally {
-      source.closeQuietly()
+      socket.source.closeQuietly()
```
yschimke, line 852: Can this just be socket.closeQuietly() now?

mockwebserver/src/main/kotlin/mockwebserver3/MockWebServer.kt
```diff
@@ -828,7 +822,7 @@ public class MockWebServer : Closeable {
 
     val connectionClose = CountDownLatch(1)
     val streams =
-      object : RealWebSocket.Streams(false, source, sink) {
+      object : RealWebSocket.Streams(false, socket.source, socket.sink) {
```
yschimke, line 825: I assume eventually you'll pass socket down?

mockwebserver/src/main/kotlin/mockwebserver3/MockWebServer.kt
```diff
@@ -449,33 +446,33 @@ public class MockWebServer : Closeable {
...
+            ) as SSLSocket
+          socket = MockWebServerSocket(sslSocket)
           sslSocket.useClientMode = false
           if (clientAuth == CLIENT_AUTH_REQUIRED) {
             sslSocket.needClientAuth = true
           } else if (clientAuth == CLIENT_AUTH_REQUESTED) {
             sslSocket.wantClientAuth = true
           }
-          openClientSockets.add(socket)
+          openClientSockets.add(sslSocket)
```
yschimke, line 475: worth putting MockWebServerSocket in openClientSockets?

mockwebserver/src/main/kotlin/mockwebserver3/MockWebServer.kt
yschimke: or is this relying on

>  Note that [asOkioSocket] returns a socket that closes the underlying [java.net.Socket] when both of its component streams are closed.

swankjesse: Yep. I’m working towards that but I’m not there yet.

mockwebserver/src/main/kotlin/mockwebserver3/MockWebServer.kt
swankjesse: EXACTLY!

mockwebserver/src/main/kotlin/mockwebserver3/MockWebServer.kt
swankjesse: It’s a bit annoying ’cause we don’t create the `MockWebServerSocket` until we know if it’s SSL or not, but we need it in that collection before that.

yschimke (approved):

PR #8973, opened 2025-07-29
Use okio.Socket internally in OkHttp connections

swankjesse:


mockwebserver/src/main/kotlin/mockwebserver3/internal/MockWebServerSocket.kt
```diff
@@ -39,9 +42,31 @@ internal class MockWebServerSocket(
 ) : Closeable,
   okio.Socket {
   private val delegate = javaNetSocket.asOkioSocket()
+  private val closedLatch = CountDownLatch(2)
```
yschimke, line 45: nice!

yschimke (approved):

PR #8974, opened 2025-07-29
Re-implement web sockets support on the new Response.socket feature

squarejesse:


okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/ws/RealWebSocket.kt
```diff
@@ -257,26 +259,28 @@ class RealWebSocket(
...
     synchronized(this) {
       this.name = name
-      this.streams = streams
+      this.socket = socket
       this.writer =
         WebSocketWriter(
-          isClient = streams.client,
-          sink = streams.sink,
+          isClient = client,
+          sink = socketSink,
```
yschimke, line 284: Why are the sink and socket passed in?

swankjesse: Great question. I added this comment.

```kotlin
  /**
   * This accepts a [BufferedSource] instead of using [Socket.source], just in case we've already
   * received data from the peer. This accepts a [BufferedSink] for symmetry with the source.
   */
```

okhttp/src/jvmTest/kotlin/okhttp3/internal/http/HttpUpgradesTest.kt
```diff
@@ -82,12 +82,12 @@ class HttpUpgradesTest {
         socket.sink.buffer().use { sink ->
           socket.source.buffer().use { source ->
             sink.writeUtf8("client says hello\n")
-            sink.flush()
+            sink.emit()
```
gesellix, line 85: Is this an optimization?

swankjesse: Kinda sort of. I wanted to confirm that we weren’t unnecessarily wrapping the streams in unnecessary buffering.

PR #8977, opened 2025-07-30
New BufferedSocket class

squarejesse:
I'm starting to push Okio's socket throughout OkHttp, and it turns out what I usually want is a BufferedSocket instead.

### DIFF
--- a/mockwebserver/src/test/java/mockwebserver3/internal/http2/Http2Server.kt
+++ b/mockwebserver/src/test/java/mockwebserver3/internal/http2/Http2Server.kt
@@ -17,6 +17,7 @@ package mockwebserver3.internal.http2
 
 import java.io.File
 import java.io.IOException
+import java.net.InetSocketAddress
 import java.net.ProtocolException
 import java.net.ServerSocket
 import java.net.Socket
@@ -28,6 +29,7 @@ import okhttp3.Protocol
 import okhttp3.Protocol.Companion.get
 import okhttp3.internal.closeQuietly
 import okhttp3.internal.concurrent.TaskRunner
+import okhttp3.internal.connection.asBufferedSocket
 import okhttp3.internal.http2.Header
 import okhttp3.internal.http2.Http2Connection
 import okhttp3.internal.http2.Http2Stream
@@ -57,7 +59,7 @@ class Http2Server(
         val connection =
           Http2Connection
             .Builder(false, TaskRunner.INSTANCE)
-            .socket(sslSocket)
+            .socket(sslSocket.asBufferedSocket(), sslSocket.peerName())
             .listener(this)
             .build()
         connection.start()
@@ -191,6 +193,11 @@ class Http2Server(
       file.name.endsWith(".png") -> "image/png"
       else -> "text/plain"
     }
+
+  private fun Socket.peerName(): String {
+    val address = remoteSocketAddress
+    return if (address is InetSocketAddress) address.hostName else address.toString()
+  }
 
   companion object {
     val logger: Logger = Logger.getLogger(Http2Server::class.java.name)
--- a/okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/connection/BufferedSocket.kt
+++ b/okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/connection/BufferedSocket.kt
@@ -1,0 +1,42 @@
+/*
+ * Copyright (C) 2025 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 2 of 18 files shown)


mockwebserver/src/test/java/mockwebserver3/internal/http2/Http2Server.kt
```diff
@@ -57,7 +59,7 @@ class Http2Server(
         val connection =
           Http2Connection
             .Builder(false, TaskRunner.INSTANCE)
-            .socket(sslSocket)
+            .socket(sslSocket.asBufferedSocket(), sslSocket.peerName())
```
yschimke, line 62: Should something like peername be an optional property of an okio socket

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/connection/BufferedSocket.kt
```diff
@@ -0,0 +1,42 @@
...
+
+import java.net.Socket as JavaNetSocket
+import okio.BufferedSink
+import okio.BufferedSource
+import okio.Socket
+import okio.Socket as OkioSocket
+import okio.asOkioSocket
+import okio.buffer
+
+interface BufferedSocket : OkioSocket {
```
yschimke, line 25: Worth adding to okio?

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/connection/RealConnection.kt
```diff
@@ -75,12 +69,10 @@ class RealConnection internal constructor(
    * The application layer socket. Either an [SSLSocket] layered over [rawSocket], or [rawSocket]
    * itself if this connection does not use SSL.
    */
-  private val socket: JavaNetSocket,
+  private val javaNetSocket: JavaNetSocket,
```
yschimke, line 72: Rather than passing around, should there  be an unwrap mechanism?

Or way to apply to an underlying type?

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/connection/BufferedSocket.kt
```diff
@@ -0,0 +1,42 @@
...
+  override val source: BufferedSource
+  override val sink: BufferedSink
+}
+
+fun JavaNetSocket.asBufferedSocket(): BufferedSocket = asOkioSocket().asBufferedSocket()
+
+fun Socket.asBufferedSocket(): BufferedSocket =
```
JakeWharton:
```suggestion
fun OkioSocket.asBufferedSocket(): BufferedSocket =
```
You have this imported twice as two different names

mockwebserver/src/test/java/mockwebserver3/internal/http2/Http2Server.kt
squarejesse: I don‘t think so.

java.net.Socket models a full TCP socket and has features for InetAddresses and such, but okio.Socket is just a source+sink.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/connection/BufferedSocket.kt
squarejesse: Maybe someday! I don’t think it’s too urgent though.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/connection/RealConnection.kt
squarejesse: I’m not quite sure yet. In this class we need both, but most of the time we should only ever need the okio.Socket 'cause that one is more general-purpose. It can model an HTTP/2 stream whereas this cannot.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/connection/BufferedSocket.kt
swankjesse: Ooooh that’s embarrassing. Fixed!

okhttp/src/jvmTest/kotlin/okhttp3/internal/http/HttpUpgradesTest.kt
```diff
@@ -82,12 +82,12 @@ class HttpUpgradesTest {
...
           socket.source.buffer().use { source ->
             sink.writeUtf8("client says hello\n")
-            sink.emit()
+            sink.flush()
 
             assertThat(source.readUtf8Line()).isEqualTo("server says hello")
 
             sink.writeUtf8("client says goodbye\n")
-            sink.emit()
+            sink.flush()
```
squarejesse, line 90: I wanted to avoid multiple layers of `BufferedSource` in this, but it’s super annoying to avoid it all the way through, and the cost is tiny.

The server can always send us data early, and so our `BufferedSource` instances may receive more than they need, and we need to keep that data.

PR #8989, opened 2025-08-01
Run OSGi tests on JDK 17

yschimke:


yschimke:
@swankjesse gradle test config only, running more than previously. requesting post review

PR #8991, opened 2025-08-02
Split android OkHttpClient tests.

yschimke:
current tests were using a single client then
testing real network (facebook, ...) and localhost insecure.

Split so we get clearer failures.


yschimke:
@swankjesse requesting post review. landing to have clearer tests for okio upgrade failure.

PR #8993, opened 2025-08-03
Add minimal HttpLoggingInterceptor support for streaming request and responses.

yschimke:
```
      /**
       * Logs streaming request and response lines.
       *
       * Example:
       * ```
       * --> POST /greeting http/1.1
       * <-- 200 OK (22ms, unknown-length body)
       * > request A
       *
       * < response B
       *
       * > request C
       *
       * < response D
       *
       * --> END POST
       * <-- END HTTP
       * ```
       */
      STREAMING,
```

Also allows level to be adjusted per Request via

```
class HttpLoggingInterceptor {
...
    var requestLevel: (Request) -> Level = { level }
```

These use cases are much more common with protocols like MCP adopting Streaming HTTP.

https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#streamable-http

> The body of the POST request MUST be one of the following:
A single JSON-RPC request, notification, or response
An array [batching](https://www.jsonrpc.org/specification#batch) one or more requests and/or notifications
An array [batching](https://www.jsonrpc.org/specification#batch) one or more responses

> The SSE stream SHOULD eventually include one JSON-RPC response per each JSON-RPC request sent in the POST body. These responses MAY be [batched](https://www.jsonrpc.org/specification#batch).

### DIFF
--- a/okhttp-logging-interceptor/src/main/kotlin/okhttp3/logging/HttpLoggingInterceptor.kt
+++ b/okhttp-logging-interceptor/src/main/kotlin/okhttp3/logging/HttpLoggingInterceptor.kt
@@ -20,17 +20,31 @@ import java.io.IOException
 import java.nio.charset.Charset
 import java.util.TreeSet
 import java.util.concurrent.TimeUnit
+import okhttp3.CompressionInterceptor
+import okhttp3.CompressionInterceptor.DecompressionAlgorithm
+import okhttp3.Gzip
 import okhttp3.Headers
 import okhttp3.HttpUrl
 import okhttp3.Interceptor
+import okhttp3.MediaType
 import okhttp3.OkHttpClient
+import okhttp3.Request
+import okhttp3.RequestBody
 import okhttp3.Response
+import okhttp3.ResponseBody
 import okhttp3.internal.charsetOrUtf8
+import okhttp3.internal.connection.RealCall
 import okhttp3.internal.http.promisesBody
 import okhttp3.internal.platform.Platform
 import okhttp3.logging.internal.isProbablyUtf8
 import okio.Buffer
+import okio.BufferedSink
+import okio.BufferedSource
+import okio.ForwardingSink
+import okio.ForwardingSource
 import okio.GzipSource
+import okio.Source
+import okio.buffer
 
 /**
  * An OkHttp interceptor which logs request and response information. Can be applied as an
@@ -44,13 +58,17 @@ class HttpLoggingInterceptor
   constructor(
     private val logger: Logger = Logger.DEFAULT,
   ) : Interceptor {
-    @Volatile private var headersToRedact = emptySet<String>()
+    @Volatile
+    private var headersToRedact = emptySet<String>()
 
-    @Volatile private var queryParamsNameToRedact = emptySet<String>()
+    @Volatile
+    private var queryParamsNameToRedact = emptySet<String>()
 
     @set:JvmName("level")
     @Volatile
     var level = Level.NONE
+
+    var requestLevel: (Request) -> Level = { level }
 
     enum class Level {
       /** No logs. */
@@ -109,6 +127,27 @@ class HttpLoggingInterceptor
        * ```
        */
       BODY,
+
+      /**
+       * Logs streaming request and response lines.
+       *
+       * Example:
+       * ```
+       * --> POST /greeting http/1.1
+       * <-- 200 OK (22ms, unknown-length body)
+       * > request A
+       *
+       * < response B
+       *
+       * > request C
+       *
+       * < response D
+       *
+       * --> END POST
+       * <-- END HTTP
+       * ```
+       */
+      STREAMING,
     }
 
     fun interface Logger {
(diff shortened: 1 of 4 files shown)


yschimke:
@swankjesse requesting post review. Happy to revert or fix forward with comments.

okhttp-logging-interceptor/src/main/kotlin/okhttp3/logging/HttpLoggingInterceptor.kt
```diff
@@ -52,6 +66,8 @@ class HttpLoggingInterceptor
     @Volatile
     var level = Level.NONE
 
+    var requestLevel: (Request) -> Level = { level }
```
swankjesse, line 69: This isn’t very Java language friendly. I’ve tried to avoid types like `Function1` in our public APIs

swankjesse: (We have overloads in OkHttpClient.Builder, but there’s also a Java way to get at those APIs.)

okhttp-logging-interceptor/src/main/kotlin/okhttp3/logging/HttpLoggingInterceptor.kt
```diff
@@ -163,29 +200,90 @@ class HttpLoggingInterceptor
 
     @Throws(IOException::class)
     override fun intercept(chain: Interceptor.Chain): Response {
-      val level = this.level
+      var request = chain.request()
+
+      val level = this.requestLevel(request)
```
swankjesse, line 205: This is two different features in one PR... configurable levels per-request, and streaming

okhttp-logging-interceptor/src/main/kotlin/okhttp3/logging/HttpLoggingInterceptor.kt
```diff
@@ -163,29 +200,90 @@ class HttpLoggingInterceptor
...
       if (level == Level.NONE) {
         return chain.proceed(request)
       }
 
-      val logBody = level == Level.BODY
-      val logHeaders = logBody || level == Level.HEADERS
+      val logHeaders = level == Level.BODY || level == Level.HEADERS
 
-      val requestBody = request.body
+      val decompressionAlgorithms = findCompressionAlgorithms(chain)
```
swankjesse, line 213: three features in one PR? configurable levels per-request, streaming, and compression?

I’ve been very reluctant to do a full-featured logging interceptor because I think it’s the wrong tool for most jobs. Using Charles or Chucker or Wireshark all seem better suited, because they don’t have to dump everything to the console. Plus they can naturally display stuff like protos and images

okhttp-logging-interceptor/src/main/kotlin/okhttp3/logging/HttpLoggingInterceptor.kt
```diff
@@ -163,29 +200,90 @@ class HttpLoggingInterceptor
...
-        val headers = request.headers
+      if (level != Level.BASIC) {
+        request = logRequest(request, level, decompressionAlgorithms)
+      }
+
+      val startNs = System.nanoTime()
+      val response: Response
+      try {
+        response = chain.proceed(request)
+      } catch (e: Exception) {
```
swankjesse, line 233: I think the better control flow for this is different:

```
var success = false
try {
  ...
  success = true
} finally {
  if (!success) { ... }
}
```

As-is you’re not catching stuff like `AssertionError` which I think is unintentional?

swankjesse: (hang on, this is just moved code. Sorry for me reviewing old code!)

okhttp-logging-interceptor/src/main/kotlin/okhttp3/logging/HttpLoggingInterceptor.kt
```diff
@@ -352,16 +476,107 @@ class HttpLoggingInterceptor
...
+        override fun close() {
+          super.close()
+          logger.log("--> END ${request.method} (streaming body)")
+        }
+      }.buffer()
+
+    body.writeTo(sink)
+  }
+
+  override fun contentLength(): Long = -1
```
swankjesse, line 534: Why isn’t this `body.contentLength()` ?

okhttp-logging-interceptor/src/main/kotlin/okhttp3/logging/HttpLoggingInterceptor.kt
```diff
@@ -352,16 +476,107 @@ class HttpLoggingInterceptor
...
+
+    private fun bodyIsEventStream(response: Response): Boolean {
       val contentType = response.body.contentType()
       return contentType != null && contentType.type == "text" && contentType.subtype == "event-stream"
     }
 
     companion object
   }
+
+private class StreamingRequestBody(
```
swankjesse, line 505: *every* request body is streaming; this is just streaming the body to the logs instead of buffering it all

okhttp-logging-interceptor/src/main/kotlin/okhttp3/logging/HttpLoggingInterceptor.kt
```diff
@@ -163,29 +200,90 @@ class HttpLoggingInterceptor
...
+          contentLength,
+        )
+      } else {
+        response
+      }
+    }
+
+    private fun findCompressionAlgorithms(chain: Interceptor.Chain): List<DecompressionAlgorithm> {
+      val compressionInterceptor =
+        (chain.call() as? RealCall)?.client?.interceptors?.find { it is CompressionInterceptor } as? CompressionInterceptor
```
swankjesse, line 273: I am uncomfortable with this . . . having the logging interceptor take over decompression is super confusing. We can get plaintext logging by applying this one earlier than that one in the pipeline?

okhttp/src/jvmTest/kotlin/okhttp3/DuplexTest.kt
```diff
@@ -69,10 +72,17 @@ class DuplexTest {
 
   private var listener = RecordingEventListener()
   private val handshakeCertificates = platform.localhostHandshakeCertificates()
+  private val loggingInterceptor =
+    HttpLoggingInterceptor({
+      // Not testing output here, but making sure it doesn't fail
+    }).apply {
+      level = Level.NONE
+    }
```
swankjesse, line 80: Because this interceptor now changes where decompression happens, the rest of this test is like, adversely impacted?

swankjesse (commented):
This is too much. Please rollback and do separate PRs for the individual features?

I’ll try to be better and not let the PRs languish unreviewed forever!

PR #9010, opened 2025-08-11
refactor(SSE): move static functions to extensions

MukjepScarlet:
Also added some comments for `TODO` blocks.

for https://github.com/square/retrofit/pull/4462

### DIFF
--- a/okhttp-sse/src/main/kotlin/okhttp3/sse/internal/ServerSentEventReader.kt
+++ b/okhttp-sse/src/main/kotlin/okhttp3/sse/internal/ServerSentEventReader.kt
@@ -22,7 +22,7 @@ import okio.BufferedSource
 import okio.ByteString.Companion.encodeUtf8
 import okio.Options
 
-class ServerSentEventReader(
+internal class ServerSentEventReader(
   private val source: BufferedSource,
   private val callback: Callback,
 ) {
@@ -119,7 +119,7 @@ class ServerSentEventReader(
   }
 
   companion object {
-    val options =
+    private val options =
       Options.of(
         // 0
         "\r\n".encodeUtf8(),


okhttp-sse/src/main/kotlin/okhttp3/sse/EventSources.kt
```diff
@@ -26,29 +27,35 @@ object EventSources {
...
-        connect(callFactory)
-      }
-    }
+    callFactory.asEventSourceFactory()
 
+  @Deprecated(
+    message = "Moved to extension function.",
+    replaceWith =
+      ReplaceWith(
+        expression = "response.toEventSource(listener)",
```
yschimke: This doesn't seem equivalent.

okhttp-sse/src/main/kotlin/okhttp3/sse/EventSource.kt
```diff
@@ -37,5 +40,45 @@ interface EventSource {
...
     ): EventSource
+
+    companion object {
+
+      /**
+       * Wraps a [Call.Factory] into [EventSource.Factory].
+       */
+      @JvmStatic
+      @JvmName("create")
+      fun Call.Factory.asEventSourceFactory(): Factory =
```
yschimke, line 50: Can you add a minimal Java test showing the API of the new functions?

okhttp-sse/src/main/kotlin/okhttp3/sse/EventSource.kt
```diff
@@ -37,5 +40,45 @@ interface EventSource {
...
+    }
+  }
+
+  companion object {
+    /**
+     * Creates a new [EventSource] from the [Call] and immediately enqueue it.
+     */
+    @JvmStatic
+    @JvmName("create")
+    fun Call.toEventSource(listener: EventSourceListener): EventSource {
```
yschimke: Effectively this is new API?

I think it seems reasonable.

MukjepScarlet: Yeah, only this one is new.

okhttp-sse/src/main/kotlin/okhttp3/sse/EventSources.kt
MukjepScarlet: `processResponse` invocation is moved to this constructor (from response)

MukjepScarlet: https://github.com/square/okhttp/pull/9010/files#diff-4db211c7595c58be4a74d26c9dfad74e727deacb936a0f5f22bcc61e5318cb26R38-R39

yschimke: That feels strange. an EventSource is something you can start, and cancel. The Call form of it enqueues the request and returns immediately.

But this one is not a blocking constructor, which seems unexpected. 

Maybe the current `processResponse` is still the right API here?

okhttp-sse/src/main/kotlin/okhttp3/sse/EventSource.kt
MukjepScarlet: I will add it later.

okhttp-sse/src/main/kotlin/okhttp3/sse/EventSources.kt
MukjepScarlet: In my opinion, the `Response` object means we have already started the progress (at least already connected successfully), so just attach a listener(callback) on it.

MukjepScarlet: `Call.toEventSource` feels more natural to me. (enqueue and then use callback), but the Response as event source is already existing in the old version... I can't delete it

yschimke: Isn't it a blocking call? Processing all events on the Response?

okhttp-sse/src/main/kotlin/okhttp3/sse/EventSource.kt
MukjepScarlet: I added a `.java` file including a simple test (maybe it should be in `samples`), and I ran it on my computer.

okhttp-sse/src/main/kotlin/okhttp3/sse/EventSources.kt
MukjepScarlet: Well it's 100% same as the original code, I changed the invocation in new commits

MukjepScarlet: The tests in `EventSourcesHttpTest` still pass. The original code `EventSources.processResponse` is also blocking but it blocks(iterates) on the enqueue thread, in the tests, a response with length-fixed body won't block as it consumes all events.

yschimke: My point is that

```
EventSources.processResponse(response, listener)
```

seems better than

```
response.toEventSource(listener)
```

Because it's not returning something that functions as an EventSource. By the time it's returned, it's completed. the call itself is processing the response.

MukjepScarlet: Okay. My opinion is that using Response as EventSource is a bit strange. If I know in my application that I need an SSE API, I will definitely use asynchronous requests instead of synchronous requests. This function... looks like only to be used for testing or a "full HTTP response" instead of a stream.

MukjepScarlet: I renamed the function in new commit. Now it returns nothing because we can't cancel the EventSource, so returning it makes no sense.

okhttp-sse/src/main/kotlin/okhttp3/sse/EventSource.kt
```diff
@@ -37,5 +40,40 @@ interface EventSource {
...
+    }
+  }
+
+  companion object {
+    /**
+     * Creates a new [EventSource] from the [Call] and immediately enqueue it.
+     */
+    @JvmStatic
+    @JvmName("create")
+    fun Call.toEventSource(listener: EventSourceListener): EventSource = RealEventSource(this, listener).also(this::enqueue)
```
JakeWharton: I don't like this name, as it hides the fact that it fires off the request using enqueue. I would want some kind of verb that coveys this in the function name

MukjepScarlet: How about:
J: EventSource.send(call)
K: Call.sendAsEventSource() (Or sendSse?)

okhttp-sse/src/main/kotlin/okhttp3/sse/EventSources.kt
yschimke: I think I prefer the original `processResponse`. I'm not sure this is worth the deprecation churn.

okhttp-sse/src/main/kotlin/okhttp3/sse/EventSource.kt
yschimke: Our existing naming would be `enqueueEventSource()`.

okhttp-sse/src/main/kotlin/okhttp3/sse/EventSources.kt
MukjepScarlet: How about `EventSource.process(Response)`(for both Kotlin and Java) ?

okhttp-sse/src/main/kotlin/okhttp3/sse/EventSource.kt
MukjepScarlet: How about the Java-side? `enqueue` looks not suitable. (The subject is EventSource instead of the Call)

okhttp-sse/src/main/kotlin/okhttp3/sse/EventSources.kt
yschimke: I guess `EventSource.process(Response, EventSourceListener)`?

I'm just not sure this is worth the pain to deprecate and rename public API. It sounds like you don't actually expect people to be calling it.

okhttp-sse/src/main/kotlin/okhttp3/sse/EventSource.kt
MukjepScarlet: Now it's EventSource.enqueue(call, listener)

yschimke:
@swankjesse requesting post API review. Seems like a good cleanup and more idiomatic Kotlin.

swankjesse:
Reverted here. https://github.com/square/okhttp/pull/9098

Discussion about why to revert on that PR.

PR #9013, opened 2025-08-12
[Not to land] Testing HttpEngine with Call Decorator

yschimke:
- [ ] Clone https://github.com/google/cronet-transport-for-okhttp/ for testing
- [ ] Update to Kotlin
- [ ] Add a test
- [ ] Migrate to Call.Decorator
- [ ] Fix cancellation cleanly


okhttp/src/commonJvmAndroid/kotlin/okhttp3/Call.kt
```diff
@@ -96,4 +96,33 @@ interface Call : Cloneable {
...
+   * - Return some [Call] implementation that will immediately fail avoiding network calls based on network or
+   *   authentication state.
+   * - Redirect the [Call], such as using an alternative [Call.Factory].
+   * - Defer execution, something not safe in an Interceptor.
+   *
+   * It should not throw an exception, instead it should return a Call that will fail on [Call.execute].
+   *
+   * A Decorator that changes the OkHttpClient should typically retain later decorators in the new client.
+   */
+  fun interface Decorator {
```
edechamps-Google, line 118: Is adding a brand new public abstraction just for this really worth it? It seems more straightforward to add cancellation support to interceptors (#7164), and treat the HttpEngine interceptor as a "special snowflake" in RealCall. Arguably this wouldn't be as clean as RealCall would need to be aware of HttpEngineInterceptor's existence, but it's not like your approach is squeaky clean either, given you have to refer to OkHttp internals in HttpEngineDecorator (e.g. manually calling `BridgeInterceptor` which is an internal class). Since neither approach achieves perfect decoupling between the HttpEngine bridge and OkHttp internals, I'd be inclined to go for the most straightforward one.

okhttp/src/androidMain/kotlin/okhttp3/android/httpengine/HttpEngineCallDecorator.kt
```diff
@@ -0,0 +1,87 @@
...
+import okhttp3.OkHttpClient
+import okhttp3.Request
+import okhttp3.internal.SuppressSignatureCheck
+import okhttp3.internal.cache.CacheInterceptor
+import okhttp3.internal.http.BridgeInterceptor
+import okhttp3.internal.http.RetryAndFollowUpInterceptor
+
+@SuppressSignatureCheck
+class HttpEngineCallDecorator(
+  internal val httpEngine: HttpEngine,
```
edechamps-Google, line 19: It is debatable whether this should use HttpEngine directly, or if it should use the Cronet API wrapper (which would then call HttpEngine). The Cronet API wrapper is more flexible and would make it possible to use other variants of Cronet (such as embedded of Play Services) if we ever want to, but the downside is it would involve OkHttp taking a new dependency on the (very small) cronet-api Maven package.

edechamps-Google, line 19: It is debatable whether we want to let the user choose which HttpEngine to use. Arguably this code should instantiate HttpEngine and should configure it by interpreting OkHttpClient settings. Otherwise we may end up backing ourselves into a corner by providing too much flexibility to the user.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Call.kt
yschimke: It exists for other reasons, and is just the best way to express this problem. Call.Factory could be similar, but isn't well adopted. I fear we lost that argument.

https://github.com/square/okhttp/pull/8376
https://github.com/square/okhttp/pull/8800

> manually calling BridgeInterceptor which is an internal class

Yep, we should fix this, and have a clean set of APIs for bridging interceptors like Cronet

> I'd be inclined to go for the most straightforward one.

I don't think we would do that if it meant treating HttpEngine as a special snowflake in RealCall.

okhttp/src/androidMain/kotlin/okhttp3/android/httpengine/HttpEngineCallDecorator.kt
yschimke: I think that dependency **is** the blocker here.

But also, a first version of this should probably be a v2.0 of the [google/cronet-transport-for-okhttp](https://github.com/google/cronet-transport-for-okhttp)

So that can do whatever makes sense?

yschimke: Yep, I'm not aware of what we expect an app developer to configure in typical cases?

Do we want them to allowlist their known hosts for HTTP/3?

But no objection from me. Your call.

PR #9019, opened 2025-08-15
Fix range and format in KDoc of `Response.isSuccessful`

dewantawsif:


okhttp/src/commonJvmAndroid/kotlin/okhttp3/Response.kt
```diff
@@ -143,7 +143,7 @@ class Response internal constructor(
   fun code(): Int = code
 
   /**
-   * Returns true if the code is in [200..300), which means the request was successfully received,
```
yschimke, line 146: It's meant to show inclusive and exclusive range boundaries. 

https://en.wikipedia.org/wiki/Interval_(mathematics)#Including_or_excluding_endpoints

dewantawsif: Ah, it didn't click when I took a quick look

PR #9023, opened 2025-08-16
Revert "Add QUERY method support"

yschimke:
Reverts square/okhttp#8550

> Shouldn't this include the request body as part of the cache key somewhere?


yschimke:
I'm holding off landing this if we can get a clean fix in place soon.

Will be simpler to review the fix

PR #9027, opened 2025-08-19
Add QUERY caching support

desiderantes:
Continuing the work from #8550, this PR adds logic for cache support on QUERY requests. I am a bit unsure on visibility for the new key method vs the old one, so I've left it unsolved. Also, on whether the cost of hashing the body would make it worth it (I guess some sort of warning should be added)


okhttp/src/commonJvmAndroid/kotlin/okhttp3/Cache.kt
```diff
@@ -752,6 +784,18 @@ class Cache internal constructor(
         .md5()
         .hex()
 
+    internal fun key(request: Request): String =
+      if (
```
yschimke, line 777: I think this logic needs to consider whether a `Request.cacheUrlOverride` is set. 

See https://github.com/square/okhttp/blob/363d3e788f497dbee0a2bb9762e69f85c4f79d16/okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/cache/CacheInterceptor.kt#L299

Perhaps that logic should move into here?

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Cache.kt
```diff
@@ -752,6 +784,18 @@ class Cache internal constructor(
...
+        request.method == "QUERY" &&
+        request.body != null &&
+        !request.body.isOneShot()
+      ) {
+        key(request.url) + "_" + Buffer().also { request.body.writeTo(it) }.md5().hex()
+      } else {
+        // Reading this kind of body would consume it, so we can't cache it.
```
yschimke: I don't think this is correct, I suspect that it's caching more generally?

Can you add a test for this?

okhttp/src/jvmTest/kotlin/okhttp3/CacheTest.kt
```diff
@@ -577,6 +577,54 @@ class CacheTest {
     assertThat(recordedRequest4.requestLine).isEqualTo("QUERY /bar HTTP/1.1")
   }
 
+  @Test
+  fun queryRequestsCacheTheBody() {
```
yschimke, line 550: Can you add separate tests covering `cacheUrlOverride` and also a one shot request body.

yschimke:
`./gradlew :spotlessApply` to cleanup formatting issues

yschimke:
Failing on existing tests

```
DnsOverHttpsTest > usesCacheOnlyIfFresh() FAILED
    java.net.UnknownHostException: google.com
        at okhttp3.dnsoverhttps.DnsOverHttps.throwBestFailure(DnsOverHttps.kt:163)
        at okhttp3.dnsoverhttps.DnsOverHttps.lookupHttps(DnsOverHttps.kt:87)
        at okhttp3.dnsoverhttps.DnsOverHttps.lookup(DnsOverHttps.kt:68)
        at okhttp3.dnsoverhttps.DnsOverHttpsTest.usesCacheOnlyIfFresh(DnsOverHttpsTest.kt:297)

        Caused by:
        java.io.IOException: canceled due to java.lang.IllegalArgumentException: method GET must not have a request body.
            at okhttp3.internal.connection.RealCall$AsyncCall.run(RealCall.kt:542)
            at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
            at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
            at java.base/java.lang.Thread.run(Thread.java:1583)
```

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Cache.kt
desiderantes: Can you explain what you mean? AFAIU oneshot request bodiess are for things like a streamable body that would not be able to be consumed twice. So we either duplicate the body (costly) or avoid caching them. Please correct me if I misunderstood.

yschimke: I mean you aren't stopping caching. You are just using a key without the body in it.

desiderantes: Yeah, that's true, I don't know what I was thinking 🤔

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Cache.kt
desiderantes: I added the cacheUrlOverride bit for QUERY in that method

yschimke:
@martinbonnin Do your worst!

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Cache.kt
```diff
@@ -191,7 +192,12 @@ class Cache internal constructor(
     get() = cache.isClosed()
 
   internal fun get(request: Request): Response? {
-    val key = key(request.url)
+    if (request.body != null && request.body.isOneShot() && request.method == "QUERY") {
+      // Don't cache one-shot QUERY requests, since we need to cache by using the body,
+      // and we can't consume the body twice
```
martinbonnin: I would probably expect this to throw instead of returning `null`, both when writing and reading from the cache. Might be worth moving this check to the `key()` function?

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Cache.kt
```diff
@@ -752,6 +790,16 @@ class Cache internal constructor(
         .md5()
         .hex()
 
+    internal fun key(request: Request): String =
+      if (
+        request.method == "QUERY" &&
+        request.body != null
+      ) {
+        key(request.url) + "_" + Buffer().also { request.body.writeTo(it) }.md5().hex()
```
martinbonnin: Several comments here:

1. Can we stream the body instead of buffering it in memory entirely? 
2. Slightly concerned that this is more doing I/O than was the case before but this code is followed by reading/writing from the cache, which is also doing I/O so probably OK?
3. Does the key size matter? I was thinking we could hash the url + body all at once to get a single md5: `md5(url + '\0' + body)` or something like so? This way we have all uniform md5 keys. One drawback of course is that we can't tell from looking at a given key whether it is a `QUERY` key or not. Maybe it's an useful property, maybe not, not sure.
4. ~~Actually, it should probably be more like `md5(method + '\0' + url + '\0' + body)` as we don't want a `QUERY` with an empty body to clash with a `GET` of the same url?~~ Well, nevermind, we can't change the existing keys for `GET` so keep using the previous one for GET and `md5(url + '\0' + body)` for QUERY should be good. The `\0` will ensure that there is no clash.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Cache.kt
```diff
@@ -609,6 +633,20 @@ class Cache internal constructor(
             .writeByte('\n'.code)
         }
 
+        if (requestMethod == "QUERY") {
+          // Write the body of a QUERY request.
```
martinbonnin: Do we need to store the request body in cache? 

Looking at the code, it looks like storing a request in the cache is already destructive as only the `vary` headers of the request are stored (not all of them). 

I think I wouldn't mind too much if `Cache.Entry.body` was missing and saving some bytes there.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Cache.kt
yschimke: I think returning null is more appropriate given below

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Cache.kt
```diff
@@ -191,7 +192,12 @@ class Cache internal constructor(
     get() = cache.isClosed()
 
   internal fun get(request: Request): Response? {
-    val key = key(request.url)
+    if (request.body != null && request.body.isOneShot() && request.method == "QUERY") {
```
yschimke: Is this enough?

```
if (request.body != null && request.body.isOneShot()) {
```

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/cache/CacheInterceptor.kt
```diff
@@ -299,12 +299,24 @@ class CacheInterceptor(
...
+      "GET", "POST" -> {
+        newBuilder()
+          .get()
+          .url(cacheUrlOverride)
+          .cacheUrlOverride(null)
+          .build()
+      }
+      "QUERY" -> {
+        newBuilder()
+          .query(body!!)
```
yschimke: should the body be stripped here also?

Or rather should QUERY be added to the case above?

yschimke: I think specifically, once a cacheUrlOverride is set, it should be treated as a GET with that override url.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Cache.kt
```diff
@@ -543,6 +550,22 @@ class Cache internal constructor(
         }
         varyHeaders = varyHeadersBuilder.build()
 
+        val bodyLength = source.readDecimalLong()
```
yschimke: Will this corrupt existing caches, changing the expected format even for existing GET entries?

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Cache.kt
yschimke: Storing a hash instead would seem a lot less disruptive to the format. Perhaps it could be stored as a synthetic header?

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Cache.kt
martinbonnin: I don't know... As a user, sounds like it's something I'd rather be aware of, in the spirit of "fail fast"? I can always disable caching for those requests if needed with `no-store` or `no-cache`.

`isSingleShot()` isn't necessarily something that users are super aware of? Or are they?

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Cache.kt
martinbonnin: Do we even need to store anything? What's the use case for reading that request body from the cache?

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Cache.kt
yschimke: You create a shared client. You send a bunch of requests through it. Some might be one shot. Some not.

If you care enough add an event listener and read the cache events

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Cache.kt
yschimke: True, it just needs to be part of the key. I think.

yschimke: If we can avoid changing the format, that would be good.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Cache.kt
martinbonnin: Works for me. Probably worth a mention in `Request.Builder.query()` along the lines of "if `body.isSingleShot()` returns true, this request won't be cached"

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Cache.kt
yschimke: I've removed this part

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/cache/CacheInterceptor.kt
yschimke: Simplified

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Cache.kt
yschimke: Reverted

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Cache.kt
yschimke: Done

yschimke:
@swankjesse can you take a look. We should land this before a new release.

martinbonnin (approved):

yschimke:

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Request.kt
```diff
@@ -317,6 +317,9 @@ class Request internal constructor(
 
     open fun patch(body: RequestBody): Builder = method("PATCH", body)
 
+    /**
+     * A QUERY request with a body. If `body.isOneShot()` is true, then caching will be disabled.
```
swankjesse, line 321: Could we just require a non-null `cacheUrlOverride` for these?

swankjesse:
I’m reverting this here: https://github.com/square/okhttp/pull/9097

Motivation & discussion on that PR.

yschimke:
Follow up in https://github.com/square/okhttp/pull/9101/files

It seems simpler and cleaner, avoids a bunch of things we debated about.

PR #9028, opened 2025-08-19
Add test demonstrating multi process

yschimke:
Not running automatically, but will fail without the explicit init.

```
class TestApplication : Application() {
  override fun onCreate() {
    super.onCreate()

    if (isSecondaryProcess()) {
      OkHttp.initialize(applicationContext)
    }
  }
```

### DIFF
--- a/android-test-app/src/main/AndroidManifest.xml
+++ b/android-test-app/src/main/AndroidManifest.xml
@@ -18,6 +18,17 @@
         <category android:name="android.intent.category.LAUNCHER" />
       </intent-filter>
     </activity>
+
+    <activity
+      android:name=".MainActivity2"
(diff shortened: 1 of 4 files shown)


yschimke:
@swankjesse requesting post review, test only.

kusius:
Is the `initialize` method available in 5.1.0 ?

yschimke:
@kusius sorry, no.

You can call the initializer directly for now

https://github.com/square/okhttp/blob/0e0352920baadd4e6473cfffa32706c765bde38a/okhttp/src/androidMain/kotlin/okhttp3/internal/platform/PlatformInitializer.kt#L26

swankjesse (commented):
LGTM!

PR #9030, opened 2025-08-21
Print URLs in HttpLoggingInterceptor when the call fails

oldergod:
When there's lots of messages, it is difficult to map the failure messages to the start messages without the url.

### DIFF
--- a/okhttp-logging-interceptor/src/main/kotlin/okhttp3/logging/HttpLoggingInterceptor.kt
+++ b/okhttp-logging-interceptor/src/main/kotlin/okhttp3/logging/HttpLoggingInterceptor.kt
@@ -247,7 +247,13 @@ class HttpLoggingInterceptor
       try {
         response = chain.proceed(request)
       } catch (e: Exception) {
-        logger.log("<-- HTTP FAILED: $e")
+        val tookMs = TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - startNs)
+        logger.log(
+          buildString {
+            append("<-- HTTP FAILED: $e.")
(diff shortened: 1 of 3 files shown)


okhttp-logging-interceptor/src/main/kotlin/okhttp3/logging/HttpLoggingInterceptor.kt
```diff
@@ -247,7 +247,12 @@ class HttpLoggingInterceptor
       try {
         response = chain.proceed(request)
       } catch (e: Exception) {
-        logger.log("<-- HTTP FAILED: $e")
+        logger.log(
+          buildString {
+            append("<-- HTTP FAILED: $e.")
+            append(" ${redactUrl(request.url)}")
```
oldergod: On success, we print `response.request.url`. Does it mean that just `request.url` might be wrong?

swankjesse: If there was a redirect, `response.request.url` has the new URL

swankjesse: Wanna include timing too?

```
append(" ${redactUrl(response.request.url)} (${tookMs}ms")
```

swankjesse (approved):

okhttp-logging-interceptor/src/main/kotlin/okhttp3/logging/HttpLoggingInterceptor.kt
oldergod: added the timing! and fixed Android tests

PR #9031, opened 2025-08-21
Move the Android test README under the androidTest directory

oldergod:
And opt-in to compile without errors


okhttp-brotli/src/test/java/okhttp3/brotli/BrotliInterceptorTest.kt
```diff
@@ -13,6 +13,9 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+@file:Suppress(
+  "INVISIBLE_REFERENCE",
+)
```
oldergod, line 18: This creates a warning
image

And the IDE is suggesting that I add `ERROR_SUPPRESSION` but it doesn't make the warning go away

android-test/src/androidTest/README.md
oldergod, line 1: I got confused because the README was at the root for the module but tests under `src/test` won't run by looking at the README.

okhttp-brotli/src/test/java/okhttp3/brotli/BrotliInterceptorTest.kt
yschimke: @swankjesse found this, it's an IDE only bug with handling of Kotlin friends. So unavoidable?

yschimke (approved):

okhttp-brotli/src/test/java/okhttp3/brotli/BrotliInterceptorTest.kt
swankjesse: It doesn’t bother me.

PR #8493, opened 2024-07-27
Upload progress recipe

seokjun7410:
Dear. @swankjesse

I improved the `upload progress recipe` discussed in #1471 #1528
It was rewritten based on the discussed requirements.

After discussion, I would like to write a Kotlin recipe
I'm awaiting your comments. 
Thank you for your time

PR #9033, opened 2025-08-23
add Kotlin version and update UploadProgress recipe

seokjun7410:
Upload Progress Recipe (#8493)
This is follow-up work. 

1. Created a Kotlin version of the UploadProgress Recipe.
2. Adjusted the indentation of Recipe #8493.
3. Added java and kt to the recipe page.

Example image below
스크린샷 2025-08-26 오전 11 39 58

### DIFF
--- a/docs/recipes.md
+++ b/docs/recipes.md
@@ -947,6 +947,210 @@ Use `Response.challenges()` to get the schemes and realms of any authentication 
       }
     ```
 
+### Upload Progress ([.kt][UploadProgressKotlin], [.java][UploadProgressJava])
+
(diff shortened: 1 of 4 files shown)


yschimke (approved):

yschimke:
```
> There were 1 lint error(s), they must be fixed or suppressed.
  samples/guide/src/main/java/okhttp3/recipes/kt/UploadPorgress.kt:L1 ktlint(standard:filename) File 'UploadPorgress.kt' contains a single class and possibly also extension functions for that class and should be named same after that class 'UploadProgress.kt'
  Resolve these lints or suppress with `suppressLintsFor`
```

Issue #9032, opened 2025-08-21
Incompatible Kotlin Version Error with OkHttp v5.1.0


satyakaikala:
I'm encountering an error when using OkHttp v5.1.0 with Kotlin v2.0.21. The error message is as follows:
"Module was compiled with an incompatible version of Kotlin. The binary version of its metadata is 2.2.0, expected version is 2.0.0."

Environment
- OkHttp version: v5.1.0
- Kotlin version: 2.0.21
- Kotlin Compiler options: `kotlin.compilerOptions.languageVersion = 1.9`, `kotlin.compilerOptions.apiVersion = 1.9`

Steps to Reproduce
1. Set up a Kotlin project with the above compiler options.
2. Add OkHttp v5.1.0 as a dependency.
3. Attempt to compile the project.

Additional Information
The issue does not occur with OkHttp v4.12.0.

JakeWharton:
The use of Kotlin 2.2 to compile is noted in the change log: https://github.com/square/okhttp/blob/master/CHANGELOG.md#version-500

That being said, we could target a slightly older version, but I'd be hesitant to provide strong guarantees here.

yschimke:
Would we need to coordinate with okio also?

It is probably a good idea to define the policy even if it's latest on the day it's released.

satyakaikala:
@JakeWharton, I appreciate the reference to the change log. I was wondering if there was a specific reason why certain compiler options, such as languageVersion and apiVersion, were not set to ensure compatibility with earlier versions, like 2.0.

I would like to suggest setting specific compiler options to facilitate smoother adoption across the industry. Given that the adoption of new Kotlin versions can be slower in some projects, it would be beneficial to consider the following recommendations:

1. **Set Minimum Compiler Options:**
   **languageVersion**: Consider setting this to a minimum of 2.0 to ensure compatibility with projects that are not yet on the latest version. 
  **apiVersion**: Align this with the _languageVersion_ to maintain consistency.
2. **Support N-2 Releases**: By supporting N-2 releases, OkHttp can provide a more gradual upgrade path for developers, allowing them to adopt new versions at a manageable pace.

JakeWharton:
It's simply never been a problem.

We're aware of how it works and what it changes, thanks. It's not quite that simple, though, because we also have dependencies.

hugoncosta:
[...]

PR #9040, opened 2025-08-28
Set apiVersion and languageVersion to 2.0 to ensure backwards compatibility

hugoncosta:
Address #9032 by setting apiVersion and languageVersion to 2.0. This ensures that the code can be compiled with Kotlin 2.2.0+ but users as low as 1.9.0 are able to run this without being forced to update. For reference, 2.2.0 supports compilation as low as 1.6.0.

Compiled locally, seems to work fine, but my laptop is not setup to compile the Android modules btw


hugoncosta:
Sorry about that - not really used to these workflows + I don't have an Android SDK on my machine. I think this should be it 🤞

hugoncosta:
I think I have found an easier path, and especially one that doesn't require us to change much - the source code is already 2.0 compatible, only the tests aren't, and they don't need to be, so why bother? The change is now focusing only on "main" source sets, leaving the tests to use any 2.0+ feature.

build.gradle.kts
```diff
@@ -195,6 +195,19 @@ subprojects {
...
+    kotlinExtension.sourceSets {
+      // Exclude all test source sets (containing Test)
+      matching { !it.name.contains("Test", ignoreCase = true) }.all {
+        languageSettings {
+          apiVersion = "2.0"
+          languageVersion = "2.0"
+        }
+      }
+    }
+  }
```
Goooler:
```suggestion
  tasks.withType<KotlinCompile>().configureEach {
    if (name.contains("test")) return@configureEach // Skip test source sets.
  
    compilerOptions {
      // Pin language level to 2.0 to ensure compatibility with older Gradle versions and other libraries that depend on okhttp.
      // https://docs.gradle.org/current/userguide/compatibility.html#kotlin
      languageVersion = org.jetbrains.kotlin.gradle.dsl.KotlinVersion.KOTLIN_2_0
      apiVersion = org.jetbrains.kotlin.gradle.dsl.KotlinVersion.KOTLIN_2_0
    }
  }
```

yschimke:
I'll wait for https://github.com/square/okio/pull/1690 to land before this

yschimke:
The Okio PR sounds promising, so I'll land this assuming it will also.  Thanks for the PR.

Are you ok with the suggested changes for matching tasks I accepted? They are also on the Okio PR.

yschimke (approved):

yschimke:
Shit, I think that suggestion broke it. 

Not sure if we need to follow this https://kotlinlang.org/docs/whatsnew21.html#new-api-for-kotlin-gradle-plugin-extensions

Goooler:
Yeah, we can try out something like 

```kt
extensions.configure<KotlinBaseExtension> {
  (this as HasConfigurableKotlinCompilerOptions<*>?)?.let {
    compilerOptions {
      apiVersion = KotlinVersion.KOTLIN_2_0
      languageVersion = KotlinVersion.KOTLIN_2_0
    }
  }
}
```

yschimke:
Applying 2.1.21 in https://github.com/square/okhttp/pull/9322

PR #9088, opened 2025-09-27
Demonstrate read timeout with event source

yschimke:


swankjesse (approved):
Nice test.

I don’t think this is the right behavior, do you?

JakeWharton:
I could see it applying during the content read (i.e., between double newlines), but not between events.

PR #9092, opened 2025-09-29
update commons-lang3 - 3.8.1 to 3.18.0

stecurran-est-tech:
Ensure Commons Lang 3-3.18.0 is used for all affected dependencies. Resolves CVE-2025-48924.
Build successful after update.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -50,6 +50,18 @@ configure<SpotlessExtension> {
 }
 
 allprojects {
+
+  configurations.all {
+    resolutionStrategy.eachDependency {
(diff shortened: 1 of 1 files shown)


yschimke:
Checkstyle and dokka shouldn't pull in commons-lang as a dependency.

So is this a false positive? Is it just a build time CVE? What's the threat?

Not sure we want to pander to broken CVE tools.

JakeWharton:
Even if they do, this isn't an attack surface or a failure mode we care about.

stecurran-est:
commons-lang3 is only a build-time dependency (Checkstyle/Doxia). Not included in runtime artifacts, so CVE alerts are false positives
@yschimke @JakeWharton Thanks for feedback.

PR #9097, opened 2025-10-02
Revert "Add QUERY caching support (#9027)"

swankjesse:
This reverts commit bc269e3d5a45f17c70c1fd3f0276644b075c5754.


swankjesse:
I don’t think this is the implementation we want for this feature. We’ve already got `cacheUrlOverride`, and I think that should be what we use for these. 

In particular we should not cache `QUERY` requests with a body if `cacheUrlOverride` is absent, and we should cache them if it is present.

yschimke:
Should we add a form of .query() taking a cache URL override?

swankjesse:
> Should we add a form of .query() taking a cache URL override?

I don’t think it’s going to be that common that we’ll need a special function.

I do think it’d be nice to have a quick sample / doc on how to create a good cache URL for Query, probably using the hashing stuff I reverted! That hashing logic was right, I just wanted to punt that work to the user.

How would you feel about say, a function like `RequestBody.sha256()`

yschimke:
Fine by me.

But I'd defer to @martinbonnin for real world usage

martinbonnin:
The caching is the main reason to use `QUERY` for me. If requests are not cached by default, I think a `query(cacheUrlOverride)` would be a nice hint that additional user action is required to enable it. That being said, it's very easy to add the `query(cacheUrlOverride)` down the road so might as well see what the reception is and decide later?

Same thing for `RequestBody.sha256()`, this can always be added down the road. ~~Let's start with some KDoc? I'll open a PR in a bit.~~

KDoc suggestion [here](https://github.com/square/okhttp/pull/9107), let me know what you think.

PR #9098, opened 2025-10-02
Revert "refactor(SSE): move static functions to extensions (#9010)"

swankjesse:
This reverts commit 09fb9a448acf8b89dcffd1c36a1f012d258e7990.


swankjesse:
Idiomatic Kotlin is fine. I like idiomatic Kotlin.

But I do not want to deprecate working code. Every time we deprecate an API we’re creating work for other people to do, and I don’t want to give our users more work to do.

Our existing APIs support Java language callers, and we care way more about Java language callers than most Kotlin language libraries.

If we want to introduce _additional_ APIs (and not _replacement_ APIs), we can do that. But they should probably call through the Java-friendly APIs.

JakeWharton (approved):

PR #9100, opened 2025-10-03
Avoid NPEs in JavaNetAuthenticator

DanielThomas:
Configuring `JavaNetAuthenticator` as the `proxyAuthenticator` when using the default ProxySelector causes NPEs to be thrown because the socket addresses returned are unresolved:

```
Exception in thread "main" java.lang.NullPointerException: address() as InetSocketAddress).address must not be null
	at okhttp3.internal.authenticator.JavaNetAuthenticator.connectToInetAddress(JavaNetAuthenticator.kt:91)
	at okhttp3.internal.authenticator.JavaNetAuthenticator.authenticate(JavaNetAuthenticator.kt:53)
	at okhttp3.internal.http.RetryAndFollowUpInterceptor.followUpRequest(RetryAndFollowUpInterceptor.kt:220)
	at okhttp3.internal.http.RetryAndFollowUpInterceptor.intercept(RetryAndFollowUpInterceptor.kt:108)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
	at okhttp3.internal.connection.RealCall.getResponseWithInterceptorChain$okhttp(RealCall.kt:201)
	at okhttp3.internal.connection.RealCall.execute(RealCall.kt:154)
```

The implementation could also NPE on a 407 from an non-proxy host, so I've handled this too.


okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/authenticator/JavaNetAuthenticator.kt
```diff
@@ -53,17 +53,21 @@ class JavaNetAuthenticator(
...
-            proxyAddress.hostName,
-            proxy.connectToInetAddress(url, dns),
-            proxyAddress.port,
-            url.scheme,
-            challenge.realm,
-            challenge.scheme,
-            url.toUrl(),
-            Authenticator.RequestorType.PROXY,
-          )
+          if (proxy.type() == Proxy.Type.HTTP) {
```
yschimke, line 56: Is this part of the same change? Is it a big fix when code == 407, but not a proxy request?

yschimke: > The implementation could also NPE on a 407 from an non-proxy host, so I've handled this too.

missed that.

yschimke: I'm still not sure about this code path. Can it only be from a HTTP Proxy?

Is this cleaner if you inline this logic into 

```
    val proxy = route?.proxy ?: Proxy.NO_PROXY
    val proxyAuthorization = response.code == 407 && proxy.type() != Proxy.Type.DIRECT
```

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/authenticator/JavaNetAuthenticator.kt
```diff
@@ -102,6 +106,13 @@ class JavaNetAuthenticator(
   ): InetAddress =
     when (type()) {
       Proxy.Type.DIRECT -> dns.lookup(url.host).first()
-      else -> (address() as InetSocketAddress).address
+      else -> {
+        val socketAddress = address() as InetSocketAddress
+        if (socketAddress.isUnresolved) {
+          dns.lookup(socketAddress.hostString).first()
```
swankjesse, line 112: Why is this resolving the DNS address? I don’t think we guarantee that we’ll use this same resolved address later when we connect

swankjesse: Ahhh, I see from your stack trace. Hmmmmm.....

swankjesse: Let’s just change this function to return `InetAddress?` ?

swankjesse: Yeah, that’s what this should do. `Authenticator.requestPasswordAuthentication` is even documented as such.

> addr – The InetAddress of the site requesting authorization, or null if not known.

swankjesse (changes requested):

PR #9101, opened 2025-10-03
Enable QUERY caching via cacheUrlOverride

yschimke:
Reuse tests from https://github.com/square/okhttp/pull/9027

But with minimal support requiring a specific override.

### DIFF
--- a/okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/cache/CacheInterceptor.kt
+++ b/okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/cache/CacheInterceptor.kt
@@ -299,7 +299,8 @@ class CacheInterceptor(
 private fun Request.requestForCache(): Request {
   val cacheUrlOverride = cacheUrlOverride
 
-  return if (cacheUrlOverride != null && (method == "GET" || method == "POST")) {
+  // Allow POST caching only when there is a cacheUrlOverride
+  return if (cacheUrlOverride != null && (HttpMethod.isCacheable(method) || method == "POST")) {
     newBuilder()
       .get()
       .url(cacheUrlOverride)
--- a/okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/cache/CacheInterceptor.kt
+++ b/okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/cache/CacheInterceptor.kt
@@ -299,7 +299,7 @@ class CacheInterceptor(
 private fun Request.requestForCache(): Request {
   val cacheUrlOverride = cacheUrlOverride
 
-  // Allow POST caching only when there is a cacheUrlOverride
+  // Allow POST and QUERY caching only when there is a cacheUrlOverride
   return if (cacheUrlOverride != null && (HttpMethod.isCacheable(method) || method == "POST")) {
     newBuilder()
       .get()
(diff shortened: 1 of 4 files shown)


yschimke:
cc @desiderantes @martinbonnin

martinbonnin:
Out of curiosity, is there a `cacheMethodOverride` or some other way that would allow me to cache `POST` requests that I know behave like `QUERY` requests but haven't migrated yet?

yschimke:
@martinbonnin yes, that's why this was added.

Specifically https://github.com/square/okhttp/issues/8113 led to https://github.com/square/okhttp/pull/8234

martinbonnin:
Perfect, even simpler 👍

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/cache/CacheInterceptor.kt
```diff
@@ -299,7 +299,8 @@ class CacheInterceptor(
 private fun Request.requestForCache(): Request {
   val cacheUrlOverride = cacheUrlOverride
 
-  return if (cacheUrlOverride != null && (method == "GET" || method == "POST")) {
+  // Allow POST and QUERY caching only when there is a cacheUrlOverride
```
swankjesse, line 302: This is perfect

okhttp/src/jvmTest/kotlin/okhttp3/CacheTest.kt
```diff
@@ -502,23 +505,175 @@ class CacheTest {
...
+        .Builder()
+        .url(server.url("/foo"))
+        .get()
+        .build()
+    val response1 = client.newCall(requestGet1).execute()
+    assertThat(response1.body.string()).isEqualTo("ABC")
+    val recordedRequest1 = server.takeRequest()
+    assertThat(recordedRequest1.requestLine).isEqualTo("GET /foo HTTP/1.1")
+
+    val requestQuery1 =
```
swankjesse, line 537: Without `.cacheUrlOverride` this isn’t actually testing that they’re cached independently?

okhttp/src/jvmTest/kotlin/okhttp3/CacheTest.kt
```diff
@@ -502,23 +505,175 @@ class CacheTest {
...
+    assertThat(recordedRequest2.requestLine).isEqualTo("QUERY /foo HTTP/1.1")
   }
 
   @Test
-  fun queryWithoutBodyNotCached() {
-    assertNotCached(
-      MockResponse.Builder().build(),
-      method = "QUERY",
-      requestBody = null,
+  fun queryRequestsCacheTheBodyWithCacheUrlOverride() {
```
swankjesse, line 550: Great test

okhttp/src/jvmTest/kotlin/okhttp3/CacheTest.kt
```diff
@@ -502,23 +505,175 @@ class CacheTest {
...
+    val response1a = client.newCall(request1).execute()
+    assertThat(response1a.body.string()).isEqualTo("ABC")
+
+    // Fifth QUERY request with body "bar" again, should be cached and return "DEF"
+    val response2a = client.newCall(request2).execute()
+    assertThat(response2a.body.string()).isEqualTo("DEF")
   }
 
+  @Test
+  fun oneshotBodyIsNotCachedForQueryRequest() {
```
swankjesse, line 617: If we added `cacheUrlOverride` would this test keep passing?

yschimke: I think the test is now invalid. I'll remove

okhttp/src/jvmTest/kotlin/okhttp3/CacheTest.kt
yschimke: Let me check. I defaulted to reusing existing tests and fixing. But I think I needed more critical review

PR #9106, opened 2025-10-03
Introduce events for the dispatcher queue

squarejesse:
So event listeners can see how long calls are waiting for resources to become available.

This is the API only. I'll do the implementation in follow-up.

Also note that I'm going to try my best to implement this event pair to only fire for calls that actually spend time in the queue. Calls that are immediately promoted out of the queue should not get events.

### DIFF
--- a/okhttp/src/commonJvmAndroid/kotlin/okhttp3/EventListener.kt
+++ b/okhttp/src/commonJvmAndroid/kotlin/okhttp3/EventListener.kt
@@ -33,6 +33,7 @@ import java.net.Proxy
  * Events are typically nested with this structure:
  *
  *  * call ([callStart], [callEnd], [callFailed])
+ *    * dispatcher queue ([dispatcherQueueStart], [dispatcherQueueEnd])
  *    * proxy selection ([proxySelectStart], [proxySelectEnd])
  *    * dns ([dnsStart], [dnsEnd])
  *    * connect ([connectStart], [connectEnd], [connectFailed])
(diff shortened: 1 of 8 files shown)


okhttp/src/commonJvmAndroid/kotlin/okhttp3/EventListener.kt
```diff
@@ -70,6 +71,30 @@ abstract class EventListener {
...
   }
 
+  /**
+   * Invoked for calls that were not executed immediately because resources weren't available. The
+   * call will remain in the queue until resources are available.
+   *
+   * Use [Dispatcher.maxRequests] and [Dispatcher.maxRequestsPerHost] to configure how many calls
+   * OkHttp performs concurrently.
+   */
+  open fun dispatcherQueueStart(
```
yschimke, line 81: Are start and end the right terms for this?

dispatcherQueued and dispatcherExecuted?

okhttp/src/commonJvmAndroid/kotlin/okhttp3/EventListener.kt
```diff
@@ -70,6 +71,30 @@ abstract class EventListener {
...
+    dispatcher: Dispatcher,
+  ) {
+  }
+
+  /**
+   * Invoked when [call] will be executed immediately.
+   *
+   * This method is invoked after [dispatcherQueueStart].
+   */
+  open fun dispatcherQueueEnd(
```
yschimke, line 92: It seems like this event should fire even if immediately executed. Could be useful signal for the call actually running.

And just not sent in a call.execute case

okhttp/src/commonJvmAndroid/kotlin/okhttp3/EventListener.kt
swankjesse: I'm thinking about it like spans in a tracing system or flame graph. Where did this call spend its first 800 milliseconds? In the dispatcher queue.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/EventListener.kt
swankjesse: Again, I'm stuck on the spans use case where this is triggering that something completed.

PR #9107, opened 2025-10-04
Add KDoc for `Request.Builder.query()`

martinbonnin:


yschimke:
If we want to add the cache key to the method we should do it now.

Avoid deprecating or two overloads

martinbonnin:
Oh I thought the question was whether we wanted 2 overloads?

```kotlin
fun query()
fun query(cacheUrlOverride: String)
```

If we don't want overloads, I'm fine with the current `query()`. Even if we do, I think I'm fine, it's symmetrical and consistent with eveything else.

yschimke:
Yeah, good question. I guess it's 

- Java users vs Kotlin with defaults
- null means remove, or just no change

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Request.kt
```diff
@@ -317,6 +317,30 @@ class Request internal constructor(
...
+     * how to cache them.
+     *
+     * A typical use case is to hash the request body:
+     *
+     * ```kotlin
+     *     val hashingSink = HashingSink.md5(blackholeSink())
+     *     hashingSink.buffer().use {
+     *       body.writeTo(it)
+     *     }
+     *     val hash = hashingSink.hash.hex()
```
swankjesse, line 333: Wanna add a function, `RequestBody.sha256()` ?

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Request.kt
```diff
@@ -317,6 +317,30 @@ class Request internal constructor(
...
+    /**
+     * Sets this request's method to `QUERY`.
+     *
+     * By default, `QUERY` requests are not cached. You can use [cacheUrlOverride] to specify
+     * how to cache them.
+     *
+     * A typical use case is to hash the request body:
+     *
+     * ```kotlin
+     *     val hashingSink = HashingSink.md5(blackholeSink())
```
swankjesse, line 329: We shouldn’t be using MD5 unless we need to for compatibility with other systems. Only SHA256 for new uses

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Request.kt
martinbonnin: https://github.com/square/okhttp/pull/9109

PR #9108, opened 2025-10-05
Allow OkHttpClient overrides in Interceptor

yschimke:
This pull request introduces significant enhancements to the `Interceptor.Chain` API in OkHttp, allowing much more fine-grained and dynamic control over the HTTP client's configuration on a per-call basis. 

Let's delve into it;

**API Extensions for Interceptor.Chain**

* Added numerous getter and setter methods to the `Interceptor.Chain` interface for properties such as DNS, proxy, cache, authenticators, SSL factories, certificate pinners, connection pool, cookie jar, hostname verifier, and retry settings. These methods allow interceptors and clients to override or access the OkHttpClient's configuration dynamically within the interceptor chain.

**Internal Refactoring to Use Chain Properties**

* Refactored `BridgeInterceptor` and `CacheInterceptor` to use the new properties from `Interceptor.Chain` instead of being directly constructed with dependencies like `CookieJar` or `Cache`. This makes interceptors more decoupled and flexible.

**Extensive test of overrides**

* `InterceptorOverridesTest` exhaustively tests that overrides can be applied by an application interceptor and observed from a network interceptor. That they can't be set in a network interceptor. And finally that bad defaults on OkHttpClient, can be saved by good overrides.


okhttp/src/commonJvmAndroid/kotlin/okhttp3/Interceptor.kt
```diff
@@ -108,5 +109,9 @@ fun interface Interceptor {
       timeout: Int,
       unit: TimeUnit,
     ): Chain
+
+    fun withDns(dns: Dns?): Chain
```
yschimke: TODO

- [ ] Add getters for existing values
- [ ] Add Proxy, TLS

okhttp/src/commonJvmAndroid/kotlin/okhttp3/OkHttpClient.kt
```diff
@@ -312,6 +312,7 @@ open class OkHttpClient internal constructor(
    * Creates an [Address] of out of the provided [HttpUrl]
    * that uses this client’s DNS, TLS, and proxy configuration.
    */
+  @Deprecated("Not intended to be public API on OkHttpClient")
```
swankjesse: Why not? I think this is out of scope for this PR.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/http/RealInterceptorChain.kt
```diff
@@ -98,6 +108,22 @@ class RealInterceptorChain(
...
+      copy(dns = dns)
+    }
+
+  override fun withSocketFactory(socketFactory: SocketFactory?): Interceptor.Chain =
+    withClientOverrides {
+      copy(socketFactory = socketFactory)
+    }
+
+  private fun withClientOverrides(overrides: ClientOverrides.() -> ClientOverrides): Interceptor.Chain {
+    check(exchange == null) { "ClientOverrides can't be adjusted in a network interceptor" }
```
swankjesse: nit: the caller doesn’t know what a ClientOverrides is . . . maybe pass a string like "socketFactory" into this function, so we can throw an error message the caller will understand?

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Interceptor.kt
swankjesse: Why is Dns nullable? We have `Dns.SYSTEM` for that

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Interceptor.kt
```diff
@@ -108,5 +109,9 @@ fun interface Interceptor {
...
+
+    fun withSocketFactory(socketFactory: SocketFactory?): Chain
```
swankjesse: Similarly, `SocketFactory.getDefault()`

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/http/RealInterceptorChain.kt
```diff
@@ -98,6 +108,22 @@ class RealInterceptorChain(
...
+
+    return copy(clientOverrides = clientOverrides)
```
swankjesse: This doesn’t appear to stack, and it needs to stack. If you `with()` on the dns and also on the socketFactory, it should not be last-one wins

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/http/RealInterceptorChain.kt
```diff
@@ -135,4 +161,39 @@ class RealInterceptorChain(
...
+      certificatePinner = useCertificatePinner,
+      proxyAuthenticator = call.client.proxyAuthenticator,
+      proxy = call.client.proxy,
+      protocols = call.client.protocols,
+      connectionSpecs = call.client.connectionSpecs,
+      proxySelector = call.client.proxySelector,
+    )
+  }
+
+  data class ClientOverrides(
```
swankjesse: I think I’d prefer adding fields to RealInterceptorChain. It avoids having to build a complex mechanism to support chaining ClientOverrides

swankjesse (changes requested):

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Interceptor.kt
yschimke: I figured null removes an override. Maybe not needed

okhttp/src/commonJvmAndroid/kotlin/okhttp3/OkHttpClient.kt
yschimke: Because I'm removing the only caller, and I'm not sure we expect people to call it?

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Interceptor.kt
```diff
@@ -108,5 +115,65 @@ fun interface Interceptor {
       timeout: Int,
       unit: TimeUnit,
     ): Chain
+
+    val dns: Dns
```
yschimke, line 147: For provoking discussion, rather than the final set.

The advantage of Call.Decorator was we didn't need to copy all these properties.

we could 
- start out small, and add as each is shown valuable. 
- open all
- be opinionated based on our ideas

okhttp/src/commonJvmAndroid/kotlin/okhttp3/OkHttpClient.kt
yschimke: Why is it here?

swankjesse: Probably a bug that it’s here. But I don’t like deprecating stuff unless we’re going to remove it later, and I definitely don’t like removing stuff later. Ultimately I believe so much in this thing that I never want to break APIs, and we should have a stronger reason to deprecate something than ‘we don’t like it anymore’.

https://www.youtube.com/watch?v=oyLBGkS5ICk

swankjesse (commented):
I think this is pretty rad actually. The most difficult part is gonna be making the tests?

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Interceptor.kt
```diff
@@ -108,5 +115,65 @@ fun interface Interceptor {
...
+
+    fun withCertificatePinner(certificatePinner: CertificatePinner): Chain
+
+    val certificateChainCleaner: CertificateChainCleaner?
+
+    fun withCertificateChainCleaner(certificateChainCleaner: CertificateChainCleaner): Chain
+
+    val connectionPool: ConnectionPool
+
+    fun withConnectionPool(connectionPool: ConnectionPool): Chain
```
swankjesse, line 294: Should we throw if you call this on a network interceptor, since you’re already connected?

yschimke: Yes, although I do wonder if that is ok, if the meaning is that the override applies for follow up calls, that do.

But for later.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/connection/RealCall.kt
```diff
@@ -276,14 +274,17 @@ class RealCall(
         RealRoutePlanner(
           taskRunner = client.taskRunner,
           connectionPool = connectionPool,
+          // TODO check if override should be used
```
yschimke: @swankjesse seems like a bug?

yschimke:
@swankjesse I'll follow up with more specific samples, such as Android Network Pinning.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Interceptor.kt
```diff
@@ -75,38 +82,228 @@ fun interface Interceptor {
   }
 
   interface Chain {
+    /**
+     * Returns the request being executed.
```
swankjesse: redundant?

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/connection/RealCall.kt
swankjesse: Indeed!

swankjesse (commented):
Love it

I would like more confidence that the override value is actually used, different from whether it’s reflected in the chain’s own properties. That’s necessary but not sufficient.

okhttp/src/jvmTest/kotlin/okhttp3/InterceptorOverridesTest.kt
```diff
@@ -0,0 +1,619 @@
...
+        } else if (override == OverrideParam.Authenticator) {
+          server.enqueue(
+            MockResponse.Builder().code(401).build(),
+          )
+        }
+      }
+        // Set the bad override directly on the client
+        .withOverride(badValue)
+        .addInterceptor { chain ->
+          // the only way to stop a bad override of a client is with a good override of an interceptor
```
yschimke: @swankjesse my plan for testing

yschimke:
@swankjesse still have a few tests to fix, but ready for review

okhttp/src/jvmTest/kotlin/okhttp3/InterceptorOverridesTest.kt
```diff
@@ -0,0 +1,858 @@
...
+      OverrideParam.WriteTimeout -> {
+        val body =
+          object : RequestBody() {
+            override fun contentType(): MediaType? = null
+
+            override fun writeTo(sink: BufferedSink) {
+              if (sink
+                  .timeout()
+                  .timeoutNanos()
+                  .nanoseconds.inWholeMilliseconds == 10L
```
yschimke, line 289: @swankjesse I tried to implement this using 

```
            .socketFactory(
              DelayingSocketFactory(onWrite = {
                Thread.sleep(100L)
              }),
```

But if I break point at Thread.sleep, then socket.sink().timeout().timeoutNanos() is always cleared (0).

I suspect a bug with write timeout, but this replicates a timeout just by checking if we have the bad timeout value.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/connection/RealCall.kt
```diff
@@ -209,9 +209,9 @@ class RealCall(
     // Build a full stack of interceptors.
     val interceptors = mutableListOf<Interceptor>()
     interceptors += client.interceptors
-    interceptors += RetryAndFollowUpInterceptor(client)
-    interceptors += BridgeInterceptor(client.cookieJar)
-    interceptors += CacheInterceptor(this, client.cache)
+    interceptors += RetryAndFollowUpInterceptor()
```
yschimke, line 212: @swankjesse this seems like a nice win.

yschimke:
@swankjesse requesting post review. I think I've addressed major comments, and taking a previous positive impression.

Happy to revert or fix forward with any feedback.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/connection/RealCall.kt
swankjesse: EXTREMELY SO

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/http/RealInterceptorChain.kt
```diff
@@ -61,41 +141,167 @@ class RealInterceptorChain(
...
 
-    return copy(connectTimeoutMillis = checkDuration("connectTimeout", timeout.toLong(), unit))
+    return copy(connectTimeoutMillis = checkDuration("connectTimeout", timeout, unit))
   }
 
   override fun readTimeoutMillis(): Int = readTimeoutMillis
 
   override fun withReadTimeout(
-    timeout: Int,
+    timeout: Long,
```
swankjesse, line 186: Ugh, this is binary-incompatible and we must revert to `Int`

okhttp/api/jvm/okhttp.api
```diff
@@ -749,12 +749,42 @@ public abstract interface class okhttp3/Interceptor$Chain {
...
+	public abstract fun getRetryOnConnectionFailure ()Z
+	public abstract fun getSocketFactory ()Ljavax/net/SocketFactory;
+	public abstract fun getSslSocketFactoryOrNull ()Ljavax/net/ssl/SSLSocketFactory;
+	public abstract fun getX509TrustManagerOrNull ()Ljavax/net/ssl/X509TrustManager;
 	public abstract fun proceed (Lokhttp3/Request;)Lokhttp3/Response;
 	public abstract fun readTimeoutMillis ()I
 	public abstract fun request ()Lokhttp3/Request;
-	public abstract fun withConnectTimeout (ILjava/util/concurrent/TimeUnit;)Lokhttp3/Interceptor$Chain;
-	public abstract fun withReadTimeout (ILjava/util/concurrent/TimeUnit;)Lokhttp3/Interceptor$Chain;
-	public abstract fun withWriteTimeout (ILjava/util/concurrent/TimeUnit;)Lokhttp3/Interceptor$Chain;
```
swankjesse, line 757: This file must never have red in it

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/http/RealInterceptorChain.kt
```diff
@@ -61,41 +141,167 @@ class RealInterceptorChain(
...
+    return copy(proxySelector = proxySelector)
+  }
+
+  override fun withProxyAuthenticator(proxyAuthenticator: Authenticator): Interceptor.Chain {
+    check(exchange == null) { "proxyAuthenticator can't be adjusted in a network interceptor" }
+
+    return copy(proxyAuthenticator = proxyAuthenticator)
+  }
+
+  override fun withSslSocketFactory(
```
swankjesse, line 259: Nice nice nice

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Interceptor.kt
```diff
@@ -82,31 +89,210 @@ fun interface Interceptor {
...
      */
     fun connection(): Connection?
 
+    /**
+     * Returns the `Call` to which this chain belongs.
+     */
     fun call(): Call
 
+    /**
+     * Returns the connect timeout in milliseconds.
```
swankjesse, line 102: 😬  https://publicobject.com/2007/08/05/stop_writing_redundant_javadoc/

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Interceptor.kt
```diff
@@ -82,31 +89,210 @@ fun interface Interceptor {
...
+
+    /**
+     * Override the [Authenticator] for the Call.Chain.
+     *
+     * @throws IllegalStateException if this is a Network Interceptor, since the override is too late.
+     */
+    fun withAuthenticator(authenticator: Authenticator): Chain
+
+    /**
+     * Returns the [CookieJar] for the OkHttpClient, or an override from the Call.Chain.
```
swankjesse, line 191: These docs describe the implementation mechanism, but not the purpose.

I’d prefer, ‘Returns the cookie jar that'll be used for this call’

okhttp/src/jvmTest/kotlin/okhttp3/InterceptorOverridesTest.kt
```diff
@@ -0,0 +1,858 @@
+package okhttp3
```
swankjesse, line 1: We should attribute copyright to somebody here.

okhttp/src/jvmTest/kotlin/okhttp3/InterceptorOverridesTest.kt
```diff
@@ -0,0 +1,858 @@
...
+        OverrideParam.ReadTimeout,
+        OverrideParam.RetryOnConnectionFailure,
+        OverrideParam.SocketFactory,
+        OverrideParam.SslSocketFactory,
+        OverrideParam.WriteTimeout,
+        OverrideParam.X509TrustManager,
+      ),
+    isDefault: Boolean,
+  ) {
+    fun <T> Override<T>.testApplicationInterceptor(chain: Interceptor.Chain): Response {
```
swankjesse, line 91: Rad

okhttp/src/jvmTest/kotlin/okhttp3/InterceptorOverridesTest.kt
```diff
@@ -0,0 +1,858 @@
...
+    body: RequestBody? = null,
+  ) {
+    with(override) {
+      client =
+        client
+          .newBuilder()
+          // Set the bad override directly on the client
+          .withOverride(badValue)
+          .addInterceptor { chain ->
+            // the only way to stop a bad override of a client is with a good override of an interceptor
```
swankjesse, line 362: The only way to stop a bad override with a gun...

swankjesse (commented):
LGTM!

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/http/RealInterceptorChain.kt
yschimke: Not sure what I was thinking

okhttp/src/jvmTest/kotlin/okhttp3/InterceptorOverridesTest.kt
iVamsi: @yschimke I updated this test and verified that it works after the fix in #9653. Can you please review?

PR #9109, opened 2025-10-06
Add RequestBody.sha256()

martinbonnin:
Follow up from https://github.com/square/okhttp/pull/9107

### DIFF
--- a/okhttp/api/android/okhttp.api
+++ b/okhttp/api/android/okhttp.api
@@ -1073,6 +1073,7 @@ public abstract class okhttp3/RequestBody {
 	public static final fun create ([BLokhttp3/MediaType;II)Lokhttp3/RequestBody;
 	public fun isDuplex ()Z
 	public fun isOneShot ()Z
+	public final fun sha256 ()Lokio/ByteString;
 	public abstract fun writeTo (Lokio/BufferedSink;)V
 }
 
@@ -1099,10 +1100,6 @@ public final class okhttp3/RequestBody$Companion {
 	public static synthetic fun create$default (Lokhttp3/RequestBody$Companion;Lokio/ByteString;Lokhttp3/MediaType;ILjava/lang/Object;)Lokhttp3/RequestBody;
 	public static synthetic fun create$default (Lokhttp3/RequestBody$Companion;Lokio/Path;Lokio/FileSystem;Lokhttp3/MediaType;ILjava/lang/Object;)Lokhttp3/RequestBody;
 	public static synthetic fun create$default (Lokhttp3/RequestBody$Companion;[BLokhttp3/MediaType;IIILjava/lang/Object;)Lokhttp3/RequestBody;
-}
-
-public final class okhttp3/RequestBodyKt {
-	public static final fun sha256 (Lokhttp3/RequestBody;)Ljava/lang/String;
 }
 
 public final class okhttp3/Response : java/io/Closeable {
--- a/okhttp/api/android/okhttp.api
+++ b/okhttp/api/android/okhttp.api
@@ -1101,6 +1101,10 @@ public final class okhttp3/RequestBody$Companion {
 	public static synthetic fun create$default (Lokhttp3/RequestBody$Companion;[BLokhttp3/MediaType;IIILjava/lang/Object;)Lokhttp3/RequestBody;
 }
 
+public final class okhttp3/RequestBodyKt {
+	public static final fun sha256 (Lokhttp3/RequestBody;)Ljava/lang/String;
+}
+
 public final class okhttp3/Response : java/io/Closeable {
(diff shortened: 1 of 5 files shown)


okhttp/api/android/okhttp.api
```diff
@@ -1101,6 +1101,10 @@ public final class okhttp3/RequestBody$Companion {
 	public static synthetic fun create$default (Lokhttp3/RequestBody$Companion;[BLokhttp3/MediaType;IIILjava/lang/Object;)Lokhttp3/RequestBody;
 }
 
+public final class okhttp3/RequestBodyKt {
+	public static final fun sha256 (Lokhttp3/RequestBody;)Ljava/lang/String;
```
martinbonnin: Remind me, what's the policy for Java interop? Should I make this unaccessible from Java or use `@JvmName` so that it reads better? Or just use a plain `RequestBody` method (not extension)?

okhttp/src/commonJvmAndroid/kotlin/okhttp3/RequestBody.kt
```diff
@@ -257,3 +260,14 @@ abstract class RequestBody {
     ): RequestBody = file.asRequestBody(contentType)
   }
 }
+
+/**
+ * Returns the SHA256 hash of the [RequestBody], as an hexadecimal String.
+ */
+fun RequestBody.sha256(): String {
```
swankjesse: Make this an extension function? It’s nicer from Java that way

swankjesse: Please return a `ByteString`, not a hex string. The caller should do `.hex()` if they want to display it, but for all we know they want to put it in a database or something

okhttp/src/commonJvmAndroid/kotlin/okhttp3/RequestBody.kt
```diff
@@ -257,3 +260,14 @@ abstract class RequestBody {
     ): RequestBody = file.asRequestBody(contentType)
   }
 }
+
+/**
+ * Returns the SHA256 hash of the [RequestBody], as an hexadecimal String.
```
swankjesse: nit: SHA-256

okhttp/src/jvmTest/kotlin/okhttp3/RequestBodyTest.kt
```diff
@@ -119,6 +119,15 @@ class RequestBodyTest {
     }
   }
 
+  @Test
+  fun testSha256() {
+    assertOnPath(content = "Hello") { path ->
```
swankjesse: `assertOnPath` seems like it introduces the file system, and I don’t think we need to?

This should work just as well:

`"Hello".toRequestBody()`

okhttp/src/commonJvmAndroid/kotlin/okhttp3/RequestBody.kt
martinbonnin: > Make this an extension function? It’s nicer from Java that way

Not sure I follow, make it a class method? Since this is an extension function already?

okhttp/src/jvmTest/kotlin/okhttp3/RequestBodyTest.kt
martinbonnin: Totally 👍

okhttp/api/android/okhttp.api
swankjesse: Just a regular polymorphic method please

swankjesse: (and not an extension)

martinbonnin: https://github.com/square/okhttp/pull/9109/commits/e34156827a0e18d136cba2bf95a3bce8f0e320f0

okhttp/src/commonJvmAndroid/kotlin/okhttp3/RequestBody.kt
```diff
@@ -95,6 +98,17 @@ abstract class RequestBody {
    */
   open fun isOneShot(): Boolean = false
 
+  /**
+   * Returns the SHA-256 hash of this [RequestBody]
+   */
+  fun sha256(): ByteString {
```
JakeWharton, line 105: Should this have a verb, like "compute", given that it's doing something expensive (from an I/O perspective)?

swankjesse: I think making it a function and not a `val` means we have freedom to do something expensive.

BUT! It should declare `@Throws(IOException::class)`

swankjesse (approved):

PR #9110, opened 2025-10-06
Use a single synchronized block in Dispatcher

swankjesse:
Previously we had synchronized blocks preceding calls to promoteAndExecute, plus the synchronized blocks in that function itself.

This is intended to make it easier to publishg the right events for dispatcherQueueStart and dispatcherQueueEnd when an enqueued call can skip the queue.

Also note that this should fix some corner-cases around unnecessary calls to idleCallback when the executor is already shut down.

### DIFF
--- a/okhttp/src/commonJvmAndroid/kotlin/okhttp3/Dispatcher.kt
+++ b/okhttp/src/commonJvmAndroid/kotlin/okhttp3/Dispatcher.kt
@@ -16,7 +16,6 @@
 package okhttp3
 
 import java.util.ArrayDeque
-import java.util.Deque
 import java.util.concurrent.ExecutorService
 import java.util.concurrent.SynchronousQueue
 import java.util.concurrent.ThreadPoolExecutor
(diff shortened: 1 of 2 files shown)


okhttp/src/commonJvmAndroid/kotlin/okhttp3/Dispatcher.kt
```diff
@@ -167,52 +156,92 @@ class Dispatcher() {
...
+          // the same host.
+          if (!enqueuedCall.call.forWebSocket) {
+            val existingCall = findExistingCallWithHost(enqueuedCall.host)
+            if (existingCall != null) enqueuedCall.reuseCallsPerHostFrom(existingCall)
+          }
         }
 
-        asyncCall.failRejected()
-      }
-      idleCallback?.run()
```
swankjesse, line 207: I believe this would have triggered unnecessary calls to the idle callback. For example, each time `maxRequests` is mutated, this would be called if the executor service was shut down, even if it was already idle.

yschimke: Nice catch

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Dispatcher.kt
```diff
@@ -167,52 +156,92 @@ class Dispatcher() {
...
+    val executorIsShutdown = executorService.isShutdown
 
-    val executableCalls = mutableListOf<AsyncCall>()
-    val isRunning: Boolean
-    synchronized(this) {
-      val i = readyAsyncCalls.iterator()
-      while (i.hasNext()) {
-        val asyncCall = i.next()
+    // Actions to take outside the synchronized block.
+    class Effects(
```
yschimke, line 172: I like this pattern across barriers

PR #9111, opened 2025-10-06
Start publishing dispatcher queue events

swankjesse:


okhttp-testing-support/src/main/kotlin/okhttp3/RecordingEventListener.kt
```diff
@@ -69,8 +69,12 @@ open class RecordingEventListener(
    */
   private val enforceOrder: Boolean = true,
 ) : EventListener() {
+  /** Events that haven't yet been removed. */
   val eventSequence: Deque<CallEvent> = ConcurrentLinkedDeque()
 
+  /** The full set of events, used to match starts with ends. */
+  private val eventsForMatching = ConcurrentLinkedDeque<CallEvent>()
```
swankjesse, line 76: (My test case consumes events before the test completes, and that was breaking matching.)

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Dispatcher.kt
```diff
@@ -233,12 +233,26 @@ class Dispatcher() {
...
-    for (i in 0 until effects.callsToReject.size) {
-      effects.callsToReject[i].failRejected()
-    }
+    var callDispatcherQueueStart = true
 
     for (i in 0 until effects.callsToExecute.size) {
-      effects.callsToExecute[i].executeOn(executorService)
+      val call = effects.callsToExecute[i]
+
+      if (call === enqueuedCall) {
```
yschimke, line 243: Maybe worth a comment. #tioli

okhttp/src/jvmTest/kotlin/okhttp3/DispatcherTest.kt
```diff
@@ -78,6 +83,9 @@ class DispatcherTest {
   fun enqueuedJobsRunImmediately() {
     client.newCall(newRequest("http://a/1")).enqueue(callback)
     executor.assertJobs("http://a/1")
+
+    assertThat(listener.eventSequence).none { it.isInstanceOf<DispatcherQueueStart>() }
```
yschimke, line 87: Nice extension of tests

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Dispatcher.kt
swankjesse:
```suggestion
      // If the newly-enqueued call is already out, skip its dispatcher queue events. We only
      // publish those events for calls that have to wait.
      if (call === enqueuedCall) {
```

swankjesse: Yeah, great idea

PR #9112, opened 2025-10-06
Make Request.toCurl work more like Chrome's 'copy as cURL'

swankjesse:
Changes:
 - single quotes
 - URL first
 - omit the default method

Also change binary detection to use isProbablyUtf8 instead, which is promoted to the main okhttp module.

Also replace the BinaryMode enum with a boolean to optionally not include the body.


swankjesse:
Also needs a follow-up to use the body’s content type if it is present. (We prefer the body’s content-type if both are included)

JakeWharton (approved):

PR #9114, opened 2025-10-07
Bump Gradle to 9.1.0

martinbonnin:


yschimke:
We reverted this just recently because of issues.  https://github.com/square/okhttp/pull/9071

The changes to move don't see simple, given we have GraalVM, OSGi, Java 9 modules, Android etc.

martinbonnin:
Sorry I missed this! 

What was the root cause exactly? 

I have read https://developer.android.com/build/releases/gradle-plugin#updating-gradle but they mention a min Gradle version, not max? I would assume AGP 8.13 is compatibile with Gradle 9? I think I have a bunch of projects using this setup without issues and all the tests are passing in this PR looks like?

JakeWharton:
Yeah I don't think anyone verified the claim. The person said it was a problem in VS Code, I think.

yschimke:
I can reproduce in VS Code

Also https://github.com/microsoft/vscode-gradle/issues/1712

yschimke:
The workaround is editing gradle.properties

```
org.gradle.parallel=false
```

yschimke:
Since the workaround unblocks VS Code, I'm going to land so we fix AGP issues against Gradle 9.

sschuberth:
Why does this not upgrade the wrapper scripts as well?

JakeWharton:
They were already upgraded in #9072 and not reverted as part of #9071. So they're already up to date.

PR #9119, opened 2025-10-07
New function, EventListener.plus()

swankjesse:
Use this to support multiple EventListeners in the same call.


JakeWharton (approved):

okhttp/src/commonJvmAndroid/kotlin/okhttp3/EventListener.kt
```diff
@@ -535,6 +535,24 @@ abstract class EventListener {
   ) {
   }
 
+  /** Returns a new `EventListener` that publishes events to this and then `other`. */
+  operator fun plus(other: EventListener): EventListener {
```
yschimke, line 539: Nice simple solution

PR #9129, opened 2025-10-09
Fix SSL Handshake issue affecting MockWebserver on Android 24

yschimke:
Only called in MWS, but fails on API 24 when called. Adds a single Junit 4 test to run on Android 21-25.

Add a test since Junit 5 wasn't running on older devices.


JakeWharton:
Do we actually utilize anything from JUnit 5 in the original tests? Would rather migrate away to 4 so this isn't a problem in the future.

Issue #9125, opened 2025-10-09
Silently throwing robolectric RunTimeException in unit test


vincent-paing:
OKHttpVersion used : `5.2.0`

Reproduction Steps
- Create an empty Android project
- Add OkHttp depdencies

```kotlin
// define a BOM and its version
implementation(platform("com.squareup.okhttp3:okhttp-bom:5.2.0"))

// define any required OkHttp artifacts without version
implementation("com.squareup.okhttp3:okhttp")
implementation("com.squareup.okhttp3:logging-interceptor")
```

- In `ExampleUnitTest`, add this `OkHttpClient.Builder().build()` in the first line of `addition_isCorrect`.

```kotlin
class ExampleUnitTest {
    @Test
    fun addition_isCorrect() {
        OkHttpClient.Builder().build()
        assertEquals(4, 2 + 2)
    }
}
```

- Run the test
- The test passes but there's runtime exception being thrown silently that's causing a a lot of noise in the test output
```
Possibly running android unit test without robolectric
java.lang.RuntimeException: Method isLoggable in android.util.Log not mocked. See https://developer.android.com/r/studio-ui/build/not-mocked for details.
	at android.util.Log.isLoggable(Log.java)
	at okhttp3.internal.platform.android.AndroidLog.enableLogging(AndroidLog.kt:127)
	at okhttp3.internal.platform.android.AndroidLog.enable(AndroidLog.kt:109)
	at okhttp3.internal.platform.PlatformRegistry.findPlatform(PlatformRegistry.kt:25)
	at okhttp3.internal.platform.Platform$Companion.findPlatform(Platform.kt:225)
	at okhttp3.internal.platform.Platform$Companion.access$findPlatform(Platform.kt:203)
	at okhttp3.internal.platform.Platform.<clinit>(Platform.kt:204)
	at okhttp3.OkHttpClient.<init>(OkHttpClient.kt:300)
	at okhttp3.OkHttpClient$Builder.build(OkHttpClient.kt:1384)
	at com.example.okhttptest.ExampleUnitTest.addition_isCorrect(ExampleUnitTest.kt:17)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(Unknown Source)
	at java.base/java.lang.reflect.Method.invoke(Unknown Source)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
[...]

PR #9137, opened 2025-10-11
Catch UnsatisfiedLinkError in AndroidLog

yschimke:
Happens with Paparazzi. see https://github.com/square/okhttp/issues/9125

### DIFF
--- a/okhttp/src/androidMain/kotlin/okhttp3/internal/platform/android/AndroidLog.kt
+++ b/okhttp/src/androidMain/kotlin/okhttp3/internal/platform/android/AndroidLog.kt
@@ -112,7 +112,7 @@ object AndroidLog {
       // Happens with non-robolectric unit tests
       System.err.println("Possibly running android unit test without robolectric")
       re.printStackTrace()
-    } catch (ule: Error) {
+    } catch (ule: UnsatisfiedLinkError) {
       // Happens with Paparazzi - with Android classes on the classpath
       System.err.println("Possibly running android unit test without robolectric")
       ule.printStackTrace()


okhttp/src/androidMain/kotlin/okhttp3/internal/platform/android/AndroidLog.kt
```diff
@@ -109,8 +109,13 @@ object AndroidLog {
         enableLogging(logger, tag)
       }
     } catch (re: RuntimeException) {
+      // Happens with non-robolectric unit tests
       System.err.println("Possibly running android unit test without robolectric")
       re.printStackTrace()
+    } catch (ule: Error) {
```
swankjesse: Catch a more targeted UnsatisfiedLinkError?

When these exceptions happen, is there a fallback we should be enabling instead?

swankjesse (approved):

okhttp/src/androidMain/kotlin/okhttp3/internal/platform/android/AndroidLog.kt
yschimke: But I guess I wanted to do either
a) Catch anything that can be thrown
b) no by some other means, e.g. is Dalvik, that it's a good android environment.

I'll change to RuntimeException

yschimke: The first RuntimeException is from

https://cs.android.com/android-studio/platform/tools/base/+/mirror-goog-studio-main:build-system/builder/src/main/java/com/android/builder/testing/MockableJarGenerator.java;l=250?q=MockableJarGenerator

yschimke: I guess second is JNI, so UnsatisfiedLinkError is specific

PR #9148, opened 2025-10-12
Factory decorator as interceptor

yschimke:
A take on Call.Decorator as an Interceptor https://github.com/square/okhttp/pull/8800

Demonstrated with an Android Network Pinning Interceptor from https://github.com/square/okhttp/pull/8376

```
class AndroidNetworkPinning : ClientForkingInterceptor<Network>() {
  override fun OkHttpClient.Builder.buildForKey(key: Network): OkHttpClient =
    dns(AndroidDns(key))
      .socketFactory(key.socketFactory)
      .apply {
        // Keep interceptors after this one in the new client
        interceptors.subList(interceptors.indexOf(this@AndroidNetworkPinning) + 1, interceptors.size).clear()
      }.build()

  override fun clientKey(request: Request): Network? = request.tag<Network>()
}
```

Introduces a ClientForkingInterceptor, which hides the nastiness of accessing the client

```
  override fun intercept(chain: Interceptor.Chain): Response {
    val client =
      (chain.call() as? RealCall)?.client ?: throw IOException("unable to access OkHttpClient")

    val key = clientKey(chain.request())
```

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -64,9 +64,9 @@ val localProperties = Properties().apply {
   }
 }
 val sdkDir = localProperties.getProperty("sdk.dir")
-//if (androidHome != null || sdkDir != null) {
+if (androidHome != null || sdkDir != null) {
   include(":android-test")
   include(":android-test-app")
-//}
+}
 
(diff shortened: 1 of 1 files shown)


yschimke:
cc @swankjesse @pyricau

PR #9159, opened 2025-10-21
Support Request Bodies on HTTP1.1 Connection Upgrades

gesellix:
Required for the Docker Engine API endpoint `/exec/{id}/start`: https://docs.docker.com/reference/api/engine/version/v1.51/#tag/Exec/operation/ExecStart

Example:

```
POST /exec/{id}/start
Content-Type: application/json
Connection: Upgrade
Upgrade: tcp

{
  "Detach": false,
  "Tty": true,
  "ConsoleSize": [
    80,
    64
  ]
}
```

### DIFF
--- a/okhttp/src/jvmTest/kotlin/okhttp3/internal/http/HttpUpgradesTest.kt
+++ b/okhttp/src/jvmTest/kotlin/okhttp3/internal/http/HttpUpgradesTest.kt
@@ -17,8 +17,8 @@ package okhttp3.internal.http
 
 import assertk.assertThat
 import assertk.assertions.containsExactly
-import assertk.assertions.hasMessage
 import assertk.assertions.isEqualTo
+import assertk.assertions.isInstanceOf
 import assertk.assertions.isNull
 import assertk.assertions.isTrue
 import kotlin.test.assertFailsWith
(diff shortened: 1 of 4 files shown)


okhttp/src/jvmTest/kotlin/okhttp3/internal/http/HttpUpgradesTest.kt
```diff
@@ -225,21 +236,23 @@ class HttpUpgradesTest {
...
       ResponseHeadersStart::class,
       ResponseHeadersEnd::class,
       FollowUpDecision::class,
       RequestBodyStart::class,
       ResponseBodyStart::class,
       ResponseBodyEnd::class,
-      RequestBodyEnd::class,
       ConnectionReleased::class,
       CallEnd::class,
+      RequestBodyEnd::class,
```
gesellix: I'm not sure how to let the `RequestBodyEnd` appear before `CallEnd` or before `ConnectionReleased`.

swankjesse: I’d like to see if I can make that work. We document that `callEnd()` or `callFailed()` will be the last function on an event listener, and I think we can retain that behavior.

gesellix:
Integration tests for Docker's `/exec/{id}/start` endpoint in interactive mode are working as expected.

PR #9163, opened 2025-10-23
New datastructure to hold Tags

squarejesse:
We have been using MutableMap and Map. These have worked fine.

But we'd like to soon support mutating tags on an existing Request object, which I'd like to do without a ton of collections or concurrency overhead. Instead we can do some lock-free stuff, which needs a simple data structure as a starting point.

### DIFF
--- a/okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/Tags.kt
+++ b/okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/Tags.kt
@@ -33,9 +33,9 @@ internal sealed class Tags {
    * Returns a tags instance that maps [key] to [value]. If [value] is null, this returns a tags
    * instance that does not have any mapping for [key].
    */
-  abstract fun plus(
-    key: KClass<*>,
-    value: Any?,
+  abstract fun <T : Any> plus(
+    key: KClass<T>,
+    value: T?,
   ): Tags
 
   abstract operator fun <T : Any> get(key: KClass<T>): T?
@@ -43,9 +43,9 @@ internal sealed class Tags {
 
 /** An empty tags. This is always the tail of a [LinkedTags] chain. */
 internal object EmptyTags : Tags() {
-  override fun plus(
-    key: KClass<*>,
-    value: Any?,
+  override fun <T : Any> plus(
+    key: KClass<T>,
+    value: T?,
   ): Tags =
     when {
       value != null -> LinkedTags(key, value, this)
@@ -61,14 +61,14 @@ internal object EmptyTags : Tags() {
  * An invariant of this implementation is that [next] must not contain a mapping for [key].
  * Otherwise, we would have two values for the same key.
  */
-private class LinkedTags(
-  private val key: KClass<*>,
-  private val value: Any,
+private class LinkedTags<K : Any>(
+  private val key: KClass<K>,
+  private val value: K,
   private val next: Tags,
 ) : Tags() {
-  override fun plus(
-    key: KClass<*>,
-    value: Any?,
+  override fun <T : Any> plus(
+    key: KClass<T>,
+    value: T?,
   ): Tags {
     // Create a copy of this `LinkedTags` that doesn't have a mapping for `key`.
     val thisMinusKey =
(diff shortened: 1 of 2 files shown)


okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/Tags.kt
```diff
@@ -0,0 +1,106 @@
...
+ * building an instance with _N_ elements is _O(N**2)_.
+ */
+internal sealed class Tags {
+  /**
+   * Returns a tags instance that maps [key] to [value]. If [value] is null, this returns a tags
+   * instance that does not have any mapping for [key].
+   */
+  abstract fun plus(
+    key: KClass<*>,
+    value: Any?,
```
JakeWharton: Is there a reason this isn't using a type parameter to ensure the KClass and value refer to the same type?

swankjesse: No reason! Done.

okhttp/src/jvmTest/kotlin/okhttp3/internal/TagsTest.kt
```diff
@@ -0,0 +1,160 @@
...
+  fun removeMiddleElement() {
+    val tags =
+      EmptyTags
+        .plus(Integer::class, 5 as Integer)
+        .plus(String::class, "a")
+        .plus(Boolean::class, true)
+        .plus(String::class, null)
+    assertThat(tags[String::class]).isNull()
+    assertThat(tags.toString())
+      .isEqualTo("{class kotlin.Int=5, class kotlin.Boolean=true}")
```
yschimke, line 113: Does the "class " prefix add much information here? I'd be tempted to even use simpleName by default, as I think your examples are short, and enterprise java is going to be more like

`{class org.springframework.aop.config.SimpleBeanFactoryAwareAspectInstanceFactory=SimpleBeanFactoryAwareAspectInstanceFactory}`

yschimke (approved):
nothing blocking

okhttp/src/jvmTest/kotlin/okhttp3/internal/TagsTest.kt
squarejesse: For now I’m just going for 100% compatibility with what `Request.toString()` was doing, which was ultimately calling `Map.toString()`. I want to introduce this new feature without changing _that_ behavior. And also we should change that behavior to be nicer.

okhttp/src/jvmTest/kotlin/okhttp3/internal/TagsTest.kt
```diff
@@ -0,0 +1,160 @@
...
+    val tags = EmptyTags.plus(String::class, "hello")
+    assertThat(tags[String::class]).isEqualTo("hello")
+  }
+
+  @Test
+  fun multipleElements() {
+    val tags =
+      EmptyTags
+        .plus(String::class, "hello")
+        .plus(Integer::class, 5 as Integer)
```
jisungbin, line 41: Hi, I’m reviewing the code for learning purposes and have a small question.

Is the reason for using `java.lang.Integer` instead of `kotlin.Int` to verify whether a Java class is correctly represented as a KClass?

yschimke: I think kotlin.Int is like java.lang.Integer.TYPE (primitive), but I guess used on generic methods it's force to java.lang.Integer.

```
inline fun <reified T> printType() {
  println(T::class.java)
}

fun main(args: Array<String>) {
  printType<kotlin.Int>()
  printType<kotlin.Int?>()
  printType<java.lang.Integer>()

  println(Int::class.java)
  println(Integer::class.java)
}

class java.lang.Integer
class java.lang.Integer
class java.lang.Integer
int
class java.lang.Integer
```

So best to avoid forcing that

```
kotlin.Int::class.java.cast(2) // java.lang.ClassCastException: Cannot cast java.lang.Integer to int
```

jisungbin: Ah, I see!

Thanks for pointing out this interesting aspect of Kotlin (probably on the JVM side).

PR #9168, opened 2025-10-25
Add tags to calls, including computable tags

swankjesse:
Originally I was planning on implementing tags on
Request only, but that design broke as soon as any
interceptor used Request.newBuilder() to create a
copy of the Request - the instance that received the
tags might not be the instance that needed them.

Instead I'm introducing this behavior on the Call
object, which behaves less like a value. I expect
that users' expected lifetimes of tags will work
naturally this way.

It is an API hazard that we have tags on both
Requests and Calls. Perhaps if given the opportunity
to do it over, I'd omit support for tags on Requests.

### DIFF
--- a/okhttp/api/jvm/okhttp.api
+++ b/okhttp/api/jvm/okhttp.api
@@ -126,6 +126,10 @@ public abstract interface class okhttp3/Call : java/lang/Cloneable {
 	public abstract fun isCanceled ()Z
 	public abstract fun isExecuted ()Z
 	public abstract fun request ()Lokhttp3/Request;
+	public abstract fun tag (Ljava/lang/Class;)Ljava/lang/Object;
+	public abstract fun tag (Ljava/lang/Class;Lkotlin/jvm/functions/Function0;)Ljava/lang/Object;
+	public abstract fun tag (Lkotlin/reflect/KClass;)Ljava/lang/Object;
+	public abstract fun tag (Lkotlin/reflect/KClass;Lkotlin/jvm/functions/Function0;)Ljava/lang/Object;
 	public abstract fun timeout ()Lokio/Timeout;
 }
 
--- a/okhttp/src/commonJvmAndroid/kotlin/okhttp3/Call.kt
+++ b/okhttp/src/commonJvmAndroid/kotlin/okhttp3/Call.kt
@@ -15,6 +15,7 @@
  */
 package okhttp3
 
+import kotlin.reflect.KClass
 import okio.IOException
 import okio.Timeout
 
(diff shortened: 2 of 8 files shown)


okhttp/src/commonJvmAndroid/kotlin/okhttp3/Call.kt
```diff
@@ -87,9 +88,79 @@ interface Call : Cloneable {
...
+   * may be computed. But only one value will be inserted, and that inserted value will be returned
+   * to all callers.
+   *
+   * If computing multiple values is problematic, use an appropriate concurrency mechanism in your
+   * [computeIfAbsent] implementation. No locks are held while calling this function.
+   */
+  fun <T : Any> tag(
+    type: KClass<T>,
+    computeIfAbsent: () -> T,
+  ): T
```
swankjesse, line 121: We have this law in Canada where packaged goods need to have both French and English on the label. 

image

Offering both a Kotlin `KClass` version and a Java `Class` version of these APIs has the same vibes. I don’t want one set of users to feel diminished vs. the other.

swankjesse: Oh and if it’s unclear, I believe the french word for Nacho is Nacho. @oldergod can confirm.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Call.kt
```diff
@@ -87,9 +88,79 @@ interface Call : Cloneable {
    */
   fun timeout(): Timeout
 
+  /**
+   * Returns the tag attached with [type] as a key, or null if no tag is attached with that key.
+   *
+   * The tags on a call are seeded from the [request tags][Request.tag]. This set will grow if new
+   * tags are computed.
+   */
+  fun <T : Any> tag(type: KClass<T>): T?
```
yschimke, line 97: Can this form be a reified type without a param?

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Call.kt
```diff
@@ -87,9 +88,79 @@ interface Call : Cloneable {
...
+  fun <T : Any> tag(
+    type: Class<T>,
+    computeIfAbsent: () -> T,
+  ): T
+
   /**
    * Create a new, identical call to this one which can be enqueued or executed even if this call
    * has already been.
+   *
+   * The tags on the returned call will equal the tags as on [request]. Any tags that were computed
```
yschimke, line 143: I like this.

okhttp/api/jvm/okhttp.api
```diff
@@ -126,6 +126,10 @@ public abstract interface class okhttp3/Call : java/lang/Cloneable {
 	public abstract fun isCanceled ()Z
 	public abstract fun isExecuted ()Z
 	public abstract fun request ()Lokhttp3/Request;
+	public abstract fun tag (Ljava/lang/Class;)Ljava/lang/Object;
```
yschimke, line 129: This seems like a disruptively breaking change. Does the interface default mechanism help?

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Call.kt
swankjesse: If I do it as an extension function. On an interface it’s not possible ’cause reified requires inline.

okhttp/api/jvm/okhttp.api
swankjesse: Agreed that it’s disruptive.

The normal use-case of this interface is that it’s implemented by OkHttp, and called by our users. In this normal case there’s no problem, as the implementations include the new functions.

The decorator use case is dangerous: users create their own `Call` implementations that delegate to our implementation, and users’ decorating implementations won’t implement these functions. But I also don’t expect such users to introduce new calls to `tag()` before they re-compile their `Call` implementations.

The most most dangerous case is when a third-party library decorates `Call`, and users introduce new calls to `tag()` without updating such a library. I will document this risk in our changelog!

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Call.kt
yschimke: This surprised me when I went to use it.

swankjesse: The lack of a reified inline function?

yschimke: Yep, needing to pass the class param

PR #9170, opened 2025-10-29
Be more strict on event lifecycle for upgrades

swankjesse:


swankjesse:
Agreed. We could do it with fewer! But I  want something that's easy to understand

PR #9174, opened 2025-10-31
Use Burst in DiskLruCacheTest (and 7 others)

swankjesse:


swankjesse:
(I wanna start using Burst for event listeners, but it seemed appropriate to migrate other stuff to it first.)

okhttp/src/jvmTest/kotlin/okhttp3/SocketChannelTest.kt
```diff
@@ -15,6 +15,8 @@
  */
 package okhttp3
 
+`import app.cash.burst.Burst
```
JakeWharton:
```suggestion
import app.cash.burst.Burst
```

okhttp/src/jvmTest/kotlin/okhttp3/SocketChannelTest.kt
```diff
@@ -93,9 +93,25 @@ class SocketChannelTest {
...
+  @Test
+  fun testHttp(
+    socketMode: SocketMode = burstValues(Channel, Standard)
+  ) {
+    testConnection(socketMode)
+  }
+
+  @Test
+  fun testHttps(
+    provider: Provider = Provider.JSSE,
```
JakeWharton, line 103: Not varying this?

JakeWharton: Oh it's probably an enum

swankjesse: Yup.

yschimke (approved):

PR #9178, opened 2025-11-04
Run all the tests in CI

swankjesse:
We haven't been. I would have expected ./gradlew test to run everything, but it doesn't. We need to explicitly run jvmTest to get all of our tests to run.


JakeWharton:
I think these should run `check` which is the aggregating task. For JVM projects it will run `test`. For multiplatform projects it will run `allTests` which will run `jvmTest`.

I also think all of the labels need to die. CI should be running everything all the time.

JakeWharton (approved):

yschimke:
I think most label conditionals were
removed. 

But also I think flakes are much lower. There was a point where each individual JVM job had 10-20% chance of failing at least one flake test.

yschimke (approved):

PR #9179, opened 2025-11-04
Fix some failing tests that CI hasn't been running

swankjesse:
Thankfully there's nothing here that indicates we've shipped a bug to our users.


yschimke (approved):

yschimke:
This isn't the first time, and as there is a good chance it was me, I'm wondering if we can test tests test things correctly.

PR #9181, opened 2025-11-04
Call.addEventListener

squarejesse:
Use this to install listeners after-the-fact, such as in Interceptors or other EventListeners.


okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/connection/RealCall.kt
```diff
@@ -126,6 +126,10 @@ class RealCall(
 
   override fun timeout(): Timeout = timeout
 
+  override fun addEventListener(eventListener: EventListener) {
+    this.eventListener += eventListener
```
JakeWharton: Does this need a lock or CAS loop? If not, docs about thread (un)safety?

squarejesse: Oooh good catch. Lemme make it thread-safe.

Issue #700, opened 2014-04-17
Rewrite Apache HttpClient shim to be v2-API based


JakeWharton:
The current implementation relies on `HttpUrlConnection` for its implementation. It should be using `Request` and `Response` directly (which it actually more closely maps to).

Issue #163, opened 2013-05-04
All of the threads should be named


swankjesse:
Currently ConnectionPool#executorService doesn't name its threads. Possibly others have this problem.

PR #9188, opened 2025-11-11
Fix TLS handshake failures on Android API 24 and 25

### LINKED ISSUE #8678: OkHttpWebsocketSession does not complete the `closeReason` future when closing the websocket in absence of connectivity
Hello! 👋 
I encountered this issue while doing some tests for the reliability of my ktor-based client application. 

**Reproduction steps**
- Initialize a ktor application with websocket and where the built-in ping-pong mechanism is disabled.
- Launch 2 coroutines: Coroutine 1 communicates through websocket with a server external to the machine, Coroutine 2 awaits the completion of the `closeReason` Future through `DefaultWebSocketSession.closeReason`.
- Disconnect the machine from the internet.
- Close the websocket.

**Actual Behaviour**
Coroutine 1 terminates while Coroutine 2 hangs forever.

**Expected Behaviour**
Coroutine 2 should also terminate, as `closeReason` should complete.

**Analysis**
Looking at the code, my understanding of the reason of this behaviour is that, in the `OkHttpWebsocketSession` class, the `closeReason` future is completed only in one of these 3 method: `onClosing`, `onClosed`, or `onFailure`. 

- `onFailure` is not called by the `RealWebSocket` implementation when the connection drops because we disabled the built-in ping/pong mechanism in favour of a custom one. 
- `onClosing` and `onClose` are not called when the websocket is closed unless a control frame (`OPCODE_CONTROL_CLOSE`) is sent by the server, event that cannot happen under these conditions.

**Additional Information**
On the other hand, the CIO engine manages to complete the `closeReason` future under the same assumptions.

**Question**
Is this a design choice or just a bug due to a very niche condition?

MohitAnuragi:
## Problem
OkHttp 4.10+ fails on Android API 24 & 25 due to buggy TLS 1.3 implementation.

## Solution
- Detect Android API 24-25 using platform checks
- Fall back to TLS 1.2 for these specific versions
- Maintain TLS 1.3 for all other platforms

## Changes
- Added `isAndroidApi24Or25()` detection in `Platform.kt`
- Created `compatibleTlsSpec()` in `ConnectionSpec.kt` 
- Updated TLS connection logic in `RealConnection.kt`

Fixes #8678

### DIFF
--- a/okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/platform/Platform.kt
+++ b/okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/platform/Platform.kt
@@ -236,5 +236,25 @@ open class Platform {
       }
       return result.readByteArray()
     }
+    @JvmStatic
+    fun isAndroidApi24Or25(): Boolean {
+    return when (platformName) {
+        "android" -> {
+            val sdkVersion = getAndroidSdkVersion()
+            sdkVersion == 24 || sdkVersion == 25
+        }
+        else -> false
+    }
+}
+
+private fun getAndroidSdkVersion(): Int {
+    return try {
+        Class.forName("android.os.Build\$VERSION")
+            .getField("SDK_INT")
+            .get(null) as Int
+    } catch (e: Exception) {
+        -1
+    }
+}
   }
 }
--- a/okhttp/src/commonJvmAndroid/kotlin/okhttp3/ConnectionSpec.kt
+++ b/okhttp/src/commonJvmAndroid/kotlin/okhttp3/ConnectionSpec.kt
@@ -379,5 +379,21 @@ class ConnectionSpec internal constructor(
     /** Unencrypted, unauthenticated connections for `http:` URLs. */
     @JvmField
     val CLEARTEXT = Builder(false).build()
+
+   @JvmStatic
+    fun compatibleTlsSpec(): ConnectionSpec {
+        return if (Platform.isAndroidApi24Or25()) {
+            // Force TLS 1.2 for buggy Android 7.0/7.1
+            Builder(true)
+                .tlsVersions(TlsVersion.TLS_1_2)
+                .cipherSuites(*APPROVED_CIPHER_SUITES.toTypedArray())
+                .supportsTlsExtensions(true)
+                .build()
+        } else {
+            COMPATIBLE_TLS
+        }
+    }
+
   }
+
 }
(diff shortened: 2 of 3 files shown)


yschimke:
Is there an upstream bug covering the issues with TLS 1.3 in Android? 

The linked bug doesn't seem to show any clear path to this conclusion, so i'd like to understand the confidence in this fix

MohitAnuragi:
Hi @yschimke ,
Thank you for the feedback!, I’m still actively exploring the TLS 1.3 issue on Android API 24–25, so I cannot confidently say that the proposed PR should be merged yet. However, I’ve gathered several upstream references that indicate TLS 1.3 instability on older Android versions:
1. Conscrypt Issue [#700 ](https://github.com/google/conscrypt/issues/700)discussing TLS 1.3 handshake failures on Android 6–7.1 (API 23–25).
 2. A similar real-world handshake issue reported in Grocy Android Issue [#163](https://github.com/patzly/grocy-android/issues/163) on Android 7.
 3. Android documentation noting that TLS 1.3 is officially enabled and stable only from Android 10 (API 29).
      https://developer.android.com/about/versions/10/behavior-changes-all#security

yschimke:
> Android documentation noting that TLS 1.3 is officially enabled and stable only from Android 10

Seems seems convincing, but I don't see mention about stable, only being default enabled.

> In Android 10 and higher, TLS 1.3 is enabled by default for all TLS connections.

yschimke:
https://github.com/patzly/grocy-android/issues/163 is the opposite, it fails because it wasn't enabled. And is fixed if you enable TLSv1.3.

yschimke:
Same for https://github.com/google/conscrypt/issues/700

> Your best option is probably to include OkHttp in your application rather than relying on the platform's HttpsURLConnection implementation.

So it's literally suggested to use OkHttp to enable TLSv1.3.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/platform/Platform.kt
```diff
@@ -236,5 +236,25 @@ open class Platform {
       }
       return result.readByteArray()
     }
+    @JvmStatic
+    fun isAndroidApi24Or25(): Boolean {
+    return when (platformName) {
```
yschimke, line 241: This probably isn't the right place to check this.

- We have an android source set with android platforms that can compile against the Android SDK
- If an Android app bundles Conscrypt or BouncyCastle, shouldn't TLSv1.3 be supported?

yschimke:
Overall we would need a failing test, that would fail on API 24 and 25 but passes with this fix.

But I'm not clear this change is justified. 

Please provide either
a) clearer documented evidence this is the right change
b) a repro, or logs showing it failing

Then we can reopen and address the structure of the PR

PR #9202, opened 2025-11-18
Make the publish action work like Okio's

squarejesse:


.github/workflows/build.yml
```diff
@@ -14,37 +14,6 @@ env:
...
-        uses: actions/setup-java@v5
-        with:
-          distribution: 'zulu'
-          java-version: 17
-
-      - name: Setup Gradle
-        uses: gradle/actions/setup-gradle@v5
-
-      - name: Upload Artifacts
-        run: ./gradlew clean publish --stacktrace
```
JakeWharton, line 41: Still need this for snapshots

yschimke (approved):

.github/workflows/build.yml
swankjesse: oooh good call. Fixed.

cmr13locotes-source: Mejor chinga. Pero atu madre . Solo joder saben

PR #9214, opened 2025-11-29
Headers operator functions

MukjepScarlet:
- Headers + Headers (Java: combine)
- Headers.Builder += Headers
- Headers.Builder -= String


yschimke:
I do wonder whether making this easy is hiding the actual complexity.

I feel like merging headers should probably be aware of rules for how to resolve duplicate keys. 

Or at least documenting that this is the behaviour, and duplicates should be removed before combine if they would make an invalid request

MukjepScarlet:
Because HTTP headers can be treated as multimap, so I think it's acceptable - or we should create something like `addIfAbsent`.

MukjepScarlet:
```kotlin
fun Headers.Builder.addIfAbsent(name: String, value: String): Headers.Builder
fun Headers.Builder.addAllIfAbsent(...): Headers.Builder

fun Headers.merge // merge: from Map, withFallback: from typesafe's Config
```

swankjesse:
What's your use case?

MukjepScarlet:
For example
```kotlin
val DEFAULT_UA = headersOf("User-Agent", "Chrome 120...")

Request.Builder().url(...).headers(DEFAULT_UA + headersOf("Authentication", "...")).build()
```

PR #9216, opened 2025-11-29
Add empty check for Headers builder

MukjepScarlet:


swankjesse:
We don't accept optimizations unless there's a motivating performance problem. In this case I don't think the optimized code is a net improvement, as it's slowing down the case where the headers are not empty.

PR #9220, opened 2025-12-04
Add compatibility with building under KGP 2.3

JakeWharton:


JakeWharton:
Hmm I thought Kotlin tasks inherited this now. Maybe that's only AGP 9.

PR #9247, opened 2026-01-10
Fix RouteFailureTest and use RepeatedTest

yschimke:
Updated tests to use @RepeatedTest annotation for consistency and performance.

Also fixes a cause of flakiness

full disclosure: fix suggested by AI from https://github.com/square/okhttp/pull/9246
but seems appropriate.


yschimke:
@swankjesse requesting post review, test only flake reduction.

swankjesse:
LGTM

PR #9248, opened 2026-01-10
Fix CacheTest on Windows

yschimke:
cache.directory.path.toPath() fights the emulateUnix.

Also avoid 10s delays on failure.


yschimke:
@swankjesse test only flake fix. Requesting post review

PR #9249, opened 2026-01-11
Avoid ServerTruncatesRequestTest windows failure

yschimke:
Avoids a permanent fails on windows, which seems unavailable due to socket differences.

```
    // "java.net.SocketException: Socket closed" thrown reading response after
    // "java.net.SocketException: Connection reset by peer" writing request
    assumeNotWindows()
```


yschimke:
@swankjesse requesting post review. Accepting defeat on supporting this with windows sockets.

PR #9250, opened 2026-01-11
Prevent flaky tests by adding sleep in HttpOverHttp2Test

yschimke:
Added a sleep to avoid flaky race conditions in test.


yschimke:
Testing on https://github.com/square/okhttp/pull/9246

yschimke:

This is not ideal, but I'd rather have it in and ensure it's not flaky when we remove it.

PR #9261, opened 2026-01-21
Fix isAscii returning true for malformed surrogates

### LINKED ISSUE #6357: isAscii is wrong for malformed surrogates
I just realized we map invalid UTF-16 chars to `?`, so my length trick is incorrect. Fix is to iterate.

https://github.com/square/okio/blob/master/okio/src/commonMain/kotlin/okio/Utf8.kt#L102

one-kash:
## Summary

The `isAscii()` extension function in `OkHostnameVerifier` used `length == utf8Size()` to determine if a string was ASCII. This was incorrect because Okio's `utf8Size()` maps malformed surrogates (unpaired high/low surrogates) to the replacement character `?` which is 1 byte in UTF-8.

This caused strings like `"\uD800.com"` (unpaired high surrogate) to incorrectly report as ASCII since:
- `length = 5` (one char for surrogate + 4 for ".com")
- `utf8Size() = 5` (surrogate mapped to '?' = 1 byte + 4 for ".com")

The fix replaces the comparison with a simple character iteration check: `all { it.code <= 127 }`

## Changes

- **OkHostnameVerifier.kt**: Changed `isAscii()` to iterate through characters and check code points
- **HostnameVerifierTest.kt**: Added test for malformed surrogates

## Test Plan

- [x] Added `malformedSurrogatesAreNotAscii` test that verifies unpaired high/low surrogates are rejected
- [x] All existing `HostnameVerifierTest` tests continue to pass

Fixes #6357


okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/tls/OkHostnameVerifier.kt
```diff
@@ -94,7 +93,7 @@ object OkHostnameVerifier : HostnameVerifier {
     }
 
   /** Returns true if the [String] is ASCII encoded (0-127). */
-  private fun String.isAscii() = length == utf8Size().toInt()
+  private fun String.isAscii() = all { it.code <= 127 }
```
yschimke, line 98: I suspect we want to avoid Iterable.all to avoid an allocation.

okhttp/src/jvmTest/kotlin/okhttp3/internal/tls/HostnameVerifierTest.kt
```diff
@@ -803,6 +803,33 @@ class HostnameVerifierTest {
...
+        .build()
+    val session = session(heldCertificate.certificatePem())
+
+    // Unpaired high surrogate - should not match any hostname
+    assertThat(verifier.verify("\uD800.com", session)).isFalse()
+    assertThat(verifier.verify("foo\uD800.com", session)).isFalse()
+
+    // Unpaired low surrogate - should not match any hostname
+    assertThat(verifier.verify("\uDC00.com", session)).isFalse()
+    assertThat(verifier.verify("foo\uDC00.com", session)).isFalse()
```
yschimke, line 827: Would any of these have matched before the change? is there some case where `length == utf8Size().toInt()` would match because of offsetting issues, longer lower case, and malformed surrogate.

PR #9263, opened 2026-01-24
Switch to Dokka v2

yschimke:
Avoid the nagging warnings, now that Dokka is stable in 2.1.0

While it fails with both dokka and java module tests, only enable these when needed.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -38,7 +38,8 @@ buildscript {
   }
 }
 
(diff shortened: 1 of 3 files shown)


yschimke:
Raised https://github.com/gradlex-org/extra-java-module-info/issues/221

yschimke:

There is decent chance it might affect release steps, but seems to be generating ok, and allows me to progress with some other build cleanup.

PR #9264, opened 2026-01-25
Update to AGP 9

yschimke:
First version with compat workarounds in place.

Will cause some issues running in Intellij, but support coming in Q1. And we are already on 8.13.1 which is not compatible.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -300,7 +300,7 @@ subprojects {
 
 /** Configure publishing and signing for published Java and JavaPlatform subprojects. */
 subprojects {
-  plugins.withId("com.vanniktech.maven.publish.base") {
+  if (plugins.hasPlugin("com.vanniktech.maven.publish.base")) {
     if (dokkaBuild) {
       apply(plugin = "org.jetbrains.dokka")
 
@@ -334,10 +334,12 @@ subprojects {
     }
 
     // Only enable lint checks for published modules
-    apply(plugin = "com.android.lint")
+    if (project.name != "okhttp") {
+      apply(plugin = "com.android.lint")
 
-    dependencies {
-      "lintChecks"(rootProject.libs.androidx.lint.gradle)
+      dependencies {
+        "lintChecks"(rootProject.libs.androidx.lint.gradle)
+      }
     }
 
     configure<MavenPublishBaseExtension> {


yschimke:
```

Could not determine the dependencies of task ':native-image-tests:lintReportJvm'.
> Could not create task ':native-image-tests:lintAnalyzeJvmMain'.
   > DefaultTaskContainer#NamedDomainObjectProvider.configure(Action) on task set cannot be executed in the current context.
```

yschimke:
```
Could not determine the dependencies of task ':native-image-tests:nativeTestCompile'.
> Could not resolve all dependencies for configuration ':native-image-tests:nativeImageTestClasspath'.
gradle/actions: Writing build results to /home/runner/work/_temp/.gradle-actions/build-results/__run_2-1769338155250.json

[Incubating] Problems report is available at: file:///home/runner/work/okhttp/okhttp/build/reports/problems/problems-report.html
   > Could not create task ':native-image-tests:generateJvmMainLintModel'.
      > DefaultTaskContainer#NamedDomainObjectProvider.configure(Action) on task set cannot be executed in the current context.
```

yschimke:

We may need to revert if it causes problems, but our gradle build needs some love, and it seems wise to do it against the latest.

android-test-app/build.gradle.kts
```diff
@@ -23,16 +22,13 @@ android {
     sourceCompatibility(JavaVersion.VERSION_11)
   }
 
-  kotlinOptions {
```
yschimke, line 26: I need to confirm this is just because AGP is now 17+, but the artifacts still work for older android clients...

build.gradle.kts
```diff
@@ -309,7 +300,7 @@ subprojects {
 
 /** Configure publishing and signing for published Java and JavaPlatform subprojects. */
 subprojects {
-  plugins.withId("com.vanniktech.maven.publish.base") {
+  if (plugins.hasPlugin("com.vanniktech.maven.publish.base")) {
```
swankjesse, line 303: Why this change? I am under the impression that `withId` does the same thing, but it’s robust to plugin installation order

yschimke: Damn, I should have commented. I'll try reverting this and see why it was breaking.

PR #9265, opened 2026-01-25
Fix JavaHttpClientTest

yschimke:
Don't check on content length, not critical.


yschimke:
Fix failure on main, for non critical test

https://github.com/square/okhttp/actions/runs/21332707775/job/61399879471

PR #9292, opened 2026-02-01
Adopt a full modern gradle build

yschimke:
Switch to a modern gradle structure.

This went too far, but looks promising to adopt and quickly iterate on.

There is some risk of non-obvious changes, to follow up putting more guardrails in place before next release.


build-logic/src/main/kotlin/okhttp.quality-conventions.gradle.kts
```diff
@@ -0,0 +1,85 @@
...
+    this.sourceSets = listOf(main)
+  }
+
+  signatures = androidSignature + jvmSignature
+  failWithoutSignatures = false
+}
+
+dependencies {
+  // Logic for signatures should be moved to the applying module or configured via extension
+  // For now, we'll keep the standard ones and allow modules to add more
```
swankjesse, line 69: 👍🏻

gradle/libs.versions.toml
```diff
@@ -1,119 +1,175 @@
...
-#noinspection UnusedVersionCatalogEntry
+graalvm-plugin = "0.11.4"
+guava = "33.4.0-jre"
+hamcrest-library = "3.0"
+http-client5 = "5.5.2"
+jetty-client = "9.4.58.v20250814"
+jlink = "0.7"
+jetbrains-annotations = "26.0.2-1"
+jnr-unixsocket = "0.38.24"
+jsoup = "1.22.1"
```
swankjesse, line 35: Seems like most of these are referenced only once, not saving much? Though I do enjoy uniformity

okcurl/build.gradle.kts
```diff
@@ -1,48 +1,60 @@
...
+  id("okhttp.jvm-conventions")
+  id("okhttp.quality-conventions")
+  id("okhttp.testing-conventions")
   id("com.gradleup.shadow")
 }
 
-val testJavaVersion = System.getProperty("test.java.version", "21").toInt()
+tasks.withType<KotlinCompile> {
+  compilerOptions {
+    jvmTarget.set(JvmTarget.JVM_17)
```
swankjesse, line 19: Where did this come from?

build.gradle.kts
```diff
@@ -1,394 +1,22 @@
...
-  }
-}
-
-plugins.withId("org.jetbrains.kotlin.jvm") {
-  val test = tasks.named("test")
-  tasks.register("jvmTest") {
-    description = "Get 'gradlew jvmTest' to run the tests of JVM-only modules"
-    dependsOn(test)
-  }
+plugins {
```
swankjesse, line 1: Nice

gradle/libs.versions.toml
yschimke: Yep - we had inconsistencies, and it was hard to know where to look for versions.

okcurl/build.gradle.kts
yschimke: From here https://github.com/square/okhttp/blob/master/build.gradle.kts#L181

It feels a lot cleaner for overrides now.

build.gradle.kts
yschimke: if anything I went overbooard. The dokka config could have stayed.

gradle/libs.versions.toml
yschimke: if anything, I'd like to have a clear pattern and apply it even more consistently.

build-logic/src/main/kotlin/okhttp.quality-conventions.gradle.kts
yschimke: I ended up reconfiguring the okcurl one, to only use jvm. But we can us this as a starting point to define how this should work.

gradle/libs.versions.toml
swankjesse: 👍🏻

PR #9295, opened 2026-02-02
Fix flaky platformTrustedCertificates test for non-Entrust platforms  

### LINKED ISSUE #9288: Consider checking for a non-Entrust certificate issuer in HandshakeCertificatesTest
HandshakeCertificatesTest has an assertion that checks fothat `CN=Entrust` is present as a certificate issuer.

https://github.com/square/okhttp/blob/861b1f7c3bee82d9360c8b4379e420491db3189b/okhttp-tls/src/test/java/okhttp3/tls/HandshakeCertificatesTest.kt#L165-L167

`CN=Entrust` may not always be present, see e.g. https://security.googleblog.com/2024/06/sustaining-digital-certificate-security.html

Amir-yazdanmanesh:
The `platformTrustedCertificates` test assumes Entrust certificates are always present in the platform's trusted certificate store. However, following Google's decision 
  to distrust Entrust certificates in Chrome (announced in 2024), newer platforms may not include Entrust as a trusted CA.                                                 
                                                                                                                                                                           
  This change updates the test to check for any major certificate authority instead of specifically Entrust, which aligns with the test's original intent: verifying that  
  platform-trusted certificates include at least one major Internet certificate issuer.                                                                                    
                                                                                                                                                                           
  Major CAs now checked:                                                                                                                                                   
  - DigiCert                                                                                                                                                               
  - Let's Encrypt / ISRG                                                                                                                                                   
  - GlobalSign                                                                                                                                                             
  - Comodo / Sectigo                                                                                                                                                       
  - GeoTrust                                                                                                                                                               
  - Entrust (still included as fallback)                                                                                                                                   
                                                                                                                                                                           
  Fixes #9288

### DIFF
--- a/okhttp-tls/src/test/java/okhttp3/tls/HandshakeCertificatesTest.kt
+++ b/okhttp-tls/src/test/java/okhttp3/tls/HandshakeCertificatesTest.kt
@@ -163,8 +163,18 @@ class HandshakeCertificatesTest {
         .toSet()
 
     // It's safe to assume all platforms will have a major Internet certificate issuer.
+    val majorIssuers = listOf(
+      "DigiCert",
+      "Let's Encrypt",
+      "ISRG",  // Internet Security Research Group (Let's Encrypt parent)
+      "GlobalSign",
+      "Comodo",
+      "Sectigo",
+      "GeoTrust",
+      "Entrust",
+    )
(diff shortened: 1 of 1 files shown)


yschimke:
I'll fix formatting in a follow up

PR #9310, opened 2026-02-04
Enable Gradle Configuration Cache — by refactoring OSGi/BND plugin

yschimke:
This commit enables Gradle's Configuration Cache feature — which was blocked by the BND (OSGi bundle) plugin's incompatibility.

The key changes are:
- A new `BndBuildAction` class is introduced. This class is a static, configuration-cache-compatible replacement for the BND plugin's default build action.
- The `Osgi.kt` build logic is refactored to use this new `BndBuildAction` instead of directly calling the BND plugin's action, which was not serializable.
- The `STABLE_CONFIGURATION_CACHE` feature preview is enabled in `settings.gradle.kts`.
- Gradle properties are updated to enable the configuration cache (`org.gradle.configuration-cache=true`).

### DIFF
--- a/build-logic/src/main/kotlin/okhttp.quality-conventions.gradle.kts
+++ b/build-logic/src/main/kotlin/okhttp.quality-conventions.gradle.kts
@@ -59,8 +59,9 @@ configure<AnimalSnifferExtension> {
 
 // Default to only published modules
 project.tasks.withType<AnimalSniffer> {
+  val hasMavenPublish = project.extensions.findByType<MavenPublishBaseExtension>() != null
   onlyIf {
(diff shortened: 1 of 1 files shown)


yschimke:
@swankjesse you touched this last, so may have a good idea. BndBuildAction edits were AI assisted, so I'm somewhat relying on tests.

yschimke:
@swankjesse requesting post review - want to have a strict build while I make any follow up changes.

decent chance I need to revert this or fix forward based on your bnd review. But tests are passing.

yschimke:
I'm more hateful of java modules

PR #9314, opened 2026-02-05
More config caching fixes

yschimke:
To address

```
> Configuration cache state could not be cached: field `__javaCompiler__` of task `:logging-interceptor:compileJava9Java` of type `org.gradle.api.tasks.compile.JavaCompile`: error writing value of type 'org.gradle.api.internal.provider.DefaultProperty'
```


yschimke:
@swankjesse follow up to previous gradle fixes

PR #9322, opened 2026-02-08
Apply kotlin version compat

### LINKED ISSUE #9032: Incompatible Kotlin Version Error with OkHttp v5.1.0

yschimke:
Copy the approach from Okio.

https://github.com/square/okio/pull/1770

Closes #9032

### DIFF
--- a/build-logic/src/main/kotlin/okhttp.jvm-conventions.gradle.kts
+++ b/build-logic/src/main/kotlin/okhttp.jvm-conventions.gradle.kts
@@ -48,6 +48,6 @@ tasks.withType<KotlinCompile> {
 }
 
 extensions.configure<TapmocExtension> {
-  // Fail the build if any api dependency exposes incompatible Kotlin metadata, Kotlin stdlib or Java bytecode version.
+  // Fail the build if any api dependency exposes incompatible Kotlin metadata, Kotlin stdlib, or Java bytecode version.
   checkDependencies()
 }


yschimke:
@JakeWharton is this correct for matching Okio?

build-logic/src/main/kotlin/okhttp.jvm-conventions.gradle.kts
```diff
@@ -41,3 +46,8 @@ tasks.withType<KotlinCompile> {
     )
   }
 }
+
+extensions.configure<TapmocExtension> {
+  // Fail the build if any api dependency exposes incompatible Kotlin metadata, Kotlin stdlib or Java bytecode version.
```
JakeWharton:
```suggestion
  // Fail the build if any api dependency exposes incompatible Kotlin metadata, Kotlin stdlib, or Java bytecode version.
```
One of my hills...

gradle/libs.versions.toml
```diff
@@ -64,6 +64,10 @@ startup-runtime = "1.2.0"
 testcontainers = "1.21.4"
 zstd-kmp-okio = "0.4.0"
 
+# Set to lower version than KGP version to maximize compatibility
+# Default to matching okio
+kotlinCoreLibrariesVersion = "2.1.21"
```
JakeWharton, line 69: Is there a reason to define this here rather than at the sole usage site?

JakeWharton (approved):
Looks good to me!

gradle/libs.versions.toml
yschimke: Consistency of versions and copying okio setup?

swankjesse (commented):
LGTM

PR #9338, opened 2026-02-21
[speculative/not to land] Add pcap logging for decorated sockets

yschimke:
Likely this goes nowhere.

Adds a RecordingSocketFactory that decorates a TCP socket and captures events as Pcap or a sealed event trait. The pcap file is produced without running anything external like tshark.

A potential debugging tool...

- after the fact flaky tests
- visibility into HTTP/2 frames
- confirming behaviour of another speculative idea - mock sockets

The last point is why I'm exploring this. See whether we can run some category of tests without real I/O and flakiness. Run in near instant time.


yschimke:
image

Issue #9335, opened 2026-02-19
Cache entries are not stored correctly on Windows, resulting in excessive requests


Nohus:
## Scenario

A server returns a resource that can be cached for 50 seconds, and an application requests it 100 times, once every second.

## Expectation

1. First request is sent, server returns 200, OkHttp stores the response in the cache.
2. 49 next requests are served from the cache.
3. Cache entry has expired, OkHttp sends a conditional request. Server responds with a 304 Not Modified. OkHttp updates the cache entry.
4. 49 next requests are served from the cache.

Total requests sent to the network: 2.

This correctly happens on Linux and macOS.

## The bug

Here is what happens on Windows:

1. First request is sent, server returns 200, OkHttp stores the response in the cache.
2. 49 next requests are served from the cache.
3. Cache entry has expired, OkHttp sends a conditional request. Server responds with a 304 Not Modified. OkHttp **fails to update** the cache entry.
4. OkHttp sends 49 conditional requests, never succeeds in updating the cache and keeps the stale response in the cache forever, resulting in needing a network request every time going forward.

Total requests sent to the network: 51

## Reproducer

The below test succeeds on Linux but fails on Windows.

```kotlin
import okhttp3.Cache
import okhttp3.OkHttpClient
import okhttp3.Request
import java.io.File
import java.io.IOException
import java.net.HttpURLConnection
import java.text.DateFormat
import java.text.SimpleDateFormat
import java.util.Date
import java.util.Locale
import java.util.TimeZone
import java.util.concurrent.TimeUnit
import mockwebserver3.MockResponse
import mockwebserver3.MockWebServer

fun main() {
  val server = MockWebServer()
  server.start()
  // First response, will return a 200 and immediately expire
  server.enqueue(
    MockResponse
      .Builder()
      .addHeader("Last-Modified: " + formatDate(0, TimeUnit.SECONDS))
      .addHeader("Cache-Control: max-age=0")
      .body("A")
      .build(),
  )
  // Second response, will return a 304 and set a new longer age
  server.enqueue(
    MockResponse
      .Builder()
      .addHeader("Cache-Control: max-age=30")
      .addHeader("Allow: GET, HEAD")
      .code(HttpURLConnection.HTTP_NOT_MODIFIED)
      .build(),
  )
  // Third response, should never be requested
  server.enqueue(
    MockResponse
      .Builder()
      .body("B")
      .build(),
  )

  val cacheDirectory = File("cache-test")
[...]

PR #9339, opened 2026-02-22
Run and Fix CacheTest emulating windows

yschimke:
Investigate and fix issues affecting windows.
From #9335

### DIFF
--- a/okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/cache/DeferredForwardingSource.kt
+++ b/okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/cache/DeferredForwardingSource.kt
@@ -1,0 +1,39 @@
+/*
+ * Copyright (c) 2025 Block, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okhttp3.internal.cache
+
+import okio.Buffer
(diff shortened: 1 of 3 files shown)


yschimke:
The fix isn't trivial.  cc @swankjesse 

We get here

```
        // Update the cache after combining headers but before stripping the
        // Content-Encoding header (as performed by initContentStream()).
        cache!!.trackConditionalCacheHit()
        cache.update(cacheResponse, response)
```

or more precisely

```
  internal fun update(
    cached: Response,
    network: Response,
  ) {
    val entry = Entry(network)
    val snapshot = (cached.body as CacheResponseBody).snapshot <-- we are still planning to use the cache response body
    var editor: DiskLruCache.Editor? = null
    try {
      editor = snapshot.edit() ?: return // edit() returns null if snapshot is not current.
      entry.writeTo(editor)
      editor.commit()
    } catch (_: IOException) {
      abortQuietly(editor)
    }
  }
```

we got the cache response body from, so lockingSourceCount is > 0

```
    private fun newSource(index: Int): Source {
      val fileSource = fileSystem.source(cleanFiles[index])
      if (civilizedFileSystem) return fileSource

      lockingSourceCount++
      return object : ForwardingSource(fileSource) {
        private var closed = false

        override fun close() {
          super.close()
          if (!closed) {
            closed = true
            synchronized(this@DiskLruCache) {
              lockingSourceCount--
              if (lockingSourceCount == 0 && zombie) {
                removeEntry(this@Entry)
              }
            }
          }
        }
      }
    }
```

So  eventually fails here

```
    if (entry != null && entry.lockingSourceCount != 0) {
      return null // We can't write this file because a reader is still reading it.
    }
```

yschimke:
At least one of the failures makes it looks expected.

```
  /**
   * Each read sees a snapshot of the file at the time read was called. This means that two reads of
   * the same key can see different data.
   */
  @Test
  fun readAndWriteOverlapsMaintainConsistency() {
    Assumptions.assumeFalse(windows) // Can't edit while a read is in progress.
```

Nohus:
@yschimke Hello, I was following this PR with much hope as I raised the original issue. Why was it closed?

yschimke:
I thought there was one further along, non draft. Let me check and reopen if not

yschimke:
closing in favour of https://github.com/lysine-dev/okhttp/pull/9446

PR #9345, opened 2026-02-28
cache: guard null fields in Entry.writeTo() to fix NPE (#8962)

### LINKED ISSUE #8962: NullPointerException in Cache.Entry.writeTo()
It’s crashing on the `handshake!!` line:

```
        if (url.isHttps) {
          sink.writeByte('\n'.code)
          sink.writeUtf8(handshake!!.cipherSuite.javaName).writeByte('\n'.code)
          writeCertList(sink, handshake.peerCertificates)
          writeCertList(sink, handshake.localCertificates)
          sink.writeUtf8(handshake.tlsVersion.javaName).writeByte('\n'.code)
        }
```

The crash looks like this:

```
	at okhttp3.Cache$Entry.writeTo(Cache.kt:633)
	at okhttp3.Cache.put$okhttp(Cache.kt:245)
	at okhttp3.internal.cache.CacheInterceptor.intercept(CacheInterceptor.kt:148)
```

I find it a bit surprising to have an HTTPS URL with no handshake at this point.

iVamsi:
Fix `NullPointerException` in `Cache.Entry.writeTo()` when an HTTPS response has a null handshake.

## Fix

Changed `if (url.isHttps)` to `if (url.isHttps && handshake != null)` so the TLS block is skipped when handshake is absent. The response still succeeds. It just won't be served from cache on subsequent requests. 

## Tests added

Three regression tests in `CacheTest`:
  - HTTPS response with null handshake does not crash `writeTo().`
  - Multiple HTTPS requests with null handshake all succeed
  - HTTPS response with null handshake is not served from cache

All verified to fail before the fix and pass after.

Fixes #8962


yschimke:
Actually `httpsResponseWithNullHandshakeIsNotServedFromCache` suggests it isn't blindly accepting.

PR #9364, opened 2026-03-11
Impose limits on HTTP/2 headers sizes

swankjesse:


okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/-HeadersCommon.kt
```diff
@@ -20,6 +20,9 @@ package okhttp3.internal
 
 import okhttp3.Headers
 
+/** This is the same as Chrome's limit. */
+const val HEADER_LIMIT = 256 * 1024L
```
JakeWharton:
```suggestion
internal const val HEADER_LIMIT = 256 * 1024L
```

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/http2/Hpack.kt
```diff
@@ -384,6 +387,10 @@ object Hpack {
         val huffmanDecode = firstByte and 0x80 == 0x80 // 1NNNNNNN
         val length = readInt(firstByte, PREFIX_7_BITS).toLong()
 
+        if (headerListByteCount + length > HEADER_LIMIT) {
+          throw IOException("header byte count limit of $HEADER_LIMIT exceeded")
```
yschimke, line 392: worth a comment that this is speculative check pre-decode, so the actual accounting is applied on the decoded value?

swankjesse: Good call. Fixed.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/-HeadersCommon.kt
swankjesse: YES! Fixed.

PR #9379, opened 2026-03-18
Gradle and vscode tweaks

yschimke:
With this running commands with VS Code gradle tasks, and a terminal or agent doesn't destroy the configuration cache. So something like `./gradlew okcurl:apiCheck` runs in 300ms

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -6,6 +6,12 @@ org.gradle.configuration-cache=true
 org.gradle.configuration-cache.problems=fail
 org.gradle.configuration-cache.parallel=true
 
+# Only configure projects reachable from the requested tasks, skipping unrelated subprojects
+org.gradle.isolated-projects=true
+
+# Watch the file system between builds to avoid full rescans on every invocation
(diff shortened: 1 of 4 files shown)


yschimke:
Tested with jetbrains, and it's faster also.

yschimke:

gradle.properties
```diff
@@ -16,7 +19,8 @@ containerTests=false
 okhttpModuleTests=false
 okhttpDokka=false
 
-org.gradle.jvmargs='-Dfile.encoding=UTF-8'
+# Increase heap and metaspace to reduce GC pressure when loading 30+ plugin classpaths
+org.gradle.jvmargs=-Xmx2g -XX:MaxMetaspaceSize=512m -Dfile.encoding=UTF-8
```
yschimke, line 23: reverted, seeing OOME in CI

swankjesse: (I would like to remove some plugins)

swankjesse (commented):
LGTM

PR #9385, opened 2026-03-23
Avoid Gradle actions v6

yschimke:
See [v6 licensing](https://github.com/gradle/actions?tab=readme-ov-file#licensing-notice)


JakeWharton:
I think the concern here has been overblown. The caching core here is the same in their proprietary plugin, presumably. If this was done correctly it should improve caching. Seems like a win.

PR #9395, opened 2026-03-30
Don't cache HTTPS responses that have no handshake

### LINKED ISSUE #8962: NullPointerException in Cache.Entry.writeTo()

iVamsi:
HTTPS cache entries always include TLS session data. If the handshake is missing — a network interceptor stripped it, or `cacheUrlOverride` turned an HTTP URL into HTTPS — we cannot write a readable entry.

#9345 prevented the NPE crash in `writeTo()`, but still wrote unreadable/malformed entries to disk. This skips the write in `put()` and `update()`.

Follow-up to #9345 / #8962


okhttp/src/commonJvmAndroid/kotlin/okhttp3/Cache.kt
```diff
@@ -238,6 +238,13 @@ class Cache internal constructor(
       return null
     }
 
+    if (response.request.url.isHttps && response.handshake == null) {
+      // An HTTPS response without a handshake cannot be serialized to the cache format because the
+      // Entry reader expects TLS session data for HTTPS URLs. This can happen when a network
+      // interceptor strips the handshake or when cacheUrlOverride rewrites an HTTP URL to HTTPS.
```
yschimke, line 244: Can you add a log via `Platform.log()` we should let the developer know something is going wrong.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Cache.kt
```diff
@@ -259,6 +266,9 @@ class Cache internal constructor(
     cached: Response,
     network: Response,
   ) {
+    if (network.request.url.isHttps && network.handshake == null) {
```
yschimke, line 274: same here, add a log.

yschimke:
@swankjesse do we agree that the result should be skipping caching when there isn't a valid handshake? I'm not certain that this would be expected. or if we should allow for this, maybe with some sentinel NoHandshake handshake?

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Cache.kt
iVamsi: Done — added Platform.get().log at WARN with redacted URLs

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Cache.kt
iVamsi: Done — added Platform.get().log at WARN with redacted URLs

iVamsi:
@swankjesse @yschimke, are any other changes needed to merge this PR?

PR #9401, opened 2026-04-04
Reject overflowing HPACK integers

oreparaz:
**Summary**. Reject malformed HPACK integers before they overflow in `Hpack.Reader.readInt()`.

**Problem**. `readInt()` decodes HPACK continuation bytes into a signed 32-bit Int without bounding the number of continuation bytes or checking for overflow. An overlong HPACK integer can therefore wrap into a negative or unrelated small positive value, and malformed input is only rejected later by downstream parsing logic.

 **Impact**. An HTTP/2 peer can use those malformed HPACK integers to cause connection-level DoS or to make OkHttp accept invalid HPACK state transitions that a compliant decoder should reject.

**Fix**. This change hardens `readInt()` by:

  - decoding into Long
  - rejecting more than 5 continuation bytes after the prefix
  - rejecting any increment that would exceed `Int.MAX_VALUE`
  - surfacing malformed values as `IOException("HPACK integer overflow")`

  Valid HPACK integers are unchanged, including the existing max 31-bit case.

**Tests**. Added regression coverage in HpackTest for malformed HPACK integers that previously overflowed or wrapped:

  - direct integer overflow
  - too many continuation bytes
  - overflow in HPACK string lengths
  - overflow that previously wrapped into a small dynamic table size

  Also updated the existing “insidious” HPACK tests to assert the new fail-fast behavior.

### DIFF
--- a/okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/http2/Hpack.kt
+++ b/okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/http2/Hpack.kt
@@ -365,19 +365,25 @@ object Hpack {
         }
 
         // This is a multibyte value. Read 7 bits at a time.
-        var result = prefixMask
+        var result = prefixMask.toLong()
         var shift = 0
+        var byteCount = 0
         while (true) {
-          val b = readByte()
-          if (b and 0x80 != 0) { // Equivalent to (b >= 128) since b is in [0..255].
-            result += b and 0x7f shl shift
-            shift += 7
-          } else {
-            result += b shl shift // Last byte.
-            break
+          // An Int.MAX_VALUE payload needs at most 5 continuation bytes after the prefix.
+          if (byteCount == 5) {
+            throw IOException("HPACK integer overflow")
           }
+          val b = readByte()
+          byteCount++
+          val increment = (b and 0x7f).toLong() shl shift
+          if (increment > Int.MAX_VALUE.toLong() - result) {
+            throw IOException("HPACK integer overflow")
+          }
+          result += increment
+          if (b and 0x80 == 0) break
+          shift += 7
         }
-        return result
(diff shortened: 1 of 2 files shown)


yschimke:
It doesn't sound like this materially changes the logic or avoids security issues, but does more closely follow the spec

https://datatracker.ietf.org/doc/html/rfc7541#section-5.1

>    This integer representation allows for values of indefinite size.  It
   is also possible for an encoder to send a large number of zero
   values, which can waste octets and could be used to overflow integer
   values.  Integer encodings that exceed implementation limits -- in
   value or octet length -- MUST be treated as decoding errors.
   Different limits can be set for each of the different uses of
   integers, based on implementation constraints.

oreparaz:
@yschimke: Agreed, this is primarily a spec-compliance / parser-hardening change, which seems more aligned with RFC 7541 §5.1. The material difference is that malformed overlong HPACK integers now fail in `readInt()` instead of overflowing and being handled later by downstream logic.

yschimke (approved):

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/http2/Hpack.kt
```diff
@@ -365,19 +365,25 @@ object Hpack {
...
 
         // This is a multibyte value. Read 7 bits at a time.
-        var result = prefixMask
+        var result = prefixMask.toLong()
         var shift = 0
+        var byteCount = 0
         while (true) {
+          // An Int.MAX_VALUE payload needs at most 5 continuation bytes after the prefix.
+          if (byteCount == 5) {
+            throw IOException("HPACK integer overflow")
```
swankjesse, line 374: nice

swankjesse (commented):
Nice tests.

LGTM

PR #9269, opened 2026-01-25
Enable AGP newDsl

yschimke:
Uses the new multiplatform plugin and AGP androidLibrary.

PR #9405, opened 2026-04-05
Move androidDeviceTest to androidTest

yschimke:
Rename the source directory for Android device tests to the standard `androidTest` location.

Moved in #9269 but seemingly that wasn't a thing?


yschimke:
@swankjesse fixing a mistaken move. But we do need to change android tests to junit 4 as a follow up

swankjesse (commented):
LGTM

PR #9407, opened 2026-04-06
Add Partitioned cookie attribute support

### LINKED ISSUE #8705: `Partioned` attribute missing in `Cookie`
[Cookie](https://github.com/square/okhttp/blob/master/okhttp/src/commonJvmAndroid/kotlin/okhttp3/Cookie.kt) does not store if a `Partioned` attribute is set.

It can be implemented just like the `secure` flag with a simple boolean.

It's needed for the "Cookies Having Independent Partitioned State" (CHIPS) feature, which all browsers support already.
More details here: https://developers.google.com/privacy-sandbox/cookies/chips

Thanks!

PS: Regarding browser support if you interested:
- Supported already by Chrome/Edge according to https://caniuse.com/mdn-http_headers_set-cookie_partitioned
  - Data seems to be incomplete however, see next points
- In Firefox since v131: https://www.mozilla.org/en-US/firefox/131.0/releasenotes/#note-790397
- Coming to Safari 18.4 (https://developer.apple.com/documentation/safari-release-notes/safari-18_4-release-notes#Networking)

eyupcanakman:
Cookie does not parse or expose the `Partitioned` attribute from `Set-Cookie` headers. The attribute is used for cross-site cookie partitioning (CHIPS) and callers have no way to read or set it.

Adds a `partitioned` boolean to `Cookie` and a `partitioned()` builder method. Both follow the `secure` and `httpOnly` pattern. Tests in `CookieTest` cover parsing, builder, toString round-trip, and equality.

Fixes #8705

### DIFF
--- a/okhttp/api/android/okhttp.api
+++ b/okhttp/api/android/okhttp.api
@@ -402,7 +402,7 @@ public final class okhttp3/Cookie {
 	public final fun -deprecated_persistent ()Z
 	public final fun -deprecated_secure ()Z
 	public final fun -deprecated_value ()Ljava/lang/String;
-	public synthetic fun <init> (Ljava/lang/String;Ljava/lang/String;JLjava/lang/String;Ljava/lang/String;ZZZZLjava/lang/String;Lkotlin/jvm/internal/DefaultConstructorMarker;)V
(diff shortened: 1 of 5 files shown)


mkurz:
Anyone?

PR #9444, opened 2026-05-15
Recover HTTP2 connection from half-open socket using pingInterval.

joelc:
A stateful firewall *CAN* silently kill a pooled HTTP2 connection at L3 (no RST, no FIN).

When this happens with okhttp, the next frame write that fills the kernel TCP send buffer blocks indefinitely.

pingInterval is documented to detect exactly this, but three issues prevented it from firing:

* The ping watchdog ran on writerQueue, so it queued behind the wedged frame write and never ticked. Moved to its own pingQueue; the ping write itself is fire-and-forget onto writerQueue.

* failConnection -> close -> shutdown -> writer.goAway() deadlocked on writer.lock held by the wedged write. failConnection now cancels the socket first to unblock the writer.

* SSLSocket.close() can block for many seconds attempting to send close_notify on a half-open connection. ConnectPlan now closes the raw TCP socket before the SSL layer on cancel.

This PR adds a container-tests repro using socat amd an iptables DROP to simulate such a firewall drop. It asserts the failure within 3x pingInterval.

(Apologies for additional pushes, was unable to run entire suite locally).


swankjesse:
Great analysis. A permanently wedged writer is likely gonna cause a bunch of liveness bugs. I'll try to address this report specifically and that issue generally.

joelc:
@swankjesse any chance we could get this reviewed please? lmk if you need me to action any further...!

PR #9452, opened 2026-05-24
tls: add Ed25519 certificate support

### LINKED ISSUE #9437: ed25519 client certificate is not sent to server in mtls set up
The thing is that the same set up with RSA key-pair set signed clientcertificate works fine. But the ed25519 setup fails.

a reproducible code example:
git@github.com:unegare/ed25519okhttpExample.git

running it against an openssl s_server faces the following error:
```
$ openssl s_server -cert ./ed25519_cert_1.pem -key ./ed25519_key_1.pem -Verify 1 -accept 0.0.0.0:3000 -www
verify depth is 1, must return a certificate
Using default temp DH parameters
ACCEPT
4017FB7DA17D0000:error:0A000076:SSL routines:tls_choose_sigalg:no suitable signature algorithm:../ssl/t1_lib.c:3252:
```
or
```
$ openssl s_server -cert ./ed25519_cert_1.pem -key ./ed25519_key_1.pem -Verify 1 -accept 0.0.0.0:3000 -www
verify depth is 1, must return a certificate
Using default temp DH parameters
ACCEPT
4017FB7DA17D0000:error:0A0000C7:SSL routines:tls_process_client_certificate:peer did not return a certificate:../ssl/statem/statem_srvr.c:3509:
```

nathannewyen:
Fixes #9437 - Ed25519 client certificates not being sent in mTLS setups.

Added Ed25519 support to okhttp-tls: new ed25519() builder method, proper key detection in decode(), and the Ed25519 OID in the DER layer.

### DIFF
--- a/okhttp-tls/src/main/kotlin/okhttp3/tls/HeldCertificate.kt
+++ b/okhttp-tls/src/main/kotlin/okhttp3/tls/HeldCertificate.kt
@@ -44,6 +44,7 @@ import okhttp3.tls.internal.der.CertificateAdapters.generalNameIpAddress
 import okhttp3.tls.internal.der.Extension
 import okhttp3.tls.internal.der.ObjectIdentifiers
 import okhttp3.tls.internal.der.ObjectIdentifiers.BASIC_CONSTRAINTS
+import okhttp3.tls.internal.der.ObjectIdentifiers.ED25519
 import okhttp3.tls.internal.der.ObjectIdentifiers.ORGANIZATIONAL_UNIT_NAME
 import okhttp3.tls.internal.der.ObjectIdentifiers.SHA256_WITH_ECDSA
 import okhttp3.tls.internal.der.ObjectIdentifiers.SHA256_WITH_RSA_ENCRYPTION
@@ -341,6 +342,16 @@ class HeldCertificate(
         keySize = 2048
       }
 
+    /**
+     * Configure the certificate to generate an Ed25519 key. Ed25519 provides about 128 bits of
+     * security and fast signing/verification. Requires Java 15 or newer at runtime.
+     */
+    fun ed25519() =
+      apply {
+        keyAlgorithm = "Ed25519"
+        keySize = 0
+      }
+
     fun build(): HeldCertificate {
       // Subject keys & identity.
       val subjectKeyPair = keyPair ?: generateKeyPair()
@@ -480,10 +491,17 @@ class HeldCertificate(
     }
 
     private fun signatureAlgorithm(signedByKeyPair: KeyPair): AlgorithmIdentifier =
-      when (signedByKeyPair.private) {
-        is RSAPrivateKey -> {
+      when {
+        signedByKeyPair.private is RSAPrivateKey -> {
           AlgorithmIdentifier(
             algorithm = SHA256_WITH_RSA_ENCRYPTION,
+            parameters = null,
+          )
+        }
+
+        signedByKeyPair.private.algorithm == "Ed25519" || signedByKeyPair.private.algorithm == "EdDSA" -> {
+          AlgorithmIdentifier(
+            algorithm = ED25519,
             parameters = null,
           )
         }
@@ -498,7 +516,7 @@ class HeldCertificate(
 
     private fun generateKeyPair(): KeyPair =
       KeyPairGenerator.getInstance(keyAlgorithm).run {
-        initialize(keySize, SecureRandom())
+        if (keySize > 0) initialize(keySize, SecureRandom())
         generateKeyPair()
       }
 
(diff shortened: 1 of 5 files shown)


okhttp-tls/src/main/kotlin/okhttp3/tls/HeldCertificate.kt
```diff
@@ -480,14 +491,21 @@ class HeldCertificate(
...
-        is RSAPrivateKey -> {
+      when {
+        signedByKeyPair.private is RSAPrivateKey -> {
           AlgorithmIdentifier(
             algorithm = SHA256_WITH_RSA_ENCRYPTION,
             parameters = null,
           )
         }
 
+        signedByKeyPair.private.algorithm == "Ed25519" || signedByKeyPair.private.algorithm == "EdDSA" -> {
```
yschimke, line 502: Given differing providers, I'm worried about case sensitivity, can you audit how it's represented in providers, and maybe make it case insensitive,  if it's not guaranteed by a spec.

okhttp-tls/src/main/kotlin/okhttp3/tls/HeldCertificate.kt
```diff
@@ -498,7 +516,7 @@ class HeldCertificate(
 
     private fun generateKeyPair(): KeyPair =
       KeyPairGenerator.getInstance(keyAlgorithm).run {
-        initialize(keySize, SecureRandom())
+        if (keySize > 0) initialize(keySize, SecureRandom())
```
yschimke, line 519: Probably worth a comment since this affects other code paths.

okhttp-tls/src/test/java/okhttp3/tls/HeldCertificateTest.kt
```diff
@@ -262,6 +262,56 @@ class HeldCertificateTest {
     assertThat(leaf.certificate.sigAlgName).isEqualTo("SHA256WITHECDSA", ignoreCase = true)
   }
 
+  @Test
+  fun ed25519() {
+    platform.assumeNotAndroid()
```
yschimke, line 268: This check is just in the tests, but if called by an app, we should have a useful error message.

Something like

Ed25519 requires JDK 15+ / a provider that supports it

okhttp-tls/src/main/kotlin/okhttp3/tls/HeldCertificate.kt
yschimke, line 502: `signedByKeyPair.private.algorithm == "EdDSA"`

Is this check complete? or are there sub algorithms?

yschimke:
Failures


HeldCertificateTest > ed25519() FAILED
    java.lang.NullPointerException: null cannot be cast to non-null type okio.ByteString
        at okhttp3.tls.internal.der.Adapters$usingTypeHint$1.toDer(Adapters.kt:450)
        at okhttp3.tls.internal.der.Adapters$sequence$codec$1.encode$lambda$1(Adapters.kt:383)
        at okhttp3.tls.internal.der.DerWriter.withTypeHint(DerWriter.kt:109)
        at okhttp3.tls.internal.der.Adapters$sequence$codec$1.encode(Adapters.kt:380)
        at okhttp3.tls.internal.der.BasicDerAdapter.toDer$lambda$0(BasicDerAdapter.kt:81)
        at okhttp3.tls.internal.der.DerWriter.write(DerWriter.kt:64)
        at okhttp3.tls.internal.der.BasicDerAdapter.toDer(BasicDerAdapter.kt:80)
        at okhttp3.tls.internal.der.Adapters$sequence$codec$1.encode$lambda$1(Adapters.kt:383)
        at okhttp3.tls.internal.der.DerWriter.withTypeHint(DerWriter.kt:109)
        at okhttp3.tls.internal.der.Adapters$sequence$codec$1.encode(Adapters.kt:380)
        at okhttp3.tls.internal.der.BasicDerAdapter.toDer$lambda$0(BasicDerAdapter.kt:81)
        at okhttp3.tls.internal.der.DerWriter.write(DerWriter.kt:64)
        at okhttp3.tls.internal.der.BasicDerAdapter.toDer(BasicDerAdapter.kt:80)
        at okhttp3.tls.internal.der.DerAdapter.toDer(DerAdapter.kt:61)
        at okhttp3.tls.HeldCertificate$Builder.build(HeldCertificate.kt:401)
        at okhttp3.tls.HeldCertificateTest.ed25519(HeldCertificateTest.kt:273)

HeldCertificateTest > ed25519SignedByEcdsa() FAILED
    org.opentest4j.AssertionFailedError: expected:<"Ed[25519]"> but was:<"Ed[DSA]">
        at app//okhttp3.tls.HeldCertificateTest.ed25519SignedByEcdsa(HeldCertificateTest.kt:312)

nathannewyen:
Thanks for running the tests! Fixed both issues — the NPE was caused by `usingTypeHint.toDer` not handling absent parameters (null value with no registered adapter for the Ed25519 OID). Added a null guard so it skips writing rather than crashing. Also loosened the algorithm name assertions to accept both `"Ed25519"` and `"EdDSA"` since different JVMs report different strings.

PR #9453, opened 2026-05-25
Fix isDotDot to recognize ..%2e and %2e.. path segments

### LINKED ISSUE #9451: HttpUrl.isDotDot misses '..%2e' and '%2e..' forms — path-traversal sandbox bypass
### Description

`HttpUrl.Builder`'s `isDotDot()` matcher recognizes some percent-encoded forms of `".."` (`%2e.`, `.%2e`, `%2e%2e`) but **misses two more**: `..%2e` and `%2e..`. These extra forms therefore are **not normalized away** during URL resolution, even though they decode to exactly the same dot-dot segment.

This can cause path-traversal sandbox checks built on top of `HttpUrl.resolve(...).encodedPath()` to be bypassed.

### Reproducer (okhttp 4.12.0)

```java
import okhttp3.HttpUrl;

HttpUrl base = HttpUrl.parse("http://example.com/static/");

// Forms that are correctly normalized (path collapses to "/"):
base.resolve("..");        //  -> http://example.com/         path=/
base.resolve("%2e%2e");    //  -> http://example.com/         path=/
base.resolve(".%2e");      //  -> http://example.com/         path=/
base.resolve("%2e.");      //  -> http://example.com/         path=/

// Forms that are NOT recognized — left in the path literally:
base.resolve("..%2e");     //  -> http://example.com/static/..%2e   path=/static/..%2e
base.resolve("%2e..");     //  -> http://example.com/static/%2e..   path=/static/%2e..
base.resolve("..%2E");     //  -> http://example.com/static/..%2E   path=/static/..%2E
```

### Root cause

`HttpUrl.kt` lines 1455-1460:

```kotlin
private fun isDotDot(input: String): Boolean {
  return input == ".." ||
      input.equals("%2e.", ignoreCase = true) ||
      input.equals(".%2e", ignoreCase = true) ||
      input.equals("%2e%2e", ignoreCase = true)
}
```

The enumeration covers 4 forms but not `..%2e` or `%2e..` (i.e. one literal dot followed by an encoded dot, then itself).

### Threat model

Code that uses `HttpUrl.resolve()` to compute the absolute form of a user-supplied path and then validates the resulting `encodedPath()` against a prefix (e.g. `"/static/"`) is bypassed:

```kotlin
val resolved = base.resolve(userInput)
require(resolved.encodedPath().startsWith("/static/"))      // passes for "..%2e"
// later: read file at resolved.encodedPath()
//   /static/..%2e   →  on most filesystems, %2e is decoded → /static/../  → traversal
```

Many HTTP servers and file-serving libraries do percent-decode `%2e` later in the pipeline, so the bypass becomes a path traversal in practice. URL specs (RFC 3986 §6.2.2.3) treat `..%2e` and `%2e..` as equivalent to `..`.

### Suggested fix

```diff
 private fun isDotDot(input: String): Boolean {
   return input == ".." ||
       input.equals("%2e.", ignoreCase = true) ||
       input.equals(".%2e", ignoreCase = true) ||
-      input.equals("%2e%2e", ignoreCase = true)
+      input.equals("%2e%2e", ignoreCase = true) ||
+      input.equals("..%2e", ignoreCase = true) ||
+      input.equals("%2e..", ignoreCase = true)
 }
```

(Two new branches; same `ignoreCase` semantics already used for the others.)

### Environment

- okhttp: 4.12.0 (also expected on 5.x; the matcher is unchanged in `okhttp/src/main/kotlin/okhttp3/HttpUrl.kt`)

Discovered via jqwik property-based testing while fuzzing the URL-resolution invariant `path-normalization(resolve(x)) is canonical`.

wyf027:
## Summary

- Extend `HttpUrl.Builder.isDotDot()` to treat `..%2e` and `%2e..` as dot-dot segments, matching RFC 3986 §6.2.2.3 normalization semantics already applied to `%2e.`, `.%2e`, and `%2e%2e`.
- Add regression tests in `HttpUrlTest.relativePath()` for the missing encoded forms (including uppercase `%2E` variants).

Fixes #9451

## Problem

`HttpUrl.resolve()` left `..%2e` and `%2e..` in the encoded path instead of collapsing them. Code that validates `encodedPath()` against a prefix (e.g. `/static/`) could be bypassed when downstream components percent-decode `%2e` later.

## Test plan

- [ ] `./gradlew :okhttp:jvmTest --tests "okhttp3.HttpUrlTest.relativePath"`
- [ ] Verify `base.resolve("..%2e")` and `base.resolve("%2e..")` normalize to the parent path instead of leaving literal encoded segments

### DIFF
--- a/okhttp/src/commonJvmAndroid/kotlin/okhttp3/HttpUrl.kt
+++ b/okhttp/src/commonJvmAndroid/kotlin/okhttp3/HttpUrl.kt
@@ -1594,7 +1594,9 @@ class HttpUrl private constructor(
       input == ".." ||
         input.equals("%2e.", ignoreCase = true) ||
         input.equals(".%2e", ignoreCase = true) ||
-        input.equals("%2e%2e", ignoreCase = true)
+        input.equals("%2e%2e", ignoreCase = true) ||
+        input.equals("..%2e", ignoreCase = true) ||
+        input.equals("%2e..", ignoreCase = true)
 
     /**
      * Cuts this string up into alternating parameter names and values. This divides a query string
--- a/okhttp/src/jvmTest/kotlin/okhttp3/HttpUrlTest.kt
+++ b/okhttp/src/jvmTest/kotlin/okhttp3/HttpUrlTest.kt
@@ -973,6 +973,10 @@ open class HttpUrlTest {
     assertThat(base.resolve("%2e.")).isEqualTo(parse("http://host/a/"))
     assertThat(base.resolve(".%2e")).isEqualTo(parse("http://host/a/"))
     assertThat(base.resolve("%2e%2e")).isEqualTo(parse("http://host/a/"))
+    assertThat(base.resolve("..%2e")).isEqualTo(parse("http://host/a/"))
+    assertThat(base.resolve("%2e..")).isEqualTo(parse("http://host/a/"))
+    assertThat(base.resolve("..%2E")).isEqualTo(parse("http://host/a/"))
+    assertThat(base.resolve("%2E..")).isEqualTo(parse("http://host/a/"))
     assertThat(base.resolve("%2E")).isEqualTo(parse("http://host/a/b/"))
     assertThat(base.resolve("%2e")).isEqualTo(parse("http://host/a/b/"))
   }


yschimke:
I suspect the bug report is AI slop, and now an AI generated fix.

Can you please explain how this is affecting you? That might help me understand why it needs a fix.

yschimke (changes requested):
Please provide context for how you hit this issue.

wyf027:
Thanks for reviewing. I do not have a production incident or concrete app impact to point to here; the motivation was the path normalization edge case itself, where encoded dot-dot variants such as ..%2e and %2e.. can be interpreted as parent-directory traversal by other URL/path normalizers.\n\nGiven your concern that this is not grounded in a real reported OkHttp issue, I do not want to invent context. I am going to close this PR unless you would prefer to keep it open as a defensive normalization fix.\n\nThe remaining Windows CI failure appears unrelated to this change: it is timing-sensitive failures in EventListenerTest/CallTest where observed timing was outside the expected delta.

wyf027:
Closing because I cannot provide the requested real-world context for this edge case, and I do not want to keep an ungrounded fix open.

PR #9454, opened 2026-05-25
Clarify Dispatcher executor sizing requirements in KDoc

### LINKED ISSUE #9443: Dispatcher Javadoc understates the executor requirement ("should be able to run maxRequests concurrently" is necessary but not sufficient)
The `Dispatcher` class doc currently states:

> Each dispatcher uses an `ExecutorService` to run calls internally. If you supply your own executor, it should be able to run [the configured maximum][maxRequests] number of calls concurrently.

— `Dispatcher.kt`, lines 32–35 on master (`74428398`)

This phrasing implies that an executor sized to `maxRequests` (e.g. `ThreadPoolExecutor(core=N, max=N, SynchronousQueue)` with `dispatcher.setMaxRequests(N)`) is sufficient. It isn't. Such an executor will reject submissions with `RejectedExecutionException` — surfaced to callers as `InterruptedIOException("executor rejected")` — under sustained traffic that approaches `maxRequests`.

### Why

When a worker thread finishes a call, it runs `Dispatcher.finished()` → `promoteAndExecute()` synchronously on its way back to `queue.take()`. The synchronized block selects calls to promote and builds `callsToExecute`; the actual `executor.execute(...)` happens *outside* the lock (`Dispatcher.kt:236–251`). At the moment of the submit, the finishing worker thread is itself still a pool slot, but is not yet in `take()`. With `SynchronousQueue` + `core == max == N`, there is no waiting consumer and no room to grow the pool, so the offer fails. With multiple near-simultaneous completions, multiple worker threads can be in this state at once; up to `maxRequests` rejections are possible per burst.

### Evidence in this repo

The default executor at `Dispatcher.kt:96–104` is `ThreadPoolExecutor(0, Int.MAX_VALUE, 60, TimeUnit.SECONDS, SynchronousQueue(), ...)`. If the doc's stated requirement were sufficient, the default would be `(maxRequests, maxRequests, ...)`. Instead, `maxPoolSize=Int.MAX_VALUE` is used precisely to absorb the transient-submit race described above.

### Suggested wording

> Each dispatcher uses an `ExecutorService` to run calls internally. If you supply your own executor, it must accept any new task at any time while up to [the configured maximum][maxRequests] calls are already running. A pool sized exactly to `maxRequests` is **not** sufficient when paired with `SynchronousQueue`, because the dispatcher submits the next promoted call from a worker thread that has not yet returned to the queue; the offer will fail and the call will surface `InterruptedIOException("executor rejected")`. Match the pattern used by the default executor (`corePoolSize=0, maxPoolSize=Int.MAX_VALUE, SynchronousQueue`), or bound the maximum at `2 × maxRequests`, or use a queueing executor.

### Referenced prior work

#4365 introduced the `executor rejected` failure path; this issue is about the documentation that leads users into the trap that PR is designed to surface.

wyf027:
## Summary

- Expand the `Dispatcher` class KDoc to explain that an executor sized exactly to `maxRequests` is not sufficient when paired with `SynchronousQueue`.
- Document why the default executor uses `maxPoolSize=Int.MAX_VALUE` and suggest safe alternatives (`2 × maxRequests` or a queueing executor).

Fixes #9443

## Problem

The previous wording ("should be able to run maxRequests number of calls concurrently") led users to configure `ThreadPoolExecutor(maxRequests, maxRequests, ..., SynchronousQueue)`, which can reject submissions with `InterruptedIOException("executor rejected")` under sustained load. This happens because `promoteAndExecute()` submits the next promoted call from a worker thread before it returns to the queue.

## Test plan

- [ ] Documentation-only change; no runtime behavior change
- [ ] KDoc renders correctly in generated docs

### DIFF
--- a/okhttp/src/commonJvmAndroid/kotlin/okhttp3/Dispatcher.kt
+++ b/okhttp/src/commonJvmAndroid/kotlin/okhttp3/Dispatcher.kt
@@ -37,8 +37,8 @@ import okhttp3.internal.unmodifiable
  * thread that has not yet returned to the queue; the offer will fail and the call will surface
  * [InterruptedIOException] with message `"executor rejected"`.
  *
- * Match the pattern used by the default executor (`corePoolSize=0`, `maxPoolSize=Int.MAX_VALUE`,
- * [SynchronousQueue]), bound the maximum at `2 × maxRequests`, or use a queueing executor.
+ * Prefer a queueing executor. Avoid [SynchronousQueue] for custom dispatchers because it forces
+ * additional worker threads instead of reusing the workers that are still returning from calls.
  */
 class Dispatcher() {
   /**
--- a/okhttp/src/commonJvmAndroid/kotlin/okhttp3/Dispatcher.kt
+++ b/okhttp/src/commonJvmAndroid/kotlin/okhttp3/Dispatcher.kt
@@ -31,8 +31,14 @@ import okhttp3.internal.unmodifiable
  * Policy on when async requests are executed.
  *
  * Each dispatcher uses an [ExecutorService] to run calls internally. If you supply your own
- * executor, it should be able to run [the configured maximum][maxRequests] number of calls
- * concurrently.
+ * executor, it must accept any new task at any time while up to [the configured maximum][maxRequests]
+ * calls are already running. A pool sized exactly to `maxRequests` is not sufficient when paired
+ * with [SynchronousQueue], because the dispatcher submits the next promoted call from a worker
+ * thread that has not yet returned to the queue; the offer will fail and the call will surface
+ * [InterruptedIOException] with message `"executor rejected"`.
+ *
+ * Match the pattern used by the default executor (`corePoolSize=0`, `maxPoolSize=Int.MAX_VALUE`,
+ * [SynchronousQueue]), bound the maximum at `2 × maxRequests`, or use a queueing executor.
  */
 class Dispatcher() {
   /**


wyf027:
I checked the failing Windows job. The failure is in timing-sensitive JVM tests (EventListenerTest/CallTest timing deltas), while the Linux/JVM/checks jobs are green. This appears unrelated to this documentation-only change.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Dispatcher.kt
```diff
@@ -31,8 +31,14 @@ import okhttp3.internal.unmodifiable
...
- * executor, it should be able to run [the configured maximum][maxRequests] number of calls
- * concurrently.
+ * executor, it must accept any new task at any time while up to [the configured maximum][maxRequests]
+ * calls are already running. A pool sized exactly to `maxRequests` is not sufficient when paired
+ * with [SynchronousQueue], because the dispatcher submits the next promoted call from a worker
+ * thread that has not yet returned to the queue; the offer will fail and the call will surface
+ * [InterruptedIOException] with message `"executor rejected"`.
+ *
+ * Match the pattern used by the default executor (`corePoolSize=0`, `maxPoolSize=Int.MAX_VALUE`,
+ * [SynchronousQueue]), bound the maximum at `2 × maxRequests`, or use a queueing executor.
```
swankjesse: I would rather we just recommend against `SynchronousQueue`. That’s a bad fit for this situation because it forces more threads than we need.

swankjesse (changes requested):

PR #9467, opened 2026-05-30
fix: defer platform detection until first use

### LINKED ISSUE #9462: Class.forName(Jdk9Platform) throws during warm up
When we try to warm up our Java binary by invoking `Class.forName` to pre-load all the classes we use, `okhttp3.internal.platform.Jdk9Platform` throws an `ExceptionInInitializerError` exception. 

The root cause is that there is a cyclic initialization between `Jdk9Platform` and its parent `Platform`:

Initialize `Jdk9Platform` > Initialize `Platform` > Initialize `Platform.Companion` > eventually calls `Jdk9Platform.buildIfSupported()` > at this point, the companion object of `Jdk9Platform` has not been initialized yet.

One possible way to solve this is to make `Platform.platform` lazy (does not call `getPlatform()` immediately).

Test:

```java
import org.junit.Test;

public class OkHttpBug {
  @Test
  public void testFailure() throws Exception {
    Class.forName("okhttp3.internal.platform.Jdk9Platform");

    // Expected Behavior:
    // The class okhttp3.internal.platform.Jdk9Platform should be loaded successfully without
    // throwing any exceptions.
    //
    // Actual Failure:
    // java.lang.ExceptionInInitializerError
    //     ...
    // Caused by: java.lang.NullPointerException: Cannot invoke
    // "okhttp3.internal.platform.Jdk9Platform$Companion.buildIfSupported()" because
    // "okhttp3.internal.platform.Jdk9Platform.Companion" is null

    System.out.println("Loaded Jdk9Platform successfully.");
  }
}
```

cyphercodes:
## Summary
- Defer OkHttp platform detection until `Platform.get()` is first used.
- Add an isolated JVM regression test proving `Class.forName("okhttp3.internal.platform.Jdk9Platform")` no longer trips the superclass initialization cycle.

Fixes #9462

## Testing
- `./gradlew :okhttp:jvmTest --tests okhttp3.internal.platform.Jdk9PlatformInitializationTest --no-daemon --rerun-tasks`
- `./gradlew :okhttp:jvmTest --tests okhttp3.internal.platform.Jdk9PlatformTest --no-daemon --rerun-tasks`
- `git diff --check`

Note: locally I had to point Gradle's daemon JVM criteria at the installed Ubuntu OpenJDK 21 to run the checks; that generated file was restored before committing.


okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/platform/Platform.kt
```diff
@@ -204,18 +204,18 @@ open class Platform {
   override fun toString(): String = javaClass.simpleName
 
   companion object {
-    @Volatile private var platform = findPlatform()
+    @Volatile private var platform = lazy { findPlatform() }
```
yschimke: This feels like a confused line now.

var + lazy + volatile.

yschimke: perhaps we make get() responsible for setting it if unset?

cyphercodes:
Updated the deferred platform initialization follow-up:

- replaced the `var + lazy + @Volatile` holder with a nullable platform cache
- made `get()` initialize and cache the platform only when unset
- fixed the missing final newline in the new JDK9 initialization test

Verification:
- `./gradlew --no-daemon --no-configuration-cache -Dorg.gradle.java.installations.auto-download=false -Dorg.gradle.java.installations.paths=/usr/lib/jvm/java-21-openjdk-amd64,/usr/lib/jvm/java-11-openjdk-amd64 :okhttp:spotlessKotlinCheck :okhttp:jvmTest --tests 'okhttp3.internal.platform.Jdk9PlatformInitializationTest'`
- `git diff --check`

JakeWharton:
Please stop pasting LLM output. If you're not going to bother to write yourself why should we be bothered to read it?

PR #9481, opened 2026-06-08
Use JDK26 (unless we're specifically testing old JDKs)

swankjesse:


.github/workflows/build.yml
```diff
@@ -18,31 +18,6 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  publish:
```
swankjesse, line 21: I should have deleted this when I did this PR:
https://github.com/square/okhttp/pull/9202

JakeWharton (approved):
Probably want this, too: https://github.com/cashapp/molecule/blob/9a48df3891fb4307032942c927e81fe5c6026e17/.github/renovate.json5#L6-L25

That way Renovate will bump it when future versions come out.

swankjesse:
Good call. I also took the liberty to switch from `.json` to `.json5`

PR #9496, opened 2026-06-21
Add RFC 1008 (QUERY Method) support

desiderantes:
Now that QUERY method graduated to a proper RFC, this change fixes the missing bits for conformance, which is redirect handling. This change allows QUERY requests to be automatically redirected reusing the body, as well as allowing redirect to GET. Fortunately, this library already implemented these behaviours for PROPFIND so it's just adding QUERY into the list of methods.

### DIFF
--- a/okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/http/RetryAndFollowUpInterceptor.kt
+++ b/okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/http/RetryAndFollowUpInterceptor.kt
@@ -315,9 +315,11 @@ class RetryAndFollowUpInterceptor : Interceptor {
       val responseCode = userResponse.code
       val isQueryAndSeeOther = method == "QUERY" && responseCode == HTTP_SEE_OTHER
       val maintainBody =
-        (HttpMethod.redirectsWithBody(method) ||
-          responseCode == HTTP_PERM_REDIRECT ||
-          responseCode == HTTP_TEMP_REDIRECT) && !isQueryAndSeeOther
+        (
+          HttpMethod.redirectsWithBody(method) ||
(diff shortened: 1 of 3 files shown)


okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/http/RetryAndFollowUpInterceptor.kt
```diff
@@ -313,11 +313,16 @@ class RetryAndFollowUpInterceptor : Interceptor {
...
-        HttpMethod.redirectsWithBody(method) ||
-          responseCode == HTTP_PERM_REDIRECT ||
-          responseCode == HTTP_TEMP_REDIRECT
-      if (HttpMethod.redirectsToGet(method) && responseCode != HTTP_PERM_REDIRECT && responseCode != HTTP_TEMP_REDIRECT) {
+        (
+          HttpMethod.redirectsWithBody(method) ||
+            responseCode == HTTP_PERM_REDIRECT ||
+            responseCode == HTTP_TEMP_REDIRECT
+        ) && !isQueryAndSeeOther
+      if ((HttpMethod.redirectsToGet(method) || isQueryAndSeeOther) &&
```
yschimke: Should we add responseCode to redirectsToGet?  Then this is all in one place.

desiderantes: I agree.

yschimke (approved):
After a change to re

desiderantes:
@yschimke 
> After a change to re

Sorry, I think this message got truncated

swankjesse (commented):
LGTM This is rad

PR #9510, opened 2026-06-25
fix missing 4.x doc links

### LINKED ISSUE #9156: Missing documentation for OkHttp 4.x
The project Web site https://square.github.io/ seems to lack documentation for OkHttp 4.x. Some pages still have links to said documentation: for example, https://square.github.io/okhttp/features/caching/ references https://square.github.io/okhttp/4.x/okhttp/okhttp3/-cache/ , but this link is broken (leads to 404).

GautamKumarOffical:
Fixes #9156

The site nav only linked to 5.x API docs with no 4.x entry, and there was no redirect fallback for old `/4.x/` URLs. Users following old links or bookmarks to 4.x API docs would get 404s.

**Changes:**
- Added a `4.x` entry under the `API` nav in `mkdocs.yml` alongside `5.x`
- Both `deploy_website.sh` and `test_docs.sh` now generate a `docs/4.x/` redirect page that maps `/4.x/*` → `/5.x/*` using JS + meta refresh fallback
- The redirect page is also committed as a static file in `docs/4.x/okhttp/okhttp3/index.html`
- On the deployed site, the gh-pages cherry-picks will overlay the real 4.x Dokka docs on top of this redirect


deploy_website.sh
```diff
@@ -30,6 +30,30 @@ cat README.md | grep -v 'project website' > docs/index.md
...
+            window.location.replace(path.substring(0, idx) + '/5.x' + path.substring(idx + 4) + window.location.search + window.location.hash);
+        }
+    </script>
+    <meta http-equiv="refresh" content="0; url=../../5.x/okhttp/okhttp3/">
+</head>
+<body>
+    <p>OkHttp 4.x API docs have moved. Redirecting to <a href="../../5.x/okhttp/okhttp3/">5.x API docs</a>&hellip;</p>
+</body>
+</html>
+REDIRECT
```
JakeWharton, line 55: Why are we writing the same code out here and below if it's also checked in?

mkdocs.yml
```diff
@@ -102,7 +102,9 @@ nav:
     - 'Providers': security/security_providers.md
     - 'Configuration History': security/tls_configuration_history.md
   - 'Works with OkHttp': works_with_okhttp.md
-  - 'API': 5.x/okhttp/okhttp3/
+  - 'API':
+    - '5.x': 5.x/okhttp/okhttp3/
+    - '4.x': 4.x/okhttp/okhttp3/
```
JakeWharton: Why would we link to the 4.x docs if it's going to immediately redirect to the 5.x ones?

GautamKumarOffical:
Hey Jake, good catches on both comments.

1. The redirect HTML is duplicated between deploy_website.sh and the checked-in file. I'll remove the checked-in copy and let deploy_website.sh generate it during build.

2. You're right about the nav link - if 4.x immediately redirects to 5.x, there's no point linking to it. I'll remove the 4.x nav entry and just keep the redirect page for old URLs bookmarked by users.

Will update the PR shortly.

GautamKumarOffical:
Done! Addressed both of Jake's review comments:

1. **Removed checked-in redirect HTML** (docs/4.x/okhttp/okhttp3/index.html) — deploy_website.sh and 	est_docs.sh already generate this during build, so the static copy was redundant.
2. **Removed the 4.x nav entry** from mkdocs.yml — reverted back to a simple API: 5.x/okhttp/okhttp3/ link since 4.x immediately redirects to 5.x.

Commit: 3555efe

JakeWharton:
That's not what I said. I'm sorry I don't have time for LLM handholding bullshit.

PR #9525, opened 2026-07-05
Handle null return from ExtendedSSLSession.getRequestedServerNames on Android SDK 27/28

### LINKED ISSUE #9523: Unhandled null return from ExtendedSSLSession.getRequestedServerNames
On some versions of Android (at least SDK 27 and 28 emulators), `ExtendedSSLSession.getRequestedServerNames` can (erroneously) return `null` instead of an empty list.

This can cause the connection to crash with `NullPointerException` when using `MockWebServer` which records the `requestedServerNames` in `RecordedRequest`.

Possibly related: https://github.com/google/conscrypt/issues/977

It's not a huge issue, we can just skip these tests on affected SDKs, but it would be nice to not have to do that.

A one line workaround is possible but unfortunately signing the CLA is a pretty heavyweight process in my organisation. I'd be able to make a pull request if the requirement to sign the CLA could be waived for a minor contribution.

tsushanth:
Fixes #9523.

On Android SDK 27 and 28, `ExtendedSSLSession.getRequestedServerNames()` can return `null` instead of an empty list (upstream bug: google/conscrypt#977). This causes a `NullPointerException` in `MockWebServer` when recording `requestedServerNames` on a `RecordedRequest`.

**Change:** one-line null guard in `Platform.getHandshakeServerNames` — replace direct `.mapNotNull` call with `(session.requestedServerNames ?: emptyList()).mapNotNull { ... }`. Already catches `UnsupportedOperationException` in the same block, so the null case fits the same defensive pattern.

### DIFF
--- a/okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/platform/Platform.kt
+++ b/okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/platform/Platform.kt
@@ -132,7 +132,7 @@ open class Platform {
   open fun getHandshakeServerNames(sslSocket: SSLSocket): List<String> {
     val session = sslSocket.session as? ExtendedSSLSession ?: return listOf()
     return try {
-      session.requestedServerNames.mapNotNull { (it as? SNIHostName)?.asciiName }
+      (session.requestedServerNames ?: emptyList()).mapNotNull { (it as? SNIHostName)?.asciiName }
     } catch (_: UnsupportedOperationException) {
       // UnsupportedOperationException – if the underlying provider does not implement the operation
       // https://github.com/bcgit/bc-java/issues/1773


yschimke (approved):

tsushanth:
The two CI failures appear to be pre-existing flakiness unrelated to this change:

- **jvm (11)**: `FastFallbackTest > preferCallConnectionOverDeferredConnection` failed with an IPv4/IPv6 connection-ordering race — the test asserts one connection type was preferred over another in a timing-sensitive scenario. The same job passes on `main` and the PR only touches `Platform.kt` (the null-check for `getRequestedServerNames`), which is on an entirely different code path.
- **android (37.0, playstore_ps16k)**: also passes on `main` at head.

Happy to re-run CI if the maintainer can trigger it, or I can rebase onto a fresh `main` commit to get a clean run.

PR #9527, opened 2026-07-06
Quote the multipart boundary when it isn't a token

### LINKED ISSUE #8068: Failed to build MultipartReader
I found that when using CloudFront CDN and OkHttp for Multipart requests, CloudFront CDN added the 'CloudFront:' field in the boundary, which caused the value of boundary to be null when building MultipartReader. It seems that the cause of the issue is due to the toMediaType method in MediaType not including ':' in its regular expression, causing it to fail. Could you kindly advise on how to solve this problem? okhttp 4.9.x

the response be like:
```
--CloudFront:ED461CAD63CDFC53A7F46FB099C72CEE
Content-Type: application/octet-stream
Content-Range: bytes xxxxx
```
and MediaType not including ':'
```
private const val TOKEN = "([a-zA-Z0-9-!#$%&'*+.^_`{|}~]+)"
```

eyupcanakman:
Fixes #8068.

`MultipartBody` builds its `Content-Type` without quoting the boundary. A boundary with a non-token character, like the colon in `abc:def`, isn't a valid token, so `toMediaType()` rejects it and `build()` throws.

This quotes the boundary when it isn't a token, per RFC 2045 section 5.1, so the Content-Type parses and round-trips through `parameter("boundary")`. A boundary containing a quote or backslash can't round-trip through the quoted-string, so it's rejected instead. Added a test for the colon boundary and one for the rejected case.

### DIFF
--- a/okhttp/src/commonJvmAndroid/kotlin/okhttp3/MultipartBody.kt
+++ b/okhttp/src/commonJvmAndroid/kotlin/okhttp3/MultipartBody.kt
@@ -35,7 +35,12 @@ class MultipartBody internal constructor(
   @get:JvmName("type") val type: MediaType,
   @get:JvmName("parts") val parts: List<Part>,
 ) : RequestBody() {
-  private val contentType: MediaType = "$type; boundary=$boundary".toMediaType()
+  private val contentType: MediaType =
+    buildString {
+      append(type)
+      append("; boundary=")
+      appendBoundary(boundary)
+    }.toMediaType()
   private var contentLength = -1L
 
   @get:JvmName("boundary")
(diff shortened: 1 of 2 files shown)


yschimke (approved):

okhttp/src/commonJvmAndroid/kotlin/okhttp3/MultipartBody.kt
```diff
@@ -354,6 +354,20 @@ class MultipartBody internal constructor(
...
+
+    /**
+     * Returns this string unchanged if it's a valid RFC 2045 token, or as a quoted-string
+     * otherwise. A boundary that contains a character like `:` isn't a token, so it must be quoted
+     * for the `Content-Type` to parse. See RFC 2045 section 5.1. A `"` or `\` can't round-trip
+     * through the quoted-string that `MediaType` reads back, so those are rejected.
+     */
+    private fun String.tokenOrQuotedString(): String {
+      if (TOKEN.matches(this)) return this
+      require('"' !in this && '\\' !in this) { "boundary contains a character that can't be quoted: $this" }
```
JakeWharton: I don't really like that this does three separate passes over the string.

And then if it has to quote it creates a string builder to produce a string only to append that to a string builder at the sole call site.

This code isn't in the hot path or anything, but it irks me a little how inefficient it is.

okhttp/src/jvmTest/kotlin/okhttp3/MultipartBodyTest.kt
```diff
@@ -102,6 +102,29 @@ class MultipartBodyTest {
...
+    assertThat(body.contentType().parameter("boundary")).isEqualTo("abc:def")
+  }
+
+  @Test
+  fun boundaryContainingQuoteIsRejected() {
+    assertFailsWith<IllegalArgumentException> {
+      MultipartBody
+        .Builder("a\"; charset=\"evil")
+        .addPart("Hello, World!".toRequestBody(null))
+        .build()
```
JakeWharton: We should only wrap the function which can fail and we need to assert on the message to ensure the code path we want to test is the one being hit (as opposed to some other validation now or in the future).

okhttp/src/commonJvmAndroid/kotlin/okhttp3/MultipartBody.kt
yschimke: I figured because the current happy path is 1 pass, then it wasn't too bad.

But 1 pass and terminating early should be possible.

eyupcanakman: Changed it to a single pass that writes straight into the `buildString`, so there's no intermediate string. It stops the token check once it knows the boundary needs quoting, and only keeps scanning to reject a `"` or `\`.

okhttp/src/jvmTest/kotlin/okhttp3/MultipartBodyTest.kt
eyupcanakman: Now only `build()` is wrapped, and it checks the message so the test can't pass on a different `IllegalArgumentException`.

PR #9536, opened 2026-07-10
Dns2, the new OkHttp DNS API

swankjesse:
Big new features are support for SVCB and being async.


okhttp/src/commonJvmAndroid/kotlin/okhttp3/dns/Dns2.kt
```diff
@@ -0,0 +1,134 @@
...
+data class DnsRequest(
+  val hostname: String,
+
+  /**
+   * The SVCB protocol name.
+   *
+   * This should be "https" for both "http://" and "https://" URLs, in order to fetch
+   * [DnsRecord.Svcb] records.
+   */
+  val protocol: String? = null,
```
swankjesse: This is weird! Leaning on ‘Attrleaf’ to qualify a DNS query
https://www.https://www.rfc-editor.org/info/rfc8552rfc-editor.org/info/rfc8552

okhttp/src/commonJvmAndroid/kotlin/okhttp3/dns/Dns2.kt
```diff
@@ -0,0 +1,134 @@
...
+
+sealed class DnsRecord private constructor() {
+  /**
+   * The hostname this record applies to.
+   *
+   * This will be different from the query's original hostname if it is served by alternate
+   * hostnames. This will be the original hostname if no alternate hostnames are in use, or if the
+   * underlying DNS resolver doesn't expose that data.
+   */
+  abstract val hostname: String
```
swankjesse: This comes up a lot in the SVCB docs (‘TargetName’) but it could also be an old-school DNS CNAME

okhttp/src/commonJvmAndroid/kotlin/okhttp3/dns/Dns2.kt
```diff
@@ -0,0 +1,134 @@
...
+  fun enqueue(callback: DnsCallback)
+  fun cancel()
+  fun isCanceled(): Boolean
+}
+
+interface DnsCallback {
+  /**
+   * @param last true if no further calls to [onRecords] will be made for this call.
+   */
+  fun onRecords(call: DnsCall, last: Boolean, records: List<DnsRecord>)
```
swankjesse: After a lot of contemplation, I think trickling values back to the caller is the only way to get best performance when DNS resolution could be recursive.

swankjesse (commented):
@yschimke I studied `DnsOverHttps` a lot in putting this together, and I’m curious about why that class makes multiple HTTP requests in order to lookup A and AAAA records, rather than making a single HTTP request with two questions. (Is that also possible?)

yschimke:
I think it practice that is basically never supported in Dns. But we should check Dns over https servers in practice

okhttp/src/commonJvmAndroid/kotlin/okhttp3/dns/Dns2.kt
yschimke: Should call also be able to tell you whether it's completed?

Or they may be different things?

okhttp/src/commonJvmAndroid/kotlin/okhttp3/dns/Dns2.kt
```diff
@@ -0,0 +1,134 @@
...
+  fun onFailure(call: DnsCall, e: IOException)
```
yschimke: Is onFailure terminal? Or can part of the DnsCall fail?

okhttp/src/commonJvmAndroid/kotlin/okhttp3/dns/Dns2.kt
```diff
@@ -0,0 +1,134 @@
...
+}
+
+// TODO: no data classes in public APIs
+data class DnsRequest(
+  val hostname: String,
```
yschimke: Is this hostname as we know?

Or Dns with a dot  - "publicobject.com."

okhttp/src/commonJvmAndroid/kotlin/okhttp3/dns/Dns2.kt
```diff
@@ -0,0 +1,134 @@
...
+
+  /**
+   * The SVCB protocol name.
+   *
+   * This should be "https" for both "http://" and "https://" URLs, in order to fetch
```
yschimke: Isn't https the default so blank?

yschimke: What about other non https ones?

mailchimp-key._domainkey.publicobject.com
google._domainkey.publicobject.com

okhttp/src/commonJvmAndroid/kotlin/okhttp3/dns/Dns2.kt
```diff
@@ -0,0 +1,134 @@
...
+   * The port to query for record types that support it, or -1 for [protocol]'s default port.
+   *
+   * In an SVCB DNS query, the protocol, port, and hostname are combined like
+   * `_8020._ftp.api.example.com`. The protocol and port segments are omitted when they hold
+   * default values.
+   */
+  val port: Int = -1,
+)
+
+sealed class DnsRecord private constructor() {
```
yschimke: Should we allow for other queries, similar to androids rawQuery?

okhttp/src/commonJvmAndroid/kotlin/okhttp3/dns/Dns2.kt
```diff
@@ -0,0 +1,134 @@
...
+   *
+   *    If an HTTPS RR query for this "https" URL re RRs or any compatible ServiceMode HTTPS RRs,
+   *    the client SHOULD behave as if it has received an HTTP 307 (Temporary Redirect) status code
+   *    with this "https" URL in the "Location" field.
+   *
+   * (TYPE_SVCB, 0x0040).
+   * (TYPE_HTTPS, 0x0041).
+   */
+  // TODO: no data classes in public APIs
+  data class Svcb(
```
yschimke: Is there value is in this being an interface? So we maybe don't need to parse all fields if not typically used?

okhttp/src/commonJvmAndroid/kotlin/okhttp3/dns/Dns2.kt
```diff
@@ -0,0 +1,134 @@
...
+    /**
+     * The socket should connect to this port, even if the URL has a different port.
+     *
+     * (`port`, 3)
+     */
+    val port: Int = -1,
+
+    /**
+     * The IP addresses of hosts known to support this configuration. If empty, assume all records
+     * with the same [DnsRecord.hostname] also support this configuration.
```
yschimke: I don't follow this

okhttp/src/commonJvmAndroid/kotlin/okhttp3/dns/Dns2.kt
swankjesse: I don’t think it’s as natural to get that information through that object. (Convince me otherwise?)

okhttp/src/commonJvmAndroid/kotlin/okhttp3/dns/Dns2.kt
swankjesse: Oooh good question. I think for the users’ benefit it should be terminal. I’ve added a note as such.

If we have a partial failure like `A` records failing and `AAAA` records succeeding, the implementation could hold back the failure until after it sends the success. That gives the application layer the opportunity to work degraded if so desired.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/dns/Dns2.kt
swankjesse: Oooh another good call out. I added more notes.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/dns/Dns2.kt
swankjesse: I’m (foolishly?) attempting to build an interface that’ll work for other protocols, so I don’t want to assume the `HTTPS` DNS query type will universally apply.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/dns/Dns2.kt
swankjesse: I don’t really like that raw API? If you’re using it you’re writing a lot of low-level DNS protocol code, but without the other benefits of being low-level, like being able to specify the DNS server or transport.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/dns/Dns2.kt
swankjesse: There’s a range of possible APIs, from low-level 'rawQuery’ to high-level `InetAddress.getAllByName()`. 

My current aspiration is to focus on connection establishment, and ignore stuff like TXT and MX records. We could lean even more into HTTP also, probably without loss of utility.

Hmm....

`HttpDns` ?!

okhttp/src/commonJvmAndroid/kotlin/okhttp3/dns/Dns2.kt
swankjesse: The overwhelming use case is OkHttp’s own use, which has a very specific set of fields we care about. If we decide we want to start honoring additional fields, we should add 'em to this class (in a binary-compatible way).

okhttp/src/commonJvmAndroid/kotlin/okhttp3/dns/Dns2.kt
swankjesse: Yeah, too complex. I added new docs in the latest commit.

yschimke (commented):
I think overall, we can't predict which records will be relevant in future. And I think rather than being Dns for HTTP only, we should model something that could evolve over time. 

So first questions that come to mind

a) is the scope purely A, AAAA, SVCB, HTTPS? or should we allow for more in future?
b) how do we make it safe when we add a new record, without breaking existing clients?

A few suggestions

1) make the result non exhaustive, I've seen this done with a private 

private object Invalid: DnsRecord()

So when clauses need an else

2) maybe support raw queries, or rather raw responses. Should this be part of DnsRecord, so any response is available raw?

3) Consider whether the abstraction is focused around OkHttp usage primarily for ease of use? or focused on modelling Dns correctly first? the service path is one example since not all fit _port._service.domain.

swankjesse:
> And I think rather than being Dns for HTTP only, we should model something that could evolve over time.

> So first questions that come to mind
> a) is the scope purely A, AAAA, SVCB, HTTPS? or should we allow for more in future?

That's a decision we gotta make! I can see a few different things to possibly build...

### 1. A general purpose DNS API.

That's gonna expose stuff like raw queries, TXT records, MX records, and recursive resolution.

I think this would be a standalone Kotlin DNS library and it could be useful for all sorts of stuff like implementing ACME.

I would like for this to exist, but not in OkHttp! I don't want the scope increase.

### 2. An API to support OkHttp’s Specific Needs Only

I think this is A, AAAA, and HTTPS records. OkHttp needs to know about alpn, ech, ports, and IP addresses. It should honor the `AliasMode` and `ServiceMode` stuff, including recursively resolving those, for best connectivity.

### 3. Split the difference

This is what I've got in this PR. It's almost all 'what does OkHttp need' plus the `protocol` field will always be "https" for us.

----

I am torn between YAGNI and it coming at a very low cost for anyone else using other protocols. Are there other protocols?! I don't expect there's many people using SVCB for other protocols.

So under a tiny bit of reflection I think we lean hard into option 2: This is an API to provide DNS information to OkHttp, and not a general-purpose DNS thing. I will drop the `protocol` parameter and simplify the docs around it.

> b) how do we make it safe when we add a new record, without breaking existing clients?

I think we do `Dns3` at that point?!

This interface is tricky because we're building an abstraction to be potentially implemented by anyone, and also potentially called by anyone.

But I'd like to assume OkHttp is the primary caller of this interface, and other callers will be rare. (Why would you call this API? Probably to do a decorator pattern?)

I'll make the `DnsRecord` class non-sealed to make an exhaustive `when` impossible.

> maybe support raw queries, or rather raw responses. Should this be part of DnsRecord, so any response is available raw?

I think raw queries makes it more difficult to implement. I'd like developers to be able to implement this interface without parsing DNS records.

 * To augment ECH information that's missing from the default DNS service? 
 * To cache more aggressively

yschimke:
@swankjesse sounds good. I didn't mind which way the answer went, just that it was self consistent.

yschimke:
@swankjesse and you should definitely do the standalone OkDns thing :)

okhttp/src/commonJvmAndroid/kotlin/okhttp3/dns/Dns2.kt
yschimke: I think we want HTTPS instead of SVCB records?

SVCB (RR type 64)
HTTPS (RR type 65)

okhttp/src/commonJvmAndroid/kotlin/okhttp3/dns/Dns2.kt
yschimke: Yeah, agreed. 

I guess if we would write this now, we might consider Flow<DnsRecord>. But we have the callback legacy, so this seems right.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/dns/Dns2.kt
yschimke: > the implementation could hold back the failure until after it sends the success

Not sure that is ideal, why introduce delays?

okhttp/src/commonJvmAndroid/kotlin/okhttp3/dns/Dns2.kt
yschimke: I don't think you need this with A, AAAA, or HTTPS records. So I wonder if we skip this complexity then?

okhttp/src/commonJvmAndroid/kotlin/okhttp3/dns/Dns2.kt
yschimke: Sounds good. OkDns can add that...

okhttp/src/commonJvmAndroid/kotlin/okhttp3/dns/Dns2.kt
yschimke: I don't think you need prefixing for HTTPS?

swankjesse: Fixed!

okhttp/src/commonJvmAndroid/kotlin/okhttp3/dns/Dns2.kt
swankjesse: I don’t expect our caller will have much to do with a partial failure. For example, if IPv6 fails after 250ms and IPv4 succeeds after 500ms, this will likely result in an ultimate success after 500 ms.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/dns/Dns2.kt
swankjesse: Fixed!

okhttp/src/commonJvmAndroid/kotlin/okhttp3/dns/Dns2.kt
swankjesse: Agreed.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/dns/Dns2.kt
yschimke: I thought you suggested the failure at 250ms was terminal?

okhttp/src/commonJvmAndroid/kotlin/okhttp3/dns/Dns2.kt
```diff
@@ -0,0 +1,173 @@
...
+  fun onRecords(call: DnsCall, last: Boolean, records: List<DnsRecord>)
+
+  /**
+   * This is a terminal event and no further calls to this callback will be made for this call.
+   */
+  fun onFailure(call: DnsCall, e: IOException)
+}
+
+// TODO: no data classes in public APIs
+data class DnsRequest(
```
yschimke: How can I implement a network specific Android DnsRequest?

Should we have tags?

swankjesse (commented):
Next steps:

Change DnsSystem to implement this, in a very basic way that doesn’t actually add any new `ServiceMetadata`. It will still implement `Dns`. Use `TaskQueue` for callback concurrency.

Build a not-public-API adapter that turns a `Dns` instance into a `Dns2` instance. It should also use `TaskQueue` for concurrency.

Change `OkHttpClient` to use `Dns2` internally instead of `Dns`. Use the adapter to make `Dns2` the canonical implementation. This will want a breaking change in `Address`, which expects a `Dns`. Can we avoid that? This may be difficult!

Implement `Dns2` in `DnsOverHttps`. It will still implement `Dns`.

Build an implementation of `Dns2` on Android’s `DnsResolver`. Enable that implementation by default on Android where it is available.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/dns/Dns2.kt
swankjesse: I think we have a Network-specific Dns2 implementation, and select that in an Interceptor.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/dns/Dns2.kt
swankjesse: The Dns2 implementation can withhold onFailure() calls. (Alternatively, we have a `last` boolean in the `onFailure` function?)

yschimke:
> This will want a breaking change in Address, which expects a Dns

OkHttp 6?

Dns2 extends Dns, and we keep identifying as Dns.

Add an extra dns2 field?

okhttp/src/commonJvmAndroid/kotlin/okhttp3/dns/Dns2.kt
yschimke: That's the delay I meant. I think we should signal that non-terminal failure, not hide it.  Also is one failure, such as HTTPS actually an overall failure?

Maybe it's HTTPS not being found, and I want to change something when ECH not configured?

Maybe it's IPv6, and I want to use HTTPS address hints?

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Dns2.kt
```diff
@@ -0,0 +1,250 @@
...
+     * @param last true if this is the last list of records for this address. That is a terminal
+     *   event and no further calls to this callback will be made for this call.
+     * @param records a possibly-empty set of records received from the name server.
+     */
+    fun onRecords(call: Call, last: Boolean, records: List<Record>)
+
+    /**
+     * This is a terminal event and no further calls to this callback will be made for this call.
+     */
+    fun onFailure(call: Call, e: IOException)
```
yschimke: Do we chain these via suppressions? Let's say HTTPS works with hints, but A and AAAA fai?

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Dns2.kt
```diff
@@ -0,0 +1,250 @@
...
+
+    override fun toString(): String {
+      return when (port) {
+        443 -> hostname
+        else -> "$hostname:$port"
+      }
+    }
+  }
+
+  sealed class Record private constructor() {
```
yschimke, line 117: Can we force this non exhaustive?

yschimke (approved):

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Dns2.kt
```diff
@@ -0,0 +1,250 @@
...
+ * Loads IP addresses and service metadata for a hostname.
+ *
+ * This interface is exclusively concerned with collecting the information necessary to establish
+ * new HTTP and HTTPS connections. Typical implementations will read `A` (IPv4), `AAAA` (IPv6) and
+ * `HTTPS` (service metadata) records only. Use a different API to read other record types, or to
+ * write DNS records.
+ *
+ * Implementations of this interface must be safe for concurrent use.
+ */
+interface Dns2 {
```
yschimke, line 33: Should we use an experimental annotation for a couple of releases?

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Dns2.kt
swankjesse: We can, but I don't think we need to. 

The design of SVCB/HTTPS records is extensible, so if new data is invented (say quantum-resistant ECH), that'll most likely be added as a new parameter on the HTTPS record instead of being a new record type.

There's a strong incentive to avoid new record types, because there's no atomicity across records. And because it takes more time to define and roll out vs. adding a new parameter.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Dns2.kt
swankjesse: I'm unsure. If we make it opt-in that makes it more difficult for us to use it in types like OkHttpClient and Address.

The impossible challenge is "what will we regret?" and I think you've come up with a few possibilities! Partial failure for example.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Dns2.kt
swankjesse: Probably.

I don't like situations where I'm digging through suppressed exceptions programmatically. So if there's some useful data exposed that way, I'm inclined to expose it in a different way.

swankjesse:
> Add an extra dns2 field?

That's where I intend to start. And if/when we do OkHttp6 we might be able to drop the old `dns` field.

PR #9537, opened 2026-07-11
Detect overflow when decoding DER variable-length integers

basavaraj-sm05:
DerReader decodes ASN.1 variable-length integers for OBJECT IDENTIFIER subidentifiers and high-number tags, building the value up with a shift left of 7 per continuation byte. That accumulation had no overflow guard (it carried a `// TODO: detect overflow`), so a subidentifier padded with enough continuation bytes silently wraps a Long and the shift can flip the sign. I ran into it decoding a certificate whose OID had an over-long arc: instead of rejecting the input, `readObjectIdentifier` handed back a bogus arc like `2.9223372036854775728` produced by the wrapped arithmetic. Since these bytes come straight off an untrusted certificate, quietly turning a malformed OID into a real-looking one seemed worth closing off. This throws a `ProtocolException` once the running value no longer fits in a Long, the same way the length decoder just above already rejects a length past `Long.MAX_VALUE`. Values up to `Long.MAX_VALUE` still decode exactly as before, and there's a regression test for the overflowing case.

### DIFF
--- a/okhttp-tls/src/main/kotlin/okhttp3/tls/internal/der/DerReader.kt
+++ b/okhttp-tls/src/main/kotlin/okhttp3/tls/internal/der/DerReader.kt
@@ -305,12 +305,16 @@ internal class DerReader(
 
   /** Used for tags and subidentifiers. */
   private fun readVariableLengthLong(): Long {
-    // TODO(jwilson): detect overflow.
     var result = 0L
     while (true) {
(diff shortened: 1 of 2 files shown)


yschimke (approved):
Thanks for the fix.

basavaraj-sm05:
gentle ping

PR #9545, opened 2026-07-14
Write DNS messages

swankjesse:


swankjesse:
This is motivated more for testing than anything. There‘s a bunch of tricky DNS HTTPS cases and I’ll need to test ’em all.

JakeWharton (approved):

PR #9548, opened 2026-07-15
read DNS RDLENGTH as an unsigned 16-bit value

basavaraj-sm05:
readResourceRecord pulls the record data length out of the RDLENGTH field, which RFC 1035 defines as an unsigned 16-bit integer, but it decodes that field with readShort().toLong(), so any length with the high bit set comes back negative. Every other 16-bit field in this reader already goes through toUShort(), so this one is the odd one out. For an A or AAAA record the wrong sign is caught by the exact-length check, but for a record type we don't decode the code falls into skip(recordDataLength), and okio's skip treats a negative count as a no-op rather than an error. That means a record legitimately carrying 32768 or more bytes never gets consumed, the reader stays parked in the middle of the message, and the leftover bytes either trip the trailing-data check or get read as the next record. I hit it while feeding the reader a response with a large unknown record sitting in front of an A record. Decoding the field as unsigned, the same way the counts and the HTTPS parameter lengths already are, keeps the stream in sync, and I added a regression test that reads a message with an oversized unsupported record.

### DIFF
--- a/okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/dns/-DnsMessageReader.kt
+++ b/okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/dns/-DnsMessageReader.kt
@@ -151,7 +151,7 @@ class DnsMessageReader(
     val type = readShort().toInt()
     val `class` = readShort().toInt()
     val timeToLive = readInt()
-    val recordDataLength = readShort().toLong()
+    val recordDataLength = readUShort().toLong()
 
     when {
       `class` == CLASS_IN && type == TYPE_A -> {
--- a/okhttp/src/commonTest/kotlin/okhttp3/internal/dns/DnsMessageReaderWriterTest.kt
+++ b/okhttp/src/commonTest/kotlin/okhttp3/internal/dns/DnsMessageReaderWriterTest.kt
@@ -157,6 +157,54 @@ class DnsMessageReaderWriterTest {
     assertThat(e).hasMessage("malformed DNS message")
   }
 
+  /**
+   * An RDLENGTH is a 16-bit unsigned field, so a record may legitimately declare up to 65535 bytes
+   * of data. A value with the high bit set must still be skipped for record types we don't decode,
+   * otherwise the reader loses sync with the stream.
+   */
+  @Test
+  fun `unsupported record with high-bit record length is skipped`() {
+    val buffer =
+      Buffer().apply {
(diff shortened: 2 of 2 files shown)


okhttp-dnsoverhttps/src/main/kotlin/okhttp3/dnsoverhttps/-DnsMessageReader.kt
```diff
@@ -140,7 +140,7 @@ internal class DnsMessageReader(
     val type = readShort().toInt()
     val `class` = readShort().toInt()
     val timeToLive = readInt()
-    val recordDataLength = readShort().toLong()
+    val recordDataLength = readShort().toUShort().toLong()
```
swankjesse:
```suggestion
    val recordDataLength = readUShort().toLong()
```

(also needs an import)

okhttp-dnsoverhttps/src/main/kotlin/okhttp3/dnsoverhttps/-DnsMessageReader.kt
swankjesse, line 28:
```suggestion
import okio.buffer
import okio.readUShort
```

swankjesse (approved):
Nice test!

okhttp-dnsoverhttps/src/main/kotlin/okhttp3/dnsoverhttps/-DnsMessageReader.kt
basavaraj-sm05: Tried this and it doesn't compile, readUShort isn't in okio yet. I checked 3.17.0 (what this repo is on) and 3.18.0, and neither has it on BufferedSource or as an extension in the okio package. I've kept readShort().toUShort().toLong() for now since it matches the count fields and SvcParam lengths above; happy to switch all the call sites over once okio grows readUShort.

basavaraj-sm05: Done. The DNS code moved to okhttp3/internal/dns on master since I opened this, so I rebased and carried the change and the regression test over to the new location. That file already imports okio.readUShort for the counts and the SvcParam lengths, so it comes out as just the one-line swap to readUShort().toLong(). Test and spotless pass on the rebased branch.

swankjesse:
Thank you! 🚀

PR #9551, opened 2026-07-16
Reject malformed UTCTime and GeneralizedTime in DER parsing

basavaraj-sm05:
The DER reader decodes UTCTime and GeneralizedTime through SimpleDateFormat, but it leaves the formatter in its default lenient mode and parses with the single-argument parse(), which only anchors at the start of the string. Because these value bytes come straight off a certificate, an out-of-range field like a month of 13 gets silently rolled over into a neighboring month instead of being rejected, and any bytes after the trailing 'Z' that still sit inside the element's declared length are quietly ignored. So two different encodings can decode to the same instant, and a malformed timestamp comes back looking like a real one, which is the same shape of problem as accepting an over-long OID arc. I noticed it while comparing these two functions against the HTTP date parser in the same codebase, which already sets isLenient=false and confirms the whole string was consumed. This switches lenience off and uses a ParsePosition so the parse has to reach the end of the value, matching that existing pattern. Valid canonical times still decode exactly as before, including the two-digit-year cutoff at 1950, and I added regression tests for the out-of-range and trailing-byte cases.

### DIFF
--- a/okhttp-tls/src/main/kotlin/okhttp3/tls/internal/der/Adapters.kt
+++ b/okhttp-tls/src/main/kotlin/okhttp3/tls/internal/der/Adapters.kt
@@ -17,7 +17,7 @@ package okhttp3.tls.internal.der
 
 import java.math.BigInteger
 import java.net.ProtocolException
-import java.text.ParseException
+import java.text.ParsePosition
 import java.text.SimpleDateFormat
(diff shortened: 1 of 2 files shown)


yschimke (approved):

basavaraj-sm05:
any update?

PR #9553, opened 2026-07-16
DnsOverHttps Fixups: event sequencing, resolvePublicAddresses, IPv6 first

swankjesse:
The biggest change here is we’re now requesting TYPE_AAAA (IPv6) records first in all DnsOverHttps calls. This should be a negligible behavior change for most applications as the two calls can execute in parallel. The order only really matters when there’s a limit on the total concurrency against the DNS service. (This is the case for our tests, which only permit 1 call at a a time.)

Fix some bugs in how events are delivered to the callback. Always deliver events on OkHttp’s dispatcher thread. This includes cancel events and resolvePublicAddress validation. Otherwise it’s surprising if the callback is *sometimes* invoked by the thread calling `enqueue()`.

### DIFF
--- a/okhttp-dnsoverhttps/src/main/kotlin/okhttp3/dnsoverhttps/-DnsOverHttpsCall.kt
+++ b/okhttp-dnsoverhttps/src/main/kotlin/okhttp3/dnsoverhttps/-DnsOverHttpsCall.kt
@@ -38,6 +38,7 @@ import okhttp3.internal.testAndSet
 internal class DnsOverHttpsCall(
   override val request: Dns2.Request,
   private val calls: List<Call>,
+  private val canceledException: IOException?,
 ) : Dns2.Call,
   Callback {
   private val state = AtomicReference<State>(State.Idle)
@@ -69,7 +70,10 @@ internal class DnsOverHttpsCall(
           ?: return // Already complete or canceled; nothing to do.
 
       val newRunningCalls = previous.runningCalls - call
-      val allFailures = previous.delayedFailures + e
+      val allFailures = when {
+        canceledException != null -> listOf(canceledException)
+        else -> previous.delayedFailures + e
+      }
       val next =
         when {
           newRunningCalls.isEmpty() -> {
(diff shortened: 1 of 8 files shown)


okhttp-dnsoverhttps/src/main/kotlin/okhttp3/dnsoverhttps/-DnsOverHttpsCall.kt
```diff
@@ -182,22 +167,33 @@ internal class DnsOverHttpsCall(
...
   override fun cancel() {
-    val previous = state.testAndSet(State.Canceled) { it is State.Running || it == State.Idle }
-    (previous as? State.Running)?.callback?.onFailure(this, IOException("canceled"))
+    for (call in calls) {
+      call.cancel()
+    }
   }
 
-  override fun isCanceled() = state.get() is State.Canceled
+  override fun isCanceled() = calls.all { it.isCanceled() }
```
yschimke: I think given RealCall, this works in practice.

But based on the interface of Call, it could be racy?

```
  /** Cancels the request, if possible. Requests that are already complete cannot be canceled. */
  fun cancel()
```

So if some are completed, and you are using some other Call impl, which I know is unlikely, it may not be correct when one call already completed?

yschimke (approved):
one question about cancellation

okhttp-dnsoverhttps/src/main/kotlin/okhttp3/dnsoverhttps/-DnsOverHttpsCall.kt
swankjesse: Ooooh, yeah I totally missed that. Good analysis. I added a test case that handles `RealCall`, but it’s arguably a bug that `RealCall` allows you to cancel something that’s already completed.

swankjesse: And I added a `volatile` to track whether this is canceled, similar to how `RealCall` does it. I previously had a `Canceled` state (and `LookupDnsCall` still does!), so I might refactor to do that also.

PR #9557, opened 2026-07-17
 Start including ECH data in RouteSelector results

swankjesse:
This changes RouteSelector to use our new async API, when it is implemented.

### DIFF
--- a/okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/connection/RouteSelector.kt
+++ b/okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/connection/RouteSelector.kt
@@ -21,10 +21,13 @@ import java.net.InetSocketAddress
 import java.net.Proxy
 import java.net.SocketException
 import java.net.UnknownHostException
+import java.util.concurrent.CompletableFuture
 import okhttp3.Address
+import okhttp3.Dns
 import okhttp3.HttpUrl
 import okhttp3.Route
 import okhttp3.internal.canParseAsIpAddress
+import okhttp3.internal.dns.LookupDnsCall
 import okhttp3.internal.immutableListOf
 import okhttp3.internal.toImmutableList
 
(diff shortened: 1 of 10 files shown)


okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/dns/-ExecuteDns.kt
```diff
@@ -0,0 +1,59 @@
...
+      val allRecords = mutableListOf<Dns.Record>()
+
+      override fun onRecords(
+        call: Dns.Call,
+        last: Boolean,
+        records: List<Dns.Record>,
+      ) {
+        allRecords += records
+        if (last) {
+          future.complete(allRecords)
```
yschimke: Should this complete exceptionally if empty?


     * @param records a possibly-empty set of records received from the name server.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/connection/RouteSelector.kt
```diff
@@ -185,30 +188,71 @@ class RouteSelector internal constructor(
...
+    call.eventListener.dnsStart(
+      call = call,
+      domainName = socketHost,
+    )
+
+    val dnsRequest = Dns.Request(socketHost)
+    val result =
+      when (val dnsCall = address.dns.newCall(dnsRequest)) {
+        is LookupDnsCall -> {
+          val inetAddresses = address.dns.lookup(socketHost)
```
yschimke, line 214: It's nice this falls back to the current prod logic.

Is there any scenario where it's important we use the dns and hostname from the LookupDnsCall? Say if someone overrides Dns with a version that calls super.newCall?

Or I'm worring too much?

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/connection/RouteSelector.kt
```diff
@@ -185,30 +188,71 @@ class RouteSelector internal constructor(
...
+
+          val hostnameToServiceMetadata =
+            records
+              .filterIsInstance<Dns.Record.ServiceMetadata>()
+              .associateBy { it.hostname }
+
+          records
+            .filterIsInstance<Dns.Record.IpAddress>()
+            .map { record ->
+              val serviceMetadata = hostnameToServiceMetadata[record.hostname]
```
yschimke, line 235: Is this correct? if the DNS record is pointing to multiple backend DNS Servers?

```
;; Multiple HTTPS records for the same domain
example.com.  3600  IN  HTTPS  1 server1.anking.com. alpn="h2,h3"
example.com.  3600  IN  HTTPS  2 server2.anking.com. alpn="h2,h3"
```

Is that possible?

yschimke (approved):
1 question about hostname filtering

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/dns/-ExecuteDns.kt
swankjesse: I don’t think so. It’s nice to differentiate between ‘no results’ and failure.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/connection/RouteSelector.kt
swankjesse: I think if your `Dns` implementation overrides those things for `newCall()`, it’s also obligated to override those for `lookup()`.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/connection/RouteSelector.kt
swankjesse: You got it. And now that I’m staring at it, I also need to handle `""` there. Fixed!

swankjesse: (I’m hoping to follow up with code in DnsOverHttps that honors `ServiceMode` and `AliasMode`, by making recursive calls. It’s a bit like CNAME handling.)

PR #9561, opened 2026-07-19
Serialize calls to Dns.Callback

swankjesse:
We had a bug where these calls could be made in parallel. Making them serial is a bit awkward but doesn't require any locks if we're clever.

### DIFF
--- a/okhttp-dnsoverhttps/src/main/kotlin/okhttp3/dnsoverhttps/internal/-DnsOverHttpsCall.kt
+++ b/okhttp-dnsoverhttps/src/main/kotlin/okhttp3/dnsoverhttps/internal/-DnsOverHttpsCall.kt
@@ -27,12 +27,34 @@ import okhttp3.Response
 import okhttp3.internal.OkHttpInternalApi
 import okhttp3.internal.testAndSet
 
-// TODO: honor resolvePrivateAddresses, resolvePublicAddresses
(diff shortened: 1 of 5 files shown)


okhttp-dnsoverhttps/src/main/kotlin/okhttp3/dnsoverhttps/internal/-DnsOverHttpsCall.kt
```diff
@@ -144,47 +141,102 @@ internal class DnsOverHttpsCall(
         }
       }
 
+    updateStateAndCallCallbacks(
+      completedCall = call,
+      newRecords = dnsRecords,
+    )
+  }
+
+  private tailrec fun updateStateAndCallCallbacks(
```
yschimke, line 150: Nice. I am curious if it's materially faster than simple synchronisation?

yschimke (approved):

okhttp-dnsoverhttps/src/main/kotlin/okhttp3/dnsoverhttps/internal/-DnsOverHttpsCall.kt
swankjesse: Probably not! But I prefer lock-free 'cause it only covers data, not execution.

PR #9562, opened 2026-07-19
Reject malformed messages that have infinite loops

swankjesse:


swankjesse:
I validated my approach by comparing vs. Bind!
https://gitlab.isc.org/isc-projects/bind9/-/blob/main/lib/dns/name.c?ref_type=heads#L1289

JakeWharton (approved):

PR #9565, opened 2026-07-19
Rename includeHttps to includeServiceMetadata

swankjesse:
'HTTPS' is way too confusing for DnsOverHttps, because it's about a DNS record type and not the transport for which records are retrieved.

### DIFF
--- a/okhttp-dnsoverhttps/src/test/java/okhttp3/dnsoverhttps/DnsOverHttpsTest.kt
+++ b/okhttp-dnsoverhttps/src/test/java/okhttp3/dnsoverhttps/DnsOverHttpsTest.kt
@@ -797,7 +797,7 @@ class DnsOverHttpsTest(
       .Builder()
       .client(bootstrapClient)
       .includeIPv6(includeIPv6)
-      .includeHttps(includeHttps)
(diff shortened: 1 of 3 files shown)


okhttp-dnsoverhttps/src/test/java/okhttp3/dnsoverhttps/DnsOverHttpsTest.kt
```diff
@@ -797,7 +797,7 @@ class DnsOverHttpsTest(
       .Builder()
       .client(bootstrapClient)
       .includeIPv6(includeIPv6)
-      .includeHttps(includeHttps)
+      .includeServiceMetadata(includeHttps)
```
JakeWharton: Update the param here on the test helper too?

JakeWharton (approved):

okhttp-dnsoverhttps/src/test/java/okhttp3/dnsoverhttps/DnsOverHttpsTest.kt
swankjesse: Good catch. Fixed!

PR #9569, opened 2026-07-21
Split the state machine out of DnsOverHttps

swankjesse:
There's a few things that are a bit awkward, but overall it seems like it'll afford swapping in UDP or Android's resolver.

### DIFF
--- a/okhttp-dnsoverhttps/src/main/kotlin/okhttp3/dnsoverhttps/internal/DnsCallStateMachine.kt
+++ b/okhttp-dnsoverhttps/src/main/kotlin/okhttp3/dnsoverhttps/internal/DnsCallStateMachine.kt
@@ -1,0 +1,338 @@
+/*
+ * Copyright (c) 2026 OkHttp Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okhttp3.dnsoverhttps.internal
+
+import java.io.IOException
+import java.net.UnknownHostException
+import java.util.concurrent.atomic.AtomicReference
+import okhttp3.Dns
+import okhttp3.Protocol
+
+/**
+ * State machine for DNS calls. This is intended for use with any transport for the queries, such
+ * as UDP or DNS over HTTPS.
+ *
+ * Concurrency
+ * -----------
+ *
+ * A few things conspire to make concurrency tricky:
+ *
+ *  * Each DNS record type is queried in parallel; [onQueryResponse] and [onQueryFailure] may be called
+ *    concurrently.
+ *  * Calls to [okhttp3.Dns.Callback] must be serialized.
+ *  * We don't want to use locks to guard access to [okhttp3.Dns.Callback] functions.
+ *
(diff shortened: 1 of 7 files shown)


okhttp-dnsoverhttps/src/main/kotlin/okhttp3/dnsoverhttps/internal/DnsCallStateMachine.kt
```diff
@@ -0,0 +1,338 @@
...
+package okhttp3.dnsoverhttps.internal
+
+import java.io.IOException
+import java.net.UnknownHostException
+import java.util.concurrent.atomic.AtomicReference
+import okhttp3.Dns
+import okhttp3.Protocol
+
+/**
+ * State machine for DNS calls. This is intended for use with any transport for the queries, such
```
yschimke, line 25: Thanks!

okhttp-dnsoverhttps/src/main/kotlin/okhttp3/dnsoverhttps/internal/DnsCallStateMachine.kt
```diff
@@ -0,0 +1,338 @@
...
+    get() = state.get().canceled
+
+  fun start(callback: Dns.Callback) {
+    val queryMessages =
+      buildList {
+        if (includeServiceMetadata) {
+          add(DnsMessage.query(call.request.hostname, TYPE_HTTPS))
+        }
+        if (includeIPv6) {
+          add(DnsMessage.query(call.request.hostname, TYPE_AAAA))
```
yschimke, line 70: If I wanted to use this for Android and I have two calls to make

a) async blocking InetAddress.getAllByName
b) DnsResolver rawQuery

is that stretching it past being a useful correct abstraction?

What would the calls be? and what would the Transport return as Q?

yschimke (approved):

okhttp-dnsoverhttps/src/main/kotlin/okhttp3/dnsoverhttps/internal/DnsCallStateMachine.kt
swankjesse: Could you use DnsResolver for TYPE_A, TYPE_AAAA, and TYPE_HTTPS ?

yschimke: No, we lose a lot of caching and RFC level sorting if we do that.

Even if we break out the RFCs and fix a bunch of things, I've been advised that will perform worse.

We should call either InetAddress.getAllByName() or Network.getAllByName()

swankjesse: Gotcha. 

Does that drawback apply to DnsOverHttps also? Might be worth fixing regardless.

I suspect the ‘perform worse’ is gonna be due to operating system-layer caching? (I can’t think of anything else that’d allow the syscall to run faster than what we can do.)

yschimke: No, sort of, maybe, yes?

perform worse is likely

- Android caching in the java layer, avoiding actual DNS work, and resetting when network changes. and other things.
- OkHttp does not correctly sort ip addresses in a way that would avoid some obvious bad patterns.

swankjesse: Let’s correctly sort IP addresses!

swankjesse: First cache PR is out, don’t threaten me with a good time.

I think DnsOverHttps should continue to do on-disk caching, but also in-memory caching. I’d like to help our users deploy DoH as widely as possible and I don’t want bad caching to be a barrier to that.

PR #9570, opened 2026-07-21
Fix broken/outdated links

Marcono1234:
(The https://github.com/square/ links are not actually broken; GitHub simply redirects them. But might be better nonetheless to directly link to https://github.com/lysine-dev/.)

Similar to https://github.com/lysine-dev/okio/pull/1827 there are a few other (non-link) references to Square where you as maintainers might know best how to or if to update them.

### DIFF
--- a/deploy_website.sh
+++ b/deploy_website.sh
@@ -6,7 +6,7 @@
 
 set -ex
 
-REPO="git@github.com:square/okhttp.git"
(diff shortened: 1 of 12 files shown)


Marcono1234:
Hmm, there was apparently already #9552; not sure why it missed those links.

PR #9573, opened 2026-07-21
Add Android ECH support

yschimke:
Initial version before dealing with NetworkSecurityPolicy,

instead just activated by developer using AndroidDns, and a server having HTTPS records with ECH.


yschimke:
@swankjesse my biggest question is how to deal with NetworkSecurityPolicy for both
a) activation - is that AndroidDns? is that just whether to include for a specific host
b) error handling / fallback

I previously ended up with EchModes (disabled, strict and so on) read from the policy

https://github.com/lysine-dev/okhttp/pull/9383/changes#diff-62659f553c144abdf5c3d578b9d14e96bf069b4ef9e7f93549b04ff38e4fcbc9

okhttp/src/androidMain/kotlin/okhttp3/android/AndroidDns.kt
```diff
@@ -0,0 +1,212 @@
...
+@RequiresApi(29)
+@SuppressSignatureCheck
+class AndroidDns
+  @JvmOverloads
+  constructor(
+    private val dnsResolver: DnsResolver = DnsResolver.getInstance(),
+    private val network: Network? = null,
+    /**
+     * True to also query the `HTTPS` record for service metadata such as ECH. Set this to false to
+     * save a query when only IP addresses are needed.
```
swankjesse: I wonder if there’s a way to phrase this documentation to encourage people to keep it on.

‘Set this to false to omit the service metadata query. This will disable privacy features in the HTTPS call.’

okhttp/src/androidMain/kotlin/okhttp3/android/AndroidDns.kt
```diff
@@ -0,0 +1,212 @@
...
+     */
+    private val includeServiceMetadata: Boolean = true,
+    // Runs inline; the executor only hands off DnsResolver's callbacks.
+    private val executor: Executor = Executor { it.run() },
+  ) : Dns {
+    /**
+     * Resolves addresses only, for callers using the legacy blocking API. This delegates to
```
swankjesse: Does this do a service metadata query?

okhttp/src/androidMain/kotlin/okhttp3/android/AndroidDns.kt
```diff
@@ -0,0 +1,212 @@
...
+        .execute()
+        .filterIsInstance<Dns.Record.IpAddress>()
+        .map { it.address }
+
+    override fun newCall(request: Dns.Request): Dns.Call = AndroidDnsCall(request)
+
+    private inner class AndroidDnsCall(
+      override val request: Dns.Request,
+    ) : Dns.Call,
+      DnsCallStateMachine.Transport<AndroidDnsCall.Query> {
```
swankjesse: The paint isn’t even dry on this API !

okhttp/src/androidMain/kotlin/okhttp3/android/AndroidDns.kt
```diff
@@ -0,0 +1,212 @@
...
+      private val stateMachine =
+        DnsCallStateMachine(
+          transport = this,
+          call = this,
+          canceledException = null,
+          // A single `A` query to represent the blocking InetAddress call for both.
```
swankjesse: 👍🏻

okhttp/src/androidMain/kotlin/okhttp3/android/AndroidDns.kt
```diff
@@ -0,0 +1,212 @@
...
+      override fun enqueue(query: Query) {
+        when (query.type) {
+          TYPE_A -> resolveAddresses(query)
+          else -> queryServiceMetadata(query)
+        }
+      }
+
+      /**
+       * Cancels a query. The system resolver doesn't call back once canceled, so this completes the
+       * query itself to keep the state machine from stalling.
```
swankjesse: Yeah, I hope this wasn’t too painful to discover.

(I believe calling back after cancelation is the only reasonable behavior)

okhttp/src/androidMain/kotlin/okhttp3/android/AndroidDns.kt
```diff
@@ -0,0 +1,212 @@
...
+          object : Task("${request.hostname} address lookup", cancelable = false) {
+            override fun runOnce(): Long {
+              try {
+                val addresses =
+                  when (network) {
+                    null -> InetAddress.getAllByName(request.hostname)
+                    else -> network.getAllByName(request.hostname)
+                  }
+                val records = addresses.map { Dns.Record.IpAddress(request.hostname, it) }
+                query.complete { stateMachine.onQueryRecords(query, records) }
```
swankjesse: neat

okhttp/src/androidMain/kotlin/okhttp3/android/AndroidDns.kt
```diff
@@ -0,0 +1,212 @@
...
+            override fun onAnswer(
+              answer: ByteArray,
+              rcode: Int,
+            ) {
+              val message =
+                try {
+                  DnsMessageReader(Buffer().write(answer)).read()
+                } catch (_: IOException) {
+                  // TODO should a malformed HTTPS record fail the lookup?
+                  // or rather, how do we hook into the Network Security Policy?
```
swankjesse: I would like for this to call `stateMachine.onFailure()` and let the `Dns.Callback` decide what to do

swankjesse: I deliberately made `Dns.Callback` accept an `IOException` and not `UnknownHostException`, so in theory users of the API can differentiate between absent data and errors.

okhttp/src/androidMain/kotlin/okhttp3/android/AndroidDns.kt
```diff
@@ -0,0 +1,212 @@
...
+              }
+            }
+
+            override fun onError(e: DnsResolver.DnsException) {
+              query.complete { stateMachine.onQueryRecords(query, listOf()) }
+            }
+          }
+
+        try {
+          dnsResolver.rawQuery(
```
swankjesse, line 192: Nice

okhttp/src/androidMain/kotlin/okhttp3/android/AndroidDns.kt
```diff
@@ -0,0 +1,212 @@
...
+            /* network = */ network,
+            /* domain = */ request.hostname,
+            /* nsClass = */ DnsResolver.CLASS_IN,
+            /* nsType = */ TYPE_HTTPS,
+            /* flags = */ DnsResolver.FLAG_EMPTY,
+            /* executor = */ executor,
+            /* cancellationSignal = */ query.cancellationSignal,
+            /* callback = */ queryCallback,
+          )
+        } catch (_: Exception) {
```
swankjesse: Why catch? This is not documented to throw? (And if it does throw, we should ask the corresponding team to fix their docs)

okhttp/src/androidMain/kotlin/okhttp3/android/AndroidDns.kt
```diff
@@ -0,0 +1,212 @@
...
+      /** One outstanding query. Each completes the state machine's query at most once. */
+      inner class Query(
+        dnsMessage: DnsMessage,
+      ) {
+        val type: Int = dnsMessage.questions.single().type
+        val cancellationSignal = if (includeServiceMetadata) CancellationSignal() else null
+        private val completed = AtomicBoolean()
+
+        fun complete(block: () -> Unit) {
+          if (completed.compareAndSet(false, true)) block()
```
swankjesse: rad

okhttp/src/androidMain/kotlin/okhttp3/android/AndroidDns.kt
```diff
@@ -0,0 +1,212 @@
...
+          // TODO consider whether this is right and how Android network policy can be enforced?
+          query.complete { stateMachine.onQueryRecords(query, listOf()) }
+        }
+      }
+
+      /** One outstanding query. Each completes the state machine's query at most once. */
+      inner class Query(
+        dnsMessage: DnsMessage,
+      ) {
+        val type: Int = dnsMessage.questions.single().type
```
swankjesse: nit:
```suggestion
        val type: Int
          get() = dnsMessage.questions.single().type
```

okhttp/src/androidMain/kotlin/okhttp3/internal/platform/android/Android17SocketAdapter.kt
```diff
@@ -0,0 +1,100 @@
...
+import okhttp3.Protocol
+import okhttp3.internal.SuppressSignatureCheck
+import okhttp3.internal.platform.Platform
+import okhttp3.internal.platform.Platform.Companion.isAndroid
+import okio.ByteString
+
+/**
+ * Socket adapter for Android 17+ platform TLS APIs.
+ *
+ * Specifically supports setEchConfigList for ECH goodness.
```
swankjesse, line 36: 💯 💯 💯 💯 💯

okhttp/src/androidMain/kotlin/okhttp3/internal/platform/android/Android17SocketAdapter.kt
```diff
@@ -0,0 +1,100 @@
...
+
+        if (hostname != null && echConfigList != null) {
+          // TODO put behind Network Policy
+
+          // The ECH config was resolved during DNS (RealCall.resolveAddresses); just apply it here.
+          // A config we can't parse is skipped, and the connection proceeds without ECH.
+          val echConfig =
+            try {
+              EchConfigList.fromBytes(echConfigList.toByteArray())
+            } catch (_: InvalidEchDataException) {
```
swankjesse: This ain’t great. Reducing privacy on misconfigured data is a bad look.

Just let this throw all the way to the caller instead?

okhttp/src/androidMain/kotlin/okhttp3/internal/platform/android/Android17SocketAdapter.kt
```diff
@@ -0,0 +1,100 @@
...
+              null
+            }
+
+          if (echConfig != null) {
+            SSLSockets.setEchConfigList(sslSocket, echConfig)
+          }
+        }
+      } catch (iae: IllegalArgumentException) {
+        // Android 17 early betas threw when ECH Config wasn't set in HTTPS record.
```
swankjesse: Got a link to anything for this? Just for my own curiosity and not necessarily to link from here

swankjesse (approved):
THIS IS IT

swankjesse:
> I previously ended up with EchModes (disabled, strict and so on) read from the policy

My preference is we assume DNS is where this should be configured. If there’s a ECH parameter on the TYPE_HTTPS record we should use it, and if there isn’t then we should assume the site operators don‘t care.

Configuring it on the client is too difficult for large organizations, because the team that ships services and apps is far detached from the team that manages DNS and HTTPS.

okhttp/src/androidMain/kotlin/okhttp3/internal/platform/android/Android17SocketAdapter.kt
yschimke: But this is (or was) a policy in Android.

They removed the required mode? So maybe enabled is now required?

https://cs.android.com/android/_/android/platform/frameworks/base/+/14b56d11ef2f36ad3335d757110d1020bcd6d00b

okhttp/src/androidMain/kotlin/okhttp3/internal/platform/android/Android17SocketAdapter.kt
yschimke: It's the one you saw before, yep, I'll link.

yschimke:
> > I previously ended up with EchModes (disabled, strict and so on) read from the policy
> 
> My preference is we assume DNS is where this should be configured. If there’s a ECH parameter on the TYPE_HTTPS record we should use it, and if there isn’t then we should assume the site operators don‘t care.
> 
> Configuring it on the client is too difficult for large organizations, because the team that ships services and apps is far detached from the team that manages DNS and HTTPS.

But there is a developer API that we should honour.

https://developer.android.com/privacy-and-security/security-config#EncryptedClientHelloSummary

okhttp/src/androidMain/kotlin/okhttp3/android/AndroidDns.kt
yschimke: I think maybe we end up defaulting to the networkSecurityConfig, but let developers opt-out, or opt-in.

For now, likely opt-in while we confirm it works.

~GREASE means it should never cause failures just by being enabled.~

okhttp/src/androidMain/kotlin/okhttp3/android/AndroidDns.kt
yschimke: Yep, good catch.

If it's enabled, but you use lookup, we just filter it out.

okhttp/src/androidMain/kotlin/okhttp3/internal/platform/android/Android17SocketAdapter.kt
yschimke: nope, I was mistaken. That was on parsing the ECH block.


      try {
        record.echConfigList?.toBytes()?.toByteString()
      } catch (e: IllegalArgumentException) {
        // The platform can throw on a malformed or absent ECH parameter.
        // https://issuetracker.google.com/issues/319957694
        null
      }

okhttp/src/androidMain/kotlin/okhttp3/android/AndroidDns.kt
yschimke: I'll ask tomorrow if that mode still exists.

okhttp/src/androidMain/kotlin/okhttp3/internal/platform/android/Android17SocketAdapter.kt
yschimke: Throw for now until we have a policy

yschimke: No wait, this was the one I meant.

https://issuetracker.google.com/issues/319957694

okhttp/src/androidMain/kotlin/okhttp3/android/AndroidDns.kt
yschimke: IT's out of scope by then, unless we make the whole thing a val?

okhttp/src/androidMain/kotlin/okhttp3/android/AndroidDns.kt
yschimke: SecurityException

yschimke:
Still tests to go

okhttp/src/androidMain/kotlin/okhttp3/android/AndroidDns.kt
yschimke: Yep, I'll update the wording. But also think if we want this we should make AndroidDns the default on API 37+ and when network security policy isn't globally disabled?

I think effectively the modes are (potentially per host)

UNSPECIFIED
DISABLED
ENABLED

So maybe that default once we are confident is the right option?

yschimke:
Need to put this catch back

```
emulator-5554 - 17 Tests 34/29 completed. (5 skipped) (1 failed)
Tests on emulator-5554 - 17 failed: There was 1 failure(s).
Finished 34 tests on emulator-5554 - 17
Execute okhttp.android.test.OkHttpTest.testUnderscoreRequest: FAILED
The input does not conform to the STD 3 ASCII rules. line: java.lang.IllegalArgumentException: Invalid input to toASCII: example_underscore_123.s3.amazonaws.com
at java.net.IDN.toASCII(IDN.java:115)
at javax.net.ssl.SNIHostName.<init>(SNIHostName.java:99)
at com.android.org.conscrypt.Platform.getSSLParameters(Platform.java:269)
at com.android.org.conscrypt.ConscryptEngine.getSSLParameters(ConscryptEngine.java:425)
at com.android.org.conscrypt.ConscryptEngineSocket.getSSLParameters(ConscryptEngineSocket.java:193)
at okhttp3.internal.platform.android.Android17SocketAdapter.configureTlsExtensions(Android17SocketAdapter.kt:63)
at okhttp3.internal.platform.Android10Platform.configureTlsExtensions(Android10Platform.kt:86)
at okhttp3.internal.connection.ConnectPlan.connectTls(ConnectPlan.kt:349)
```

okhttp/src/androidMain/kotlin/okhttp3/android/AndroidDns.kt
yschimke: BTW if we ignore or override the NSP, Conscrypt still checks it.

https://cs.android.com/android/platform/superproject/+/android-latest-release:external/conscrypt/common/src/main/java/org/conscrypt/NativeSsl.java;l=569?q=enableEchBasedOnPolicy

swankjesse:
Where does Android implement the policy in that XML file? Presumably we act as if it's always enabled, and the XML config overrides our preferences?

okhttp/src/androidMain/kotlin/okhttp3/android/AndroidDns.kt
swankjesse: Ahh

yschimke:
> Where does Android implement the policy in that XML file? Presumably we act as if it's always enabled, and the XML config overrides our preferences?

But then we do extra work fetching ECH records that won't be used. 

But I think the test is that we do the correct, secure, optimal thing with DnsOverHttps enabled? So it can't be inside AndroidDns then?

Not sure where exactly - one of these

https://cs.android.com/android/platform/superproject/+/android-latest-release:external/conscrypt/repackaged/platform/src/main/java/com/android/org/conscrypt/ConscryptNetworkSecurityPolicy.java

https://cs.android.com/android/platform/superproject/+/android-latest-release:external/conscrypt/nsc/src/android/security/NetworkSecurityPolicy.java

cc @swankjesse not sure if you see non threaded comments.

okhttp/src/androidMain/kotlin/okhttp3/android/AndroidDns.kt
```diff
@@ -0,0 +1,211 @@
...
+     * Resolves addresses only, for callers using the legacy blocking API. [Dns] cannot carry
+     * HTTPS/ECH metadata, so this skips that query rather than paying for it and discarding it.
+     */
+    override fun lookup(hostname: String): List<InetAddress> =
+      AndroidDnsCall(Dns.Request(hostname), includeServiceMetadata = false)
+        .execute()
+        .filterIsInstance<Dns.Record.IpAddress>()
+        .map { it.address }
+
+    override fun newCall(request: Dns.Request): Dns.Call = AndroidDnsCall(request, includeServiceMetadata = includeServiceMetadata)
```
yschimke: @swankjesse should includeServiceMetadata move to Dns.Request? so we can honour NetworkSecurityPolicy?

swankjesse: I don't think so. I consider that to be a property of the service and not the data.

okhttp/src/androidMain/kotlin/okhttp3/android/AndroidDns.kt
```diff
@@ -0,0 +1,233 @@
...
+import okio.Buffer
+
+/**
+ * A [Dns] backed by Android's system resolver, with ECH support from Android's [DnsResolver].
+ *
+ * IP addresses come from the system resolver — [InetAddress.getAllByName], or
+ * [Network.getAllByName] when a [network] is set. Internally this drives OkHttp's
+ * [StateMachineDnsCall] through a private transport: a single `A` query stands in for the blocking
+ * address lookup (which returns both address families), plus an optional `HTTPS` query for service
+ * metadata such as ECH.
```
swankjesse, line 51: Great docs

okhttp/src/androidMain/kotlin/okhttp3/android/AndroidDns.kt
```diff
@@ -0,0 +1,233 @@
...
+       * Asks [DnsResolver] for the `HTTPS` record. The platform has no typed API for this below
+       * API 36, so we request the raw message and decode it with OkHttp's own [DnsMessageReader] —
+       * the same decoder `DnsOverHttps` uses.
+       *
+       * Failures are reported to the [StateMachineDnsCall] rather than swallowed, so callers can
+       * tell an absent `HTTPS` record from a query that errored.
+       *
+       * `WrongConstant` is suppressed because okhttp's [TYPE_HTTPS] is the DNS wire value (65), the
+       * same as the platform's `DnsResolver.TYPE_HTTPS`. We use ours so the query stays valid on
+       * API 29+; `DnsResolver.TYPE_HTTPS` was only added in API 37.
```
swankjesse, line 162: oooh neat

PR #9579, opened 2026-07-22
 New CachingTransport

swankjesse:


okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/dns/CachingTransport.kt
```diff
@@ -0,0 +1,263 @@
...
+ * If this receives multiple equivalent queries, it combines them into a single query on the
+ * underlying transport.
+ */
+// TODO: evict old entries from cache using State.lastRequestedAt
+@OkHttpInternalApi
+@OptIn(ExperimentalTime::class) // We know Clock and Instant will be stable in Kotlin 2.3.
+class CachingTransport<Q>(
+  private val taskRunner: TaskRunner,
+  private val delegate: Transport<Q>,
+  private val clock: Clock,
```
JakeWharton: Do we care that this can go backwards or jump forwards? Should it be `TimeSource` instead and everything be relative to time marks?

JakeWharton (approved):

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/dns/CachingTransport.kt
swankjesse: Yeah, Good call.

PR #9581, opened 2026-07-23
More tests for the new DNS cache

swankjesse:


okhttp/src/commonTest/kotlin/okhttp3/internal/dns/DnsCallStateMachineTest.kt
```diff
@@ -231,6 +588,87 @@ class DnsCallStateMachineTest {
...
+    testDnsCallStateMachine {
+      val call0 =
+        newCall(
+          request = Dns.Request(hostname = "lysine.dev"),
+          caching = true,
+          includeIPv6 = false,
+          includeServiceMetadata = false,
+        )
+      call0.enqueue()
+      transport
```
yschimke, line 602: nice MockDnsServer api.

yschimke (approved):

PR #9583, opened 2026-07-23
Start evicting elements from the DNS memory cache

swankjesse:


okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/dns/MemoryCache.kt
```diff
@@ -0,0 +1,125 @@
...
+package okhttp3.internal.dns
+
+import java.util.PriorityQueue
+import java.util.concurrent.ConcurrentHashMap
+import kotlin.time.ComparableTimeMark as Time
+import kotlin.time.TimeSource
+
+/**
+ * This cache is configured with a target [maxSize], but it will not evict entries until it
+ * has twice as many entries.
```
swankjesse, line 25: Also considered evicting whenever we were 25% greater than the target. Both 1.25x and 2x have the same big-O runtime. 1.25x is better on memory, 2.x is better on CPU.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/dns/MemoryCache.kt
```diff
@@ -0,0 +1,125 @@
...
+ *
+ * This retains entries in least-recently-used order.
+ *
+ * This evicts in big batches because each eviction must traverse the entire cache.
+ */
+abstract class MemoryCache<K : Any, V : Any>(
```
yschimke, line 31: Can we have a way to mark global changes that would invalidate the whole cache?

Let's say the default network changes, we should not used cached values.

I think for network selection we can have different caches. So it's mainly for global changes

yschimke (approved):

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/dns/MemoryCache.kt
swankjesse: Yeah, can do. evictAll() !

PR #9584, opened 2026-07-24
Rearrange new DNS classes without changing behavior

swankjesse:
Transport is now DnsQuery.Factory
CachingTransport is now DnsCache
StateMachineDnsCall.Query is now DnsQuery

This is intended to make it possible to have end-users get a DnsCache instance without being exposed to the implementation details of our transports and things.


okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/dns/DnsQuery.kt
```diff
@@ -0,0 +1,43 @@
...
+import java.io.IOException
+import okhttp3.internal.OkHttpInternalApi
+
+/**
+ * A transport-layer DNS query for a single record type.
+ *
+ * This is different from `Dns.Call` which returns multiple record types.
+ */
+@OkHttpInternalApi
+interface DnsQuery {
```
yschimke, line 29: Nice

yschimke (approved):

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/dns/DnsCache.kt
```diff
@@ -66,24 +66,23 @@ import okhttp3.internal.dns.StateMachineDnsCall.Transport
  */
 @OkHttpInternalApi
 @OptIn(ExperimentalTime::class) // We know Clock and Instant will be stable in Kotlin 2.3.
-class CachingTransport<Q>(
+class DnsCache(
```
yschimke, line 69: If we want android apps using this with system sources, we will need to hook into network changes.

For DNS over Https it seems fine.

yschimke (approved):

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/dns/DnsCache.kt
swankjesse: Yeah, I think we wanna scope the cache to the network.

PR #9594, opened 2026-07-25
Enable in-memory caching in DnsOverHttps and AndroidDns

swankjesse:
The default cache size is 1,000, which is effectively unbounded for mobile apps, which likely call a few dozen API servers, ad networks, and analytics systems.

I don't expect DnsOverHttps to be deployed in other environments. If so, it's possible this cache will be problematic, as even 10-second stale results can be problematic for server environments. I'll put an appropriate warning about this in our release notes.

### DIFF
--- a/okhttp/src/androidMain/kotlin/okhttp3/android/AndroidDns.kt
+++ b/okhttp/src/androidMain/kotlin/okhttp3/android/AndroidDns.kt
@@ -27,6 +27,7 @@ import java.net.InetAddress
 import java.net.UnknownHostException
 import java.util.concurrent.Executor
 import okhttp3.Dns
+import okhttp3.DnsCache
 import okhttp3.internal.OkHttpInternalApi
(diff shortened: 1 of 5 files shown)


okhttp/src/androidMain/kotlin/okhttp3/android/AndroidDns.kt
```diff
@@ -66,7 +68,7 @@ class AndroidDns(
 
   /** Drives [StateMachineDnsCall] using the system resolver and [DnsResolver]. */
   private val queryFactory =
-    DnsQuery.Factory { question ->
+    dnsCache.`-delegate`.wrap { question ->
```
yschimke, line 71: I think if we do this, we are also responsible for evicting on network changes. So it's if it's the default, we need to ship that wiring also.

This is fine until some oblivious well intentioned grunt makes it the default in an unrelated PR.

yschimke (approved):

PR #9604, opened 2026-07-26
Add proper EchTests

yschimke:
Run ECH tests with AndroidDns and DnsOverHttps.

note: doesn't run in CI because Remote tag.

### DIFF
--- a/android-test/src/androidTest/java/okhttp/android/test/EchTest.kt
+++ b/android-test/src/androidTest/java/okhttp/android/test/EchTest.kt
@@ -18,15 +18,18 @@ package okhttp.android.test
 import android.annotation.SuppressLint
 import android.net.ssl.EchConfigMismatchException
 import android.os.Build
+import app.cash.burst.Burst
 import assertk.assertThat
 import assertk.assertions.contains
 import assertk.assertions.doesNotContain
 import assertk.assertions.isEqualTo
 import assertk.assertions.isFalse
 import assertk.assertions.isTrue
+import okhttp3.HttpUrl.Companion.toHttpUrl
 import okhttp3.OkHttpClient
 import okhttp3.Request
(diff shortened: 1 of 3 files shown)


yschimke:
Testing strategy borrowed from https://github.com/curl/curl/blob/master/docs/ECH.md

android-test/src/androidTest/java/okhttp/android/test/EchTest.kt
```diff
@@ -15,52 +15,152 @@
...
-  private fun <T> OkHttpClient.sendRequest(
-    request: Request,
-    fn: (Response) -> T,
-  ): T {
-    val response = newCall(request).execute()
+  /**
+   * Disabled by policy.
+   */
+  @Test
+  fun policyDisabledHostDoesNotUseEch() {
```
swankjesse, line 130: NEAT

swankjesse (approved):

android-test/src/androidTest/java/okhttp/android/test/EchTest.kt
```diff
@@ -15,52 +15,152 @@
...
+  }
 
-    val tlsEchBody =
-      client.sendRequest(Request.Builder().url("https://tls-ech.dev/").build()) {
-        it.body.string()
+  @Test
+  fun wrongPublicNameIsNotRetried() {
+    val rejection = client.echRejectionFrom("https://wrong.tls-ech.dev/")
+
+    // TODO retry with these, then assert "You are using ECH" like tlsEchDevUsesEch.
```
yschimke, line 92: @swankjesse thoughts on handling this?

swankjesse: Discussed offline. I want to be extremely judicious at implementing retries, as it exposes us to attacks.

PR #9607, opened 2026-07-26
Add a test with network pinning

yschimke:
Applies the proposed approach from #8376 

And uses it test only to confirm we can use AndroidDns with a network on API 37.

These should probably move out of tests.


yschimke:
Wait for https://github.com/lysine-dev/okhttp/pull/9608 to land

android-test/src/androidTest/java/okhttp/android/test/AndroidNetworkPinning.kt
```diff
@@ -0,0 +1,76 @@
...
+
+    return effectiveChain.proceed(request)
+  }
+
+  /**
+   * ECH needs the `HTTPS` record, which is only reachable through `DnsResolver.rawQuery()` and only
+   * consulted by the platform from API 37. Below that there's nothing to gain from the extra query,
+   * so [AndroidNetworkDns] does the plain address lookup.
+   */
+  private fun dnsForNetwork(network: Network): Dns =
```
swankjesse, line 59: Off topic, but I think we structure `AndroidNetworkDns` very carefully, stuff like this might not be necessary

swankjesse (approved):

android-test/src/androidTest/java/okhttp/android/test/AndroidNetworkPinning.kt
yschimke: Yep, happy to get some guidance on how to structure it, and I'll hold us to correct on Android.

PR #9611, opened 2026-07-26
Handle ECH Retry

yschimke:
A few assorted changes
- grab the ech retry config and use it to retry when we get a mismatch
- avoid ech retry more than once
- avoid downgrading from ech to non ech


yschimke:
@swankjesse I don't think Conscrypt ECH PR supported this

https://github.com/google/conscrypt/pull/1406/changes#diff-808df7e6867f40b6d68b137bb7e1be92b26fb0b99f20df4ed016e7f4015a42e3

android-test/src/androidTest/java/okhttp/android/test/EchTest.kt
```diff
@@ -77,21 +75,34 @@ class EchTest(
...
+    assertThat(body).doesNotContain("not using ECH")
   }
 
   @Test
-  fun wrongPublicNameIsNotRetried() {
-    val rejection = client.echRejectionFrom("https://wrong.tls-ech.dev/")
+  fun differentPublicHostnameIsVerifiedBeforeRetry() {
+    // The outer certificate authenticates public.tls-ech.dev,
+    // so the retry config may be used if it matches.
+    // https://www.rfc-editor.org/rfc/rfc9849.html#section-6.1.6
```
yschimke, line 94: Think I got think backwards, re-reading.

yschimke:
Quick search for evidence

https://chromium.googlesource.com/chromium/src/+/refs/tags/143.0.7499.47/net/url_request/http_with_dns_over_https_unittest.cc#449

https://github.com/mozilla/gecko-dev/blob/master/security/nss/gtests/ssl_gtest/tls_ech_unittest.cc#L2137-L2160

https://go.dev/src/crypto/tls/handshake_client_test.go#L2808

https://github.com/google/boringssl/blob/22a0079b189c391b95689813a41982ce11876f0a/ssl/test/runner/ech_tests.go#L2192-L2265

yschimke:
I'll ask at work tomorrow.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/dns/-DnsMessage.kt
```diff
@@ -68,6 +68,11 @@ data class DnsMessage(
   }
 }
 
+internal data class EchRetryConfig(
```
swankjesse: Move this to its own file

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/connection/ConnectPlan.kt
```diff
@@ -239,9 +260,37 @@ class ConnectPlan internal constructor(
...
+                  socketAddress = route.socketAddress,
+                  echConfigList = offeredEchRetryConfig.configList,
+                ),
+              echRetryConfig = offeredEchRetryConfig,
+            )
+          }
+
+          echRetryConfig != null && offeredEchRetryConfig != null -> {
+            // TODO: Should we treat an untrusted or missing ECH retry config as an ordinary
+            // SSLException and try another connection spec?
```
swankjesse: I believe that if we have ECH information, we shouldn’t let ourselves be downgraded.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/connection/ConnectPlan.kt
```diff
@@ -204,7 +211,21 @@ class ConnectPlan internal constructor(
         retryTlsConnection = tlsEquipPlan.nextConnectionSpec(connectionSpecs, sslSocket)
```
swankjesse: I think we make `nextConnectionSpec()` accept an `EchRetryConfig`, rather than having two competing mechanisms for deciding how to retry

swankjesse: Even better, maybe `nextConnectionSpec()` should accept an `SSLException`

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/connection/ConnectPlan.kt
```diff
@@ -204,7 +211,21 @@ class ConnectPlan internal constructor(
...
 
         connectionSpec.apply(sslSocket, isFallback = tlsEquipPlan.isTlsFallback)
-        connectTls(sslSocket, connectionSpec)
+        try {
+          connectTls(sslSocket, connectionSpec)
+        } catch (e: SSLException) {
+          val echRetryConfig = Platform.get().getEchRetryConfig(e)
+          if (
+            echRetryConfig != null &&
+            route.address.hostnameVerifier!!.verify(
```
swankjesse: This surprises me. Is there an attack that this defends against?

swankjesse (commented):
I’d like to be particularly careful with landing this as it introduces new attack vectors and is not yet possible to test with MockWebServer + JVM Conscrypt.

I think it’s definitely good to have special recovery code for ECH handshake failures. But my first instinct is that special recovery code should mostly be about reducing the opportunity for downgrade attacks.

This is attempting to do something else, and I would like to understand the problem it solves before we land this.

android-test/src/androidTest/java/okhttp/android/test/EchTest.kt
bjiang7: Yes, you should only be verifying the public hostname against the certificate, not against the request. Conscrypt isn't able to do the hostname validation itself, which is why it's necessary to do it here in OkHttp.

FYI @tweksteen

bjiang7:
Just chiming in here to state that the ECH recovery flow is a pretty important part of the spec, and so should be implemented if OkHttp is to say it supports ECH.

The reason why is because there's no other way to recover from a mismatch in ECH config between the client and server (i.e. synchronization issues). It's a performance hit to have to fall into this codepath (and shouldn't be happening in most cases), but it is needed so the consequences of things being out of sync is latency rather than outages.

I also asked BoringSSL a long time ago about this when doing the Android implementation, because I had the same question of how important the recovery flow was. Their stance was "This behavior is pretty crucial to allow servers to safely deploy ECH because DNS records can be stale and we need to make it possible for servers to recover from mismatches. The existence of any client that gets this wrong means a server now cannot deploy ECH in a rollback-safe way."

swankjesse:
Making it safer for server operators to confidently deploy ECH is good.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/connection/ConnectPlan.kt
yschimke: cc @bjiang7

android-test/src/androidTest/java/okhttp/android/test/EchTest.kt
yschimke: Just to confirm, at this point our SSLSession is dead, but we can safely read the certificates that were presented.

But are the certificates on the SSLSession trusted? Does Conscrypt already clean the certs?

okhttp/src/jvmTest/kotlin/okhttp3/InterceptorOverridesTest.kt
```diff
@@ -242,23 +241,15 @@ class InterceptorOverridesTest {
...
       OverrideParam.RetryOnConnectionFailure -> {
         enableTls()
-        var first = true
+
+        server.enqueue(MockResponse.Builder().failHandshake().build())
         client =
           client
             .newBuilder()
             .connectionSpecs(listOf(ConnectionSpec.RESTRICTED_TLS, ConnectionSpec.MODERN_TLS))
-            .eventListener(
```
yschimke, line 252: this wasn't valid, and after the refactor, isn't effective.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/connection/ConnectPlan.kt
bjiang7: Yes, it's necessary for the client to authenticate the public name (i.e. validate the certificate) before using any retry configs supplied by the server, as that's the way to "prove" the retry configs are genuine and actually owned by the server.

https://www.rfc-editor.org/rfc/rfc9849.html#name-authenticating-for-the-publ

okhttp/src/jvmTest/kotlin/okhttp3/internal/connection/RetryConnectionTest.kt
```diff
@@ -69,6 +90,98 @@ class RetryConnectionTest {
...
+      factory.newHttpsAddress(
+        hostnameVerifier = { hostname, _ ->
+          verifiedHostnames += hostname
+          true
+        },
+      )
+    val routePlanner = factory.newRoutePlanner(client, address)
+    val route = factory.newRoute(address)
+    val connectionSpecs = listOf(ConnectionSpec.MODERN_TLS, ConnectionSpec.COMPATIBLE_TLS)
+    val socket = createSocketWithEnabledProtocols(TlsVersion.TLS_1_2, TlsVersion.TLS_1_1)
```
swankjesse: This one’s curious 'cause it’s missing TLS 1.3

okhttp/src/jvmTest/kotlin/okhttp3/internal/connection/RetryConnectionTest.kt
```diff
@@ -69,6 +90,98 @@ class RetryConnectionTest {
...
+    val address = factory.newHttpsAddress(hostnameVerifier = { _, _ -> true })
+    val routePlanner = factory.newRoutePlanner(client, address)
+    val connectionSpecs = listOf(ConnectionSpec.MODERN_TLS)
+    val socket = createSocketWithEnabledProtocols(TlsVersion.TLS_1_2)
+    val attempt0 =
+      routePlanner
+        .planConnect()
+        .planWithCurrentOrInitialConnectionSpec(connectionSpecs, socket)
+
+    // A new DNS result must not influence a retry of the previous ECH configuration.
```
swankjesse, line 165: Great test

okhttp/src/jvmTest/kotlin/okhttp3/internal/connection/RetryConnectionTest.kt
```diff
@@ -69,6 +90,98 @@ class RetryConnectionTest {
     assertThat(retryTlsHandshake(retryableException)).isTrue()
   }
 
+  @Test fun echRetryConfigIsUsedOnceWithoutTlsFallback() {
```
swankjesse, line 106: Maybe another test, echRetryConfigIsNotRecursive() ?

okhttp/src/jvmTest/kotlin/okhttp3/internal/connection/RetryConnectionTest.kt
```diff
@@ -69,6 +90,98 @@ class RetryConnectionTest {
...
+    val attempt1 = attempt0.nextConnectionSpec(connectionSpecs, socket, echRetryException)
+
+    assertThat(attempt1).isNotNull()
+    assertThat(attempt1!!.route.socketAddress.address).isEqualTo(originalAddresses[0])
+    assertThat(attempt1.route.socketAddress.address).isNotEqualTo(newAddress)
+    dns.assertRequests(hostname)
+    socket.close()
+  }
+
+  @Test fun untrustedEchRetryConfigIsNotRetried() {
```
swankjesse, line 177: I believe the spec wants us to retry if the public name verifies

okhttp/src/jvmTest/kotlin/okhttp3/internal/connection/RouteSelectorTest.kt
```diff
@@ -133,6 +134,43 @@ class RouteSelectorTest(
     dns.assertRequests(uriHost)
   }
 
+  /** https://www.rfc-editor.org/rfc/rfc9848.html#section-5.1 */
+  @Test fun echAddressesDoNotFallBackToNonEch() {
```
swankjesse, line 138: I believe they do fall back to non-ech, if the public name from DNS verifies

swankjesse (changes requested):
This is excellent. Some feedback and I’d like another look before merge. I promise to not make you wait so long on followups

okhttp/src/jvmTest/kotlin/okhttp3/internal/connection/RouteSelectorTest.kt
yschimke: I don't think Conscrypt tells us that? Unless we check the session that just failed a handshake?

swankjesse: Rather, we fall back to no ECH if there's no retry config and our hostname verification succeeds.

android-test/src/androidTest/java/okhttp/android/test/EchTest.kt
```diff
@@ -77,21 +75,33 @@ class EchTest(
   }
 
   @Test
-  fun staleEchConfigIsNotRetried() {
-    val rejection = client.echRejectionFrom("https://stale.tls-ech.dev/")
+  fun staleEchConfigIsRetried() {
+    val body = client.get("https://stale.tls-ech.dev/")
```
bjiang7: Just FYI stale.tls-ech.dev uses port 444 and wrong.tls-ech.dev uses port 445. Not sure if you're using the default port of 443, but if so that should probably be changed to make sure that you're not accidentally getting the default domain of plain tls-ech.dev.

Another option is adding an assertion that you're hitting the correct domain since it also appears on the page (stale.tls-ech.dev vs tls-ech.dev).

android-test/src/androidTest/java/okhttp/android/test/EchTest.kt
bjiang7: >But are the certificates on the SSLSession trusted?

The certificates on the SSLSession should be trusted since we did handshake with ClientHelloOuter, so TrustManager would've already authenticated the certificates. BoringSSL also only emits the SSL_R_ECH_REJECTED error at the very end of the client handshake, so it would've had to go through all the certificate validation first as otherwise it would've instead sent a SSL_R_CERTIFICATE_VERIFY_FAILED error instead of SSL_R_ECH_REJECTED.

>Does Conscrypt already clean the certs?

Conscrypt doesn't clean the certificates, it just decodes things into an X509Certificate array and passes it around. It doesn't clean any of the certs up in the SSL session.

okhttp/src/androidMain/kotlin/okhttp3/internal/platform/Android10Platform.kt
```diff
@@ -86,6 +90,23 @@ class Android10Platform :
...
+  @SuppressLint("NewApi")
+  internal override fun getEchRetryConfig(exception: SSLException): EchRetryConfig? {
+    if (Build.VERSION.SDK_INT < 37 || exception !is EchConfigMismatchException) return null
+
+    // From https://cs.android.com/android/platform/superproject/+/android-latest-release:external/conscrypt/platform/src/main/java/org/conscrypt/Platform.java;bpv=0
+    // we can get neither, publicHostname only, or both
+    return EchRetryConfig(
+      publicHostname = exception.publicHostname ?: return null,
+      // An absent retry config list is how a server securely disables ECH.
+      // TODO can Conscrypt hand us an empty list, and does it mean the same thing?
```
bjiang7: No, Conscrypt would either hand you null or an actual EchConfigList obj. We do some minimal validation on the Conscrypt side to check that the list isn't empty or null and that its length matches up, before creating any EchConfigList object. 

The rest of the validation is done by BoringSSL when we pass it the EchConfigList (that's where all the "exciting" validation logic comes in, e.g. version checking, etc.). That was a conscious design decision because we didn't want to duplicate this logic and have it become out of sync with BoringSSL.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/dns/EchRetryConfig.kt
```diff
@@ -0,0 +1,41 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okhttp3.internal.dns
+
+import okio.ByteString
+
+/**
+ * ECH Retry config. Generally sent by a server when there is a
```
bjiang7: The retry config is normally sent by a server when the ECH configurations have become out of sync (e.g. the TTL on them expired, they rotated to a new ECH config), not from mismatched A/AAAA and HTTPS records, so I would update this KDoc.

For example, Cloudflare has 1 given ECH configuration at a time and rotates it every hour. They have a grace period of 4 hours after they've rotated it, so in theory if you got this old config (e.g. if you cached it past its TTL or something) and tried to use it after the grace period expired, this is where the server should be sending you a retry config.

https://datatracker.ietf.org/meeting/126/materials/slides-126-iepg-sessa-encrypted-client-hello-ech-an-active-measurement-perspective-00

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/dns/EchRetryConfig.kt
```diff
@@ -0,0 +1,41 @@
...
+ * mismatch between A/AAAA and HTTPS Records.
+ *
+ * If a new [configList] is present, the server securely replaced our ECH configuration, and it
+ * must only be used when [publicHostname] can be validated against the certificate
```
bjiang7: Optional: you can reference RFC 9849 6.1.7 for the necessity of authenticating the public name.

okhttp/src/jvmTest/kotlin/okhttp3/internal/connection/RetryConnectionTest.kt
yschimke: TODO

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/dns/EchRetryConfig.kt
yschimke: done

okhttp/src/androidMain/kotlin/okhttp3/internal/platform/Android10Platform.kt
yschimke: done

android-test/src/androidTest/java/okhttp/android/test/EchTest.kt
yschimke: done

okhttp-testing-support/src/main/kotlin/okhttp3/FakeDns.kt
```diff
@@ -275,7 +275,7 @@ class FakeDns(
           }
 
           is ResourceRecord.IpAddress -> {
-            val ipAddressRecord = Dns.Record.IpAddress(request.hostname, resourceRecord.address)
+            val ipAddressRecord = Dns.Record.IpAddress(resourceRecord.name, resourceRecord.address)
```
swankjesse, line 278: ooooh tricky

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/connection/ConnectPlan.kt
```diff
@@ -200,11 +206,13 @@ class ConnectPlan internal constructor(
         val tlsEquipPlan = planWithCurrentOrInitialConnectionSpec(connectionSpecs, sslSocket)
         val connectionSpec = connectionSpecs[tlsEquipPlan.connectionSpecIndex]
 
-        // Figure out the next connection spec in case we need a retry.
-        retryTlsConnection = tlsEquipPlan.nextConnectionSpec(connectionSpecs, sslSocket)
```
swankjesse, line 204: I’m walking through this, and trying to cover all our bases.

This sets up a mechanism to attempt the next connection spec in sequence, and it’s replaced with code to attempt the next ECH config.

If the ECH config fails, do we attempt the next connection spec? I suppose we don’t per the ECH doc, which is also weird!

> If the server rejects ECH, the client proceeds with the handshake, authenticating for ECHConfig.contents.public_name as described in [Section 6.1.7](https://www.rfc-editor.org/rfc/rfc9849.html#auth-public-name). If authentication or the handshake fails, the client MUST return a failure to the calling application.

(We shouldn’t fall back using our normal fallback mechanism, and that’s difficult to test)

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/connection/ConnectPlan.kt
```diff
@@ -239,10 +247,6 @@ class ConnectPlan internal constructor(
       call.eventListener.connectFailed(call, route.socketAddress, route.proxy, null, e)
       connectionPool.connectionListener.connectFailed(route, call, e)
 
-      if (!retryOnConnectionFailure || !retryTlsHandshake(e)) {
```
swankjesse, line 242: Note to self... this behavior isn’t removed, we just compute a null `retryTlsConnection` elsewhere

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/connection/ConnectPlan.kt
```diff
@@ -489,12 +493,66 @@ class ConnectPlan internal constructor(
...
    */
   internal fun nextConnectionSpec(
     connectionSpecs: List<ConnectionSpec>,
     sslSocket: SSLSocket,
+    sslException: SSLException,
+  ): ConnectPlan? {
+    if (!retryOnConnectionFailure) return null
+
+    val offeredEchRetryConfig = Platform.get().getEchRetryConfig(sslException)
+    if (offeredEchRetryConfig != null) {
```
swankjesse, line 507: This condition surprises me. The logic looks incorrect for the case where the server securely disables ECH. In particular I would expect the `offeredEchRetryConfig` to be null in the case where the server rejects ECH (distinct from `offeredEchRetryConfig.configList == null`)

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/dns/EchRetryConfig.kt
```diff
@@ -0,0 +1,42 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okhttp3.internal.dns
+
+import okio.ByteString
+
+/**
+ * ECH retry config. Sent by a server when the ECH configuration we offered has fallen out of sync
```
swankjesse, line 21: This doc says ‘Sent by a server’ but we need a mechanism to lookup the public name when the server replies _without this extension_. 

Does this class represent the `encrypted_client_hello` extension? Or is it a higher-level object that describes the TLS client’s state?

(I think it’s probably something describing the client state, because I don’t think the publicHostname is included in the `encrypted_client_hello` extension)

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/connection/ConnectPlan.kt
yschimke: I think we should clarify, one badly configured server/region is more likely than poorly attempted hacking, so I don't like reducing one of the "free" wins of distributed computing, retries.

But the change was really intended to just be to include the possible mismatch failure in the next choice, so it can't happen before connectTls.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/dns/EchRetryConfig.kt
yschimke: https://cs.android.com/android/platform/superproject/+/android-latest-release:external/boringssl/src/include/openssl/ssl.h;l=4352?q=SSL_get0_ech_name_override&ss=android%2Fplatform%2Fsuperproject:external%2F&start=11

```
// SSL_get0_ech_name_override, if |ssl| is a client and the server rejected ECH,
// sets |*out_name| and |*out_name_len| to point to a buffer containing the ECH
// public name. Otherwise, the buffer will be empty.
//
// When offering ECH as a client, this function should be called during the
// certificate verification callback (see |SSL_CTX_set_custom_verify|). If
// |*out_name_len| is non-zero, the caller should verify the certificate against
// the result, interpreted as a DNS name, rather than the true server name. In
// this case, the handshake will never succeed and is only used to authenticate
// retry configs. See also |SSL_get0_ech_retry_configs|.
OPENSSL_EXPORT void SSL_get0_ech_name_override(const SSL *ssl,
                                               const char **out_name,
                                               size_t *out_name_len);
```

```
  std::string_view ech_name_override = GetECHNameOverride();
  if (!ech_name_override.empty()) {
    // If ECH was offered but not negotiated, BoringSSL will ask to verify a
    // different name than the origin. If verification succeeds, we continue the
    // handshake, but BoringSSL will not report success from SSL_do_handshake().
    // If all else succeeds, BoringSSL will report |SSL_R_ECH_REJECTED|, mapped
    // to |ERR_R_ECH_NOT_NEGOTIATED|. |ech_name_override| is only used to
    // authenticate GetECHRetryConfigs().
    DCHECK(!ssl_config_.ech_config_list.empty());
    used_ech_name_override_ = true;
```

okhttp-testing-support/src/main/kotlin/okhttp3/FakeDns.kt
yschimke: it seems more correct?

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/connection/ConnectPlan.kt
yschimke: Does this cover the fallback concern for you?

```
    // If this was already in response to an ech retry, we are done for this
    // connection
    if (echRetryConfig != null || !retryTlsHandshake(sslException)) return null
```

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/connection/ConnectPlan.kt
yschimke: From the validation comments here https://github.com/lysine-dev/okhttp/pull/9611#discussion_r3690110107

Which one do you think is which?

okhttp-testing-support/src/main/kotlin/okhttp3/FakeDns.kt
swankjesse: Yes, the new code is a nice improvement.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/connection/ConnectPlan.kt
swankjesse: Yeah, I am mostly eager to write a bunch of test cases for this because the logic is too difficult for me to follow statically.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/connection/ConnectPlan.kt
swankjesse: I think it might be a problem with documentation and naming, and not a problem with logic.

If the server doesn’t include an ECH extension, there’s no ‘offered ECH retry config’. In particular there shouldn’t be anything _offered_ there, perhaps _derived_.

But we do need an object to extract the public name from the EchConfigList object we passed in.

swankjesse: (The class `EchRetryConfig` is documented as _sent by a server_ but I don’t think the `publicName` can be sent by a server; that needs to come from our passed-in EchConfigList)

okhttp/src/androidMain/kotlin/okhttp3/internal/platform/Android10Platform.kt
```diff
@@ -86,6 +90,24 @@ class Android10Platform :
...
+  internal override fun getEchRetryConfig(exception: SSLException): EchRetryConfig? {
+    if (Build.VERSION.SDK_INT < 37 || exception !is EchConfigMismatchException) return null
+
+    // From https://cs.android.com/android/platform/superproject/+/android-latest-release:external/conscrypt/platform/src/main/java/org/conscrypt/Platform.java;bpv=0
+    // we can get neither, publicHostname only, or both. Conscrypt only hands us an EchConfigList
+    // if it is non-empty and self-consistent; BoringSSL does the real validation (version checks
+    // and such) when we hand the list back to it.
+    return EchRetryConfig(
+      publicHostname = exception.publicHostname ?: return null,
+      // An absent retry config list is how a server securely disables ECH.
```
swankjesse, line 103: Nice

okhttp/src/jvmTest/kotlin/okhttp3/internal/connection/RetryConnectionTest.kt
```diff
@@ -39,12 +50,35 @@ class RetryConnectionTest {
...
+      configList = null,
+      publicHostname = "public.tls-ech.dev",
+    )
 
   @RegisterExtension
   val clientTestRule = OkHttpClientTestRule()
 
   @RegisterExtension
-  val platform = PlatformRule()
+  val platform =
```
swankjesse, line 70: This is rad

swankjesse (approved):

PR #9614, opened 2026-07-27
Support @StartStop fields in nested tests

### LINKED ISSUE #9546: MockWebServer @StartStop field on outer class not started for @Nested JUnit 5 tests
## Problem

When a test class uses `@Nested` inner classes to group tests, a `@StartStop MockWebServer` field declared on the outer class is never started. Any test inside a nested class that touches the server (e.g. `mockWebServer.url(...)`) fails with:

```
java.lang.IllegalStateException: call start() first
    at mockwebserver3.MockWebServer.getSocketAddress(MockWebServer.kt:150)
    at mockwebserver3.MockWebServer.getHostName(MockWebServer.kt:156)
    at mockwebserver3.MockWebServer.url(MockWebServer.kt:200)
```

## Minimal reproduction

```kotlin
class NestedStartStopTest {
    @StartStop
    private val mockWebServer = MockWebServer()

    @Nested
    inner class Group {
        @Test
        fun `server is running`() {
            // fails: IllegalStateException: call start() first
            mockWebServer.url("/")
        }
    }
}
```

Running the same assertion in a top-level `@Test` on the outer class works as expected.

Environment: `com.squareup.okhttp3:mockwebserver3-junit5:5.3.2`, JUnit Jupiter 5.x.

## Root cause

In [`StartStopExtension.beforeEach`](https://github.com/square/okhttp/blob/127cfe25168931aa600fff29c43f96d8d1dd480c/mockwebserver-junit5/src/main/kotlin/mockwebserver3/junit5/internal/StartStopExtension.kt#L50-L67):

```kotlin
override fun beforeEach(context: ExtensionContext) {
  val testInstance = context.testInstance.get()
  val instanceFields =
    findAnnotatedFields(
      context.requiredTestClass,
      StartStop::class.java,
    ) { !Modifier.isStatic(it.modifiers) }
  for (field in instanceFields) {
    ...
    val server = field.get(testInstance) as? MockWebServer ?: continue
    ...
    server.start()
  }
}
```

For a nested test:
- `context.testInstance.get()` returns the **inner-class** instance, not the outer one that holds the field.
- `context.requiredTestClass` is the **inner** `@Nested` class, so `findAnnotatedFields` never sees fields declared on enclosing classes.

Result: the outer `@StartStop` field is invisible to the extension, `server.start()` is never invoked, and any nested test that uses the server hits `IllegalStateException`.

## Suggested fix

Iterate `context.requiredTestInstances.allInstances` (which contains outer instances in order) and, for each, enumerate `@StartStop`-annotated fields on that instance's class:

```kotlin
override fun beforeEach(context: ExtensionContext) {
  val store = context.getStore(Namespace.create(StartStop::class.java))
  for (testInstance in context.requiredTestInstances.allInstances) {
    val fields = findAnnotatedFields(
      testInstance.javaClass,
      StartStop::class.java,
    ) { !Modifier.isStatic(it.modifiers) }
    for (field in fields) {
      field.setAccessible(true)
      val server = field.get(testInstance) as? MockWebServer ?: continue
      store.put(field, server)
      server.start()
    }
  }
}
```

This matches the pattern JUnit 5 recommends for extensions that need to work with `@Nested`.

## Current workarounds

- Manual `@BeforeEach`/`@AfterEach` on the outer class calling `mockWebServer.start()`/`.close()`. Outer-class `@BeforeEach` runs before each nested test.
- `@TestInstance(Lifecycle.PER_CLASS)` + `@BeforeAll`/`@AfterAll` — server shared across all nested tests.
- Duplicate the `@StartStop` field on every `@Nested` class.
- Drop `@Nested` and flatten the test class.

arpitagarwal1301:
Fixes #9546

## Summary

- start `@StartStop` servers declared on enclosing instances for JUnit 5 `@Nested` tests
- discover fields from each matching test instance and retain each server independently in the extension store
- cover startup, non-annotated fields, and after-test cleanup for a nested test

## Validation

- focused `StartStopTest` regression
- `./gradlew :mockwebserver3-junit5:check`
- `git diff --check`

### DIFF
--- a/mockwebserver-junit5/src/main/kotlin/mockwebserver3/junit5/internal/StartStopExtension.kt
+++ b/mockwebserver-junit5/src/main/kotlin/mockwebserver3/junit5/internal/StartStopExtension.kt
@@ -51,24 +51,24 @@ internal class StartStopExtension :
   }
 
   override fun beforeEach(context: ExtensionContext) {
-    // Requires API 24
-    val testInstance = context.testInstance.get()
     val store = context.getStore(Namespace.create(StartStop::class.java))
 
-    val instanceFields =
-      findAnnotatedFields(
-        context.requiredTestClass,
-        StartStop::class.java,
-      ) { !Modifier.isStatic(it.modifiers) }
+    for (testInstance in context.requiredTestInstances.allInstances) {
+      val instanceFields =
+        findAnnotatedFields(
+          testInstance.javaClass,
+          StartStop::class.java,
+        ) { !Modifier.isStatic(it.modifiers) }
 
-    for (field in instanceFields) {
-      field.setAccessible(true)
-      val server = field.get(testInstance) as? MockWebServer ?: continue
+      for (field in instanceFields) {
+        field.setAccessible(true)
+        val server = field.get(testInstance) as? MockWebServer ?: continue
 
-      // Put the instance in the store, so JUnit closes it for us in afterEach.
-      store.put(field, server)
+        // Put the instance in the store, so JUnit closes it for us in afterEach.
+        store.put(server, server)
 
-      server.start()
+        server.start()
+      }
     }
   }
 }
--- a/mockwebserver-junit5/src/test/java/mockwebserver3/junit5/StartStopTest.kt
+++ b/mockwebserver-junit5/src/test/java/mockwebserver3/junit5/StartStopTest.kt
@@ -23,6 +23,7 @@ import mockwebserver3.Dispatcher
 import mockwebserver3.MockResponse
 import mockwebserver3.MockWebServer
 import mockwebserver3.RecordedRequest
+import org.junit.jupiter.api.Nested
 import org.junit.jupiter.api.Test
 import org.junit.jupiter.api.extension.AfterAllCallback
 import org.junit.jupiter.api.extension.RegisterExtension
@@ -60,6 +61,18 @@ class StartStopTest {
     assertThat(serverD.started).isTrue()
     assertThat(serverE.started).isTrue()
     assertThat(serverF.started).isFalse()
+  }
+
+  @Nested
+  inner class NestedTest {
+    @Test
+    fun happyPath() {
+      testInstances += this@StartStopTest
+
+      assertThat(serverA.started).isTrue()
+      assertThat(serverB.started).isTrue()
+      assertThat(serverC.started).isFalse()
+    }
   }
 
   private companion object {


kdelay (commented):
I checked this out locally (bca7b58, JDK 21, macOS arm64) and it does what it says: `:mockwebserver3-junit5:test` is green on the branch, and restoring only the old `StartStopExtension.beforeEach` while keeping the new test fails 3 tests (`NestedTest.happyPath`, `NestedTest.executionError`, `StartStopTest.executionError`), so the new test really pins the behaviour.

One thing that may be worth calling out explicitly: the store-key change from `field` to `server` is load-bearing, not cosmetic, and nothing in the current test would catch it if it were reverted. It becomes observable as soon as the outer class and the `@Nested` class inherit the `@StartStop` field from a shared base:

```kotlin
abstract class SharedBase {
  @StartStop val server = MockWebServer().apply { dispatcher = ClosableDispatcher() }
}

class CollisionTest : SharedBase() {
  @Test fun outerTest() {}

  @Nested inner class Child : SharedBase() {
    @Test fun childTest() {}
  }
}
```

`findAnnotatedFields` returns the same `java.lang.reflect.Field` for both instances, so with `store.put(field, server)` the child's entry overwrites the outer's. I measured it: both servers start either way, but with the `field` key only the child's server is closed (outer dispatcher `closed=false` in an `AfterAllCallback`), while with `store.put(server, server)` both are closed. That is a leak the current diff fixes silently; a second test for it would keep it from regressing.

Other cases I ran against the branch, all behaving as expected:

- Two levels of nesting (`@Nested` inside `@Nested`): outer, middle and innermost servers are all started.
- Two `@StartStop` fields referencing the same `MockWebServer`: fine, since `start()` returns early for an already-bound address and the single store entry closes it once.
- The outer instance's servers are closed after a nested test, so iterating `allInstances` does not leak in the plain case.

Two notes that are not about this change:

- Unrelated pre-existing behaviour: `@TestInstance(PER_CLASS)` with two test methods fails with `IllegalStateException: close() already called`, because `beforeEach` restarts a server that the previous test's store already closed. It reproduces identically on `main`, so it is not introduced here.
- The red `loom` check is `DuplexTest.duplexWithRedirect` in `:okhttp:allTests`, a module this PR does not touch. `main`'s own `loom` runs are red too with a different set of tests each time (run 30692116591: `BasicLoomTest`/`BasicMockServerTest`/`BasicProxyTest`; run 30546012738: `EventListenerTest_Call.successfulSecureConnect`), which looks like existing flakiness on that leg rather than a regression from this branch.

JakeWharton:
Hi! Thanks for the PR, but we are imminently about to ban LLM descriptions and comments, as well as LLM-authored code. Your PR description is pretty clearly LLM authored and that makes me think the code is too (I didn't review the diff).

You are welcome to use whatever tools you want for investing problems, but we expect all code to be written by hand along with comments and descriptions. If you can meet these requirements, you're welcome to make a new PR with a fix.

Issue #6594, opened 2021-03-19
WebSocketWriter flagged by Veracode CWE-331


dsuresh-ap:
Our veracode scan flagged `WebSocketWriter.random` as a vulnerability due to its use of `Random`.

https://cwe.mitre.org/data/definitions/331.html

According to [Websocket Protocol](https://tools.ietf.org/html/rfc6455#section-10.3):

> For example, each masking could be drawn from a cryptographically strong random number generator.  If the same key is used or a decipherable pattern exists for how the next key is chosen, the attacker can send a message that, when masked, could appear to be an HTTP request (by taking the message the attacker wishes to see on the wire and masking it with the next masking key to be used, the masking key will effectively unmask the data when the client applies it).

According to this specification the library should be using `SecureRandom` to prevent brute force attacks.

Version: 4.0.9

JakeWharton:
This came up before in https://github.com/square/okhttp/issues/4974. The randomness is used to defeat proxies (specifically caching proxies) and isn't for cryptographic purposes.

We could probably switch if it will help silence this kind of automated tooling.

tcolligan:
@JakeWharton Even though https://tools.ietf.org/html/rfc6455#section-5.3 says the following?

`The masking key is a 32-bit value chosen at random by the client.
   When preparing a masked frame, the client MUST pick a fresh masking
   key from the set of allowed 32-bit values.  The masking key needs to
   be unpredictable; thus, the masking key MUST be derived from a strong
   source of entropy, and the masking key for a given frame MUST NOT
   make it simple for a server/proxy to predict the masking key for a
   subsequent frame.  The unpredictability of the masking key is
   essential to prevent authors of malicious applications from selecting
   the bytes that appear on the wire.  RFC 4086 [RFC4086] discusses what
   entails a suitable source of entropy for security-sensitive
   applications.`

JakeWharton:
I guess if you're using HTTP and not HTTPS across an untrusted network. It's a pretty annoying spec inclusion given the fact that TLS obviates it and HTTP is only used on trusted networks (such as local unix domain sockets).

swankjesse:
@tcolligan please reopen if you have an example where this would improve security in practice. Otherwise I think it's just an overly sensitive detector.

PR #9624, opened 2026-07-30
use SecureRandom for web socket mask keys

basavaraj-sm05:
OkHttp masks every outbound web socket frame with a four-byte key taken from the java.util.Random that newWebSocket creates for each connection, and the Sec-WebSocket-Key nonce comes from that same instance. RFC 6455 section 5.3 asks for the mask to come from a strong source of entropy and says the key for one frame must not make it simple for a server or proxy to predict the key for a later one, which is the whole point of masking: it keeps a party who can influence message content from choosing the bytes an intermediary sees on a ws:// stream. java.util.Random is a 48-bit linear congruential generator, so the first mask a peer observes hands it the top 32 bits of state and the remaining 16 fall to a 65,536 candidate search that the next frame confirms. I drove WebSocketWriter with a plain Random to check, and after two frames it predicts the rest of the mask stream exactly, enough to read a later frame plaintext straight off the wire, so the property the spec rules out is the one we currently have. What made me look was that MockWebServer already builds its RealWebSocket with SecureRandom even though a server never masks, while the client that does mask does not. SecureRandom is a java.util.Random subclass so nothing downstream changes, and the writer tests that want deterministic masks pass their own Random in.

### DIFF
--- a/okhttp/src/commonJvmAndroid/kotlin/okhttp3/OkHttpClient.kt
+++ b/okhttp/src/commonJvmAndroid/kotlin/okhttp3/OkHttpClient.kt
@@ -18,8 +18,8 @@ package okhttp3
 import java.net.Proxy
 import java.net.ProxySelector
 import java.net.Socket
+import java.security.SecureRandom
 import java.time.Duration
-import java.util.Random
 import java.util.concurrent.ExecutorService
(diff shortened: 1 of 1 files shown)


JakeWharton:
See #6594

PR #9628, opened 2026-08-04
Refactor OkHttp to use applyJavaModules()

### LINKED ISSUE #9103: Gradle refactor `applyJavaModules()`
Please refactor this so it’s the same `applyJavaModules()` call we have in all of our other modules? Having two copy-pastes of this configuration will confuse me for all time

_Originally posted by @swankjesse in https://github.com/square/okhttp/pull/8767#discussion_r2399970028_

LouisLou2:
## Summary

Refactor OkHttp's JPMS configuration to use the shared `applyJavaModules()` helper, consistent with the other modules.

Extend the helper to support Kotlin Multiplatform projects while preserving the existing JVM behavior and OkHttp's multi-release JAR configuration.

Closes #9103

## Testing

- `./gradlew :build-logic:spotlessCheck :okhttp:spotlessCheck :okhttp:jvmJar :module-tests:test -PokhttpModuleTests=true`
- Verified that the JVM JAR contains `META-INF/versions/9/module-info.class`
- Verified that the manifest contains `Multi-Release: true`

### DIFF
--- a/build-logic/src/main/kotlin/JavaModules.kt
+++ b/build-logic/src/main/kotlin/JavaModules.kt
@@ -15,13 +15,17 @@
  */
 
 import me.champeau.mrjar.MultiReleaseExtension
+import org.gradle.api.JavaVersion
 import org.gradle.api.Project
 import org.gradle.api.plugins.JavaPluginExtension
+import org.gradle.api.tasks.bundling.Jar
 import org.gradle.api.tasks.compile.JavaCompile
+import org.gradle.jvm.toolchain.JavaLanguageVersion
 import org.gradle.jvm.toolchain.JavaToolchainService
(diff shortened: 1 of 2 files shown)


JakeWharton:
Hi! Thanks for the PR, but we are imminently about to ban LLM descriptions and comments, as well as LLM-authored code. Your PR description is pretty clearly LLM authored and that makes me think the code is too (I didn't review the diff).

You are welcome to use whatever tools you want for investing problems, but we expect all code to be written by hand along with comments and descriptions. If you can meet these requirements, you're welcome to make a new PR with a fix.

LouisLou2:
Got it. Recreated the PR with a simplified, hand-written description and commit.

PR #9632, opened 2026-08-06
Update MockServer container image to 7.5.0

### LINKED ISSUE #9629: Container tests fail because MockServer client and server versions do not match
The container tests currently use different MockServer client and server versions.

#9622 updated `mockserver-client-java-no-dependencies` from 7.4.0 to 7.5.0, but
`BasicMockServerTest.MOCKSERVER_IMAGE` still uses the
`mockserver/mockserver:mockserver-7.4.0` image.

MockServer rejects requests when the client and server major or minor versions
do not match. As a result, all container tests fail while registering
expectations, before making an OkHttp request.

## Reproduction

From the current `main` branch:

```shell
./gradlew :container-tests:test -PcontainerTests=true
```

This produces failures such as:

```
org.mockserver.client.ClientException:
Client version "7.5.0" major and minor versions do not match
server version "7.4.0"
```

The Loom configuration reproduces the same problem:

```
./gradlew :container-tests:test \
  -PcontainerTests=true \
  -Pokhttp.platform=loom \
  -Ptest.java.version=24
```

All 12 container tests fail in the Loom configuration. The same failure can
also be seen in the Loom job for #9622:

[https://github.com/lysine-dev/okhttp/actions/runs/30490484586/job/90706867182](https://github.com/lysine-dev/okhttp/actions/runs/30490484586/job/90706867182)

## Expected behavior

The MockServer client and container image use compatible versions, and the
container tests pass.

## Proposed fix

Update the container image to match the client version:

```
-.withTag("mockserver-7.4.0")
+.withTag("mockserver-7.5.0")
```

I verified this change with both commands above. All 12 container tests pass
with the Loom configuration.

LouisLou2:
Updates the container image for MockServer from 7.4.0 to 7.5.0.

The client was bumped in #9622, but the server image stayed on 7.4.0.

Fixes #9629

Tested via:

./gradlew :container-tests:test -PcontainerTests=true

./gradlew :container-tests:test -PcontainerTests=true -Pokhttp.platform=loom -Ptest.java.version=24 (12/12 passed)

### DIFF
--- a/container-tests/build.gradle.kts
+++ b/container-tests/build.gradle.kts
@@ -21,6 +21,7 @@ tasks.withType<Test> {
   jvmArgs(
     "-Dokhttp.platform=$platform",
   )
+  systemProperty("mockserver.version", libs.versions.mockserver.client.get())
 
   if (platform == "loom") {
     jvmArgs(
--- a/container-tests/src/test/java/okhttp3/containers/BasicMockServerTest.kt
+++ b/container-tests/src/test/java/okhttp3/containers/BasicMockServerTest.kt
@@ -78,10 +78,15 @@ class BasicMockServerTest {
   }
 
   companion object {
+    private val mockServerVersion =
+      checkNotNull(System.getProperty("mockserver.version")) {
+        "mockserver.version is not set"
+      }
+
     val MOCKSERVER_IMAGE: DockerImageName =
       DockerImageName
         .parse("mockserver/mockserver")
-        .withTag("mockserver-7.4.0")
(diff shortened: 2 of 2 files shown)


JakeWharton:
We should inject this version from the version catalog or this is just going to happen over and over again.

LouisLou2:
> We should inject this version from the version catalog or this is just going to happen over and over again.

Updated—now using the version catalog as the single source of truth to drive the test task and container image tag.

JakeWharton (approved):

PR #9633, opened 2026-08-06
 Refactor OkHttp to use applyJavaModules() 

### LINKED ISSUE #9103: Gradle refactor `applyJavaModules()`

LouisLou2:
This extends applyJavaModules() to support KMP, preserving the existing multi-release JAR setup.

Closes #9103

Testing
Ran:
```
./gradlew :build-logic:spotlessCheck :okhttp:spotlessCheck :okhttp:jvmJar :module-tests:test -PokhttpModuleTests=true
```

### DIFF
--- a/build-logic/src/main/kotlin/JavaModules.kt
+++ b/build-logic/src/main/kotlin/JavaModules.kt
@@ -15,13 +15,17 @@
  */
 
 import me.champeau.mrjar.MultiReleaseExtension
+import org.gradle.api.JavaVersion
 import org.gradle.api.Project
 import org.gradle.api.plugins.JavaPluginExtension
+import org.gradle.api.tasks.bundling.Jar
 import org.gradle.api.tasks.compile.JavaCompile
+import org.gradle.jvm.toolchain.JavaLanguageVersion
 import org.gradle.jvm.toolchain.JavaToolchainService
 import org.gradle.kotlin.dsl.configure
 import org.gradle.kotlin.dsl.getByType
 import org.gradle.kotlin.dsl.named
+import org.gradle.kotlin.dsl.register
 import org.jetbrains.kotlin.gradle.tasks.KotlinJvmCompile
 
 fun Project.applyJavaModules(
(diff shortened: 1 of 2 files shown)


build-logic/src/main/kotlin/JavaModules.kt
```diff
@@ -38,29 +66,79 @@ fun Project.applyJavaModules(
...
+      configureModuleInfoCompilation(moduleName, compileKotlinTask)
+    }
+
+  tasks.named<Jar>("jvmJar").configure {
+    manifest {
+      attributes(mapOf("Multi-Release" to true))
+    }
+
+    from(compileJavaModuleInfo.map { it.destinationDirectory }) {
+      into("META-INF/versions/$javaModuleVersion/")
```
swankjesse:
```suggestion
      into("META-INF/versions/9/")
```

build-logic/src/main/kotlin/JavaModules.kt
```diff
@@ -38,29 +66,79 @@ fun Project.applyJavaModules(
...
+private fun Project.applyJavaModulesMultiplatform(
+  moduleName: String,
+  javaModuleVersion: Int,
+  enableValidation: Boolean,
+) {
+  val compileJavaModuleInfo =
+    tasks.register<JavaCompile>("compileJavaModuleInfo") {
+      val compileKotlinTask = tasks.getByName("compileKotlinJvm") as KotlinJvmCompile
+      val targetDir = compileKotlinTask.destinationDirectory.dir("../java$javaModuleVersion")
+      val sourceDir = file("src/jvmMain/java$javaModuleVersion")
```
swankjesse:
```suggestion
      val sourceDir = file("src/jvmMain/java9")
```

build-logic/src/main/kotlin/JavaModules.kt
```diff
@@ -38,29 +66,79 @@ fun Project.applyJavaModules(
...
+
+private fun Project.applyJavaModulesMultiplatform(
+  moduleName: String,
+  javaModuleVersion: Int,
+  enableValidation: Boolean,
+) {
+  val compileJavaModuleInfo =
+    tasks.register<JavaCompile>("compileJavaModuleInfo") {
+      val compileKotlinTask = tasks.getByName("compileKotlinJvm") as KotlinJvmCompile
+      val targetDir = compileKotlinTask.destinationDirectory.dir("../java$javaModuleVersion")
```
swankjesse:
```suggestion
      val targetDir = compileKotlinTask.destinationDirectory.dir("../java9")
```

swankjesse (approved):
I’d prefer to hardcode all of these values to `9`, cause that way searching for `java9` turns up results. (We’re unlikely to have separate directories for other Java versions here)

LouisLou2:
> I’d prefer to hardcode all of these values to `9`, cause that way searching for `java9` turns up results. (We’re unlikely to have separate directories for other Java versions here)

@swankjesse Agreed. I've applied all three suggestions.

swankjesse:
Thanks!

PR #9634, opened 2026-08-06
Start @StartStop MockWebServer for @Nested JUnit 5 tests

### LINKED ISSUE #9546: MockWebServer @StartStop field on outer class not started for @Nested JUnit 5 tests

HDPark95:
Fixes #9546.

A `@StartStop MockWebServer` field declared on an outer class is never started when the test runs inside a `@Nested` inner class, so a nested test that uses the server fails with `IllegalStateException: call start() first`.

`StartStopExtension.beforeEach` resolved instance fields from `context.testInstance` (the innermost `@Nested` instance) and `context.requiredTestClass` (the nested class). A `@StartStop` field on an enclosing class is on neither, so it was never found or started. Top-level tests were unaffected because the test instance and its declaring class line up.

The fix walks `context.requiredTestInstances.allInstances`, which returns every instance in the enclosing hierarchy, and starts the annotated fields on each. Behavior is unchanged for non-nested tests, where `allInstances` is the single test instance.

Added `NestedStartStopTest`, which declares a `@StartStop` server on the outer class and one on a `@Nested` inner class and asserts both are started from a nested test. It fails before the change and passes after.

### DIFF
--- a/mockwebserver-junit5/src/main/kotlin/mockwebserver3/junit5/internal/StartStopExtension.kt
+++ b/mockwebserver-junit5/src/main/kotlin/mockwebserver3/junit5/internal/StartStopExtension.kt
@@ -51,24 +51,26 @@ internal class StartStopExtension :
   }
 
   override fun beforeEach(context: ExtensionContext) {
-    // Requires API 24
-    val testInstance = context.testInstance.get()
     val store = context.getStore(Namespace.create(StartStop::class.java))
 
-    val instanceFields =
-      findAnnotatedFields(
-        context.requiredTestClass,
-        StartStop::class.java,
-      ) { !Modifier.isStatic(it.modifiers) }
+    // A @Nested test declares its @StartStop fields on an enclosing instance, so walk every
+    // instance in the hierarchy instead of just the innermost one from context.testInstance.
+    for (testInstance in context.requiredTestInstances.allInstances) {
+      val instanceFields =
+        findAnnotatedFields(
+          testInstance.javaClass,
+          StartStop::class.java,
+        ) { !Modifier.isStatic(it.modifiers) }
 
-    for (field in instanceFields) {
-      field.setAccessible(true)
-      val server = field.get(testInstance) as? MockWebServer ?: continue
+      for (field in instanceFields) {
+        field.setAccessible(true)
+        val server = field.get(testInstance) as? MockWebServer ?: continue
 
-      // Put the instance in the store, so JUnit closes it for us in afterEach.
-      store.put(field, server)
+        // Put the instance in the store, so JUnit closes it for us in afterEach.
+        store.put(field, server)
 
-      server.start()
+        server.start()
+      }
     }
   }
 }
--- a/mockwebserver-junit5/src/test/java/mockwebserver3/junit5/NestedStartStopTest.kt
+++ b/mockwebserver-junit5/src/test/java/mockwebserver3/junit5/NestedStartStopTest.kt
@@ -1,0 +1,43 @@
+/*
+ * Copyright (C) 2025 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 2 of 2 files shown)


swankjesse (approved):

HDPark95:
Thanks for the review!

The two red checks look unrelated to this change, which is isolated to `mockwebserver3-junit5` (`StartStopExtension` plus a new `NestedStartStopTest`):

- `loom` fails in `:container-tests:test`, and that job is also red on the latest `main` (the #9638 merge), so it predates this PR.
- `android (37.0, …, playstore_ps16k)` fails only on `okhttp.android.test.OkHttpTest.testRequestWithSniRequirement` with an emulator `Process crashed`; the other Android matrices (21/23/29/34) and every JVM matrix (8/11/17/21) are green.

Happy to rebase onto `main` if a fresh run would help.

swankjesse:
Thank you!

HDPark95:
Thanks for the quick review and merge!

PR #9638, opened 2026-08-07
Add overdue tests for FakeNetwork

swankjesse:
The test that uses real OkHttp is good to confirm it works, but it isn't useful for debugging problems with FakeNetwork.

### DIFF
--- a/okhttp-testing-support/src/main/kotlin/okhttp3/sockets/FakeSocket.kt
+++ b/okhttp-testing-support/src/main/kotlin/okhttp3/sockets/FakeSocket.kt
@@ -24,7 +24,10 @@ import java.net.SocketAddress
 import java.net.SocketException
 import java.net.SocketOption
 import java.util.concurrent.atomic.AtomicReference
-import okhttp3.internal.connection.BufferedSocket
+import okio.Buffer
+import okio.Sink
+import okio.Source
+import okio.buffer
 
 /**
  * A client or server socket.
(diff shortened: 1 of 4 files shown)


okhttp-testing-support/src/test/kotlin/okhttp3/sockets/FakeNetworkTest.kt
```diff
@@ -0,0 +1,264 @@
...
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.extension.RegisterExtension
+
+@Tag("Slowish")
+class FakeNetworkTest {
+  @RegisterExtension
+  @JvmField
+  val clientTestRule = OkHttpClientTestRule()
+
+  val taskRunner = TaskRunner.INSTANCE
```
yschimke, line 44: It's good you can use either, but would you typically use TaskFaker to be 100% deterministic?

yschimke (approved):

okhttp-testing-support/src/test/kotlin/okhttp3/sockets/FakeNetworkTest.kt
swankjesse: YES! I need to make TaskFaker able to fake out wait + notify.

okhttp-testing-support/src/main/kotlin/okhttp3/sockets/FakeSocket.kt
```diff
@@ -293,3 +309,56 @@ internal class FakeSocket(
     }
   }
 }
+
+internal class SocketSource(
```
swankjesse, line 313: FYI the new tests showed that there’s a race closing sockets which caused some tests to flake.

The fix was straightforward-enough . . . the source & sink offered by `FakeSocket` needs to track its own closed state. Previously I was relying on `Socket.cancel()` to do that, but it closes the peer stream asynchronously and we don’t want that.

PR #9639, opened 2026-08-07
Synthethise IpAddress records from ServiceMetadata hints

yschimke:
We should consider the ip address hints as valid for dns routing, arguably we should even prefer them as they are less likely to have a mismatch.

### DIFF
--- a/okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/dns/-StateMachineDnsCall.kt
+++ b/okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/dns/-StateMachineDnsCall.kt
@@ -242,8 +242,11 @@ class StateMachineDnsCall(
         val mutableReturnedIpAddresses = previous.returnedIpAddresses!!.toMutableSet()
         deduplicatedNewRecords =
           newRecords.filter { record ->
-            // Include the hostname: an address for an alternate service is a distinct result.
-            record !is Dns.Record.IpAddress || mutableReturnedIpAddresses.add(record)
+            when (record) {
+              // ServiceMetadata hints and IpAddress (A/AAAA) may overlap
+              is Dns.Record.IpAddress -> mutableReturnedIpAddresses.add(record)
+              is Dns.Record.ServiceMetadata -> true
+            }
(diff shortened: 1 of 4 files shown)


okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/dns/-StateMachineDnsCall.kt
```diff
@@ -224,9 +236,27 @@ class StateMachineDnsCall(
           else -> previous.pendingExceptions
         }
 
+      val returnedIpAddresses: Set<Dns.Record.IpAddress>?
+      val deduplicatedNewRecords: List<Dns.Record>
```
yschimke, line 240: This feels a lot more complex, but if we wait for someone receiving Dns.Record, then I think all clients need to deal with deduping.

Unless we decide that the routing logic later should deal with those?

swankjesse (commented):
What’s the motivation here? We already expose this data through the API in the existing `ServiceMetadata` type.

(If we’d like to start connecting to those IP addresses, we should, but I think that work goes in `RouteSelector`)

The RFC uses the word _hints_ for these 'cause it wants us to confirm them with DNS calls.

> Failure to use A and/or AAAA response addresses could negatively impact load balancing or other geo-aware features and thereby degrade client performance.

https://datatracker.ietf.org/doc/rfc9460/

yschimke:
Yep, that makes sense. I think it's important we include it. I'm not sure you need overlap with A or AAAA, but you should probably use both?

I think the spec says if we already have them locally? Is that the cache?

swankjesse:
Given enough confidence in our tests, I’d like something like this:

- If we already have a fresh A or AAAA record, use that
- If we’re waiting on a DNS query for an A or AAAA record, connect to the hinted address immediately, instead
- When the A or AAAA records arrive, use those addresses for new calls if they don’t overlap. (Perhaps via `noNewExchanges`.)

yschimke:
@swankjesse might be good to survey user agents and see what they all do. I believe at least one common mobile stack just uses hints if they get any.

PR #9640, opened 2026-08-08
Completely fake out TLS for testing

swankjesse:
This covers the happy path of OkHttp + MockWebServer with HTTPS.

This doesn't actually perform anything cryptographic; certificates and protocols and things are transmitted in-memory rather than over the wire.

This replaces the Okio.inMemorySocketPair() after the fake TLS handshake; that way there's no risk of data leaking from the pre-encrypted session to the post-encrypted session.

The useful capability this unlocks is testing TLS scenarios like ECH retries, before we have a server-side TLS stack that can support them.

### DIFF
--- a/okhttp-testing-support/src/main/kotlin/okhttp3/sockets/InsecureHandshaker.kt
+++ b/okhttp-testing-support/src/main/kotlin/okhttp3/sockets/InsecureHandshaker.kt
@@ -1,0 +1,106 @@
+/*
+ * Copyright (c) 2026 OkHttp Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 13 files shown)


okhttp-testing-support/src/main/kotlin/okhttp3/sockets/InsecureHandshaker.kt
```diff
@@ -0,0 +1,106 @@
...
+        Handshaker.ClientAuth.None -> {
+          listOf()
+        }
+      }
+
+    val serverCertificates = server.handshakeCertificates.keyManager.serverCertificates(keyType)
+
+    val (clientSocket, serverSocket) = inMemorySocketPair(maxBufferSize = 1024 * 1024)
+
+    return Handshaker.Result(
```
yschimke, line 68: Fun to see. I still hope we can kick over Tls Socket code to those suckers who work on okio. :)

This could be a nice fake over there.

yschimke (approved):

okhttp-testing-support/src/main/kotlin/okhttp3/sockets/InsecureHandshaker.kt
swankjesse: I think your `SocketAdapter` abstraction is really the thing we need around these parts. A good enough `SocketAdapter` API might make it so we can confirm Conscrypt conforms to our Socket expectations, without actually running all the tests against it.

swankjesse: (And similarly for all other TLS things)

PR #9641, opened 2026-08-08
Not to merge: hack all of CallTest working with FakeNetwork

swankjesse:


swankjesse:
I was using `CallTest` to exercise `FakeNetwork` and `FakeTls`, and both fared well.

Some challenges: 
 * Finding all of the places we use the default SSL stack. For example, in `MockWebServer.processHandshakeFailure()`.
 * Getting `Socket.close()` to work successfully for both async cancellation and graeful close cases.
 * Tests that assert on exact exception message text.

I didn’t do the work to make `FakeSocket` work properly with socket leak detection. I’m unsure why that didn’t just work.

PR #9643, opened 2026-08-09
reject SETTINGS_HEADER_TABLE_SIZE larger than 2^31 - 1

basavaraj-sm05:
readSettings decodes each SETTINGS value with a signed readInt(), and the size-shaped settings are meant to be range checked right where they're read. SETTINGS_INITIAL_WINDOW_SIZE and SETTINGS_MAX_FRAME_SIZE both reject a value whose high bit is set, but the SETTINGS_HEADER_TABLE_SIZE branch is empty, so a peer that advertises a header table size of 2^31 or more lands a negative int in the Settings. That value clears the != -1 guard in Http2Writer.applyAndAckSettings and reaches hpackWriter.resizeHeaderTable, where minOf(negative, 16384) stays negative and maxDynamicTableByteCount goes negative with it; adjustDynamicTableByteCount then runs evictToRecoverBytes against an empty dynamic table and walks off the end into a null slot, so the write side throws NullPointerException rather than a protocol error and the encoder's byte accounting is left broken. I noticed it lining up which settings check their value against which don't. Rejecting a negative header table size in the reader, the same way the two neighboring settings already do, keeps the bad value out of the encoder, and valid sizes decode unchanged since okhttp caps its own table at 16 KiB regardless.

### DIFF
--- a/okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/http2/Http2Reader.kt
+++ b/okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/http2/Http2Reader.kt
@@ -266,6 +266,9 @@ class Http2Reader(
       when (id) {
         // SETTINGS_HEADER_TABLE_SIZE
         1 -> {
+          if (value < 0) {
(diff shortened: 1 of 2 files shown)


basavaraj-sm05:
gentle ping

PR #9647, opened 2026-08-10
Set publish workflow JDK version inline

### LINKED ISSUE #9547: publish workflow references a missing .github/workflows/.java-version (setup-java fails; recent publish runs failed)
Hi, and thank you for OkHttp.

The publish workflow references a file that doesn't exist, and its recent runs have been failing.

`.github/workflows/publish.yml` (triggered on `push: tags: '**'`) configures Java via a version file:

```yaml
- uses: actions/setup-java@v5
  with:
    distribution: 'temurin'
    java-version-file: .github/workflows/.java-version
```

But `.github/workflows/.java-version` does not exist:

```
$ gh api repos/square/okhttp/contents/.github/workflows/.java-version
# → 404 Not Found
```

`setup-java` fails when `java-version-file` points to a missing file, and the recent `publish` runs reflect that — the two "Prepare for release 5.4.0" runs both completed as **failure** in ~12–14s (fast-fail consistent with the setup step), e.g. run 27162061368.

**Suggested fix:** add `.github/workflows/.java-version` containing the intended JDK version (the build workflow uses Java 21 for the primary matrix, so `21` is a likely value), or switch the step to an inline `java-version:`. I didn't want to guess the exact version and open a PR that might pick the wrong one — happy to send a one-line PR once you confirm the intended version.

For transparency: I used AI assistance to spot and draft this; I verified the 404 and the failing runs myself.

iVamsi:
publish.yml points at a .java-version file that does not exist, so tag publishes fail at setup-java. 

Fixes #9547.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -20,7 +20,7 @@ jobs:
       - uses: actions/setup-java@v5
         with:
           distribution: 'temurin'
-          java-version-file: .github/workflows/.java-version
+          java-version: 21
 
       - run: ./gradlew publish
         env:


JakeWharton:
The fix is to create the file, not inline.

iVamsi:
> The fix is to create the file, not inline.

I added inline to match other workflows, but I'm happy to follow up with the file if you prefer that.

PR #9650, opened 2026-08-10
Change DnsOverHttps to query the correct port per Attrleaf

swankjesse:


okhttp/src/commonTest/kotlin/okhttp3/internal/dns/StateMachineDnsCallTest.kt
```diff
@@ -80,6 +80,96 @@ class StateMachineDnsCallTest {
...
+      }
+    }
+  }
+
+  @Test
+  fun `attrleaf is not used for http port 80`(caching: Boolean = true) {
+    testStateMachineDnsCall {
+      val call =
+        newCall(
+          request = Dns.Request(hostname = "lysine.dev", port = 80),
```
yschimke: Why not for 80? Isn't 443 the default?

yschimke (approved):

okhttp/src/commonTest/kotlin/okhttp3/internal/dns/StateMachineDnsCallTest.kt
swankjesse: hmmm... there's a special rule where the presence of an HTTPS record is supposed to automatically promote HTTP to HTTPS.

But I can do that a nicer way, by picking port 443 when we create the query. Let me do that...

PR #9653, opened 2026-08-11
Flush HTTP/1 request bodies before detachTimeout

### LINKED ISSUE #9228: writeTimeoutMillis not active during RequestBody.writeTo
> @swankjesse I tried to implement this using 
> 
> ```
>             .socketFactory(
>               DelayingSocketFactory(onWrite = {
>                 Thread.sleep(100L)
>               }),
> ```
> 
> But if I break point at Thread.sleep, then socket.sink().timeout().timeoutNanos() is always cleared (0).
> 
> I suspect a bug with write timeout, but this replicates a timeout just by checking if we have the bad timeout value. 

 _Originally posted by @yschimke in [#9108](https://github.com/square/okhttp/pull/9108/changes#r2616266797)_

iVamsi:
HTTP/1 keeps the connection sink buffered. Closing a request body called detachTimeout() and cleared the write timeout; then finishRequest() flushed the remaining bytes with the timeout already gone.

Flush in KnownLengthSink/ChunkedSink.close() before detachTimeout() so that write still has write timeout. Also switched InterceptorOverridesTest's WriteTimeout case to a delayed socket write (with a request body), same idea as ReadTimeout.

Fixes #9228.

### DIFF
--- a/okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/http1/Http1ExchangeCodec.kt
+++ b/okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/http1/Http1ExchangeCodec.kt
@@ -335,6 +335,8 @@ class Http1ExchangeCodec(
     override fun close() {
       if (closed) return
       closed = true
+      // Flush before detachTimeout().
+      socket.sink.flush()
       detachTimeout(timeout)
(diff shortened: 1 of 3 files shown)


swankjesse (approved):
Nice test

PR #9654, opened 2026-08-11
Get the FakeNetwork via OkHttp's existing Platform thing

swankjesse:


okhttp/src/jvmTest/kotlin/okhttp3/FakeNetworkOkHttpTest.kt
```diff
@@ -20,46 +20,43 @@ import assertk.assertions.isEqualTo
...
 
-  @StartStop
-  private val server =
-    MockWebServer()
-      .apply {
-        serverSocketFactory = network.serverSocketFactory
-      }
+  private val handshakeCertificates = platformRule.localhostHandshakeCertificates()
+
+  // We can't create these until after platformRule runs. Sigh.
```
swankjesse, line 44: I think the fix for this is to do a small bit of DI in PlatformRule

okhttp-tls/src/test/java/okhttp3/tls/HandshakeCertificatesTest.kt
```diff
@@ -180,7 +179,7 @@ class HandshakeCertificatesTest {
   }
 
   private fun startTlsServer(): InetSocketAddress {
-    val serverSocketFactory = ServerSocketFactory.getDefault()
+    val serverSocketFactory = Platform.get().serverSocketFactory
```
yschimke, line 182: overall these seem like a net positive.

yschimke (approved):

PR #9656, opened 2026-08-12
Handle unsupported AliasMode records

swankjesse:


okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/dns/-StateMachineDnsCall.kt
```diff
@@ -105,26 +105,27 @@ class StateMachineDnsCall(
         return
       }
 
-      val queries =
-        questions.map { question ->
+      val questionToQuery =
+        questions.associateWith { question ->
```
yschimke, line 109: TIL The returned map preserves the entry iteration order of the original array.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/dns/-StateMachineDnsCall.kt
```diff
@@ -196,14 +209,23 @@ class StateMachineDnsCall(
               echConfigList = resourceRecord.echConfigList,
             )
           }
+        }
+
+        TYPE_A, TYPE_AAAA -> {
+          resourceRecords.mapNotNull { resourceRecord ->
+            // Discard resource records that don't fit the query.
```
yschimke, line 216: this seems unlikely and worth warning about, but I assume we think it won't happen, so ignore?

Arguably ew should throw away A answers for a AAAA query, but I'm guessing this is a smart cast.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/dns/-StateMachineDnsCall.kt
```diff
@@ -177,10 +179,21 @@ class StateMachineDnsCall(
...
-          is ResourceRecord.Https -> {
+    val dnsRecords: List<Dns.Record> =
+      when (question.type) {
+        TYPE_HTTPS -> {
+          resourceRecords.mapNotNull { resourceRecord ->
+            // Discard resource records that don't fit the query.
+            if (resourceRecord !is ResourceRecord.Https) return@mapNotNull null
+
+            // OkHttp doesn't yet implement AliasMode resource records. If any AliasMode record is
+            // returned, we must ignore ALL returned resource records.
```
yschimke, line 190: TODO log something observable?

yschimke (approved):

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/dns/-StateMachineDnsCall.kt
swankjesse: Yeah even if we did warn, it's unlikely the recipient of the warning would be able to do something with it.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/dns/-StateMachineDnsCall.kt
swankjesse: I don't think the recipient of the warning is the person who can do anything about it. 

There's probably a DNS record linter tool for DNS admins to learn that their records are malformed.

PR #9658, opened 2026-08-12
Update EchTest to reflect current behavior

swankjesse:
Prior to these changes the test fails for me.


swankjesse:
... and I think we might need to further exercise retries.

android-test/src/androidTest/java/okhttp/android/test/EchTest.kt
yschimke, line 51: Ahhh, shit. Not running on CI because of this.

yschimke (approved):

PR #9659, opened 2026-08-13
Don't close the multipart sink when a part closes its stream

### LINKED ISSUE #7692: `okhttp3.RequestBody.Companion#gzip` throws `java.lang.IllegalStateException: closed` if used in `MultipartBody`
[Reproducer](https://gist.github.com/korniltsev/817171107399602e19defa82c49da532) . Tried it on  `5.0.0-alpha.11`

I believe I can workaround it in our project locally by creating a vendored copy of GzipSink which does not close upstream sink but flushes it with some help of ugly dirty java tricks accessing kotlin internals it should work but it would be nice to be able to use the original extension method in MultipartBody

Stacktrace:
<details>
Exception in thread "main" java.lang.IllegalStateException: closed
	at okio.RealBufferedSink.write(RealBufferedSink.kt:176)
	at okhttp3.MultipartBody.writeOrCountBytes(MultipartBody.kt:162)
	at okhttp3.MultipartBody.writeTo(MultipartBody.kt:99)
	at okhttp3.internal.http.CallServerInterceptor.intercept(CallServerInterceptor.kt:64)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
	at okhttp3.internal.cache.CacheInterceptor.intercept(CacheInterceptor.kt:95)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
	at okhttp3.internal.http.BridgeInterceptor.intercept(BridgeInterceptor.kt:84)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
	at okhttp3.internal.http.RetryAndFollowUpInterceptor.intercept(RetryAndFollowUpInterceptor.kt:65)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
	at okhttp3.internal.connection.RealCall.getResponseWithInterceptorChain$okhttp(RealCall.kt:205)
	at okhttp3.internal.connection.RealCall.execute(RealCall.kt:158)
	at MainKt.main(Main.kt:22)
</details>

iVamsi:
A request body closing its sink was also closing the stream that OkHttp was still writing to. In a multipart request, that meant the next part (or even the bytes after the first part) failed with IllegalStateException: closed.

The part still needs to finish its stream. This change lets it do that without closing the stream MultipartBody was handed.

Fixes #7692.


JakeWharton:
This seems like it fixes a very specific variant of this bug, but not the whole class of bug.

iVamsi:
@JakeWharton Yeah, gzip was just one body that closed the sink. Moved the fix to MultipartBody.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/MultipartBody.kt
```diff
@@ -160,7 +162,12 @@ class MultipartBody internal constructor(
       if (countBytes) {
         byteCount += contentLength
       } else {
-        body.writeTo(sink)
+        object : ForwardingSink(sink) {
+          override fun close() {
+            // GzipSink.close() writes the gzip footer, then closes whatever it wraps.
+            // Swallow that close so we can write the trailing CRLF and later parts.
+          }
+        }.buffer().use(body::writeTo)
```
JakeWharton:
```suggestion
        class NoCloseSink(sink: Sink) : ForwardingSink(sink) {
          override fun close() {}
        }
        body.writeTo(NoCloseSink(sink).buffer())
```

okhttp/src/commonJvmAndroid/kotlin/okhttp3/RequestBody.kt
```diff
@@ -41,7 +41,7 @@ abstract class RequestBody {
   @Throws(IOException::class)
   open fun contentLength(): Long = -1L
 
-  /** Writes the content of this request to [sink]. */
+  /** Writes the content of this request to [sink]. This must not close [sink]. */
```
JakeWharton:
```suggestion
  /** Writes the content of this request to [sink]. This should not close [sink]. */
```
Either we guarantee this by crashing, or we relax the wording precisely because we _are_ handling `close` appropriately.

okhttp/src/jvmTest/kotlin/okhttp3/MultipartBodyTest.kt
```diff
@@ -254,6 +259,92 @@ class MultipartBodyTest {
...
+        .Builder("123")
+        .addPart(ClosingBody("hello"))
+        .addPart("world".toRequestBody(null))
+        .build()
+    assertThat(body.boundary).isEqualTo("123")
+    assertThat(body.type).isEqualTo(MultipartBody.MIXED)
+    assertThat(body.contentType().toString())
+      .isEqualTo("multipart/mixed; boundary=123")
+    assertThat(body.parts.size).isEqualTo(2)
+    assertThat(body.contentLength()).isEqualTo(-1)
```
JakeWharton:
```suggestion
```
This is other test problems

okhttp/src/jvmTest/kotlin/okhttp3/MultipartBodyTest.kt
```diff
@@ -254,6 +259,92 @@ class MultipartBodyTest {
...
+        .Builder("123")
+        .addPart(GzipRequestBody("part1".toRequestBody(null)))
+        .addPart(GzipRequestBody("part2".toRequestBody(null)))
+        .build()
+    assertThat(body.boundary).isEqualTo("123")
+    assertThat(body.type).isEqualTo(MultipartBody.MIXED)
+    assertThat(body.contentType().toString())
+      .isEqualTo("multipart/mixed; boundary=123")
+    assertThat(body.parts.size).isEqualTo(2)
+    assertThat(body.contentLength()).isEqualTo(-1)
```
JakeWharton:
```suggestion
```
Other test problems

okhttp/src/jvmTest/kotlin/okhttp3/MultipartBodyTest.kt
```diff
@@ -254,6 +259,92 @@ class MultipartBodyTest {
...
+      assertThat(part1Body).isEqualTo("part1")
+
+      val part2 = reader.nextPart()!!
+      val part2Body =
+        GzipSource(part2.body).use {
+          it.buffer().readUtf8()
+        }
+      assertThat(part2Body).isEqualTo("part2")
+      assertThat(reader.nextPart()).isNull()
+    }
```
JakeWharton: I would rather see this all replaced with a `readUtf8()` like the above test. We don't care about parsing, we care that the entire request was written.

JakeWharton (approved):
Just some small tweaks

okhttp/src/commonJvmAndroid/kotlin/okhttp3/MultipartBody.kt
JakeWharton: The `use` is pointless here, because it guarantees a `close()` call which we're already going out of our way to suppress.

iVamsi: .buffer() adds a RealBufferedSink on top of NoCloseSink. .use closes it so the last segment is flushed. NoCloseSink.close() is still a no-op.

Without .use, 9 of 14 MultipartBodyTest tests failed.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/RequestBody.kt
iVamsi: Updated.

okhttp/src/jvmTest/kotlin/okhttp3/MultipartBodyTest.kt
iVamsi: Removed

okhttp/src/jvmTest/kotlin/okhttp3/MultipartBodyTest.kt
iVamsi: Removed

okhttp/src/jvmTest/kotlin/okhttp3/MultipartBodyTest.kt
iVamsi: Updated

okhttp/src/commonJvmAndroid/kotlin/okhttp3/MultipartBody.kt
JakeWharton: Ah, yeah. Makes sense. The whole try/finally of use is probably overkill, but it's not a hot path and it's less code.

PR #9660, opened 2026-08-13
Complete fixed-length duplex request before reporting length mismatch

### LINKED ISSUE #9648: Closing a partial fixed-length duplex request body does not complete the request side
This was observed with OkHttp 5.3.2 while working on https://github.com/ktorio/ktor/pull/5791. The current OkHttp `main` branch contains the same relevant implementation.

When a fixed-length duplex request is cancelled before all declared bytes are written, closing its sink throws:

```text
ProtocolException: unexpected end of stream
```

The exception is expected, but the request side is not marked complete, leaving the connection allocated:

```text
connectionCount() == 1
idleConnectionCount() == 0
```

### How to reproduce

The observed sequence is:

1. Start an HTTP/2 duplex request whose body returns a positive `contentLength()`.
2. Write only part of the request body and receive the server's final response.
3. Cancel the call and close the partially written request sink.
4. Observe the length-mismatch exception and that the connection does not become idle.

`Exchange.RequestBodySink.close()`throws before calling `complete(...)`:

https://github.com/lysine-dev/okhttp/blob/61423f472da24e0ccc42b6a2c0863fb27932fea5/okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/connection/Exchange.kt#L294-L306

A subsequent `close()` cannot complete the request because `closed` is already `true`.

Should the request side be completed with failure before reporting the length mismatch, so the cancelled call can release its connection allocation?

### Workaround

As a workaround in Ktor, after request cancellation we attempt to call `flush()` before the sink is closed. Because the HTTP/2 stream has already been reset, `RequestBodySink.flush()` fails and invokes `complete(e)`, which marks the request side complete.

Unfortunately this relies on OkHttp’s internal implementation and only works when `flush()` fails after cancellation.

Sskift:
## Problem

When a fixed-length duplex request body is closed before all declared bytes are written, `Exchange.RequestBodySink.close()` throws a `ProtocolException` ("unexpected end of stream") **before** calling `complete(...)`. As a result the request side is never marked complete, so the HTTP/2 connection allocation is never released:

```
connectionCount() == 1
idleConnectionCount() == 0
```

See issue #9648 for the original report.

## Fix

Complete the request side with failure before reporting the length mismatch, so the cancelled call can release its connection allocation:

```kotlin
if (contentLength != -1L && bytesReceived != contentLength) {
  throw complete(ProtocolException("unexpected end of stream"))!!
}
```

## Test

Added `DuplexTest.fixedLengthDuplexRequestCompletedWhenClosedEarly`, which writes a partial fixed-length duplex body, receives the server's final response, then closes the sink. Before the fix the test fails (connection stays allocated, `idleConnectionCount() == 0`); after the fix it asserts `idleConnectionCount() == 1`.

Fixes #9648.

### DIFF
--- a/okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/connection/Exchange.kt
+++ b/okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/connection/Exchange.kt
@@ -295,7 +295,7 @@ class Exchange(
       if (closed) return
       closed = true
       if (contentLength != -1L && bytesReceived != contentLength) {
-        throw ProtocolException("unexpected end of stream")
+        throw complete(ProtocolException("unexpected end of stream"))!!
       }
       try {
         super.close()
--- a/okhttp/src/jvmTest/kotlin/okhttp3/DuplexTest.kt
+++ b/okhttp/src/jvmTest/kotlin/okhttp3/DuplexTest.kt
@@ -736,6 +736,59 @@ class DuplexTest {
       .contains("StreamResetException: stream was reset: CANCEL")
   }
 
+  @Test
+  fun fixedLengthDuplexRequestCompletedWhenClosedEarly() {
+    enableProtocol(Protocol.HTTP_2)
+    val body =
+      MockSocketHandler()
+        .receiveRequest("partial")
+        .sendResponse("response A\n")
+        .sleep(2, TimeUnit.SECONDS)
+    server.enqueue(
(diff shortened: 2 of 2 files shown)


JakeWharton:
Thank you for your PR. Unfortunately, we are about to ban LLM use for both code and issue/PR descriptions as well as comments. Hopefully that policy will be up in our repos today.

PR #9665, opened 2026-08-13
Schedule 4.x unsupported in 24 months

swankjesse:


swankjesse:
5.x is extremely compatible with 4.x, so users can safely upgrade to 5.x.

yschimke:
That's an argument for 2027-08-31

yschimke (approved):

PR #9666, opened 2026-08-14
New 'No Generative Tools' section in contributing doc

swankjesse:
Also drop the link to Square's CLA and bug bounty, which are no longer part of this project.

### DIFF
--- a/docs/contribute/contributing.md
+++ b/docs/contribute/contributing.md
@@ -1,10 +1,12 @@
 Contributing
 ============
 
+Our goal with this project is to build a simple, fast, and secure component that you can trust. We
+have scoped the project deliberately to address common use cases easily, and made it extensible so
+it'll handle exotic use cases too.
+
 Keeping the project small and stable limits our ability to accept new contributors. We are not
 seeking new committers at this time, but some small contributions are welcome.
-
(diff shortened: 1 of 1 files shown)


JakeWharton (approved):

yschimke (approved):
Thanks for writing this up.

docs/contribute/contributing.md
```diff
@@ -1,19 +1,38 @@
...
+
+We require all contributors to do likewise. Building is more fun when everyone is making an effort.
+
+We’ll immediately reject LLM-generated contributions to protect the culture of our project. We ban
+repeat offenders.
+
+Some narrow exceptions to this policy:
+
+ * Non-English speakers may use machine translation tools if they are disclosed.
+ * Local autocomplete. (Avoid tools that are metered in tokens.)
```
yschimke, line 33: This seems stricter than I expected, it would rule out even asking an Agent to rebase and fix the merge conflicts. But ok, I'll need to avoid even that sort of usage.

JakeWharton: That seems like a good property. Resolving merge conflicts is arguably more challenging than just writing a normal change. You first have to understand both sides of the change and then determine the correct reconciliation.

PR #9668, opened 2026-08-14
Disable Loom CI shard

JakeWharton:
It fails 100% of the time, so any real failures today are being ignored anyway.


yschimke (approved):
I'll follow up on https://github.com/lysine-dev/okhttp/issues/9667

JakeWharton:
Yeah we jinxed on this!

JakeWharton:
Forgot to use a loom-themed branch name 🤦

PR #9669, opened 2026-08-14
Use BouncyCastle BOM

JakeWharton:
They patch these independently, so just rely on the BOM to figure out what's latest.


JakeWharton:
Oh okay I figured that shit was just my local env

yschimke (approved):

swankjesse (approved):

swankjesse:
Spotless doesn’t like your tab!

JakeWharton:
Yeah something is wrong with my IntelliJ. It keeps mixing up indent settings and not honor editorconfig when I have multiple projects open.

PR #9671, opened 2026-08-14
Test ECH + HTTP proxy

swankjesse:
We had a bug where we incorrectly sent the proxy's ECH config list to the origin server. That wouldn't have worked.

### DIFF
--- a/okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/connection/RouteSelector.kt
+++ b/okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/connection/RouteSelector.kt
@@ -25,6 +25,7 @@ import okhttp3.Address
 import okhttp3.Dns
 import okhttp3.HttpUrl
 import okhttp3.Route
+import okhttp3.internal.OkHttpInternalApi
 import okhttp3.internal.canParseAsIpAddress
 import okhttp3.internal.dns.LookupDnsCall
 import okhttp3.internal.dns.execute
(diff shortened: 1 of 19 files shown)


okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/connection/RouteSelector.kt
```diff
@@ -249,7 +251,11 @@ class RouteSelector internal constructor(
                 address = address,
                 proxy = proxy,
                 socketAddress = InetSocketAddress(record.address, socketPort),
-                echConfigList = serviceMetadata?.echConfigList,
+                echConfigList =
+                  when (proxy.type()) {
+                    Proxy.Type.DIRECT -> serviceMetadata?.echConfigList
+                    else -> null
+                  },
```
swankjesse, line 258: This was a pretty nasty one. We sent the proxy’s ECH config list to the origin server

yschimke: Good catch.

yschimke (approved):

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/connection/RouteSelector.kt
swankjesse: This would have manifested as a connectivity problem for clients that use an HTTP proxy, if the HTTP proxy itself has an ECH record. Possibly not likely in practice? Unsure.

swankjesse: Also, I love tests. That’s how we build complex things safely!

PR #9678, opened 2026-08-14
Only one contributing.md doc

swankjesse:
This originally worked but our .gitignore got out of sync. We gotta make our website like Retrofit's.


JakeWharton (approved):
Oh is it copied as part of the build? I didn't even think to look for that.

PR #9680, opened 2026-08-14
Don't recover after an ECH public_name fails to verify

swankjesse:


okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/connection/ConnectPlan.kt
```diff
@@ -510,10 +511,14 @@ class ConnectPlan internal constructor(
     val nextEchRetryPlan = Platform.get().echRetryPlan(sslException)
     if (nextEchRetryPlan != null) {
       // Validate the publicHostname against the session certificate. The session is protected by
-      // the outer client hello (e.g. cloudflare-ech.com), not the origin server.
+      // the outer client hello (e.g. cloudflare-ech.com), not the origin server. If this fails,
+      // we must not retry further.
```
yschimke, line 515: Why is it risky to try another route? Is this in the spec, can we link it?

Let's say we get 3 possible IP addresses

IP Address A 1 - this DNS result is stale and now points at some unrelated server
IP Address A 2
HTTPS ip address hint

Does this mean A1 still points at something answering cloudflare-ech.com for the outer, but that is misconfigured for ECH? or that we connected to cdn.aws.com and it didn't have the records?

yschimke (approved):
`no fallback after ech failure` explains it, but I just haven't internalised it.

okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/connection/ConnectPlan.kt
swankjesse: This is what I’m using to guide this behavior:

> If a client does not retry, it MUST report an error to the calling application.

https://www.rfc-editor.org/rfc/rfc9849.html#section-6.1.6

That said, the _retrying_ here is related to ECH retries, not necessarily retries for other routes.

I’d like ship this. If any user reports that this harms connectivity we should do more analysis and decide whether we want more connectivity at a cost of potential downgrade attacks. But not having done the analysis yet my preference is to reduce connectivity.

PR #9682, opened 2026-08-15
Flip wiring of Dns.SYSTEM

swankjesse:
Previously Dns.SYSTEM was a constant, and Platform.platformDns() was a hook to return something different.

With this change, Dns.SYSTEM forwards to whatever Platform.systemDns currently returns.

### DIFF
--- a/okhttp/src/commonJvmAndroid/kotlin/okhttp3/Dns.kt
+++ b/okhttp/src/commonJvmAndroid/kotlin/okhttp3/Dns.kt
@@ -13,13 +13,17 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+@file:OptIn(OkHttpInternalApi::class)
+
 package okhttp3
 
 import java.io.IOException
 import java.net.InetAddress
 import java.net.UnknownHostException
+import okhttp3.internal.OkHttpInternalApi
 import okhttp3.internal.concurrent.TaskRunner
 import okhttp3.internal.dns.LookupDnsCall
+import okhttp3.internal.platform.Platform
 import okhttp3.internal.toCanonicalHost
 import okio.ByteString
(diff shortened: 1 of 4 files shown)


okhttp/src/commonJvmAndroid/kotlin/okhttp3/Dns.kt
```diff
@@ -279,22 +283,18 @@ fun interface Dns {
...
     /**
-     * A DNS that uses [InetAddress.getAllByName] to ask the underlying operating system to
-     * lookup IP addresses. Most custom [Dns] implementations should delegate to this instance.
+     * The current process's host DNS service.
+     *
+     * Most custom [Dns] implementations should delegate to this instance.
      */
     @JvmField
-    val SYSTEM: Dns = DnsSystem()
+    val SYSTEM: Dns =
```
yschimke, line 291: Should we provide a Dns.InetAddressDns to allow getting back the existing behaviour?

yschimke, line 291: This did make me question the meaning of "SYSTEM", I was figuring SYSTEM remained as is, and OkHttpClient defaulted to Platform.dns()

But I think I'm ok with it.

yschimke (approved):

okhttp/src/commonJvmAndroid/kotlin/okhttp3/Dns.kt
swankjesse: I don’t think we need that yet. `Dns.SYSTEM` currently always does that.

I’d like to make `SYSTEM` point at `AndroidDns` with `includeServiceMetadata=true` later. Once the calling code has more parallelism.

PR #9686, opened 2026-08-15
Revert "Use a fixed size for DoH request body"

yschimke:
This reverts commit 0070c2e7b4c162ee03abf1c88fd94083e94697f4.

closes #9685

PR #9687, opened 2026-08-15
Use a fixed size for DoH request body

yschimke:
closes #9686 

Cloudflare seems unique in rejecting chunked request bodies


okhttp-dnsoverhttps/src/main/kotlin/okhttp3/dnsoverhttps/internal/-DnsOverHttpsQuery.kt
```diff
@@ -166,12 +166,17 @@ internal fun DnsMessage.asQueryParameter(): String {
 }
 
 internal class QueryRequestBody(
-  private val query: DnsMessage,
+  query: DnsMessage,
 ) : RequestBody() {
+  private val content = Buffer().also { DnsMessageWriter(it).write(query) }.readByteString()
```
swankjesse: If we're doing this anyway, there's an API that turns a ByteString into a RequestBodg that we could use.

swankjesse (approved):

PR #9691, opened 2026-08-16
downgrade a public-suffix Domain cookie to host-only

basavaraj-sm05:
When a Set-Cookie header carries a Domain attribute, Cookie.parse checks that the domain isn't a public suffix, so a server can't scope a cookie to something like `github.io` or a `*.elb.amazonaws.com` label and have it ride along to every host underneath. The problem is that the check is gated on `urlHost.length != domain.length`, so it only runs when the Domain is a proper suffix of the request host and is skipped whenever the Domain equals the host exactly. A response from a host that is itself a public suffix slips right through: `https://github.io/` returning `Set-Cookie: a=b; Domain=github.io` never reaches the public-suffix lookup, hostOnly stays false, and the cookie then matches every `*.github.io` host in `Cookie.matches`. RFC 6265 section 5.3 covers this exact case by turning the cookie into a host-only cookie rather than a shared domain cookie. I ran into it reading the length comparison next to the existing `domainIsPublicSuffix` test, which only walks the proper-suffix path and never the equal-length one. The fix runs the public-suffix lookup whenever a Domain is present, still refuses a proper-suffix public suffix as before, and marks the cookie host-only when the domain equals the host; registrable domains like example.com are untouched because their eTLD+1 is non-null.

### DIFF
--- a/okhttp/src/commonJvmAndroid/kotlin/okhttp3/Cookie.kt
+++ b/okhttp/src/commonJvmAndroid/kotlin/okhttp3/Cookie.kt
@@ -556,17 +556,19 @@ class Cookie private constructor(
       val urlHost = url.host
       if (domain == null) {
         domain = urlHost
-      } else if (!domainMatch(urlHost, domain)) {
-        return null // No domain match? This is either incompetence or malice!
-      }
-
-      // A public suffix can't be used as a cookie domain. If it's the request host itself the
-      // cookie is downgraded to host-only; if it's a proper suffix of the host it's refused.
-      if (PublicSuffixDatabase.get().getEffectiveTldPlusOne(domain) == null) {
-        if (urlHost.length != domain.length) {
-          return null
+      } else {
+        if (!domainMatch(urlHost, domain)) {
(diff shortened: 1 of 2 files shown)


swankjesse (commented):
Great analysis. One hazard with this code change is it causes us to evaluate `PublicSuffixDatabase.get()` more eagerly, and that’s a potentially expensive operation.

Will investigate

basavaraj-sm05:
You're right, and it was actually worse than the diff suggests: since domain defaults to urlHost when no Domain attribute is present, the patch as written ran the lookup for every cookie, Domain or not. I've pushed a change that nests the public-suffix check inside the Domain branch, so parses without a Domain attribute never touch PublicSuffixDatabase, same as today. The remaining delta from main is the Domain-equals-host case, and I don't see a way around that one: deciding whether to downgrade to host-only means asking whether the host is a public suffix, so a server that echoes its own host in Domain now pays the one-time list load it previously skipped. After that it's just a few binary searches per parse.

basavaraj-sm05:
any update here?

PR #9729, opened 2026-08-24
Cap connection-failure retries at 21 attempts

### LINKED ISSUE #9727: Lock convoy between Http2Writer monitor and Okio's global AsyncTimeout lock causes stuck requests and unbounded heap growth in `recoveredFailures`
## OkHttp version
4.12.0

## Okio version
3.6.0

## Description

In a production Spring Boot service (JDK 21, using OkHttp as the transport for an HTTP/2 API client), we experienced an incident where two Tomcat request threads never returned, eventually consuming ~945MB (~72%) of the JVM heap and leaving hundreds of other threads `BLOCKED`/`WAITING` for over 2 hours until the instance was manually restarted.

We reconstructed the following causal chain from heap dumps (Eclipse MAT) and full thread dumps (including virtual threads):

### 1. `RetryAndFollowUpInterceptor.recoveredFailures` grew unbounded on a single call

Two request threads accumulated recovered connection-failure exceptions without limit:

okhttp3.internal.http.RetryAndFollowUpInterceptor.intercept()
  -> recoveredFailures: ArrayList<IOException>

Heap contents for the two stuck threads:

| Thread | Retained heap | recoveredFailures size |
|---|---:|---:|
| exec-A | 781,957,496 bytes (59.4%) | ~40,000 |
| exec-B | 162,947,616 bytes (12.4%) | ~8,310 |

Dominant object types across the heap: `[java.net](http://java.net/).SocketException` (~48,319), `[java.io](http://java.io/).IOException` (~48,310), `java.lang.StackTraceElement` (~9.67M).

Because the request threads never returned, these local variables stayed reachable from the thread stack GC root and were never collected.

### 2. Threads holding `Http2Writer` were blocked on the same global `AsyncTimeout.lock`

At the same time, both stuck threads held a **different** `Http2Writer` monitor each, while both were waiting to acquire the **same** Okio companion-object lock:

exec-A -> holds Http2Writer@0x...590 -> waiting on AsyncTimeout.lock
exec-B -> holds Http2Writer@0x...bd0 -> waiting on AsyncTimeout.lock

Other request threads then queued up `BLOCKED` behind whichever `Http2Writer` monitor they needed next. Standard JVM deadlock detection reported nothing (no classic cyclic wait), but functionally this was a persistent lock convoy: the same lock relationship was observed in two thread dumps taken ~2h36m apart.

### 3. The call-level timeout mechanism itself could not fire

  -> AsyncTimeout.timedOut() -> RealCall.cancel()
  -> internally needs AsyncTimeout.lock to update the timeout queue
  -> lock already contended by the writer threads above
  -> cancellation itself stalls

So the safety mechanism that should have terminated the call was blocked by the same global lock it needed to update its own scheduling state.

### 4. Amplification factors on our side (application-level, not OkHttp's fault, listed for context)

- An HTTP client SDK layered on top of OkHttp had its own retry (2 attempts) enabled, in addition to `retryOnConnectionFailure(true)`.
- Our application-level timeouts (`Future.get(timeout)`) only stopped the *caller's* wait; they never called `Call.cancel()` on the underlying OkHttp call, so nothing forced the retry loop to end from our side either.
- JDK 21 virtual thread pinning was also observed (`VirtualThread.parkOnCarrierThread()` while holding a `synchronized` monitor), which we believe worsened propagation but was not the primary driver of the
heap growth.

adityaanikam:
﻿Fixes #9727

RetryAndFollowUpInterceptor's connection failure retry loop had no cap, unlike its sibling follow up loop (redirects and auth challenges), which already gives up after MAX_FOLLOW_UPS. A route selector that keeps reporting a usable route available let recover() retry indefinitely.

In a production incident, two threads accumulated tens of thousands of recovered failures each on a single stuck call, growing recoveredFailures without bound, while both held a per connection Http2Writer monitor and contended for Okio's global AsyncTimeout lock. Because the calls never returned, the timeout mechanism meant to cancel them needed that same contended lock to update its own scheduling state, so cancellation stalled too. Full details and thread dump analysis are in the issue.

Added MAX_RECOVERED_FAILURES, mirroring MAX_FOLLOW_UPS's value and placement, and thrown the same way an unrecoverable failure already is. This bounds how long any single call can spend retrying and contending for locks a stuck attempt is still holding, regardless of how the route selector behaves.

Added a regression test providing more usable routes (25) than the cap allows, and asserting the exact request count rather than only that some IOException is eventually thrown: without the fix, the loop consumes routes until exhaustion (25 requests); with it, the cap fires at 21. Verified locally by reverting only the interceptor change and confirming the test fails with exactly that 21 vs 25 mismatch, then confirmed the full CallTest suite (217 tests) still passes with the fix applied.

I am in the process of signing the Individual Contributor License Agreement per CONTRIBUTING.md.

### DIFF
--- a/okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/http/RetryAndFollowUpInterceptor.kt
+++ b/okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/http/RetryAndFollowUpInterceptor.kt
@@ -77,6 +77,7 @@ class RetryAndFollowUpInterceptor : Interceptor {
           call.eventListener.retryDecision(call, e, isRecoverable)
           if (!isRecoverable) throw e.withSuppressed(recoveredFailures)
           recoveredFailures += e
+          if (recoveredFailures.size > MAX_RECOVERED_FAILURES) throw e.withSuppressed(recoveredFailures)
           newRoutePlanner = false
           continue
         }
@@ -353,5 +354,13 @@ class RetryAndFollowUpInterceptor : Interceptor {
      * curl, and wget follow 20; Safari follows 16; and HTTP/1.0 recommends 5.
      */
     private const val MAX_FOLLOW_UPS = 20
+
+    /**
+     * How many connection failures should we retry before giving up? Without a cap, a route
+     * selector that keeps yielding usable routes (or keeps cycling through an exhausted set)
+     * lets this loop retry indefinitely, growing [recoveredFailures] without bound and repeatedly
+     * contending for locks a stuck attempt is still holding.
+     */
+    private const val MAX_RECOVERED_FAILURES = 20
   }
 }
--- a/okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/http/RetryAndFollowUpInterceptor.kt
+++ b/okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/http/RetryAndFollowUpInterceptor.kt
@@ -76,8 +76,8 @@ class RetryAndFollowUpInterceptor : Interceptor {
           val isRecoverable = recover(e, call, chain, request)
           call.eventListener.retryDecision(call, e, isRecoverable)
           if (!isRecoverable) throw e.withSuppressed(recoveredFailures)
-          recoveredFailures += e
           if (recoveredFailures.size > MAX_RECOVERED_FAILURES) throw e.withSuppressed(recoveredFailures)
+          recoveredFailures += e
           newRoutePlanner = false
           continue
         }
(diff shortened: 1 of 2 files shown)


okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/http/RetryAndFollowUpInterceptor.kt
```diff
@@ -77,6 +77,7 @@ class RetryAndFollowUpInterceptor : Interceptor {
           call.eventListener.retryDecision(call, e, isRecoverable)
           if (!isRecoverable) throw e.withSuppressed(recoveredFailures)
           recoveredFailures += e
+          if (recoveredFailures.size > MAX_RECOVERED_FAILURES) throw e.withSuppressed(recoveredFailures)
```
swankjesse, line 79: Swap this line with the one above? Otherwise e is in its own suppressed list.

swankjesse (changes requested):

PR #9736, opened 2026-08-29
Delete container tests

yschimke:
Running outside of OkHttp project.

Instead of running per build, running on a schedule in a vibe coded nightmare to quarantine these typically flaky tests

https://yschimke.github.io/okhttp-testbed/


swankjesse (commented):
LGTM

